PR #32, opened 2020-07-13
Create desktop app

ashu8912:


frontend/package.json
```diff
@@ -41,8 +44,28 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
+    "start-electron": "export BROWSER=none && react-scripts start",
+    "build-electron": "npm run build && cp -r electron/. build/electron",
+    "package-electron": "electron-builder build --mac --win --linux -c.extraMetadata.main=build/electron/main.js --publish never",
+    "dev-electron": "export ELECTRON_START_URL=http://localhost:3000 && electron .",
     "lint": "eslint -c .eslintrc.yml --ext .js,.ts,.tsx src/"
   },
+  "build": {
```
joaquimrocha, line 53: This should be called `build-electron`

frontend/electron/main.js
```diff
@@ -0,0 +1,27 @@
+const { app, BrowserWindow } = require('electron');
+const path = require('path');
+const url = require('url');
+let mainWindow;
```
joaquimrocha, line 4: Keep these lines together, but the rest should be separated by a newline.

frontend/.gitignore
```diff
@@ -10,7 +10,7 @@
 
 # production
 /build
-
+/dist
```
joaquimrocha, line 13: Please don't remove the newline. Just add this entry under /build without replacing the newline.

frontend/package.json
joaquimrocha: I'd still like this to be called `build-electron` or `build-desktop`.

frontend/.env
```diff
@@ -0,0 +1,2 @@
+/*Update this with backend services hosted url while building*/
```
joaquimrocha, line 1: Add a space around the text (`/* Update` and `building */`),

joaquimrocha: Also, I think this patch needs to explain why we may need to change this URL.

frontend/package.json
```diff
@@ -18,6 +18,7 @@
     "@types/react-router-dom": "^5.1.5",
     "@types/webpack-env": "^1.15.2",
     "ansi-to-react": "^5.1.0",
+    "is-electron": "^2.2.0",
```
joaquimrocha, line 21: Do we really need another dependency just for this?
The package hasn't been updated for 2 years and is fairly simple:
https://github.com/cheton/is-electron/blob/master/index.js

Maybe we can just pass it a `REACT_APP` var to determine this is an electron app?

frontend/src/App.tsx
```diff
@@ -147,30 +170,13 @@ function App() {
...
-              <ActionsNotifier />
-            </div>
-          </Router>
+          {isElectron() ?
+            <HashRouter>
+              {routerChildren}
+            </HashRouter> :
+            <BrowserRouter>
+              {routerChildren}
+            </BrowserRouter>}
```
joaquimrocha, line 179: You could create a `Router` var/component that depending on the value of the `isElectron()` would return one router or another.
Then the changes are minimal.

joaquimrocha (changes requested):
Left some comments.
Also, I don't know how I missed it last time but this app is just serving the frontend. In a desktop app we want the backend to be running from it too. That's also why I had mentioned on Slack that the electron plumbing should be done in an `app` folder or something, outside of the frontend, and just call the frontend to be built, just like it should for the backend.

There is still some work to be done but it's going in the right direction.

ashu8912:
@joaquimrocha I didn't understand the plumbing part, you said the electron app should be able to run the backend with it, but as far as I understand our headlamp backend services will be hosted somewhere and our electron app will just interact with the backend hosted endpoint. and when running locally I think we need to add few more scripts in the makefile specific to the desktop app.

frontend/package.json
ashu8912: hmm, I don't like the approach of setting the REACT_APP to determine the electron app because than we will have to take care of so many env vars and yes as you mentioned checking whether isElectron is quite simple and I think I should just copy the helper method from https://github.com/cheton/is-electron/blob/master/index.js(coz its also a standard way to check electron environment)

frontend/src/App.tsx
ashu8912: Yeah that would be more elegant thanks!!

frontend/package.json
ashu8912: This is the build specification for the electron app it has to be called build as it is what the electron packager looks for

app/electron/main.js
```diff
@@ -0,0 +1,38 @@
...
+let mainWindow;
+
+const isDev = process.mainModule.filename.indexOf('app.asar') === -1 ||
+    process.mainModule.filename.indexOf('app') === -1;
+
+function createWindow () {
+  let serverFilePath;
+  if(!isDev) {
+     serverFilePath = path.join(process.resourcesPath,'./electron/server');
+     spawn(serverFilePath)
```
joaquimrocha, line 14: Add a `;` at the end of this line. Actually I think the linter needs to be pointed to `app/` now.

app/electron/main.js
```diff
@@ -0,0 +1,38 @@
...
+    };
```
joaquimrocha, line 15: This is not correctly aligned and needs no `;` at the end.

app/electron/main.js
```diff
@@ -0,0 +1,38 @@
...
+    protocol: 'file:',
+    slashes: true,
+  });
+  mainWindow = new BrowserWindow({ width: 800, height: 600 });
+  mainWindow.loadURL(startUrl);
+  mainWindow.on('closed', () => {
+    mainWindow = null;
+  });
+}
+app.on('ready', createWindow);
```
joaquimrocha, line 28: Add a newline on top of this first `app.on(…)`

app/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "headlamp",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "copy-server-binary": "cd ../ && make backend && cp -r ./backend/server ./app/electron/",
+    "start-electron": "npm run copy-server-binary && cd ../ && make run-backend  & cd ../frontend/ && export BROWSER=none && npm start",
+    "build-electron": "npm run copy-server-binary && cd ../frontend/ && npm run build && cp -r ../app/electron/. build/electron",
+    "package-electron": "mkdir -p build && cp -r ../frontend/build/ ./build && electron-builder build --mac --win --linux -c.extraMetadata.main=build/electron/main.js --publish never",
+    "serve-electron-dev": "export ELECTRON_START_URL=http://localhost:3000 && cd ../frontend/ && cp -r ../app/electron/. electron && electron ."
```
joaquimrocha, line 10: Nitpick: I don't think we need to use the `electron` keyword in the commands now that we have this separated, i.e. this whole package.json file refers to the electron app. So unless the command clashes with an existing one (I guess `start` and `build`), don't add electron. i.e. maybe simply `package`, `serve-dev` or just `dev`.

app/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "headlamp",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "copy-server-binary": "cd ../ && make backend && cp -r ./backend/server ./app/electron/",
+    "start-electron": "npm run copy-server-binary && cd ../ && make run-backend  & cd ../frontend/ && export BROWSER=none && npm start",
+    "build-electron": "npm run copy-server-binary && cd ../frontend/ && npm run build && cp -r ../app/electron/. build/electron",
+    "package-electron": "mkdir -p build && cp -r ../frontend/build/ ./build && electron-builder build --mac --win --linux -c.extraMetadata.main=build/electron/main.js --publish never",
```
joaquimrocha, line 9: As mentioned on Slack, for this to work as intended, it should copy the contents of the `frontend/build`, not the whole folder as it will copy it into the existing `build` one, i.e. `cp -r ../frontend/build/* ./build`

app/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "react-redux": "^7.1.3",
+    "react-router-dom": "^5.1.2",
+    "react-scripts": "^3.4.1",
+    "react-scroll": "^1.7.14",
+    "recharts": "^1.8.5",
+    "redux": "^4.0.4",
+    "redux-saga": "^1.1.3",
+    "swagger-parser": "^8.0.3",
+    "typescript": "^3.8.3",
+    "xterm": "^4.3.0"
```
joaquimrocha, line 64: I think we don't need all this (which is problematic as now we need to keep this and the frontend's package.json in sync) if we just build the frontend in its own folder rather than copying the whole project here. i.e. in the `build-electron` command above, we'd just go into frontend, build it, and copy it. Something like `pushd ../frontend/; npm build; popd; mkdir -p build; cp -R ../frontend/build/* ./build/`

frontend/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "headlamp",
   "version": "0.1.0",
   "private": true,
+  "main": "electron/main.js",
```
joaquimrocha, line 5: Hmm, this is changing the `frontend/package.json` . Do we really need this here? I would expect we wouldn't need to change frontend at all.

frontend/src/helpers/index.ts
```diff
@@ -0,0 +1,19 @@
+//helper method to determine whether app is running in electron environment
+export function isElectron():boolean {
```
joaquimrocha, line 2: Add a space after `:`.

frontend/src/helpers/index.ts
```diff
@@ -0,0 +1,19 @@
+//helper method to determine whether app is running in electron environment
```
joaquimrocha, line 1: Please add a `space` after `//` and a notice about where this code was taken from:
`// This snipet was taken form the XYZ project (ABC licensed): https://github.com/…`

frontend/.env
```diff
@@ -0,0 +1,2 @@
+/* Update this with backend services hosted url while building */
```
joaquimrocha, line 1: In the commit message, add a `.` after project: `project. By default it's…`

joaquimrocha (changes requested):
It's almost there now. I left some comments and also I think there should be another commit removing the `Zoom` from the `Window` menu (it's not doing anything), and remove all the entries in `Help`, leaving only `Open an issue` (link to https://github.com/kinvolk/headlamp/issues ) and `About` (link to https://github.com/kinvolk/headlamp ); later we will have to make these links configurable by the .env file, but it's fine having them hardcoded for now.

app/electron/main.js
```diff
@@ -136,6 +137,10 @@ function createWindow () {
 
 app.on('ready', createWindow);
 app.on('window-all-closed', function () {
+  if (!isDev) {
```
joaquimrocha, line 140: Instead of relying on this property again, it's better to check the `serverProcess` itself (and initialize it to `null` when declaring it).

frontend/src/components/account/Auth.tsx
```diff
@@ -21,9 +21,9 @@ interface ReactRouterLocationStateIface {
 
 function Auth() {
   const location = useLocation();
-  const history = useHistory();
   const clusterConf = useClustersConf();
   const {from = { pathname: '/' }} = location.state as ReactRouterLocationStateIface;
+  const history = useHistory();
```
joaquimrocha, line 26: This change doesn't belong here, right?

joaquimrocha (changes requested):
Last changes request I think.

app/electron/main.js
ashu8912: done

frontend/src/components/account/Auth.tsx
ashu8912: done

joaquimrocha (approved):
This is working great now! Thanks!

PR #37, opened 2020-07-16
Add lint fix script

ashu8912:


ashu8912:
hmm, I missed the script in makefile, fair enough closing this.

Issue #34, opened 2020-07-14
Prettify the JSON and YAML views


joaquimrocha:
We should highlight the JSON and YAML data in the several views that show it. E.g. https://www.prettifier.net/json/

**Acceptance Criteria:**
-  [x] The JSON data is highlighted
-  [ ] The YAML data is highlighted

ashu8912:
fixes #38

ashu8912:
@joaquimrocha this can be closed now

joaquimrocha:
We can close it. But we should assign the dark theme everywhere because of that issues with the singleton editor theme.

PR #38, opened 2020-07-16
Pretiffy json and yaml configs

ashu8912:


ashu8912:
fixes #34

joaquimrocha (changes requested):
I have tested this and there are a couple of issues when the content is too wide.
Screenshot from 2020-08-28 18-07-25
Screenshot from 2020-08-28 18-07-16

I was also wondering if we should just use the editor monaco editor (making sure the highlight is on) with the readOnly mode on. So we'd not have to bring more dependencies.
On that note, maybe we also highlight the YAML in the current editor we use.

This is just an idea at this point, since I am sure yet whether it'll look good.

ashu8912:
@joaquimrocha for the highlighting to work in react-monaco-editor it requires to either eject the app and configure the webpack config manually or use react-app-rewired 
https://github.com/react-monaco-editor/react-monaco-editor#usage-with-create-react-app
But I found a better wrapper of monaco for react which prevents us from ejecting or using react-rewired (for highlighting) https://github.com/suren-atoyan/monaco-react

joaquimrocha:
@ashu8912 , the https://github.com/suren-atoyan/monaco-react project looks cool.
I think we should integrate it.

frontend/src/components/common/Resource.tsx
```diff
@@ -331,23 +332,32 @@ export function SectionGrid(props: SectionGridProps) {
 
 export function DataField(props: TextFieldProps) {
   const { label, value, ...other } = props;
+  let language = ((label as string).split('.').pop() as string);
+  if (language !== 'json'){
```
joaquimrocha, line 336: add a space between `)` and `{`.

frontend/src/components/common/Resource.tsx
```diff
@@ -331,23 +332,32 @@ export function SectionGrid(props: SectionGridProps) {
 
 export function DataField(props: TextFieldProps) {
   const { label, value, ...other } = props;
+  let language = ((label as string).split('.').pop() as string);
```
joaquimrocha, line 335: I think the changes to the DataField are not related to changing the monaco editor, so they should be done in a different commit that explains why we're doing this change (the actual prettify purpose).

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -109,16 +110,18 @@ export default function EditorDialog(props: EditorDialogProps) {
     return onSave === null;
   }
 
-  function onChange(newValue: string) {
-    setCode(newValue);
+  function onChange(ev:Monaco.editor.IModelContentChangedEvent,
```
joaquimrocha, line 113: Add a space after `:`

joaquimrocha (changes requested):
The change looks good. Left some small comments. Also, it'd read slightly better if you use `we'd need` in
```
To pretiffy and highlight code in react-monaco-editor we need to eject
the app and than manually configure the webpack.
```

frontend/src/components/common/Resource.tsx
```diff
@@ -331,23 +332,32 @@ export function SectionGrid(props: SectionGridProps) {
...
+        InputLabelProps={{
+          shrink: true,
+          style: {fontSize: '1.3rem'}
+        }}
+        variant="outlined"
+        fullWidth
+        multiline
+        rowsMax="20"
+        {...other}
+      />
```
joaquimrocha, line 352: I don't think we want a `TextField` anymore since we're not using it. What we need is an https://material-ui.com/api/input-label/ with the Editor within.
Screenshot from 2020-09-08 12-44-52

frontend/src/components/common/Resource.tsx
```diff
@@ -331,23 +332,32 @@ export function SectionGrid(props: SectionGridProps) {
...
+      <Editor
+        value={value as string}
+        height="40vh"
+        theme="vs-dark"
```
joaquimrocha, line 356: The Editor shoud also have some changes in order for it to look a bit better. The theme in this case should be white, and its height should be dependent on its contents (let's try to use a 1/3 of the current default height if the lines are < 10). Unless it can be made automatically (adapt to the height of the contents, with a maximum height of what you have set up).

If possible let's try to hide the line numbers and the cursor (the main purpose of this view is to read the contents).
If a light theme for the Editor looks better we could also add it but I am not yet 100% sure on that.

joaquimrocha (changes requested):

joaquimrocha (approved):

PR #39, opened 2020-07-16
Show error when no documentation

### LINKED ISSUE #33: Show error message when documentation is not found
When creating a new resource, if the kind is not set up yet (or is the wrong kind), the documentation tab should show an error message instead of keeping the spinner showing forever.

**Acceptance Criteria:**
- [ ] When the resource has a kind and documentation is found, the documentation is shown
- [ ] If there is an error fetching the documentation, it should be shown in red with a message like: `Cannot load documentation: CAUSE HERE`.

ashu8912:
fixes #33

### DIFF
--- a/frontend/src/components/common/EditorDialog.tsx
+++ b/frontend/src/components/common/EditorDialog.tsx
@@ -293,15 +293,24 @@ function DocsViewer(props: {docSpecs: any}) {
   const { docSpecs } = props;
   const classes = useStyles();
   const [docs, setDocs] = React.useState<object | null>(null);
+  const [docsError, setDocsError] = React.useState<string | null>(null);
 
   React.useEffect(() => {
+    if (docSpecs.error) {
+      setDocsError(`Cannot load documentation: ${docSpecs.error}`);
+      return;
+    }
     if (!docSpecs.apiVersion || !docSpecs.kind) {
+      setDocsError('Cannot load documentation: Please make sure the YAML is valid and has the kind and apiVersion set.');
       return;
     }
 
     getDocDefinitions(docSpecs.apiVersion, docSpecs.kind)
       .then(result => {
         setDocs(result?.properties || {});
+      })
+      .catch((err) => {
+        setDocsError(`Cannot load documentation: ${err}`);
       });
   },
   [docSpecs]);
@@ -333,13 +342,10 @@ function DocsViewer(props: {docSpecs: any}) {
   }
 
   return (
-    docSpecs.error ?
-      <Empty>
-        Error getting documentation!
-        Please make sure the YAML is valid and has the kind and apiVersion set.
-      </Empty>
-      : docs === null ?
-        <Loader />
+    docs === null && docsError === null ?
+      <Loader />
+      : !_.isEmpty(docsError) ?
+        <Empty color="error">{docsError}</Empty>
         : _.isEmpty(docs) ?
           <Empty>No documentation for type {docSpecs.kind.trim()}.</Empty>
           :
--- a/frontend/src/components/common/EmptyContent.tsx
+++ b/frontend/src/components/common/EmptyContent.tsx
@@ -1,13 +1,13 @@
 import Box from '@material-ui/core/Box';
-import Typography from '@material-ui/core/Typography';
+import Typography, { TypographyProps } from '@material-ui/core/Typography';
 import React from 'react';
 
-export default function Empty(props: React.PropsWithChildren<{}>) {
+export default function Empty(props: React.PropsWithChildren<{color: TypographyProps['color']}>) {
   return (
     <Box padding={2}>
       {React.Children.map(props.children, child => {
         if (typeof child === 'string') {
-          return <Typography color="textSecondary" align="center">{child}</Typography>;
+          return <Typography color={props.color} align="center">{child}</Typography>;
         }
         return child;
       })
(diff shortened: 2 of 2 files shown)


frontend/src/components/common/EditorDialog.tsx
```diff
@@ -292,16 +292,28 @@ const useStyles = makeStyles(theme => ({
...
+        `);
       return;
     }
 
     getDocDefinitions(docSpecs.apiVersion, docSpecs.kind)
       .then(result => {
         setDocs(result?.properties || {});
+      })
+      .catch((err) => {
+        setDocsError(`${docsErrorCommonString}: ${err}`);
```
joaquimrocha, line 316: This one will end up with 2 `:` in the message.

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -292,16 +292,28 @@ const useStyles = makeStyles(theme => ({
 function DocsViewer(props: {docSpecs: any}) {
   const { docSpecs } = props;
   const classes = useStyles();
+  const docsErrorCommonString = 'Cannot load documentation:';
```
joaquimrocha, line 295: I understand why you used a var here, but this is a message and will likely get translated later. As a rule one shouldn't do things like concatenation of translatable strings programatically like this patch does, so maybe just repeat the text.

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -333,13 +345,10 @@ function DocsViewer(props: {docSpecs: any}) {
...
-      <Empty>
-        Error getting documentation!
-        Please make sure the YAML is valid and has the kind and apiVersion set.
-      </Empty>
-      : docs === null ?
-        <Loader />
+    docs === null && docsError === null ?
+      <Loader />
+      : !_.isEmpty(docsError) ?
+        <Empty><Typography color="secondary" align="center">{docsError}</Typography></Empty>
```
joaquimrocha, line 351: Hmm, you shouldn't need the `typography/secondary` elem here, right? Should all be in the `<Empty>` component.

ashu8912: @joaquimrocha i added the Typography inside Empty element because right now our Empty Component doesn't have the facility to pass color from outside because its hardcoded there.And in this case I need to pass the red color in case of errors

joaquimrocha: Ah, but you're using the secondary color. I think you shoud use the `error` color.
Maybe a quick patch just passing the color down to the empty would be better than wrapping this component.

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -293,15 +293,26 @@ function DocsViewer(props: {docSpecs: any}) {
   const { docSpecs } = props;
   const classes = useStyles();
   const [docs, setDocs] = React.useState<object | null>(null);
-
```
joaquimrocha, line 296: Please don't remove newlines like this. You should add your new state line to the top of the newline.

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -293,15 +293,26 @@ function DocsViewer(props: {docSpecs: any}) {
...
   React.useEffect(() => {
+    if (docSpecs.error) {
+      setDocsError(`Cannot load documentation: ${docSpecs.error}`);
+      return;
+    }
     if (!docSpecs.apiVersion || !docSpecs.kind) {
+      setDocsError(
+        `Cannot load documentation: 
+        Please make sure the YAML is valid and has the kind and apiVersion set.
+        `);
```
joaquimrocha, line 306: Nitpick: Are you using the backquotes here for preserving the space or was it a mistake? You can use a longer than 100 chars line if it's a string (with single quotes).

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -333,13 +344,10 @@ function DocsViewer(props: {docSpecs: any}) {
...
-    docSpecs.error ?
-      <Empty>
-        Error getting documentation!
-        Please make sure the YAML is valid and has the kind and apiVersion set.
-      </Empty>
-      : docs === null ?
-        <Loader />
+    docs === null && docsError === null ?
+      <Loader />
+      : !_.isEmpty(docsError) ?
```
joaquimrocha, line 349: Nitpick: You can easily just check for the `docsErorr` without using the `isEmpty` (as will only be valid if it's a non-empty string here).
But you don't really have to change this, it's just for future reference.

frontend/src/components/common/EmptyContent.tsx
```diff
@@ -1,17 +1,19 @@
 import Box from '@material-ui/core/Box';
-import Typography from '@material-ui/core/Typography';
+import Typography, { TypographyProps } from '@material-ui/core/Typography';
 import React from 'react';
 
-export default function Empty(props: React.PropsWithChildren<{}>) {
+export default function Empty(props: {color: TypographyProps['color'];children: React.ReactNode}) {
```
joaquimrocha, line 5: Can't you use?
```suggestion
export default function Empty(props: React.PropsWithChildren<{color: TypographyProps['color']}>) {
```

frontend/src/components/common/EditorDialog.tsx
```diff
@@ -347,7 +347,7 @@ function DocsViewer(props: {docSpecs: any}) {
     docs === null && docsError === null ?
       <Loader />
       : !_.isEmpty(docsError) ?
-        <Empty>{docsError}</Empty>
+        <Empty color="error">{docsError}</Empty>
```
joaquimrocha, line 350: BTW, this change shouldn't be part of this commit.

PR #41, opened 2020-07-30
Handle errors for metrics server 

ashu8912:


frontend/src/components/cluster/Charts.tsx
```diff
@@ -7,6 +7,7 @@ import { PercentageCircle, PercentageCircleProps } from '../common/Chart';
 
 interface ResourceCircularChartProps extends Omit<PercentageCircleProps, 'data'> {
   items: KubeObjectInterface[] | null;
+  metricsError: Error | null;
```
joaquimrocha, line 10: This should probably just be called `error` since we don't do anything special about metrics, and maybe we can use the charts for things beyond displaying metrics. i.e. charts are agnostic to metrics, they're just used to display it, so we can call this `error`.

frontend/src/components/node/Details.tsx
```diff
@@ -59,10 +59,11 @@ export default function NodeDetails() {
 interface ChartsSectionProps {
   node: Node | null;
   metrics: KubeMetrics[] | null;
+  metricsError: Error | null;
 }
 
 function ChartsSection(props: ChartsSectionProps) {
-  const { node, metrics } = props;
+  const { node, metrics, metricsError } = props;
```
joaquimrocha, line 66: Same with the name here, should just be `error`.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -399,14 +399,17 @@ export async function apply(body: KubeObjectInterface): Promise<JSON> {
   }
 }
 
-export async function metrics(url: string, onMetrics: (arg: KubeMetrics[]) => void) {
+export async function metrics(
+  url: string, onMetrics: (arg: KubeMetrics[]) => void, onError :(arg:Error) => void
```
joaquimrocha, line 403: Add a space after the two `:`.

joaquimrocha: You should also break it differently I think: leave the url + onMetrics on the previous line, and then add the `onError` arg in the next one, aligned to the opening `(`.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -399,14 +399,17 @@ export async function apply(body: KubeObjectInterface): Promise<JSON> {
...
+){
```
joaquimrocha, line 404: Add a space between `)` and `{`.

frontend/src/lib/k8s/node.ts
```diff
@@ -45,8 +45,8 @@ class Node extends makeKubeObject<KubeNode>('node') {
     return this.jsonData!.spec;
   }
 
-  static useMetrics(onMetrics: (metricsList: KubeMetrics[]) => void) {
-    useConnectApi(metrics.bind(null, '/apis/metrics.k8s.io/v1beta1/nodes', onMetrics));
+  static useMetrics(onMetrics: (metricsList: KubeMetrics[]) => void, onError : (err:Error)=>void) {
```
joaquimrocha, line 48: Add a space after `:`, and around `=>`.

ashu8912:
fixes #41

joaquimrocha:
This is now being handled in #72 .

PR #43, opened 2020-08-14
Build plugin outside frontend tree

ashu8912:


frontend/src/plugin/loadPlugins.js
```diff
@@ -0,0 +1,20 @@
...
+    var head = document.getElementsByTagName('head')[0];
+    (head || document.body).appendChild( script );
+  };
+  (function run(){
+    if (array.length !== 0){
+      loader(array.shift(), run);
+    } else {
+      callback && callback();
+    }
+  })();
```
joaquimrocha, line 18: I find this code a bit difficult to follow. Maybe you can explain this better or point to where you got this if you copied bits.

joaquimrocha (changes requested):
The `loadPlugins` module should be in Typescript, like the rest of the project.

joaquimrocha:
Also for the first commit. Please use uppercase API.
And WRT the message, the backend was not built with the plugins, as the message says; it's simply configured to serve them. So an example message could be:
```
backend: Add API endpoint for listing the available plugins

This patch adds a backend API route to list all the frontend plugins
available, so the frontend can then request and import them.
```

frontend/src/setupProxy.js
```diff
@@ -0,0 +1,10 @@
+const { createProxyMiddleware } = require('http-proxy-middleware');
+module.exports = function(app) {
+  app.use(
+    '/plugins',
+    createProxyMiddleware({
+      target: 'http://localhost:4654',
+      changeOrigin: true,
+    })
+  );
+};
```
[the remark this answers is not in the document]
ashu8912: In case of config we use the apiProxy from the k8s lib, we require the above because we want to proxy "when in dev mode all requests to /plugins to the backend" this is just a development related change, we did similar thing in Nebraska as well for /api and /config

frontend/src/plugin/loadPlugins.js
ashu8912: So we want to fetch all the pluginScripts and the above function job is to be called whenever a previous script in the array is loaded, with the updated array(by deleting the first element using array.shift and now for the loadScript the updated first element is in focus), and we do it until our array has become empty(in case of empty we do our work which we want to do when all script are loaded i.e., in our case initialize plugins).
I think the IIFE part is confusing here It is in no way different from calling the run function manually in the next line.

frontend/src/plugin/loadPlugins.ts
```diff
@@ -0,0 +1,20 @@
...
+    script.src = src;
+    script.onload = (script as any).onreadystatechange = function() {
+      (script as any).onreadystatechange = script.onload = null;
+      handler();
+    };
+    const head = document.getElementsByTagName('head')[0];
+    (head || document.body).appendChild(script);
+  };
+  (function run(){
+    if (array.length !== 0){
```
joaquimrocha, line 13: Add a space between `)` and `{`.

joaquimrocha (changes requested):
Almost there. I have tested and it works fine but a couple of things need to be improved (see comments).

PR #46, opened 2020-09-08
Don't auto refresh tables automatically when new data is passed to it

### LINKED ISSUE #44: Don't refresh tables if they're not in the first page
In views which use tables (every list view), if the data is updated, then the list refreshes to show those changes. This represents a UX a problem if the user is beyond the first page of the table, as it will refresh and show the 1st page again.
Refreshing and keep showing the previous page the user was seeing is also not a good solution as there may be new entries before or in the current page, which would "displace" the results the user may be interested in checking out.

A better solution thus is to just not update the table when the data changes if the table is in a page > 1, but instead show a "Refresh" button on top of the table, so the user can be aware that there are new changes, and click the button.

**Acceptance Criteria:**
- [ ] When the table's data changes, if the user has changed the page at all (i.e. even if changed to page 2 and then page 1), the current table page and data are not changed
- [ ] When the table's data changes, a refresh button is shown on top of the table
- [ ] When the user clicks the refresh button, the actual table's data is changed and thus the user is taken to page 1, with the new data showing (and the refresh button is no longer shown)
- [ ] The refresh button will be centralized horizontally and have this icon, the [label](https://iconify.design/icon-sets/mdi/) `Refresh`, and the tooltip `Click to display the new data in the table`.

joaquimrocha:
fixes #44 

Here's a screenshot of what happens when the data is changed:
Screenshot from 2020-09-08 17-05-08

### DIFF
--- a/frontend/src/components/common/SimpleTable.tsx
+++ b/frontend/src/components/common/SimpleTable.tsx
@@ -1,5 +1,8 @@
+import refreshIcon from '@iconify/icons-mdi/refresh';
 import squareIcon from '@iconify/icons-mdi/square';
-import { InlineIcon } from '@iconify/react';
+import { Icon, InlineIcon } from '@iconify/react';
+import { Button } from '@material-ui/core';
+import Box from '@material-ui/core/Box';
 import { makeStyles } from '@material-ui/core/styles';
 import Table from '@material-ui/core/Table';
 import TableBody from '@material-ui/core/TableBody';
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/SimpleTable.tsx
```diff
@@ -77,30 +81,55 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    filteredData = currentData.filter(filterFunction);
   }
 
   return (
-    (!data || data.length === 0) ?
+    (!currentData || currentData.length === 0) ?
       <Empty>{emptyMessage || 'No data to be shown.'}</Empty>
       :
       <React.Fragment>
+        { // Show a refresh button is the data is not up to date, so we allow the user to keep
```
ashu8912, line 120: A tiny tiny comment fix
```suggestion
        { // Show a refresh button **if** the data is not up to date, so we allow the user to keep
```

ashu8912 (approved):

PR #54, opened 2020-09-15
Improve plugins development

joaquimrocha:
This PR allows to build local plugins + changes the default plugins dir.

### DIFF
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -12,7 +12,7 @@ func main() {
 	insecure := flag.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	// @todo: Make this a uint and validate the values
 	port := flag.String("port", "4654", "Port to listen from")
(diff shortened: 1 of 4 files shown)


frontend/src/redux/reducers/ui.tsx
```diff
@@ -84,7 +84,7 @@ function reducer(state = INITIAL_STATE, action: Action) {
     }
     case UI_DETAILS_VIEW_SET_HEADER_ACTION: {
       const headerActions = {...newFilters.views.details.headerActions};
-      headerActions[action.actionName as string] = action.action;
+      headerActions[action.name as string] = action.action;
```
ashu8912, line 87: This change doesn't seem to belong here

joaquimrocha: Good catch. I have also added that to another PR. This was mistake.

ashu8912 (approved):
Tested and it looks fine to me

ashu8912:
@joaquimrocha we can merge this one

PR #56, opened 2020-09-17
Add oidc support

ashu8912:


backend/cmd/headlamp.go
```diff
@@ -132,6 +136,80 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		resp := struct {
+			OAuth2Token   *oauth2.Token
+			IDTokenClaims *json.RawMessage // ID Token payload is just JSON.
+		}{oauth2Token, new(json.RawMessage)}
+
+		if err := idToken.Claims(&resp.IDTokenClaims); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		if err != nil {
```
knrt10, line 200: I think the error is already checked. No need for this.

knrt10 (commented):
Was just checking. Some little suggestions. Also, I think you guys can add golanglint-ci to the project. Because I see there are many linting errors for go code.

backend/cmd/headlamp.go
ashu8912: nice catch thanks @knrt10

backend/cmd/headlamp.go
```diff
@@ -132,6 +136,77 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+			return
+		}
+
+		var redirectUrl string
+		if config.devMode {
+			redirectUrl = "http://localhost:3000/"
+		} else {
+			redirectUrl = fmt.Sprintf("http://localhost:%1s/", config.port)
+		}
+		redirectUrl += "auth?token=" + oauth2Token.AccessToken
```
joaquimrocha, line 207: This has no information about the cluster for which we're getting the token. So if the user starts authenticating in a cluster, then switches to a different one, and in the popup window completes the authentication, I am guessing it results in the frontend setting up the token for the wrong cluster.

backend/cmd/cluster.go
```diff
@@ -41,12 +67,36 @@ func GetClustersFromKubeConfigFile(kubeConfigPath string) ([]Cluster, error) {
...
+
+		authConfig.ClientID = clientID
+		authConfig.ClientSecret = clientSecret
+		authConfig.IdpIssuerURL = idpIssuerURL
+		authConfig.Type = authType
+	} else {
+		authConfig.Type = "token"
+	}
+	authConfigs[OWN_CLUSTER_NAME] = &authConfig
+
```
joaquimrocha, line 99: I don't think the in-cluster config will have the auth provider set up like this.
Instead, we should access the related config as arguments to the server (`--oidc-client-id`, `--oidc-secret`, ...).

backend/cmd/headlamp.go
```diff
@@ -212,6 +287,19 @@ func GetDefaultKubeConfigPath() string {
 	return filepath.Join(os.Getenv("HOME"), ".kube", "config")
 }
 
+func getClusterAuthType(w http.ResponseWriter, r *http.Request) {
```
joaquimrocha, line 290: Should we instead include the cluster auth types together with the cluster config we already have an endpoint for?

frontend/src/App.tsx
```diff
@@ -117,15 +117,18 @@ function RouteSwitcher() {
...
-            path={getRoutePath(route)}
-            sidebar={route.sidebar}
-            requiresAuth={!route.noAuthRequired}
-            requiresCluster={!route.noCluster}
-            exact={!!route.exact}
-            children={<route.component />}
-          />
+          route.name === 'OidcAuth' ?
+            <Route path={route.path} component={() => <route.component/>} key={index}/>
+            :
```
joaquimrocha, line 122: The route you configured should be used here. I don't understand why you need this distinction.

frontend/src/components/account/Auth.tsx
```diff
@@ -23,7 +23,7 @@ function Auth() {
   const location = useLocation();
   const history = useHistory();
   const clusterConf = useClustersConf();
-  const {from = { pathname: '/' }} = location.state as ReactRouterLocationStateIface;
+  const {from = { pathname: '/login' }} = location.state as ReactRouterLocationStateIface;
```
joaquimrocha, line 26: With this change, once we're authenticated, we're directed to this login page again after logging in.

frontend/src/components/authbuttons/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { generatePath, useHistory } from 'react-router-dom';
+import { request } from '../../lib/k8s/apiProxy';
+import { getCluster, getClusterPrefixedPath } from '../../lib/util';
+import { ClusterDialog } from '../cluster/Chooser';
+import OauthPopup from '../oidcauth/OauthPopup';
+
+function AuthButtons(){
+  const history = useHistory();
+  const [authType, setAuthType] = React.useState<string | null>(null);
+  localStorage.setItem('selected_cluster', getCluster() as string);
```
joaquimrocha, line 12: There should be a setter for this. If there's not please create it, otherwise it'll be difficult to maintain.

frontend/src/components/authbuttons/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  function oidcAuthHandler(){
+    history.replace({
+      pathname: generatePath(getClusterPrefixedPath(), {cluster: localStorage.getItem('selected_cluster') as string}),
+      state: {
+        from: '/login'
+      }
+    });
+  }
+  function oauthWindowCloseHandler(){
+    console.log('closed');
```
joaquimrocha, line 31: I think this was left over from debugging.

frontend/src/components/authbuttons/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+  React.useEffect(() => {
+    request(`/authtype?cluster=${getCluster()}`, {}, false, false)
```
joaquimrocha, line 34: If we had this coming from the cluster config, then maybe we wouldnt need it again.

frontend/src/components/authbuttons/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      useCover
+      disableEscapeKeyDown
+      disableBackdropClick
+    >
+      <Box display="flex" justifyContent="center">
+        {
+          authType === 'oidc' ?
+            <Box m={1}>
+              <OauthPopup onCode={oidcAuthHandler}
+                url={`http://127.0.0.1:4654/oidc?cluster=${getCluster()}`}
```
joaquimrocha, line 54: The port is now `4466`.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,75 @@
+import React, {PureComponent, ReactChild} from 'react';
+
+type Props = {
```
joaquimrocha, line 3: Nitpick: This should be an interface AFAIU.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,75 @@
...
+  width: number;
+  height: number;
+  url: string;
+  title: string;
+  onClose: () => any;
+  onCode: (params: any) => any;
+  children?: ReactChild;
+}
+
+export default class OauthPopup extends PureComponent<Props> {
```
joaquimrocha, line 13: Why did you use a class component instead of a functional one?

frontend/src/components/oidcauth/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { FunctionComponent } from 'react';
+import {useLocation} from 'react-router-dom';
+import { setToken } from '../../lib/auth';
+
+const OIDCAuth:FunctionComponent<{}> = () => {
```
joaquimrocha, line 5: Add a space after the colon.

frontend/src/components/oidcauth/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+  const location = useLocation();
+  const token = new URLSearchParams(location.search).get('token');
+  const cluster = localStorage.getItem('selected_cluster');
+  localStorage.setItem('auth_status', 'success');
+  setToken(cluster as string, token);
```
joaquimrocha, line 10: Surround these by newlines.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const windowFeatures = `toolbar=0,scrollbars=1,status=1,resizable=0,location=1,menuBar=0,width=${width},height=${height},top=${top},left=${left}`;
+
+    this.externalWindow = window.open(
+        url,
+        title,
+        windowFeatures
+    );
+
+    const storageListener = () => {
+      console.log('storage listener trigerred');
```
joaquimrocha, line 46: Remove this and the other logs, as they seem to be debug logs instead.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,75 @@
...
+      try {
+        if (localStorage.getItem('auth_status')) {
+          console.log('trying to close window');
+          onCode(localStorage.getItem('auth_status'));
+          localStorage.removeItem('auth_status');
+          this.externalWindow.close();
+          window.removeEventListener('storage', storageListener);
+        }
+      } catch (e) {
+        console.log('handle error', e);
```
joaquimrocha, line 56: You should lot this but maybe give it a bit more context, otherwise it seems like a debug log.

frontend/src/lib/router.tsx
```diff
@@ -304,13 +306,28 @@ export const ROUTES: {
...
+    component: () => <OIDCAuth/>
+  },
   login: {
     path: '/login',
     exact: true,
     name: 'Login',
     sidebar: null,
     noAuthRequired: true,
-    component: () => <Auth />
+    component: () => <AuthButtons />
```
joaquimrocha, line 330: IMO the `AuthButtons` should be called `AuthChooser`, and the `Auth` should be called, `AuthToken`.
We also need to make sure that we skip the `AuthChooser` if only one auth type is available.

joaquimrocha (changes requested):

backend/cmd/headlamp.go
```diff
@@ -132,6 +136,77 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+			ClientID:     authConfig.ClientID,
+			ClientSecret: authConfig.ClientSecret,
+			Endpoint:     provider.Endpoint(),
+			RedirectURL:  fmt.Sprintf("http://127.0.0.1:%1s/callback", config.port),
+			Scopes:       []string{oidc.ScopeOpenID, "profile", "email", "groups"},
+		}
+		http.Redirect(w, r, oauthConfig.AuthCodeURL(state), http.StatusFound)
+	}).Queries("cluster", "{cluster}")
+
+	r.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ashu8912: hmm, yeah but with the change we will have to register this `oidc-cb` or `oidc-callback` as the auth callback endpoint instead of /callback.
Also yeah the callback can be called with POST method that will be used to post the refresh token and fetch a new token with that

backend/cmd/headlamp.go
ashu8912: hmm, we will have to store on the server info about which cluster has started the auth flow(this can be stored as a global property which will be changed by an api call) and than in the callback we can pass the current auth asking cluster in the query param

backend/cmd/headlamp.go
ashu8912: yes we should do that, it would reduce one extra api call to fetch the auth type

frontend/src/components/authbuttons/index.tsx
ashu8912: ok

frontend/src/components/authbuttons/index.tsx
ashu8912: yes

backend/cmd/kubeconfig.go
```diff
@@ -30,7 +32,28 @@ func GetContextsFromKubeConfigFile(kubeConfigPath string) ([]Context, error) {
...
+		if authProvider != nil {
+			authType := authProvider.Name
+			clientID := authProvider.Config["client-id"]
+			clientSecret := authProvider.Config["client-secret"]
+			idpIssuerURL := authProvider.Config["idp-issuer-url"]
+
+			authConfig.ClientID = clientID
+			authConfig.ClientSecret = clientSecret
+			authConfig.IdpIssuerURL = idpIssuerURL
+			authConfig.Type = authType
```
joaquimrocha, line 48: You can assign each of these with the `authProvider.*` directly, i.e. no need for those temporary vars.

backend/cmd/kubeconfig.go
```diff
@@ -30,7 +32,28 @@ func GetContextsFromKubeConfigFile(kubeConfigPath string) ([]Context, error) {
...
+		} else {
+			authConfig.Type = "token"
```
joaquimrocha, line 50: This is not correct, as you see in the minikube case, where it may have a client certificate.

backend/cmd/cluster.go
```diff
@@ -12,10 +13,28 @@ import (
...
+	config   *clientcmdapi.Cluster
+	AuthType string `json:"auth_type"`
+}
+
+type AuthConfig struct {
+	Type         string `json:"type"`
+	ClientID     string
+	ClientSecret string
+	IdpIssuerURL string
+}
```
joaquimrocha, line 30: I think this needs to be called `OidcConfig`, as it's only useful for that type of authentication. It could even be directly the `oauth2.Config` objects I ask to cache below.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+function AuthChooser(){
+  const history = useHistory();
+  const location = useLocation();
+  const clusters = useClustersConf();
+  const dispatch = useDispatch();
+  const [testingAuth, setTestingAuth] = React.useState(false);
+  const {from = { pathname: createRouteURL('cluster') }} = (location.state || {}) as ReactRouterLocationStateIface;
+  const clusterName = getCluster() as string;
+
+  let clusterAuthType:string = '';
```
joaquimrocha, line 40: Add a space after the `:` (maybe we need to add a rule for this).
Actually, since you're initializing this as a string, you don't need to specify the type here.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  if (clusters[clusterName]){
```
joaquimrocha, line 41: Add a space between `)` and `{`. Ditto for the rest of the patch.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    clusterAuthType = clusters[clusterName].auth_type;
+  }
+
+  function tokenAuthHandler(){
```
joaquimrocha, line 45: Nitpick: This method name seems to result some handler but instead if just goes to a different URL. Consider renaming it as `handleTokenAuth`, or `authWithToken`. Same for the rest of the handlers.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    history.push({
+      pathname: generatePath(getClusterPrefixedPath('token'), {cluster: clusterName as string}),
```
joaquimrocha, line 47: I think you can simply use `getRoutePath('token')` instead of this, right?

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    history.replace({
+      pathname: generatePath(getClusterPrefixedPath(), {cluster: clusterName as string}),
+    });
+  }
+  function oauthWindowCloseHandler(){
+    //gets called when oauth window popup is closed
+  }
+
+  React.useEffect(() => {
+    request(`registerselectedauthcluster?cluster=${clusterName}`, {}, false, false);
```
joaquimrocha, line 61: As mentioned, we should not have this step as it may lead to issues if multiple users and multiple clusters are being used at the same time.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  title="Headlamp Cluster Authentication"
+                >
+                  <ColorButton>Sign In</ColorButton>
+                </OauthPopup>
+              </Box>
+              : null
+          }
+          <Box m={2}>
+            <ColorButton onClick={tokenAuthHandler}>Use A Token</ColorButton>
+          </Box>
```
joaquimrocha, line 150: There should be a back button here, as the mockups show.

frontend/src/components/account/Auth.tsx
```diff
@@ -8,29 +8,24 @@ import Snackbar from '@material-ui/core/Snackbar';
 import TextField from '@material-ui/core/TextField';
 import { Location } from 'history';
 import React from 'react';
-import { useDispatch } from 'react-redux';
```
joaquimrocha, line 11: (Related to the commit message)

Use the CamelCase components' names.
Capitalize OIDC.
Add a space between OIDC and `(`.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onClose: () => {},
+  width: 500,
+  height: 500,
+  url: '',
+  title: ''
+};
+
+const OauthPopup: React.FC<OauthPopupProps> = (props) => {
+  let externalWindow: Window | null;
+  React.useEffect(() => {
```
joaquimrocha, line 23: Add a newline above this.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return () => {
+      if (externalWindow) {
+        externalWindow.close();
+      }
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
joaquimrocha, line 29: I usually add this like:
```
},
[]);
```

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,79 @@
...
+        title,
+        windowFeatures
+    );
+
+    const storageListener = () => {
+      try {
+        if (localStorage.getItem('auth_status')) {
+          onCode(localStorage.getItem('auth_status'));
+          localStorage.removeItem('auth_status');
+          if (externalWindow){
```
joaquimrocha, line 50: Add a space between `)` and `{`.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,79 @@
...
+    externalWindow = window.open(
+        url,
+        title,
+        windowFeatures
+    );
+
+    const storageListener = () => {
+      try {
+        if (localStorage.getItem('auth_status')) {
+          onCode(localStorage.getItem('auth_status'));
```
joaquimrocha, line 48: Use a const var for not repeating the `getItem...`

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -0,0 +1,79 @@
...
+        }
+      } catch (e) {
+        console.log('error occured while closing auth window', e);
+        window.removeEventListener('storage', storageListener);
+      }
+    };
+
+    window.addEventListener('storage', storageListener);
+
+    if (externalWindow){
```
joaquimrocha, line 63: Space needed as mentioned above.

frontend/src/components/oidcauth/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const OIDCAuth: FunctionComponent<{}> = () => {
+  const location = useLocation();
+  const urlSearchParams = new URLSearchParams(location.search);
+  const token = urlSearchParams.get('token');
+  const cluster = urlSearchParams.get('cluster');
+
+  localStorage.setItem('auth_status', 'success');
+  localStorage.setItem(cluster as string, (token as string));
+  return (
```
joaquimrocha, line 14: Add a newline above this.

joaquimrocha (changes requested):
The auth is working (thanks!) but the code needs just changes. See my comments.

backend/cmd/cluster.go
```diff
@@ -13,24 +17,63 @@ import (
...
+		oauthConfig = &oauth2.Config{
+			ClientID:     headlampConfig.oidcClientID,
+			ClientSecret: headlampConfig.oidcClientSecret,
+			Endpoint:     provider.Endpoint(),
+			RedirectURL:  "/oidc-callback",
+			Scopes:       []string{oidc.ScopeOpenID, "profile", "email", "groups"},
+		}
+	}
+
+	oauth2ConfigsCache[OwnClusterName] = &OauthConfig{oauthConfig, verifier, OwnClusterName}
```
joaquimrocha, line 68: What happens when the `oauthConfig` is not correctly initialized? (i.e. if we don't have OIDC enabled)
I think we should only add the cache if we have a valid OauthConfig for it.

backend/cmd/headlamp.go
```diff
@@ -142,6 +148,73 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		http.Redirect(w, r, oauthConfig.Config.AuthCodeURL(state), http.StatusFound)
+	}).Queries("cluster", "{cluster}")
+
+	r.HandleFunc("/oidc-callback", func(w http.ResponseWriter, r *http.Request) {
+		if len(oauthRequestQueue) == 0 {
+			http.Error(w, "no request found for this callback", http.StatusBadRequest)
+			return
+		}
+		var oauthConfig *OauthConfig
+		oauthConfig, oauthRequestQueue = oauthRequestQueue[0], oauthRequestQueue[1:]
```
joaquimrocha, line 173: So I imagine the match by URL is not possible here? (i.e. `r.URL` to match the clusters, as we had talked about)

backend/cmd/server.go
```diff
@@ -13,16 +13,23 @@ func main() {
 	// @todo: Make this a uint and validate the values
 	port := flag.String("port", "4466", "Port to listen from")
 	pluginDir := flag.String("plugins-dir", "./plugins", "Specify the plugins directory to build the backend with")
+	// For inCluster config we need to get the oidc properties from the flags
+	oidcClientID := flag.String("oidc-client-id", "", "ClientID for OIDC")
+	oidcClientSecret := flag.String("oidc-client-secret", "", "ClientSecret for OIDC")
+	oidcIdpIssuerURL := flag.String("oidc-idp-issuer-url", "", "Identity provider issuer URL for oidc")
```
joaquimrocha, line 19: You should validate the use of this with `useInCluster`. i.e. if `useInCluster` is false but we are using the `oidc` options, there should be an error.

frontend/src/components/authchooser/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          cluster.useToken = useToken;
+          dispatch(setConfig({clusters: {...clusters}}));
+          // If we don't require a token, then we just move to the attempted URL or root.
+          if (!useToken) {
+            history.replace(from);
+          }
+
+          setTestingAuth(false);
+        });
+      if (clusterAuthType === 'token' && cluster.useToken){
```
joaquimrocha, line 106: I thought the cluster auth type now would be empty.

joaquimrocha (changes requested):
Some comments there.
Small nitpick: Check as well whether the cluster cache should be that map, or if we should just add pointer in the `Cluster` struct to the oidc object. If you think this makes sense (less things to maintain in case we later do a refresh). I am fine with the map if it's risky otherwise or so.

PR #58, opened 2020-09-17
Add plugins NPM pkg

joaquimrocha:
Now that we support external plugins, the infrastructure to compile
those plugins was still not easy to configure. In order to make that
easier and also to provide other useful modules (like types, libs,
etc.) related to the plugins, this patch creates a new NPM package that
developers can use for building the Headlamp plugins externally.

### DIFF
--- a/plugins-pkg/.gitignore
+++ b/plugins-pkg/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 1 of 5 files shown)


ashu8912 (commented):
LGTM overall, just one comment => we need a gitgnore for the package

PR #64, opened 2020-09-22
backend: Add new rules to golangci and update the code to honor them

joaquimrocha:
We should have similar lint rules across our projects, so this patch
adds the same linters as the Lokomotive channel does, and changes the
code so it passes the lint.


johananl (commented):
The changes look good to me. I have the following concerns though:

- Are we planning to manually sync the linting rules with other projects over time? On the other hand I'm not sure it would make sense to have "central" linting rules as different projects could have different needs.
- Building `golangci-lint` locally results in writes to `go.mod` and `go.sum`. Looks like these changes aren't committed at the moment. We should commit them to avoid a constant diff when running the linter.
- `backend/tools/golangci-lint` isn't gitignored. We should rebase over latest `master` which contains a gitignore file and add this path to the file.

johananl:
I've rebased over `master` and pushed two commits to update `.gitignore` and commit the diff to `go.mod` and `go.sum.` Please feel free to discard these commits in case I've missed something and they are unnecessary.

joaquimrocha (commented):
@johananl , your changes LGTM.
I will wait until we merge other backend PRs so I only have to update this one instead.

PR #66, opened 2020-09-23
docker: Include a plugins dir in the image if it exists

joaquimrocha:


johananl:
Rebased over master.

Dockerfile
```diff
@@ -16,10 +16,15 @@ RUN cd ./backend && go build -o ./server ./cmd/
 
 RUN cd ./frontend && npm install && npm run build
 
+# Create a plugins folder if none exists so we copy whatever exists to the
+# image.
+RUN mkdir -p ./plugins
```
johananl, line 21: Looks like this is necessary only when there is no local plugins directory when building the image.
I think a better solution is to actually commit an empty `./plugins` dir to the repo with a `.gitkeep` file in it and possibly a README which explains its purpose. Since this seems to be a "special" directory which (I assume) can be used for including optional plugins in the image, it makes sense IMO to explicitly have it in the repo. We then don't need this line in the Dockerfile.

johananl (approved):

johananl (commented):
Sorry, meant to comment rather than approve.

Dockerfile
joaquimrocha: It's a good point. I will test this and change the approach.

joaquimrocha: Actually, on a second thought. We do have a `-pkugin-dir` option in the backend, so I am not sure if we should include a "default" empty directory just for this.
Maybe the solution is to pass the dir as an arg to the Dockerfile.

joaquimrocha:
I decided to merge this as it's actually a pain to conditionally include dirs/files in the Dockerfile (and I don't like having an empty dir just for creating a docker image). I will open an issue to work out something else though.

PR #67, opened 2020-09-25
Makefile: Specify Docker build path and ignore node_modules

johananl:
This PR makes `make image` work. I'm not sure if the produced image works but at least we get an image.
It also adds `node_modules` to `.dockerignore` since this directory is huge and there is no need to send it to the Docker daemon when building images.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+frontend/node_modules
--- a/Makefile
+++ b/Makefile
@@ -43,4 +43,5 @@ image:
 	--no-cache \
 	--build-arg IMAGE_BASE=$(DOCKER_IMAGE_BASE) \
 	-t $(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
-	Dockerfile
+	Dockerfile \
+	.


joaquimrocha (changes requested):
I think Ashu had pointed it out but I am not sure what happened after that.
I use podman and it doesn't complain, maybe that;s why I never noticed it.

Anyway, good catch. I'd like only to ask for a slightly less generic title. Something along the lines of:
`Makefile: Add missing dir to image target's docker command`

(With generic titles things take longer to figure out later when we're scrolling through lists of similar messages)

ashu8912:
fixes #31

johananl:
> I think Ashu had pointed it out but I am not sure what happened after that.
> I use podman and it doesn't complain, maybe that;s why I never noticed it.
> 
> Anyway, good catch. I'd like only to ask for a slightly less generic title. Something along the lines of:
> `Makefile: Add missing dir to image target's docker command`
> 
> (With generic titles things take longer to figure out later when we're scrolling through lists of similar messages)

I tried to encompass both of the commits without being too elaborate. Hope it's better now.

joaquimrocha:
> > I think Ashu had pointed it out but I am not sure what happened after that.
> > I use podman and it doesn't complain, maybe that;s why I never noticed it.
> > Anyway, good catch. I'd like only to ask for a slightly less generic title. Something along the lines of:
> > `Makefile: Add missing dir to image target's docker command`
> > (With generic titles things take longer to figure out later when we're scrolling through lists of similar messages)
> 
> I tried to encompass both of the commits without being too elaborate. Hope it's better now.

I actually meant the commit's title, not the PR one. Sorry if it was not clear.

johananl:
@joaquimrocha Fixed.

PR #68, opened 2020-09-25
Use context auth

### LINKED ISSUE #4: Authentication with a kubeconfig (without token)
When starting a Lokomotive cluster without gangway, I get a kubeconfig with a client certificate but without token. We should implement a way in the dashboard to authenticate with a kubeconfig, without a dashboard.

joaquimrocha:
Allow to use client certificates for authenticating + make token requirement dependent on whether the auth fails.

fixes #4


ashu8912 (approved):
These are some nice set of changes. Tested and the changes LGTM

PR #72, opened 2020-09-29
Permissions check & optional metrics-server

joaquimrocha:
We need to check whether the user has the permissions to list objects in the cluster, otherwise most of the views the user has direct access to from the sidebar will be confusing.

As part of this PR we also make the metrics-server optional.

### DIFF
--- a/frontend/src/lib/util.ts
+++ b/frontend/src/lib/util.ts
@@ -1,9 +1,11 @@
 import TimeAgo from 'javascript-time-ago';
 import en from 'javascript-time-ago/locale/en';
+import React from 'react';
 import { matchPath } from 'react-router';
 import { isElectron } from '../helpers';
 import { FilterState } from '../redux/reducers/filter';
 import { useTypedSelector } from '../redux/reducers/reducers';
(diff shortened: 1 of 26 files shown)


frontend/src/lib/util.ts
```diff
@@ -107,3 +109,17 @@ export function getCluster(): string | null {
...
+  const [error, setError] = React.useState<ApiError | null>(null);
+
+  React.useEffect(() => {
+    if (!!error && !!dependentSetter) {
+      dependentSetter(null);
+    }
+  },
+  [error, dependentSetter]);
+
+  // Adding "as any" here because it was getting difficult to validate the setter type.
```
ashu8912, line 123: Why do you need to set the any type on setError isn't it implicit to the useErrorState function?

ashu8912 (approved):
Tested!! LGTM

frontend/src/lib/util.ts
joaquimrocha: I could validate it by setting the right return type in the function, but that's opening more and more places that need the return types set. Let's get this in now and change it later if a better solution is found.

PR #74, opened 2020-09-29
Fix exceptions in cluster actions

### LINKED ISSUE #73: Handle exceptions when performing cluster actions
When performing cluster actions such as edit/delete/create, apparently the exceptions aren't being handled.
I have fixed this locally but I'm filing this issue to track it as I cannot finish it now.

joaquimrocha:
fixes #73

### DIFF
--- a/frontend/src/components/common/EditButton.tsx
+++ b/frontend/src/components/common/EditButton.tsx
@@ -37,8 +37,6 @@ export default function EditButton(props: EditButtonProps) {
         cancelledMessage: `Cancelled changes to ${item.metadata.name}.`,
         successMessage: `Applied changes to ${item.metadata.name}.`,
         errorMessage: `Failed to apply changes to ${item.metadata.name}.`,
-        successOptions: {variant: 'success'},
-        errorOptions: {variant: 'error'},
         cancelUrl,
         ...options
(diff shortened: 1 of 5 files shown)


frontend/src/components/common/EditButton.tsx
```diff
@@ -20,12 +20,30 @@ export default function EditButton(props: EditButtonProps) {
...
+    }
+  }
+
+  async function updateFunc(newItem: KubeServiceAccount) {
+    try {
+      await item.update(newItem);
+    } catch (err) {
+      setErrorMessage(makeErrorMessage(err));
+      setOpenDialog(true);
+      throw err;
```
ashu8912, line 42: we are already in the catch block why do you want to throw err from here? 
it will break the app if i am correct

joaquimrocha: We catch it here so we show the dialog, etc. then we throw it again so it gets caught and treated in the saga (so it displays a notification). This is usually called error propagation.

ashu8912: cool if its caught in the saga than its fine

ashu8912 (approved):

PR #78, opened 2020-10-06
Add route info for clusterroles and clusterrolebindings

ashu8912:


joaquimrocha:
I would swear I had fixed this before!
Maybe I forgot to include it in a PR...

I had fixed it by overriding:
```
get listRoute() {
    return 'roleBindings';
}
```

But I think your solution is also good and may imply less changes in the future. So let's use it.

PR #79, opened 2020-10-06
 Fix loader for role and roleBinding list view

ashu8912:


joaquimrocha:
I think this is already fixed by some of the changes merged this week. Can you verify?

ashu8912:
True this fix already in master

PR #83, opened 2020-10-09
useCluster + Logout button

joaquimrocha:
Add a useCluster hook and a logout button.
These functionalities weren't supposed to go in the same PR but the hook was needed also for the log out functionality so here it goes.


ashu8912:
@joaquimrocha there are conflicts that need to be resolved here

ashu8912 (approved):

PR #84, opened 2020-10-09
frontend: Set right color for workload charts' foreground

joaquimrocha:
The color should be black and not blue.


ashu8912 (approved):
LGTM

PR #92, opened 2020-10-16
frontend: Don't show a cluster name if there's only one cluster

joaquimrocha:
If there is only one cluster configured, then we should show the cluster
name in the UI as the user context should be enough, and when deployed
in-cluster it was just showing "Cluster: main".


ashu8912 (approved):
Tested LGTM!! Thank you :D

PR #93, opened 2020-10-16
backend: Use https scheme for all OIDC redirect URL

joaquimrocha:
The URL scheme used for the OIDC redirect URL was dependent on having
the TLS presence in the /oidc request, but TLS may have been delegated
to the k8s deployment. So this patch enforces https unless the host
being used is the localhost one. This is not a great solution but we
can use it for now and think about a more complete one later.


ashu8912:
@joaquimrocha the linter is failing

PR #95, opened 2020-10-19
Fix infinite update roles

ashu8912:


frontend/src/components/role/BindingList.tsx
```diff
@@ -27,12 +27,12 @@ export default function RoleBindingList() {
...
-  function setupRoleBindings(newBindgins: RoleBinding[] | null) {
-    setRoleBindings(newBindgins, 'RoleBinding');
+  function setupRoleBindings(newBindings: RoleBinding[] | null) {
+    setRoleBindings(newBindings, 'RoleBinding');
   }
 
-  function setupClusterRoleBindings(newBindgins: RoleBinding[] | null) {
-    setRoleBindings(newBindgins, 'ClusterRoleBinding');
+  function setupClusterRoleBindings(newBindings: RoleBinding[] | null) {
+    setRoleBindings(newBindings, 'ClusterRoleBinding');
```
joaquimrocha, line 38: :joy: Thanks!

frontend/src/components/role/BindingList.tsx
```diff
@@ -19,6 +19,9 @@ export default function RoleBindingList() {
   const filterFunc = useFilterFunc();
 
   function setRoleBindings(newBindings: RoleBinding[] | null, kind: string) {
+    if (!newBindings) {
+      return;
+    }
```
joaquimrocha, line 24: I have a bit of a doubt that this is the root of the problem but I'd need to verify my claim. Is this triggering the infinite loop, or do we continuously get errors and the infinite look is just more evident because this function call gets triggered?
If we keep getting this triggered because the other setters are being called repeatedly by the `useErrorState`, then I'd check that hook and see if the problem is that we have the passed callback in the dependencies of its effect. We can discuss better synchronously tomorrow.

ashu8912: If you look below in this function we are doing a setState(see setBindings(data)) and that is what is trigerring this infinite loop because the component gets re rendered again and this makes the useErrorState gets called and than in the callback of useErrorState is this setupRoleBindings which calls to setBindings function which gets trigerred again and so on,Now we want to control this if our dependent setter is called with null that means there was an error and we don't want to set Bindings again

joaquimrocha: Right. So what I understand here is that `useErrorState` should be triggered on every rewrite, right? Because it's a hook (has a side-effect and states in itself). So the problem here is that the error keeps getting triggered. My suspicion is that it's because of having the setter in the effect's dependencies like that.

joaquimrocha (approved):

PR #96, opened 2020-10-19
WIP: Add page and rowsPerPage to url

ashu8912:


frontend/src/components/common/SimpleTable.tsx
```diff
@@ -72,37 +73,68 @@ export default function SimpleTable(props: SimpleTableProps) {
...
   } = props;
+  const location = useLocation();
+  const history = useHistory();
   const [page, setPage] = React.useState(0);
   const [currentData, setCurrentData] = React.useState(data);
   const rowsPerPageOptions = props.rowsPerPage || [5, 10, 50];
   const [rowsPerPage, setRowsPerPage] = React.useState(rowsPerPageOptions[0]);
   const classes = useTableStyle();
 
+  function addQuery(queryObj:{[key:string]: string}) {
```
joaquimrocha, line 84: There should be a space after which `colon`.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -72,37 +73,68 @@ export default function SimpleTable(props: SimpleTableProps) {
...
-
     // If the currentData is not up to date and we are in the first page, then update
     // it directly. Otherwise it will require user's intervention.
     if ((!currentData || currentData.length === 0) || page === 0) {
       setCurrentData(data);
     }
   },
   // eslint-disable-next-line
-  [data, currentData]);
+  [data])
```
joaquimrocha, line 124: Why did should this be removed? We need to take the value of currentdata, otherwise it will always be the first state's version.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -72,37 +73,68 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    setCurrentData(data);
+  }
+
+  React.useEffect(() => {
+    const params = new URLSearchParams(location.search);
+    setPage(parseInt(params.get('page') as string) || 0);
+    setRowsPerPage(parseInt(params.get('rowsPerPage') as string) || rowsPerPageOptions[0]);
+  },
+  // eslint-disable-next-line
+  [location])
```
joaquimrocha, line 137: I am very surprised that this doesn't result in an infinite loop (maybe because we're setting the history and not the location?).

Seems like the same case we had a while ago with Nebraska where we were not taking the the URL as the single source of truth but in fact setting the state in both places (URL + component states like `page`/`rowsPerPage`).
However, probably we need to keep the statuses like you're doing because I'd like this feature to be optional (as if we add two tables to the same view, it'll be messy!).

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -72,37 +73,68 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+
+  function refreshClickHandler(data : SimpleTableProps['data']){
+    addQuery({'page': '0', 'rowsPerPage': rowsPerPage.toString()});
+    setCurrentData(data);
+  }
+
+  React.useEffect(() => {
+    const params = new URLSearchParams(location.search);
+    setPage(parseInt(params.get('page') as string) || 0);
+    setRowsPerPage(parseInt(params.get('rowsPerPage') as string) || rowsPerPageOptions[0]);
```
joaquimrocha, line 134: We should verify this against the data we have, because if we type page=2 rowsPerPage=500 or so, it will still set that in the table (here's an example with 50 on page 2):
Screenshot from 2020-10-20 16-17-49

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -72,37 +73,68 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    const pathname = location.pathname;
+    const searchParams = new URLSearchParams(location.search);
+    for (const key in queryObj) {
+      const value = queryObj[key];
+      if (value) {
+        searchParams.set(key, value);
+      } else {
+        searchParams.delete(key);
+      }
+    }
```
joaquimrocha, line 94: I don't love noisy URLs. So I'd prefer if we only set these params if they differ from whatever we set as default.

joaquimrocha (changes requested):
Left some comments.
Also, we should add the `namespace` filter (set from the filter header) from the URL as well, since it's an important filter to use.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -125,6 +125,12 @@ export default function SimpleTable(props: SimpleTableProps) {
     const tableNameUrlParam = getFilterValueByNameFromURL('tableName', history) || null;
     if (!tableNameUrlParam || tableNameUrlParam === tableName) {
       let rowsPerPageVal: number = parseInt(getFilterValueByNameFromURL('rowsPerPage', location) as string) || rowsPerPageOptions[0];
+      if (data && rowsPerPageVal && data.length < rowsPerPageVal) {
+        // get me the nearest rowsPerPageVal from the rowsPerPageOptions
+        rowsPerPageVal = rowsPerPageOptions.sort((a, b) => {
+          return Math.abs(a - rowsPerPageVal) - Math.abs(b - rowsPerPageVal);
+        })[0];
+      }
```
joaquimrocha, line 133: This check is missing the page value, as if we use `rowsPerPage=2&page=30000` with a data length of 3, it's the page that will create problems.
So the check should be`if (data && data.length < (rowsPerPageVal * pageVal))`, and the action in case this is true should just be:
1) Use the default `rowsPerPage` if the value provided is not in `rowsPerPageOptions`;
2) And show the last page of the table (i.e. floor of `data.length / rowsPerPage`).

joaquimrocha (changes requested):
Almost where we want it. Left some comments.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -71,38 +74,58 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+  }
+
+  React.useEffect(() => {
+    const tableNameUrlParam = getFilterValueByNameFromURL('tableName', history) || null;
+    if (!tableNameUrlParam || tableNameUrlParam[0] === tableName) {
+      const rowsPerPageUrlParam = getFilterValueByNameFromURL('rowsPerPage', location);
+      const pageUrlParam = getFilterValueByNameFromURL('page', location);
+      rowsPerPageUrlParam ? setRowsPerPage(parseInt(rowsPerPageUrlParam[0])) :
+        setRowsPerPage(rowsPerPageOptions[0]);
+      pageUrlParam ? setPage(parseInt(pageUrlParam[0])) : 0;
```
joaquimrocha: Seems like the ternary operation should be inside `setPage`? (otherwise the `0` is not used)
Maybe it should also only set the page/rows if the values change (otherwise changing the row also changes the page, and vice-versa, getting us extra refreshes that don't need to be done).

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -71,38 +74,58 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    setCurrentData(data);
+  }
+
+  React.useEffect(() => {
+    const tableNameUrlParam = getFilterValueByNameFromURL('tableName', history) || null;
+    if (!tableNameUrlParam || tableNameUrlParam[0] === tableName) {
+      const rowsPerPageUrlParam = getFilterValueByNameFromURL('rowsPerPage', location);
+      const pageUrlParam = getFilterValueByNameFromURL('page', location);
+      rowsPerPageUrlParam ? setRowsPerPage(parseInt(rowsPerPageUrlParam[0])) :
+        setRowsPerPage(rowsPerPageOptions[0]);
```
joaquimrocha: Use the ternary op inside `setRowsPerPage`.

frontend/src/helpers/index.ts
```diff
@@ -20,3 +20,67 @@ export function isElectron(): boolean {
...
+  const searchParams = new URLSearchParams(location.search);
+  const filtersInURLString = searchParams.get('filters');
+  if (!!filtersInURLString) {
+    const filterObjStringArray = filtersInURLString.split(';');
+
+    for (const filterStr of filterObjStringArray) {
+      const filterPropertyArr = filterStr.split('=');
+      if (filterPropertyArr[0] === key) {
+        const filterValueStringArr = filterPropertyArr[1].split(',');
+        return filterValueStringArr;
```
joaquimrocha, line 34: Just return the result of the split, no need for this const var.

frontend/src/helpers/index.ts
```diff
@@ -20,3 +20,67 @@ export function isElectron(): boolean {
 
   return false;
 }
+
+export function getFilterValueByNameFromURL(key: string, location: any): string[] {
+  const searchParams = new URLSearchParams(location.search);
+  const filtersInURLString = searchParams.get('filters');
+  if (!!filtersInURLString) {
```
joaquimrocha, line 27: Invert the logic with an early `[]` return to make the code easier to read.

frontend/src/components/common/Autocomplete.tsx
```diff
@@ -9,6 +9,8 @@ import Typography from '@material-ui/core/Typography';
 import Autocomplete from '@material-ui/lab/Autocomplete';
 import React from 'react';
 import { useDispatch } from 'react-redux';
+import { useHistory, useLocation } from 'react-router-dom';
```
joaquimrocha, line 12: In the commit message, use URL in all caps.

frontend/src/components/common/Autocomplete.tsx
```diff
@@ -20,6 +22,9 @@ export function NamespacesAutocomplete() {
   const theme = useTheme();
   const filter = useTypedSelector(state => state.filter);
   const [namespaces, setNamespaces] = React.useState<Namespace[]>([]);
+  const queryParamDefaultObj = {'namespace': '', 'page': '0'};
```
joaquimrocha, line 25: Do we need a page here? (and below)

joaquimrocha (changes requested):
Left some comments still but getting very close. Thanks..

frontend/src/helpers/index.ts
ashu8912: So @joaquimrocha  here we want this to be like this only cause the return [] also acts as a default return when the filters are there in url but the property that you are asking for was not present

frontend/src/components/common/Autocomplete.tsx
ashu8912: aah so this shouldn't be here but one thing in SimpleTable Component we must look for changes in the url namespace filter and whenever it changes we must reset the page to 0 because the namespace filtering will triger but if don't reset the page to 0 and if we our page was previously pointing to no 2 but because of the filter we don't have that much data we will show no data because of being on the wrong page

ashu8912: @joaquimrocha this won't be required anymore created and used checkOutOfBoundUrlParam function which will handle this

frontend/src/helpers/index.ts
joaquimrocha: Maybe I couldn't express what I meant:
You have a big block in `if (!!filtersInURLString) { …`. You could invert the logic and early return:
```javascript
  if (!filterInURLString) {
    return [];
  }

  const …
```

Anyway it's a nitpick.

joaquimrocha (changes requested):
Trying it out now that the code LGTM I noticed however that the URL encoding is not as I expected. Sorry I couldn't notice this from just reading the code. I am getting:
`http://…/c/acela/pods?filters=namespace%3Ddefault%2Cdex%2Cexternal-dns%3B%3B`
but I meant that the namespace should be a global filter in the sense that it's safe for us to make it it's own key in the query, and we can use commas for the separation, unless I am missing some issue with the URL encoding:
`http://.../c/acela/pods?namespace=external-dns,default,dex,gangway`

joaquimrocha:
> Trying it out now that the code LGTM I noticed however that the URL encoding is not as I expected. Sorry I couldn't notice this from just reading the code. I am getting:
> `http://…/c/acela/pods?filters=namespace%3Ddefault%2Cdex%2Cexternal-dns%3B%3B`
> but I meant that the namespace should be a global filter in the sense that it's safe for us to make it it's own key in the query, and we can use commas for the separation, unless I am missing some issue with the URL encoding:
> `http://.../c/acela/pods?namespace=external-dns,default,dex,gangway`

@ashu8912 I have tested the PR again and my second concern (that the namespace filter is not applied globally) still stands.

About the encoded chars, maybe we could make it work with `+` instead. This is what GMail is doing for words AFAICT (and it looks better than the encoded semi-colons).

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -21,10 +23,13 @@ interface SectionFilterHeaderProps extends SectionHeaderProps {
 
 export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
   const { noNamespaceFilter = false, noSearch = false, ...headerProps } = props;
+  const location = useLocation();
+  const namespaceInURL = getFilterValueByNameFromURL('namespace', location, true);
```
joaquimrocha, line 27: Maybe we should just use the value of the namespace filter in Redux here, as it should be the one source of truth.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -21,10 +23,13 @@ interface SectionFilterHeaderProps extends SectionHeaderProps {
...
   const filter = useTypedSelector(state => state.filter);
   const dispatch = useDispatch();
 
-  const hasNamespaceFilters = !noNamespaceFilter && filter.namespaces.size > 0;
+  const hasNamespaceFilters = namespaceInURL.length !== 0 ||
+  (!noNamespaceFilter && filter.namespaces.size > 0);
```
joaquimrocha, line 32: This logic will activate the filter header even when there's no namespace filter to be shown (some of the views don't have a namespace to be used).

frontend/src/components/common/Autocomplete.tsx
```diff
@@ -36,14 +41,25 @@ export function NamespacesAutocomplete() {
...
+  React.useEffect(() => {
+    const selectedNamespace = getFilterValueByNameFromURL('namespace', location, true);
+    if (!selectedNamespace) {
+      return;
+    }
+    dispatch(setNamespaceFilter(selectedNamespace));
+  },
+  // eslint-disable-next-line
+  []);
+
```
joaquimrocha, line 53: I am not sure this is the best place for this logic, since this means that if we access the cluster's root URL with namespaces, example: http://localhost:3000/c/acela/?namespace=dex,gangway
and then we move to the Workloads, the namespaces filtering is not applied (because the link in the sidebar won't have it).
However, if we choose the namespaces (e.g. from workloads) and go to the cluster view (which, BTW, will show an empty list of events as it's using the namespaces for filtering) and then click the Workloads view again, it will apply the filters.

joaquimrocha (changes requested):
Left a few comments related to some issues I found out about the namespace logic.
The rowsPerPage and page number are not working for me as well. Let's keep this out for now (you can keep a WIP branch) and just focus on the namespace in this PR instead.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -41,6 +45,11 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
   });
 
   React.useEffect(() => {
+    const namespace = getFilterValueByNameFromURL('namespace', location);
+    if (namespace.length > 0) {
+      dispatch(setNamespaceFilter(namespace));
+      setShowFilters(true);
```
joaquimrocha, line 51: This will show the search filter as well, on pages that may not even have the namespace filter. So you need to check for whether the `noNamespaceFilter` is set.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -41,6 +45,11 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
   });
 
   React.useEffect(() => {
+    const namespace = getFilterValueByNameFromURL('namespace', location);
+    if (namespace.length > 0) {
+      dispatch(setNamespaceFilter(namespace));
```
joaquimrocha, line 50: Before we dispatch, we should check whether the current namespace is already the same value as the URL, to avoid reacting to this change by other components.

The fundamental idea is that redux is the one source of truth, the URL is simply the entry point when loading the page and for copying the URL (and from that moment, should be kept up to date with the redux value).

frontend/src/components/common/Autocomplete.tsx
```diff
@@ -26,6 +26,11 @@ export function NamespacesAutocomplete() {
   const location = useLocation();
   const history = useHistory();
 
+  React.useEffect(() => {
+    getFilterValueByNameFromURL('namespace', location);
+  },
+  []);
+
```
joaquimrocha: Should we just take the value from Redux?

frontend/src/components/cluster/Overview.tsx
```diff
@@ -75,7 +75,7 @@ const useStyles = makeStyles(theme => ({
 
 function EventsSection() {
   const classes = useStyles();
-  const filterFunc = useFilterFunc();
+  const filterFunc = useFilterFunc(false);
```
joaquimrocha, line 78: I understand why you added this patch, but I release now that we have a fundamental mistake in the events' page: it is missing a namespace column. I will open an issue.

joaquimrocha (changes requested):
Almost there. Now the page settings are working, but a couple of things in the code need to be refined.

joaquimrocha:
@ashu8912 , I have tested the latest version and the page and rowsPerPage is still having issues. In the Workloads view, if I go to page 3 in the acela dev cluster and just copy & paste the URL, it will not show me any results.
If the page is also given as a value greater than what we support, then it also is not set to the last or the first page.

Screenshot from 2021-02-18 09-57-50
Screenshot from 2021-02-18 10-12-22

joaquimrocha:
Let's close this one but keep the branch, since I think we will not get to it in the near future.

PR #99, opened 2020-10-21
Add Sort to table columns

ashu8912:


frontend/src/components/common/SimpleTable.tsx
```diff
@@ -42,6 +44,7 @@ interface SimpleTableColumn {
   cellProps?: {
     [propName: string]: any;
   };
+  sortFunction?: (isIncreasingOrder: boolean) => (arg1: any, arg2: any) => number;
```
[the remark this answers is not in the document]
joaquimrocha: Another thing we could do to make this even more automatic is to provide a default sorting function and try to apply it to at least strings/numbers. E.g. we could use `sort` instead of `sortFunc` and it'd be either the `sortFunc` type you had, and `| boolean`.
If the `sort` prop is `true`, then we just use the default one (this should only be set for columns we are using "basic" values in, that we know our default sort function can consule), so either the `datum` or the result of the `getter` should be a known/sortable one. This requires a bit of work but maybe saves a lot of code/maintenance?

frontend/src/components/common/Resource.tsx
```diff
@@ -474,10 +484,20 @@ export function ConditionsTable(props: ConditionsTableProps) {
...
+    const columnInitiatingSort = getColumns().filter((col) => {
+      return col.label === sortObj.label;
+    })[0].sortFunction;
+    return columnInitiatingSort && columnInitiatingSort(sortObj.isIncreasingOrder);
+  }
+
   return (
     <SimpleTable
-      data={(resource && resource.status && resource.status.conditions) || {}}
+      data={(resource && resource.status && resource.status.conditions.sort(getSortFunc())) || {}}
```
joaquimrocha, line 498: You don't need to sort it here in the condution, because sorting shouldn't change the data, just the order.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -62,6 +65,7 @@ export interface SimpleTableProps {
   rowsPerPage?: number[];
   emptyMessage?: string;
   errorMessage?: string | null;
+  sortObjSetter? : Dispatch<SetStateAction<{ label: string; isIncreasingOrder: boolean }>>;
```
joaquimrocha, line 68: Let's discuss this one on Slack because I don't understand exactly why it's needed, and it seems to add a lot of complexity to the implementation.

joaquimrocha (changes requested):

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -104,6 +112,34 @@ export default function SimpleTable(props: SimpleTableProps) {
   // eslint-disable-next-line
   [data, currentData]);
 
+  React.useEffect(() => {
+    if (data) {
+      let applySort = undefined;
+      const columnAskingForSort = columns.find((col) => col.label === sortObj.label);
```
joaquimrocha, line 118: Don't use the label as it may not be unique and thus not a good way for distinguishing between columns. I think it's better to use the column index, this also makes getting the column more efficient.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -71,12 +77,14 @@ export default function SimpleTable(props: SimpleTableProps) {
...
     emptyMessage = null,
     errorMessage = null,
+    isDataAlreadySortedInOneOfTheOrder
   } = props;
   const [page, setPage] = React.useState(0);
   const [currentData, setCurrentData] = React.useState(data);
   const rowsPerPageOptions = props.rowsPerPage || [5, 10, 50];
   const [rowsPerPage, setRowsPerPage] = React.useState(rowsPerPageOptions[0]);
   const classes = useTableStyle();
+  const [sortObj, setSortObj] = React.useState({'label': columns[0].label, 'isIncreasingOrder': true});
```
joaquimrocha, line 87: The `isIncreasingOrder` should be its own state AFAIU.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -104,6 +112,34 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    if (data) {
+      let applySort = undefined;
+      const columnAskingForSort = columns.find((col) => col.label === sortObj.label);
+      const sortFunction = columnAskingForSort?.sort;
+
+      if (typeof sortFunction === 'boolean' && sortFunction){
+        if (sortObj.isIncreasingOrder){
+          setCurrentData(data.sort());
+        } else {
+          setCurrentData(data.sort().reverse());
```
joaquimrocha, line 125: This function is very likely to be inneficient because at least in Headlamp will mostly be comparing classes.
So we should provide the default sorting function that will use the column's `getter`/`datum` and compare the results, i.e. if the getter is `getter: (item: Event) => item.name`, then we need to do something like (but with types :smile_cat: ):
```
function defaultSortingFunction(column) {
  function defaultSortingReal(item1, item2) {
    let value1 = column.getter(item1);
    let value2 = column.getter(item2);

    let compareValue = 0;
    if (value1 < value2) {
      compareValue = -1;
    } else if (value1 > value2) {
      compareValue = 1;
    }

    return compareValue * (isAscendingOrder ? 1 : -1);
  }

  return defaultSortingReal;
}

...

data.sort(defaultSortingFunction(col[colIndex]));

File: frontend/src/components/common/SimpleTable.tsx [hunk 7]
```diff
@@ -155,6 +191,33 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+                        <>
+                          <IconButton onClick={() => setSortObj({label, isIncreasingOrder: true})}>
+                            <Icon icon={menuUp}/>
+                          </IconButton>
+                          <IconButton onClick={() => setSortObj({label, isIncreasingOrder: false})}>
+                            <Icon icon={menuDown}/>
+                          </IconButton>
+                        </>
+                      }
+                    </>}
```
joaquimrocha, line 220: Maybe we should abstract this into a `ColumnSortButton` component.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -116,15 +116,44 @@ function EventsSection() {
...
+            getter: event => event.involvedObject.kind,
+            sort: (e1: Event, e2: Event) => {
+              if (e1.involvedObject.kind < e2.involvedObject.kind) {
+                return 1;
+              } else if (e1.involvedObject.kind > e2.involvedObject.kind) {
+                return -1;
+              } else {
+                return 0;
+              }
+            }
```
joaquimrocha, line 128: This is the thing we can easily make automatic as written in my previous patch's comments.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -116,15 +116,44 @@ function EventsSection() {
...
           },
           {
             label: 'Name',
             getter: event => event.involvedObject.name,
+            sort: (e1: Event, e2: Event) => {
+              const e1Name = e1.involvedObject.name.toUpperCase();
+              const e2Name = e2.involvedObject.name.toUpperCase();
```
joaquimrocha, line 135: I don't think uppercasing is necessary. For the rest, please see my previous comments.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -116,15 +116,44 @@ function EventsSection() {
...
             getter: event => timeAgo(event.metadata.creationTimestamp),
+            sort: (e1: Event, e2: Event) => {
+              if (e1.metadata.creationTimestamp < e2.metadata.creationTimestamp) {
+                return 1;
+              } else if (e1.metadata.creationTimestamp > e2.metadata.creationTimestamp) {
+                return -1;
+              } else {
+                return 0;
+              }
+            }
```
joaquimrocha, line 156: This one we need to override (because the getter will return `timeAgo` which is not sortable). Note that the function can be highly simplified as `(e1: Event, e2: Event) => e2.metadata.creationTimestamp - e1.metadata.creationTimestamp`.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -136,12 +165,14 @@ function EventsSection() {
               >
                 <Box>{makeStatusLabel(event)}</Box>
               </LightTooltip>,
+            sort: false
```
joaquimrocha, line 168: The default value is something I am not too confident giving a value for, but I am leaning towards it being `false`. i.e. if a caller wants to have sorting by default, they will need to just set `sort: true`.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -136,12 +165,14 @@ function EventsSection() {
...
           },
         ]
           :
           []
         }
         data={events ? events.sort(sortFunc) : null}
+        isDataAlreadySortedInOneOfTheOrder
```
joaquimrocha, line 175: Instead of this property, maybe a more flexible and simpler way to do this is to just set whether we need a table to be sorted by a table. For example: `defaultSortingColumn: number`. This way, if it's set to `defaultSortingColumn: -2`, we can sort decendingly (or ascendingly if it's `2`) using the 2nd column (we cannot use 0 because we cannot use minus with it).

joaquimrocha (changes requested):
It's almost there but there are some more things we can do to make it simpler.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -62,6 +67,56 @@ export interface SimpleTableProps {
...
+        <Icon icon={menuUp}/>
+      </IconButton>
+      <IconButton onClick={() => {
+        setSortColIndex(sortColIndex);
+        setIsIncreasingOrder(true);
+      }}
+      >
+        <Icon icon={menuDown}/>
+      </IconButton>
+    </>;
```
joaquimrocha, line 119: What's the difference between this block and the one above? (Maybe it's still early here and I am not seeing it)

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -71,12 +126,16 @@ export default function SimpleTable(props: SimpleTableProps) {
...
     errorMessage = null,
+    defaultSortingColumn
   } = props;
   const [page, setPage] = React.useState(0);
   const [currentData, setCurrentData] = React.useState(data);
   const rowsPerPageOptions = props.rowsPerPage || [5, 10, 50];
   const [rowsPerPage, setRowsPerPage] = React.useState(rowsPerPageOptions[0]);
   const classes = useTableStyle();
+  const [isIncreasingOrder, setIsIncreasingOrder] = React.useState(!!(defaultSortingColumn &&
+    defaultSortingColumn > 0));
```
joaquimrocha, line 137: Does this mean that if we have a defaultSortingColumn than we assume it's in increasing order, otherwise it's not? Why is that (instead of just choosing true or false)?

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -104,6 +163,64 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    if (typeof sortFunction === 'boolean' && sortFunction){
+      setCurrentData([...data.sort(defaultSortingFunction(columnAskingForSort as
+        SimpleTableGetterColumn))]);
+      return;
+    }
+
+    if (typeof sortFunction === 'function') {
+      applySort = (arg1: any, arg2: any) => {
+        const orderChanger = isIncreasingOrder ? 1 : -1;
+        return sortFunction(arg1, arg2) as any * orderChanger;
```
joaquimrocha, line 201: Should it be `as number` instead?

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -104,6 +163,64 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+      };
+      const sortedData = data.sort(applySort);
+      setCurrentData([...sortedData]);
```
joaquimrocha, line 204: AFAIU, this will set the `currentData` different from the `data`, so as soon as we change the sorting order we will get the `Refresh` button shown. Same if the same data is set up in the table when the current data is the same but sorted.
This is the reason why we use `filteredData` as a separate variable rather than filtering `currentData`.

I am not sure if sorting the `currentData` before the `filteredData` (i.e. not storing sorted data) is going to impact the performance a lot, but if it does, then maybe the solution is to either set the `data` as `currentData` on a sorting request when it's still equal to `currentData`  (so you don't miss the new data if it changed), or to use a third state variable called `displayData` (for holding the filtered/sorted data, but this seems to create confusion as well).

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -155,6 +272,15 @@ export default function SimpleTable(props: SimpleTableProps) {
                     {...otherProps}
                   >
                     {label}
+                    {sort &&
+                    <ColumnSortButtons
+                      defaultSortingColumn={defaultSortingColumn}
+                      isIncreasingOrder={isIncreasingOrder}
+                      setIsIncreasingOrder={setIsIncreasingOrder}
+                      sortColIndex={i}
+                      setSortColIndex={setSortColIndex}
```
joaquimrocha, line 281: Maybe we just have to pass it `isIncreasingOrder` (for the right icon to be shown) and then use a `onClick` and just have the callback use the column for sorting here (instead of passing 5 properties).

frontend/src/components/cluster/Overview.tsx
```diff
@@ -135,13 +128,14 @@ function EventsSection() {
...
-              </LightTooltip>,
+              </LightTooltip>
           },
         ]
           :
           []
         }
-        data={events ? events.sort(sortFunc) : null}
+        data={events}
+        defaultSortingColumn={2}
```
joaquimrocha, line 138: I thought we were going to use the idea that the `defaultSortingColumn` is indexed from 1 so we can use negative values for decreasing sorting without having a 2nd property passed.

joaquimrocha (changes requested):
Getting very close but I think we need some more improvements.

frontend/src/components/common/SimpleTable.tsx
ashu8912: so if we have defaultSortingColumn than we check whether its value is greater than 0 or not and if that's true we know the default sorting order is increasing

frontend/src/components/common/SimpleTable.tsx
ashu8912: in case the table is not default sorted we show both buttons

frontend/src/components/common/SimpleTable.tsx
ashu8912: yes you are right, but data is a prop to the SimpleTable component and we should never modify prop directly in a component, Thus we are only left with the confusing solution which is to use one more state called displayData

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -62,6 +67,56 @@ export interface SimpleTableProps {
...
+          <Icon icon={menuUp}/>
+        </IconButton> :
+        <IconButton onClick={() => {
+          setSortColIndex(sortColIndex);
+          setIsIncreasingOrder(true);
+        }}
+        >
+          <Icon icon={menuDown}/>
+        </IconButton>
+      }
```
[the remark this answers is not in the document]
ashu8912: Thanku this is more readable :D

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -71,12 +101,17 @@ export default function SimpleTable(props: SimpleTableProps) {
...
   } = props;
   const [page, setPage] = React.useState(0);
   const [currentData, setCurrentData] = React.useState(data);
+  const [displayData, setDisplayData] = React.useState(data);
   const rowsPerPageOptions = props.rowsPerPage || [5, 10, 50];
   const [rowsPerPage, setRowsPerPage] = React.useState(rowsPerPageOptions[0]);
   const classes = useTableStyle();
+  const [isIncreasingOrder, setIsIncreasingOrder] = React.useState(!!(defaultSortingColumn &&
+    defaultSortingColumn > 0));
+  const [sortColIndex, setSortColIndex] = React.useState(0);
```
joaquimrocha, line 114: By default we shouldn't sort anything (so we keep the current behavior of just showing whatever the data has).
The way I understand this logic is that we have this `sortColIndex` but also `defaultSortingColumn` and then we use the latter once below, and the other if the user clicks it.

This think we can deeply simplify this by using just the `sortColIndex` that will take the value of `defaultSortingColumn` (or `defaultSortingColumn != 0 : abs(defaultSortingColumn) - 1 : null` if we keep a separate `isIncreasingOrder` state for the order alone, being `null` the "don't sort anything" state).

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -99,17 +134,85 @@ export default function SimpleTable(props: SimpleTableProps) {
     // it directly. Otherwise it will require user's intervention.
     if ((!currentData || currentData.length === 0) || page === 0) {
       setCurrentData(data);
+      const sortedData = getSortData();
+      setDisplayData(sortedData || data);
```
joaquimrocha, line 138: Don't use a single const just once here. Use the `getSortData()` directly.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -99,17 +134,85 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+      return compareValue * (isIncreasingOrder ? 1 : -1);
+    }
+
+    return defaultSortingReal;
+  }
+
+  function getSortData() {
+    if (!data) {
+      return null;
+    }
```
joaquimrocha, line 165: Check whether we want to sort anything at all.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -99,17 +134,85 @@ export default function SimpleTable(props: SimpleTableProps) {
...
+    const columnAskingForSort = columns[defaultSortingColumn - 1];
+    if (!columnAskingForSort) {
+      return;
+    }
+    setDisplayData(data.slice().sort(defaultSortingFunction(
+      columnAskingForSort as SimpleTableGetterColumn)));
+  },
+  // eslint-disable-next-line
+  []);
```
joaquimrocha, line 208: See my previous comments. We shouldn't need two sorting effects.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -155,6 +263,13 @@ export default function SimpleTable(props: SimpleTableProps) {
                     {...otherProps}
                   >
                     {label}
+                    {sort &&
+                    <ColumnSortButtons
+                      isIncreasingOrder={isIncreasingOrder}
+                      isDefaultSorted={Boolean(defaultSortingColumn)}
```
joaquimrocha, line 269: We are always passing `true` here (unless we have `defaultSortingColumn` as `0`).
Shouldn't we instead pass whether this column is the default one (so we should the buttons' according to that)?

I think what we want for each `ColumnSortButtons` is:
Is this column the currently sorting one?
 Yes -> then show the Up/Down according to the required state (by default is Increasing Order)
 No -> show both buttons so the user can choose

When changing to a different column sorting, the new order shouldn't have to do with the current one (because we're talking about different columns).
We can still manage the `isIncreasingOrder` state outside of the sort buttons as you're doing if you prefer, but we should reset the state when switching to a different column (i.e. would be decided in `sortClickHandler` above).

joaquimrocha (changes requested):
We can simplify it even more. See my comments.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -62,6 +67,31 @@ export interface SimpleTableProps {
   rowsPerPage?: number[];
   emptyMessage?: string;
   errorMessage?: string | null;
+  defaultSortingColumn? : number;
+}
+
+interface ColumnSortButtonProps {
+   isDefaultSorted: boolean;
+   isIncreasingOrder: boolean;
+   clickHandler: () => void;
```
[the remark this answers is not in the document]
ashu8912: Or we can also just pass the correct boolean true for increasing and false for decreasing in the clickHandler without taking the isIncreasingOrder from SimpleTable component to ColumnSortButton

frontend/src/components/common/SimpleTable.tsx
ashu8912: So Boolean(defaultSortingColumn) always return true but returns false only if the column index passed is 0 or this prop wasn't passed in that case undefined i think this is what we want no?

frontend/src/components/common/SimpleTable.tsx
ashu8912: So this is where our onMount effect is helpful in that we check whether there is a defaultSortingColumn and if not don't do anything with data(don't sort)

joaquimrocha (changes requested):
I still think the simplifications I had in mind need to be added. In fact, trying it out I think it's showing the problem of a shared state among columns (given by the default sorting column):
Screenshot from 2020-11-06 22-00-38

All of the column buttons are showing the up or down icons, when only the column being used for the sort should show the up or down icon. The others should have both (as far as I understood from what the intention in the code was).

joaquimrocha:
@ashu8912 , Please check the changes I propose in the fix up commit which simplify the patch by not having two sort effects and rely solely on the `sortColIndex` for setting up the sort.
I have also changed the 2 buttons into a "swap" button when the column is not being used for the sorting, as having two buttons was taking too much space and didn't look great. And the buttons are smaller now.

Please verify the code and test it if you can.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -108,9 +104,11 @@ export default function SimpleTable(props: SimpleTableProps) {
...
   const [rowsPerPage, setRowsPerPage] = React.useState(rowsPerPageOptions[0]);
   const classes = useTableStyle();
-  const [isIncreasingOrder, setIsIncreasingOrder] = React.useState(
-    isNaN(defaultSortingColumn as number) || (defaultSortingColumn && defaultSortingColumn > 0));
-  const [sortColIndex, setSortColIndex] = React.useState(0);
+  const [isIncreasingOrder, setIsIncreasingOrder] =
+    React.useState(!defaultSortingColumn || defaultSortingColumn > 0);
+  // We use a -1 value here if no sorting should be done by default.
+  const [sortColIndex, setSortColIndex] =
+    React.useState(defaultSortingColumn ? defaultSortingColumn - 1 : -1);
```
ashu8912, line 111:
```suggestion 
    React.useState(defaultSortingColumn ? Math.abs(defaultSortingColumn) - 1 : -1);
```

joaquimrocha (approved):
We could have abstracted the metadata.name comparison into a different function in the base class but let's add that later.

PR #101, opened 2020-10-29
Adds Shrink Sidebar according to the new design and also navigation tabs

ashu8912:


frontend/src/components/common/Tab.tsx
```diff
@@ -20,25 +20,40 @@ interface TabsProps {
...
+  resetState?: boolean;
+  defaultIndex?:number | boolean;
   onTabChanged?: (tabIndex: number) => void;
 }
 
 export default function Tabs(props: TabsProps) {
-  const {tabs, tabProps = {}, onTabChanged = null} = props;
-  const [tabIndex, setTabIndex] = React.useState(0);
+  const {tabs, tabProps = {}, isDefaultSelectedRequired = true,
+         resetState = false, defaultIndex = 0, onTabChanged = null} = props;
```
joaquimrocha, line 31: Use one prop per line since they're a lot now.

frontend/src/components/common/Tab.tsx
```diff
@@ -20,25 +20,40 @@ interface TabsProps {
...
+  const [tabValue, setTabValue] = React.useState<number | boolean>(
+    defaultIndex > -1 ? 0 : defaultIndex);
```
joaquimrocha, line 33: Is this the same as `min(defaultIndex, 0)`?

frontend/src/components/common/Tab.tsx
```diff
@@ -20,25 +20,40 @@ interface TabsProps {
   tabProps?: {
     [propName: string]: any;
   };
+  isDefaultSelectedRequired?: boolean;
```
joaquimrocha, line 23: The title of this patch should be less generic. Also, if you need to pass it a default index or nothing, why not just use `number | null` and watch if that property changes so you can also control it from outside?

frontend/src/redux/actions/actions.tsx
```diff
@@ -82,6 +84,14 @@ export function setSidebarSelected(selected: SidebarType['selected']) {
   return { type: UI_SIDEBAR_SET_SELECTED, selected };
 }
 
+export function setWhetherSidebarOpen(isSidebarOpen: boolean) {
+  return { type: UI_SIDEBAR_SET_WHETHER_OPEN, isSidebarOpen};
```
joaquimrocha, line 88: Please use `UI_SIDEBAR_SET_EXPANDED`.

frontend/src/redux/actions/actions.tsx
```diff
@@ -82,6 +84,14 @@ export function setSidebarSelected(selected: SidebarType['selected']) {
...
+}
+
+export function setSidebarParentIconClicked() {
+  return { type: UI_SIDEBAR_SET_PARENT_ICON_CLICKED};
+}
+
```
joaquimrocha, line 94: I am not sure why we need this one.

frontend/src/components/Sidebar.tsx
```diff
@@ -381,12 +384,12 @@ function ListItemLink(props: ListItemLinkProps) {
 }
 
 export default function Sidebar() {
-  const classes = useStyle();
   const sidebar = useTypedSelector(state => state.ui.sidebar);
+  const dispatch = useDispatch();
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const items = React.useMemo(() => prepareRoutes(), [sidebar.entries]);
   const [open, setOpen] = React.useState(false);
-
```
joaquimrocha: Don't remove this newline.

frontend/src/App.tsx
```diff
@@ -258,6 +259,8 @@ function App() {
               <main className={classes.content}>
                 <div className={classes.toolbar} />
                 <Container maxWidth="lg">
+                  <NavigationTabs/>
+                  <Divider />
```
joaquimrocha, line 263: This divider should be shown only if the tabs are in use (otherwise it will look out of place).
Screenshot from 2020-11-04 12-43-49

frontend/src/components/Sidebar.tsx
```diff
@@ -410,10 +413,13 @@ export default function Sidebar() {
       }}
     >
       <div className={classes.toolbar}>
-        <Button onClick={() => setOpen(!open)}>
+        <Button onClick={() => {
+          dispatch(setWhetherSidebarOpen(!open));
+          setOpen(!open);}}
```
joaquimrocha, line 418: Add the `}}` in the next line please.

joaquimrocha: Maybe we could stop using the open state and rely on redux in this case?

frontend/src/components/Sidebar.tsx
```diff
@@ -600,3 +606,78 @@ export function useSidebarItem(itemName: string | null) {
   []);
 }
 
+function searchNameInSubList(sublist : SidebarEntry['subList'], name: string): boolean {
+  if (!sublist){
```
joaquimrocha, line 610: Add a space after `)`.

frontend/src/components/Sidebar.tsx
```diff
@@ -600,3 +606,78 @@ export function useSidebarItem(itemName: string | null) {
...
+    return false;
+  }
+  for (let i = 0;i < sublist.length;i++) {
```
joaquimrocha, line 613: Add a space after `;`.

frontend/src/components/Sidebar.tsx
```diff
@@ -600,3 +606,78 @@ export function useSidebarItem(itemName: string | null) {
...
+  return false;
+}
+
+function findParentOfSubList(list: SidebarEntry[], name: string | null): SidebarEntry | null{
+  if (!name) {
+    return null;
+  }
+
+  let parent = null;
+  for (let i = 0;i < list.length;i++) {
```
joaquimrocha, line 627: Add a space after `;`.

frontend/src/components/Sidebar.tsx
```diff
@@ -600,3 +606,78 @@ export function useSidebarItem(itemName: string | null) {
...
+      pathname = generatePath(getClusterPrefixedPath(url), {cluster: getCluster()!});
+    } else {
+      pathname = createRouteURL(subList[index].name);
+    }
+    history.push({pathname});
+  }
+  const tabRoutes = subList.map((item: any) => {
+    return {'label': item.label, component: <></>};
+  });
+
```
joaquimrocha, line 673: This function has several logical separations/blocks. Please try to group them by using a newline where appropriate.

frontend/src/components/Sidebar.tsx
```diff
@@ -600,3 +606,78 @@ export function useSidebarItem(itemName: string | null) {
...
+  defaultIndex = subList.findIndex((item) => item.name === sidebar.selected);
+  return (
+    <Tabs
+      tabs={tabRoutes}
+      isDefaultSelectedRequired={isDefaultSelectedInTabs}
+      resetState={sidebar.isIconClicked}
+      onTabChanged={(index) => {tabChangeHandler(index);}}
+      defaultIndex={defaultIndex}
```
joaquimrocha, line 681: Right. I think this is too complicated. I'd rather just pass it the `tabIndex` to determine the default tab, and then if the tabs are to be hidden, you can return `null` here, so the next time it's rebuilt, the default tab will be set correctly.
You can also move the divider here so it'll not be shown when the tabs should be hidden.

joaquimrocha (changes requested):
Cool stuff! I like how you managed to include the tabs and show/hide them accordingly.
I left some comments that I think would make the changes simpler. Otherwise, I'd also kindly ask you to go through the PR and remove/adapt the commits (even/especially if they are mine, since they're old).

Some UI comments:
Please add more spacing between the tabs and the section where needed:
Screenshot from 2020-11-04 12-45-01

The version icon should have the kubernetes icon on top of the version when shrunk (maybe this can be done automatically with CSS):
Screenshot from 2020-11-04 12-45-46

joaquimrocha:
I forgot to say this too: please store the sidebar state (expanded = true/false) in the local storage (stored from a saga) so we can keep the state between sessions.
The schema for storing should be something like:
```json
settings: {
  ui: {
    sidebar: {
      expanded: true
    }
  }
}

```

frontend/src/components/Sidebar.tsx
```diff
@@ -599,3 +608,84 @@ export function useSidebarItem(itemName: string | null) {
...
+  const tabRoutes = subList.map((item: any) => {
+    return {'label': item.label, component: <></>};
+  });
+
+  defaultIndex = subList.findIndex((item) => item.name === sidebar.selected);
+  return (
+    <>
+      <Tabs
+        tabs={tabRoutes}
+        resetState={sidebar.isIconClicked}
```
joaquimrocha, line 716: Are you sure this is needed? I don't love having this in the `Tabs` component and would prefer if we rely on an `index` property from outside.

frontend/src/components/Sidebar.tsx
```diff
@@ -336,9 +338,15 @@ function VersionButton() {
...
       >
-        {clusterVersion.gitVersion}
+        <Box display={sidebar.isSidebarOpen ? 'flex' : 'block'} alignItems="center">
+          <Box>
+            <Icon color="#adadad" icon={kubernetesIcon} className={classes.versionIcon} />
+          </Box>
+          <Box>
+            {clusterVersion.gitVersion}
+          </Box>
+        </Box>
```
joaquimrocha, line 349: It would be great if we could just rely on CSS instead of having to use the sidebar-open logic, etc.
Maybe using a Grid?

frontend/src/redux/actions/actions.tsx
joaquimrocha: If clicking on the sidebar changes the route, do we really need this?

joaquimrocha (changes requested):
Left some comments still.
I tried it again and it's working great, so it's a matter of improving the code now.

I noticed also that this part hasn't yet been fixed (needs a space below the tabs or above the charts' area):
Screenshot from 2020-11-06 12-09-03

And the tooltips for the sidebar could double in size as they're very tiny.

joaquimrocha:
@ashu8912 Please check the quick and WIP changes I have added in a new commit. I think this may help understand the ideas I tried to communicate earlier today. Cheers.

frontend/src/components/common/Tab.tsx
```diff
@@ -20,25 +20,42 @@ interface TabsProps {
...
+    tabProps = {},
+    defaultIndex = 0,
+    onTabChanged = null
+  } = props;
+  const [tabValue, setTabValue] = React.useState<TabsProps['defaultIndex']>(
+    defaultIndex && Math.min(defaultIndex as number, 0));
 
   function handleTabChange(event: any, newValue: number) {
-    setTabIndex(newValue);
+    setTabValue(newValue);
```
joaquimrocha, line 38: There's no need for this because what we're setting is the index and thus changing the name to value makes the code less clear IMO.

frontend/src/redux/reducers/ui.tsx
```diff
@@ -65,14 +68,14 @@ function reducer(state = INITIAL_STATE, action: Action) {
       newFilters.sidebar = {
         ...newFilters.sidebar,
         selected: action.selected,
-        isVisible: !!action.selected,
+        isVisible: !!action.selected
```
joaquimrocha, line 71: This is not needed and actually adds noise to the diff when we add a new line to this object. Same below.

joaquimrocha (changes requested):

frontend/src/components/Sidebar.tsx
joaquimrocha: Let's keep this for now then.

joaquimrocha (approved):

PR #103, opened 2020-11-03
Add Dark Theme

ashu8912:


frontend/src/App.tsx
```diff
@@ -33,9 +34,6 @@ import store from './redux/stores/store';
 
 const useStyle = makeStyles(theme => ({
   root: {
-    display: 'flex',
```
joaquimrocha, line 108: Would this patch be simpler if we were using Redux to set/retrieve the theme name? Themes like this patch changes a lot of code just for the theme name to be changed by an inner component.

joaquimrocha (changes requested):
Please check my comments.

BTW, just like the sidebar, I think we should use the theme setting in Redux and store it locally but a bit more structured in a schema like this:
```
settings: {
  ui: {
    theme: 'dark'
  }
}
```

I think we should add a saga that stores locally whatever is changed in this settings Redux state (so it works for whatever we add there in the future). I am fine with this part being done in a different PR.

frontend/src/App.tsx
ashu8912: No, So here the problem is not that we aren't getting the correct theme, we are but we are using the useStyles for App before we could even get the new theme via ThemeProvider that's why this patch renders the entire app in a container and that container is wrapped by ThemeProvider and in that container itself we use the useStyles and not outside the ThemeProvider. I hope i didn't confused you :sweat_smile:

ashu8912: @joaquimrocha i would prefer to use the settings redux state to store it locally to be done in a different pr

joaquimrocha (approved):
Awesome! LGTM!

PR #104, opened 2020-11-06
Add network status alert notification bar

ashu8912:


frontend/src/App.tsx
```diff
@@ -255,6 +256,7 @@ function App() {
               <CssBaseline />
               <TopBar />
               <Sidebar />
+              <AlertNotification/>
```
joaquimrocha, line 259: The mockups show the bar in the central area:
Screenshot from 2020-11-06 22-15-49

Trying this PR it shows on top of the whole view. I don't dislike it in a global position, and I know it's also simpler to execute this way but it will show on top of the cluster button. And if we're having a problem in just one cluster, changing to a different one will be a bit more difficult.

frontend/src/components/common/AlertNotification.tsx
```diff
@@ -0,0 +1,113 @@
...
+    return setInterval(() => {
+      setError(null);
+      testAuth().then(() => {
+        setError(false);
+      })
+        .catch((err) => {
+          const error = new Error(err);
+          setError(error.message);
+          setNetworkStatusCheckTimeFactor(networkStatusCheckTimeFactor =>
+            networkStatusCheckTimeFactor + 1);
```
joaquimrocha, line 29: Why is this using a function here? Shouldn't we just use `++networkStatusCheckTimeFactor`?

frontend/src/components/common/AlertNotification.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export default function AlertNotification(){
+  const routes = useTypedSelector(state => state.ui.routes);
+  const [networkStatusCheckTimeFactor, setNetworkStatusCheckTimeFactor] = React.useState(0);
+  const [error, setError] = React.useState<null | string | boolean>(null);
+  const [intervalID, setIntervalID] = React.useState<NodeJS.Timeout | null>(null);
+
+  function registerSetInterval(): NodeJS.Timeout {
+    return setInterval(() => {
+      setError(null);
```
joaquimrocha, line 21: I think if you set the error as null here, it will force 2 refreshes: one where the error bar is removed, and another when the result from the testAuth arrives. Can you double check this?

joaquimrocha (changes requested):
Left a couple of comments but it's already in a very good proposal.

frontend/src/components/common/AlertNotification.tsx
ashu8912: as setState is asynchronous we can't rely on networkStatusCheckTimeFactor to be the new state value whenever we are incrementing, The callback to setState ensures that it is always called with the latest state value

frontend/src/components/common/AlertNotification.tsx
ashu8912: setting the null there means please dont show the alert notification until i have achieved the result from the testAuth method otherwise while its going to make an ajax call it will show that there is an issue we know there is an issue only after the request is made

joaquimrocha (changes requested):
Just marking as requiring changes (discussed on Slack this week).

ashu8912:
the linter is failing for electron code i have already added a fix for that in #178

joaquimrocha:
@ashu8912 , I have merged the #178 and re-run the CI jobs here but it's still failing.
I have also noticed that the red bar is still showing on the main area instead of on the entire page. This is fine as my request for it to be in the entire page had to do with it being easier to re-use in the login screen.
However, in the login screen I don't see the bar.

@illume , if you can, please review this PR's code as I cannot do it ATM.

joaquimrocha:
Another thing is that if the network is off, we see the bar, and when it comes back online, the bar disappears but the views showing "Error" remain as such.

I am starting to think we need something more than just the bar, like a central setting showing the status, and requests taking that into account when performing their tasks.

joaquimrocha (approved):

PR #105, opened 2020-11-09
frontend: Show plugins' header actions before others

joaquimrocha:
We were showing the plugins' header actions after the default ones but
this means that the UI is less consistent among views, e.g.: if a plugin
added a header button in pods' details but not in the deployments', then
the other buttons would be pushed left. This was even more evident if
the plugin took a while to add the button (even leading to a misclick).

This patch makes the plugins' header action components be added to the
left, so the rest of the buttons don't move in such case.


ashu8912 (approved):
The linter was failing so i added the fix other than that  the changes LGTM

joaquimrocha:
Weird. I had also force pushed the long line linter fix...

PR #107, opened 2020-11-11
Store sidebar shrink status to localStorage

ashu8912:


ashu8912:
This is a temporary solution for now and in future this will be replaced by a proper ui setting panel with ui related settings that needs to be stored in the browserStorage

joaquimrocha (approved):
LGTM.

PR #108, opened 2020-11-13
Add make targets for desktop app builds

ashu8912:
This patch adds make targets for desktop app build for all three
platforms.

# How to use
run make app to generate desktop app builds for all three platforms 
use app-win for windows build generation
use app-mac for mac specific build generation
use app-linux for linux build generation


joaquimrocha (approved):
I think we still need to run `npm install` on the forntend when building the app, but we can add this later.

PR #115, opened 2020-11-17
Update _index.md

danielabbatt:
Fix title formatting which looks wrong when viewed as markdown on github

Documentation change only


joaquimrocha:
Hi @danielabbatt , we're using our docs rendered that's based on the Kubernetes documentation and needs to use front-matter for titles, rather than simple markdown.
See https://kinvolk.io/docs/headlamp/latest/

PR #118, opened 2020-11-17
frontend: Decode secrets when showing them

joaquimrocha:
The secrets should be decoded from base64 when the user requests to
view them.


ashu8912 (approved):
Tested,LGTM

PR #121, opened 2020-11-17
Service should come before the deployment

just1689:
# Service before deployment

When a pod is scheduled Kubernetes checks if it will match existing services. If so it annotates the pod with the service name. This is used by sidecars for various use cases 

# How to use
NA

# Testing done
NA

### DIFF
--- a/kubernetes-headlamp.yaml
+++ b/kubernetes-headlamp.yaml
@@ -1,3 +1,15 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: headlamp
+  namespace: kube-system
+spec:
+  ports:
+    - port: 80
+      targetPort: 4466
+  selector:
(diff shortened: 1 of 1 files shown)


joaquimrocha (changes requested):
Looks great! Thank you!

If you don't mind, can you use the commit messages' format we recommend in our [guidelines](https://github.com/kinvolk/contribution/blob/master/topics/git.md#authoring-prs)?

Something like:
```
kubernetes-headlamp.yaml: Services should come before the deployment

This allows the pod to be annotated with the matching service name when it is scheduled.
```

Thanks!

joaquimrocha:
@just1689 , @ashu has provided the small commit message fix and I have merged it into master.
I will hence close this PR. Thank you again for your contribution and I hope to see more improvements by you in the future.

PR #125, opened 2020-11-18
Don't stop the server if no config is found

ashu8912:
we shouldn't use log.Fatal if there is no kubeconfig available as it gives a call to os.Exit and thus stops the server.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -36,6 +36,7 @@ type HeadlampConfig struct {
 
 type clientConfig struct {
 	Clusters []Cluster `json:"clusters"`
+	Error    string    `json:"error"`
 }
 
 type spaHandler struct {
@@ -77,6 +78,7 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func StartHeadlampServer(config *HeadlampConfig) {
 	kubeConfigPath := ""
 
+	var configError string
 	// If we don't have a specified kubeConfig path, and we are not running
 	// in-cluster, then use the default path.
 	if config.kubeConfigPath != "" {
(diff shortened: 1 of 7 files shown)


backend/cmd/headlamp.go
```diff
@@ -91,7 +93,7 @@ func StartHeadlampServer(config *HeadlampConfig) {
 	if config.useInCluster {
 		context, err := GetOwnContext(config)
 		if err != nil {
-			log.Println("Failed to get in-cluster config", err)
+			configError = fmt.Sprintf("Failed to get in-cluster config")
```
joaquimrocha, line 96: You don't need Sprintf here.

backend/cmd/headlamp.go
```diff
@@ -103,7 +105,7 @@ func StartHeadlampServer(config *HeadlampConfig) {
 
 		contextsFound, err := GetContextsFromKubeConfigFile(kubeConfigPath)
 		if err != nil {
-			log.Println("Failed to get contexts from", kubeConfigPath, err)
+			configError = fmt.Sprintf("Failed to get contexts from %s", kubeConfigPath)
```
joaquimrocha, line 108: Maybe we could send the error as a `{ message: string, code: number }` so we can check for the code in the UI and show a nicer error message.

joaquimrocha (changes requested):
Thanks. Left some comments. The message can also be improved, but I think it's okay to do that in a different PR (and not immediately).

backend/cmd/headlamp.go
ashu8912: @joaquimrocha i think the error string in itself is easy to manage because if it's not there than no error if it's there than there is an error why do we need the code???

joaquimrocha: Because the code can tell us what the error is: e.g. if we use for example error codes, we can send the error that the server had when reading the config file, e.g. `ENOENT`. And thus we can quickly check if the error code is this in the frontend and make a better message for the user: `The kube config file was not found. Please read this documentation on how to set up K8s. [ Link to the docs ]`

If we only have the message, then we're configuring something that is to be facing the user (error message) from the backend, and we don't have to e.g. add JSX code to the error message in the backend.

joaquimrocha:
Closed in favor of #134

PR #131, opened 2020-11-26
Desktop: Kill the server running process on app quit

ashu8912:
On desktop app quit kill the backend server on app close
# Testing done
Yes

### DIFF
--- a/app/electron/main.js
+++ b/app/electron/main.js
@@ -167,17 +167,23 @@ autoUpdater.on('update-downloaded', () => {
 });
 
 app.on('ready', createWindow);
-app.on('window-all-closed', function () {
+
+app.on('quit', function () {
   if (serverProcess) {
     serverProcess.stdin.pause();
     serverProcess.kill();
-  }
-  if (process.platform !== 'darwin') {
-    app.quit();
+    serverProcess = null;
   }
 });
+
 app.on('activate', function () {
   if (mainWindow === null) {
     createWindow();
   }
+});
+
+app.once('window-all-closed', app.quit);
+
+app.once('before-quit', () => {
+  mainWindow.removeAllListeners('close');
 });


joaquimrocha (changes requested):
Thanks @ashu8912 .
I think this is a bit overkill (pun intended). We basically need to eventually change how Headlamp's port is managed when running locally.  Because, what happens if there's already a service running in the `4466` port?

This doesn't have to be fixed for the next version, but we also shouldn't just directly kill the port (what if you have an important service running in that port!?).

Looking at the code we had already, maybe we just need to call `serverProcess.kill()` when the app quits (which means that maybe the problem is that the `window-all-closed` event doesn't happen when we quit the app directly, i.e. we may just need `app.on('quit, …` + `serverProcess.kill()`). Of course this is all theory at this point, but tomorrow I can try it in the Macbook.

joaquimrocha:
I check whether this assumption (that only the `quit` happens when quitting the app, and not the `windows-all-closed`) would happen on Linux too and indeed it does! So the fix for this issue should be adding the `on('quit'` and kill the server process in there.

app/electron/main.js
```diff
@@ -167,17 +167,22 @@ autoUpdater.on('update-downloaded', () => {
...
 
 app.on('ready', createWindow);
-app.on('window-all-closed', function () {
+
+app.on('quit', function () {
   if (serverProcess) {
     serverProcess.stdin.pause();
     serverProcess.kill();
   }
-  if (process.platform !== 'darwin') {
```
joaquimrocha, line 175: Maybe we want to set the serverProcess to `null` after killing it above? So if the signal happens twice for some reason we don't call kill twice.

joaquimrocha (changes requested):

joaquimrocha (approved):

PR #133, opened 2020-11-27
backend: Don't print an error when the plugins folder doesn't exist

joaquimrocha:
Plugins are optional in Headlamp, so we shouldn't print an error if the
folder doesn't exist.


ashu8912 (approved):
LGTM

PR #137, opened 2020-12-03
Link service account Details view to secret details view

ashu8912:
On the service account details view page we want to connect the secret name to its own page this will reduce copy pasting and manually looking for that secret name in secrets list.

### DIFF
--- a/frontend/src/components/serviceaccount/Details.tsx
+++ b/frontend/src/components/serviceaccount/Details.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useParams } from 'react-router-dom';
 import ServiceAccount from '../../lib/k8s/serviceAccount';
+import { Link } from '../common';
 import { MainInfoSection } from '../common/Resource';
 
 export default function ServiceAccountDetails() {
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/SimpleTable.tsx
```diff
@@ -338,7 +338,7 @@ const useStyles = makeStyles(theme => ({
 
 export interface NameValueTableRow {
   name: string | JSX.Element;
-  value?: string | JSX.Element;
+  value?: string | JSX.Element | JSX.Element[];
```
joaquimrocha, line 341: You don't need this if you wrap the whole value in a fragment or another element, right?
It's not a big deal but if we don't need it now, we shouldn't add it.

frontend/src/components/serviceaccount/Details.tsx
```diff
@@ -15,7 +16,13 @@ export default function ServiceAccountDetails() {
       extraInfo={item && [
         {
           name: 'Secrets',
-          value: item.secrets.map(({name}) => name).join(', ')
+          value: item.secrets.map(({name}, index) => <React.Fragment key={`${name}__${index}`}>
+            <Link routeName={'secret'} params={{namespace, name}}>
+              {name}
+            </Link>
+            {index !== item.secrets.length - 1 && ','}
```
joaquimrocha, line 23: Maybe just add one per line. I think that will look better if there are more than one link.
If you wrap the result of `item.secrets.map` in a fragment, then you don't need the previous commit.

joaquimrocha (changes requested):

joaquimrocha (approved):

PR #138, opened 2020-12-04
backend: make it possible to serve SSL

aholler:
This adds 4 new optional command line parameters:

  -ssl-cert filename
  -ssl-cert-key filename
  -ssl-cert-ca filename
  -ssl-client-cert-required

To serve SSL, the first two parameters are needed, without them HTTP is
served (like before).

The parameter -ssl-cert-ca might be used if the given SSL-certificate
uses a root CA which is not known by default.

The parameter -ssl-client-cert-required is by default false but might
be used to require that clients are having a valid client certificate.

So in order to serve SSL without something else which does SSL
termination, just mount the certificate(s) into the container (e.g. by
using a secret) and add the parameters -ssl-cert and -ssl-cert-key to
the args in the yaml with the deployment.

# How to use

# Testing done

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -22,16 +22,20 @@ import (
 )
 
 type HeadlampConfig struct {
-	useInCluster     bool
-	devMode          bool
-	insecure         bool
-	kubeConfigPath   string
-	port             string
-	staticDir        string
-	pluginDir        string
-	oidcClientID     string
-	oidcClientSecret string
-	oidcIdpIssuerURL string
+	useInCluster          bool
+	devMode               bool
+	insecure              bool
+	sslClientCertRequired bool
+	kubeConfigPath        string
+	port                  string
+	staticDir             string
+	pluginDir             string
+	oidcClientID          string
+	oidcClientSecret      string
+	oidcIdpIssuerURL      string
+	sslCert               string
+	sslCertKey            string
+	sslCertCA             string
 }
 
 type clientConfig struct {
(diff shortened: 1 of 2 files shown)


joaquimrocha:
Hi @aholler . Thanks for the PR, it looks good and clean in a quick first read.

We do try to keep changes arguments/options to a minimum necessary, so I would like to ask you more details about your use-case. For TLS we recommend to [configure the ingress](https://kinvolk.io/docs/headlamp/latest/installation/in-cluster/) appropriately; can you tell me a bit more why adding the options to Headlamp's server would be more desirable for you than using the ingress solution? Cheers.

aholler:
The first argument for serving SSL directly is to avoid the need for an ingress controller at all.
The second argument is security. headlamp is able to display secrets and if you use an ingress controller, the secrets will go unecrypted from headlamp to the ingress controller.

Best regards,

Alexander

iaguis:
Not sure this solves a common use case and I think we should carefully consider adding features that we'd have to maintain later.

> The first argument for serving SSL directly is to avoid the need for an ingress controller at all.

That's a good point but I wonder how common this use case is. Are you running apps this way or is it a theoretical concern?

>  The second argument is security. headlamp is able to display secrets and if you use an ingress controller, the secrets will go unecrypted from headlamp to the ingress controller.

So in this case you're not trusting the internal cluster network, again I'm not sure how common that is. If you don't trust your internal cluster network, I think this should be solved in a more generic way by using pod-to-pod encryption in your CNI provider or using a service mesh with mTLS enabled.

aholler:
Ingress controllers are a non-standard component of Kubernetes.

To avoid the rest of the discussion: It's ok if you don't like those 40 lines and, maybe, I should look out for some more secure dashboard which offers e2e-encryption.

joaquimrocha:
Hi @aholler , I understand your point. And your code is nice, and it's not a controversial change. Still, we need to keep changes under check and make sure they make sense for a good number of our users.
So I will close this PR for now, but I have opened #145 in order to keep an open issue about it and thus measure whether other members of the community need this feature.
If we turn out to be wrong about this, I am more than happy to include your patch. I hope you understand.

PR #140, opened 2020-12-08
Add documentation for metrics server and the notice we show when it is not installed

ashu8912:


joaquimrocha:
I've adapted a bit the text. Thanks @ashu8912 .

PR #141, opened 2020-12-10
Add dedicated list view to different  workload types

ashu8912:
This patch adds seperate list views for workload types such as deployments,jobs,statefulset etc.
Testing
Done

### DIFF
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -144,6 +144,27 @@ const LIST_ITEMS: SidebarEntry[] = [
         name: 'ReplicaSets',
         label: 'Replica Sets'
       },
+      {
+        name: 'DaemonSets',
+        label: 'Daemon Sets'
+      },
+      {
+        name: 'StatefulSets',
+        label: 'Stateful Sets'
+      },
+      {
+        name: 'Jobs',
+        label: 'Jobs'
+      },
+      {
+        name: 'Deployments',
+        label: 'Deployments'
+      },
+      {
+        name: 'CronJobs',
+        label: 'CronJobs'
+
+      }
     ]
   },
   {
--- a/frontend/src/components/daemonset/List.tsx
+++ b/frontend/src/components/daemonset/List.tsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import DaemonSet from '../../lib/k8s/daemonSet';
+import { useFilterFunc } from '../../lib/util';
+import { Link } from '../common';
+import { SectionBox } from '../common/SectionBox';
+import SectionFilterHeader from '../common/SectionFilterHeader';
+import SimpleTable from '../common/SimpleTable';
+
+export default function DaemonSetList() {
+  const [daemonSets, error] = DaemonSet.useList();
+  const filterFunc = useFilterFunc();
+
+  return (
+    <SectionBox
+      title={
+        <SectionFilterHeader
+          title="Daemon Sets"
+        />
+      }
+    >
+      <SimpleTable
+        rowsPerPage={[15, 25, 50]}
+        filterFunction={filterFunc}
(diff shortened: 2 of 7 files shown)


frontend/src/components/deployments/List.tsx
```diff
@@ -0,0 +1,100 @@
...
+          {
+            label: 'Replicas',
+            getter: (deployment) => deployment.spec.replicas || 0
+          },
+          {
+            label: 'Age',
+            getter: (deployment) => deployment.getAge(),
+            sort: (d1: Deployment, d2: Deployment) =>
+              new Date(d2.metadata.creationTimestamp).getTime() -
+              new Date(d1.metadata.creationTimestamp).getTime()
```
joaquimrocha, line 88: This should be the last column, in order to be consistent with the rest of the views.

frontend/src/components/deployments/List.tsx
```diff
@@ -0,0 +1,100 @@
...
+      return (
+        <Box pr={0.5} display="inline-block">
+          <StatusLabel status=''>
+            <span title={message} key={type}>
+              {type}
+            </span>
+          </StatusLabel>
+        </Box>);
+    });
+  }
```
joaquimrocha, line 45: Add a newline after this.

frontend/src/components/job/List.tsx
```diff
@@ -0,0 +1,76 @@
...
+          {
+            label: 'Completions',
+            getter: (job) => getCompletions(job)
+          },
+          {
+            label: 'Age',
+            getter: (job) => job.getAge(),
+            sort: (j1: Job, j2: Job) =>
+              new Date(j2.metadata.creationTimestamp).getTime() -
+              new Date(j1.metadata.creationTimestamp).getTime()
```
joaquimrocha, line 64: The age should be the last column, for consistency.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -111,6 +111,10 @@ makeKubeObject<T extends (KubeObjectInterface | KubeEvent)>(objectName: string):
       return this.jsonData![prop];
     }
 
+    getPods() {
+      return this.jsonData!.status.currentNumberScheduled;
+    }
+
```
joaquimrocha, line 117: I think this shouldn't belong to the base class, because not all subclasses have this in their jsonData.
In the K8s API reference I see this is only in the DaemonSet spec: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#daemonset-v1-apps

frontend/src/components/Sidebar.tsx
```diff
@@ -144,6 +144,22 @@ const LIST_ITEMS: SidebarEntry[] = [
...
+        label: 'Stateful Sets'
+      },
+      {
+        name: 'Jobs',
+        label: 'Jobs'
+      },
+      {
+        name: 'Deployments',
+        label: 'Deployments'
+      }
```
joaquimrocha, line 162: You're missing the CronJob from the list.

frontend/src/components/daemonset/List.tsx
```diff
@@ -0,0 +1,59 @@
...
+              return 0;
+            }
+          },
+          {
+            label: 'Namespace',
+            getter: (daemonSet) => daemonSet.getNamespace()
+          },
+          {
+            label: 'Pods',
+            getter: (daemonSet) => daemonSet.getPods()
```
joaquimrocha, line 44: This method has been added in a later commit than where it is used.

joaquimrocha (changes requested):
Left a couple of comments.

Also, maybe the `listRoute` of each of the types represented in the views should be changed to be the respective list view. This will make things a bit confusing if someone reaches the details view from the workload generic view, but we can find a solution for that later.

joaquimrocha (approved):

PR #144, opened 2020-12-15
Login improvements

joaquimrocha:
Fixes some issues and improves the UX a bit in the login dialog.

### DIFF
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -12,6 +12,7 @@ import { getCluster, getClusterPrefixedPath } from '../../lib/util';
 import { setConfig } from '../../redux/actions/actions';
 import { ClusterDialog } from '../cluster/Chooser';
 import { Loader } from '../common';
+import Empty from '../common/EmptyContent';
 import OauthPopup from '../oidcauth/OauthPopup';
 
 const ColorButton = withStyles((theme) => ({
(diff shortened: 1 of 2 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -45,8 +49,24 @@ export async function request(path: string, params: RequestParams = {},
...
+
   const url = combinePath(BASE_HTTP_URL, fullPath);
-  const response = await fetch(url, opts);
+  let response: Response = new Response(undefined,
+                                        {status: 502, statusText: 'Unreachable'});
+
+  try {
+    response = await fetch(url, {signal: controller.signal, ...opts});
+  } catch(err) {
+    // AbortController sets the error code as 20
```
knrt10, line 62:
```suggestion
    // AbortController sets the error code as 20.
```

frontend/src/components/authchooser/index.tsx
```diff
@@ -90,24 +97,37 @@ function AuthChooser(){
...
             if (!useToken) {
               history.replace(from);
             }
 
-            setTestingAuth(false);
-
             // If we reach this point, then we know whether or not we need a token. If we don't,
             // just redirect.
-            if (!cluster.useToken) {
+            if (cluster.useToken === false) {
```
knrt10, line 130: How about just?

```suggestion
            if (useToken === false) {
```

knrt10 (changes requested):
Some suggestions. Feel free to resolve if it does'nt seems right

frontend/src/lib/k8s/apiProxy.ts
joaquimrocha: I use periods for comments that are longer than one line. One single line is optional.

frontend/src/components/authchooser/index.tsx
joaquimrocha: This is deliberate. I want to make sure we're analysing the `cluster`'s property and not the variable. This is usually safer code in this context, as if later we don't assign the `useToken` directly, we still rely on what the cluster's property is.

PR #152, opened 2021-01-07
Update backend target to produce a server excutable when on windows

ashu8912:
# How to use
package the app on windows and try running the desktop app 
earlier it was only showing a infinite spinner because of server not running, this is fixed now.
# Testing done
yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,16 @@
 GO111MODULE=on
 export GO111MODULE
 
+SERVER_EXE_EXT ?=
 DOCKER_CMD ?= docker
 DOCKER_REPO ?= docker.io/joaquim
 DOCKER_IMAGE_NAME ?= headlamp
 DOCKER_IMAGE_VERSION ?= $(shell git describe --tags --always --dirty)
 DOCKER_IMAGE_BASE ?= alpine:3.11.3
 
+ifeq ($(OS), Windows_NT)
+	SERVER_EXE_EXT = .exe
+endif
 all: backend frontend
 
 tools/golangci-lint: backend/go.mod backend/go.sum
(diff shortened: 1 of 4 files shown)


app/electron/main.js
```diff
@@ -127,6 +128,9 @@ function createWindow () {
   let serverFilePath;
   if (!isDev) {
     serverFilePath = path.join(process.resourcesPath, './electron/server');
+    if (isWin) {
+      execSync(`ren ${serverFilePath} server.exe`)
+    }
```
joaquimrocha, line 133: I don't think this is the place for this. We'd be running this as an external command and on every run of the app.
Ideally we should compile the server to have that extension on Windows already (as it seems to be needed), so it's something I'd check to see how it could be added in the backend, not the electron app.

joaquimrocha (changes requested):
Left a comment.
(There's also a typo in the commit title, but I think that after addressing my comment, the whole commit title will be different)

Makefile
```diff
@@ -27,8 +27,12 @@ app-mac:
 
 .PHONY: backend
 backend:
+ifeq ($(OS), Windows_NT)
```
joaquimrocha, line 30: On Windows maybe we get more than the `Windows_NT` value.
In fact, I have read that the `OS` may be available on Windows prior to Windows 7 (i.e. XP, NT). We only have to support >=7 AFAIU.
Have you checked maybe the `OSTYPE` variable which also exists on Linux?

A way to simplify the code (as to not repeat the whole command below) is to create a `SERVER_EXEC_EXT` variable, and assign it to .`exe` on Windows + leaving it empty otherwise. Then you use just one command with `... go build -o ./server${SERVER_EXE_EXT} ./cmd`.

joaquimrocha (changes requested):
The approach looks good in general. I left a comment that may improve it a bit more.
Also, you can squash the gitignore commit with this one as they're very related, but I think you're missing some changes to the electron app so it calls the new server exec name (or does it do it automagically?).

joaquimrocha (approved):

PR #154, opened 2021-01-11
Add headless mode

ashu8912:
Adds headless mode to the desktop app which allows user to open the app in there default browser
# How to use
Run the headlamp desktop app with headless flag
example:
On linux use 
./Headlamp.AppImage headless to start the desktop app in headless mode(opening in your default browser)

# Testing done
Yes

### DIFF
--- a/app/electron/main.js
+++ b/app/electron/main.js
@@ -4,12 +4,14 @@ const path = require('path');
 const url = require('url');
 const { autoUpdater } = require('electron-updater');
 const log = require('electron-log');
+const open = require('open')
 
 const yargs = require('yargs');
 const args = yargs.option('headless', {
   describe: 'Open Headlamp in the default web browser instead of its app window'
 }).parse();
 const isHeadlessMode = args.headless;
+const defaultPort = 4466;
 
 function startServer(flags = []) {
   const serverFilePath = path.join(process.resourcesPath, './electron/server');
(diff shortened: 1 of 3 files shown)


app/electron/main.js
```diff
@@ -176,14 +193,9 @@ app.on('quit', function () {
...
-    createWindow();
-  }
-});
-
-app.once('window-all-closed', app.quit);
-
-app.once('before-quit', () => {
-  mainWindow.removeAllListeners('close');
-});
+const isHeadlessMode = process.argv[1] === 'headless';
```
joaquimrocha, line 197: Please use `yargs` for this. Let's give it a nice description and have the base so we can add other options later (like the `port`).

app/electron/main.js
```diff
@@ -196,6 +197,9 @@ app.on('quit', function () {
 const isHeadlessMode = process.argv[1] === 'headless';
 if (isHeadlessMode) {
   serverProcess = startServer(['-html-static-dir', path.join(process.resourcesPath, './build')]);
+  (async () => {
+    await open('http://localhost:4466')
```
joaquimrocha, line 201: Please keep the default port in a const variable and then use it here. It will be also useful for when we add the port option.

Another thing: this commit's title mentions that it adds the open package but that's not the main goal of this commit, so it's a better message to write: `desktop: Open the system's defined browser in headless mode` (you already mention the new dependency in the commit's body so it's not needed to repeat in the title).

app/electron/main.js
joaquimrocha: We also want to use it as an option (`--headless`), not a simple argument.

joaquimrocha (changes requested):

joaquimrocha (changes requested):
There are some improvements to be done in the use of `yargs`, but I did try it and it worked very well. Thanks!

PR #155, opened 2021-01-11
App server logging

illume:
# App server logging

Log errors when something goes wrong executing or running the server process in the app.
With this PR the app logs stderr, and stdout from the `server` process.
Additionally the signal, and return code of the `server` process are logged.

related: https://github.com/kinvolk/headlamp/issues/112

What's not in this PR:
- updating the UI when there's a `server` related error.
- restarting or retrying server process start

Should we be logging the server.stderr URLs it prints out?

# How to use

1. log when server process crashes or is killed.

```bash
make app-mac
```

run with logging enabled.
```bash
app/dist/mac/Headlamp.app/Contents/MacOS/Headlamp --enable-logging
```

In a separate terminal you can kill the `server` process.
```bash
while true; do killall server ; done
```

Now you should see an error log that the process terminated.

Screenshot 2021-01-11 at 09 55 56

2. When running, and quitting without the server process it should not log an error.

Screenshot 2021-01-11 at 09 56 49

3. remove the server process.

To replicate an issue executing the server process, remove the file.

```bash
rm app/dist/mac/Headlamp.app/Contents//Resources/electron/server
```

Then an error should be logged.

Screenshot 2021-01-11 at 09 45 52

# Testing done

See above on the testing done on Mac.

### DIFF
--- a/app/electron/main.js
+++ b/app/electron/main.js
@@ -123,11 +123,47 @@ const isDev = process.mainModule.filename.indexOf('app.asar') === -1 ||
     process.mainModule.filename.indexOf('app') === -1;
 
 let serverProcess = null;
+let intentionalQuit = false;
+let serverProcessQuit = false;
+
+/**
+ * add some error handlers to the serverProcess.
+ * @param  {ChildProcess} serverProcess to attach the error handlers to.
+ */
+function attachServerEventHandlers (serverProcess) {
+  serverProcess.on('error', (err) => {
+    log.error(`server process failed to start: ${err}`);
+  });
+  serverProcess.stdout.on('data', (data) => {
+    log.info(`server process stdout: ${data}`);
+  });
+  serverProcess.stderr.on('data', (data) => {
+    const sterrMessage = `server process stderr: ${data}`;
+    if (data && data.indexOf && data.indexOf("Requesting") !== -1) {
+      // The server prints out urls it's getting, which aren't errors.
+      log.info(sterrMessage);
+    } else {
+      log.error(sterrMessage);
+    }
+  });
+  serverProcess.on('close', (code, signal) => {
+    const closeMessage = `server process process exited with code:${code} signal:${signal}`;
+    if (!intentionalQuit) {
+      // @todo: message mainWindow, or loadURL to an error url?
+      log.error(closeMessage);
+    } else {
+      log.info(closeMessage);
+    }
+    serverProcessQuit = true;
+  });
+}
+
 function createWindow () {
   let serverFilePath;
   if (!isDev) {
     serverFilePath = path.join(process.resourcesPath, './electron/server');
     serverProcess = spawn(serverFilePath, {shell: true});
+    attachServerEventHandlers(serverProcess);
   }
 
   const startUrl = process.env.ELECTRON_START_URL || url.format({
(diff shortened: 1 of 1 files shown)


app/electron/main.js
```diff
@@ -123,11 +123,37 @@ const isDev = process.mainModule.filename.indexOf('app.asar') === -1 ||
...
+    serverProcess.on('close', (code, signal) => {
+      const closeMessage = `server process process exited with code:${code} signal:${signal}`;
+      if (!intentionalQuit) {
+        //TODO: message mainWindow, or loadURL to an error url?
+        log.error(closeMessage);
+      } else {
+        log.info(closeMessage);
+      }
+      serverProcessQuit = true;
+    });
```
ashu8912, line 159: @illume can we move these event handlers on serverProcess into a seperate function maybe something like ```attachServerEventHandlers```(which resides in the global scope of main.js) because otherwise than in the headless mode feature  https://github.com/kinvolk/headlamp/pull/154 I will have to copy the entire server handlers again

illume: Good idea. Thanks @ashu8912 

I pushed a commit for this.

app/electron/main.js
```diff
@@ -170,8 +196,16 @@ app.on('ready', createWindow);
...
   if (serverProcess) {
-    serverProcess.stdin.pause();
-    serverProcess.kill();
+    if (serverProcessQuit) {
+      log.error('server process already not running');
+    } else {
+      intentionalQuit = true;
+      log.info('stopping server process...');
+      serverProcess.stdin.pause();
+      // TODO: should we try and end the process a bit more gracefully?
```
joaquimrocha, line 215: We've been using `// @todo: `. Not too important but maybe we keep it consistent so it's easy to find in the future.
Maybe the phrasing could also be styled as a todo instead of a question.

app/electron/main.js
```diff
@@ -126,37 +126,44 @@ let serverProcess = null;
 let intentionalQuit = false;
 let serverProcessQuit = false;
 
+/**
+ * add some error handlers to the serverProcess.
+ * @param  {ChildProcess} serverProcess to attach the error handlers to.
+ */
+const attachServerEventHandlers = (serverProcess) => {
```
joaquimrocha, line 133: We've been using `function ...` instead of the arrow functions.
Maybe we should move to arrows functions but for consistency for now, let's keep the functions.

joaquimrocha (changes requested):

app/electron/main.js
illume: Ok, thanks. It's done.

app/electron/main.js
illume: Ok, thanks. It's done.

app/electron/main.js
```diff
@@ -205,7 +212,7 @@ app.on('quit', function () {
       intentionalQuit = true;
       log.info('stopping server process...');
       serverProcess.stdin.pause();
-      // TODO: should we try and end the process a bit more gracefully?
+      // @todo: should we try and end the process a bit more gracefully?
```
joaquimrocha, line 215: This doesn't belong in this commit. Please add it to the commit that introduced the todo.

joaquimrocha (changes requested):
Left just a comment in the code. Also please remove the periods on commit titles.

app/electron/main.js
illume: I squished the commits into one now. Can you please check it again?

illume:
I have one open question left...
> Should we be logging the server.stderr URLs it prints out?

Here's an example:

Screenshot 2021-01-11 at 13 57 34

joaquimrocha:
> I have one open question left...
> 
> > Should we be logging the server.stderr URLs it prints out?
> 
> Here's an example:
> 
> Screenshot 2021-01-11 at 13 57 34

I think so. If it gets to be too much, we can silence it through levels in the future.

joaquimrocha (approved):

PR #163, opened 2021-01-12
Makefile: Improve the app related targets

joaquimrocha:
Instead of repeating the "npm run build" command in all the 3 app
related targets, we now create an app-build target for that, and also
run "npm install" in it, as it's a needed first step.


ashu8912 (approved):
Tested and LGTM!!

ashu8912:
The linter is failing for electron related code but i think @illume already pushed that fix in master.

PR #166, opened 2021-01-13
app: Update how we check dev mode for desktop app

ashu8912:
If we are in dev mode we set an env flag otherwise we assume we are in prod mode.

Testing Done 
Yes.


illume (approved):
👍 works for me

PR #168, opened 2021-01-15
Add  storybooks (PercentBar, AuthChooser, ConfirmButton, LogViewer) for frontend

illume:
# more pieces for the storybook

- Charts/PercentBar
- AuthChooser
- ConfirmButton
- LogViewer 

# How to use

```
cd frontend
npm run storybook
```


joaquimrocha (changes requested):
There are many style changes in unrelated commits. Some make sense to me (as it's conventions we use but somehow aren't yet added in the eslint file), but others seem more arbitrary to the rules we have.

If you have new formatting/style rules you want to propose, that's fine but let's discuss them and have them in a separate PR/commits.
Otherwise it's very difficult to review the actual changes this PR proposes, and the same for future ones if we don't use the same rules.

illume:
I reformatted the commits.

joaquimrocha (approved):

PR #170, opened 2021-01-18
Check for nil in context parsing

marians:
# Mitigate panic in backend/cmd/kubeconfig.go:52

Attempts to fix one of the problems causing https://github.com/kinvolk/headlamp/issues/112, especially the panic described in https://github.com/kinvolk/headlamp/issues/112#issuecomment-762148606

# How to use

1. Create a kubeconfig context referencing a user that does not exist

```yaml
...
contexts:
- context:
    cluster: my-existing-cluster
    user: non-existing
  name: my-context
...
```

2. Execute the backend

# Testing done

```nohighlight
$ make backend
cd backend && go build -o ./server ./cmd

$ backend/server

2021/01/18 17:42:18 Not adding context my-context because authInfo is nil!
*** Headlamp Server ***
  API Routers:
	...
```

### DIFF
--- a/backend/cmd/kubeconfig.go
+++ b/backend/cmd/kubeconfig.go
@@ -49,6 +49,11 @@ func GetContextsFromKubeConfigFile(kubeConfigPath string) ([]Context, error) {
 		authInfo := config.AuthInfos[value.AuthInfo]
 		authType := ""
 
+		if authInfo == nil {
+			log.Printf("Not adding context %v because authInfo is nil!\n", key)
+			continue
+		}
+
 		authProvider := authInfo.AuthProvider
 		if authProvider != nil {
 			authType = "oidc"


backend/cmd/kubeconfig.go
```diff
@@ -49,6 +49,11 @@ func GetContextsFromKubeConfigFile(kubeConfigPath string) ([]Context, error) {
 		authInfo := config.AuthInfos[value.AuthInfo]
 		authType := ""
 
+		if authInfo == nil {
+			log.Printf("Not adding context %v because authInfo is nil!\n", key)
+			continue
+		}
+
```
joaquimrocha, line 56: Although this solution bypasses the issue, I think it's worth checking the `value.AuthInfo` before as well.
I am not sure if an empty `AuthInfo` in the `Context` struct is possible (likely it is), if it's empty, maybe we still add the context as later on we're checking for the validity of `AuthInfo` when using it.
If it's not empty but the AuthInfo in question is not found in the list (an invalid kubeconfig, I guess), then we do skip it.

In order words: let's see if using this logic *only* when the `value.AuthInfo != ""` makes sense.

joaquimrocha (commented):
Left a comment. @illume , maybe you can check whether it makes sense. I'm not marking as requesting changes because I am not sure yet if what I am asking for is possible.

illume:
Cool, thanks! I'll have a look tomorrow morning :)

backend/cmd/kubeconfig.go
illume: I might be misunderstanding you (or be bad at go)... but I think the PR is ok as is. Because...

I tried modifying the kube config in a few ways like suggested and could reproduce the problem, and found the PR fixes the problem.

There's two cases to test:
1. If the `user: ` key is missing from the yaml config, then the value.AuthInfo == "". 
2. if the `user: ` is not a user in `config.AuthInfos`.

It's possible to add a `if value.AuthInfo == ""` check like the code below, but I think it's redundant. clusterConfig has the same issue (if `cluster:` is missing) and the nil check handles that missing key, and non-existant value case too. This is another reason why I think the error handling should stay as it is - it more closely matches clusterConfig handling.

```golang
		if value.AuthInfo == "" {
			log.Printf("Not adding context %v because AuthInfo is empty!\n", key)
			continue
		}
		authInfo := config.AuthInfos[value.AuthInfo]
		authType := ""

		if authInfo == nil {
			log.Printf("Not adding context %v because authInfo is nil!\n", key);
			continue
		}
```

illume: @marians what do you think about changing the log to this?

```golang
log.Printf("Not adding context %v because user:%v: could not be found!\n", key, value.AuthInfo);
```

illume (approved):
👍

backend/cmd/kubeconfig.go
illume: > I am not sure if an empty AuthInfo in the Context struct is possible

@joaquimrocha I tried removing the user, and context parts of the context, and it was still there. Maybe it's possible... but I couldn't find out how by mangling the config file :)

illume: I think @joaquimrocha is interested in the case where there is no authInfo, but we want to let it try to connect to the cluster anyway.

I'm no expert, but I did some research, and couldn't find a way to do no-authentication authentication with Kubernetes.

The code would be something like below if the no-authentication use case is to be supported. Note, it doesn't crash... for me. The user gets a nice error message inside Headlamp saying they can't authenticate.

```golang
	for key, value := range config.Contexts {
		clusterConfig := config.Clusters[value.Cluster]
		if clusterConfig == nil {
			log.Printf("Not adding context %v because cluster doesn't exist!\n", key)
			continue
		}

		authInfo := config.AuthInfos[value.AuthInfo]
		authType := ""

		if authInfo == nil && value.AuthInfo != ""  {
			log.Printf("Not adding context: %v because user: %v could not be found!\n", key, value.AuthInfo)
			continue
		}

		if authInfo != nil {
			authProvider := authInfo.AuthProvider
			if authProvider != nil {
				authType = "oidc"

				var oidcConfig OidcConfig
				oidcConfig.ClientID = authProvider.Config["client-id"]
				oidcConfig.ClientSecret = authProvider.Config["client-secret"]
				oidcConfig.IdpIssuerURL = authProvider.Config["idp-issuer-url"]

				oidcConfigCache[key] = &oidcConfig
			}
		}

		cluster := Cluster{key, clusterConfig.Server, clusterConfig, authType}

		contexts = append(contexts, Context{key, cluster, authInfo})
	}
```

joaquimrocha: Thanks @illume . I do prefer to have the context show up in the UI and the message saying it cannot authenticate. This is a good hint for the user to look better at the configuration (until we have a better way to add contexts from a kube config file).

Let's integrate this solution + log a message as well when no auth context is found.

joaquimrocha: @marians Let us know if you want to update the PR or whether you prefer us to take it from here. Either way, it's been great to have you help us find a solution for this issue!

marians: Please take over. Thanks

PR #172, opened 2021-01-19
Port app code to typescript

ashu8912:


app/electron/main.ts
```diff
@@ -1,33 +1,33 @@
...
+import { MenuItemConstructorOptions } from 'electron/main';
+
 const args = yargs.option('headless', {
   describe: 'Open Headlamp in the default web browser instead of its app window'
 }).parse();
 const isHeadlessMode = args.headless;
 const defaultPort = 4466;
 
-function startServer(flags = []) {
+function startServer(flags:string[] = []):ChildProcessWithoutNullStreams {
```
joaquimrocha, line 17: Add a space after the `:`. Same for the other places where this is occurring.

app/electron/main.ts
```diff
@@ -1,33 +1,33 @@
...
 
-let serverProcess = null;
-let intentionalQuit = false;
-let serverProcessQuit = false;
+let serverProcess: ChildProcessWithoutNullStreams | null;
+let intentionalQuit:boolean;
+let serverProcessQuit:boolean;
 
 function startElecron() {
-  autoUpdater.logger = log;
```
joaquimrocha, line 26: This seems to be changed functionality. Do we get the logs reported as before?

app/electron/main.ts
```diff
@@ -202,7 +205,7 @@ app.on('quit', function () {
     } else {
       intentionalQuit = true;
       log.info('stopping server process...');
-      serverProcess.stdin.pause();
+      serverProcess.stdin.destroy();
```
joaquimrocha, line 208: Can you explain this change? Doesn't seem related to the port.

app/package.json
```diff
@@ -2,15 +2,16 @@
   "name": "headlamp",
   "version": "0.2.0",
   "description": "",
-  "main": "electron/main.js",
+  "main": "electron/src/main.js",
```
joaquimrocha, line 5: We don't have a `main` defined in the frontend. Not sure if we need it here either (it's fine if we do though).
Please notice that below you still have places showing `build/electron/main.js`.

joaquimrocha (changes requested):
Left some comments in the code. Also, please squash the commits (adding typescript + babel without adding the related code changes doesn't give us an advantage in this case).

illume:
Ya TypeScript in the app. Thanks :)

It runs for me locally. I'll review.

app/package.json
illume: Perhaps just change it to electron/src/main.ts ? (note the ts). It's kind of a nice documentation for the entry point.

https://docs.npmjs.com/cli/v6/configuring-npm/package-json#main

app/electron/main.ts
```diff
@@ -1,33 +1,33 @@
...
+let serverProcessQuit:boolean;
 
 function startElecron() {
-  autoUpdater.logger = log;
-  autoUpdater.logger.transports.file.level = 'info';
+  log.transports.file.level = 'info';
   log.info('App starting...');
 
-  let mainWindow;
+  let mainWindow: BrowserWindow | null;
```
illume, line 30: I think this should be `BrowserWindow | undefined` rather than null?

app/electron/main.ts
ashu8912: it was an unnecessary step we just want to access transports from log see the updated code

app/electron/main.ts
ashu8912: so no pause method exists on stdin(type Writable)(typescript helped to get this) we want to use destory instead

app/electron/main.ts
ashu8912: I think for objects, having null type makes more sense

app/package.json
ashu8912: main entry tells the electron cli where to look for the entry file for electron codebase and that's why i changed it to ```electron/src/main.js``` because this is where we have the compiled js code from typescript.
about ```build/electron/main.js``` that is used by electron-builder and its a place where we copy the electron codebase while the build step so it doesn't require the change.

joaquimrocha (approved):

PR #179, opened 2021-01-22
Check for nil and empty username in context parsing (#170)

illume:
# Mitigate panic in backend/cmd/kubeconfig.go:52

Continuing on from the PR https://github.com/kinvolk/headlamp/pull/170 related to https://github.com/kinvolk/headlamp/issues/112#issuecomment-762148606

There was a segfault in the backend when the user was not there, or not existing.

# How to use

1. Create a kubeconfig context referencing a user that does not exist

```yaml
...
contexts:
- context:
    cluster: my-existing-cluster
    user: non-existing
  name: my-context
...
```

Another case is where there is no user mentioned at all.

```yaml
...
contexts:
- context:
    cluster: my-existing-cluster
  name: my-context
...
```

2. Execute the backend

# Testing done

```nohighlight
$ make backend run-backend

2021/01/21 15:09:48 Not adding context: my-context because user:non-existing could not be found!
*** Headlamp Server ***
  API Routers:
	...
```

It shows this for when the user is not there, and anonymous users are not allowed auth.

Screenshot 2021-01-22 at 18 18 36

### DIFF
--- a/backend/cmd/kubeconfig.go
+++ b/backend/cmd/kubeconfig.go
@@ -49,16 +49,23 @@ func GetContextsFromKubeConfigFile(kubeConfigPath string) ([]Context, error) {
 		authInfo := config.AuthInfos[value.AuthInfo]
 		authType := ""
 
-		authProvider := authInfo.AuthProvider
-		if authProvider != nil {
-			authType = "oidc"
+		if authInfo == nil && value.AuthInfo != "" {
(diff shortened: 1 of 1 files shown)


ashu8912 (approved):
Tested & LGTM!!!
Thankyou @illume and @marians

joaquimrocha (approved):

PR #181, opened 2021-01-22
Support relative certificate-authority paths in config

illume:
# certificate-authority paths can be relative to config

for https://github.com/kinvolk/headlamp/issues/180

Where relative paths are also supported in the kubectl config.

related https://github.com/kinvolk/headlamp/issues/112

# How to use

In .kube/config we add a relative path.

```yaml
- cluster:
    certificate-authority: ca.crt
```

# Testing done

Tried a relative path "ca.crt", and an absolute path "/Users/rene/.kube/ca.crt", and it loads the correct file.

### DIFF
--- a/backend/cmd/cluster.go
+++ b/backend/cmd/cluster.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"path"
+	"path/filepath"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


joaquimrocha (changes requested):
The code looks fine but the commit message should be improved for consistency and ease of understanding it in the future without having to check the code.

We try to use git commit titles that describe in the imperative mood:
https://chris.beams.io/posts/git-commit/#imperative

And it's worth adding an explanation of what + why this commit is added.

illume:
Updated the commit message

joaquimrocha (approved):

PR #183, opened 2021-01-26
Add login screen back for OIDC configurations

illume:
So that the login screen is shown again as it was in v0.1.3

See https://github.com/kinvolk/headlamp/issues/182

It would be worthwhile to refactor this code so that the states are clear and tested (as started in https://github.com/kinvolk/headlamp/pull/168 ) but this is not done in this PR.

# How to use

Login against an OIDC configuration.
See that there is a choice between Login, and using a token.

Screenshot 2021-01-26 at 16 33 48

# Testing done

Before it was going to the enter token state.
Now it's showing a choice.

### DIFF
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -80,8 +80,10 @@ function AuthChooser(){
     let cancelled = false;
 
     // If we haven't yet figured whether we need to use a token for the current
-    // cluster, then we check here.
-    if (cluster.useToken === undefined) {
+    //   cluster, then we check here.
+    // With clusterAuthType == oidc,
+    //   they are presented with a choice of login or enter token.
+    if (clusterAuthType !== 'oidc' && cluster.useToken === undefined) {
       let useToken = true;
 
       setTestingAuth(true);


frontend/src/components/authchooser/index.tsx
```diff
@@ -81,7 +81,9 @@ function AuthChooser(){
 
     // If we haven't yet figured whether we need to use a token for the current
     // cluster, then we check here.
-    if (cluster.useToken === undefined) {
+    // oidc should first see if they want to try and login.
+    const noLoginCheck = (clusterAuthType === 'oidc' && testingAuth || clusterAuthType !== 'oidc');
```
joaquimrocha, line 85: Shouldn't we simply check for `clusterAuthType !== 'oidc'`?
Or (probably better) just not redirect to the token prompt within the next if block when the clusterAuthType is `oidc`?

The overall problem seems to be that this logic was for checking whether we use a certificate (i.e. no token required) or need a token, but the oidc option has been forgotten here. Redirecting to the token directly is what we need iff the token is the only auth option (i.e. we don't want to show a dialog with just one option to the user).

illume: Yes, you're right. I changed it to just do: `if (clusterAuthType !== 'oidc' && cluster.useToken === undefined)`.

illume: Both clicking the signin button, and also entering a token worked with this change.

joaquimrocha (approved):
Thanks! This change seems easier to understand.

One thing I forgot to comment on though: the commit title and body are acceptable but could be improved.
The title is okay but the situation is not so trivial that by reading this commit body in 1 month I'd understand what exactly the issue was (or I'd have to run v0.1.3 to understand what's going on, or try to find the PR).

Hope this makes sense. Approving and leaving the commit body improvement to your criteria.

illume:
I've improved the git commit message to describe the issue some more.

PR #186, opened 2021-01-27
Add prettier code formatting tool for frontend

### LINKED ISSUE #185: Add prettier
**Current situation**
We're relying on eslint for the coding style, and stule issues are also caught by CI, but in order to reduce some back and forth, we should add prettier support.

**Impact**
Users will have autoomatic formating of code without having to run `make frontend-fixlint`.

illume:
Based on config from create react app prettier docs.
https://create-react-app.dev/docs/setting-up-your-editor/#formatting-code-automatically

Files that are git staged get eslint and prettier run on them.
This is done via lint-staged, and husky with git hooks.

`npm run format` formats code with prettier.

We use eslint-config-prettier to smooth integration with eslint.

Both eslint, and prettier:

- use @kinvolk/eslint-config 0.3.0
- are configured by package.json.

# How to use

Files that are git staged get prettier run on them.

See [Editor integration](https://prettier.io/docs/en/editors.html) section of prettier docs.

# Testing done

- `npm run format`
- `npm run lint`
- `vi src/App.tsx && git commit .` # should show eslint and prettier being run with git pre-commit hook.

See the new `@kinvolk/eslint-config` shared config PR here: https://github.com/kinvolk/eslint-config/pull/3 (which has instructions on how to test).

fixes #185

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,14 +19,17 @@
     "@types/semver": "^7.3.4",
     "@types/webpack-env": "^1.16.0",
     "ansi-to-react": "^6.1.4",
+    "husky": "^4.3.8",
     "javascript-time-ago": "^2.3.4",
     "js-base64": "^3.6.0",
     "js-cookie": "^2.2.1",
     "js-yaml": "^3.14.0",
+    "lint-staged": "^10.5.4",
     "lodash": "^4.17.20",
     "monaco-editor": "^0.21.2",
     "notistack": "^1.0.3",
     "openapi-types": "^7.2.3",
+    "prettier": "^2.2.1",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-hotkeys-hook": "^3.0.3",
@@ -45,7 +48,8 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint -c .eslintrc.yml --ext .js,.ts,.tsx src/ ../app/electron",
+    "lint": "eslint -c package.json --ext .js,.ts,.tsx src/ ../app/electron",
+    "format": "prettier --config package.json --write src/ ../app/electron",
     "storybook": "start-storybook -p 6006 -s public",
     "build-storybook": "build-storybook -s public"
   },
@@ -61,6 +65,27 @@
       "last 1 safari version"
     ]
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "src/**/*.{js,jsx,ts,tsx}": [
+      "eslint -c package.json --fix"
+    ],
+    "src/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
+      "prettier --config package.json --write"
+    ]
+  },
+  "eslintConfig": {
+    "extends": [
+      "@kinvolk",
+      "prettier",
+      "prettier/react"
+    ]
+  },
+  "prettier": "@kinvolk/eslint-config/prettier-config",
   "devDependencies": {
     "@iconify/icons-mdi": "^1.1.10",
     "@iconify/react": "^1.1.3",
@@ -76,6 +101,7 @@
     "@typescript-eslint/eslint-plugin": "^4.14.2",
     "@typescript-eslint/parser": "^4.14.2",
     "eslint": "^7.19.0",
+    "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -8,15 +8,17 @@ import path from 'path';
 import url from 'url';
 import yargs from 'yargs';
 
-const args = yargs.option('headless', {
-  describe: 'Open Headlamp in the default web browser instead of its app window'
-}).parse();
+const args = yargs
+  .option('headless', {
+    describe: 'Open Headlamp in the default web browser instead of its app window',
+  })
+  .parse();
 const isHeadlessMode = args.headless;
 const defaultPort = 4466;
 
 function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
   const serverFilePath = path.join(process.resourcesPath, './electron/server');
-  return spawn(serverFilePath, [...flags], {shell: true});
+  return spawn(serverFilePath, [...flags], { shell: true });
 }
 
 let serverProcess: ChildProcessWithoutNullStreams | null;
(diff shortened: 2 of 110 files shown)


frontend/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true,
```
joaquimrocha, line 2: This is not the only thing we'd need prettier to change in order to keep most of the rules we have.
Running prettier I see lots of changes with one especially that I'd like it to change: it's enforcing 80 chars but we've been enforcing 100.

Can you check whether this is possible? @ashu8912 had mentioned that prettier could reuse the eslint rules we've defined.

illume:
There's some prettier documentation here on [Integrating with Linters](https://prettier.io/docs/en/integrating-with-linters.html), in which they recommend [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier).

It also comes with a script for checking conflicting rules. Which finds this conflict.
```bash
npx eslint-config-prettier `find src -type f -name \* | grep ".ts\|.js\|.tsx"`
The following rules are unnecessary or might conflict with Prettier:

- @typescript-eslint/member-delimiter-style
```

The docs for this rule: [@typescript-eslint/member-delimiter-style](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-delimiter-style.md)

joaquimrocha:
AFAICS running prettier didn't change the members' delimiter, so we can delete that rule if that's correct and it's the only thing impeding prettier to use our eslint rules.

illume:
To make sure eslint would not fix prettier changes, I tried this:
```
$ ./node_modules/.bin/prettier --write src
# 2800 or so changed lines
$ git commit . -m "prettier changes"
$ yarn lint --fix src
$ git status
On branch prettier
nothing to commit, working tree clean
```

We can either 
1. make one big code formatting commit
2. do it as we go, and only format files files as they change

Probably 2 is the safest option, with less noise too. If we went with 1 big commit, there's some trick you can do to get git blame working again.

illume:
> AFAICS running prettier didn't change the members' delimiter, so we can delete that rule if that's correct and it's the only thing impeding prettier to use our eslint rules.

Yeah, true. It does seem prettier leaves those delimiters alone. So if we stick with using `;`, then prettier should not rewrite them.

illume:
Why not add the [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) now? 

> "Runs Prettier as an ESLint rule and reports differences as individual ESLint issues."

This would mean we'd have to apply those 2800 lines of changes so `make lint` would be happy.

frontend/.prettierrc
```diff
@@ -0,0 +1,4 @@
...
+  "printWidth": 100
```
illume, line 3: The printWidth is slightly different to the eslint rule. It's the preferred length, vs a max limit with eslint. More details here: https://prettier.io/docs/en/options.html#print-width

ashu8912: > To make sure eslint would not fix prettier changes, I tried this:
> 
> ```
> $ ./node_modules/.bin/prettier --write src
> # 2800 or so changed lines
> $ git commit . -m "prettier changes"
> $ yarn lint --fix src
> $ git status
> On branch prettier
> nothing to commit, working tree clean
> ```
> 
> We can either
> 
> 1. make one big code formatting commit
> 2. do it as we go, and only format files as they change
> 
> Probably 2 is the safest option, with less noise too. If we went with 1 big commit, there's some trick you can do to get git blame working again.

I am not sure whether step 2 will be good here because otherwise than each time somebody is doing a different pr it will bring in the noise of formatting there. It's best if we have the formatting done here in a separate commit.
let's see what @joaquimrocha has to say.

illume: Yes, I think joaquimrocha has this view too. Let's go with 1. make one big code formatting commit.

illume:
I made `estlint --fix` run in the lint-staged, and added a `npm run format` command.

I was going to add the app/ config, which is a duplicate. Then I considered, that we'd need to use this config in each app going forward. We can do this with a [shared configuration](https://prettier.io/docs/en/configuration.html#sharing-configurations) in a package like a `@kinvolk/prettier-config`.

However, what if we had no prettier configuration and left all style decisions up to prettier?

The benefits of leaving all style decisions up to prettier are:
- no prettier config needed for each app
- code is closer to 'standard' JS ecosystem style

The downsides are:
- there are more changes, because of the single quote.

```
$ git diff | wc -l
   14772
```

- a) go with prettier for all style decisions, and no config.
- b) keep single quotes, and add a shared `@kinvolk/prettier-config` package.

I would probably go for a), because it's a one time cost.

illume:
From chat about prettier config...

- Material UI: https://github.com/mui-org/material-ui/blob/next/prettier.config.js
- CRA: https://github.com/facebook/create-react-app/blob/master/.prettierrc
- react too: https://github.com/facebook/react/blob/master/.prettierrc.js
- redux: https://github.com/reduxjs/redux/blob/master/.prettierrc.json

> so I think we should really check again how to minimize the changes between prettier + eslint
> and if we can have a @kinvolk/prettier or whatever to share the config as we do for eslint, let's do that (edited)

frontend/package.json
```diff
@@ -46,6 +50,7 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint -c .eslintrc.yml --ext .js,.ts,.tsx src/ ../app/electron",
+    "format": "prettier --write src/ ../app/electron",
```
joaquimrocha, line 53: Just running `npm run format` wasn't picking up the config.

```suggestion
    "format": "prettier --config .prettierrc --write src/ ../app/electron",
```

frontend/package.json
```diff
@@ -61,6 +66,19 @@
...
     ]
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "src/**/*.{js,jsx,ts,tsx}": [
+      "lint --fix"
```
joaquimrocha, line 76: This one also didn't work here because there's no lint executable. Does this run in yours?

```suggestion
      "npm run lint -- --fix"
```

joaquimrocha (changes requested):
Please check my comments, suggestions, and squash them with your commits if you agree with them.
I have also added a commit with the formatting.

Please create the shared package so we can start applying these to Nebraska as well.

illume:
There's a shared config PR here: kinvolk/eslint-config#3 which needs to be merged/published first (testing instructions in there).

illume:
I'm working on dependency upgrades on headlamp and kinvolk/eslint-config first, so prettier/@typescript-eslint can work (which requires a newer dependency).

joaquimrocha (approved):

illume:
The kinvolk/eslint-config 0.4.0 PR https://github.com/kinvolk/eslint-config/pull/4, then the dependencies update PR https://github.com/kinvolk/headlamp/pull/190 should be merged before this one is done again.

PR #189, opened 2021-01-29
Add ability to run Headlamp with a base-url like "/headlamp"

illume:
# Add ability to run Headlamp with a base-url like "/headlamp"

So can use Headlamp at `example.com/headlamp/` rather than just at `headlamp.example.com`.

Also some documentation on how to use Headlamp with docker and minikube in development.

# How to use

### in dev mode
```
./backend/server -dev -base-url /headlamp
PUBLIC_URL="/headlamp" make run-frontend
```

### static build mode
```
cd frontend && npm install 
npm run buildbaseurl
cd ..
./backend/server -dev -base-url /headlamp -html-static-dir frontend/build
```
Then go to http://localhost:4466/headlamp/

### docker mode

Append the `--base-url /headlamp` to the docker run arguments.

### kubernetes

Build the docker image, and then modify your kubernetes deployment file to add `-base-url /headlamp` to the containers args.

```yaml
        args:
          - "-in-cluster"
          - "-base-url=/headlamp"
```

# Testing done

Tried running headlamp with a base-url 
- [x] in-cluster on a local minikube with a docker image
- [x] in dev mode
- [x] using a static html build
- [x] running in app mode on linux is not broken (but of course base-url in app mode makes no sense, so not tested)
- [x] loading plugins works with base-url and without base-url

It also checks the base-url argument starts with a '/'.
```
$ ./backend/server -dev -base-url broken/headlamp
2021/02/08 17:35:29 base-url needs to start with a '/'
```

Some tests were added. The tests can be run like this:
```
cd backend/cmd && go test -v
cd frontend && npm run test
```

The documentation on how to use Headlamp with docker and minikube in development added to this PR can also be used to test.

---

Related to https://github.com/kinvolk/headlamp/issues/150


backend/cmd/server.go
```diff
@@ -38,5 +39,6 @@ func main() {
 		oidcClientID:     *oidcClientID,
 		oidcClientSecret: *oidcClientSecret,
 		oidcIdpIssuerURL: *oidcIdpIssuerURL,
+		baseURL:          *baseURL,
```
joaquimrocha, line 42: Don't we need to make sure (in this block or another) that the base URL has a starting `/`?

joaquimrocha (changes requested):
AFAIU, the `process.env.PUBLIC_URL` will work when running the desktop version, or at build time, but not in-cluster because we run the static code (without node and locally). And I understood the user that filed this issue was running Headlamp in-cluster.
IIRC the way to solve this is to replace some var in the HTML that the backend serves, but I am not sure if there's another way.

Another comment, please add more context as to why we're adding this feature (at least in the first commit). You don't have to repeat the entire issue's description, but it's useful to have some more context about why this would be needed.

backend/cmd/server.go
illume: This validation has been added.

illume:
> AFAIU, the process.env.PUBLIC_URL will work when running the desktop version, or at build time, but not in-cluster because we run the static code (without node and locally). And I understood the user that filed this issue was running Headlamp in-cluster. IIRC the way to solve this is to replace some var in the HTML that the backend serves, but I am not sure if there's another way.

Ah, ok. I'm looking into this now.

.dockerignore
```diff
@@ -1 +1,6 @@
-frontend/node_modules
+**/node_modules
+.git
+app/dist
+app/build
```
joaquimrocha, line 4: Maybe we should ignore the whole app as it's not used for docker.

Makefile
```diff
@@ -3,7 +3,7 @@ export GO111MODULE
 
 SERVER_EXE_EXT ?=
 DOCKER_CMD ?= docker
-DOCKER_REPO ?= docker.io/joaquim
+DOCKER_REPO ?= docker.io/kinvolk
```
joaquimrocha, line 6: This was on my todo list :slightly_smiling_face: 
Can you please move it into a different commit though (let's keep things atomic)?

backend/cmd/headlamp.go
```diff
@@ -54,24 +59,74 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	path = filepath.Join(h.staticPath, path)
 
 	// check whether a file exists at the given path
 	_, err = os.Stat(path)
 	if os.IsNotExist(err) {
 		// file does not exist, serve index.html
-		http.ServeFile(w, r, filepath.Join(h.staticPath, h.indexPath))
+		staticPath := filepath.Join(h.staticPath, h.indexPath)
+		http.ServeFile(w, r, staticPath)
+
```
joaquimrocha, line 73: This change doesn't seem necessary/relevant to the commit.

backend/cmd/headlamp.go
```diff
@@ -54,24 +59,74 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		return
 	} else if err != nil {
 		// if we got an error (that wasn't that the file doesn't exist) stating the
 		// file, return a 500 internal server error and stop
 		http.Error(w, err.Error(), http.StatusInternalServerError)
+
```
joaquimrocha, line 79: Shouldn't be here.

docs/development/_index.md
```diff
@@ -54,6 +54,109 @@ options can be appended to the main command as arguments.
...
+
+$ docker run --network="host" -p 127.0.0.1:4466:4466/tcp --mount type=bind,source="/home/rene/.minikube",target=$HOME/.minikube --mount type=bind,source="$HOME/.kube",target=/root/.kube kinvolk/headlamp:development /headlamp/server -html-static-dir /headlamp/frontend -plugins-dir=/headlamp/plugins
+```
+
+Then go to https://localhost:4466
+
+
+### Minikube "incluster"
+
+This is if you want to use Headlamp incluster, and test it locally on minikube with a local docker image.
```
joaquimrocha, line 81: I think we've been using `in-cluster` throughout Headlamp.

docs/development/_index.md
```diff
@@ -54,6 +54,109 @@ options can be appended to the main command as arguments.
...
+
+#### docker image in the minikub docker environment
```
joaquimrocha, line 83: We should try not to refer docker too much and rather use "Container image".

docs/development/_index.md
```diff
@@ -54,6 +54,109 @@ options can be appended to the main command as arguments.
...
+    spec:
+      containers:
+      - command:
+        - /headlamp/server
+        - -html-static-dir
+        - /headlamp/frontend
+        - -in-cluster
+        - -plugins-dir=/headlamp/plugins
+        - -base-url
+        - /headlamp
```
joaquimrocha, line 129: This section doesn't seem to be released to the base-url but you're adding this option here.

docs/development/_index.md
```diff
@@ -54,6 +54,109 @@ options can be appended to the main command as arguments.
...
+
+$ kubectl get service headlamp -n kube-system
+NAME       TYPE       CLUSTER-IP      EXTERNAL-IP   PORT(S)          AGE
+headlamp   NodePort   10.99.144.210   <none>        4466:30712/TCP   6m57s
+
+$ minikube service headlamp -n kube-system --url
+http://192.168.49.2:30342
+```
+
+Go to the url printed, and get your token to login.
```
joaquimrocha, line 157: Please use URL when in text form.

docs/installation/base-url.md
```diff
@@ -0,0 +1,46 @@
+# Add ability to run Headlamp with a base-url like "/headlamp"
+
+So can use Headlamp at `http://example.com/headlamp` rather than at `http://headlamp.example.com/`.
```
joaquimrocha, line 3: I think this needs a bit more context as the example `headlamp.example.com` may lead to doubts IMO.

docs/installation/base-url.md
```diff
@@ -0,0 +1,46 @@
+# Add ability to run Headlamp with a base-url like "/headlamp"
```
joaquimrocha, line 1: Remove the the example from the section title

frontend/src/components/authchooser/index.tsx
```diff
@@ -152,7 +152,9 @@ function AuthChooser() {
       haveClusters={!!clusters && Object.keys(clusters).length > 1}
       error={error}
       oauthUrl={`${
-        isDevMode || isElectron() ? 'http://localhost:4466/' : '/'
+        isDevMode || isElectron()
+          ? 'http://localhost:4466' + (getBaseUrl() ? getBaseUrl() : '/')
```
joaquimrocha, line 156: I think it should work with two `//` in the URL path. Otherwise let's change it so we don't run the `getBaseUrl()` twice.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -7,15 +7,19 @@
...
 import { getCluster } from '../util';
 import { ResourceClasses } from '.';
 import { KubeMetrics, KubeObjectClass, KubeObjectInterface } from './cluster';
 
 const isDev = process.env.NODE_ENV !== 'production';
 const backendServicesURL =
-  isElectron() || isDev ? 'http://localhost:4466/' : window.location.origin;
+  isElectron() || isDev
+    ? 'http://localhost:4466' + (getBaseUrl() ? getBaseUrl() : '/')
+    : window.location.origin + getBaseUrl();
```
joaquimrocha, line 20: Same comments as above.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -7,15 +7,19 @@
...
+console.log('backendServicesURL', backendServicesURL);
```
joaquimrocha, line 21: Left over as debug?

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -58,10 +62,10 @@ export async function request(
   const id = setTimeout(() => controller.abort(), timeout);
 
   const url = combinePath(BASE_HTTP_URL, fullPath);
+  const requestData = { signal: controller.signal, ...opts };
   let response: Response = new Response(undefined, { status: 502, statusText: 'Unreachable' });
-
```
joaquimrocha, line 62: Should not be here (unless prettier did this :stuck_out_tongue: ).

docs/development/_index.md
```diff
@@ -54,6 +54,109 @@ options can be appended to the main command as arguments.
...
+
+#### Create a deployment yaml.
+
+```bash
+$ kubectl create deployment headlamp -n kube-system --image=kinvolk/headlamp:development -o yaml --dry-run -- /headlamp/server -html-static-dir /headlamp/frontend -in-cluster -plugins-dir=/headlamp/plugins > minikube-headlamp.yaml
+```
+
+To use the local docker image we change the imagePullPolicy to Never.
+
+```yaml
```
joaquimrocha, line 100: We have a Kubernetes YAML file, maybe we should reuse it here and just have users change the needed parts (maybe through `sed` as we also have in other docs). Otherwise we have to keep this up to date, not a big deal, just seems like a good idea IMO.

joaquimrocha (changes requested):
Thanks @illume . I still need to try this PR, gave it a quick review.
The same commit is adding different things though, so it should be separated in atomic ones (the documentation to run minikube, the base-url changes, the base-url docs, other fixes here and there).

.dockerignore
illume: This done.

Makefile
illume: This done.

backend/cmd/headlamp.go
illume: Moved it back to how it was.

docs/development/_index.md
illume: This is done.

docs/development/_index.md
illume: I changed it to Container image in several places.

docs/development/_index.md
illume: This is done.

docs/installation/base-url.md
illume: This is done.

docs/installation/base-url.md
illume: This is done.

frontend/src/components/authchooser/index.tsx
illume: I changed this around to use a getAppUrl() function and cleaned it up.

frontend/src/lib/k8s/apiProxy.ts
illume: This is done.

frontend/src/lib/k8s/apiProxy.ts
illume: Yop. Fixed, thanks.

frontend/src/lib/k8s/apiProxy.ts
illume: I guess it must have removed the line.

docs/development/_index.md
illume: I couldn't find a simple expression to add the imagePullPolicy line to the yaml in the right spot nicely without a complex confusing regex. Also that command isn't available as a command line argument. I'm not familiar with the documentation system, but some other ones can include from files/urls, but couldn't see how to do that either. With that, maybe including this yaml is the better way.

docs/development/_index.md
illume: Thanks. I removed this.

backend/cmd/headlamp.go
illume: The extra line? I think the golang lint tool asked for it.

illume:
> it should be separated in atomic ones

It's been separated into separate commits now.

As well, I've addressed all the notes... except two open ones that I'm unsure of.

illume:
@joaquimrocha I've added some tests for the base-url stuff in the frontend and backend.

```
cd backend/cmd && go test -v
cd frontend && npm run test
```

illume:
I fixed the new merge conflict.

docs/installation/base-url.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Add ability to run Headlamp with a base-url
```
joaquimrocha, line 2: This should be called: `Run Headlamp with a custom base URL` or something like that. Otherwise it looks like a PR :slightly_smiling_face: 
Screenshot from 2021-02-25 14-25-20

docs/installation/base-url.md
```diff
@@ -0,0 +1,55 @@
...
+cd ..
+./backend/server -dev -base-url /headlamp -html-static-dir frontend/build
+```
+
+Then go to http://localhost:4466/headlamp/ in your browser.
+
+
+### Docker mode
+
+Append `--base-url /headlamp` to the docker run command. Note the extra "-".
```
joaquimrocha, line 42: Does it really require the two dashes?

docs/development/_index.md
```diff
@@ -44,16 +44,161 @@ and in a different terminal instance:
...
+```
+
+```bash
+make app-mac
+```
+
+See the generated app files in app/dist/
+
+
+
```
joaquimrocha, line 77: Is it common to add 3 newlines? Usually we add two but I am not sure what's standard.

Makefile
joaquimrocha, line 6: Actually make this `quay.io/kinvolk`

joaquimrocha (changes requested):
Left some comments.

docs/installation/base-url.md
illume: Fixed, thanks.

docs/development/_index.md
illume: Updated the file to 2 lines before a header. Thanks.

Makefile
illume: Fixed.

docs/installation/base-url.md
illume: I found this a bit strange, and I'm not sure why... but yes it does.

frontend/src/helpers/index.ts
```diff
@@ -71,3 +72,69 @@ export function addQuery(
...
+  url += baseUrl ? baseUrl + '/' : '/';
+
+  return url;
+}
+
+declare global {
+  interface Window {
+    headlampBaseUrl?: string;
+  }
+}
```
joaquimrocha, line 109: Isn't this going to clash with the same Window interface we have declared in the src/plugin/index.tsx?

frontend/package.json
```diff
@@ -48,7 +49,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "if-env PUBLIC_URL react-scripts build || cross-env PUBLIC_URL=./ react-scripts build --max_old_space_size=768 && rimraf build/frontend/index.baseUrl.html",
```
illume, line 52: There's only one build script in package.json again.

frontend/src/helpers/index.ts
illume: This just extends the type, so it's fine to extend the type multiple times (AFAIK).

I'm testing the plugins still work with this branch, and will leave this item open until that's done.

illume: Plugins work now.

illume:
Testing base-url with plugins done.

joaquimrocha (commented):
Just one extra thing I noticed now, and then I think we will be able to merge.

backend/cmd/headlamp.go
```diff
@@ -133,7 +202,11 @@ func StartHeadlampServer(config *HeadlampConfig) {
 	pluginListURLS := make([]string, 0, len(files))
 
 	for _, f := range files {
-		pluginFileURL := filepath.Join("/plugins", f.Name(), "main.js")
+		if f.Name() == ".gitignore" {
+			continue
+		}
+
+		pluginFileURL := filepath.Join(config.baseURL+"/plugins", f.Name(), "main.js")
```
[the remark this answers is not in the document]
illume: Fixed. Thanks.

PR #192, opened 2021-02-08
Allow viewing resource YAML even if its read-only token

### LINKED ISSUE #175: Add YAML view dialog for resource
**Current situation**
Users can see a resource's details view that shows information about that resource, but maybe we're still missing some information from it.
One way of allowing users to see all the information for a resource is to show the YAML associated with it. This currently only possible for users with write permissions as they can use the edit dialog for it.

**Impact**
Users may only be able to see the limited information we're showing for a resource, in its details view.

**Ideal future situation**
Ideally we'd have a "View YAML" header action that any users with view credentials can use.

**Implementation options**
We should have a YAML button (with an eye icon, or a YAML document icon, or similar) and the tooltip "View YAML" as a header action. Upon activating it, a dialog with a read-only dialog is shown with the YAML code.

Let's make this exclusive for users who don't have the update button showing (i.e. either you see the edit button or the view button).

ashu8912:
Users logged in via a read only token can also see resource YAML(in read-only mode)

fixes #175

### DIFF
--- a/frontend/src/components/common/ViewButton.tsx
+++ b/frontend/src/components/common/ViewButton.tsx
@@ -1,0 +1,41 @@
+import eyeIcon from '@iconify/icons-mdi/eye';
+import eyeOff from '@iconify/icons-mdi/eye-off';
+import Icon from '@iconify/react';
+import { IconButton, Tooltip } from '@material-ui/core';
+import React from 'react';
+import { KubeObject } from '../../lib/k8s/cluster';
+import EditorDialog from './EditorDialog';
+
+interface ViewButtonProps {
+    item: KubeObject;
+}
+
+function ViewButton(props: ViewButtonProps) {
+  const {item} = props;
+  const [toggle, setToggle] = React.useState(false);
+  function handleButtonClick() {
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/ViewButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <>
+      <Tooltip title="View YAML">
+        <IconButton
+          edge="end"
+          aria-label="toggle field visibility"
+          onClick={handleButtonClick}
+          onMouseDown={event => event.preventDefault()}
+        >
+          <Icon icon={toggle ? eyeOff : eyeIcon} />
```
joaquimrocha, line 29: We don't really need a different icon because we have a modal anyway when the editor is opened. It's a small change but let's remove it.

joaquimrocha (changes requested):
I was going to suggest using the initial description I had give: just adapting the EditButton so it makes the editor read-only & change the icon, as a lot of code can be shared.
But we can still keep this approach of having the YAML view button as its own component though, so we may reuse it later if needed.

Let a few comments in the code though.

PR #193, opened 2021-02-08
Add some documentation on frontend dependencies

illume:
Adds some documentation on dependencies used in Headlamp and for what.


joaquimrocha:
@illume , as discussed in our 1:1 I think this article would require us to maintain it which is a burden not outweighed by the benefits it brings.

So instead we should have a guide about how to implement common things throughout Headlamp, which also talk about how some of the dependencies are used: which states are stored in Redux, how to create a cancellable cluster action, how to create a table with pagination + sorting, etc.

Feel free to close this PR, or overwrite it with a guide from the proposal above.

PR #194, opened 2021-02-08
Add namespace column to events table.

### LINKED ISSUE #191: Add namespace column to the events table
**Description**

The events' table is missing a namespace column.

**Impact**

Users cannot filter by namespace, but it will get the namespacing filtering setting in the background, showing no entries in the table in this case.

**Expected behavior**
There is a namespace column and header filter in the events table.

ashu8912:
fixes #191


illume (approved):
:+1: :tada:

PR #200, opened 2021-02-17
plugin demo

illume:

# How to use

# Testing done


frontend/src/plugin/index.tsx
```diff
@@ -86,10 +86,23 @@ function loadDevPlugins() {
   }
 }
 
+async function loadEnabledPlugins() {
+  // const enabledPlugins = await fetch('/plugins/enabled')
+  const enabledPlugins = ['greenify'];
+
+  for (const plugin of enabledPlugins) {
+    const aPlugin = await import(`./plugins-enabled/${plugin}`);
```
joaquimrocha, line 94: Even we use a list of plugins that have to get enabled as a second integration point, I think we want them in a generic place and just have a list of which ones are enabled to decide when we go through them.

PR #202, opened 2021-02-19
Add stories, frontend coverage, fix mui random classnames in snapshots

illume:
# Some storybook improvements

- Stop Material UI from using using random classNames in storyshots (so do not have to regenerate other unrelated snapshots when randomly assigned class names change)
- Setup coverage testing, run on CI
- Add Resource/ViewButton.stories
- Reorganize Resource components together (to find related components)
- Rename Tab to Tabs (file name should be component name)

# How to use

```
cd frontend
CI=true npm run test -- --coverage
npm run storybook
```


frontend/src/components/common/Tabs.tsx
```diff
@@ -10,13 +10,13 @@ function a11yProps(index: number) {
   };
 }
 
-export interface Tab {
+export interface TabItem {
```
illume, line 13: This was renamed because Tab interface was shadowing the imported Tab.

joaquimrocha: You could import it as `MuiTab`

frontend/src/components/common/Resource/ConfirmButton.tsx
```diff
@@ -1,6 +1,6 @@
 import Button, { ButtonProps } from '@material-ui/core/Button';
 import React from 'react';
-import { ConfirmDialog } from './Dialog';
+import { ConfirmDialog } from '../Dialog';
```
joaquimrocha, line 3: I don't think the ConfirmButton has something to do with a cluster resource. What am I missing?

frontend/src/components/common/Tabs.tsx
illume: Ah yes, that's better.

frontend/src/components/common/Resource/ConfirmButton.tsx
illume: Yes, my mistake. Moved it back.

joaquimrocha (approved):

PR #203, opened 2021-02-22
Add namespace to url

ashu8912:


frontend/src/helpers/index.ts
```diff
@@ -32,3 +32,43 @@ export function isElectron(): boolean {
...
+  return filterValue.split('~');
+}
+
+export function addQuery(
+  queryObj: { [key: string]: string },
+  queryParamDefaultObj: { [key: string]: string } = {},
+  history: any,
+  location: any,
+  tableName = '',
+  isGlobalFilter: boolean = false
```
joaquimrocha, line 52: This is not being used.

joaquimrocha (changes requested):
I found a way to keep the namespaces separated by + signs in the URL, turns out we need to separate by spaces.
I also removed an unnecessary change you added.

Please review + squash these two changes to their respective commits.

frontend/src/helpers/index.ts
ashu8912: fixed

PR #204, opened 2021-02-23
Add "make docs" with typedoc for docs/developer/api

illume:
We generate `api` under docs/development.

Using typedoc for the generated `api` files.

# How to use

`make docs`

# Testing done

`make docs` outputs static markdown, and you can view that in the following folders:

- docs/development/api

With the website using the docs branch at (https://github.com/kinvolk/docs/pull/14) inside the website repo you can do `make docs run` and have it serve the api folder with hugo.
- http://localhost:1313/docs/headlamp/0.2/development/api/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ backend/server
 backend/server.exe
 backend/tools
 app/electron/src/*
+docs/development/storybook/
--- a/Makefile
+++ b/Makefile
@@ -63,3 +63,7 @@ image:
 	-t $(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
 	Dockerfile \
 	.
+
+.PHONY: docs
+docs:
+	cd frontend && npm install && npm run build-typedoc
--- a/docs/development/api/_index.md
+++ b/docs/development/api/_index.md
@@ -1,0 +1,12 @@
+---
+title: "API"
+linkTitle: "API"
+slug: "_index"
+---
+
+## Table of contents
+
+### Modules
+
+- [k8s](modules/k8s.md)
+- [util](modules/util.md)
--- a/docs/development/api/interfaces/_index.md
+++ b/docs/development/api/interfaces/_index.md
@@ -1,0 +1,3 @@
+---
+title: "Interfaces"
+---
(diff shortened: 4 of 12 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
 backend/server
 backend/server.exe
 backend/tools
-app/electron/src/*
\ No newline at end of file
+app/electron/src/*
+docs/development/api-reference/
```
joaquimrocha, line 5: Let's call this simply `docs/development/api`.

joaquimrocha (commented):
The API docs should ideally be generated in markdown, maybe this helps: https://www.npmjs.com/package/typedoc-plugin-markdown

Also commented on the kinvolk/docs repo.

joaquimrocha (changes requested):
We still need some work in the integration:

Screenshot from 2021-02-25 12-10-20

We need to lose the breadcrumb is generates and add the front-matter to the doc files:
```
---
title: THIS TITLE NEEDS TO BE ADDED
---
```

illume:
Indeed. I have a discussion with the typedoc-plugin-markdown author on how best to proceed generating the front-matter. Will come back to this PR when that discussion is finished.

joaquimrocha:
> Indeed. I have a discussion with the typedoc-plugin-markdown author on how best to proceed generating the front-matter. Will come back to this PR when that discussion is finished.

Great! Thanks.

BTW, regarding Storybook, I am still a bit reluctant that it'll slow down the website build, so let's keep it out and just add insttructions in the docs on how to run it.
I guess running it locally also makes sense if you're looking into how to develop things anyway.

illume:
There's a fancy new theme for Hugo now thanks to the typedoc-plugin-markdown author which adds front-matter. This branch now uses the first release of the theme. It still has some rough edges with links in the side bar and slugs, which I'll debug.

illume:
This now:
- does not build storybook with `make docs`
- looks good in the website
- uses the new hugo theme for the typedoc markdown, with some small fixes to the output until/if that plugin is improved at some point in the future.

The broken test is because of an unrelated issue, see temp hotfix in PR: https://github.com/kinvolk/headlamp/pull/211

illume:
There's a new typedoc-hugo plugin for typedoc markdown that fixes the remaining issues, and I removed the tidy-up script.

joaquimrocha:
I have added an extra option to the typedoc's config that hides the TOC on pages (since we already have one being generated by default).

PR #216, opened 2021-03-11
App using plugins, and live reload of plugins

illume:
- [x] the app can load plugins as well
- [x] we can use a user config plugins folder (eg. `~/.config/Headlamp/plugins`)
- [x] live reload happens when the plugins folder changes

# How to use

```bash
make backend run-backend
make run-frontend
# copy a plugin into ~/.config/Headlamp/plugins
```

# Testing done

- [x] app can load plugins from the user config plugins folder
- [x] live reload happens when creating/updating/deleting plugins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,16 @@ RUN cd ./backend && go build -o ./server ./cmd/
 
 RUN cd ./frontend && npm install && npm run build
 
-# Create a plugins folder if none exists so we copy whatever exists to the
-# image.
+# Backwards compatibility, move plugin folder to only copy matching plugins.
+RUN mv plugins plugins-old
+
 RUN mkdir -p ./plugins
+
+# Backwards compatibility, copy any matching plugins found inside "./plugins-old" into "./plugins".
+# They should match plugins-old/MyFolder/main.js, otherwise they are not copied.
+RUN find plugins-old -mindepth 2 -maxdepth 2 -type f  | grep -i main.js$ | xargs -i dirname {} | xargs -i cp -a {} ./plugins/
+
+RUN find .plugins -mindepth 2 -maxdepth 2 -type f  | grep -i main.js$ | xargs -i dirname {} | xargs -i cp -a {} ./plugins/
 
 FROM $IMAGE_BASE
 
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -14,7 +14,7 @@ func main() {
 	insecure := flag.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	// @todo: Make this a uint and validate the values
 	port := flag.String("port", "4466", "Port to listen from")
-	pluginDir := flag.String("plugins-dir", "./plugins", "Specify the plugins directory to build the backend with")
+	pluginDir := flag.String("plugins-dir", "", "Specify the plugins directory to build the backend with")
 	// For inCluster config we need to get the oidc properties from the flags
 	oidcClientID := flag.String("oidc-client-id", "", "ClientID for OIDC")
 	oidcClientSecret := flag.String("oidc-client-secret", "", "ClientSecret for OIDC")
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -14,7 +14,7 @@ func main() {
 	insecure := flag.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	// @todo: Make this a uint and validate the values
 	port := flag.String("port", "4466", "Port to listen from")
-	pluginDir := flag.String("plugins-dir", "", "Specify the plugins directory to build the backend with")
+	pluginDir := flag.String("plugins-dir", defaultPluginDir(), "Specify the plugins directory to build the backend with")
 	// For inCluster config we need to get the oidc properties from the flags
 	oidcClientID := flag.String("oidc-client-id", "", "ClientID for OIDC")
 	oidcClientSecret := flag.String("oidc-client-secret", "", "ClientSecret for OIDC")
--- a/docs/development/_index.md
+++ b/docs/development/_index.md
@@ -208,5 +208,5 @@ a deployment of Headlamp using the Docker image can mount a plugins folder
 and point to it by using the mentioned option.
 
 An alternative is to build an image that ships some plugins in it. For that,
-just create a plugins folder in the Headlamp project directory as the Dockerfile
+just create a ".plugins" folder in the Headlamp project directory as the Dockerfile
 will include it and point to it by default.
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -68,6 +68,13 @@ export async function request(
     clearTimeout(id);
   }
 
+  // The backend signals through this header that it wants a reload.
+  // See plugins.go
+  const headerVal = response.headers.get('X-Reload');
+  if (headerVal && headerVal.indexOf('reload') !== -1) {
+    window.location.reload();
+  }
+
   if (!response.ok) {
     const { status, statusText } = response;
     if (autoLogoutOnAuthError && status === 401 && opts.headers.Authorization) {
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ backend/server.exe
 backend/tools
 app/electron/src/*
 docs/development/storybook/
+.plugins
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -167,6 +167,9 @@ func StartHeadlampServer(config *HeadlampConfig) {
 		kubeConfigPath = GetDefaultKubeConfigPath()
 	}
 
+	config.pluginDir = defaultPluginDir(config.pluginDir)
+	log.Printf("plugins-dir: %s\n", config.pluginDir)
+
 	var contexts []Context
 
 	// In-cluster
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -167,8 +167,12 @@ func StartHeadlampServer(config *HeadlampConfig) {
 		kubeConfigPath = GetDefaultKubeConfigPath()
 	}
 
-	config.pluginDir = defaultPluginDir(config.pluginDir)
 	log.Printf("plugins-dir: %s\n", config.pluginDir)
+
+	if !config.useInCluster {
+		// in-cluster mode is unlikely to want reloading plugins.
+		go watchForChanges(config.pluginDir)
+	}
 
 	var contexts []Context
 
@@ -448,6 +452,8 @@ func proxyHandler(url *url.URL, proxy *httputil.ReverseProxy) func(http.Response
 		request.URL.Host = url.Host
 		request.URL.Path = mux.Vars(request)["api"]
 		request.URL.Scheme = url.Scheme
+
+		pluginReloadResponse(writer)
 
 		log.Println("Requesting ", request.URL.String())
 		proxy.ServeHTTP(writer, request)
--- a/backend/cmd/plugins.go
+++ b/backend/cmd/plugins.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"log"
+	"os"
+	"path"
+)
+
+// folderExists(path) returns true if the folder exists.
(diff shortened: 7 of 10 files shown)


backend/cmd/server.go
```diff
@@ -14,7 +44,7 @@ func main() {
 	insecure := flag.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	// @todo: Make this a uint and validate the values
 	port := flag.String("port", "4466", "Port to listen from")
-	pluginDir := flag.String("plugins-dir", "./plugins", "Specify the plugins directory to build the backend with")
+	pluginDir := flag.String("plugins-dir", "", "Specify the plugins directory to build the backend with")
```
joaquimrocha, line 47: Nitpick: In order for it to inform users about what the default dir is (without having to check the docs), we could get the default dir and use it here as the default.

backend/cmd/pluginwatcher.go
```diff
@@ -0,0 +1,104 @@
...
+			done <- struct{}{}
+		}()
+	}
+}
+
+// func main() {
+// 	done := make(chan bool)
+// 	go findChanges("/tmp/plugins")
+// 	<-done
+// }
```
joaquimrocha, line 104: Leftover from debug?

backend/cmd/server.go
```diff
@@ -69,6 +69,8 @@ func main() {
 	*pluginDir = defaultPluginDir(*pluginDir)
 	log.Printf("plugins-dir: %s\n", *pluginDir)
 
+	go watchForChanges(*pluginDir)
```
joaquimrocha, line 72: Should we add this only if we're not in-cluster? Just thinking it's unlikely that in-cluster deployments will update the plugins without relaunching the service.
ATM either way seems fine and the extra IO should not be too much, but I'd prefer to err on the side of caution and have it the monitor only locally.

Another option is to add it only if the `-dev` option is passed, but we'd have to expose that to the desktop app's CLI, and later on we'd likely have to revisit that since we may want to allow desktop users to just replace the plugins on the fly (not only for development but for regular use, I mean).

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -68,6 +68,13 @@ export async function request(
     clearTimeout(id);
   }
 
+  // The backend signals through this CORS safe header that it wants a reload.
+  // See pluginwatcher.go
+  const headerVal = response.headers.get('Content-Language');
```
joaquimrocha, line 73: In https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS I see this:
```WebKit Nightly and Safari Technology Preview place additional restrictions on the values allowed in the Accept, Accept-Language, and Content-Language headers. If any of those headers have ”nonstandard” values, WebKit/Safari does not consider the request to be a “simple request”. What values WebKit/Safari consider “nonstandard” is not documented, except in the following WebKit bugs:
```

AFAIU, shouldn't be a problem except for users who use the headless mode (if we restrict the plugin watch to desktop use) with a non-final version of Safari, but I wonder if we'll hit an issue in the future.

Doing a quick search, have you tried adding a dedicated allowed header?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

joaquimrocha (changes requested):
Left a few comments but it's looking good!

backend/cmd/server.go
illume: Good idea. This is done.

backend/cmd/pluginwatcher.go
illume: Yeah. It's been removed. Thanks.

backend/cmd/server.go
illume: Indeed. I think -in-cluster option is best, and have changed it to not watch when in-cluster is on. I thought about -dev too, and had the same thought about having it on by default with the app.

So yeah... turning it off for in-cluster it is :)

illume:
Thanks for the review. I pushed some changes encapsulating the plugins functions into plugins.go, and also addressed a couple of your notes.

I'll look into the config folder name, and the header now.

frontend/src/lib/k8s/apiProxy.ts
illume: I used Access-Control-Expose-Headers in the end, which works easily in dev mode and with the app.

backend/cmd/server.go
```diff
@@ -3,9 +3,39 @@ package main
...
+	pluginDirDefault := "./plugins"
+
+	userConfigDir, err := os.UserConfigDir()
+	if err != nil {
+		log.Printf("error getting user config dir: %s\n", err)
+
+		return pluginDirDefault
+	}
+
+	pluginsConfigDir := path.Join(userConfigDir, "Headlamp", "plugins")
```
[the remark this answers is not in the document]
illume: I poked around the Electron docs and searched the interwebs, and couldn't see how to change this easily. I think it's tied to the app name.

illume:
I noticed one bug in the headless app mode not reladed to this PR... When reloading the page it errors and shows as a blank page. The reloading works fine in the app and with other ways of running Headlamp however.

illume:
I added the backwards compatible copying of plugins from ./plugins to ./.plugins and use the .plugins folder as a default if it exists. So this way the [lokomotive build.sh](https://github.com/kinvolk/lokomotive-web-ui/blob/master/build.sh) should still work without changes.

backend/cmd/server.go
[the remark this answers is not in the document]
illume: This issue should probably be handled outside of this PR.

Dockerfile
```diff
@@ -14,17 +14,27 @@ WORKDIR /headlamp
...
 
 COPY --from=base-build /headlamp/backend/server /headlamp/server
 COPY --from=base-build /headlamp/frontend/build /headlamp/frontend
+# Backwards compatibility, "./plugins" was the old name.
 COPY --from=base-build /headlamp/plugins /headlamp/plugins
+COPY --from=base-build /headlamp/.plugins /headlamp/.plugins
+
+# Backwards compatibility, copy any plugins found inside "./plugins" into ".plugins".
+# They should match plugins/MyFolder/main.js, otherwise they are not copied.
+RUN find plugins -maxdepth 2 -type f  | grep -i main.js$ | xargs -i dirname {} | xargs -i cp -a {} .plugins/
```
joaquimrocha, line 37: Unless my tests didn't work (I tested with the lokomotive-web-ui plugins), I ended up with /headlamp/.plugins empty.

docs/development/_index.md
```diff
@@ -106,7 +106,7 @@ $ DOCKER_IMAGE_VERSION=development make image
 ...
 Successfully tagged kinvolk/headlamp:development
 
-$ docker run --network="host" -p 127.0.0.1:4466:4466/tcp --mount type=bind,source="/home/rene/.minikube",target=$HOME/.minikube --mount type=bind,source="$HOME/.kube",target=/root/.kube kinvolk/headlamp:development /headlamp/server -html-static-dir /headlamp/frontend -plugins-dir=/headlamp/plugins
+$ docker run --network="host" -p 127.0.0.1:4466:4466/tcp --mount type=bind,source="/home/rene/.minikube",target=$HOME/.minikube --mount type=bind,source="$HOME/.kube",target=/root/.kube kinvolk/headlamp:development /headlamp/server -html-static-dir /headlamp/frontend -plugins-dir=/headlamp/.plugins
```
joaquimrocha, line 109: I meant we'd use `.plugins` in the repo dir, not in the final image.

docs/development/_index.md
```diff
@@ -170,7 +170,7 @@ spec:
         - -html-static-dir
         - /headlamp/frontend
         - -in-cluster
-        - -plugins-dir=/headlamp/plugins
+        - -plugins-dir=/headlamp/.plugins
```
joaquimrocha, line 173: This is a breaking change for any deployments who just update the image tag, whereas before I said we could break just with how it was generated, having .plugins in the repo file. I am happy to clarify a bit better tomorrow.

joaquimrocha (changes requested):
Left a few comments.

illume:
I added the ".plugins" folder to gitignore

Dockerfile
```diff
@@ -14,17 +14,27 @@ WORKDIR /headlamp
 
 RUN cd ./backend && go build -o ./server ./cmd/
 
-RUN cd ./frontend && npm install && npm run build
-
 # Create a plugins folder if none exists so we copy whatever exists to the
 # image.
+RUN mkdir -p ./.plugins
+
+# Backwards compatibility, Create the plugins folder if none exists.
```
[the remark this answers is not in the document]
illume: Thanks. I used the mv plugins plugins-old way.

Dockerfile
```diff
@@ -14,17 +14,27 @@ WORKDIR /headlamp
 
 RUN cd ./backend && go build -o ./server ./cmd/
 
-RUN cd ./frontend && npm install && npm run build
```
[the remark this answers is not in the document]
illume: Fixed the commit title.

illume:
In the docker image, it keeps things in "headlamp/plugins" again. I had to update the find command.

Dockerfile
```diff
@@ -16,10 +16,21 @@ RUN cd ./backend && go build -o ./server ./cmd/
...
-# Create a plugins folder if none exists so we copy whatever exists to the
-# image.
+# Backwards compatibility, move plugin folder to only copy matching plugins.
+RUN mv plugins plugins-old
+
 RUN mkdir -p ./plugins
 
+# Backwards compatibility, copy any plugins found inside "./plugins" into ".plugins".
+# They should match plugins/MyFolder/main.js, otherwise they are not copied.
+RUN find plugins-old -maxdepth 2 -type f  | grep -i main.js$ | xargs -i dirname {} | xargs -i cp -a {} ./plugins/
```
joaquimrocha, line 26: This should never be the case AFAIU. We assume that plugins-old will contain plugins in the old form of `MyPlugin/dist/main.js`, which should be covered by the other script you run below.

Dockerfile
```diff
@@ -16,10 +16,21 @@ RUN cd ./backend && go build -o ./server ./cmd/
...
+
+RUN find .plugins -maxdepth 2 -type f  | grep -i main.js$ | xargs -i dirname {} | xargs -i cp -a {} ./plugins/
```
joaquimrocha, line 28: Should we be a bit more strict here and copy the files only if they are in `.plugins/*/main.js`? (i.e. why do we allow a depth of 2? Are we expecting `.plugins/main.js`?)

Dockerfile
```diff
@@ -16,10 +16,21 @@ RUN cd ./backend && go build -o ./server ./cmd/
...
+
+# Backwards compatibility, plugins could also be in a MyPlugin/dist/main.js file.
+RUN for i in $(find ./plugins-old/*/dist/main.js); do plugin_name=$(echo $i|cut -d'/' -f3); mkdir -p plugins/$plugin_name; cp $i plugins/$plugin_name; done
```
joaquimrocha, line 31: I would run this before running the new way (the .plugins), so any new way would overwrite a forgotten old format.
Either that or fail if the path exists when copying (basically to prevent overwriting plugins from  an existing old format, but running before should be enough).

joaquimrocha (changes requested):
This works already for the image :partying_face: 
I left just a couple of comments on the Dockerfile to simplify it but I am positive we'll merge it next.

Dockerfile
illume: I moved this to before.

Dockerfile
illume: I added a mindepth 2 so it only matches plugins/MyFolder/main.js

Dockerfile
illume: plugins-old could also theoretically contain the [documented MyPlugin/main.js form](https://kinvolk.io/docs/headlamp/0.2/development/plugins/building/#shipping--deploying-plugins).

I wonder if we should support .plugins/MyPlugin/main.js and .plugins/MyPlugin/dist/main.js ? The dist form is useful because that's where the plugins are output by default, and then maybe people just point their plugins-dir at where they build their plugins.

illume: I removed the dist folder handling from the Dockerfile.

joaquimrocha (approved):

PR #217, opened 2021-03-11
App refactoring

joaquimrocha:
The app was relying on the use of the node_modules from the frontend and was shipping the electron js twice by mistake.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -2,26 +2,22 @@
   "name": "headlamp",
   "version": "0.2.1",
   "description": "",
-  "main": "electron/src/main.js",
+  "main": "electron/main.js",
   "homepage": ".",
   "productName": "Headlamp",
   "scripts": {
-    "compile-electron": "babel electron --out-dir electron/src/ --extensions '.ts'",
-    "copy-server-binary": "cd ../ && make backend && cp -r ./backend/server ./app/electron/",
-    "start": "npm run copy-server-binary && cd ../ && make run-backend  & cd ../frontend/ && export BROWSER=none && npm start",
-    "build": "npm run compile-electron && npm run copy-server-binary && cd ../frontend/ && PUBLIC_URL='./' npm run build && cp -r ../app/electron/src/. build/electron",
-    "package": "mkdir -p build && cp -r ../frontend/build/* ./build && electron-builder build -c.extraMetadata.main=build/electron/main.js --publish never",
-    "serve-dev": "npm run compile-electron && export ELECTRON_DEV=1 && export ELECTRON_START_URL=http://localhost:3000 && electron .",
-    "publish": "mkdir -p build && cp -r ../frontend/build/* ./build && electron-builder build -c.extraMetadata.main=build/electron/main.js --publish always"
+    "compile-electron": "babel electron --out-dir electron/ --extensions '.ts'",
+    "start": "cd ../ && make backend && make run-backend  & cd ../frontend/ && export BROWSER=none && npm start",
+    "build": "npm run compile-electron && electron-builder --dir --publish never",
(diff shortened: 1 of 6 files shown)


app/package.json
```diff
@@ -2,33 +2,35 @@
...
   },
   "build": {
     "appId": "com.kinvolk.headlamp",
-    "extraMetadata": {
-      "main": "build/electron/main.js"
-    },
+    "beforeBuild": "./scripts/build-backend.js",
     "afterSign": "mac/scripts/notarize.js",
-    "asar": false,
+    "asar": true,
```
illume, line 19: We set asar false before to get source maps working. I'll check if sourcemaps are still working with the new setup.

app/package.json
```diff
@@ -2,33 +2,35 @@
...
+    "artifactName": "${productName}-${version}-${os}-${arch}.${ext}",
```
illume, line 20: Do we really want these?

In the other PR https://github.com/kinvolk/headlamp/pull/214 I set this to just 
```json
    "artifactName": "${name}.${ext}",
```

Which makes eg. "Headlamp.app".

illume:
I'm getting this error with `make app-linux`
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    Error: Cannot find module '/home/rene/dev/kinvolk/headlamp/app/scripts/build-backend.js'
```

illume:
Looking at the release notes for breaking changes in electron, doesn't look like there's anything they mentioned we're using.
- https://www.electronjs.org/releases/stable#release-notes-for-v1200
- https://www.electronjs.org/releases/stable?page=3#release-notes-for-v1100

illume:
With the linux app, it doesn't seem to find the JavaScript.

app/package.json
joaquimrocha: Ah right. I didn't remember this. Thanks.

app/package.json
joaquimrocha: I do want it in this format as it matches the name we push when we release.

joaquimrocha:
> With the linux app, it doesn't seem to find the JavaScript.

I have pushed the missing file.

app/package.json
illume: The sourcemaps are working now. So we can keep the asar: true.

illume:
I wiped the node_modules folders again and this time `make app-linux` is loading. I clicked about the app a bit and could not see any issues.

I'll try some other configurations.

illume:
After quitting the app, the server process is still running for me. I had to manually kill it. Starting the app again after this means the app is nonfunctional.

illume (approved):
:+1:

illume:
I tried the new commit with the linux builds (the tar.gz and AppImage) - it builds and runs.

illume:
I see this warning when building:
```
  • description is missed in the package.json  appPackageFile=/home/rene/dev/kinvolk/headlamp/app/package.json
```

How about this? 

app/package.json
```json
  "description": "Headlamp is an easy-to-use and extensible Kubernetes web UI. Headlamp was created to be a Kubernetes web UI that has the traditional functionality of other web UIs/dashboards available (i.e. to list and view resources) as well as other features.",
```

illume:
With a fresh clone, and `make app-linux` it has these warnings:
```
  • file source doesn't exist  from=/home/rene/dev/kinvolk/tmp/headlamp/frontend/build
  • file source doesn't exist  from=/home/rene/dev/kinvolk/tmp/headlamp/backend/server
  • default Electron icon is used  reason=application icon is not set
```

joaquimrocha:
Right. That's because electron-builder is filtering out the node_modules inside each node_modules/DEPENDENCY..

PR #218, opened 2021-03-12
Fix baseURLReplace on read only fs

illume:
This issue caused the AppImage build to fail to run.

# Testing done

### with the AppImage on linux (failed previously)

```bash
rm -rf app/dist
make app-linux 
cd app/dist
./Headlamp-0.2.1-linux-x86_64.AppImage --headless
```

### base-url in dev mode

```bash
./backend/server -dev -base-url /headlamp
PUBLIC_URL="/headlamp" make run-frontend
```
### base-url in static mode

```bash
cd frontend
npm install 
npm run build
cd ..
./backend/server -dev -base-url /headlamp -html-static-dir frontend/build
```

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -16,6 +16,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	oidc "github.com/coreos/go-oidc"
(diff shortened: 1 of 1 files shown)


backend/cmd/headlamp.go
```diff
@@ -110,13 +111,35 @@ func copyReplace(src string, dst string,
...
 // make sure the base-url is updated in the index.html file.
 func baseURLReplace(staticDir string, baseURL string) {
 	indexBaseURL := path.Join(staticDir, "index.baseUrl.html")
 	index := path.Join(staticDir, "index.html")
 
+	if baseURL == "" && strInFile("headlampBaseUrl=\".\"", index) {
+		// The index.html does not need resetting from a different baseURL.
+		return
+	}
+
```
joaquimrocha, line 137: IIUC, the `index.baseUrl.html` will always have the original value of `index.html`, right? Also, the `index.baseUrl.html` should only exist if we override the baseUrl (at least with these new changes).

So an extra check which prevents the reading would be to check if the indexBaseUrl file exists and early return if it doesn't.

joaquimrocha (changes requested):
Left a comment.
I am also thinking that maybe instead of writing the index file in general for the base URL, there should be a way to serve it as a []bytes which means we could keep the replacement in memory instead of creating a new file. But that can be handled in a different PR later.

backend/cmd/headlamp.go
illume: Good idea. I added the extra check to see if the file is missing first to avoid the read.

joaquimrocha (approved):

PR #224, opened 2021-03-18
backend: Don't continue OIDC logic on errors

joaquimrocha:
As reported in https://github.com/kinvolk/headlamp/issues/209#issuecomment-801329837 , a mismatch between the OIDC issuer's returned URL and the one the client is configured with was crashing the server because the errors were not being treated.

I looked also into how to be tolerant to the URL difference of a trailing slash, but I noticed that it's a hard comparison within the OIDC lib we use, and any examples on Github were not working around this, so I think for now it's safer to keep the error reporting alone and expect users to set up the matching URLs (any solution for a tolerant comparison should be implemented in the OIDC lib IMO).

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -264,10 +264,14 @@ func StartHeadlampServer(config *HeadlampConfig) {
 		oidcAuthConfig, err := GetClusterOidcConfig(cluster)
 		if err != nil {
 			log.Printf("Error getting %s cluster oidc config %s", cluster, err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
 		}
 		provider, err := oidc.NewProvider(ctx, oidcAuthConfig.IdpIssuerURL)
 		if err != nil {
 			log.Printf("Error while fetching the provider from %s error %s", oidcAuthConfig.IdpIssuerURL, err)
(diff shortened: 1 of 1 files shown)


illume:
I wonder what the error message is, and if it's useful for figuring out that the URL could do without a '/' at the end?
If it's not useful, we could print a hint.
```golang
log.Printf("Error while fetching the provider from %s error %s. Does your URL have an extra / at the end?", oidcAuthConfig.IdpIssuerURL, err)
```

joaquimrocha:
> I wonder what the error message is, and if it's useful for figuring out that the URL could do without a '/' at the end?
> If it's not useful, we could print a hint.
> 
> ```go
> log.Printf("Error while fetching the provider from %s error %s. Does your URL have an extra / at the end?", oidcAuthConfig.IdpIssuerURL, err)
> ```

I don't think we can add that hint to all errors as they may be unrelated.

The reporter shared the error they got (which IMO is good enough):
`Error while fetching the provider from https://sso-test.test.mydomain.com/auth/realms/sso/ error oidc: issuer did not match the issuer returned by provider, expected "https://sso-test.test.mydomain.com/auth/realms/sso/" got "https://sso-test.test.mydomain.com/auth/realms/sso"`

illume (approved):
:+1:

PR #225, opened 2021-03-19
Fix copying innexistent plugins

joaquimrocha:
Copying the plugins was failing if the plugins folder was empty, because
    in the Dockerfile we tried to copy from /headlamp/plugins but we first
    had moved any plugins folder to plugins-old and if the folder was empty,
    then the copy command wouldn't create the final plugins directory again.
    
    Fixing that just required creating the mentioned plugins folder after
    the move, but there was also the problem that the copy command was
    copying everything in each plugins folder and not just the main.js file.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,5 @@
 .git
 app
 backend/tools/
+docs/
+**/*.stories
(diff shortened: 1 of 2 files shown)


illume (approved):
:+1:

PR #226, opened 2021-03-22
Auto image publishing

joaquimrocha:
* Add image publishing on a new tag.
* Build an image for every frontend/backend change.
* Fix an issue when building the image (again)


illume (approved):
:tada: :balloon:

PR #227, opened 2021-03-22
Only install frontend production dependencies when making container

illume:
So that only the required deps to build are installed before building the docker image. Move some deps out of devDependencies because the build depends on them.

Also includes dependency updates:
- monaco editor
- storybook
- react-scripts
- audit fixes
- ...

Many(all?) of the unfixable audit issues are related to waiting on a new react-scripts release: https://github.com/facebook/create-react-app/issues/11012

# How to use

`make image`

# Testing done

`make image` works.


frontend/package.json
```diff
@@ -44,7 +44,21 @@
...
+    "@typescript-eslint/parser": "^4.14.2",
+    "eslint": "^7.19.0",
+    "eslint-config-prettier": "^7.2.0",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint-plugin-simple-import-sort": "^7.0.0",
+    "@types/js-base64": "^2.3.1",
+    "@types/js-yaml": "^3.12.4",
+    "prettier": "^2.2.1"
```
joaquimrocha, line 61: I don't think these belong to deps, right?
This means that we will ship these as dependencies when we do production builds, and they won't be used at all, unless I am missing something?
(About the `iconify`, it feels like it should be a dependency, but the way it works and how it's recommended, is that it can stay as a dev dependency AFAIU)

illume: It just means these dependencies are required for doing production builds. For example, if you remove the iconify icons from there it fails to compile (with `npm install --only=prod && npm run build`).

frontend/src/components/common/__snapshots__/ActionsNotifier.stories.storyshot
```diff
@@ -11,14 +11,14 @@ exports[`Storyshots ActionsNotifier Some 1`] = `
   id="root"
 >
   <div
-    class="makeStyles-root makeStyles-bottom makeStyles-left makeStyles-bottom makeStyles-left makeStyles-reverseColumns"
+    class="makeStyles-bottom makeStyles-left makeStyles-root"
```
illume, line 14: Seems that some of the styles were duplicated before, and are only in there once now -- good.

However it seems `makeStyles-reverseColumns` was removed... so this needs checking.

illume: This is an internal to notistack-snackbar style change. Nothing to worry about.

PR #229, opened 2021-03-24
Use redux for theme

joaquimrocha:
The main motivation for this change is to not having plugins' components in the top bar being placed between the theme change and the account buttons.

On the other hand, we should in any case have the theme being set through redux as we may need it later.


illume (approved):
:+1:

PR #233, opened 2021-03-31
Add monitor-link plugin example from lokomotive-web-ui

illume:
Adding monitor-link plugin from here: https://github.com/kinvolk/lokomotive-web-ui/tree/master/plugins/monitor-link as a plugin/example

I updated the package.json, but it would be good to first do a npm install once the headlamp-plugin has been published.


joaquimrocha:
This plugin should remain the lokomotive-web-ui until we have an official list/repo to have the plugins.

PR #236, opened 2021-04-07
Add a --disable-gpu option for the app

illume:
For people who have buggy graphics drivers.

# How to use

`headlamp --disable-gpu`

Then you should see this message:
```
› Disabling GPU hardware acceleration.
```

Electron docs: https://www.electronjs.org/docs/api/app#appdisablehardwareacceleration


joaquimrocha (approved):
LGTM.

PR #243, opened 2021-04-19
Add app.yml workflow

illume:
For testing the linux app build to start with.

I looked into seeing if there's a "--fail-with-warnings" that the electron builder gives us, which I hoped to use to catch issues like the icons missing one... but it doesn't exist.

I want to add a very simple test to that PR to see if the executable runs and then exits. A --run-and-exit type command, which runs and then exits. So we can confirm executable is working.

### DIFF
--- a/app/mac/scripts/notarize.js
+++ b/app/mac/scripts/notarize.js
@@ -2,8 +2,8 @@ require('dotenv').config();
 const { notarize } = require('electron-notarize');
 
 exports.default = async function notarizing(context) {
-  const { electronPlatformName, appOutDir } = context;  
-  if (electronPlatformName !== 'darwin') {
+  const { electronPlatformName, appOutDir } = context;
+  if (electronPlatformName !== 'darwin' || process.env.CI === 'true') {
     return;
   }
 
--- a/app/package.json
+++ b/app/package.json
@@ -44,13 +44,6 @@
             "x64",
             "arm64"
           ]
-        },
-        {
-          "target": "pkg",
-          "arch": [
-            "x64",
-            "arm64"
-          ]
         }
       ],
       "hardenedRuntime": true,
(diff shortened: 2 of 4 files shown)


joaquimrocha:
Can we do that without a command? Just maybe looking for the PID and killing it?

illume:
Yes, maybe that could work too. I did a quick search-foo for and electron and signal handlers, and couldn't find much but I'm not clear if they would be a mess with electron/chromium. A command line argument(and/or environment variable) will give the possibility for the JavaScript code to run, and for the app to close gracefully at a specific point. Then we can check the return code for success.

joaquimrocha:
I have pushed a few changes to fix the linux build + add the other 2 platforms.

app/mac/scripts/notarize.js
```diff
@@ -2,8 +2,8 @@ require('dotenv').config();
 const { notarize } = require('electron-notarize');
 
 exports.default = async function notarizing(context) {
-  const { electronPlatformName, appOutDir } = context;  
-  if (electronPlatformName !== 'darwin') {
+  const { electronPlatformName, appOutDir } = context;
+  if (electronPlatformName !== 'darwin' || process.env.CI === 'true') {
```
illume, line 6: Not related to this PR... but this reminds me: If you don't have the notarization files the app-mac build currently breaks. I'll make an issue/PR.

app/package.json
```diff
@@ -65,9 +58,7 @@
     },
     "win": {
       "target": [
-        "zip",
-        "msi",
```
illume, line 69: Another semi related comment, but msi is deprecated from around 2018/2019 I think.

illume:
Just realized this is my own PR, so I can't review it...
so here's a 👍 from me.

app/mac/scripts/notarize.js
joaquimrocha: You mean the notarization credentials, right? This line is runing them off for the CI. An alternative is that disable it if the APPLE* stuff used below is not set up.

PR #253, opened 2021-05-13
Improve ingress docs

joaquimrocha:
- frontend: Remove unneeded newlines from ingress sample
- docs: Offer a simpler ingress alternative for in-cluster deployment
- docs: Re-organize the in-cluster section


alban:
When the Kubernetes cluster does not have contour or any other ingress controller, it would be good to document alternative methods to access in-cluster Headlamp:

1. `kubectl proxy --port=8080`, and then use this URL in the browser:
  http://localhost:8080/api/v1/namespaces/${HEADLAMP_NAMESPACE}/services/${HEADLAMP_SERVICE}/proxy/
2. `kubectl port-forward service/${HEADLAMP_SERVICE} 8080:80`, and then use this URL in the browser:
  http://localhost:8080

I see that the docs currently mentions `minikube service headlamp -n kube-system --url`, but that only works with minikube. The kubectl proxy/port-forward methods should be more generic and work on any clusters.

joaquimrocha:
Good point @alban , I will make the changes next week.

vbatts (approved):
LGTM 
(as far as the re-arrange goes. The additions alban mentions are good too)

joaquimrocha:
I have added some changes. Maybe you can take a look, @alban .

docs/installation/in-cluster/_index.md
```diff
@@ -0,0 +1,57 @@
...
+kubectl apply -f ./headlamp-ingress.yaml
+```
+
+## Exposing Headlamp with port-forwarding
+
+If you want to quickly access Headlamp (after having its service running) and
+don't want to set up an ingress for it, you can run use port-forwarding as follows:
+
+```bash
+kubectl port-forward service/headlamp 8080:80
```
alban, line 48: Does it work? I think `-n kube-system` should be added since https://raw.githubusercontent.com/kinvolk/headlamp/master/kubernetes-headlamp.yaml is deploying Headlamp in the kube-system namespace.

But I think it would be cleaner for headlamp to have its own namespace (maybe in a follow-up PR).

docs/installation/in-cluster/_index.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+and then you can access `localhost:8080` in your browser.
+
+## Accessing Headlamp
+
+Once Headlamp is up and running, be sure to enable access to it either by creating
+a [service account](../#create-a-service-account-token) or by setting up\
```
alban, line 56:
```suggestion
a [service account](../#create-a-service-account-token) or by setting up
```

docs/installation/in-cluster/_index.md
joaquimrocha: I did try it before and added the namespace, but forgot to add it when I copied your statement :facepalm: 
Thanks for catching this.

alban (approved):
LGTM!
Thanks

PR #254, opened 2021-05-18
Add more resource info to Pod details

joaquimrocha:
- frontend: Add link to node in Pod details view
- frontend: Use StatusLabel component for the Pod status in details view

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -1,8 +1,6 @@
 import chevronLeft from '@iconify/icons-mdi/chevron-left';
 import eyeIcon from '@iconify/icons-mdi/eye';
 import eyeOff from '@iconify/icons-mdi/eye-off';
-import menuDown from '@iconify/icons-mdi/menu-down';
-import menuUp from '@iconify/icons-mdi/menu-up';
 import { Icon } from '@iconify/react';
 import { Button, InputLabel } from '@material-ui/core';
 import Box from '@material-ui/core/Box';
@@ -10,25 +8,21 @@ import Divider from '@material-ui/core/Divider';
 import Grid, { GridProps } from '@material-ui/core/Grid';
 import IconButton from '@material-ui/core/IconButton';
 import Input, { InputProps } from '@material-ui/core/Input';
-import { makeStyles } from '@material-ui/core/styles';
 import { TextFieldProps } from '@material-ui/core/TextField';
-import Typography, { TypographyProps } from '@material-ui/core/Typography';
+import Typography from '@material-ui/core/Typography';
 import Editor from '@monaco-editor/react';
 import { Base64 } from 'js-base64';
 import _ from 'lodash';
 import * as monaco from 'monaco-editor';
 import React from 'react';
 import { Link as RouterLink, useLocation } from 'react-router-dom';
-import { ResourceClasses } from '../../../lib/k8s';
 import {
   KubeCondition,
   KubeContainer,
   KubeObject,
   KubeObjectInterface,
-  KubeOwnerReference,
 } from '../../../lib/k8s/cluster';
 import { createRouteURL, RouteURLProps } from '../../../lib/router';
-import { localeDate } from '../../../lib/util';
 import { useTypedSelector } from '../../../redux/reducers/reducers';
 import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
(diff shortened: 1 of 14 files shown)


joaquimrocha:
I have updated this PR with a new entry in the metadata display table that shows owner refs/controlled by labels.

docs/platforms.md
```diff
@@ -0,0 +1,20 @@
+# Tested Platforms
```
illume, line 1: Was this doc added accidentally?

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -58,6 +60,33 @@ interface MetadataDisplayProps {
 export function MetadataDisplay(props: MetadataDisplayProps) {
   const { resource, extraRows } = props;
 
+  function makeOwnerReferences(ownerReferences: KubeOwnerReference[]) {
```
illume, line 63: Could drop the `make` part here - since it's basically a component it should be Capitalized.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -37,6 +37,16 @@ export interface KubeMetadata {
   selfLink: string;
   labels?: StringDict;
   annotations?: StringDict;
+  ownerReferences?: KubeOwnerReference[];
+}
+
+export interface KubeOwnerReference {
```
illume, line 43: Documenting the interface fields would be helpful. 

```typescript
  /** */
  apiVersion: string;
```

illume (changes requested):
Could you please add a storybook example to show the change?

docs/platforms.md
joaquimrocha: Yes. I noticed the mistake and added it to the right branch: https://github.com/kinvolk/headlamp/pull/256
But forgot to force push this one. Updated now.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: This is basically a copy from the [K8s reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/). I don't think there's a lot of value by adding the docs. Maybe adding a link + context in a comment at the top should be enough.

joaquimrocha:
@illume , I have now added a story for the resource's MetadataDisplay component. Please check it out.

illume:
In 7a1bd26ebefa2e6d6858584eddaa0b5ca2e1976a I added config item so that timezone and locale are set to a default en-us/UTC when `npm run test`. So the tests pass locally, and on CI for me. I'd suggest this change would be good to apply to this PR so the tests pass.

ps. Maybe it can be fixed in a more elegant way in the i18n branch, where this date formatting function usage will be replaced with formatting through the i18next way. There's a way to also mock things with jest (and in combination with storybook as in here: https://github.com/storybookjs/storybook/issues/3809#issuecomment-592759605) as well. I'll be pushing these to the i18n branch.

pps. Perhaps we should display the timezone when showing dates? Since it's perhaps semi-confusing thing knowing which timezone a date is in (UTC, some other random server time, or browser time). I know for me, setting servers to UTC I sometimes assume the dates will be also displayed in UTC. I know deciding how date formatting should be done is probably out of scope of this PR... just thought I'd mention it.

joaquimrocha:
> In [7a1bd26](https://github.com/kinvolk/headlamp/commit/7a1bd26ebefa2e6d6858584eddaa0b5ca2e1976a) I added config item so that timezone and locale are set to a default en-us/UTC when `npm run test`. So the tests pass locally, and on CI for me. I'd suggest this change would be good to apply to this PR so the tests pass.

Cool. Hadn't noticed that.
 
> ps. Maybe it can be fixed in a more elegant way in the i18n branch, where this date formatting function usage will be replaced with formatting through the i18next way. There's a way to also mock things with jest (and in combination with storybook as in here: [storybookjs/storybook#3809 (comment)](https://github.com/storybookjs/storybook/issues/3809#issuecomment-592759605)) as well. I'll be pushing these to the i18n branch.

Let's do that later.

> pps. Perhaps we should display the timezone when showing dates? Since it's perhaps semi-confusing thing knowing which timezone a date is in (UTC, some other random server time, or browser time). I know for me, setting servers to UTC I sometimes assume the dates will be also displayed in UTC. I know deciding how date formatting should be done is probably out of scope of this PR... just thought I'd mention it.

For showing more details, I think we should do that on hover, either on the date itself or on a watch icon next to it. But yeah, I agree having a quick way to know the TZ is important when dealing with services.

joaquimrocha:
@illume , I squashed your commit with mine so we don't have a broken tests commit that we could avoid. I added you as co-author in the commit. Thanks.

PR #256, opened 2021-05-19
docs: Add platforms list

### LINKED ISSUE #248: Add matrix of tested platforms
We should maintain a list of platforms (public clouds, K8s flavors, etc.) where we have tested Headlamp or plan on testing, so anyway can contribute their experience with it (whether everything seems to work fine, any gotchas when deploying, etc.).

joaquimrocha:
We want to keep a list of platforms where we've tested Headlamp in order
to share not only works and what doesn't, but also any useful remarks
about the deployment.

@vbatts : We have tried it on Azure as well but need to do it again in
order to capture any gotchas (hence it showing as `?`).

I have also chosen to add it to the docs instead of the README as it looks out of place and a misleading (from a "tested" to a "supported" vibe somehow) there.

fixes #248

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -1,0 +1,24 @@
+---
+title: Tested Kubernetes Platforms
+linkTitle: Platforms
+weight: 300
+---
+
+This section shows the different platforms where Headlamp has been tested (in-cluster) or is intended to be tested, and useful observations about it.
+If you have tested Headlamp on a different flavor or Kubernetes, please file a PR or [issue](https://github.com/kinvolk/headlamp/issues/new/choose) to add your remarks to the list.
+
+The "works" column refers to the overall Kubernetes related functionality when running in the respective platform; it may have 3 different values:
+- ✔️ : Has been tried and works fine to the extent of what has been tested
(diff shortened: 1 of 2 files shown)


docs/platforms.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Tested Platforms
```
illume, line 2: Perhaps this title?
> Tested Kubernetes Platforms

Because, the other platforms are desktop/mobile environments eg. Ubuntu/MacOS/etc.

joaquimrocha: I did wonder if we should add the other platforms in the list as well. It would be as extensive a list, since most platform diversity is on Linux, and I think we should ultimate promote Flatpak for that.
If we're not adding the desktop platforms, then I agree with changing the title.

Let's hear from @vbatts about whether this is the format he had envisioned.

illume: Maybe related to this topic is CPU architecture... There's x64, but no armX docker images for example.

joaquimrocha:
Updated with some Azure comments/instructions + links for the platforms.

illume (approved):
👍 thanks

vbatts (approved):
LGTM

PR #257, opened 2021-05-21
Add react-i18next basics for i18n

illume:
For the issue: https://github.com/kinvolk/headlamp/issues/246

This sets up i18n using [react-i18next](https://react.i18next.com/), and uses translation in a couple of components as a test.
It disables other locales for now, only having 'en'.

- react.i18next 
- uses dynamic imports to only load the locale files that are necessary (ie, does not load 20 languages and only uses one).
- integration with Material UI locales
- integration with Monaco editor locales
- Number formatting, and date formatting can be done in an i18n way.
- some docs on how things work and how to localize components. See docs/development/i18n.md
- see src/i18n/config.ts and src/i18n/locales
- defaults to 'en', has folders and setup for other locales but only allows 'en' for now.

## Next steps in later PRs

- a language switcher GUI (probably in the sidebar or header)
- use translation in all other components, so English is done first.
- as other locales/languages are complete we add them
- improve Material UI integration: don't hard code the locales material ui supports. (maybe dynamically load, and automatically map the locales to our ones).

## How to use

Note that the last commit in this branch only allows 'en'. Try the previous commit, and change some src/i18n/locales text after changing the browser language.

As well as being setup to detect browser language and use that locale if it exists, one other way is to change the URL: `?lng=de`. 

In development mode i18next debug mode is on, so it prints some diagnostic messages in the console.

To see the Monaco language, edit a resource and then press alt+shift+F1 to see the accessibility instructions in the selected language.

## Testing done

Try changing the browser language to see the browser language detection work (see dev console).
Loads other locale data when loading that locale, not all at once.
Browser URL locale switching works.
Regression testing still passes, and the changed components look ok in the app.


illume:
There are two other ways we could do this.

1. keep languages in `public/i18n/{locale}.json` and have them lazy loaded. The advantage of this is that all of the languages are not loaded even if they are not used. This is important to keep the page weight down when it's being used as a website (as apposed to an electron app).
2. Have the localizations near the components. `Component/i18n/{locale}.json`. This is good for developers of components, and for code-savy translators. Because it's easier to translate something if you can more easily know what component it is for. It's also easy to find. However, then the translations are spread out all over the code base - not in one place. This makes it harder for people to go to one place and do a translation.

It seems we decided to have localizations contributed and managed through github, and not use some translation workflow management tool.

joaquimrocha:
I'd be more inclined to approach 2 or 3. I understand that we don't have a lot of text but approach 2 seems to be very similar in terms of effort, and in a state where it'll not hinder the performance should we add more text/translations in the future.

Approach 3 seems good if we end up extracting the core of the common components into their own lib. But I think this is less important, or more difficult to preview, at the moment.

illume:
I'm more leaning towards method 2.

- An estimate of the size based on only 5 languages shows it would add significant overhead not only to users on the web, but to development speed because of the extra time typescript/webpack take.
- it's more compatible with translation services (like phrase app) if we want to translate to more languages later.
- separating files by namespace makes it pretty easy enough to find things and keep things separate by components.

I've pushed the Component.i18n way... but have been changing it locally to method 2.

frontend/public/locales/README.md
```diff
@@ -0,0 +1,127 @@
+# i18n Internationali[sz]ed Locali[sz]ation
```
joaquimrocha, line 1: It's a good joke that people like me will miss and point out :facepalm: 

```suggestion
# Internationalization / Localization
```

frontend/public/locales/README.md
```diff
@@ -0,0 +1,127 @@
...
+      numReady: podsReady.length,
+      numItems: items.length,
+    });
+```
+
+Number formatting is being done with [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat).
+
+## Date formatting
+
+TODO: not implmented.
```
joaquimrocha, line 95: lil typo here. And below as well.

frontend/public/locales/config.ts
```diff
@@ -0,0 +1,56 @@
...
+//   en: {
+//     cluster,
+//   },
+//   de: {
+//     clusterDe,
+//   },
+//   es: {
+//     clusterEs,
+//   },
+// } as const;
```
joaquimrocha, line 20: Can we make this somehow more automatic? (maybe listing the folder, or at least automating the `cluster{LANG}` resource imports/uses)

frontend/public/locales/es/cluster.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Charts": {
+    "PodsStatusCircleChart": {
+      "title": "Podso"
```
joaquimrocha, line 4: "El Podso" :joy:

frontend/src/i18n/README.md
```diff
@@ -0,0 +1,111 @@
+# i18n Internationali[sz]ed Locali[sz]ation
```
joaquimrocha, line 1: Did you forget to delete this one?

joaquimrocha (commented):
Looks good to me overall.
One thing that I think it's missing from the docs (or maybe you intended to add later) is how to work on a new or existing locale (you do have instructions of how things work, but not a guide IIUC).

Also, maybe we move the locale's README to a docs/ article and have the README just intro + link there.

illume:
> Also, maybe we move the locale's README to a docs/ article and have the README just intro + link there.

Done.

The main problem I was having was with the static file loader and with things like storybook, typechecking, testing, and also requiring using React.Suspense(which is clunky and still marked experimental). With all those problems, I switched it to using dynamic modules... so webpack generates a JS file for each locale - and everything is nicely integrated it seems. I haven't finished converting all the components to use translations yet however, I think this can come in later PRs. 

Use dynamic imports (webpack code splitting) to load javascript bundles:
- https://www.i18next.com/misc/creating-own-plugins#backend
- https://webpack.js.org/guides/code-splitting/

PR #268, opened 2021-06-14
Change i18n from logical keys to "textual" ones, add app localization, add PT/ES localizations

joaquimrocha:
These changes move the i18n keys from logical ones (this.title) to textual ones (This title). The reason is that:
1. We get to see the actual human string in the code (maybe be useful for grepping, etc.);
2. We get to see what we're translating in the very translation file (no need to look into one source language file).
3. We don't mix keys/textual when using the `<Trans>` tag.
4. We instantly update any strings needing changes because the keys change when the text changes.

Also adds:
- a new command to create/update the i18n json files: `make i18n`. For creating new translations, create a locale directories with the new language code + run the `make i18n` command.
- Spanish and Portuguese localizations
- Electron app localizations (including menus)
- A menu in the app to select the localization/language

Screenshot 2021-06-21 at 12 01 29


joaquimrocha:
@illume , This is very WIP ATM, but I'd love your opinion on the direction (one file does use `<Trans>` in blocks that have no inner tags for example).

illume:
Like it. Much more readable.

The downside of using strings as keys are if there are big strings, or if the strings change a lot. That's not the case with Headlamp, and if it becomes that way we could always switch back to keys then.

joaquimrocha:
I have updated the catalog now with a few more namespaces. I will finish the PT version if it looks good, but we need to add a language picker as well.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -411,13 +416,14 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
...
   const { resource } = props;
+  const { t } = useTranslation('glossary');
 
   if (!resource) {
     return null;
   }
 
   return (
-    <SectionBox title="Conditions">
+    <SectionBox title={t('Containers')}>
```
illume, line 426: Is this supposed to be Conditions, and not Containers?

frontend/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,742 @@
+import circleSlice2 from '@iconify/icons-mdi/circle-slice-2';
```
illume, line 1: There's a Sidebar/Sidebar.tsx now from the a11y branch.

illume: I removed this from the commit where the old Sidebar.tsx was added (`frontend: Mark all user-facing strings as translatable`) and rebased.

joaquimrocha:
Status: We should make the app translatable as well and hopefully joining the translations as well (i.e. have them all in the frontend/ if possible).
One thing that's needed for knowing (in Electron) when the user changes the locale (in React) is to use the IP (electron.ipcRenderer) to exchange that info, but it seems we need changes in order to make electron available to the app.
I did a few attempts (e.g. based on [this one](https://phrase.com/blog/posts/building-an-electron-app-with-internationalization-i18n/)) but still no luck so I'll leave it for tomorrow.

illume (approved):
👍 🎉

illume:
I updated the PR description with the latest PR additions and a screenie.

PR #271, opened 2021-06-18
Use Browser or OS light/dark theme preference

illume:
If the user makes a preference by selecting the theme from the UI, then it uses that preference instead. Once the user preference is selected it always uses that preference from then on.
    
The localStorage key was prefixed with headlamp to try and avoid nameclashes. The older key was ignored, since it was set without a user selecting a preference. This way the browser/OS preference is used until they change it manually.
    
Also make the theme preference change reactive. So it changes when the os/browser preference changes.

https://user-images.githubusercontent.com/9541/122554287-13de9680-d039-11eb-905d-08b9e5b6bdd7.mov

To test, use the browser/OS to set a darkmode/lightmode theme preference. There is no way to clear the user preference except to delete the local storage "themePreference" key.

For the [Accessibility improvements](https://github.com/kinvolk/headlamp/issues/245) issue.

### DIFF
--- a/frontend/src/components/common/Resource/EditorDialog.tsx
+++ b/frontend/src/components/common/Resource/EditorDialog.tsx
@@ -22,6 +22,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import getDocDefinitions from '../../../lib/docs';
 import { KubeObjectInterface } from '../../../lib/k8s/cluster';
+import { getThemeName } from '../../../lib/themes';
 import ConfirmButton from '../ConfirmButton';
 import Empty from '../EmptyContent';
 import Loader from '../Loader';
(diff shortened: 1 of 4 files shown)


frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -205,7 +208,7 @@ export default function EditorDialog(props: EditorDialogProps) {
       <Box paddingTop={2} height="100%">
         <Editor
           language="yaml"
-          theme="vs-dark"
+          theme={themeName === 'dark' ? 'vs-dark' : 'light'}
```
joaquimrocha, line 211: The reason why we hardcoded the dark one was that Monaco Ed was not updating/behaving correctly when we switched the theme. Please double-check this before we merge this change in.

joaquimrocha (approved):
Good stuff!
Please check my comment about the Monaco Ed. not getting updated when switching themes. If that's still the case, please keep the dark variant of the editor as before.
If it all works fine, feel free to merge.

frontend/src/components/common/Resource/EditorDialog.tsx
illume: It updates the theme and works correctly now. I guess the last monaco dependency update must have fixed it.

PR #272, opened 2021-06-22
Fix Electron integration, show lang selector on front page, generate LocaleSelect options from config

illume:
# Fix Electron integration, show lang selector on front page, generate LocaleSelect options from config

## Testing done

- With pt selected, when starting app fresh the pt locale is used for menu items and other parts
- there is a locale selecter on the front page (under the cluster list)

Screenshot 2021-06-22 at 09 42 59

### DIFF
--- a/frontend/src/i18n/LocaleSelect/LocaleSelect.tsx
+++ b/frontend/src/i18n/LocaleSelect/LocaleSelect.tsx
@@ -4,7 +4,6 @@ import Select from '@material-ui/core/Select';
 import { createStyles, makeStyles, Theme, useTheme } from '@material-ui/core/styles';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import helpers from '../../helpers';
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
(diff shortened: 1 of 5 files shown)


illume:
I think it's better to leave a label off the locale selector... but unsure. What do you think?

Screenshot 2021-06-22 at 09 45 31

frontend/src/i18n/electronI18n.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/** Integration with Electron app, so it can change locale information. */
+export function useElectronI18n() {
+  const { i18n } = useTranslation();
+
+  useEffect(() => {
+    i18n.on('languageChanged', tellAppAboutLanguage);
+    return () => {
+      i18n.off('languageChanged', tellAppAboutLanguage);
```
joaquimrocha, line 21: I chose to do it explicitly when the language was changed or on load because I thought we'd also have the languages at some point added to the app menu and then we could end up in calling the language set up circularly if we have it react to the language changed.
In any case this looks cleaner and we can look at the issue I described above when we have it.

frontend/src/i18n/LocaleSelect/LocaleSelect.tsx
```diff
@@ -39,9 +39,10 @@ export default function LocaleSelect(props: LocaleSelectProps) {
         onChange={changeLng}
         inputProps={{ 'aria-label': t('Select locale') }}
       >
-        <MenuItem value={'en'}>en</MenuItem>
-        <MenuItem value={'pt'}>pt</MenuItem>
-        <MenuItem value={'es'}>es</MenuItem>
+        {i18n?.options?.supportedLngs &&
+          i18n.options.supportedLngs
+            .filter(lng => lng !== 'cimode')
+            .map(lng => <MenuItem value={lng}>{lng}</MenuItem>)}
```
joaquimrocha, line 49: Yup! I wanted to replace this and completely forgot. Thanks.

joaquimrocha (approved):

PR #274, opened 2021-06-22
Add a LocaleSelect to AuthChooser

illume:
... so there is a way to change the locale if only the auth chooser dialog is chosen.


joaquimrocha (approved):
LGTM. Thanks a lot!

PR #281, opened 2021-06-28
Use ghcr.io instead of quay.io for Headlamp's container image

### LINKED ISSUE #280: Move images to Github
We want to move all our images to Github's container registry. This implies:
- [X] Create the package in the kinvolk org
- [ ] Change the workflows/dependencies to point there.

joaquimrocha:
fixes #280

### DIFF
--- a/.github/workflows/container-publish.yml
+++ b/.github/workflows/container-publish.yml
@@ -2,7 +2,7 @@ name: Publish container
 on:
   push:
     tags:
-    - '*'
+    - 'v*'
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: kinvolk/headlamp
--- a/.github/workflows/container-publish.yml
+++ b/.github/workflows/container-publish.yml
@@ -3,21 +3,34 @@ on:
   push:
     tags:
     - '*'
+env:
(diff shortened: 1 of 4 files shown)


illume:
How do we handle the transition for people using the old image location? Can we just mention it's moved in the release notes?

I tried to test it with:
`docker run --network="host" -p 127.0.0.1:4466:4466/tcp --mount type=bind,source="$HOME/.kube",target=/root/.kube ghcr.io/kinvolk/headlamp:latest /headlamp/server -html-static-dir /headlamp/frontend -plugins-dir=/headlamp/plugins`

... and get the error message:
```
docker: Error response from daemon: Head https://ghcr.io/v2/kinvolk/headlamp/manifests/latest: unauthorized.
```

Do we need to [Configure a package's access control and visibility](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility)?

illume (approved):
👍

joaquimrocha:
> Do we need to [Configure a package's access control and visibility](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility)?

Good catch! Fixed.

PR #283, opened 2021-07-07
Security updates

joaquimrocha:
- frontend: Update refractor for fixing security issues
- frontend: Update prismjs for security fixes
- app: Update glob-parent for security fixes
- plugins/headlamp-plugin: Update packages with npm audit fix


illume (approved):
thanks 👍

PR #289, opened 2021-07-15
Add update popup and release notes

### LINKED ISSUE #260: Show release notes on app whenever a new release is installed
## Current situation

We ping github releases to look if there is a new version of the app  and whenever a new version is found we auto download it,
A good way to show the user that the app was updated and the new fixes or feature which are added would be via a release note modal.

## Ideal future situation
A good way to communicate with user about  what all changes the app has gone through so that he/she doesn't have to go 
 to github and see what all fixes were done.

## **Implementation options
autoUpdater.on('update-downloaded', (info) => {
  //Get release notes 
  info.releaseNotes (add it to some modal that can be displayed on the main screen)
})

ashu8912:
fixes #260

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -7,6 +7,7 @@ import Store from 'electron-store';
 import { i18n as I18n } from 'i18next';
 import open from 'open';
 import path from 'path';
+import semver from 'semver';
 import url from 'url';
 import yargs from 'yargs';
 import i18n from './i18next.config';
@@ -295,15 +296,19 @@ function startElecron() {
     mainWindow.webContents.on('dom-ready', () => {
       const octokit = new Octokit();
 
-      async function fetchLatestRelease() {
-        const response = await octokit.request('GET /repos/{owner}/{repo}/releases/latest', {
+      async function fetchRelease() {
+        const githubReleaseURL = `GET /repos/{owner}/{repo}/releases`;
+        // get me all the releases -> default decreasing order of releases
+        const response = await octokit.request(githubReleaseURL, {
           owner: 'kinvolk',
           repo: 'headlamp',
         });
-
-        if (response.data.name !== appVersion) {
+        const latestRelease = response.data.find(
+          release => !release.name.startsWith('headlamp-helm')
+        );
+        if (semver.gt(latestRelease.name, appVersion) && !process.env.FLATPAK_ID) {
           mainWindow.webContents.send('update_available', {
-            downloadURL: response.data.html_url,
+            downloadURL: latestRelease.html_url,
           });
         }
         /*
@@ -313,14 +318,23 @@ function startElecron() {
         const storedAppVersion = store.get('app_version');
         if (!storedAppVersion) {
           store.set('app_version', appVersion);
-        } else if (storedAppVersion !== appVersion) {
-          mainWindow.webContents.send('showReleaseNotes', { releaseNotes: response.data.body });
+        } else if (semver.lt(storedAppVersion as string, appVersion)) {
+          // get the release notes for the version with which the app was built with
+          const githubReleaseURL = `GET /repos/{owner}/{repo}/releases/tags/v${appVersion}`;
+          const response = await octokit.request(githubReleaseURL, {
+            owner: 'kinvolk',
+            repo: 'headlamp',
+          });
+          mainWindow.webContents.send('showReleaseNotes', {
+            releaseNotes: response.data.body,
+            appVersion,
+          });
           // set the store version to latest so that we don't show release notes on
           // every start of app
           store.set('app_version', appVersion);
         }
       }
-      fetchLatestRelease();
+      fetchRelease();
     });
 
     mainWindow.on('closed', () => {
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,8 +1,8 @@
+import { Octokit } from '@octokit/core';
 import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
-import { autoUpdater } from 'electron-updater';
 import { i18n as I18n } from 'i18next';
 import open from 'open';
 import path from 'path';
@@ -288,6 +288,27 @@ function startElecron() {
       },
     });
     mainWindow.loadURL(startUrl);
+
+    mainWindow.webContents.on('dom-ready', () => {
+      const octokit = new Octokit();
+      const appVersion = app.getVersion();
+
+      async function fetchLatestRelease() {
+        const response = await octokit.request('GET /repos/{owner}/{repo}/releases/latest', {
+          owner: 'kinvolk',
+          repo: 'headlamp',
+        });
+
+        if (response.data.name !== appVersion) {
+          mainWindow.webContents.send('update_available', {
+            downloadURL: response.data.html_url,
+            releaseNotes: response.data.body,
+          });
+        }
+      }
+      fetchLatestRelease();
+    });
+
     mainWindow.on('closed', () => {
       mainWindow = null;
     });
@@ -302,37 +323,11 @@ function startElecron() {
       }
     });
 
-    autoUpdater.checkForUpdatesAndNotify();
-
     if (!isDev) {
       serverProcess = startServer();
       attachServerEventHandlers(serverProcess);
     }
   }
-  autoUpdater.autoDownload = true;
-
-  function sendStatusToWindow(text: string) {
-    log.info(`Sending status to window: ${text}`);
-    if (mainWindow) {
-      mainWindow.webContents.send('message', text);
-    }
-  }
-
-  autoUpdater.on('update-not-available', () => {
-    sendStatusToWindow('Update not available.');
-  });
-
-  autoUpdater.on('checking-for-update', () => {
-    sendStatusToWindow('Checking for update...');
-  });
-
-  autoUpdater.on('update-available', () => {
-    sendStatusToWindow('Update is available');
-  });
-
-  autoUpdater.on('update-downloaded', () => {
-    sendStatusToWindow('Update downloaded');
-  });
 
   if (disableGPU) {
     log.info('Disabling GPU hardware acceleration.');
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -282,8 +282,9 @@ function startElecron() {
       width,
       height,
       webPreferences: {
-        nodeIntegration: true,
-        contextIsolation: false,
+        nodeIntegration: false,
+        contextIsolation: true,
+        preload: `${__dirname}/preload.js`,
       },
     });
     mainWindow.loadURL(startUrl);
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -303,6 +303,10 @@ function startElecron() {
       return { action: 'deny' };
     });
 
+    ipcMain.on('disableUpdateChecking', () => {
+      store.set('disable_update_check', true);
+    });
+
     mainWindow.webContents.on('dom-ready', () => {
       const octokit = new Octokit();
 
@@ -344,7 +348,10 @@ function startElecron() {
           store.set('app_version', appVersion);
         }
       }
-      fetchRelease();
+      const isUpdateCheckingDisabled = store.get('disable_update_check');
+      if (!isUpdateCheckingDisabled) {
+        fetchRelease();
+      }
     });
 
     mainWindow.on('closed', () => {
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,6 +1,6 @@
 import { Octokit } from '@octokit/core';
 import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
-import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen } from 'electron';
+import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
 import Store from 'electron-store';
@@ -292,6 +292,16 @@ function startElecron() {
       },
     });
     mainWindow.loadURL(startUrl);
+
+    mainWindow.webContents.setWindowOpenHandler(({ url }) => {
+      // allow all urls starting with app startUrl to open in electron
+      if (url.startsWith(startUrl)) {
+        return { action: 'allow' };
+      }
+      // otherwise open url in a browser and prevent default
+      shell.openExternal(url);
+      return { action: 'deny' };
+    });
 
     mainWindow.webContents.on('dom-ready', () => {
       const octokit = new Octokit();
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -3,6 +3,7 @@ import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
+import Store from 'electron-store';
 import { i18n as I18n } from 'i18next';
 import open from 'open';
 import path from 'path';
@@ -10,6 +11,8 @@ import url from 'url';
 import yargs from 'yargs';
 import i18n from './i18next.config';
 
+const appVersion = app.getVersion();
+const store = new Store();
 const args = yargs
   .option('headless', {
     describe: 'Open Headlamp in the default web browser instead of its app window',
@@ -291,7 +294,6 @@ function startElecron() {
 
     mainWindow.webContents.on('dom-ready', () => {
       const octokit = new Octokit();
-      const appVersion = app.getVersion();
 
       async function fetchLatestRelease() {
         const response = await octokit.request('GET /repos/{owner}/{repo}/releases/latest', {
@@ -302,8 +304,20 @@ function startElecron() {
         if (response.data.name !== appVersion) {
           mainWindow.webContents.send('update_available', {
             downloadURL: response.data.html_url,
-            releaseNotes: response.data.body,
           });
+        }
+        /*
+  check if there is already a version in store if it exists don't store the current version
+  this check will help us later in determining whether we are on the latest release or not
+  */
+        const storedAppVersion = store.get('app_version');
+        if (!storedAppVersion) {
+          store.set('app_version', appVersion);
+        } else if (storedAppVersion !== appVersion) {
+          mainWindow.webContents.send('showReleaseNotes', { releaseNotes: response.data.body });
+          // set the store version to latest so that we don't show release notes on
+          // every start of app
+          store.set('app_version', appVersion);
         }
       }
       fetchLatestRelease();
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -339,8 +339,9 @@ function startElecron() {
             owner: 'kinvolk',
             repo: 'headlamp',
           });
+          const [releaseNotes] = response.data.body.split('<!-- end-release-notes -->');
           mainWindow.webContents.send('showReleaseNotes', {
-            releaseNotes: response.data.body,
+            releaseNotes,
             appVersion,
           });
           // set the store version to latest so that we don't show release notes on
--- a/app/electron/preload.ts
+++ b/app/electron/preload.ts
@@ -4,13 +4,23 @@ import { contextBridge, ipcRenderer } from 'electron';
 contextBridge.exposeInMainWorld('desktopApi', {
   send: (channel, data) => {
     // allowed channels
-    const validChannels = ['updateAvailable', 'locale', 'showReleaseNotes'];
+    const validChannels = [
+      'updateAvailable',
+      'locale',
+      'showReleaseNotes',
+      'disableUpdateChecking',
+    ];
     if (validChannels.includes(channel)) {
       ipcRenderer.send(channel, data);
     }
   },
   receive: (channel, func) => {
-    const validChannels = ['updateAvailable', 'locale', 'showReleaseNotes'];
+    const validChannels = [
+      'updateAvailable',
+      'locale',
+      'showReleaseNotes',
+      'disableUpdateChecking',
+    ];
     if (validChannels.includes(channel)) {
       // Deliberately strip event as it includes `sender`
       ipcRenderer.on(channel, ({}, ...args) => func(...args));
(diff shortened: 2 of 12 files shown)


app/electron/main.ts
```diff
@@ -282,11 +282,33 @@ function startElecron() {
...
-        nodeIntegration: true,
-        contextIsolation: false,
+        nodeIntegration: false,
+        contextIsolation: true,
+        preload: `${__dirname}/preload.js`
       },
     });
     mainWindow.loadURL(startUrl);
+
+    mainWindow.webContents.on('dom-ready', () => {
```
illume, line 292: Could this be outside the app, so it can run for the website version too? Or should it only be visible for the app version?

frontend/package.json
```diff
@@ -59,6 +61,7 @@
     "redux-saga": "^1.1.3",
     "rimraf": "^3.0.2",
     "semver": "^7.3.4",
+    "styled-components": "^5.3.0",
```
illume, line 64: I'm not sure if we should add styled-components. Since we're already using the jss way everywhere else.

frontend/src/App.tsx
```diff
@@ -245,49 +271,80 @@ function AppContainer() {
...
-                <RouteSwitcher />
-              </Container>
-            </Box>
-          </main>
-          <ActionsNotifier />
-        </Box>
-      </Router>
-    </SnackbarProvider>
+    <>
+      <Modal open={showReleaseNotes} BackdropComponent={Backdrop} style={modalStyle}>
```
illume, line 287: I feel this should be in a separate component.

frontend/src/components/updatedesktopapp/Popup.tsx
```diff
@@ -0,0 +1,88 @@
+import simpleClose from '@iconify/icons-mdi/close';
+import informationOutline from '@iconify/icons-mdi/information-circle-outline';
+import { InlineIcon } from '@iconify/react';
+import styled from 'styled-components';
+
+const PopupContainer = styled.div`
```
illume, line 6: Not sure about using styled-components as well.

illume:
What do you think about only showing this once per release? Perhaps storing the "I have seen the release notes" version in localStorage when they close the modal.

illume:
IMHO, I think for all new components we should add storybook files.

app/electron/main.ts
ashu8912: I am not sure do we want to show whether there is a release or not on web version?

ashu8912: cc @joaquimrocha

joaquimrocha: We don't. At least for now I don't think it's a good UX to just show the updates for the web version.

For the web I imagine we could have a notification mentioning there is a new version available, with a link for refreshing (similar to WhatsApp web does).
However, showing the release notes is a UX that at least I'd like to give control to the operators for whether to show them or not, and until we do that, I prefer not to have it shown.

joaquimrocha:
@ashu8912 Can you share a gif of the update UX, and instructions on how to simulate the update?
I am not 100% sure since I haven't tried it yet but I think the update notification should probably follow the snackbar notifications we have for cluster actions.

frontend/src/App.tsx
```diff
@@ -233,9 +235,49 @@ function ThemeChangeButton() {
   );
 }
 
+function UpdatePopup() {
```
illume, line 238: Can this also go inside a folder with the other related module? Maybe `releases/UpdateSnack`?

frontend/src/App.tsx
```diff
@@ -245,13 +287,23 @@ function AppContainer() {
 
   localStorage.setItem('sidebar', JSON.stringify({ shrink: isSidebarOpen }));
 
+  React.useEffect(() => {
```
illume, line 290: This could be encapsulated with the other releases based code?

joaquimrocha:
Hi @ashu8912 , I have just tried the builds you did for testing this new feature.
Nice work. Here's a couple of comments:
* When clicking More, we should open the system's browser and not a window of the app with the new release. (The reason is that it feels more natural + the user can stop Headlamp at any point without stopping the ongoing download of the new version + the user can see the progress for the download if it's using the regular browser)
* Since the notification keeps getting shown, maybe we need a way to shut it forever. How about we make the buttons as:
```
[ More ] [ Do not notify again ] [ X ]
```
X should be a cross/close icon.

This should be disabled for flatpak BTW. You can maybe just disable this if the `FLATPAK_ID` env var is set ().

joaquimrocha:
(Pressed the button too soon.)

I am not confident on the workflow for keeping the release diff which is being shown when the app's new version is first run.
IMO, until we have a better plan for showing the release notes within the app, I think we should just show a notification (once!) saying "You have upgraded to version X.Y.Z [ Close ]"
X.Y.Z is a link to the release notes for the tag on github, again opening in the external browser.

illume (approved):
👍 nice one

frontend/src/App.tsx
```diff
@@ -33,6 +33,7 @@ import {
 import { ClusterTitle } from './components/cluster/Chooser';
 import ActionsNotifier from './components/common/ActionsNotifier';
 import AlertNotification from './components/common/AlertNotification';
+import ReleaseNotesModal from './components/releasenotes';
```
joaquimrocha, line 35: Nitpick on title: typo on `separate`

app/electron/main.ts
```diff
@@ -313,14 +313,20 @@ function startElecron() {
         const storedAppVersion = store.get('app_version');
         if (!storedAppVersion) {
           store.set('app_version', appVersion);
-        } else if (storedAppVersion !== appVersion) {
+        } else if (parseInt(storedAppVersion as string) <= parseInt(appVersion)) {
```
joaquimrocha: This will not do a correct comparison for the versions as `parseInt` returns `2` for `2.1.0`. There's a `semver` package that can be used for the comparison.

app/package.json
```diff
@@ -68,6 +68,7 @@
     },
     "files": [
       "electron/main.js",
+      "electron/preload.js",
```
joaquimrocha, line 71: Add this change to the commit that introduces the file.

app/electron/main.ts
```diff
@@ -315,7 +315,7 @@ function startElecron() {
           owner: 'kinvolk',
           repo: 'headlamp',
         });
-        if (response.data.name !== appVersion) {
+        if (response.data.name >= appVersion && !response.data.name.startsWith('headlamp-helm')) {
```
joaquimrocha: This means the update_available will only work until a chart is published. The logic for getting the latest release needs to be to get the latest N releases (or all, if there's no such API) and get the latest in the list that's starting with `v*`.

app/electron/main.ts
```diff
@@ -334,7 +338,10 @@ function startElecron() {
             owner: 'kinvolk',
             repo: 'headlamp',
           });
-          mainWindow.webContents.send('show_release_notes', { releaseNotes: response.data.body, appVersion });
+          mainWindow.webContents.send('show_release_notes', {
+            releaseNotes: response.data.body,
+            appVersion,
+          });
```
joaquimrocha: nitpick: This change seems to be unrelated.

frontend/src/App.tsx
```diff
@@ -4,7 +4,7 @@ import logoutIcon from '@iconify/icons-mdi/logout';
 import darkIcon from '@iconify/icons-mdi/weather-night';
 import lightIcon from '@iconify/icons-mdi/weather-sunny';
 import { Icon } from '@iconify/react';
-import { Box, Button } from '@material-ui/core';
+import { Box } from '@material-ui/core';
```
joaquimrocha: Nitpick: Typo on `separate` in the title.

joaquimrocha (changes requested):
I have added a few nitpicks and comments.
I realized only after a few commits that I commented on code that was changed later into the comments I was adding. I decided not to go back and delete them because that'd take a bit, so please ignore them.

This can be avoided by making sure that PRs do not update code that they have introduced themselves.

app/tsconfig.json
```diff
@@ -15,7 +15,7 @@
       "module": "ESNext",
       "moduleResolution": "node",
       "noFallthroughCasesInSwitch": true,
-      "outDir": "electron/src/",
+      "outDir": "electron/",
```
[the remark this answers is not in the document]
ashu8912: no the outDir flag on the babel script does work, it picks up the babel one always so we actually don't need an outDir property here

app/package.json
ashu8912: done

app/electron/main.ts
```diff
@@ -302,9 +304,21 @@ function startElecron() {
...
+  this check will help us later in determining whether we are on the latest release or not
+  */
+        const storedAppVersion = store.get('app_version');
+        if (!storedAppVersion) {
+          store.set('app_version', appVersion);
+        } else if (storedAppVersion !== appVersion) {
+          mainWindow.webContents.send('show_release_notes', { releaseNotes: response.data.body });
+          // set the store version to latest so that we don't show release notes on
+          // every start of app
+          store.set('app_version',appVersion);
```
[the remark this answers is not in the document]
ashu8912: fixed, yes the linter doesn't seem to take this into consideration

app/electron/main.ts
```diff
@@ -302,9 +304,21 @@ function startElecron() {
         if (response.data.name !== appVersion) {
           mainWindow.webContents.send('update_available', {
             downloadURL: response.data.html_url,
-            releaseNotes: response.data.body,
           });
         }
+        /*
+  check if there is already a version in store if it exists don't store the current version
+  this check will help us later in determining whether we are on the latest release or not
+  */
```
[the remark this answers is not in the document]
ashu8912: it's the prettier that is formatting it this way

frontend/package.json
```diff
@@ -47,6 +47,7 @@
     "react-dom": "^17.0.1",
     "react-hotkeys-hook": "^3.0.3",
     "react-i18next": "^11.11.1",
+    "react-markdown": "^6.0.2",
```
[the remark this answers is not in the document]
ashu8912: done

frontend/src/App.tsx
ashu8912: done

frontend/src/App.tsx
ashu8912: fixed

app/electron/main.ts
ashu8912: fixed

app/electron/main.ts
```diff
@@ -302,9 +304,21 @@ function startElecron() {
...
+        const storedAppVersion = store.get('app_version');
+        if (!storedAppVersion) {
+          store.set('app_version', appVersion);
+        } else if (storedAppVersion !== appVersion) {
+          mainWindow.webContents.send('show_release_notes', { releaseNotes: response.data.body });
```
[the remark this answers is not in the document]
ashu8912: done

app/electron/preload.ts
```diff
@@ -0,0 +1,19 @@
...
+contextBridge.exposeInMainWorld('desktopApi', {
+  send: (channel, data) => {
+    // whitelist channels
+    const validChannels = ['update_available'];
+    if (validChannels.includes(channel)) {
+      ipcRenderer.send(channel, data);
+    }
+  },
+  receive: (channel, func) => {
+    const validChannels = ['update_available'];
```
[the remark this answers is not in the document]
ashu8912: done

app/electron/preload.ts
```diff
@@ -0,0 +1,19 @@
+import { contextBridge, ipcRenderer } from 'electron';
+// Expose protected methods that allow the renderer process to use
+// the ipcRenderer without exposing the entire object
+contextBridge.exposeInMainWorld('desktopApi', {
+  send: (channel, data) => {
+    // whitelist channels
```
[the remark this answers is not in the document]
ashu8912: fixed

app/electron/preload.ts
```diff
@@ -0,0 +1,19 @@
+import { contextBridge, ipcRenderer } from 'electron';
+// Expose protected methods that allow the renderer process to use
+// the ipcRenderer without exposing the entire object
+contextBridge.exposeInMainWorld('desktopApi', {
+  send: (channel, data) => {
+    // allowed channels
+    const validChannels = ['updateAvailable', 'locale'];
```
joaquimrocha: updateAvailable doesn't really belong to this patch though, as it's only used later.

app/package.json
```diff
@@ -108,6 +108,7 @@
     "@babel/runtime": "^7.12.5",
     "@octokit/core": "^3.5.1",
     "electron-log": "^4.2.4",
+    "electron-store": "^8.0.0",
```
joaquimrocha, line 112: This should be squashed with the patch that uses the store.

app/electron/main.ts
```diff
@@ -3,13 +3,16 @@ import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
+import Store from 'electron-store';
```
joaquimrocha, line 6: The commit title references saving the app version but the important part of this commit is that it sends the release info to the renderer. So maybe reflect this in the title to make it a bit easier to understand.

frontend/src/App.tsx
```diff
@@ -264,6 +264,15 @@ function UpdatePopup() {
...
             More
           </Button>
+          <Button
+            color="inherit"
+            onClick={() => {
+              desktopApi.send('disableUpdateChecking');
+              setShow(false);
+            }}
+          >
+            Do not notify again
```
joaquimrocha: This needs to be translatable.

frontend/src/App.tsx
```diff
@@ -233,6 +234,44 @@ function ThemeChangeButton() {
...
+        'aria-describedby': 'updatePopup',
+      }}
+      message={`There is an available update`}
+      action={
+        <React.Fragment>
+          <Button color="secondary" href={`${updateDownloadURL}`} target="_blank">
+            More
+          </Button>
+          <Button color="primary" onClick={() => setShow(false)}>
+            Close
```
joaquimrocha: Forgot to notice in my last review but these need to be translatable.

frontend/src/App.tsx
```diff
@@ -233,6 +234,44 @@ function ThemeChangeButton() {
...
+      anchorOrigin={{
+        vertical: 'bottom',
+        horizontal: 'right',
+      }}
+      open={show}
+      autoHideDuration={100000}
+      ContentProps={{
+        'aria-describedby': 'updatePopup',
+      }}
+      message={`There is an available update`}
```
joaquimrocha: This needs to be translatable.

joaquimrocha: Also please change it to "An update is available"

app/electron/main.ts
```diff
@@ -295,13 +296,17 @@ function startElecron() {
     mainWindow.webContents.on('dom-ready', () => {
       const octokit = new Octokit();
 
-      async function fetchLatestRelease() {
-        const response = await octokit.request('GET /repos/{owner}/{repo}/releases/latest', {
+      async function fetchRelease() {
+        const githubReleaseURL = `GET /repos/{owner}/{repo}/releases/latest`;
+        const response = await octokit.request(githubReleaseURL, {
```
joaquimrocha: This commit should probably be squashed with the one that introduces this code, because without it the original code is broken. So even though these are complex changes, the functionality needs to be there from the beginning or we may try a broken commit later.

app/electron/main.ts
```diff
@@ -311,18 +311,19 @@ function startElecron() {
       const octokit = new Octokit();
 
       async function fetchRelease() {
-        const githubReleaseURL = `GET /repos/{owner}/{repo}/releases/latest`;
+        const githubReleaseURL = `GET /repos/{owner}/{repo}/releases`;
+        // get me all the releases -> default decreasing order of releases
```
joaquimrocha, line 315: These changes should also be in the original commit that introduces the logic.

If you want to highlight them, you can use the commit body.

app/electron/main.ts
```diff
@@ -295,13 +296,17 @@ function startElecron() {
...
+        const response = await octokit.request(githubReleaseURL, {
           owner: 'kinvolk',
           repo: 'headlamp',
         });
-
-        if (response.data.name !== appVersion) {
+        if (
+          semver.gt(response.data.name, appVersion) &&
+          !response.data.name.startsWith('headlamp-helm') &&
+          !process.env.FLATPAK_ID
```
joaquimrocha: I am fine having this here but you have a commit where it mentions not showing it for Flatpak.
So maybe remove the flatpak dedicated commit and add a code comment here to make it more clear what this does.

frontend/src/components/updatepopup/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+            onClick={() => {
+              desktopApi.send('disableUpdateChecking');
+              setShow(false);
+            }}
+          >
+            Do not notify again
+          </Button>
+          <Button color="primary" onClick={() => setShow(false)}>
+            Close
+          </Button>
```
joaquimrocha, line 44: These have strings that need to be translated.

frontend/src/components/releasenotes/index.tsx
```diff
@@ -24,7 +24,9 @@ export default function ReleaseNotesModal(props: ReleaseNotesModalProps) {
   const releaseNotesStyle = {
     padding: '1rem',
     maxHeight: '80%',
-    overflow: 'scroll',
+    minWidth: '50%',
+    minHeight: '50%',
+    overflow: 'auto',
```
joaquimrocha, line 29: Also add this to the code that introduces the window.

app/electron/main.ts
```diff
@@ -339,8 +339,9 @@ function startElecron() {
             owner: 'kinvolk',
             repo: 'headlamp',
           });
+          const [releaseNotes] = response.data.body.split('<!-- end-release-notes -->');
```
joaquimrocha, line 342: :+1:

joaquimrocha (changes requested):
It's looking very close but still a few things to tackle.

app/electron/main.ts
```diff
@@ -282,8 +282,8 @@ function startElecron() {
       width,
       height,
       webPreferences: {
-        nodeIntegration: true,
-        contextIsolation: false,
+        nodeIntegration: false,
+        contextIsolation: true,
```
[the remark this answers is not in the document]
ashu8912: squashed the preload integration with this commit

app/electron/preload.ts
ashu8912: moved to the right place

app/electron/main.ts
ashu8912: highlighted the important stuff first

frontend/src/App.tsx
ashu8912: fixed

frontend/src/components/releasenotes/index.tsx
ashu8912: done

frontend/src/components/updatepopup/index.tsx
ashu8912: done

app/electron/main.ts
ashu8912: done

frontend/src/components/releasenotes/index.tsx
```diff
@@ -26,10 +32,20 @@ export default function ReleaseNotesModal(props: ReleaseNotesModalProps) {
   return (
     <Modal open={showReleaseNotes} BackdropComponent={Backdrop} style={modalStyle}>
       <Paper style={releaseNotesStyle}>
-        <ReactMarkdown>{releaseNotes}</ReactMarkdown>
-        <Box display="flex" justifyContent="flex-end">
-          <Button onClick={() => setShowReleaseNotes(false)}>Cancel</Button>
+        <Box display="flex" justifyContent="center">
+          <Box flexGrow={2}>
+            <Typography variant="h4">Release Notes({appVersion}) </Typography>
```
joaquimrocha, line 37: This needs to be translated.

frontend/src/components/releasenotes/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    minHeight: '50%',
+    overflow: 'auto',
+  };
+
+  return (
+    <Modal open={showReleaseNotes} BackdropComponent={Backdrop} style={modalStyle}>
+      <Paper style={releaseNotesStyle}>
+        <ReactMarkdown>{releaseNotes}</ReactMarkdown>
+        <Box display="flex" justifyContent="flex-end">
+          <Button onClick={() => setShowReleaseNotes(false)}>Cancel</Button>
```
joaquimrocha: Needs to be translated.

frontend/src/App.tsx
```diff
@@ -233,6 +234,45 @@ function ThemeChangeButton() {
...
+        'aria-describedby': 'updatePopup',
+      }}
+      message={t(`An update is available`)}
+      action={
+        <React.Fragment>
+          <Button color="secondary" href={`${updateDownloadURL}`} target="_blank">
+            {t('More')}
+          </Button>
+          <Button color="primary" onClick={() => setShow(false)}>
+            {t('Close')}
```
joaquimrocha: These should be namespaced with `frequent|WORD`

frontend/src/App.tsx
```diff
@@ -265,6 +265,15 @@ function UpdatePopup() {
...
             {t('More')}
           </Button>
+          <Button
+            color="inherit"
+            onClick={() => {
+              desktopApi.send('disableUpdateChecking');
+              setShow(false);
+            }}
+          >
+            Do not notify again
```
joaquimrocha: This also needs to be translated.

joaquimrocha (changes requested):
This is very close now. I noticed you translated some of the things I noted in this last review in later commits. That's fine. Please double check that every new string is translatable but be sure to add the namespace, e.g. `frequent|Close`. You can search for the strings and see if they already exist in a namespace file that makes sense (`frequent` for example).

frontend/src/components/updatepopup/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      anchorOrigin={{
+        vertical: 'bottom',
+        horizontal: 'right',
+      }}
+      open={show}
+      autoHideDuration={100000}
+      ContentProps={{
+        'aria-describedby': 'updatePopup',
+      }}
+      message={t(`An update is available`)}
```
joaquimrocha, line 29: No backquotes needed.

frontend/src/components/updatepopup/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+            color="inherit"
+            onClick={() => {
+              desktopApi.send('disableUpdateChecking');
+              setShow(false);
+            }}
+          >
+            {t('Do not notify again')}
+          </Button>
+          <Button color="primary" onClick={() => setShow(false)}>
+            {t('frquent|Close')}
```
joaquimrocha, line 45: There's a typo here.

ashu8912: oops

joaquimrocha (approved):

PR #293, opened 2021-07-26
charts: Bump chart version so we get it released

joaquimrocha:
We want to release again because now we do have the gh-page
published.


joaquimrocha:
Merging since it's a fairly straightforward change.

PR #294, opened 2021-07-27
Sort almost everything

joaquimrocha:
- frontend: Allow to set the SimpleTable's sort function as a getter
- frontend: Enable sorting in (almost) every column


illume (approved):
Thanks! 👍 🎉

Issue #301, opened 2021-08-09
Use Browser Storage to store all app related configs


ashu8912:
## Current situation
Instead of managing two stores for app and frontend we should stick to using one storage type which is easier to manage 

## Impact
we have one storage type which is easier to manage

## **Implementation options
maybe use package to get stuff from browser storage and remove the electron-store dependency
https://www.npmjs.com/package/electron-browser-storage#security-concerns

PR #304, opened 2021-08-16
frontend & desktop: Move releaseNotes,download update to client storage

ashu8912:
Use client storage for releaseNotes and update available logic


illume:
@ashu8912 seems prettier didn't run...
```
Checking formatting...
[warn] ../app/electron/main.ts
[warn] ../app/electron/preload.ts
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

illume:
... seems there's an error when this runs under test. Probably just needs a guard on the `desktopApi`, which sometimes doesn't exist.

```typescript
    TypeError: Cannot read property 'receive' of undefined

      255 | 
      256 |   React.useEffect(() => {
    > 257 |     desktopApi.receive('appVersion', (currentBuildAppVersion: string) => {
          |                ^
```

ashu8912:
fixes #301

ashu8912:
@illume about the prettier issue i found that the lint-staged hook doesn't runs prettier on electron src directory,i will open a seperate isssue for this

frontend/src/App.tsx
```diff
@@ -252,13 +255,56 @@ function AppContainer() {
...
+          ) {
+            setReleaseDownloadURL(latestRelease.html_url);
+          }
+          /*
+    check if there is already a version in store if it exists don't store the current version
+    this check will help us later in determining whether we are on the latest release or not
+    */
+          const storedAppVersion = localStorage.getItem('app_version');
+          if (!storedAppVersion) {
+            localStorage.setItem('app_version', currentBuildAppVersion);
```
joaquimrocha, line 285: Please abstract this in a function and replace this and the other call.

frontend/src/App.tsx
```diff
@@ -252,13 +255,56 @@ function AppContainer() {
 
   React.useEffect(() => {
     if (desktopApi) {
-      desktopApi.receive(
-        'showReleaseNotes',
-        (data: { releaseNotes: string; appVersion: string }) => {
-          setReleaseNotes(data.releaseNotes);
-          setAppVersion(data.appVersion);
+      desktopApi.receive('appVersion', (currentBuildAppVersion: string) => {
+        setAppVersion(currentBuildAppVersion);
```
joaquimrocha, line 259: Is the `appVersion` state actually needed here? I'd assume that since we store it in the localStorage an fetch it every time, there's no need to duplicate the state here. i.e. it can call `fetchRelease` with the `currentBuildAppVersion` and achieve the same AFAICS.

joaquimrocha (changes requested):
This PR moves more code than I expected initially, but it simplifies things together with it, so it looks good.

Left just a couple of comments to simplify it further. Good job.

frontend/src/App.tsx
ashu8912: done

frontend/src/App.tsx
ashu8912: done

joaquimrocha (approved):
LGTM.

illume (approved):
🎉 🎈

PR #307, opened 2021-08-19
Make release notes more resilient

joaquimrocha:
- app: Allow to override the app version with an env var
- frontend: Ensure the current app version is always set
- frontend: Catch any errors when getting the release information


ashu8912 (approved):
thanks, these are some nice set of changes.

PR #313, opened 2021-08-25
Fix showing the version in the release notes

joaquimrocha:
- frontend: Set the release notes' version string as translatable
- frontend: Add a margin between the release notes title and body
- frontend: Use frequent namespace on TopBar translations
- i18n: Use release namespace for an update release string
- i18n: Add missing translations


ashu8912 (approved):
LGTM!!

PR #315, opened 2021-08-27
Add facility to override details view

ashu8912:
Allow to append more details view via plugins

### DIFF
--- a/frontend/src/helpers/renderHelpers.tsx
+++ b/frontend/src/helpers/renderHelpers.tsx
@@ -1,0 +1,22 @@
+import { SectionBox } from '../components/common';
+import { useTypedSelector } from '../redux/reducers/reducers';
+
+export default function DetailsViewPluginRenderer(props: { resource: any }) {
+  const { resource } = props;
+  const detailViews = useTypedSelector(state => state.ui.views.details.pluginAppendedDetailViews);
+  return (
+    <>
+      {detailViews.map((item, index) => {
+        const pluginDetailsObj = item.sectionFunc(resource);
+        if (pluginDetailsObj) {
+          const { title, component: DetailsView } = pluginDetailsObj;
+          return (
+            <SectionBox title={title} key={`${item.sectionName}__${index}`}>
+              <DetailsView resource={resource} />
+            </SectionBox>
+          );
+        }
+      })}
+    </>
+  );
+}
--- a/frontend/src/plugin/registry.tsx
+++ b/frontend/src/plugin/registry.tsx
@@ -1,11 +1,20 @@
+import { KubeObject } from '../lib/k8s/cluster';
 import { Route } from '../lib/router';
 import {
   setAppBarAction,
+  setDetailsView,
   setDetailsViewHeaderAction,
   setRoute,
   setSidebarItem,
 } from '../redux/actions/actions';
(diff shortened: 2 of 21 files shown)


frontend/src/helpers/renderHelpers.tsx
```diff
@@ -0,0 +1,13 @@
+import { useTypedSelector } from '../redux/reducers/reducers';
+
+export default function DetailsViewRenderMapper(props: { resource: any }) {
+  const { resource } = props;
+  const detailViews = useTypedSelector(state => state.ui.views.details.pluginAppendedDetailViews);
+  return (
+    <>
+      {detailViews.map((item, index) => {
+        return <item.renderComponent resource={resource} key={`${item.sectionName}__${index}`} />;
```
joaquimrocha, line 9: This needs to be added to a SectionBox, and it's title also added as the title. That's why I mentioned we should move the title out of the plugin registration and into the return of the sectionFunc.

joaquimrocha (changes requested):

frontend/src/redux/reducers/ui.tsx
```diff
@@ -42,6 +43,10 @@ export interface UIState {
       headerActions: {
         [name: string]: HeaderActionFunc;
       };
+      pluginAppendedDetailViews: Array<{
+        sectionName: string;
+        sectionFunc: () => { title: string; component: (props: any) => JSX.Element | null };
```
joaquimrocha, line 48: This should be `(resource: KubeObject) => { title?: string; component?: (props: any) => JSX.Element | null } | null | undefined;`

title can be optional if we can take the sectionName. Otherwise, let's keep it mandatory as you had.
But the whole result can be null | undefined, because if you want not to have anything showing, you maybe just return null/undefined...

frontend/src/plugin/registry.tsx
```diff
@@ -101,4 +102,26 @@ export default class Registry {
...
+  /**
+   * Append the specified title and component to the details view.
+   * @param sectionName a unique name for it
+   * @param sectionFunc - a function that returns your detail view component with props
+   *                      passed into it and the section title
+   *
+   * @example
+   *
+   * ```JSX
+   * register.registerDetailsViewSection("Block I/O Latency", () => { title: 'Block I/O Latency', component: (props) => <BioLatency {...props}/>});
```
joaquimrocha, line 115: The section name could be examplified as `biolatency`. I am not sure what we'll do with it yet, but maybe it's good to have it as an identifier.

frontend/src/helpers/renderHelpers.tsx
```diff
@@ -0,0 +1,19 @@
+import { SectionBox } from '../components/common';
+import { useTypedSelector } from '../redux/reducers/reducers';
+
+export default function DetailsViewRenderMapper(props: { resource: any }) {
```
joaquimrocha, line 4: I'd prefer a name that's more direct and thus easier to understand, so we can read:
`<DetailsViewPluginsRenderer />`

joaquimrocha (changes requested):
@ashu8912 , this is very close to what's needed. Left a couple of comments.

joaquimrocha (approved):
LGTM now.

PR #316, opened 2021-08-30
backend: better way of determining the URL scheme

zdzichu:
# Better way of determining scheme of an URL for OIDC callback

Instead of guessing (which would return HTTP for localhost and HTTPS for every other situation – which is not always correct), check TLS property of a request object. It is `nil` for plain connection, and not `nil` if HTTPS is used.

## How to use

Not a new functionality.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -164,10 +164,15 @@ func baseURLReplace(staticDir string, baseURL string) {
 func getOidcCallbackURL(r *http.Request, config *HeadlampConfig) string {
 	urlScheme := r.URL.Scheme
 	if urlScheme == "" {
-		// @todo: Find a better way to get the scheme for the URL.
-		if r.Host == "localhost:"+config.port {
+		// check proxy headers first
+		fwdProto := r.Header.Get("X-Forwarded-Proto")
+
+		switch {
+		case fwdProto != "":
+			urlScheme = fwdProto
+		case r.Host == "localhost:"+config.port:
 			urlScheme = "http"
-		} else {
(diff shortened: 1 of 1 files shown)


zdzichu:
Actually, I don't think the new way is correct. If headlamp is run behind reverse proxy, it will not have access to TLS session.
Maybe the only way is to have baseUrl defined and extract the scheme from it?

joaquimrocha:
AFAIU, this logic is only run if the URL scheme is empty. So if the /oidc endpoint is called with a defined scheme, it'll be used.
Is this an issue?

zdzichu:
This is an issue because apparently in Go `URL.Scheme` is always empty (https://github.com/golang/go/issues/28940)
I've hit this problem, because I have Headlamp  (in my internal network) served over plain HTTP, and the `getOidcCallbackURL` function keep returning an URL with HTTPS (because it is *not* localhost).

joaquimrocha:
Ouch! Maybe we need to define a CLI arg for completely overridding the OIDC callback URL.

joaquimrocha (approved):
This LGTM. Thanks @zdzichu !
BTW, feel free to file any issues with bugs or suggestions you may have from your daily use of Headlamp.

joaquimrocha:
Should have run the CI first.
You may be able to check the lint issues with `make backend-lint`.

zdzichu:
Sorry about that. I've pleased the linter by turning `if-else-if` block into `switch`. CI is green now.

joaquimrocha:
Awesome! Thanks @zdzichu !

PR #317, opened 2021-09-08
Provide patch functionality

ashu8912:
Allow patch requests and provide handlers for it

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -7,6 +7,7 @@
  * Copyright © 2020 Kinvolk GmbH
  */
 
+import { OpPatch } from 'json-patch';
 import helpers from '../../helpers';
 import { getToken, logout } from '../auth';
 import { getCluster } from '../util';
@@ -98,7 +99,7 @@ export async function request(
 
     const error: Error & { status?: number } = new Error(message);
     error.status = status;
-    throw error;
+    return Promise.reject(error);
   }
 
   return response.json();
@@ -117,6 +118,7 @@ export function apiFactory(group: string, version: string, resource: string) {
       streamResult(url, name, cb, errCb),
     post: (body: KubeObjectInterface) => post(url, body),
     put: (body: KubeObjectInterface) => put(`${url}/${body.metadata.name}`, body),
+    patch: (body: OpPatch[], name: string) => patch(`${url}/${name}?pretty=true`, body),
     delete: (name: string) => remove(`${url}/${name}`),
     isNamespaced: false,
   };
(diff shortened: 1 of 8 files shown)


ashu8912:
The CI is failing because of an issue with jest on node.js v16
https://github.com/nodejs/node/issues/40030

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -98,7 +99,7 @@ export async function request(
 
     const error: Error & { status?: number } = new Error(message);
     error.status = status;
-    throw error;
+    return Promise.reject(error);
```
joaquimrocha, line 102: Why is this need? And doesn't this require changing the code calling/catching this function?

ashu8912: With throw we have problems of not being able to handle the errors 
There are two solutions to it either wrap the entire function in a promise and then throw from it or return an error wrapped in Promise.reject so that our catch can handle it
See https://stackoverflow.com/a/33446005/8527880

joaquimrocha: What I mean is that we've been using it this way. So I imagine we'd have to change some of the calling code that may be catching the exception already.

ashu8912: I checked i think we don't need to change other parts of code for this as at all places we use the .catch method on promise

frontend/src/helpers/renderHelpers.tsx
```diff
@@ -1,18 +1,26 @@
...
     <>
       {detailViews.map((item, index) => {
         const pluginDetailsObj = item.sectionFunc(resource);
         if (pluginDetailsObj) {
-          const { title, component: DetailsView } = pluginDetailsObj;
+          const { title } = pluginDetailsObj;
           return (
             <SectionBox title={title} key={`${item.sectionName}__${index}`}>
-              <DetailsView resource={resource} />
+              {prepareDetailsComponent(pluginDetailsObj.component, resource)}
```
illume, line 23: Could this be something like this instead? (to avoid having to make the prepareDetailsComponent function)
```tsx
          const { title, component: Component } = pluginDetailsObj;
          return (
            <SectionBox title={title} key={`${item.sectionName}__${index}`}>
              <Component resource={resource} />
          ...
```

frontend/src/lib/k8s/resource.ts
```diff
@@ -0,0 +1,14 @@
+import { OpPatch } from 'json-patch';
+import { apiFactoryWithNamespace } from './apiProxy';
+
+export default class Resource {
+  static className = 'Resource';
```
joaquimrocha, line 5: Is class `Resource` correct? We use the className for a number of operations, and "CustomResource" is a more appropriate name as Resource is too generic.

joaquimrocha: This should be applied to the file name, class, and className BTW. So we can use `k8s.CustomResource` instead of `k8s.Resource`.

joaquimrocha (changes requested):
Left a comment about naming, but the rest is looking good. Also please check René's comments.

joaquimrocha (approved):

PR #319, opened 2021-09-13
Fix render issues and add more modules to pluginLib

ashu8912:
## Testing Done
add a component via a plugin  registerDetailsViewSection method place a useEffect with unmount handler inside it to see if it gets called each time the parent state changes

### DIFF
--- a/frontend/src/helpers/renderHelpers.tsx
+++ b/frontend/src/helpers/renderHelpers.tsx
@@ -1,22 +1,25 @@
+import { useMemo } from 'react';
 import { SectionBox } from '../components/common';
+import { KubeObject } from '../lib/k8s/cluster';
 import { useTypedSelector } from '../redux/reducers/reducers';
 
-export default function DetailsViewPluginRenderer(props: { resource: any }) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -18,7 +18,7 @@ export default function RouteSwitcher() {
           route.name === 'OidcAuth' ? (
             <Route
               path={route.path}
-              component={() => (
+              render={() => (
```
joaquimrocha, line 21: I need to understand better why we need this. AFAIU, if we don't unmount when a route is changed, it means we'll increase the memory footprint and the whole UI may become much slower after some use.

ashu8912: any state change will cause the component to get re rendered in case of a component prop a new component is rendered each time rather than rendering with some parts changed
https://reactrouter.com/web/api/Route/render-func

ashu8912:
more info #323

illume (approved):
👍

PR #320, opened 2021-09-13
plugin: Pass apiProxy and recharts to pluginLib

ashu8912:


frontend/src/plugin/index.tsx
```diff
@@ -99,6 +99,7 @@ declare global {
 }
 
 window.pluginLib = {
+  apiProxy: require('../lib/k8s/apiProxy'),
```
joaquimrocha, line 102: I more inclined to wrapping a function to create a class that will use the `apiFactory` and `apiFactoryWithNamespace` but this export may still be useful. One comment though. It should be exported in a consistent way with the rest: `ApiProxy` instead of `apiProxy`.

joaquimrocha (changes requested):

joaquimrocha (approved):

PR #321, opened 2021-09-20
headlamp-plugin: Make the package arg of the build command optional

joaquimrocha:
We rely on just calling the "headlamp-plugin build" command without
any arguments, but the build command expected a positional argument as
required, so the default and advised case for building plugins was
failing.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -47,3 +47,6 @@ jobs:
       run: |
         make frontend-build
 
+    - name: Test plugins
+      run: |
+        make plugins-test
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,9 @@ frontend-tsc:
 frontend-test:
 	cd frontend && npm run test -- --coverage
 
+plugins-test:
+	cd plugins/headlamp-plugin && ./test-headlamp-plugin.sh
+
 image:
 	$(DOCKER_CMD) build \
 	--build-arg IMAGE_BASE=$(DOCKER_IMAGE_BASE) \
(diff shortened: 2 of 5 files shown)


illume:
interestingly, the CI found a bug when using npm 7. Where npm 7 gets into a loop resolving dependencies, and keeps processing until it runs out of memory. See https://github.com/npm/cli/issues/3359 and https://github.com/npm/cli/issues/3208

There's a work around for running it on CI with npm 7, but users of npm 7 will encounter this bug.

I tried updating the plugin dependencies, and still get the issue. There's another work around or two to try(which I will do next)... but ultimately the bug in npm 7 needs fixing (or maybe it's a bug in one of the dependencies).

illume:
Adding a package-lock.json generated with npm v6 to the template lets npm v7 work.

illume (approved):
👍 🎉

illume:
Probably we should release a new `@kinvolk/headlamp-plugin` version.

ashu8912 (approved):

joaquimrocha (commented):
@illume , It all LGTM. The only thing I noticed is that we're not testing the `headlamp-plugin build` (no argument) which is what failed initially, or maybe I missed it?

illume:
@joaquimrocha yeah. It's not directly calling `headlamp-plugin build`... but it does so through `npm run build` which calls that. I'll update the test to call `headlamp-plugin build` with no argument directly.

illume:
The `headlamp-plugin build` is tested better now, and also added a test for `headlamp-plugin extract`.

joaquimrocha (commented):
Thanks!

PR #328, opened 2021-09-27
Fix authorization and update K8s APi versioning

joaquimrocha:


illume (approved):
:+1: Nice one.

I wonder if the API version update will break (m)any people? It seems some v1beta1 APIs are being removed according to here: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

joaquimrocha:
@illume , that's a good point. I think we should probably try both versions of the APIs until we decide to drop very old versions...
I will check how we can do this.

joaquimrocha:
Besides the cryptic typescript, this should be a transparent enough change to support multiple API versions.
@illume, please give it a look.

illume:
Hey hey. Nice!

I'm wondering how to test this. I guess trying it with an old minikube or so.

illume (changes requested):
I'd like to test this first.

joaquimrocha:
You can try an old minikube from last year, or you can also change the APIs versions locally in Headlamp and see whether things keep working (because if the wrong ones come first, they should 404).

illume:
I tried as far back as minikube v1.5.1 with Kubernetes v1.16.2.

- ingress works with the fallback, first 404ing twice the v1 ingress URL (probably should only 404 once) then using the beta one.
- does cronjob also need to be updated to v1 with a fallback? https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125

illume:
Do we want to support k8s v1.8 and v1.9? 
(If so, then we have some more fall backs to add where we are using v1/ APIs that aren't available then.)

joaquimrocha:
> Do we want to support k8s v1.8 and v1.9? (If so, then we have some more fall backs to add where we are using v1/ APIs that aren't available then.)

Nah, that's too far back I think.
I will update it for cronjob. Maybe all v1beta1 should be changed to v1, for future compatibility, but maybe that opens unforeseen issues :stuck_out_tongue:

illume (approved):

PR #336, opened 2021-10-13
Fix plugin related issues

ashu8912:
This PR fixes and adds a few things that are required for the plugins to work correctly.

1. The ```createRouteURL``` method in ```frontend/src/lib/router.tsx``` doesn't considers routes from redux store which could be a problem when a plugin wants to render a link and the only way the plugin can communicate it's route params is through the store so we have to make the createRouteURL method aware of the store routes.
2. We need the Editor component exported because it will come in useful when we want to modify a kube object.
3. We need  APIs such as ```useLocation``` or ```useHistory``` etc. from react-router package so it makes sense to export it as well

### DIFF
--- a/frontend/src/lib/router.tsx
+++ b/frontend/src/lib/router.tsx
@@ -45,6 +45,7 @@ import PersistentVolumeList from '../components/storage/VolumeList';
 import WorkloadDetails from '../components/workload/Details';
 import WorkloadOverview from '../components/workload/Overview';
 import LocaleSelect from '../i18n/LocaleSelect/LocaleSelect';
+import store from '../redux/stores/store';
 import CronJob from './k8s/cronJob';
 import Deployment from './k8s/deployment';
 import Job from './k8s/job';
@@ -432,7 +433,8 @@ export interface RouteURLProps {
 }
 
 export function createRouteURL(routeName: string, params: RouteURLProps = {}) {
-  const route = getRoute(routeName);
+  const storeRoutes = store.getState().ui.routes;
+  const route = getRoute(routeName) || (storeRoutes && storeRoutes[routeName]);
 
   if (!route) {
     return '';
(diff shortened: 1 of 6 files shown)


frontend/src/lib/router.tsx
```diff
@@ -431,8 +431,12 @@ export interface RouteURLProps {
...
 }
 
-export function createRouteURL(routeName: string, params: RouteURLProps = {}) {
-  const route = getRoute(routeName);
+export function createRouteURL(
+  routeName: string,
+  params: RouteURLProps = {},
+  storeRoutes?: { [path: string]: any }
+) {
+  const route = getRoute(routeName) || (storeRoutes && storeRoutes[routeName]);
```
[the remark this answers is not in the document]
ashu8912: Fixed

joaquimrocha (approved):

joaquimrocha:
@illume , do you want to give one last look at this one?

frontend/src/lib/router.tsx
```diff
@@ -432,7 +433,8 @@ export interface RouteURLProps {
 }
 
 export function createRouteURL(routeName: string, params: RouteURLProps = {}) {
-  const route = getRoute(routeName);
+  const storeRoutes = store.getState().ui.routes;
+  const route = getRoute(routeName) || (storeRoutes && storeRoutes[routeName]);
```
illume, line 437: I wonder why getRoute(routeName) is chosen over the store routes here?

illume:
I'm not sure what I should be testing. The PR could use a description and the PR title could mention which issues it addresses.

frontend/src/lib/router.tsx
ashu8912: So it's like first search in the list of routes we have if that's undefined check if it's in storeRoutes

ashu8912:
@illume i have added a description for the PR

PR #343, opened 2021-10-26
Add pod list deployments

ashu8912:
Add the list of pods which belongs to a deployment under deployment view and remove container section from Deployment Detail view

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -649,6 +649,8 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
   const { resource } = props;
   const { t } = useTranslation('glossary');
 
+  let title = '…';
+
   function getContainers() {
     if (!resource) {
       return [];
@@ -658,8 +660,10 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
 
     if (resource.spec) {
       if (resource.spec.containers) {
+        title = t('Containers');
         containers = resource.spec.containers;
       } else if (resource.spec.template && resource.spec.template.spec) {
+        title = t('Container Spec');
         containers = resource.spec.template.spec.containers;
       }
     }
@@ -690,7 +694,7 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
 
   return (
     <>
-      <SectionBox title={t('Containers')} />
+      <SectionBox title={title} />
       <>
         {numContainers === 0 ? (
           <SectionBox>
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -35,7 +35,7 @@ import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
 import SectionHeader, { HeaderStyleProps } from '../../common/SectionHeader';
 import SimpleTable, { NameValueTable, NameValueTableRow } from '../../common/SimpleTable';
-import { PodListRenderer } from '../../pod/List';
+import { PodListProps, PodListRenderer } from '../../pod/List';
 import { LightTooltip } from '..';
 import Empty from '../EmptyContent';
 import { DateLabel, HoverInfoLabel, StatusLabel, StatusLabelProps } from '../Label';
@@ -647,7 +647,6 @@ export function OwnedPodsSection(props: OwnedPodsSectionProps) {
   }
 
   const filteredPods = getOwnedPods();
-  console.log('filtered pods', filteredPods);
   return <PodListRenderer hideColumns={hideColumns} pods={filteredPods} error={error} />;
 }
 export function ContainersSection(props: { resource: KubeObjectInterface | null }) {
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -717,7 +717,7 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
   );
 }
 
-export function ReplicasSection(props: { resource: KubeObjectInterface | null }) {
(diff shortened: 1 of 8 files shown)


frontend/src/components/pod/List.tsx
```diff
@@ -39,7 +47,7 @@ export default function PodList() {
       <SimpleTable
         rowsPerPage={[15, 25, 50]}
         filterFunction={filterFunc}
-        errorMessage={Pod.getErrorMessage(error)}
+        errorMessage={Pod.getErrorMessage(error as ApiError)}
```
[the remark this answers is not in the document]
ashu8912: The thing is we want to allow the error to be an optional entity to the Component thus error has types undefined | null | ApiError.
The function Pod.getErrorMessage takes a type ApiError | null as argument that is why we need a explicit typecast here and we are safe here because there is a value check ultimately inside the getErrorMessage method

frontend/src/components/pod/List.tsx
```diff
@@ -21,9 +22,16 @@ export function makePodStatusLabel(pod: Pod) {
...
   return <StatusLabel status={status}>{phase}</StatusLabel>;
 }
+interface PodListProps {
+  pods?: Pod[];
+  error?: ApiError | null;
+}
+export default function PodList(props: PodListProps) {
+  let { pods, error } = props;
+  if (!pods) {
+    [pods, error] = Pod.useList();
```
joaquimrocha, line 33: I thought hooks couldn't be used conditionally 🤔

joaquimrocha (changes requested):
What's the impact in the performance of the affected views?
I am thinking maybe we need to somehow cache the pods (maybe in redux) in order to make the views more efficient.
Otherwise just browsing through the workloads can become very slow and repeating unnecessary work.

frontend/src/components/pod/List.tsx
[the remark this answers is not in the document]
illume: If Pod.getErrorMessage should handle undefined, then the type should be changed there to say it can. Another way to fix it is to consider if error should be `APIError | null | undefined`... and instead just be `APIError | null` (maybe looking at the hook above to be non-conditional could clarify the types here).

frontend/src/components/pod/List.tsx
ashu8912: Fixed

ashu8912:
I agree that we are at many places doing jobs that can be easily cached and thus improve the views performance, maybe we can deal with all such cases in a seperate PR, WDYT??

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -439,6 +441,38 @@ export function ContainerInfo(props: { container: KubeContainer }) {
...
+      return [];
+    }
+    return resource?.spec.template.metadata.labels;
+  }
+
+  function getOwnedPods() {
+    if (!pods) {
+      return [];
+    }
+    const resourceTemplateLabel = Object.values(getResourceTemplateLabel());
```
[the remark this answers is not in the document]
ashu8912: updated, thanks

frontend/src/components/pod/List.tsx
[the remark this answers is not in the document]
ashu8912: Fixed with how we were handling apiPods

frontend/src/components/pod/List.tsx
```diff
@@ -21,9 +22,15 @@ export function makePodStatusLabel(pod: Pod) {
 
   return <StatusLabel status={status}>{phase}</StatusLabel>;
 }
+interface PodListProps {
+  pods?: Pod[];
+  error?: ApiError | null;
+}
+export default function PodList(props: PodListProps) {
+  const { pods, error } = props;
+  const [apiPods, apiError] = Pod.useList();
```
joaquimrocha, line 32: This means we'll always attempt to fetch the pod list, regardless of whether the pods are passed to it.
I think maybe we could just split out the PodList in PodListRenderer (this one accepts pods as props) and PodList (this one uses Pod.useList and passes the result to a PodListRenderer). This way you can call PodListRenderer in the deployment details.

joaquimrocha (changes requested):
Clarified the approach used here.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -617,6 +618,30 @@ export function ContainerInfo(props: ContainerInfoProps) {
   );
 }
 
+export function OwnedPodsSection(props: { resource: KubeObjectInterface | null }) {
+  const { resource } = props;
+
+  if (!resource) {
+    return null;
+  }
+  const [pods, error] = Pod.useList();
```
joaquimrocha, line 627: I think this constitutes still using a hook conditionally.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -35,8 +35,8 @@ import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
 import SectionHeader, { HeaderStyleProps } from '../../common/SectionHeader';
 import SimpleTable, { NameValueTable, NameValueTableRow } from '../../common/SimpleTable';
-import { LightTooltip } from '..';
 import { PodListRenderer } from '../../pod/List';
+import { LightTooltip } from '..';
```
joaquimrocha, line 39: Should this be here?

frontend/src/components/pod/List.tsx
```diff
@@ -64,7 +69,6 @@ export default function PodList() {
     }
     return _.sumBy(pod.status.containerStatuses, container => container.restartCount);
   }
-
```
joaquimrocha, line 67: Shouldn't be here.

joaquimrocha (changes requested):
I left some more comments but I will test it now and hopefully add some fixups.

joaquimrocha:
Awesome @ashu8912 , we got this in!

PR #346, opened 2021-11-02
Add 404 route and implement the NotFound Component

ashu8912:


joaquimrocha (changes requested):
This PR has strings that need to be translatable.

joaquimrocha (approved):

PR #347, opened 2021-11-02
Limit the window size for desktop app, setup jest test framework

ashu8912:
## Testing done
still needs to be tested on larger screens

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -10,6 +10,7 @@ import path from 'path';
 import url from 'url';
 import yargs from 'yargs';
 import i18n from './i18next.config';
+import windowSize from './windowSize';
 
 const args = yargs
   .command('$0 [kubeconfig]', '', yargs => {
@@ -66,6 +67,19 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
   serverArgs.concat(flags);
 
   return spawn(serverFilePath, serverArgs, options);
+}
+
+/**
+ * Are we running inside WSL?
+ * @returns true if we are running inside WSL.
+ */
+function isWSL(): boolean {
+  try {
+    const data = fs.readFileSync('/proc/version', { encoding: 'utf8', flag: 'r' });
+    return data.indexOf('icrosoft') !== -1;
+  } catch {
+    return false;
+  }
 }
 
 let serverProcess: ChildProcessWithoutNullStreams | null;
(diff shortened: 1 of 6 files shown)


app/electron/main.ts
```diff
@@ -258,6 +258,24 @@ function setMenu(i18n: I18n) {
   Menu.setApplicationMenu(menu);
 }
 
+function calculateWindowSize() {
+  const { width, height } = screen.getPrimaryDisplay().workAreaSize;
+  const [maxWidth, maxHeight] = [1920, 1080];
+  const pixelRatio = screen.getPrimaryDisplay().scaleFactor;
```
illume, line 264: scaleFactor is a Number which can be a float. So a 1.5x screen should be returned here as `1.5`.

So the case statement below should be changed into a series of ranges.

illume:
Started a table so we can set some expectations of what good sizes would be.

| Size         | Scale     | Window size | Name        |
|--------------|-----------|-------------|-------------|
| 1366×768     | 100%      |             | Popular     |
| 1920x1080    | 100%      |             | HD          |
| 1920x1080    | 200%      |             | 4K zoomed   |
| 2880x1800    | 100%      |             | Mac retina  |
| 1440x900     | 200%      |             | Mac retina  |
| 3240x2160    | 200%      |             | SB3         |
| 3240x2160    | 100%      |             | SB3 0 scale |
| 3840x2160    | 100%      |             | 4K          |

illume:
> for displays larger then 1920x1080 we should limit the windows size to 1920x1080

Limiting it to 1920x1080 on large screens sounds ok to me. I wonder if anyone uses Headlamp as a dashboard? For a dashboard mode, full screen would be important. Can't think of another reason on large screens why someone would prefer a giant window with Headlamp as it is.

illume:
I made a test using a table of expected results. But haven't checked if those results are correct.

One change I also made I wanted to ask you about... The height should probably not be limited. Instead it should probably use the full height of the work area, because some views could be quite long.  What do you think?

illume:
I put some margin around the height and width... mainly because on WSL it wasn't fitting on screen. But maybe other systems have toolbars/docks. It kind of looks a bit squished height wise on my Mac when the dock is hiding.

Also put the max height back, because on large screens it's kind of silly.

app/electron/main.ts
```diff
@@ -342,7 +338,11 @@ function startElecron() {
         protocol: 'file:',
         slashes: true,
       });
-    const { width, height } = calculateWindowSize();
+
+    // WSL has a problem with full size window placement, so make it smaller.
+    const windowMargin = isWSL();
+    const { width, height } = windowSize(screen.getPrimaryDisplay().workAreaSize, windowMargin);
```
joaquimrocha, line 344: Nit: Reading this, I'd assume windowMargin is a numeric value. It's good practice for booleans to be indicated as such by  using `is` or `with` or similar prefixes.

app/package.json
```diff
@@ -14,7 +14,8 @@
     "package": "npm run copy-icons && npm run compile-electron && electron-builder build --publish never",
     "dev": "npm run compile-electron && cross-env ELECTRON_DEV=1 electron .",
     "dev-only-app": "npm run compile-electron && cross-env ELECTRON_DEV=1 ELECTRON_START_URL=http://localhost:3000 EXTERNAL_SERVER=true electron .",
-    "i18n": "npx --no-install i18next ./electron/main.ts -c ./electron/i18next-parser.config.js"
+    "i18n": "npx --no-install i18next ./electron/main.ts -c ./electron/i18next-parser.config.js",
+    "test": "jest"
```
joaquimrocha, line 18: Did you forget to include jest in package.json? I see it only in the lock file.

illume: My clumsy clumsy fingers! Thanks.

app/electron/main.ts
illume: Nit squashed withThanks.

joaquimrocha (approved):

PR #351, opened 2021-11-12
Upload app artifacts & draft release

joaquimrocha:
This is a PR towards automating the release process a bit more.
It builds and uploads the apps as an action artifact, so we can test them, and creates a release draft with the format we've been using.

### DIFF
--- a/.github/workflows/app-artifacts-linux.yml
+++ b/.github/workflows/app-artifacts-linux.yml
@@ -1,0 +1,44 @@
+name: Build and upload Linux app artifact
+
+on:
+  workflow_dispatch:
+    inputs:
+      buildBranch:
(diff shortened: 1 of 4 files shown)


ashu8912 (approved):

illume:
Nice one.

I looks like these workflows are to be run on dispatch rather on making a tag.

Probably you considered adding artifact uploading to the app.yml?

joaquimrocha:
Yes. This only semi-automates things. It allows us to build and download the apps for testing, but e.g. Mac binaries need to be signed and I am not yet comfortable with the solutions I saw for doing that from GH actions.

Probably, for simplification, we can drop the app.yml and keep the new individual app actions. That would also allow us to have binaries for PRs in case we want to quickly try them.

PR #352, opened 2021-11-12
github: Simplify the release draft action

joaquimrocha:
Turns out the action used before is discontinued, so this one replaces it and also simplifies the logic.


illume (approved):
👍

ashu8912 (approved):

PR #363, opened 2021-11-25
Fix showing CRD objects

joaquimrocha:
- frontend: Fix hook usage in EditorDialog
- frontend: Fix listing objects in CRDs


ashu8912 (approved):
Tested and LGTM!!

PR #364, opened 2021-11-25
frontend: Allow to apply unknown resource configurations

joaquimrocha:
Headlamp was using its resource classes in order to apply the
configurations desired by the user (in the Create/Edit dialog).
However, if the user used a configuration of a Custom Resource
Definition, it would fail as the kind would not be known to Headlamp.

This patch fixes that by not requiring the use of resource classes
for the mentioned uses.

fixes #364


ashu8912 (approved):
LGTM!!

PR #367, opened 2021-11-29
Add a CustomResource details view

### LINKED ISSUE #365: Add a Custom Resource details view
## Current situation

We list the custom resources for each definition but when clicking on one of them, the YAML viewer is used, instead of a dedicated view, which would allow the user to edit/remove the resource, besides viewing it in a simpler way.

## Impact

Users cannot remove/edit custom resources.

## Ideal future situation

There is a custom resource details view which shows the custom resource objects' details, showing an edit and remove buttons.

joaquimrocha:
- frontend: Fix hook usage in EditorDialog
- frontend: Fix listing objects in CRDs
- frontend: Add a dedicated CustomResource details view

fixes #365


joaquimrocha:
Wrong branch to apply to. So I closed it.

PR #371, opened 2021-11-30
desktop: Use a different kubeconfig

### LINKED ISSUE #357: [desktop] : can't specify the K8s config file
## Description
I am using many different K8s config files with only one context in them but headlamp seems to be using the default ~/.kube.config file.
That would be nice to be able to choose the config file.

## Impact

can use headlamp if i have to copy my individual config files to the default `config` file.

### LINKED ISSUE #370: Allow to load different non-default k8s config files in the desktop app
## Current situation

We open the default k8s config file when loading the desktop app, but we should allow a non-default K8s file to be used.
There's another issue (#357 ) related to this and the user suggests loading multiple files from the default location.

## Impact

This makes the use of the desktop app less flexible.

## Ideal future situation

We should be able to allow the user to open a different kube config file, by specifying it in the command line.

## **Implementation options

We could investigate whether loading multiple files at once, from a given directory, is a good idea.

joaquimrocha:
This PR adds the possibility of running Headlamp with a different kube config, by using either a `KUBECONFIG` env var, or the new positional argument: `headlamp /path/to/my/different/kubeconfig`.

fixes #370 
fixes #357 

cc/ @sgandon , it's not loading an entire dir by default, but I hope this makes sense.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -310,10 +310,17 @@ function startElecron() {
   setMenu(i18n);
 
   function createWindow() {
+    let frontendPath = '';
+    if (isDev) {
+      frontendPath = path.resolve('..', 'frontend', 'build', 'index.html');
+    } else {
+      frontendPath = path.join(process.resourcesPath, 'frontend', 'index.html');
+    }
+
     const startUrl =
       process.env.ELECTRON_START_URL ||
       url.format({
-        pathname: path.join(process.resourcesPath, 'frontend', 'index.html'),
+        pathname: frontendPath,
         protocol: 'file:',
         slashes: true,
       });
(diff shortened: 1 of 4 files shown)


ashu8912:
Screenshot 2021-12-09 at 12 02 57 PM
@joaquimrocha I tested the PR and the app shows a blank screen with the above error on closing the app

joaquimrocha:
> @joaquimrocha I tested the PR and the app shows a blank screen with the above error on closing the app

Turns out my changes still required having the frontend running. So I added another commit to address that. Now either there's frontend/build (used when running `npm run dev` in app/) or we need to start the frontend as well as the backend when using `npm run dev-only-app`.

ashu8912 (approved):
Tested and it works fine now

PR #373, opened 2021-12-13
Fix i18n warnings

joaquimrocha:
There were a lot of i18n warnings, at first sight only due to the English catalog having only keys, not values, but then there were more things.
So I have dug into this rabbit hole today.


frontend/src/App.tsx
```diff
@@ -22,15 +25,21 @@ function AppWithRedux(props: React.PropsWithChildren<{}>) {
     initializePlugins();
   }, [themeName]);
 
-  return <ThemeProviderNexti18n theme={themes[themeName]}>{props.children}</ThemeProviderNexti18n>;
+  return (
+    <I18nextProvider i18n={i18n}>
+      <ThemeProviderNexti18n theme={themes[themeName]}>{props.children}</ThemeProviderNexti18n>
```
illume, line 30: I think ThemeProviderNexti18n does I18nextProvider.

frontend/src/App.tsx
```diff
@@ -22,15 +25,21 @@ function AppWithRedux(props: React.PropsWithChildren<{}>) {
...
+  );
 }
 
 function App() {
   return (
     <Provider store={store}>
-      <AppWithRedux>
-        <AppContainer />
-      </AppWithRedux>
+      <React.Suspense fallback={<Loader title="Loading…" />}>
```
illume, line 38: I think React.Suspense is still an experimental feature (https://reactjs.org/docs/concurrent-mode-suspense.html). We had trouble with it long ago too (I seem to recall).

frontend/src/components/404/index.tsx
```diff
@@ -16,14 +16,16 @@ export default function NotFoundComponent() {
         style={{ minHeight: '100vh' }}
       >
         <Grid item>
-          <img src={headlampBrokenImage} alt={t('404 broken headlamp')} />
+          <img src={headlampBrokenImage} alt="" />
```
illume, line 19: We need some alt text here I think.

joaquimrocha: I don't think so, because the image doesn't add any context needed that is not already given by the other textual info.
So AFAIU, the best practice here is to use either `alt=""` or `aria-hidden="true"`.

frontend/src/App.tsx
joaquimrocha: I have figured it out in the end without having to use the suspend.

PR #377, opened 2021-12-17
frontend: Fix fetching the resource in the RoleBindingDetails view

joaquimrocha:
The RoleBinding and ClusterRoleBinding resources were being used
without and with the namespace, respectively, but it should be the
other way around.


ashu8912 (approved):
Tested, LGTM!!

PR #379, opened 2021-12-21
Show more detailed container statuses

joaquimrocha:
- frontend: Add a container status label to the Pod details view
- frontend: Make the Pod status better show its real status
- frontend: Show more details about the container status in Pod details

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -1,3 +1,4 @@
+import alertIcon from '@iconify/icons-mdi/alert-outline';
 import chevronLeft from '@iconify/icons-mdi/chevron-left';
 import eyeIcon from '@iconify/icons-mdi/eye';
 import eyeOff from '@iconify/icons-mdi/eye-off';
@@ -22,15 +23,18 @@ import { ApiError } from '../../../lib/k8s/apiProxy';
 import {
   KubeCondition,
   KubeContainer,
+  KubeContainerStatus,
   KubeObject,
   KubeObjectInterface,
 } from '../../../lib/k8s/cluster';
+import { KubePod } from '../../../lib/k8s/pod';
 import { createRouteURL, RouteURLProps } from '../../../lib/router';
 import { useTypedSelector } from '../../../redux/reducers/reducers';
 import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
 import SectionHeader, { HeaderStyleProps } from '../../common/SectionHeader';
 import SimpleTable, { NameValueTable, NameValueTableRow } from '../../common/SimpleTable';
+import { LightTooltip } from '..';
 import Empty from '../EmptyContent';
 import { DateLabel, HoverInfoLabel, StatusLabel, StatusLabelProps } from '../Label';
 import Link, { LinkProps } from '../Link';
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -440,7 +444,7 @@ function LivenessProbes(props: { liveness: KubeContainer['livenessProbe'] }) {
   return (
     <Box display="flex" flexDirection="column">
       <LivenessProbeItem>
-        {`http-get, path: ${liveness?.httpGet?.path}, port: ${liveness?.httpGet?.port}, 
+        {`http-get, path: ${liveness?.httpGet?.path}, port: ${liveness?.httpGet?.port},
```
ashu8912, line 448: do we need this change here??

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -37,9 +35,11 @@ import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
 import SectionHeader, { HeaderStyleProps } from '../../common/SectionHeader';
 import SimpleTable, { NameValueTable, NameValueTableRow } from '../../common/SimpleTable';
+import { LightTooltip } from '..';
 import Empty from '../EmptyContent';
 import { DateLabel, HoverInfoLabel, StatusLabel, StatusLabelProps } from '../Label';
 import Link, { LinkProps } from '../Link';
+import { useMetadataDisplayStyles } from '.';
```
ashu8912, line 42: why this change??

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -37,9 +35,11 @@ import Loader from '../../common/Loader';
 import { SectionBox } from '../../common/SectionBox';
 import SectionHeader, { HeaderStyleProps } from '../../common/SectionHeader';
 import SimpleTable, { NameValueTable, NameValueTableRow } from '../../common/SimpleTable';
+import { LightTooltip } from '..';
```
ashu8912, line 38: looks like this should not belong here

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -19,16 +19,14 @@ import * as monaco from 'monaco-editor';
...
-import { LightTooltip } from '..';
 import DetailsViewPluginRenderer from '../../../helpers/renderHelpers';
 import { ApiError } from '../../../lib/k8s/apiProxy';
 import {
   KubeCondition,
   KubeContainer,
   KubeContainerStatus,
   KubeObject,
-  KubeObjectInterface
+  KubeObjectInterface,
```
ashu8912, line 29: this should not belong here

ashu8912 (commented):
Tested and it LGTM, just a few comments

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: Not sure. It does have a trailing whitespace that should not be there, but it's not part of this patch. Thanks.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: True. I think I copied & paste the entire file due to changing a few things when arranging the commits, and may have removed more than I thought.

PR #380, opened 2021-12-22
frontend: Correctly filter ready pods in the overview chart

joaquimrocha:
The overview chart was showing pods that are running or not, but it's
more interesting and useful to show the pods that are ready and not
ready instead.

### DIFF
--- a/frontend/src/components/cluster/Charts.tsx
+++ b/frontend/src/components/cluster/Charts.tsx
@@ -4,6 +4,7 @@ import _ from 'lodash';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { KubeObject } from '../../lib/k8s/cluster';
+import Pod from '../../lib/k8s/pod';
 import { parseCpu, parseRam, TO_GB, TO_ONE_CPU } from '../../lib/units';
 import { HeaderLabel } from '../common';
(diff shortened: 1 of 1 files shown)


frontend/src/components/cluster/Charts.tsx
```diff
@@ -164,9 +165,14 @@ export function PodsStatusCircleChart(props: Pick<ResourceCircularChartProps, 'i
...
 
-  const podsReady = (items || []).filter(pod =>
-    ['Running', 'Succeeded'].includes(pod.status!.phase)
-  );
+  const podsReady = (items || []).filter((pod: Pod) => {
+    if (pod.status!.phase === 'Succeeded') {
+      return true;
+    }
+
+    const readyCondition = pod.status.conditions.find(condition => condition.type === 'Ready');
```
ashu8912, line 173: I think we should change this to 
```const readyCondition = pod.status?.conditions.find(condition => condition.type === 'Ready');``` to not allow any unknown property invocation

ashu8912:
Overall the pr LGTM!! just left a small comment

frontend/src/components/cluster/Charts.tsx
joaquimrocha: The pod is supposed to have the status, but I will change it with defensiveness in mind.

PR #381, opened 2021-12-22
frontend: Add links to the events list

joaquimrocha:
The events list shows important occurrences in the cluster, and it's
very useful to have a way to go right to the resource in question.

So this patch adds a link to the involved object, when possible.

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -440,7 +440,7 @@ function LivenessProbes(props: { liveness: KubeContainer['livenessProbe'] }) {
   return (
     <Box display="flex" flexDirection="column">
       <LivenessProbeItem>
-        {`http-get, path: ${liveness?.httpGet?.path}, port: ${liveness?.httpGet?.port}, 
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -440,7 +440,7 @@ function LivenessProbes(props: { liveness: KubeContainer['livenessProbe'] }) {
   return (
     <Box display="flex" flexDirection="column">
       <LivenessProbeItem>
-        {`http-get, path: ${liveness?.httpGet?.path}, port: ${liveness?.httpGet?.port}, 
+        {`http-get, path: ${liveness?.httpGet?.path}, port: ${liveness?.httpGet?.port},
```
ashu8912, line 443: I think this should not belong here

ashu8912:
I have tested this and the links work perfectly thanks, just a small comment

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: Yeah. Our lint checks should have detected this whitespace before we pushed it to main, but I will add a new commit just fixing that.

PR #382, opened 2021-12-22
frontend: Simplify and fix advanced search

joaquimrocha:
The advanced filter search (searching by non-default criteria) was
not working correctly because it was mixing non-filtered matches with
the matched ones, e.g. if we selected a namespace and then used the
advance filter matching (by matching the "kind" in the workload
overview), it was showing the correctly matched kind, but with data
from all namespaces.

This patch not only fixes that issue, but also simplifies how the
advanced filter is implemented (in the filter function rather than
in the SimpleTable component) and uses JSONPath to make it a bit more
powerful as well.


ashu8912 (approved):
Tested

PR #383, opened 2021-12-22
Fix focus on search

joaquimrocha:
- frontend: Fix focusing the filter's search entry
- frontend: Make strings in the SectionFilterHeader translatable


ashu8912 (approved):
tested, LGTM

PR #388, opened 2022-01-06
plugins: Allow to use shared modules with the import syntax

### LINKED ISSUE #324: Import module types to plugins
## Current situation

Right now all the modules we attach to pluginLib are passed as type any because of which there is no way we can use all the typescript benefits from those modules such as strict types, hinting on objects, etc

We should have concrete types on modules that are present on the pluginLib

joaquimrocha:
This will allow plugin writers to import React and other modules
without having to install them and without having webpack bundle them
as they will be available when importing the plugin's code in Headlamp.

TBD:
- [x] Install the shared modules in @kinvolk/headlamp-plugin so they show up in its node_modules
- [x] Make VSCode find the shared modules (in the @kinvolk/headlamp-plugin/node_modules), using a jsconfig.json file
- [x] Share also the Headlamp's modules, not just npm shared modules (non-headlamp ones)
- [x] Offer a better way to call registerPlugin

fixes #324 

cc/ @illume , @ashu8912


joaquimrocha:
This should be good to test now. It ships the shared modules we promote in Headlamp's plugins (`@material-ui`, `react`, etc.), type declarations for Headlamp's modules are shipped, js/tsconfig is configured to find the modules, webpack config will replace shared module imports by the respective globals.

I'd like to be sharing less type declarations than we are but they seem to be tied together as dependencies. Also we should check whether we need to ship the transpiled js files.

plugins/headlamp-plugin/config/webpack.config.js
```diff
@@ -6,6 +20,26 @@ module.exports = {
   resolve: {
     extensions: ['.tsx', '.ts', '.js', '.jsx']
   },
+  externals: function ({ request }, callback) {
```
illume, line 23: 👍 Nice.

https://webpack.js.org/configuration/externals/#function

plugins/headlamp-plugin/template/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "allowSyntheticDefaultImports": true,
+    "jsx": "react-jsx",
+    "moduleResolution": "node",
+    "baseUrl": ".",
+    "paths": {
```
illume, line 8: 👍 

https://www.typescriptlang.org/tsconfig#paths
>A series of entries which re-map imports to lookup locations relative to the baseUrl, there is a larger coverage of paths in the handbook.
> paths lets you declare how TypeScript should resolve an import in your require/imports.

https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

plugins/headlamp-plugin/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.3.2",
   "description": "The needed infrastructure for building Headlamp plugins.",
   "bin": "bin/headlamp-plugin.js",
+  "scripts": {
```
illume, line 6: Seems like this needs to be run any time there is a change in the frontend (types).

So we should add a CI check which runs `npm run build` and checks for git changes.

frontend/src/plugin/index.ts
```diff
@@ -0,0 +1,107 @@
+import helpers from '../helpers';
```
illume, line 1: Nice separation.  Would be nice to add a note to the top of each file explaining what is in the git commit message.

```ts
/**
 * The lib.ts file should carry the bits to be used by plugins whereas
 * the index.ts should have the functions that Headlamp itself needs for
 * loading the plugins.
*/

/**
 * The index.ts should have the functions that Headlamp itself needs for
 * loading the plugins.
 * The lib.ts file should carry the bits to be used by plugins whereas
*/

```

docs/development/plugins/functionality.md
```diff
@@ -8,38 +8,46 @@ the user interface and experience.
...
-Apart from the external modules above, the `pluginLib` contains of course
-modules that are related to Headlamp, so developers can use the cluster and
-web UI related functionality. Those modules are:
-
-* K8s
-* CommonComponents
-
-This means that you can just declare a `const React = window.pluginLib.React` in order to use
-React without having to import it.
+The `@kinvolk/headlamp-plugin` module ships a library
```
illume, line 11: 👍

illume (changes requested):
Looking good. I just read through it so far. Left a few notes.

The example plugin needs updating? plugins/examples/pod-counter.

How can I test this? I was going to run through the docs creating a plugin again and see how that goes. I think that will be enough.

plugins/headlamp-plugin/package.json
joaquimrocha: @illume , I have added the CI check for `npm run build`. You mentioned you had done something similar to what's needed for the type check, so maybe you can add that commit or point me in the right direction.

joaquimrocha: I actually removed the patch now because I forgot that we're not tracking the generated types. We should still add the check for package versions.

ashu8912 (commented):
The pr LGTM!!
I just have this  one small question/doubt

1. From what I understand this PR helps us in resolving imports via headlamp-plugin one thing that is unclear to me is so we in the webpack config mentioned externals where we say to webpack please don't look for these in the builds rather look it onto the global pluginLib that we expose from headlamp,  Now if that's the case why do we still have some of the dependencies on the headlamp-plugin side shouldn't those be resolved via the global object, because to me it looks like a mix of both the things where we want to have everything in headlamp-plugin itself and then also expect to resolve the imports from the global object(via headlamp-plugin).

plugins/headlamp-plugin/config/webpack.config.js
```diff
@@ -1,3 +1,17 @@
+const externalModules = {
```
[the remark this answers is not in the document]
joaquimrocha: The way this approach works is:
1. Headlamp's frontend already contains all the dependencies it needs for its UI, and plugins should use those too (not bundle them).
2. headlamp-plugin requires installing the dependencies useful for plugins, which should match the imports added to window.pluginLib and should match the versions in Headlamp's frontend. The reason why we add them here is to make them available for the plugin's editor to resolve the imports, e.g. for VS Code to find the modules and offer completion. These modules are not used to bundle the plugin code, i.e. they're not transpiled.
3. external modules is a way to tell *webpack* how to translate the imports. We do not want to bundle the dependencies code in the plugins, because that would risk using a different version of modules that already exist in Headlamp, and increase bundle size unnecessarily. So webpack translates e.g. `import {Button} from @material-ui/core` to `const { Button } = pluginLib.MuiCore`.

Conclusion: the way the code works should be exactly as if the user was still using `pluginLib`, the headlamp-plugin dependencies are just for helping with the coding experience.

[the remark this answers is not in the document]
joaquimrocha: We could eventually make Headlamp itself look for dependencies from the plugins/headlamp-plugin . But let's think about that in the future.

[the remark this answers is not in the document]
ashu8912: Thanks @joaquimrocha this does answer my question.

ashu8912 (approved):

illume:
I was thinking how we could have a test for this so it keeps working...

How about running the type checker and linter on the example plugin? This would help us know this functionality keeps working.

The plugin test uses npm link and npx, so it works with the version of the plugin code that's checked in and not the version that's released. See headlamp/plugins/headlamp-plugin/test-headlamp-plugin.sh

frontend/src/plugin/lib.ts
```diff
@@ -0,0 +1,141 @@
...
+ * loading the plugins.
+ */
+
+/**
+ * ## Simple plugin example
+ *
+ *
+ * @example
+ * // Here's a very simple JavaScript plugin example, that does nothing.
+ * import { Headlamp } from '@kinvolk/headlamp-plugin/lib';
```
illume, line 13: How about just `import { registerPlugin } from '@kinvolk/headlamp-plugin/lib';` instead? I don't think the Headlamp class is needed here.

frontend/src/plugin/lib.ts
```diff
@@ -0,0 +1,141 @@
...
+   *     // do some stuff with register
+   *     // use some libraries in window.pluginLib
+   *     return true;
+   *   }
+   * }
+   *
+   * Headlamp.registerPlugin("aPluginIdString", myPlugin)
+   * ```
+   */
+  static registerPlugin(pluginId: string, pluginObj: Plugin) {
```
illume, line 134: Regarding [323](https://github.com/kinvolk/headlamp/issues/323#issuecomment-1020125006), I'm wondering if since we will change the API here... is the time to change the registerPlugin interface to address that problem? (Perhaps to use a react component rather than a plugin object with an initialize method).

frontend/src/plugin/lib.ts
joaquimrocha: I wonder if it's useful to have it as a namespace kind of helper.
We could also default export a list of plugins, like:

```
export default PLUGINS = [
   MyPlugin
];
```

illume (approved):
🎉 thanks

PR #390, opened 2022-01-12
frontend: Fix make docs, add building docs to CI

illume:
# docs fix

- `make docs` works again, 
- and is now run in CI (to hopefully catch it if it breaks again).

There were some dependencies updated in https://github.com/kinvolk/headlamp/pull/337 which broke typedoc. Updated to newer versions of typedoc and friends fixes it.


joaquimrocha:
Good stuff. Thanks.

PR #394, opened 2022-01-14
Compiling Headlamp app on Windows fixes

illume:
# There were a few issues when building the app on windows

export, mkdir -p, and `''` quoting were problematic

Additionally. the github app windows build now doesn't use bash with npm. This is a test which shows this PR works without bash on windows now.

## Testing done

- [x] `npm run dev` starts and seems to work
- [x] generated app installs, starts and seems to work 

ps. Note: server.exe will hang around on app close. So make sure to kill that with task manager if you are testing.  There's a PR for that here: https://github.com/kinvolk/headlamp/pull/395

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -8,8 +8,8 @@
   "scripts": {
     "compile-electron": "babel electron --out-dir electron/ --extensions '.ts'",
     "start": "cd ../ && make backend && make run-backend  & cd ../frontend/ && export BROWSER=none && npm start",
-    "prod-deps": "mkdir -p prod_deps && cd ./prod_deps && cp ../package.json ../package-lock.json . && npm i --only=prod && rm -rf ./node_modules/.bin",
-    "copy-icons": "mkdir -p build/icons && cp ../frontend/build/*.png ../frontend/build/*.ico ../frontend/build/*.icns ../frontend/build/*.svg build/icons",
+    "prod-deps": "mkdirp prod_deps && cd ./prod_deps && cp ../package.json ../package-lock.json . && npm i --only=prod && rm -rf ./node_modules/.bin",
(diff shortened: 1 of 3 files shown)


joaquimrocha (approved):
While it's good that we can make the app work on Windows without bash, I wonder if it's worth doing it. I am concerned with the fact that future changes to the build scripts will always need to use extra tools to ensure compatibility, instead of just using bash as a wrapper/standard for building it.

Is there a good argument for not using bash?

illume:
Bad DX. Bash is not standard on windows. Even when using bash, npm doesn't use bash by default.

ashu8912:
> Bad DX. Bash is not standard on windows. Even when using bash, npm doesn't use bash by default.

That is true at times I have to change the shell for windows manually to bash for npm scripts to work

app/package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": ".",
   "productName": "Headlamp",
   "scripts": {
-    "compile-electron": "babel electron --out-dir electron/ --extensions '.ts'",
+    "compile-electron": "babel electron --out-dir electron/ --extensions .ts",
```
ashu8912, line 9: @illume How is this probelmatic, I tried with both ways without and with quotes and it works

ashu8912 (commented):
Does ```cross-env``` needs to be an app dependency?? 
Shouldn't it be a dev-dependency?

ashu8912 (commented):
same here mkdirp should be a dev-dependency

app/package.json
illume: It runs, but does not compile the scripts (reports 0 compiled and should be something like 4). Note, this is when using cmd.

illume:
@ashu8912 `cross-env` and `mkdirp` are required to build for production. For that reason I think they are correctly placed as a dependency.

ashu8912 (approved):

app/package.json
ashu8912: aah I was using the wrong shell, with cmd it is a problem

Issue #160, opened 2021-01-12
configuration (server port)


illume:
**Current situation**

Currently the server process port is hard coded. Perhaps people will want to use another one.

Configuration of other things may be needed.

**Impact**

Some people won't be able to run Headlamp if it's not configured for their environment.

Each configuration increases the size of the testing matrix.

**Ideal future situation**

People can configure Headlamp so they are comfy.

**Implementation options**

There's already some command line options (for headless mode).

How could they configure this (and other things)?

- environment variables
- command line
- configure in the app (for people not using the command line)
- a configuration file (perhaps writable by the app but not when hosted)
- Headlamp can run in different ways: app, hosted, headless

knrt10:
User can now configure [port](https://github.com/headlamp-k8s/headlamp/blob/main/backend/pkg/config/config.go#L28) in headlamp. So, closing this.

PR #395, opened 2022-01-14
Fix windows app not closing server.exe on quit

illume:
# Now when the app is quit the server.exe process is also stopped.

Related to issue: https://github.com/kinvolk/headlamp/issues/389

## How to use

Check task manager for a server.exe process still around when the app quits.

Or with cmd you can find it with:
```
tasklist | findstr /c:"server.exe"
```

## Testing done

- [x] when app is quit it kills server.exe as well

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,4 +1,4 @@
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams, exec, spawn } from 'child_process';
 import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
@@ -73,7 +73,7 @@ let intentionalQuit: boolean;
 let serverProcessQuit: boolean;
 
 function quitServerProcess() {
-  if (!serverProcess || serverProcessQuit) {
+  if ((!serverProcess || serverProcessQuit) && process.platform !== 'win32') {
     log.error('server process already not running');
     return;
   }
@@ -84,6 +84,9 @@ function quitServerProcess() {
     // Negative pid because it should kill the whole group of processes:
     //    https://azimi.me/2014/12/31/kill-child_process-node-js.html
     process.kill(-serverProcess.pid);
+  } else if (process.platform === 'win32' && serverProcess) {
+    // Otherwise on Windows the process will stick around.
+    exec('taskkill /pid ' + serverProcess.pid + ' /T /F');
   }
 
   serverProcess.stdin.destroy();
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -6,6 +6,7 @@ on:
     - 'frontend/**'
     - 'Makefile'
     - '.github/**'
+    - 'app/**'
   push:
     branches:
       - main
@@ -13,6 +14,7 @@ on:
     - 'frontend/**'
     - Makefile
     - '.github/**'
+    - 'app/**'
 
 jobs:
   build:


app/electron/main.ts
```diff
@@ -1,4 +1,4 @@
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams, exec,spawn } from 'child_process';
```
joaquimrocha, line 1: Nit: This is missing a space after the new coma.
Maybe we should add prettier to the app as well.

joaquimrocha (approved):
Left a nit, but LGTM otherwise.

app/electron/main.ts
illume: Oh. Thanks. 

Prettier was working for the app previously... will look into it.

illume: CI wasn't checking it because that is done by the frontend workflow, and this PR only changed the app. So I fixed that.

prettier is broken when using windows git, because it reformats everything with line endings. I'll leave that issue for another day outside this PR.

ashu8912:
Screenshot (1)
@illume I was testing the PR but I still see this process floating around

ashu8912:
```
[2022-01-17 21:08:15.338] [info]  server process stdout: *** Headlamp Server ***
  API Routers:
	localhost:4466/clusters/acela/{api...} -> https://acela.dev.lokomotive-k8s.net:6443

[2022-01-17 21:08:15.342] [info]  server process stdout: 	localhost:4466/clusters/minikube/{api...} -> https://127.0.0.1:55953

[2022-01-17 21:08:15.347] [error] server process stderr: 2022/01/17 21:08:15 listen tcp :4466: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.

[2022-01-17 21:08:15.359] [error] server process process exited with code:1 signal:null
```
also electron log

joaquimrocha:
Maybe we should have a dialog informing the user that we couldn't run the server because another one is running (and occupying the needed port which is why the server fails to launch). (Besides making sure the server is killed on quit)

illume:
@joaquimrocha wrote:
> Maybe we should have a dialog informing the user that we couldn't run the server because another one is running (and occupying the needed port which is why the server fails to launch). (Besides making sure the server is killed on quit)

Yeah, we should do something like that. 

I did start this issue to discuss some options: https://github.com/kinvolk/headlamp/issues/397
Probably displaying a dialog with the error message would be enough.
However, we could also:
- offer to kill that server.exe
- offer to run on a separate port (or just do that automatically)

ashu8912:
> @ashu8912 thanks for trying it out. Can you please confirm there was no server.exe running before you started Headlamp? This PR doesn't cover closing existing server.exe. It only should kill the server.exe started by itself.

@illume can you please throw more light on what do you mean by killing only the server.exe started by itself?

illume:
> > @ashu8912 thanks for trying it out. Can you please confirm there was no server.exe running before you started Headlamp? This PR doesn't cover closing existing server.exe. It only should kill the server.exe started by itself.
> 
> @illume can you please throw more light on what do you mean by killing only the server.exe started by itself?

Sure. Perhaps there's already a server.exe running on your system from a previous run of Headlamp.

1. make sure no server.exe is running
```
tasklist | findstr /c:"server.exe"tasklist | findstr /c:"server.exe"
# taskkill /pid XXXX /T /F
```
2. run Headlamp
3. close headlamp
4. check no server.exe is running


kill-server-exe

joaquimrocha:
> * offer to kill that server.exe

I worried we cannot guarantee that the server.exe is ours 🙂 

> * offer to run on a separate port (or just do that automatically)

This would be ideal (and I thought we had an issue for it), but it requires passing that info over to the client code (doable but as not trivial).

illume:
Oh, there is an old issue about being able to configure the server port: https://github.com/kinvolk/headlamp/issues/160

illume:
@joaquimrocha @ashu8912 I'd like to merge this one in, and then start working on 
- allow using separate server port #160
- tell the user if the port is taken, and offer the user to use another port as part of #397

That ok?

PR #396, opened 2022-01-17
frontend: Prevent repeated details view rendering

ashu8912:
Details View that is meant to be rendered by a plugin was rendering repeated views at times, This patch prevents rendering of a section more than once. 

## Testing done
To reproduce this behaviour on main try changing the theme and you will see multiple rendering of the details view, on this branch the above behaviour is not reproducible

### DIFF
--- a/frontend/src/redux/reducers/ui.tsx
+++ b/frontend/src/redux/reducers/ui.tsx
@@ -163,8 +163,18 @@ function reducer(state = INITIAL_STATE, action: Action) {
     }
     case UI_SET_DETAILS_VIEW: {
       const { sectionName, action: sectionFunc } = action;
-      const detailViews = [...newFilters.views.details.pluginAppendedDetailViews];
-      detailViews.push({ sectionName, sectionFunc });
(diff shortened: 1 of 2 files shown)


frontend/src/redux/reducers/ui.tsx
```diff
@@ -163,8 +163,18 @@ function reducer(state = INITIAL_STATE, action: Action) {
...
+      const detailViews = [...pluginDetailViews];
+      // only push if it's a unique section
+      if (
+        detailViews.length === 0 ||
+        !detailViews.find(
+          (view: { sectionName: string; sectionFunc: sectionFunc }) =>
+            view.sectionName === sectionName
+        )
+      ) {
+        detailViews.push({ sectionName, sectionFunc });
```
joaquimrocha, line 176: I think this cures the symptom, not the cause, of somehow getting the register* functions called more than once for existing plugins. Maybe it's worth investigating further to realize exactly why this is happening.

If this solution should still be added, then I think the correct thing to do may be to replace the entry in the details views, so we can replace any older code with new one.

joaquimrocha:
Closing in favor of #403 which is a different and cleaner approach.

PR #399, opened 2022-01-19
frontend: Fix applying non-grouped API resources

joaquimrocha:
Some APIs do not have a group and we were parsing the YAML added to
the creation dialog in a way that it expected to have a group.


illume:
Do you know where this is specified? (resourceDef.apiVersion) I'd like to see if this handles the different formats... but not sure what they look like.

joaquimrocha:
> Do you know where this is specified? (resourceDef.apiVersion) I'd like to see if this handles the different formats... but not sure what they look like.

I believe this is the place: https://kubernetes.io/docs/reference/using-api/#api-groups

ashu8912 (approved):

PR #400, opened 2022-01-19
frontend: Make workload details go back to the correct list view

joaquimrocha:
The back link from their details view would go to the workloads'
list view but it should go to the resources' own list view.


illume (approved):
👍 🎉 thanks

PR #402, opened 2022-01-20
Better info for workloads

joaquimrocha:
This PR adds more information to different workflow details views as well as a couple of bug fixes related to that.

### DIFF
--- a/frontend/src/components/cronjob/Details.tsx
+++ b/frontend/src/components/cronjob/Details.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import DetailsViewPluginRenderer from '../../helpers/renderHelpers';
+import CronJob from '../../lib/k8s/cronJob';
+import Job from '../../lib/k8s/job';
+import { DetailsGrid } from '../common/Resource';
+import { JobsListRenderer } from '../job/List';
+import { getLastScheduleTime, getSchedule } from './List';
+
+export default function CronJobDetails() {
+  const { namespace, name } = useParams<{ namespace: string; name: string }>();
+  const { t, i18n } = useTranslation('glossary');
+
+  const [jobs, jobsError] = Job.useList();
+
+  function filterOwnedJobs(jobs?: Job[] | null) {
+    if (!jobs) {
+      return null;
+    }
+
+    return jobs.filter(job => {
+      type OwnerRef = {
+        name: string;
+        kind: string;
+      };
+      return !!job.metadata?.ownerReferences?.find(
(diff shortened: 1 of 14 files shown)


frontend/src/components/pod/List.tsx
```diff
@@ -22,7 +22,7 @@ export function makePodStatusLabel(pod: Pod) {
     status = 'error';
   } else if (phase === 'Succeeded' || phase === 'Running') {
     const readyCondition = pod.status.conditions.find(condition => condition.type === 'Ready');
-    if (readyCondition?.status === 'True') {
+    if (readyCondition?.status === 'True' || phase === 'Succeeded') {
```
ashu8912, line 25: This makes me think should we add a blue color for all the succeeded ones because now running and succeeded have the same color

frontend/src/components/daemonset/Details.tsx
```diff
@@ -29,6 +29,10 @@ export default function DaemonSetDetails() {
             name: t('Selector'),
             value: <MetadataDictGrid dict={item.spec.selector.matchLabels || {}} />,
           },
+          {
+            name: t('Node Selector'),
```
ashu8912, line 88: Thanks, While testing this I found a issue with the DaemonSet details view looks like the back button is broken there it should take us to the list view but it right now takes us to the workload overview view

ashu8912: can we add a fix for that as well in this PR itself??

frontend/src/components/cronjob/Details.tsx
```diff
@@ -0,0 +1,73 @@
...
+          {
+            name: t('Last Schedule'),
+            value: getLastScheduleTime(item),
+          },
+        ]
+      }
+      sectionsFunc={item => (
+        <>
+          {item && (
+            <>
```
ashu8912, line 64: I think we don't need this fragment here

frontend/src/components/daemonset/Details.tsx
ashu8912: On further testing looks like the back button is broken for 
```DaemonSets, StatefulSets, Jobs, CronJobs```

ashu8912 (commented):
Just a few comments and nitpicks otherwise it is looking good to me, Thanks

frontend/src/components/pod/List.tsx
joaquimrocha: Let's worry about that later. I think Running and Succeeded can be green for now. But it's a good remark.

frontend/src/components/cronjob/Details.tsx
joaquimrocha: Good catch. I will update it.

joaquimrocha:
Updated the PR after your comments. Thanks @ashu8912 .

ashu8912 (approved):

PR #404, opened 2022-01-20
frontend: Use the current theme for the ConfigMaps' editor

joaquimrocha:
The ConfigMap details view was using the editor to display data always
in the dark theme. This patch makes it aware of the theme changes and
lets it change accordingly.


illume (approved):
👍thanks :)

illume:
ps. Whilst testing this, I noticed there's some responsive issues on that page too. I'll address them in a separate PR.

PR #411, opened 2022-01-28
Fix UI warnings

joaquimrocha:
- frontend: Replace Grid's justify prop by justifyContent
- frontend: Don't use deprecated props for dialogs
- frontend: Don't use deprecated onBackdropClick property on dialogs


illume (approved):
👍 🎉 thanks

PR #413, opened 2022-02-01
Typedoc plugin upgrade, frontend package lock v2, removed node 14 from frontend CI check

illume:
Update to newer version of typedoc-plugin-markdown that's compatible with other newer packages (newer node/ts/eslint/etc). Required to update so can [update to CRA 5/webpack5/storybook 4.6](https://github.com/kinvolk/headlamp/pull/405).

Here's the [changelog for new typedoc-plugin-markdown](https://github.com/tgreyuk/typedoc-plugin-markdown/blob/master/packages/typedoc-plugin-markdown/CHANGELOG.md#31114-2022-02-14)

Also:
- removed node 14 CI workflow, node 18 isn't out yet and 17 is too new
- updated package-lock.json to version 2 with node 16, some dependencies were updated as a result


docs/development/api/classes/plugin_lib.Headlamp.md
```diff
@@ -0,0 +1,53 @@
...
+
+This class is a more convenient way for plugins to call registerPlugin in order to register
+themselves.
+
+## Constructors
+
+### constructor
+
+• **new Headlamp**()
+
```
joaquimrocha, line 17: We only have a method in this class and it's a static one. Should we still mention the constructor here?

docs/development/api/classes/plugin_lib.Headlamp.md
```diff
@@ -0,0 +1,53 @@
...
+▸ `Static` **registerPlugin**(`pluginId`, `pluginObj`): `void`
+
+Got a new plugin to add? Well, registerPlugin is your friend.
+
+**`example`**
+
+```javascript
+const myPlugin = {
+  initialize: (register) => {
```
joaquimrocha, line 30:
```suggestion
  initialize: (registry) => {
```

PR #415, opened 2022-02-02
WIP: frontend: Centralize k8s listing requests

joaquimrocha:
As the user navigates the UI, there may be several calls to k8s
server for listing and getting resources. This means that a quick
navigation may retrieve new objects that were already in memory.
This not only implies more IO but may also lead to more memory usage
by Headlamp on the client side.

This will also allow us to do things like searching through a global cache of resources more easily.

Tasks to be done:
- [X] Adding a way to centralize and refcount the requests (POC commit with a small lifetime for the cache of 2s is done)
- [ ] Increase the ref=0 lifetime (maybe 10 seconds is enough)
- [ ] Verify whether we need to use timestamps in the cache
- [ ] Update the cache when using `useGet`


ashu8912:
@joaquimrocha wow this is a big PR, I will review it once I finish the notification part is done

joaquimrocha:
This is now accomplished by react-query. Closing.

PR #424, opened 2022-02-09
frontend: Upgrade to new iconify 3.1.3

joaquimrocha:
Also move to a string-based icon description.

### DIFF
--- a/frontend/src/components/App/__snapshots__/TopBar.stories.storyshot
+++ b/frontend/src/components/App/__snapshots__/TopBar.stories.storyshot
@@ -56,20 +56,7 @@ exports[`Storyshots TopBar No Token 1`] = `
         <span
           class="MuiIconButton-label"
         >
-          <svg
-            focusable="false"
-            height="1em"
-            preserveAspectRatio="xMidYMid meet"
-            style="transform: rotate(360deg);"
-            viewBox="0 0 24 24"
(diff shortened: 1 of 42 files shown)


frontend/src/components/App/__snapshots__/TopBar.stories.storyshot
```diff
@@ -56,20 +56,7 @@ exports[`Storyshots TopBar No Token 1`] = `
         <span
           class="MuiIconButton-label"
         >
-          <svg
```
illume, line 59: Looks like there's no SVG anymore, and just a span instead.

joaquimrocha:
@illume , I would like to merge this but we can eventually break existing plugins because of this required change:
```javascript
import Icon from '@iconify/react';
```

Do you think that we can do any magic on the headlamp-plugin level to still support the legacy default import? Or some other solution?

illume:
Yeah I think we can do some sort of fixer for that in a `headlamp-plugin update` style thing. Added it to the migration issue: https://github.com/kinvolk/headlamp/issues/426#issuecomment-1058714643

I tested this with latest main, and it's ok. The storyshots issue isn't important IMHO, and I hope it will be fixed with other dependency updates coming up (if it even really needs fixing at all).

illume (approved):
🎉🎈

PR #427, opened 2022-02-14
Dockerfile: Fix alpine version typo, and separate front and back building

illume:
So that if a backend part is changed, then the frontend image is not rebuilt and vice versa.
Also if a frontend/src change is made, docker build doesn't do an npm install again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,19 @@ ENV GOPATH=/go \
 RUN apk update && \
 	apk add git nodejs npm go ca-certificates make musl-dev bash icu-data
 
-COPY ./ /headlamp/
+FROM base-build as backend
+
+COPY ./backend /headlamp/backend
 
 WORKDIR /headlamp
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,21 +8,24 @@ ENV GOPATH=/go \
...
+FROM base-build AS frontendinstall
+COPY frontend/package*.json /headlamp/frontend/
+WORKDIR /headlamp
+RUN cd ./frontend && npm install --only=prod
 
+FROM frontendinstall AS frontend
 COPY ./frontend /headlamp/frontend
-
 WORKDIR /headlamp
-
```
joaquimrocha: I'd keep these empty lines, for clarity.

joaquimrocha (approved):
Approved with a comment. Feel free to merge after changing.

Dockerfile
illume: Ok, thanks.

Issue #277, opened 2021-06-24
[RFE] Allow to modify style/branding parts from plugins


joaquimrocha:
## Current situation

If we want to set up a new logo/name for a Headlamp deployment, we have to change the project's code.

## Impact

We want Headlamp to be a flexible/customizable web UI for Kubernetes and that includes giving more creative options in terms of branding too.

## Ideal future situation

We have methods to set up a name and logo in the UI. Other functionalities are to be defined.

## Additional information

This new functionality is such that we should be able to set a different project logo depending on a variety of triggers: cluster's info, number of cluster, etc.

We should also be able to build the lokomotive-web-ui without having to modify Headlamp's source code.

joaquimrocha:
We can consider this done since there are a lot more options for branding like menu + logo overrides.

PR #429, opened 2022-02-15
Add functionality to modify app logo from plugin

ashu8912:
Setup headlamp to handle logo change via plugin


plugins/headlamp-plugin/package.json
```diff
@@ -28,7 +28,7 @@
     "react-router-dom": "^5.3.0",
     "recharts": "^2.1.8",
     "redux": "^4.1.2",
-    "typescript": "^4.5.4",
```
illume, line 31: Seems typescript was removed accidentally.

joaquimrocha (changes requested):
Branding is a term that encompasses several aspects about the visuals of a product.
So we should not really mention branding as the thing the plugin is changing if it's changing just the logo.

plugins/headlamp-plugin/package.json
ashu8912: fixed

ashu8912:
One part of #277

frontend/src/plugin/registry.tsx
```diff
@@ -126,4 +128,16 @@ export default class Registry {
...
+  /**
+   *
+   * @param branding all the branding change related resources that are specified in the ```JSX BrandingProps``` interface
+   *
+   * @example
+   * ```JSX
+   * register.registerBrandingChange({smallScreenLogo: File, largeScreenLogo: File})
+   * ```
+   */
+  registerBrandingChange(branding: BrandingProps) {
```
joaquimrocha, line 140: I thought I had left already a comment in a different review explaining that "branding" is a broad term, and that this plugin changes the app logo which should be specific name. Maybe `registerMainLogo`.

joaquimrocha (changes requested):

ashu8912:
@joaquimrocha the CI is failing here I am not sure why it didn't take the patch about not using localeTime in stories

frontend/src/components/pod/__snapshots__/Pod.stories.storyshot
```diff
@@ -213,7 +213,7 @@ exports[`Storyshots pods/PodListRenderer Pods 1`] = `
               <td
                 class="MuiTableCell-root MuiTableCell-body"
               >
-                just now
+                1 month
```
joaquimrocha, line 216: This is what's making the CI fail.

frontend/src/plugin/registry.tsx
```diff
@@ -16,6 +17,17 @@ interface SectionFuncProps {
...
 
+export type getAppLogoFunc = (props: {
+  logoType?: LOGO_TYPE;
+  theme?: string;
+  [key: string]: any;
+}) => JSX.Element | null;
+
+export enum LOGO_TYPE {
+  'SMALL' = 'SMALL',
+  'LARGE' = 'LARGE',
```
joaquimrocha, line 28: This looks a bit odd. Maybe we can make it be just be an object? (so it doesn't have to be used from Ts)

@illume , do you have a suggestion for this?

frontend/src/redux/reducers/ui.tsx
```diff
@@ -188,6 +197,14 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
       newState.sidebar = { ...newState.sidebar, ...setInitialSidebarOpen() };
       return newState;
     }
+    case UI_APP_LOGO_BRANDING: {
+      newFilters.branding.logo = action.appLogoComponent;
```
joaquimrocha, line 201: This is not a component, it's a function that returns a component.

joaquimrocha (changes requested):
Left a few comments, but it's almost there.

frontend/src/plugin/registry.tsx
```diff
@@ -126,4 +138,17 @@ export default class Registry {
...
+   *
+   * @param appLogoGeneratorFunc a logo generator function which returns a react component, this component takes two optional props
+   * logoType which is an enum @enum {LOGO_TYPE} that has two values SMALL and LARGE for screen sizes and the other argument is the theme ```JSX Theme``` interface
+   *
+   * @example
+   * ```JSX
+   * register.registerAppLogoChange(("SMALL") => <img src="/small_logo" />)
+   * ```
+   */
+  registerAppLogoChange(appLogoGeneratorFunc: getAppLogoFunc) {
```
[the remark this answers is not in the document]
illume: What about `registerAppLogoComponent`? 
(This would require the function signature being changed.)

frontend/src/redux/reducers/ui.tsx
illume: I made a suggestion elsewhere that we pass in a component instead. This would need to be kept if we use that suggestion.

frontend/src/plugin/registry.tsx
illume: It can be written a bit more clearly like this: `"small" | "large"`

If we pass in a component instead this would turn into the props interface. The function type here seems to already use props.

frontend/src/components/Sidebar/HeadlampButton.tsx
```diff
@@ -56,11 +61,24 @@ export default function HeadlampButton({ open, onToggleOpen, mobileOnly }: Headl
...
-          component={open ? LogoWithTextLight : LogoLight}
-          viewBox="0 0 auto 32"
-        />
+        {
+          // Till all plugins are not loaded show empty content for logo as we might have logo coming from a plugin
+          pluginLoadState === false ? (
+            <EmptyContent />
+          ) : AppLogoGeneratorComponent ? (
+            <AppLogoGeneratorComponent
+              logoType={open ? LOGO_TYPE.LARGE : LOGO_TYPE.SMALL}
```
illume, line 70: You could get rid of LOGO_TYPE here if you used `"small" : "large"` and relied on typescript to check the strings.

frontend/src/plugin/registry.tsx
```diff
@@ -126,4 +138,17 @@ export default class Registry {
   registerDetailsViewSection(sectionName: string, sectionFunc: sectionFunc) {
     store.dispatch(setDetailsView(sectionName, sectionFunc));
   }
+  /**
+   *
+   * @param appLogoGeneratorFunc a logo generator function which returns a react component, this component takes two optional props
+   * logoType which is an enum @enum {LOGO_TYPE} that has two values SMALL and LARGE for screen sizes and the other argument is the theme ```JSX Theme``` interface
```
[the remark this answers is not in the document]
ashu8912: @joaquimrocha I am a little confused are you suggesting to return a component or a function that returns a component? Using a component made more sense because All the state related info can be passed to the component via prop.
cc: @illume

illume:
@ashu8912 I pushed a couple of small type check fixes causing the ci to fail `npm run tsc`.

frontend/src/plugin/registry.tsx
```diff
@@ -126,4 +133,17 @@ export default class Registry {
...
     store.dispatch(setDetailsView(sectionName, sectionFunc));
   }
+  /**
+   *
+   * @param appLogoGeneratorFunc a logo generator function which returns a react component, this component takes two optional props
+   * logoType which is a string literal that can take two values small or large for screen sizes and the other argument is the theme ```JSX Theme``` interface
+   *
+   * @example
+   * ```JSX
+   * register.registerAppLogoChange(() => <YOUR_COMPONENT logoType="small"/>)
```
illume, line 143: This example seems wrong?

frontend/src/plugin/registry.tsx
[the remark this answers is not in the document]
illume: I made a suggestion here: https://github.com/kinvolk/headlamp/pull/429#issuecomment-1068571834 resolving this to clean things up.

ashu8912:
Thanks @illume this is indeed a better approach.

Adding my comment here for why 
1. Storing directly the component into the store is a problem
2. Storing a function that returns a component into store is a problem

For the first case we would run into problems of inline component rendering where our store component gets rerendered each time the component that is asking it from the store is rerendered.

For the second case we would overcome the first problem because now our component reference is retained but this is still not recommended because functions are non-serializable and storing non serializable properties in redux store is not recommended.

illume:
Looks like the fixes for the typescript errors were overwritten. See `npm run tsc`.

joaquimrocha:
> @ashu8912 I pushed a couple of small type check fixes causing the ci to fail `npm run tsc`.
> Ok... most of how we could do it is here (untested and not very refined...):
> 
> ```tsx
> // Example How we use the API from the plugin
> 
> function MyLogo({logoType, theme}: {
>   logoType?: 'small' | 'large';
>   theme?: string;
>   [key: string]: any;
> }) {
>   return 
> }
> registerAppLogo(MyLogo)
> 
> 
> // Implementation of registerAppLogo
> registerAppLogo(Component: null | ReactElement) {

I am a bit confused. This Component arg is null or a ReactElement, but its being used above with a function, which AFAIU is not a ReactElement, it just does return one, but we're not calling the function.

>   setPluginAppLogo(Component)
>   // note we don't send the component to the redux action
>   store.dispatch(setAppLogo(component === null ? false : true));
> }
> 
> // Here we set the logo component singleton
> let pluginLogo = null; 
> function setPluginAppLogo(component : null | ReactElement) { 
>   pluginLogo = component
> }
> function getPluginAppLogo() : null | ReactElement {
>   return pluginLogo;
> }

This does accomplish what we want to do. If using functions as we've been doing in Redux is not recommendable, then maybe we can just have a global action store to abstract cases like the logo and others? The idea would be to have keys stored in redux, and using those to access this global object where the functions are defined.

We sort of had names for actions in the beginning in order to have a more controllable way of using such keys (though it was intended inside Redux). But of course, using unique keys (generated or logical, like "main-app-logo") may make sense instead.

> 
> // Now we select the state, and if there's a logo, we can get that logo.
> const pluginLogo = useTypedSelector(state => state.ui.branding.logo);
> const arePluginsLoaded = useTypedSelector(state => state.ui.pluginsLoaded);
> let PluginAppLogoComponent;

Right. With my suggestion, this could be a hook, like:
const getPluginAppLogo = usePluginComponent('name-of-my-key');
 
> if (pluginLogo) {
>   PluginAppLogoComponent = getPluginAppLogo();
> }
> ```

illume:
Yeah, a usePluginComponent with keys would be a nice way to extend things via plugins.
```tsx
const getPluginAppLogo = usePluginComponent('name-of-my-key');
```

> I am a bit confused. This Component arg is null or a ReactElement, but its being used above with a function, which AFAIU is not a ReactElement, it just does return one, but we're not calling the function.

MyLogo there is a functional component, and it would take all the props a logo needs.

frontend/src/redux/actions/actions.tsx
```diff
@@ -20,6 +20,15 @@ export const UI_APP_BAR_SET_ACTION = 'UI_APP_BAR_SET_ACTION';
...
 export const UI_RESET_PLUGIN_VIEWS = 'UI_RESET_PLUGIN_VIEWS';
 export const UI_PLUGINS_LOADED = 'UI_PLUGINS_LOADED';
+export const UI_BRANDING_APP_LOGO_COMPONENT = 'UI_BRANDING_APP_LOGO_COMPONENT';
+
+export interface BrandingProps {
+  logo: React.ComponentType<{
+    logoType: 'small' | 'large';
+    themeName: string;
+    [key: string]: any;
+  }> | null;
```
joaquimrocha, line 30: I still find this a bit breaking from the current way we've used, but we probably require a full review into how to unify and make sure the API is as future proof as possible.
If @illume is fine with this, let's let it through.

frontend/src/redux/actions/actions.tsx
```diff
@@ -20,6 +20,15 @@ export const UI_APP_BAR_SET_ACTION = 'UI_APP_BAR_SET_ACTION';
 export const UI_THEME_SET = 'UI_THEME_SET';
 export const UI_RESET_PLUGIN_VIEWS = 'UI_RESET_PLUGIN_VIEWS';
 export const UI_PLUGINS_LOADED = 'UI_PLUGINS_LOADED';
+export const UI_BRANDING_APP_LOGO_COMPONENT = 'UI_BRANDING_APP_LOGO_COMPONENT';
```
joaquimrocha, line 23: All the other action names here have verbs, so we should be consistent and call this `UI_BRANDING_SET_APP_LOGO`.

joaquimrocha (changes requested):
Still had a few comments going through this, but it's getting there.

frontend/src/components/Sidebar/HeadlampButton.tsx
```diff
@@ -44,6 +46,12 @@ export default function HeadlampButton({ open, onToggleOpen, mobileOnly }: Headl
   const isSmall = useMediaQuery('(max-width:600px)');
   const classes = useStyle({ isSidebarOpen: open, isSmall: isSmall });
   const { t } = useTranslation('sidebar');
+  const arePluginsLoaded = useTypedSelector(state => state.ui.pluginsLoaded);
+  let PluginAppLogoComponent;
+
+  if (arePluginsLoaded) {
+    PluginAppLogoComponent = useTypedSelector(state => state.ui.branding.logo);
+  }
```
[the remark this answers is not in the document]
ashu8912: oops, fixed

illume:
I think the plugin API for changing the logo is fine 👍

Probably it would be nice going forward if we use new plugin APIs inside an example plugin - but I don't think that should block this being merged. For the implementation, I'd prefer it was done using a more generic extension point as discussed in the meeting `const AppLogo = usePluginComponent('logo', DefaultLogo);`. But I don't think this should block this being merged either.

joaquimrocha (approved):
LGTM.
Let's wait on getting #458 to be fixed before we merge this, so I don't have to fiddle with git before publishing a new headlamp-plugin.

joaquimrocha:
> LGTM. Let's wait on getting #458 to be fixed before we merge this, so I don't have to fiddle with git before publishing a new headlamp-plugin.

Since the fix was in the backend, then there's no problem with merging this, as we won't have a new release for headlamp-plugin.

PR #430, opened 2022-02-16
backend: Update how we handle http callback URLs for oidc

ashu8912:
The conditions used to decide urlScheme for an http callback URI for oidc was broken

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -150,7 +150,7 @@ func getOidcCallbackURL(r *http.Request, config *HeadlampConfig) string {
 		switch {
 		case fwdProto != "":
 			urlScheme = fwdProto
-		case r.Host == "localhost:"+config.port:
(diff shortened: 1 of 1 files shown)


backend/cmd/headlamp.go
```diff
@@ -150,7 +150,7 @@ func getOidcCallbackURL(r *http.Request, config *HeadlampConfig) string {
 		switch {
 		case fwdProto != "":
 			urlScheme = fwdProto
-		case r.Host == "localhost:"+config.port:
+		case strings.HasPrefix(r.Host, "localhost:") || r.TLS == nil:
```
joaquimrocha, line 153: The commit title/body doesn't tell much about why this is needed. i.e. what was wrong, what does this fix?

joaquimrocha (changes requested):

backend/cmd/headlamp.go
ashu8912: Updated

joaquimrocha (approved):

PR #431, opened 2022-02-17
.github: Remove file rename from app-artifacts-mac action

joaquimrocha:
The newly used electron-build version no longer requires that rename
as it honors the passed architecture.


illume (approved):
👍 thanks

Issue #393, opened 2022-01-13
oidc: No permissions to list pods (AWS Cognito)


javlonsodikov:
## Description

oidc: No permissions to list pods.

## Impact

Cannot use headlamp after OIDC login

## Environment and steps to reproduce

1. **Set-up**: headlamp, configure oidc with AWS cognito  
2. **Task**: Login to the headlamp 
3. **Action(s)**: 
  a. open main page
  b. press "Sign In"
  c. type OIDC login password, login

4. **Error**: 
     page redirects to the main headlamp page but "No permissions to list pods." errors messages are shown. 
## Expected behavior

Expected to see cluster details 

## Additional information

What else should be configured?

joaquimrocha:
Hi @javlonsodikov , does Headlamp keep redirecting you to the login, or does it feel like logging in works but it's the listing that fails?
If that's the case, please double check you have the right role to list pods, as Headlamp will check those permissions in order to determine if any cluster resource can be accessed/updated/deleted/etc.

javlonsodikov:
I would say both: after logging in I see following: 
![image](https://user-images.githubusercontent.com/5047093/149385813-b3db3dc0-5896-4e6d-a338-5e8bc415eb16.png)

Whenever I click any other sections in the UI its redirects to login page

also in the chrome console I can see 
```
{kind: "Status", apiVersion: "v1", metadata: {}, status: "Failure",…}
apiVersion: "v1"
code: 403
details: {group: "metrics.k8s.io", kind: "nodes"}
kind: "Status"
message: "nodes.metrics.k8s.io is forbidden: User \"system:anonymous\" cannot list resource \"nodes\" in API group \"metrics.k8s.io\" at the cluster scope"
metadata: {}
reason: "Forbidden"
status: "Failure"
```

javlonsodikov:
here is the authorization Bearer 
```
{
  "sub": "fb6cf411-aff6-4630-964f-899a03a94b45",
  "cognito:groups": [
    "cluster-admin",
  ],
  "iss": "https://cognito-idp.us-east-1.amazonaws.com/<id>",
  "version": 2,
  "client_id": "<client-id>",
  "origin_jti": "<masted_jti>",
  "token_use": "access",
  "scope": "phone openid profile email",
  "auth_time": 1642158722,
  "exp": 1642162322,
  "iat": 1642158722,
  "jti": "<jti>",
  "username": "fb6cf411-aff6-4630-964f-899a03a94b45"
}
```

[...]

PR #437, opened 2022-02-28
backend: Fix oidc redirect url token type

ashu8912:
In an oidc auth flow the id token is used for authentication and that is
what should be send in the redirect URI of the oidc callback so that
id_token is used to make request to the kubernetes cluster.
Access token should not be used because they are part of oauth 2.0 flow
and are used for authorization purposes.

## Testing done
Tested with cognito and EKS setup 
The fix is contained in this image  ```docker.io/ashu8912/headlamp-test```

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -356,7 +356,7 @@ func StartHeadlampServer(config *HeadlampConfig) {
 				redirectURL = "/"
 			}
 
-			redirectURL += fmt.Sprintf("auth?cluster=%1s&token=%2s", state, oauth2Token.AccessToken)
+			redirectURL += fmt.Sprintf("auth?cluster=%1s&token=%2s", state, rawIDToken)
 			http.Redirect(w, r, redirectURL, http.StatusSeeOther)
 		} else {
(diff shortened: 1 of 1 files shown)


ashu8912:
fixes #393

illume:
Thanks @ashu8912 :) Indeed...

> An id token proves that a user has been authenticated. 
-- https://auth0.com/blog/id-token-access-token-what-is-the-difference/

Good!

The part I haven't checked is the /auth route. Where is that documented? I want to understand that and also check the token input documentation.


- Probably that code should be constructing URLs with https://pkg.go.dev/net/url#URL Using string interpolation can be problematic if not escaping input.
- How about using the name idToken for the id token?

illume (approved):
thanks 🎉

ashu8912:
> The part I haven't checked is the /auth route. Where is that documented? I want to understand that and also check the token input documentation.
auth route is a dummy callback route on frontend that gets the token appended onto its query param

joaquimrocha:
@ashu8912 has confirmed this also fixes OIDC issues when using KeyCloack. Merging.

PR #442, opened 2022-03-10
Add notification area

ashu8912:
Add a notification area for headlamp on the topbar and show the relevant events in the notification dialog


frontend/src/redux/reducers/config.tsx
```diff
@@ -18,6 +24,43 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
     case CONFIG_NEW:
       newState.clusters = { ...action.config.clusters };
       break;
+    case CLUSTER_SET_NOTIFICATIONS: {
+      const clusterName: string = action.clusterName;
+      const notifications = action.notifications;
+      // if there are no notifications set on cluster just set it to an empty array
+      if (newState.clusters && !newState.clusters[clusterName].notifications) {
```
joaquimrocha, line 31: Seems to me like the `newState.clusters[clusterName]` may not exist.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -15,6 +15,9 @@ export interface Cluster {
   name: string;
   useToken?: boolean;
   [propName: string]: any;
+  notifications?: KubeEvent[];
```
joaquimrocha, line 18: Notifications will be for more than KubeEvents. So they should support info related to a KubeEvent but not be limited to it.
i.e. it's better to implement a `Notification` type and have a function that converts a KubeEvent to it.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+function prepareNotificationsList(
+  notifications: KubeEvent[] | undefined,
+  apiError: ApiError | undefined,
+  clickEventHandler: (event?: any) => void
+) {
+  const classes = useStyles();
+  if (!notifications || notifications.length === 0) {
+    return (
+      <MenuItem onClick={clickEventHandler} className={classes.notificationItem}>
+        You don't have any notifications right now
```
joaquimrocha, line 50: Needs to be translatable.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: Notifications is also a global feature, meaning each notification **may** be related to a cluster, but we want these to be independent from clusters as we're showing them all together.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    //only take us to events page if it's not a error notification from fetching events
+    if (!apiError) {
+      history.replace('/');
+    }
+  }
+
+  return (
+    <>
+      <IconButton onClick={handleClick}>
+        {notifications && notificationsAlert && notifications.length !== 0 ? (
```
joaquimrocha, line 121: The badge should be set on the notifications based on whether we have "unread" notifications, and this should be done without having to check a redux status, in order for the functionality to be encapsulated and handled within the component.

joaquimrocha (changes requested):
I left a few comments.
We should only have one dispatch notification related function. The notification itself should have info about what needs to be displayed, severity of it, etc.

I have tried the patch and noticed I always get the "Not found" notification + not sure how/when the badge is set to the bell (always get it at the beginning).

image

frontend/src/redux/reducers/ui.tsx
```diff
@@ -188,6 +192,24 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
...
+    case UI_SET_NOTIFICATIONS: {
+      const notifications = action.notifications;
+      /* There are two ways user can send notifications either a complete set of array or a single notification
+         when handling the array notifications we want to only have unique set of notifications pushed into the UI notifications config
+      */
+      if (Array.isArray(notifications)) {
+        const uniqueNotifications = _.uniqBy(
+          [...notifications].concat(newFilters.notifications),
+          function (item) {
+            return item.id;
```
illume, line 204: You could remove the extra dependency and all the id stuff by using:
```ts
return `${item.date},${item.message}`
```

There's no need for a UUID, since no UUID properties are being used.

frontend/src/lib/notification.tsx
```diff
@@ -0,0 +1,34 @@
...
+      this.message = message;
+    }
+    if (date) {
+      this.date = date;
+    }
+    // generate the id based on the message and the date attached to a notification
+    this.id = getUuidByString(`${date},${message}`);
+  }
+
+  generateID() {
```
illume, line 31: This isn't used. Additionally, there's no need for it to be a UUID.

frontend/src/lib/notification.tsx
```diff
@@ -0,0 +1,34 @@
...
+      this.type = type;
+    }
+    if (message) {
+      this.message = message;
+    }
+    if (date) {
+      this.date = date;
+    }
+    // generate the id based on the message and the date attached to a notification
+    this.id = getUuidByString(`${date},${message}`);
```
illume, line 28: You could remove the extra dependency by using:
```ts
this.id = btoa(`${item.date},${item.message}`)
```

There's no need for a UUID, since no UUID properties are being used.

frontend/src/lib/themes.ts
```diff
@@ -130,6 +131,7 @@ const darkTheme = createTheme({
     resourceToolTip: {
       color: 'rgba(255, 255, 255, 0.87)',
     },
+    notificationBorderColor: 'rgba(255,255,255,0.12)',
```
illume, line 134: I wonder if you checked these for contrast/vision?
(It should hopefully print out in the browser console if there's an issue, but maybe it's useful to check with the browser tools.)

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        {notificationAlert ? (
+          <Badge variant="dot" color="error">
+            <Icon icon={bellIcon} />
+          </Badge>
+        ) : (
+          <Icon icon={bellIcon} />
+        )}
+      </IconButton>
+      <Menu
+        id="simple-menu"
```
illume, line 151: Is this id used for anything? I know this component should probably only be used once at the moment... (so it's not that important). But if this component is used multiple times this would be a problem repeating the same id.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    // push notification only if we have some cluster selected + there are events
+    const recentNotificationFromStorage = localStorage.getItem('recent_notification');
+    if (recentNotificationFromStorage) {
+      const parsedRecentNotification = JSON.parse(recentNotificationFromStorage);
+      if (parsedRecentNotification?.id !== notifications[0].id) {
+        setNotificationAlert(true);
+      }
+    }
+    // we store the recent event in storage which we will use to check if there were new notifications loaded from last time the user checked the notification tab
+    localStorage.setItem('recent_notification', JSON.stringify(notifications[0]));
```
illume, line 118: Are the notification ordering stable and consistent or eventually consistent? If it's eventually consistent(logs, queues, third party systems), then I'm not entirely sure this logic will work out.

frontend/src/lib/notification.tsx
ashu8912: TIL about btoa, Thanks I used UUID to uniquely identify each notification but btoa would do the same job here I will revert to using btoa

frontend/src/lib/k8s/cluster.ts
ashu8912: Fixed

frontend/src/components/common/Notifications/index.tsx
ashu8912: Fixed

frontend/src/lib/notification.tsx
ashu8912: This would be a helper within the class that will allow plugin developer to create ids on notification when they use this class

frontend/src/components/common/Notifications/index.tsx
ashu8912: When adding these notifications in redux store we always store them in order of the time they occured so the most recent ones stay on top of that list, so it eventually is consistent.But I am not sure if this is the problem you point to can you please elaborate more??

joaquimrocha:
I have converted to draft, pending the style changes (we should have mockups next Monday).

ashu8912:
@joaquimrocha @illume new improvements are added from the last time, 

1. events section now accepts a filter from queryParams
2. improvement in notification area design
3. persistent notifications even on browser refresh
4. handle seen/unseen of notifications

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+        className={classes.notificationItem}
+      >
+        <Grid container spacing={4}>
+          <Grid item direction="column" md={notification.seen ? 12 : 11}>
+            <Grid item md={12}>
+              <Typography style={{ fontWeight: notification.seen ? 'normal' : 'bold' }}>
+                {prepareNotificationLinkComponent(notification)}
+              </Typography>
+            </Grid>
+            <Grid item>{timeAgo(notification.date)} ago</Grid>
```
joaquimrocha, line 99: This is not translatable. I think `timeAgo` can already give the full sentence with "ago" in it, but you'd have to check if that's so.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const notifications = useTypedSelector(state => state.ui.notifications);
+  const dispatch = useDispatch();
+  const [events] = Event.useList();
+  const { t } = useTranslation();
+
+  useEffect(() => {
+    console.log('cluster path');
+    if (events && events.length !== 0) {
+      const importantEvents: Notification[] = events
+        .filter((event: KubeEvent) => event.type === 'Normal')
```
joaquimrocha, line 124: I think we had talked about showing only warnings by default, right?
Maybe we can ask some of the Kubernetes volks what would make sense to them.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    );
+  });
+}
+
+export default function Notifications() {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
+  const notifications = useTypedSelector(state => state.ui.notifications);
+  const dispatch = useDispatch();
+  const [events] = Event.useList();
```
joaquimrocha, line 117: I get the feeling we shouldn't mix the events here and just have this logic handle Notifications in a pure way. But I don't know where I'd put the events' logic right now... So maybe we can keep this as a first version.

frontend/src/components/common/Notifications/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+  }
+  const areThereUnseenNotifications =
+    notifications.filter(notification => notification.seen !== true).length > 0;
+
+  return (
+    <>
+      <IconButton onClick={handleClick}>
+        {areThereUnseenNotifications ? (
+          <Badge variant="dot" color="error">
+            <Tooltip title={`${t('frequent|You have unread notifications')}`}>
```
joaquimrocha, line 181: Nice tooltip!
Please use the notifications namespace.

plugins/headlamp-plugin/config/webpack.config.js
```diff
@@ -10,6 +10,7 @@ const externalModules = {
   notistack: 'pluginLib.Notistack',
   '@kinvolk/headlamp-plugin/lib/CommonComponents': 'pluginLib.CommonComponents',
   '@kinvolk/headlamp-plugin/lib': 'pluginLib',
+  notification: 'pluginLib.Notification',
```
joaquimrocha, line 13: This is not needed here. You already added the Notification to the pluginLib above.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -126,7 +130,11 @@ function EventsSection() {
               ]
             : []
         }
-        data={events}
+        data={
+          eventsFilter
+            ? events.filter((event: Event) => event.metadata.uid === eventsFilter)
+            : events
+        }
```
joaquimrocha, line 137: I'm not sure this is the best way as it will only show one event without chances to seeing the rest without going to the view without the filter. Could we maybe add the event ID as a special search field and then use that views search entry?

joaquimrocha (changes requested):

joaquimrocha:
I don't know why this comment hasn't shown up in the review, but besides what I wrote in the code review I also found the following issues:
1. The notifications text should not be shown behind the red dot.
image
2. There are too many notifications in my demo cluster, so the notifications' loading is very slow, we should consider limiting it even if this means losing a lot of notifications ATM.
3. Clicking the menu item should be the same as clicking the link, not making it read/unread. It's clicking the red dot that should make it red/unread, this is how another famous social network's notifications work. This also means that we should probably not have the link effect. When I mentioned we'd need links I was talking about when we introduce different actions for notifications (maybe I didn't notice this link behavior in Tejas' design, sorry).

frontend/src/lib/notification.tsx
```diff
@@ -0,0 +1,32 @@
+export enum NOTIFICATION_TYPE {
+  DIALOG = 'DIALOG',
+  SNACKBAR = 'SNACKBAR',
+  BOTH = 'BOTH',
```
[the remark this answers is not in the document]
ashu8912: `I guess that for users using plain Js, this is transpiled as an object and users would do NotitificationType.BOTH?`

yes

frontend/src/lib/notification.tsx
ashu8912: I was thinking of adding this earlier but i don't think there will ever be need from outside to change the ID of a notification

frontend/src/redux/reducers/ui.tsx
```diff
@@ -188,6 +193,40 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
...
+          }
+        );
+        newFilters.notifications = uniqueNotifications;
+      } else {
+        newFilters.notifications.push(notifications);
+      }
+      // This way we make sure notifications are always in a consistent order
+      newFilters.notifications.sort((n1: Notification, n2: Notification) => {
+        return new Date(n2.date).getTime() - new Date(n1.date).getTime();
```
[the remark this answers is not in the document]
ashu8912: n1.date and n2.date here are not objects they are either a date string or a timestamp(number)

frontend/src/redux/reducers/ui.tsx
```diff
@@ -114,6 +118,7 @@ export const INITIAL_STATE: UIState = {
   theme: {
     name: '',
   },
+  notifications: [],
```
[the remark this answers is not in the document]
ashu8912: A map actually doesn't serve any purpose here, we want a list of notifications and it's serializable as well.

frontend/src/redux/reducers/ui.tsx
```diff
@@ -188,6 +193,40 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
...
+      if (Array.isArray(notifications)) {
+        const uniqueNotifications = _.uniqBy(
+          [...notifications].concat(newFilters.notifications),
+          function (item) {
+            return item.id;
+          }
+        );
+        newFilters.notifications = uniqueNotifications;
+      } else {
+        newFilters.notifications.push(notifications);
```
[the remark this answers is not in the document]
ashu8912: notifications is a list I just need to push the new object into it, why would we copy it?? One thing i realized we would like to have the filter here as well which doesn't allow a notification that is already there to be pushed again

frontend/src/components/App/TopBar.stories.tsx
```diff
@@ -6,8 +6,11 @@ import { createStore } from 'redux';
 import { PureTopBar, PureTopBarProps } from './TopBar';
 
 // eslint-disable-next-line no-unused-vars
-const store = createStore((state = { config: {} }, action) => state, {
+const store = createStore((state = { config: {}, ui: { notifications: [] } }, action) => state, {
   config: {},
+  ui: {
+    notifications: [],
+  },
```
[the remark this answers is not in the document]
ashu8912: we are create a new store here which expects notifications to be passed in with default values

frontend/src/components/common/Notifications/index.tsx
ashu8912: oops pushed this by mistake while testing, warning events rarely happen so i changed it to normal for testing, will revert back to normal because we don't wan't to show every normal event out there because they are of no concern to the user

plugins/headlamp-plugin/config/webpack.config.js
ashu8912: I don't get this, we have other modules here which are also added to the pluginLib how does it work here??

frontend/src/redux/reducers/ui.tsx
```diff
@@ -220,6 +221,11 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
       newFilters.notifications.sort((n1: Notification, n2: Notification) => {
         return new Date(n2.date).getTime() - new Date(n1.date).getTime();
       });
+      // limit showing not more then max_notification_count notifications for now till we have a dedicated page for notifications
+      if (newFilters.notifications.length >= max_notification_count) {
+        // we remove the oldest available notifications when the notifications length is crossing max_notification_count threshold
+        newFilters.notifications.pop();
```
joaquimrocha, line 227: I understand this should work, but I am getting the UI to be unusable still with this change.
It can be that localStorage in my tests still had loads of notifications and that was getting loaded...
I noticed that limiting the length earlier does work for me. i.e. having the following above:
```
[...notifications].slice(0, max_notification_count).concat(newFilters.notifications),
          function (item) {
            return item.id;
          }
```

joaquimrocha (changes requested):
I still have issues related to the UI being slowed due to this change (see my comment in the limit).
Even with the limit on, I feel the UI is not as responsive, maybe this has to do with this setting the storage * 100. We should throttle this and/or maybe doing it in a saga since it's async.

Some last minute UX issues I noticed:
1. When clicking the read/unread, it is propagating the click, so the event is getting selected and shown (it shouldn't, since the user just wants to mark it as read).
2. When clicking the notification, the popup stays opened, so it's almost unnoticeable that we set up the event in the event's page. If we close the popup, this will be a bit better.
3. We should make the popup grow up to 75% of the page's height.
4. You can reuse some date component we have (I think it's in the conditions component) to display a calendar icon + tooltip for the full date, so we show the "3 minutes ago" label but the user has a way to see the full date.

frontend/src/redux/reducers/ui.tsx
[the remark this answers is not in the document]
joaquimrocha: I am saying you can do `new Date(n2.date) - new Date(n1.date)`.

joaquimrocha (commented):
I found a few more things.

frontend/src/redux/reducers/ui.tsx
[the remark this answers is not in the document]
ashu8912: new Date() returns number | string so that would require an extra step of handling a string returned Date

joaquimrocha:
@ashu8912 I think last time you showed me this PR working, I could still notice some slowness which you believed was just caused by your machine overload but I think it's just that we render too many items in the list. So we need to lazy load them. You have an example in https://v4.mui.com/components/lists/#virtualized-list and also we use something like this for the package selection dialog in Nebraska.

(Adding the comment just for better tracking the progress here)

joaquimrocha (commented):
Left a comment in the new commit.
BTW, I realized also that the Notifications component should probably be inside the `App` folder, rather than the `common` (this is exported to the plugins, and the notifications component should only be interacted with through a dispatch call AFAIU).

joaquimrocha (approved):
The changes and the UX LGTM.
I had to add two updates to the commits:
1. Prettified a change, maybe prettier is not running automatically as a hook in your env, @ashu8912 
2. Changed the events filter from `=== Normal` to `!== Normal` as I am sure the former was forgotten after testing.

It is now very fast and useful! Good job.

One thing I noticed is that it pulls events from all clusters! This is actually what we wanted in the long run, but I couldn't see where you changed it to behave as such... I wonder if it's an indication of a bug in the request layer. Can you confirm if this is intentional, @ashu8912 ?

frontend/src/lib/notification.tsx
illume: Seems this was removed. Resolving this conversation.

frontend/src/redux/reducers/ui.tsx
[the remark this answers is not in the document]
illume: Seems this is settled. The sort by date code seems to work. I've resolved this conversation.

frontend/src/components/common/Notifications/index.tsx
illume: This part isn't there anymore... resolving conversation.

frontend/src/lib/notification.tsx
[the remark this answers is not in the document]
illume: This was changed. Resolving conversation.

frontend/src/lib/notification.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+enum NotificationType {
+  DIALOG = 1 << 0,
+  SNACKBAR = 1 << 1,
+  ALL = DIALOG | SNACKBAR,
+}
+
+export class Notification {
+  static NotificationType = NotificationType;
+  type: NotificationType = NotificationType.DIALOG;
```
illume, line 12: Could this class be documented? It would be great if all interfaces and functions could be documented... but these in particular are exposed to the public.

frontend/src/redux/reducers/ui.tsx
```diff
@@ -206,6 +211,47 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
...
+        );
+        newFilters.notifications = uniqueNotifications;
+      } else {
+        // check if this notification is already present if not add it
+        if (
+          newFilters.notifications.findIndex(
+            (notification: Notification) => notification.id === notifications.id
+          ) === -1
+        ) {
+          newFilters.notifications.push(notifications);
```
illume, line 234: (These was a previous comment about this...)

The notifications needs to be copied here. Because the newFilters returned from the reducer needs to be copied. This is passing a reference to the data the action passes in.

frontend/src/redux/reducers/ui.tsx
```diff
@@ -206,6 +211,47 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
       newFilters.branding.logo = component;
       break;
     }
+    case UI_SET_NOTIFICATIONS: {
+      const notifications = action.notifications;
+      /* There are two ways user can send notifications either a complete set of array or a single notification
+         when handling the array notifications we want to only have unique set of notifications pushed into the UI notifications config
+      */
+      if (Array.isArray(notifications)) {
```
illume, line 219: Should this support the following?
- single notifications adding to the existing notification
- multiple notifications overwriting the existing ones

I see that Notifications/index@Notifications Event.useList always overwrites them, so the single adding notification function will not work. 

So, something needs to be fixed here.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+        .filter((event: KubeEvent) => event.type !== 'Normal')
+        .map((event: KubeEvent) => {
+          const notificationIndexFromStore = notifications.findIndex(
+            notification => notification.id === event.metadata.uid
+          );
+          if (notificationIndexFromStore !== -1) {
+            return notifications[notificationIndexFromStore];
+          }
+          const notification = new Notification();
+          notification.date = new Date(event.metadata.creationTimestamp).getTime();
```
illume, line 146: Should the Notification constructor be used here, and the constructor should be fixed to require the arguments that are required?

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+          notification.id = event.metadata.uid;
+          notification.message = event.message;
+          notification.url = createRouteURL('cluster') + `?eventsFilter=${notification.id}`;
```
illume, line 149: This could be: `notification.url = createRouteURL('cluster',  {eventsFilter: notification.id});`

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+    clickEventHandler(notification);
+  }
+
+  function notificationItemClickHandler(notification: Notification) {
+    notification.url && history.push(notification.url);
+    notification.seen = true;
+    clickEventHandler(notification, true);
+  }
+
+  function Row(props: { index: number; style: object }) {
```
illume, line 74: nit: I saw here the type for props: https://mui.com/material-ui/react-list/#virtualized-list

```ts
import { FixedSizeList, ListChildComponentProps } from 'react-window';
function renderRow(props: ListChildComponentProps) {
```

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+    },
+  },
+  errorItem: {
+    '&': {
+      color: theme.palette.error,
+    },
+  },
+}));
+
+function prepareNotificationsList(
```
illume, line 44: Just out of curiosity, why did you choose to make this a function and not a component?

frontend/src/redux/reducers/ui.tsx
illume: Ok, I got confused... for the events adding it looks at the existing event in the store are also left there. (I wonder if this notification logic should be in one place?)

Note: There's a updateUINotification which takes a single Notification for updating.

plugins/headlamp-plugin/config/webpack.config.js
```diff
@@ -14,6 +14,7 @@ const externalModules = {
   // how it's used in Headlamp's source code.
   '@kinvolk/headlamp-plugin/lib/k8s': 'pluginLib.K8s',
   '@kinvolk/headlamp-plugin/lib': 'pluginLib',
+  '@kinvolk/headlamp-plugin/types/lib/notification': 'pluginLib.Notification',
```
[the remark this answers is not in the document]
illume: Looks like this was updated. Resolving conversation.

frontend/package.json
```diff
@@ -158,6 +159,7 @@
     "@storybook/node-logger": "^6.3.10",
     "@storybook/preset-create-react-app": "^3.2.0",
     "@storybook/react": "^6.3.10",
+    "@types/react-window": "^1.8.5",
```
illume, line 162: Yeah, this should go to dependencies. Anything required for `npm run build` is a dependency not a dev dependency.

frontend/src/redux/reducers/ui.tsx
ashu8912: fixed

frontend/src/components/App/Notifications/index.tsx
ashu8912: Adding it as a second argument will not work for our case because we want to pass the filter as queryParam but the second argument of ```createRouteURL``` processes urlParams

docs/development/plugins/functionality.md
```diff
@@ -17,7 +17,7 @@ The main ones are:
 * Registry: The plugin registry, used to register functionality
 * Headlamp: To register plugins
 * CommonComponents: React components commonly used in the Headlamp UI
-
+* Notification: This module contains two exported members one is Notification a class which can be used to prepare notifications that are accepted by headlamp and the other one is setNotificationsInStore use this to fill the headlamp redux store with the desired notification.
```
joaquimrocha, line 20: Do not remove the newline. Add this before (and also needs a grammar check).
```suggestion
* Notification: This module contains two exported members. One is `Notification`, a class which can be used to prepare notifications that are accepted by Headlamp, and the other is `setNotificationsInStore` which is used to fill the Headlamp redux store with the desired notification.
```

It's unclear to me, reading this, what exactly setNotificationsInStore does or why would we need to call it.

frontend/src/components/common/Notifications/index.tsx
joaquimrocha: It's still showing the "Normal" events.

joaquimrocha (changes requested):
There are still some unanswered or unaddressed comments.
Also, I have just noticed that we are not doing anything with the NotificationType, so maybe this is something that only makes sense to expose later, rather than having a spec that's not implemented but is exported.

frontend/package.json
ashu8912: fixed

frontend/src/redux/reducers/ui.tsx
ashu8912: fixed

frontend/src/components/App/Notifications/index.tsx
ashu8912: thank you, fixed

frontend/src/components/App/Notifications/index.tsx
ashu8912: updated this, I actually at first made it a pure function but i think it is not a pure function because it has hook calls inside it.

frontend/src/lib/notification.tsx
ashu8912: I have created a mention of the class in the plugin doc, is that what you meant @illume ??

frontend/src/components/App/Notifications/index.tsx
ashu8912: fixed with the necessary args

frontend/src/lib/notification.tsx
illume: I'm meant documented with tsdoc strings.

docs/development/plugins/functionality.md
illume: Probably an example plugin will help explaining what it does.

PR #443, opened 2022-03-14
plugins/headlamp-plugin: Do not ignore types when packing

joaquimrocha:
We want to ignore the presence of generated types (types folder) in
git, but we want them to be shipped in the module. However, npm does
check .gitignore if a .npmignore file is missing, which meant that no
types were being packed/shipped.

This commit adds a .npmignore file that doesn't ignore the types,
effectively making npm pack the mentioned types.

### DIFF
--- a/plugins/headlamp-plugin/package-lock.json
+++ b/plugins/headlamp-plugin/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@kinvolk/headlamp-plugin",
-  "version": "0.4.2",
+  "version": "0.4.3",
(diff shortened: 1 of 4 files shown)


illume:
Also plugins/examples/pod-counter/ will need a dependency version bump to 0.4.3

illume (approved):
Thanks! 🎉

You can see the files included with: ```npm publish --dry-run```, and with this patch the files are in there. (I always assumed that `npm pack` was equivalent... but apparently it's not.)

The pod-counter example version needs a bump, but otherwise looks good.

PR #446, opened 2022-03-16
plugins/headlamp-plugins: Use internal tsconfig for plugins

### LINKED ISSUE #433: headlamp-plugin should use internal tsconfig.json
## Description

Currently headlamp-plugin create adds a tsconfig.json file through a template. However this leaves around an extra file in each plugin and makes things harder to maintain. Especially if things need to change in the file.

Instead headlamp-plugin should use the typescript config internally.

## Impact

Easier to keep plugins maintained.

joaquimrocha:
headlamp-plugin was adding a tsconfig.json file through a template.
However this left around a file in each plugin which could not be
easily updated.

This files prevents that by shipping a plugins-tsconfig.json in the
headlamp-plugin module, and including it from the templated tsconfig.

fixes #433

### DIFF
--- a/plugins/headlamp-plugin/config/plugins-tsconfig.json
+++ b/plugins/headlamp-plugin/config/plugins-tsconfig.json
@@ -1,0 +1,30 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "allowSyntheticDefaultImports": true,
+    "jsx": "react-jsx",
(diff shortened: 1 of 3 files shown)


joaquimrocha:
I have tested it quickly here, but would appreciate anyone confirming this works fine and is the best way to tackle #433.

Issue #440, opened 2022-03-03
OIDC: state is too short minimum 8 characters expected


javlonsodikov:
## Description

Internal OIDC provider complains that `state` is too short.  

## Impact

OIDC not letting to authorize 

## Environment and steps to reproduce
usual setup

## Expected behavior

OIDC provider should allow to login 

## Additional information
in headlamp.go file I added UUID after state as a temporary fix. 

```		
             oauthRequestMap[cluster] = &OauthConfig{Config: oauthConfig, Verifier: verifier, Ctx: ctx}
		state := uuid.New().String()
		cookie := http.Cookie{
			Name:    "state",
			Value:   state,
			Expires: time.Now().Add(time.Minute),
		}
		http.SetCookie(w, &cookie)
		uuidWithoutHyphens := strings.Replace(state, "-", "", -1)
		http.Redirect(w, r, oauthConfig.AuthCodeURL(cluster+":"+uuidWithoutHyphens), http.StatusFound)
	}).Queries("cluster", "{cluster}")

	r.HandleFunc("/oidc-callback", func(w http.ResponseWriter, r *http.Request) {
		mstate := r.URL.Query().Get("state")
		if mstate == "" {
			http.Error(w, "invalid state request ", http.StatusBadRequest)
			return
		}
		state := strings.Split(mstate, ":")[0]
		// nolint: nestif
```

illume:
The PR #447 was merged, so this should be fixed and in main and in the next Headlamp release.

PR #447, opened 2022-03-18
backend: Encode state in base64, use this rather than plain cluster name

ashu8912:
## Testing done
Testing not done yet, this would be an opportunity to test with the recent oidc fixes on keycloak and dex as well

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


ashu8912:
fixes #440

joaquimrocha (commented):
> State in oidc provides us security

I find this comment a bit confusing since we're talking about encoding/decoding and security here and not exposing the state, which sounds like encrypting it but it's of course not what this is about. So I'd change the message and say something like: "The OIDC state is used for verifying the incoming OIDC callback requests and we should make sure it's transmitted without issues, regardless of its value, so this patch starts using that state in encoded in base64. This also fixes an issues related to the state's length, as it needs to be longer than 8 chars, and this new base64 encoding grants that."

backend/cmd/headlamp.go
```diff
@@ -307,14 +308,20 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		*/
+		state := base64.StdEncoding.EncodeToString([]byte(cluster))
+		oauthRequestMap[state] = &OauthConfig{Config: oauthConfig, Verifier: verifier, Ctx: ctx}
+		http.Redirect(w, r, oauthConfig.AuthCodeURL(state), http.StatusFound)
 	}).Queries("cluster", "{cluster}")
 
 	r.HandleFunc("/oidc-callback", func(w http.ResponseWriter, r *http.Request) {
 		state := r.URL.Query().Get("state")
+		decodedState, err := base64.StdEncoding.DecodeString(state)
+		if err != nil {
```
joaquimrocha, line 322: We should probably refer this error in the HTTP error and/or log it, so eventual bug reports are more complete.

joaquimrocha (changes requested):
Left a comment about logging/returning an error relating to decoding.
After that we can merge this.

backend/cmd/headlamp.go
ashu8912: fixed

joaquimrocha (approved):

PR #448, opened 2022-03-21
frontend: Do not use objects for storing plugins functions

joaquimrocha:
Throughout the Redux code, we were using objects to store a map of
name -> something, but keys' order is not kept in objects, so we could
not predict which actions were being called after others.

This patch replaces that use with a Map subclass, so we can still use
it very similarly to what we used to, but now we guarantee that the
values can be used in insertion order.

As a side-effect, the serialization of a Map is more predictable than
that of an object (the Map is always {} by default), so we have a more
predictable use of the data within Redux (this comment has to do with
the fact that we shouldn't use unserializable data in Redux, but that
should be for cases where one's thinking of storing/persisting Redux's
values, which is not the case for this particular UI data).

@ashu8912 , after we discussed today that the unserializable data in Redux was only an issue if we had to serialize it (for possibly persist the data), I almost wrote to you to mention we shouldn't worry about changing our Redux use ATM and just focus on other things. However, I did notice that the use of objects may be an issue in Redux just because we lose any insertion order, and that order may be useful to keep, even if nowadays we don't do anything particular with it.
cc/ @illume

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -14,35 +14,33 @@ export default function RouteSwitcher() {
 
   return (
     <Switch>
-      {Object.values(routes)
-        .concat(defaultRoutes)
-        .map((route, index) =>
-          route.name === 'OidcAuth' ? (
-            <Route
-              path={route.path}
-              component={() => (
-                <PageTitle title={route.name ? route.name : route.sidebar}>
-                  <route.component />
-                </PageTitle>
-              )}
-              key={index}
-            />
-          ) : (
-            <AuthRoute
-              key={index}
-              path={getRoutePath(route)}
-              sidebar={route.sidebar}
-              requiresAuth={!route.noAuthRequired}
-              requiresCluster={!route.noCluster}
-              exact={!!route.exact}
-              children={
-                <PageTitle title={route.name ? route.name : route.sidebar}>
-                  <route.component />
-                </PageTitle>
-              }
-            />
-          )
-        )}
+      {[...routes.values()].concat(defaultRoutes).map((route, index) =>
+        route.name === 'OidcAuth' ? (
+          <Route
+            path={route.path}
+            component={() => (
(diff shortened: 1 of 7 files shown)


joaquimrocha:
BTW, doing this, I noticed we don't have testing for (at least all) the plugins' functionality. Is this correct @illume ? Do you think should add it as part of the Storybook somehow, or have full test plugins and integrate them in the testing somehow?

ashu8912 (approved):
Tested, LGTM!!, we can merge it once the conflict is resolved

illume (approved):
Not against merging this... but I'm not sure this is an issue. I didn't test this, because I don't see the how to test part of the PR description. For many years Object .values/.keys has a stable ordering: https://stackoverflow.com/questions/41761732/do-object-keys-and-object-values-methods-return-arrays-that-preserve-the-sam

joaquimrocha:
> Not against merging this... but I'm not sure this is an issue. I didn't test this, because I don't see the how to test part of the PR description. For many years Object .values/.keys has a stable ordering: https://stackoverflow.com/questions/41761732/do-object-keys-and-object-values-methods-return-arrays-that-preserve-the-sam

The stability of Object.values/.keys against an object is not the issue here. The issue is that we should preserve insertion (into the object) order.
I was double checking again, because I was pretty sure javascript objects are not ordered, but I have made a mistake indeed. 

Turns out that objects do not keep the keys' order if they're provided as numbers, even though their converted to string:
```javascript
> let c = {3:'First', 1:'Second'}
undefined
> Object.values(c)
[ 'Second', 'First' ]
> Object.keys(c)
[ '1', '3' ]
```
(Order was changed)

```javascript
> let o = {
... 'zkey': 'First inserted',
... 'akey': 'Last inserted',
... }
> Object.values(o)
[ 'First inserted', 'Last inserted' ]
> Object.keys(o)
[ 'zkey', 'akey' ]
```

(Order was kept)

So you're right @illume that we shouldn't need it for the order, if we guarantee that we're passing strings as keys.
The unstable way we can serialize an object is not an enough reason for keeping this change. I can close this one.

PR #453, opened 2022-03-28
Allow listing resources per namespace

### LINKED ISSUE #450: [RFE] Extend the KubeObject hooks to support listing objects in a single namespace
## Current situation

We can list objects easily using the `MyKubeObject.useList` hooks, but this works only when the user wants to list all objects from all namespaces.
It may be that users can only or want only to list objects in a certain namespace, so we should have that as an option.

## Ideal future situation

We can pass the namespace on which to list the objects.

joaquimrocha:
- frontend: Allow using a namespace in KubeObject's useList and useApiList
- frontend: Support multiple namespaces in KubeObject's use(Api)List

fixes #450

### DIFF
--- a/frontend/src/lib/k8s/cluster.ts
+++ b/frontend/src/lib/k8s/cluster.ts
@@ -49,13 +49,18 @@ export interface KubeOwnerReference {
   uid: string;
 }
 
+export interface ApiListOptions {
+  namespace?: string;
+}
+
 // We have to define a KubeObject implementation here because the KubeObject
 // class is defined within the function and therefore not inferable.
 export interface KubeObjectIface<T extends KubeObjectInterface | KubeEvent> {
   apiList: (onList: (arg: InstanceType<KubeObjectIface<T>>[]) => void) => any;
   useApiList: (
     onList: (arg: InstanceType<KubeObjectIface<T>>[]) => void,
-    onError?: (err: ApiError) => void
+    onError?: (err: ApiError) => void,
+    opts?: ApiListOptions
   ) => any;
   useApiGet: (
     onGet: (...args: any) => void,
@@ -64,7 +69,7 @@ export interface KubeObjectIface<T extends KubeObjectInterface | KubeEvent> {
     onError?: (err: ApiError) => void
   ) => void;
   useList: (
-    onList?: (...arg: any[]) => any
+    opts?: ApiListOptions
   ) => [any[], ApiError | null, (items: any[]) => void, (err: ApiError | null) => void];
   getErrorMessage: (err?: ApiError | null) => string | null;
   new (json: T): any;
@@ -138,14 +143,15 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
 
     static apiList<U extends KubeObject>(
       onList: (arg: U[]) => void,
-      onError?: (err: ApiError) => void
+      onError?: (err: ApiError) => void,
+      opts?: ApiListOptions
     ) {
       const createInstance = (item: T) => this.create(item) as U;
 
       const args: any[] = [(list: T[]) => onList(list.map((item: T) => createInstance(item) as U))];
 
       if (this.apiEndpoint.isNamespaced) {
-        args.unshift(null);
+        args.unshift(opts?.namespace || null);
       }
 
       if (onError) {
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Hello,

Would it be ok to add one or more of? 
- some documentation for what the feature does (perhaps just on the interface field)
- a test
- using it somewhere in the code base (perhaps in a plugin example)
- the how to test part of the PR description template being filled out

joaquimrocha:
> 

Sure. Thanks.

illume:
The examples given `useList('default') and useList(['default', 'kube-system'])` don't seem to match the argument signature. I think they should be: `useList({namespace: 'default'})`.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -157,18 +165,56 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+          listCallback(allObjs);
+        }
+      }
+
+      const listCalls = [];
+      if (!!opts?.namespace) {
+        let namespaces: string[] = [];
+        if (typeof opts.namespace === 'string') {
+          namespaces = [opts.namespace];
+        } else {
```
illume, line 197: This could help a user of this API know there is an issue with their input...
```ts
    } else if (Array.isArray(opts.namespace)) {
        namespaces = opts.namespace as string[];
    } else {
        throw Error("namespace should be a string or array of strings")
    }
```

frontend/src/lib/k8s/cluster.ts
```diff
@@ -138,14 +143,17 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
 
     static apiList<U extends KubeObject>(
       onList: (arg: U[]) => void,
-      onError?: (err: ApiError) => void
+      onError?: (err: ApiError) => void,
+      opts?: {
+        namespace?: string;
```
illume, line 148: Is this type correct? I'd think it should be a ApiListOptions, with `namespace?: string | string[];`

frontend/src/lib/k8s/cluster.ts
```diff
@@ -157,18 +165,56 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+        let namespaces: string[] = [];
+        if (typeof opts.namespace === 'string') {
+          namespaces = [opts.namespace];
+        } else {
+          namespaces = opts.namespace as string[];
+        }
+
+        for (const namespace of namespaces) {
+          listCalls.push(
+            this.apiList(objList => onObjs(namespace, objList as U[]), onError, { namespace })
```
illume, line 203: I wonder what happens with multiple errors? Before onError could be called once, now it can be called multiple times. Not sure it matters.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -157,18 +165,56 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
     ) {
+      const [objs, setObjs] = React.useState<{ [key: string]: U[] }>({});
       const listCallback = onList as (arg: U[]) => void;
-      useConnectApi(this.apiList(listCallback, onError));
+
+      function onObjs(namespace: string, objList: U[]) {
+        if (objList.length > 0) {
+          let newObjs: typeof objs = {};
+          // Set the objects so we have them for the next API response...
+          setObjs(objs => {
```
illume, line 178: Here objs would be more clear if it was called prevObjs.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -157,18 +165,56 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+            newObjs = { ...objs, [namespace || '']: objList };
+            return newObjs;
+          });
+
+          let allObjs: U[] = [];
+          Object.values(newObjs).map(currentObjs => {
+            allObjs = allObjs.concat(currentObjs);
+          });
+
+          listCallback(allObjs);
```
illume, line 188: This would now be called once for each namespace, each time with more.

illume:
Looking through it, I see it makes multiple API calls and gathers the results. So now it will call the callback multiple times. I think this would work ok most of the time, and only be a problem if there is a slow or errored response(s) for some reason - in this case it will probably display incorrectly without the user being aware. Could be fixed in another iteration I guess... up to you. Or maybe it's more desirable to show the results as they come in as is.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: I will add this.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: Yeah. This needs some more thought but I think we can proceed with what we have and look into this later. Thanks.

joaquimrocha:
> The examples given `useList('default') and useList(['default', 'kube-system'])` don't seem to match the argument signature. I think they should be: `useList({namespace: 'default'})`.

Sorry, I have just updated it.

illume (approved):
👍

PR #460, opened 2022-03-29
Export all k8s modules

### LINKED ISSUE #449: [RFE] Export makeKubeObject from Cluster class
## Current situation

When trying to use `makeKubeObject` function to abstract the k8s resource got the `'Property 'makeKubeObject' does not exist on type 'typeof import("/node_modules/@kinvolk/headlamp-plugin/types/lib/k8s/index")`.

## Impact

Can't abstract custom Kubernetes resources in a proper way. Instead, I am extending som resource class for overriding the `apiEndpoint` with custom `apiFactoryWithNamespace(..)`;

## Ideal future situation

All methods needed for creating custom k8s resource classes are exported and typed.

## Additional information

Please consider other class members potentially needed for creating custom k8s resource classes.

joaquimrocha:
I wonder if we should export `customresource.ts`. I think it's mixing static with instance members in it, and if we have `makeCustomResourceClass` class maybe it's no longer needed (as `patch` should work there)? Any thoughts @ashu8912 ?

fixes #449


joaquimrocha:
@ashu8912 , I have removed the customresource module due to the concerns I express in this issue's description.
Please take a look. It'd be nice if you could verify that you can create and use CRs with the makeCustomResourceClass.

illume (changes requested):
`make docs` needs to be run to generate the API markdown.

(Would be better if we can have some sort of example/doc/how to test for this.)

frontend/src/lib/k8s/cluster.ts
```diff
@@ -243,6 +244,18 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
       return this.apiEndpoint.put(data);
     }
 
+    patch(body: OpPatch[]) {
```
illume, line 247: I don't understand what this code does, or why it is different to delete and put with regards to args handling?

plugins/headlamp-plugin/lib/index.ts
```diff
@@ -5,4 +5,5 @@ import { Headlamp, Plugin } from '../types/plugin/lib';
 import Registry from '../types/plugin/registry';
 import * as CommonComponents from './CommonComponents';
 
-export { ApiProxy, K8s, CommonComponents, Utils, Plugin, Registry, Headlamp };
+// We export k8s (lowercase) since someone may use it as we do in the Headlamp source code.
```
illume, line 8: Nice change.

Probably we should (in a separate PR) change this in the examples, and deprecate uppercase K8s?

frontend/src/lib/k8s/cluster.ts
illume: Started some docs for patch()...

There's a few things that could be improved in my docs:
- I had trouble finding what this method returns because it's being typed as any.
- Added a link to OpPatch docs I found, because the docs for json-patch aren't there and the typing means you need to read the node-modules source code.
- wasn't sure of a realistic example.

```ts
    /**
     * Patches the cluster with a set of operations.
     * 
     * @param body an array of patch operations
     * @returns ? TODO: type is being resolved as any. The apiFactory.patch needs typing.
     * 
     * @see {@link https://www.npmjs.com/package/json-patch#patch-operations} OpPatch docs
     * @example
     * patch([{op: 'add', path: '/foo', value: 'bar'}]) // what's a more realistic example?
     */
    patch(...)
```

illume: I think I'd like to continue documenting and such in this PR: https://github.com/kinvolk/headlamp/issues/465

illume (approved):
Now I think generating docs and writing them separately is a better idea.

PR #461, opened 2022-03-30
backend: Fix plugins path on windows

### LINKED ISSUE #458: headlamp-plugin is not copying plugins to the correct location in Windows
## Description

A user has reported that headlamp-plugin is copying the plugins into `C:\Users\MYUSER\AppData\Roaming\Headlamp\Config\plugins` (which has an extra `Config` folder than the path that Headlamp reads plugins from).

## Impact

Headlamp cannot load plugins on Windows.

## Expected behavior

We need to make sure that the functions used to get the config folder are consistent between the backend and the `headlamp-plugin` tool.

illume:
## How to use

build headlamp on windows, then use a headlamp plugin

fixes #458

### DIFF
--- a/backend/cmd/plugins.go
+++ b/backend/cmd/plugins.go
@@ -5,8 +5,8 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
-	"path"
 	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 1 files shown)


backend/cmd/plugins.go
```diff
@@ -29,20 +29,29 @@ func defaultPluginDir() string {
...
+	// https://pkg.go.dev/os#UserConfigDir
 	userConfigDir, err := os.UserConfigDir()
 	if err != nil {
 		log.Printf("error getting user config dir: %s\n", err)
 
 		return pluginDirDefault
 	}
 
-	pluginsConfigDir := path.Join(userConfigDir, "Headlamp", "plugins")
+	pluginsConfigDir := filepath.Join(userConfigDir, "Headlamp", "plugins")
```
joaquimrocha, line 49: Nit: I usually try to avoid extra memory usage (unless golang is smarter than I think).
So I'd not join the path here until I know we're not on Windows.

joaquimrocha (approved):
Approved with a nit. I understand this is not HPC code, so feel free to ignore if you prefer.

PR #470, opened 2022-04-01
app: Allow to open about:blank URLs

### LINKED ISSUE #455: URLs not being opened from the desktop app
## Description

A user has reported to me that they cannot open external links from the UI in the desktop app (Windows version in this case).
We need to investigate if this is happening and fix it if it is.

## Impact

Users may not be able to open links to external resources (e.g. Kubernetes documentation).

joaquimrocha:
Some plugins may want to show popups, e.g. for authentication, so we
should allow such popups.

### How to test:
In a plugin's component that gets loaded, add something like:
```javascript
React.useEffect(() => {
    window.open('about:blank')
  },
  [])
```
Launch headlamp desktop.

You should see an empty popup window from the app, not the external browser.

fixes #455

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -358,8 +358,9 @@ function startElecron() {
     mainWindow.loadURL(startUrl);
 
     mainWindow.webContents.setWindowOpenHandler(({ url }) => {
-      // allow all urls starting with app startUrl to open in electron
-      if (url.startsWith(startUrl)) {
+      // Allow all app's domain URLs and also ones that are "about:blank"
+      // (so plugins can open e.g. popups for authentication).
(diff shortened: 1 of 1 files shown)


joaquimrocha:
@lijianzhi01 Let me know if this works well for you.

lijianzhi01:
Popup did work, but cross origin problem shows up. Desktop app load landing page through file protocol. But popup page is opened from landing page with http protocol. MSAL in landing page will manipulate popup window. If parent and child page use different protocol, location, and port, it would be failed to authenticate.

joaquimrocha:
Making it a draft now since we'll soon now whether we need it.

PR #474, opened 2022-04-04
frontend: Add router to pluginLib

ashu8912:
We want all the route related methods to be used by plugin developers
thus exporting this in the pluginLib makes sense.

### DIFF
--- a/docs/development/api/classes/lib_k8s_clusterRole.ClusterRole.md
+++ b/docs/development/api/classes/lib_k8s_clusterRole.ClusterRole.md
@@ -30,7 +30,7 @@ slug: "lib_k8s_clusterRole.ClusterRole"
 
 #### Defined in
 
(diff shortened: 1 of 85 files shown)


joaquimrocha:
@ashu8912, how about this fixup patch?

@illume , I was having doubts whether to expose the `ROUTES` object directly as this commit makes it or introduce a new `getDefaultRoutes()` exported function instead. Because maybe for developers, doing `Routes.getDefaultRoutes()` is better than `Routes.ROUTES`. WDYT?

illume:
Good idea. `Routes.getDefaultRoutes()` sounds better.

joaquimrocha:
Updated it. Not sure why Ashu's avatar is not showing on the first commit, since I kept the authorship...

illume (approved):
👍 🎈

PR #478, opened 2022-04-05
Bundled app plugins

illume:
Bundle plugins with the app version of Headlamp.  So different builds can be made that include plugins.

For https://github.com/kinvolk/headlamp/issues/417

## How to use

To build a Headlamp app with a set of plugins, first extract some plugins into the .plugins folder in the root of the headlamp repo.

```bash
npm install
npm run build
cd ../..

npx @kinvolk/headlamp-plugin extract ./plugins/examples/ ./.plugins
ls -la .plugins
make app-linux
```

## Testing done

See above.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -64,6 +64,11 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
     serverArgs = serverArgs.concat(['--kubeconfig', fullPath]);
   }
 
+  const secondPlugins = path.join(process.resourcesPath, '.plugins');
+  if (fs.existsSync(secondPlugins)) {
+    serverArgs.concat(['-plugins-dir2', secondPlugins]);
+  }
+
   serverArgs.concat(flags);
 
   return spawn(serverFilePath, serverArgs, options);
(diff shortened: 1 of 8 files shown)


joaquimrocha (changes requested):
@illume , I think the concept of having a plugins-dir and a plugins-dir2 is a bit odd and can be confusing for users. If this was an internal concept for building, it'd be fine, but it's also a flag that's exposed in the `server`, so we'll need to keep supporting it.

Seeing the changes I immediately thought "why not supporting loading multiple plugins dir in the same options"? Maybe you considered these two issues with this approach (?):
1. We'd need to supply the current default dir if we didn't want to support both whatever custom plugins dir is supplied + the default one.
2. We'd be watching all plugin dirs, and maybe you want the shipped ones not to be watched?

I think that for this initial step on build able to ship plugins in a custom app build, we can use the current `plugins-dir` option and just make sure we can build an app by overriding it if a non-empty `.plugins` dir exists. You have half of this already, so I'd prefer to not make a compromise on a new flag and just allow the app build to use the existing one.

If that's limiting for desktop app users (maybe they want to use the same app to load plugins from different locations), we can add a different option later (or they can use vanilla Headlamp). But that's something we can sort in a later iteration.

Let me know if this makes sense.

illume:
@joaquimrocha can you please expand on your suggestion?

PR #479, opened 2022-04-05
Build app with plugins

joaquimrocha:
- app: Use bundled plugins from .plugins folder
- FIXUP: Use plugins-dir

@illume , here's a quick cherry-pick of your app/ commit + a change to use the default plugins flag.

All I'd add that's missing here is maybe checking if the plugin is empty in order to not override the flag in such case.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -64,6 +64,11 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
     serverArgs = serverArgs.concat(['--kubeconfig', fullPath]);
   }
 
+  const secondPlugins = path.join(process.resourcesPath, '.plugins');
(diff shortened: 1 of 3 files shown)


app/electron/main.ts
```diff
@@ -64,6 +64,11 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
     serverArgs = serverArgs.concat(['--kubeconfig', fullPath]);
   }
 
+  const secondPlugins = path.join(process.resourcesPath, '.plugins');
+  if (fs.existsSync(secondPlugins)) {
```
illume, line 68: @joaquimrocha it seems to check if .plugins is there already.

illume:
I continued this here: https://github.com/kinvolk/headlamp/pull/480

PR #485, opened 2022-04-08
Fix plugin imports again

joaquimrocha:
This is another attempt at making the plugins' imports easy.

#### Test

1. Do the following in VSCode (and use the values/types).
```typescript
import { KubeObjectInterface } from '@kinvolk/headlamp-plugin/lib/K8s/cluster';
import { makeCustomResourceClass } from '@kinvolk/headlamp-plugin/lib/K8s/crd';
```
2. Set your plugin's tsconfig as the following:
```json
{
  "extends": "./node_modules/@kinvolk/headlamp-plugin/config/plugins-tsconfig.json"
}
```
3. Click + hover the module part of the import to see that it's resolved. (And also, no errors should be reported/underlined by VSCode)
4. Change an import to a non-existing module to check if there's an error reported by VSCode.
5. Run `npm run tsc` in the plugin to see that there are no complaints about the imports.

If there are VSCode complaints, in a .ts file, press Ctrl+Shift+p and choose `Typescript: Restart TS Server`
(I noticed that VSCode sometimes is picking rules that are no longer accurate, so the TS restart should fix that)

### DIFF
--- a/plugins/headlamp-plugin/config/plugins-tsconfig.json
+++ b/plugins/headlamp-plugin/config/plugins-tsconfig.json
@@ -27,5 +27,5 @@
     "rootDirs": ["node_modules/@kinvolk/headlamp-plugin/"],
     "skipLibCheck": true
   },
-  "include": ["./src/**/*"]
+  "include": ["./../../../../src/**/*"]
 }
--- a/plugins/headlamp-plugin/config/plugins-tsconfig.json
+++ b/plugins/headlamp-plugin/config/plugins-tsconfig.json
@@ -4,7 +4,7 @@
     "allowSyntheticDefaultImports": true,
     "jsx": "react-jsx",
     "moduleResolution": "node",
-    "baseUrl": ".",
+    "baseUrl": "../../../../",
     "paths": {
       "*": [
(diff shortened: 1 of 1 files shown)


joaquimrocha:
cc/ @andrii-batiuk , check this out as I think it may fix things for you. Notice that we still haven't published a new headlamp-plugin module with these changes, so either copy the PR's files into your node_modules/@kinvolk/headlamp-plugin/... or maybe linking will work.

joaquimrocha:
> @joaquimrocha Yeah, it fixed the issue. I've just edited my `tsconfig` in accordance with `plugins/headlamp-plugin/config/plugins-tsconfig.json` (but without `"baseUrl": "../../../../"`, obviously)

The `../../../../` will make things work when we release these changes and just have each plugin's tsconfig extending it.
Ashu did find a few things we should improve so I am working on that.

joaquimrocha:
I have removed just one useless line from my patch (made it look into a file that didn't exist, with no side-effects but I still removed it).

joaquimrocha:
Published with headlamp-plugin 0.4.6 .

PR #486, opened 2022-04-08
Headless app fix

illume:
- backend: Fix serving the static frontend only if needed
- app: Fix passing static dir arg for headless mode

## How to use

```bash
make app-linux
cd app/dist/linux-unpacked
./headlamp --headless
```

## Testing done

- `--headless` mode with app works
- `make run-frontend` / `make run-backend` still works


joaquimrocha (approved):
The changes LGTM. Take into account however that if I run this from WSL, I see that the loaded page is a folder in the Windows machine and this seems very much like a security issue. I thought this would be either introduced or fixed by your changes but it's the same in both main and this branch.

PR #491, opened 2022-04-11
plugins: Show how to use styles inside a plugin

illume:
Adds an example of styling inside a plugin.

## How to use

```
npm run start
```


plugins/examples/pod-counter/src/index.tsx
```diff
@@ -1,24 +1,30 @@
-import React from 'react';
-import { Headlamp, Plugin, Registry, K8s } from '@kinvolk/headlamp-plugin/lib';
+import { Headlamp, K8s,Plugin, Registry } from '@kinvolk/headlamp-plugin/lib';
```
joaquimrocha, line 1: npm run format yields changes here.

joaquimrocha (changes requested):

plugins/examples/pod-counter/src/index.tsx
joaquimrocha: I thought I might as well just push the updated commit since it's a simple change.

PR #495, opened 2022-04-21
Rename backend server to headlamp-server

### LINKED ISSUE #487: The name of server.exe might be conflicted with other process
## Description

In my computer, other process called server.exe is running, headlamp failed to start. Please replace server.exe to headlamp_server.exe or some other name.

joaquimrocha:
The former name was too generic, which made it difficult to spot, and
could clash on Windows as well.

This patch renames it as headlamp-server but keeps as /headlamp/server
symlink in the container image, for backward compatibility.

fixes #487

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
-backend/server
-backend/server.exe
+backend/headlamp-server
+backend/headlamp-server.exe
 backend/tools
 app/electron/src/*
 docs/development/storybook/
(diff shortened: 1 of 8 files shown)


illume:
Broken mac build fixed in here: https://github.com/kinvolk/headlamp/pull/498

illume (approved):
Thanks! 👍

PR #496, opened 2022-04-21
WIP: Headless close backend when close last browser window

illume:
for https://github.com/kinvolk/headlamp/issues/493

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -18,6 +18,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	oidc "github.com/coreos/go-oidc"
 	"github.com/gorilla/handlers"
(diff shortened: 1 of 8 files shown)


joaquimrocha:
I am a bit worried that this maybe lead to undesired exiting of the headlamp process, the rationale being: you open the app as a user, you should close it in a similar manner. For desktop apps, one clicks the icon which opens the app, then we close the window or call quit on the app and it quits the process. For a CLI initiated app though, usually either the process is forked/run in the background, or the user is supposed to close it as they started it (Ctrl+C normally). I also think it may be more normal for a user to close a browser tab and then open another for the same service, an experience that would be broken if the next tab has no server for it to reach.

An alternative would be to add a quit timeout and also a CLI option `--auto-quit=false` to prevent this behavior.

joaquimrocha:
Since the original issue was closed as not planned. Let's close this PR for cleanup reasons.
cc/ @illume

PR #497, opened 2022-04-22
Add clusters dynamically

### LINKED ISSUE #456: [RFE] Add functionality to allow setting up a cluster dynamically in the backend
## Current situation

Headlamp's backend simply reads the in-cluster configuration, or existing kubeconfig files.
This means that new clusters cannot be dynamically created which prevents functionality like: loading a cluster that's defined by a plugin or allowing the user to load a kubeconfig from the UI.

## Ideal future situation

We allow to dynamically set up new clusters in the backend, so we open new possibilities for cluster management in Headlamp.

## Implementation options

I think that we can have an API endpoint in the backend that allows to set up new clusters by passing it the cluster's details.
This endpoint should be off by default when running in-cluster, but there should be a CLI option to turn it on anyway, in case a Headlamp deployment wants to allow that.

joaquimrocha:
- [X] Backend allows to set up a cluster dynamically
- [X] Clusters set up dynamically show up in consequent calls to /config
- [x] Backend has tests for the new functionality
- [x] Plugins can add a cluster dynamically (we have a way to dispatch such a request from the frontend)

fixes #456 

## How to test

- Set up a new cluster, and remove details from your `~/.kube/config` 
- add the cluster details to the `plugins/examples/dynamic-clusters/src/index.tsx` source
- cd `plugins/examples/dynamic-clusters/ && npm start` the plugin
- start headlamp
- click the "New Cluster" button in the app bar at the top
- headlamp should reload with the new cluster

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -24,6 +24,17 @@ export interface RequestParams {
   [prop: string]: any;
 }
 
+export interface ClusterRequest {
+  /** The name of the cluster (has to be unique, or it will override an existing cluster) */
+  name: string;
+  /** The cluster URL */
+  server: string;
+  /** Whether the server's certificate should not be checked for validity */
+  insecureTLSVerify?: boolean;
+  /** The certificate authority data */
+  certificateAuthorityData?: string;
+}
+
 export async function request(
   path: string,
   params: RequestParams = {},
@@ -675,4 +686,13 @@ export async function testAuth() {
   return post('/apis/authorization.k8s.io/v1/selfsubjectrulesreviews', { spec }, false, {
     timeout: 5 * 1000,
   });
+}
+
+export async function setCluster(clusterReq: ClusterRequest) {
+  return request(
+    '/cluster',
+    { method: 'POST', body: JSON.stringify(clusterReq), headers: JSON_HEADERS },
+    false,
+    false
+  );
 }
--- a/plugins/examples/dynamic-clusters/package.json
+++ b/plugins/examples/dynamic-clusters/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "dynamic-clusters",
+  "version": "0.0.1",
+  "description": "A Headlamp (the extensible Kubernetes web UI) plugin that shows how to set up clusters dynamically",
+  "scripts": {
+    "start": "headlamp-plugin start",
+    "build": "headlamp-plugin build",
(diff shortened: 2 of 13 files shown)


joaquimrocha:
CI is failing because of the unreleased new function in the Headlamp class (for plugins).

illume:
Looks like there's a problem with the types somehow.
```
Error: src/index.tsx(9,18): error TS2339: Property 'setCluster' does not exist on type 'typeof Headlamp'.
make: *** [Makefile:76: plugins-test] Error 2
```

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -24,6 +24,13 @@ export interface RequestParams {
   [prop: string]: any;
 }
 
+export interface ClusterRequest {
```
illume, line 27: It would be good if these had tsdoc strings to explain what they are for.

plugins/examples/dynamic-clusters/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "dynamic-clusters",
+  "version": "0.0.1",
+  "description": "Your Headlamp plugin",
```
illume, line 4: The plugin description should go here.

plugins/examples/dynamic-clusters/src/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+function ClusterCreationButton() {
+  return (
+    <Button
+      onClick={() => {
+        Headlamp.setCluster({
+          name: 'my-plugin-set-cluster',
+          server: 'http://phony.abcde-example-k8s-cluster.k8scluster',
+        }).then(() => {
+          // We should find a better way to make sure the clusters info is reloaded.
```
illume, line 13: I think window.location.reload is good enough, so probably the comment can be removed?

plugins/examples/dynamic-clusters/src/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+          window.location.reload();
+        });
```
illume, line 15: I'm not sure if we need to show how errors are handled here. This could be added later to this PR.

illume (changes requested):
Looks good to me. There's the type error in CI that needs to be fixed though.

I left some other comments of things that could be improved, but I don't think any of them should block this being merged.

illume (approved):
🎈🎉 thanks

PR #503, opened 2022-04-27
Add better a11y error reporting, fix storybook, fix reported a11y issues

illume:
- Issues are logged _last_ in the console. To make them easier to read whilst the console is noisy.
- If any issues remain, an alert() is done to warn developers when Headlamp is run in development mode. 

Hopefully these measures will help prevent more issues escaping review, since the console log was quite noisy and can be forgotten.

Additionally:
- Started a section in the documentation on a11y tools.
- fixes a number of the reported a11y issues
- storybook was broken because of an iconify issue (added test of storybook to CI)
- apart from a11y messages in the console, it removes a few more lines from the dev console related to i18n

## How to use

During development
- "axe react" a11y messages should be printed at the end of the development console now (rather than somewhere in the middle).
- an alert message is shown if any a11y issues are reported

To disable the alert message during development, use the following:
```bash
REACT_APP_SKIP_A11Y=true make run-frontend
```

You can either introduce an a11y issue, or checkout the first commit to see the behaviour (because it will report issues).

## Testing done

- [x] go to each route, load the page and no alert and no a11y issues reported
- [x] with REACT_APP_SKIP_A11Y=true no alert is shown
- [x] storybook builds and runs

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useTranslation } from 'react-i18next';
 import { Redirect, Route, RouteProps, Switch } from 'react-router-dom';
 import { getToken } from '../../lib/auth';
 import { useClustersConf } from '../../lib/k8s';
@@ -8,6 +9,8 @@ import { useTypedSelector } from '../../redux/reducers/reducers';
 import { useSidebarItem } from '../Sidebar';
 
 export default function RouteSwitcher() {
+  const { t } = useTranslation('frequent');
+
   // The NotFoundRoute always has to be evaluated in the last place.
   const defaultRoutes = Object.values(ROUTES).concat(NotFoundRoute);
   const routes = useTypedSelector(state => state.ui.routes);
@@ -21,7 +24,7 @@ export default function RouteSwitcher() {
             <Route
               path={route.path}
               component={() => (
-                <PageTitle title={route.name ? route.name : route.sidebar}>
+                <PageTitle title={t(route.name ? route.name : route.sidebar)}>
                   <route.component />
                 </PageTitle>
               )}
--- a/frontend/src/components/account/Auth.tsx
+++ b/frontend/src/components/account/Auth.tsx
@@ -1,10 +1,9 @@
-import { Link } from '@material-ui/core';
 import Box from '@material-ui/core/Box';
 import Button from '@material-ui/core/Button';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogContentText from '@material-ui/core/DialogContentText';
-import DialogTitle from '@material-ui/core/DialogTitle';
+import Link from '@material-ui/core/Link';
 import Snackbar from '@material-ui/core/Snackbar';
 import TextField from '@material-ui/core/TextField';
 import React from 'react';
@@ -15,6 +14,7 @@ import { useClustersConf } from '../../lib/k8s';
 import { ApiError, testAuth } from '../../lib/k8s/apiProxy';
 import { getCluster, getClusterPrefixedPath } from '../../lib/util';
 import { ClusterDialog } from '../cluster/Chooser';
+import { DialogTitle } from '../common/Dialog';
 
 export default function AuthToken() {
   const history = useHistory();
@@ -101,7 +101,7 @@ export function PureAuthToken({
   }
 
   return (
-    <Box>
+    <Box component="main">
       <ClusterDialog useCover onClose={onClose} aria-labelledby="authtoken-dialog-title">
         <DialogTitle id="authtoken-dialog-title">{title}</DialogTitle>
         <DialogContent>
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -1,5 +1,5 @@
 import { InlineIcon } from '@iconify/react';
-import { Box, Button, DialogTitle, withStyles } from '@material-ui/core';
+import { Box, Button, withStyles } from '@material-ui/core';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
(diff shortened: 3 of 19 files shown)


frontend/src/components/account/Auth.tsx
```diff
@@ -101,9 +110,13 @@ export function PureAuthToken({
...
 
   return (
-    <Box>
+    <Box component="main">
       <ClusterDialog useCover onClose={onClose} aria-labelledby="authtoken-dialog-title">
-        <DialogTitle id="authtoken-dialog-title">{title}</DialogTitle>
+        <DialogTitle id="authtoken-dialog-title">
+          <Typography variant="h1" className={classes.title}>
+            {title}
+          </Typography>
```
joaquimrocha, line 118: Same comment about styling the whole component.

frontend/src/components/cluster/Chooser.tsx
```diff
@@ -101,6 +102,11 @@ const useStyles = makeStyles(theme => ({
     alignItems: 'center',
     display: 'flex',
   },
+  h1: {
+    fontSize: '1.25rem',
+    fontWeight: 500,
+    lineHeight: 1.6,
+  },
```
joaquimrocha, line 109: Ditto.

joaquimrocha: An alternative would be to use a wrapper component so we don't have to call typography + style all the time.

I'm curious to learn more about why the h1 is needed instead of h2 since I found some examples that use h2 like
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

joaquimrocha (commented):
Finally got to it.
The alert may seem a bit overkill at first sight, but I think it's excellent way for us not to let a11y issues slip through and I like the idea a lot. Good job!

I left some comments which I think may help us get the a11y right in the future with less changes, and some nits. Hope they make sense.

frontend/src/index.tsx
```diff
@@ -7,7 +7,26 @@ ReactDOM.render(<App />, document.getElementById('root'));
 
 if (process.env.NODE_ENV !== 'production') {
   const axe = require('@axe-core/react');
-  axe(React, ReactDOM, 500);
+  const axeCore = require('axe-core');
+
+  if (process.env.REACT_APP_SKIP_A11Y !== 'true') {
```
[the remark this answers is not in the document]
illume: It has the REACT_APP prefix because those environment vars are passed in to create react apps react-scripts. See [Adding Custom Environment Variables
](https://create-react-app.dev/docs/adding-custom-environment-variables/). When I tried something like HEADLAMP_xxx it isn't available.

frontend/src/lib/router.tsx
```diff
@@ -430,6 +430,7 @@ export const ROUTES: {
 export const NotFoundRoute = {
   path: '*',
   exact: true,
+  name: `Whoops! This page doesn't exist`,
```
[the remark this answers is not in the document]
illume: This string is translated because it's used on the 404 page.

I didn't look through every single title. Checked many and they seemed already translated (I guess because the side bar ones are translated already).

frontend/src/components/authchooser/index.tsx
```diff
@@ -234,16 +242,18 @@ export function PureAuthChooser({
...
           <Loader title={t('Testing auth')} />
         </Box>
       ) : (
-        <Box display="flex" flexDirection="column" alignItems="center">
+        <Box component="main" display="flex" flexDirection="column" alignItems="center">
           <DialogTitle ref={focusedRef} id="authchooser-dialog-title">
-            {title}
+            <Typography variant="h1" className={classes.title}>
+              {title}
+            </Typography>
```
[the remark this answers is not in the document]
illume: Yeah, all the dialog views need a good heading structure (h1, h2 etc) and navigation landmarks (main). It already correctly disabled the other content on the page. So with this change we have a main, and we have a h1.

Changing the font sizing keeps it as before whilst allowing the use of the h1 tag. The design of the header tags should be separate from the use of semantically correct document structure.

This is a bug in the Material UI component (luckily it was reported long ago and is fixed in newer versions). Ultimately, we should move to newer Material UI versions to get these and other fixes... but that's a bit bigger of a change because of the major version changes. When we update to the new version we can remove all this code.

frontend/src/components/cluster/Chooser.tsx
illume: I guess that example is assuming the document also has a h1 tag in the document. For ours, the dialog component sets aria-hidden=true on the other parts of the doc, leaving it with no h1 header. Only the dialog is visible in the document at this point to readers, and so we need a h1 tag for the dialog title.

illume: > An alternative would be to use a wrapper component so we don't have to call typography + style all the time.

Cool, I'll do that wrapper. It will be easier to delete later.

joaquimrocha: Yes, that's the idea. This makes things easier to fix now and also easier to replace when we move to a newer MUI. Thanks.

frontend/src/lib/router.tsx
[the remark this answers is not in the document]
illume: This is translated in the frontend/src/components/App/RouteSwitcher.tsx PageTitle component when the route is switched.

illume:
I added a DialogTitle wrapper which contains the a11y fixes (h1, styles and such) so it can be reused.

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -21,7 +24,7 @@ export default function RouteSwitcher() {
             <Route
               path={route.path}
               component={() => (
-                <PageTitle title={route.name ? route.name : route.sidebar}>
+                <PageTitle title={t(route.name ? route.name : route.sidebar)}>
```
illume, line 27: This is where it is translated.

joaquimrocha: I was thinking there's some static analysis on i18next-parser that gets the strings for translation, and that this was not recommended, but I am not 100% sure this is the case.
Maybe the one we discussed works in this example but we'll see lots of missing messages for the ones we don't have a translation for?

illume: Ah right... good point.

Yes, I'll check that. 
(Maybe it needs to have a separate PageTitle in each route so the tool can pick up the `t('')`.)

illume:
I'm going to leave the dynamic title translation in for this PR, and do that work after this is merged in, and the dependency PR and other PR(cluster chooser stories) is also done. They both depend on this PR being merged. It's at least an improvement over not translating the titles.

PR #513, opened 2022-05-18
Improve terminal user experience

fondoger:
image

Main features:
* Tty window auto resize
* Add shell support for different containers (Windows & Linux)

### DIFF
--- a/frontend/src/components/common/Terminal.tsx
+++ b/frontend/src/components/common/Terminal.tsx
@@ -5,7 +5,6 @@ import DialogActions from '@material-ui/core/DialogActions';
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogTitle from '@material-ui/core/DialogTitle';
 import FormControl from '@material-ui/core/FormControl';
-import Grid from '@material-ui/core/Grid';
 import InputLabel from '@material-ui/core/InputLabel';
 import MenuItem from '@material-ui/core/MenuItem';
 import Select from '@material-ui/core/Select';
@@ -20,12 +19,27 @@ const decoder = new TextDecoder('utf-8');
 const encoder = new TextEncoder();
 
 const useStyle = makeStyles(() => ({
+  dialogTitle: {
+    paddingBottom: 0,
+  },
   dialogContent: {
-    height: '80%',
-    minHeight: '80%',
+    paddingBottom: 20,
+    '& .xterm .xterm-viewport': {
+      width: 'initial !important', // BugFix: https://github.com/xtermjs/xterm.js/issues/3564#issuecomment-1004417440
+    },
+    '& #xterm-container': {
+      overflow: 'hidden',
+      width: '100%',
+      '& .terminal.xterm': {
+        padding: 10,
+      },
+    },
   },
   containerFormControl: {
     minWidth: '11rem',
+  },
+  xtermGrid: {
+    width: '100%',
   },
 }));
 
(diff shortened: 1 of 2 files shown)


joaquimrocha:
I think we need to split this commit in two:
1. Improve the window resizing
2. Add shell support for Windows containers

@fondoger , let me know if you can split this (can be in the same PR, but would be 2 commits). Otherwise, we can take it from here.

frontend/src/components/common/Terminal.tsx
```diff
@@ -150,35 +278,82 @@ export default function Terminal(props: TerminalProps) {
...
-          </Grid>
-          <Grid item>
-            <div id="xterm" ref={x => setTerminalContainerRef(x)} className="terminal" />
-          </Grid>
-        </Grid>
+      <DialogContent
+        className={classes.dialogContent}
+        style={{ height: '100%', display: 'flex', flexDirection: 'column', padding: 0 }}
+      >
+        <div
```
illume, line 295: A small nitpick... Probably `Box` could be used here instead of div.

illume (approved):
Thanks! Some very nice usability improvements. 🎈🎉

I saw @joaquimrocha already requested some changes which sound reasonable to me too. Left a couple of small comments for your consideration as well :)

frontend/src/lib/k8s/pod.ts
```diff
@@ -65,8 +66,8 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
     return cancel;
   }
 
-  exec(container: string, onExec: StreamResultsCb, options: ExecOptions = {}) {
-    const { command = ['sh'], reconnectOnFailure = undefined } = options;
+  exec(container: string, shell: string, onExec: StreamResultsCb, options: ExecOptions = {}) {
+    const { command = [shell], reconnectOnFailure = undefined } = options;
```
joaquimrocha, line 70: I don't think we need to add `shell` as an argument, when it can be taken from ExecOptions, which is more future-proof.

frontend/src/components/common/Terminal.tsx
```diff
@@ -116,24 +218,50 @@ export default function Terminal(props: TerminalProps) {
...
+        xtermRef.current.dispose();
+        if (execRef.current) {
+          execRef.current.cancel();
+        }
+        window.removeEventListener('resize', handler);
       };
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [container, terminalContainerRef]
+    [container, terminalContainerRef, shells.available[shells.currentIdx]]
```
joaquimrocha, line 264: Should we use the plain `shells.available` instead of `shells.available[shells.currentIdx]`?

frontend/src/components/common/Terminal.tsx
```diff
@@ -116,24 +218,50 @@ export default function Terminal(props: TerminalProps) {
...
+      const handler = () => {
+        fitAddonRef.current.fit();
+        console.log('resize');
+      };
+      window.addEventListener('resize', handler);
 
       return function cleanup() {
-        xterm.dispose();
-        exec.cancel();
+        console.log('[Terminal] run cleanup.');
```
joaquimrocha, line 255: Make this a debug log please.

frontend/src/components/common/Terminal.tsx
```diff
@@ -116,24 +218,50 @@ export default function Terminal(props: TerminalProps) {
...
+        );
 
-      const exec = item.exec(container, (items: ArrayBuffer) => onData(xterm, items));
+        setupTerminal(terminalContainerRef, xtermRef.current, fitAddonRef.current);
+      })();
 
-      setupTerminal(terminalContainerRef, xterm, fitAddon, exec);
+      const handler = () => {
+        fitAddonRef.current.fit();
+        console.log('resize');
```
joaquimrocha, line 250: Make this a debug log and add more context to the message.

frontend/src/components/common/Terminal.tsx
```diff
@@ -85,24 +123,88 @@ export default function Terminal(props: TerminalProps) {
...
+    }
+    if (!shells.automaticallySwitch) {
+      return false;
+    }
+
+    if (isShellNotFoundError(channel, text)) {
+      if (shells.currentIdx + 1 >= shells.available.length) {
+        xterm.write('Could not estable connection.');
+      } else {
+        xterm.write('Retrying connectting with shell: ' + shells.available[shells.currentIdx]);
```
joaquimrocha, line 199: There's a typo at `connectting` -> `connecting`.
And this message should be translatable.

frontend/src/components/common/Terminal.tsx
```diff
@@ -85,24 +123,88 @@ export default function Terminal(props: TerminalProps) {
...
+    if ((xterm as any).connected) {
+      return false;
+    }
+    if (!shells.automaticallySwitch) {
+      return false;
+    }
+
+    if (isShellNotFoundError(channel, text)) {
+      if (shells.currentIdx + 1 >= shells.available.length) {
+        xterm.write('Could not estable connection.');
```
joaquimrocha, line 197: This message should be translatable and say `Could not establish connection.`

frontend/src/components/common/Terminal.tsx
```diff
@@ -85,24 +123,88 @@ export default function Terminal(props: TerminalProps) {
...
     xterm.write(text);
   }
 
+  function isShellNotFoundError(channel: number, text: string): boolean {
+    // Linux container Error
+    if (channel === 3) {
+      try {
+        const error = JSON.parse(text);
+        if (error.code === 500 && error.status === 'Failure' && error.reason === 'InternalError') {
+          if (error.details.causes[0].message.includes('executable file not found')) {
```
joaquimrocha, line 172: I am not sure if looking at the error message is the correct way to identify the errors. Maybe we can just consider the error codes/statuses/reason instead of the message?

frontend/src/components/common/Terminal.tsx
```diff
@@ -85,24 +123,88 @@ export default function Terminal(props: TerminalProps) {
...
     }
 
+    // send resize command to server once connection is establised
+    if (!(xterm as any).connected && !!text) {
+      xterm.clear();
+      (async function () {
+        send(4, `{"Width":${xterm.cols},"Height":${xterm.rows}}`);
+      })();
+      (xterm as any).connected = true;
+      console.log('Connected!!!');
```
joaquimrocha, line 160: Make this a debug log.

frontend/src/components/common/Terminal.tsx
```diff
@@ -40,43 +54,67 @@ export default function Terminal(props: TerminalProps) {
...
-      // We just send the key strokes to the socket; the actual writing into the
-      // terminal will be done when the data arrives.
-      send(code, exec);
+    // Allow copy/paste in terminal
+    xterm.attachCustomKeyEventHandler(arg => {
+      if (arg.ctrlKey && arg.code === 'KeyC' && arg.type === 'keydown') {
+        const selection = xterm.getSelection();
+        if (selection) {
+          return false;
+        }
```
joaquimrocha, line 104: This means Ctrl+C without a selection will still abort the process, right? If so, this is very good! (I don't want to break the regular Unix Ctrl+C shortcut)

frontend/src/components/common/Terminal.tsx
```diff
@@ -40,43 +54,67 @@ export default function Terminal(props: TerminalProps) {
...
-        return;
-      }
+    xterm.onData(data => {
+      send(0, data);
+    });
 
-      // Send a newline when pressing enter
-      const code = event.domEvent.key === 'Enter' ? '\n' : event.key;
+    xterm.onResize(size => {
+      console.log('xterm resized: ' + size);
```
joaquimrocha, line 94: Please make this a debug log.

frontend/src/components/common/Terminal.tsx
```diff
@@ -40,43 +54,67 @@ export default function Terminal(props: TerminalProps) {
...
+
+  function getAvailableShells() {
+    const selector = item.spec?.nodeSelector || {};
+    const os = selector['kubernetes.io/os'] || selector['beta.kubernetes.io/os'];
+    if (os === 'linux') {
+      return ['bash', 'sh'];
+    } else if (os === 'windows') {
+      return ['powershell.exe', 'cmd.exe'];
+    }
+    return ['powershell.exe', 'bash', 'cmd.exe', 'sh'];
```
joaquimrocha, line 74: I know this is the "last resort" statement, but I'd reorder this as `sh, bash, powershell, cmd` because ATM, Linux containers are much more widely common than Windows, meaning we'll have a better guess by using a Unix-first ordering.

joaquimrocha (changes requested):
I gave a full review now and left some comments.
Overall, the changes are very nice. I think we can just make them a bit more maintainable in some parts of it.

PR #516, opened 2022-05-24
Handle custom auth for desktop app

ashu8912:
We add a custom protocol that can be used as the redirect URI, this is because we want to allow users to use there default browser for auth rather than the electron chromium window. Right now we only allow microsoft login auth to open in the default browser.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -415,6 +415,55 @@ function startElecron() {
       mainWindow = null;
     });
 
+    // Force Single Instance Application
+    const gotTheLock = app.requestSingleInstanceLock();
+    if (gotTheLock) {
+      app.on('second-instance', () => {
+        // Someone tried to run a second instance, we should focus our window.
+        if (mainWindow) {
+          if (mainWindow.isMinimized()) mainWindow.restore();
+          mainWindow.focus();
+        }
+      });
+    } else {
+      app.quit();
+      return;
+    }
+
+    /*
+    if a library is trying to open a url other than app url in electron take it 
+    to the default browser
+    */
+    mainWindow.webContents.on('will-navigate', (event, url) => {
+      if (url.startsWith(startUrl)) {
+        return;
+      }
+      shell.openExternal(url);
+    });
+
+    app.on('open-url', (event, url) => {
+      mainWindow?.focus();
+      let urlObj;
+      try {
+        urlObj = new URL(url);
+      } catch (e) {
+        dialog.showErrorBox(
+          i18n.t('Invalid URL'),
+          i18n.t('Application opened with an invalid URL: {{ url }}', { url })
+        );
+        return;
+      }
+
+      const urlParam = urlObj.hostname;
+      let baseUrl = startUrl;
+      // this check helps us to avoid adding multiple / to the startUrl when appending the incoming url to it
+      if (baseUrl.endsWith('/')) {
+        baseUrl = baseUrl.slice(0, startUrl.length - 1);
+      }
+      // load the index.html from build and route to the hostname received in the protocol handler url
+      mainWindow?.loadURL(baseUrl + '#' + urlParam + urlObj.search);
+    });
+
     i18n.on('languageChanged', () => {
       setMenu(i18n);
     });


app/electron/main.ts
```diff
@@ -386,6 +388,42 @@ function startElecron() {
...
+        }
+      });
+    } else {
+      app.quit();
+      return;
+    }
+
+    mainWindow.webContents.on('will-navigate', (event, url) => {
+      shell.openExternal(url);
+    });
```
joaquimrocha, line 416: Should we check that the URL doesn't have our app's internal base URL one? (in which case we should open it within Electron?)

app/electron/main.ts
```diff
@@ -384,6 +374,12 @@ function startElecron() {
       mainWindow.webContents.send('appVersion', appVersion);
     });
 
+    mainWindow.webContents.on('will-navigate', (e, url) => {
+      if (url.startsWith('https://login.microsoftonline.com')) {
+        shell.openExternal(url);
+      }
+    });
```
joaquimrocha, line 381: You already had a function adding this callback but not filtering the URL, should you remove the other one?

app/electron/main.ts
```diff
@@ -421,7 +417,14 @@ function startElecron() {
 
     app.on('open-url', (event, url) => {
       mainWindow.focus();
-      dialog.showErrorBox('Welcome Back', `You arrived from: ${url}`);
+      const urlObj = new URL(url);
+      const authCode = urlObj.hash.split('&')[0];
```
joaquimrocha, line 421: I think you should parse the URL here and check for the needed elements in a more robust way. (Unless the parse lib is less powerful than I thought and you tried it already)

ashu8912:
To test this PR with msal

1.  setup azure ad app registration
2. set redirect URI to be headlamp://redirect choosing the desktop auth flow
3. Right now headlamp communicates the auth_token or an auth_code(in case of PKCE flow) via the auth_token and auth_code event type, To get these on the renderer side you can just use the global desktopApi receive function and listen for these events from your auth flow

app/electron/main.ts
```diff
@@ -386,6 +388,42 @@ function startElecron() {
...
+
+    // Force Single Instance Application
+    const gotTheLock = app.requestSingleInstanceLock();
+    if (gotTheLock) {
+      app.on('second-instance', (e, argv) => {
+        // Someone tried to run a second instance, we should focus our window.
+
+        // Protocol handler for win32
+        // argv: An array of the second instance’s (command line / deep linked) arguments
+        if (process.platform == 'win32') {
```
joaquimrocha, line 409: This is the change breaking the CI.

app/electron/main.ts
```diff
@@ -1,5 +1,5 @@
 import { ChildProcessWithoutNullStreams, exec, spawn } from 'child_process';
-import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
+import { app, BrowserWindow, dialog,ipcMain, Menu, MenuItem, screen, shell } from 'electron';
```
joaquimrocha: This needs a space after the `,`.

joaquimrocha:
Moving to the next milestone.

ashu8912:
@joaquimrocha i think this will require a fresh review

joaquimrocha (changes requested):
Thanks for updating the PR @ashu8912 . However, I have noticed many comments have not been addressed, and I was hoping that the URL handling could be more naturally handled by the app, instead of adding new event handlers.

app/electron/main.ts
```diff
@@ -442,13 +442,22 @@ function startElecron() {
       return;
     }
 
+    // if a library is trying to open a microsoft login url in electron take it to the browser
     mainWindow.webContents.on('will-navigate', (event, url) => {
+      if (url.startsWith('https://login.microsoftonline.com')) {
+        shell.openExternal(url);
+      }
```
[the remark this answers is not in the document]
ashu8912: done updated this to check if it starts with app url then open in electron window otherwise use browser window

app/electron/main.ts
```diff
@@ -442,13 +442,22 @@ function startElecron() {
...
+    ipcMain.on('open_url_in_external_browser', (e, url) => {
       shell.openExternal(url);
     });
 
     app.on('open-url', (event, url) => {
-      mainWindow.focus();
-      dialog.showErrorBox('Welcome Back', `You arrived from: ${url}`);
+      mainWindow?.focus();
+      // pass whatever we received in the custom protocol url to the renderer process
+      mainWindow?.webContents.send('open-url', url);
```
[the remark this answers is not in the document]
ashu8912: fixed, now we handle the URL at this handler only and open it in the electron window, initially I used the events coz thought of leveraging the router helper methods to push into the route history. But this works too and has less moving parts

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -14,6 +14,13 @@ export default function RouteSwitcher() {
   // The NotFoundRoute always has to be evaluated in the last place.
   const defaultRoutes = Object.values(getDefaultRoutes()).concat(NotFoundRoute);
   const routes = useTypedSelector(state => state.ui.routes);
+  const history = useHistory();
+  React.useEffect(() => {
+    window.desktopApi?.receive('open-url', (url: string) => {
+      const urlObj = new URL(url);
+      history.replace({ pathname: '/' + urlObj.pathname.replace('//', ''), search: urlObj.search });
+    });
+  }, []);
```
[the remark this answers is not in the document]
ashu8912: fixed now we handle all of this in electron world

app/electron/main.ts
```diff
@@ -442,13 +442,22 @@ function startElecron() {
...
 
+    // if a library is trying to open a microsoft login url in electron take it to the browser
     mainWindow.webContents.on('will-navigate', (event, url) => {
+      if (url.startsWith('https://login.microsoftonline.com')) {
+        shell.openExternal(url);
+      }
+    });
+
+    // this event is for if someone wants to open a url in the external browser window
+    ipcMain.on('open_url_in_external_browser', (e, url) => {
```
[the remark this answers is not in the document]
ashu8912: removed the event now this part is handled by the part of code which handles opening whether a electron window or a browser window based on startURL

joaquimrocha (approved):

app/electron/main.ts
```diff
@@ -415,6 +415,49 @@ function startElecron() {
...
+      try {
+        const urlObj = new URL(url);
+        const urlParam = urlObj.hostname;
+        let massagedStartUrl = startUrl;
+        // this check helps us to avoid adding multiple / to the startUrl when appending the incoming url to it
+        if (massagedStartUrl.endsWith('/')) {
+          massagedStartUrl = massagedStartUrl.slice(0, startUrl.length - 1);
+        }
+        // load the index.html from build and route to the hostname received in the protocol handler url
+        mainWindow?.loadURL(massagedStartUrl + '#' + urlParam + urlObj.search);
```
joaquimrocha, line 455: We should just include the code we may expect to throw exceptions. 
So that would just be the urlObj creation.

joaquimrocha (changes requested):
Actually saw something that needs to be addressed.

app/electron/main.ts
ashu8912: @joaquimrocha there are other pieces in this code segment that are dependent on urlObj so I think they also will not work if URL is invalid.

app/electron/main.ts
```diff
@@ -415,6 +415,52 @@ function startElecron() {
...
+      shell.openExternal(url);
+    });
+
+    app.on('open-url', (event, url) => {
+      mainWindow?.focus();
+      let urlObj;
+      try {
+        urlObj = new URL(url);
+      } catch (e) {
+        throw new Error(`Invalid URL: ${url}`);
```
joaquimrocha, line 450: Before it was showing an error message? Now it's throwing an error which won't be caught. Shouldn't it show an error message and then return?

app/electron/main.ts
```diff
@@ -415,6 +415,52 @@ function startElecron() {
...
+      }
+      if (urlObj) {
```
joaquimrocha, line 452: If there's an early return in the catch, maybe this is not needed. If Ts complains about it still, then you can check and early return before doing this new block.

joaquimrocha (changes requested):
It's almost there but I think we are now not catching an error.

app/electron/main.ts
ashu8912: got it updated accordingly, thanks

PR #518, opened 2022-05-26
Fix frontend "npm start"

illume:
There was an issue with the dependencies, and `make run-frontend` wasn't working. 
(That's fixed in the first commit.)

Additionally, there were a number of errors and warnings introduced with the Notifications component. This PR removes the ones emitted on load but leaves other ones which become apparent when interacting with the component. There are also some other console warnings that could be removed, but those are left for another time.

I expect the es/pt locale files will need some attention :)

## How to use

`cd frontend && npm install && npm start`

Should run Headlamp with no big warning/error alert.

## Testing done

`cd frontend && npm install && npm start` runs.


joaquimrocha (approved):

joaquimrocha:
It works, with 10 warnings related to missing files.
@illume , is this something that impacts the rest of the build somehow?

illume:
I can't see any other effect. The warning seems to come from that component failing with typescript.

Probably, I'd like to address that separately to this PR (probably with a PR+another patch or replacing the dependency if upstream continues to be stalled).

PR #519, opened 2022-05-26
Update: nodeSelector

till:
The other has been deprecated since kubernetes v1.14

# update node selector

Updates the node selector to get rid off this warning:

```
service/headlamp created
Warning: spec.template.spec.nodeSelector[beta.kubernetes.io/os]: deprecated since v1.14; use "kubernetes.io/os" instead
deployment.apps/headlamp created
```

### DIFF
--- a/kubernetes-headlamp.yaml
+++ b/kubernetes-headlamp.yaml
@@ -41,4 +41,4 @@ spec:
           initialDelaySeconds: 30
           timeoutSeconds: 30
       nodeSelector:
-        'beta.kubernetes.io/os': linux
+        'kubernetes.io/os': linux


joaquimrocha:
@yolossn or @knrt10 , can you verify what's the best approach here?

knrt10:
It makes sense, but make sure that the node also has the labels assigned. I checked it on minkube it did had the required label `kubernetes.io/os`.

So I guess the change makes sense.

till:
I am not sure I understand the _discussion_. The label was there before but emits a deprecation warning.

I just updated it. I tested it on a k0s setup. It has the label as well.

till:
@joaquimrocha Thank you!

PR #527, opened 2022-05-30
Fix multiple kubeconfig bug

yolossn:


backend/cmd/headlamp.go
```diff
@@ -209,14 +210,27 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
-		var err error
-
-		contextsFound, err := GetContextsFromKubeConfigFile(kubeConfigPath)
-		if err != nil {
-			log.Println("Failed to get contexts from", kubeConfigPath, err)
+		var kubeConfigs []string
+		if runtime.GOOS == "windows" {
+			kubeConfigs = strings.Split(kubeConfigPath, ";")
+		} else {
+			kubeConfigs = strings.Split(kubeConfigPath, ":")
```
joaquimrocha, line 217: Nit: I'd assign the right delimiter to a var and avoid duplicating the logic (i.e. calling just one Split).

backend/cmd/headlamp.go
```diff
@@ -209,14 +210,27 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+		for _, kubeConfig := range kubeConfigs {
+			kubeConfig, err := absPath(kubeConfig)
+			if err != nil {
+				log.Println("Failed to get current user", err)
+				continue
+			}
+
+			contextsFound, err := GetContextsFromKubeConfigFile(kubeConfig)
+			if err != nil {
+				log.Println("Failed to get contexts from", kubeConfig, err)
```
joaquimrocha, line 229: Should we print these errors to stderr?

joaquimrocha (approved):
Left a couple of comments. One is a nit, the other (stderr) may require a bit more closer check.
Approving nonetheless since it looks good.

Would be great if we could have tests for this too, but if that's complicated for now, we can merge and do that later.

backend/cmd/headlamp.go
yolossn: All the logs are in Println, Ill fix it as a part of another PR ?

yolossn:
> Would be great if we could have tests for this too, but if that's complicated for now, we can merge and do that later.

Will require a refactor and to get this tested, lets do it later ?

PR #528, opened 2022-05-30
frontend: Add mark all as read and clear functionalities to notificaton

### LINKED ISSUE #517: [RFE] Allow to clear notifications
## Current situation

We have notifications showing, with the ability to mark them as read, but no possibility to hide/remove them.

## Impact

Users who don't want to see any of the past notifications, cannot choose to remove/hide them.

## Ideal future situation

We have a way to remove individual notifications or all of them.

## UI

Mockup of the new notifications UI, showing clear/remove options

## Implementation options

We should probably not actually delete the notifications but rather just hide them when the user clicks the X/clear buttons. The reason is that if we actually clear them from memory, then chances are that the user may see them come back if the logic is reloaded.

ashu8912:
This patch adds the clear and mark all as read functionality to notification popup

## How to use
Go to notification popup click on the Mark as All Read button and you will see all notifications marked as read, click on Clear button and you will see all notifications getting cleared.

fixes #517


frontend/src/components/App/Notifications/index.tsx
```diff
@@ -228,6 +252,35 @@ export default function Notifications() {
...
+          <Grid container justifyContent="space-between">
+            <Grid item>
+              <Box mx={1}>
+                <Typography style={{ fontWeight: 'bold' }}>{t('Notifications')}</Typography>
+              </Box>
+            </Grid>
+            <Grid item>
+              <Box display={'flex'} justifyContent="space-between">
+                <Box px={1}>
+                  <Link
```
illume, line 265: All features should be usable with the keyboard. If you try tabbing through, it's not possible to use the keyboard for these links (I think changing them to buttons would fix it).

illume (changes requested):
👍 Cool, thanks.

Left a few comments.

Would you mind addressing tooltip readability issues as reported by axe in the dev console?

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -131,11 +142,12 @@ export default function Notifications() {
   const dispatch = useDispatch();
   const [events] = Event.useList();
   const { t } = useTranslation();
+  const theme = useTheme();
 
   useEffect(() => {
     if (events && events.length !== 0) {
       const importantEvents: Notification[] = events
-        .filter((event: KubeEvent) => event.type !== 'Normal')
+        .filter((event: KubeEvent) => event.type === 'Normal')
```
joaquimrocha, line 150: This was added by mistake.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -228,6 +252,35 @@ export default function Notifications() {
...
+              </Box>
+            </Grid>
+            <Grid item>
+              <Box display={'flex'} justifyContent="space-between">
+                <Box px={1}>
+                  <Link
+                    style={{ textDecoration: 'none', cursor: 'pointer' }}
+                    onClick={handleNotificationMarkAllRead}
+                  >
+                    {t('frequent|Mark as All Read')}
```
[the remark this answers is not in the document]
joaquimrocha: And it should say `Mark all as read`.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -179,6 +191,18 @@ export default function Notifications() {
...
+  function handleNotificationMarkAllRead() {
+    const massagedNotifications = notifications.map(notification => {
+      notification.seen = true;
+      return notification;
+    });
+    dispatch(setUINotifications(massagedNotifications));
+  }
+
+  function handleNotificationClear() {
+    dispatch(setUINotifications([]));
```
joaquimrocha, line 203: This will have the effect I had warned about: upon refreshing the page, we get all the notifications back again. We should instead have a way to hide them, so for the user it looks like they've been deleted, but we need them around in order to know if they should be displayed.

joaquimrocha:
I have updated this PR to fix a couple of issues and ended up doing the following:
1. When refreshing the page, the notifications now show the same state as before (notifications which were marked as read will be shown as read, same for clear);
2. Disable the "Mark all as read" and "Clear" if their actions cannot be applied to the current notifications' state
3. Replace the use of a badge for the "mark as read" button by an icon button
4. Ensure that if there are less events than the previous shown notifications, unmatched older notification are still shown (it was only showing the same number of notifications as the events since the latter was used for filtering)
5. Add a tooltip to the "mark as read" button
6. Use the Empty component for displaying the empty state
7. Ensure tab navigation is possible in the notifications' popup

All notifications clear screenshot

joaquimrocha:
@illume Maybe you can take a look.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -133,43 +161,56 @@ export default function Notifications() {
   const { t } = useTranslation();
 
   useEffect(() => {
+    let notificationsToShow: Notification[] = [];
+    let currentNotifications = notifications;
+    let changed = false;
+
+    if (currentNotifications.length === 0) {
+      currentNotifications = JSON.parse(localStorage.getItem('notifications') || '[]');
+      changed = true;
```
illume, line 170: This block here causes the App.test.jsx to loop forever. Doesn't loop in the app though it seems.

joaquimrocha: Happens when there are no events. I have updated the PR with a fix and the tests passed now, so I will merge it. Thanks.

PR #542, opened 2022-06-06
i18n: Update using make i18n

joaquimrocha:
I can add the actual translations later as searching for what makes sense in k8s requires checking the docs for consistency, etc.


illume (approved):
👍

PR #544, opened 2022-06-07
app github: Build windows app without bash

illume:
## Bash, but also... not bash.

Before we were building the app on windows using bash.

- [x] We should support building the app on windows without bash. However, bugs keep creeping in because we only test making it with bash on CI.

## Testing done

This now works on windows with cmd (and make installed):
```bash
make app-win

# Also...
cd app
npm run copy-icons
npm run prod-deps
npm run copy-plugins
```

### DIFF
--- a/.github/workflows/app-artifacts-win.yml
+++ b/.github/workflows/app-artifacts-win.yml
@@ -27,8 +27,8 @@ jobs:
       with:
         args: install make
     - name: App Windows
+      shell: pwsh
       run: |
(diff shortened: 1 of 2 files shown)


illume:
Moving this item to be done in another PR: 

> The headlamp-plugin tests are only tested with bash on linux. This means we are not testing portable configurations to make sure headlamp-plugin works properly. (With windows and powershell for example).

joaquimrocha:
This seems to work fine now. Thanks.

PR #545, opened 2022-06-07
app: Kill Headlamp server process synchronously

joaquimrocha:
When the app is quitting, if we try to kill the Headlamp server
process asynchronously, there's a race condition which may end up in
the electron process being quit before the server kill instruction is
actually done. The result was that the Headlamp server would be left
running in Windows most of the times.

Testing done:
1. Launch the app, verify there's a headlamp-server process running
2. Quit the app: verify there's no headlamp-server process running


illume (approved):
👍 thanks

PR #547, opened 2022-06-08
Fix (again) + help quitting server

joaquimrocha:
I reflected more on the discussion about whether we should ask users to terminate running Headlamp server processes and remembered that any users of current Headlamp versions on Windows will hit this problem until they update to the new version.
So I started fixing this, working from Linux, and noticed that Linux was also leaving its headlamp-server processes running after the app quit. It also had an issue (we were killing the PID using its negative version, for killing a process group, but on Linux the process doesn't run detached, so there was an uncaught exception...)

Therefore, I worked again on making sure that we can start and finish the headlamp-server process from within the app without resorting to killing it by PID (since we do have a subprocess wrapper for it). This was not working because we were:
1. Running detached only on windows (so the negative kill in Linux was broken)
2. Running with a shell on non-windows, so killing the process would just kill the `sh -c` command (parent) and not the actual `headlamp-server`

Here is the dialog asking the user whether to kill ongoing processes:
confirmation dialog for quitting other headlamp servers

## Testing
I have tested running on both Linux and Windows. Not yet on Mac. It'd be also nice if someone tried it on Ubuntu as I believe it used to behave differently from Fedora.

- [x] Test on Ubuntu
- [X] Test on Windows
- [x] Test on Mac

Test 1:
1. Run `make run-backend` to have a running process
3. Launch the Headlamp app (or `cd app && npm run dev`) -> There should be a dialog asking to kill the process or quit
4. Try the "kill" option, the app should start after that
5. Repeat 2 and try the "quit" option, the app should quit

Test 2:
1. Run the app and quit it at slightly different times
2. Do this 10 times. There should be no dialogs or dangling headlamp-server processes

Test 3 (cannot kill):
1. Modify the app code where it checks whether the headlamp-server has been successfully killed (look for attempts and change it to 0)
2. Run the app -> The dialog should now say that it couldn't kill the process, and upon accepting it, it should quit the app.

Test 4:
1. Run headlamp-server as root
2. Then run the Headlamp app as a user -> It should show a dialog saying it couldn't kill the processes

cannot kill dialog

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,7 +1,7 @@
-import { ChildProcessWithoutNullStreams, execSync, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
 import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
-import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
+import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import fs from 'fs';
 import { i18n as I18n } from 'i18next';
 import open from 'open';
@@ -91,14 +91,6 @@ function quitServerProcess() {
 
   intentionalQuit = true;
   log.info('stopping server process...');
-  if (process.platform !== 'win32') {
-    // Negative pid because it should kill the whole group of processes:
-    //    https://azimi.me/2014/12/31/kill-child_process-node-js.html
-    process.kill(-serverProcess.pid);
-  } else if (process.platform === 'win32' && serverProcess) {
-    // Otherwise on Windows the process will stick around.
-    execSync('taskkill /pid ' + serverProcess.pid + ' /T /F');
-  }
 
   serverProcess.stdin.destroy();
   // @todo: should we try and end the process a bit more gracefully?
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -363,7 +363,6 @@ function startElecron() {
         preload: `${__dirname}/preload.js`,
       },
     });
-    mainWindow.loadURL(startUrl);
 
     mainWindow.webContents.setWindowOpenHandler(({ url }) => {
       // allow all urls starting with app startUrl to open in electron
@@ -466,6 +465,9 @@ function startElecron() {
       serverProcess = startServer();
       attachServerEventHandlers(serverProcess);
     }
+
+    // Finally load the frontend
+    mainWindow.loadURL(startUrl);
   }
 
   if (disableGPU) {
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,7 +1,8 @@
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
-import { app, BrowserWindow, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
-import log from 'electron-log';
+import { ChildProcessWithoutNullStreams, execSync, spawn } from 'child_process';
+import { app, BrowserWindow, dialog, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
+import log from 'electron-log';
+import find_process from 'find-process';
 import fs from 'fs';
 import { i18n as I18n } from 'i18next';
 import open from 'open';
@@ -299,6 +300,24 @@ function setMenu(i18n: I18n) {
   Menu.setApplicationMenu(menu);
 }
 
+async function getRunningHeadlampPIDs() {
+  const processes = await find_process('name', 'headlamp-server.*');
+  if (processes.length === 0) {
+    return null;
+  }
+
+  return processes.map(pInfo => pInfo.pid);
+}
+
+function killProcess(pid: number) {
+  if (process.platform === 'win32') {
+    // Otherwise on Windows the process will stick around.
+    execSync('taskkill /pid ' + pid + ' /T /F');
+  } else {
+    process.kill(pid, 'SIGHUP');
+  }
+}
+
 function startElecron() {
   log.transports.file.level = 'info';
   log.info('App starting...');
@@ -315,7 +334,7 @@ function startElecron() {
 
   setMenu(i18n);
 
-  function createWindow() {
+  async function createWindow() {
     let frontendPath = '';
     if (isDev) {
       frontendPath = path.resolve('..', 'frontend', 'build', 'index.html');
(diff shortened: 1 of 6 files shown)


joaquimrocha:
I have now tested in WSL, Ubuntu, Windows, and Mac, and it worked fine. Hope it's the last we hear from this problem.

app/electron/main.ts
```diff
@@ -383,6 +402,63 @@ function startElecron() {
...
+        }
+      }
+
+      // If we reach here, then we attempted to kill headlamp-server. Let's make sure it's killed
+      // before starting our own, or else we may end up in a race condition (failing to start the
+      // new one before the existing one is fully killed).
+      if (!!runningHeadlamp) {
+        let stillRunning = true;
+        const waitTime = 250; //ms
+        const attempts = (10 * 1000) / 250; // 4 times per sec, 10 secs max
```
illume, line 431: Attempts is 40 here. The comment is a puzzle to me.

app/electron/main.ts
```diff
@@ -383,6 +402,63 @@ function startElecron() {
...
+        for (let tries = 0; tries < attempts && stillRunning; tries++) {
+          console.debug(
+            `Checking if Headlamp is still running after we asked it to be killed; try=${
+              tries + 1
+            }/${attempts}`
+          );
+          // Wait 250ms
+          await setTimeout(() => {}, waitTime);
```
illume, line 439: For the most part, I guess the process will be killed in much less than 250ms. An improvement could be to not wait before checking the first time. Another improvement could be to back off a bit more each time (2ms, 4ms, 8ms, 16ms, 32ms, 64ms, 128ms) so as to reduce the wait time.

All that said, 250ms is not so long to wait... so this seems fine to me.

app/electron/main.ts
```diff
@@ -383,6 +402,63 @@ function startElecron() {
...
+        }
+
+        // If we couldn't kill the process, warn the user and quit.
+        const processes = await getRunningHeadlampPIDs();
+        if (!!processes) {
+          dialog.showMessageBoxSync({
+            type: 'warning',
+            title: i18n.t('Failed to quit the other running process'),
+            message: i18n.t(
+              `Could not quit the other running process, PIDs: {{ process_list }}. Please stop that process and relaunch the app.`,
```
illume, line 452: Probably need to add these to the locale files.

How about mentioning the process name ("headlamp-server") in the message too?

> Could not quit the "headlamp-server" background process, ...

app/electron/main.ts
```diff
@@ -383,6 +402,63 @@ function startElecron() {
     });
 
     if (!useExternalServer) {
+      const runningHeadlamp = await getRunningHeadlampPIDs();
+      if (!!runningHeadlamp) {
+        const resp = dialog.showMessageBoxSync(mainWindow, {
+          // Avoiding mentioning Headlamp here because it may run under a different name depending on branding (plugins).
+          title: i18n.t('Other process running'),
+          message: i18n.t(
+            'Looks like another process is already running. Continue by stopping that process automatically, or quit?'
```
illume, line 411: To someone who doesn't know how Headlamp works, probably they need to know Headlamp runs a background process.

app/electron/main.ts
```diff
@@ -307,6 +300,32 @@ function setMenu(i18n: I18n) {
   Menu.setApplicationMenu(menu);
 }
 
+async function getRunningHeadlampPIDs() {
+  const processes = await find_process('name', 'headlamp-server.*');
```
illume, line 304: find_process also works with port. In the future that could be used to detect other not-headlamp-processes squatting on our port.

illume (approved):
👍 looks mighty fine to me.

I suggested some potential improvements, but I don't think they are necessarily needed before merging.

app/electron/main.ts
joaquimrocha: There was a race between killing + the process effectively getting killed (maybe my computer is too fast). I can do an iterative back-off but I don't think it's worth anything <10ms...

app/electron/main.ts
joaquimrocha: I think that the info that "another process is running" in the context of Headlamp (or whatever branding) is less confusing than talking about background processes.

app/electron/main.ts
joaquimrocha: Makes sense. But a fix for that would be to run it with another port or show a warning. We can do that in a different PR.

illume:
Probably we should make sure the frontend is stopped before trying to stop the backend. If the frontend is making requests frequently (as it does), then it's possible the frontend will keep the backend running. Because the backend will recieve new requests.

Asking the backend to elegantly stop is something we could do. Of course a SIGHUP (and windows equivalent) **should** do that. However, maybe we can directly ask the backend to stop instead. So it can close down elegantly. I don't know if the backend does actually gracefully exit now via SIGHUP.

After giving the backend time to stop elegantly, what about doing a force quit (`kill -9` and so)?

app/electron/main.ts
joaquimrocha: I have avoided mentioning "headlamp" so users who completely override the branding don't get users confused. Not sure if this is a good exception or not...

joaquimrocha:
> Probably we should make sure the frontend is stopped before trying to stop the backend. If the frontend is making requests frequently (as it does), then it's possible the frontend will keep the backend running. Because the backend will recieve new requests.
> 
> Asking the backend to elegantly stop is something we could do. Of course a SIGHUP (and windows equivalent) **should** do that. However, maybe we can directly ask the backend to stop instead. So it can close down elegantly. I don't know if the backend does actually gracefully exit now via SIGHUP.
> 
> After giving the backend time to stop elegantly, what about doing a force quit (`kill -9` and so)?

I think it's risky to now go and implement all that (since I tested these changes thoroughly), besides delaying the release.
Let's do that in another PR.

joaquimrocha:
I have updated it to:
* implement some sort of exponential back-off
* fixed an issue in the waiting time logic (also reduced it to 3secs most)
* avoid waiting if there was a problem trying to kill the process (permissions for example)
* load the frontend only after we successfully kill the server (unless we're running with an external server)

app/electron/main.ts
illume: "Continue by stopping" confused me at first.

I'm not sure if "Continue by terminating..." would be better.

["Terminate and Continue"] button?

hrmm. I'm not sure.

app/electron/main.ts
illume: Yeah, ok. Probably better to avoid mentioning Headlamp.

app/electron/main.ts
```diff
@@ -383,9 +393,81 @@ function startElecron() {
...
+        let timeWaited = 0;
+        const maxWaitTime = 3000; // ms
+        // @todo: Use an iterative back-off strategy for the wait (so we can start by waiting for shorter times).
+        for (let tries = 1; timeWaited < maxWaitTime && stillRunning; tries++) {
+          console.debug(
+            `Checking if Headlamp is still running after we asked it to be killed; ${tries} ${timeWaited}/${maxWaitTime}ms wait.`
+          );
+
+          // Wait (10 * powers of 2) ms with a max of 250 ms
+          const waitTime = Math.min(10 * tries ** 2, 250); // ms
```
illume, line 439: Cool... So now it's: 10, 40, 90, 160, 250. That seems good to me.

joaquimrocha: More like 10, 40, 80, 160, 250.

app/electron/main.ts
joaquimrocha: I have replaced stopping by terminating. I kept the buttons because it's confusing and long to have "Terminate and continue"

PR #548, opened 2022-06-10
Fix crash bug due to null reference in log viewer

fondoger:

## How to use

## Testing done


illume (approved):
👍 thanks for he fix.

joaquimrocha:
@fondoger , thanks. I don't think the type annotations are necessary for those functions since their respective states have been declared with the `boolean` type. Let me know if this was still failing for you.

I will separate the fix from the type changes since they should be in different commits anyway.

PR #549, opened 2022-06-14
WIP: Plugin docs improvements

illume:

## How to use

## Testing done


frontend/src/plugin/registry.tsx
```diff
@@ -124,44 +134,272 @@ export default class Registry {
...
+   * }
+   *
+   * registerClusterChooser(MyClusterChooser)
    * ```
+   *
+   * @see {@link http://github.com/kinvolk/headlamp/plugins/examples/clusterchooser/ Cluster Chooser Example}
+   *
    */
-  registerClusterChooserComponent(component: clusterChooserButtonComponent | null) {
+  registerClusterChooser(component: React.ComponentType<ClusterChooserProps> | null) {
```
joaquimrocha, line 206: IIRC we had a way to override the cluster chooser (dialog) or the cluster chooser button.
Is this still possible?

frontend/src/plugin/registry.tsx
```diff
@@ -124,44 +134,272 @@ export default class Registry {
...
   }
+
+  /**
+   * @deprecated Please use registerClusterChooser. This will be removed in headlamp-plugin 0.4.11.
+   */
+  registerClusterChooserComponent(component: React.ComponentType<ClusterChooserProps> | null) {
+    console.log(
+      'registerClusterChooserComponent is deprecated. Please use registerClusterChooser.'
+    );
+    store.dispatch(setClusterChooserButtonComponent(component));
```
joaquimrocha, line 217: Maybe this should call the new function, so we don't repeat code?

plugins/examples/change-logo/src/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const show = 'simple';
+if (show === 'simple') {
+  registerAppLogo(SimpleLogo);
+} else if (show === 'text') {
+  registerAppLogo(<p>My Logo</p>);
+} else {
+  registerAppLogo(ReactiveLogo);
+}
```
joaquimrocha, line 59: I wonder if we should return/export code like this in a function as we talked about at some point.

illume:
These are my remaining items left todo.

- [ ] Add PR description (start from change-logo)
- [ ] Fix details-view readme
- [ ] Make a consistent Props/Type for Header actions that is used by many of them. (ActionButton). Should there be a component of ours which imports the material ui stuff, and provides a simpler interface?
- [ ] update "`@example`" in registerDetailsViewHeaderAction, registerDetailsViewSection after the consistent interface is introduced.
- [ ] Add screenshot for details-view
- [ ] Update functionality.md for details-view
- [ ] Add new headlamp-plugin release version commit 0.5.0
- [ ] Add `make docs` commit

frontend/src/components/common/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,52 @@
+import { Icon, IconifyIcon } from '@iconify/react';
+import IconButton from '@material-ui/core/IconButton';
+import Tooltip from '@material-ui/core/Tooltip';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export interface ActionButtonProps {
+  /** Label is a description of the action.
+   *
+   * Internally it is passed into t() i18n translation.
```
joaquimrocha, line 10: This works for existing translations but will not work for ones that need to be parsed into a catalog.
So I'd actually not pass this into a translation function but just describe how to translate strings.

frontend/src/components/common/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,52 @@
+import { Icon, IconifyIcon } from '@iconify/react';
+import IconButton from '@material-ui/core/IconButton';
+import Tooltip from '@material-ui/core/Tooltip';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export interface ActionButtonProps {
```
joaquimrocha, line 7: It's not clear to me what an ActionButton does. If this is about adding a tooltip, maybe we can override the IconButton and Button and receive a tooltip prop instead?

frontend/src/components/common/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onClick,
+  color,
+  width,
+  edge,
+}: ActionButtonProps) {
+  const { t } = useTranslation();
+  return (
+    <Tooltip title={t(label)}>
+      <IconButton
+        aria-label={t(label)}
```
joaquimrocha, line 41: I can imagine many examples where the tooltip for a button shouldn't match the label of that button. E.g.:
label -> Delete.
tooltip -> Will only work if you have the permissions.

frontend/src/components/common/Resource/ViewButton.tsx
```diff
@@ -14,23 +12,13 @@ export interface ViewButtonProps {
...
 function ViewButton({ item, initialToggle = false }: ViewButtonProps) {
   const [toggle, setToggle] = React.useState(initialToggle);
-  const { t } = useTranslation('resource');
 
   function handleButtonClick() {
     setToggle(toggle => !toggle);
   }
   return (
     <>
-      <Tooltip title={t('View YAML') as string}>
```
joaquimrocha, line 24: Adding to what I mentioned before, this gets picked up by i18next-parser, but with the new ActionButton it shouldn't. So the translation is lost unless it exists already (meaning it needs to be parsed from elsewhere).

joaquimrocha (changes requested):
This was quite some endeavor! Thanks a lot for all this effort in simplifying things @illume .
I have tried the examples and everything looks fine AFAICS on VS Code as well. I have left a few comments about the things I have noticed could be improved IMO.

If the ActionButton requires some debate, we can extract it from this PR so it doesn't hold things back.

BTW, running into the issue with the pod-counter I realized that maybe we should have a way to catch those exceptions coming from plugins and then turn off the plugin in question. This is made more easily/dynamically if we had a way to identify where each plugin component/function came from but that requires dropping the plain `register*` functions AFAIU.
Mentioning this because this PR is a big step in plugin development and I want to make sure we have some future-proofability in the code.

BTW2, AFAIU this is backward compatible, right? If it isn't then we need to add and enforce those version checks we've talked about.

Thanks again!

plugins/examples/clusterchooser/README.md
```diff
@@ -0,0 +1,24 @@
+# Example Plugin: ClusterChooserButton override
```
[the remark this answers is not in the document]
illume: Thanks, fixed.

docs/development/plugins/functionality.md
```diff
@@ -60,7 +60,7 @@ what we have so far:
 
 ### Changing The Logo
 
-- API reference for [registerAppLogo](https://kinvolk.github.io/headlamp/docs/latest/development/api/classes/plugin_registry.registry/#registerapplogo)
+- API reference for [registerAppLogo](https://kinvolk.github.io/headlamp/docs/latest/development/api/modules/plugin_registry/#registerapplogo)
```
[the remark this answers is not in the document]
illume: Thanks, fixed.

frontend/src/components/common/ActionButton/ActionButton.tsx
illume: Thanks. A longDescription has been added for this case.

frontend/src/components/common/Resource/ViewButton.tsx
illume: Thanks. Fixed.

plugins/examples/clusterchooser/README.md
```diff
@@ -0,0 +1,24 @@
...
+git clone git@github.com:kinvolk/headlamp.git
+cd headlamp/plugins/examples/clusterchooser/
+npm install
+npm start
+# See the changed cluster chooser button in the top middle of Headlamp
+```
+
+The main code for the plugin is in [src/index.tsx](src/index.tsx).
+
+<img src="" />
```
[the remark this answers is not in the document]
illume: Thanks, fixed.

docs/development/plugins/functionality.md
```diff
@@ -58,6 +58,11 @@ The plugin registers makes functionality available to plugins in an easy way.
 The idea is to make more and more functionality available to plugins. Here is
 what we have so far:
 
+### Changing The Logo
+
+- API reference for [registerAppLogo](https://kinvolk.github.io/headlamp/docs/latest/development/api/classes/plugin_registry.registry/#registerapplogo)
```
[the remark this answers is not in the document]
illume: Thanks, fixed.

plugins/examples/pod-counter/README.md
```diff
@@ -0,0 +1,19 @@
+# Example Plugin: Pod Counter
```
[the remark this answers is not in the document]
illume: This is fixed. Turns out that instantiating components inside loops is against the rule of hooks.

PR #550, opened 2022-06-14
frontend: Fix unexpected undefined value in OwnedPodsSection

joaquimrocha:
Just hit an error when apparently the labels of a pod are null/undefined but they are not expected to be so in the OwnedPodsSection.


illume (approved):
👍

PR #558, opened 2022-06-15
github: Add emojis to the release draft creation action

joaquimrocha:
These emojis have been used for a couple of releases. Let's add
them. Also adds a link to the container image URL.


illume (approved):
👍 emojis

PR #559, opened 2022-06-15
frontend: Do not duplicate children in dynamic sidebar items

joaquimrocha:
When items were set up in the sidebar from a plugin, the object they
were using was the one set up in redux. This meant that any children
of that sidebar entry would be appended to the object's "subList" in
redux and hence that list would grow with multiple equal entries
every time the sidebar was rendered.

This patch prevents that by deep cloning the object before using it.

### DIFF
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import { SidebarEntry } from '../../redux/reducers/ui';
 import store from '../../redux/stores/store';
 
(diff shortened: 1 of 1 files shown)


joaquimrocha:
I think there may still be a problem rendering dynamic sidebar sub-0 level entries (I couldn't see them rendered in a quick test), so I added that as a todo in the example list: https://github.com/kinvolk/headlamp/issues/537

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -135,7 +136,7 @@ function prepareRoutes(t: (arg: string) => string) {
       placement = parent['subList'];
     }
 
-    placement.push(item);
+    placement.push(_.cloneDeep(item));
```
illume, line 139: Perhaps cloneDeep at the top of the loop, and use the clonedItem everywhere?

This way parent will also be cloned.

illume (changes requested):
Parent cloning needed too?

frontend/src/components/Sidebar/prepareRoutes.ts
joaquimrocha: The parent will come from the routes list or the cloned item, so I don't think it will be used without being a new/cloned object. But I have added the clone to the beginning of the item use, as it's safer for future changes to have it there. Thanks.

illume (approved):
👍

PR #565, opened 2022-06-20
app: Do not create a one-click installer for Windows

joaquimrocha:
We want to provide Windows users with installation options.


illume (approved):
👍 good idea. Thanks.

PR #566, opened 2022-06-21
Allow to override app menus from plugins

joaquimrocha:
@illume , I have not touched the README of the app-menus plugin since it should be consistent with whatever new format you're going to land.

Testing:
- Run the app-menus plugin and see that it creates a new menu called "Chat with us".

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -93,6 +93,10 @@ function quitServerProcess() {
   intentionalQuit = true;
   log.info('stopping server process...');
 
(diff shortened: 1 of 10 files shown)


illume (approved):
👍 this is really nice. Thanks!

illume:
> @illume , I have not touched the README of the app-menus plugin since it should be consistent with whatever new format you're going to land.

Cool, I'll merge it in and do that in the other PR.

PR #567, opened 2022-06-24
frontend: Expose Resource and fix issue in PageGrid

joaquimrocha:
PageGrid was resulting in an error when it was given just a single
child. This patch corrects the types for the sections and children
properties of the PageGrid and updates how the sections are used as
children.

Besides, the Resource components were not being exported, so plugins couldn't use them.

@illume , Please take this into account for the plugin improvements you're working on.
Here's an example that will compile correctly but fail to load in the browser unless this PR is used.

```typescript
import { Headlamp, Plugin, Registry } from '@kinvolk/headlamp-plugin/lib';
import { SectionBox } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
import { PageGrid } from '@kinvolk/headlamp-plugin/lib/CommonComponents/Resource';
import React from 'react'

function Overview() {
  return (
    <PageGrid>
      <SectionBox>
        Hello world!!!
      </SectionBox>,
    </PageGrid>
  );
}

class FeedbackSidebar extends Plugin {
  initialize(registry: Registry) {
    console.log('feedback initialized');

    registry.registerSidebarItem(null, 'feedback', 'Feedback', '/feedback', {
      icon: 'mdi:comment-quote',
    });
    registry.registerSidebarItem('cluster', 'more-feedback', 'Feedback!', '/feedback');

    registry.registerRoute({
      path: '/feedback',
      sidebar: 'Feedback',
      component: () => <Overview />

    });

    return true;
  }
}

Headlamp.registerPlugin('sidebar', new FeedbackSidebar());
```


illume (approved):
👍 thanks

illume:
- ps. I'll fix the import path in the other PR, so it matches the headlamp source a bit more closely.
`import { PageGrid } from '@kinvolk/headlamp-plugin/lib/components/common/Resource';`
- pps. I'll use the PageGrid in one of the examples like you have there in the other PR too.
- ppps. I think when we finish doing stories for everything, part of refactoring will be to make sure we export all of the interfaces. So those interfaces can be used in the plugins as well as the stories.
- pppppppppppppps. Probably we should get some e2e testing done (maybe with minikube). This would catch these types of errors where compiling the plugin works, but it doesn't run because something isn't exported.

PR #573, opened 2022-06-30
frontend: Add refresh token logic

yolossn:
This patch adds logic to check if the token
is about to expire and refreshes it.

Signed-off-by: yolossn <nssvlr@gmail.com>

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -8,16 +8,21 @@
  */
 
 import { OpPatch } from 'json-patch';
+import { decodeToken } from 'react-jwt';
 import helpers from '../../helpers';
-import { getToken, logout } from '../auth';
+import { getToken, logout, setToken } from '../auth';
 import { getCluster } from '../util';
 import { KubeMetrics, KubeObjectInterface } from './cluster';
+import { KubeToken } from './token';
 
 const BASE_HTTP_URL = helpers.getAppUrl();
 const BASE_WS_URL = BASE_HTTP_URL.replace('http', 'ws');
 const CLUSTERS_PREFIX = 'clusters';
 const JSON_HEADERS = { Accept: 'application/json', 'Content-Type': 'application/json' };
 const DEFAULT_TIMEOUT = 2 * 60 * 1000; // ms
+const MIN_LIFESPAN_FOR_TOKEN_REFRESH = 10; // sec
+
+let isTokenRefreshInProgress = false;
 
 export interface RequestParams {
   timeout?: number; // ms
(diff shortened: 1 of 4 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -35,6 +37,64 @@ export interface ClusterRequest {
...
+  // convert expiry seconds to date object
+  const expiry = decodedToken.exp;
+  const now = new Date().valueOf();
+  const expDate = new Date(0);
+  expDate.setUTCSeconds(expiry);
+
+  // calculate time to expiry in minutes
+  const diff = (expDate.valueOf() - now) / 60000;
+  // If the token is about to expire, refresh it
+  if (diff < 10) {
```
joaquimrocha, line 61: Maybe using a const instead of the "magic" 10 number makes it easier to understand: `const minLifeSpanForRefresh = 10; // sec`.

BTW, 10 mins is probably not very effective, i.e. If I quit Headlamp for 15 mins and I am unlucky, I would have to re-login.
If the normal lifespan of tokens is 24h, maybe we can make the min lifespan for a refresh be 2h or 1.5h?

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -35,6 +37,64 @@ export interface ClusterRequest {
...
+    let namespace = '';
+    let serviceAccountName = '';
+    if (decodedToken && decodedToken['kubernetes.io']) {
+      namespace = decodedToken['kubernetes.io']['namespace'] || '';
+      const serviceAccount = decodedToken['kubernetes.io']['serviceaccount'] || {};
```
joaquimrocha, line 66: Maybe it's worth assigning `decodedToken['kubernetes.io']` to a variable and use that in these 3 places?

frontend/src/lib/k8s/apiProxy.ts
joaquimrocha, line 61: Also, I think early returning if the token is young enough may make the code cleaner.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -55,6 +115,9 @@ export async function request(
   let fullPath = path;
   if (useCluster && cluster) {
     const token = getToken(cluster);
+
+    await refreshToken(token);
```
joaquimrocha, line 119: Could this possibly lead to a token being refreshed in less than a fraction of seconds if we have "simultaneous" requests throughout Headlamp? Maybe we'll have to guard for that, otherwise we may get a token refreshed multiple times in a very short time span.

joaquimrocha (changes requested):
Thanks for these changes.
Besides the code comments, I trust this will only try to refresh tokens in K8s >=1.24 because of the decoded token's structure, right? If not, maybe we need to make sure we're not breaking or leading to preventable error loggings in sub-1.24.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -8,16 +8,21 @@
...
+import { KubeToken } from './token';
 
 const BASE_HTTP_URL = helpers.getAppUrl();
 const BASE_WS_URL = BASE_HTTP_URL.replace('http', 'ws');
 const CLUSTERS_PREFIX = 'clusters';
 const JSON_HEADERS = { Accept: 'application/json', 'Content-Type': 'application/json' };
 const DEFAULT_TIMEOUT = 2 * 60 * 1000; // ms
+const MIN_LIFESPAN_FOR_TOKEN_REFRESH = 10; // sec
+
+var TOKEN_REFRESH_INPROGRESS = false;
```
joaquimrocha, line 25: This shouldn't be all caps. All caps are used here just for constants.

joaquimrocha: Also, maybe separate `INPROGRESS` -> `IN_PROGRESS`.

illume: let instead of var

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -35,6 +40,77 @@ export interface ClusterRequest {
...
+  const expDate = new Date(0);
+  expDate.setUTCSeconds(expiry);
+
+  // calculate time to expiry in minutes
+  const diff = (expDate.valueOf() - now) / 60000;
+  // If the token is not about to expire return
+  if (diff > MIN_LIFESPAN_FOR_TOKEN_REFRESH) {
+    return;
+  }
+  let namespace = '';
```
illume, line 67: For variables which are initialized in one place and not changed, it's more readable to declare them `const` IMHO. Then it's easier to read knowing that the variable won't change somewhere else.

```ts
const namespace = decodedToken && decodedToken['kubernetes.io'] && decodedToken['kubernetes.io']['namespace'] || '';
```

However if the service name or namespace are empty, should we continue? I think this should be a fail case.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -35,6 +40,77 @@ export interface ClusterRequest {
...
+  // convert expiry seconds to date object
+  const expiry = decodedToken.exp;
+  const now = new Date().valueOf();
+  const expDate = new Date(0);
+  expDate.setUTCSeconds(expiry);
+
+  // calculate time to expiry in minutes
+  const diff = (expDate.valueOf() - now) / 60000;
+  // If the token is not about to expire return
+  if (diff > MIN_LIFESPAN_FOR_TOKEN_REFRESH) {
```
illume, line 64: From reading the code it's not clear to me what units diff, or MIN_LIFESPAN_FOR_TOKEN_REFRESH are.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -35,6 +40,77 @@ export interface ClusterRequest {
...
+    `api/v1/namespaces/${namespace}/serviceaccounts/${serviceAccountName}/token`
+  );
+  const tokenData = {
+    kind: 'TokenRequest',
+    apiVersion: 'authentication.k8s.io/v1',
+    metadata: { creationTimestamp: null },
+    spec: { expirationSeconds: 86400 },
+  };
+
+  const response = await fetch(tokenUrl, {
```
illume, line 95: There should be some error handling with fetch: See https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful

Needs a try/catch or a .catch() added.

frontend/src/lib/k8s/apiProxy.ts
ashu8912: @illume  converted both properties to seconds and commented as well

frontend/src/lib/k8s/apiProxy.ts
ashu8912: added

joaquimrocha (approved):
LGTM. We should double check whether the `await refreshToken...` needs to be prevented when using OIDC though. Maybe @yolossn can check it out tomorrow.

PR #574, opened 2022-07-01
Add scaling UI

### LINKED ISSUE #530: [RFE] Add special UI for setting up the replica number in deployments
## Current situation

Users have to modify the YAML in order to change the replicas amount in deployments.

## Impact

Editing the YAML just for this is a bit cumbersome and can lead to issues.

## Ideal future situation

We have a dedicated UI element in the deployments' details view for setting up the desired replicas.

## Implementation options

TBD

joaquimrocha:
This PR adds a scaling button to the header actions of Deployments, StatefulSets, and ReplicaSets. Doing this it became clear that we really need to try to find an alternative to the current API proxy.

Header actions screenshot

Scaling dialog screenshot

fixes #530

### DIFF
--- a/frontend/src/components/common/Resource/ScaleButton.tsx
+++ b/frontend/src/components/common/Resource/ScaleButton.tsx
@@ -1,0 +1,205 @@
+import { Icon } from '@iconify/react';
+import { DialogContentText } from '@material-ui/core';
+import Button from '@material-ui/core/Button';
+import Dialog, { DialogProps } from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import Fab from '@material-ui/core/Fab';
+import Grid from '@material-ui/core/Grid';
+import IconButton from '@material-ui/core/IconButton';
+import MuiInput from '@material-ui/core/Input';
+import { makeStyles, styled } from '@material-ui/core/styles';
+import Tooltip from '@material-ui/core/Tooltip';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useDispatch } from 'react-redux';
+import { useLocation } from 'react-router-dom';
+import { KubeObject } from '../../../lib/k8s/cluster';
+import { CallbackActionOptions, clusterAction } from '../../../redux/actions/actions';
+
+interface ScaleButtonProps {
+  item: KubeObject;
+  options?: CallbackActionOptions;
+}
+
+export default function ScaleButton(props: ScaleButtonProps) {
+  const dispatch = useDispatch();
+  const { item, options = {} } = props;
+  const [openDialog, setOpenDialog] = React.useState(false);
+  const [visible, setVisible] = React.useState(false);
+  const location = useLocation();
+  const { t } = useTranslation('resource');
+
+  async function updateFunc(numReplicas: number) {
+    try {
+      await item.scale(numReplicas);
+    } catch (err) {
+      throw err;
+    }
+  }
+
+  const applyFunc = React.useCallback(updateFunc, [item]);
+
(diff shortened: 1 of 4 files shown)


frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  if (!visible || !['Deployment', 'StatefulSet', 'ReplicaSet'].includes(item.kind)) {
+    return null;
+  }
+
+  return (
+    <React.Fragment>
+      <Tooltip title={t('frequent|Scale') as string}>
+        <IconButton aria-label={t('frequent|scale')} onClick={() => setOpenDialog(true)}>
+          <Icon icon="mdi:overscan" />
```
illume, line 93: Not sure this is a good icon. Looks close to a fullscreen button too me.

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+  onClose: () => void;
+  errorMessage?: string;
+}
+
+const Input = styled(MuiInput)({
+  '& input[type=number]': {
+    MozAppearance: 'textfield',
+    textAlign: 'center',
+  },
+  '& input::-webkit-outer-spin-button, & input::-webkit-inner-spin-button': {
```
illume, line 113: What is this for, and why webkit only?

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const [numReplicas, setNumReplicas] = React.useState<number>(getNumReplicas());
+  const { t } = useTranslation(['frequent', 'resource']);
+  const classes = useScaleDialogStyle();
+  const theme = useTheme();
+  const desiredNumReplicasLabel = 'desired-number-replicas-label';
+  const numReplicasForWarning = 100;
+
+  function getNumReplicas() {
+    if (!resource?.spec) {
+      return -1;
```
illume, line 147: Why is -1 used here?

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+            </DialogContentText>
+          </Grid>
+          <Grid item container alignItems="center" spacing={1}>
+            <Grid item sm="auto" xs={12}>
+              <DialogContentText id={desiredNumReplicasLabel}>
+                {t('resource|Desired number of replicas:')}
+              </DialogContentText>
+            </Grid>
+            <Grid item spacing={2} sm="auto" className={classes.replicasSwitcher}>
+              <Fab
```
illume, line 172: Could use an aria-label.

illume (changes requested):
Thanks. Looking good👍

Left some questions/comments. Would also like to see a story for this component if possible?

frontend/src/components/common/Resource/ScaleButton.tsx
joaquimrocha: Hmmm, it would indicate something weird happened (if we end up applying this to an unwanted resource).
I will change the label according to this value in the UI.

frontend/src/components/common/Resource/ScaleButton.tsx
joaquimrocha: This hides the up/down arrows from the number type input. It works fine in Firefox too.

joaquimrocha:
@illume  BTW, for the i18n, there are missing translations to be added in main, so we can just merge this and then add them all when this is ready.

frontend/src/components/common/Resource/ScaleButton.tsx
joaquimrocha: Changed to content-copy

ashu8912 (approved):
Tested this PR on local LGTM!!!

PR #578, opened 2022-07-05
WIP: Improve pod matcher

joaquimrocha:
This is based on a PR by @fondoger . As originally written:

Pod list of a deployment is determined by its selector.
Example Deployment selector:
```
selector:
  matchLabels:
    component: redis
  matchExpressions:
    - {key: tier, operator: In, values: [cache]}
    - {key: environment, operator: NotIn, values: [dev]}
```

We can't simply fetch all pods and filter by labels only.

## Update to this PR

This PR has a commit to fix an issue in the original one (it was not passing all the query options when using `useList`).
We do not want to merge this right away since it's a bit complex and so we should make sure we have tests in place for verifying the conversion of selectors to queries, or replace it with a lib that does this for us.


joaquimrocha:
This functionality is already in, so closing the PR.

PR #580, opened 2022-07-06
Export resource chart and lib

joaquimrocha:
# Export ResourceCircularChart and lib

This PR is about exporting the ResourceCircularChart, used to create the cluster's overview charts, and also to export the Headlamp's lib/* modules.

- [X] Export ResourceCircularCart
- [X] Export units and auth modules

## Testing done

In a plugin (with the new version of headlamp-plugin), add:
```typescript
import { auth, units } from '@kinvolk/headlamp-plugin/lib/Utils';

console.log('VERIFY', units, auth)
```

Look for the logs to see if they show valid modules.

PS. We need to really integrate testing in plugins imports.

### DIFF
--- a/frontend/src/components/cluster/Charts.tsx
+++ b/frontend/src/components/cluster/Charts.tsx
@@ -1,9 +1,8 @@
 import '../../i18n/config';
 import { useTheme } from '@material-ui/core/styles';
 import _ from 'lodash';
-import React from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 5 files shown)


illume (approved):
Great.

I feel using this in a storybook or a plugin example would be instructive.

joaquimrocha:
> Great.
> 
> I feel using this in a storybook or a plugin example would be instructive.

Yes. I have opened #586 to track that. It can be a good first issue too I think.

PR #587, opened 2022-07-12
Refresh token fix

yolossn:


joaquimrocha (approved):
I think we should add a getter function in util for the cluster info and then use it for getting the auth_type here, but it's fine to merge this and do that later.

PR #590, opened 2022-07-12
frontend: Fix image text wrap for deployment detail view

ashu8912:


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -574,7 +574,7 @@ export function ContainerInfo(props: ContainerInfoProps) {
         name: t('Image'),
         value: (
           <>
-            <Typography>{container.image}</Typography>
+            <Typography style={{ overflowWrap: 'anywhere' }}>{container.image}</Typography>
```
joaquimrocha, line 577: It'd be preferable if we add this to the classes and use a className here, since we have already the classes being declared within this component.

ashu8912: updated by adding a new class preventOverflow

joaquimrocha (approved):

PR #601, opened 2022-07-15
app: Add chocolatey spec + install script

joaquimrocha:
This patch allows us to build a chocolatey package for Headlamp.


illume (approved):
👍

PR #603, opened 2022-07-18
frontend: Make the namespaces filter values occupy just one line

### LINKED ISSUE #593: [RFE] Namespaces filter UI/UX improvement
## Current situation

At the moment namespaces filter user experience is quite uncomfortable.
The main problem is multi-namespaces handling. If you pick more than one namespace, the content below e.g. tables are being pushed from top to bottom and if you pick more namespaces it just continues pushing the content to bottom. Also, select option styling expects namespaces to be short, but if they are 30+ chars, select option takes 2-3 lines to show that namespace name.

## Impact

It makes the namespaces filter interface uncomfortable

## Ideal future situation

When user picks whatever quantity of items the content below stays where it is, select options styling lets user to interact with long namespaces in a comfortable way

## Implementation options

In my opinion, the select should show the picked item name if it is only one, if more, it should just show the amount of names.
Maybe select should be a little wider and select's options should have less font-size in order to look good with long names

## Additional information

Attached some screenshots that shows the problem and possible ways it could be solved

Screenshot 2022-07-13 121219
Screenshot 2022-07-13 121304
Screenshot 2022-07-13 121426
Screenshot 2022-07-13 121521
Screenshot 2022-07-13 121554
Screenshot 2022-07-13 121954

joaquimrocha:
When multiple namespaces were selected, the input could show several
lines and that made the header for the contents very large vertically,
which didn't look good.

This patch changes how the namespaces are shown, so that only up to
a certain number of chars are visible for the namespace name and after
that it shows how many other namespaces are selected.

Here are examples:
Long name namespace selected + another one:
Screenshot of a long name namespace selected + another one

A single and short namespace selected:
Screenshot of a single and short namespace selected

If two namespaces' names fit the char limit (12 chars), then they may be displayed with the `+N` part.

fixes #593

### DIFF
--- a/frontend/src/components/common/NamespacesAutocomplete.tsx
+++ b/frontend/src/components/common/NamespacesAutocomplete.tsx
@@ -26,7 +26,8 @@ export function PureNamespacesAutocomplete({
   filter,
 }: PureNamespacesAutocompleteProps) {
   const theme = useTheme();
-  const { t } = useTranslation('glossary');
+  const { t } = useTranslation(['glossary', 'frequent']);
+  const maxNamespacesChars = 12;
 
   return (
     <Autocomplete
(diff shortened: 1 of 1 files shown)


ashu8912 (approved):
Tested the PR, LGTM!!!!

PR #604, opened 2022-07-18
shorter resource age

### LINKED ISSUE #263: [RFE] Show timezone information for every relevant date/time
## Current situation

We show dates for resources/events (creation, etc.) and those are adapted to the user's current timezone but that's rather implicit.

## Impact

Not being sure about which timezone a date/time is may lead to confusion/doubts when managing services with Headlamp.

## Ideal future situation

We have a way to show users what timezone a date/time is in.

## **Implementation options

We decorate dates with a dotted underline, which on hover will display a tooltip with a full version of the date and time, including timezone info

or, we show a watch icon next to the date, which on hover does what's described above.

### LINKED ISSUE #576: [RFE] Use a shorter version of time/age labels
## Current situation

We use "3 months" for the age column if something has happened 3 months ago. Since that column and values are an abbreviation of a full date/time anyway, we should use `3mo` for the previous example, which allows more space to be used by the remaining columns and even to add more columns.

## Ideal future situation / Implementation options

Dates in the age column are shown with what's called the "mini" format in the javascript-time-ago package.
On hover, those dates should show the full date/time in a tooltip.

**Important:** I have already part of this implementation done. So please reach out if you pick this up.

joaquimrocha:
This PR greatly simplifies show we can consistently show a table of resources of the same kind, by not having to list the resource before. Or by not having to repeat the logic for common columns like age, name, namespace, etc.

How to test:
- [ ] Go to all list views: check that there's a much shorter label now
- [ ] Hover on the age label: verify is shows the full date with a timezone now
- [ ] Go to all list views, verify that the links worked as usual
- [ ] Go to the Node view details: check that the uptime is the same as before
- [ ] Go to any details view: check that the creation date has the same format as before but now shows the timezone

fixes #576
fixes #263


joaquimrocha:
@illume , I was running into issues when the UI was refreshed by changes (not by hard refreshed, which had no issues) but I hope it had to do with these hooks being used conditionally:
https://github.com/kinvolk/headlamp/pull/604/commits/da16549cccf591cf6989d8040e3bfc52ce589a28

frontend/src/components/cluster/Chooser.tsx
```diff
@@ -38,8 +38,8 @@ export interface ClusterTitleProps {
 }
 
 export function ClusterTitle(props: ClusterTitleProps) {
-  const cluster = props.cluster || useCluster();
-  const clusters = props.clusters || useClustersConf();
+  const cluster = props.cluster;
```
illume, line 41: The useCluster/useClusterConf could be called, and then later:

```ts
const clusterUse = useCluster();
const clustersUse = useClustersConf()
...

const cluster = props.cluster || clusterUse;
const clusters = props.clusters || clustersUse;
```

This way the props could still be used to override.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -299,17 +299,20 @@ export default function SimpleTable(props: SimpleTableProps) {
...
-                        &nbsp;
-                      </React.Fragment>
-                    )}
-                    {'datum' in col ? row[col.datum] : col.getter(row)}
-                  </TableCell>
-                ))}
+                {columns.map((col, i) => {
+                  const { cellProps = {} } = col;
+                  return (
+                    <TableCell key={`cell_${i}`} {...cellProps}>
```
illume, line 305: ps. Using array index for the key is problematic. It causes problems when the data changes. (I know this problem was here before this change, just mentioning it)

frontend/src/components/crd/List.tsx
```diff
@@ -1,25 +1,19 @@
-import React from 'react';
 import { useTranslation } from 'react-i18next';
 import CRD from '../../lib/k8s/crd';
-import { timeAgo, useFilterFunc } from '../../lib/util';
 import { Link } from '../common';
+import ResourceTable from '../common/Resource/Table';
```
illume, line 4: Maybe the component file should be ResourceTable.

frontend/src/components/pod/__snapshots__/Pod.stories.storyshot
```diff
@@ -212,8 +213,15 @@ exports[`Storyshots pods/PodListRenderer Pods 1`] = `
...
               <td
                 class="MuiTableCell-root MuiTableCell-body"
+                style="text-align: right;"
               >
-                just now
+                <p
+                  class="MuiTypography-root makeStyles-noWrap MuiTypography-body1"
+                  title="4/25/2022, 12:00:00 AM UTC"
+                >
+                  3mo
```
illume, line 222: I wonder if this will be "3mo" next month? I don't think this test will be stable.

ps. TIL: I wasn't entirely sure "mo" was widely understood as an abbreviation for month. After doing some googling... a number of ranking pages do seem to put it first.

illume:
I really like that ResourceTable has replaced so much code. Adding the timezone and shortening the resource age are nice touches too.

frontend/src/components/cluster/Chooser.tsx
joaquimrocha: I think the way to do this is to use just the props and then have two higher level components which take either props or nothing. That way we can have the hooks or props properly. (This is what I did for the resource table, BTW)

At this point, if it's not being overridden, I prefer not to have it available.

frontend/src/components/common/SimpleTable.tsx
joaquimrocha: I am not aware of a better way to assign indexes in this case. The name for cells is not unique, so this sounds okay according with https://reactjs.org/docs/lists-and-keys.html .

frontend/src/components/pod/__snapshots__/Pod.stories.storyshot
joaquimrocha: Good point. I will fix this in a different PR.

joaquimrocha: It's fixed in #630

ashu8912 (approved):
Tested. LGTM!! 
Lets wait for @illume final inputs on the code.

frontend/src/components/cluster/Chooser.tsx
illume: Yeah, just taking props is better.

frontend/src/components/common/SimpleTable.tsx
illume: Yeah, should be fine.

illume:
Where does it get the timezone from in the tooltip? (browser?, where `headlamp_server` is running?)
(it's showing CEST for me like the backend)

Not sure if some folks still like seeing UTC  on all the server related times these days. When discussing server related date-times in teams that's best IMHO.

illume:
Seems there is more margin around the table now?
image


Previously there was less space outside the table:

image


Ah, the branch needs to be rebased on main... can you please do a rebase?

illume:
I compared views (and their tables) between the old version(main), and this branch... and they seem to be correct. ✔️
(We don't really have storybook/tests for the SimpleTable, and in most places it is used.)

Clicking on Namespaces link doesn't work for me. It doesn't go to the details view. 🛑

joaquimrocha:
> Where does it get the timezone from in the tooltip? (browser?, where `headlamp_server` is running?) (it's showing CEST for me like the backend)

It uses the localeDateTime function we have (which uses the toLocaleString). So the tz comes from the user's browser IIUC.

> Not sure if some folks still like seeing UTC on all the server related times these days. When discussing server related date-times in teams that's best IMHO.

It's a good question. I'd argue that having the local timezone is better but for working cross teams, it's true that using UTC makes sense.

Maybe we should add the local VS UTC as a user preference? (different PR though)

joaquimrocha:
@illume , as for not getting to the Namespace details, maybe it had to do with the issue we had in `main` last week. I have rebased now and I can access the namespaces. Can you double check things work for you?

illume (approved):
Great, thanks! 🎉🎈

PR #605, opened 2022-07-19
Wait for plugins to load before showing views

### LINKED ISSUE #454: [RFE] Ensure routes are rendered only after the plugins are loaded
## Current situation

Plugins may define or override routes. If they override an existing route but the user navigates to that route before the plugins are loaded, they may still see the default route components being rendered until the plugin is loaded and replaces it.

## Impact

It prevents plugins from providing a stable UX for the user.

## Ideal future situation

We only render routes after all plugins are rendered, which allows for the plugins to fully replace routes.

illume:
Fixes #454

The plugins can change a route, so waiting for the plugin to load stops flash-of-route-that-has-changed problem.

## How to use

Change a route with a plugin, and then the default view should not be shown before the plugin is loaded.

## Testing done

Use the following plugin that overrides "namespaces" and go to http://localhost:3000/c/minikube/namespaces

There should be "Namespaces is overridden." text.

```tsx
import { registerRoute } from '@kinvolk/headlamp-plugin/lib';
import { SectionBox } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
import Typography from '@material-ui/core/Typography';

// A component to go along with the URL path.
registerRoute({
  path: '/namespaces',
  sidebar: 'namespaces',
  name: 'namespaces',
  exact: true,
  component: () => (
    <SectionBox title="Feedback" textAlign="center" paddingTop={2}>
      <Typography>Namespaces is overridden.</Typography>
    </SectionBox>
  ),
});
```

Here's another one to show the 404 page doesn't flash. Go to http://localhost:3000/woof
```tsx
import { registerRoute } from '@kinvolk/headlamp-plugin/lib';
import { SectionBox } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
import Typography from '@material-ui/core/Typography';

// go to http://localhost:3000/woof
registerRoute({
  path: '/woof',
  exact: true,
  sidebar: null,
  noCluster: true,
  noAuthRequired: true,
  component: () => (
    <SectionBox title="Feedback" textAlign="center" paddingTop={2}>
      <Typography>There should be no 404 page flashed.</Typography>
    </SectionBox>
  ),
});
```

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -10,6 +10,6 @@ test('renders without crashing', async () => {
   );
 
   await waitFor(() => {
-    expect(getByText(/Wait while fetching clusters/i)).toBeInTheDocument();
+    expect(getByText(/Skip to main content/i)).toBeInTheDocument();
   });
 });
--- a/frontend/src/components/App/Layout.tsx
+++ b/frontend/src/components/App/Layout.tsx
@@ -3,6 +3,7 @@ import Container from '@material-ui/core/Container';
 import CssBaseline from '@material-ui/core/CssBaseline';
 import Link from '@material-ui/core/Link';
 import { makeStyles } from '@material-ui/core/styles';
+import { useTypedSelector } from '../../redux/reducers/reducers';
 import ActionsNotifier from '../common/ActionsNotifier';
 import AlertNotification from '../common/AlertNotification';
 import Sidebar, { NavigationTabs } from '../Sidebar';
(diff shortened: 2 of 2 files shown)


ashu8912 (approved):
LGTM!! too just a small comment on how do we load the RouteSwitcher, like till the time we check if the plugins are loaded we should show some sort of a Loader something like getting your view ready or preparing things, but yeah this can be done in a different PR too for now what if we have is already good enough

PR #606, opened 2022-07-19
Add error handling/boundaries for plugins

illume:
# Add error handling/boundaries for plugins

Stopped plugins with errors crashing Headlamp:
- On plugin load (script syntax error, or other error)
- When components are rendering

Extracted plugin loading and executing into a Plugins component (moving the rest of it there will come in a future PR).

Wrapped `<ErrorBoundary>` around where plugins can insert components into the app.

Fixes https://github.com/kinvolk/headlamp/issues/597

## How to Test

Introduce some errors into the example plugins.

- [x] Add an error into a plugin component (`throw 'error';`) and Headlamp should still load.
- [x] Add an error into a compiled plugin, and Headlamp should still load.

```tsx
import { registerAppBarAction } from '@kinvolk/headlamp-plugin/lib';

registerAppBarAction(() => {
  throw 'meow';
});
```


frontend/src/components/common/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+export default class ErrorBoundary extends Component<ErrorBoundaryProps, State> {
+  state: State = { error: null };
+  static getDerivedStateFromError(error: Error | null) {
+    return { error };
+  }
+
+  //   componentDidCatch(error, errorInfo) {
+  //     logError(error, errorInfo);
+  //   }
```
joaquimrocha, line 44: Maybe unintendedly left?

frontend/src/redux/actions/actions.tsx
```diff
@@ -77,6 +77,7 @@ export interface Action {
 type SidebarType = UIState['sidebar'];
 
 export type HeaderActionType = ((...args: any[]) => JSX.Element | null) | null | ReactElement;
+// export type HeaderActionType = React.FunctionComponent | null | ReactElement;
```
joaquimrocha, line 80: Is this intended?

frontend/src/plugin/Plugins.tsx
```diff
@@ -0,0 +1,46 @@
...
+        (function (str: string) {
+          try {
+            const result = eval(str);
+            return result;
+          } catch (e) {
+            console.error(e);
+          }
+        }.call({}, source));
+      });
+      await initializePlugins();
```
joaquimrocha, line 44: We're calling `store.dispatch({ type: UI_RESET_PLUGIN_VIEWS });` within this method which runs after loading the external plugins AFAICS. Won't this be a problem now that the plugins perform code right when they're loaded (before it worked on the assumption we'd load the code but only initialize the plugins after resetting views, etc.)?

frontend/src/plugin/Plugins.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      sources.forEach(source => {
+        (function (str: string) {
+          try {
+            const result = eval(str);
+            return result;
+          } catch (e) {
+            console.error(e);
+          }
+        }.call({}, source));
```
joaquimrocha, line 42: I'm surprise this works fine like this. Before we were injecting the scripts into the document. Not sure how things differ or what's best, just curious.

frontend/src/plugin/Plugins.tsx
```diff
@@ -0,0 +1,46 @@
...
+        )
+      );
+
+      sources.forEach(source => {
+        (function (str: string) {
+          try {
+            const result = eval(str);
+            return result;
+          } catch (e) {
+            console.error(e);
```
joaquimrocha, line 40: Maybe prefix the error so we have that context when we see it?

frontend/src/plugin/Plugins.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const fetchData = async () => {
+      const pluginsScriptPaths = (await fetch(`${helpers.getAppUrl()}plugins/list`).then(resp =>
+        resp.json()
+      )) as string[];
+
+      const sources = await Promise.all(
+        pluginsScriptPaths.map(path =>
+          fetch(`${helpers.getAppUrl()}${path}`).then(resp => resp.text())
+        )
+      );
```
joaquimrocha, line 30: Nice! I guess we get some speed improvements by having these (sort of) parallel.

frontend/src/components/common/index.test.ts
```diff
@@ -12,6 +12,7 @@ const avoidCheck = [
   'AlertNotification',
   'ConfirmButton',
   'NamespacesAutocomplete',
+  'ErrorBoundary',
```
joaquimrocha, line 15: 👍

joaquimrocha (changes requested):
This is pretty cool! I left a couple of comments.
As a second iteration, maybe we could have a mode where we use a visible and adequate fallback, so users can see at least what plugin has generated the errors and where.

Also, maybe on the plugin defined routes' components we should always show a fallback component with an error message?

frontend/src/plugin/Plugins.tsx
illume: Thanks. I moved this so the initialization is done first before loading the plugins.

illume (commented):
> Hrmm. Github changed so my replies to your comments are now done with a review, and all my comments were pending and not sent right away. Kind of strange... but ok github.

I've addressed all your comments but left them unresolved so you can see.

frontend/src/components/common/ErrorBoundary/ErrorBoundary.tsx
illume: Thanks. Removed it.

(Was leaving it there for a later, but it's easy to find in the docs so removed it.)

joaquimrocha (approved):

PR #608, opened 2022-07-20
backend: Add proxy endpoint that forwards all request to provided URL

ashu8912:


backend/cmd/headlamp.go
```diff
@@ -275,6 +275,23 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	config.handleClusterRequests(r)
 
+	r.HandleFunc("/customapiproxy", func(w http.ResponseWriter, r *http.Request) {
```
joaquimrocha, line 278: Instead of custom (here and in the rest of the cases), I think it makes more sense to use `external`.
And we still need to have this turned off by default, pending the new server option we have discussed.

joaquimrocha (changes requested):
Left a couple of comments.

backend/cmd/headlamp.go
```diff
@@ -275,6 +275,23 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+		url, err := url.Parse(r.Header.Get("CustomAPI"))
+		if err != nil {
+			log.Fatal("Failed to get URL from server", err)
+		}
+
+		proxy := httputil.NewSingleHostReverseProxy(url)
+		r.Host = url.Host
+		r.URL.Host = url.Host
+		r.URL.Path = ""
```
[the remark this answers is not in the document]
ashu8912: yup this is not required altogether coz we already add the request URI here

backend/cmd/headlamp.go
```diff
@@ -275,6 +276,30 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	config.handleClusterRequests(r)
 
+	r.HandleFunc("/externalapiproxy", func(w http.ResponseWriter, r *http.Request) {
```
yolossn, line 279:
```suggestion
	r.HandleFunc("/externalproxy", func(w http.ResponseWriter, r *http.Request) {
```

backend/cmd/headlamp.go
```diff
@@ -275,6 +276,30 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+		}
+		isURLContainedInProxyURLs := false
+		for _, proxyURL := range config.proxyURLs {
+			if strings.HasPrefix(url.Host, proxyURL) {
+				isURLContainedInProxyURLs = true
+			}
+		}
+		if !isURLContainedInProxyURLs {
+			http.Error(w, "no allowed proxy url match, request denied ", http.StatusBadRequest)
+		}
```
yolossn, line 292: From our previous discussion, this was supposed to be a wildcard match instead of a Host match.

FYI, this is how we check for valid redirect URLs in Nebraska.

https://github.com/kinvolk/nebraska/blob/main/backend/pkg/auth/oidc.go#L248-L258

backend/cmd/headlamp.go
```diff
@@ -275,6 +276,30 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	config.handleClusterRequests(r)
 
+	r.HandleFunc("/externalapiproxy", func(w http.ResponseWriter, r *http.Request) {
+		url, err := url.Parse(r.Header.Get("ExternalAPI"))
```
yolossn, line 280:
```suggestion
		url, err := url.Parse(r.Header.Get("proxy-to"))
```

backend/cmd/headlamp.go
yolossn, line 292: From our previous discussion, I supposed this will be a wildcard match instead of a Host match.

FYI, this is how we check for valid redirect URLs in Nebraska.

https://github.com/kinvolk/nebraska/blob/main/backend/pkg/auth/oidc.go#L248-L258

yolossn, line 292: From our previous discussion, I thought this will be a wildcard match instead of a Host match.

FYI, this is how we check for valid redirect URLs in Nebraska.

https://github.com/kinvolk/nebraska/blob/main/backend/pkg/auth/oidc.go#L248-L258

backend/cmd/headlamp.go
yolossn: I made a similar suggestion, I think "/proxy" and the query param "proxy-to" makes more sense.

backend/cmd/headlamp.go
joaquimrocha: It's nice that we have several, but they should also be wildcard matches.

backend/cmd/headlamp.go
ashu8912: Updated, thanks @yolossn

backend/cmd/headlamp.go
ashu8912: update to externalproxy

backend/cmd/headlamp.go
ashu8912: updated

yolossn (approved):
LGTM

backend/cmd/headlamp.go
```diff
@@ -185,7 +187,7 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
 	}
 }
 
-// nolint:gocognit,funlen
+// nolint:gocognit,funlen,gocyclo
```
joaquimrocha, line 190: We should make sure things are readable and these checks exist to help us figure that out. But we can do this as part of the backend improvements that @yolossn will work on.

joaquimrocha (approved):

PR #618, opened 2022-07-21
Fix ActionButton color and width problems

illume:
This fixes the Create resource button being invisible.

## How to use

Open Headlamp see the Create + icon is visible.

## Testing done

`npm run storybook` also shows the ActionButtons with different colors / widths.


joaquimrocha (approved):
And... It's back! Thanks. LGTM.

PR #512, opened 2022-05-18
Delay sending appVersion to fix update checking not triggering

fondoger:
…gered

# [Title: describe the change in one sentence]

[ describe the change in 1 - 3 paragraphs ]

## How to use

[ describe what reviewers need to do in order to validate this PR ]

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

PR #619, opened 2022-07-21
Fix new version checks

joaquimrocha:
Today I ran into the issue as described in #512 . Indeed the appVersion is sent only when the DOM is ready, but the component that acts on the received appVersion was not ready yet.

To fix this, this PR changes how the appVersion is sent. Now it's sent only when requested, guaranteeing that the component is in fact ready.


illume (approved):
👍

PR #620, opened 2022-07-21
Allow to build without checking for updates

joaquimrocha:
This is a proposal for adding a way to build Headlamp's app without having it check for updates.

@illume , I am open to a better alternative to having the options in app/package.json.

Testing:
- [ ] Run `make run-backend` and `make run-frontend`, then go to the app folder and run `HEADLAMP_APP_VERSION=0.1.0 npm run dev-only-app`: It should check for updates and show the small popup notification on the lower right saying a new version is available
- [ ] Change the `checkForUpdates` setting to `false` in app/package.json and run the same command as above again: the updat check shouldn't happen this time

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -145,6 +145,7 @@
     "@types/semver": "^7.3.8",
     "copyfiles": "^2.4.1",
     "cross-env": "^7.0.3",
+    "dotenv": "^16.0.1",
     "electron-log": "^4.4.1",
     "find-process": "^1.4.7",
     "i18next": "^20.6.1",
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,4 +1,5 @@
 import { ChildProcessWithoutNullStreams, execSync, spawn } from 'child_process';
+import dotenv from 'dotenv';
 import { app, BrowserWindow, dialog, ipcMain, Menu, MenuItem, screen, shell } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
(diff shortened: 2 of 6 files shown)


illume:
TLDR; Yeah, this should work to provide a build time config for this check-for-updates feature.

With options in package.json, I was thinking about merging. If someone was using a diff, they might need to regularly update the diff every time we change the package.json by a few lines. But if they're using git(and they are), then it should be no problems rebasing.

An alternative way to use environment variables could be to look at the environment variable at build time. Or see if an options.json is there and can be require()'d. This way, the project doing the build would have their options.json, or .env file there with their configurations.

illume (approved):
👍 thanks.

illume:
I don't think our app/ supports .env files. It's only supported in CRA I think: https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env

- https://github.com/facebook/create-react-app/blob/f0a837c1f07ebd963ddbba2c2937d04fc1b79d40/packages/react-scripts/config/env.js#L36
- https://github.com/motdotla/dotenv-expand
- https://github.com/motdotla/dotenv

Adding support for that would allow people to define build time config in a app/.env file, and then not have to modify the package.json.

joaquimrocha:
@illume , I took the approach in the new WIP commit now, but after running the built app, I think it just looks for the .env file in the current directory (didn't also work after I copied the .env file to the app files). So we may need to specify the location.

joaquimrocha:
I have changed it so now the .env is shipped in the resources folder + loaded with that custom path.

app/package.json
```diff
@@ -18,6 +18,9 @@
     "test": "jest",
     "copy-plugins": "npx --no-install rimraf build/.plugins && mkdirp build/.plugins && copyfiles ../.plugins build/.plugins"
   },
+  "headlampOptions": {
```
illume, line 21: This isn't needed now.

illume:
There's that config from the package.json can be removed, otherwise LGTM.

illume:
Docs somewhere for HEADLAMP_CHECK_FOR_UPDATES?

joaquimrocha:
I have added the docs in #628.

PR #621, opened 2022-07-21
app: Update electron and electron-builder

joaquimrocha:
Apparently versions >16 of electron may end up showing artifacts due
to GPU issues in WSL2 but it works fine outside of it, and we need to
update dependencies.


illume (approved):
👍

PR #625, opened 2022-07-22
Fix logo getting trimmed

### LINKED ISSUE #622: Headlamp's logo gets trimmed in cluster chooser
## Description

This is happening frequently:
Logo cut off in cluster chooser

joaquimrocha:
This fixes #622 .
The problem was that the logo component was added to an SvgIcon which is something that is intended to be used with icons, not with generic images. Adding the component directly, fixed it.

Not sure if there are any a11y implications, but that's not an important part of the dialog for understand what it is, so I guess it's fine.

I have also added a patch to drop the typography from DialogTitle components when the respective attribute is present.


illume (approved):
👍

PR #626, opened 2022-07-22
Fix exporting circular chart

joaquimrocha:
It was being exported only by default from its module, so it was not getting name-exported by common/index.tsx.


illume (approved):
👍 thanks

PR #627, opened 2022-07-22
Add checks for exported common components

joaquimrocha:
This new test allows us to check whether we have modules in `common/` that cannot be imported from `common/index.tsx`. It may require more complexity for adding whatever corner cases I may have forgotten, but it's a good start IMO.

By running it I have noticed that Charts.tsx was not being exported and started exporting it.


illume (approved):
👍 🎉 love the test :)

Issue #546, opened 2022-06-08
[RFE] Only scroll log viewer down automatically if the view is at the bottom of it


fondoger:
https://github.com/kinvolk/headlamp/blob/620e79d80c4330a0914ebbb6238b23e27b192efd/frontend/src/components/common/LogViewer.tsx#L62

```ts
  React.useEffect(() => {
    // @todo: Only scroll down automatically if the view hasn't been scrolled up by the user yet.
    if (logsBottomRef && logsBottomRef.current) {
      logsBottomRef.current.scrollIntoView();
    }
  }, [logs]);
```

**Update (by Joaquim):** This is about:
- [ ] By default the log viewer shows the logs as they happen, scrolling to the bottom automatically like a terminal view
- [ ] If the user scrolls the view up, even the tiniest bit, it should remain showing that portion of the log even if more logs arrive
- [ ] If the user scrolls down to the bottom again, then new updates will keep the view at the bottom (i.e. it will show the latest contents)

ashu8912:
This is fixed in main

PR #629, opened 2022-07-25
frontend: Don't scroll log viewer to the bottom

ashu8912:
We want to let the log view scroll to remain at the current viewing
point only and not scroll to the bottom of the log.

## Testing done
Go to the log viewer in pod detail view and you will see the logs show up without getting scrolled to the bottom

### DIFF
--- a/frontend/src/components/common/LogViewer.tsx
+++ b/frontend/src/components/common/LogViewer.tsx
@@ -49,6 +49,8 @@ export function LogViewer(props: LogViewerProps) {
   const classes = useStyle({ isFullScreen });
   const logsBottomRef = React.useRef<HTMLDivElement>(null);
   const { t } = useTranslation('frequent');
+  const [isScrolledUp, setIsScrolledUp] = React.useState(false);
+  const scrollMargin = 15;
 
   function downloadLog() {
     const element = document.createElement('a');
@@ -61,11 +63,32 @@ export function LogViewer(props: LogViewerProps) {
   }
 
   React.useEffect(() => {
-    // @todo: Only scroll down automatically if the view hasn't been scrolled up by the user yet.
-    if (logsBottomRef && logsBottomRef.current) {
+    if (logsBottomRef?.current && !isScrolledUp) {
       logsBottomRef.current.scrollIntoView();
     }
   }, [logs]);
+
+  function handleScroll(event: any) {
+    if (logsBottomRef?.current) {
+      /* 
+        - By default the log viewer shows the logs as they happen, scrolling to the bottom automatically like a terminal view
+        - If the user scrolls the view up, even the tiniest bit, it should remain showing that portion of the log even if more logs arrive
+        - If the user scrolls down to the bottom again, then new updates will keep the view at the bottom (i.e. it will show the latest contents)
+      */
+      const wrapperPosition = logsBottomRef.current.getBoundingClientRect().bottom;
+      const scrollPosition = event.target.getBoundingClientRect().bottom + scrollMargin;
+      /*  compare if the terminal wrapper bottom is in the scroll limit then the user is at 
+          the bottom of the screen
+      */
+      const scrollAtBottom = Math.trunc(wrapperPosition) < Math.trunc(scrollPosition);
+      if (scrollAtBottom) {
+        setIsScrolledUp(false);
+        return;
+      }
+    }
+
+    setIsScrolledUp(true);
+  }
 
   return (
     <Dialog
@@ -92,7 +115,7 @@ export function LogViewer(props: LogViewerProps) {
             />
           </Grid>
         </Grid>
-        <Box className={classes.terminal}>
+        <Box className={classes.terminal} onScroll={handleScroll}>
           <pre>
             {logs.map((item, i) => (
               <Ansi className={classes.terminalCode} key={i} linkify={false}>


joaquimrocha:
@ashu8912 , this is not what #546 describes. We do want the view to be scrolled to the bottom, but only if it's already at the bottom.
The problem we want to solve though, is that if the user gets all the logs, then is scrolling up looking for something and suddenly more logs arrive, we shouldn't force the view to scroll to the bottom at that moment, because we'll disrupt the user's reading.
That's why we want to only scroll to the bottom at the beginning + if the view is already scrolled to the bottom.

frontend/src/components/common/LogViewer.tsx
```diff
@@ -59,14 +61,31 @@ export function LogViewer(props: LogViewerProps) {
     document.body.appendChild(element);
     element.click();
   }
-
```
joaquimrocha, line 62: Do not remove this line.

joaquimrocha (changes requested):
Tested and it works fine. Left a couple of comments in the code which I hope make things more readable.

frontend/src/components/common/LogViewer.tsx
ashu8912: fixed

frontend/src/components/common/LogViewer.tsx
```diff
@@ -59,14 +61,31 @@ export function LogViewer(props: LogViewerProps) {
...
   React.useEffect(() => {
     // @todo: Only scroll down automatically if the view hasn't been scrolled up by the user yet.
     if (logsBottomRef && logsBottomRef.current) {
-      logsBottomRef.current.scrollIntoView();
+      if (!isScrolledUp) {
```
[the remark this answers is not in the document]
ashu8912: updated

frontend/src/components/common/LogViewer.tsx
```diff
@@ -59,14 +61,31 @@ export function LogViewer(props: LogViewerProps) {
...
     }
   }, [logs]);
 
+  function handleScroll(event: any) {
+    if (logsBottomRef && logsBottomRef.current) {
+      const wrapperPosition = logsBottomRef.current.getBoundingClientRect().bottom;
+      const scrollPosition = event.target.getBoundingClientRect().bottom + scrollMargin;
+      /*  the div we have placed at the bottom of the log viewer 
+         we allow a difference of scrollMargin if it crosses it we know 
+        that we are no longer at the bottom of the log viewer */
```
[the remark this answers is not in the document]
ashu8912: fixed and updated the comment

frontend/src/components/common/LogViewer.tsx
```diff
@@ -59,14 +61,31 @@ export function LogViewer(props: LogViewerProps) {
...
     if (logsBottomRef && logsBottomRef.current) {
-      logsBottomRef.current.scrollIntoView();
+      if (!isScrolledUp) {
+        logsBottomRef.current.scrollIntoView();
+      }
     }
   }, [logs]);
 
+  function handleScroll(event: any) {
+    if (logsBottomRef && logsBottomRef.current) {
```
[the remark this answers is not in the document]
ashu8912: updated

frontend/src/components/common/LogViewer.tsx
```diff
@@ -61,12 +63,27 @@ export function LogViewer(props: LogViewerProps) {
...
   React.useEffect(() => {
-    // @todo: Only scroll down automatically if the view hasn't been scrolled up by the user yet.
-    if (logsBottomRef && logsBottomRef.current) {
+    if (logsBottomRef?.current && !isScrolledUp) {
       logsBottomRef.current.scrollIntoView();
     }
   }, [logs]);
 
+  function handleScroll(event: any) {
+    if (logsBottomRef?.current) {
```
illume, line 72: Reading this, I'm confused what this is supposed to do...

How about adding the description from the issue?
```
- By default the log viewer shows the logs as they happen, scrolling to the bottom automatically like a terminal view
- If the user scrolls the view up, even the tiniest bit, it should remain showing that portion of the log even if more logs arrive
- If the user scrolls down to the bottom again, then new updates will keep the view at the bottom (i.e. it will show the latest contents)
```

frontend/src/components/common/LogViewer.tsx
```diff
@@ -61,12 +63,27 @@ export function LogViewer(props: LogViewerProps) {
...
+      const wrapperPosition = logsBottomRef.current.getBoundingClientRect().bottom;
+      const scrollPosition = event.target.getBoundingClientRect().bottom + scrollMargin;
+      /*  compare if the terminal wrapper bottom is in the scroll limit then the user is at 
+          the bottom of the screen
+      */
+      if (Math.trunc(wrapperPosition) < Math.trunc(scrollPosition)) {
```
illume, line 78: How about this?
```ts
const scrollAtBottom = Math.trunc(wrapperPosition) < Math.trunc(scrollPosition);
if (scrollAtBottom) {
...
```

frontend/src/components/common/LogViewer.tsx
ashu8912: updated

frontend/src/components/common/LogViewer.tsx
ashu8912: fixed, thanks

joaquimrocha (approved):

PR #632, opened 2022-07-25
Width cosmetic fixes

joaquimrocha:
These changes do:
* Remove an extra left margin when on sm
* Prevent the image ID from expanding the details views horizontally
* Remove the now unneeded overflow-wrap from the image name
* Add the navigation tabs when the sidebar is the narrow static version (width=sm)
* Wrap section titles so they fit better on small screens

Testing:
- [ ] Go to a pod's details view: verify that the image and image ID have not forced the page width to go beyond normal
- [ ] Enter inspect mode (Ctrl+Shift+I), click on the mobile sizes button and redeuce the page's width: There should not be a sudden extra margin when the width is between `sm` and `md`
- [ ] Verify that in the narrow sidebar above, you can see the navigation tabs
- [ ] Verify the sidebar works as expected in all other resolutions

### DIFF
--- a/frontend/src/components/App/Layout.tsx
+++ b/frontend/src/components/App/Layout.tsx
@@ -14,7 +14,7 @@ const useStyle = makeStyles(theme => ({
   content: {
     flexGrow: 1,
     marginLeft: 'initial',
-    [theme.breakpoints.between('sm', 'md')]: {
+    [theme.breakpoints.only('sm')]: {
       marginLeft: drawerWidthClosed,
     },
   },
--- a/frontend/src/components/Sidebar/HeadlampButton.tsx
+++ b/frontend/src/components/Sidebar/HeadlampButton.tsx
@@ -42,9 +42,16 @@ export interface HeadlampButtonProps {
   mobileOnly?: boolean;
   /** Called when sidebar toggles between open and closed. */
   onToggleOpen: () => void;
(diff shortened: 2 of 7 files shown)


frontend/src/components/App/Layout.tsx
```diff
@@ -14,7 +14,7 @@ const useStyle = makeStyles(theme => ({
   content: {
     flexGrow: 1,
     marginLeft: 'initial',
-    [theme.breakpoints.between('sm', 'md')]: {
+    [theme.breakpoints.between('xs', 'sm')]: {
```
ashu8912, line 17: @joaquimrocha do we need this margin on any screen?? cause having a blank space on mobile screen beside sidebar is still not a good UX

ashu8912:
On a small screen this is how it looks with the margin
Screenshot 2022-07-28 at 11 48 04 AM

ashu8912:
One more thing that i found while testing this PR, is that the section box title needs to be wrapped as well when it overflows
Screenshot 2022-07-28 at 11 48 36 AM

frontend/src/components/App/Layout.tsx
joaquimrocha: Good cathc. Thanks.

illume (approved):
👍
great!

PR #633, opened 2022-07-28
Fix routes

joaquimrocha:
I have noticed that my pluralName PR's last changes introduced a bug in the routes. So this PR fixes that and adds a new test to check for the routes.


illume:
Ah, this is the previous PR: https://github.com/kinvolk/headlamp/pull/610

illume (approved):
👍

PR #634, opened 2022-07-28
Add support for volumeMounts, PVC and initContainers in the Helm chart

yolossn:


charts/headlamp/values.yaml
```diff
@@ -22,6 +22,9 @@ nameOverride: ""
 # -- Overrides the full name of the chart
 fullnameOverride: ""
 
+# -- An optional list of init containers to be run before the main containers are started.
```
knrt10, line 25:
```suggestion
# -- An optional list of init containers to be run before the main containers.
```

charts/headlamp/values.yaml
```diff
@@ -76,6 +76,12 @@ service:
   # -- Kubernetes Service port
   port: 80
 
+# -- Headlamp containers volume mounts
+volumeMounts: []
+
+# -- Headlamp pods volumes
```
knrt10, line 83:
```suggestion
# -- Headlamp pod's volumes
```

charts/headlamp/templates/pvc.yaml
```diff
@@ -0,0 +1,21 @@
...
+  {{- end }}
+spec:
+  {{- range .Values.persitentVolumeClaim.accessModes}}
+  accessModes:
+    - {{ . }}
+  {{- end}}
+  resources:
+    requests:
+      storage: {{ .Values.persitentVolumeClaim.size }}
+{{- end}}
```
knrt10, line 21: We should add option for `storageClassName` and `selector` if available.

charts/headlamp/values.yaml
```diff
@@ -82,6 +82,17 @@ volumeMounts: []
...
+  # -- Enable Persistent Volume Claim
+  enabled: false
+  # -- Annotations to add to the persistent volume claim (if enabled)
+  annotations:
+    {}
+  # -- AccessModes for the persistent volume claim, eg: ReadWriteOnce, ReadOnlyMany, ReadWriteMany etc.
+  accessModes: []
+  # -- Size of the persistent volume claim, eg: 10Gi. Required if enabled is true.
+  size: ""
+
```
knrt10, line 96: As suggested above. Adding option for `storageClassName` and `selector` would be nice for the user.

charts/headlamp/values.yaml
```diff
@@ -37,6 +37,7 @@ config:
     issuerURL: ""
     # -- OIDC scopes to be used
     scopes: ""
+  pluginDir: "/headlamp/plugins"
```
knrt10, line 40:
```suggestion
  pluginsDir: "/headlamp/plugins"
```

charts/headlamp/templates/deployment.yaml
```diff
@@ -37,7 +37,9 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - "-in-cluster"
-            - "-plugins-dir=/headlamp/plugins"
+            {{- with .Values.config.pluginDir}}
```
knrt10, line 40: As the flag is named `plugins-dir`, I think it would make sense the variable to be `pluginsDir`. WDYT?

knrt10 (changes requested):
Nice work @yolossn. Some NITs, rest looks good

knrt10 (approved):
+1

PR #636, opened 2022-07-28
Center cluster chooser

joaquimrocha:
Before:
Screenshot of topbar before changes

After:
Screenshot of topbar after changes

Testing:
- [ ] Go to Headlamp, verify that the cluster button is always center, no matter whether the sidebar is open or closed
- [ ] Verify that the topbar behaves as expected in different resolutions

### DIFF
--- a/frontend/src/components/App/TopBar.tsx
+++ b/frontend/src/components/App/TopBar.tsx
@@ -197,7 +197,6 @@ export function PureTopBar({
   const { t } = useTranslation('frequent');
   const theme = useTheme();
   const isSmall = useMediaQuery(theme.breakpoints.down('sm'));
-  const isMedium = useMediaQuery(theme.breakpoints.down('sm'));
   const dispatch = useDispatch();
 
   const openSideBar = !!(isSidebarOpenUserSelected === undefined ? false : isSidebarOpen);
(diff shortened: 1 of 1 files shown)


ashu8912 (commented):
There are two small issues here

1. The cluster chooser button is not by default centered on the first render, if we shrink the sidebar and than expand it than the center functionality takes into place
2. The transition of cluster chooser button from sidebar shrink to expand or vice versa should be smooth as if it slides to fill its centered position otherwise it gives a laggy experience if you render it directly at the two points.

joaquimrocha:
> There are two small issues here
> 
> 1. The cluster chooser button is not by default centered on the first render, if we shrink the sidebar and than expand it than the center functionality takes into place

Good catch. I have addressed this.

> 2. The transition of cluster chooser button from sidebar shrink to expand or vice versa should be smooth as if it slides to fill its centered position otherwise it gives a laggy experience if you render it directly at the two points.

This is not easily fixed, and we have also other places where this happens, like expanding items in the sidebar. We should maybe list all such examples to fix them.

ashu8912 (approved):
Tested, LGTM!!! thanks

illume (approved):
👍 thanks

PR #639, opened 2022-08-01
Add MSI Installer Generator

lijianzhi01:
Add scripts to generate MSI installer


joaquimrocha:
@lijianzhi01 , This is nice. Thanks. I will merge it into a temp branch and improve a few things (I think we can add a few more options coming from the app/package.json).

PR #640, opened 2022-08-01
Show build version on frontend

### LINKED ISSUE #384: Display Headlamp's version in the UI
## Current situation

We have no easy way to understand what version of Headlamp is being run.

## Impact

It's useful to quickly understand what version it is, for users to include it in bug reports, etc.

## Ideal future situation

We show the version (e.g. `X.Y.Z desktop` or `X.Y.Z in-cluster`) as a `Headlamp version` entry in the version dialog (when clicking the k8s version on the bottom left).

## **Implementation options

Ideally we only have a single source for determining the version. Usually this is done using git, as in the Nebraska project, but the truth is that this means always making sure that git is present (not building from a tarball, which is uncommon but still useful if we get into other desktop platforms/formats) and clean when building the project.
Another option is to take the version set in `app/package.json`, as it's the one we currently update when releasing.

In any case, the version should be included as part of the config endpoint in the server, i.e. the server tells the frontend which version of Headlamp it's using.

joaquimrocha:
This PR adds a new "version" dialog to the frontend which ATM displays the product name (Headlamp by default), version number and git version (commit ref).

This is useful for users to be able to report the version when filing issues, etc.
In the future is may double or be replaced by an about dialog.

## Testing 
Placing the trigger for this dialog was a bit tricky. So **ATM**:
- [ ] There's an (i) button in the Cluster chooser's title: click it and the chooser dialog should be closed (unless in the home view) and the version dialog should be shown
- [ ] There's a "Headlamp X.Y.Z" link in the "user menu" (when in a cluster view): click it and the version dialog should be shown

fixes #384


frontend/src/redux/stores/store.tsx
```diff
@@ -6,6 +6,12 @@ import reducers from '../reducers/reducers';
 import { INITIAL_STATE as UI_INITIAL_STATE } from '../reducers/ui';
 import rootSaga from '../sagas/sagas';
 
+window.config = {
```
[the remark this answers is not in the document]
joaquimrocha: I was supposed to delete the one in the App.tsx but forgot. I had it there initially but moved it to the store because this way we get it in the tests as well.

ashu8912:
Found one more small issue while testing, the version shouldn't appear if we have multiple clusters right on the user menu??
Also if there is no version specified that shouldn't even appear. Coz this white line is more confusing to what it actually is
Screenshot 2022-08-09 at 1 41 42 PM

frontend/src/redux/stores/store.tsx
illume, line 9: Probably good to prefix this. `headlampConfig`.

frontend/src/components/App/TopBar.tsx
```diff
@@ -179,6 +193,7 @@ export function PureTopBar({
   const { t } = useTranslation('frequent');
   const isSmall = useMediaQuery('(max-width:960px)');
   const isMedium = useMediaQuery('(max-width:960px)');
+  const dispatch = useDispatch();
```
illume, line 196: To keep this separated from redux, a `versionOnClick` could be passed in to the PureTopBar.

illume (changes requested):
Looking good. I left a few notes.

frontend/src/components/App/TopBar.tsx
joaquimrocha: I used dispatch because we have two places where we can trigger the dialog from, the cluster chooser + the user menu.
So this is a global way of triggering the build dialog.

frontend/src/components/App/TopBar.tsx
```diff
@@ -236,6 +252,19 @@ export function PureTopBar({
...
+        component="a"
+        onClick={() => {
+          dispatch(setVersionDialogOpen(true));
+          handleMenuClose();
+        }}
+        dense
+        className={classes.versionLink}
+      >
+        <ListItemText>
+          {window.headlampConfig.HEADLAMP_PRODUCT_NAME} {window.headlampConfig.HEADLAMP_VERSION}
```
illume, line 265: A cleaner way to do this would be to put this information in the store.

That would also work with a story test.

illume:
There's a conflict now from another PR I merged.

I think before merging this PR should:
- use the store for state (with an action that puts the headlampConfig information into the store).
- have a story for this feature

joaquimrocha:
> There's a conflict now from another PR I merged.

I will fix it.

> * use the store for state (with an action that puts the headlampConfig information into the store).

I don't think this info belongs to the store, as it's info we set up at build time, not something to be set dynamically. Later if needed it there, I think we can add it without an issue. For plugins, we can wrap this in a function (`Headlamp.getVersion` or something, so even if we change where it's stored, we don't break it).

> * have a story for this feature

We're have to update the snapshot every time a version changes or override it for the story. I can try doing the override solution.

illume (approved):
Looking gooood to me. Thanks! 🎉

PR #641, opened 2022-08-01
Add msi installer

joaquimrocha:
## Testing
Install Wix and ensure that light.exe and candle.exe are in the Path.
- [ ] Run: `make`, then move to the app folder and run `npm run package-msi`: Test install the MSI package that's been generated in the dist/ folder.


joaquimrocha:
We should add a Github action later, to generate the MSI as well as the regular NSIS.

app/windows/msi/build.js
```diff
@@ -0,0 +1,43 @@
...
+});
+
+const msiOptions = {
+  appDirectory: APP_DIR,
+  outputDirectory: OUT_DIR,
+  description: info.description,
+  exe: exeName,
+  name: nameOptions.productName,
+  manufacturer: info.author.name,
+  version: process.env.npm_package_version,
```
illume, line 30: I'm not sure about `process.env.npm_package_version`. Can you please explain why this is used?

app/windows/msi/build.js
```diff
@@ -0,0 +1,43 @@
...
+  arch: 'x86',
+};
+
+// Generate the exe name from electron-builder's artifactName
+let exeName = info.build.artifactName.split('.')[0];
+Object.entries(nameOptions).forEach(([key, value]) => {
+  exeName = exeName.replace(`\${${key}}`, value);
+});
+
+const msiOptions = {
```
illume, line 23: I reviewed the configuration options which are documented here: https://github.com/felixrieseberg/electron-wix-msi#configuration

> shortName (optional, string) - A short name for the app, used wherever spaces and special characters are not allowed. Will use the name if left undefined.

I wonder if we need a shortName here? It's a place for where spaces and special characters are not allowed. For "Headlamp" it is ok to not have one, but for names like "Sxxxx Bxxxx" it's not.

> upgradeCode (string, optional) - A unique UUID used by your app to identify itself. This module will generate one for you, but it is important to reuse it to enable conflict-free upgrades.

This `upgradeCode` seems important to keep track of and check.

> arch (string, optional) - Defines the architecture the MSI is build for. Values can be either x86 or x64. Default's to x86 if left undefined.

How is arch handled here? We need a x64 only? Seems we are setting 'x86'.

Perhaps we could override some of the images shown in the installer?https://github.com/felixrieseberg/electron-wix-msi#ui-configuration-optional

app/windows/msi/build.js
```diff
@@ -0,0 +1,43 @@
+// You need to install wix tool set first
+// Reference https://ourcodeworld.com/articles/read/927/how-to-create-a-msi-installer-in-windows-for-an-electron-framework-application
```
illume, line 2: Can you please add the docs link?
https://github.com/felixrieseberg/electron-wix-msi#configuration

illume (changes requested):
Looking good so far. I left some questions and notes.

For the first commit, can you please add the `app:` prefix?

app/windows/msi/build.js
joaquimrocha: This is just the app's version. We can use the info.version instead, though it should be the same.

app/windows/msi/build.js
joaquimrocha: All good points.
I will address all but the images, which we can add later.

joaquimrocha:
Hopefully it's good now.

docs/development/_index.md
```diff
@@ -66,11 +66,18 @@ make app-linux
...
+make app-mac
 ```
 
 ```bash
-make app-mac
+make app-win
+```
+
+For Windows, by default it will produce an installer using [NSIS (Nullsoft Scriptable Install System)](https://sourceforge.net/projects/nsis/).
+If you prefer an `.msi` installer, then be sure to install the [WiX toolset](https://wixtoolset.org/), have its `light.exe` and `candle.exe` in the Windows path, and run:
```
illume, line 77: I have wixtoolset installed with chocolatey, but there is no light.exe or candle.exe. Do you know where it should be?

illume: I guess the chocolatey install was broken. The installer worked.

joaquimrocha: I installed with WiX but had to add those executables to Windows's Path.

docs/development/_index.md
```diff
@@ -66,11 +66,18 @@ make app-linux
...
+
+```bash
```
illume, line 79:
```
set PATH=%PATH%;C:\Program Files (x86)\WiX Toolset v3.11\bin
make app-win-msi
```

illume:
There's no branding/headlamp logo on the MSI installer:
(but I think you know that?)
image

Here's a screenie of the uninstaller without the logo:
image

Note: no icon on msi file in explorer:
image

The installer doesn't add a startbar shortcut, but does add a desktop icon.

---

The installer works, but I cannot run the Headlamp program (from desktop shortcut or from cmd line).  There's no error message (or any logging at all).

Does it work for you?

joaquimrocha:
> The installer works, but I cannot run the Headlamp program (from desktop shortcut or from cmd line). There's no error message (or any logging at all).
> 
> Does it work for you?

It did work from me. I launched it from the entry in the Start menu *I think*. I will try again.

As for the lack of branding. This is not an installer we're publishing ATM yet, so we can add all those images later.

joaquimrocha:
@illume , I have verified that there was indeed an issue with running the app after installation. It had to do with the `exe` name given, which I thought was just the installer name but it's the app's executable. I updated that and left a comment.

I also don't see a start menu now. Not sure why... But this is a detail I think we can refine later.

illume:
> I also don't see a start menu now. Not sure why... But this is a detail I think we can refine later.

Cool. What's left open for the MSI based installer?
- protocol support (adding the registry key you mentioned)
- branding
- start menu item (I believe it had a start menu item for me... so I don't know what's going on there)

Should these be done in this PR or other ones?

docs/development/_index.md
illume: Would you mind adding this set PATH command in to the docs there?

An improvement for later might be to check if WiX Toolset is there and add it to the path inside the make target/or npm script.

joaquimrocha:
> > I also don't see a start menu now. Not sure why... But this is a detail I think we can refine later.
> 
> Cool. What's left open for the MSI based installer?
> 
> * protocol support (adding the registry key you mentioned)
> * branding
> * start menu item (I believe it had a start menu item for me... so I don't know what's going on there)

Yes.

> Should these be done in this PR or other ones?

I think this PR can already be useful for some people, and has minimal impact to the UX since we're not publishing the .msi just now. So let's merge it in and I will open an issue to track the remaining tasks, after we merge this PR.

PR #643, opened 2022-08-02
plugins/headlamp-plugin: Add upgrade command

illume:
An upgrade command for upgrading plugins (and folders of plugins).

For https://github.com/kinvolk/headlamp/issues/426

- [x] Check if headlamp-plugin is outdated when starting `npm start`
- [x] When package is upgraded, make sure the version is also updated (For example `^0.5.0` should be `^0.5.1`)
- [x] Upgrade @kinvolk/headlamp-plugin if it is outdated. On one folder or many.
- [x] Add extra config inside package.json (like the new eslint/prettier config)
- [x] Add extra required config files (like a previous version did with a tsconfig file)
- [x] Add tests

### Testing this PR before it's released:

```bash
node plugins/headlamp-plugin/bin/headlamp-plugin.js upgrade plugins/examples
```

#### Upgrade multiple folders

```bash
npx @kinvolk/headlamp-plugin upgrade plugins/examples
```

#### Upgrade current folder

```bash
cd plugins/examples/sidebar
npx @kinvolk/headlamp-plugin upgrade
```

## Testing done

- [x] run it against our plugins/examples/
- [x] Add a few tests to test-headlamp-plugin.sh

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -44,7 +44,10 @@ function create(name, link) {
 
   console.log(`Creating folder :${dstFolder}:`);
 
-  fs.copySync(templateFolder, dstFolder, { errorOnExist: true, overwrite: false });
+  fs.copySync(templateFolder, dstFolder, {
+    errorOnExist: true,
+    overwrite: false,
+  });
 
   function replaceFileVariables(path) {
     fs.writeFileSync(
@@ -65,7 +68,9 @@ function create(name, link) {
   // This can be used to make testing locally easier.
   if (link) {
     console.log('Linking @kinvolk/headlamp-plugin');
-    child_process.spawnSync('npm', ['link', '@kinvolk/headlamp-plugin'], { cwd: dstFolder });
+    child_process.spawnSync('npm', ['link', '@kinvolk/headlamp-plugin'], {
+      cwd: dstFolder,
+    });
   }
 
   console.log('Installing dependencies...');
@@ -85,7 +90,9 @@ function create(name, link) {
   if (link) {
     // Seems to require linking again with npm 7+
     console.log('Linking @kinvolk/headlamp-plugin');
-    child_process.spawnSync('npm', ['link', '@kinvolk/headlamp-plugin'], { cwd: dstFolder });
+    child_process.spawnSync('npm', ['link', '@kinvolk/headlamp-plugin'], {
+      cwd: dstFolder,
+    });
   }
 
   console.log(`"${dstFolder}" created.`);
@@ -99,37 +106,9 @@ function create(name, link) {
 }
 
 /**
- * Copies the built plugin to the app config folder ~/.config/Headlamp/plugins/
+ * Compile callback for webpack to show any warnings + errors.
  */
-function copyToPluginsFolder(config) {
-  const packageJson = JSON.parse(fs.readFileSync('package.json', 'utf8'));
-
-  // @todo: should the whole package name be used here,
-  //    and the load be fixed to use? What about namespace packages?
-  const packageName = packageJson.name.split('/').splice(-1)[0];
-  const paths = envPaths('Headlamp', { suffix: '' });
-  const configDir = fs.existsSync(paths.data) ? paths.data : paths.config;
-
-  config.plugins = [
-    new FileManagerPlugin({
-      events: {
-        onEnd: {
-          copy: [
-            {
-              source: './dist/*',
-              destination: path.join(configDir, 'plugins', packageName),
-            },
-          ],
-        },
-      },
-    }),
-  ];
-}
-
-/**
- * Compile and show any warnings + errors
- */
-function compile(err, stats) {
+function compileMessages(err, stats) {
   if (err && err.message) {
     console.log(err);
     return;
@@ -230,12 +209,62 @@ function extract(pluginPackagesPath, outputPlugins) {
  * @returns {0} Exit code, where 0 is success.
  */
 function start() {
+  /**
+   * Copies the built plugin to the app config folder ~/.config/Headlamp/plugins/
+   *
+   * Adds a webpack config plugin for copying the folder.
+   */
+  function copyToPluginsFolder(webpackConfig) {
+    const packageJson = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+
+    // @todo: should the whole package name be used here,
+    //    and the load be fixed to use? What about namespace packages?
+    const packageName = packageJson.name.split('/').splice(-1)[0];
+    const paths = envPaths('Headlamp', { suffix: '' });
+    const configDir = fs.existsSync(paths.data) ? paths.data : paths.config;
+
+    webpackConfig.plugins = [
+      new FileManagerPlugin({
+        events: {
+          onEnd: {
+            copy: [
+              {
+                source: './dist/*',
+                destination: path.join(configDir, 'plugins', packageName),
+              },
+            ],
+          },
+        },
+      }),
+    ];
+  }
+
+  /**
+   * Inform if @kinvolk/headlamp-plugin is outdated.
+   */
+  function informIfOutdated() {
+    console.log('Checking if headlamp-plugin is up to date...');
+    const outdated = getNpmOutdated();
+    if ('@kinvolk/headlamp-plugin' in outdated) {
+      const url = `https://github.com/kinvolk/headlamp/releases`;
+      console.warn(
+        '    @kinvolk/headlamp-plugin is out of date. Run the following command to upgrade \n' +
+          `    See release notes here: ${url}` +
+          '    npx @kinvolk/headlamp-plugin upgrade'
+      );
+    } else {
+      console.log('    @kinvolk/headlamp-plugin is up to date.');
+    }
+  }
+
+  informIfOutdated();
+
   console.log('Watching for changes to plugin...');
   config.watch = true;
   config.mode = 'development';
   process.env['BABEL_ENV'] = 'development';
   copyToPluginsFolder(config);
-  webpack(config, compile);
+  webpack(config, compileMessages);
 
   return 0;
 }
@@ -263,7 +292,7 @@ function build(packageFolder) {
 
     process.chdir(folder);
     console.log(`Building "${folder}" for production...`);
-    webpack(config, compile);
+    webpack(config, compileMessages);
     console.log(`Done building: "${folder}".`);
     process.chdir(oldCwd);
     return true;
@@ -302,6 +331,7 @@ function build(packageFolder) {
  * @returns {0 | 1} Exit code, where 0 is success, 1 is failure.
  */
 function format(packageFolder) {
+  // @todo: this should work on a folder of packages
   try {
     child_process.execSync('prettier --config package.json --write src', {
       stdio: 'inherit',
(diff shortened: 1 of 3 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -316,6 +348,205 @@ function format(packageFolder) {
...
+    checkKeys.forEach(key => {
+      if (JSON.stringify(packageJson[key]) !== JSON.stringify(templatePackageJson[key])) {
+        packageJson[key] = templatePackageJson[key];
+        changed = true;
+        console.log(`Updated package.json field "${key}"`);
+      }
+    });
+
+    if (changed) {
+      fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, '  ') + '\n');
```
joaquimrocha, line 456: Should we write a backup file with the previous contents? I know people should use git anyways but...

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -316,6 +348,205 @@ function format(packageFolder) {
...
+    if (skipPackageUpdates === true) {
+      console.log('Skipping package updates.');
+      return true;
+    }
+
+    const outDated = getNpmOutdated();
+    if ('@kinvolk/headlamp-plugin' in outDated) {
+      // Upgrade the @kinvolk/headlamp-plugin
+
+      const cmd = 'npm install @kinvolk/headlamp-plugin@latest --save';
```
joaquimrocha, line 470: Won't this get the version to be `latest` in the package.json? Probably doesn't matter with package-lock.json, but I wonder if this is the case.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -316,6 +348,205 @@ function format(packageFolder) {
...
+    const oldCwd = process.cwd();
+
+    process.chdir(folder);
+    console.log(`Upgrading "${folder}"...`);
+
+    addMissingTemplateFiles();
+    addMissingConfiguration();
+    if (!upgradeHeadlampPlugin(skipPackageUpdates)) {
+      return false;
+    }
```
joaquimrocha, line 504: This block led me to think that skipPackageUpdates would be used in the function we're passing it to and if it's false, then the function would also return false; after all, the package would not have been updated. Which would make this whole logic very confusing for the user.
Then I read the upgradeHeadlampPlugin and realized it's returning false just on error.

I think the code would be clearer IMO if you just check the `skipPackageUpdates` here (as `upgradeHeadlampPlugin` is not called from anywhere else).

plugins/headlamp-plugin/test-headlamp-plugin.sh
```diff
@@ -53,3 +53,10 @@ npm run lint-fix
 
 # test type script error checks
 npm run tsc
+
+# test upgrade adds missing files
+rm jsconfig.json tsconfig.json src/headlamp-plugin.d.ts
+node bin/headlamp-plugin.js upgrade --no-package-updates
```
joaquimrocha, line 59: The option seems to be called `skip-package-updates`.

plugins/headlamp-plugin/test-headlamp-plugin.sh
```diff
@@ -53,3 +53,10 @@ npm run lint-fix
...
+stat jsconfig.json
+stat tsconfig.json
+stat src/headlamp-plugin.d.ts
```
joaquimrocha, line 62: I wonder if it's possible to add a test for a headlamp-plugin version upgrade.

joaquimrocha (changes requested):
Great work! I left a couple of questions + noticed there's an outdated arg name being used.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: I don't think so.

`npm install` doesn't for example.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: I checked this. It sets the version to the latest version, not to "latest". Now it sets it to "^0.5.1" for example.

plugins/headlamp-plugin/test-headlamp-plugin.sh
illume: Thanks.

plugins/headlamp-plugin/test-headlamp-plugin.sh
illume: I added a test for this.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Good point. I moved the skipPackageUpdates check outside that function.

plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha: Ok.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -230,12 +209,64 @@ function extract(pluginPackagesPath, outputPlugins) {
...
+  /**
+   * Copies the built plugin to the app config folder ~/.config/Headlamp/plugins/
+   *
+   * Adds a webpack config plugin for copying the folder.
+   */
+  function copyToPluginsFolder(webpackConfig) {
+    const packageJson = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+
+    // @todo: should the whole package name be used here,
+    //    and the load be fixed to use? What about namespace packages?
```
joaquimrocha, line 221: Good point. Maybe making the namespace as part of the target path? `@kinvolk/podcounter-headlamp-plugin` -> `kinvolk-podcounter-headlamp-plugin`?

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -230,12 +209,64 @@ function extract(pluginPackagesPath, outputPlugins) {
...
+  function informIfOutdated() {
+    console.log('Checking if headlamp-plugin is up to date...');
+    const outdated = getNpmOutdated();
+    if ('@kinvolk/headlamp-plugin' in outdated) {
+      // const latest = outdated['@kinvolk/headlamp-plugin'].latest;
+      // const url = `https://github.com/kinvolk/headlamp/releases/tag/headlamp-plugin-${latest}`;
+      // const url = `https://github.com/kinvolk/headlamp/releases`;
+      console.warn(
+        '    @kinvolk/headlamp-plugin is out of date. Run the following command to upgrade \n' +
+          //  `    See release notes here: ${url}` +
```
joaquimrocha, line 254: I guess the commented out code should be removed?

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -316,6 +348,208 @@ function format(packageFolder) {
...
+ *  {}
+ * ```
+ */
+function getNpmOutdated() {
+  let result = null;
+  try {
+    result = child_process.execSync('npm outdated --json', {
+      encoding: 'utf8',
+    });
+    // console.log(result);
```
joaquimrocha, line 382: Remove the comment, or make it print a meaningful message.

plugins/headlamp-plugin/test-headlamp-plugin.sh
```diff
@@ -53,3 +53,22 @@ npm run lint-fix
...
+python3 -c "import sys;sys.stdout.write(\"\".join([l if '\"@kinvolk/headlamp-plugin\"' not in l else '    \"@kinvolk/headlamp-plugin\": \"^0.4.9\"\n' for l in open(\"package.json\", \"r\").readlines()]))" > /tmp/package.json.bak && mv /tmp/package.json.bak package.json
+
+# test upgrade updates the package line, and the old version is not in there
+node ../bin/headlamp-plugin.js upgrade
+echo "
+import sys
+old_version = '0.4.9'
+if old_version in open('package.json').read():
+  sys.exit(1)
+" | python3
```
joaquimrocha, line 74: I was wondering why this is not using just sed/grep and other shell tools, but remembered we need this to run also in Windows without installing a bunch of tools.
Will we have Python in the Windows CI though? I wonder if this should just be js since we have node already.

joaquimrocha (changes requested):
Found a couple of things but overall it's looking good.

plugins/headlamp-plugin/test-headlamp-plugin.sh
illume: Yes, I'll fix the tests to use nodejs.

illume: (Also so the tests can be run on windows cmd like we spoke about another time)

plugins/headlamp-plugin/test-headlamp-plugin.sh
```diff
@@ -53,3 +53,22 @@ npm run lint-fix
...
+
+# test upgrade adds missing files
+rm jsconfig.json tsconfig.json src/headlamp-plugin.d.ts
+node ../bin/headlamp-plugin.js upgrade --skip-package-updates
+stat jsconfig.json
+stat tsconfig.json
+stat src/headlamp-plugin.d.ts
+
+# change @kinvolk/headlamp-plugin version in package.json to an old one "^0.4.9"
+python3 -c "import sys;sys.stdout.write(\"\".join([l if '\"@kinvolk/headlamp-plugin\"' not in l else '    \"@kinvolk/headlamp-plugin\": \"^0.4.9\"\n' for l in open(\"package.json\", \"r\").readlines()]))" > /tmp/package.json.bak && mv /tmp/package.json.bak package.json
```
illume, line 65: This one liner has perl vibes.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Do you think the message is ok? Or should I add the release notes in there too?

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Thanks. I removed it.

plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha: I think that having the URL would be more complete.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: That seems fine... I think. Or expanding the loading parts of the code so it can work with namespaces like `@kinvolk/package` properly. Need to think it through a bit more, with how that could affect namespace collisions.

It's not the topic of this PR so I've created an issue for this one: https://github.com/kinvolk/headlamp/issues/680

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Thanks. Yeah, agreed. I added the release URL in the message.

plugins/headlamp-plugin/test-headlamp-plugin.sh
illume: Converting tests to js was done in the other PR. This one now uses JS for these new tests instead of bash+python.

joaquimrocha (approved):

PR #646, opened 2022-08-04
frontend: Fix showing cluster chooser on top bar

joaquimrocha:
In a recent change, we stopped using the cluster related hooks in the
PureTopBar component as it led to conditionally using hooks. However,
those hooks (and info from them) need to be used from the wrapping
component (TopBar), otherwise no cluster chooser button will be shown.

### DIFF
--- a/frontend/src/components/App/TopBar.tsx
+++ b/frontend/src/components/App/TopBar.tsx
@@ -15,7 +15,7 @@ import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import LocaleSelect from '../../i18n/LocaleSelect/LocaleSelect';
 import { getToken, setToken } from '../../lib/auth';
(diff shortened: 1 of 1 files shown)


ashu8912 (commented):
This works and the code LGTM too, but I have a small doubt, in the commit it says "we stopped using the cluster related hooks in the PureTopBar component" was that the case??  Coz PureTopBar component was receiving cluster and clusterConfig as prop earlier as well i think, it's just that we weren't passing them to the component and that's why it didn't show up.
@joaquimrocha  can you please help me understand this better? I am sure i am missing something here.

illume (approved):
👍

illume:
Previously if they weren't supplied, the use* calls were made inside. Then they were removed inside, and now they are passed in... fixing the bug.

Hopefully that clears it up.

PR #647, opened 2022-08-04
Add basic test of container image on minikube

illume:
Runs minikube, and then checks the service URL is serving minikube.

## How to use

Look at the build container check.

## Testing done

- [x] It tests it on a github check. See "Test Service URLs" part of "Build container and test".
- [x] Ran it multiple times, and it doesn't fail. So hopefully not a flaky test!


yolossn (approved):
LGTM

joaquimrocha (approved):

PR #648, opened 2022-08-05
Add Debian package

joaquimrocha:
Testing:
- [ ] Run make app-linux: then install the resulting Debian package
- [ ] Run the app after it's installed to verify everything works

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -38,6 +38,9 @@
       "perMachine": false,
       "allowToChangeInstallationDirectory": true
     },
+    "deb": {
+      "artifactName": "${name}_${version}-1_${arch}.${ext}"
+    },
     "linux": {
       "target": [
         {
@@ -54,6 +57,12 @@
             "x64",
             "armv7l",
             "arm64"
+          ]
+        },
+        {
+          "target": "deb",
+          "arch": [
+            "x64"
           ]
         }
       ],


app/package.json
```diff
@@ -19,7 +19,7 @@
     "copy-plugins": "npx --no-install rimraf build/.plugins && mkdirp build/.plugins && copyfiles ../.plugins build/.plugins"
   },
   "build": {
-    "appId": "com.kinvolk.headlamp",
```
illume, line 22: Will this change affect upgrades?

app/package.json
```diff
@@ -55,6 +58,12 @@
             "armv7l",
             "arm64"
           ]
+        },
+        {
+          "target": "deb",
+          "arch": [
+            "x64"
```
illume, line 65: Was only x64(and no arm) intentional?

illume (commented):
I left a couple of questions...

illume:
It doesn't run for me on WSL2, but then neither does the AppImage app anymore. It flashes black and then crashes to a white screen.

I also tried with `--disable-gpu` and no luck. I guess this is the WSL issue you mentioned from the electron upgrade.

```bash
17:19:17-rene~/dev/headlamp/app/dist (joaquimrocha-add-debian-package)$ /usr/bin/headlamp
17:19:28.277 › App starting...
Check for updates:  true
[463274:0809/171928.385256:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[463274:0809/171928.385365:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
[463274:0809/171928.387171:ERROR:bus.cc(398)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[463274:0809/171928.387251:ERROR:bus.cc(398)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[463274:0809/171928.408257:ERROR:bus.cc(398)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
arguments passed to backend server []
17:19:28.468 › server process stdout: *** Headlamp Server ***
  API Routers:
        localhost:4466/clusters/minikube/{api...} -> https://127.0.0.1:49169
        localhost:4466/clusters/anotherkube2/{api...} -> https://127.0.0.1:49169

17:19:28.469 › server process stderr: 2022/08/09 17:19:28 plugins-dir: /home/rene/.config/Headlamp/plugins

libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
[463303:0809/171928.788226:ERROR:viz_main_impl.cc(186)] Exiting GPU process due to errors during initialization
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
[463364:0809/171928.837465:ERROR:gpu_memory_buffer_support_x11.cc(44)] dri3 extension not supported.
```


ps. Building make app-linux is extremely slow on my slow laptop... I temporarily deleted the targets from the app/package.json which is fine.

illume:
Hrmm. The issue is fixed when I remove the plugins... I guess it's not related to this PR. Something to investigate on main before we release.

app/package.json
joaquimrocha: Hmm, good question. I don't think this being used anywhere, but we should double check.

app/package.json
joaquimrocha: Yes. Let's go with x64 for now...

joaquimrocha:
> Hrmm. The issue is fixed when I remove the plugins... I guess it's not related to this PR. Something to investigate on main before we release.

I have noticed in the past couple of days that Headlam's showing fine on WSL2 (not showing the artifacts I had seen).

joaquimrocha:
@illume , turns out this leaves 2 Headlamp entries alright.
Screenshot showing two Headlamp entries in installer programs

joaquimrocha:
I have removed the appID change. We should figure out later how to do the appID change properly.

illume (approved):
👍

illume:
> turns out this leaves 2 Headlamp entries alright.

Ah, cool.

PR #652, opened 2022-08-08
Allow Port forwarding from the headlamp UI

### LINKED ISSUE #536: [RFE] Add ability to port-forward to a pod directly from Headlamp
It will be a nice addition to port-forward to an pod directly from the Headlamp UI. 
Refer:
https://github.com/kubernetes-client/javascript/blob/6c5a58f51b8cdb01c59cbd5b433013beb4940df2/examples/typescript/port-forward/port-forward.ts

ashu8912:

In the container info section of pod details we now have a button to allow port forwarding to the target port this button is situated right beside the port option
Screenshot 2022-08-08 at 8 48 15 PM

## Testing done
Screenshot 2022-08-08 at 8 48 07 PM

## How To Use
Go to Pod detail view and go to container section if there is an available port on that container we can port forward to it by pressing the Forward Button link, on clicking it we see a loader and then the port forwarded link is shown to us with a stop button that can be pressed to stop the port forward.

fixes #536


ashu8912:
- [x] Allow to port forward a pod from desktop app
- [ ] Store the port forwarding state of a pod
- [ ] Show all the running port-forwards in a new list view under networks tab
- [ ] Ask the user to start or not all the port-forwards once they reopen the app

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -573,6 +637,29 @@ export function ContainerInfo(props: ContainerInfoProps) {
       env[envVar.name] = value;
     });
 
+    function preparePortForwardComponent(containerPort: number) {
+      if (!helpers.isElectron()) {
+        return null;
+      }
+      return !portForwardInitiated ? (
+        <Button onClick={() => handlePortForward(namespace, podname, containerPort)}>
+          Port Forward
```
joaquimrocha, line 646: All buttons like this this need to be translatable.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -555,6 +562,63 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+          cluster === getCluster() &&
+          podname === props.podname &&
+          namespace === props.namespace
+        ) {
+          setIsCurrentPodPortForwarding(true);
+          setPortForwardServerURL(`http://localhost:${port}`);
+        }
+        setPortForwardStarted(true);
+      }
+    );
```
joaquimrocha, line 586: If you listen to this in a higher component and store the info in redux, then you could just use the redux hook in this component to verify the states.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -755,7 +757,12 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
         ) : (
           containers.map((container: any, i: number) => (
             <SectionBox key={i} outterBoxProps={{ pt: 1 }}>
-              <ContainerInfo container={container} status={statuses[container.name]} />
+              <ContainerInfo
+                podname={resource?.metadata.name}
+                namespace={resource?.metadata.namespace}
```
joaquimrocha, line 850: Should we just pass the resource? And then within it use what we need from it instead of passing two props?

app/electron/preload.ts
```diff
@@ -4,13 +4,32 @@ import { contextBridge, ipcRenderer } from 'electron';
...
-    const validChannels = ['setMenu', 'locale', 'appConfig'];
+    const validChannels = [
+      'setMenu',
+      'locale',
+      'appConfig',
+      'port-forward',
+      'port-forward-initiated',
+      'port-forward-started',
+      'port-forward-stopped',
+      'port-forward-stop',
```
joaquimrocha, line 15: A couple of these (initiated/started) would not be needed as they are part of the app -> renderer only (receive end).
Vice-versa for the receive ones below.

I think this could be made simpler if you have just one method `port-forward` and then you have an `action` parameter passed together with any others, where you identify what's going on (also notice the case without dashes, for consistency):

```typescript
desktopApi.send('portForward', {action: 'start', port...})
```

```typescript
desktopApi.receive('portForward', ({action, port}) => ...)
```

app/electron/port-forward.ts
```diff
@@ -0,0 +1,52 @@
...
+    server.close(() => {
+      mainWindow.webContents.send('port-forward-stopped', {
+        podname,
+        namespace,
+        cluster: kc.getCurrentCluster()?.name,
+      });
+    });
+  });
+  server.on('error', (e: any) => {
+    dialog.showErrorBox('Error in creating a port forward tunnel', e.message);
```
joaquimrocha, line 50: This message needs translation.

app/electron/port-forward.ts
```diff
@@ -0,0 +1,52 @@
...
+    kc.loadFromDefault();
+  }
+  const forward = new k8s.PortForward(kc);
+  mainWindow.webContents.send('port-forward-initiated', {
+    podname,
+    namespace,
+    cluster: kc.getCurrentCluster()?.name,
+  });
+  const server = net.createServer(socket => {
+    console.log('port forward initiated request');
```
joaquimrocha, line 26: Please log with more info so we know which ports/pods are requesting it.

app/electron/port-forward.ts
```diff
@@ -0,0 +1,52 @@
...
+  namespace: string,
+  podname: string,
+  targetPort: number
+) {
+  const kc = new k8s.KubeConfig();
+  if (kubeconfig) {
+    kc.loadFromFile(kubeconfig);
+  } else {
+    kc.loadFromDefault();
+  }
```
joaquimrocha, line 18: I'm slightly concerned we're reading from the kubeconfig here, not only creating 2 points where we do this (the other is the server), but also we'd need to make sure that the cluster we're requesting the port-forward for is matched in the kubeconfig (which can be a problem with dynamic clusters).

I see that internally the lib uses an [API endpoint in the k8s server](https://github.com/kubernetes-client/javascript/blob/6c5a58f51b8cdb01c59cbd5b433013beb4940df2/src/portforward.ts#L44) and in the PortForward constructor we can give it a WebSocketHandler which we could create using a cooked up KubeConfig (dynamically, from the info we have).
Not sure if doing this is better than adding a special forward in the backend (because now it's already forwarding any calls we want, so it'd be a matter of using a different port + handler for that), but I think we'll find a good solution that doesn't involve re-reading the kubeconfig.

joaquimrocha (changes requested):

joaquimrocha:
@ashu8912 , this is indeed a good start. I think there's still some things to consider which hopefully I covered in my review. Any questions, let me know.

frontend/src/components/common/Resource/Resource.tsx
ashu8912: updated

frontend/src/components/common/Resource/Resource.tsx
ashu8912: this is now solved by using the backend api

frontend/src/components/common/Resource/Resource.tsx
ashu8912: updated

app/electron/preload.ts
ashu8912: this is now solved by the backend api

app/electron/port-forward.ts
ashu8912: this is not needed now

app/electron/port-forward.ts
ashu8912: updated this we now read from the backend api

app/electron/port-forward.ts
ashu8912: this is no longer needed as we now use the backend api

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -572,6 +656,31 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+    function preparePortForwardComponent(containerPort: number) {
+      if (!helpers.isElectron()) {
+        return null;
+      }
+      return !portForward ? (
+        <Button
+          onClick={() => handlePortForward(namespace, resource?.metadata.name, containerPort)}
+        >
+          {t('frequent|Port Forward')}
+        </Button>
```
joaquimrocha, line 668: We need to set up some feedback for the user to know this is happening. Maybe just having a spinner after we click the button until we hear back from the call?

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -554,6 +594,50 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+          const portForwardsList = [];
+          portForwardsList.push(JSON.stringify(data));
+          return;
+        }
+        const parsedPortForwards = JSON.parse(portForwards);
+        parsedPortForwards.push(data);
+        localStorage.setItem('portforwards', JSON.stringify(parsedPortForwards));
+      })
+      .catch(error => {
+        console.log(error);
```
joaquimrocha, line 624: We should show an error dialog to warn the user that it didn't work.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -554,6 +594,50 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+    if (!podname || !namespace || !containerPort) {
+      return;
+    }
+    fetch(
+      `${helpers.getAppUrl()}/startportforward?cluster=${cluster}&namespace=${namespace}&pod=${podname}&targetPort=${containerPort}`,
+      {
+        method: 'GET',
+        headers: new Headers({ Authorization: `Bearer ${localStorage.getItem('token')}` }),
+      }
+    )
```
joaquimrocha, line 607: This should be abstracted into a function in apiProxy.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -510,10 +513,47 @@ export interface ContainerInfoProps {
...
+    port: string;
+  } | null>(null);
+  const cluster = getCluster();
+  const namespace = resource?.metadata.namespace;
+
+  React.useEffect(() => {
+    if (!helpers.isElectron()) {
+      return;
+    }
+    fetch(`${helpers.getAppUrl()}/listportforward?cluster=${cluster}`)
```
joaquimrocha, line 536: This should also be abstracted into a function in apiProxy.

joaquimrocha (changes requested):
Thanks for the changes @ashu8912 .
I have finally tried them and unfortunately they don't work for me as I get an error:
`Error: failed to forward ports: error upgrading connection: Unauthorized`

I have also left a couple of comments in the code.

ashu8912:
@joaquimrocha can you mention the steps you followed to run the port forward??

joaquimrocha:
(FTR, responded today in a call)

frontend/src/components/common/Resource/Resource.tsx
ashu8912: We do a show a loader using the portForward object

frontend/src/components/common/Resource/Resource.tsx
ashu8912: updated

frontend/src/components/common/Resource/Resource.tsx
ashu8912: fixed

backend/cmd/headlamp.go
```diff
@@ -355,6 +449,133 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		http.Redirect(w, r, oauthConfig.AuthCodeURL(state), http.StatusFound)
 	}).Queries("cluster", "{cluster}")
 
+	r.HandleFunc("/startportforward", func(w http.ResponseWriter, r *http.Request) {
```
yolossn, line 452: Instead of using separate end points for `starting` and `deleting` portforward use a single endpoint like `/portforward` and use `r.Method()` and if the method is `POST` start a portforward and `GET` to list portforward and `DELETE` to delete the portforward.

yolossn: In `POST` method use the request body to get the request details instead of query params.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+	return l.Addr().(*net.TCPAddr).Port, nil
+}
+
+// nolint: funlen
+func (c *HeadlampConfig) startPortForward(p PortForwardPayload, token string) {
+	ports := []string{fmt.Sprintf(p.Port + ":" + p.TargetPort)}
+	ctxtProxy, ok := c.contextProxies[p.Cluster]
+
+	if !ok {
+		log.Fatalf("cluster %s not found", p.Cluster)
```
yolossn, line 682: don't use `Fatalf` return the error and fail the request, in this case even if the port forward is not setup the endpoint will not return an error.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		return
+	}
+
+	var caData []byte
+
+	var err error
+
+	if caData, err = ctxtProxy.context.cluster.getCAData(); err != nil {
+		log.Fatalf("error getting caData" + err.Error())
```
yolossn, line 691: same, bottle up the error and return the error in the API response.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+	roundTripper, upgrader, err := spdy.RoundTripperFor(rConf)
+	if err != nil {
+		log.Printf("Error: failed to create round tripper: %s", err)
+	}
+
+	requestURL := fmt.Sprintf("%s/api/v1/namespaces/%s/pods/%s/portforward", rConf.Host, p.Namespace, p.Pod)
+
+	reqURL, err := url.Parse(requestURL)
+	if err != nil {
+		log.Panicln("url parse err", requestURL, err)
```
yolossn, line 719: return error, don't panic.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+			KeyData:  ctxtProxy.context.authInfo.ClientKeyData,
+			CertData: ctxtProxy.context.authInfo.ClientCertificateData,
+		},
+	}
+
+	fmt.Println("caData is", caData)
+	fmt.Println("\n\nkeyFile is", ctxtProxy.context.authInfo.ClientKey)
+	fmt.Println("\n\ncertFile is", ctxtProxy.context.authInfo.ClientCertificate)
+	roundTripper, upgrader, err := spdy.RoundTripperFor(rConf)
+	if err != nil {
```
yolossn, line 711: if err is not nil, error has to be returned.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+	reqURL, err := url.Parse(requestURL)
+	if err != nil {
+		log.Panicln("url parse err", requestURL, err)
+	}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, reqURL)
+	stopChan, readyChan := make(chan struct{}), make(chan struct{}, 1)
+	out, errOut := new(bytes.Buffer), new(bytes.Buffer)
+	forwarder, err := portforward.New(dialer, ports, stopChan, readyChan, out, errOut)
+	if err != nil {
```
yolossn, line 726: return error is missed.

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+			Port:      p.Port,
+			Out:       out,
+			ErrOut:    errOut,
+		}
+		portforwardstore(p)
+	}
+
+	clientset, err := kubernetes.NewForConfig(rConf)
+	if err != nil {
+		log.Printf("kpoward: failed to create clientset: %s", err)
```
yolossn, line 757: I don't see `kpoward` being used, is the log out of date?

backend/cmd/headlamp.go
```diff
@@ -437,6 +657,131 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+	}
+	/* check every PodAvailabilityCheckTimer seconds if the pod for which we started a portforward is running
+	if not then we close the channel
+	*/
+	ticker := time.NewTicker(PodAvailabilityCheckTimer * time.Second)
+
+	go func() {
+		for range ticker.C {
+			pod, err := clientset.CoreV1().Pods(p.Namespace).Get(p.Pod, v1.GetOptions{})
+			if err != nil {
```
yolossn, line 767: This will stop the port forward even if Get pod fails due to network issues. Better error matching is required here.

backend/cmd/headlamp.go
yolossn: In this implementation port forward is only supported for `pods` we will have to support `services`.

yolossn (changes requested):

backend/cmd/headlamp.go
```diff
@@ -68,6 +102,45 @@ func resetPlugins() {
 	pluginListURLs = nil
 }
 
+var portForwards = make(map[string][]PortForward)
```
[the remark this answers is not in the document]
ashu8912: updated

backend/cmd/headlamp.go
```diff
@@ -128,6 +202,26 @@ func copyReplace(src string, dst string,
 	}
 }
 
+func validatePortForward(p PortForwardPayload) error {
```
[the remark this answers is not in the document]
ashu8912: updated

backend/cmd/headlamp.go
ashu8912: changed this to /portforward post get and for list /portforward/list and for delete /portforward/delete

backend/cmd/headlamp.go
ashu8912: fixed

backend/cmd/headlamp.go
ashu8912: fixed

backend/cmd/headlamp.go
ashu8912: updated

backend/cmd/headlamp.go
ashu8912: updated

backend/cmd/headlamp.go
ashu8912: updated

backend/cmd/headlamp.go
```diff
@@ -440,6 +689,148 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		Host:        ctxtProxy.context.cluster.config.Server,
+		BearerToken: token,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAData:   caData,
+			KeyFile:  ctxtProxy.context.authInfo.ClientKey,
+			CertFile: ctxtProxy.context.authInfo.ClientCertificate,
+			KeyData:  ctxtProxy.context.authInfo.ClientKeyData,
+			CertData: ctxtProxy.context.authInfo.ClientCertificateData,
+		},
+	}
```
joaquimrocha, line 745: I'd like us to not have to create a config and use all these + dialer, etc. but we can do that later as we talked on the call.

backend/cmd/headlamp.go
```diff
@@ -440,6 +689,148 @@ func StartHeadlampServer(config *HeadlampConfig) {
...
+		requestURL = fmt.Sprintf("%s/api/v1/namespaces/%s/pods/%s/portforward", rConf.Host, p.Namespace, p.Pod)
+	} else if p.Service != "" {
+		svc, err := clientset.CoreV1().Services(p.Namespace).Get(p.Service, v1.GetOptions{})
+		if err != nil {
+			log.Printf("Error: failed to get service: %s", err)
+			return fmt.Errorf("failed to get service %s", p.Service)
+		}
+		pods, err := getPodsForSvc(svc, p.Namespace, clientset.CoreV1())
+		p.Pod = pods.Items[0].Name
+		requestURL = fmt.Sprintf("%s/api/v1/namespaces/%s/pods/%s/portforward", rConf.Host, p.Namespace, pods.Items[0].Name)
```
joaquimrocha, line 769: You could assign the pod name to a var and use this hardcoded request just once.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const { t } = useTranslation();
+
+  if (!helpers.isElectron()) {
+    return null;
+  }
+  function handlePortForward() {
+    if (!namespace || !containerPort || !cluster) {
+      return;
+    }
+    setError(null);
```
joaquimrocha, line 47: Nit: Please add a newline before and after this block.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const massagedResourceName = name || '';
+    const serviceName = !isPod ? massagedResourceName : '';
+    const podName = isPod ? massagedResourceName : '';
+    startPortForward(cluster, namespace, podName, containerPort, serviceName)
```
joaquimrocha, line 51: Nit: Please add a newline before this block.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -0,0 +1,174 @@
...
+          }
+        </Box>
+      )}
+    </Box>
+  ) : (
+    <>
+      {isCurrentPodOrServicePortForwarding &&
+        (portForward ? (
+          <>
+            <Button onClick={portForwardStopHandler}>Stop</Button>
```
joaquimrocha, line 164: This button is not translated.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -0,0 +1,174 @@
...
+    });
+  }, []);
+
+  if (isPod && !isPodRunning) {
+    return null;
+  }
+
+  return !portForward ? (
+    <Box display="flex">
+      <Button onClick={handlePortForward}>{t('frequent|Port Forward')}</Button>
```
joaquimrocha, line 141: This needs to be allow only one click as I could do several clicks before the button disappears which allowed me to create 4 working port forwardings, only one of which (the last one) was visible in the UI.

frontend/src/components/common/Resource/PortForward.tsx
joaquimrocha: These buttons should also get an aria-label as I think it's difficult for screen reader users to understand the whole context without the visual cues.

joaquimrocha (changes requested):

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: updated

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: updated now it shows a loader when we start the portforward

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: fixed

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: fixed

backend/cmd/headlamp.go
ashu8912: fixed

backend/cmd/headlamp.go
ashu8912: I will create an issue to revisit the portforward implementation

backend/cmd/headlamp.go
```diff
@@ -358,6 +472,142 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		if _, err = w.Write(jsonPayload); err != nil {
+			http.Error(w, "failed to write json payload to response write "+err.Error(), http.StatusInternalServerError)
+		}
+	}).Methods("POST")
+
+	r.HandleFunc("/portforward/delete", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ashu8912: updated

illume (changes requested):
Hello,

- Could you please update the PR description? I know this is a long PR already so I guess the description would need to be updated now. A "how to use" section in the PR would be nice, as would a paste of the most current design (if there is one).
- There's a number of open conversations. I guess it's on @joaquimrocha to close them if they've been sufficiently addressed?
- seems there's some CI issues still in there

With those I should be able to test it out and review more easily.

in general...
- functions and interface fields should be documented
- frontend components should have storybook tests
- i18n t() should be used for strings
- backend functions should have tests

ashu8912:
Screenshot 2022-11-14 at 11 01 23 PM
Screenshot 2022-11-14 at 11 01 15 PM

joaquimrocha:
@ashu8912 , I tried the PR and it works fine for the Pod's port forwarding. The Service forwarding was still not working correctly when I tried that named port (non-numeric).

In order to make progress and be more efficient about the feature, I am proposing we drop the Service's port forwarding for now.

I also added a fix up commit that I'd like you to check. This makes the style closer to the latest mockups.

joaquimrocha:
This is reviewed and tested. We should merge after merging its base PR (#843 ).

joaquimrocha:
This was deleted automatically after I merged + deleted the base branch, but it should have been automatically rebased on main...
Maybe this is a bug on Github. Reopening so we merge this.

backend/cmd/headlamp.go
yolossn: This is fixed

PR #653, opened 2022-08-09
Add name for cluster chooser route

illume:
Add name for cluster chooser route

## How to use

Go to / with a cluster chooser.

## Testing done

See the page has a `<title>Choose a cluster</title>`

### DIFF
--- a/frontend/src/lib/router.tsx
+++ b/frontend/src/lib/router.tsx
@@ -84,6 +84,7 @@ const defaultRoutes: {
   chooser: {
     path: '/',
     exact: true,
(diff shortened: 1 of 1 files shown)


frontend/src/lib/router.tsx
```diff
@@ -84,6 +84,7 @@ const defaultRoutes: {
   chooser: {
     path: '/',
     exact: true,
+    name: 'Choose a cluster',
```
joaquimrocha, line 87: We should eventually find a way to get these translated.

joaquimrocha (approved):

PR #660, opened 2022-08-11
frontend: Catch and log Plugin initialize() errors

illume:
So that plugins that use initialize() will not crash the system if they have an error in them.

## Testing done

Introduce an error into a plugin that uses the Plugin `initialize()` method of registration (like plugin/examples/app-menus) and it should not stop other plugins running.

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -81,10 +81,15 @@ export async function initializePlugins() {
 
   // Initialize every plugin in the order they were loaded.
   return new Promise(resolve => {
-    for (const plugin of Object.values(window.plugins)) {
-      // @todo: what should happen if this fails?
-      // @todo: The return code is not checked? What is it for?
-      plugin.initialize(new Registry());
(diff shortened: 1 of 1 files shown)


frontend/src/plugin/index.ts
```diff
@@ -81,10 +81,15 @@ export async function initializePlugins() {
...
   // Initialize every plugin in the order they were loaded.
   return new Promise(resolve => {
-    for (const plugin of Object.values(window.plugins)) {
+    for (const pluginName of Object.keys(window.plugins)) {
       // @todo: what should happen if this fails?
       // @todo: The return code is not checked? What is it for?
-      plugin.initialize(new Registry());
+      const plugin = window.plugins[pluginName];
+      try {
+        plugin.initialize(new Registry());
```
joaquimrocha, line 89: Maybe move the todo comments to the top of this line now?

illume: Good idea. Thanks. Done.

joaquimrocha (approved):

PR #664, opened 2022-08-12
plugins: Add useClusterURL, noAuthRequired to sidebar example

illume:
noCluster has been confusing to a few people. This is an attempt to improve things.

- [x] Add useClusterURL, noAuthRequired to sidebar example.
- [x] Deprecate Route.noCluster in favor of useClusterURL. Old plugins can still use noCluster.
- [x] Fix registerRoute to take a useClusterURL option (instead of noCluster option).
- [x] Add lots of comments to the sidebar example

Why useClusterURL and not noCluster?

- noCluster is confusing because it's a negatively named variable
- useClusterURL is used elsewhere (when registering sidebar entries)
- useClusterURL has URL in the name, so it's a bit more descriptive

## How to use

```bash
cd plugins/examples/sidebar
npm i
npm start
# Then go to Headlamp (make run-frontend).
```

Read sidebar example; does useClusterURL make sense?

Is /no-cluster-link not a 404?

## Testing done

Does useClusterURL option make more sense now?

### DIFF
--- a/frontend/src/lib/router.tsx
+++ b/frontend/src/lib/router.tsx
@@ -61,8 +61,15 @@ export interface Route {
   exact?: boolean;
   /** Human readable name. Capitalized and short. */
   name?: string;
-  /** In case this route does *not* need a cluster prefix and context. */
+  /**
+   * In case this route does *not* need a cluster prefix and context.
+   * @deprecated please use useClusterURL.
+   */
   noCluster?: boolean;
+  /**
+   * Should URL have the cluster prefix? (default=true)
+   */
+  useClusterURL?: boolean;
   /** This route does not require Authentication. */
   noAuthRequired?: boolean;
   /** The sidebar group this Route should be in, or null if it is in no group. */
@@ -86,7 +93,7 @@ const defaultRoutes: {
     exact: true,
     name: 'Choose a cluster',
     sidebar: null,
-    noCluster: true,
+    useClusterURL: false,
     noAuthRequired: true,
     component: () => (
       <Chooser useCover open>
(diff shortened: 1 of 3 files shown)


illume:
registerSidebarEntry should take a `noCluster` option

illume:
registerSidebarEntry takes useClusterURL... which uses the opposite to noCluster. 

I could add the better sounding option for symmetry. Probably a good time to do it before it's widely used.
Keeping the noCluster option for backwards compatibility but deprecating it.

plugins/examples/sidebar/src/index.tsx
```diff
@@ -85,3 +94,28 @@ registerRoute({
...
+  useClusterURL: false, // Do not put "/c/mycluster/" in the URL
+});
+
+// This component rendered at URL: /no-cluster-link
+registerRoute({
+  path: '/no-cluster-link',
+  sidebar: null,
+  name: 'no-cluster-link',
+  exact: true,
+  useClusterURL: false, // This one does not have a "/c/mycluster/" in the URL
```
illume, line 114: You can test this PR is still backwards compatible with the old "noCluster: true" option.

Set "noCluster: true" here.

frontend/src/lib/router.tsx
```diff
@@ -439,11 +446,30 @@ export function getRoute(routeName: string) {
...
+ * @returns true when a cluster URL contains cluster in the URL. eg. /c/minikube/my-url
+ *   false, the URL does not contain the cluster. eg. /my-url
+ */
+export function getRouteUseClusterURL(route: Route): boolean {
+  if (route.useClusterURL === undefined && route.noCluster !== undefined) {
+    console.warn('Route.noCluster is deprecated. Please use route.useClusterURL instead.');
+    return route.noCluster;
+  }
+  if (route.useClusterURL === undefined) {
+    // default is true, so undefined === true.
```
joaquimrocha, line 462: This is why I chose `noCluster` in the past. `undefined` is often used as `false` too, but I agree negative vars, especially boolean ones, are confusing.

joaquimrocha (approved):

plugins/examples/sidebar/src/index.tsx
ashu8912: @illume what's the difference b/w noCluster and useClusterURL?? Does the noCluster option also doesn't allow to put a url without "/c/mycluster" ??

illume: noCluster was the name of the option previously. In this PR noCluster is being deprecated.

```ts
useClusterURL: true IS noCluster: false
useClusterURL: false IS noCluster: true
```

Now registerRoute, and registerSidebarEntry both use the useClusterURL option.

ashu8912: got it thanks, also For a view that is not the normal /c/mycluster should we also allow an option to not show the appbar??

illume: Do you need that option?

illume: Ashu added an issue for no-app-bar https://github.com/kinvolk/headlamp/issues/681

PR #666, opened 2022-08-12
frontend: Ensure timeAgo always uses the date + 90 days when testing

joaquimrocha:
We need to compare static dates when testing, otherwise snapshots
become outdated.


illume (approved):
🎉 magic. Thank you.

PR #669, opened 2022-08-17
frontend: Fix MetadataDisplay to skip ResourceClasses

illume:
Ones that we can't get a routeName from skip.

## How to use

See https://github.com/kinvolk/headlamp/issues/649

## Testing done

...

### DIFF
--- a/frontend/src/components/common/Resource/MetadataDisplay.tsx
+++ b/frontend/src/components/common/Resource/MetadataDisplay.tsx
@@ -45,21 +45,35 @@ export function MetadataDisplay(props: MetadataDisplayProps) {
       return undefined;
     }
 
-    return ownerReferences.map((ownerRef, i) => (
-      <>
-        {ownerRef.kind in ResourceClasses ? (
-          <Link
-            routeName={new ResourceClasses[ownerRef.kind]().detailsRoute}
-            params={{ name: ownerRef.name, namespace: resource.metadata.namespace }}
-          >
-            {ownerRef.kind}: {ownerRef.name}
-          </Link>
-        ) : (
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -45,21 +45,35 @@ export function MetadataDisplay(props: MetadataDisplayProps) {
...
-        {i < numItems - 1 && <br />}
-      </>
-    ));
+    return ownerReferences
+      .map((ownerRef, i) => {
+        if (ownerRef.kind in ResourceClasses) {
+          let routeName;
+          try {
+            routeName = new ResourceClasses[ownerRef.kind]().detailsRoute;
+          } catch (e) {
```
joaquimrocha, line 54: An error shouldn't occur here, AFAIU. Or is there something I am missing?

joaquimrocha (approved):
Looks like a nice change. We do have to think about what we can do for unknown resources like in this case. This is a good solution for the owner ref.
I left a comment about the exception, but if it really makes sense to have it, then it's fine for me.

frontend/src/components/common/Resource/MetadataDisplay.tsx
illume: My guess is `ResourceClasses[ownerRef.kind]` isn't callable, so `new ResourceClasses[ownerRef.kind]()` fails.

joaquimrocha: Okay. It should be callable as that's what we're setting those to be. But I am fine using it defensively here.

PR #671, opened 2022-08-18
Fix frontend version

joaquimrocha:
Somehow the code to generate the .env was not included and thus the version dialog was not getting correctly filled.
This PR adds those changes + a test.

Testing:
- [ ] Ensure there's no frontend/.env file, then run `make run frontend` and check the UI to see if the version is correctly filled in
- [ ] Ensure there's no frontend/.env file, then run `make frontend` and check with the version dialog with the app

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,12 @@ RUN cd ./backend && go build -o ./headlamp-server ./cmd/
 
 # Keep npm install separated so source changes don't trigger install
 FROM base-build AS frontendinstall
+
+# We need .git and app/ in order to get the version and git version for the frontend/.env file
+# that's generated when building the frontend.
+COPY ./.git /headlamp/.git
+COPY app/package.json /headlamp/app/package.json
+
 COPY frontend/package*.json /headlamp/frontend/
 COPY frontend/patches/* /headlamp/frontend/patches/
 WORKDIR /headlamp
@@ -29,6 +35,9 @@ COPY ./frontend /headlamp/frontend
 WORKDIR /headlamp
 
 RUN cd ./frontend && npm run build
+
+RUN echo "*** Built Headlamp with version: ***"
+RUN cat ./frontend/.env
 
 # Backwards compatibility, move plugin folder to only copy matching plugins.
 RUN mv plugins plugins-old || true
--- a/frontend/make-env.js
+++ b/frontend/make-env.js
@@ -1,0 +1,28 @@
+'use strict';
+
+// Creates the .env file
+
+const fs = require('fs');
+const {execSync} = require('child_process');
+const appInfo = require('../app/package.json');
+
+const gitVersion = execSync('git rev-parse HEAD').toString().trim();
+
+const envContents = {
+  REACT_APP_HEADLAMP_VERSION: appInfo.version,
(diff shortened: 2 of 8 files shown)


frontend/package.json
```diff
@@ -86,8 +86,8 @@
     "xterm-addon-fit": "^0.5.0"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "if-env PUBLIC_URL react-scripts build || cross-env PUBLIC_URL=./ react-scripts build --max_old_space_size=768 && rimraf build/frontend/index.baseUrl.html",
+    "start": "npm run make-version; react-scripts start",
```
illume, line 89: Maybe && here would be better.

There's a pre hook feature... See here: https://docs.npmjs.com/cli/v6/using-npm/scripts

So prestart and prebuild. But I don't think this works here because it's setting environment variables?

illume (commented):
Cool. Left a comment about using &&

frontend/package.json
joaquimrocha: Cool. I will add it to pre. The `&&` I am not sure it works on Windows.

ashu8912 (approved):
Tested, LGTM!!!

frontend/make-env.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
```
illume, line 2: Was thinking... is this script needed for development? Or only on CI?

If only on CI, it could exit early and not write the .env file... by checking if the "`CI`" env var is set.

frontend/package.json
illume: `&&` works with cmd and powershell

frontend/make-env.js
joaquimrocha: It's needed in both.

Dockerfile
```diff
@@ -18,6 +18,10 @@ RUN cd ./backend && go build -o ./headlamp-server ./cmd/
 
 # Keep npm install separated so source changes don't trigger install
 FROM base-build AS frontendinstall
+
+# We need .git in order to get the git version for the frontend/.env file
+COPY ./.git /headlamp/.git
```
illume, line 23: Seems a lot for just the version.

Dockerfile
```diff
@@ -18,6 +18,10 @@ RUN cd ./backend && go build -o ./headlamp-server ./cmd/
...
+COPY app /headlamp/app
```
illume, line 24: I'm curious why the app is needed in the image.

Dockerfile
joaquimrocha: I tried copying just a few dirs/files, but git can only fully work if it's got some minimum structure, and it's not possible unless I copy everything AFAIU.
I could run the version command from outside and pass the version to the Dockerfile, but then it's dependent on running the Dockerfile with an external command, not sure it's great for orthogonality.

illume: I guess it won't slow down the workflow so much. My .git is only 64MB so far, and only takes 1s to copy with a cold cache/0.2s for a warm file cache. 

We can address this later if needed IMHO.

illume (approved):
🎉

PR #683, opened 2022-08-26
app: Add Windows signing

joaquimrocha:
## Testing done

I will add instructions separately since this is an admin-only action.

### DIFF
--- a/.github/workflows/app-artifacts-win.yml
+++ b/.github/workflows/app-artifacts-win.yml
@@ -7,14 +7,23 @@ on:
         description: 'Headlamp ref/branch/tag'
         required: true
         default: 'main'
+      signBinaries:
+        description: Sign the binaries
+        default: false
+        type: boolean
 
 jobs:
   build-windows:
-    runs-on: windows-2019
+    runs-on: windows-2022
     steps:
     - uses: actions/checkout@v2.3.3
       with:
+        path: build-helper
+    - uses: actions/checkout@v2.3.3
+      with:
+        repository: kinvolk/headlamp
         ref: ${{ github.event.inputs.buildBranch }}
+        path: headlamp
     - name: Setup nodejs
       uses: actions/setup-node@v1
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/app-artifacts-win.yml
```diff
@@ -26,14 +35,43 @@ jobs:
       uses: crazy-max/ghaction-chocolatey@v1
       with:
         args: install make
+    - name: Azure CLI script
```
illume, line 38: suggestion: `name: Download certificates from Azure keyvault`

Link(s) to the relevant documentation sections could be helpful.

app/windows/codesign.js
```diff
@@ -0,0 +1,166 @@
...
+    .map(dir => filesJson(dir, files[dir]))
+    .filter(f => f.SignRequestFiles.length > 0);
+
+  const filePath = path.join(os.tmpdir(), fileName);
+  fs.writeFileSync(filePath, JSON.stringify(SIGN_JSON_TEMPLATE, undefined, 2));
+
+  return filePath;
+}
+
+/** Signs the given file, or all files in a given directory if that's what's passed to it.
```
illume, line 129: nit, newline before Signs.

app/windows/codesign.js
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        files[dirName].push(file);
+      }
+    });
+  }
+  return files;
+}
+
+function createJson(pathToSign, fileName = 'test_SignInput.json') {
```
illume, line 68: Can we include what type of json structure in the name? Or add a function comment block to describe it.

Also, can you please include a URL to the docs for this json structure?

app/windows/codesign.js
```diff
@@ -0,0 +1,166 @@
...
+        console.log(`stderr: ${stderr}`);
+        return;
+      }
+      console.log(`stdout: ${stdout}`);
+    }
+  );
+}
+
+module.exports = {
+  createJson,
```
illume, line 164: nit: I guess createJson doesn't need to be exported.

app/windows/sign.js
```diff
@@ -0,0 +1,23 @@
+const { sign } = require('./codesign');
+
+exports.default = async function (configuration) {
+  const esrpTool = process.env.ESRP_PATH;
+  if (!esrpTool) {
+    // Ignore if the signing tool is not available.
```
illume, line 6: nit: Probably a warning here would be good.

``console.log(`Warning: exiting because esrpTool doesn't exist at: ${esrpTool}`)``

app/windows/codesign.js
```diff
@@ -0,0 +1,166 @@
+const { exec } = require('child_process');
+const path = require('path');
+const os = require('os');
+const fs = require('fs');
+
+const SIGN_JSON_TEMPLATE = {
+  Version: '1.0.0',
+  DriEmail: ['joaquim.rocha@microsoft.com'],
```
illume, line 8: I'm not sure if this should be a personal email?

app/windows/codesign.js
```diff
@@ -0,0 +1,166 @@
...
+
+function getFileList(rootDir) {
+  let files = {};
+  let dirs = ['.'];
+  while (dirs.length > 0) {
+    const dirName = dirs.shift();
+    const curDir = path.join(rootDir, dirName);
+
+    fs.readdirSync(curDir).forEach(file => {
+      if (file === 'node_modules') {
```
illume, line 48: Any other folders to skip? I guess .git isn't in here.

illume (commented):
🎉 Nice one.
Left a few questions and comments.

I tried to find out where this was being run in the Checks... and looked in the app windows one, and couldn't find logs for it. Where do I see the output being run?

app/windows/codesign.js
joaquimrocha: There are no public docs to this. So I will add them in our internal guides to building signed binaries.

app/windows/codesign.js
joaquimrocha: I added it since I thought it was helpful later. But I can remove it for now.

app/windows/sign.js
joaquimrocha: Thing is that this is called per file, so we'd see dozens of such warnings. This is purposedly silent as users may not have the tool around.

app/windows/codesign.js
joaquimrocha: This is the contact. I guess I can add hello at kinvolk.

app/windows/codesign.js
joaquimrocha: Right. This may require fine tuning later. For now the signing goes file by file, but it's useful to be able to generate the signing for an entire dir like this.

app/windows/codesign.js
illume: Cool thanks. I'll have a look there.

app/windows/sign.js
illume: Ok, understood. I guess the action will fail if the tool isn't downloaded/there, so we'd be aware of that.

joaquimrocha: The action has an option for using signing or not. But if it's required to use and there's no tool, it shall fail.

PR #688, opened 2022-08-29
headlamp-plugin upgrade improvements - check and audit after upgrade, fail better, README

illume:
headlamp-plugin upgrade improvements - check after upgrade, fail better, README.

- Check after upgrade: run lint, tsc, and such after upgrading.
- Run npm audit --fix after upgrading.
- Add the upgrade command to the README.md
- Turn on the type checker so it type checks headlamp-plugin.js

## How to use

Run `node ./plugins/headlamp-plugin/bin/headlamp-plugin.js upgrade plugins/examples/pod-counter` or on another package.

## Testing done

Ran on some existing plugins, and there's a test in CI for the upgrade command already.

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-
+// @ts-check
 'use strict';
 
 const webpack = require('webpack');
@@ -20,7 +20,7 @@ const headlampPluginPkg = require('../package.json');
  * Then runs npm install inside of the folder.
  *
  * @param {string} name - name of package and output folder.
- * @param {bool} link - if we link @kinvolk/headlamp-plugin for testing
+ * @param {boolean} link - if we link @kinvolk/headlamp-plugin for testing
  * @returns {0 | 1 | 2 | 3} Exit code, where 0 is success, 1, 2, and 3 are failures.
  */
 function create(name, link) {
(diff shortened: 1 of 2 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -498,50 +509,89 @@ function upgrade(packageFolder, skipPackageUpdates) {
...
-
-    folders.forEach(folder => {
-      const folderToBuild = path.join(packageFolder, folder.name);
-      if (!upgradePackage(folderToBuild, skipPackageUpdates)) {
-        console.error(`"${folderToBuild}" does not contain a package. Not upgrading.`);
+  function upgradeFolderOfPackages(packageFolders, skipPackageUpdates) {
+    let failed = '';
+
+    for (const folder of packageFolders) {
+      if (failed) {
```
ashu8912, line 556: umm, I think we here are testing failed value for each next iteration of the loop, will this work if there were no failures till the second last packageFolders but the last one was the failed one??

ashu8912 (commented):
Just one comment, otherwise this looks and works like charm, thanks.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Good question. 

For the case of last one failed:

- It will return false, and shows a message about the package with an error. The log is emitted inside upgradePackage(...) when there is a failure.
- It will Not show a message about skipping any, because none will be skipped.

ashu8912: aah i see yupp this makes sense, thanks

ashu8912 (approved):
OOps, this was already tested forgot to press approve

illume:
Thanks @ashu8912

PR #693, opened 2022-08-31
plugins: Add registerRouteFilter and registerSidebarEntryFilter

### LINKED ISSUE #445: [RFE] Allow to remove default capabilities/routes
## Current situation

We allow to create and even replace routes through plugins, but we don't allow to remove them.

## Impact

There may be vendors that do not want to offer a certain route or entry in the sidebar.

## Ideal future situation

We have a way to remove a route and entry in the sidebar. Maybe also other features, like default header actions, sections, etc.

## Implementation options

One way would be to allow "deregistering" routes/sections/etc. But this is TBD. Suggestions are welcome.

illume:
Add registerRouteFilter and registerSidebarEntryFilter,
which allows plugins to remove routes, and sidebar menu items.

```tsx
import { registerRouteFilter, registerSidebarEntryFilter } from '@kinvolk/headlamp-plugin/lib';

// Remove workloads top level sidebar item
registerSidebarEntryFilter(entry => (entry.name === 'workloads' ? null : entry));
// Remove workloads route
registerRouteFilter(route => (route.path === '/workloads' ? null : route));
```

Closes #445

## How to use

See plugins/examples/sidebar

```bash
cd plugins/headlamp-plugin
npm run build && npm run pack
cd ../examples/sidebar
npm i ../../headlamp-plugin/*.tgz
npm start
# run headlamp to see the Workloads+Namespaces routes and sidebar items gone.
```

## Testing done

- [x] top level sidebar item is removed
- [x] second level sidebar item removed
- [x] route removed
- [x] works with no filters active.

### DIFF
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -122,6 +122,7 @@ function prepareRoutes(t: (arg: string) => string) {
   ];
 
   const items = store.getState().ui.sidebar.entries;
+  const filters = store.getState().ui.sidebar.filters;
   // @todo: Find a better way to avoid modifying the objects in LIST_ITEMS.
   const routes: SidebarItemProps[] = JSON.parse(JSON.stringify(LIST_ITEMS));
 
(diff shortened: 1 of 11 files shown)


frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -122,11 +122,13 @@ function prepareRoutes(t: (arg: string) => string) {
   ];
 
   const items = store.getState().ui.sidebar.entries;
+  const filters = store.getState().ui.sidebar.filters;
   // @todo: Find a better way to avoid modifying the objects in LIST_ITEMS.
   const routes: SidebarItemProps[] = JSON.parse(JSON.stringify(LIST_ITEMS));
 
   for (const i of Object.values(items)) {
     const item = _.cloneDeep(i);
+
```
ashu8912, line 131: small nitpick, this empty line was not there earlier

illume: Thanks. This is fixed.

PR #701, opened 2022-09-12
Build: Create gh action to create PR

yolossn:
This patch creates a PR to update homebrew and minikube versions. The GitHub action will get triggered once the [Publish Container](https://github.com/kinvolk/headlamp/blob/main/.github/workflows/container-publish.yml) workflow completes, this ensures that the container used in minikube is the latest one. Refer to the following example PRs created by the workflow
https://github.com/yolossn/minikube/pull/4
https://github.com/yolossn/homebrew-cask/pull/5

To Dos

- [x] Create a fork of [Homebrew/homebrew-cask](https://github.com/Homebrew/homebrew-cask)
- [x] Create a GH token with the following access and add to secrets under the key `KINVOLK_REPOS_TOKEN`
- [x] Right now the PRs are in draft and assigned to the github actor who initiated the action, have to make decide if the PRs should be left in draft or not and add compulsory assignees.

image

### DIFF
--- a/.github/workflows/pr-to-update-homebrew.yml
+++ b/.github/workflows/pr-to-update-homebrew.yml
@@ -1,0 +1,86 @@
+name: PR to update homebrew
+
+# This action will run after a tag starting with "v" is published
+on:
+  push:
+    tags:
+    - 'v*'
+  workflow_dispatch:
+
+env:
+  LATEST_HEADLAMP_TAG: latest
+jobs:
+  create_pr_to_upgrade_homebrew:
+    name: Create PR to upgrade homebrew
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout headlamp repo 
+        uses: actions/checkout@v2 
+        with:
+          path: headlamp
+          fetch-depth: 0
+      - name: Configure Git 
+        run: | 
+          user=${{github.actor}}
+          if [ -z $user ]; then
+            user=yolossn
+          fi
+          git config --global user.name "$user"
+          git config --global user.email "$user@users.noreply.github.com"
+      - name: Get headlamp latest tag 
+        run: |
+          cd headlamp 
+          latestTag=$(git tag --list --sort=version:refname 'v*' | tail -1)
+          echo "LATEST_HEADLAMP_TAG=$latestTag" >> $GITHUB_ENV
+          echo $latestTag
+      - name: Sync homebrew-cask fork from upstream
+        run: |
+          gh repo sync kinvolk/homebrew-cask
+        env:
+          GITHUB_TOKEN: ${{ secrets. KINVOLK_REPOS_TOKEN }}
+      - name: Check out homebrew-cask repo 
+        uses: actions/checkout@v2 
+        with: 
+          repository: kinvolk/homebrew-cask
+          path: homebrew-cask
+          token: ${{ secrets. KINVOLK_REPOS_TOKEN }}
+      - name: Update headlamp version in homebrew-cask
+        run: |
+          user=${{github.actor}}
+          if [ -z $user ]; then
+            user=yolossn
+          fi
+          cd homebrew-cask
+          HEADLAMP_VERSION=${LATEST_HEADLAMP_TAG:1}
+          wget "https://github.com/kinvolk/headlamp/releases/download/$LATEST_HEADLAMP_TAG/checksums.txt"
+          ARM_SHA=$(cat checksums.txt | grep .arm64.dmg | awk -F" " '{print $1}')
+          INTEL_SHA=$(cat checksums.txt | grep .x64.dmg | awk -F" " '{print $1}')
+          git checkout -b "update_headlamp_$HEADLAMP_VERSION"
+          sed -i "s/version\ .*/version \"$HEADLAMP_VERSION\"/g" ./Casks/headlamp.rb
+          if [ $ARM_SHA ]; then
+            echo "replacing ARM SHA"
+            sed -i "s/sha256 arm:\   \".*/sha256 arm:\   \"$ARM_SHA\",/g" ./Casks/headlamp.rb
+          fi
+          if [ $INTEL_SHA ]; then
+            echo "replacing Intel SHA"
+            sed -i "s/  intel:\ \".*/  intel:\ \"$INTEL_SHA\"/g" ./Casks/headlamp.rb
+          fi
+          git diff
+          rm ./checksums.txt
+          git add ./Casks/headlamp.rb
+          git status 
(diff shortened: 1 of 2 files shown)


.github/workflows/create-pr-to-update-release-version.yml
```diff
@@ -0,0 +1,108 @@
...
+    steps:
+      - name: Checkout headlamp repo 
+        uses: actions/checkout@v2 
+        with:
+          path: headlamp
+          fetch-depth: 0
+      - name: Configure Git 
+        run: | 
+          git config --global user.name 'yolossn'
+          git config --global user.email 'sannagaraj@microsoft.com'
```
joaquimrocha, line 25: Not sure if we can infer the user from the action triggering. This way, whoever releases the version will be the one who's getting the user as the PR author. Not super important but just a thought for helping with trailing the actual requests.

.github/workflows/create-pr-to-update-release-version.yml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          cd headlamp 
+          latestTag=$(git tag --list --sort=version:refname 'v*' | tail -1)
+          echo "LATEST_HEADLAMP_TAG=$latestTag" >> $GITHUB_ENV
+          echo $latestTag
+      - name: Sync homebrew-cask fork from upstream
+        run: |
+          gh repo sync kinvolk/homebrew-cask
+        env:
+          GITHUB_TOKEN: ${{ secrets.TOKEN }}
```
joaquimrocha, line 36: This could also be used as the token of the user who's triggered the request AFAIU. (I'm not entirely sure this works for tag-triggered actions but maybe it's worth a shot)
https://docs.github.com/en/actions/security-guides/automatic-token-authentication

.github/workflows/create-pr-to-update-release-version.yml
```diff
@@ -0,0 +1,108 @@
+name: PR to update homebrew and minikube
+
+# This action will run after the "Publish container" is succesfully completed
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
+on:
+  workflow_run:
+    workflows: ["Publish container"]
+    types:
+      - completed
```
joaquimrocha, line 9: Nice!

joaquimrocha (commented):
Thanks a lot for this PR. I left a couple of comments regarding the authentication/authorship which can be further automated IMO, but I am not entirely sure this is the case.

.github/workflows/create-pr-to-update-release-version.yml
yolossn: The token that is automatically added doesn't work because it has the scope particular to this repository alone and we need a token which access other repositories too. Using the token that is automatically added returns an unauthorised error.

.github/workflows/create-pr-to-update-release-version.yml
yolossn: Only the username is available from the github context, but this step requires both username and email address.
[Refer](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

.github/workflows/create-pr-to-update-release-version.yml
joaquimrocha: Okay. Then let's call the secret KINVOLK_REPOS_TOKEN or something more particular, since we have already many secrets.

yolossn: Done 👍🏽

.github/workflows/create-pr-to-update-release-version.yml
joaquimrocha: Looks like a no-reply GH email can be easily retrieved using the GH rest API.
My insistence on this is that if I create a token for `KINVOLK_REPOS_TOKEN`, I think it gets associated with me. So either we use my own info or we go with the dynamically filled user name + email.

I know you've replied to why we'd need to use a separate token (permissions), but apparently there's a way we can tweak the GITHUB_TOKEN's permissions, e.g.:
https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp

Have you looked into this? If you have, let me know, otherwise I think it's worth pursuing, so we can have the user's associated token being used instead of one associated with another user.

yolossn: I checked and couldn't find a way to tweak the GITHUB_TOKEN to access other repos in the organisation, as the token is scoped to this particular repository. For configuring git to use the initiator the GitHub API returns the default email in my case it returns my personal email id. Will it be okay to signoff the commit using personal email ?

joaquimrocha (commented):
@yolossn , I have just noticed that this updates the release tag for the homebrew but it won't get the checksums (for the 2 architectures) of the Mac OS apps.
So maybe this bit needs to be in too (and for that, we cannot just rely on the fact that the container action has run, but rather that the Mac OS binaries are published).

yolossn:
> @yolossn , I have just noticed that this updates the release tag for the homebrew but it won't get the checksums 

Good catch! I will better split this action into two separate ones (minikube and homebrew) and find a way to fetch and update the checksums too.

joaquimrocha (changes requested):
When the Homebrew action is called, there will be no checksums.txt yet because I create the draft and only add the checksums after all assets are uploaded.
I usually only publish the tag after I have everything ready, so I'd say let's trigger this on v* tags instead.

yolossn:
> When the Homebrew action is called, there will be no checksums.txt yet because I create the draft and only add the checksums after all assets are uploaded. I usually only publish the tag after I have everything ready, so I'd say let's trigger this on v* tags instead.

Updated

joaquimrocha (approved):
LGTM. We should later add the ability to manually trigger this PR, in case it every fails (for example, if we publish the checksums.txt and hold on publishing one of the binaries).

.github/workflows/pr-to-update-homebrew.yml
```diff
@@ -0,0 +1,82 @@
+name: PR to update homebrew
+
+# This action will run after the "Publish container" is succesfully completed
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
```
joaquimrocha, line 4: Actually just noticed that this description is no longer accurate.

joaquimrocha (changes requested):

.github/workflows/pr-to-update-homebrew.yml
yolossn: Fixed it, have to complete the ToDo list in the PR before merging.

yolossn:
Updated the comments and added manual trigger

PR #702, opened 2022-09-12
Add cluster name to notification area

ashu8912:
Adds cluster name to each notification list item in the popup

## Testing done
Go to notification popup and you will find cluster name on each list item

### DIFF
--- a/frontend/src/components/App/Notifications/index.tsx
+++ b/frontend/src/components/App/Notifications/index.tsx
@@ -71,6 +71,7 @@ function NotificationsList(props: {
   const classes = useStyles();
   const history = useHistory();
   const theme = useTheme();
+  const config = useTypedSelector(state => state.config);
 
   if (!notifications || notifications.length === 0) {
     return <Empty>{t(`notifications|You don't have any notifications right now`)}</Empty>;
@@ -132,7 +133,21 @@ function NotificationsList(props: {
             </Grid>
           )}
           <Grid item md={12}>
-            <DateLabel date={notification.date} />
+            <Box display={'flex'} alignItems="center">
+              {Object.entries(config?.clusters || {}).length > 1 && notification.cluster && (
+                <Box
+                  border={1}
+                  p={0.5}
+                  mr={1}
+                  textOverflow="ellipsis"
+                  overflow={'hidden'}
+                  whiteSpace="nowrap"
+                >
+                  {notification.cluster}
+                </Box>
+              )}
+              <DateLabel date={notification.date} />
+            </Box>
           </Grid>
         </Grid>
       </ListItem>
@@ -186,7 +201,7 @@ export default function Notifications() {
 
           const message = event.message;
           const date = new Date(event.metadata.creationTimestamp).getTime();
-          const notification = new Notification(message, date);
+          const notification = new Notification({ message, date });
           notification.id = event.metadata.uid;
           notification.url = createRouteURL('cluster') + `?eventsFilter=${notification.id}`;
 
--- a/frontend/src/lib/notification.tsx
+++ b/frontend/src/lib/notification.tsx
@@ -1,5 +1,17 @@
 import { setUINotifications } from '../redux/actions/actions';
 import store from '../redux/stores/store';
+import { getCluster } from './util';
+
+interface NotificationOptions {
+  message?: string;
+  date?: number | string | Date;
+  cluster?: string | null;
+}
+
+/** @deprecated use ${NotificationOptions} instead */
+type NotificationMessageString = string;
+/** @deprecated use ${NotifictionOptions} instead */
+type OldNotificationDateArg = number | string;
 
 export class Notification {
   message: string = '';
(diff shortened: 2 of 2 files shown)


frontend/src/lib/notification.tsx
```diff
@@ -8,14 +9,18 @@ export class Notification {
   url?: string;
   date: number | string = new Date().getTime();
   deleted: boolean = false;
+  cluster: string | null = getCluster();
 
-  constructor(message?: string, date?: number | string) {
+  constructor(message?: string, date?: number | string, cluster?: string | null) {
```
joaquimrocha, line 14: Since all these are optional, maybe it's a good idea to make them an options object instead of parameters.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -132,7 +132,12 @@ function NotificationsList(props: {
             </Grid>
           )}
           <Grid item md={12}>
-            <DateLabel date={notification.date} />
+            <Box display={'flex'} alignItems="center">
+              <Box border={1} p={0.5} mr={1}>
+                {notification.cluster}
+              </Box>
```
joaquimrocha, line 138: The cluster name should only be shown if we have more than one cluster available.

joaquimrocha (changes requested):
LGTM but I think we can improve things slightly.
Also, please provide a screenshot of the resulting UI.

frontend/src/lib/notification.tsx
ashu8912: updated

frontend/src/components/App/Notifications/index.tsx
ashu8912: fixed

ashu8912:
Screenshot 2022-09-14 at 3 09 57 PM

joaquimrocha:
Hi @ashu8912 , I have noticed the following issues:
1. The name is missing in some cases (I think maybe these are notifications that existed already before, if this is the case, then the name box should not be shown at all):
Screenshot
2. We have to limit/elipsize the cluster's name, otherwise it may push the date beyond visible:
Screenshot2

joaquimrocha (changes requested):
Now it's works great!
I noticed however that Notification is exported for plugins to use, so we should make sure we still support the previous constructor but with a deprecated notice when used.
(Look for "deprecate" in the source code to see examples if needed, or let me know if you have questions)

frontend/src/lib/notification.tsx
```diff
@@ -8,15 +9,22 @@ export class Notification {
...
+      if (cluster) {
+        this.cluster = cluster;
+      }
+      // generate the id based on the message and the date attached to a notification
     }
-    if (date) {
-      this.date = date;
-    }
-    // generate the id based on the message and the date attached to a notification
     this.id = btoa(unescape(encodeURIComponent(`${this.date},${this.message}`)));
```
illume, line 28: Would cluster be needed in making the ID too?

joaquimrocha: Good catch. I think it's a good idea.

ashu8912: updated

frontend/src/lib/notification.tsx
```diff
@@ -1,5 +1,17 @@
 import { setUINotifications } from '../redux/actions/actions';
 import store from '../redux/stores/store';
+import { getCluster } from './util';
+
+interface NotificationOptions {
+  message?: string;
+  date?: number | string;
```
joaquimrocha, line 7: Let's make this `number | string | Date` so we cover everything Date takes.

frontend/src/lib/notification.tsx
```diff
@@ -1,5 +1,17 @@
...
+  cluster?: string | null;
+}
+
+/**  @deprecated use ${NotificationOptions} instead */
```
joaquimrocha, line 11: You have an extra space after ** or a missing one in line 13.

frontend/src/lib/notification.tsx
```diff
@@ -8,16 +20,36 @@ export class Notification {
...
-    }
-    if (date) {
-      this.date = date;
+  constructor(
+    messageOrOptions?: NotificationOptions | NotificationMessageString,
+    date?: OldNotificationDateArg
+  ) {
+    if (typeof messageOrOptions === 'string') {
+      console.warn(
+        `Notification constructor with string is deprecated use NotificationOptions as args instead`
```
joaquimrocha, line 31:
```suggestion
        `Notification constructor with a string arg is deprecated. Please use NotificationOptions as args instead`
```

frontend/src/lib/notification.tsx
```diff
@@ -8,16 +20,36 @@ export class Notification {
...
+      if (message) {
+        this.message = message;
+      }
+      if (date) {
+        this.date = date;
+      }
+      if (cluster) {
+        this.cluster = cluster;
+      }
+      // generate the id based on the message and the date attached to a notification
```
joaquimrocha, line 50: This should be in the next line, outside of the if blocks.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -132,7 +133,24 @@ function NotificationsList(props: {
             </Grid>
           )}
           <Grid item md={12}>
-            <DateLabel date={notification.date} />
+            <Box display={'flex'} alignItems="center">
+              {config &&
+                config.clusters &&
+                Object.entries(config.clusters).length > 1 &&
```
joaquimrocha, line 139: Can all this be condensed as
```typescript
Object.entries(config?.clusters || {}).length > 1
```

joaquimrocha (commented):
Works well now. I have a couple of comment just to tidy up the code a bit.

Issue #654, opened 2022-08-10
[RFE] Ensure the cluster button is not extended depending on the cluster name


joaquimrocha:
## Current situation

If a cluster has a very long name, it will show up in the cluster button with that name, making the cluster button very wide beyond what's reasonable. E.g.:
![Screenshot with an example of the above](https://user-images.githubusercontent.com/1029635/183942770-ff4e54f3-d558-45c0-8d6d-86003d5b1576.png)

## Impact

This makes the UX very poor around the cluster chooser.

## Ideal future situation

We limit this button's label (and the cluster chooser dialog's buttons) so the cluster name has some ellipses + there's a titletooltip showing the actual cluster's name.

PR #703, opened 2022-09-14
frontend: Fix cluster button extending issue

yolossn:
Fixes: #654

Before:
image

After:
image

Before:
image

After:
image

### DIFF
--- a/frontend/src/components/cluster/ClusterChooser.tsx
+++ b/frontend/src/components/cluster/ClusterChooser.tsx
@@ -16,6 +16,13 @@ const useClusterTitleStyle = makeStyles(theme => ({
     '&:hover': {
       color: theme.palette.text.primary,
     },
+    maxWidth: '20em',
+  },
+  clusterName: {
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    display: 'block',
   },
 }));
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/cluster/ClusterChooser.tsx
```diff
@@ -25,7 +32,11 @@ export default function ClusterChooser({ clickHandler, cluster }: ClusterChooser
 
   return (
     <Button size="large" variant="contained" onClick={clickHandler} className={classes.button}>
-      <Trans t={t}>Cluster: {{ cluster }}</Trans>
+      <Trans t={t}>
+        <span className={classes.clusterName} title={cluster}>
+          Cluster: {{ cluster }}
+        </span>
+      </Trans>
```
joaquimrocha, line 39: AFAIU, the Trans function should be in the span and wrapping just the text. Otherwise we'll get the tags in the translation and it's not needed because they wrap the whole phrase.
In other words, tags are fine within `<Trans>` if they're not just wrapping the full text, eg.:
`<Trans>My not bold word and my <b>bold word</b>!</Trans>`

joaquimrocha (changes requested):
Nice changes! Left just a small comment. Once that's done, it's good to go.

joaquimrocha (approved):

Issue #696, opened 2022-09-06
[RFE] Consider to get config from environment variables


blobor:
## Current situation

Headlamp backend service expects configuration to be passed through arguments to the container, which is mostly fine until we have a case for passing secrets. Specifically, I'm talking about the `clientSecret` parameter.

```yaml
      containers:
        - name: {{ .Chart.Name }}
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}
          image: "{{ .Values.image.registry}}/{{ .Values.image.repository }}:{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}"
          imagePullPolicy: {{ .Values.image.pullPolicy }}
          args:
            - "-in-cluster"
            {{- with .Values.config.oidc.clientSecret }}
            - "-oidc-client-secret={{ . }}"
```

As a "workaround" (or a feature), we can [pass arguments from environment variables](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments), which could be loaded from secrets.

## Ideal future situation

It would be great if we had an option to pass the configuration straight from environment variables or could be combined with arguments.

```yaml
      containers:
        - name: {{ .Chart.Name }}
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}
          image: "{{ .Values.image.registry}}/{{ .Values.image.repository }}:{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}"
          imagePullPolicy: {{ .Values.image.pullPolicy }}
          args:
            - "-in-cluster"
          env:
            - name: OIDC_CLIENT_SECRET
               valueFrom:
                 secretKeyRef:
                   name: some-random-secure-store
                   key: headlampClientSecret
```

## Implementation options

As I see from a code, Headlamp uses a built-in `flags` package.
As an option, it could use [github.com/jessevdk/go-flags](https://pkg.go.dev/github.com/jessevdk/go-flags) package, which can load config from env variables and flags passed.
Or, good old [Viper package](github.com/spf13/viper)

illume:
Yeah, good idea. I agree env vars would be nice to support... especially for secrets as you say.

illume:
Hrmm.

I guess rather than redoing all the flags with environment variables and a different library,
we could instead add that single clientSecret environment variable.

joaquimrocha:
I think it's a good idea to use a different flags package that supports both params and env vars. Let's do this.

PR #706, opened 2022-09-16
backend: Add env support for config

yolossn:
Fixes: https://github.com/kinvolk/headlamp/issues/696


yolossn:
The build is failing because the latest version of `golangci-lint` uses io/fs which is available only in versions above 1.16, we will have to upgrade the go version of the project or use pre built binary of `golangci-lint` instead of building it from source? 
Upgrading the go version of the project might take more time as we have to check the compatibility of the packages etc. 
I think using `golangci-lint` github action could be a better approach for this https://github.com/kinvolk/headlamp/issues/707.
cc: @joaquimrocha

illume:
Can you please document this somewhere? Probably at least the case where people want to pass the OIDC client secret through an environment variable mentioned in #696 inside oidc.md

yolossn:
@joaquimrocha Finally the CI is passing, Can you please take a look?

backend/pkg/config/config.go
```diff
@@ -0,0 +1,202 @@
...
+	if c.BaseURL != "" && !strings.HasPrefix(c.BaseURL, "/") {
+		return errors.New("base-url needs to start with a '/' or be empty")
+	}
+
+	return nil
+}
+
+// Parse Loads the config from flags and env.
+// env vars should start with HEADLAMP_CONFIG_ and use _ as separator
+// If a value is set both in flags and env then env takes priority.
```
joaquimrocha, line 54: Hmm, is this the default behavior in most programs?
I was under the impression that it's be default-config-file < env vars < flags .

joaquimrocha: I have quickly checked and it seems that Azure and AWS CLIs do assign a lower priority to env vars than flags. The reasoning may be that usually when you run a command directly, you're hardcoding a value/option you want to use, not the default or the environment predefined ones.

backend/pkg/config/config.go
```diff
@@ -0,0 +1,202 @@
...
+	f.String("oidc-client-secret", "", "ClientSecret for OIDC")
+	f.String("oidc-idp-issuer-url", "", "Identity provider issuer URL for OIDC")
+	f.String("oidc-scopes", "profile,email",
+		"A comma separated list of scopes needed from the OIDC provider")
+
+	return f
+}
+
+// Gets the default plugins-dir depending on platform.
+func defaultPluginDir() string {
```
joaquimrocha, line 141: Should this really be defined here? I'd assume this belongs to plugins.go . Maybe it does belong here, but I am wondering how much of this is configuration and how much is a default value (same for the default kube config).

backend/pkg/config/config_test.go
```diff
@@ -0,0 +1,96 @@
+package config_test
```
joaquimrocha, line 1: Awesome! Thanks for the tests.
I wonder if we could cover all flags before and after, just to make sure we don't have to test all of them manually.

docs/installation/in-cluster/oidc.md
```diff
@@ -10,10 +10,10 @@ Headlamp supports OIDC for cluster users to effortlessly log in using a "Sign in
...
 For OIDC to be used, Headlamp needs to know how to configure it, so you have to provide the different OIDC-related arguments to Headlamp from your OIDC provider. Those are:
 
- * the client ID: `-oidc-client-id`
- * the client secret: `-oidc-client-secret`
- * the issuer URL: `-oidc-idp-issuer-url`
- * (optionally) the OpenId scopes: `-oidc-scopes`
+ * the client ID: `-oidc-client-id` or env var `HEADLAMP_CONFIG_OIDC_CLIENT_ID`
+ * the client secret: `-oidc-client-secret` or env var `HEADLAMP_CONFIG_OIDC_CLIENT_SECRET`
+ * the issuer URL: `-oidc-idp-issuer-url` or env var `HEADLAMP_CONFIG_OIDC_IDP_ISSUER_URL`
+ * (optionally) the OpenId scopes: `-oidc-scopes` or env var `HEADLAMP_CONFIG_OIDC_SCOPES`
```
joaquimrocha, line 16: 👍

joaquimrocha (changes requested):
Left a few comments but this is looking very good and close to landing. Thanks @yolossn !

backend/pkg/config/config.go
yolossn: Changed the priority order

backend/pkg/config/config.go
yolossn: These are default configuration values, so it makes sense to reside in the `config` package.

backend/pkg/config/config_test.go
yolossn: We will have to set up integration tests for the before and after, also we won't be able to test all the config values in integration test apart from things like PORT, etc. Unfortunately, I think we have to settle with manual tests for this once.

joaquimrocha (approved):
LGTM. Thanks!

PR #717, opened 2022-09-23
Fix simple table data type

joaquimrocha:
- frontend: Don't allow objects as SimpleTable data (should be arrays)
- Fix critical crash bug due to empty object in simple table


illume (approved):
👍 thanks

PR #722, opened 2022-09-28
Fix table filter

### LINKED ISSUE #721: Table search/filter doesn't work
## Description

All tables have a search entry for filtering content but this is not working ATM.

## Impact

Users have to browser when they're looking for things.

joaquimrocha:
fixes #721 

How to test:
- [ ] Go to any table that has a filter and press Ctrl+Shift+F, then type something: the table should immediately filter contents based on the string you inputted.

### DIFF
--- a/frontend/src/components/common/SimpleTable.stories.tsx
+++ b/frontend/src/components/common/SimpleTable.stories.tsx
@@ -1,9 +1,12 @@
 import { Meta, Story } from '@storybook/react/types-6-0';
-import React from 'react';
 import { Provider } from 'react-redux';
+import { MemoryRouter } from 'react-router-dom';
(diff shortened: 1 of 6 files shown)


illume (approved):
👍

frontend/src/components/common/SimpleTable.stories.tsx
```diff
@@ -126,7 +131,155 @@ Datum.args = {
...
+        labels: {
+          mykey3: 'myvalue3',
+        },
+      },
+    },
+  ],
+  defaultSortingColumn: 1,
+};
+
+export const NameSearch = TemplateWithFilter.bind({});
```
illume, line 253: 👍

PR #733, opened 2022-09-30
app: Add script to help push assets to releases

joaquimrocha:
This is a maintenance script used to help push assets to the releases and update its checksums.txt file.

This may be used from the CI in the future.

How to test:
- [ ] Don't use the --force option! (unless you use a different/guinea repo)
- [ ] Create a draft release and follow the instructions from --help . Combine the different options (except --force) and see if it does what it states.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -34,6 +34,8 @@
         "fs-extra": "^8.1.0",
         "i18next-parser": "^5.0.0",
         "jest": "^27.4.7",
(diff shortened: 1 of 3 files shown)


illume (approved):
Nice improvement towards automating all the things. Thanks.

I looked around for existing tools for this, and couldn't find anything standard-ish/good to do this. If we want to have a checksums.txt file, this seems a good addition!

I didn't test it (and there's no tests in here).

Issue #712, opened 2022-09-23
[RFE] Allow to apply several resources at once


joaquimrocha:
## Current situation

Headlamp only expects one resource to be applied at a time.

## Impact

This differs from kubectl, so it may confuse users.

## Ideal future situation

We should be able to apply several resources at once like kubectl does.

ashu8912:
fixed by #809

PR #736, opened 2022-10-04
frontend: Allow to use JSON for defining K8s resources

### LINKED ISSUE #713: [RFE] Allow to apply resources defined in json
## Current situation

Headlamp supports only resource definitions in YAML, but it should also support JSON.

## Impact

Users who define resources in JSON cannot apply them from Headlamp.

joaquimrocha:
Besides allowing users to set code in JSON, these changes also set the
error bit much more accurately and often as a result of having to parse
the code anyway.

fixes #713 

How to test:
- [ ] Click the + button, add a YAML resource and apply it: it should work as usual;
- [ ] Replace the contents by JSON and apply it: it should work as YAML did
- [ ] Change pieces of both language versions in the editor to make it not parse correctly: according to whether it looks like YAML or JSON, there should be an error displayed about it and at most it changes every half a second


joaquimrocha:
@illume , this is failing with the heap allocation. Should we increase the max-old-space setting? Do you have a recommendation for this?

illume:
Before I could apply yaml like this: https://github.com/kinvolk/headlamp/blob/main/kubernetes-headlamp.yaml now it doesn't work, and shows the error: "expected a single document in the stream, but found more"

yolossn (approved):

yolossn:
Works as expected for a single JSON but failed for multiple JSON documents like @illume mentioned.

joaquimrocha:
> Before I could apply yaml like this: https://github.com/kinvolk/headlamp/blob/main/kubernetes-headlamp.yaml now it doesn't work, and shows the error: "expected a single document in the stream, but found more"
> 
> Note, kubectl works with multiple documents in JSON as well. Like this [kubernetes-headlamp.json](https://gist.github.com/illume/42042b25d7b8237a1c957f39335ebc09#file-kubernetes-headlamp-json)
> 
> ED: sorry, before the code didn't work either. An error is only shown after pressing apply. So it's not a regression.

Right. We have this other issue for the multiple streams: #712

PR #743, opened 2022-10-12
frontend: Fix filter/search when items are numbers

joaquimrocha:
When items to be matched in the filter are numbers, the filter was
not matching them because it first verified whether the matched items
were string.

How to test:
- [ ] Check the new story added for it


joaquimrocha:
@illume , the plugins testing is failing. Maybe you could take a look?

illume (approved):
👍

PR #744, opened 2022-10-12
Add a new entry to show endpoints

joaquimrocha:
These changes add new list/details views for k8s Endpoints resources.
This is shown under the new Endpoints sub-entry of the sidebar's Network level.

Still to do:
- [ ] Add list of endpoints to the service details


frontend/src/components/endpoints/EndpointList.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+      );
+    },
+  ],
+} as Meta;
+
+const Template: Story = () => {
+  return <EndpointList />;
+};
+
+export const Default = Template.bind({});
```
illume, line 66: Are the different states of EndpointList covered?

frontend/src/components/endpoints/EndpointList.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import { Meta, Story } from '@storybook/react/types-6-0';
+import Endpoints, { KubeEndpoint } from '../../lib/k8s/endpoints';
+import { TestContext } from '../../test';
+import { generateK8sResourceList } from '../../test/mocker';
+import EndpointList from './List';
+
+Endpoints.useList = () => {
```
illume, line 7: Assigning it here would seem to modify the component, so if it's used elsewhere it will also use this stub function.

frontend/src/components/endpoints/EndpointDetails.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+    Endpoints.useGet = args.useGet;
+  }
+  if (!!args.useList) {
+    Endpoints.useList = args.useList;
+  }
+
+  return <EndpointDetails />;
+};
+
+export const Default = Template.bind({});
```
illume, line 95: Perhaps "Items" would describe the state better?
(I know I used "Default" in other stories, but I think Default is a bad name.)

illume (approved):
Nice one. Works for me. 🎉 

I get the same results as with kubectl.
(As an aside... apart from the Age field, which for me says 4w, and kubectl gives 27d)

I left a few notes for your consideration.

frontend/src/components/endpoints/EndpointList.stories.tsx
joaquimrocha: I know. I will work on a way to make sure this happens only on mount and it's re-established to the original on unmount.
But for know should be fine because only the storybook runs this logic?

frontend/src/components/endpoints/EndpointList.stories.tsx
joaquimrocha: What do you mean?

joaquimrocha:
I have renamed Default to Items and also switched the Endpoints place with Ingresses in the sidebar.

PR #745, opened 2022-10-12
Simplify testing

joaquimrocha:
These changes allow to deeply simplify the stories we have and to create them also for views which use the list/get API with data coming from the URL (namespace, name).

see an example at #744


illume (approved):
👍

PR #756, opened 2022-10-17
Add notification and cluster-chooser page

### LINKED ISSUE #644: [RFE] Create Notifications area
## Current situation

We have the notifications popup which show a fixed number of notifications.

## Impact

If we happen to have more than that number of notifications, it is not shown. Also, the popup is a small and maybe difficult to read everything for all users.

## Ideal future situation

Besides the notification's popup, we have also a notifications page which shows all the notifications. In the notifications' popup there's a link to the notifications' page.

## Additional information

The design only makes sense to be delivered together with #16 .

## Design

Notifications mockup

ashu8912:
This patch adds two route entries /notifications and /cluster-chooser  and also add a "view all notifications" to notification popup.

## How to use
Go to notification popup click on view all notification and you will be taken to notification list view

fixes #644

### DIFF
--- a/frontend/src/components/Sidebar/Sidebar.tsx
+++ b/frontend/src/components/Sidebar/Sidebar.tsx
@@ -15,7 +15,7 @@ import CreateButton from '../common/Resource/CreateButton';
 import HeadlampButton from './HeadlampButton';
 import NavigationTabs from './NavigationTabs';
 import prepareRoutes from './prepareRoutes';
-import SidebarItem, { SidebarEntryProps } from './SidebarItem';
+import SidebarItem, { SidebarEntryProps, SidebarItemProps } from './SidebarItem';
 import VersionButton from './VersionButton';
 
 export const drawerWidth = 330;
@@ -65,6 +65,14 @@ const useStyle = makeStyles(theme => ({
   },
 }));
 
+const specialSidebarOptions: SidebarItemProps[] = [
+  {
+    name: 'notifications',
+    icon: 'mdi:bell',
+    label: 'Notifications',
+    url: '/notifications',
+  },
+];
 export default function Sidebar() {
   const sidebar = useTypedSelector(state => state.ui.sidebar);
   const isSidebarOpen = useTypedSelector(state => state.ui.sidebar.isSidebarOpen);
@@ -72,16 +80,24 @@ export default function Sidebar() {
     state => state.ui.sidebar.isSidebarOpenUserSelected
   );
   const arePluginsLoaded = useTypedSelector(state => state.ui.pluginsLoaded);
-
   const namespaces = useTypedSelector(state => state.filter.namespaces);
+  const [isSpecialSidebarOpen, setSpecialSidebarOpen] = React.useState(false);
   const dispatch = useDispatch();
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const { t, i18n } = useTranslation(['glossary', 'frequent']);
   const items = React.useMemo(
-    () => prepareRoutes(t),
-    [sidebar.entries, i18n.language, arePluginsLoaded]
+    () => (isSpecialSidebarOpen ? specialSidebarOptions : prepareRoutes(t)),
+    [sidebar.entries, i18n.language, arePluginsLoaded, isSpecialSidebarOpen]
   );
   const search = namespaces.size !== 0 ? `?namespace=${[...namespaces].join('+')}` : '';
+
+  React.useEffect(() => {
+    if (specialSidebarOptions.map(item => item.name).includes(location.pathname.split('/')[1])) {
+      setSpecialSidebarOpen(true);
+    } else {
+      setSpecialSidebarOpen(false);
+    }
+  }, [location]);
 
   // Use the location to make sure the sidebar is changed, as it depends on the cluster
   // (defined in the URL ATM).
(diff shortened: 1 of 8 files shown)


ashu8912:
Screenshot 2022-10-17 at 12 59 41 PM
Screenshot 2022-10-17 at 12 58 53 PM
Screenshot 2022-10-17 at 12 58 42 PM

joaquimrocha:
@ashu8912 , the clusters entry is supposed to go on top of the notifications entry in the sidebar:
https://github.com/kinvolk/headlamp/issues/644

illume:
> the clusters entry is supposed to go on top of the notifications entry in the sidebar:

I'll test this after this is addressed.

joaquimrocha:
@tsompura, today we have realized that clicking the clusters sidebar entry to show the cluster chooser dialog looks a bit quirky as the background will show an empty view.
Not sure exactly how to fix this before working on the full-blown multi-cluster overview. Let us know if you have an idea for fixing it.

tsompura:
@joaquimrocha Is it possible to return to the current cluster dashboard or to the previous screen from which you arrived at this page? (like browser back button). In my opinion, the user should not be redirected to the cluster selection popup and restart the journey, which should return to the same cluster.

joaquimrocha:
> @joaquimrocha Is it possible to return to the current cluster dashboard or to the previous screen from which you arrived at this page? (like browser back button).

It is possible but maybe it feels still a bit quirky. We can try it.
What happens if the user arrives at the view by direct link though? Should it in that case go to the root view (cluster selection)?

Another question we may ask is: Would there be anything besides the full-blown aggregated multi-cluster view we could use for this cluster view?

tsompura:
@joaquimrocha Yes, if the user arrives via direct link, we should redirect to the default cluster dashboard (currently in cluster link) OR, for the time being, we can use popup view if we don't have more setting options.

Not getting on your another question, However, when we have a multi-cluster view, we should still redirect the user to the current cluster dashboard and provide a link option for a cluster overview. In that case, the links in the left bar should be "Back to Dashboard (or Back)" and "Cluster Overview."

Assume a user is reviewing or analyzing a section that has two or three levels deeper (Workload > Pods > Detail view, etc.) with a selected cluster, and some notification arrives, and clicking on to redirect the detail page, and from that page, the user should have the option to switch back to the previous section or, at the very least, the current dashboard, and should not start the journey again, in my opinion.

joaquimrocha:
I have noticed that the notifications in the page are missing a link to the associated event.

tsompura:
@joaquimrocha Updated mockup with link added "Mark all as read" & "Clear all"
https://www.figma.com/proto/o37HxZC9aE22iQvvaAEyJ3/Notification-UI?page-id=513%3A2&node-id=621%3A2&viewport=2965%2C688%2C0.64&scaling=min-zoom

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -65,21 +65,29 @@ const useStyle = makeStyles(theme => ({
   },
 }));
 
+const specialSidebarOptions: SidebarItemProps[] = [
+  {
+    name: 'notifications',
+    icon: 'mdi:bell',
+    label: 'Notifications',
```
joaquimrocha, line 77: This needs to be translatable.

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -65,21 +65,29 @@ const useStyle = makeStyles(theme => ({
...
+];
 export default function Sidebar() {
   const sidebar = useTypedSelector(state => state.ui.sidebar);
   const isSidebarOpen = useTypedSelector(state => state.ui.sidebar.isSidebarOpen);
   const isSidebarOpenUserSelected = useTypedSelector(
     state => state.ui.sidebar.isSidebarOpenUserSelected
   );
   const arePluginsLoaded = useTypedSelector(state => state.ui.pluginsLoaded);
-
+  const isSpecialSidebarOpen = useTypedSelector(state => state.ui.isSpecialSidebarOpen);
```
joaquimrocha, line 88: Maybe what determines if the special sidebar is open or not is the sidebar item that the route declares it needs. Otherwise we always to special case whether to set the special sidebar or not.

joaquimrocha (changes requested):

joaquimrocha:
Some other problems I am seeing:

1. Not sure how but I got these notifications without a message and the tooltip shows as "undefined" (we should at least make sure the tooltip is not shown if no message is available):
Screenshot showing a tooltip with the word undefined in it

2. The notifications don't a link to go to their respective URL like the popup does. This is a UX problem.

3. If I use the popup and click one of the notifications there, it'll take me to the Cluster overview but keep the special sidebar. This shows that the current way to define whether the special or normal sidebar are set should be improved:
Cluster overview with special sidebar

joaquimrocha:
Also, when clicking "back", I get 
Notifications with normal sidebar

frontend/src/components/Sidebar/Sidebar.tsx
ashu8912: updated

frontend/src/components/Sidebar/Sidebar.tsx
ashu8912: updated

joaquimrocha: Where?

PR #760, opened 2022-10-19
frontend: List pods in OwnedPodsSection with query parameters

lijianzhi01:
# frontend: List pod in OwnedPodsSection with query parameters

In Resource.ts, list pod in OwnedPodsSection with query parameters. If users only have namespace access, list all pods isn't available.

## How to use

Instead of filtering pods list, OwnedPodsSection query pods with label selectors and field selectors.

## Testing done

Test list pods in deployment with query parameters.

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -681,7 +681,7 @@ export function OwnedPodsSection(props: OwnedPodsSectionProps) {
 }
 
 // Label selector examples: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering
-// deployment selector example: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering
+// deployment selector example: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements
 // Possible operators: https://github.com/kubernetes/apimachinery/blob/be3a79b26814a8d7637d70f4d434a4626ee1c1e7/pkg/selection/operator.go#L24
 // Format rule for expressions: https://github.com/kubernetes/apimachinery/blob/be3a79b26814a8d7637d70f4d434a4626ee1c1e7/pkg/labels/selector.go#L305
 function getLabelSelector(item: KubeObjectInterface): string | undefined {
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -729,7 +729,7 @@ function getLabelSelector(item: KubeObjectInterface): string | undefined {
     }
 
     let sorted = [...(expr.values ?? [])].sort().join(',');
-    if (expr.operator == 'In' || expr.operator == 'NotIn') {
+    if (expr.operator === 'In' || expr.operator === 'NotIn') {
       sorted = '(' + sorted + ')';
     }
     segment += sorted;
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -728,19 +728,11 @@ function getLabelSelector(item: KubeObjectInterface): string | undefined {
         continue;
     }
 
-    switch (expr.operator) {
-      case 'In':
-      case 'NotIn':
-        segment += '(';
+    let sorted = [...(expr.values ?? [])].sort().join(',');
+    if (expr.operator == 'In' || expr.operator == 'NotIn') {
+      sorted = '(' + sorted + ')';
     }
-    const sorted = [...(expr.values ?? [])].sort();
-    segment += sorted.join(',');
-    switch (expr.operator) {
-      case 'In':
-      case 'NotIn':
-        segment += ')';
-    }
-
+    segment += sorted;
     segments.push(segment);
   }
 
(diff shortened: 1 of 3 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -486,6 +515,7 @@ export function remove(url: string, requestOptions = {}) {
 export async function streamResult(
   url: string,
   name: string,
+  queryParams: QueryParameters | undefined,
```
joaquimrocha, line 518: The `streamResult` function is exported, which means this breaks API for plugins.
Since the `queryParams` can be undefined, we can easily avoid braking API by adding as optional, to the end of the params. i.e `queryParams?: QueryParamters` after `errCb`.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -520,7 +551,12 @@ export async function streamResult(
   }
 }
 
-export async function streamResults(url: string, cb: StreamResultsCb, errCb: StreamErrCb) {
+export async function streamResults(
+  url: string,
+  queryParams: QueryParameters | undefined,
```
joaquimrocha, line 556: Same about breaking API.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -164,9 +166,7 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
     static apiList<U extends KubeObject>(
       onList: (arg: U[]) => void,
       onError?: (err: ApiError) => void,
-      opts?: {
-        namespace?: string;
-      }
```
joaquimrocha, line 169: This was only one single namespace. With your changes, it can be also an array of strings.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -393,6 +418,10 @@ function simpleApiFactoryWithNamespace(
   }
 }
 
+function asQuery(queryParams?: QueryParameters): string {
+  return '?' + new URLSearchParams(queryParams).toString();
```
joaquimrocha, line 422: For the requests to be exactly the same as before when the query is empty, the queryParams should be checked for validity (whether it's undefined or empty) before we return `'?'+..`, otherwise you may have `?` without any query.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: Maybe you just do something like:
`{
namespace?: string;
} | Omit<ApiListOptions, 'namespace'>`

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -669,32 +669,86 @@ export interface OwnedPodsSectionProps {
 export function OwnedPodsSection(props: OwnedPodsSectionProps) {
   const { resource, hideColumns } = props;
 
-  const [pods, error] = Pod.useList();
```
joaquimrocha, line 672: (Not related to this line)
Please prefix the commit name with `frontend: ` if it only touched the frontend.

joaquimrocha (changes requested):
Thanks @lijianzhi01 !
I left a couple of comments in the PR. Let me know if you can address them or prefer that we do it.
Also, it'd be very cool to separate the logic that converts the matchExpressions to a string into `lib/k8s/index.ts` and add a test for it in `index.tests.ts`. Again, let me know if you can do this or prefer us to take it from here.

frontend/src/components/common/Resource/Resource.tsx
lijianzhi01: Aha, fixed it.

frontend/src/lib/k8s/apiProxy.ts
lijianzhi01: Make sense!

frontend/src/lib/k8s/apiProxy.ts
lijianzhi01: Indeed!

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -669,32 +669,86 @@ export interface OwnedPodsSectionProps {
...
-    if (!pods) {
-      return null;
-    }
+  const [pods, error] = Pod.useList(queryData);
+
+  return <PodListRenderer hideColumns={hideColumns} pods={pods} error={error} />;
+}
+
+// Label selector examples: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering
+// deployment selector example: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering
```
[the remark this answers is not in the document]
lijianzhi01: Link updated!

frontend/src/lib/k8s/apiProxy.ts
lijianzhi01: Has moved the queryParams to the last position!

frontend/src/lib/k8s/cluster.ts
lijianzhi01: I set QueryParams as a member of opts. It looks like
`{ namespace?: string; queryParams?: QueryParams}`

lijianzhi01:
> 

I have resolved your comments except converting the matchExpressions to a string into lib/k8s/index.ts and adding a test for it in index.tests.ts. I think you can take from here.

joaquimrocha:
Thanks @lijianzhi01 . We'll take it from here.

PR #762, opened 2022-10-19
frontend: Replace lastTerminationState with lastState

lijianzhi01:
# Replace lastTerminationState with lastState

The lastTerminationState of KubeContainerStatus is not existed. I checked the doc here. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/

## How to use

Replace lastTerminationState  to lastState

## Testing done


lijianzhi01:
It should be part of format pod running status.

Issue #734, opened 2022-09-30
Release notes view seems to delay/block Headlamp's startup


joaquimrocha:
## Description

Today I experimented a weird issue: Headlamp 0.12.1 (released yesterday) would start by showing a blank page, then a black page, and slowly the contents would appear. The same would not occur when using the very commit prior to the release bump.

This seems to be related to the release notes, so we should check whether network conditions affect it and make it so that the notes can be skipped if they're not ready in like a second. Feedback should also be shown to the user while the release notes are being fetched (a spinner + message, and a skip button I think).

joaquimrocha:
> I'd like to keep this open so the solution can be improved. There's still a 5 second delay with no feedback if there's a network issue.

Good point. Let's keep this open.

joaquimrocha:
Fixed with #832

PR #763, opened 2022-10-19
Fix github release fetch issue

ashu8912:


ashu8912:
fixes #734

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -21,7 +21,9 @@ export default function ReleaseNotes() {
             return;
           }
 
-          const octokit = new Octokit();
+          const octokit = new Octokit({
+            timeout: 5000,
+          });
```
joaquimrocha, line 26: Was this the one place causing the delay?
Could we also get in trouble by fetching the actual release ( 2 lines below) if the network gets hyper slow then?

app/electron/main.ts
```diff
@@ -37,6 +37,8 @@ const defaultPort = 4466;
 const isDev = process.env.ELECTRON_DEV || false;
 const useExternalServer = process.env.EXTERNAL_SERVER || false;
 const shouldCheckForUpdates = process.env.HEADLAMP_CHECK_FOR_UPDATES !== 'false';
+// make it global so that it doesn't get garbage collected
+let mainWindow: BrowserWindow | null;
```
joaquimrocha, line 41: When why would this be gc'd?

joaquimrocha (commented):
This patch seems to deal with 2 different issues. So they should be separated commits.
I'd like to understand better if there's a problem with the window's life cycle as you seem to indicate.

app/electron/main.ts
ashu8912: on app close only

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
ashu8912: So that is handled in, this object only every request made using this instance will have a timeout of 5 ms

joaquimrocha: Okay.

joaquimrocha:
I have split the commit.

app/electron/main.ts
joaquimrocha: I think my question was written in a confusing/wrong way.
Can you elaborate on what was the issue with having the window be declared in the startElectron method? (since it's only used there)

ashu8912: It's an extra problem i was looking for which can sometime make the main process block and hence laggy experience, This has to do with local scope defining of mainWindow which can get garbage collected on function execution,

Making it a global scope one prevents us from mistakenly getting the mainWindow varibale getting garbage collected.

ashu8912: checkout this 
https://stackoverflow.com/questions/71810712/electron-app-freezes-in-production-build-when-there-are-no-errors-and-no-memory

ashu8912: yeah this is more of a defensive solution to prevent any unnecessary gotchas.
https://github.com/electron/electron-quick-start/issues/221

PR #764, opened 2022-10-19
frontend: Print pod status following kubectl

lijianzhi01:
# Print pod status following kubectl

Print pod status following kubectl convention. https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L759

## How to use
The pod status in Headlamp cannot describe the full picture of the container status. It is confusing when user see running pod status marked with yellow. So, I change status and restarts by following the link I provide above. 

In this PR.
image

Current situation.
image

This is pretty wired 
## Testing done

### DIFF
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -127,7 +127,7 @@ interface PodTableRow {
   lastRestartDate: Date;
 }
 
-// Source Code: https://github.com/kubernetes/kubernetes/blob/33bf8a8ebd33eae0d45c18feb1c05baa6a055cde/pkg/printers/internalversion/printers.go#L558
+// Source Code: https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L759
 function getPodTableInfo(pod: KubePod): PodTableRow {
   let restarts = 0;
   const totalContainers = (pod.spec.containers ?? []).length;
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -1,9 +1,9 @@
 import { Icon } from '@iconify/react';
 import { Box } from '@material-ui/core';
-import _ from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { ApiError } from '../../lib/k8s/apiProxy';
-import Pod from '../../lib/k8s/pod';
+import Pod, { KubePod } from '../../lib/k8s/pod';
+import { timeAgoByDate } from '../../lib/util';
 import { LightTooltip, SectionFilterHeader } from '../common';
 import { StatusLabel, StatusLabelProps } from '../common/Label';
 import ResourceTable, { ResourceTableProps } from '../common/Resource/ResourceTable';
@@ -14,6 +14,7 @@ export function makePodStatusLabel(pod: Pod) {
   let status: StatusLabelProps['status'] = '';
   let tooltip = '';
 
+  const podRow = getPodTableInfo(pod.jsonData);
   if (phase === 'Failed') {
     status = 'error';
   } else if (phase === 'Succeeded' || phase === 'Running') {
@@ -32,7 +33,7 @@ export function makePodStatusLabel(pod: Pod) {
     <LightTooltip title={tooltip} interactive>
       <Box display="inline">
         <StatusLabel status={status}>
-          {phase}
+          {podRow.reason}
           {(status === 'warning' || status === 'error') && (
             <Box
               aria-label="hidden"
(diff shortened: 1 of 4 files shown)


lijianzhi01:
The link provided is old. But the code in pod/List.ts is following the latest version of code. I'll replace link in the code.

frontend/src/lib/util.ts
```diff
@@ -230,6 +235,59 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
   return [error, setError as any];
 }
 
+// Source Code: https://github.com/kubernetes/kubernetes/blob/2f275b72b2248bb7c522327115acb17223b8387b/staging/src/k8s.io/apimachinery/pkg/util/duration/duration.go#L48
+export function humanDuration(milliseconds: number): string {
```
joaquimrocha, line 239: This will output very similar output to the `humanize-duration` package we've started using recently.
I don't see a need for this function since we already output a similar format + we allow for localized versions if needed.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -516,32 +517,34 @@ export interface KubeMetrics {
...
-      signal: number;
-      startedAt: number;
-    };
-    waiting: {
-      message: string;
-      reason: string;
-    };
-  };
+  lastState: ContainerStateApplyConfiguration; // aka: lastTerminationState
+  state: ContainerStateApplyConfiguration;
```
joaquimrocha, line 547: This should be called `ContainerState`.

frontend/src/lib/k8s/pod.ts
```diff
@@ -15,6 +15,7 @@ export interface KubePod extends KubeObjectInterface {
     nodeSelector?: {
       [key: string]: string;
     };
+    initContainers?: any[];
```
joaquimrocha, line 18: I believe this should be a `KubeContainer[]`.

frontend/src/components/pod/List.tsx
```diff
@@ -111,3 +118,135 @@ export default function PodList() {
...
+function getPodTableInfo(pod: KubePod): PodTableRow {
+  let restarts = 0;
+  const totalContainers = (pod.spec.containers ?? []).length;
+  let readyContainers = 0;
+  let lastRestartDate = new Date(0);
+
+  let reason = pod.status.phase;
+  if (!!pod.status.reason) {
+    reason = pod.status.reason;
+  }
```
joaquimrocha, line 140: This can be `let reason = pod.status.reason ? pod.status.reason :  pod.status.phase;`

frontend/src/components/pod/List.tsx
```diff
@@ -111,3 +118,135 @@ export default function PodList() {
...
+      } else if (container.ready && !!container.state.running) {
+        hasRunning = true;
+        readyContainers++;
+      }
+    }
+
+    // change pod status back to "Running" if there is at least one container still reporting as "Running" status
+    if (reason === 'Completed' && hasRunning) {
+      if (hasPodReadyCondition(pod.status.conditions)) {
+        alert('Running');
```
joaquimrocha, line 214: This was probably left over from debugging?

frontend/src/components/pod/List.tsx
```diff
@@ -111,3 +118,135 @@ export default function PodList() {
...
+  }
+
+  // Instead of `pod.deletionTimestamp`. Important!
+  const deletionTimestamp = pod.metadata.deletionTimestamp;
+
+  if (!!deletionTimestamp && pod.status.reason === 'NodeLost') {
+    reason = 'Unknown';
+  } else if (!!deletionTimestamp) {
+    reason = 'Terminating';
+  }
```
joaquimrocha, line 229: This could be a nested if just checking for the `pod.status.reason`

frontend/src/components/pod/List.tsx
```diff
@@ -111,3 +118,135 @@ export default function PodList() {
...
+  return {
+    restartsStr: restartsStr,
+    totalContainers: totalContainers,
+    readyContainers: readyContainers,
+    reason: reason,
+    lastRestartDate: lastRestartDate,
+  };
+}
+
+function hasPodReadyCondition(conditions: any): boolean {
```
joaquimrocha, line 245: I think we have a type for the conditions.

frontend/src/components/pod/List.tsx
```diff
@@ -127,7 +127,7 @@ interface PodTableRow {
   lastRestartDate: Date;
 }
 
-// Source Code: https://github.com/kubernetes/kubernetes/blob/33bf8a8ebd33eae0d45c18feb1c05baa6a055cde/pkg/printers/internalversion/printers.go#L558
+// Source Code: https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L759
```
joaquimrocha, line 130: I know @illume asked for this, but I am not sure if having a link to a branch is better than to the actual commit, since the code will likely change.

lijianzhi01:
> Hi @lijianzhi01 . Thanks for this PR. I left a few comments. I am fine taking the PR from here if you prefer. Just want to know if you are fine with us keeping the current abbreviated dates which are very similar to the ones in the function you introduced.

Cool, please go ahead from my PR. It will be efficient to work this out if you start from here. I may spend lot more time than you to finish this.

frontend/src/lib/util.ts
lijianzhi01: I like what you mentioned here. Let's use humanize-duration package. The purpose of this function is to display output the same as kubectl does. But some users told me they'd prefer to **output a similar format + we allow for localized versions**.

frontend/src/components/pod/List.tsx
lijianzhi01: Yes, it is. Please delete it.

PR #765, opened 2022-10-19
frontend: Fix unexpected scroll on filter change

yolossn:
Before:

https://user-images.githubusercontent.com/22731013/196707719-c14c87de-b21b-488b-9653-d4ed74f7180e.mp4

After:

https://user-images.githubusercontent.com/22731013/196707641-e2089d51-3d93-4aec-a8c0-b7dcb1be3f6b.mp4

### DIFF
--- a/frontend/src/components/common/SectionFilterHeader.tsx
+++ b/frontend/src/components/common/SectionFilterHeader.tsx
@@ -32,10 +32,6 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
 
   const [showFilters, setShowFilters] = React.useState<boolean>(hasNamespaceFilters || hasSearch);
 
-  React.useEffect(() => {
-    setShowFilters(hasSearch);
-  }, [hasSearch]);
-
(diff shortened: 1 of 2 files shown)


ashu8912:
@joaquimrocha @yolossn what was the issue here?? Why would it scroll, sorry i tried to understand this from the PR changes but couldn't understand the scroll behaviour

joaquimrocha:
I believe it's something that has to do with focus or loss of it, but in any case, the current solution is good and innocent enough that it shouldn't be risky to add it.

illume (approved):
Works for me.

I did spot a separate bug on the Cluster view. It doesn't search the Type field. So if I type in "Pod" it fails to find any results. If you don't fix it in here, I'll raise another bug.

joaquimrocha:
> I did spot a separate bug on the Cluster view. It doesn't search the Type field. So if I type in "Pod" it fails to find any results. If you don't fix it in here, I'll raise another bug.

Let's do a separate PR.

PR #768, opened 2022-10-20
frontend: Fix duplicated protocol in reference

lijianzhi01:
# Duplicated protocol in reference

In connecStream, the original way will cause duplicated protocol pushed into variables which additionalProtocols refer to.


joaquimrocha (approved):
Makes sense. Thanks!

PR #776, opened 2022-10-24
frontend: Fix reconnectOnFailure according to comments

lijianzhi01:
According to code comment, set reconnectOnFailure as true by default if the failCb has not been provided. So, it should be single exclamation mark.


joaquimrocha (approved):
Good catch. Thanks.

PR #777, opened 2022-10-24
Fetch pods by labels

joaquimrocha:
# frontend: List pod in OwnedPodsSection with query parameters

In Resource.ts, list pod in OwnedPodsSection with query parameters. If users only have namespace access, list all pods isn't available.

Branch created from #760 so we can rework the PR a bit.
cc/ @lijianzhi01 

## How to use

Instead of filtering pods list, OwnedPodsSection query pods with label selectors and field selectors.

## Testing done

* Test list pods in deployment with query parameters. 
* Run `make frontend-test` to run the unit tests for the query conversion function.

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -394,14 +394,11 @@ function simpleApiFactoryWithNamespace(
       queryParams?: QueryParameters
     ) => streamResult(url(namespace), name, cb, errCb, queryParams),
     post: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
-      post(url(body.metadata.namespace as string) + asQuery(queryParams), body),
+      post(url(body.metadata.namespace!) + asQuery(queryParams), body),
     patch: (body: OpPatch[], namespace: string, name: string, queryParams?: QueryParameters) =>
       patch(`${url(namespace)}/${name}` + asQuery({ ...queryParams, ...{ pretty: 'true' } }), body),
     put: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
-      put(
-        `${url(body.metadata.namespace as string)}/${body.metadata.name}` + asQuery(queryParams),
-        body
-      ),
+      put(`${url(body.metadata.namespace!)}/${body.metadata.name}` + asQuery(queryParams), body),
     delete: (namespace: string, name: string, queryParams?: QueryParameters) =>
       remove(`${url(namespace)}/${name}` + asQuery(queryParams)),
     isNamespaced: true,
@@ -468,7 +465,7 @@ function apiScaleFactory(apiRoot: string, resource: string) {
   return {
     get: (namespace: string, name: string) => request(url(namespace, name)),
     put: (body: { metadata: KubeMetadata; spec: { replicas: number } }) =>
-      put(url(body.metadata.namespace as string, body.metadata.name), body),
+      put(url(body.metadata.namespace!, body.metadata.name), body),
   };
 
   function url(namespace: string, name: string) {
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -43,6 +43,12 @@ export interface ClusterRequest {
   certificateAuthorityData?: string;
 }
 
+export interface QueryParameters {
+  labelSelector?: string;
+  fieldSelector?: string;
+  [prop: string]: any;
+}
+
 //refreshToken checks if the token is about to expire and refreshes it if so.
 async function refreshToken(token: string | null) {
   if (!token || isTokenRefreshInProgress) {
@@ -311,13 +317,18 @@ function singleApiFactory(group: string, version: string, resource: string) {
   const apiRoot = getApiRoot(group, version);
   const url = `${apiRoot}/${resource}`;
   return {
-    list: (cb: StreamResultsCb, errCb: StreamErrCb) => streamResults(url, cb, errCb),
-    get: (name: string, cb: StreamResultsCb, errCb: StreamErrCb) =>
-      streamResult(url, name, cb, errCb),
-    post: (body: KubeObjectInterface) => post(url, body),
-    put: (body: KubeObjectInterface) => put(`${url}/${body.metadata.name}`, body),
-    patch: (body: OpPatch[], name: string) => patch(`${url}/${name}?pretty=true`, body),
-    delete: (name: string) => remove(`${url}/${name}`),
+    list: (cb: StreamResultsCb, errCb: StreamErrCb, queryParams?: QueryParameters) =>
+      streamResults(url, cb, errCb, queryParams),
+    get: (name: string, cb: StreamResultsCb, errCb: StreamErrCb, queryParams?: QueryParameters) =>
+      streamResult(url, name, cb, errCb, queryParams),
+    post: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
+      post(url + asQuery(queryParams), body),
+    put: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
+      put(`${url}/${body.metadata.name}` + asQuery(queryParams), body),
+    patch: (body: OpPatch[], name: string, queryParams?: QueryParameters) =>
+      patch(url + asQuery({ ...queryParams, ...{ pretty: 'true' } }), body),
+    delete: (name: string, queryParams?: QueryParameters) =>
+      remove(`${url}/${name}` + asQuery(queryParams)),
     isNamespaced: false,
   };
 }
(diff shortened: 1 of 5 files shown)


frontend/src/lib/k8s/index.ts
```diff
@@ -173,6 +173,86 @@ export function useConnectApi(...apiCalls: (() => CancellablePromise)[]) {
...
+
+  const matchExpressions = labelSelector.matchExpressions ?? [];
+  for (const expr of matchExpressions) {
+    let segment = '';
+    if (expr.operator === 'DoesNotExist') {
+      segment += '!';
+    }
+
+    let needsParensWrap = false;
+    let expectedValuesLength = -1; // -1 means no limit
```
lijianzhi01, line 196: Nice category! Code is more readable.

frontend/src/lib/k8s/index.ts
```diff
@@ -173,6 +173,86 @@ export function useConnectApi(...apiCalls: (() => CancellablePromise)[]) {
...
+  }
+
+  const matchExpressions = labelSelector.matchExpressions ?? [];
+  for (const expr of matchExpressions) {
+    let segment = '';
+    if (expr.operator === 'DoesNotExist') {
+      segment += '!';
+    }
+
+    let needsParensWrap = false;
```
lijianzhi01, line 195: After checking this variable, I have understood what you mean about wrap.

frontend/src/lib/k8s/index.test.ts
```diff
@@ -55,3 +55,148 @@ describe('Class tests', () => {
     });
   });
 });
+
+const labelSelector: LabelSelector = {
```
lijianzhi01, line 59: I remember that @joaquimrocha  and @illume mentioned about test infra for Headlamp. Are these test cases prepared for that?

joaquimrocha: Unfortunately testing the label selector effectively would have to involved the k8s API server (as mocking the results wouldn't make much sense in this case).
We maybe have a way to do this now that we just added e2e testing.

illume (changes requested):
Can you please add a PR description? I see some new fields/params added... would it be possible to document those?

I'll review again after those are added.

ashu8912 (approved):
Tested the PR, LGTM
Will let @illume look into the code structure for a better look

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -369,16 +380,30 @@ function simpleApiFactoryWithNamespace(
...
+      namespace: string,
+      name: string,
+      cb: StreamResultsCb,
+      errCb: StreamErrCb,
+      queryParams?: QueryParameters
+    ) => streamResult(url(namespace), name, cb, errCb, queryParams),
+    post: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
+      post(url(body.metadata.namespace as string) + asQuery(queryParams), body),
+    patch: (body: OpPatch[], namespace: string, name: string, queryParams?: QueryParameters) =>
+      patch(`${url(namespace)}/${name}` + asQuery({ ...queryParams, ...{ pretty: 'true' } }), body),
```
illume, line 399: I wonder if we should be using URL and URL.searchParams instead of stringly adding them together. Would be a bit safer.

frontend/src/lib/k8s/index.ts
```diff
@@ -173,6 +173,86 @@ export function useConnectApi(...apiCalls: (() => CancellablePromise)[]) {
   );
 }
 
+// Label selector examples: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering
```
illume, line 176: These links could be part of the function documentation...
```ts
/**
 * See {@link https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering|Label selector examples}
 */
```

frontend/src/lib/k8s/index.ts
illume, line 196: I was confused reading the code below where it's used about what -1 meant. Eventually I found the comment, but Instead of -1, we could use a string instead and a union type.

Something like...

```ts
expectedValuesLength : string | number = 'no limit'
...
    } else if (expectedValuesLength === 'no limit') {
...
```

illume (commented):
Added a few more comments, but I still need to read the format docs linked from labelSelectorToQuery.

frontend/src/lib/k8s/index.test.ts
```diff
@@ -55,3 +55,148 @@ describe('Class tests', () => {
...
+    const query = labelSelectorToQuery({
+      matchExpressions: [
+        {
+          key: 'label2',
+          operator: 'In',
+          values: ['value2', 'value3'],
+        },
+      ],
+    });
+    expect(query).toBe('label2 in (value2,value3)');
```
illume, line 96: Is there supposed to be a space?

- `'label2 in (value2,value3)'`
- VS `'label2 in (value2, value3)'`

I ask because the examples I see have spaces. https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering

frontend/src/lib/k8s/index.ts
```diff
@@ -173,6 +173,86 @@ export function useConnectApi(...apiCalls: (() => CancellablePromise)[]) {
...
+      case 'In':
+        segment += ' in ';
+        needsParensWrap = true;
+        break;
+      case 'NotIn':
+        segment += ' notin ';
+        needsParensWrap = true;
+        break;
+      case 'GreaterThan':
+        segment += '>';
```
illume, line 221: I think these GreaterThan and LessThan are wrong. See https://github.com/kubernetes/apimachinery/blob/be3a79b26814a8d7637d70f4d434a4626ee1c1e7/pkg/selection/operator.go#L31

frontend/src/lib/k8s/index.ts
```diff
@@ -173,6 +173,86 @@ export function useConnectApi(...apiCalls: (() => CancellablePromise)[]) {
...
+        expectedValuesLength = 1;
+        break;
+      case 'LessThan':
+        segment += '<';
+        expectedValuesLength = 1;
+        break;
+      case 'Exists':
+      case 'DoesNotExist':
+        expectedValuesLength = 0;
```
illume, line 230: Should these ones(Exists, and DoesNotExist) modify the segment variable?

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -369,16 +380,30 @@ function simpleApiFactoryWithNamespace(
...
+    ) => streamResults(url(namespace), cb, errCb, queryParams),
+    get: (
+      namespace: string,
+      name: string,
+      cb: StreamResultsCb,
+      errCb: StreamErrCb,
+      queryParams?: QueryParameters
+    ) => streamResult(url(namespace), name, cb, errCb, queryParams),
+    post: (body: KubeObjectInterface, queryParams?: QueryParameters) =>
+      post(url(body.metadata.namespace as string) + asQuery(queryParams), body),
```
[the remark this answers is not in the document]
lijianzhi01: Namespace in metadata is defined as string type. So, I think it would be fine to cast namespace to string. https://github.com/kinvolk/headlamp/blob/f9002aa769caa2ffd00480981b5b51aa27e17069/frontend/src/lib/k8s/cluster.ts#L35

frontend/src/lib/k8s/apiProxy.ts
lijianzhi01: In asQuery, URL searchParams is used. Could you please explain what is stringly adding them and what does safer mean?

frontend/src/lib/k8s/index.ts
lijianzhi01: I did some experiment. `<` and `>` are ok. `lt` and `gt` are not allowed in our kubectl.

illume: This seems to be some sort of internal label, and ">" and "<" works in testing. I'll mark this as resolved.

frontend/src/lib/k8s/apiProxy.ts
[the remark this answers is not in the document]
joaquimrocha: Turns out we only use a full (http/s) URL later in the request, so we cannot just call this with `new URL` AFAICS.
I agree with removing the `as string` and passing the assertion instead (`metadata.namespace!`).

frontend/src/lib/k8s/index.ts
joaquimrocha: I don't love the type overload for just interpreting a number, so I added a const that represents the no-limits.

frontend/src/lib/k8s/index.test.ts
joaquimrocha: Both work with kubectl. Let's keep the one without a space since this will save a few bytes when sending the request over.

frontend/src/lib/k8s/index.test.ts
illume: I think this conversation can be resolved... but not sure. Can one of you please resolve it, or let me know what needs to be done/discussed?

frontend/src/lib/k8s/apiProxy.ts
illume: We had a little conversation about this elsewhere... but I'll comment here too.

In general, it is safer if objects like URL() were used to construct the URLs instead of constructing them with string interpolation.
- The URL object handles a number of cases which are hard to do correctly. One is that query params can be repeated in a URL but using a JS dict doesn't allow duplicates. Another place it can help is safely joining paths together. Inserting values from untrusted input is safer than just using string interpolation (here we already use the searchParams though for the query params).
- the type checker when using an API vs using strings.
- The code is easier to review/audit.

Probably we have other parts in the codebase already that use strings for constructing URLs, so it could be done as a different PR or put in an issue I guess.

frontend/src/lib/k8s/apiProxy.ts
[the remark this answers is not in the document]
illume: I don't follow what you mean exactly, but it sounds like you fixed it. Resolving this convo.

illume:
@joaquimrocha I added the PR description from the other PR, but I guess that's not perhaps entirely correct for this PR. Can you please review it?

frontend/src/lib/k8s/apiProxy.ts
joaquimrocha: What I meant in https://github.com/kinvolk/headlamp/pull/777#discussion_r1012785684 is that we cannot use the `URL` class here because we're still dealing with a partial URL (or rather its path + query), so creating an instance of `URL` with the point of validating it here would fail (the actual full URL is created later, when performing the actual request).
And since we already use `URLSearchParams` in `asQuery`, I don't think this logic is very unsafe as it is, or not more than before anyway.
So any further improvements to safety should be done at the request level and can be done in a separate PR.

illume (approved):
👍 RAD</80'sSkaterVoice>

PR #779, opened 2022-10-24
frontend: Fix app crash due to null reference

lijianzhi01:
Fix app crash due to ingress might be null

### DIFF
--- a/frontend/src/lib/k8s/service.ts
+++ b/frontend/src/lib/k8s/service.ts
@@ -51,7 +51,7 @@ class Service extends makeKubeObject<KubeService>('service') {
     return _.uniq(
       (
         this.status?.loadBalancer?.ingress?.map(
-          (ingress: KubeLoadBalancerIngress) => ingress.hostname || ingress.ip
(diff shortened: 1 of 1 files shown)


frontend/src/lib/k8s/service.ts
```diff
@@ -51,7 +51,7 @@ class Service extends makeKubeObject<KubeService>('service') {
     return _.uniq(
       (
         this.status?.loadBalancer?.ingress?.map(
-          (ingress: KubeLoadBalancerIngress) => ingress.hostname || ingress.ip
+          (ingress: KubeLoadBalancerIngress) => ingress?.hostname || ingress?.ip
```
joaquimrocha, line 54: Hi @lijianzhi01 . I would assume `ingress` to be an array and never have `null` in.
Are we preprocessing that array somehow and leading to those null refs?

lijianzhi01: Yes. Let me see if I can find any evidence for null value of ingress.

lijianzhi01:
I cannot find any proof of this. So, I will close this PR and open it when I find anything.

PR #782, opened 2022-10-25
Improve pod status

joaquimrocha:
Taking over the PR at #764 

# Print pod status following kubectl

Print pod status following kubectl convention. https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L759

- Cache the detailed pod status results in Pod
- Add new stories for the Pod list and delete the old ones
- Print pod status following kubectl's format, for a more descriptive reason why the pod may be failing
- Improve a few types (and optionality) in cluster.ts and pod.ts
- Replace time type in Kube* definitions

## How to use
The pod status in Headlamp cannot describe the full picture of the container status. It is confusing when user see running pod status marked with yellow. So, I change status and restarts by following the link I provide above. 

In this PR.
image

Current situation.
image

This is pretty weird

## Testing done

- [x] storybook examples of the different states
- [x] works locally with a smoke test


lijianzhi01:
Hi @joaquimrocha, I am just curious about __snapshot__ folder. What it is and why you add it? https://github.com/kinvolk/headlamp/pull/782/commits/1f911bcf05554be54690f516e041c3f4b6f83303#diff-88efd7f329208609bff6555d507f650a22c43a527c92125b5acf624cacdc2130

frontend/src/lib/util.ts
```diff
@@ -230,6 +235,59 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
   return [error, setError as any];
 }
 
+// Source Code: https://github.com/kubernetes/kubernetes/blob/2f275b72b2248bb7c522327115acb17223b8387b/staging/src/k8s.io/apimachinery/pkg/util/duration/duration.go#L48
```
lijianzhi01, line 238: You add this function back. Why not use `humanize-duration` package?

joaquimrocha: This was a mistake. I'll remove it.

illume (changes requested):
Looking good. Left a few questions and suggestions.

I'll add a PR description.

frontend/src/lib/k8s/pod.ts
```diff
@@ -57,8 +59,23 @@ type newGetLogs = (
   logsOptions: LogOptions
 ) => () => void;
 
+type PodDetailedStatus = {
+  restarts: number;
```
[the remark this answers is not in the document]
joaquimrocha: This is just the type of the return object we're using to describe the status we want to show.
I called it detailed status but could be called something else.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -549,32 +552,35 @@ export interface KubeMetrics {
   };
 }
 
+export interface ContainerState {
```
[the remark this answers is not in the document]
joaquimrocha: This is coming from the K8s [spec/API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/). Unfortunately I cannot understand well from it which fields are mandatory or optional.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,142 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
 
     return stream(url, onExec, { additionalProtocols, isJson: false, ...streamOpts });
   }
+
+  // Implementation based on: https://github.com/kubernetes/kubernetes/blob/33bf8a8ebd33eae0d45c18feb1c05baa6a055cde/pkg/printers/internalversion/printers.go#L558
+  getDetailedStatus(): PodDetailedStatus {
```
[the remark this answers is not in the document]
joaquimrocha: I added extracted one small piece of code that was repeated in the function into a helper method.
Unfortunately the rest of the code kind of makes sense together. I started extracting the initContainers' status into its own function, but the logic of reconciling the results + not having the code flow the same would differ from the printers.go function we took the inspiration from, making any future comparison more complicated. Thus I'd rather leave it like this.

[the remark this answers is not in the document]
illume: It's really not readable to me.

frontend/src/lib/k8s/cluster.ts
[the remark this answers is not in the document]
illume: Thanks, I'll check it to see if I can figure it out.

ashu8912:
Tested the PR, and it works fine for me

frontend/src/lib/k8s/pod.ts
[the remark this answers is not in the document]
joaquimrocha: I think we may have more problems understanding the logic if later we need to compare it with the Go code.
If you have a way to make it more readable without compromising that, please send a fixup.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,151 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
...
+            reason = 'Init:' + container.state.terminated.reason;
+          }
+          message = container.state.terminated.message || '';
+          initializing = true;
+          break;
+        case !!container.state.waiting?.reason &&
+          container.state.waiting.reason !== 'PodInitializing':
+          reason = 'Init:' + container.state.waiting.reason;
+          initializing = true;
+          message = container.state.waiting.message || '';
```
illume, line 210: This message part isn't in the golang function.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,151 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
...
+      break;
+    }
+
+    if (!initializing) {
+      restarts = 0;
+      let hasRunning = false;
+      for (let i = (this.status?.containerStatuses?.length || 0) - 1; i >= 0; i--) {
+        const container = this.status?.containerStatuses[i];
+
+        restarts += container.restartCount;
```
illume, line 225: Here golang function converts to an int, but we don't.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,151 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
...
+        lastRestartDate = this.getLastRestartDate(container, lastRestartDate);
```
illume, line 226: LastRestartDate isn't in the golang function.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,151 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
...
+
+        if (!!container.state.waiting?.reason) {
+          reason = container.state.waiting.reason;
+          message = container.state.waiting.message || '';
```
illume, line 230: The messages here aren't in the golang func.

frontend/src/lib/k8s/pod.ts
```diff
@@ -117,6 +134,151 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
...
+          message = container.state.terminated.message || '';
+        } else if (container.ready && !!container.state.running) {
+          hasRunning = true;
+          readyContainers++;
+        }
+      }
+
+      // change pod status back to "Running" if there is at least one container still reporting as "Running" status
+      if (reason === 'Completed' && hasRunning) {
+        if (this.hasPodReadyCondition(this.status?.conditions)) {
```
illume, line 249: This is different logic to the golang function.

illume (commented):
There's a number of differences inside the getDetailedStatus function different to the golang one. They should be documented, because the function description made me believe it was the same.

The int conversion one might be a bug.

frontend/src/lib/k8s/pod.ts
joaquimrocha: Right. This is an implementation based on the golang one.
Is this something problematic in your opinion? (Any alternatives if so?)

frontend/src/lib/k8s/pod.ts
joaquimrocha: Good point. I have checked, and this is an integer in the jsonData. So it should be fine here.
The initContainers had an issue though (not being actually defined in the status).

frontend/src/lib/k8s/pod.ts
joaquimrocha: Right. You asked me to simplify this whole function and I identified that the last restart date was being repeated, so I abstracted a couple of blocks with this function. It's small enough that we don't lose a lot of connection with the original function, as we would if we try to abstract the much more complex two container blocks.

illume (approved):
👍 thanks

PR #784, opened 2022-10-26
Add support for HPA,PDB,ResourceQuotas and PriorityClasses

yolossn:
To Do:

- [x] i18n changes
- [x] Add storybook

### DIFF
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -26,6 +26,10 @@ function prepareRoutes(t: (arg: string) => string) {
           label: t('glossary|Config Maps'),
         },
         {
+          name: 'horizontalPodAutoscalers',
+          label: t('glossary|HPAs'),
+        },
+        {
           name: 'resourceQuotas',
           label: t('glossary|Resource Quotas'),
         },
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -37,6 +37,10 @@ function prepareRoutes(t: (arg: string) => string) {
           name: 'podDisruptionBudgets',
           label: t('glossary|Pod Disruption Budgets'),
         },
+        {
+          name: 'priorityClasses',
+          label: t('glossary|Priority Classes'),
+        },
       ],
     },
     {
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -33,6 +33,10 @@ function prepareRoutes(t: (arg: string) => string) {
           name: 'resourceQuotas',
           label: t('glossary|Resource Quotas'),
         },
+        {
+          name: 'podDisruptionBudgets',
+          label: t('glossary|Pod Disruption Budgets'),
+        },
       ],
     },
     {
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -25,6 +25,10 @@ function prepareRoutes(t: (arg: string) => string) {
           name: 'configMaps',
           label: t('glossary|Config Maps'),
         },
+        {
+          name: 'resourceQuotas',
+          label: t('glossary|Resource Quotas'),
+        },
       ],
     },
     {
--- a/frontend/src/components/horizontalPodAutoscaler/Details.tsx
+++ b/frontend/src/components/horizontalPodAutoscaler/Details.tsx
@@ -1,0 +1,54 @@
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import HPA from '../../lib/k8s/hpa';
+import { DetailsGrid, Link, ObjectEventList, SimpleTable } from '../common';
+
+export default function HpaDetails() {
+  const { namespace, name } = useParams<{ namespace: string; name: string }>();
+  const { t } = useTranslation(['hpa', 'frequent']);
+
+  return (
+    <DetailsGrid
+      resourceType={HPA}
+      name={name}
+      namespace={namespace}
+      extraInfo={item =>
+        item && [
+          {
+            name: t('hpa|Reference'),
+            value: (
+              <Link kubeObject={item.referenceObject}>
+                {item.referenceObject?.kind}/{item.referenceObject?.metadata.name}
+              </Link>
+            ),
+          },
+          {
+            name: t('hpa|Metrics'),
+            value: (
+              <SimpleTable
+                data={item.metrics(t)}
+                columns={[
+                  { label: t('frequent|Name'), getter: item => item.definition },
+                  { label: t('hpa|(Current/Target)'), getter: item => item.value },
+                ]}
+              />
+            ),
+          },
+          {
+            name: t('hpa|MinReplicas'),
+            value: item.spec.minReplicas,
+          },
+          {
+            name: t('hpa|MaxReplicas'),
+            value: item.spec.maxReplicas,
+          },
+          {
+            name: t('hpa|Deployment pods'),
+            value: `${item.status.currentReplicas} current / ${item.status.desiredReplicas} desired`,
+          },
+        ]
+      }
+      sectionsFunc={item => item && <ObjectEventList object={item} />}
+    />
+  );
+}
--- a/frontend/src/components/horizontalPodAutoscaler/List.tsx
+++ b/frontend/src/components/horizontalPodAutoscaler/List.tsx
@@ -1,0 +1,71 @@
+import { useTranslation } from 'react-i18next';
+import HPA from '../../lib/k8s/hpa';
+import { Link, SectionBox, SectionFilterHeader, StatusLabel } from '../common';
+import ResourceTable from '../common/Resource/ResourceTable';
+
+export default function HpaList() {
+  const { t } = useTranslation(['glossary', 'hpa', 'frequent']);
+
+  return (
+    <SectionBox title={<SectionFilterHeader title={t('glossary|Horizontal Pod Autoscalers')} />}>
+      <ResourceTable
+        resourceClass={HPA}
+        columns={[
+          'name',
+          'namespace',
+          {
+            label: t('hpa|Reference'),
+            getter: item => (
+              <Link kubeObject={item.referenceObject}>
+                {item.referenceObject?.kind}/{item.referenceObject?.metadata.name}
+              </Link>
+            ),
+          },
+          {
+            label: t('hpa|Targets'),
+            getter: (hpa: HPA) => {
+              const value: JSX.Element[] = [];
+              const metrics = hpa.metrics(t);
+              if (metrics.length) {
+                value.push(
+                  <StatusLabel
+                    key={metrics[0].name}
+                    id={metrics[0].name}
+                    status=""
+                    title={metrics[0].name}
+                  >
+                    {metrics[0].shortValue}
+                  </StatusLabel>
+                );
+                if (hpa.metrics.length > 1) {
+                  value.push(
+                    <StatusLabel key="more..." status="">
+                      {metrics.length - 1} {t('frequent|more...')}
+                    </StatusLabel>
+                  );
+                }
+              }
+              return <>{value}</>;
+            },
+          },
+          {
+            label: t('hpa|MinReplicas'),
+            getter: item => item.spec.minReplicas,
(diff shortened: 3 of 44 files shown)


illume:
Could you please rebase this against main? Running the tests should then be warning free now.

(Probably you'll need to regenerate the story snapshots and check them again)

joaquimrocha:
> Could you please rebase this against main? Running the tests should then be warning free now.
> 
> (Probably you'll need to regenerate the story snapshots and check them again)

This is based on the fetch-pods-by-labels. So we'd have to rebase both.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -133,7 +133,8 @@ export async function request(
   path: string,
   params: RequestParams = {},
   autoLogoutOnAuthError: boolean = true,
-  useCluster: boolean = true
+  useCluster: boolean = true,
+  queryParams?: QueryParameters
```
joaquimrocha, line 137: I wonder if we should deprecated the current version of request and use all these options as part of the `RequestParams` so it scales better.

frontend/src/components/resourceQuota/Details.tsx
```diff
@@ -0,0 +1,43 @@
...
+                columns={[
+                  {
+                    label: t('resourceQuota|Resource'),
+                    getter: item => item.name,
+                  },
+                  {
+                    label: t('resourceQuota|Used'),
+                    getter: item => item.used,
+                  },
+                  {
```
joaquimrocha, line 31: I think we may have these and other labels already translated to the same or very similar meaning elsewhere. If that's the case, we should reuse those translations rather than creating new ones for this new namespace.

frontend/src/components/resourceQuota/Details.tsx
```diff
@@ -0,0 +1,43 @@
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import ResourceQuota from '../../lib/k8s/resourceQuota';
+import { DetailsGrid, SimpleTable } from '../common';
+
+export default function ResourceQuotaDetails() {
+  const { namespace, name } = useParams<{ namespace: string; name: string }>();
+  const { t } = useTranslation();
```
joaquimrocha, line 8: You should preload the namespaces you use in this view, I think this is why you may have warnings.
i.e. `useTranslation(['frequent', 'resourceQuota', ...]);`

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -25,6 +25,10 @@ function prepareRoutes(t: (arg: string) => string) {
           name: 'configMaps',
           label: t('glossary|Config Maps'),
         },
+        {
+          name: 'horizontalPodAutoscalers',
+          label: t('glossary|Horizontal Pod Autoscalers'),
```
joaquimrocha, line 30: Maybe `HPAs` as otherwise it'll be too long?

frontend/src/lib/k8s/hpa.ts
```diff
@@ -0,0 +1,316 @@
...
+    const metrics: HPAMetrics[] = [];
+    for (let iter = 0; iter < this.spec.metrics.length; iter++) {
+      const spec = this.spec.metrics[iter];
+      const status = this.status.currentMetrics?.[iter];
+      let value = '';
+      switch (spec.type) {
+        case 'External':
+          if (spec.external && status?.external) {
+            value = `${
+              status ? metricValueStatus(status.external.current) : '<unknown>'
```
joaquimrocha, line 191: Unknown should be translatable.

frontend/src/lib/k8s/hpa.ts
```diff
@@ -0,0 +1,316 @@
...
+          }
+          break;
+        case 'Object':
+          if (spec.object && status?.object) {
+            value = `${
+              status ? metricValueStatus(status.object.current) : '<unknown>'
+            }/${metricTargetValue(spec.object.target)}`;
+            metrics.push({
+              name: spec.object.metric.name,
+              definition: `"${spec.object.metric.name}" on ${spec.object.describedObject.kind}/${
```
joaquimrocha, line 210: This should be translatable.

frontend/src/components/podDisruptionBudget/List.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <SectionBox title={<SectionFilterHeader title={t('glossary|Pod Disruption Budget')} />}>
+      <ResourceTable
+        resourceClass={PDB}
+        columns={[
+          'name',
+          'namespace',
+          {
+            label: t('pdb|Min Available'),
+            getter: (item: PDB) => item.spec.minAvailable || 'N/A',
```
joaquimrocha, line 18: I think we have a translated N/A.

joaquimrocha (changes requested):
Quick review (have to test later). Left a couple of comments.

frontend/src/components/resourceQuota/List.tsx
```diff
@@ -0,0 +1,40 @@
...
+              });
+              return <>{requests}</>;
+            },
+          },
+          {
+            label: t('resourceQuota|Limit'),
+            getter: (item: ResourceQuota) => {
+              const limits: JSX.Element[] = [];
+              item.limits.forEach((limit: string) => {
+                limits.push(<StatusLabel status="">{limit}</StatusLabel>);
```
joaquimrocha, line 30: This and similar may need a `key` (check for warnings).

frontend/src/lib/k8s/apiProxy.ts
yolossn: If my understanding is correct,In the `request` implementation `RequestParams` is used to pass `RequestHeaders` and the newly added `QueryParams` cannot be merged together with that.

frontend/src/components/resourceQuota/Details.tsx
yolossn: I reverified, `Resources` exists in the `glossary` but this is `Resource`.

joaquimrocha:
I have tested it and I think the events' table should reuse bits of the general Events sections and putting the date as the last column, like other tables have.

List of events in an HPA resource

E.g. using the colored status label component as the general Events section is using:
List of events in cluster

I will give another look at the code.

frontend/src/lib/k8s/apiProxy.ts
joaquimrocha: We should rename the `params` as `headers` then. This function's parameters are also getting very long, so we should eventually create a `request(path: string, params: RequestParams);` where request params are all these parameters supplied as an object (as it will scale better for the future).

This means also deprecating the current way.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -165,7 +166,11 @@ export async function request(
   const controller = new AbortController();
   const id = setTimeout(() => controller.abort(), timeout);
 
-  const url = combinePath(BASE_HTTP_URL, fullPath);
+  let url = combinePath(BASE_HTTP_URL, fullPath);
+  if (queryParams) {
+    const query = new URLSearchParams(queryParams);
+    url = `${url}?${query.toString()}`;
+  }
```
joaquimrocha, line 173: We have as of very recently this function called `asQuery` which does most of this.

frontend/src/lib/k8s/hpa.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  static async getEvents(namespace: string, name: string) {
+    const path = `/api/v1/namespaces/${namespace}/events`;
+    const params: QueryParameters = {
+      fieldSelector: `involvedObject.name=${name},involvedObject.kind=HorizontalPodAutoscaler,involvedObject.namespace=${namespace}`,
+      limit: 500,
+    };
+    const response = await request(path, {}, true, true, params);
+    return response.items;
+  }
```
joaquimrocha, line 342: This is a generic events fetcher function, so AFAIU, we should define it outside of the HPA class.
A probably better idea is to use it right from the event, like: `const events = Event.useList({namespace, name})` which would fetch it as regularly but using the query params, and making it a hook (be careful with likely wanting to call it conditionally, on the existence of namespace/name, as that logic needs to be moved into the hook).

frontend/src/components/horizontalPodAutoscaler/Details.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const { t } = useTranslation(['hpa', 'frequent']);
+
+  const [events, setEvents] = useState([]);
+  async function fetchEvents() {
+    const events = await HPA.getEvents(namespace, name);
+    setEvents(events);
+  }
+  useEffect(() => {
+    fetchEvents();
+  }, []);
```
joaquimrocha, line 19: See comment about using `Event.useList`

frontend/src/components/horizontalPodAutoscaler/List.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    status=""
+                    title={metrics[0].name}
+                  >
+                    {metrics[0].shortValue}
+                  </StatusLabel>
+                );
+                if (hpa.metrics.length > 1) {
+                  value.push(
+                    <StatusLabel key="more..." status="">
+                      {metrics.length - 1} more...
```
joaquimrocha, line 43: Needs translation.

frontend/src/components/horizontalPodAutoscaler/List.tsx
```diff
@@ -0,0 +1,71 @@
...
+          },
+          {
+            label: t('hpa|MaxReplicas'),
+            getter: item => item.spec.maxReplicas,
+            sort: (hpa: HPA) => hpa?.spec.maxReplicas,
+          },
+          {
+            label: t('glossary|Replicas'),
+            getter: item => item.status.currentReplicas,
+            sort: (hpa: HPA) => hpa?.status.currentReplicas,
```
joaquimrocha, line 64: Hmm, I think you can just use `sort: true` and it'll use the value it has (from the getter). Wouldn't that work here?

frontend/src/lib/k8s/cluster.ts
```diff
@@ -55,6 +56,16 @@ export interface ApiListOptions extends QueryParameters {
   namespace?: string | string[];
 }
 
+export interface KubeManagedFields {
+  apiVersion: string;
+  fieldsType: string;
+  fieldsV1: { map: Map<string, string[]> };
```
joaquimrocha, line 62: Is it going to be a map, or `{[key: string]: string[]}`.
I am asking because I don't think we have maps anywhere else in the types.

frontend/src/lib/k8s/hpa.ts
```diff
@@ -0,0 +1,366 @@
...
+    return this.jsonData!.spec;
+  }
+
+  get status(): HpaStatus {
+    return this.jsonData!.status;
+  }
+
+  metrics(t: Function): HPAMetrics[] {
+    const metrics: HPAMetrics[] = [];
+    for (let iter = 0; iter < this.spec.metrics.length; iter++) {
```
joaquimrocha, line 184: I wonder if we should divide each case in the switch into different functions.

frontend/src/lib/k8s/cluster.ts
yolossn: Changing this to `{[key:string]:string[]}` as we are not dependent on the `Map` functionalities in this case.

frontend/src/lib/k8s/hpa.ts
yolossn: Moved this to a separate component.

frontend/src/lib/router.tsx
```diff
@@ -397,6 +399,20 @@ const defaultRoutes: {
     sidebar: 'secrets',
     component: () => <SecretDetails />,
   },
+  resourceQuotas: {
+    path: '/resourcequotas',
+    exact: true,
+    name: 'Resource Quotas',
+    sidebar: 'resourceQuotas',
```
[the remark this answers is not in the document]
yolossn: Fixed this

yolossn:
@joaquimrocha can you PTAL?

joaquimrocha:
I couldn't get to it today. Will do it tomorrow.

joaquimrocha:
There's still a problem with the sidebar sections. I will add some fixup patches:
ResourceQuotas with the sidebar section not activated

should be:

ResourceQuotas with the sidebar section activated

joaquimrocha:
I had commented that the sidebar was still incorrect, but I was testing an outdated version of it.
Now it LGTM! Great job! Thanks.

PR #786, opened 2022-10-28
frontend: Optimize KubeObjectClass.useGet so it updates iff needed

### LINKED ISSUE #783: [useList/useApiGet hooks] [streamResult/streamResults functions] Callback fires twice on same resource/s
## Description

When we use useList or useApiGet hooks, we use streamResult/streamResults functions under the hood which get success callback as a parameter.
StreamResult function makes an initial request and fires success callback on received data and then creates a socket which immediately gets a first message with the same resource and again fires success callback  which makes our views render twice on the same data.
I think stream functions could compare resourceVersion of the object it gets on first request and the object it gets from first message and fire the callback only if it gets a new object.
Some similar logic described in streamResults function in switch construction in MODIFIED case.

## Impact

Performance

## Expected behavior

Fire a success callback only when new data is received

joaquimrocha:
This function was setting its state every time new data was returned
by the API function, but that data may be the same as before, which
would still trigger a change in any component using it. Thus, this
patch makes it so that the state only changes if the resourceVersion
changes.

fixes #783

### DIFF
--- a/frontend/src/lib/k8s/cluster.ts
+++ b/frontend/src/lib/k8s/cluster.ts
@@ -310,13 +310,29 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
       const [error, setError] = useErrorState(setObj);
 
       function onGet(item: U | null) {
+        // Only set the object if we have we have a different one.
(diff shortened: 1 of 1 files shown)


ashu8912 (approved):
LGTM!!

PR #790, opened 2022-11-02
frontend: Add filter by type for events section

### LINKED ISSUE #769: [RFE] Allow to search by involved object type in Events table
## Current situation

We show an events' table in the cluster overview which we can search but not by the involved object's type, and this would be a good way to filter out events by the user.

## Ideal future situation

Typing replicaset or pod would filter the table to only show the ReplicaSet or Pod objects, respectively.

ashu8912:

Allow to search by type for events table

## How to use

Go to cluster overview page go to events table search with a resource type and it would filter the table based on resource type

fixes #769

### DIFF
--- a/frontend/src/components/cluster/Overview.tsx
+++ b/frontend/src/components/cluster/Overview.tsx
@@ -8,6 +8,7 @@ import { useLocation } from 'react-router';
 import Event from '../../lib/k8s/event';
 import Node from '../../lib/k8s/node';
 import Pod from '../../lib/k8s/pod';
+import { useFilterFunc } from '../../lib/util';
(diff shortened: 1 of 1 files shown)


ashu8912:
fixes #790

joaquimrocha:
Tested and works well. Also LGTM.

lijianzhi01 (commented):
This is pretty cool. But I want to mention two things according to our customers. 
1. Filter is hard to notice. Some users do not even know filter exists. 
2. There are several dimensions to filter data already in filter. But the UI does not have any tips for it. So, users need to guess what can be searched or confused about it.

PR #791, opened 2022-11-03
Render logs using xterm

joaquimrocha:
This PR changes how the logs are rendered, it namely:
* Allows selected text to remain selected even if new logs arrive
* Allows copying logs
* Improves the performance by only rendering the logs that have not yet been shown
* Adds search to logs
* Allows to stop following logs (pauses the stream)

fixes https://github.com/kinvolk/headlamp/issues/785

Testing:
* There's a new story for checking logs in the Storybook
* Go to a pod that has logs or create a [pod that logs constantly](https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/debug/counter-pod.yaml) and click the logs button

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -45,6 +45,7 @@ function PodLogViewer(props: PodLogViewerProps) {
   const [container, setContainer] = React.useState(getDefaultContainer());
   const [showPrevious, setShowPrevious] = React.useState<boolean>(false);
   const [showTimestamps, setShowTimestamps] = React.useState<boolean>(true);
+  const [follow, setFollow] = React.useState<boolean>(true);
   const [lines, setLines] = React.useState<number>(100);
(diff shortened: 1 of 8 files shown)


frontend/src/components/pod/Details.tsx
```diff
@@ -216,9 +276,14 @@ export function VolumeDetails(props: VolumeDetailsProps) {
   );
 }
 
-export default function PodDetails() {
+export interface PodDetailsProps {
+  showLogs?: boolean;
```
illume, line 280: Perhaps showLogs should be renamed to showLogsDefault, because that's what it is at the moment. If that prop is changed then the PodDetails component won't use that change.

illume (approved):
👍 thanks

frontend/src/components/pod/Details.tsx
joaquimrocha: Got it. Will change.

Issue #748, opened 2022-10-13
[RFE] Allow to persist query parameters


joaquimrocha:
## Current situation

We already persist some query parameters like the `namespace` filter across views (even though this is not fully working because of #739 ) but this is not easy to do for any other examples nor do we allow plugins to have URL query params they want to persist across views.

Recently we were also asked by a plugin developer whether we could allow certain parameters to persist.

## Impact

Even if plugins can keep state outside of the URL, having certain query parameters in the URL is good for being able to share the exact state.

## Ideal future situation

We have a way to persist URL params.

## Implementation options

I thought of 2 ways we could do this:
  1. We pre-process the URL before generating routes (and navigating across views), so we/plugins can add any needed query before the URL is returned or;
  2. We keep a list of "URL params to persist" which is configurable by Headlamp and plugins, so when navigate to a route, we check whether the current URL has those params and we set them (by using their name + value taken from either the current URL or other config).

Let's debate this a bit and choose a good way forward.

cc/ @callmevladik, @lijianzhi01

illume:
I think https://github.com/kinvolk/headlamp/issues/21 should be addressed as a part of this effort (or first perhaps). There's plenty of state/URL mismatches. To do end to end tests better for some parts of the app we'll need these URL fixes as well from #21.

PR #793, opened 2022-11-05
Make Sure we Persist url params

ashu8912:
When we change views and there is a namespace filter applied we should persist it in the URL
The SectionFilterHeader takes care of if there should be a namespace filter to be applied on its initial render
## How to use
GO to a namespace filter, select a few namespaces and now switch views and we have the namespace filter in the URL now for all the views that have the namespace filter, which was only for some certain views before


joaquimrocha (commented):
The namespace filter was already being persisted in some views, and not in others. Your patch seems to make it persistent on all, but didn't touch how it was previously being set up in the views that did have it. Can you make sure we only have/need one way of adding these params?

ashu8912:
> The namespace filter was already being persisted in some views, and not in others. Your patch seems to make it persistent on all, but didn't touch how it was previously being set up in the views that did have it. Can you make sure we only have/need one way of adding these params?

@joaquimrocha fixed this, so we were actually not passing the search prop to the children render of the SidebarItem, that's why it wasn't working for the child sidebar items

joaquimrocha (approved):

joaquimrocha:
fixes #748

PR #794, opened 2022-11-07
Go back from details views using actual previous route

joaquimrocha:
This is related to #793 . Right now, the back link of details views in Headlamp was either passed to the route's state, or it'd generate it as being the resource's list view route. This means that going into a resource's details view and back will not keep things like URL's search parameters, state, etc.

This PR fixes that by using the actual previous location for that back link button.

I tried using document.referer but it was always empty. Also, history.length/go/back functions cannot be used thoroughly here since they keep the full history of the browser, i.e. calling `back()` (`go(-1)`) from `history` will take the user to the previous page, regardless of whether the page belongs to the site (headlamp) and that make for a very confusing UX (as the back link button looks like it's only related to Headlamp).


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -185,13 +193,22 @@ export function DetailsGrid(props: DetailsGridProps) {
   const { sectionsFunc, resourceType, name, namespace, children, ...otherMainInfoSectionProps } =
     props;
   const location = useLocation<{ backLink: NavLinkProps['location'] }>();
+  const previousLocation = usePreviousRoute();
 
-  const backLink: string | undefined = React.useMemo(() => {
+  const backLink: string | Location | undefined = React.useMemo(() => {
```
lijianzhi01, line 198: Just curious about useMemo with empty dependencies?

lijianzhi01:
Resource list page does not have page number in query parameter. If go back from detail page, the table shows the first page. If this pr is not prepared for what I mentioned above, please ignore.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: I think this means the function inside will only be defined on component mount. Maybe it's not needed as the advantage we get is not much though maybe it protects for adding more heavy function logic in the future...

joaquimrocha:
> It would be better to set table page number in query parameters. If go back from detail page, the table shows the first page.

Yes, but this will be done in a different PR.

frontend/src/components/common/Resource/Resource.tsx
ashu8912: I think we don't need the useMemo here cause a useMemo with empty array means it will be computed on every render, maybe in future when we have some extra optimization to be done then this makes more sense.

ashu8912 (approved):
A small question

ashu8912:
I have a small question/confusion related to the UX of the back button, I don't completely understand what purpose does the back button solely presents, like if i go into a nested chain of links i am not taken back all the way back to where i started clicking the links, this makes back button a little confusing to me.

joaquimrocha:
> I have a small question/confusion related to the UX of the back button, I don't completely understand what purpose does the back button solely presents, like if i go into a nested chain of links i am not taken back all the way back to where i started clicking the links, this makes back button a little confusing to me.

I did it this way because the back button is supposed to take you to the page you were before that one, but IIUC, you're saying that if I go to: List of Nodes  -> Select a node (back button -> List of nodes) -> Select a pod in the list of pods in node (back button -> list of nodes) -> Select node where pod's running (back button -> Pod details again).
I guess this can be confusing.

I will try to address this next week.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: That's not what the empty array does. An empty array means it only runs on mount and cleans up on unmount (i.e. runs just once). See the [notes](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) in this section.
For the hook to run on every render, regardless of its dependencies, the array argument would have to be skipped/undefined.

joaquimrocha:
> I did it this way because the back button is supposed to take you to the page you were before that one, but IIUC, you're saying that if I go to: List of Nodes -> Select a node (back button -> List of nodes) -> Select a pod in the list of pods in node (back button -> list of nodes) -> Select node where pod's running (back button -> Pod details again). I guess this can be confusing.
> 
> I will try to address this next week.

So, the solution I thought about for fixing this is to have a stack of the routes. Essentially it's like a history clone that we can track. We need to track full locations in this stack and not just a history count because we'd have to match locations when the user goes back more than 1 level in history.

However, this solution doesn't work because users can go back a more than 1 level, and in that case, just matching the locations  is good only if there are no repeated locations in the stack (going to the same URL more than once).

Therefore, either we blindly go back using the browser's method: the only downside is that if you go `Search Engine > Headlamp details view`, then the back button will go to the search engine.
Or we still use the stack and at least prevent the above case (unless the use never jumps back more than one level)...

I implemented the latter and instead of keeping the full history, just keep a numeric value to understand whether we have a previous Headlamp route or not.

Any comments? Should we just go back using the browser's method and leave the added complexity of checking the previous route?

cc/ @ashu8912 , @illume

illume:
Not sure if you saw these? I see [useNavigate](https://reactrouter.com/en/main/hooks/use-navigate) being recommended for back buttons (`navigate(-1)`). There's also a way to do breadcrumbs mentioned in the docs of useMatches: https://reactrouter.com/en/main/hooks/use-matches#breadcrumbs

I think it would be useful to outline some examples of expected behaviours to test. Then we can make sure we try the different cases (like the ones you've already mentioned in these comments).

But yeah. The browser back is limited and has that poor case you mentioned. Not so sure it's a good one to use.

joaquimrocha:
> Not sure if you saw these? I see [useNavigate](https://reactrouter.com/en/main/hooks/use-navigate) being recommended for back buttons (`navigate(-1)`). There's also a way to do breadcrumbs mentioned in the docs of useMatches: https://reactrouter.com/en/main/hooks/use-matches#breadcrumbs

This requires a new version of react-router but it works the same as `goBack()`/`go(-1)`, AFAIU.
About the breadcrumbs, I think it just uses the matched routes unless I am mistaken, and in that case it won't help our case since we are jumping to different routes from ones that aren't necessarily prefixes of them.

> I think it would be useful to outline some examples of expected behaviours to test. Then we can make sure we try the different cases (like the ones you've already mentioned in these comments).

The behaviors are simple:
1. User arrives at a resource's details view directly (without navigating from within Headlamp); clicks the back button -> goes to the list view of that resource.
2. User arrives at a resource's details view from within Headlamp; clicks the back button -> goes to the previous location, keeping path, search, and state (which is what the browser back button does).
3. User arrives at a list view, clicks a resource to view its details view A; from here, goes into another details view B from a page link; clicks the back button -> goes to the previous details view A (same as case 2); clicks again the back button -> initially, my PR was sending the user to the details view B (view where they came from), but @ashu8912 argued that it should instead keep following the history (like a browser back button) so it'd send user to the list view (which kind of makes some sense).

> But yeah. The browser back is limited and has that poor case you mentioned. Not so sure it's a good one to use.

Do you think it'd make sense to keep my initial PR? i.e. case 3 would never go back to the list view.

joaquimrocha:
After talking to Ashu, this implementation makes sense even if not perfect. I will merge it.

PR #809, opened 2022-11-11
Apply multiple resources at once 

### LINKED ISSUE #712: [RFE] Allow to apply several resources at once

### LINKED ISSUE #737: Docs viewer is no longer working
## Description

The docs viewer takes the resource definition currently set in the editor dialog and uses it to retrieve the docs dynamically.
I have noticed this is not working ATM.

## Impact

Users can no longer see the docs.

ashu8912:
This patch allows the editor to apply multiple resources at once and also shows docs for all the available resources

fixes #737 
fixes #712

### DIFF
--- a/frontend/src/components/common/Resource/CreateButton.tsx
+++ b/frontend/src/components/common/Resource/CreateButton.tsx
@@ -15,42 +15,57 @@ export default function CreateButton() {
   const location = useLocation();
   const { t } = useTranslation(['resource', 'frequent']);
 
-  const applyFunc = async (newItem: KubeObjectInterface) => {
-    try {
-      await apply(newItem);
-    } catch (err) {
-      let msg = t('Something went wrong…');
-      if (err instanceof Error) {
-        msg = err.message;
-      }
-      setErrorMessage(msg);
-      setOpenDialog(true);
-      throw err;
-    }
+  const applyFunc = async (newItems: KubeObjectInterface[]) => {
+    await Promise.allSettled(newItems.map(newItem => apply(newItem))).then((values: any) => {
+      values.forEach((value: any, index: number) => {
+        if (value.status === 'rejected') {
+          let msg;
+          const kind = newItems[index].kind;
+          const name = newItems[index].metadata.name;
+          const apiVersion = newItems[index].apiVersion;
+          if (newItems.length === 1) {
+            msg = t('Failed to create {{ kind }} {{ name }}.', { kind, name });
+          } else {
+            msg = t('Failed to create {{ kind }} {{ name }} in {{ apiVersion }}.', {
+              kind,
+              name,
+              apiVersion,
+            });
+          }
+          setErrorMessage(msg);
+          setOpenDialog(true);
+          throw msg;
+        }
+      });
+    });
   };
 
-  function handleSave(newItemDef: KubeObjectInterface) {
+  function handleSave(newItemDefs: KubeObjectInterface[]) {
     const cancelUrl = location.pathname;
-
-    if (!newItemDef.metadata?.name) {
-      setErrorMessage(t('Please set a name to the resource!'));
-      return;
+    // check if all yaml objects are valid
+    for (let i = 0; i < newItemDefs.length; i++) {
+      if (!newItemDefs[i].metadata?.name) {
+        setErrorMessage(t(`Invalid: One or more of the resource doesn't have a name property`));
+        return;
+      }
+      if (!newItemDefs[i].kind) {
+        setErrorMessage(t('Invalid: Please set a kind to the resource!'));
+        return;
+      }
     }
-
-    if (!newItemDef.kind) {
-      setErrorMessage(t('Please set a kind to the resource!'));
-      return;
-    }
-
-    const newItemName = newItemDef.metadata.name;
-
+    // all resources name
+    const resourceNames = newItemDefs.map(newItemDef => newItemDef.metadata.name);
     setOpenDialog(false);
     dispatch(
-      clusterAction(() => applyFunc(newItemDef), {
-        startMessage: t('Applying {{ newItemName }}…', { newItemName }),
-        cancelledMessage: t('Cancelled applying {{ newItemName }}.', { newItemName }),
-        successMessage: t('Applied {{ newItemName }}.', { newItemName }),
-        errorMessage: t('Failed to apply {{ newItemName }}.', { newItemName }),
+      clusterAction(() => applyFunc(newItemDefs), {
+        startMessage: t('Applying {{ newItemName }}…', { newItemName: resourceNames.join(',') }),
+        cancelledMessage: t('Cancelled applying {{ newItemName }}.', {
+          newItemName: resourceNames.join(','),
+        }),
+        successMessage: t('Applied {{ newItemName }}.', { newItemName: resourceNames.join(',') }),
+        errorMessage: t('Failed to apply {{ newItemName }}.', {
+          newItemName: resourceNames.join(','),
+        }),
         cancelUrl,
       })
     );
--- a/frontend/src/components/common/Resource/DocsViewer.tsx
+++ b/frontend/src/components/common/Resource/DocsViewer.tsx
@@ -4,7 +4,6 @@ import { makeStyles } from '@material-ui/core/styles';
 import Typography from '@material-ui/core/Typography';
 import TreeItem from '@material-ui/lab/TreeItem';
 import TreeView from '@material-ui/lab/TreeView';
-import _ from 'lodash';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import getDocDefinitions from '../../../lib/docs';
@@ -22,32 +21,53 @@ const useStyles = makeStyles(() => ({
 function DocsViewer(props: { docSpecs: any }) {
   const { docSpecs } = props;
   const classes = useStyles();
-  const [docs, setDocs] = React.useState<object | null>(null);
-  const [docsError, setDocsError] = React.useState<string | null>(null);
+  const [docs, setDocs] = React.useState<
+    (
+      | {
+          data: null;
+          error: any;
+          kind: string;
+        }
+      | {
+          data: any;
+          error: null;
+          kind: string;
+        }
+      | undefined
+    )[]
+  >([]);
+  const [docsLoading, setDocsLoading] = React.useState(false);
   const { t } = useTranslation('resource');
 
   React.useEffect(() => {
-    setDocsError(null);
-
-    if (docSpecs.error) {
-      t('Cannot load documentation: {{ docsError }}', { docsError: docSpecs.error });
-      return;
-    }
-    if (!docSpecs.apiVersion || !docSpecs.kind) {
-      setDocsError(
-        t(
-          'Cannot load documentation: Please make sure the YAML is valid and has the kind and apiVersion set.'
-        )
-      );
-      return;
-    }
-
-    getDocDefinitions(docSpecs.apiVersion, docSpecs.kind)
-      .then(result => {
-        setDocs(result?.properties || {});
+    setDocsLoading(true);
+    // fetch docSpecs for all the resources specified
+    Promise.allSettled(
+      docSpecs.map((docSpec: { apiVersion: string; kind: string }) => {
+        return getDocDefinitions(docSpec.apiVersion, docSpec.kind);
+      }) as PromiseSettledResult<any>[]
+    )
+      .then(values => {
+        const docSpecsFromApi = values.map((value, index) => {
+          if (value.status === 'fulfilled') {
+            return {
+              data: value.value,
+              error: null,
+              kind: docSpecs[index].kind,
+            };
+          } else if (value.status === 'rejected') {
+            return {
+              data: null,
+              error: value.reason,
+              kind: docSpecs[index].kind,
+            };
+          }
+        });
+        setDocsLoading(false);
+        setDocs(docSpecsFromApi);
       })
-      .catch(err => {
-        setDocsError(t('Cannot load documentation: {{err}}', { err }));
+      .catch(() => {
+        setDocsLoading(false);
       });
   }, [docSpecs]);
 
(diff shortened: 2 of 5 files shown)


frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -15,42 +15,51 @@ export default function CreateButton() {
...
-        errorMessage: t('Failed to apply {{ newItemName }}.', { newItemName }),
+      clusterAction(() => applyFunc(newItemDefs), {
+        startMessage: t('Applying {{ newItemNames }}…', { newItemNames: resourceNames.join(',') }),
+        cancelledMessage: t('Cancelled applying {{ newItemNames }}.', {
+          newItemNames: resourceNames.join(','),
+        }),
+        successMessage: t('Applied {{ newItemNames }}.', { newItemNames: resourceNames.join(',') }),
+        errorMessage: t('Failed to apply {{ newItemNames }}.', {
+          newItemNames: resourceNames.join(','),
+        }),
```
joaquimrocha, line 62: I understand why you renamed `newItemName` but this will require the translations to be updated AFAIU. So if you change just the value as you did but keep the singular name, we don't have to update the translations.

frontend/src/components/common/Resource/DocsViewer.tsx
```diff
@@ -22,32 +21,54 @@ const useStyles = makeStyles(() => ({
...
-        setDocs(result?.properties || {});
+    setDocsLoading(true);
+    // fetch docSpecs for all the resources specified
+    Promise.allSettled(
+      docSpecs.map((docSpec: { apiVersion: string; kind: string }) => {
+        return getDocDefinitions(docSpec.apiVersion, docSpec.kind);
+      })
+    )
+      .then(values => {
+        //@ts-ignore
```
joaquimrocha, line 51: I'd prefer to avoid ts-ignores. If you need help fixing the types, we can look into it together.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -134,10 +140,10 @@ export default function EditorDialog(props: EditorDialogProps) {
     }
   }
 
-  function getObjectFromCode(code: string): KubeObjectInterface | null {
+  function getObjectsFromCode(code: string): KubeObjectInterface[] | null {
     try {
-      const codeObj = yaml.load(code);
-      return codeObj as KubeObjectInterface;
+      const codeObjs = yaml.loadAll(code);
+      return codeObjs as KubeObjectInterface[];
```
joaquimrocha, line 146: We now also support json. So I guess this is why we're having conflicts.

joaquimrocha (changes requested):

frontend/src/components/common/Resource/DocsViewer.tsx
```diff
@@ -13,7 +13,6 @@ import Loader from '../Loader';
 
 const useStyles = makeStyles(() => ({
   root: {
-    height: 216,
```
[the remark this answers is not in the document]
ashu8912: If we set a certain height to it then each tree view expands over the other present doc view, this is solved by giving a maxHeight to the DocViewer so that it scrolls as content grows.

frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -15,42 +15,51 @@ export default function CreateButton() {
...
+  const applyFunc = async (newItems: KubeObjectInterface[]) => {
+    // @ts-ignore
+    await Promise.allSettled(newItems.map(newItem => apply(newItem))).then((values: any) => {
+      values.forEach((value: any, index: number) => {
+        if (value.status === 'rejected') {
+          let msg = t('Something went wrong…');
+          const kind = newItems[index].kind;
+          const name = newItems[index].metadata.name;
+          const apiVersion = newItems[index].apiVersion;
+          msg = value.reason + ` kind: ${kind} name: ${name} apiVersion: ${apiVersion}`;
```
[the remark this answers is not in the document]
ashu8912: Doubt:
umm, i am not sure about it, i have changed the function signature now, so will it be fine if we treat a single entry in the array differently then multiple entries???

frontend/src/components/common/Resource/CreateButton.tsx
ashu8912: updated

frontend/src/components/common/Resource/EditorDialog.tsx
ashu8912: Fixed

joaquimrocha:
I have noticed that with these changes, if you switch to the docs tab, the dialog shrinks vertically. This is more evident when not having any resource defined, in which case it also doesn't show an error message:
Screenshot showing a tiny docs dialog

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -323,7 +323,7 @@ export default function EditorDialog(props: EditorDialogProps) {
                   {
                     label: t('frequent|Documentation'),
                     component: (
-                      <Box p={2} className={classes.scrollable} height="600px">
+                      <Box p={2} className={classes.scrollable} maxHeight={600}>
```
joaquimrocha, line 339: This is the reason why the dialog shrinks when docs have no content.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -130,14 +132,24 @@ export default function EditorDialog(props: EditorDialogProps) {
...
+    const firstChar = !!trimmedCode ? trimmedCode[0] : '';
+    if (['{', '['].includes(firstChar) || ['}', ']'].includes(lastChar)) {
+      return true;
+    }
+    return false;
+  }
+
   function onChange(value: string | undefined): void {
-    // Clear any ongoing attempts to check the code.
+   // Clear any ongoing attempts to check the code.
```
joaquimrocha, line 146: This is misaligned now.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -190,7 +194,7 @@ export default function EditorDialog(props: EditorDialogProps) {
     if (!format || (!res.obj && looksLikeJson(code))) {
       res.format = 'json';
       try {
-        res.obj = JSON.parse(code) as KubeObjectInterface;
+        res.obj = JSON.parse(code) as KubeObjectInterface[];
```
joaquimrocha, line 197: This will mask the case when the object is a single resource and not multiple.
Also, have you checked if kubectl accepts multiple resources as JSON?

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -210,34 +214,43 @@ export default function EditorDialog(props: EditorDialogProps) {
...
+      if(!Array.isArray(codeObjs[0])) {
+        setDocSpecs([codeObjs[0]])
+      } else {
+      const docSpecsFromCodeObjs = codeObjs[0]?.map((resource: KubeObjectInterface) => ({
+        kind: resource.kind,
+        apiVersion: resource.apiVersion,
+        error: resource === null,
+      }));
+      // @ts-ignore
+      setDocSpecs(docSpecsFromCodeObjs);
```
joaquimrocha, line 228: This is misaligned. Not sure why prettier hasn't fixed it.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -210,34 +214,43 @@ export default function EditorDialog(props: EditorDialogProps) {
...
   function onUndo() {
     setCode(originalCode);
   }
 
   function handleSave() {
-    // Verify the YAML / JSON even means anything before trying to use it.
-    const { obj, format, error } = getObjectFromCode(code);
-    if (!!error) {
-      setError(t('Error parsing the code: {{error}}', { error: error.message }));
+    // Verify the YAML even means anything before trying to use it.
```
joaquimrocha, line 238: This is misaligned.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -210,34 +214,43 @@ export default function EditorDialog(props: EditorDialogProps) {
...
-    const { kind, apiVersion } = (codeObj || {}) as KubeObjectInterface;
-    if (codeObj === null || (!!kind && !!apiVersion)) {
-      setDocSpecs({
-        error: codeObj === null,
-        kind,
-        apiVersion,
-      });
+    const { obj: codeObjs } = getObjectsFromCode(code);
+    if (codeObjs && codeObjs?.length > 0) {
+      if(!Array.isArray(codeObjs[0])) {
```
joaquimrocha, line 219: What's the case where the first object of the array is an array?

joaquimrocha (changes requested):

frontend/src/components/common/Resource/EditorDialog.tsx
ashu8912: This is when we have an array of multiple resources, the editor loads them at index 0 all inside it, if we have a single object as a resource then it stores it as the object only and not as array. That;s why we have to add this check

joaquimrocha: So the parsing will be:
* If we have a list A -> loaded as [A, ]
* If we have a single resource R -> R

This looks odd.

ashu8912: This is actually wrong i think i messed up somewhere with the format and parsing, I have a fix

frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -15,42 +15,53 @@ export default function CreateButton() {
...
-      }
-      setErrorMessage(msg);
-      setOpenDialog(true);
-      throw err;
-    }
+  const applyFunc = async (newItems: KubeObjectInterface[]) => {
+    await Promise.allSettled(newItems.map(newItem => apply(newItem))).then((values: any) => {
+      values.forEach((value: any, index: number) => {
+        if (value.status === 'rejected') {
+          let msg = t('Something went wrong…');
```
joaquimrocha, line 22: This value is always replaced below, so it will never be used AFAICS.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -152,45 +164,31 @@ export default function EditorDialog(props: EditorDialogProps) {
...
-    if (!res.obj) {
-      res.format = 'yaml';
-      try {
-        res.obj = yaml.load(code) as KubeObjectInterface;
-        return res;
-      } catch (e) {
-        res.error = new Error((e as Error).message || t('Invalid YAML'));
-      }
-    }
-
```
joaquimrocha, line 189: By doing this, you're switching the order in which we evaluate the code, which is risky as it can create more unforeseen problems.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -152,45 +164,31 @@ export default function EditorDialog(props: EditorDialogProps) {
...
     if (!format || (!res.obj && looksLikeJson(code))) {
       res.format = 'json';
       try {
-        res.obj = JSON.parse(code) as KubeObjectInterface;
+        res.obj = JSON.parse(code) as KubeObjectInterface[];
```
joaquimrocha, line 191: The `as` change here doesn't convert the value of the obj, only forces it to evaluate as that type (this is why it doesn't complain even when you parse a single JSON object).
As a result, you can try a JSON single object and it's currently not possible to handle it.

A fix for this is to first parse the code and check if it's an array, if not, add it as an array.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -247,6 +237,11 @@ export default function EditorDialog(props: EditorDialogProps) {
     if (format !== code.format) {
       setCode({ code: code.code, format });
     }
+
+    if (!getObjectsFromCode(code)) {
+      setError(t("Error parsing the code. Please verify it's valid YAML!"));
```
joaquimrocha, line 242: Let's add `...valid YAML or JSON!`

joaquimrocha (changes requested):
Almost there but applying single object for JSON is not working now.

frontend/src/components/common/Resource/EditorDialog.tsx
ashu8912: updated

frontend/src/components/common/Resource/EditorDialog.tsx
ashu8912: fixed

frontend/src/components/common/Resource/EditorDialog.tsx
ashu8912: this was removed by mistake, updated this

frontend/src/components/common/Resource/CreateButton.tsx
ashu8912: changed the msg to be appended with this string now

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -197,6 +197,16 @@ export default function EditorDialog(props: EditorDialogProps) {
...
     }
 
+    if (!res.obj) {
+      res.format = 'yaml';
+      try {
+        res.obj = yaml.loadAll(code) as KubeObjectInterface[];
+        return res;
+      } catch (e) {
+        res.error = new Error((e as Error).message || t('Invalid YAML'));
+      }
```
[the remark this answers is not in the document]
ashu8912: this was added to this commit by mistake

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -152,48 +164,51 @@ export default function EditorDialog(props: EditorDialogProps) {
...
+    const res: { obj: KubeObjectInterface[] | null; format: string; error: Error | null } = {
       obj: null,
       format,
       error: null,
     };
 
-    if (!res.obj) {
-      res.format = 'yaml';
+    if (!format || (!res.obj && looksLikeJson(code))) {
+      res.format = 'json';
```
joaquimrocha, line 189: I don't understand why this switched the order of processing YAML first and then JSON (I think this way we may attempt to convert to JSON all the time when the user may try YAML first).

joaquimrocha (approved):
Left a comment which I think we need to address but that requires testing again, so let's try what we have now.

PR #810, opened 2022-11-11
frontend: Prevent crash in labelSelector

joaquimrocha:


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -667,7 +667,7 @@ export function OwnedPodsSection(props: OwnedPodsSectionProps) {
 
   const queryData = {
     namespace: resource.kind === 'Namespace' ? resource.metadata.name : undefined,
-    labelSelector: labelSelectorToQuery(resource?.spec?.selector),
+    labelSelector: resource?.spec?.selector ? labelSelectorToQuery(resource?.spec?.selector) : '',
```
illume, line 670: Ah. The type checker is turned off for spec(it has an any type). All those any like `[otherProps: string]: any;` in cluster.ts are doing us a disservice.

ashu8912, line 670: can we move the logic of returning a empty string in case of no selector is provided into the ```labelSelectorToQuery``` function itself ???

ashu8912 (commented):
small nitpick

ashu8912 (approved):

illume (approved):
👍

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: I thought about that, but this implies accepting `LabelSelector | undefined` or similar (and then do the validation access || undefined), and in this case I think we should keep it expecting a valid labelSelector and filter out on the caller's side.

PR #817, opened 2022-11-15
Limit notifications

### LINKED ISSUE #815: Notifications: local storage exceeded quota
## Description

Failed to execute 'setItem' on 'Storage': Setting the value of 'notifications' exceeded the quota.

image

## Impact

Whole page crush.

image

Link to test how many chars your local storage can store. https://arty.name/localstorage.html

## Solution

Try to split notification to several chunks with a index array.

joaquimrocha:
This PR limits the message field of the notifications to 250 chars, and the number of notifications stored to 200 (by default).

Maybe we should limit more fields in the notification class, but this was an obvious one that we should control.

**How to test:**
- [ ] First check if there are more than 200 notifications in your current headlamp session (Local storage of the Application tab in Edge dev tools)
- [ ] Run this branch and wait for a notifiation to be received; check the stored notifications again: they should be now 200 (at most)
- Frontend tests are also added
fixes #815

### DIFF
--- a/frontend/src/lib/notification.tsx
+++ b/frontend/src/lib/notification.tsx
@@ -14,7 +14,7 @@ type NotificationMessageString = string;
 type OldNotificationDateArg = number | string;
 
 export class Notification {
-  message: string = '';
+  private _message: string = '';
   id: string;
   seen: boolean = false;
   url?: string;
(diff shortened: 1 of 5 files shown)


frontend/src/lib/notification.tsx
```diff
@@ -55,6 +55,18 @@ export class Notification {
     // generate the id based on the message and the date attached to a notification
     this.id = btoa(unescape(encodeURIComponent(`${this.date},${this.message},${this.cluster}`)));
   }
+
+  set message(message: string) {
+    this._message = message;
+    if (this._message.length > 250) {
```
illume, line 61: I wonder if this message shortening is needed if we are already limiting it to 200 notifications.  It's in general good to apply limits, so probably good to keep it anyway. It's not clear to me how long the message could be or what the typical size is.

illume (approved):
👍 thanks

frontend/src/lib/notification.tsx
joaquimrocha: I had a repeated notification which was less than 50 chars long. Very anecdotal but still all the notifications I had fit in well under 250 chars.
I think it's always okay to increase these limits in a controlled way, but I wanted to limit the chances of notifications hitting that limit, and the most variable size part of it is the message.

PR #831, opened 2022-11-17
frontend: Specify the locale for localeDates when testing

joaquimrocha:
Snapshots were failing if the locale of the developers' machines is
not the same.

These changes specify the mentioned locale and update the snapshots.


ashu8912 (approved):
Works tested with the add-notification-area PR

illume (approved):
👍

PR #832, opened 2022-11-17
frontend: Add feedback via update popup for release fetch

ashu8912:
We have a 5 second window for a update fetch, this patch adds feedback for all the actions that happen in that 5 sec window like fetching release and also if the release fetch failed and shows them in the update Snackbar

### DIFF
--- a/frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
+++ b/frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
@@ -7,8 +7,16 @@ import UpdatePopup from './UpdatePopup';
 
 export default function ReleaseNotes() {
   const { desktopApi } = window;
-  const [releaseNotes, setReleaseNotes] = React.useState<string>();
+  const [releaseNotes, setReleaseNotes] = React.useState<string>('');
   const [releaseDownloadURL, setReleaseDownloadURL] = React.useState<string | null>(null);
+  const [fetchingRelease, setFetchingRelease] = React.useState<boolean>(false);
+  const [releaseFetchFailed, setReleaseFetchFailed] = React.useState<boolean>(false);
+  const [skipFetch, setSkipFetch] = React.useState(false);
+
+  // network controller this makes sure if the octockit request is still lying around we
+  // abort it on fetch release skip press button click
+  const controller = new AbortController();
+  const signal = controller.signal;
 
   React.useEffect(() => {
     if (desktopApi) {
@@ -22,16 +30,24 @@ export default function ReleaseNotes() {
           }
 
           const octokit = new Octokit({
-            timeout: 5000,
+            timeout: 10000,
+            signal,
           });
 
           async function fetchRelease() {
+            // attach a timeout which checks after 5 second of fetching release
+            // if the release request was not successful
+            const timeoutID = setTimeout(() => {
+              setFetchingRelease(true);
+            }, 5000);
+
             const githubReleaseURL = `GET /repos/{owner}/{repo}/releases`;
             // get me all the releases -> default decreasing order of releases
             const response = await octokit.request(githubReleaseURL, {
               owner: 'kinvolk',
               repo: 'headlamp',
             });
+
             // Get the latest release that is not headlamp-plugin or headlamp-helm.
             const latestRelease = response.data.find(
               release => !release.name?.startsWith('headlamp-')
@@ -50,25 +66,28 @@ export default function ReleaseNotes() {
             let releaseNotes = '';
             if (storedAppVersion && semver.lt(storedAppVersion as string, currentBuildAppVersion)) {
               // get the release notes for the version with which the app was built with
-
               const githubReleaseURL = `GET /repos/{owner}/{repo}/releases/tags/v${currentBuildAppVersion}`;
               try {
                 const response = await octokit.request(githubReleaseURL, {
                   owner: 'kinvolk',
                   repo: 'headlamp',
                 });
-
                 const [notes] = response.data.body.split('<!-- end-release-notes -->');
                 if (!!notes) {
                   releaseNotes = notes;
                 }
               } catch (err) {
+                setReleaseFetchFailed(true);
                 console.error(
                   `Error getting release notes for version ${currentBuildAppVersion}:`,
                   err
                 );
               }
             }
+
+            // If all of the above was done before we need to warn the user, we don't need to warn them.
+            clearTimeout(timeoutID);
+            setFetchingRelease(false);
 
             // set the store version to the current so that we don't show release notes on
             // every start of app
@@ -84,13 +103,13 @@ export default function ReleaseNotes() {
           const isUpdateCheckingDisabled = JSON.parse(
             localStorage.getItem('disable_update_check') || 'false'
           );
-          if (!isUpdateCheckingDisabled) {
+          if (!isUpdateCheckingDisabled && !fetchingRelease && !skipFetch) {
             fetchRelease();
           }
         }
       );
     }
-  }, []);
+  }, [fetchingRelease, skipFetch]);
 
   React.useEffect(() => {
     desktopApi?.send('appConfig');
@@ -98,9 +117,21 @@ export default function ReleaseNotes() {
 
   return (
     <>
-      {releaseDownloadURL && <UpdatePopup releaseDownloadURL={releaseDownloadURL} />}
+      {
+        <UpdatePopup
+          releaseDownloadURL={releaseDownloadURL || ''}
+          fetchingRelease={fetchingRelease}
+          releaseFetchFailed={releaseFetchFailed}
+          skipUpdateHandler={() => {
+            // abort the octockit request to fetch github release
+            controller.abort();
+            setSkipFetch(false);
+            setFetchingRelease(false);
+          }}
+        />
+      }
       {releaseNotes && (
-        <ReleaseNotesModal appVersion={helpers.getAppVersion()} releaseNotes={releaseNotes} />
+        <ReleaseNotesModal releaseNotes={releaseNotes} appVersion={helpers.getAppVersion()} />
       )}
     </>
   );
--- a/frontend/src/components/common/ReleaseNotes/ReleaseNotesModal.tsx
+++ b/frontend/src/components/common/ReleaseNotes/ReleaseNotesModal.tsx
@@ -7,7 +7,7 @@ import { useTranslation } from 'react-i18next';
 import ReactMarkdown from 'react-markdown';
 
 export interface ReleaseNotesModalProps {
-  releaseNotes?: string;
+  releaseNotes: string;
   appVersion: string | null;
 }
 
@@ -28,6 +28,7 @@ export default function ReleaseNotesModal(props: ReleaseNotesModalProps) {
     minWidth: '50%',
     minHeight: '50%',
     overflow: 'auto',
+    maxWidth: '80%',
   };
 
   return (
@@ -50,7 +51,7 @@ export default function ReleaseNotesModal(props: ReleaseNotesModalProps) {
           className="markdown-body"
           style={{ color: theme.palette.text.primary, fontFamily: 'inherit' }}
         >
-          {releaseNotes && <ReactMarkdown>{releaseNotes}</ReactMarkdown>}
+          <ReactMarkdown>{releaseNotes}</ReactMarkdown>
         </Box>
       </Paper>
     </Modal>
--- a/frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
+++ b/frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
@@ -1,6 +1,16 @@
-import { Button, Snackbar } from '@material-ui/core';
+import { Icon } from '@iconify/react';
+import { Box, Button, makeStyles, Snackbar } from '@material-ui/core';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+
+const useStyles = makeStyles(() => ({
+  root: {
+    '& .MuiSnackbarContent-root': {
+      backgroundColor: 'rgb(49, 49, 49)',
+      color: '#fff',
+    },
+  },
+}));
 
 function UpdatePopup(props: {
   releaseDownloadURL?: string | null;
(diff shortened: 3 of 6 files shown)


ashu8912:
fixes #734

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,57 @@ import { Button, Snackbar } from '@material-ui/core';
...
-function UpdatePopup(props: { releaseDownloadURL: string }) {
+function UpdatePopup(props: {
+  releaseDownloadURL?: string | null;
+  fetchingRelease?: boolean;
+  releaseFetchFailed?: boolean;
+  skipUpdateCheck?: boolean;
+}) {
   const [show, setShow] = React.useState(true);
-  const { releaseDownloadURL } = props;
+  const { releaseDownloadURL, fetchingRelease, releaseFetchFailed, skipUpdateCheck } = props;
```
illume, line 12: Would be nice if there were storybook for these new states.

illume (approved):
👍 thanks

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,57 @@ import { Button, Snackbar } from '@material-ui/core';
...
 
+  if (fetchingRelease) {
+    return (
+      <Snackbar
+        open={show}
+        message={t('release|fetching release...')}
+        ContentProps={{
+          'aria-describedby': t('release|fetching release message'),
+        }}
+      />
```
joaquimrocha, line 23: This should have a skip button, because if it's taking too long, the user may want to skip fetching the info.

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,57 @@ import { Button, Snackbar } from '@material-ui/core';
...
+    );
+  }
+
+  if (releaseFetchFailed) {
+    return (
+      <Snackbar
+        open={show}
+        message={t('release|failed to fetch release')}
```
joaquimrocha, line 31: Please use `release|Failed to fetch release information.`

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,57 @@ import { Button, Snackbar } from '@material-ui/core';
...
+    return (
+      <Snackbar
+        open={show}
+        message={t('release|skipping update check')}
+        ContentProps={{
+          'aria-describedby': t('release|skipping update check message'),
+        }}
+      />
+    );
+  }
```
joaquimrocha, line 50: This shouldn't be shown.

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
joaquimrocha: It should also probably be delayed for 1/2 a sec (to give time for the release just to be checked without noise for the user) and hidden once the release info request is finished.

joaquimrocha (changes requested):

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,52 @@ import { Button, Snackbar } from '@material-ui/core';
...
+}) {
   const [show, setShow] = React.useState(true);
-  const { releaseDownloadURL } = props;
+  const [releaseFetchShow, setReleaseFetchShow] = React.useState(true);
+  const { releaseDownloadURL, fetchingRelease, releaseFetchFailed } = props;
   const { t } = useTranslation('frequent');
 
+  if (fetchingRelease) {
+    return (
+      <Snackbar
```
joaquimrocha, line 17: This and the next Snackbar declared below do not show up in the same location as the current release related snackbar. They should be consistent.

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,52 @@ import { Button, Snackbar } from '@material-ui/core';
...
+        open={releaseFetchShow}
+        message={t('release|fetching release...')}
+        ContentProps={{
+          'aria-describedby': t('release|fetching release message'),
+        }}
+        action={
+          <React.Fragment>
+            <Button color="primary" onClick={() => setReleaseFetchShow(false)}>
+              {t('frequent|Skip')}
+            </Button>
```
joaquimrocha, line 27: This doesn't skip fetching the release, just doesn't show this snackbar. More details in the final review comment.

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,52 @@ import { Button, Snackbar } from '@material-ui/core';
...
-  const { releaseDownloadURL } = props;
+  const [releaseFetchShow, setReleaseFetchShow] = React.useState(true);
+  const { releaseDownloadURL, fetchingRelease, releaseFetchFailed } = props;
   const { t } = useTranslation('frequent');
 
+  if (fetchingRelease) {
+    return (
+      <Snackbar
+        open={releaseFetchShow}
+        message={t('release|fetching release...')}
```
joaquimrocha, line 19: This should be `release|Fetching release information…`

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
```diff
@@ -2,11 +2,52 @@ import { Button, Snackbar } from '@material-ui/core';
...
+        ContentProps={{
+          'aria-describedby': t('release|fetching release message'),
```
joaquimrocha, line 21: This property is not supposed to get a description but rather an ID:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
ashu8912: fixed

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
ashu8912: fixed

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
ashu8912: updated

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
ashu8912: updated

joaquimrocha:
Testing the PR, I see this popup which keeps static:
Popup showing "Fetching release information"

joaquimrocha (changes requested):
I cannot test this like we could a while ago (by overriding the version with a `HEADLAMP_APP_VERSION` environment variable), so something is still not working correct.

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -85,20 +95,31 @@ export default function ReleaseNotes() {
             localStorage.getItem('disable_update_check') || 'false'
           );
           if (!isUpdateCheckingDisabled) {
-            fetchRelease();
+            if (!fetchingRelease && !skipFetch) {
```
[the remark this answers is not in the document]
ashu8912: fixed

frontend/src/components/common/ReleaseNotes/UpdatePopup.tsx
illume: There is a skip/dismiss button now. However I don't see where the "delay for 1/2 a sec" is.

illume:
Cool. There were updates (including to HEADLAMP_APP_VERSION). I marked as resolved the topics that were open that were fixed. There's just the "delay for 1/2 a sec" topic still open as far as I can see.

@joaquimrocha I think its ready for another review.

joaquimrocha (changes requested):
I have tested the latest changes and noticed that:
1. The "Fetching release information" is always shown after 10 secs (no matter whether the info was fetched or not);
2. The release notes' styling are broken somehow (not happening in `main`): 
Release notes without any contrast

ashu8912:
@joaquimrocha i have updated the fetching release issue, i was trying to reproduce the release notes modal but couldn't produce it

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -49,6 +62,17 @@ export default function ReleaseNotes() {
...
+              // if the release request was not successful
+              setTimeout(() => {
+                if (
+                  !releaseDownloadURLRef.current &&
+                  !releaseFetchFailedRef.current &&
+                  !releaseNotesRef.current
+                ) {
+                  setFetchingRelease(true);
+                }
+              }, 10000);
```
joaquimrocha, line 75: So I realize this will never happen because we have the 5 sec timeout on Octokit.
Maybe a good idea is to have this popup show up at about 3 secs after the UI is shown, so let's think 5 secs are enough. But we should then have the Octokit timeout as 10 secs. Let me know if this makes sense.

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -49,6 +62,17 @@ export default function ReleaseNotes() {
             const storedAppVersion = helpers.getAppVersion();
             let releaseNotes = '';
             if (storedAppVersion && semver.lt(storedAppVersion as string, currentBuildAppVersion)) {
+              // attach a timeout which checks after 10 second of fetching release
```
joaquimrocha, line 65: We already use Octokit before this to get the latest release. So if that's slow, I guess we'll be slow to get here.
This means we should likely have also the "fetching" popup being spawned after a timeout before that initial Octokit call.

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -49,6 +62,17 @@ export default function ReleaseNotes() {
...
+              // if the release request was not successful
+              setTimeout(() => {
```
joaquimrocha, line 67: For making this extra clean, we could assign the handler to a var and remove it after the Octokit call is successful.

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -77,30 +103,43 @@ export default function ReleaseNotes() {
             // Calling this after setting the version above, so the release notes have the right version
             // set when they show it.
             if (!!releaseNotes) {
+              releaseNotesRef.current = releaseNotes;
```
joaquimrocha, line 106: We should also call `setFetchingRelease(false)` here since it means that we've reached what we told the user we were doing, no need for the popup to be there.

frontend/src/components/common/ReleaseNotes/ReleaseNotesModal.tsx
```diff
@@ -7,7 +7,7 @@ import { useTranslation } from 'react-i18next';
 import ReactMarkdown from 'react-markdown';
 
 export interface ReleaseNotesModalProps {
-  releaseNotes?: string;
```
joaquimrocha, line 10: I don't think this commit is doing what the title says it's doing. But the title kind of hints to the problem I experienced in the screenshot I shared (if the text is white on white background -> screenshot makes sense).

joaquimrocha (changes requested):
It requires minimal changes now. Please see my comments.

frontend/src/components/common/ReleaseNotes/ReleaseNotesModal.tsx
ashu8912: updated

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
ashu8912: fixed

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
ashu8912: done

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
ashu8912: added at both places

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
ashu8912: fixed

joaquimrocha (changes requested):

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -22,16 +33,32 @@ export default function ReleaseNotes() {
...
+            // attach a timeout which checks after 5 second of fetching release
+            // if the release request was not successful
+            const timeoutID = setTimeout(() => {
+              if (
+                !releaseDownloadURLRef.current &&
+                !releaseFetchFailedRef.current &&
+                !releaseNotesRef.current &&
+                latestRelease &&
+                semver.gt(latestRelease.name as string, currentBuildAppVersion) &&
+                !process.env.FLATPAK_ID
```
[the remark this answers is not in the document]
ashu8912: lets do this in a different pr

frontend/src/components/common/ReleaseNotes/ReleaseNotes.tsx
```diff
@@ -22,16 +33,32 @@ export default function ReleaseNotes() {
...
           async function fetchRelease() {
+            // attach a timeout which checks after 5 second of fetching release
+            // if the release request was not successful
+            const timeoutID = setTimeout(() => {
+              if (
+                !releaseDownloadURLRef.current &&
+                !releaseFetchFailedRef.current &&
+                !releaseNotesRef.current &&
+                latestRelease &&
+                semver.gt(latestRelease.name as string, currentBuildAppVersion) &&
```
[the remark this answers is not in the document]
ashu8912: updated

joaquimrocha (approved):

PR #833, opened 2022-11-17
small storybook fixes for SimpleTable

illume:
- Fix storybook config to add port to mocked URLs
- Fix errCb guard to see if it's a function

This was the error shown in the console.
```ts
Failed to load resource: net::ERR_CONNECTION_REFUSED
apiProxy.ts:204 
        
       Unable to parse error json at url: http://localhost:4466/api/v1/namespaces Object with request data: Object
request @ apiProxy.ts:204
apiProxy.ts:583 
        
       Error in api request Object
run @ apiProxy.ts:583
apiProxy.ts:584 
        
       Uncaught (in promise) TypeError: errCb is not a function
    at run (apiProxy.ts:584:1)
```

## How to use

- Go to SimpleTable story http://localhost:6006/?path=/story/simpletable--name-search
- See no error (after this PR).

## Testing done

see ^

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -540,7 +540,7 @@ export async function streamResult(
       socket = stream(watchUrl, x => cb(x.object), { isJson: true });
     } catch (err) {
       console.error('Error in api request', { err, url });
-      if (errCb) errCb(err as ApiError, cancel);
+      if (errCb && typeof errCb === 'function') errCb(err as ApiError, cancel);
     }
   }
 
@@ -581,7 +581,9 @@ export async function streamResults(
       socket = stream(watchUrl, update, { isJson: true });
     } catch (err) {
       console.error('Error in api request', { err, url });
-      if (errCb) errCb(err as ApiError, cancel);
+      if (errCb && typeof errCb === 'function') {
+        errCb(err as ApiError, cancel);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -540,7 +540,7 @@ export async function streamResult(
       socket = stream(watchUrl, x => cb(x.object), { isJson: true });
     } catch (err) {
       console.error('Error in api request', { err, url });
-      if (errCb) errCb(err as ApiError, cancel);
+      if (errCb && typeof errCb === 'function') errCb(err as ApiError, cancel);
```
joaquimrocha, line 543: The `StreamErrCb` is a function, so I am surprised we need this check. Shouldn't Ts catch this?

illume: It's getting `{}` for some reason when it crashes. But I couldn't understand the code well enough to track it down.

Also, I think that errCb is optional, despite what the type says. Because other functions that take errCb are optional. I guess there are some `any` in there and that improving the typing would perhaps find the issue.

joaquimrocha: Please add a `// @todo` comment for us to track down the real cause of this and we then merge this if it fixes the issue.

illume: Added the todo comment.

PR #835, opened 2022-11-18
frontend: Don't wrap header cells by default

joaquimrocha:
Otherwise the sort buttons are sometimes shown in the next line and
make the UI look very quirky.


illume:
Can we make this more specific to the sort buttons? Because no wrapping with long headers could be a problem on low width views (mobile/etc).

joaquimrocha:
> Can we make this more specific to the sort buttons? Because no wrapping with long headers could be a problem on low width views (mobile/etc).

Do you mean only not wrapping if there's a sort button?

ashu8912 (approved):
LGTM!!

joaquimrocha:
This may be riskier than it seems for some users if we put it out so close to the release. Let's park it for now till the next release and see if @illume has a solution.

Issue #322, opened 2021-09-22
Re-read the kubeconfig file when needed


joaquimrocha:
## Current situation

If we have Headlamp (desktop) running and change the kubeconf file, e.g. to add, remove, or change a cluster, then Headlamp will not be aware of it until it is reloaded.

## Impact

If users want to add a new cluster and see it reflected in Headlamp, they cannot do so until they restart the program. Moreover, they are not even warned that they need to restart Headlamp.
Another related issue is that e.g. if Minikube is configured but not started, and we start Headlamp before starting Minikube, then the connections to Minikube will fail even after Minikube has been started.

## Ideal future situation

There is a way to quickly reflect the changes made to kubeconfig, in a running Headlamp.

## **Implementation options

We should detect changes to the kubeconfig file and display the reload banner+button (I believe we already have a similar banner related to internet connection outage). Upon clicking the reload button, the backend should re-read the kubeconfig.

## Additional information

This should be implemented only in the desktop version.

PR #841, opened 2022-11-23
Docker desktop setup instructions and change exposed port

yolossn:
Steps to test
[[Hard Mode]]
- `docker build . -t ghcr.io/kinvolk/headlamp:v0.13.0` in project home directory
- `cd  docker-extension`
- `docker build . -t headlamp-dd`
-  `docker extension install headlamp-dd`
-  Go to docker desktop dashboard and test

[[Easy mode]] 
- `docker extension install yolossn/headlamp-dd-extension`
-  Go to docker desktop dashboard and test

Flows:
1. The user has a `~/.kube/config` file with a proper cluster config -> Headlamp opens to the dashboard UI directly
2. The user has a `~/.kube/config` file but the cluster is not accessible -> The user gets the following message
`Please make sure the Kubernetes cluster is running and accessible, Recieved 'Bad Gateway' when trying to connect`
3. The user doesn't have a `~/.kube/config` file, the user is taken a screen which shows the steps to enable Kubernetes cluster in docker desktop. Once the cluster is up and running and accessible the user is redirected to the dashboard UI.


frontend/src/components/authchooser/index.tsx
```diff
@@ -261,9 +261,14 @@ export function PureAuthChooser({
             <Box alignItems="center" textAlign="center">
               <Box m={2}>
                 <Empty>
-                  {t('Failed to get authentication information: {{ errorMessage }}', {
-                    errorMessage: error!.message,
-                  })}
+                  {error && error.message === 'Bad Gateway'
+                    ? t(
+                        "Please make sure the Kubernetes cluster is running and accessible, Recieved '{{ errorMessage }}' when trying to connect",
```
illume, line 266: Please add these to the other locales too (so joaquimrocha can review/edit them more easily).

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
+import { Icon } from '@iconify/react';
```
illume, line 1: The folder name should match the component name. It's best practice to name the filename also after the component too... and just rexport it from the index.

Can you please add a DockerDesktopSetup.stories.tsx?

frontend/src/helpers/index.ts
```diff
@@ -37,6 +37,21 @@ function isElectron(): boolean {
   return false;
 }
 
+declare global {
+  interface Window {
+    ddClient: any | undefined;
+  }
+}
+
+// isDockerDesktop checks if ddClient is available in the window object
```
illume, line 46: This should follow jsdoc/tsdoc formatting https://tsdoc.org/

```ts
/**
 * isDockerDesktop checks if ddClient is available in the window object
 * if it is available then it is running in docker desktop
 * 
 * @see https://docker-desktop-docs.example.com
 * 
 * @returns true if Headlamp is running inside docker desktop
 */
function isDockerDesktop(): boolean {
  if (window?.ddClient === undefined) {
    return false;
  }
  return true;
}
```

illume (commented):
Looking GTM :)

Left a few notes for your consideration. Didn't try it yet.

backend/cmd/headlamp.go
```diff
@@ -190,7 +190,36 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
 	}
 }
 
-//nolint:gocognit,funlen,gocyclo
+func getContextFromKubeConfig(path string) []Context {
```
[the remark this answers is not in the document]
yolossn: I am not sure which spec/code you are referring to here. Are you referring to the kubeconfig spec ?

backend/cmd/headlamp.go
```diff
@@ -307,6 +314,35 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		proxy.ServeHTTP(w, r)
 	})
 
+	//nolint:gosec,scopelint
+	r.HandleFunc("/config/clusters/refresh", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
yolossn: This endpoint doesn't require authentication/authorization. It just rechecks if new contexts were added to the kubeconfigs and creates proxy for them.

docker-extension/docker-compose.yml
```diff
@@ -3,9 +3,9 @@ version: '3'
 services:
   headlamp:
     image: ghcr.io/kinvolk/headlamp:v0.13.0
-    command: ["--kubeconfig","/headlamp/config/config"]
+    command: ["--dev","--kubeconfig","/headlamp/config/config"]
```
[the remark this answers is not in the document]
yolossn: This is because the docker-desktop extension runs the frontend and the backend service separately. Right now the port is also changed to `64446` (to use a not-so-common port there by reducing the probability of port clash) without enabling the `dev` mode the frontend is getting `CORS` issue.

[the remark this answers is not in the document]
illume: What's the CORS issue? Dev mode disables all the CORS/security features. I'm not sure if that's what we want to do.

backend/cmd/headlamp.go
```diff
@@ -307,6 +314,35 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+		contexts := getContextFromKubeConfig(config.kubeConfigPath)
+
+		refreshedClusterProxies := 0
+
+		for _, context := range contexts {
```
illume, line 323: I don't think this handles contexts being removed?

What about if the context changes, but the name remains the same as an old one?

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+      history.push({
+        pathname: generatePath(getClusterPrefixedPath(), {
+          cluster: config['clusters'][0].name,
+        }),
+      });
+      setAuthenticated(true);
+    }
+  }
+
+  useEffect(() => {
```
illume, line 47: Refreshing config would be useful for the app/local version of headlamp too. This is a long running annoying thing for me.

I'm wondering if we enable the refresh config behaviour for all Headlamp?

backend/cmd/headlamp.go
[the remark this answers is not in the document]
illume: Ah yes, ok.

backend/cmd/headlamp.go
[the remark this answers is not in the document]
illume: Sorry, not sure what I was thinking with that spec part of the comment! Was probably thinking... this function needs a comment, and dreaming of another function. Anyway...

What about getContextFromKubeConfigs? (with the extra s at the end)

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+              <Button size="small" onClick={handleBack} disabled={activeStep === 0}>
+                <Icon icon="material-symbols:arrow-back-ios-new-rounded" />
+              </Button>
+            }
+          />
+        </Paper>
+      </DialogContentText>
+      <Box m={1} display="flex" justifyContent="flex-end" alignItems="flex-end">
+        <Button size="large" onClick={refreshConfigAndTestAuth} color="primary">
+          Done
```
illume, line 129: Done needs a translation.

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <Redirect to={'/'} />
+  ) : (
+    <ClusterDialog open useCover>
+      <DialogTitle>Enable Kubernetes in Docker Desktop to continue</DialogTitle>
+      <DialogContentText>
+        <Paper square elevation={3}>
+          <Card className={classes.card}>
+            <Typography variant="h6">{`${activeStep + 1}.${steps[activeStep].label}`}</Typography>
+            <Typography>{steps[activeStep].description}</Typography>
+            <img
```
illume, line 101: A material ui component should be used instead of img.

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const handleBack = () => {
+    setActiveStep(prevActiveStep => prevActiveStep - 1);
+  };
+
+  return authenticated ? (
+    <Redirect to={'/'} />
+  ) : (
+    <ClusterDialog open useCover>
+      <DialogTitle>Enable Kubernetes in Docker Desktop to continue</DialogTitle>
```
illume, line 95: This needs a translation.

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useEffect(() => {
+    const interval = setInterval(refreshConfigAndTestAuth, 30000);
+    refreshConfigAndTestAuth();
+    return () => clearInterval(interval);
+  }, []);
+
+  const steps = [
+    {
+      label: 'Go to Docker Desktop Settings',
```
illume, line 55: All these texts need translations.

frontend/src/lib/k8s/index.ts
```diff
@@ -135,6 +136,13 @@ export function useClustersConf(): ConfigState['clusters'] {
   return clusters;
 }
 
+export async function refreshConfig() {
+  const response = await post('/config/clusters/refresh', {}, false, {
```
illume, line 140: This probably needs to be wrapped in a try/catch in case it fails.

frontend/src/lib/k8s/index.ts
```diff
@@ -135,6 +136,13 @@ export function useClustersConf(): ConfigState['clusters'] {
...
+    timeout: 5 * 1000,
+  });
+  return response as ConfigState['clusters'];
```
illume, line 143: Probably should check if there's a 200 return code.

backend/cmd/headlamp.go
yolossn: Removing contexts can be quite tricky because now we support dynamic cluster. I will make changes to replace the ones with same name.

frontend/src/components/dockerDesktopSetup/index.tsx
yolossn: Makes sense, Created a new issue for the same
https://github.com/kinvolk/headlamp/issues/850

frontend/src/components/dockerDesktopSetup/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+        pathname: generatePath(getClusterPrefixedPath(), {
+          cluster: config['clusters'][0].name,
+        }),
+      });
+      setAuthenticated(true);
+    }
+  }
+
+  useEffect(() => {
+    const interval = setInterval(refreshConfigAndTestAuth, 30000);
```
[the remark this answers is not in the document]
yolossn: This can be coupled with
https://github.com/kinvolk/headlamp/issues/850

frontend/src/lib/k8s/index.ts
yolossn: The underlying `post` function which uses `request` takes care of status code checks and returns only `response.json()` as result.

backend/cmd/headlamp.go
```diff
@@ -307,6 +314,35 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		proxy.ServeHTTP(w, r)
 	})
 
+	//nolint:gosec,scopelint
```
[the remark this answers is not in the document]
yolossn: Reduced the `nolint` as much as possible. Most of them are because of the way existing code is structured. Fixing this will require a lot of yak shaving, so skipping this as of now as refactoring the backend is on my list of things to improve.

frontend/src/lib/k8s/index.ts
illume: Opening this back up. 

> The underlying post function which uses request takes care of status code checks and returns only response.json() as result.
Not true. It still needs error handling.

backend/cmd/headlamp.go
illume: It's not clear to me what you changed?

If it's not handling any of the cases, and can't be fixed... then can you please document what it can and can't do?

backend/cmd/headlamp.go
```diff
@@ -307,6 +316,36 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
 	}
 }
 
+// getContextFromKubeConfigs returns the context from the kubeconfig files.
```
illume, line 319: Comment says it returns a singular, which is a bit confusing.

illume (commented):
Looking good. Left a couple of notes.

backend/cmd/headlamp.go
yolossn: Added comments. Earlier there wasn't a way to differentiate if a context was created from kubeConfig or Dynamic Cluster request or For InCluster. Now I have added a "source" key to the context struct. And all contexts with the "KUBECONFIG"  source are dropped and the new contexts are set up for the clusters found in the kubeConfig.

frontend/src/lib/k8s/index.ts
yolossn: I have added a `try catch` block for handling errors but not sure how to check the response status code because the `request` function only returns `response.json()`. Can you point me to an example that I can refer to?

Reference:
https://github.com/kinvolk/headlamp/blob/d6df0869cd37af55f06f239219fcb4c1062517fd/frontend/src/lib/k8s/apiProxy.ts#L478-L487

https://github.com/kinvolk/headlamp/blob/d6df0869cd37af55f06f239219fcb4c1062517fd/frontend/src/lib/k8s/apiProxy.ts#L132

https://github.com/kinvolk/headlamp/blob/d6df0869cd37af55f06f239219fcb4c1062517fd/frontend/src/lib/k8s/apiProxy.ts#L218

frontend/src/components/DockerDesktopSetup/DockerDesktopSetup.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+      );
+    },
+  ],
+} as Meta;
+
+const Template: Story = args => {
+  return <DockerDesktopSetup refreshConfig={false} {...args} />;
+};
+
+export const DockerDesktopDefault = Template.bind({});
```
illume, line 25: Having the different states in here means they can be tested (both for regressions with storyshot snapshot testing, and with the UI).

> The idea is that you do all the messy “connected” logic in a single wrapper component in your app outside of Storybook
https://storybook.js.org/docs/react/writing-stories/build-pages-with-storybook#pure-presentational-pages

How to split this into a presentational component and a container component? All the state/effect/api calls/handlers would happen in a DockerDesktopSetup component. The DockerDesktopSetupPure component would take props for those state, only presentational parts. DockerDesktopSetup would pass all of these into DockerDesktopSetupPure. Then the storybook would test the DockerDesktopSetupPure component. Finally, you shouldn't have to have any mocks in there to test DockerDesktopSetupPure.

frontend/src/helpers/index.ts
```diff
@@ -98,6 +119,9 @@ function getAppUrl(): string {
     exportFunctions.isDevMode() || exportFunctions.isElectron()
       ? 'http://localhost:4466'
       : window.location.origin;
+  if (exportFunctions.isDockerDesktop()) {
+    url = 'http://localhost:64446';
+  }
```
joaquimrocha, line 124: I wonder if we should just set up an env var with the port and check it from here instead?

backend/cmd/headlamp.go
```diff
@@ -424,6 +451,43 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		proxy.ServeHTTP(w, r)
 	})
 
+	// the endpoint removes all the cluster contexts created from the KubeConfig,
+	// and reloads the contexts from the KubeConfig file.
+	// This is useful when the KubeConfig file is updated.
+	r.HandleFunc("/config/clusters/refresh", func(w http.ResponseWriter, r *http.Request) {
```
joaquimrocha, line 457: Can this be abused by blasting this endpoint (resulting in IO)?
An alternative would be to watch the directory for changes (we have an issue #322  ) but we have to think about the implications to the UX (browsing stuff while suddenly the config changes), so maybe one quick way to avoid this is to also pass the server a `--watch-kubeconfigs` option to listen to such changes only at will.

WDYT?
cc/ @illume / @ashu8912 / @yolossn

frontend/src/components/authchooser/index.tsx
joaquimrocha, line 266: There's a typo, and this can be simplified:

```suggestion
                        "Failed to connect. Please make sure the Kubernetes cluster is running and accessible. Error: {{ errorMessage }}",
```

joaquimrocha (changes requested):

frontend/src/lib/k8s/index.ts
illume: Yeah, the request checks the response.ok already. It's kind of an unfortunate API choice in some ways, since you can't check the response yourself. But it also means we don't need to handle all those cases ourselves. If you need to handle something other than 'ok' (200, 201 etc), then not using request() would be appropriate.

The request() returns `Promise.reject(error);` a few lines above the `return response.json()`.  So all not-ok responses should be handled by your try/catch.

The fetch API returns the response most of the time, except in exceptional circumstances (networks errors basically). https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

backend/cmd/headlamp.go
illume: I'm not sure about adding a --watch-kubeconfigs option. It would be useful for the Headlamp app in general to watch for kubeconfig changes(not just as a docker desktop extension), and I think @yolossn made an issue for that.

Auth for this URLs needs solving along with all of them when running as an app IMHO.

backend/cmd/headlamp.go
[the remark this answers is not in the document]
illume: Thanks. I've added an issue here https://github.com/kinvolk/headlamp/issues/855 for these specific cleanups.

illume:
Hello. Thanks for addressing all of those things. I went through and resolved a bunch of the open conversations.

frontend/src/components/DockerDesktopSetup/DockerDesktopSetup.stories.tsx
yolossn: Created a pure component and updated the storybook 👍🏽

backend/cmd/headlamp.go
yolossn: Even if we choose to add a watcher the refresh endpoint does make sense, we can also extend this to let the user manually refresh the app. As far as abuse is concerned we have to take care of this across the whole app as @illume pointed out.

illume: I agree that limiting the scope of this feature would be good. The /refresh URL shouldn't be available for in-cluster deployments IMHO. Because I can't think of a use case where that makes sense.

Having this on ONLY when running in docker extension is probably where we should start. Because the other configurations will not make use of this feature yet. In future PRs they may, but they don't yet. If this is best accomplished by adding a `--watch-kubeconfigs`, then that's what I vote for.

illume:
I reran the failing CI job, so I guess it's failing for real.

backend/cmd/headlamp.go
yolossn: AFAIK there isn't a way for the backend to find if it is running as a docker extension, in frontend the checks have been added to load the docker setup page only when it is running as a docker extension. So the /refresh endpoint will not be called when headlamp is running in in-cluster mode.

yolossn:
> I reran the failing CI job, so I guess it's failing for real.

Fixed it, thanks for pointing out.

backend/cmd/headlamp.go
joaquimrocha: > Auth for this URLs needs solving along with all of them when running as an app IMHO.

Right. But so far we've been careful and more restrictive with the things we allow in-cluster.
One thing is a user having services who are abusing Headlamp's endpoints but somehow had to make it to the user's machine in the first place. Another thing is to have that be the case for a true server use-case (hosted Headlamp).

So I am still a bit reluctant in adding this endpoint. Of course we can mark it as experimental and avoid it in-cluster, but if we would like anyway to watch for kube config, why not just watch for it as part of this PR (or as a separate one before this)? (the frontend can already get the config which would indicate that something changed by comparison as a lower bar)

Slightly related, in case you can thing of other implications:
In the near future we will have a welcome screen where the user is supposed to select the clusters they want to use in Headlamp and it shall keep a copy of that context in Headlamp's own config (so we can manage it independently from the main kube configs). Maybe there's an option we can give to the user where they just keep using the default kube configs in which case watching for kube configs would still be necessary.

frontend/src/lib/router.tsx
```diff
@@ -114,6 +115,15 @@ const defaultRoutes: {
...
   },
+  dockerDesktopSetup: {
+    path: '/docker-desktop-setup',
+    exact: true,
+    name: 'Docker Desktop Setup',
+    sidebar: null,
+    useClusterURL: false,
+    noAuthRequired: true,
+    component: () => <DockerDesktopSetup />,
+  },
```
joaquimrocha, line 126: I'd have preferred a link in order to save us time in the future, but since you already spent some time making this look good and since the UX is probably better, we can keep it, but please add this route only when running in DD.

joaquimrocha (changes requested):

yolossn:
Most part of why the solution was designed this particular way is because the idea was to solve "config refresh on change" only when the application is running in docker-desktop. Now that the wider goal is to fix the "config refresh on change" across the application, I would like to solve this in a different PR first and then alter this solution for docker-desktop.
cc: @joaquimrocha @illume @ashu8912

yolossn:
Planning to solve with the long term solution that was discussed as that will solve the kubeconfig refresh issue across the application instead of just in docker desktop environment.

PR #843, opened 2022-11-23
Fix navigation when URL params are enabled

joaquimrocha:
- frontend: Don't replace the location when updating URL params
- frontend: Optimize reflecting default states in the URL

How to test:
- [ ] Make sure the test instructions in #808 still apply
- [ ] Repeat the instructions (the ones which are possible) but using Headlamp in app mode


illume (approved):
👍 thanks for the detailed explanations

PR #844, opened 2022-11-24
frontend: Don't show bottom buttons in special sidebar

joaquimrocha:
The special sidebar should not have the plus nor the version buttons,
since they are tied to the concept of having a current cluster.


ashu8912 (approved):
Tested, LGTM!!!

illume (approved):
Cool. Makes sense. Thanks.

PR #854, opened 2022-11-30
frontend: The Plugin.initialize return value isn't used, remove

illume:
The return code for `Plugin.initialize()`:
- was never used by anything. Don't require a return value
- is now `void | boolean` not `void` for backwards compatibility

## How to use

Just a code cleanup.

## Testing done

- [x] CI tests pass (including plugin tests).

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -88,7 +88,6 @@ export async function initializePlugins() {
       const plugin = window.plugins[pluginName];
       try {
         // @todo: what should happen if this fails?
-        // @todo: The return code is not checked? What is it for?
         plugin.initialize(new Registry());
       } catch (e) {
         console.error(`Plugin initialize() error in ${pluginName}:`, e);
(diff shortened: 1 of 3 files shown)


joaquimrocha:
The reason why I added this return value was for being able to disable the plugins who report that they couldn't be initialized that way. e.g. My plugin only works on desktop -> disable after checking that.

illume:
> The reason why I added this return value was for being able to disable the plugins who report that they couldn't be initialized that way. e.g. My plugin only works on desktop -> disable after checking that.

Ah ok. I guessed since it was never used it'd be ok to remove.

Probably later we could add a function to signal a plugin which wants to disable itself, if we end up needing to do that. The benefit of that is that it'd work with plugins that aren't Plugin subclasses.

Are you ok if we remove this return value?

joaquimrocha:
> Are you ok if we remove this return value?

Your change is fine since it doesn't do away with the return type. I will merge it.

PR #856, opened 2022-12-01
frontend SimpleTable: Make nowrap only on sort having header cells

joaquimrocha:
So that other header cells still have the opportunity to wrap if they want to. Meaning there's less chance of a problem with header cells that used to fit with wrapping.

**Important:** The PR #835 was closed by mistake after I pushed what I believed was just the PR rebased on main. This PR has the changes @illume had made to it + rebased on main, so it should be fine.


illume (approved):
👍

PR #860, opened 2022-12-07
frontend: Fix editing resources

### LINKED ISSUE #858: Regression: Cannot edit resources
## Description

Since the changes that allow adding multiple resources, editing a resource is not working because the function is expecting one single resource but is getting a list of resources.

## Impact

This is severe enough that we should release a bug fix version.

## Environment and steps to reproduce

Just try to edit any resource in the cluster.

joaquimrocha:
The onSave function from the EditorDialog is now called with an array
of resources instead of just one resource, but the EditButton logic
was not changed to accommodate that.

This patch fixes that as well as the typing for the EditButton which
was expecting a KubeServiceAccount by mistake.

fixes #858 

How to test:
-  [ ] Edit any resource: it should allow to do so.

### DIFF
--- a/frontend/src/components/common/Resource/EditButton.tsx
+++ b/frontend/src/components/common/Resource/EditButton.tsx
@@ -2,8 +2,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useLocation } from 'react-router-dom';
-import { KubeObject } from '../../../lib/k8s/cluster';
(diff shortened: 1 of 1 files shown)


joaquimrocha:
One tricky issue related to this is that we broke compatibility by having the `EditorDialog.onSave` start being called with a list of resources instead of 1 single resource.
The `onSave` is not really typed, so we can argue that it always had to be checked, but it's still something we need to avoid.

@callmevladik , please note that if your plugins make use of the `EditorDialog` component, they may be broken by #858 .

illume (approved):
👍

PR #864, opened 2022-12-09
Add portforward to service

ashu8912:
Adds start portforward button to service details view

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"io/ioutil"
@@ -21,6 +22,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	oidc "github.com/coreos/go-oidc"
@@ -59,28 +61,34 @@ type HeadlampConfig struct {
 
 const PodAvailabilityCheckTimer = 5 // seconds
 
+const (
+	RUNNING = "Running"
+	STOPPED = "Stopped"
+)
+
 type PortForward struct {
-	ID        string `json:"id"`
-	closeChan chan struct{}
-	Pod       string `json:"pod"`
-	Service   string `json:"service"`
-	Namespace string `json:"namespace"`
-	Cluster   string `json:"cluster"`
-	Port      string `json:"port"`
-	Out       *bytes.Buffer
-	ErrOut    *bytes.Buffer
-	OutString string `json:"outString"`
-	ErrString string `json:"errString"`
+	ID               string `json:"id"`
+	closeChan        chan struct{}
+	Pod              string `json:"pod"`
+	Service          string `json:"service"`
+	ServiceNamespace string `json:"serviceNamespace"`
+	Namespace        string `json:"namespace"`
+	Cluster          string `json:"cluster"`
+	Port             string `json:"port"`
+	TargetPort       string `json:"targetPort"`
+	Status           string `json:"status"`
+	Error            string `json:"error"`
 }
 
 type PortForwardPayload struct {
-	ID         string `json:"id"`
-	Namespace  string `json:"namespace"`
-	Pod        string `json:"pod"`
-	Service    string `json:"service"`
-	TargetPort string `json:"targetPort"`
-	Cluster    string `json:"cluster"`
-	Port       string `json:"port"`
+	ID               string `json:"id"`
+	Namespace        string `json:"namespace"`
+	Pod              string `json:"pod"`
+	Service          string `json:"service"`
+	ServiceNamespace string `json:"serviceNamespace"`
+	TargetPort       string `json:"targetPort"`
+	Cluster          string `json:"cluster"`
+	Port             string `json:"port"`
 }
 
 type clientConfig struct {
@@ -111,41 +119,32 @@ func resetPlugins() {
 	pluginListURLs = nil
 }
 
-const (
-	PODS     = "pods"
-	SERVICES = "services"
-)
-
-var portForwards = make(map[string]map[string]map[string]PortForward)
+var portForwards = make(map[string][]PortForward)
 
 func portforwardstore(p PortForward) {
-	if portForwards[p.Cluster] == nil {
-		portForwards[p.Cluster] = make(map[string]map[string]PortForward)
+	// check if we already have a portforward with the same id if yes update it
+	for index, v := range portForwards[p.Cluster] {
+		if v.ID == p.ID {
+			portForwards[p.Cluster][index] = p
+			return
+		}
 	}
 
-	if p.Service != "" {
-		if portForwards[p.Cluster][SERVICES] == nil {
-			portForwards[p.Cluster][SERVICES] = make(map[string]PortForward)
-		}
-
-		portForwards[p.Cluster][SERVICES][p.ID] = p
-	} else if p.Pod != "" {
-		if portForwards[p.Cluster][PODS] == nil {
-			portForwards[p.Cluster][PODS] = make(map[string]PortForward)
-		}
-
-		portForwards[p.Cluster][PODS][p.ID] = p
-	}
+	portForwards[p.Cluster] = append(portForwards[p.Cluster], p)
 }
 
-func deletePortForward(cluster string, id string, podOrService string) error {
-	clusterPortForwardsForPodsOrService, ok := portForwards[cluster][podOrService]
+func stopOrDeletePortForward(cluster string, id string, isStopRequest bool) error {
+	clusterPortForwards, ok := portForwards[cluster]
 	if ok {
-		for key, v := range clusterPortForwardsForPodsOrService {
+		for index, v := range clusterPortForwards {
 			if v.ID == id {
-				v.closeChan <- struct{}{}
-
-				delete(clusterPortForwardsForPodsOrService, key)
+				if !isStopRequest {
+					portForwards[cluster] = append(clusterPortForwards[:index], clusterPortForwards[index+1:]...)
+				} else {
+					v.Status = STOPPED
+					v.closeChan <- struct{}{}
+					clusterPortForwards[index] = v
+				}
 
 				return nil
 			}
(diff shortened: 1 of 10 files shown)


ashu8912:
Still looking into the named port issue

backend/cmd/headlamp.go
```diff
@@ -60,28 +62,34 @@ type HeadlampConfig struct {
...
-	Pod        string `json:"pod"`
-	Service    string `json:"service"`
-	TargetPort string `json:"targetPort"`
-	Cluster    string `json:"cluster"`
-	Port       string `json:"port"`
+	ID               string `json:"id"`
+	Namespace        string `json:"namespace"`
+	Pod              string `json:"pod"`
+	Service          string `json:"service"`
+	ServiceNamespace string `json:"serviceNamespace"`
```
yolossn, line 89: Instead of having `Namespace`, `Pod`, `Service` and `ServiceNamespace` I think it is better to have a `Namespace`, `Name`  and `Type` (which takes they value `service` or `pod`)

backend/cmd/headlamp.go
```diff
@@ -105,41 +113,32 @@ func resetPlugins() {
 	pluginListURLs = nil
 }
 
-const (
-	PODS     = "pods"
-	SERVICES = "services"
-)
-
-var portForwards = make(map[string]map[string]map[string]PortForward)
+var portForwards = make(map[string][]PortForward)
```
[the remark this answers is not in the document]
ashu8912: The problem with one more nesting of accessing the Portforward is that on the frontend it becomes complex to mutate the list of items, with an array it is actually easy. And with writing to localStorage it becomes more difficult with how we mutate a single property in the map compared to an array item.

backend/cmd/headlamp.go
ashu8912: We have case like where we have podName, but there is service name too for that this would not be enough

joaquimrocha: So the service namespace can be different than the pod's namespace? I would assume that the service and the pod's name would be the same...

ashu8912: I found usecases where a service in a different namespace can control a pod from a different namespace, thus having these two seperate covers all the usecases for the service

frontend/src/components/service/Details.tsx
```diff
@@ -70,6 +71,13 @@ export default function ServiceDetails() {
                         <ValueLabel>{port}</ValueLabel>
                         <InlineIcon icon="mdi:chevron-right" />
                         <ValueLabel>{targetPort}</ValueLabel>
+                        <PortForward
+                          containerPort={targetPort}
+                          service={item}
+                          isPod={false}
```
joaquimrocha, line 77: I guess `isPod` is the same as `!service`, so I don't think we need this property (makes things simpler).
Probably the same can be done for name and namespace. The more logic we abstract in the component, the less we have to worry about maintaining how it is called.

frontend/src/lib/router.tsx
```diff
@@ -89,6 +89,8 @@ export interface Route {
   component: () => JSX.Element;
   /** Hide the appbar at the top. */
   hideAppBar?: boolean;
+  /** Desktop app specific route */
+  isDesktopAppOnlyRoute?: boolean;
```
joaquimrocha, line 94: This is already about routes, so it can be called `desktopOnly`.

frontend/src/lib/router.tsx
```diff
@@ -519,6 +521,14 @@ const defaultRoutes: {
     noAuthRequired: true,
     component: () => <NotificationList />,
   },
+  portforwards: {
+    path: '/portforwards',
+    exact: true,
+    name: 'PortForwards',
```
joaquimrocha, line 528: Should be Port Forwards

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -114,6 +114,11 @@ function prepareRoutes(t: (arg: string) => string) {
           name: 'ingresses',
           label: t('glossary|Ingresses'),
         },
+        {
+          name: 'portforwards',
+          label: t('glossary|PortForwards'),
```
joaquimrocha, line 119: Should be Port Forwards I think.

frontend/src/components/portforward/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+          },
+          {
+            label: 'Actions',
+            getter: portforward => {
+              const filteredOptions = options.filter((option: any) => {
+                if (portForwardInAction?.error) {
+                  return option === 'Delete';
+                }
+                if (portForwardInAction?.status === PORT_FORWARD_RUNNING_STATUS) {
+                  return option !== 'Start';
```
joaquimrocha, line 236: There are lots of strings that need to be translatable in this UI, including the ops, and when that's done, this logic needs to be changed.

frontend/src/components/portforward/index.tsx
```diff
@@ -187,7 +185,27 @@ export default function PortForwardingList() {
...
+                <Box display={'flex'} alignItems="center">
+                  <Box
+                    mr={1}
+                    onClick={() => {
+                      window.open(`http://localhost:${portforward.port}`, '_blank');
+                    }}
+                    style={{ color: '#2774B3', cursor: 'pointer' }}
+                  >
+                    {portforward.port}
+                  </Box>
```
joaquimrocha, line 203: Box components translate to divs, so we shouldn't use them for reactive components unless nothing else applies.
In this case, MUI has a Link component which is what we want to use.

frontend/src/components/portforward/index.tsx
```diff
@@ -190,19 +189,31 @@ export default function PortForwardingList() {
                   <Box
                     mr={1}
                     onClick={() => {
+                      if (portforward.status === PORT_FORWARD_STOP_STATUS) {
+                        return;
+                      }
```
joaquimrocha, line 194: This doesn't disable the hint that we can click the link. There should be another way to disable the link.

joaquimrocha (changes requested):
Found a few more things but I will take care of them.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -1,78 +1,167 @@
...
+      const portForwardsInStorage = localStorage.getItem(PORT_FORWARDS_STORAGE_KEY);
+      const parsedPortForwards = JSON.parse(portForwardsInStorage || '[]');
+      parsedPortForwards.forEach((portforward: any) => {
+        const isPortForwardInStorage = portForwards.find((pf: any) => pf.id === portforward.id);
+        if (!isPortForwardInStorage) {
+          portforward.status = PORT_FORWARD_STOP_STATUS;
+          massagedPortForwards.push(portforward);
+        }
+      });
+      localStorage.setItem(PORT_FORWARDS_STORAGE_KEY, JSON.stringify(massagedPortForwards));
```
joaquimrocha, line 114: I noticed another issue with the PR and it is related to this block:
1. Create a container port;
2. Restart the backend;
3. Go to the pod that had the container port -> it'll allow the user to start a new port.

This is related to how the ports concept is being managed here (where it's taking the backend's response as the source of truth).

Instead, it's the ports in the *local storage* (used created) that need to be the source of truth in what comes to what ports we have forwarded, as that reflects the user's intentions; the *backend* should be the source of the *state* for those ports.

So the logic should be rewritten as:
1. Get ports from the local storage;
2. Set the current portforward if it matches the current pod/service.
3. Set up the port's status from the backend response.

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -1,78 +1,167 @@
...
 
 export default function PortForward(props: PortForwardProps) {
-  const { isPod, containerPort, namespace, name, isPodRunning } = props;
+  const { containerPort, resource } = props;
+  const isPod = resource?.metadata?.kind !== 'Service';
+  const service = !isPod ? (resource as Service) : undefined;
+  const namespace = resource?.metadata?.namespace;
+  const name = resource?.metadata?.name;
   const [isCurrentPodOrServicePortForwarding, setIsCurrentPodOrServicePortForwarding] =
     React.useState(false);
```
joaquimrocha, line 71: Is this needed when we can just check if the `portForward` is set?

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -1,78 +1,167 @@
...
   function handlePortForward() {
-    if (!namespace || !containerPort || !cluster) {
+    if (!namespace || !cluster) {
       return;
     }
 
     setError(null);
 
     const massagedResourceName = name || '';
+    const massagedResourceNamespace = isPod ? namespace : pods[0].metadata.namespace;
```
joaquimrocha, line 138: I believe this can be called `podNamespace`

joaquimrocha (changes requested):
@ashu8912 , Here's another couple of things I found in the PR.

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: fixed by checking this later on a data where we have storage and backend portforward data, so there is no inconsistency

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: removed

frontend/src/components/common/Resource/PortForward.tsx
ashu8912: fixed

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -1,78 +1,167 @@
...
   const [isCurrentPodOrServicePortForwarding, setIsCurrentPodOrServicePortForwarding] =
     React.useState(false);
   const [error, setError] = React.useState(null);
   const [portForward, setPortForward] = React.useState<PortForwardState | null>(null);
   const [loading, setLoading] = React.useState(false);
   const cluster = getCluster();
-  const { t } = useTranslation('frequent');
+  const { t } = useTranslation(['frequent', 'resource']);
+  const [pods, podsFetchError] = Pod.useList({ labelSelector: getPodsSelectorFilter(service) });
+  const massagedContainerPort =
```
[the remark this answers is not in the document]
ashu8912: updated, also taking this as feedback to future variable namings.

PR #868, opened 2022-12-12
Allow terminal to reconnect after failure

### LINKED ISSUE #866: Exec to a pod results with "Failed to run cmd.exe" after few minutes
## Description

When trying to exec to any pod everything goes smoothly, but after almost 3 minutes "Failed to run cmd.exe" message appears making further terminal use unusable.

## Impact

User cannot use terminal longer than few minutes.

## Environment and steps to reproduce

To better understand a problem, small video file is attached with all required steps.

[https://youtu.be/_fMO1Ha9Qoc](https://youtu.be/_fMO1Ha9Qoc)

## Expected behavior

Exec into pod shouldn’t result in unusable terminal after a few minutes.

## Additional information

Tested on Headlamp version 0.14.0 using Google Chrome on Mac machine.

joaquimrocha:
When connecting or connected to a container (via exec), if it fails and Headlamp runs through the list of available shells, then there's no way to retry them (except for refreshing the page).

See #866 for a report where this happens. Even though this patch doesn't try to fix the cause for which the connection is lost after a couple of minutes (not even sure if we can avoid it), it does allow the user to recover because after it runs through the list of terminals it will display a message "Press enter to reconnect" and retry all shells again once the user presses enter.

In the future we shall add UI to select/enter a shell command and to directly refresh it, but for now this is a very non-intrusive solution.

fixes #866 

**How to test:**
- [ ] Connect to a container, then shut the network off and let the terminal get aware of that which should make it run through the list of terminals till the end. Turn the connection on and press Enter to reconnect as the instructions in the terminal say.
- [ ] Connect to a container that doesn't have a known shell and let all shells fail: see how the instructions are shown for the user to reconnect and press Enter to see it attempt it again.

### DIFF
--- a/frontend/src/components/common/Terminal.tsx
+++ b/frontend/src/components/common/Terminal.tsx
@@ -172,8 +172,11 @@ export default function Terminal(props: TerminalProps) {
       (async function () {
         send(4, `{"Width":${xterm.cols},"Height":${xterm.rows}}`);
       })();
-      xtermc.connected = true;
-      console.debug('Terminal is now connected');
+      // On server error, don't set it as connected
+      if (channel !== 3) {
+        xtermc.connected = true;
+        console.debug('Terminal is now connected');
+      }
     }
 
     if (isSuccessfulExitError(channel, text)) {
--- a/frontend/src/components/common/Terminal.tsx
+++ b/frontend/src/components/common/Terminal.tsx
@@ -18,6 +18,14 @@ import { Dialog } from './Dialog';
 const decoder = new TextDecoder('utf-8');
 const encoder = new TextEncoder();
 
+enum Channel {
+  StdIn = 0,
+  StdOut,
+  StdErr,
+  ServerError,
+  Resize,
+}
+
 const useStyle = makeStyles(theme => ({
   dialogContent: {
     height: '100%',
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/Terminal.tsx
```diff
@@ -172,8 +172,11 @@ export default function Terminal(props: TerminalProps) {
       (async function () {
         send(4, `{"Width":${xterm.cols},"Height":${xterm.rows}}`);
       })();
-      xtermc.connected = true;
-      console.debug('Terminal is now connected');
+      // On server error, don't set it as connected
+      if (channel !== 3) {
```
illume, line 176: I don't understand how `channel !== 3` means not an error.

frontend/src/components/common/Terminal.tsx
```diff
@@ -210,6 +223,11 @@ export default function Terminal(props: TerminalProps) {
       } else {
         xterm.write(t('Failed to connect…') + '\r\n');
       }
+
+      xterm.write('\r\n' + t('Press the enter key to reconnect.') + '\r\n');
```
illume, line 230: Is there a reason for doing this in the terminal, and not with... say a mui dialog/button?

frontend/src/components/common/Terminal.tsx
joaquimrocha: There's a list of what channels mean, just above that block.

joaquimrocha: Probably an enum would make this more readable.

frontend/src/components/common/Terminal.tsx
joaquimrocha: I did have a simple "Reconnect" button, but that's new UI we're adding and I'd like to do that in addition to allowing the user to set up the shell binary to run. So this is an unintrusive change IMO (we were already writing into the terminal when errors were found).

frontend/src/components/common/Terminal.tsx
```diff
@@ -18,6 +18,14 @@ import { Dialog } from './Dialog';
 const decoder = new TextDecoder('utf-8');
 const encoder = new TextEncoder();
 
+enum Channel {
```
illume, line 21: 👍 good idea

illume (approved):
👍

PR #870, opened 2022-12-14
Add easier auth buttons + make logs/exec dependent on permissions

### LINKED ISSUE #767: Access to pod logs and exec should be checked
## Description

Pod logs and exec are kind of k8s resources. The access to Pod logs and exec should be checked before display. Sometimes user only has namespace level access. 
image

joaquimrocha:
This PR adds a component that verifies the permissions for some action and only renders its children if those are valid.
It replaces the repeated logic for checking for permissions for actions such as edit/delete/etc.

This PR also makes the logs and exec actions based on permissions.

fixes #767 

**How to test:**
- [ ] Set up a token for a service account whose Role (create a new one based on the default _view_ one) is _view_ only; verify that in the role, there's no "pods/log" nor "pods/exec" in the list of the resources: go to a pod and verify that the logs and exec buttons are not visible (only the view button is visible)
- [ ] Edit the role and add `pods/log` and `pods/exec` to the a rule that has at least the `get` verb: verify that the buttons for a pod are visible upon refresh

### DIFF
--- a/frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
+++ b/frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
@@ -58,44 +58,10 @@ exports[`Storyshots Pod/PodDetailsView Error 1`] = `
               >
                 <div
                   class="MuiGrid-root MuiGrid-item"
-                >
-                  <button
-                    aria-label="logs"
-                    class="MuiButtonBase-root MuiIconButton-root"
-                    tabindex="0"
-                    title="Show Logs"
-                    type="button"
-                  >
-                    <span
-                      class="MuiIconButton-label"
-                    >
-                      <span />
-                    </span>
-                    <span
-                      class="MuiTouchRipple-root"
-                    />
-                  </button>
-                </div>
+                />
                 <div
                   class="MuiGrid-root MuiGrid-item"
-                >
-                  <button
-                    aria-label="terminal"
-                    class="MuiButtonBase-root MuiIconButton-root"
-                    tabindex="0"
-                    title="Terminal / Exec"
-                    type="button"
-                  >
-                    <span
-                      class="MuiIconButton-label"
-                    >
-                      <span />
-                    </span>
-                    <span
-                      class="MuiTouchRipple-root"
-                    />
-                  </button>
-                </div>
+                />
                 <div
                   class="MuiGrid-root MuiGrid-item"
                 />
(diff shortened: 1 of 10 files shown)


frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
```diff
@@ -58,44 +58,10 @@ exports[`Storyshots Pod/PodDetailsView Error 1`] = `
               >
                 <div
                   class="MuiGrid-root MuiGrid-item"
-                >
```
illume, line 61: These stories should now be tested with the AuthVisible true/false states. Otherwise we can't see these in the storybook anymore.

frontend/src/components/common/Resource/AuthBasedComponent.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export interface AuthBasedComponentProps extends React.PropsWithChildren<{}> {
+  item?: KubeObject;
+  authVerb?: string;
+  subresource?: string;
+  onError?: (err: Error) => void;
+  onUnauthorized?: () => void;
+}
+
+export default function AuthBasedComponent(props: AuthBasedComponentProps) {
```
illume, line 12: How about `AuthVisible` instead? `AuthBasedComponent` doesn't really spell what it does.

illume (changes requested):
This will be a nice improvement, thanks!

(Left a few notes.)

frontend/src/components/common/Resource/AuthBasedComponent.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { KubeObject } from '../../../lib/k8s/cluster';
+
+export interface AuthBasedComponentProps extends React.PropsWithChildren<{}> {
+  item?: KubeObject;
+  authVerb?: string;
```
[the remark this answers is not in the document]
joaquimrocha: You are right! They should be required. Maybe I copy pasted the props and only changed the name/type and forgot to change optionality. Good catch!

frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
joaquimrocha: Right. But the states being tested are the same as before, i.e. we need to make sure the UI looks the way it should for the status it has: no permissions. And this is the same as before (it's just the UI changes).

So we should add more stories for testing the UI when the state is such that the buttons must be visible. I think that can be added outside of this PR.

frontend/src/components/common/Resource/AuthBasedComponent.tsx
joaquimrocha: I struggled with the component name a bit. I had it as `AuthBasedButton` which makes sense as a name, but it wraps any component, not really just a button, so I renamed it.

I like your `AuthVisible` suggestion and will make the changes.

illume (approved):
Nice one. Thanks for that. 🎉

PR #875, opened 2022-12-19
Kubeconfig refresh

### LINKED ISSUE #322: Re-read the kubeconfig file when needed

yolossn:
Fixes #322

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -19,7 +19,6 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"time"
@@ -94,9 +93,16 @@ type spaHandler struct {
 	baseURL    string
 }
 
+const (
+	KubeConfig = 1 << iota
+	DynamicCluster
+	InCluster
+)
+
 type contextProxy struct {
 	context *Context
 	proxy   *httputil.ReverseProxy
+	source  int // Source indicates if contextProxy is configured from kubeconfig or dynamic cluster or incluster.
 }
 
 var pluginListURLs []string
@@ -316,9 +322,13 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	if !config.useInCluster {
 		// in-cluster mode is unlikely to want reloading plugins.
 		go watchForChanges(config.pluginDir)
+		// in-cluster mode is unlikely to want reloading kubeconfig.
+		go watchForKubeConfigChanges(config)
 	}
 
 	var contexts []Context
+
+	config.contextProxies = make(map[string]contextProxy)
 
 	// In-cluster
 	if config.useInCluster {
@@ -327,33 +337,20 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			log.Println("Failed to get in-cluster config", err)
 		}
 
-		contexts = append(contexts, *context)
+		proxy, err := config.createProxyForContext(*context)
+		if err != nil {
+			log.Printf("Error setting up proxy for context %s: %s\n", context.Name, err)
+		}
+
+		config.contextProxies[context.Name] = contextProxy{
+			context,
+			proxy,
+			InCluster,
+		}
 	}
 
 	// KubeConfig clusters
-	if kubeConfigPath != "" {
-		delimiter := ":"
-		if runtime.GOOS == "windows" {
-			delimiter = ";"
-		}
-
-		kubeConfigs := strings.Split(kubeConfigPath, delimiter)
-
-		for _, kubeConfig := range kubeConfigs {
-			kubeConfig, err := absPath(kubeConfig)
-			if err != nil {
-				log.Printf("Failed to resolve absolute path of :%s, error: %v\n", kubeConfig, err)
-				continue
-			}
-
-			contextsFound, err := GetContextsFromKubeConfigFile(kubeConfig)
-			if err != nil {
-				log.Println("Failed to get contexts from", kubeConfig, err)
-			}
-
-			contexts = append(contexts, contextsFound...)
-		}
-	}
+	contexts = append(contexts, getContextFromKubeConfigs(kubeConfigPath)...)
 
 	if config.staticDir != "" {
 		baseURLReplace(config.staticDir, config.baseURL)
@@ -367,8 +364,6 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		baseRoute := mux.NewRouter()
 		r = baseRoute.PathPrefix(config.baseURL).Subrouter()
 	}
-
-	config.contextProxies = make(map[string]contextProxy)
 
 	fmt.Println("*** Headlamp Server ***")
 	fmt.Println("  API Routers:")
@@ -390,6 +385,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			config.contextProxies[context.Name] = contextProxy{
 				context,
 				proxy,
+				KubeConfig,
 			}
 		}
 	}
@@ -1002,6 +998,7 @@ func (c *HeadlampConfig) addClusterSetupRoute(r *mux.Router) {
 		c.contextProxies[clusterReq.Name] = contextProxy{
 			&context,
 			proxy,
+			DynamicCluster,
 		}
 
 		if isReplacement {
--- a/backend/cmd/kubeconfig.go
+++ b/backend/cmd/kubeconfig.go
@@ -2,10 +2,15 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
+	"os"
+	"runtime"
 	"strings"
+	"time"
 
 	oidc "github.com/coreos/go-oidc"
+	"github.com/fsnotify/fsnotify"
 	"golang.org/x/oauth2"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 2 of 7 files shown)


frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,134 @@ const useStyle = makeStyles(theme => ({
 
 export interface LayoutProps {}
 
+const CLUSTER_FETCH_INTERVAL = 5 * 1000; // ms
```
yolossn, line 57:
```suggestion
const CLUSTER_FETCH_INTERVAL = 10 * 1000; // ms
```

illume (changes requested):
Could you please add a test for the golang part, and a story for the react component?

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+          dispatch(setConfig(configToStore));
+        } else {
+          Object.keys(clustersToConfig).every((key: string) => {
+            const clustersCopy = _.cloneDeep(clusters[key]);
+            // remove the cluster status from the comparison
+            delete clustersCopy.useToken;
+            if (_.isEqual(clustersCopy, clustersToConfig[key])) {
+              return true;
+            } else {
+              dispatch(setConfig(configToStore));
```
illume, line 167: It seems this would dispatch multiple times if there are multiple changes, but we probably only want to dispatch once when multiple ones have changed?

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+    );
+    fetchConfig();
+    return () => {
+      if (window.clusterConfigFetchHandler) {
+        clearInterval(window.clusterConfigFetchHandler);
+      }
+    };
+  }, []);
+
+  const fetchConfig = () => {
```
illume, line 142: This function could use documentation about what/why it's doing.

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
 export default function Layout({}: LayoutProps) {
   const classes = useStyle();
   const arePluginsLoaded = useTypedSelector(state => state.ui.pluginsLoaded);
-
+  const dispatch = useDispatch();
+  const clusters = useTypedSelector(state => state.config.clusters);
   const { t } = useTranslation('frequent');
+  const [isClusterNotAvailable, setIsClusterNotAvailable] = useState(false);
+
+  useEffect(() => {
```
illume, line 126: Probably all of this new logic should go into it's own component. New components should go into their own file named after the component. In this case it would make sense if all the related parts are in components inside a folder of components, and then only one component is imported and used. So that all the pieces implementing this feature are together in one place.

backend/cmd/headlamp.go
```diff
@@ -94,9 +93,18 @@ type spaHandler struct {
...
+	DynamicCluster
+	InCluster
+)
+
+// Source indicates if the context proxy is for a cluster
+// configured from kubeconfig or dynamic cluster or incluster.
 type contextProxy struct {
 	context *Context
 	proxy   *httputil.ReverseProxy
+	source  int
```
illume, line 107: Usually the documentation for a field would come after the field.

```golang
source   int // If the context proxy is ...
```

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+  const cluster = getCluster();
+  const { t } = useTranslation('frequent');
+  const isSidebarOpen = useTypedSelector(state => state.ui.sidebar.isSidebarOpen);
+  const history = useHistory();
+
+  if (!cluster) {
+    return null;
+  }
+
+  if (!isClusterNotAvailable) {
```
joaquimrocha, line 78: Nit: You could condense this as `!cluster || !isClusterNotAvailable`.

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+    return null;
+  }
+  return (
+    <>
```
joaquimrocha, line 82: AFAIU, you don't need the fragment here since you only have one element (Box) inside it.

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+          color={'error.main'}
+          bgcolor={theme.palette.common.white}
+          ml={1}
+          p={0.5}
+          onClick={() => {
+            history.replace('/');
+            changeClusterNotAvailableStateHandler(false);
+          }}
+        >
+          {t('frequent| Take me to cluster chooser')}
```
joaquimrocha, line 111: You don't need the `frequent` here, since it's the default namespace when you declared the `useTranslation`. If you keep this, please remove the space after the pipe.

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,133 @@ const useStyle = makeStyles(theme => ({
...
+          textAlign: 'center',
+        }}
+        bgcolor={'error.main'}
+        zIndex={1400}
+        width={'100%'}
+        p={0.5}
+        alignItems="center"
+      >
+        <Box ml={isSidebarOpen ? '-25vw' : 0}>
+          {t(`Something went wrong with cluster ${cluster}`)}
```
joaquimrocha, line 98: Should be
```
          {t(`Something went wrong with cluster {{ cluster }}`, {cluster})}
```

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -44,8 +44,10 @@ export default function RouteSwitcher() {
             requiresAuth={!route.noAuthRequired}
             requiresCluster={getRouteUseClusterURL(route)}
             exact={!!route.exact}
-            children={<RouteComponent route={route} />}
-            key={index}
+            clusters={useClustersConf()}
```
joaquimrocha, line 47: This is a hook, so be careful if it's being used conditionally here (hooks must be used conditionally). Maybe bring it earlier in the component's code.

joaquimrocha (changes requested):

backend/cmd/kubeconfig.go
```diff
@@ -123,3 +128,127 @@ func GetOwnContext(config *HeadlampConfig) (*Context, error) {
...
+		log.Println("Error watching for kube config changes:", err)
+		return
+	}
+	defer watcher.Close()
+
+	done := make(chan bool)
+
+	ticker := time.NewTicker(tickerDuration)
+
+	err = watcher.Add(path)
```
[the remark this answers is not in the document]
yolossn: Aah, good catch. Ill make the changes

backend/cmd/kubeconfig.go
```diff
@@ -123,3 +128,127 @@ func GetOwnContext(config *HeadlampConfig) (*Context, error) {
...
+	if err != nil {
+		log.Printf("Couldn't add %s to watcher: %v", path, err)
+	}
+
+	go func() {
+		for {
+			select {
+			// when the kubeconfig file is removed the watcher stops listening
+			// so this is a workaround to check if the file exists and add it to the watcher
```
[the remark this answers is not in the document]
yolossn: This cant be added when using the `fsnotify.Rename` or `fsnotify.Remove` event because the file is automatically removed from the watcher when the file is removed so we won't get those events in the first place.

frontend/src/components/App/Layout.tsx
ashu8912: added a comment about what it does

frontend/src/components/App/Layout.tsx
ashu8912: umm i am not sure if we want to create a new component out of this, the new addition just add a new fetchConfig function and passes new set of state to the already present RouteSwitcher component

frontend/src/components/App/Layout.tsx
ashu8912: updated

frontend/src/components/App/Layout.tsx
ashu8912: updated

frontend/src/components/App/Layout.tsx
ashu8912: fixed

frontend/src/components/App/RouteSwitcher.tsx
ashu8912: fixed, thanks

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,136 @@ const useStyle = makeStyles(theme => ({
...
+      () => {
+        fetchConfig();
+      },
+      CLUSTER_FETCH_INTERVAL,
+      clusters
+    );
+    fetchConfig();
+    return () => {
+      if (window.clusterConfigFetchHandler) {
+        clearInterval(window.clusterConfigFetchHandler);
```
joaquimrocha, line 132: I think this variable should be reset in moment, to avoid future misinterpretations.

frontend/src/components/App/Layout.tsx
```diff
@@ -44,11 +54,136 @@ const useStyle = makeStyles(theme => ({
...
+            const clustersCopy = _.cloneDeep(clusters[key]);
+            // remove the cluster status from the comparison
+            delete clustersCopy.useToken;
+            if (_.isEqual(clustersCopy, clustersToConfig[key])) {
+              return true;
+            } else {
+              isConfigDifferent = true;
+              return false;
+            }
+          });
```
joaquimrocha, line 170: This could be abstracted into its own function (e.g. `isClustersConfigDifferent(clusters, newClusters)`), to make it simpler to understand.

backend/cmd/kubeconfig.go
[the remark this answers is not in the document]
joaquimrocha: So when does these triggers happen? Maybe they're happening for directories only?

joaquimrocha (changes requested):

yolossn:
> So when does these triggers happen? Maybe they're happening for directories only?

The trigger is a 10 second ticker, every ten seconds the number of files being watched and the no of files in the config is compared it the no of files is less then the files are added to the watcher again.

joaquimrocha:
@yolossn , I have added a fixup commit that fixes 4 things:
    1. Corrects the translations' namespace;
    2. A bug when the clusters change in the config but the list has the same length (happens when replacing a cluster or renaming one)
    3. Avoids deep-cloning the cluster info when the useToken is not used.
    4. Improves the error banner: Replaces a clickable Box by an actual Button (important for the UI + a11y), disregards the sidebar in order to fix the alignment of the text, since the route is moved to login and in there there's no sidebar anyway.

Please review/test to see if it's all good.

yolossn:
Thanks for improvements.
Tested it and found that the Error banner is visible even when the cluster exists in the `/config` response
image

backend/cmd/kubeconfig.go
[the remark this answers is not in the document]
yolossn: Discussed this privately and updated the comment to make it clear.

yolossn:
I fixed the Error banner issue. Can someone test and review the changes? cc: @joaquimrocha @ashu8912

joaquimrocha:
This is working pretty well and is very useful. Thanks!

PR #876, opened 2022-12-20
Docker desktop setup instructions and change exposed port

yolossn:
Steps to test
[[Hard Mode]]
- `docker build . -t ghcr.io/kinvolk/headlamp:v0.15.0` in project home directory
- `cd  docker-extension`
- `docker build . -t headlamp-dd`
-  `docker extension install headlamp-dd`
-  Go to docker desktop dashboard and test

[[Easy mode]] 
- `docker extension install yolossn/headlamp-dd-extension`
-  Go to docker desktop dashboard and test

Flows:
1. The user has a `~/.kube/config` file with a proper cluster config -> Headlamp opens to the dashboard UI directly
2. The user has a `~/.kube/config` file but the cluster is not accessible -> The user gets the following message
`Failed to connect. Please make sure the Kubernetes cluster is running and accessible. Error: {{ errorMessage }}`
3. The user doesn't have a `~/.kube/config` file, the user is shown a message to setup kubernetes cluster or load a kube config file.

### DIFF
--- a/docker-extension/docker-compose.yml
+++ b/docker-extension/docker-compose.yml
@@ -1,11 +1,10 @@
 version: '3'
-
 services:
   headlamp:
-    image: ghcr.io/headlamp-k8s/headlamp:v0.13.0
-    command: ["--kubeconfig","/headlamp/config/config"]
+    image: ghcr.io/kinvolk/headlamp:v0.13.0
+    command: ["--kubeconfig","/headlamp/config/config", "--port","64446"]
     restart: unless-stopped
     volumes:
       - ~/.kube/:/headlamp/config:ro
     ports:
-      - 4466:4466
+      - 64446:64446
--- a/docker-extension/docker-compose.yml
+++ b/docker-extension/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   headlamp:
-    image: ghcr.io/kinvolk/headlamp:v0.13.0
+    image: ghcr.io/headlamp-k8s/headlamp:v0.15.1
     command: ["--kubeconfig","/headlamp/config/config", "--port","64446"]
     restart: unless-stopped
     volumes:
--- a/frontend/src/i18n/locales/pt/auth.json
+++ b/frontend/src/i18n/locales/pt/auth.json
@@ -14,5 +14,6 @@
   "Use A Token": "Use um token",
   "Failed to get authentication information: {{ errorMessage }}": "Falha ao obter informação de autenticação: {{ errorMessage }}",
   "No permissions to list pods.": "Sem permissões para listar pods",
-  "Redirecting to main page…": "A redireccionar para a página principal…"
+  "Redirecting to main page…": "A redireccionar para a página principal…",
+  "Failed to connect. Please make sure the Kubernetes cluster is running and accessible. Error: {{ errorMessage }}": ""
 }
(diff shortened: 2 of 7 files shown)


frontend/src/components/cluster/Chooser.tsx
```diff
@@ -416,23 +412,39 @@ function Chooser(props: ClusterDialogProps) {
...
+                    </DialogContentText>
+                    {helpers.isElectron() && (
+                      <DialogContentText>
+                        {t('Or try running Headlamp with a different kube config.')}
+                      </DialogContentText>
+                    )}
+                  </React.Fragment>
                 )}
-              </>
+              </React.Fragment>
```
illume, line 438: React.Fragment should be spelled `<></>` these days.

frontend/src/components/cluster/Chooser.tsx
```diff
@@ -416,23 +412,39 @@ function Chooser(props: ClusterDialogProps) {
...
             )}
           </React.Fragment>
         ) : (
-          <ClusterList clusters={clusterList} onButtonClick={handleButtonClick} />
+          <React.Fragment>
+            <DialogTitle id="chooser-dialog-title" focusTitle>
```
illume, line 443: Why isn't the title at the top anymore? There should always be a title, and now it seems there isn't always.

frontend/src/i18n/locales/pt/auth.json
```diff
@@ -14,5 +14,6 @@
   "Use A Token": "Use um token",
   "Failed to get authentication information: {{ errorMessage }}": "Falha ao obter informação de autenticação: {{ errorMessage }}",
   "No permissions to list pods.": "Sem permissões para listar pods",
-  "Redirecting to main page…": "A redireccionar para a página principal…"
+  "Redirecting to main page…": "A redireccionar para a página principal…",
+  "Failed to connect. Please make sure the Kubernetes cluster is running and accessible. Error: {{ errorMessage }}": "Failed to connect. Please make sure the Kubernetes cluster is running and accessible. Error: {{ errorMessage }}"
```
illume, line 18: I think ones which are not translated shouldn't be English, but should stay empty.

Hopefully Joaquim can add translations for these two... maybe in this PR.

illume (changes requested):
Left a few little notes.

frontend/src/components/cluster/Chooser.tsx
yolossn: The `Choose a cluster` title is being displayed even when there is no cluster to select from, so made it appear only when the clusters count is not equal to 0.

frontend/src/i18n/locales/pt/auth.json
joaquimrocha: @yolossn , this PR seems to require just the translation changes before we give a final review. Can you add that change?

docker-extension/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   headlamp:
-    image: ghcr.io/kinvolk/headlamp:v0.13.0
+    image: ghcr.io/kinvolk/headlamp:v0.15.0
```
ashu8912, line 4: IG this needs to be `ghcr.io/headlamp-k8s/headlamp:v0.15.0`

PR #877, opened 2022-12-22
Dockerfile: Implement least privileged user

illume:
So it's not run as root, it runs as a "headlamp" user.

It's a docker security best practice according to these guides:
- https://snyk.io/blog/10-docker-image-security-best-practices/
- https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md

Also, make the container workflow run when Dockerfile is changed.

## How to use

It should run as before.

## Testing done

- [x] CI passes e2e test 
- [x] works locally on minikube
- [x] add test for using plugins from mounts
- [ ] add test for using podman

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,5 +64,10 @@ COPY --from=backend-build --link /headlamp/backend/headlamp-server /headlamp/hea
 COPY --from=frontend --link /headlamp/frontend/build /headlamp/frontend
 COPY --from=frontend --link /headlamp/plugins /headlamp/plugins
 
+RUN addgroup -S headlamp && adduser -S headlamp -G headlamp
+RUN chown -R headlamp:headlamp /headlamp
+USER headlamp
+
 EXPOSE 4466
+
 ENTRYPOINT ["/headlamp/headlamp-server", "-html-static-dir", "/headlamp/frontend"]
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DOCKER_CMD ?= docker
 DOCKER_REPO ?= ghcr.io/headlamp-k8s
 DOCKER_EXT_REPO ?= docker.io/headlamp
 DOCKER_IMAGE_NAME ?= headlamp
+DOCKER_PLUGINS_IMAGE_NAME ?= plugins
 DOCKER_IMAGE_VERSION ?= $(shell git describe --tags --always --dirty)
 DOCKER_PLATFORM ?= local
 
@@ -88,6 +89,15 @@ image:
 	--platform=$(DOCKER_PLATFORM) \
 	-t $(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
 	Dockerfile \
+	.
+
+.PHONY: build-plugins-container
+build-plugins-container:
+	$(DOCKER_CMD) buildx build \
+	--pull \
+	--platform=linux/amd64 \
+	-t $(DOCKER_REPO)/$(DOCKER_PLUGINS_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
+	Dockerfile.plugins \
 	.
 
 docker-ext:
--- a/e2e-tests/kubernetes-headlamp-ci.yaml
+++ b/e2e-tests/kubernetes-headlamp-ci.yaml
@@ -1,6 +1,19 @@
-# For testing headlamp in CI.
-kind: Service
+# Source: headlamp/templates/pvc.yaml
 apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: headlamp
+  namespace: kube-system
+spec:
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 1Gi
+---
+# Source: headlamp/templates/service.yaml
+apiVersion: v1
+kind: Service
 metadata:
   name: headlamp
   namespace: kube-system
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -62,5 +62,12 @@ COPY --from=frontend /headlamp/plugins /headlamp/plugins
 # renamed it as "headlamp-server".
 RUN cd /headlamp && ln -s ./headlamp-server ./server
 
+RUN addgroup -S headlamp && adduser -S headlamp -G headlamp
+RUN chown -R headlamp:headlamp /headlamp
```
yolossn, line 66: If the user mounts the plugins in a directory other than `/headlamp` will this affect?

illume: Good point. Not sure, I will look into it.

illume: The plugins are [documented to be copied](https://headlamp.dev/docs/latest/development/plugins/building/#shipping-and-deploying-plugins) into `/headlamp/plugins` from a .plugins folder. So this does not affect the plugins copied in. Since headlamp/plugins is inside /headlamp and they have the headlamp:headlamp user.

I don't think this is a problem... can you please explain it some more?

yolossn: Yeah the documented/recommended way is to copy the plugins to `/headlamp/plugins` but the user can create a volumeMount to a directory other than `/headlamp/plugins` and then point to that plugins directory using the `--plugins-dir` flag. For example in [this](https://headlamp.dev/blog/2022/10/get-up-to-speed-deploying-headlamp-with-plugins/) blog post the plugins-dir is `/build/plugins`. In that case will the headlamp user/group need any other rights to access the plugin files.

illume: Understood, thanks.

joaquimrocha:
@illume , is the discussion over? Do these changes need to be updated?
Also, it'd be nice if these changes were tested with podman too.

illume:
> @illume , is the discussion over? Do these changes need to be updated?
>  Also, it'd be nice if these changes were tested with podman too.

No, the conversation is still unresolved. Last week I tried following the tutorial, but got lost. Since there's no tests for using mounts that way for plugins (or using podman either)... I think it's best some tests for those are added to this PR.

I've updated the todo list, and put this into draft mode.

knrt10:
@illume if it's okay, I can look into this?

illume:
Hi @knrt10 That would be great, thank you :)

Dockerfile
knrt10: Clarifying some things, once we create `headlamp` user in the Dockerfile, we can use the same user an group in k8s context using `securityContext`. We need to run the container as the user `100` and group `101` as the container needs to access the headlamp directory.

For example in [values.yaml](https://github.com/headlamp-k8s/headlamp/blob/main/charts/headlamp/values.yaml#L66) we can just do.

```bash
securityContext:
  # capabilities:
  #   drop:
  #   - ALL
  # readOnlyRootFilesystem: true
  runAsNonRoot: true
  privileged: false
  runAsUser: 100
  runAsGroup: 101
```

It doesn't matter how you use init containers since they are build differently and does not have `headlamp` user implemented.

knrt10:
I don't know why `values.yaml` changes are not applying in the CI, it's working fine when I install locally using helm. Let me debug more.

Also regarding CI on running charts test, I think it should build dockerfile and test against it. Since if any changes to dockerfile is made it still pulls `latest` headlamp-image. What are your thoughts folks?

cc @headlamp-k8s/devs

knrt10:
Failing charts test is unrelated, once we merge this PR and new docker image is pushed. I think they would pass. Since e2e tests are passing.

e2e-tests/kubernetes-headlamp-ci.yaml
```diff
@@ -10,34 +23,66 @@ spec:
...
-        k8s-app: headlamp
+        app.kubernetes.io/name: headlamp
+        app.kubernetes.io/instance: headlamp
     spec:
+      initContainers:
+      - command:
+        - /bin/sh
+        - -c
+        - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/
+        image: ghcr.io/yolossn/headlamp-plugins:latest
```
illume, line 55: I guess this needs to change?

illume: What is different in that image than the default headlamp image? Why can't it be used?
(ghcr.io/headlamp-k8s/headlamp:latest)

knrt10: cc @joaquimrocha can we push the same at `headlamp-k8s` repository?

knrt10: Sure, done. Please review

e2e-tests/kubernetes-headlamp-ci.yaml
```diff
@@ -10,34 +23,66 @@ spec:
...
       - name: headlamp
-        imagePullPolicy: Never
-        image: ghcr.io/headlamp-k8s/headlamp:latest
+        securityContext:
+          privileged: false
+          runAsGroup: 101
+          runAsNonRoot: true
+          runAsUser: 100
+        image: "ghcr.io/headlamp-k8s/headlamp:latest"
+        imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
knrt10: If we use `Always` it won't use the built image, it will download the image from the link. I think it should be `Never`, then it will use the one that we recently built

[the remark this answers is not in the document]
illume: Awesome, thanks.

illume:
> Failing charts test is unrelated, once we merge this PR and new docker image is pushed. I think they would pass. Since e2e tests are passing.

@yolossn can you please have a look at what to do about the charts failing in here?

yolossn:
> @yolossn can you please have a look at what to do about the "Lint and Test Charts" CI job failing in here?

This is because the Helm test pulls the latest docker image from ghcr and the changes added in the Dockerfile is not reflecting. I discussed this with @knrt10  He is changing the testing logic to use the local docker image.

e2e-tests/tests/headlamp.spec.ts
```diff
@@ -16,3 +16,13 @@ test("headlamp is there and so is minikube", async ({ page }) => {
 
   await expect(page).toHaveURL(/.*main/);
 });
+
+test('GET /plugins/list returns plugins list', async ({ page }) => {
+  const response = await page.goto('/plugins/list');
```
yolossn, line 21: Should this be `/plugins` ? When I run the backend from this branch I get 404 on `/plugins/list`

e2e-tests/tests/headlamp.spec.ts
```diff
@@ -16,3 +16,13 @@ test("headlamp is there and so is minikube", async ({ page }) => {
...
+  expect(response).toBeTruthy();
+
+  const text = await response.text();
+
+  expect(text).toBeTruthy();
+  expect(text.length).toBeGreaterThan(0);
```
yolossn, line 27: The endpoint returns an array of strings, when the array is empty it will return `[]` and the text length > 0 will still pass, the test case can be improved.

yolossn (changes requested):

e2e-tests/tests/headlamp.spec.ts
knrt10: Sure, make sense.

e2e-tests/tests/headlamp.spec.ts
knrt10: Weird, on web browser it gives the desired result

Screenshot 2023-05-09 at 7 31 59 AM

e2e-tests/tests/headlamp.spec.ts
```diff
@@ -16,3 +16,16 @@ test("headlamp is there and so is minikube", async ({ page }) => {
 
   await expect(page).toHaveURL(/.*main/);
 });
+
+test('GET /plugins/list returns plugins list', async ({ page }) => {
+  const response = await page.goto('/plugins/list');
+  expect(response).toBeTruthy();
+
+  const text = await response.text();
```
illume, line 24: This could be redone using some notes from the API testing guide? https://playwright.dev/docs/api-testing

Something like (untested):
```ts
  const plugins= await request.get(''/plugins/list'');
  expect(await plugins.json().length).toBeGreaterThan(0)
```

knrt10: thing is plugins.json() is returning error. I think it’s not sending the response in incorrect json format. That why I had to use text and check

illume: Seems you found a bug. I guess it's maybe not returning the correct headers or something.

knrt10: Also the value is like this.

```sh
"[\"plugins/app-menus/main.js\",\"plugins/change-logo/main.js\",\"plugins/cluster-chooser/main.js\",\"plugins/details-view/main.js\",\"plugins/dynamic-clusters/main.js\",\"plugins/pod-counter/main.js\",\"plugins/sidebar/main.js\"]
```

Notice there is no `"` in the end.

knrt10: So I checked it's not a bug. I was not building latest backend code for my headlamp. It is indeed returning. `application/json` in headers. And API endpoint is changed to `/plugins` and how the response is sent is also changes. So refactoring the code accordingly

Makefile
```diff
@@ -90,6 +91,15 @@ image:
 	Dockerfile \
 	.
 
+.PHONY: plugins
```
yolossn, line 94: I think this Make recipe name can be a bit verbose. Apart from that rest looks good.

yolossn (approved):

illume:
Great work :) Thanks @knrt10 (and also thanks @yolossn )

PR #892, opened 2023-01-03
Allow Overriding token management

ashu8912:
We want to have the ability for the plugins to override the getToken and setToken methods and this patch provides the mechanism to do that.

### DIFF
--- a/frontend/src/plugin/registry.tsx
+++ b/frontend/src/plugin/registry.tsx
@@ -13,6 +13,7 @@ import {
   setClusterChooserButtonComponent,
   setDetailsView,
   setDetailsViewHeaderAction,
+  setFunctionsToOverride,
   setRoute,
   setRouteFilter,
   setSidebarItem,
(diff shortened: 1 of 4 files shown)


frontend/src/components/App/TopBar.tsx
```diff
@@ -48,12 +49,17 @@ export default function TopBar({}: TopBarProps) {
   const history = useHistory();
 
   function hasToken() {
-    return !!cluster ? !!getToken(cluster) : false;
+    const getTokenMethodToUse = store.getState().ui.functionsToOverride.getToken;
```
joaquimrocha, line 52: In our conversation, what I had mentioned was that you could use the `store.getState().ui.functionsToOverride.getToken` inside the getToken(...)'s implementation. I see that maybe that will get us into trouble if plugin developers want to call the original method within their overridden method (which could end up calling itself), but maybe that's fine if we document this fact.

The reason why I think we have to really call the overridden method from the `getToken` is that developers maybe expect that this method will have the overridden behavior if any plugin overrides it.
And it also makes the whole code simpler.

joaquimrocha (changes requested):
Left a comment which should simplify the changes a lot.

Also, I appreciate that it is useful to have a single way to define overridden functions, and we can keep that internally, but when it comes to exposing it to the user, this method (`registerFunctionsToOverride`) will likely confuse the user (what can they override? why not function xYZ?) so we should expose this as `registerGetToken`.

If it's really useful to have the generic register method exposed, then maybe `registerFunctionOverride(...)` is fine, but it should receive and document what keys to set (and maybe there should be an enum of keys). So this gets way more complex than we probably want to make it ATM.

frontend/src/components/App/TopBar.tsx
ashu8912: updated

frontend/src/plugin/registry.tsx
```diff
@@ -384,3 +389,18 @@ export function registerAppLogo(logo: AppLogoType) {
...
+ * @example
+ * ```ts
+ * registerFunctionsToOverride({
+ *  getClusterName: () => 'my-cluster',
+ *  getToken: () => 'token'
+ * });
+ */
+export function registerFunctionsToOverride(override: FunctionsToOverride) {
+  store.dispatch(setFunctionsToOverride(override));
+}
```
joaquimrocha, line 406: What I meant in my previous review is that we shouldn't export this function since we don't have a clear plan yet to use it. So we can keep the internal redux logic you added, but let's export a `registerSetTokenFunction()` and similar for `getToken`.

Later if we have a clear plan or many functions to override, we can bring back this `registerFunctionsToOverride` function easily (since we'll keep the redux internals).

joaquimrocha (changes requested):
I had to clarify my previous review. Let me know if this makes sense.

joaquimrocha:
@ashu8912 , I have added a fixup commit, can you please verify it makes sense?

joaquimrocha (approved):

Issue #898, opened 2023-01-09
Failing to load preload.js


joaquimrocha:
## Description

We use the preload.js script to inject the desktopApi object in the global window object as it's necessary for IPC in the app.
The script is not getting loaded for some reason, so none of the IPC related functionality is working (fetching release info, changing languages, etc.).

![VM4 sandbox_bundle:2 Unable to load preload script: /home/jrocha/headlamp/app/electron/preload.js
(anonymous) @ VM4 sandbox_bundle:2
VM4 sandbox_bundle:2 Error: module not found: @babel/runtime/helpers/interopRequireDefault
    at preloadRequire (VM4 sandbox_bundle:2:82516)
    at <anonymous>:4:30
    at runPreloadScript (VM4 sandbox_bundle:2:83356)
    at VM4 sandbox_bundle:2:83615
    at VM4 sandbox_bundle:2:83770
    at ___electron_webpack_init__ (VM4 sandbox_bundle:2:83774)
    at VM4 sandbox_bundle:2:83897](https://user-images.githubusercontent.com/1029635/211375273-0726e28c-7d99-47d8-a48f-18b8f4f6590c.png)

PR #901, opened 2023-01-10
app: Remove @babel/runtime instead use regenerator-runtime

ashu8912:
There was an issue with @babel/runtime not loading the correct module when running the app. This was fixed by using @babel/polyfill moreover we don't need the entire @babel/runtime and @babel/plugin-transform-runtime when we only have two typescript files to be compiled.

## Testing done
Run the app and all the preload related settings will start working now


ashu8912:
fixes #898

joaquimrocha (approved):
Very cool. Thanks.

Issue #899, opened 2023-01-09
Doc Viewer not loading for the desktop app build


ashu8912:
## Description

While viewing the doc viewer in the latest release of the desktop app build it fails to load with an error message saying buffer not found, we added a past fix for this but looks like there is some issue with how the buffer package is pollyfilled for the desktop app.

## Impact

user can not see the doc viewer for the desktop app.

## Environment and steps to reproduce

1. Go to the editor 
2. Paste a valid YAML file
3. Go to  Doc Viewer on the editor
4. The doc viewer will not loads the docs

## Expected behavior

The doc viewer should show the doc for the particular resource.

PR #905, opened 2023-01-13
frontend: Fix doc viewer breaking

ashu8912:
With a recent CRA update node related modules are not shipped with CRA anymore. This broke the doc viewer as it was using the buffer module. We already had a patch earlier that added the polyfill but one problem with polyfilling it in a place like index.tsx is that when building the frontend for production, the polyfill is not included. The correct place to polyfill is to include it at the place where it is being used.


ashu8912:
fixes #899

PR #907, opened 2023-01-16
FIx a small issue with how we check for NaN values in useURLState

ashu8912:
The NaN check was done in a wrong way  for useURLState, this PR fixes that


joaquimrocha (approved):
LGTM. Let's wait for the CI to validate before merging.

PR #915, opened 2023-01-23
Fix notifications

joaquimrocha:
The message was being stored as _message in the local storage due to being its variable name in the class and it would not get correctly assigned nor would the stored notifications be unmarshalled as Notification objects.
    
This patch fixes that and guarantees that the notifications' marshalling is under control.

How to test:
 [ ] Verify that the notifications do have a title when they before didn't.


ashu8912 (approved):
Tested, LGTM

PR #917, opened 2023-01-25
Move settings in redux store

joaquimrocha:
These changes move the settings entries to the config redux story.
More context in the commit messages.

A new story is also added.


ashu8912 (approved):
LGTM!!!

PR #920, opened 2023-01-26
frontend: Format age column in event list (in resources' section)

joaquimrocha:
This patch fixes how the age columns is getting calculated and also
moves it to be the last column, in order to be consistent with how we
order the columns across Headlamp.

### DIFF
--- a/frontend/src/components/common/ObjectEventList.tsx
+++ b/frontend/src/components/common/ObjectEventList.tsx
@@ -1,9 +1,9 @@
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { KubeObject } from '../../lib/k8s/cluster';
-import Event from '../../lib/k8s/event';
-import { timeAgo } from '../../lib/util';
-import { SectionBox, SimpleTable } from '../common';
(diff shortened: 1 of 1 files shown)


yolossn:
The `age` column was not kept as the last column to match the results produced by kubectl describe output, unlike the Kubernetes objects creation time, the event time here plays an important role so I think th existing order makes sense.

Attaching kubectl output for reference:
image

joaquimrocha:
We're not trying to match the order of columns from the kubectl output as the kubectl also doesn't show the age as the last col in other cases. But I'd still like to keep it consistent with the rest of Headlamp.
If later turns out we should match the column order from kubectl, that's a decision we'd have to make for the entirety of the UI.

yolossn:
I agree there is no need to match the kubectl output, but in this case it makes sense to have age column in the existing order as there is more significance to the time when the event occurred. It boils down to prioritizing functionality over consistency.

joaquimrocha:
As mentioned in the team call, I don't think we are sacrificing functionality, rather I think we're improving it by having a more consistent way for looking for the age col.

PR #925, opened 2023-01-31
Fix portforward list

ashu8912:
We were not filtering out the list view of portforward based on the cluster they are part of, this patch fixes that and only shows portforwards that are part of the current cluster.


illume (approved):
👍

joaquimrocha (approved):

joaquimrocha:
I've tried it and have the exact same issue in the backend when starting a new forward.

ashu8912:
@joaquimrocha #927 fixes the above, but we still need this PR

PR #926, opened 2023-01-31
frontend: Add pluginLib.Crd for exposing lib/k8s/crd

illume:
So plugins can use it.

window.pluginLib.Crd should be available.

This is required so tests can use lib/k8s/crd. To make it more convenient for future headlamp-plugin versions, the headlamp needs to expose this. Related to https://github.com/headlamp-k8s/headlamp/pull/795


joaquimrocha (approved):
FTR: This is meant as a work around till we properly fix all exports from the crd module (and others).

PR #929, opened 2023-02-01
Fix sidebar on locale

joaquimrocha:
How to test:
- Go to settings, choose a non-English locale
- Go to the cluster chooser or some other area
- Click settings again, and then click the Settings sidebar entry or the Notifications one -> it should still show the special sidebar (Cluster + Notifications + Settings) as opposed to the regular sidebar

### DIFF
--- a/frontend/src/components/Sidebar/Sidebar.tsx
+++ b/frontend/src/components/Sidebar/Sidebar.tsx
@@ -69,19 +69,19 @@ export default function Sidebar() {
   const { t, i18n } = useTranslation(['glossary', 'frequent']);
   const specialSidebarOptions: SidebarItemProps[] = [
     {
-      name: t('frequent|back'),
+      name: 'clusters',
       icon: 'mdi:hexagon-multiple',
       label: t('glossary|Cluster'),
       url: '/',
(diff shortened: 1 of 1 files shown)


frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -69,19 +69,19 @@ export default function Sidebar() {
   const { t, i18n } = useTranslation(['glossary', 'frequent']);
   const specialSidebarOptions: SidebarItemProps[] = [
     {
-      name: t('frequent|back'),
+      name: 'back',
       icon: 'mdi:hexagon-multiple',
       label: t('glossary|Cluster'),
```
illume, line 74: I wonder if you checked if the label has the same problem?

joaquimrocha: The label doesn't have the same problem because that's the user-facing text and it should be translated.
We could rename this item as `clusters` instead of `back` though. Let me do that.

ashu8912 (approved):
Tested with 18n-fixes branch, LGTM!!

PR #933, opened 2023-02-02
app: Update repo org in push-release-assets script

joaquimrocha:
Otherwise the assets don't actually get pushed.


illume:
I re-ran the failed build-windows job. Here's the failing one if anyone is interested: https://github.com/headlamp-k8s/headlamp/actions/runs/4076804620/jobs/7025028937

illume (approved):
👍

PR #935, opened 2023-02-03
helm: update chart URL to headlamp-k8s namespace

milas:
This only updates the instances of the `kinvolk.github.io` URLs to `headlamp-k8s.github.io` as GitHub Pages doesn't do automatic redirects which means the docs don't work to add the Helm repo.

Any `github.com/kinvolk/headlamp` references have been left as-is.


yolossn (approved):
Thanks for the contribution

PR #937, opened 2023-02-03
ci: Update image name in container publish action

yolossn:


joaquimrocha:
I didn't expect these things to fail due to the org moving. We should grep for kinvolk and make sure we update any GitHub org references.

PR #938, opened 2023-02-06
frontend: Add start and stop to cronJob

ashu8912:
Our users might want to suspend or resume a cronJob so we allow doing that from the UI via header actions to CronJob detail view


ashu8912:
https://user-images.githubusercontent.com/30126128/216942064-c5ee0b09-ff54-4d60-816a-eb913b59f308.mov

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -182,11 +182,19 @@ export interface DetailsGridProps
   name: string;
   namespace?: string;
   sectionsFunc?: (item: KubeObject) => React.ReactNode;
+  actions?: JSX.Element[] | ((item: any) => any);
```
joaquimrocha, line 185: This is not needed because this `DetailsGridProps` already extends `MainInfoSectionProps` which supports the `actions` property.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -12,7 +18,30 @@ export default function CronJobDetails() {
...
   const [jobs, jobsError] = Job.useList();
+  const [cronJob, cronJobError] = CronJob.useGet(name, namespace);
+  const [isCronSuspended, setIsCronSuspended] = useState(false);
+  const [isCheckingCronSuspendStatus, setIsCheckingCronSuspendStatus] = useState(true);
+  const dispatch = useDispatch();
 
+  function applyFunc(newItem: KubeObjectInterface) {
+    return apply(newItem)
+      .then(value => {
+        console.log('success value is ', value);
```
joaquimrocha, line 29: left over from debug?

frontend/src/components/cronjob/Details.tsx
```diff
@@ -12,7 +18,30 @@ export default function CronJobDetails() {
...
+        setIsCronSuspended(!isCronSuspended);
```
joaquimrocha, line 30: A best practice for doing this, due to how the values may differ from expected, is to do:
`setIsCronSuspended(isCronSuspended => !isCronSuspended);`

joaquimrocha: Actually is this even necessary? AFAIK, the proper value here will be done when the `cronJob` object gets updated.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -12,7 +18,30 @@ export default function CronJobDetails() {
...
+      })
+      .finally(() => {
+        setIsCheckingCronSuspendStatus(false);
```
joaquimrocha, line 33: Ditto for this.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -12,7 +18,30 @@ export default function CronJobDetails() {
...
+  }
+
+  useEffect(() => {
+    if (cronJob) {
+      setIsCronSuspended(cronJob.spec.suspend);
+      setIsCheckingCronSuspendStatus(false);
+    }
+  }, [cronJob]);
+
+  console.log('cronJob suspend', isCronSuspended);
```
joaquimrocha, line 44: Left over from debug?

frontend/src/components/cronjob/Details.tsx
```diff
@@ -31,11 +60,74 @@ export default function CronJobDetails() {
...
+  function handleCron(suspend: boolean) {
+    const suspendingOrStarting = suspend ? 'Suspend' : 'Resum';
+    cronJob.spec.suspend = suspend;
+    // when we apply the cronjob, we need to remove the resourceVersion because a new resourceVersion is generated
+    // when the cronjob is applied
+    delete cronJob.metadata.resourceVersion;
+    setIsCheckingCronSuspendStatus(true);
+    dispatch(
+      clusterAction(() => applyFunc(cronJob.jsonData), {
+        startMessage: t(`resource|${suspendingOrStarting}ing CronJob {{ newItemName }}…`, {
```
joaquimrocha, line 72: The translation function works by matching the strings after the `|` separator with the catalog of translations we have.
The catalog is populated by scanning (parsing) the code and extracting the values you have added to the function.

This means the string in `t()` needs to be a plain string and not a literal. So computing strings within that function will not work with translations. Please have different strings for different cases if you need to branch out here.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -31,11 +60,74 @@ export default function CronJobDetails() {
...
+        description={t('frequent|Pause')}
+        onClick={() => {
+          handleCron(true);
+        }}
+        icon="mdi:pause-circle"
+        iconButtonProps={{
+          disabled: isCheckingCronSuspendStatus,
+        }}
+      />
+    );
```
joaquimrocha, line 121: I'd probably have just one button change its label/icons/function depending on what state we have.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -31,11 +60,74 @@ export default function CronJobDetails() {
 
   const ownedJobs = filterOwnedJobs(jobs);
 
+  function handleCron(suspend: boolean) {
+    const suspendingOrStarting = suspend ? 'Suspend' : 'Resum';
+    cronJob.spec.suspend = suspend;
+    // when we apply the cronjob, we need to remove the resourceVersion because a new resourceVersion is generated
+    // when the cronjob is applied
+    delete cronJob.metadata.resourceVersion;
```
joaquimrocha, line 68: I wonder if we should just remove the resource version in the `apply` function itself (so it works for all json).

joaquimrocha (changes requested):
Looks great! Thanks for the video! I left a few comments on the code.

frontend/src/components/cronjob/Details.tsx
ashu8912: yupp you are right we don't need this here

frontend/src/components/common/Resource/Resource.tsx
ashu8912: updated

frontend/src/components/cronjob/Details.tsx
ashu8912: fixed

frontend/src/components/cronjob/Details.tsx
ashu8912: fixed

frontend/src/components/cronjob/Details.tsx
ashu8912: fixed

frontend/src/components/cronjob/Details.tsx
ashu8912: updated

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -185,8 +185,15 @@ export interface DetailsGridProps
...
 export function DetailsGrid(props: DetailsGridProps) {
-  const { sectionsFunc, resourceType, name, namespace, children, ...otherMainInfoSectionProps } =
-    props;
+  const {
+    sectionsFunc,
+    resourceType,
+    name,
+    namespace,
+    children,
+    actions,
```
joaquimrocha, line 194: You don't need the `actions` here because they're already included in the `...otherMainInfoSectionProps`.

frontend/src/components/cronjob/Details.tsx
```diff
@@ -12,6 +19,22 @@ export default function CronJobDetails() {
   const { t, i18n } = useTranslation('glossary');
 
   const [jobs, jobsError] = Job.useList();
+  const [cronJob, cronJobError] = CronJob.useGet(name, namespace);
```
joaquimrocha, line 22: We should avoid doing this because it'll fetch the details twice from the server: once here, and another time as part of the DetailsGrid. I think that since you need to do all these changes, probably you should replace the DetailsGrid by MainInfoSection (which does less things automatically but it's still easy to use).

joaquimrocha (changes requested):
Spotted some other issues.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -800,14 +804,18 @@ export async function apply(body: KubeObjectInterface): Promise<JSON> {
...
   try {
+    delete bodyToApply.metadata.resourceVersion;
     return await apiEndpoint.post(bodyToApply);
   } catch (err) {
     // Check to see if failed because the record already exists.
     // If the failure isn't a 409 (i.e. Confilct), just rethrow.
     if ((err as ApiError).status !== 409) throw err;
 
+    // Preserve the resourceVersion if its an update request
+    bodyToApply.metadata.resourceVersion = resourceVersion;
```
joaquimrocha, line 818: We should avoid modifying the original structure. One way to avoid it, just like we do on line 797 is to clone the body (maybe we can clone it for all cases even...).

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -73,7 +73,7 @@ export default function EditorDialog(props: EditorDialogProps) {
   const [originalCode, setOriginalCode] = React.useState({ code: '', format: item ? 'yaml' : '' });
   const [code, setCode] = React.useState(originalCode);
   const [lastCodeCheckHandler, setLastCodeCheckHandler] = React.useState(0);
-  const [previousVersion, setPreviousVersion] = React.useState('');
+  const [previousVersion, setPreviousVersion] = React.useState<string>('');
```
joaquimrocha, line 76: Nit: This is unnecessary as it's inferred from the arg.

frontend/src/lib/k8s/apiProxy.ts
ashu8912: updated by cloning at source

frontend/src/components/cronjob/Details.tsx
ashu8912: updated with MainInfoSection

frontend/src/components/common/Resource/Resource.tsx
ashu8912: updated

frontend/src/components/cronjob/Details.tsx
joaquimrocha: I squashed a change to remove the back ticks (this causes a warning) + remove the capitalization of verbs (Supended -> suspended)

PR #939, opened 2023-02-06
frontend: Add a warning filter  switch to events table

ashu8912:
Users want to filter simply by clicking a switch to list all the warning type events, this patch adds that to the events table.

### DIFF
--- a/frontend/src/components/common/SectionHeader.tsx
+++ b/frontend/src/components/common/SectionHeader.tsx
@@ -1,3 +1,4 @@
+import { Box } from '@material-ui/core';
 import Grid from '@material-ui/core/Grid';
 import { makeStyles } from '@material-ui/core/styles';
 import { Variant } from '@material-ui/core/styles/createTypography';
@@ -28,6 +29,7 @@ export interface SectionHeaderProps {
   actions?: React.ReactNode[] | null;
   noPadding?: boolean;
   headerStyle?: HeaderStyle;
+  titleSideActions?: React.ReactNode[];
 }
 
 export default function SectionHeader(props: SectionHeaderProps) {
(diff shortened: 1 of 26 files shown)


ashu8912:
Screenshot 2023-02-06 at 11 38 38 AM
Screenshot 2023-02-06 at 11 38 48 AM

ashu8912:
https://user-images.githubusercontent.com/30126128/216931390-18f1cbe3-c20d-4ca4-8dd0-8a5190e71940.mp4

joaquimrocha:
The mock ups were actually not approved yet so this will need some changes as the current UX is not optimal.

joaquimrocha:
Here is the updated mockup. Also, it shouldn't hide the switch when the filter is opened (the filter should work together with the warning switch).
Mockup showing the warning switch aligned to the left

ashu8912:
@joaquimrocha updated this with the new designs

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -153,6 +154,7 @@ export default function SimpleTable(props: SimpleTableProps) {
     columns,
     data,
     filterFunction = null,
+    actionFilterFunction = null,
```
joaquimrocha, line 157: I'd rather keep a single filter function. This seems too intimate to this PR's use-case and it's not clear what exactly `actionFilterFunction` does or why it's needed.

frontend/src/components/common/SectionHeader.tsx
```diff
@@ -28,6 +29,7 @@ export interface SectionHeaderProps {
   actions?: React.ReactNode[] | null;
   noPadding?: boolean;
   headerStyle?: HeaderStyle;
+  actionsBesideTitle?: React.ReactNode[];
```
joaquimrocha, line 32: Let's call this `titleSideActions` please.

joaquimrocha (changes requested):
It's almost there. Left new comments.

frontend/src/components/common/SimpleTable.tsx
ashu8912: there is ```filterFunc``` which uses the search capabilities from store using hooks like ```useFilterFunc()```  but then we need another filterFunction which works beside the normal filtering, that's why this new prop

frontend/src/components/common/SectionHeader.tsx
ashu8912: updated

joaquimrocha (approved):

frontend/src/components/common/SimpleTable.tsx
joaquimrocha: I forgot to insist on getting a solution for this. I will try to fix it in another PR.

joaquimrocha: @ashu8912 I have added the fix in #979 . Please take a look.

PR #942, opened 2023-02-06
Add events table to each view

### LINKED ISSUE #911: Add events section to all resource details views
It may be useful to see any events related to a resource when checking its details view.
This should be easy to do by leveraging the common component we're using for most resource details views.

The events section should be shown at the bottom of the resource.

ashu8912:
This patch adds events list view to each detail view for a resource

fixes #911

### DIFF
--- a/frontend/src/components/configmap/Details.tsx
+++ b/frontend/src/components/configmap/Details.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import ConfigMap from '../../lib/k8s/configMap';
+import { ObjectEventList } from '../common';
 import Empty from '../common/EmptyContent';
 import Loader from '../common/Loader';
 import { DataField, MainInfoSection, PageGrid } from '../common/Resource';
@@ -34,6 +35,7 @@ export default function ConfigDetails() {
         )}
       </SectionBox>
       <DetailsViewSection resource={item} />
+      {item && <ObjectEventList object={item} />}
     </PageGrid>
   );
 }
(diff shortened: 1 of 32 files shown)


illume (approved):
Hrmm... I'm for it. Seems to be a good idea to me.

(You could perhaps merge the two commits)

frontend/src/components/configmap/Details.tsx
```diff
@@ -33,6 +34,7 @@ export default function ConfigDetails() {
           ))
         )}
       </SectionBox>
+      <EventsSection />
```
joaquimrocha, line 37: The events we want to show in the resource details view is the ones related to that resource alone (as #911 describes).
We already have a [component](https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/common/ObjectEventList.tsx) that does this for [HPA](https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/horizontalPodAutoscaler/Details.tsx) and a few other details view.

So we can either add it to every view, or maybe a better idea is to have it be added to the DetailsGrid after DetailsViewSection (if a prop `withEvents` is true, as by default it should be false so we don't break the use of DetailsGrid by plugins).

joaquimrocha (changes requested):
Thanks for starting this, @ashu8912 . Please check my comment as I think the intent of the #911 is not covered here.

frontend/src/components/configmap/Details.tsx
ashu8912: updated

joaquimrocha (changes requested):
Thanks for the changes @ashu8912 .
I have noticed that there's just a small UX issue whenloading the views where `withEvents` is used, because instead of seeing just the `loading` message, it shows that + the events table. See the screenshot below.
Screenshot showing the issue described above



The way to fix this is to add an extra check for the `!!item` before the `<ObjectEventList ...`.

joaquimrocha (changes requested):
The UX now is great! Thanks.
However, the tests are failing because it ends up in an error due to the events not being reachable (no cluster connected). I think this can be fixed by overriding the ObjectEventList (or maybe `Event.objectEvents`) in order to return phony events.

Maybe it's a good idea to override the mentioned function in `TestContext` (and make sure stories use it), so all tests are updated automatically instead of going story by story.

joaquimrocha (approved):

PR #951, opened 2023-02-10
frontend: Rearrange the Sidebar entries to distribute the load

### LINKED ISSUE #921: RFC: Rearrange the sidebar entries to improve the UX
As Headlamp's functionality is expanded and more resources are supported as first class, i.e. they get their own section in the UI by default, the sidebar should be reshuffled a bit in order to balance the distribution of sections and thus improve the UX.

Therefore, a solution could be to add a Config top-level section in the sidebar, which would contain:
* ConfigMaps
* Secrets
* HPA
* Resource Quotas
* Pod Disruption Budgets
* Leases
* Priority Classes

This would balance out the sidebar a bit better as it would make the Cluster section lighter and have this new way of focusing on config/limiting related sections.

Thoughts?

joaquimrocha:
This patch moves certain items from the Cluster sidebar entry (and
the Secrets from the Security entry) to a new Configuration entry.

Screenshot:
Headlamp sidebar showing a new Configuration entry

fixes #921


illume (approved):
Good idea. Works for me. 👍

PR #954, opened 2023-02-13
Replace kinvolk by headlamp-k8s in a lot of content

joaquimrocha:
We have recently moved the org from kinvolk to headlamp-k8s but we
left certain things still mentioning the kinvolk org because there
are redirects in place, but they don't work for certain things like
git clones, container registries, etc.


illume:
I'm looking into the CI failure (issue with the service URLs).

joaquimrocha:
@yolossn , can we also change the org for the Docker extension (maybe it's good to do it earlier than later)?

PR #956, opened 2023-02-13
frontend: Rebuild package-lock.json to include "resolved" field

joaquimrocha:
Somehow the "resolved" field in each package of package-lock.json has been deleted and this not only makes the builds possibly less consistent but also makes the process of generating flatpaks very complicated (because the build tools expect the "resolved" field to exist).


joaquimrocha:
Closing in favor of #966

PR #964, opened 2023-02-15
frontend: Fix formatting owner references for CRs

joaquimrocha:
how to test:
- [ ] Go to a CR that has an owner reference and check that there's no `$` showing up in the UI


ashu8912:
Before
Screenshot 2023-02-16 at 3 42 59 PM

After
Screenshot 2023-02-16 at 3 42 31 PM

PR #967, opened 2023-02-16
backend: Read token from the kube config file

joaquimrocha:
This may be needed for something users who have set up the token in the kubeconfig file.

How to test:
  1. Duplicate the context for a cluster you have running + copy its user auth info as well (all in kube config);
  2. Remove any client-certificate* data in the user auth info, and set the `token: ` field to be a valid token.
  3. Access that cluster in the UI -> Headlamp should not as for a token

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -873,6 +873,14 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
 			return
 		}
 
+		// Try to add the Authorization header if it's not already set from the client.
+		if r.Header.Get("Authorization") == "" {
+			token := c.contextProxies[clusterName].context.authInfo.Token
+			if token != "" {
+				r.Header.Add("Authorization", "Bearer "+token)
+			}
+		}
+
 		handler := proxyHandler(server, ctxtProxy.proxy)
 		handler(w, r)
 	})


joaquimrocha:
I got the confirmation from @gberche-orange that this works.

illume:
I think the test might need to be updated.

joaquimrocha:
> I think the test might need to be updated.
> 
> Because the test inserts the token, and it waits for the "Token" title ("/token" path). Now it's going to the "Login" title instead ("/login" path).

It's weird because for minikube (which doesn't set a token), this PR shouldn't have any difference. So definitely we need to test if this influences when running in-cluster.

backend/cmd/headlamp.go
```diff
@@ -873,6 +873,14 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
 			return
 		}
 
+		// Try to add the Authorization header if it's not already set from the client.
```
yolossn, line 876: I think instead of checking if token exists on every request this has to be solved when the context proxy is created. Like how the tls is setup to the proxy in `createProxyForContext` function.
https://github.com/headlamp-k8s/headlamp/blob/53fd76ffda97886749b1d132df4a3a73a6428c75/backend/cmd/headlamp.go#L903

yolossn: the [ReverseProxy](https://pkg.go.dev/net/http/httputil#ReverseProxy) has a `Director` function which can be used to modify the request before it is sent.

yolossn (commented):
we should consider handling the proxy part similar to how kubectl does by building on top of other kubernetes SDKs. If not we will have to play catchup or handle multiple cases like this 

https://github.com/kubernetes/kubectl/blob/feae475adee49103ba6033f33ce2a1c39024a527/pkg/proxy/proxy_server.go#L195

joaquimrocha:
> we should consider handling the proxy part similar to how kubectl does by building on top of other kubernetes SDKs. If not we will have to play catchup or handle multiple cases like this
> 
> https://github.com/kubernetes/kubectl/blob/feae475adee49103ba6033f33ce2a1c39024a527/pkg/proxy/proxy_server.go#L195

Yes. We should probably start using the transport part of k8s client-go if it allows us to replace the logic of loading and preparing every it ourselves.

joaquimrocha:
We have a user ( @atefhares ) who's built an image out of this branch and tried it with a kubeconfig that had a token set up but it still doesn't apparently pick it up, so we also need to test these changes with the image.

backend/cmd/headlamp.go
joaquimrocha: What I meant in today's discussion, for @illume 's better context, is that if the token in the kube config is expired, then I guess Headlamp asks the user for a token. If the user then adds a token (coming from the client side), we should take that as a higher relevance than the kube config one. That's why I always verify the request in my PR.

So we should double check that if the token in the kube config fails to get access, that the frontend takes the user to the login/token-request page.

illume:
@joaquimrocha @yolossn from the chat the other day, maybe you had some other comments for this issue? (I’ll work on this in a couple of hours.)

joaquimrocha:
Closing since the original issue is closed.

PR #972, opened 2023-02-17
ci: Fix invalid tag in container publish action

yolossn:


joaquimrocha (changes requested):
The commit has a patch about IG which doesn't seem related to this change.

yolossn:
> The commit has a patch about IG which doesn't seem related to this change.

Sorry, that was a mistake. Fixed it.

joaquimrocha (approved):

PR #975, opened 2023-02-21
backend: Add test for externalproxy endpoint

ashu8912:


yolossn (commented):
The code looks good, one suggestion will be to try and convert this to a table test. Also the CI is failing.

PR #979, opened 2023-02-22
Remove extra simple table filtering

joaquimrocha:
In #939 , I didn't notice I was merging an extra filter property into the SimpleTable. This is unnecessary and would create confusion for users.

This PR removes that and reimplements the filter by using the default function + the warning check.

How to test:
1. Verify the combination of search/filter for events and the warning switch work fine.


illume (approved):
Thanks👍 

the filter works for me

PR #987, opened 2023-02-28
WIP: Make plugin info (package.json) available to frontend

illume:
For https://github.com/headlamp-k8s/headlamp/issues/952

Makes the plugin information from the plugins package.json available to the frontend.


frontend/src/plugin/Plugins.tsx
```diff
@@ -23,21 +23,53 @@ export default function Plugins() {
...
+              }
+            }
+            return resp.json();
+          })
+        )
+      );
+
+      const sources = await sourcesPromise;
+      const packageInfos = await packageInfosPromise;
+      console.log('packageInfos', packageInfos);
```
joaquimrocha, line 71: Left over from debug?

frontend/src/plugin/Plugins.tsx
```diff
@@ -23,21 +23,53 @@ export default function Plugins() {
...
+          fetch(`${helpers.getAppUrl()}${path}/package.json`).then(resp => {
+            if (!resp.ok) {
+              if (resp.status !== 404) {
+                return Promise.reject(resp);
+              }
+              {
+                console.warn(
+                  'Missing package.json. ' +
+                    `Please upgrade the plugin at ${path}` +
+                    ' by running "headlamp-plugin extract" again'
```
joaquimrocha, line 54: Maybe refer which from which headlamp-plugin version should it be used?

backend/cmd/plugins.go
```diff
@@ -135,7 +135,7 @@ func pluginURLsForDir(pluginDir string, baseURL string) ([]string, error) {
 			continue
 		}
 
-		pluginFileURL := filepath.Join(baseURL, f.Name(), "main.js")
+		pluginFileURL := filepath.Join(baseURL, f.Name())
```
joaquimrocha, line 138: We should probably also output a warning in this function if the package.js doesn't exist (as we do for main.js).

backend/cmd/plugins.go
```diff
@@ -155,15 +155,15 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 		pluginListURLs = make([]string, 0)
 	}
 
-	r.HandleFunc("/plugins/list", func(w http.ResponseWriter, r *http.Request) {
+	r.HandleFunc("/plugins/paths", func(w http.ResponseWriter, r *http.Request) {
```
joaquimrocha, line 158: I think `paths` represent a logical concept of how we organize plugins that's not necessarily interesting for what we're trying to do.
The frontend needs a list of the plugins available. If we want to not use the old path (probably a good idea), then just `/plugins` probably works?

frontend/src/plugin/Plugins.tsx
```diff
@@ -28,13 +28,13 @@ export default function Plugins() {
      *   .initialize() plugins that register (not all do).
      */
     async function fetchAndExecute() {
-      const pluginsScriptPaths = (await fetch(`${helpers.getAppUrl()}plugins/list`).then(resp =>
+      const pluginsScriptPaths = (await fetch(`${helpers.getAppUrl()}plugins/paths`).then(resp =>
```
joaquimrocha: I wonder if some users have been deploying the frontend/backend in a separate way, which would mean that dropping the old route would break the frontend.
(I don't think this is a big issue, but we should not forget to mention it in the release notes)

frontend/src/plugin/Plugins.tsx
```diff
@@ -23,21 +23,53 @@ export default function Plugins() {
...
+        pluginPaths.map(path =>
+          fetch(`${helpers.getAppUrl()}${path}/package.json`).then(resp => {
+            if (!resp.ok) {
+              if (resp.status !== 404) {
+                return Promise.reject(resp);
+              }
+              {
+                console.warn(
+                  'Missing package.json. ' +
+                    `Please upgrade the plugin at ${path}` +
```
joaquimrocha, line 53: We can refer to the plugins names given by the plugin endpoint without treating it as a path. So `Please upgrade the plugin ${path}` (and renaming path to `pluginName` or something) would make more sense to me.

joaquimrocha (changes requested):
Awesome! Thanks for addressing this and unblocking the plugin management task.
Left a few comments.

plugins/headlamp-plugin/test-headlamp-plugin-published.js
```diff
@@ -27,6 +27,7 @@ function testHeadlampPluginPublished(pluginVersion) {
   // test extraction works
   run('npx @kinvolk/headlamp-plugin extract ./ .plugins');
   checkFileExists(path.join(curDir, '.plugins', 'headlamp-myfancy', 'main.js'));
+  // checkFileExists(path.join(curDir, '.plugins', 'headlamp-myfancy', 'package.json'));
```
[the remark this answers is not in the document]
illume: Yes, sorry... that will be uncommented once there's a new version published. This test uses the published version of headlamp-plugin.

backend/cmd/plugins.go
illume: It's changed to /plugins now.

frontend/src/plugin/Plugins.tsx
illume: This has been removed.

frontend/src/plugin/Plugins.tsx
illume: A message about the version was added.

frontend/src/plugin/Plugins.tsx
illume: This was changed.

frontend/src/plugin/Plugins.tsx
illume: I don't think it's worth mentioning. We've never mentioned other breakages caused by only updating parts of headlamp at once.

If they are using a non standard config, and only upgrade one part of the app... then yeah, it would break.

There's no known instances of anyone deploying things separately anyway. AFAIK.

backend/cmd/plugins.go
illume: I added a warning if there's no package.json.

We can turn the package.json into a hard requirement in a few versions IMHO. So plugins will keep working at the moment even though there's no package.json. So for now, it returns the plugin if there's no package.json.

illume:
@joaquimrocha thanks for the notes. I've made updates, and also left a comment for each of them. Feel free to unresolve any of the convos if there's something left to do.

illume:
@joaquimrocha I fixed that /paths issue, and retested. Thanks.

joaquimrocha (approved):

PR #991, opened 2023-03-02
Little improvements for Settings

illume:
A few little improvements for Settings...

- [x] Fix mobile layout issues (Also back button on other views like notifications)
- [x] Follow component conventions. Have a Settings.tsx, and one component per file
- [x] Add some working stories


frontend/src/lib/router.tsx
```diff
@@ -2,12 +2,12 @@ import { generatePath } from 'react-router';
 import NotFoundComponent from '../components/404';
 import AuthToken from '../components/account/Auth';
 import NotificationList from '../components/App/Notifications/List';
-import AppSettings, { SettingsButton } from '../components/App/settings';
-import { ClusterSettings, ClustersSettings } from '../components/App/settings/clusters';
 import AuthChooser from '../components/authchooser';
 import Chooser from '../components/cluster/Chooser';
 import KubeConfigLoader from '../components/cluster/KubeConfigLoader';
 import Overview from '../components/cluster/Overview';
+import Settings, { SettingsButton } from '../components/common/Settings';
```
joaquimrocha, line 10: My understanding is that if we add Settings to `common/`, then it'll be exported to plugins (so we have to maintain it in a backward-compatible way) and that's not really desirable or there's no real need to export it ATM, AFAIU.
WDYT?

joaquimrocha:
LGTM.

PR #994, opened 2023-03-06
Add accessible namespaces

### LINKED ISSUE #679: [RFE] Add accessible namespaces option
## Current situation

If a user doesn't have permission to list namespaces but only has access to specific namespaces, he basically can't use Headlamp.

## Ideal future situation

Headlamp could add an option for predefining "accessible namespaces" for a user via parameters (like env variables) or user/cluster settings (via user input before login).

image

## Additional information

Similar behavior has a `Lens` tool.
image

ashu8912:
How to test:
- [ ] Verify that the cluster settings button exists in the user menu (when in cluster)
- [ ] Verify that the cluster settings exists in the "failed to connect to cluster" login dialog
- [ ] Verify that the cluster settings exists in the "token requested" login dialog
- [ ] Go to the cluster settings in 2 different clusters and set different values for the default namespace and the allowed namespaces; then changes routes/refresh and go back to the cluster settings: verify that the values added are persistent
- [ ] Set the default namespace for a cluster and apply a simple resource in that cluster without specifying the namespace: verify it takes the namespace set as default
- [ ] Use a token that grants access only to a certain namespace; then set that namespace as the allowed one; then go to all list views: verify that only resources in the allowed namespace are shown
- [ ] Use a token that grants access only to a certain namespace; then set that namespace as the allowed one; then go to a list view and open the namespace filter: verify that only the allowed namespace is shown in the filter

fixes #679

### DIFF
--- a/frontend/src/components/App/TopBar.tsx
+++ b/frontend/src/components/App/TopBar.tsx
@@ -16,6 +16,7 @@ import { useHistory } from 'react-router-dom';
 import helpers from '../../helpers';
 import { getToken, setToken } from '../../lib/auth';
 import { useCluster, useClustersConf } from '../../lib/k8s';
+import { createRouteURL } from '../../lib/router';
 import {
   HeaderActionType,
   setVersionDialogOpen,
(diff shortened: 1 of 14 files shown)


joaquimrocha:
@ashu8912 has tested this. So let's merge after CI passes.

PR #995, opened 2023-03-07
Fix sidebar not reacting to filters

joaquimrocha:
We were not updating the sidebar when the plugin-added filters are updated after the sidebar is first rendered, so this made it impossible to register filters from within components' side effects.

How to test:
- [ ] Run the sidebar example plugin: it will render a sidebar entry "Click me and I will disappear"; click it and verify that the sidebar entry disappears


ashu8912 (approved):
Tested, LGTM!!!

ashu8912:
@joaquimrocha now that i think of it should we add the route filter logic to special sidebar options that are rendered??

joaquimrocha:
> @joaquimrocha now that i think of it should we add the route filter logic to special sidebar options that are rendered??

We need to do that but this filter logic is not really how I intended it to be. It was supposed to be the "list processor" behavior that now I am trying to apply to more places. So let's keep non-main sidebars in mind when we create a list processor plugin method for the sidebar.

PR #997, opened 2023-03-08
Allow plugins to process the list of header actions

joaquimrocha:
Instead of allowing to just add (append) header actions, we want plugins to be able to process the list of header actions in details views. The idea is that plugins supply a function that is given the current list of header actions and should return a list of header actions to display, allowing them to add/remove/update such actions.

An example is provided in the details-view example plugin.

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -26,6 +26,7 @@ import {
 import Pod, { KubePod } from '../../../lib/k8s/pod';
 import { createRouteURL, RouteURLProps } from '../../../lib/router';
 import { getThemeName } from '../../../lib/themes';
+import { DefaultHeaderAction, HeaderAction } from '../../../redux/actions/actions';
 import { useTypedSelector } from '../../../redux/reducers/reducers';
 import { useHasPreviousRoute } from '../../App/RouteSwitcher';
 import Loader from '../../common/Loader';
@@ -77,7 +78,11 @@ export interface MainInfoSectionProps {
     | ((resource: KubeObject | null) => NameValueTableRow[] | null)
     | NameValueTableRow[]
     | null;
-  actions?: ((resource: KubeObject | null) => React.ReactNode[] | null) | React.ReactNode[] | null;
+  actions?:
+    | ((resource: KubeObject | null) => React.ReactNode[] | null)
+    | React.ReactNode[]
+    | null
+    | HeaderAction[];
   headerStyle?: HeaderStyleProps['headerStyle'];
   noDefaultActions?: boolean;
   /** The route or location to go to. If it's an empty string, then the "browser back" function is used. If null, no back button will be shown. */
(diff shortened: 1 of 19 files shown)


lijianzhi01:
It looks good to me. Thumb up.

frontend/src/plugin/registry.tsx
```diff
@@ -280,6 +285,12 @@ export function registerDetailsViewHeaderAction(headerAction: DetailsViewHeaderA
   store.dispatch(setDetailsViewHeaderAction(headerAction));
 }
 
+export function registerDetailsViewHeaderActionsProcessor(
```
illume, line 288: This needs to be documented.

plugins/examples/details-view/src/index.tsx
```diff
@@ -26,3 +28,36 @@ registerDetailsViewSection(({ resource }: DetailsViewSectionProps) => {
   }
   return null;
 });
+
+registerDetailsViewHeaderActionsProcessor({
+  id: 'my-custom-header-actions-processor',
+  processor: (resource, actions) => {
```
illume, line 34: Some commentary on what this does and why would be useful. Passing in a named function(still inline), instead of an anonymous one could help with this.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -157,7 +193,7 @@ export function MainInfoSection(props: MainInfoSectionProps) {
         <SectionHeader
           title={title || (resource ? resource.kind : '')}
           headerStyle={headerStyle}
-          actions={allActions}
+          actions={allActions.filter(action => !!action)}
```
illume, line 196: Should this be done inside SectionHeader? I'm wondering, because this seems to be an API change one what SectionHeader.actions accepts. If it is an API change in this way, it could be fixed by having SectionHeader take care of this filtering.

frontend/src/redux/actions/actions.tsx
```diff
@@ -147,8 +167,29 @@ export function setRouteFilter(filterFunc: (entry: Route) => Route | null) {
...
+}
+
+export function addDetailsViewHeaderActionsProcessor(
+  actionProcessor: HeaderActionsProcessor | HeaderActionsProcessor['processor']
+) {
+  let headerActionsProcessor = actionProcessor as HeaderActionsProcessor;
+  if (headerActionsProcessor.id === undefined) {
+    headerActionsProcessor = {
+      id: '',
+      processor: actionProcessor as HeaderActionsProcessor['processor'],
```
illume, line 189: I wonder if `as` is hiding a bug.

illume (changes requested):
👍 

I left a few notes. Would you mind adding docs to some of those things? Then I'll have another look.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: I don't think this is an API change. Actually things should work without this change. I was trying to avoid passing [null, null, null] for example but I don't think this does that even, because only when rendering do the components turn to null.
Due to this we do get empty <div> in the resulting HTML but I don't think we can avoid it...

frontend/src/redux/actions/actions.tsx
joaquimrocha: I'll do a type check for function so this is not needed. Thanks.

joaquimrocha:
I have updated the PR and also fixed an issue @lijianzhi01 reported (that if a plugin returns an empty list of actions from its processor, the pod view still showed logs and exec).

@lijianzhi01 also pointed out maybe the logs and exec should be in the default actions. I think there's a point to it even if the actions aren't available in all details views as this is the case with scale and restart as well. @illume , WDYT?

@illume , I would also like you to double check whether we should simplify the types of header actions, as we have JSX.Element, ReactElement, ReactNode, etc. and maybe just ReactNode does the job?

plugins/examples/details-view/src/index.tsx
joaquimrocha: I fixed this by updating the ID so it describes better what it does.

illume (approved):
👍 thanks

PR #998, opened 2023-03-08
frontend: Always show overridden cluster chooser button

joaquimrocha:
We don't show the default cluster chooser button if only one cluster
is available, but if the button is overridden by a plugin, we
shouldn't make that change.


lijianzhi01:
It works for me. Thanks @joaquimrocha

PR #999, opened 2023-03-09
Disable GPU on ARM Linux

joaquimrocha:
How to test:
- Run the app in a Linux+ARM device: verify that the text is shown


joaquimrocha:
@vbatts has a verified this works on an actual ARM Linux machine. Thanks!

@ashu8912 , can you quickly check the code?

Issue #983, opened 2023-02-23
Add init container list to pod details view


joaquimrocha:
Somehow we're not showing the [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) list in the pod details view.

We should show the init containers under the regular ones in that view. If there are no init containers set up, they should however, not show up.

joaquimrocha:
Already in 0.16.0

PR #1007, opened 2023-03-13
frontend init-containers: Includes init containers to pod containers

vyncent-t:
For issue #983 

Creates an additional field for Init containers under the regular containers when in the pod view.

image

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -739,6 +739,10 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
     return containers;
   }
 
+  function getInitContainers() {
+    return resource?.spec?.initContainers || [];
+  }
+
   function getStatuses() {
     if (!resource || resource.kind !== 'Pod') {
       return {};
@@ -757,6 +761,7 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
   }
 
   const containers = getContainers();
+  const initContainers = getInitContainers();
   const statuses = getStatuses();
   const numContainers = containers.length;
 
@@ -780,6 +785,21 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null 
           ))
         )}
       </>
+
+      {initContainers.length > 0 && (
+        <>
+          <SectionBox title={t('resource|Init Containers')} />
+          {initContainers.map((initContainer: KubeContainer, i: number) => (
+            <SectionBox key={`init_container_${i}`} outterBoxProps={{ pt: 1 }}>
+              <ContainerInfo
+                resource={resource}
+                container={initContainer}
+                status={statuses[initContainer.name]}
+              />
+            </SectionBox>
+          ))}
+        </>
+      )}
     </>
   );
 }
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -739,6 +739,20 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
...
+      return [];
+    }
+
+    let initContainers: KubeContainer[] = [];
+
+    if (resource.spec.initContainers) {
+      initContainers = resource.spec.initContainers;
+    }
+
+    return initContainers;
```
joaquimrocha, line 753: I think this can be simplified with `return resource?.spec?.initContainers || [];`

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -780,6 +795,25 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
...
         )}
       </>
+
+      {!initContainers ? (
+        <SectionBox>
+          <Empty>{t('resource|No data to be shown.')}</Empty>
+        </SectionBox>
+      ) : (
+        <>
+          <SectionBox title={t(`resource| Init Containers`)} />
```
joaquimrocha, line 805: The namespace separator (`|`) shouldn't have spaces around it.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -780,6 +795,25 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
...
+          {initContainers.map((initContainer: any, i: number) => (
+            <SectionBox key={i} outterBoxProps={{ pt: 1 }}>
```
joaquimrocha, line 807: I know you just used what's written for the regular containers, but the `key` should be textual, so `key={`init_container_${i}`}` is better; and `initContainer` should have the proper type (`KubeContainer` I think).

joaquimrocha: Feel free to also modify the regular container's key+type in a different commit if you want.

joaquimrocha (changes requested):
Left some comments in the code.
Also, the commit title shouldn't have `init containers` in the prefix, just frontend is enough.

If you want to reuse the exact same code, you could also pass an "isInit" property to the `ContainersSection`, and then add a second `<ContainerSection isInit={true}...`.

vyncent-t:
[x] updated the build snapshots
[x] commit title edit
[x] simplified with return resource?.spec?.initContainers || []
[x] namespace separator (|) shouldn't have spaces around it
[x] contextual edits

- left the original context container as is, will change with another commit as mentioned

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -780,6 +785,25 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
           ))
         )}
       </>
+
+      {initContainers.length === 0 ? (
+        <SectionBox>
+          <Empty>{t('resource|No data to be shown.')}</Empty>
+        </SectionBox>
```
joaquimrocha, line 792: We should not display anything if there are no initContainers.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -780,6 +785,25 @@ export function ContainersSection(props: { resource: KubeObjectInterface | null
...
+      ) : (
+        <>
+          <SectionBox title={t(`resource|Init Containers`)} />
```
joaquimrocha, line 795: Please use single quotes here.

joaquimrocha (changes requested):

joaquimrocha (approved):

PR #1009, opened 2023-03-14
Add ability to remove clusters

yolossn:


frontend/src/components/cluster/Chooser.tsx
```diff
@@ -446,7 +453,7 @@ function Chooser(props: ClusterDialogProps) {
         ) : (
           <ClusterList clusters={clusterList} onButtonClick={handleButtonClick} />
         )}
-        {helpers.isElectron() ? (
+        {helpers.isElectron() || helpers.isDevMode() ? (
```
joaquimrocha, line 456: Ditto.

frontend/src/components/cluster/ManageClusters.tsx
```diff
@@ -0,0 +1,128 @@
+import { DialogTitle } from '@material-ui/core';
```
joaquimrocha, line 1: Since there are some more complex UX decisions we need to make here (like: if we don't allow removing from the default kubeconfig, then users will be puzzled when they don't see certain clusters in the list for removal) and the UI needs to be refined, I have just realized that now that we have a cluster settings view, we could just add a "Remove Cluster" button to the bottom right of that view, which is a much less complex UX and allows us to unblock the "cluster addition" feature for this version.

joaquimrocha (changes requested):

frontend/src/components/cluster/ManageClusters.tsx
yolossn: Do you mean move the `Manage Cluster (pen)` Icon next to the settings?

joaquimrocha:
All looks good now. I have updated the last commit with just styling + confirmation button:
cluster settings showing the delete cluster option

PR #1020, opened 2023-03-15
Update i18n and some deps

joaquimrocha:
This PR is the same of #1010 without the last commit (since it Storybook breaks CI + make i18n yields empty strings in the English translation) and with the i18n updated to the best I could.


illume (approved):
👍 thanks

- I tried make run-frontend / run- backend locally against minikube, and it works in my limited testing.
- Didn't review the i18n text changes, but things still seem to work as far as I could see changing languages.
- I tried app-linux and "dist/Headlamp-0.15.0-linux-arm64.AppImage" and it seems to work with limited testing in WSL2 ubuntu 22.04.1 LTS. However, note the version in the produced filenames is wrong. (I guess this gets bumped in the release action... I'm not sure).
- There's a bug with the headlamp-server not being killed with "dist/Headlamp-0.15.0-linux-arm64.AppImage" generated with this branch. But also with the previous 0.14.0 release, so it's not a regression that should block this release IMHO. It's also not related to this PR. For some reason killing it requires me to use sudo... weird.

joaquimrocha:
> * There's a bug with the headlamp-server not being killed with "dist/Headlamp-0.15.0-linux-arm64.AppImage" generated with this branch. But also with the previous 0.14.0 release, so it's not a regression that should block this release IMHO. It's also not related to this PR. For some reason killing it requires me to use sudo... weird.

I noticed this yesterday too. Somehow, if I enable the Headlamp addon in minikube, I get this headlamp-server process.
Santhosh couldn't reproduce it so I assumed this has to do with running minikube with the podman driver (are you using this driver?).
Also, it explains Headlamp not being able to kill it because it's a process started as root (whereas Headlamp runs as the normal user).

PR #1036, opened 2023-03-19
backend: Use client-go to generate tls

yolossn:
fixes:https://github.com/headlamp-k8s/headlamp/issues/961


joaquimrocha:
5 lines in, 57 out in one patch. Love it.
Also tested and it works the same as the current backend AFAICS.

joaquimrocha (approved):

PR #1039, opened 2023-03-20
frontend: Cache icons for airgapped environment

lijianzhi01:
This is proposal for caching icons from iconify. Below requests will not be sent any more.
image

image

IconBundles.ts contains more icons than Headlamp used. So, Headlamp can remove those unnecessary icons.


joaquimrocha:
@illume , I think we only need the mdi ones. Let's take over this PR (I've already talked with Julian).

PR #1040, opened 2023-03-20
frontend: Cache font from google api for airgapped environment

lijianzhi01:
Cache font from google api for airgapped environment. This is not a complete, because it needs backend set response header with text/css.


lijianzhi01:
I just code font style in index.html. It can save one request from browser.

joaquimrocha:
@lijianzhi01 Does this really work in airgap? I see it still referencing the fonts from Google's servers in your changes.
I'd expect us to ship the fonts if we want them in an airgapped env.

lijianzhi01:
> @lijianzhi01 Does this really work in airgap? I see it still referencing the fonts from Google's servers in your changes. I'd expect us to ship the fonts if we want them in an airgapped env.

You are right. I am not fully cache those fonts. I will cache fonts instead.

lijianzhi01:
Hi @joaquimrocha, I have cached each kinds of fonts in public folder. 
image

illume (approved):
Looks fine to me. A good improvement, thanks!

I'll tweak a few things in another branch and merge.

PR #1049, opened 2023-03-22
Fix docker ext release

yolossn:
The docker extension release action is also responsible for updating the base image used by the docker extension, due to a bug that was introduced when manual trigger was added to this github action the automatic updating part didn't work. This patch fixes it


yolossn:
Tested this by manually running the github action
https://github.com/headlamp-k8s/headlamp/actions/runs/4486610733/jobs/7889256019

joaquimrocha (approved):

PR #1050, opened 2023-03-22
frontend: Improve HPA, PDB, Resource Quota

lijianzhi01:


frontend/src/lib/k8s/resourceQuota.ts
```diff
@@ -42,35 +42,113 @@ class ResourceQuota extends makeKubeObject<KubeResourceQuota>('resourceQuota') {
...
+      Object.keys(this.spec.hard).forEach(key => {
+        if (key.startsWith('requests.')) {
+          req.push(
+            `${key}: ${this.normalizeUnit(key, this.status.used[key])}/${this.normalizeUnit(
+              key,
+              this.spec.hard[key]
+            )}`
+          );
+        }
+      });
```
yolossn, line 55: looks like this should also accept `cpu` and `memory` since it is same as `requests.cpu` and `requests.memory`.
[Refer](https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota)

frontend/src/lib/k8s/resourceQuota.ts
```diff
@@ -42,35 +42,113 @@ class ResourceQuota extends makeKubeObject<KubeResourceQuota>('resourceQuota') {
...
+          name: key,
+          hard: `${this.normalizeUnit(key, this.status.hard[key])} (${this.status.hard[key]})`,
+          used: `${this.normalizeUnit(key, this.status.used[key])} (${this.status.used[key]})`,
+        });
       });
-    });
     return stats;
   }
+
+  normalizeUnit(key: string, quantity: string) {
```
yolossn, line 88: I don't see kubectl normalizing the units, I am not sure if we should go ahead and add this. @joaquimrocha WDYT?

yolossn (changes requested):

frontend/src/lib/k8s/resourceQuota.ts
joaquimrocha: If kubectl is not doing, my inclination is to also not do it here. Maybe we can add a setting in the future to turn this on, but I would like to keep the current behavior if it matches kubectl.

lijianzhi01: I think we should keep the same with kubectl. Maybe add a setting in the future.

lijianzhi01: I got cases from our partners. The main purpose they use GUI to check resource quota is they want to know the usage of CPU and Memory. 
Kubectl doesn't normalize this unit indeed, but customers don't care the raw output, they need a straightforward way to know what how much they have used. 
Here is raw output from Headlamp, they have to do the math by converting the unit one by one.  
image
Alos the same as resource quota list, the unnormalized data only bring inconvenient to users.
image

joaquimrocha:
Closing as #1117 ismerged.

PR #1052, opened 2023-03-23
Improve sidebar

joaquimrocha:
This PR improves the sidebar, more particularly, it stops special casing the "home" sidebar and now supports having several sidebars, creating new ones, and controlling them from plugins.
It also fixes smaller issues like the special casing of the sidebar matching routes by name or cloning default routes by stringifying/unstringifying them...

This was part of the multi-cluster improvements but it's important enough that we can treat it separately.

Unfortunately, due to not easily being able to import types from plugins, I couldn't use the `DefaultSidebars.HOME` constant to refer to the home sidebar. Maybe @illume has a solution for this.

How to test:
- [ ] Verify there are no regressions to the default sidebars' behavior
- [ ] Verify there are no regressions to how the sidebar plugins behave (except for the new functionality)

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -64,7 +64,15 @@ function RouteComponent({ route }: { route: RouteType }) {
   }, [route.hideAppBar]);
 
   return (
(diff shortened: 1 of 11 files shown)


illume:
Rebased on main fixing conflicts.

illume (approved):
👍 thanks

PR #1053, opened 2023-03-24
docs: Fix KUBECONFIG section on multiple config files

illume:
Headlamp now supports reading from multiple kubeconfig files.

### DIFF
--- a/docs/installation/desktop/_index.md
+++ b/docs/installation/desktop/_index.md
@@ -27,12 +27,17 @@ KUBECONFIG=/my/different/kubeconfig /path/to/headlamp
 
 ### Use several kube config files
 
-If you need to use more than one kube config file, please merge them into
-one and use the resulting one as indicated above.
+If you need to use more than one kube config file at the same time, you can list
+each config file path with a separator.
 
-You can easily merge kubee config files the following way:
+On unix:
 ```bash
-KUBECONFIG=kubeconfig1:kubeconfig2:kubeconfig3 kubectl config view --raw > mynewconfig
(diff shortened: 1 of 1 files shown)


docs/installation/desktop/_index.md
```diff
@@ -27,12 +27,17 @@ KUBECONFIG=/my/different/kubeconfig /path/to/headlamp
...
 
-If you need to use more than one kube config file, please merge them into
-one and use the resulting one as indicated above.
+If you need to use more than one kube config file at the same time, you can list
+each config file path with a separator.
 
-You can easily merge kubee config files the following way:
+On unix:
 ```bash
-KUBECONFIG=kubeconfig1:kubeconfig2:kubeconfig3 kubectl config view --raw > mynewconfig
```
joaquimrocha, line 35: Maybe we should still keep this as an alternative because users maybe don't really want to launch Headlamp from the CLI?

illume: I don't think so. This is a workaround solution that needs more effort to use from the command line than just specifying the config files with KUBECONFIG.

Should users be able to select different config files from the app? Probably. But this PR is just about fixing the wording to show that the kubectl standard method of supporting multiple kube config files through KUBECONFIG, and not having to use this work around anymore.

yolossn (approved):

docs/installation/desktop/_index.md
yolossn: Also this is docs is for the desktop application, so the user can import kubeconfigs.

PR #1055, opened 2023-03-24
Add cluster from sidebar

### LINKED ISSUE #1054: Add cluster button missing when only 1 cluster is available
We allow the user to load clusters/contexts from kubeconfig files when running the desktop version. However, this option is not available in the UI when only one cluster is configured because it is only available in the cluster chooser (and this dialog is only shown when more than one cluster is configured).

Acceptance Criteria:
- [ ] We have a button available in the UI for loading a kube config (a good place is likely the home sidebar)
- [ ] The button is only visible in the desktop version

joaquimrocha:
fixes #1054 

A screenshot of the UI after this PR

### DIFF
--- a/frontend/src/lib/router.tsx
+++ b/frontend/src/lib/router.tsx
@@ -608,7 +608,7 @@ const defaultRoutes: {
     component: () => <PortForwardingList />,
   },
   loadKubeConfig: {
-    path: '/loadKubeConfig',
+    path: '/load-kube-config',
     exact: true,
     name: 'Load KubeConfig',
     sidebar: null,


yolossn:
Everything looks good but there is a small issue with the flow. After clicking on `Add cluster` if the user click back in the kubeconfig loader view it jumps back to home screen.

joaquimrocha:
> Everything looks good but there is a small issue with the flow. After clicking on `Add cluster` if the user click back in the kubeconfig loader view it jumps back to home screen.

You are right. However, the way to do this through react-router (using state) is not working somehow. So I'd rather leave this detail for later.

PR #1056, opened 2023-03-24
Installer docs

joaquimrocha:
Windows was the one desperately needing an update, but the others also needed some care.


vyncent-t:
It looks pretty good to me! All the links work and the wording is direct and instructional enough to easily follow.

PR #1058, opened 2023-03-27
frontend: Add more details in Pod list

lijianzhi01:
kubectl get pod -o wide, you will get nodeName, podIP, nodeIP, which are important for troubleshooting. Imagine 
1. You have hundreds of Pod, you only want to know the one with specific podIP.
2. One of nodes does not work, you want to sort hundreds of pods with nodeName.

image


frontend/src/components/pod/List.tsx
```diff
@@ -77,6 +77,42 @@ export function PodListRenderer(props: PodListProps) {
...
+          if (p1.spec.nodeName < p2.spec.nodeName) {
+            return -1;
+          } else if (p1.spec.nodeName > p2.spec.nodeName) {
+            return 1;
+          }
+          return 0;
+        },
+      },
+      {
+        label: t('frequent|Node IP'),
```
joaquimrocha, line 105: This is not showing in the regular `kubectl get pods -o wide`, so I think we should enable this from a plugin instead of by default.

illume:
Arg. Sorry, I messed up your PR somehow trying to push a fix for the test failure. But it added a whole bunch of other commits for some reason. I'll fix it.

PR #1065, opened 2023-03-30
WIP: frontend: Use redux-toolkit old redux is unmaintained

illume:
Start using redux toolkit incrementally. This means we can use our old redux code mostly as it was before. Then we can slowly change it to use the new style over time.


illume:
This was merged into the https://github.com/headlamp-k8s/headlamp/pull/990 PR to be continued there.

PR #1066, opened 2023-03-31
Add network policy view

ashu8912:
Acceptance Criteria

- [x] Go to networks tab on sidebar, there should be a new option now under networks tab called ```Network Policies```
- [x] There is network policy list view when we click on `Network Policies` tab under network tab in sidebar
- [x] In the network policy list view clicking the name of network policy should take us to the network policy detail view

### DIFF
--- a/frontend/src/components/networkpolicy/Details.tsx
+++ b/frontend/src/components/networkpolicy/Details.tsx
@@ -1,0 +1,187 @@
+import { Box, Typography } from '@material-ui/core';
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router';
+import { matchExpressionSimplifier, matchLabelsSimplifier } from '../../lib/k8s';
+import { LabelSelector } from '../../lib/k8s/cluster';
+import NetworkPolicy, {
+  NetworkPolicyEgressRule,
+  NetworkPolicyIngressRule,
+  NetworkPolicyPort,
+} from '../../lib/k8s/networkpolicy';
+import { DetailsGrid, NameValueTable, SectionBox, useMetadataDisplayStyles } from '../common';
+
+export function NetworkPolicyDetails() {
+  const { t } = useTranslation('glossary');
+  const { name, namespace } = useParams<{ name: string; namespace: string }>();
+  const classes = useMetadataDisplayStyles();
+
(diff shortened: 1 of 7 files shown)


frontend/src/lib/k8s/index.ts
joaquimrocha, line 1: I'd like to understand *why* we need to simplify the _matchLabel_. I feel like the logic is very similar to the one we already have, so maybe there's a way to reuse/share some of the code, otherwise we have to maintain two very similar versions of things.

frontend/src/components/networkpolicy/Details.tsx
```diff
@@ -0,0 +1,190 @@
...
+                      return <></>;
+                    }
+                    const { cidr, except = [] } = to.ipBlock || {};
+                    if (!cidr) {
+                      return <></>;
+                    }
+                    if (cidr && except.length === 0) {
+                      return <>{`cidr: ${cidr}`}</>;
+                    }
+                    return <>{`cidr: ${cidr}, except: ${except.join(', ')}`}</>;
```
joaquimrocha, line 155: Please make this translatable (for the `except`)

joaquimrocha (dismissed):
Thanks for the changes Ashu. Left a couple of comments but it's looking already very close.

frontend/src/lib/k8s/index.ts
ashu8912: updated now we use one method which is reused

frontend/src/components/networkpolicy/Details.tsx
```diff
@@ -0,0 +1,190 @@
...
+                  name: t('To'),
+                  value: '',
+                },
+                {
+                  name: t('ipBlock'),
+                  value: item.to.map(to => {
+                    if (!to.ipBlock) {
+                      return <></>;
+                    }
+                    const { cidr, except = [] } = to.ipBlock || {};
```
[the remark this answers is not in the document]
ashu8912: updated

frontend/src/components/networkpolicy/Details.tsx
ashu8912: updated

illume (approved):
👍 thanks

PR #1077, opened 2023-04-12
ci: add multi-platform (arm64) Docker images

### LINKED ISSUE #369: Build & publish ARM container images
## Current situation

We do publish ARM versions of the Headlamp app, but we don't have an ARM version for the container image.

## Impact

Users who want to deploy Headlamp in an ARM cluster, cannot do it.

## Ideal future situation

We have the Headlamp ARM images published to the registry as well.

milas:
Refactor the Docker image to support cross-compiling and update the CI pipeline to publish arm64 images in addition to the amd64 images already being published.

There's also some general improvements here, e.g. the Go builds copy `go.{mod,sum}` and run `go mod download`, cache volumes, use of `COPY --link`, removal of unnecessary base images.

This removes the symlink for `/headlamp/server`, which has been around for almost a year for compatibility. (Keeping it will add a dependency on QEMU to run the `cd` / `ln` commands, as it's the only command run in the final image, which is for the _target_ arch, not the _build_ arch.)

CI also uses GitHub Actions cache now - free speed-up!

The builds are also done with `pull: true` to ensure that the latest Go / Node / Alpine patch versions get used.

Additionally, the images publish provenance metadata.

Fixes #369.

### DIFF
--- a/.github/workflows/container-publish.yml
+++ b/.github/workflows/container-publish.yml
@@ -74,5 +74,10 @@ jobs:
       with:
         context: .
         push: true
+        pull: true
+        platforms: linux/amd64,linux/arm64
         tags: ${{ env.CONTAINER_TAGS }}
         labels: ${{ steps.meta.outputs.labels }}
+        provenance: true
+        cache-from: type=gha
+        cache-to: type=gha,mode=max
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,36 @@
-ARG IMAGE_BASE=alpine:3.17.2
-FROM $IMAGE_BASE as base-build
+# syntax=docker/dockerfile:1
+FROM --platform=${BUILDPLATFORM} golang:1.19 as backend-build
 
-RUN apk update && \
-	apk add git nodejs npm go ca-certificates make musl-dev bash icu-data
+WORKDIR /headlamp
 
-FROM golang:1.19 as backend-build
-
+ARG TARGETOS
+ARG TARGETARCH
 ENV GOPATH=/go \
     GOPROXY=https://proxy.golang.org \
 	GO111MODULE=on\
 	CGO_ENABLED=0\ 
-	GOOS=linux 
+	GOOS=${TARGETOS}\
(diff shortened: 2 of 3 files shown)


illume:
Thanks. Some very nice improvements.


I had to learn about some of the options and techniques you used in here. For the benefit of other folks who may read this, here's the documentation I found on them.

### COPY --link

https://docs.docker.com/engine/reference/builder/#copy---link

### Run cache docs. 

https://docs.docker.com/build/cache/#use-the-dedicated-run-cache

### Options used in the "docker/build-push-action@v4" action 

https://github.com/docker/build-push-action#inputs

provenance | Bool/String | Generate [provenance](https://docs.docker.com/build/attestations/slsa-provenance/) attestation for the build (shorthand for --attest=type=provenance)
-- | -- | --
cache-from | List | List of [external cache sources](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from) (e.g., type=local,src=path/to/dir) (e.g., type=local,src=path/to/dir)
cache-to | List | List of [cache export destinations](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to) (e.g., type=local,dest=path/to/dir)

### buildx

https://docs.docker.com/engine/reference/commandline/buildx/

### build --pull

https://docs.docker.com/engine/reference/commandline/build/

`--pull Always attempt to pull a newer version of the image`

milas:
Ahhh I meant to preemptively apologize that I went a bit overboard 😅 

(Disclaimer: I work at Docker Inc but am doing this in my free time...very late at night)

If there's any parts you want me to split out into separate PR(s), I'm happy to so

I should also mention that I've actually been running Headlamp on arm64 [building the Docker image manually from my computer] without any issues for months on a low-powered home cluster and haven't seen any issues, it's great

illume (approved):
👍 thanks

illume:
> I should also mention that I've actually been running Headlamp on arm64 [building the Docker image manually from my computer] without any issues for months on a low-powered home cluster and haven't seen any issues, it's great

That's great to hear :)

PR #1079, opened 2023-04-13
Helm plugin

ashu8912:
To run this PR

`cd plugins/examples/helm` run `npm start ` to start the plugin

Building the backend
`make backend`
`cd backend`
`./headlamp-server -proxy-urls "https://artifacthub.io/*" -dev`

We need the backend to run with the artifacthub proxy `https://artifacthub.io/*` because the helm charts are directly fetched from the hub.

Edge cases that are still not fixed

- [x] In the release detail view the upgrade and rollback option are not added 

Some remaining items to do:
- [x] backend lint errors
- [x] the Helm endpoint should be restricted to desktop only (not in-cluster)
- [x] Tests for helm related backend code
- [x] backend helm code TODOs
- [x] commit messages for helm backend code need adding (some are empty with just "wip")
- [x] probably some of the lint fixes could be squashed. They are kept separate for now to make debugging easier in case something was broken whilst lint-fixing.


illume:
Would you mind fixing the ci errors?

How about adding https://artifacthub.io/* to the default server for now? Or add it by the headlamp-server call in app mode.

illume:
@ashu8912 I fixed the remaining lint issues in backend, and merged fixes from the pre-release branch from the other day.

The CI is failing on the lint with it timing out for some reason.
>  Error: golangci-lint exit with code 4
>  Ran golangci-lint in 180963ms

I'm not sure why it's doing that. Hrmm.

illume:
Ok, CI passes now :) There was a weird golang thing where `repo := repo` needed to be done inside a for loop which was messing things up.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+	defaultNewConfigFileMode   os.FileMode = os.FileMode(0644)
+	defaultNewConfigFolderMode os.FileMode = os.FileMode(0770)
+)
+
+// add repository
+
+type AddUpdateRepoRequest struct {
+	Name string `json:"name"`
+	URL  string `json:"url"`
+	// TODO: Figure out how to support auth
```
illume, line 24: There's a TODO here.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// Adds a repository with name, url to the helm config. Returns error if there is one.
+func AddRepositoryToHelm(name string, url string) error {
+	err := createFileIfNotThere(settings.RepositoryConfig)
+	if err != nil {
+		zlog.Error().Err(err).Str("action", "add_repo").Msg("failed to create empty RepositoryConfig file")
+	}
+
+	// TODO: Lock repo file
```
illume, line 48: There's a TODO here.

plugins/examples/helm/src/api/releases.tsx
```diff
@@ -0,0 +1,57 @@
...
+  });
+}
+
+export function createRelease(name: string, namespace: string, value: string, chart: string) {
+  return request(`/helm/release/install?namespace=${namespace}`, {
+    method: 'POST',
+    body: JSON.stringify({
+      name: name,
+      namespace: namespace,
+      //@todo fix this
```
illume, line 46: There's a TODO here.

plugins/examples/helm/src/components/charts/EditorDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+                  label="Namespaces"
+                  placeholder="Select Namespace"
+                />
+              )}
+            />
+          )}
+        </Box>
+      </Box>
+      <Box pt={2} height="100%" my={2}>
+        {/* @todo: The editor onChange doesn't fire on an undo redo event */}
```
illume, line 181: There's a TODO here.

plugins/examples/helm/src/components/charts/List.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <TextField
+        style={{
+          width: '20vw',
+          margin: '0 1rem',
+        }}
+        id="outlined-basic"
+        label="Search"
+        variant="outlined"
+        value={search}
+        // @todo: Find a better way to handle search autofocus
```
illume, line 98: There's a TODO here.

backend/pkg/helm/handler.go
```diff
@@ -0,0 +1,161 @@
...
+	defer a.Unlock()
+
+	for key, value := range a.store {
+		if time.Since(value.updatedAt) > 5*time.Minute {
+			delete(a.store, key)
+		}
+	}
+}
+
+// TODO FIXME: can this be refactored without init?
```
illume, line 80: There's a TODO here.

plugins/examples/helm/src/api/releases.tsx
```diff
@@ -0,0 +1,57 @@
...
+      value: value,
+      chart: chart,
+    }),
+  });
+}
+
+export function getActionStatus(name: string, action: string) {
+  return request(`/helm/action/status?name=${name}&action=${action}`, {
```
illume, line 54: URLs need to be constructed using safe methods without string interpolation.

backend/pkg/helm/charts.go
```diff
@@ -0,0 +1,74 @@
+package helm
+
+// import (
```
illume, line 3: There's a TODO here, with code commented out.

backend/pkg/helm/release.go
```diff
@@ -0,0 +1,744 @@
...
+}
+
+// Returns (releases, error) given the request and helm configuration.
+//
+//nolint:gocognit
+func getReleases(req ListReleaseRequest, config *action.Configuration) ([]*release.Release, error) {
+	// Get list client
+	listClient := action.NewList(config)
+
+	// Removing all these if assignments is not possible, so we disable gocognit linter
```
illume, line 51: I'm not sure if all of these assignments can be done in a cleaner way? 

Is there a single line way to do this? I know there's no ternary operators... but something like this? (untested)
`listClient.AllNamespaces = req.AllNamespaces != nil && *req.AllNamespaces || nil`

illume (changes requested):
Left some notes for mostly backend related things.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+	defaultNewConfigFolderMode os.FileMode = os.FileMode(0770)
+)
+
+// add repository
+
+type AddUpdateRepoRequest struct {
+	Name string `json:"name"`
+	URL  string `json:"url"`
+	// TODO: Figure out how to support auth
+	// like username, password, certfile etc
```
illume, line 25: I looked into this a bit, and it isn't about auth for our helm routes. In this PR the helm routes are currently completely unprotected. Which is obviously a big security issue, since it means any network connection can now run helm commands.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	return nil
+}
+
+// Adds a repository with name, url to the helm config. Returns error if there is one.
+func AddRepositoryToHelm(name string, url string) error {
+	err := createFileIfNotThere(settings.RepositoryConfig)
+	if err != nil {
+		zlog.Error().Err(err).Str("action", "add_repo").Msg("failed to create empty RepositoryConfig file")
```
illume, line 45: There should be a `return err` here.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, err
+	}
+
+	return os.Create(p)
+}
+
+func (h *Handler) ListRepo(w http.ResponseWriter, r *http.Request) {
+	err := createFileIfNotThere(settings.RepositoryConfig)
+	if err != nil {
+		zlog.Error().Err(err).Str("action", "list_repo").Msg("failed to create empty RepositoryConfig file")
```
illume, line 144: There should be a `return` here.

backend/go.mod
```diff
@@ -5,30 +5,31 @@ go 1.12
...
 	github.com/go-openapi/swag v0.22.3 // indirect
+	github.com/go-playground/validator/v10 v10.12.0
 	github.com/gobwas/glob v0.2.3
 	github.com/google/gnostic v0.6.9 // indirect
-	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/uuid v1.3.0
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
-	github.com/imdario/mergo v0.3.13 // indirect
+	github.com/gorilla/schema v1.2.0
```
illume, line 16: We're adding a dependency which is unmaintained here. I haven't looked into why this has been added or what alternatives there may or may not be.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,261 @@
...
+
+		return
+	}
+}
+
+// Remove repository name.
+func (h *Handler) RemoveRepo(w http.ResponseWriter, r *http.Request) {
+	err := createFileIfNotThere(settings.RepositoryConfig)
+	if err != nil {
+		zlog.Error().Err(err).Str("action", "remove_repo").Msg("failed to create empty RepositoryConfig file")
```
illume, line 188: There should be a `return` here.

backend/cmd/headlamp.go
```diff
@@ -510,16 +513,45 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+			return
+		}
+		// We may want to filter some headers, otherwise we could just use a shallow copy
+		// proxyReq.Header = req.Header
+		proxyReq.Header = make(http.Header)
+		for h, val := range r.Header {
+			proxyReq.Header[h] = val
+		}
+
+		// Disable caching
```
illume, line 531: I'm wondering if disabling caching should be up to the client? If the client wants caching stuff to be cached, then this wouldn't allow that.

backend/cmd/headlamp.go
```diff
@@ -510,16 +513,45 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+		w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
+		w.Header().Set("Expires", time.Unix(0, 0).Format(http.TimeFormat))
+		w.Header().Set("Pragma", "no-cache")
+		w.Header().Set("X-Accel-Expires", "0")
+
+		client := http.Client{}
+		resp, err := client.Do(proxyReq)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadGateway)
```
illume, line 540: Should we also log these errors?

backend/cmd/headlamp.go
```diff
@@ -960,7 +992,92 @@ func (c *HeadlampConfig) startPortForward(p PortForwardPayload, token string) er
...
+//nolint:gocognit
+func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
+	router.PathPrefix("/clusters/{clusterName}/helm/{.*}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		helmHandler, err := getHelmHandler(c, w, r)
+		if err != nil {
+			return
+		}
+
+		// we used nolint:gocognit in this function because...
+		//  Perhaps there's a better way to dispatch these?
```
illume, line 1027: Is there a better way to dispatch these?

backend/pkg/helm/release.go
```diff
@@ -0,0 +1,763 @@
...
+		actionState.SetStatus("uninstall", req.Name, "failed", err)
+
+		return
+	}
+
+	actionState.SetStatus("uninstall", req.Name, "success", err)
+}
+
+type RollbackReleaseRequest struct {
+	Name      string `json:"name" validate:"required"`
```
illume, line 335: I wonder if these fields need to be validated further by us, or if the helm code does more validation?

backend/go.mod
joaquimrocha: Let's remove this when we get rid of gorilla.

backend/cmd/headlamp.go
joaquimrocha: This is fine to improve later IMO.

backend/cmd/headlamp.go
joaquimrocha: It's a good idea.

backend/cmd/headlamp.go
joaquimrocha: This is fine to improve later IMO.

backend/pkg/helm/charts.go
joaquimrocha: is this whole file needed? Seems just like a bunch of commented out code...
Where is the TODO?

backend/pkg/helm/handler.go
joaquimrocha: This is fine to improve later IMO.

backend/pkg/helm/repository.go
joaquimrocha: This is fine to improve later IMO.

backend/pkg/helm/repository.go
joaquimrocha: This is restricted to desktop. So it's like the dynamic clusters. AFAIU, when trying to apply the charts, it'd fail anyway if no valid credentials to the API server.

backend/pkg/helm/repository.go
joaquimrocha: Okay. Please add it.

backend/pkg/helm/repository.go
joaquimrocha: This is fine to improve later IMO.

backend/pkg/helm/repository.go
joaquimrocha: Please add it.

backend/pkg/helm/repository.go
joaquimrocha: Please add it.

backend/pkg/helm/charts.go
illume: The comment is 'wip'. I don't know why it's here.

backend/pkg/helm/repository.go
illume: It's not clear to me if this could cause corruption of helm config if more than one helm code caller runs at the same time (either multiple headlamp actions or helm binary whilst headlamp is doing actions). The github linked shows locking, and it seems to me this is required by API users. Needs further investigation.

illume:
I think the plugin bundling commits could be made in a separate PR? They seem to be self contained enough.

backend/cmd/headlamp.go
illume: I'm not so sure. Previously we didn't disable caching, so this could perhaps break something that relied on it.

backend/pkg/helm/repository.go
```diff
@@ -0,0 +1,265 @@
...
+// Adds a repository with name, url to the helm config. Returns error if there is one.
+func AddRepositoryToHelm(name string, url string) error {
+	err := createFileIfNotThere(settings.RepositoryConfig)
+	if err != nil {
+		zlog.Error().Err(err).Str("action", "add_repo").Msg("failed to create empty RepositoryConfig file")
+		return err
+	}
+
+	// TODO: Lock repo file
+
```
illume, line 50: As well as file locking, missing from here is checking if the repo exists already. As is done here: 
 https://github.com/helm/helm/blob/39ca699ca790e02ba36753dec6ba4177cc68d417/cmd/helm/repo_add.go#L102

backend/pkg/helm/repository.go
yolossn: This auth part is not related to headlamp, it is used when adding a private repository which has its own custom authentication system.

backend/pkg/helm/repository.go
yolossn: closing duplicate comment in favor of https://github.com/headlamp-k8s/headlamp/pull/1079#discussion_r1173458174

backend/cmd/headlamp.go
yolossn: Due to the current structure of backend I couldn't find better ways to call the helm handler functions. Looking into ways to improve this and the URL structure for helm endpoints.

backend/cmd/headlamp.go
yolossn: @ashu8912  Can you expand on the issues that you faced with the external proxy that led to disabling cache?

backend/pkg/helm/release.go
yolossn: We can create a separate function to initialize the listClient from the request params as a work around for the gocognit linter.

backend/pkg/helm/charts.go
yolossn: We were unsure if the charts list endpoint will be required or not but turns out this is required for the release upgrade.

backend/pkg/helm/repository.go
yolossn: Since the file lock is added there won't be corruption of helm repo file due to concurrency. Are there any other issues that are on your mind that needs to be investigated?

backend/cmd/headlamp.go
```diff
@@ -1186,9 +1333,11 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 
 func (c *HeadlampConfig) addClusterSetupRoute(r *mux.Router) {
 	// We do not support this feature when in-cluster
-	if c.useInCluster {
+	if !c.enableDynamicClusters {
+		fmt.Println("Dynamic clusters endpoints: disabled.")
 		return
 	}
+	fmt.Println("Dynamic clusters endpoints: enabled.")
```
yolossn, line 1340: Is there any specific reason for adding the enableDynamicClusters flag? The existing logic already makes sure that the dynamic clusters doesn't work for incluster mode.

backend/pkg/helm/repository.go
illume: To obtain feature parity, we’d need something like this. This is most likely used by internal repos. I asked a mate at one bigco and they have auth on all their internal package repos. This is best practice, so I expect many other places to require this for private repos too.

IMHO we could delay this todo for another time.

backend/pkg/helm/repository.go
illume: I noticed the helm binary doesn’t lock on anything but the add repo call. Which is a bug in helm. Probably we should lock it everywhere it modifies. Probably locking is more important with a web app because it’s easier to fire off multiple requests (double clicks on buttons, impatient users etc).

backend/pkg/helm/handler.go
illume: Yes, let’s leave this one for later.

backend/pkg/helm/repository.go
yolossn: AFAIU Apart from the add repo and update repo most of the functions are read only. Ill check once if there are any other places where we need to add locks.

backend/pkg/helm/repository.go
joaquimrocha: Since this is fixed now, resolving the dicussion.

backend/cmd/headlamp.go
joaquimrocha: This was answered in today's call.

backend/cmd/headlamp.go
ashu8912: So while proxying to get the chart detail  the /externalproxy endpoint was cached, So even if i was on a different detail view i would get data from the last cached /externalproxy

illume: From Ashu's description, we should NOT be caching in this endpoint. Insteatd setting those cache headers in the client. Because clients are responsible for setting client side caching headers.

It seems this code has been moved out of this PR, so I'll resolve this convo.

backend/pkg/helm/release.go
illume: This code cleanup could happen in future work too IMHO. Resolving this convo.

backend/pkg/helm/repository.go
illume: It looks like @yolossn added locks to the other other methods. 🎉  So resolving this convo.

illume:
> Added tests to Helm functions and marked them with integration build tag. Use the following command to run the test

Nice work on adding tests! 🎉🎈

illume:
> Some README are not rendering everything correctly. See Headlamp readme, which has tables that aren't rendered nicely.

For this TODO in the PR description can be taken out of here. Since the frontend part of this plugin code is now not in this PR. (We still need to render things correctly of course... just not in this PR)

illume:
Auth on these endpoints will be done on a separate PR.

backend/cmd/headlamp.go
joaquimrocha: Answered in yesterday's call. Resolving the conversation.

PR #1080, opened 2023-04-14
Pre release 17

illume:
`npx @kinvolk/headlamp-plugin build plugins/examples/helm/`
`npx @kinvolk/headlamp-plugin extract plugins/examples/helm/ .plugins`


joaquimrocha:
Maybe this should be closed now?

PR #1090, opened 2023-04-19
frontend: Clone the clusters conf to prevent direct state changes

joaquimrocha:
The auth chooser was getting an error when checking for a cluster's
auth because of how the cluster's state is set. It sets the state
directly and then calls dispatch, but since the object being modified
is the same that redux holds, redux detected this and it threw an
exception.

To prevent this case, the useClustersConf hook now returns a copy of
the state instead.

### DIFF
--- a/frontend/src/lib/k8s/index.ts
+++ b/frontend/src/lib/k8s/index.ts
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import React from 'react';
 import { useLocation } from 'react-router-dom';
 import { ConfigState } from '../../redux/reducers/config';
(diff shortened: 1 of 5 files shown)


illume:
CI is passing now. Regarding this failure: https://github.com/headlamp-k8s/headlamp/actions/runs/4753252945/jobs/8444609724?pr=1090#step:8:51
Seems a type was escaping that the `cloneDeep()` kept, so I added it back to how it was before so the type was the same as before.
```ts
  const newFilters = { ..._.cloneDeep(state) };
```

illume (approved):
👍 thanks

PR #1091, opened 2023-04-19
frontend: Fix plugins settings sidebar entry

joaquimrocha:
It was not migrated to the new way sidebars are defined, so it was
no longer visible. This patch removes the legacy code and sets up
the sidebar properly.


illume (approved):
👍 thanks

Issue #1088, opened 2023-04-18
registerAppBarAction issue


illume:
There's an issue with this...

![issue-action](https://user-images.githubusercontent.com/9541/232829826-4b411f8d-8c57-42a5-bb74-17bb37d7a857.png)

Which is a problem with our use of redux, where often we are mutating state. Unfortunately because everything is all tangled together simply deep cloning all the state each time is a bit of a performance hit. Ultimately the fix would be to move this into a redux slice. But that's a more time consuming refactor. So, maybe doing a deep clone of state is enough for now.

joaquimrocha:
Yeah, seems like the redux update uncovered all these issues. I also opened #1090 to handle another case.
I wonder if reverting that redux change for the release makes sense.

illume:
@lijianzhi01 There's a fix in main for this now. Thanks @joaquimrocha and @vyncent-t

PR #1092, opened 2023-04-19
WIP: frontend registerAppBarAction: fixes state mutation error

vyncent-t:
Fixes for issue #1088 

Currently:

image

Similar to the screenshot in the original issue, I am able to replicate the error in the browser console by parsing through the error messages at the root dir of my local dev env. The original screenshot delivers the error for plugins\cop, but I am using the headlamp-pod-counter plugin (the only plugin I ran) and the same error is delivered.

I implemented the suggested change Rene had mentioned when overviewing the issue, changing the newFilters const variable to be a clone. 

By changing  `const newFilters = { ...state }` to `const newFilters = _.cloneDeep(state)` within the reducer function in headlamp/frontend/src/redux/reducers/ui.tsx, and reloading the home dir in the dev env, I no longer see the error message.

I will continue to look for alternative solutions if this is not a whole fix, including implementation of a redux slice.

### DIFF
--- a/frontend/src/redux/reducers/ui.tsx
+++ b/frontend/src/redux/reducers/ui.tsx
@@ -144,7 +144,8 @@ export const INITIAL_STATE: UIState = {
 };
 
 function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
-  const newFilters = { ...state };
+  // const newFilters = { ...state };
+  const newFilters = _.cloneDeep(state);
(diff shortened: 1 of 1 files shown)


illume:
Thanks @vyncent-t It's good to know this fixes the issue.

@joaquimrocha made a fix like this in https://github.com/headlamp-k8s/headlamp/pull/1090 for a different error, and then added this change in as well after.

PR #1093, opened 2023-04-19
Allow plugins to override tables

### LINKED ISSUE #1044: Allow to modify resource table list
As a plugin developer, we got cases:
* Customers want us to add view pod log, exec, edit and delete in the to table
image
* Operators wants us to add DNS creatation status from monitoring data to Service List table to make troubleshooting more easier
* Operators wants to add pod numbers into Node List table for debug
* And etc
So, we want to Headlamp can provide plugin an ability to add, delete, replace columns in Resource List Table

joaquimrocha:
This PR adds the changes and an example plugin on modifying existing tables throughout Headlamp.
The way this is done is by having ResourceTables identified with a new ID (or automatically done when they receive the resourceClass argument) and then plugins can declare their processors and use this ID within the processors' function to decide whether to modify the columns that the table receives.

fixes #1044


lijianzhi01:
Hey Joaquim, this PR works well in all our cases. Thanks.

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Menu, MenuItem, Typography } from '@material-ui/core';
+import React from 'react';
+import { useHistory } from 'react-router-dom';
+
+function ContextMenu(props: { pod: Pod }) {
+  const { pod } = props;
+  const [anchorEl, setAnchorEl] = React.useState(null);
+  const history = useHistory();
+
+  function handleClick(event: any) {
```
illume, line 13: I don't think it's clear what handleClick is doing. "Our action button" could be verb-ish instead I guess to help understanding?

plugins/examples/tables/README.md
```diff
@@ -0,0 +1,14 @@
+# Tables
+
+This example plugin overrides the pod list view's table to add a context menu.
```
illume, line 3: There's a place in docs/plugins/development/functionality.md to add more detail. Including a screenshot in docs/plugins/development/images/

illume (changes requested):
Very good. I think the API is nice.

I left some notes for your consideration.

Would you mind rebasing on main? (there's quite a few changes merged since this was done)

frontend/src/components/pod/List.tsx
```diff
@@ -122,6 +122,7 @@ export function PodListRenderer(props: PodListProps) {
         columns={getDataCols()}
         data={pods}
         reflectInURL={reflectTableInURL}
+        id="pods"
```
illume, line 125: What do you think about adding a prefix for these ids? 

'pods' seems like it might perhaps clash in the future.
('table-pods', 'headlamp-table-pods')?

Or maybe it doesn't matter... I'm not sure.

joaquimrocha: It's a good question. I think having a `headlamp-` prefix can be a good thing to prevent any lack of namespacing we may need down the road. Let me change that.

plugins/examples/tables/README.md
joaquimrocha: Adding that.

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <MenuItem
+          onClick={() => {
+            history.push(pod.getDetailsLink());
+          }}
+        >
+          <Typography>Details</Typography>
+        </MenuItem>
+        <MenuItem
+          onClick={() => {
+            alert('Boom! Pod deleted! (just kidding)');
```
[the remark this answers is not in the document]
joaquimrocha: I am a comedy genius.
But it's just a message...

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+          }}
+        >
+          <Typography>Delete</Typography>
+        </MenuItem>
+      </Menu>
+    </>
+  );
+}
+
+registerResourceTableColumnsProcessor(function ageRemover({ id, columns }) {
```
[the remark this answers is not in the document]
joaquimrocha: Yeah. That was the first idea, but a context menu is much more interesting. Good catch.

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  console.debug('Resource table to process:', id);
```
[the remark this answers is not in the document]
joaquimrocha: Not accidental, but I will remove it.

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (id === 'pods') {
```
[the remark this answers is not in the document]
joaquimrocha: This is a good question. I had the console.debug because users could quickly get a print of all the tables processed by the example plugin for the view they're at.
I think it's useful to map IDs -> tables-in-views, but let's do that in another PR later if that's the best approach.
I think anyone developing a plugin using this processor will just check which names do those tables have dynamically anyway.

frontend/src/redux/actions/actions.tsx
```diff
@@ -199,6 +216,19 @@ export function addDetailsViewHeaderActionsProcessor(
   return { type: UI_DETAILS_VIEW_ADD_HEADER_ACTIONS_PROCESSOR, action: headerActionsProcessor };
 }
 
+export function addResourceTableColumnsProcessor(
+  tableProcessor: TableColumnsProcessor | TableColumnsProcessor['processor']
+) {
+  let tableColsProcessor = tableProcessor as TableColumnsProcessor;
+  if (tableColsProcessor.id === undefined && typeof tableProcessor === 'function') {
+    tableColsProcessor = {
+      id: '',
```
illume, line 225: Why is this an empty string, and in the other place it's generated?

plugins/examples/tables/src/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      </Menu>
+    </>
+  );
+}
+
+// Add a context menu to the pods table.
+// You can see it by going to the pods list view: there will be a 3-dot menu for every row.
+registerResourceTableColumnsProcessor(function setupContextMenuForPodsList({ id, columns }) {
+  // Not checking for the table ID will add the context menu to all resource tables.
+  if (id === 'headlamp-pods') {
```
illume, line 54: How do users of the function know which table ids are available, or how to find them?

In another processor there's an object with the default ids they can import. Maybe tell them in the docs for id where to look, and the convention id naming follows.

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -12,8 +13,12 @@ type SimpleTableColumn = SimpleTableProps['columns'][number];
 type ColumnType = 'age' | 'name' | 'namespace' | 'type' | 'kind';
 
 export interface ResourceTableProps extends Omit<SimpleTableProps, 'columns'> {
-  /** The columns to be rendered, like used in SimpleTable, or by name */
+  /** The columns to be rendered, like used in SimpleTable, or by name. */
   columns: (SimpleTableColumn | ColumnType)[];
+  /** ID for the table. Will be used by plugins to identify this table. */
+  id?: string;
```
illume, line 19: Can you please add the naming convention here?

frontend/src/redux/reducers/ui.tsx
```diff
@@ -222,6 +233,16 @@ function reducer(state = _.cloneDeep(INITIAL_STATE), action: Action) {
       newFilters.views.details.headerActionsProcessors = processors;
       break;
     }
+    case UI_ADD_TABLE_COLUMNS_PROCESSOR: {
+      const processor = action.action;
+      const generatedID = processor.id || `generated-id-${Date.now().toString(36)}`;
```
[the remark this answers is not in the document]
illume: I left a separate comment in actions.tsx where it's empty in this PR.

illume (commented):
Thanks for those updates.

I resolved a bunch of convos, and left a few more notes for your consideration. 

It's maybe worth a rebase against main.

frontend/src/redux/actions/actions.tsx
joaquimrocha: This is just because I want to do the generation in the reducer, not here. So we have it in one place. Same for the header actions processor.

plugins/examples/tables/src/index.tsx
joaquimrocha: The ids do get automatically generated for known kinds of pods.
Keeping a list of the ids will be hard to maintain IMO, and it will not of course include any IDs for tables that plugins add, so developers will likely list all ids before selecting the ones they want to use.

But I agree that better docs are needed. Let's just add that in a different PR.

frontend/src/components/common/Resource/ResourceTable.tsx
joaquimrocha: Will do.

plugins/examples/tables/src/index.tsx
illume: I'd prefer if the docs were updated here.

How about this?

> How do you find the table id? They follow a convention. `headlamp-${pluralName}`. For example, Pods would be `headlamp-pods` and for Namespaces it would be `headlamp-namespaces`. Plugins can add their own table ids, so you can either look at what the plugin uses for the id.

illume (commented):
The part after the headlamp- prefix needs to be described. Added some suggested docs in one convo.

illume (approved):
🎉 thanks!

All the convos are resolved. Just waiting for tests.

lijianzhi01:
Love this PR. It means lot to customization.

PR #1097, opened 2023-04-21
backend: Support a static plugins dir

joaquimrocha:
This is so we can have bundled plugins without affecting the user's
own plugin dir. For it is set by an env var.

### DIFF
--- a/backend/cmd/plugins.go
+++ b/backend/cmd/plugins.go
@@ -103,9 +103,25 @@ func watchSubfolders(path string) {
 // Returns pluginListURLs, nil if there is no problem.
 // returns nil, err if there's an error.
 func (c *HeadlampConfig) getPluginListBasePaths() ([]string, error) {
+	var pluginListURLStatic []string
+
+	if c.staticPluginDir != "" {
+		var err error
+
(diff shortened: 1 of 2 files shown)


backend/cmd/plugins.go
```diff
@@ -184,4 +200,10 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 	}
 
 	r.PathPrefix("/plugins/").Handler(pluginHandler)
+
+	if config.staticPluginDir != "" {
+		staticPluginsHandler := http.StripPrefix(config.baseURL+"/static-plugins/",
```
illume, line 205: Probably we need similar no-cache code to above for this staticPluginsHandler too?

This part I mean:
```go
	if !config.useInCluster {
		pluginHandler = serveWithNoCacheHeader(pluginHandler)
	}
```

joaquimrocha: I left that out on purpose because these are supposed to be static. So we ship it once and don't care if those contents have changed. If it turns out it's needed later, we can add it, but now I don't see a need.

illume (approved):
👍 thanks

PR #1105, opened 2023-04-26
backend: Fix externalproxy endpoint, also make it not cache requests

ashu8912:


joaquimrocha (changes requested):
Please just explain how it was failing (in the commit body) so it's clear why this was done.

joaquimrocha (approved):

PR #1106, opened 2023-04-26
frontend: Add node version and services selector into list

lijianzhi01:
To keep the same output with kubectl, I made following changes:
* Add selector for services list
image

* Add node version to node list
image


lijianzhi01:
I have set kubelet version as node version in table list. I left label parts because it will require UI design just as you said.

PR #1108, opened 2023-04-26
frontend: Prevent multiple auth test requests at login

### LINKED ISSUE #1100: Too many requests when clusters cannot be reached
We we select a cluster that's down, I see too many requests fired. This is likely a side effect from the quick response in the auth frontend. So maybe it needs some sort of exponential backoff or so.

**Update:** Some more info after looking into authchooser/index.tsx : it needs to be rewritten because not only is this the cause of the issue I wrote above, it's also making multiple requests even when things are working.

joaquimrocha:
Due to how the test-auth logic in the AuthChooser was designed, it
led to many requests to the API server, especially if the response
was not valid.

This is due to the clusters dependency in the related useEffect hook
being constantly changed. Even though this is worth an investigation
on its own, making sure these requests only happen when needed is
needed, so this patch ensures only one test auth request happens
unless the clusters are actually changed.

fixes #1100 

How to test:
- [ ] Open the dev tools; Go to the cluster chooser and choose an unreachable cluster (a shutdown cluster for example): make sure only one request to the selfsubjectrules endpoint is triggered from the login page
- [ ] Do the same for an existing cluster and make sure the logic advances into that working cluster (loads the cluster home view)

### DIFF
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -1,5 +1,6 @@
 import { InlineIcon } from '@iconify/react';
 import { Box, Button, withStyles } from '@material-ui/core';
+import _ from 'lodash';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
@@ -48,6 +49,8 @@ function AuthChooser({ children }: AuthChooserProps) {
     {}) as ReactRouterLocationStateIface;
   const clusterName = getCluster() as string;
   const { t } = useTranslation('auth');
+  const clustersRef = React.useRef<typeof clusters>(null);
+  const cancelledRef = React.useRef(false);
 
   let clusterAuthType = '';
   if (clusters && clusters[clusterName]) {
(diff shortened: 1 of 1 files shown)


yolossn:
Tested and it works. A small doubt, Can't we make this logic simple by sending a verifyAuthHandler to PureAuthChooser and run it whenever the selected cluster is changed, instead of dealing with cluster reference etc.

frontend/src/components/authchooser/index.tsx
```diff
@@ -80,15 +91,16 @@ function AuthChooser({ children }: AuthChooserProps) {
         setTestingAuth(true);
 
         let errorObj: Error | null = null;
-        setError(errorObj);
+
+        console.debug('Testing auth at authchooser');
```
illume, line 95: console.debug left in on purpose?

joaquimrocha: yes. If we see millions of prints, we'll know :)

joaquimrocha:
> Tested and it works. A small doubt, Can't we make this logic simple by sending a verifyAuthHandler to PureAuthChooser and run it whenever the selected cluster is changed, instead of dealing with cluster reference etc.

At this point, I think we should merge this PR as it prevents that problem. And we can simplify the code later.
Thanks.

PR #1113, opened 2023-04-28
frontend backend app: Prevent unauthorized access to helm endpoints

illume:
This prevents some backend endpoints from being called unless there is a token passed in via the X-HEADLAMP_BACKEND-TOKEN header. The app generates the token and then sets HEADLAMP_BACKEND_TOKEN environment variable which the backend uses. This also makes available a getHeadlampAPIHeaders() function which can be used by the frontend and plugins to add headers to API endpoints that the backend serves. App creates token and sends it to the backend and frontend. This prevents unauthorized connections to the helm chart endpoints. The threat was that any connection from a Browser or localhost process could call helm and dynamic cluster creation endpoints.

Here's the steps of how the token gets generated, passed around and checked:

- [x] app/ generates a token (`backendToken`)
- [x] app starts backend with `backendToken` (via HEADLAMP_BACKEND_TOKEN env var)
- [x] backend opens frontend URL with ?backendToken=XXX
- [x] frontend takes `?backendToken=XXX` in app mode
  - [x] `getHeadlampAPIHeaders()` API function made available to frontend
  - [x] plugins have `getHeadlampAPIHeaders()` API function made available to frontend
- [x] backend checks API calls have backendtoken `X-HEADLAMP_BACKEND-TOKEN header == HEADLAMP_BACKEND_TOKEN env var`
  - [x] helm chart api calls
  - [x] dynamic cluster endpoint creation
- [x] use `getHeadlampAPIHeaders` for setCluster and deleteCluster in the frontend/.

# how to use

Here's how you use the headers for `fetch()` in a plugin:
```ts
import { getHeadlampAPIHeaders } from '@kinvolk/headlamp-plugin/lib';
fetch(url, {headers: getHeadlampAPIHeaders()})
```

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -11,6 +11,7 @@ import { OpPatch } from 'json-patch';
 import _ from 'lodash';
 import { decodeToken } from 'react-jwt';
 import helpers from '../../helpers';
+import { getHeadlampAPIHeaders } from '../../helpers';
 import store from '../../redux/stores/store';
 import { getToken, logout, setToken } from '../auth';
 import { getCluster } from '../util';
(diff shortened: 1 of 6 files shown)


joaquimrocha (changes requested):
Thanks. The backend code makes sense but I don't see where the requests are being modified to include the backend-token (in the frontend).
I also tested by loading a cluster from a kubeconfig and got a Forbidden code as result.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -872,14 +873,23 @@ export async function testAuth() {
 export async function setCluster(clusterReq: ClusterRequest) {
   return request(
     '/cluster',
-    { method: 'POST', body: JSON.stringify(clusterReq), headers: JSON_HEADERS },
+    {
+      method: 'POST',
+      body: JSON.stringify(clusterReq),
+      headers: { ...JSON_HEADERS, ...getHeadlampAPIHeaders() },
```
joaquimrocha, line 879: Looks good! I wonder if it'd be easier to have a request variant that already puts these headers in `headlampApiRequest`.
Not super convinced this is much more useful, so take it as a light suggestion and feel free to decide.

joaquimrocha (approved):

PR #1117, opened 2023-05-02
frontend: Improve HPA, PDB, Resource Quota

yolossn:
Cleaned https://github.com/headlamp-k8s/headlamp/pull/1050 and updated snapshots

Changes:

This PR contains following changes:
1. Fix more lable missing in HPA target
image

2. Remove duplicated HPA detail event table and add contdition table
image

3. Improve unit of request and limit in resource quota and make them human readable
image

4. Improve UX fro RQ and HPA
image

5. Fix null reference error in resource quota status

### DIFF
--- a/frontend/src/components/horizontalPodAutoscaler/List.tsx
+++ b/frontend/src/components/horizontalPodAutoscaler/List.tsx
@@ -1,9 +1,24 @@
+import { Chip, createStyles, makeStyles, Theme } from '@material-ui/core';
 import { useTranslation } from 'react-i18next';
 import HPA from '../../lib/k8s/hpa';
-import { Link, SectionBox, SectionFilterHeader, StatusLabel } from '../common';
+import { Link, SectionBox, SectionFilterHeader } from '../common';
 import ResourceTable from '../common/Resource/ResourceTable';
 
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      display: 'flex',
+      justifyContent: 'left',
+      flexWrap: 'wrap',
+      '& > *': {
+        margin: theme.spacing(0.5),
+      },
+    },
+  })
+);
+
 export default function HpaList() {
+  const classes = useStyles();
   const { t } = useTranslation(['glossary', 'hpa', 'frequent']);
 
   return (
@@ -27,25 +42,14 @@ export default function HpaList() {
               const value: JSX.Element[] = [];
               const metrics = hpa.metrics(t);
               if (metrics.length) {
-                value.push(
-                  <StatusLabel
-                    key={metrics[0].name}
-                    id={metrics[0].name}
-                    status=""
-                    title={metrics[0].name}
-                  >
-                    {metrics[0].shortValue}
-                  </StatusLabel>
-                );
+                value.push(<Chip label={metrics[0].shortValue} variant="outlined" />);
                 if (metrics.length > 1) {
                   value.push(
-                    <StatusLabel key="more..." status="">
-                      {metrics.length - 1} {t('frequent|more…')}
-                    </StatusLabel>
+                    <Chip label={metrics.length - 1 + t('frequent|more…')} variant="outlined" />
                   );
                 }
               }
-              return <>{value}</>;
+              return <div className={classes.root}>{value}</div>;
             },
           },
           {
--- a/frontend/src/components/horizontalPodAutoscaler/List.tsx
+++ b/frontend/src/components/horizontalPodAutoscaler/List.tsx
@@ -37,7 +37,7 @@ export default function HpaList() {
                     {metrics[0].shortValue}
                   </StatusLabel>
                 );
-                if (hpa.metrics.length > 1) {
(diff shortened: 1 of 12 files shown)


lijianzhi01:
Hey @yolossn , I have tested this PR. It looks good to me. Thanks for your improvement.

frontend/src/lib/k8s/resourceQuota.ts
```diff
@@ -42,35 +42,114 @@ class ResourceQuota extends makeKubeObject<KubeResourceQuota>('resourceQuota') {
...
-      stats.push({
-        name: key,
-        hard: this.status.hard[key],
-        used: this.status.used[key],
+    this.status.hard &&
+      Object.keys(this.status.hard).forEach(key => {
+        stats.push({
+          name: key,
+          hard: `${this.normalizeUnit(key, this.status.hard[key])} (${this.status.hard[key]})`,
+          used: `${this.normalizeUnit(key, this.status.used[key])} (${this.status.used[key]})`,
```
joaquimrocha, line 82: If the result within parenthesis is the same as outside, then we should not include the parenthesis, otherwise it's `0 (0)` and it's a bit confusing.
See this example when using this PR (also shows `5 (5)`):
Screenshot with the example

frontend/src/lib/k8s/resourceQuota.ts
```diff
@@ -42,35 +42,114 @@ class ResourceQuota extends makeKubeObject<KubeResourceQuota>('resourceQuota') {
...
+
+        if (bytes === 0) {
+          normalizedQuantity = '0 Bytes';
+        } else {
+          const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+          const k = 1000;
+          const dm = 2;
+          const i = Math.floor(Math.log(bytes) / Math.log(k));
+          normalizedQuantity = parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
+        }
```
joaquimrocha, line 144: Should this logic be abstracted in the utils/helpers modules?

joaquimrocha (changes requested):

frontend/src/components/horizontalPodAutoscaler/List.tsx
```diff
@@ -27,25 +42,14 @@ export default function HpaList() {
...
                   value.push(
-                    <StatusLabel key="more..." status="">
-                      {metrics.length - 1} {t('frequent|more…')}
-                    </StatusLabel>
+                    <Chip label={metrics.length - 1 + t('frequent|more…')} variant="outlined" />
                   );
                 }
               }
-              return <>{value}</>;
+              return <div className={classes.root}>{value}</div>;
```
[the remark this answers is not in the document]
yolossn: fixed

frontend/src/lib/k8s/resourceQuota.ts
yolossn: Instead of showing both normalized and unnormalized values I changed it to just show normalized values and the unnormalized value is shown as tooltip.

joaquimrocha: To clarify. I think it's fine, and even probably recommended, to see the values right away. I just don't want all the values to be repeated. `21.27 GB (20 Gi)` is fine. `5 cores (5 cores)` or `0 (0)` is what I wanted to avoid. In this case we should just show `5 cores`.

I have tested your latest changes and the tooltip on hover effect is not clear (no indication it'll have that info unless the user actually hovers) so you might as well keep both values like before but special case when the values are the same, as mentioned.
Also, I see that in some cases the number + unit are separated by a space, in others they are not (and we should pluralize `core`):
Screenshot showing the changes

yolossn: Fixed it.
image
image

frontend/src/components/resourceQuota/Details.tsx
```diff
@@ -27,11 +28,21 @@ export default function ResourceQuotaDetails() {
                   },
                   {
                     label: t('resourceQuota|Used'),
-                    getter: item => item.used,
+                    getter: item =>
+                      compareUnits(item.used, normalizeUnit(item.name, item.used)) ? (
+                        <>{normalizeUnit(item.name, item.used)}</>
+                      ) : (
+                        <>{`${item.used} (${normalizeUnit(item.name, item.used)})`}</>
+                      ),
```
joaquimrocha, line 36: This code can be made more readable by having
```typescript
getter: item => {
  const normalizedUnits = normalizeUnit(item.name, item.used);
  ...

  return ...
}
```

And I don't think you need `<></>` as it looks like there's only one child.

Ditto for the other changes.

joaquimrocha (changes requested):

joaquimrocha (approved):

joaquimrocha:
Thanks for the changes @yolossn . I have merged it.

Issue #1115, opened 2023-04-28
Fix plugin link in Settings screen for small window size


vyncent-t:
Originally noticed by Tejas

<img width="436" alt="image" src="https://user-images.githubusercontent.com/78232183/235165289-79d4413e-59e9-4838-b4a1-5e93a7d2ed5d.png">

Currently:
This issue seems to exist only for desktop at the present (as the plugin feature under settings is only available in app).
When navigating to the settings page, resizing the window to a smaller size moves the submenu from the sidebar to the main page, which should allow page tab navigation but is not functional (neither from plugings page to settings or from settings to plugings).

I am working on a fix for this at the moment.

PR #1118, opened 2023-05-02
frontend NavigationTabs: Plugin Settings Nav Links Fix

vyncent-t:
Fixes for issue #1115

Essentially, this an issue on the desktop app when the user is within a small screen view and on the settings page, if you click the navigation tab while in settings or plugins they do not work.

image

I was able to track the issue down while trouble shooting it in web environment and it seems to do a few things.

The first lead with this issue occurs within the console
image

Currently:
PLEASE NOTE: Currently debugging this issue from the web client, the router.tsx file has been edited to work with web client, under the plugins route field, disabled is temporarily set as `disabled: helpers.isElectron()` and will be reset to `disabled: !helpers.isElectron()`

- I have tried turning on and off useCluster field as well as trying to use noCluster (although its deprecated) within the router.tsx file although neither of which seems to impact the issue.
- I am currently debugger walking through the NavigationTabs.tsx file attempting to locate any console logs to further pin point the error, the console logs currently do not print.
- It seems the error may be related to a null field being used within the paths being created by either the function getClusterPrefixedPath() or the getCluster() function being used within the NavigationTabs.tsx file 
image

### DIFF
--- a/frontend/src/components/Sidebar/NavigationTabs.tsx
+++ b/frontend/src/components/Sidebar/NavigationTabs.tsx
@@ -76,19 +76,27 @@ export default function NavigationTabs() {
     return null;
   }
 
+  /**
+   * This function is used to handle the tab change event.
+   *
+   * @param index The index of the tab that was clicked.
+   * @returns void
+   */
   function tabChangeHandler(index: number) {
     if (!subList) {
       return;
     }
-    let pathname;
 
     const url = subList[index].url;
-    if (url) {
-      pathname = generatePath(getClusterPrefixedPath(url), { cluster: getCluster()! });
+    if (url && getCluster()) {
+      history.push({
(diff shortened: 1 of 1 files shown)


vyncent-t:
Current WIP: This applies a draft fix for the nav tabs on the settings page basing the logic off the conditions used within SidebarItem.tsx 

It is currently set to run the cluster and url links first, if no cluster is found then run based off the url provided, this then allows the nav tabs to work without being dependent on the cluster base url set up. I have tried it within my local desktop env and the nav tabs with cluster prefixes work as intended and the settings nav tabs also work as intended. I will expand on this foundation for a more set approach.

TO DO:

- [ ] remove notes and console logs
- [ ] explore better alternatives

illume (approved):
👍 thanks

PR #1121, opened 2023-05-02
frontend: Use single-icon button for sidebar's add-cluster

### LINKED ISSUE #1120: Fix add-cluster button in narrow sidebar
When the sidebar is collapsed while in desktop mode, the add-cluster button is cut-off as it's too wide for the narrow sidebar.
So we should use a single icon button for when the sidebar is collapsed.

Screenshot showing the issue

joaquimrocha:
fixes #1120 

**Wide button**

Wide button

**Icon-only button**

Icon-only button

How to test:
- [ ] Check the several states of the sidebar: wide, collapsed by user, tablet-wide, mobile-wide + click the sidebar: the button should be, respectively, wide (with text), icon-only, icon-only, and wide again.
- [ ] Start Headlamp directly into these modes: verify that the button is the right variant

### DIFF
--- a/frontend/src/components/Sidebar/Sidebar.tsx
+++ b/frontend/src/components/Sidebar/Sidebar.tsx
@@ -10,11 +10,13 @@ import clsx from 'clsx';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks

PR #1122, opened 2023-05-04
Misc improvements for loading a cluster

joaquimrocha:
How to test:
- [ ] Run `make run-backend`: verify that the backend prints info that the Helm + dynamic-clusters are enabled
- [ ] Run the backend without enabling the dynamic-clusters; in the desktop app, start adding a cluster and select a value kube config and then a value cluster: it should show an error and go back to the cluster selection with the error banner (because the backend will not have the dynamic-cluster)
  - [ ] Click `back`: It should go back to the kube config loader view and no longer show the error dialog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,9 @@ frontend-build-storybook:
 	cd frontend && npm run build-storybook
 
 run-backend:
-	./backend/headlamp-server -dev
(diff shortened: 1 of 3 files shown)


yolossn (approved):
LGTM

PR #1125, opened 2023-05-05
frontend: Abstract AppLogo with plugins' logic into its own component

joaquimrocha:
The AppLogo logic was being used as a very simple component just
displaying the original logo and thus was used in two different places
with the logic for checking for a plugin overridden logo and error
boundary, etc.
These changes abstract all that into a component that can just be
added anywhere that it's needed.
The file is also moved from Sidebar/AppLogo to App/AppLogo, as it's a
more logical place for this component (the component is also in the
chooser, not just in the sidebar).

(This abstraction will be useful for when we implement the new design where the logo may be moved elsewhere)

How to test:
- [ ] Verify the logo is correctly displayed in both the sidebar and the chooser
- [ ] Do the same when the sidebar is collapsed
- [ ] Do the same but after running the `change-logo` example plugin


illume (approved):
👍 nice improvement. Thanks.

PR #1126, opened 2023-05-05
backend: Fix --html-static-dir cmd line option on windows

### LINKED ISSUE #1094: Backend html-static-dir doesn't work in Windows
**Command**: headlamp_server.exe --dev=true --proxy-urls=* --html-static-dir=frontend/build
**Path generated in line 191**: C:\Users\***\Desktop\***\frontend\build\C:
**Error Message**: CreateFile C:\Users\***\Desktop\****\frontend\build\C:\index: The filename, directory name, or volume label syntax is incorrect.
image
image

illume:
The path cleaning didn't work on windows.

Additionally, support unix paths for html-static-dir like 'frontend/build'.

Fixes #1094

## How to test?

- build on windows
- use cmd (not bash) and run `backend\headlamp-server.exe --dev=true --proxy-urls=* --html-static-dir=frontend/build`
- http://localhost:4466/static/ should work
- directory traversal should not work. Eg. `http://localhost:4466/../../../etc/passwd`

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -72,6 +72,8 @@ const (
 	STOPPED = "Stopped"
 )
 
+const isWindows = runtime.GOOS == "windows"
+
 type PortForward struct {
 	ID               string `json:"id"`
 	closeChan        chan struct{}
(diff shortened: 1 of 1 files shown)


illume:
Thanks @yolossn 

Yeah, I think chomey browsers do that automagically.

Here's an example URL that is handled by the browser but with curl with give you a 404.
```bash
curl https://github.com/headlamp-k8s/headlamp/pull/../1126
```

lijianzhi01:
Thanks @illume , it works for me as well.

yolossn (approved):

PR #1135, opened 2023-05-18
frontend: Add virtualization to SimpleTable component

ashu8912:


joaquimrocha:
We are using a new Table (Material-React-Table) so this no longer applies.

PR #1140, opened 2023-05-22
github: Trigger flatpak version update on tag publishing

joaquimrocha:


.github/workflows/trigger-flatpak-update.yml
```diff
@@ -0,0 +1,18 @@
+name: Trigger Flatpak Headlamp version update
+
+# See https://github.com/flathub/io.kinvolk.Headlamp
+on:
```
yolossn, line 4: Adding a manual trigger here will be useful to debug when things fail. Apart from that looks good.

yolossn (approved):

.github/workflows/trigger-flatpak-update.yml
joaquimrocha: Good point!

joaquimrocha: Done.

PR #1149, opened 2023-05-25
frontend: Do not show namespace filter in resources' namespace

joaquimrocha:
How to test:
- [ ] Go to the pods list in a namespace or node view: there should be a namespace table and namespace filter available for the pod table
- [ ] Go to the general pods list view + pod list view of a resource like deployment, replicaset, etc.: verify that no namespace column and no namespace filter is available for the pod table


ashu8912 (approved):
Tested, LGTM!!

PR #1150, opened 2023-05-26
Fix always visible sidebar

joaquimrocha:
- frontend: Allow setting up sidebar as null when no sidebar is needed
- frontend: Refactor and add more cases to the sidebar stories


lijianzhi01:
Thanks @joaquimrocha , it works for me. Good fix.

ashu8912 (approved):
LGTM!!

PR #1151, opened 2023-05-26
frontend plugins: Simplify headerActions, make adding new plugin APIs simpler, top bar action processors

illume:
This refactors action buttons a bit, and the register functions a bit. The idea is there are less files that the functionality touches, making further modifications require less effort and hunting around.

- [x] frontend: Refactor headerAction stuff into actionButtonSlice
- [x] frontend Resource: Refactor MainInfoSection into own component
- [x] plugins/examples/details-view: Bump headlamp-plugin to 0.6.0-alpha.27
- [x] frontend plugin: Improve index to export all registry exports (so new registry function just need to be exported, not also edited in a separate file)

### Acceptance:

- [x] There's a function registerTopBarActionsProcessor or similar which allows plugins to add their own processor
- [x] Top bar actions can be easily identified from the actions processor
- [x] The default top bar actions need switching to {id: '',  action: ...}
- [x] Top bar actions can be shuffled/updated/removed from the actions processor
- [x] There's an example in the examples plugin (see pod-counter)
- [x] There's documentation for the new method (the processor variant needs to be added as an example)

### How to test?

- [ ] Go to plugins/examples/pod-counter and `npm start`. See the notification icon moved to the end.
- [ ] Try it in mobile view too.

### To add to a plugin

```bash
cd yourplugin
npm install @kinvolk/headlamp-plugin@0.6.0-alpha.28 -D

# maybe optional...
# npx @kinvolk/headlamp-plugin@0.6.0-alpha.28 upgrade .

```

#### Example

```ts
import { DefaultAppBarAction, registerAppBarAction } from '@kinvolk/headlamp-plugin/lib';

// We can also reorder the actions in the app bar.
registerAppBarAction((_, actions) => {
  // Remove the notifications action button
  const newActions = actions.filter(action => action.id !== DefaultAppBarAction.NOTIFICATION);

  // Move the notification action to the end.
  const notificationAction = actions.filter(
    action => action.id === DefaultAppBarAction.NOTIFICATION
  )[0];
  newActions.push(notificationAction);

  return newActions;
});

```


illume:
@joaquimrocha @lijianzhi01 I'm finishing off some of the last checklist items this morning for this.

illume:
@joaquimrocha @lijianzhi01 This is ready to review.

lijianzhi01:
Hi @illume , it works for me. I used your example, which has just fit into my case.

plugins/examples/pod-counter/src/index.tsx
```diff
@@ -9,11 +9,19 @@ function PodCounter() {
...
+// We can also reorder the actions in the app bar.
+registerAppBarAction((_, actions) => {
+  // Remove the notifications action button
+  const newActions = actions.filter(action => action.id !== DefaultAppBarAction.NOTIFICATION);
+
+  // Add an extra pod counter action button
+  // newActions.push({action: <PodCounter />, id: 'pod-counter
+
+  // Move the notification action to the end.
+  const notificationAction = actions.filter(
```
joaquimrocha, line 21: Why not use `find`?

plugins/examples/pod-counter/src/index.tsx
```diff
@@ -9,11 +9,19 @@ function PodCounter() {
...
-//   },
-//   id: 'pod-counter-processor',
-// });
+// We can also reorder the actions in the app bar.
+registerAppBarAction((_, actions) => {
+  // Remove the notifications action button
+  const newActions = actions.filter(action => action.id !== DefaultAppBarAction.NOTIFICATION);
+
+  // Add an extra pod counter action button
+  // newActions.push({action: <PodCounter />, id: 'pod-counter
```
joaquimrocha, line 18: Is this intentionally left here? (the code comment is not finished).
Maybe we could add a label or something as an example here.

plugins/examples/pod-counter/src/index.tsx
```diff
@@ -9,11 +9,19 @@ function PodCounter() {
...
-// registerAppBarAction({
-//   processor: (_, actions) => {
-//     const newActions = actions.filter(action => action.id !== 'pod-counter')
-//     newActions.push({action: <PodCounter />, id: 'pod-counter'})
-//     return newActions;
-//   },
-//   id: 'pod-counter-processor',
-// });
+// We can also reorder the actions in the app bar.
+registerAppBarAction((_, actions) => {
```
joaquimrocha, line 13: You have previously asked me to name the functions in such examples, so it became clearer for the user. Should this one be `function reorderNotifications`.

plugins/examples/pod-counter/src/index.tsx
```diff
@@ -9,11 +9,19 @@ function PodCounter() {
...
+  const newActions = actions.filter(action => action.id !== DefaultAppBarAction.NOTIFICATION);
+
+  // Add an extra pod counter action button
+  // newActions.push({action: <PodCounter />, id: 'pod-counter
+
+  // Move the notification action to the end.
+  const notificationAction = actions.filter(
+    action => action.id === DefaultAppBarAction.NOTIFICATION
+  )[0];
+  newActions.push(notificationAction);
```
joaquimrocha, line 24: We're not caring about RTL languages here but maybe it's fine since it's an example and we don't want to make it complex.

plugins/examples/pod-counter/src/index.tsx
joaquimrocha: BTW, the reason why I made the header actions processor receive the `KubeObject` as first arg was that they're in the context of a details view which will have an object like that.
I think that if we never pass the 1st arg to the app bar processor, then we should remove that first arg.

joaquimrocha (changes requested):
Looking good! I have tested it and it worked. Left a couple of comments in the code.
I also did notice there's some different spacing in the top bar actions when the notifications button gets re-added to the end of it. Maybe worth investigating just to double check the components are being added in the same way.

Screenshot showing the spacing problem

frontend/src/redux/reducers/reducers.tsx
```diff
@@ -13,7 +13,7 @@ const reducers = combineReducers({
   clusterAction: clusterAction,
   config: config,
   plugins: pluginsReducer,
-  headerAction: headerActionReducer,
+  actionButtons: actionButtons,
```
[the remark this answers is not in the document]
illume: headerActions is still there, just under actionButtons.

actionButtons.headerActions
actionButtons.headerActionsProcessors
actionButtons.appBarActions
actionButtons.appBarActionsProcessors

plugins/examples/pod-counter/src/index.tsx
illume: > BTW, the reason why I made the header actions processor receive the KubeObject as first arg was that they're in the context of a details view which will have an object like that.
I think that if we never pass the 1st arg to the app bar processor, then we should remove that first arg.

I'd like to rethink the processor input so it can take a dictionary.

```
registerAppBarAction(({ actions }) => {
```

So it can be extended later. 

I can imagine people wanting different buttons depending on the resource.

But I think that can come in a different refactor?

illume (commented):
@joaquimrocha there's a couple of open items. Can you please have a look?

frontend/src/components/App/TopBar.tsx
```diff
@@ -29,8 +34,28 @@ import Notifications from './Notifications';
 
 export interface TopBarProps {}
 
+export function useAppBarActionsProcessed() {
+  const appBarActions = useTypedSelector(state => state.actionButtons.appBarActions);
+  const appBarActionsProcessors = useTypedSelector(
+    state => state.actionButtons.appBarActionsProcessors
+  );
+  return processAppBarActions(appBarActions, appBarActionsProcessors);
```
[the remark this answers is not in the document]
illume: This changed to not process in there. Do you think it's ok without memo now?

illume:
> I also did notice there's some different spacing in the top bar actions

I missed this one... ^ looking into it now.

illume:
@joaquimrocha I updated it to use ({actions})

frontend/src/redux/reducers/reducers.tsx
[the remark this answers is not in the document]
joaquimrocha: Yeah. I think it's still inaccurate but I won't block on this.

joaquimrocha (approved):
Thanks a lot for these changes. Ideally we'd not add commits that fix stuff introduced in commits of this same PR, for preventing extra work later when e.g. bisecting.
Given our timeline though, I will exceptionally approve as is.

illume:
@lijianzhi01 note, the usage changed and so did the dependency required. Please see updated PR description.

PR #1157, opened 2023-05-30
frontend: Check useClusterURL in NavigationTabs

joaquimrocha:
NavigationTabs were only checking if a cluster exists in getCluster()
and were ignoring whether the respective sidebar entry actually needs
a cluster for creating the URL to begin with.
This patch uses the mentioned boolean parameter of the sidebar entry
before attempting to use the cluster, thus fixing this issue.


ashu8912 (approved):
Tested, code also LGTM!!

Issue #1156, opened 2023-05-30
OIDC URL Redirect not respecting base-url


Jaguar-Kwok:
In the subpath, I set the base-url to /headlamp , however after dex OIDC, when headlamp handling /oidc-callback, the redirect path do not respect base-url.
<img width="431" alt="paint" src="https://github.com/headlamp-k8s/headlamp/assets/86696392/2a94c746-8dc8-4911-b440-9f11b5a281b5">

Issue #292, opened 2021-07-23
oidc login does not respect base uri


mhaddon:
If you run headlamp on a server, and configure both the oidc login and base uri, then when you go to login the authentication popup is not respecting the configured base uri.

joaquimrocha:
Hi @mhaddon , we (assumed we) have fixed this issue in #238 which went out in version 0.4.0. Can you confirm you're running that version?
Once it's confirmed we'll look into the issue.

Thanks in advance.

mhaddon:
It still seems to be an issue.

I am using 0.4.0.

joaquimrocha:
Thanks for confirming. We'll look into it, @mhaddon .

illume:
Fixed in:
- #1158

PR #1158, opened 2023-06-01
backend: Consider the base URL for OIDC-callback's redirect URL

Jaguar-Kwok:
#1156 #292 
A quick fix for the issue. Sorry for opening duplicated issue.


illume:
Thanks for the fix!

I'm not sure the backend test failure is related to this. I heard there might be some problem in the main branch. I'll get back to you about that. So if the test failure doesn't make sense to you don't worry too much, it might be something else :)

PR #1164, opened 2023-06-05
frontend apiProxy: Add some docs and debug logging

illume:
- frontend apiProxy: Add some docs and debug logging

Added some helpers to allow us to sprinkle plenty of verbose debug calls in there.

To add some verbose debugging to a module:
```ts
import { isDebugVerbose } from './helpers';
if (isDebugVerbose('k8s/apiProxy')) {
    console.debug('k8s/apiProxy', {dataToLog});
}
```
You can also include a symbol name:
```ts
import { isDebugVerbose } from './helpers';
if (isDebugVerbose('k8s/apiProxy@refreshToken')) {
    console.debug('k8s/apiProxy@refreshToken', {dataToLog});
}
```
In that example:
- 'k8s/apiProxy' is the module name.
- 'refreshToken' is the function symbol name.
To turn verbose debugging on via code in that module:
```ts
import { debugVerbose } from './helpers';
debugVerbose('k8s/apiProxy')
// or for everything in refreshToken:
debugVerbose('k8s/apiProxy@refreshToken')
```
To turn it on for multiple modules via environment variable:
```bash
REACT_APP_DEBUG_VERBOSE="k8s/apiProxy i18n/config" make run-frontend
```
To turn it on via environment variable for all modules:
```bash
REACT_APP_DEBUG_VERBOSE="all" make run-frontend
```
To add some verbose debugging to a module:
```ts
import { isDebugVerbose } from './helpers';
if (isDebugVerbose('k8s/apiProxy')) {
    console.debug('k8s/apiProxy', {dataToLog});
}
```
You can also include a symbol name:
```ts
import { isDebugVerbose } from './helpers';
if (isDebugVerbose('k8s/apiProxy@refreshToken')) {
    console.debug('k8s/apiProxy@refreshToken', {dataToLog});
}
```
In that example:
- 'k8s/apiProxy' is the module name.
- 'refreshToken' is the function symbol name.
To turn verbose debugging on via code in that module:
```ts
import { debugVerbose } from './helpers';
debugVerbose('k8s/apiProxy')
// or for everything in refreshToken:
debugVerbose('k8s/apiProxy@refreshToken')
```
To turn it on for multiple modules via environment variable:
```bash
REACT_APP_DEBUG_VERBOSE="k8s/apiProxy i18n/config" make run-frontend
```
To turn it on via environment variable for all modules:
```bash
REACT_APP_DEBUG_VERBOSE="all" make run-frontend
```

Some notes:

- Gives us the line number and file of the log in developer console. If it was in a wrapper function it just shows the wrapper function line number.
- Turned off by default, and the message doesn't even get constructed if it's off. This is important do high frequency messages so not impact performance.
- ON/OFF via environment variable REACT_APP_DEBUG_VERBOSE='k8s/apiProxy'
- ON/OFF via code debugVerbose('k8s/apiProxy').  So can easily turn it on when debugging.
- Also can turn on just a function debugVerbose('k8s/apiProxy@refreshToken')

### DIFF
--- a/frontend/src/helpers/index.ts
+++ b/frontend/src/helpers/index.ts
@@ -37,6 +37,90 @@ function isElectron(): boolean {
   return false;
 }
 
+/** used by isDebugVerbose and debugVerbose */
+const verboseModDebug: string[] = [];
+
+/**
+ * To allow us to include verbose debug information for a module.
+ *
+ * - Gives us the line number and file of the log in developer console.
+ *   If it was in a wrapper function it just shows the wrapper function line number.
+ * - Turned off by default, and the message doesn't even get constructed if it's off.
+ *   This is important do high frequency messages so not impact performance.
+ * - ON/OFF via environment variable REACT_APP_DEBUG_VERBOSE='k8s/apiProxy'
+ * - ON/OFF via code debugVerbose('k8s/apiProxy').
+ *   So can easily turn it on when debugging.
+ * - Also can turn on just a function debugVerbose('k8s/apiProxy@refreshToken')
+ *
+ * @param modName only show verbose debugging for this module name.
+ * @returns true if verbose debugging should be done.
+ *
+ * @example
+ *
+ * To add some verbose debugging to a module:
+ * ```ts
+ * import { isDebugVerbose } from './helpers';
+ * if (isDebugVerbose('k8s/apiProxy')) {
+ *     console.debug('k8s/apiProxy', {dataToLog});
+ * }
+ * ```
+ *
+ * You can also include a symbol name:
+ * ```ts
+ * import { isDebugVerbose } from './helpers';
+ * if (isDebugVerbose('k8s/apiProxy@refreshToken')) {
+ *     console.debug('k8s/apiProxy@refreshToken', {dataToLog});
+ * }
+ * ```
+ *
+ * In that example:
+ * - 'k8s/apiProxy' is the module name.
+ * - 'refreshToken' is the function symbol name.
+ *
+ * To turn verbose debugging on via code in that module:
+ * ```ts
+ * import { debugVerbose } from './helpers';
+ * debugVerbose('k8s/apiProxy')
+ *
+ * // or for everything in refreshToken:
+ * debugVerbose('k8s/apiProxy@refreshToken')
+ * ```
+ *
+ * To turn it on for multiple modules via environment variable:
+ * ```bash
+ * REACT_APP_DEBUG_VERBOSE="k8s/apiProxy i18n/config" make run-frontend
+ * ```
+ *
+ * To turn it on via environment variable for all modules:
+ * ```bash
+ * REACT_APP_DEBUG_VERBOSE="all" make run-frontend
+ * ```
+ */
+export function isDebugVerbose(modName: string): boolean {
+  if (verboseModDebug.filter(mod => modName.indexOf(mod) > 0)) {
+    return true;
+  }
+
+  return (
+    process.env.REACT_APP_DEBUG_VERBOSE === 'all' ||
+    !!(
+      process.env.REACT_APP_DEBUG_VERBOSE &&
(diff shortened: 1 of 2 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -28,6 +28,17 @@ const MIN_LIFESPAN_FOR_TOKEN_REFRESH = 10; // sec
 
 let isTokenRefreshInProgress = false;
 
+/**
+ * Like console.debug, except it's off by default and prefixed with 'apiProxy'.
+ */
+function debug_log(...args: any[]) {
```
joaquimrocha, line 34: Should this be lower camel cased (for consistency)?

joaquimrocha (approved):
Nice debugging. I added a question about consistency in the function's name, but otherwise should be good to go.

frontend/src/helpers/index.ts
```diff
@@ -37,6 +37,90 @@ function isElectron(): boolean {
...
+}
+
+/**
+ * onDebugVerbose turns on verbose debugging for a module.
+ *
+ * @param modName turn on verbose debugging for this module name.
+ *
+ * @see isDebugVerbose
+ */
+export function onDebugVerbose(modName: string): void {
```
joaquimrocha, line 120: `on`DebugVerbose seems like we're responding to the fact that we have a verbose debug mode (like onMessage, onCancel, that we have in other places) but this seems to turn on debug, so maybe it should be called `enableVerboseDebug`.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -10,15 +10,17 @@
...
+import helpers, { getHeadlampAPIHeaders, isDebugVerbose } from '../../helpers';
 import store from '../../redux/stores/store';
 import { getToken, logout, setToken } from '../auth';
 import { getCluster } from '../util';
 import { ResourceClasses } from '.';
 import { KubeMetadata, KubeMetrics, KubeObjectInterface } from './cluster';
 import { KubeToken } from './token';
 
+// This module has verbose debugging sprinkled throughout.
+// onDebugVerbose('k8s/apiProxy');
```
joaquimrocha, line 22: If this is here so it can be uncommented, then maybe part of the comment above it should be "Uncomment the following line to enable verbose debugging."

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -91,7 +93,10 @@ async function refreshToken(token: string | null) {
     return;
   }
 
-  console.debug('Refreshing token');
+  if (isDebugVerbose('k8s/apiProxy@refreshToken')) {
+    console.debug('k8s/apiProxy@refreshToken', 'Refreshing token');
+  }
```
joaquimrocha, line 98: Seems like all uses of `isDebugVerbose` are also printing a message with the same prefix as the param as the former function, so maybe this should be turned into a `printVerboseDebug('k8s/apiProxy@refreshToken', 'Refreshing token')`, so it internally does the check already?

joaquimrocha (changes requested):
Thanks. I have added a couple of comments.

frontend/src/lib/k8s/apiProxy.ts
illume: This trade off was done to give better log messages. If they are wrapped in a helper function then the log message says it is from the file where the helper is. This way the log message includes the file and line number where the debug log is, and stepping into the debugger is easier too.

illume:
Updates from convos were made:

- onDebugVerbose -> debugVerbose
- Updated apiProxy comment about "uncomment to enable verbose debugging".

joaquimrocha (approved):
I think it'd be fine if we just check for verbosity without any arg and print the messages if true.
We can always filter in the console. But anyway, it's not a strong push back. I will approve it.

PR #1165, opened 2023-06-06
frontend ResourceTable: Throttle data updates to 1Hz

illume:
Throttles the data updates on tables to once per second. When new data comes in it is shown right away, but if more comes in within the next second it is not shown until 1 second has past. So there should be new data right away in most cases, and it should be limited to how fast it updates otherwise.

### How to test

- [ ]  create many events per second on a cluster and view the cluster overview page
- [ ] filtering still works ok

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -1,4 +1,5 @@
 import { useTheme } from '@material-ui/core/styles';
+import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { KubeObject } from '../../../lib/k8s/cluster';
 import { useFilterFunc } from '../../../lib/util';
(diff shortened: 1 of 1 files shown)


illume:
@ashu8912 I added a small improvement. It now doesn't throttle for the first couple of times. This is because often the first value is `null` whilst loading happens. Now there there is no throttling until real data comes in. Before there could be an unnecessary wait added for up to 0.8 seconds because the first value to enter is `null`.

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -39,16 +40,57 @@ export default function ResourceTable(
...
+  //
+  //   numEffected == 0,  null, or undefined whilst loading.
+  //   numEffected == 1,  real data.
+  const numEffected = useRef(0);
+
+  useEffect(() => {
+    const now = Date.now();
+
+    if (now >= lastEffected.current + interval || numEffected.current < 2) {
+      numEffected.current = numEffected.current + 1;
```
joaquimrocha, line 66: Would it be more readable if `numEffected` was just `doThrottle = useRef(false)` and we'd set it to true if `!!value`.
So it be `if (now >= lastEffected.current + interval || !doThrottle.current)`.

Would this be logically equivalent?

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -39,16 +40,57 @@ export default function ResourceTable(
...
+  }, [value, interval]);
+
+  return throttledValue;
+}
+
 function TableFromResourceClass(props: ResourceTableFromResourceClassProps) {
   const { resourceClass, id, ...otherProps } = props;
   const [items, error] = resourceClass.useList();
+  // throttle the update of the table to once per second
+  const throttledItems = useThrottle(items, 1000);
```
joaquimrocha, line 86: I think this is so useful that I think we should probably add it to the `useList` itself and have a param to it that passes `{throttle: timeInMs}` to it.

joaquimrocha (approved):
Approved with a comment to hopefully make it more readable but also we should integrate it into useList IMO and have all calls throttled by default.

frontend/src/components/common/Resource/ResourceTable.tsx
illume: This could come in a later PR.

frontend/src/components/common/Resource/ResourceTable.tsx
illume: 🤷 The way it's done makes it more apparent it's a counter.

joaquimrocha: It's a difficult to understand 1-2 counter that checks for whether we've finally have content, but in the interest of time I will merge it.

PR #1167, opened 2023-06-09
frontend: Fix setting up header actions when they're not identified

joaquimrocha:
When header actions are not in the { id, action } format, we create
this format for them but were adding the action itself as a function
() => action. If the action at this point was a component, setting it
up as a function was resulting in always recreating the component which
led to the side effect of having those header actions blinking whenever
the related resource was changed.

This patch fixes that by setting up the action directly instead of
using the function format.


ashu8912 (approved):
Nice, LGTM!!

illume (approved):
👍 thanks.

Issue #1160, opened 2023-06-02
Alternative methods to provide OIDC configuration


maxsargentdev:
Hi,

I have noticed that the only way to provide the OIDC configuration (which some of is sensitive) is directly in the values file.

It would be good to be able to support loading this configuration from environment variables or files. If you are using a GitOps approach for managing your deployment of headlamp the current setup forces you into bad practice.

Is this a welcome change? If so I will make it.

Thanks,
Max Sargent

maxsargentdev:
I did not read enough of the source, I have found the HEADLAMP_CONFIG_ prefix now, thanks.

maxsargentdev:
Re-opening, this change will now be a request for:

- Trivial change to the helm chart to support setting environment variables from a secret
- Updating the koanf usage so that it can read the settings from a flat yaml file mounted at a pre-specified location

Thanks.

joaquimrocha:
Thanks @maxsargentdev . We will look into this.

PR #1170, opened 2023-06-10
charts: OIDC config via secrets

knrt10:
When installing charts, oidc was being provided as config and since this contains sensitive information it should have been provided as secret.

Fixes: #1160

### DIFF
--- a/charts/headlamp/README.md
+++ b/charts/headlamp/README.md
@@ -67,11 +67,11 @@ $ helm install my-headlamp headlamp/headlamp --namespace kube-system
 
 ### Headlamp Configuration
 
-| Key | Type | Default | Description |
-|-----|------|---------|-------------|
-| config.baseURL | string | `""` | base url path at which headlamp should run |
-| config.oidc.clientID | string | `""` | OIDC client ID |
-| config.oidc.clientSecret | string | `""` | OIDC client secret |
-| config.oidc.issuerURL | string | `""` | OIDC issuer URL |
-| config.oidc.scopes | string | `""` | OIDC scopes to be used |
-| config.pluginsDir | string | `"/headlamp/plugins"` | directory to look for plugins |
+| Key                      | Type   | Default               | Description                                |
+|--------------------------|--------|-----------------------|--------------------------------------------|
+| config.baseURL           | string | `""`                  | base url path at which headlamp should run |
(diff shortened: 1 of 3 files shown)


yolossn (commented):
I think even in this solution if a person is able to do `kubectl top` on the headlamp pod the secrets will be shown. We must use the ENV support in Headlamp to tackle this. 
For more info refer.
https://github.com/headlamp-k8s/headlamp/blob/890f19c8570d297bbf9e2470d6af49dbb37efe78/backend/pkg/config/config.go#L53-L60

knrt10:
I had a talk with @yolossn and we came to conclusion that he got confused `kubectl top` with `docker top`. With the context of the issue the current change will work with Headlamp as a security feature.

charts/headlamp/README.md
```diff
@@ -70,8 +70,13 @@ $ helm install my-headlamp headlamp/headlamp --namespace kube-system
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | config.baseURL | string | `""` | base url path at which headlamp should run |
-| config.oidc.clientID | string | `""` | OIDC client ID |
-| config.oidc.clientSecret | string | `""` | OIDC client secret |
-| config.oidc.issuerURL | string | `""` | OIDC issuer URL |
-| config.oidc.scopes | string | `""` | OIDC scopes to be used |
 | config.pluginsDir | string | `"/headlamp/plugins"` | directory to look for plugins |
+
```
yolossn, line 74: For the sake of backward compatibility cant we have the values as `config.oidc.*` and use it as secrets instead of args?

knrt10: Makes sense. will do it

charts/headlamp/values.yaml
```diff
@@ -29,6 +29,7 @@ config:
   # -- base url path at which headlamp should run
   baseURL: ""
   oidc:
+    create: false
```
yolossn, line 32: I think we can achieve the same output without having to introduce a new variable here. We can use the `with` helm template operator to achieve this.
https://helm.sh/docs/chart_template_guide/variables/

knrt10: Sure, updated

yolossn (approved):
lgtm

PR #1172, opened 2023-06-13
frontend: Only process notifications when the bell icon is clicked

ashu8912:


illume (approved):
Nice fix. 🎉

frontend/src/helpers/index.ts
```diff
@@ -204,9 +204,35 @@ const defaultMaxNotificationsStored = 200;
...
 function storeNotifications(notifications: Notification[], options: NotificationStoreOptions = {}) {
   const { max = defaultMaxNotificationsStored } = options;
-  localStorage.setItem('notifications', JSON.stringify(notifications.slice(0, max)));
+  const throttledSetItem: (notifications: Notification[]) => void = throttle(
+    (notifications: Notification[]) => {
+      localStorage.setItem('notifications', JSON.stringify(notifications.slice(0, max)));
+    },
+    5000
+  );
+  throttledSetItem(notifications);
```
joaquimrocha, line 216: I think this will only apply the items after 5s but we should rather delay it only when there's been a call less than 5s before. i.e. A 1st call, and any calls that happen > 5s since the previous one should be done immediately.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -178,6 +178,10 @@ export default function Notifications() {
   const history = useHistory();
 
   useEffect(() => {
+    // only perform processing if the bell icon is clicked
+    if (!anchorEl) {
+      return;
+    }
```
joaquimrocha, line 184: AFAIU this still affects the normal functioning of the notifications. (or maybe you left this here by mistake)

joaquimrocha (dismissed):

frontend/src/helpers/index.ts
ashu8912: This PR doesn't work on the throttling part now, that is something i would like to handle in a different PR

illume:
I tested this, and it does fix notification marking in the full page view.

Issue #670, opened 2022-08-17
OIDC + dex, expected audience \"kubernetes\" got [\"headlamp\"]]"


matt-deboer:
## Description

This not really a bug, just a signpost for others who may hit the same issue. When configuring headlamp with OIDC using dex, and had configured a static client for 'headlamp', which was working successfully. I had another static client configured for 'kubernetes' for use by the apiservers' oidc configuration. When authenticating to headlamp, my user had no permission to list anything. 

In examining the kube apiserver logs, I saw the following error: 
```
"Unable to authenticate the request" err="[invalid bearer token, oidc: verify token: oidc: expected audience \"kubernetes\" got [\"headlamp\"]]"
```
To work around this, I added `audience:server:client_id:kubernetes` to the `-oidc-claims` arg specified for the headlamp binary, and added `trustedPeers: [ 'headlamp']` to the static client config (in dex) for the 'kubernetes' static client, which resulted in tokens being issued with `aud: ['headlamp', 'kubernetes']` which were accepted by the kube apiserver, and then all worked as desired.

I now realize that this also could have worked if the headlamp and the apiserver shared a single static client id in dex, with redirectURIs specified for each (which is maybe what the documentation was trying to say to begin with?)

The only potential fix is maybe more clarification in the docs about these 2 possible configuration scenarios.

joaquimrocha:
Thanks @matt-deboer ! We shall add the relevant bits to our docs.

PR #1175, opened 2023-06-14
OIDC doc fix

knrt10:
Fixes: #670

### DIFF
--- a/docs/development/oidc.md
+++ b/docs/development/oidc.md
@@ -1,0 +1,84 @@
+---
+title: Testing OIDC
+weight: 5
+---
+
+OIDC (OpenID Connect) is a protocol that allows web applications to authenticate users through an identity provider and obtain basic profile information about the user.
+
+## Testing OIDC with Kubernetes API Server
+
+This guide will walk you through the process of testing OIDC on a Minikube cluster using an Azure App Registration. OIDC allows you to authenticate and interact with the cluster using kubectl.
+
+### Create an Application on Azure
+
+- Create an [application on Azure](https://learn.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app) and specify the redirect URL as `http://localhost:8000`.
+- Generate a secret for the application under "Certificates and Secrets" and make note of its value.
+
+### Start Minikube
+
+- Start Minikube without enabling OIDC: `minikube start`. This is done to avoid RBAC (Role-Based Access Control) issues during startup. However, we can enable OIDC once the cluster is running.
+
+### SSH into Minikube
+
+- Once the Minikube cluster is started, SSH into it: `minikube ssh`.
+- Install the Vim text editor: `sudo apt-get update && sudo apt-get install vim`.
+
+### Edit the kube-apiserver Configuration
+
+- Edit the kube-apiserver configuration file using Vim: `sudo vi /etc/kubernetes/manifests/kube-apiserver.yaml`.
+- Add the following lines to the configuration file:
+
+```bash
+    - --oidc-issuer-url=https://sts.windows.net/<tenantID from application>/
+    - --oidc-client-id=<Application Client ID>
+    - --oidc-username-claim=email
+```
+
+- Save the changes to the configuration file and exit Vim.
+
+### Wait for API Server Restart
+
+- Wait for approximately 1 minute for the API server to restart and apply the new configuration.
+
+### Install kubelogin
+
+- Install [kubelogin](https://github.com/int128/kubelogin), which is a tool for testing Kubernetes authentication:
+
+```bash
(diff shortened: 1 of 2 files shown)


docs/development/oidc.md
```diff
@@ -0,0 +1,85 @@
+---
+title: OIDC
```
illume, line 2: How about something with a bit more?

> Testing OIDC with Kubernetes API Server: Step-by-Step Guide

Probably not that^... but something to indicate it's about testing OIDC and Headlamp.

knrt10: Updated, please review.

docs/development/oidc.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Testing OIDC
+weight: 5
+---
+
+OIDC is a protocol that allows web applications to authenticate users through an identity provider and obtain basic profile information about the user.
+
+## Testing OIDC with Kubernetes API Server
+
+This guide will walk you through the process of testing OIDC (OpenID Connect) on a Minikube cluster using an Azure App Registration. OIDC allows you to authenticate and interact with the cluster using kubectl:
```
joaquimrocha, line 10: `(OpenID Connect)` this should be on the 1st use of OIDC, line 6.

joaquimrocha: Please remove the last `:`. It works without it.

docs/development/oidc.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Create an Application on Azure
+
+- Create an application on Azure and specify the redirect URL as `http://localhost:8000`.
```
joaquimrocha, line 14: It'd be great to add links to Azure's docs every time we refer to something where there are docs about if they help users to more easily understand what to do. A link to the docs on creating this application would be very useful.

docs/development/oidc.md
```diff
@@ -0,0 +1,84 @@
...
+    - --oidc-issuer-url=https://sts.windows.net/<tenantID from application>/
+    - --oidc-client-id=<Application Client ID>
+    - --oidc-username-claim=email
+```
+
+- Save the changes to the configuration file and exit Vim.
+
+### Wait for API Server Restart
+
+- Wait for approximately 1 minute for the API server to restart and apply the new configuration.
```
joaquimrocha, line 41: Is there a way we can check whether the API server has been restarted?

docs/development/oidc.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Install Kubelogin
+
+- Install Kubelogin, which is a tool for testing Kubernetes authentication:
+
+```bash
+# Homebrew (macOS and Linux)
+brew install int128/kubelogin/kubelogin
+```
```
joaquimrocha, line 50: I don't know of anyone who uses homebrew on Linux.
You can link to the kubelogin installation docs, and then say "Here's how to do it with Homebrew as an example:"

docs/development/oidc.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Test Access to Pods
+
+- Now you can test access to pods using the OIDC user. Try running the following command to get a list of pods:
+
+```bash
+kubectl get po -A
+```
+
+By following these steps, you can test OIDC with the Kubernetes API Server and authenticate using the OIDC user you created.
```
joaquimrocha, line 84: The guide looks good but I get the impression this is more about generally using OIDC and do not see where's the bridge to Headlamp. i.e. will Headlamp just pick up the OIDC config from the moment you add it to kubeconfig? Will the "Login" button in Headlamp work?

joaquimrocha (dismissed):
Added a few more comments.

docs/development/oidc.md
knrt10: Unfortunately no, since any request made will result in sever timeout, even if one watches it by checking pods like `kubectl get po -A -w`

docs/development/oidc.md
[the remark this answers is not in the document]
knrt10: Thank you, updated

docs/development/oidc.md
illume: I guess restarting minikube is the other option?

> Wait for approximately 1 minute for the API server to restart and apply the new configuration. Or run `minikube stop && minikube start` to restart it manually.

docs/development/oidc.md
knrt10: To do that, one need to configure Headlamp as well. Will reference accessing headlamp doc to it.

docs/development/oidc.md
knrt10: There is no need to restart since on change to api server file will automatically restart it

illume (approved):
🎉

PR #1176, opened 2023-06-14
frontend: Only get non-normal events for notifications

joaquimrocha:
This will make the number of events to process much smaller and thus
be more optimal.


illume (approved):
👍

PR #1177, opened 2023-06-14
frontend: Show number of warnings in Overview

joaquimrocha:


frontend/src/components/cluster/Overview.tsx
```diff
@@ -125,7 +131,7 @@ function EventsSection() {
           titleSideActions={[
             <FormControlLabel
               checked={isWarningEventSwitchChecked}
-              label={t('Warnings')}
+              label={t('Only warnings ({{ numWarnings }})', { numWarnings })}
```
ashu8912, line 134: @joaquimrocha we have a filter also to the table where we show events, but this count won't change in case I apply a filter. Should we change that, or should we change this to say total warnings??

frontend/src/components/cluster/Overview.tsx
```diff
@@ -77,6 +77,7 @@ function EventsSection() {
   const [isWarningEventSwitchChecked, setIsWarningEventSwitchChecked] = React.useState(
     Boolean(localStorage.getItem(EVENT_WARNING_SWITCH_FILTER_STORAGE_KEY))
   );
+  const [events, eventsError] = Event.useList();
```
ashu8912, line 80: we use `ResourceTable` down below that fetches the events again, can we change that to something like `MainInfoSection` so it uses this same events

ashu8912 (changes requested):

PR #1188, opened 2023-06-23
Customizable columns

### LINKED ISSUE #366: [RFE] Customizable columns in various list views (pods, nodes, etc.)
## Current situation

Currently list of pods or nodes shows predefined set of columns:
- Pods show name, namespace, restarts, status and age.
- Nodes show name, readiness, cpu, memory and age.

These lists missing other information that can be very useful in overall summary in various situations. For example, user can't see nodes uptime or pod's node placement in overall list and has to click on each item to check.
Similar statement can be done for other resource types.

## Impact

Lack of important information makes it harder to assess state of overall state of cluster and see patterns without resorting to kubectl.

## Ideal future situation

Being able to add/remove/rearrange columns in the list views, and filter by custom columns will help a lot in analyzing the overall status. Have a broader list of supported columns, that can be added/removed from views, as well as ability to memorize those per view will be very useful.

joaquimrocha:
This PR adds a "column chooser" to all the list views in Headlamp, so users can choose to hide/show certain columns.

Acceptance Criteria:
- [ ] For all list views, there's a column-chooser button next to the filter button:
Screenshot showing the column-chooser button
- [ ] When clicked, the column-chooser button shows a popup with all the columns in the respective table
- [ ] Clicking to hide/show columns does effectively hide/show them
- [ ] When moving to another view and then back, Headlamp "remembers" the hidden/shown columns

fixes #366


frontend/src/components/common/Resource/ResourceTableColumnChooser.tsx
```diff
@@ -0,0 +1,87 @@
...
+      }}
+    >
+      <List>
+        {currentColumns.map((column, index) => {
+          const labelId = `column-index-${index}`;
+
+          return (
+            <ListItem
+              key={labelId}
+              role={undefined}
```
ashu8912, line 65: small doubt what does role undefined signify here??

frontend/src/helpers/index.ts
```diff
@@ -277,6 +302,8 @@ const exportFunctions = {
   storeClusterSettings,
   loadClusterSettings,
   getHeadlampAPIHeaders,
+  storeTableSettings,
```
ashu8912, line 305: This commit and the previous commit has a lot of same changes, maybe squash them??

ashu8912 (approved):
I have looked at the code as well as tested this on each resource view,
Small doubts otherwise this LGTM, some real nice changes thanks @joaquimrocha 
1. Right now customizing column is limited to ResourceTable only but should we add this to other tables as well like SimpleTable or NamespaceTable??
2. The identity of a table right now is attached to the resource type it is, should we attach it to the view instead?? Example if i customize the column in Pod table anywhere Pod table is used the changes will be reflected with the changes we have as of now.

frontend/src/components/common/Resource/ResourceTableColumnChooser.tsx
joaquimrocha: Good catch. I will delete it.

joaquimrocha:
Thanks for the review Ashu.

> 1. Right now customizing column is limited to ResourceTable only but should we add this to other tables as well like SimpleTable or NamespaceTable??

This was done on purpose like this. The SimpleTable is increasing in complexity and for having it support the tables, it'd require getting its own `id` prop and others.
So my view is that the `SimpleTable` already shows the columns we want to show or hide, by providing a different list of columns. So the mechanism to select them in the UI should be done at a higher level.
Since the ResourceTable already has an id and is being used in a lot of places, we can use it to set up the column chooser.
About the NamespaceTable, I didn't add support for choosing its rows, since the new design doesn't need it, but since the ColumnChooser is its own independent component, we can easily add it later.

> 2. The identity of a table right now is attached to the resource type it is, should we attach it to the view instead?? Example if i customize the column in Pod table anywhere Pod table is used the changes will be reflected with the changes we have as of now.

I am not sure I understand this one, but each table should have its own ID. e.g. if a pods table is in 2 different places, it should have 2 different ids (unless it's representing the exact same view). If we have an id being used in different locations, then we should rename the id.

frontend/src/helpers/index.ts
joaquimrocha: I think you mean the previous 2 commits. I want to have them separated in case we end up reverting parts of this.

ashu8912 (approved):

ashu8912:
@joaquimrocha there is a merge conflict other then that this LGTM

PR #1197, opened 2023-06-27
Fix drain, cordon icon

ashu8912:
Steps to reproduce 
1. Go to node detail view and press either the drain node or the cordon node icon
2. Pressing the drain or cordon node icon shows a notistack saying cancel the drain or cordon

Acceptance Criteria

- [ ] Pressing the cancel button on notistack cancels the drain or cordon process with a message 
- [ ] When the cancel button is clicked the icons return back to there enable state which were earlier staying disabled even on cancel click


joaquimrocha:
Nice change to the Redux Sagas logic. I think we had a request for something like this a while ago. LGTM.

PR #1203, opened 2023-07-03
frontend: Fix warning switch always on bug.

ashu8912:
Issue:
go to cluster overview page, click on events warning switch, 
now click again to turn it off, 
switch to some other page and come back to cluster overview page, you will find the switch is always on if switched once.

This is fixed by parsing the value of event switch before making it as a default value


joaquimrocha (approved):

joaquimrocha (approved):
LGTM. Thanks @ashu8912 .

PR #1208, opened 2023-07-03
Use cluster health in alert notification

joaquimrocha:
- frontend: Add testClusterHealth function
- frontend: Use testClusterHealth in the AlertNotification

How to test:
- [ ] Try accessing a the pods list view when the cluster is down
- [ ] The banner no longer shows up in the login, token, and settingsCluster routes (because the first 2 routes already inform the user about the issue, and the 3rd one doesn't need the banner's context)


illume (approved):
👍  thanks

PR #1216, opened 2023-07-07
docs:add headlamp & plugins

jingchu000:
docs:add headlamp & plugins & README_CN

fixes https://github.com/headlamp-k8s/headlamp/issues/930

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -75,6 +75,7 @@ const (
 
 const isWindows = runtime.GOOS == "windows"
 
+// PortForward Information for internal use for each port that is being forwarded.
 type PortForward struct {
 	ID               string `json:"id"`
 	closeChan        chan struct{}
@@ -89,6 +90,8 @@ type PortForward struct {
 	Error            string `json:"error"`
 }
 
+// PortForwardPayload Request parameters for Port forwarding.
+// RequestURL:/portforward  POST.
 type PortForwardPayload struct {
 	ID               string `json:"id"`
 	Namespace        string `json:"namespace"`
@@ -136,6 +139,10 @@ func portforwardstore(p PortForward) {
 	portForwards[p.Cluster] = append(portForwards[p.Cluster], p)
 }
 
+// stopOrDeletePortForward stops or deletes a port forward by its cluster and id.
+// It takes three parameters: cluster is the name of the cluster, id is the unique identifier of the port forward,
+// isStopRequest is a boolean value indicating whether to stop or delete the port forward.
+// It returns an error value indicating whether the operation is successful or not.
 func stopOrDeletePortForward(cluster string, id string, isStopRequest bool) error {
 	clusterPortForwards, ok := portForwards[cluster]
 	if ok {
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ app/electron/src/*
 docs/development/storybook/
 .plugins
 node_modules
+.idea/
--- a/backend/cmd/plugins.go
+++ b/backend/cmd/plugins.go
@@ -165,6 +165,10 @@ func pluginBasePathListForDir(pluginDir string, baseURL string) ([]string, error
 	return pluginListURLs, nil
 }
 
+// addPluginRoutes adds plugin routes to a router.
+// It serves plugin list base paths as json at “/plugins”.
+// It serves plugin static files at “/plugins/” and “/static-plugins/”.
+// It disables caching and reloads plugin list base paths if not in-cluster.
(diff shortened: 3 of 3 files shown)


joaquimrocha:
Thanks @jingchu000 .
@yolossn / @illume , can you check the backend? I think the docs need some updates and the commits need to be split.

jingchu000:
@illume

README_CN.md
```diff
@@ -0,0 +1,83 @@
+# Headlamp <img align="right" width=384 src="docs/headlamp_light.svg">
+
+Headlamp 是一个使用简单且可扩展的Kubernetes web UI。
```
illume, line 3: @lijianzhi01 I wonder if you would mind reviewing this file?

backend/cmd/headlamp.go
```diff
@@ -39,6 +36,10 @@ import (
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
 
+	"github.com/headlamp-k8s/headlamp/backend/pkg/cache"
```
illume, line 39: For my own interest, can you please explain why these import lines were moved?

backend/cmd/headlamp.go
```diff
@@ -75,6 +76,7 @@ const (
 
 const isWindows = runtime.GOOS == "windows"
 
+// PortForward Information forwarded by Port forwarding
```
illume, line 79: I'm not sure about this one:
> PortForward Information forwarded by Port forwarding

How about something like this instead?
> PortForward Information for internal use for each port that is being forwarded.

It adds that it is an internal structure of the ports being forwarded.

backend/cmd/headlamp.go
```diff
@@ -136,6 +140,10 @@ func portforwardstore(p PortForward) {
 	portForwards[p.Cluster] = append(portForwards[p.Cluster], p)
 }
 
+// stopOrDeletePortForward stops or deletes a port forward by its cluster and id.
+// It takes three parameters: cluster is the name of the cluster, id is the unique identifier of the port forward,
+// isStopRequest is a boolean value indicating whether to stop or delete the port forward.
+// It returns an error value indicating whether the operation is successful or not.
```
illume, line 146: For your consideration... maybe this instead?
> It returns an error if there is one, or nil if there is not.

illume (changes requested):
Thank you for this.

I left a few comments about some issues the linter found, a few other small things, and also asked someone else to review your README_CN.md.

Joaquim suggested that the commits should be split up into separate ones per topic or change. For example, one commit for the README_CN, one for the headlamp.go stopOrDeletePortForward change. However, if you prefer to not split it up into different commits... let me know and when you've finished all of your changes I can assist doing this for you.

backend/cmd/headlamp.go
```diff
@@ -89,6 +91,8 @@ type PortForward struct {
 	Error            string `json:"error"`
 }
 
+// PortForwardPayload Request parameters for Port forwarding
+// RequestURL:/portforward  POST
```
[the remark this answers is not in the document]
jingchu000: OK

backend/cmd/headlamp.go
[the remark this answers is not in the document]
jingchu000: OK

backend/cmd/headlamp.go
```diff
@@ -148,7 +156,6 @@ func stopOrDeletePortForward(cluster string, id string, isStopRequest bool) erro
 					v.closeChan <- struct{}{}
 					clusterPortForwards[index] = v
 				}
-
```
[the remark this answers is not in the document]
jingchu000: OK

backend/cmd/headlamp.go
jingchu000: Developer tools caused, recovered

backend/cmd/headlamp.go
jingchu000: OK

jingchu000:
Which question is not still there, could you please prompt me?

jingchu000:
If it's more difficult, should I remove the README_CN?

backend/cmd/headlamp.go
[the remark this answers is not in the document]
jingchu000: Okay, it has been updated

jingchu000:
@illume  Is this okay?

illume (approved):
👍 thanks.

illume:
> If it's more difficult, should I remove the README_CN?

Good idea, thanks. We have some CN speaking collaborators, so if you want to submit a PR for the README_CN someone will get to it in time.

PR #1217, opened 2023-07-07
docs: Add section about filing security bugs

joaquimrocha:
The email is a redirect.


illume:
What do you think @blixtra ? 
Can I merge this in, and we can potentially add the github integrations later?

PR #1236, opened 2023-07-11
frontend NamespaceAutocomplete: Fix disappearing typed value

### LINKED ISSUE #1228: Namespace filter clears the input text as the user types it
A user has reported that the namespace filter (in any namespace scoped list view) has a bug where it clears the text being inputted.

### Steps to reproduce:
1. Go to a namespaced list view (e.g. Pods) 
2. Click the filter button, select the namespace text entry and start typing (e.g. `kube`)
**Expected result:** The namespaces whose names contain the text inputted are displayed in the popup; the inputted text keeps being displayed, so the user can adapt it.
**Actual result:** The namespaces matching the text do show up briefly, but then the text is cleared out, so all the namespaces available are shown again

joaquimrocha:
The AutoComplete component resets the typed (input) value when the
multiple selection is enabled, under certain conditions, so this patch
fixes that by controlling that input's value ourselves.

fixes #1228

### DIFF
--- a/frontend/src/components/common/NamespacesAutocomplete.tsx
+++ b/frontend/src/components/common/NamespacesAutocomplete.tsx
@@ -23,12 +23,27 @@ export interface PureNamespacesAutocompleteProps {
 
 export function PureNamespacesAutocomplete({
   namespaceNames,
-  onChange,
+  onChange: onChangeFromProps,
   filter,
 }: PureNamespacesAutocompleteProps) {
   const theme = useTheme();
   const { t } = useTranslation(['glossary', 'frequent']);
+  const [namespaceInput, setNamespaceInput] = React.useState<string>('');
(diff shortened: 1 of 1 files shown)


illume:
Hrmm. I can't reproduce the problem. 

Is it still there for you on main?

joaquimrocha:
I've asked @vyncent-t to check it out on his side. Thanks.

vyncent-t:
I was able to recreate the issue on the current main branch by following the process detailed in #1228 , it took the written word for about a second before attempting a filter and then wiping it. If you do not refresh and try entering the word again it will attempt and wipe after about a minute or so.

vyncent-t (approved):

illume (approved):
👍 thanks

PR #1240, opened 2023-07-13
Wide UI

joaquimrocha:
This PR increases the width of the inner container so we can pack the information given by the `-o wide` option for `kubectl`.
It also changes how the table is displayed to be a grid instead.

Known problems:
- [ ] The bar charts on the Node list view show the popup as being clipped by the cell's contents: hover on the chart to reproduce

### DIFF
--- a/frontend/src/components/App/Home/__snapshots__/index.stories.storyshot
+++ b/frontend/src/components/App/Home/__snapshots__/index.stories.storyshot
@@ -193,7 +193,7 @@ exports[`Storyshots Home/Home Base 1`] = `
               class="MuiPaper-root MuiTableContainer-root makeStyles-tableContainer MuiPaper-outlined MuiPaper-rounded"
             >
               <table
-                class="MuiTable-root makeStyles-table"
+                class="MuiTable-root makeStyles-table makeStyles-table"
               >
                 <thead
                   class="MuiTableHead-root"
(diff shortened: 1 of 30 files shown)


illume:
This breaks the mobile view when there are many columns. Before on mobile you could scroll to see them all, and now not so.
Nodes for example: http://localhost:3000/c/minikube/nodes

image

joaquimrocha:
I think I have covered the issues that existed in this PR. And opened a new one on top of it for adding more columns to each list view: #1263

illume:
Please rebase against main. I think that failure is fixed in main.

joaquimrocha:
> Please rebase against main. I think that failure is fixed in main.

It was rebased but there was an issue, which is now fixed.

illume:
Seems there's some unnecessary cut off of the last column in mobile view.

image

illume:
How about instead of cutting off the text,
set a maximum width for a column,
so that all the text can be read?

image

illume:
There's an uncessary scroll bar for the paging navigation.

image

joaquimrocha:
This is the base for the new wide UI (with more columns). I opened it as a different PR because I thought we could add this in the release (when I thought maybe we released without the extra new columns) so the table cells wouldn't wrap which makes the UI also not great when it has little horizontal space.
But maybe it makes no sense to test it now since you're testing the columns which may change with #1263 ; sorry I didn't realize this earlier.

I will close this and have it be part of the mentioned PR.

PR #1241, opened 2023-07-14
frontend: TopBar: Fix icon centering in mobile

illume:
before:
image

after:

Screenshot 2023-07-14 125330


vyncent-t:
Looks like this fixes it on my end, just tried it!

vyncent-t (approved):

PR #1242, opened 2023-07-14
frontend: TopBar: Fix mobile menu to close when popup opens

illume:
So that people can interact with the new popup. 

To test this, touch the notification icon from the mobile menu. Note how you can not scroll the list of notifications? With this fix people can scroll the list of notifications when it pops up.


vyncent-t:
The mobile menu moves and allows scrolling, looks like this works!

vyncent-t (approved):

PR #1245, opened 2023-07-17
frontend: SidebarItem: Fix focus background color

illume:
When tabbing with the keyboard through the links, it was not
obvious with color which link was focused in light mode theme.

Fixes https://github.com/headlamp-k8s/headlamp/issues/1234


vyncent-t:
looks great in light and dark mode!

vyncent-t (approved):

PR #1247, opened 2023-07-18
Reuse exec connection

ashu8912:


illume:
What's the benefit of doing this I wonder? 

How many connections are we making?

ashu8912:
This is useful in inspektor-gadget scenario where we don't want to spin up new websockets for the same URL again and again, as there is a limit to number of socket connection a browser can handle at any point.

illume:
I wonder if we need to worry about garbage collection of the connections? Adding them into a map would seem to prevent garbage collection when they aren't used anymore.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -850,11 +857,18 @@ export function stream(url: string, cb: StreamResultsCb, args: StreamArgs) {
...
     if (connection) connection.close();
     isCancelled = true;
+    OPEN_SOCKET_CONNECTIONS.delete(url);
   }
 
   function connect() {
     if (connectCb) connectCb();
-    connection = connectStream(url, cb, onFail, isJson, additionalProtocols);
+    if (OPEN_SOCKET_CONNECTIONS.has(url)) {
+      const { socket, close } = OPEN_SOCKET_CONNECTIONS.get(url)!;
```
joaquimrocha, line 866: I think this may not be enough to verify it's the same connect stream.
If I call the same URL but with a different callback, onFail, isJson, or additionalProtocols, then this code will not have the expected behavior.

Making URL+isJson+additionalProtocols the actual key for the map should be easy, but checking for the same cb+onFail may be tricky and it's not very efficient to have different connections just for different callbacks, so maybe a solution is to have a generic callback & onFail that will be in charge of calling the user set callback/onFail (which we keep track of). We must not forget to update this list of callbacks/onFails if the user has cancelled their connection.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -850,11 +857,18 @@ export function stream(url: string, cb: StreamResultsCb, args: StreamArgs) {
   function cancel() {
     if (connection) connection.close();
     isCancelled = true;
+    OPEN_SOCKET_CONNECTIONS.delete(url);
```
joaquimrocha, line 860: This means that if we have currently 2 users (places in the code, not actual different people) for this connection, and one cancels the connection, the others will have it cancelled too.

See my other comment for keeping track of the list of callbacks/onFails, and this list can act as a ref count for making sure we only effectively stop the connection when there's no list of tracked callers anymore.

joaquimrocha (changes requested):
The code needs some changes to warrant that the logic is the same as before except for effectively starting the connections.

This is some complexity that I wonder whether we actually need this, as opposed to reducing any unnecessary socket connections throughout the program.

joaquimrocha:
We changed how connections are handled and we have the multiplexer. This shouldn't be needed.

PR #1248, opened 2023-07-18
Event list optimization

ashu8912:
Acceptance criteria

- [x] Event list now fetches max events upto 2000
- [ ] Filtering for a event that doesn't belong to this 2000 events should be requested to the server

### DIFF
--- a/frontend/src/lib/k8s/event.ts
+++ b/frontend/src/lib/k8s/event.ts
@@ -23,6 +23,19 @@ export interface KubeEvent {
 class Event extends makeKubeObject<KubeEvent>('Event') {
   static apiEndpoint = apiFactoryWithNamespace('', 'v1', 'events');
 
+  // Max number of events to fetch from the API
+  private static maxEventsLimit = 2000;
+
+  // Getter to get the max number of events that are to be fetched
+  static get maxLimit() {
+    return this.maxEventsLimit;
+  }
+
+  // Setter to set the max number of events that are to be fetched
+  static set maxLimit(limit: number) {
+    this.maxEventsLimit = limit;
+  }
+
   get spec() {
     return this.getValue('spec');
   }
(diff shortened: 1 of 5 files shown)


frontend/src/lib/k8s/event.ts
```diff
@@ -69,7 +71,7 @@ class Event extends makeKubeObject<KubeEvent>('Event') {
           return `${k}=${fieldSelector[k]}`;
         })
         .join(','),
-      limit: '500',
+      limit: EVENTS_LIMIT,
```
joaquimrocha, line 74: I am not very confident with this change, as I am not sure how this can impact the details view. But it's probably fine.

frontend/src/lib/k8s/event.ts
```diff
@@ -20,6 +20,8 @@ export interface KubeEvent {
   [otherProps: string]: any;
 }
 
+export const EVENTS_LIMIT = 2000;
```
joaquimrocha, line 23: I'm not sure we should export a const like this. Maybe a static getter/setter for `maxNumEvents` in the Event class, and documented would be better? So users could also easily modify it?

joaquimrocha (changes requested):
Great job @ashu8912 . I like changes but think we can improve the EVENTS_LIMIT role/use a bit.

frontend/src/lib/k8s/event.ts
ashu8912: I think the previous limit didn't had any effect as we were not using in the api request we were making

frontend/src/lib/k8s/event.ts
ashu8912: updated

frontend/src/lib/k8s/event.ts
```diff
@@ -69,7 +82,7 @@ class Event extends makeKubeObject<KubeEvent>('Event') {
           return `${k}=${fieldSelector[k]}`;
         })
         .join(','),
-      limit: '500',
+      limit: this.maxEventsLimit,
```
joaquimrocha, line 85: This should use the getter.

joaquimrocha (changes requested):
Just one little thing to change. Then it's good.

frontend/src/lib/k8s/event.ts
ashu8912: fixed

joaquimrocha (approved):

PR #1263, opened 2023-07-20
Add more columns to list views

### LINKED ISSUE #1082: Resource list table should show columns the same as kubectl get [resource] -o wide
Resource list table should have more information, which can get from kubectl -o wide. I have listed missing columns here:

1. [Node] Role, version(kubelet, kubeproxy), Internal IP, External IP, OS-Image, Kernel Version, container runtimeImage
2. [Services] Selector, Ports
3. [Ingresses] Adress, Ports, Class
4. [Rolebinding] Role, Users, Groups, ServiceAccounts
5. [Pod] Nominated Node, Readiness Gates
6. [Replicaset] Containers, images, selector
7. [Daemonset] Containers, Images, Selector, Desired, current, ready, up-to-date, available, node-selector
8. [Statefulset] Containers, Images, 
9. [Deployment] Containers, Images, Selector, Up-to-date, Available,
10. [Cronjobs] Containers, Images, Selector, Active
11. [Jobs] Containers, Images, Selector, Duration

joaquimrocha:
This PR adds more columns to each list view in order to try to cover the kubectl get RESOURCE_TYPE -o wide  option.
Whenever it was impossible to fit visually, the least important columns have been hidden by default.

This is based on the #1240 PR.

fixes #1082


illume:
- For pods, I see "NOMINATED NODE   READINESS GATES" columns are missing from kubectl output?  `k get pods -A -o wide`
- "Node name" is just called "Node" in kubectl.
- the order of the columns is slightly different
- the values in the restart column do a ceiling rather than rounding... so 9 days displays as 2w ago. Feels misleading IMHO. The Age column does round them (37 days is show as 1 month), so there is some inconsistency with how the dates are shown.

illume:
At this browser width I can't read a lot of the data because it is cut off with "...". I mentioned in the other issue, it might be better to do a max-width. Not displaying the data breaks the view IMHO.

image

illume:
Notice how there are two dots (not three) at the end of the 2d age? Probably there should be no dots.

The other values are wider with 3 chars rather than 2 and don't show any dots.

image

joaquimrocha:
> Notice how there are two dots (not three) at the end of the 2d age? Probably there should be no dots.
> 
> The other values are wider with 3 chars rather than 2 and don't show any dots.
> 

I think the 2 dots is because the cell is being cut of. I will work next time (Monday) on increasing the minimum width of cells. However, I think ellipsizing is fine given we have the tooltip displaying the full name and it allows the user to copy that info as well.
Have wide scrollbars is not a great UX too. Ideally we'd transform this table into a grid with Key / Value vertically displayed, but that's outside the scope of this PR.

joaquimrocha:
> Whilst going to serviceaccounts page directly http://localhost:3000/c/minikube/serviceaccounts I get a blank screen. I also get an error if navigating there from another URL.

I will check what's going on next time (Monday).

joaquimrocha:
> Notice how there are two dots (not three) at the end of the 2d age? Probably there should be no dots.
> 
> The other values are wider with 3 chars rather than 2 and don't show any dots.
> 
> image

Which view is this, @illume ? I thought it was the regular pods list view but cannot reproduce it.

joaquimrocha:
> Whilst going to serviceaccounts page directly http://localhost:3000/c/minikube/serviceaccounts I get a blank screen. I also get an error if navigating there from another URL.

Should be fixed now.

> * For pods, I see "NOMINATED NODE   READINESS GATES" columns are missing from kubectl output?  `k get pods -A -o wide`

I have added these.

> * "Node name" is just called "Node" in kubectl.

I have changed it (though this was already in main, not introduced here).

> * the order of the columns is slightly different

That's okay. I am not trying to mimic the order.

> * the values in the restart column do a ceiling rather than rounding... so 9 days displays as 2w ago. Feels misleading IMHO. The Age column does round them (37 days is show as 1 month), so there is some inconsistency with how the dates are shown.

This is the case in main, not introduced here. I will open a separate PR.

illume:
Seems there's some conflicts that need to be resolved.

illume:
From storybook cluster/Overview:
Seems like when you shrink the browser window to be less wide there's some overlap of the columns...
image

illume:
I can't read the notifications message.
(happens on main branch too)

image

illume:
The headings "Namespace" and "Readiness Gates" are cut off.

image

illume:
At full browser width I still get a scroll bar on pods, but it seems there should be enough room.

image

joaquimrocha:
> At full browser width I still get a scroll bar on pods, but it seems there should be enough room.

I don't notice this at a 1920x1080 resolution.
I have now pushed a change to wrap words at any character, making words break at any char but which allows them to be read.

illume:
> I don't notice this at a 1920x1080 resolution.
> I have now pushed a change to wrap words at any character, making words break at any char but which allows them to be read.

I still see a scrollbar. At 1612 browser width.

joaquimrocha:
> > I don't notice this at a 1920x1080 resolution.
> > I have now pushed a change to wrap words at any character, making words break at any char but which allows them to be read.
> 
> I still see a scrollbar. At 1612 browser width.

At some point we can expect users to see a scroll bar and we should add for a minimum viewport width from which users shouldn't see a scrollbar. Hopefully it's better now.

illume:
Cool. The table scrollbar isn't there in those cases now.

illume:
At around 1300 width there is a window scrollbar and the table controls are not visible.

image

joaquimrocha:
> At around 1300 width there is a window scrollbar and the table controls are not visible.

This is not caused by the changes to the table, rather made more evident by it. Fixing it requires changing how the app container boxes are set up, so it should be outside of the scope of this PR.

illume:
Probably having it like the smaller view would be good. 
So the controls are visible, but the data is scrolled inside the table view.

I'm happy to move this issue, and other ones out of this PR? So we have more of a chance for other people to play with what is there. Then perhaps it can be fixed Monday, before the release, but separately to this PR. What do you think?

illume (approved):
👍 thanks

joaquimrocha:
> I'm happy to move this issue, and other ones out of this PR? So we have more of a chance for other people to play with what is there. Then perhaps it can be fixed Monday, before the release, but separately to this PR. What do you think?

Good idea. If you can work on this, that'd be great, thanks! I will merge the PR.

PR #1264, opened 2023-07-20
Misc fixes to k8s classes to the ResourceClasses object and new tests

joaquimrocha:
This PR adds a few fixes to the k8s classes we have as ResourceClasses in Headlamp.
It also adds a new test to prevent related regressions. This new test checks whether certain classes 1. have the listRoute correctly set, 2. are namespaced or not; and 3. are being tested or not.

How to test:
- [ ] Create resources of the LimitRange, Lease, PriorityClass, and RuntimeClass (add the YAML to the creation dialog after pressing Create): verify that they can be created
- [ ] Run the k8s/index.test.ts test (make frontend-test)

### DIFF
--- a/frontend/src/components/priorityClass/Details.tsx
+++ b/frontend/src/components/priorityClass/Details.tsx
@@ -1,6 +1,6 @@
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
-import PriorityClasses from '../../lib/k8s/priorityClasses';
+import PriorityClass from '../../lib/k8s/priorityClass';
 import { DetailsGrid, ObjectEventList } from '../common';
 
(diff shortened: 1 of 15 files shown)


vyncent-t:
Tried with the following templates and then changing some fields but was unable to add it

image

image

joaquimrocha:
@vyncent-t Good catch! It was indeed missing a commit, and in the meanwhile I added more tests and fixes. Thanks.
I have updated the description for extra checks.

illume:
@vyncent-t can you please review this again?

vyncent-t:
Just tried again with the templates from the screen shots and it works!

illume:
Thanks @vyncent-t

PR #1275, opened 2023-07-26
frontend Sidebar: Add a different color to selected sub-entries

joaquimrocha:
This is in order for the currently selected entry to be more actively distinguishable.

Old/current:

old screenshot

New:
new screenshot


illume (approved):
👍 ah yes!

PR #1277, opened 2023-07-26
frontend: Creates tests for icons

### LINKED ISSUE #1109: Create test for checking for missing local icons
Recently (#1039) we added support for shipped icons in Headlamp so they work in air gapped envs.
However, since this requires keeping a list of icons, we should have a test in place for checking for used icons that are missing from the cache files. I imagine this test needs to look for (roughly) `icon="*"` and then verify whether the wildcard match corresponds to an icon in the cache files; otherwise, report an error.

vyncent-t:
This PR improves Headlamp's offline mode support by implementing a Jest test to analyze used and shipped icons. This improvement enables the application icons to function in air gapped environments with limited external resource access.

The Jest test ensures that all necessary icons are bundled and available locally within the application. With the required icons readily accessible, users can experience a better offline usage.

Done
- [x] Created logic to compare used icons to cached icons

Next
- [ ] Create better error functionality (currently only displays message)

fixes #1109


frontend/src/components/App/icons.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        return false;
+      }
+    }
+  }
+};
+
+describe('Icon tests', () => {
+  test('Check icons', async () => {
+    const result = await checkIcons();
+    expect(result).toBe(true);
```
joaquimrocha, line 38: The logic for testing is correct. This is exactly what we talked about.
It can be improved though, to make the tests more understandable when they fail, by having the function for checking icons create a test (`test('Check icon ${match[0]}', async () => { expect(usedIcons.includes(matches[0])).toBe(true) }`) for each icon that the regex matches. So when the test fails, we see exactly which icon failed.

joaquimrocha (approved):

frontend/src/components/App/icons.test.ts
illume: It seems to print the failed icon near the test failure...

From the test failure:
> Icon mdi:more-vert is not used in ./src/components/App/Home/index.tsx

But maybe that message could be improved by telling people what they need do?

illume:
This doesn't seem to take into account aliases?

For example, in frontend/src/components/App/icons.ts there's a `more-vert` which is an alias for `dots-vertical`.

vyncent-t:
> This doesn't seem to take into account aliases?
> 
> For example, in frontend/src/components/App/icons.ts there's a `more-vert` which is an alias for `dots-vertical`.

Great point, thanks for mentioning it I missed it! Just put in a fix that parses the alias list too

vyncent-t:
WIP push
The last push did not work with alias or lengthy names, working on a better fix to locate every icon

joaquimrocha (changes requested):

vyncent-t:
Included changes mentioned!

vyncent-t:
I manually added the 35/36 missing icons that the test prints to the icons.ts file, with the last being a potential fix

Noticed this error

image


Which seems to be linked to this bit, currently not sure if this is meant to be this way or if I should fix the spelling

image
image


Running into a small blocker from the client when trying to start a job in suspended, prevents me from viewing in browser so it could be working as intended.

There is also no icon in the iconify catalog for "mdi:pause-icon" or "mdi:pauseIcon" there is an icon for the default pause icon if I should replace it (source https://icon-sets.iconify.design/mdi/pause/)

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "devDependencies": {
+    "@types/jest": "^29.5.3"
+  },
+  "dependencies": {
+    "glob": "^10.3.3"
```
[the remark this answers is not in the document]
vyncent-t: done!

joaquimrocha:
Good job. I'll add a test for the job and then we can try it.

vyncent-t:
I tried this template for suspended jobs and it works, thank you!

source (https://kubernetes.io/blog/2021/04/12/introducing-suspended-jobs/)

image

frontend/src/components/App/icons.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  const unusedIcons = [];
+
+  for (const file of files) {
+    const content = await fs.readFileSync(file, 'utf8');
+    // const regex = /mdi:\w+/g;
+    // \w+ matches any word character (equal to [a-zA-Z0-9_])
+    // + Quantifier — Matches between one and unlimited times, as many times as possible, giving back as needed (greedy)
+    // g modifier: global. All matches (don't return after first match)
+    // this will only read for words, not for words with a dash in it
+
```
joaquimrocha, line 23: Please remove this comment.

frontend/src/components/App/icons.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    // this regex will match all mdi:icon-name
+    // const regex = /mdi:[\w-]+/g;
+    // mdi: matches the characters mdi: literally (case sensitive)
+    // Match a single character present in the list below [\w-]+
+    // + Quantifier — Matches between one and unlimited times, as many times as possible, giving back as needed (greedy)
+    // \w matches any word character (equal to [a-zA-Z0-9_])
+    // - matches the character - literally (case sensitive)
+    // g modifier: global. All matches (don't return after first match)
+    // const regex = /mdi:[\w-]+/g;
```
joaquimrocha, line 32: Remove the commented out code (`const regex = ...`).
Also I think just having a comment explaining what exactly this matches, is fine.
Something like `Find all matches of a word starting with 'mdi:'`.

frontend/src/components/App/icons.test.ts
```diff
@@ -0,0 +1,61 @@
...
+          icon: match[0],
+          file,
+        });
+      }
+    }
+  }
+
+  for (let i = 0; i < unusedIcons.length; i++) {
+    error(
+      `Icon ${unusedIcons[i].icon} from file ${unusedIcons[i].file} is not cached for offline use `
```
joaquimrocha, line 49: Nit: Showing the line number would also be cool `file-name:line-number`, but if it's not directly given by the regex.exec, let's not complicate the more logic just for that.

joaquimrocha (changes requested):
Just noticed some duplicated comments, so left a suggestion.

frontend/src/components/App/icons.test.ts
vyncent-t: Removed!

frontend/src/components/App/icons.test.ts
vyncent-t: Edited to be one line of summary for the regex

frontend/src/components/App/icons.test.ts
vyncent-t: Interesting! I will push the notes edits and will try this if possible I'll push it next!

vyncent-t: Tried to find an easy way to implement this friday but could but was not forward enough to put in

joaquimrocha (approved):

PR #1279, opened 2023-07-27
frontend plugin: Check plugins' compatibility with the plugin engine

### LINKED ISSUE #423: Create a compatibility validation for plugins
## Current situation

So far, we left the responsibility of ensuring Headlamp's plugins are compatible with the Headlamp core to whoever is running/deploying the plugins.
However, if we are to make the installation of plugins more dynamic, we need to ensure that the compatibility of plugins vs Headlamp is clear.

## Ideal future situation

When loading a plugin, Headlamp can check whether the plugin is compatible with it and warn the user otherwise, instead of trying to load the plugin straight away.

## Implementation options

There are several ways we can do this. While Headlamp is in 0.x version, maybe relying on the Headlamp version itself for comparison with a plugins' inner compat version is not a good idea.
So maybe we can have a plugin-engine version associated with Headlamp, as this allows us to decouple the Headlamp version from the plugin engine which can be useful in other cases too.

- [ ] Deny loading incompatible plugins: #1279 
- [ ] Have UI showing the user that some plugins are incompatible

joaquimrocha:
This patch adds a way for checking for plugins' compatibility with the plugin engine the current Headlamp build uses, and uses the new functions for filtering plugins when loading them initially.

Tests for the compatibility functions are also added.

How to test:
- [ ] Verify that (without changes), plugins are loaded
- [ ] Change the `compatibleHeadlampPluginVersion` to something like `0.0.1`, then load headlamp with plugins: verify that plugins are not loaded
- [ ] Change the mentioned version for `>=0.0.1`, realize the plugins are loaded

fixes #423

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -4,10 +4,18 @@
  * The lib.ts file should carry the bits to be used by plugins whereas
  */
 
+import semver from 'semver';
 import helpers from '../helpers';
 import { Headlamp, Plugin } from './lib';
 import { PluginInfo } from './pluginsSlice';
 import Registry, * as registryToExport from './registry';
+
+/** The version of the plugin engine for which this headlamp build is compatible.
+ * If the plugin engine version is not compatible, the plugin will not be loaded.
+ * Can be set to a semver range, e.g. '>= 0.6.0' or '0.6.0 - 0.7.0'.
+ * If set to an empty string, all plugin versions will be loaded.
+ */
+const compatibleHeadlampPluginVersion = '';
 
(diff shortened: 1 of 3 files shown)


illume:
I'm not sure about this one. It feels more complicated than comparing against @kinvolk/headlamp-plugin version.

joaquimrocha:
> I'm not sure about this one. It feels more complicated than comparing against @kinvolk/headlamp-plugin version.

Can you elaborate on that?

joaquimrocha:
Closed in favor of the changes in #668

Issue #1266, opened 2023-07-23
Setting baseURL breaks liveness/readiness probes


celcion:
Hello!

Sorry if it was already mentioned, but it looks like setting baseURL changes the base path globally which makes liveness and readiness probes (that statically configured to "/" in helm chart) to return 404 and cause pod to crashloop.

Chart version: 0.13.0
Image version: 0.18.0

My values.yaml
```
          config:
            oidc:
              clientID: "myid"
              clientSecret: "mysecret"
              issuerURL: "https://oidc-issuer.local"
              scopes: "profile,email,groups"
            baseURL: /headlamp
          ingress:
            enabled: true
            annotations:
              kubernetes.io/ingress.class: nginx
            hosts:
            - host: my-host.local
              paths:
              - type: Prefix
                path: /headlamp
            tls:
            - secretName: kubernetes-dashboard-tls
              hosts:
              - my-host.local
```
Pod events:
```
11m         Normal    Pulling            pod/headlamp-57c9c965c-xzgtg   Pulling image "ghcr.io/headlamp-k8s/headlamp:v0.18.0"
11m         Normal    Pulled             pod/headlamp-57c9c965c-xzgtg   Successfully pulled image "ghcr.io/headlamp-k8s/headlamp:v0.18.0" in 6.443236552s (13.401954993s including waiting)
10m         Normal    Created            pod/headlamp-57c9c965c-xzgtg   Created container headlamp
10m         Normal    Started            pod/headlamp-57c9c965c-xzgtg   Started container headlamp
10m         Warning   Unhealthy          pod/headlamp-57c9c965c-xzgtg   Readiness probe failed: HTTP probe failed with statuscode: 404
10m         Warning   Unhealthy          pod/headlamp-57c9c965c-xzgtg   Liveness probe failed: HTTP probe failed with statuscode: 404
10m         Normal    Killing            pod/headlamp-57c9c965c-xzgtg   Container headlamp failed liveness probe, will be restarted
10m         Normal    Pulled             pod/headlamp-57c9c965c-xzgtg   Container image "ghcr.io/headlamp-k8s/headlamp:v0.18.0" already present on machine
6m19s       Warning   BackOff            pod/headlamp-57c9c965c-xzgtg   Back-off restarting failed container headlamp in pod headlamp-57c9c965c-xzgtg_infra(153c875b-d03f-4e6d-b4c5-0fd44e07ded1)
```

joaquimrocha:
[...]

PR #1280, opened 2023-07-31
charts: Fix liveness and readiness probes

knrt10:
Previously when user changed baseURL for the application it was not reflecting with liveness and readiness probes. Due to which the application was crashing.

Fixes: #1266


illume (approved):
👍 thanks

Issue #1268, opened 2023-07-24
enhancement: make host in ingress rules clickable


smuu:
Make the host in the ingress rules a clickable link to open the endpoint without the need to copy & paste the URL.

![image](https://github.com/headlamp-k8s/headlamp/assets/18609909/9983d17b-438a-4fd6-b185-76f0c49753d9)

PR #1283, opened 2023-07-31
frontend: Create nav link for ingress

vyncent-t:
Related to issue #1268 

## Description
This PR addresses the enhancement requested in issue #1268 . The purpose was to transform the host in the ingress rules into a clickable link, thereby eliminating the need for users to manually copy and paste the URL to open the endpoint.

## Changes
image

- [x]  Functional links within the field Host
- [x]  Additional links for path field for specific routing
- [x] Created logic for determining url protocols based on ingress tls, defaults to http when not using tls

## Notes
- To validate this change, I created example pod, service, and host with the host set to bing. The changes work for formatted URL stings only, but will print '*' if needed and will not print any links if no host is written.
- When using tls and host is within fields, protocol is set to https, if not using tls or if it is not found, defaults to http

Using tls
image

Not using tls
image

### DIFF
--- a/frontend/src/components/ingress/Details.tsx
+++ b/frontend/src/components/ingress/Details.tsx
@@ -1,11 +1,156 @@
+import { Box, Link, Typography } from '@material-ui/core';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import Ingress, { IngressRule } from '../../lib/k8s/ingress';
+import { IngressBackend } from '../../lib/k8s/ingress';
 import { useSettings } from '../App/Settings/hook';
 import LabelListItem from '../common/LabelListItem';
 import { DetailsGrid } from '../common/Resource';
 import { SectionBox } from '../common/SectionBox';
 import SimpleTable from '../common/SimpleTable';
+
+/**
+ * Is https used in the ingress item
+ */
+function isHttpsUsed(item: Ingress, url: String) {
+  const hostList: string[] = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`);
+  const isHttps = hostList.includes(`${url}`);
+
+  return isHttps;
+}
+
+export interface LinkStringFormatProps {
+  url: string;
+  item: Ingress;
+  urlPath?: string;
+}
+
+/**
+ * Format the url to be used in the Link component
+ */
+export function LinkStringFormat({ url, item, urlPath }: LinkStringFormatProps) {
+  let urlProtocol;
+  let formatURL;
+
+  /**
+   * Renders the "host" field as we intentionally do not supply a path.
+   * If the url is set to * or undefined/empty, and the optional path is not used, print url
+   */
+  if ((url === '*' || !url) && !urlPath) {
+    return <Typography>{url || ''}</Typography>;
+  }
+
+  /**
+   * Renders the "path" field as we intentionally supply both a path and url.
+   */
+  if ((url === '*' || !url) && urlPath) {
+    return <Typography>{urlPath}</Typography>;
+  } else {
+    /*
+     * If the ingress does not use tls, then the url should be http
+     */
+    if (!item.jsonData.spec?.tls) {
+      urlProtocol = 'http://';
+    } else {
+      /*
+       * If the url is in the tls array, then the url should be https
+       */
+      isHttpsUsed(item, url) ? (urlProtocol = 'https://') : (urlProtocol = 'http://');
+    }
+
+    /*
+     * Since we cannot access the prefix from the ingress object, we have to access it from the rules array
+     */
+    const rules: any[] = item.spec.rules;
+    let currentPathType;
+    if (rules) {
+      for (let i = 0; i < rules.length; i++) {
+        for (let j = 0; j < rules[i].http.paths.length; j++) {
+          if (rules[i].http.paths[j].path === urlPath) {
+            currentPathType = rules[i].http.paths[j].pathType;
+          }
+        }
+      }
+    }
+
+    function isValidURL(url: string): boolean {
+      try {
+        new URL(url);
+      } catch (e) {
+        return false;
+      }
+      return true;
+    }
+
+    /**
+     * Adding the protocol to the url string is required for the URL constructor to work because data.host is only the host name
+     */
+    if (!isValidURL(urlProtocol + url)) {
+      console.debug(`Ingress' host ${url} is not a valid URL; displaying it without a link.`);
+      return <Typography>{url}</Typography>;
+    }
+
+    formatURL = new URL(urlProtocol + url);
+
+    if (formatURL && urlPath) {
+      return (
+        <Box style={{ display: 'flex', marginBottom: '5px' }}>
+          <Link
+            href={`${formatURL.protocol}//${formatURL.hostname}${urlPath}`}
+            style={{ marginRight: '5px' }}
+          >
+            {urlPath}
+          </Link>
+          {`(${currentPathType})`}
+        </Box>
+      );
+    }
+
+    return <Link href={formatURL.toString()}>{`${formatURL.toString()}`}</Link>;
+  }
+}
+
+export interface BackendFormatProps {
+  backend: IngressBackend[];
+}
+
+/**
+ * Used to format the backend to be used in the backends field
+ */
+
+export function BackendFormat({ backend }: BackendFormatProps) {
+  const backendList: string[] = [];
+  for (const backendItem of backend) {
+    let currentBackend: string = '';
+
+    if (!!backendItem.resource) {
+      currentBackend = backendItem.resource.kind + ':' + backendItem.resource.name;
+    }
+    if (!!backendItem.service) {
+      currentBackend =
+        backendItem.service.name +
+        ':' +
+        (backendItem.service.port.number ?? backendItem.service.port.name ?? '').toString();
+    }
+
+    backendList.push(currentBackend);
+  }
+
+  return (
+    <Box
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        flexWrap: 'nowrap',
+        marginBottom: '5px',
+      }}
+    >
+      {backendList.map(backendItem => (
+        <Typography>{backendItem}</Typography>
+      ))}
+    </Box>
+  );
+}
 
 export default function IngressDetails() {
   const { namespace, name } = useParams<{ namespace: string; name: string }>();
@@ -80,33 +225,21 @@ export default function IngressDetails() {
               columns={[
                 {
                   label: t('Host'),
-                  getter: (data: IngressRule) => data.host || '*',
+                  getter: (data: IngressRule) => (
+                    <LinkStringFormat url={data.host || '*'} item={item} />
+                  ),
                 },
                 {
                   label: t('Path'),
-                  getter: (data: IngressRule) => data.http.paths.map(({ path }) => path).join(', '),
+                  getter: (data: IngressRule) =>
+                    data.http.paths.map(({ path }) => (
+                      <LinkStringFormat url={data.host || '*'} item={item} urlPath={path} />
+                    )),
                 },
                 {
                   label: t('Backends'),
                   getter: (data: IngressRule) => (
-                    <LabelListItem
-                      labels={data.http.paths.map(({ backend }) => {
-                        if (!!backend.resource) {
-                          return backend.resource.kind + ':' + backend.resource.name;
-                        }
-                        if (!!backend.service) {
-                          return (
-                            backend.service.name +
-                            ':' +
-                            (
-                              backend.service.port.number ??
-                              backend.service.port.name ??
-                              ''
-                            ).toString()
-                          );
-                        }
-                      })}
-                    />
+                    <BackendFormat backend={data.http.paths.map(({ backend }) => backend)} />
                   ),
                 },
               ]}
--- a/frontend/src/lib/k8s/ingress.ts
+++ b/frontend/src/lib/k8s/ingress.ts
@@ -16,6 +16,7 @@ export interface IngressRule {
   http: {
     paths: {
       path: string;
+      pathType?: string;
       backend: IngressBackend;
     }[];
   };
@@ -26,7 +27,7 @@ interface LegacyIngressBackend {
   servicePort: string;
 }
 
-interface IngressBackend {
+export interface IngressBackend {
   service?: {
     name: string;
     port: {
@@ -43,7 +44,7 @@ interface IngressBackend {
 
 export interface KubeIngress extends KubeObjectInterface {
   spec: {
-    ingressClassName: string;
+    ingressClassName?: string;
     rules: IngressRule[] | LegacyIngressRule[];
     tls?: {
       hosts: string[];
--- a/frontend/src/components/ingress/LinkStringFormat.stories.tsx
+++ b/frontend/src/components/ingress/LinkStringFormat.stories.tsx
@@ -1,0 +1,156 @@
+import { Meta, Story } from '@storybook/react/types-6-0';
+import React from 'react';
+import Ingress from '../../lib/k8s/ingress';
+import { LinkStringFormat, LinkStringFormatProps } from './Details';
+
+export default {
+  title: 'ingress/LinkStringFormat',
+  component: LinkStringFormat,
+  argTypes: {},
+} as Meta;
+
+const Template: Story<LinkStringFormatProps> = args => <LinkStringFormat {...args} />;
+
+const noPath = new Ingress({
+  kind: 'Ingress',
+  metadata: {
+    name: 'test',
+    namespace: 'test',
+    uid: '',
+    creationTimestamp: '',
+  },
+  spec: {
+    rules: [
+      {
+        host: 'examplehost.com',
+        http: {
+          paths: [
+            {
+              path: '/',
+              pathType: 'Prefix',
+              backend: {
+                service: {
+                  name: 'test',
+                  port: {
+                    number: 80,
+                  },
+                },
+              },
+            },
+          ],
+        },
+      },
+    ],
+  },
+});
+
+const onePath = new Ingress({
+  kind: 'Ingress',
+  metadata: {
+    name: 'test',
+    namespace: 'test',
+    uid: '',
+    creationTimestamp: '',
+  },
+  spec: {
+    rules: [
+      {
+        host: 'examplehost.com',
+        http: {
+          paths: [
+            {
+              path: '/pathA',
+              pathType: 'Prefix',
+              backend: {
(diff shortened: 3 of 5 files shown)


frontend/src/components/common/LabelListItem.tsx
```diff
@@ -6,7 +6,7 @@ export interface LabelListItemProps {
 }
 
 export default function LabelListItem(props: LabelListItemProps) {
-  const { labels } = props;
+  const { labels = [] } = props;
```
joaquimrocha, line 9: This should be in a separate commit as it's an unrelated fix.

frontend/src/components/ingress/Details.tsx
```diff
@@ -74,7 +75,8 @@ export default function IngressDetails() {
               columns={[
                 {
                   label: t('Host'),
-                  getter: (data: IngressRule) => data.host || '*',
+                  getter: (data: IngressRule) =>
+                    <Link routeName={`${data.host}`}>{data.host}</Link> || '*',
```
joaquimrocha, line 79: The routeName is supposed to be a route's name from the object in router.tsx.
Also you need to validate whether the host is an actual URL before adding it as a link prop.

frontend/src/lib/k8s/ingress.ts
```diff
@@ -84,7 +84,7 @@ class Ingress extends makeKubeObject<KubeIngress>('ingress') {
 
     const rules: IngressRule[] = [];
 
-    this.spec!.rules.forEach(({ http, host }) => {
+    this.spec!.rules?.forEach(({ http, host }) => {
```
joaquimrocha, line 87: This needs to be a separate commit.

joaquimrocha (changes requested):

frontend/src/components/common/LabelListItem.tsx
vyncent-t: Put it in the fix branch!

frontend/src/lib/k8s/ingress.ts
vyncent-t: Put it in the fix branch!

frontend/src/components/ingress/Details.tsx
```diff
@@ -36,6 +37,29 @@ export default function IngressDetails() {
...
+      const formatURL = new URL(`http://${url}`);
+      if (formatURL) {
+        return (
+          <Link
+            href={`${formatURL.protocol}//${formatURL.hostname}`}
+          >{`${formatURL.protocol}//${formatURL.hostname}`}</Link>
+        );
+      }
+    } catch (e) {
+      throw new Error('Invalid URL');
```
joaquimrocha, line 59: I think we should log the occurrence, though here it's just for debug purposes (console.debug("Ingress' host is not a valid URL; displaying it without a link.')) as it may fail for some reason. And it should return just the string IMO.

frontend/src/components/ingress/Details.tsx
```diff
@@ -74,7 +98,7 @@ export default function IngressDetails() {
               columns={[
                 {
                   label: t('Host'),
-                  getter: (data: IngressRule) => data.host || '*',
+                  getter: (data: IngressRule) => isFormatURL(data.host || '*'),
```
joaquimrocha, line 101: Should we include the path too? There's something to be decided about that here.

frontend/src/components/ingress/Details.tsx
```diff
@@ -36,6 +37,29 @@ export default function IngressDetails() {
...
+  /** this function is used to format the url to be used in the Link component */
+  function isFormatURL(url: any) {
+    if (url === '*') {
+      return '*';
+    }
+    try {
+      /** https://developer.mozilla.org/en-US/docs/Web/API/URL/URL
+       * new URL() constructor creates a new URL object instance only if the string passed is a valid URL
+       * if the string is not a valid URL, then an error is thrown
+       * note: adding the protocol to the url string is required for the URL constructor to work because data.host is only the host name */
```
joaquimrocha, line 49: Except for the last line (without the note prefix), I don't think the comment is needed as it's kind of the API doc.

frontend/src/components/ingress/Details.tsx
```diff
@@ -36,6 +37,29 @@ export default function IngressDetails() {
     );
   }
 
+  /** this function is used to format the url to be used in the Link component */
+  function isFormatURL(url: any) {
+    if (url === '*') {
+      return '*';
+    }
```
joaquimrocha, line 44: This could be removed since we'd catch it with the exception anyway, but exceptions should be avoided, so it's good! 👍

frontend/src/components/ingress/Details.tsx
```diff
@@ -36,6 +37,29 @@ export default function IngressDetails() {
     );
   }
 
+  /** this function is used to format the url to be used in the Link component */
+  function isFormatURL(url: any) {
```
joaquimrocha, line 41: The name of the function seems like it returns a boolean, but it returns a component. So it should be called something like LinkOrString or something like that.

joaquimrocha (changes requested):
Needs a couple of fixes. Also, Ingresses may have an https link if they have TLS set up. So we should likely add that too. Ping me if you need more details.

frontend/src/components/ingress/Details.tsx
vyncent-t: Should we add the same path as the field below this?

illume:
https://kubernetes.io/docs/concepts/services-networking/ingress/

frontend/src/components/ingress/Details.tsx
illume: I think you'd almost always want the path in the link. It doesn't seem to me very common that someone would want to visit just the host part.

illume:
Missing from here is the 'prefix' information for the path. We should display this, because it's quite important... since it can change the meaning of it quite a lot.

frontend/src/components/ingress/Details.tsx
vyncent-t: Fixed it!

frontend/src/components/ingress/Details.tsx
vyncent-t: Fixed it!

vyncent-t:
> Not exactly in scope for this... but...
> 
> Missing from the ingress info is the 'prefix' information for the path. We should display this, because it's quite important... since it can change the meaning of it quite a lot.

Made a field for it!

vyncent-t:
> @vyncent-t as you suggested... How does wrapping work for the link? A good test for the storybook might be a very long host, and a very long path.

frontend/src/components/ingress/Details.tsx
vyncent-t: There is a link for the host field and for the path field!

frontend/src/components/ingress/Details.tsx
```diff
@@ -36,6 +37,24 @@ export default function IngressDetails() {
     );
   }
 
+  /** this function is used to format the url to be used in the Link component */
+  function LinkStringFormat(url: any) {
+    try {
+      /** adding the protocol to the url string is required for the URL constructor to work because data.host is only the host name */
+      const formatURL = new URL(`http://${url}`);
```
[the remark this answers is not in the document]
vyncent-t: Done!

vyncent-t:
May need to see if multiple paths break any links... will look into this

vyncent-t:
Can anyone provide context for this build error?
image

vyncent-t:
Is there a way to fix this if I do not use mac?

vyncent-t:
> May need to see if multiple paths break any links... will look into this

Fixed this, links will now print in a list.
image

illume:
It would be nice to have a story for the states as well.

vyncent-t:
WIP: this commit creates a working story using json dummydata

vyncent-t:
Deleted old story file that was left in

vyncent-t:
WIP
- Changed the backend field to list view
- Removed host from path link, links in path only render path strings but direct to full link
- Added context to path links

Need to remove comments and console logs before review

image

vyncent-t:
Commit fixes issue with rendering links that are either "*" or ""

vyncent-t:
Commit fixes issue with rendering links that are either "*" or ""

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+  /** this function is used to determine the protocol used in the formatURL for LinkedStringFormat */
+  function getProtocol(item: any, url: any) {
+    const hostList: string[] = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`);
+    const isHttps = hostList.includes(`${url}`);
+
+    return isHttps;
+  }
+
+  /** this function is used to format the url to be used in the Link component */
+  function LinkStringFormat(url: any, item?: any, urlPath?: any) {
```
joaquimrocha, line 55: The item is always used, as far as I can see, so it shouldn't be optional.
And the types should be precise.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
     );
   }
 
+  /** this function is used to determine the protocol used in the formatURL for LinkedStringFormat */
+  function getProtocol(item: any, url: any) {
```
joaquimrocha, line 47: This function is called `getProtocol` which suggests it returns the protocol, but it returns whether https is used.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+    if ((url === '*' || url === '' || url === undefined) && urlPath) {
+      return <Typography>{urlPath}</Typography>;
+    } else {
+      /* if the ingress does not use tls, then the url should be http */
+      if (!item.jsonData.spec?.tls) {
+        urlProtocol = 'http://';
+      } else {
+        /* if the ingress does use tls, then we compare the url to the hosts in the tls array to determine if the url should be http or https */
+        getProtocol(item, url) ? (urlProtocol = 'https://') : (urlProtocol = 'http://');
+      }
```
joaquimrocha, line 71: This logic can be re-written much more clearly if the getProtocol function returns the actual protocol.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+  /** this function is used to format the url to be used in the Link component */
+  function LinkStringFormat(url: any, item?: any, urlPath?: any) {
+    let urlProtocol;
+
+    if ((url === '*' || url === '' || url === undefined) && !urlPath) {
+      return <Typography>{url}</Typography>;
+    }
+
+    if ((url === '*' || url === '' || url === undefined) && urlPath) {
+      return <Typography>{urlPath}</Typography>;
```
joaquimrocha, line 63: Can't this be rewritten as:
```typescript
if (!url || url === '*') {
  return <Typography>{urlPath || url}</Typography>;
}
```
If url is undefined, we probably want to use it as `''` instead, so:
```typescript
if (!url || url === '*') {
  return <Typography>{urlPath || url || ''}</Typography>;
}
```

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+      /* if the ingress does not use tls, then the url should be http */
+      if (!item.jsonData.spec?.tls) {
+        urlProtocol = 'http://';
+      } else {
+        /* if the ingress does use tls, then we compare the url to the hosts in the tls array to determine if the url should be http or https */
+        getProtocol(item, url) ? (urlProtocol = 'https://') : (urlProtocol = 'http://');
+      }
+
+      /* since we cannot access the prefix from the ingress object, we have to access it from the itemData array */
+      const itemData: any[] = item.jsonData.spec?.rules;
```
joaquimrocha, line 74: You can actually use `item.spec.rules`. Because there's a spec getter in the class.
And the itemDatas should be called rules IMO.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+      let currentPathType;
+      if (itemData) {
+        itemData.forEach((rule: any) => {
+          rule.http.paths.forEach((path: any) => {
+            if (path.path === urlPath) {
+              currentPathType = path.pathType;
+            }
+          });
+        });
```
joaquimrocha, line 83: I think using `for` loops here would be more efficient in not worse case scenarios, as you can break the loop once the path is found.

joaquimrocha (dismissed):
Left a few comments in the code. Overall it's functionally fine but we can simplify the code a good deal.

frontend/src/components/ingress/Details.tsx
vyncent-t: I will change the name to reflect it being a Boolean return since it does not return a protocol object or string

frontend/src/components/ingress/Details.tsx
vyncent-t: I added notes to the relevant pieces for this to explain why one renders the host field and the other renders the path

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+        }
+      } catch (e) {
+        console.debug("Ingress' host is not a valid URL; displaying it without a link.");
+        return url;
+      }
+    }
+  }
+
+  function backendFormat(backend: any) {
+    const backendList: any[] = [];
```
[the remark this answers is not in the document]
vyncent-t: Ah right! I had reformatted the previous backend print logic to work without the label hover

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,110 @@ export default function IngressDetails() {
...
+        itemData.forEach((rule: any) => {
+          rule.http.paths.forEach((path: any) => {
+            if (path.path === urlPath) {
+              currentPathType = path.pathType;
+            }
+          });
+        });
+      }
+
+      try {
```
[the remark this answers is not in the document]
vyncent-t: Changed to be based on if else on the new URL creating a valid url or undefined

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,113 @@ export default function IngressDetails() {
...
+  /** this function is used to determine the protocol used in the formatURL for LinkedStringFormat */
+  function isHttpsUsed(item: Ingress, url: String) {
+    const hostList: string[] = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`);
+    const isHttps = hostList.includes(`${url}`);
+
+    return isHttps;
+  }
+
+  /** this function is used to format the url to be used in the Link component */
+  function LinkStringFormat(url: String, item: Ingress, urlPath?: String) {
```
illume, line 55: Components should take a `props`.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,113 @@ export default function IngressDetails() {
     );
   }
 
+  /** this function is used to determine the protocol used in the formatURL for LinkedStringFormat */
```
illume, line 46: Other comments use a different format and style.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,113 @@ export default function IngressDetails() {
...
+  function isHttpsUsed(item: Ingress, url: String) {
+    const hostList: string[] = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`);
+    const isHttps = hostList.includes(`${url}`);
+
+    return isHttps;
+  }
+
+  /** this function is used to format the url to be used in the Link component */
```
illume, line 54: Other functions use a different style of commenting.

frontend/src/components/ingress/Details.tsx
```diff
@@ -42,6 +43,113 @@ export default function IngressDetails() {
...
+            {`(${currentPathType})`}
+          </Box>
+        );
+      } else if (formatURL) {
+        return <Link href={formatURL.toString()}>{`${formatURL.toString()}`}</Link>;
+      }
+    }
+  }
+
+  function backendFormat(backend: any) {
```
illume, line 119: This one needs a comment.

illume (changes requested):
Left a few notes.

frontend/src/components/ingress/Details.tsx
illume, line 55: How about adding a story for LinkStringFormat? You could add the different states, which would be a nice test.

frontend/src/components/ingress/Details.tsx
illume, line 119: This should be a React component (with a props argument).

vyncent-t:
added stories for refactor

frontend/src/components/ingress/Details.tsx
vyncent-t: added!

frontend/src/components/ingress/Details.tsx
vyncent-t: done!

vyncent-t:
Fixed multiple path story

frontend/src/components/ingress/Details.tsx
```diff
@@ -1,12 +1,157 @@
...
 import Ingress, { IngressRule } from '../../lib/k8s/ingress';
+import { IngressBackend } from '../../lib/k8s/ingress';
 import { useSettings } from '../App/Settings/hook';
 import LabelListItem from '../common/LabelListItem';
 import { DetailsGrid } from '../common/Resource';
 import { SectionBox } from '../common/SectionBox';
 import SimpleTable from '../common/SimpleTable';
 
+/**
+ * This function is used to determine the protocol used in the formatURL for LinkedStringFormat
```
illume, line 13: How about this?

```suggestion
Is https used in the ingress item?
```

"This function is used to" is a bit redundant.

frontend/src/components/ingress/Details.tsx
```diff
@@ -1,12 +1,157 @@
...
+}
+
+export interface LinkStringFormatProps {
+  url: string;
+  item: Ingress;
+  urlPath?: string;
+}
+
+/**
+ * This function is used to format the url to be used in the Link component
```
illume, line 29: Maybe get rid of this part: "This function is used to" which is a bit redundant for this component.

illume (approved):
Looks good to me. Thanks. 🎉

Nice addition of those tests.

I left a couple of minor nit picks about two comments for your consideration.

Feel free to merge it in when you're ready.

PR #1285, opened 2023-07-31
frontend: Fixes ingress render

vyncent-t:
This PR resolves an issue that was causing the Ingress tab to fail to load, resulting in a blank screen for users. The problem stemmed from a bug in the rendering logic for the Ingress resources.


illume (approved):
👍 thanks

joaquimrocha (approved):

PR #1286, opened 2023-08-01
Add collapse to events table + fix simple table

joaquimrocha:
These changes fix a few issues in the simple table:

Here's the table before the fix (ignore the blueboxes):
Empty table screenshot before the fix

And here it is after the fix:
Empty table screenshot after fix

And they add an expand button to the events' message.

Storybook stories are added.

### DIFF
--- a/frontend/src/components/cluster/Overview.tsx
+++ b/frontend/src/components/cluster/Overview.tsx
@@ -15,6 +15,7 @@ import Empty from '../common/EmptyContent';
 import { PageGrid } from '../common/Resource';
 import ResourceListView from '../common/Resource/ResourceListView';
 import { SectionBox } from '../common/SectionBox';
+import ShowHideLabel from '../common/ShowHideLabel';
 import { LightTooltip } from '../common/Tooltip';
 import { CpuCircularChart, MemoryCircularChart, PodsStatusCircleChart } from './Charts';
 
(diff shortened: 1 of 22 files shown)


illume:
All of the messages I see are under 256 chars. I guess there's a limit somewhere around that much. I'd suggest showing all of the message unless it's longer than that? Or even up to 500/1000?

A lot of the detail seems to be at the end of the messages. 

Here's two examples:
> Failed to pull image "ghcr.io/headlamp-k8s/headlamp:latest": rpc error: code = Unknown desc = Error response from daemon: Get "https://ghcr.io/v2/": dial tcp 140.82.121.34:443: i/o timeout

> Back-off restarting failed container storage-provisioner in pod storage-provisioner_kube-system(2cf3d8e8-113a-4760-8963-9a8fa7585458)

If we cut off too much data, then a click is required to see each of the ones you want to inspect... which would be tiresome.

joaquimrocha:
> All of the messages I see are under 256 chars. I guess there's a limit somewhere around that much. I'd suggest showing all of the message unless it's longer than that? Or even up to 500 or 1000?

I ended up doing a 256 char limit cut for whether to show the expand button or not, and the button is now appended to the text after the ellipsis, which looks better.

ashu8912 (approved):
Tested locally this looks neat now, thanks

PR #1297, opened 2023-08-04
headlamp-plugin: v0.6.1, latest api docs, plugin/examples updated 

illume:


joaquimrocha (approved):
LGTM but I wonder if the semver bump is correct as we have introduced new methods, etc. that are now newly available to the plugin lib. So this would mean 0.7.0 rather than 0.6.1, right?

illume:
>  I wonder if the semver bump is correct as we have introduced new methods, etc. that are now newly available to the plugin lib. So this would mean 0.7.0 rather than 0.6.1, right?

Yes, you're right. We could do better next time, or I could do another 0.7.0 release.

PR #1301, opened 2023-08-04
Fix editing resources without a resource version

### LINKED ISSUE #1168: [EditorDialog] Cannot update anything in the editor
**Description**

User is not able to edit anything in the editor when passed "item" resource has no resourceVersion field inside metadata

**Impact**

User is not able to edit anything in the editor when passed "item" resource has no resourceVersion field inside metadata

**Set-up:**
1. Pass this object as an "item" prop to EditorDialog component
`{
                                apiVersion: 'v1',
                                kind: 'TestKind',
                                spec: {
                                    testProp: 'testValue',
                                },
                                metadata: {
                                    name: 'testName',
                                    namespace: 'testNamespace',
                                },
                            }`
2. Try to update anything in the editor. If you can't - that's it. 
3. Try to add resourceVersion field into metadata 
`{
    apiVersion: 'v1',
    kind: 'TestKind',
    spec: {
        testProp: 'testValue',
    },
    metadata: {
        name: 'testName',
        namespace: 'testNamespace',
        resourceVersion: '123',
    },
} `. Now you can

joaquimrocha:
We need to test this with the list of objects to be set.

**Update:** After @callmevladik clarified the issue (#1168 ) , I was able to reproduce and fix it.

fixes #1168

### DIFF
--- a/frontend/src/components/common/Resource/DocsViewer.tsx
+++ b/frontend/src/components/common/Resource/DocsViewer.tsx
@@ -102,20 +102,26 @@ function DocsViewer(props: { docSpecs: any }) {
       {docsLoading ? (
         <Loader title={t('Loading documentation')} />
       ) : (
-        docs.map((docSpec: any) => {
+        docs.map((docSpec: any, idx: number) => {
           if (!docSpec.error && !docSpec.data) {
             return (
-              <Empty>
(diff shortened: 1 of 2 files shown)


ashu8912 (approved):
I tested this for the List kind too and it is working, and the code LGTM

PR #1303, opened 2023-08-07
Add support for dynamic cluster in shared env

### LINKED ISSUE #1249: Allow to set up dynamic clusters safely in a shared environment
Right now we do not enable the dynamic clusters feature except in a non-shared environment (desktop) but we'd like to do so since some may deploy Headlamp as a web service but would like users to be able to set up clusters individually, through a plugin for example.
The problem with allowing this feature is a shared environment is that, since we do not have an authentication system coupled with headlamp (auth is handled by the API server), any added clusters would be visible to all users; it'd also allow for denial of service abuse since one could just keep pushing new clusters.

So we should investigate if there's a way to do this safely while avoiding having to have a user authentication coupled with headlamp itself.

Some suggestions for this include:
1. Keeping the cluster info local (client-side/browser) and always passing the cluster info on every request. Caveats: too much data on every request, and/or insuficient data dependending on how the cluster is to be configured.
2. Using an associated session ID so users only see the clusters set up for their session. Caveats: would not prevent DOS attacks.

knrt10:
This adds support for dynamic cluster in Headlamp for non-shared environment. If enabled, user can see clusters created by them using kubeconfig. The clusters added are completely stateless and others won't be able to see them.

TODO

- [x] Make headlamp calls stateless
- [x] Add and update tests
- [x] Add e2e tests 
- [x] Fix todos of code

fixes #1249

### How to test
- Open 3 different terminals and run 3 different things
  - Go to dir `cd plugins/examples/dynamic-clusters`
  - Run above edited plugin: `npm i && npm start`
  - Run frontend: `cd frontend && npm start`
  - Run backend with dynamic cluster enabled: `cd backend && go run ./cmd -dev -enable-dynamic-clusters`
  - It would be better to test if if you have more than 1 cluster in your environment.
  - Open `http://localhost:3000/` and click on `New Cluster` from the top right of toolbar.
  
  ### Demo
 
https://github.com/headlamp-k8s/headlamp/assets/24803604/a3b0e919-7fb8-42f2-b3ac-512ea5a7b25d


backend/cmd/headlamp.go
```diff
@@ -1047,10 +1051,174 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+			if restConf == nil {
+				log.Printf("Error: clientConfig is nil: %s", err)
+				http.NotFound(w, r)
+				return
+			}
+
+			roundTripper, err := rest.TransportFor(restConf)
+			if err == nil {
+				proxy.Transport = roundTripper
+			}
```
yolossn, line 1216: Use the `LoadContextsFromAPIConfig` from the kubeconfig package to create a Context and use the underlying `SetupProxy` function instead of repeating the logic here.

backend/cmd/headlamp.go
```diff
@@ -1047,10 +1051,174 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+					}
+					clusterFound = true
+					break
+				}
+			}
+
+			for _, user := range kubeconfig.Users {
+				if user.Name == clusterName {
+					if user.User.ClientCertificate != "" {
+						clientCertificateData = readFileContents(user.User.ClientCertificate)
```
yolossn, line 1130: You can use the [clientcmd.Load](https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#Load) function to load the apiConfig from the kubeconfig file contents.

backend/cmd/headlamp.go
```diff
@@ -1102,14 +1270,29 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
+			clusters = append(clusters, Cluster{
+				Name:     context.Name,
+				Server:   context.Cluster.Server,
+				AuthType: context.AuthType(),
+				Metadata: map[string]interface{}{
+					"source": context.SourceStr(),
+				},
+				SessionID: sessionID,
+			})
+		}
```
yolossn, line 1295:
```suggestion
		if context.Source == kubeconfig.DynamicCluster && context.SessionID != sessionID {
                  continue
		} 
		clusters = append(clusters, Cluster{
				Name:     context.Name,
				Server:   context.Cluster.Server,
				AuthType: context.AuthType(),
				Metadata: map[string]interface{}{
					"source": context.SourceStr(),
				},
				SessionID: sessionID,
			})
```

yolossn (dismissed):
Left few comments, Apart from this I think the same logic has to be added to few endpoints like drain node and cordon. cc: @ashu8912

backend/cmd/headlamp.go
knrt10: @yolossn IIUC, I don't think we are loading context here. We are creating API client kubeconfig, so the `LoadContextsFromAPIConfig` could not be used. Am I missing something?

backend/cmd/headlamp.go
```diff
@@ -1121,19 +1214,70 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
 				"source": context.SourceStr(),
 			},
+			SessionID: sessionID,
 		})
 	}
 
 	return clusters
 }
 
+func (c *HeadlampConfig) getStatelessClusters(sessionID string, kubeConfigs []string) []Cluster {
```
yolossn, line 1237: The function `getStatelessClusters` is kind of confusing, I think it should be something like `parseKubeConfig` or `parseClusterFromKubeConfig`. Also I think the sessionID is not required here.

backend/cmd/cluster.go
```diff
@@ -24,3 +25,7 @@ type ClusterReq struct {
 	Metadata                 map[string]interface{} `json:"meta_data"`
 	KubeConfig               *string                `json:"kubeconfig,omitempty"`
 }
+
+type KubeconfigRequest struct {
```
yolossn, line 29: The struct name can be better here.

backend/cmd/headlamp.go
```diff
@@ -120,6 +120,10 @@ type OauthConfig struct {
 	Ctx      context.Context
 }
 
+type statelessClientConfig struct {
```
yolossn, line 123: I think we can reuse the `clientConfig` struct here instead of this new one as it is the same.

backend/cmd/headlamp.go
```diff
@@ -829,6 +834,37 @@ func StartHeadlampServer(config *HeadlampConfig) {
 	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", config.port), handler)) //nolint:gosec
 }
 
+// CheckAndRemoveClusters checks for clusters marked as "dynamic_cluster"
+// and removes them if they are older than 10 minutes.
+func CheckAndRemoveClusters(kubeConfigStore kubeconfig.ContextStore) {
```
yolossn, line 839: Looks like the context is being removed irrespective of the last accessed time, if this is the required logic then it makes more sense to create a function like `AddContextWithLifespan` to make things a bit more organized and leverage the caches `SetWithTTL` functionality.

https://github.com/headlamp-k8s/headlamp/blob/41738dff59a9af0c3ad77529d5981f9b38169afb/backend/pkg/kubeconfig/contextStore.go#L10-L16

yolossn:
I think we have got the Session ID part wrong. In this current logic the session ID is created by the client and the backend is not involved either in creating or deleting the session. If any random UUID is accepted by the backend as a valid session ID then anyone can mimic and keep adding dynamic clusters, ideally the session has to be created by the backend here.

backend/cmd/headlamp.go
knrt10: I agree on the name but sessionID is required because it is used on the frontend side to filter cluster.

knrt10: I understand your point now. It makes sense to remove it. Thanks

backend/cmd/headlamp.go
knrt10: Thanks for the idea. Made the necessary changes.

backend/cmd/headlamp.go
```diff
@@ -1066,10 +1102,67 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+
+			config, err := clientcmd.Load(kubeConfigByte)
+			if err != nil {
+				http.Error(w, "Error loading kubeconfig", http.StatusBadRequest)
+				return
+			}
+
+			var contexts []kubeconfig.Context
+			var setupErrors []error
+			contexts, setupErrors = kubeconfig.LoadContextsFromAPIConfig(config, true)
```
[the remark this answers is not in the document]
knrt10: Done

backend/cmd/headlamp.go
```diff
@@ -1134,6 +1177,49 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+func (c *HeadlampConfig) parseClusterFromKubeConfig(kubeConfigs []string) []Cluster {
+	clusters := []Cluster{}
+
+	for _, kubeConfig := range kubeConfigs {
+		var contexts []kubeconfig.Context
+
+		var setupErrors []error
```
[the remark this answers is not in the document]
knrt10: yes, we should return error. Fixed

backend/pkg/kubeconfig/contextStore.go
```diff
@@ -62,3 +65,13 @@ func (c *contextStore) GetContext(name string) (*Context, error) {
 func (c *contextStore) RemoveContext(name string) error {
 	return c.cache.Delete(context.Background(), name)
 }
+
+// AddContextWithTTL adds a context to the store with a ttl.
+func (c *contextStore) AddContextWithTTL(headlampContext *Context, ttl time.Duration) error {
+	return c.cache.SetWithTTL(context.Background(), headlampContext.Name, headlampContext, ttl)
```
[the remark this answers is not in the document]
knrt10: Since we removed concept of sessionID, do it think the having combination of `name + server` as a key would make sense?

backend/cmd/headlamp.go
```diff
@@ -1066,11 +1071,80 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
 	})
 }
 
-func handleClusterAPI(c *HeadlampConfig, router *mux.Router) {
+func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen,gocognit
```
illume, line 1074: Can you please break this down into smaller parts? Then remove the nolint part.

frontend/src/helpers/index.ts
```diff
@@ -391,6 +416,82 @@ export function getHeadlampAPIHeaders(): { [key: string]: string } {
   };
 }
 
+// Export function to store cluster kubeconfig
```
illume, line 419: We should use "tsdoc" style documentation.

frontend/src/helpers/index.ts
```diff
@@ -391,6 +416,82 @@ export function getHeadlampAPIHeaders(): { [key: string]: string } {
...
+  if (storedClusterKubeconfigsJSON) {
+    const storedClusterKubeconfigs: string[] = JSON.parse(storedClusterKubeconfigsJSON);
+
+    return storedClusterKubeconfigs;
+  }
+
+  return [];
+}
+
+export function findKubeconfigByClusterName(clusterName: string) {
```
illume, line 452: This needs documentation.

frontend/src/redux/configSlice.ts
```diff
@@ -9,6 +9,17 @@ export interface ConfigState {
   clusters: {
     [clusterName: string]: Cluster;
   } | null;
+  /**
+   * Statelss Clusters is a map of cluster names to cluster objects.
```
illume, line 13: Can you please explain here why they are they stateless here? I don't get the difference between these and clusters.

frontend/src/redux/configSlice.ts
```diff
@@ -9,6 +9,17 @@ export interface ConfigState {
...
+   * Null indicates that the clusters have not been loaded yet.
+   */
+  statelessClusters: {
+    [clusterName: string]: Cluster;
+  } | null;
+
+  allClusters: {
```
illume, line 20: Needs documentation.

frontend/src/helpers/index.ts
```diff
@@ -391,6 +416,82 @@ export function getHeadlampAPIHeaders(): { [key: string]: string } {
...
+  return null; // If not found
+}
+
+export function getUserId(): string {
+  // Retrieve userId ID from sessionStorage
+  let userId = sessionStorage.getItem('userId');
+
+  // If no userId ID exists, generate one
+  if (!userId) {
+    userId = uuidv4();
```
illume, line 484: Can you please explain why we generate the id here on the client side?

frontend/src/helpers/index.ts
```diff
@@ -391,6 +416,82 @@ export function getHeadlampAPIHeaders(): { [key: string]: string } {
...
+        const encodedKubeconfig = btoa(JSON.stringify(kubeconfigObject));
+        return encodedKubeconfig;
+      }
+    }
+  }
+
+  return null; // If not found
+}
+
+export function getUserId(): string {
```
illume, line 478: This needs documentation.

frontend/src/lib/k8s/index.ts
```diff
@@ -86,9 +86,19 @@ classList.forEach(cls => {
 export const ResourceClasses = resourceClassesDict;
 
 // Hook for getting or fetching the clusters configuration.
-export function useClustersConf(): ConfigState['clusters'] {
-  const clusters = _.cloneDeep(useTypedSelector(state => state.config.clusters));
-  return clusters;
+export function useClustersConf(): ConfigState['allClusters'] {
```
illume, line 89: This seems to be only used once, so it should live near where it's used.

frontend/src/helpers/index.ts
```diff
@@ -391,6 +416,82 @@ export function getHeadlampAPIHeaders(): { [key: string]: string } {
...
+  // Retrieve userId ID from sessionStorage
+  let userId = sessionStorage.getItem('userId');
```
illume, line 480: It's not clear to me why sessionStorage is being used here. Maybe the documentation would explain this.

Consider adding a "headlamp-" prefix to prevent possible clashes.

backend/cmd/headlamp.go
```diff
@@ -1154,6 +1280,38 @@ func (c *HeadlampConfig) getConfig(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (c *HeadlampConfig) parseKubeConfig(w http.ResponseWriter, r *http.Request) {
```
illume, line 1283: Missing documentation.

backend/cmd/headlamp.go
```diff
@@ -1134,6 +1209,57 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+func (c *HeadlampConfig) parseClusterFromKubeConfig(kubeConfigs []string) ([]Cluster, []error) {
```
illume, line 1212: Missing documentation.

illume (commented):
Nice one. I didn't get to test it, but just looked through the code.

- For the backend, it would be preferable if the related functionality was together as much as possible. Maybe in a new file/pkg if that seems appropriate.
- The backend code needs tests added.
- The frontend related code is also missing tests.

For the frontend code... It would be better if the functionality for this feature was clustered together, rather than spread amongst many different files. Consider making a folder or a component where all of this functionality can live.  Then if code needs to be placed somewhere else that file can import from a central place where this feature is. If it's organized by feature, in the future changing this feature will be easier than having to try and figure out where and in which files are related to it.

Please avoid adding anything more to helpers if possible. This file is an anti-pattern, because it doesn't name any particular functionality and is just a random selection of. In general things should be placed somewhere with a name that gives a hint to their functionality.

backend/cmd/headlamp.go
```diff
@@ -1066,11 +1071,80 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
 
-		kContext, err := c.kubeConfigStore.GetContext(clusterName)
+		// checking if kubeConfig exists, if not check if the request headers for kubeConfig information
+		kubeConfig := r.Header.Get("KUBECONFIG")
+		userID := r.Header.Get("X-HEADLAMP-USER-ID")
+		// unique key for the context
+		key := clusterName + userID
+
+		if kubeConfig != "" && c.enableDynamicClusters { //nolint:nestif
+			contexts, err := kubeconfig.LoadContextsFromBase64String(kubeConfig, kubeconfig.DynamicCluster)
```
yolossn, line 1086:
```suggestion
		if kubeConfig != "" && c.enableDynamicClusters { //nolint:nestif
		
			kubeConfig := r.Header.Get("KUBECONFIG")
		        userID := r.Header.Get("X-HEADLAMP-USER-ID")
		        // unique key for the context
		        key := clusterName + userID


			contexts, err := kubeconfig.LoadContextsFromBase64String(kubeConfig, kubeconfig.DynamicCluster)
```

backend/cmd/headlamp.go
```diff
@@ -1066,11 +1071,80 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+		queryParams := u.Query()
+
+		// Check if "X-HEADLAMP-USER-ID" parameter is present in the websocket URL.
+		userIDparam := queryParams.Get("X-HEADLAMP-USER-ID")
+		if userIDparam != "" {
+			contextKey = clusterName + userIDparam
+		}
+
+		// Remove the "X-HEADLAMP-USER-ID" parameter from the websocket URL.
+		delete(queryParams, "X-HEADLAMP-USER-ID")
```
yolossn, line 1145: If this logic is specific to websockets please add it inside an if block

yolossn, line 1145: After the query param is deleted but I dont see the URL being rewritten to request.RawURL after deleting.

backend/cmd/headlamp.go
knrt10: In this, we cannot check kubeConfig even before getting it? Also we cannot just check `enableDynamicClusters` because even it's enabled user can choose not to send kubeConfig.

backend/cmd/headlamp.go
knrt10: Updated

frontend/src/helpers/index.ts
knrt10: Added the comment why this function is needed.

frontend/src/lib/k8s/index.ts
knrt10: I am sorry I don't understand what you mean my used only once? I workflow is we are combining normal clusters and stateless clusters. Scenario can be user don't have it enabled then it will only return allClusters as normal clusters from all sources. But if it is enabled then we combine the two.

Also `useClustersConf` is used throught the frontend. Please correct me if I am missing something?

cc @illume

frontend/src/redux/configSlice.ts
knrt10: Indeed, while the two types of clusters share many similarities, it's important to distinguish them as 'stateless clusters' for future considerations. Stateless clusters are handled primarily on the frontend, using Redux for retrieval and processing, while normal clusters are fetched from the backend proxy cache. This separation allows us to implement specific features or functionality for stateless clusters in a more streamlined manner.

backend/cmd/headlamp.go
```diff
@@ -760,6 +808,58 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
 }
 
+// Handles stateless cluster requests if kubeconfig is set and dynamic clusters are enabled.
+// It returns context key which is used to store the context in the cache.
+func (c *HeadlampConfig) handleStatelessReq(r *http.Request, kubeConfig string) (string, error) {
+	var key string
+
+	var contextKey string
+
+	userID := r.Header.Get("X-HEADLAMP-USER-ID")
```
yolossn, line 818: check for empty userID is missing

backend/cmd/headlamp.go
```diff
@@ -719,11 +724,54 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+			key, err := c.handleStatelessReq(r, kubeConfig)
+			if err != nil {
+				log.Println("Error: handling stateless cluster request: ", err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			contextKey = key
+		} else {
+			contextKey = clusterName
+		}
```
yolossn, line 746: If KUBECONFIG is present in all requests made by a client then the handleStatelessReq will be called everytime irrespective of the cluster being present in the kubeConfigStore.

yolossn (dismissed):

frontend/src/components/App/Layout.tsx
```diff
@@ -117,23 +162,27 @@ export default function Layout({}: LayoutProps) {
    */
   const fetchConfig = () => {
     const clusters = store.getState().config.clusters;
+    const statelessClusters = store.getState().config.statelessClusters;
     request('/config', {}, false, false)
       .then((config: Config) => {
         const clustersToConfig: ConfigState['clusters'] = {};
         config?.clusters.forEach((cluster: Cluster) => {
           clustersToConfig[cluster.name] = cluster;
+          if (clustersToConfig[cluster.name].meta_data.source === 'dynamic_cluster') {
```
yolossn, line 171: I think this check is not necessary as the /config endpoint doesn't expose any internal clusters. There can be a case where we have dynamic_cluster that is not stateless.

yolossn (commented):
The backend changes look good. @illume can you take a look at the frontend changes?

backend/cmd/headlamp.go
```diff
@@ -719,11 +724,54 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen
 	router.PathPrefix("/clusters/{clusterName}/{api:.*}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var contextKey string
 		clusterName := mux.Vars(r)["clusterName"]
 
-		kContext, err := c.kubeConfigStore.GetContext(clusterName)
+		// checking if kubeConfig exists, if not check if the request headers for kubeConfig information
+		kubeConfig := r.Header.Get("KUBECONFIG")
+
+		if kubeConfig != "" && c.enableDynamicClusters {
```
yolossn, line 735: the enableDynamicClusters check is not required here as it is already checked
https://github.com/headlamp-k8s/headlamp/blob/1bb8e942e75f0ef843d30ea3be4d9e6b2e6c9049/backend/cmd/headlamp.go#L926

e2e-tests/tests/dynamicCluster.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+  if (stderr) {
+    console.error('Error fetching Minikube kubeconfig:', stderr);
+    return;
+  }
+
+  // Parse the kubeconfig JSON
+  const kubeconfig = JSON.parse(stdout);
+  // Update the existing cluster and context names to "test"
+  kubeconfig.clusters[0].name = 'test';
+  kubeconfig.clusters[0].cluster.server = 'https://10.96.0.1:443';
```
yolossn, line 55: the server url is hard coded here, can fail if minikube starts in another address due to some reason.

knrt10: It won't since it is default in minikube if not changes. The `10.96.0.0/12`: Used by service cluster IP’s and the first service that is created is that of minikube when it bootstraps the cluster, so it won't be a problem. https://minikube.sigs.k8s.io/docs/handbook/vpn_and_proxy/

backend/cmd/headlamp.go
```diff
@@ -922,11 +1113,10 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *HeadlampConfig) addClusterSetupRoute(r *mux.Router) {
-	// We do not support this feature when in-cluster
-	if !c.enableDynamicClusters {
```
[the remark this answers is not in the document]
knrt10: yes, this can be added again, I forgot to revert it back since we went for stateless cluster. Thank you for noticing.

knrt10:
Thank you for the review @yolossn. Addressed the changes.

yolossn (approved):
Reviewed the backend part and tested everything locally. Someone please take a look at the frontend changes once.

joaquimrocha (changes requested):
@knrt10 , I gave it a look and was seeing things I think you had addressed. So I noticed you added different commits.
Since @yolossn has already reviewed it, I think it makes sense to squash some of the commits. We need commits to be atomic, but when changing things introduced by commits that are in the PR itself, we should rather squash those commits together, so there's no intermediate states in the main tree.

knrt10:
@joaquimrocha thank you, make sense. Done

illume:
👀

backend/cmd/headlamp.go
```diff
@@ -719,11 +720,57 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
 	})
 }
 
-func handleClusterAPI(c *HeadlampConfig, router *mux.Router) {
+func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen
```
illume, line 728: This function is too long.

frontend/src/redux/configSlice.ts
```diff
@@ -9,6 +9,22 @@ export interface ConfigState {
   clusters: {
     [clusterName: string]: Cluster;
   } | null;
+  /**
+   * Stateless Clusters is a map of cluster names to cluster objects.
```
illume, line 13: Can you please answer these in the docs: 
- What is a stateless cluster? 
- Why is it separate from allClusters?

frontend/src/components/App/Layout.tsx
```diff
@@ -156,13 +163,87 @@ export default function Layout({}: LayoutProps) {
           });
 
           if (isConfigDifferent) {
-            dispatch(setConfig(configToStore));
+            dispatch(setConfig({ ...configToStore }));
```
illume, line 205: There's quite a lot of logic here that should probably be somewhere else.

backend/cmd/headlamp.go
```diff
@@ -760,6 +807,58 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
 	handleClusterAPI(c, router)
 }
 
+// Handles stateless cluster requests if kubeconfig is set and dynamic clusters are enabled.
+// It returns context key which is used to store the context in the cache.
+func (c *HeadlampConfig) handleStatelessReq(r *http.Request, kubeConfig string) (string, error) {
```
illume, line 814: I feel like this should all be in package rather than in the headlamp.go? @yolossn what do you think?

yolossn: have made sure new logic that is introduced is part of existing packages wherever possible. I don't think this stateless part alone needs a separate package.

backend/cmd/headlamp.go
knrt10: Fixed.

frontend/src/components/App/Layout.tsx
knrt10: Can you please explain a bit more? Do you mean the part about `statelessClusters` or this whole function. Also to move elsewhere do you mean in the same file or a separate file? Thanks

knrt10:
Thank you for the review @illume. I have made the necessary changes. Can you please review it again?

Also regarding:

> Also, one possible improvement for context... rather than just "frontend:" the component/s that it affects can be added like "frontend: Layout:". Because 90% of commits are in frontend, and it doesn't give you any context on what modules are changed without it.

As per [git commit standards](https://www.conventionalcommits.org/en/v1.0.0/) followed by others which is not an official standard per say but I don't think when writing commit headlines, we need to add which files are changed. We can easily see that with `git whatchanged`. WDYT?

frontend/src/components/App/Layout.tsx
illume: There's still a lot of logic related to this in the Layout component.

frontend/src/redux/configSlice.ts
knrt10: Fixed it in the docs

illume:
> Thank you for the review @illume. I have made the necessary changes. Can you please review it again?
> 
> Also regarding:
> 
> > Also, one possible improvement for context... rather than just "frontend:" the component/s that it affects can be added like "frontend: Layout:". Because 90% of commits are in frontend, and it doesn't give you any context on what modules are changed without it.
> 
> As per [git commit standards](https://www.conventionalcommits.org/en/v1.0.0/) followed by others which is not an official standard per say but I don't think when writing commit headlines, we need to add which files are changed. We can easily see that with `git whatchanged`. WDYT?

We follow the commit guidelines linked from the contributing documentation, see here: https://github.com/kinvolk/contribution/blob/master/topics/git.md#commit-format

illume (commented):
Looking real good. I left a few notes for your consideration.

frontend/src/components/App/Layout.tsx
knrt10: Got it, updating the same.

backend/cmd/headlamp_test.go
```diff
@@ -109,7 +109,7 @@ func makeJSONReq(method, url string, jsonObj interface{}) (*http.Request, error)
 	return http.NewRequestWithContext(context.Background(), method, url, bytes.NewBuffer(jsonBytes))
 }
 
-func getResponse(handler http.Handler, method, url string, body interface{}) (*httptest.ResponseRecorder, error) {
+func getResponse(handler http.Handler, method, url string, body interface{}) (*httptest.ResponseRecorder, error) { //nolint:lll
```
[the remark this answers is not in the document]
knrt10: Fixed

backend/cmd/headlamp.go
knrt10: Agreed. Since the new logic about stateless cluster does not need specific implementation, creating a new package is not necessary.

illume:
(I can't use the review button for some reason. Hopefully this comment works.)

- It would be good to add some documentation for a person trying to use this feature.
- Is this feature meant to only be used by a plugin? 
- Probably best a separate example plugin (as the one in your PR description) could be added?

illume:
Seems you missed updating one commit message.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,344 @@
+import { indexedDB as indexedDBtest } from 'fake-indexeddb';
```
illume, line 1: This testing module shouldn't really be imported here.

Maybe it's not even necessary? If the tests can run without it, it would be good to avoid adding the dependency.

illume (changes requested):
It would be good to remove that extra frontend dependency if possible. If it's not at all possible to remove the dependency, then only importing it inside a test would be better.

backend/cmd/headlamp.go
illume: If not a package, then I still feel the code should be separated into its own file.

Keeping it in headlamp.go mixes it in with a lot of things and it makes it hard to know what goes with what.

illume (dismissed):
I left some more notes. This time with a security focus.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,344 @@
...
+    context: {
+      cluster: string;
+      user: string;
+    };
+  }>;
+  'current-context': string;
+}
+
+/**
+ * Store the kubeconfig for a stateless cluster in IndexedDB.
```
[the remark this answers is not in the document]
knrt10: Since we are storing **kubeconfig** in base64 encoded format, the localstorage has [limit](https://stackoverflow.com/questions/2989284/what-is-the-max-size-of-localstorage-values/) of storing ~5Mb of characters. And a user can create multiple stateless clusters which would only result in storage issues later on, so it was suggested and decided to use [indexDB](https://stackoverflow.com/questions/5692820/maximum-item-size-in-indexeddb) since there's isn't a specific limit for a size of a single item, only a global limit. And Maximum storage size depends on 2 things, Browser and Disk space.

backend/cmd/headlamp.go
```diff
@@ -760,6 +827,59 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
+
+	for _, context := range contexts {
+		context := context
+
+		if context.Name == clusterName { //nolint:nestif
+			// check context is present
+			_, err := c.kubeConfigStore.GetContext(key)
+			if err != nil {
+				if err.Error() == "key not found" {
+					context.Internal = true
```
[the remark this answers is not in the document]
knrt10: Stateless clusters are not to be shown to the other users. We are storing them in proxy cache, so they were available in `/config` endpoint. To fix this we are marking them as internal clusters. Will comment in the code as well.

backend/cmd/headlamp.go
```diff
@@ -719,11 +725,72 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+			}
+			contextKey = key
+		} else {
+			contextKey = clusterName
+		}
+
+		// This means the connection is from websocket so there won't be kubeconfig header.
+		// We get the value of X-HEADLAMP-USER-ID from the parameter and append it to the cluster name
+		// to get the context key. This is to ensure that the context key is unique for each user.
+		if r.Header.Get("Upgrade") == "websocket" {
```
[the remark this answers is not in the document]
knrt10: Yes, it's been tested and it does work. When user clicks on any page new websockets connections are created and even if `enableDynamicClusters` is set or not we need a key to get the proxy value from the cache.

backend/cmd/headlamp.go
```diff
@@ -719,11 +725,72 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
 
+// handleWebsocket handles websocket requests. It returns context key
+// which is used to store the context in the cache. The context key is
+// unique for each user. It is found in the "X-HEADLAMP-USER-ID" parameter
+// in the websocket URL.
+func handleWebsocket(w http.ResponseWriter, r *http.Request, clusterName string) string {
+	var contextKey string
+	// Parse the URL
+	u, err := url.Parse(r.URL.String())
+	if err != nil {
```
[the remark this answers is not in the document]
knrt10: Will update it.

backend/cmd/headlamp.go
```diff
@@ -771,6 +891,11 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 
 	for _, context := range contexts {
 		context := context
+
+		if context.Internal {
```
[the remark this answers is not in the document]
knrt10: https://github.com/headlamp-k8s/headlamp/pull/1303#discussion_r1423262134

Continuing this. Since `/config` show's all the clusters saved in the cache, we don't want others to see when someone created a stateless cluster since their proxy is stored in cache.

backend/cmd/headlamp.go
```diff
@@ -719,11 +725,72 @@ func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
...
+		log.Println("Error: parsing URL: ", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+
+		return ""
+	}
+
+	// Get the query parameters
+	queryParams := u.Query()
+	// Check if "X-HEADLAMP-USER-ID" parameter is present in the websocket URL.
+	userIDparam := queryParams.Get("X-HEADLAMP-USER-ID")
```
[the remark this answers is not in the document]
knrt10: It was discussed internally between @yolossn and myself. When there is a websocket connection we cannot send kubeconfig as header like we are doing in a normal request. Websocket do not support it. Websockets connection are only open when a cluster proxy has been saved in cache and when a user tries to open let say home page of a particular cluster. Since we didn't have kubeconfig sent from the cluster we needed a way to know whether the cluster that user has clicked on was a normal cluster or stateless one. 

This param isolate's stateless cluster created by one user alone.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,344 @@
...
+ * cluster name and this headlamp-userId to create a unique ID for a cluster. If we don't
+ * do it then if 2 different users have a cluster with the same name, then the
+ * proxy will be overwritten.
+ * @returns headlamp-userId from localStorage
+ */
+export function getUserIdFromLocalStorage(): string {
+  let headlampUserId = localStorage.getItem('headlamp-userId');
+
+  if (!headlampUserId) {
+    headlampUserId = uuidv4();
```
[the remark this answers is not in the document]
knrt10: We do not need a secure token per say, we needed a random uuid to store. Which is then combined with cluster name. Since it's random I don't think it would be easy to guess or hack. 

Please correct me if I don't understand your concern regarding this.

backend/cmd/headlamp.go
```diff
@@ -922,11 +1132,14 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *HeadlampConfig) addClusterSetupRoute(r *mux.Router) {
-	// We do not support this feature when in-cluster
+	// Do not add the route if dynamic clusters are disabled
 	if !c.enableDynamicClusters {
 		return
 	}
+	// Get stateless cluster
+	r.HandleFunc("/parseKubeConfig", c.parseKubeConfig).Methods("POST")
```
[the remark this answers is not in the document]
knrt10: I think it make sense. Since it is taking kubeconfig as in input. And then it parses the base64 encoded kubeconfig and returns it in desired format. It is not replacing any existing config. It is just parsing the values and returning it to frontend.

backend/cmd/headlamp.go
knrt10: Done.

frontend/src/stateless/index.ts
illume: This code is for testing, and it should only be imported by the test code.

illume (commented):
I left some more notes. Mostly I'm worried about the query param

backend/cmd/headlamp.go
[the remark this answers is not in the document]
knrt10: Ok, I will send it as a message to add security. Thanks for the suggestion

frontend/src/stateless/index.ts
knrt10: Thanks for the suggestion, moving this to a test file and injecting it here.

backend/cmd/headlamp.go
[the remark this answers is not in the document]
joaquimrocha: @illume , I want to understand this issue better.
1. Can the info be leaked only in non-secure connections?
2. Can someone get straight access to the cluster, or just be able to see that the cluster exists (but they still require a token)?

We are hoping no-one exposes Headlamp to the internet through a non-secure connection of course.

frontend/src/stateless/index.ts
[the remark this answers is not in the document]
knrt10: As discussed, we have now used crypto package here instead of uuid.

backend/cmd/headlamp.go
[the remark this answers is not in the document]
knrt10: This is now fixed. We are now not using query parameters anymore. We now send the uuid using websocket protocol headers which solves all the security issue. Please refer https://github.com/headlamp-k8s/headlamp/pull/1303/commits/36613c84b6e7830c8061f8fcef715bb81590f592

illume:
Hi,

a couple of things regarding the next round of testing/reviewing.

1) I would strongly prefer if the example plugin was part of this PR. There are some testing infrastructure that runs on it. Additionally, I will be able to test it and understand it more easily.

2) For the next review round, can you please leave the conversations unresolved and write there instead that you think it's done. This PR is too big for me to go through each resolved conversation to check if it's been changed and needs retesting. So it's better if you leave them open and then I can close them if I also think they are resolved. Please write in the ones you think are resolved now, and I will mark the conversations as resolved.

frontend/src/components/App/Layout.tsx
```diff
@@ -115,54 +113,39 @@ export default function Layout({}: LayoutProps) {
    */
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/lib/k8s/index.ts
```diff
@@ -88,9 +88,19 @@ classList.forEach(cls => {
...
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+ * KubeconfigObject is the object that is stored in indexDB as string format.
+ * It is a JSON encoded version of the kubeconfig file.
+ * It is used to store the kubeconfig for stateless clusters.
+ * This is basically a k8s client - go Kubeconfig object.
+ * @see storeStatelessClusterKubeconfig
+ * @see getStatelessClusterKubeConfigs
+ * @see findKubeconfigByClusterName
+ */
+interface KubeconfigObject {
+  apiVersion: string;
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+  }
+  window.crypto.getRandomValues(buffer);
+  return Array.from(buffer, byte => byte.toString(16).padStart(2, '0'))
+    .join('')
+    .slice(0, length);
+}
+
+/**
+ * Compares the cluster config from the backend and the redux store
+ * and returns true if the present stored config is different from the fetched one.
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+ * @param clusters
+ * @param clustersToConfig
+ * @param isStateless
+ */
+export function processClusterComparison(
+  clusters: ConfigState['clusters'],
+  clustersToConfig: any,
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/**
+ * Compares the cluster config from the backend and the redux store
+ * and returns true if the present stored config is different from the fetched one.
+ * @param clusters
+ * @param clustersToConfig
+ * @param isStateless
+ */
+export function processClusterComparison(
```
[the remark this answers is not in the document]
knrt10: I think the e2e has enough coverage for now. I plan to add more e2e tests later once the PR is merged to test edge cases. I think it would be enough to cover the functionality.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+    })
+    .catch(error => {
+      console.error('Error importing fake-indexeddb:', error);
+    });
+}
+
+/**
+ * Config is the object that is stored in localStorage as string format.
+ */
+interface Config {
```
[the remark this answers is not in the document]
knrt10: We need this since Spread types may only be created from object types. Modified its name for better clarity.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+      user: string;
+    };
+  }>;
+  'current-context': string;
+}
+
+/**
+ * Store the kubeconfig for a stateless cluster in IndexedDB.
+ * @param kubeconfig - The kubeconfig to store.
+ * @returns void
```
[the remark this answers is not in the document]
knrt10: Thanks. Fixed

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+ * @throws Error if IndexedDB is not supported.
+ * @throws Error if the kubeconfig is invalid.
+ */
+export function storeStatelessClusterKubeconfig(kubeconfig: any): Promise<void> {
+  return new Promise<void>(async (resolve, reject) => {
+    let request;
```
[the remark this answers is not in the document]
knrt10: Thanks, done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+    request.onupgradeneeded = function (event: any) {
+      const db = event.target ? event.target.result : null;
+      if (db) {
+        if (!db.objectStoreNames.contains('kubeconfigStore')) {
+          db.createObjectStore('kubeconfigStore', { keyPath: 'id', autoIncrement: true });
+        }
+      }
+    };
+
+    request.onsuccess = function (event: any) {
```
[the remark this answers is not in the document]
knrt10: Done

backend/cmd/headlamp_test.go
```diff
@@ -492,7 +496,7 @@ func TestDrainAndCordonNode(t *testing.T) {
 		}
 
 		rr, err = getResponse(tc.handler, "GET",
-			fmt.Sprintf("/drain-node-status?cluster=%s&nodeName=%s", drainNodePayload.Cluster, drainNodePayload.NodeName), nil)
+			fmt.Sprintf("/drain-node-status?cluster=%s&nodeName=%s", drainNodePayload.Cluster, drainNodePayload.NodeName), nil) //nolint:lll
```
illume, line 499: For each time we disable a linter/type checker it should have an explanation of why it was disabled as a comment. Or it should be fixed.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -269,6 +273,38 @@ func LoadContextsFromMultipleFiles(kubeConfigs string, source int) ([]Context, e
 	return contexts, errors.Join(errs...)
 }
 
+// LoadContextsFromBase64String loads contexts from the given kubeconfig string.
+func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, error) {
```
illume, line 277: I don't see a test for this?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -236,10 +238,12 @@ func LoadContextsFromAPIConfig(config *api.Config) ([]Context, []error) {
 			AuthInfo:    authInfo,
 		}
 
-		err := context.SetupProxy()
-		if err != nil {
-			errors = append(errors, fmt.Errorf("couldnt setup proxy for context: %q, err:%q", contextName, err))
-			continue
+		if !skipProxySetup {
```
illume, line 241: Is there a test for this?

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -12,6 +12,11 @@ import _ from 'lodash';
 import { decodeToken } from 'react-jwt';
 import helpers, { getHeadlampAPIHeaders, isDebugVerbose } from '../../helpers';
 import store from '../../redux/stores/store';
+import {
+  findKubeconfigByClusterName,
+  getUserIdFromLocalStorage,
+  storeStatelessClusterKubeconfig,
+} from '../../stateless/index';
```
illume, line 19: Nit, the /index can be removed from here.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1334,7 +1351,7 @@ interface StreamParams {
   additionalProtocols?: string[];
 }
 
-function connectStreamWithParams(
+async function connectStreamWithParams(
```
illume, line 1354: I see this was changed to async, but I don't see any call sites changed. Can you please confirm if the call sites are also changed? (do they now need await?)

It would be good if you can document at least the changes to this function.

frontend/src/lib/k8s/index.ts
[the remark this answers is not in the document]
illume: This needs to be a tsdoc style comment.

plugins/examples/dynamic-clusters/src/index.tsx
```diff
@@ -1,24 +1,102 @@
-import { Headlamp, registerAppBarAction } from '@kinvolk/headlamp-plugin/lib';
-import { Button } from '@mui/material';
+import { ApiProxy, Headlamp, registerAppBarAction } from '@kinvolk/headlamp-plugin/lib';
+import { Button, Modal, Paper, Stack, TextField, Typography } from '@mui/material';
+import { useEffect, useState } from 'react';
+
```
illume, line 4: Can you please update the README to explain what it demonstrates. Also maybe some more commentary could be helpful in here to describe what it demonstrates.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,471 @@
...
+    cluster: {
+      server: string;
+    };
+  }>;
+  // users is a list of the users.
+  users: Array<{
+    // name is the name of the user.
+    name: string;
+    // user is the user information.
+    user: {};
```
illume, line 110: I think user needs a better type.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,471 @@
...
+ * @param event - The event that is fired when the IndexedDB is upgraded.
+ * @see storeStatelessClusterKubeconfig
+ * @see getStatelessClusterKubeConfigs
+ * @see findKubeconfigByClusterName
+ * **/
+function handleDatabaseUpgrade(event: DatabaseEvent) {
+  const db = event.target ? event.target.result : null;
+  // Create the object store if it doesn't exist
+  if (db) {
+    if (!db.objectStoreNames.contains('kubeconfigStore')) {
```
illume, line 140: nit, this could be `if (db && !db. ...)`

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,471 @@
...
+}
+
+/**
+ * Store the kubeconfig for a stateless cluster in IndexedDB.
+ * @param kubeconfig - The kubeconfig to store.
+ * @returns promise that resolves when the kubeconfig is successfully added.
+ * @throws Error if IndexedDB is not supported.
+ * @throws Error if the kubeconfig is invalid.
+ */
+export function storeStatelessClusterKubeconfig(kubeconfig: any): Promise<void> {
```
illume, line 165: Is there a better type than any here?

frontend/src/stateless/index.ts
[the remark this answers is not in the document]
illume: `@returns` is for a description

> `@returns true if the present stored config is different from the fetched one.`

illume (changes requested):
Left a few more notes.

I still have about 7 more files to review, and then to test.

backend/cmd/headlamp_test.go
knrt10: Fixed

backend/pkg/kubeconfig/kubeconfig.go
knrt10: It doesn't need specific testing. The function that calls this `LoadContextsFromAPIConfig` has tests already.

frontend/src/stateless/index.ts
knrt10: I think, It is not required, since it can be anything. And the use of KubeconfigObject is to get the config loaded in right format. And `jsyaml.load` does the same. It would be unnecessary to add type checks for users here

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,372 @@
...
+import { request } from '../lib/k8s/apiProxy';
+import { Cluster } from '../lib/k8s/cluster';
+import { ConfigState, setStatelessConfig } from '../redux/configSlice';
+import store from '../redux/stores/store';
+
+let indexedDBtest: any;
+
+// Import 'fake-indexeddb' only in the testing environment
+if (process.env.NODE_ENV === 'test') {
+  import('fake-indexeddb')
```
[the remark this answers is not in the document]
knrt10: I did not try it, let me check. Don't you think it would require much of a change in this if we change it now? It could be added in the scope of a different PR?

frontend/src/lib/k8s/apiProxy.ts
knrt10: Added documentation. I checked, the calls do not need await, they are working as expected

backend/cmd/headlamp_test.go
illume: I still see `//nolint:lll` without an explanation.

knrt10: Done now

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,468 @@
...
+}
+
+/**
+ * ParsedConfig is the object that is fetched from the backend.
+ * It has cluster information. The values are the same as the KubeconfigObject.
+ * @see KubeconfigObject
+ * @see fetchStatelessClusterKubeConfigs
+ */
+interface ParsedConfig {
+  [prop: string]: any;
```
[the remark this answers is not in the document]
knrt10: Edited

frontend/src/stateless/index.ts
knrt10: Sure done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,468 @@
...
+  return new Promise<void>(async (resolve, reject) => {
+    const request =
+      process.env.NODE_ENV === 'test'
+        ? indexedDBtest.open('kubeconfigs', 1)
+        : indexedDB.open('kubeconfigs', 1);
+
+    // The onupgradeneeded event is fired when the database is created for the first time.
+    request.onupgradeneeded = handleDatabaseUpgrade;
+
+    // The onsuccess event is fired when the database is opened.
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,468 @@
...
+  return new Promise<string[]>(async (resolve, reject) => {
+    const request =
+      process.env.NODE_ENV === 'test'
+        ? indexedDBtest.open('kubeconfigs', 1)
+        : indexedDB.open('kubeconfigs', 1);
+
+    // The onupgradeneeded event is fired when the database is created for the first time.
+    request.onupgradeneeded = handleDatabaseUpgrade;
+
+    // The onsuccess event is fired when the database is opened.
```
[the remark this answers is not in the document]
knrt10: Done

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,468 @@
...
+      } else if (Object.keys(clustersToConfig).length !== Object.keys(statelessClusters).length) {
+        dispatch(setStatelessConfig({ ...configToStore }));
+      }
+    })
+    .catch((err: Error) => {
+      console.error('Error getting config:', err);
+    });
+}
+
+const exportFunctions = {
```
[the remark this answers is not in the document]
knrt10: I found that we were using export default "funtion name" throughout the headlamp code, so did it like that.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1334,7 +1351,19 @@ interface StreamParams {
   additionalProtocols?: string[];
 }
 
-function connectStreamWithParams(
+/**
+ *
+ * @param path
```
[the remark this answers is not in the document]
knrt10: Fixed

frontend/src/lib/k8s/apiProxy.ts
knrt10: Thanks, fixed it

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1352,11 +1382,16 @@ function connectStreamWithParams(
   }
 
   let fullPath = path;
+  let url = '';
   if (cluster) {
     fullPath = combinePath(`/${CLUSTERS_PREFIX}/${cluster}`, path);
+    const kubeconfig = await findKubeconfigByClusterName(cluster);
```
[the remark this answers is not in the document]
knrt10: Make sense, fixed it

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1375,7 +1410,10 @@ function connectStreamWithParams(
     if (!socket) {
       return;
     }
-    socket.close();
+
+    if (socket) {
```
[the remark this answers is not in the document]
knrt10: Yep, make sense. Fixed it. Thanks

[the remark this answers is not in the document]
knrt10: I am sorry I don't understand. We are only closing the socket when close function is called and there is a socket.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,478 @@
...
+    // result is the cursor. It is used to iterate through the object store.
+    result: IDBCursorWithValue | null;
+  };
+}
+
+/**
+ * KubeconfigObject is the object that is stored in indexDB as string format.
+ * It is a JSON encoded version of the kubeconfig file.
+ * It is used to store the kubeconfig for stateless clusters.
+ * This is basically a k8s client - go Kubeconfig object.
```
illume, line 86: A link to the documentation in here for this would be good. https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/

There you will also find some field documentation. Please consider if there's anything that should be changed compared to your descriptions. Maybe it's better to just use what they provide.

There are some fields which are missing here compared to what the documentation says. The missing fields should be added here, or a `@todo` note added mentioning that they are missing.

frontend/src/stateless/index.ts
illume: The user type is documented here: https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#AuthInfo

Please add that type, or document with a `@todo` comment that it needs to be improved (with the above URL).

plugins/examples/dynamic-clusters/README.md
```diff
@@ -2,10 +2,23 @@
...
+### Static Clusters (Default):
+
+- If dynamic cluster is disabled (default setting), Headlamp prompts users to manually enter information about the cluster, including the cluster name and server details.
+- Users can submit the form, and Headlamp will load the cluster based on the provided information.
+
+<img width="584" alt="dynamic cluster is disabled" src="https://github.com/headlamp-k8s/headlamp/assets/24803604/8b67f739-dd41-4114-9f2f-93f8cf2bbf0a">
+
+### Dynamic Clusters (Enable with -enable-dynamic-clusters flag):
+
+- When dynamic cluster is enabled on the backend using the -enable-dynamic-clusters flag, Headlamp allows users to paste the base64 encoded value of their kubeconfig.
```
illume, line 16: btw, what is the reason it is base64 encoded?

plugins/examples/dynamic-clusters/src/index.tsx
```diff
@@ -1,24 +1,102 @@
...
-          })
-          .catch(e => {
-            console.log('Error setting cluster', e);
-          });
-      }}
-    >
-      New cluster
-    </Button>
+        .catch((e: any) => {
+          console.error('Error setting cluster', e);
```
illume, line 42: Could you please display this error to the user?

That will let me test the error handling more easily.

frontend/src/stateless/index.ts
```diff
@@ -0,0 +1,478 @@
...
+interface CursorSuccessEvent extends Event {
+  // target is the request that generated the event.
+  target: EventTarget & {
+    // result is the cursor. It is used to iterate through the object store.
+    result: IDBCursorWithValue | null;
+  };
+}
+
+/**
+ * KubeconfigObject is the object that is stored in indexDB as string format.
```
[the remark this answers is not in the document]
knrt10: Added

frontend/src/stateless/index.ts
knrt10: Make sense, done

frontend/src/stateless/index.ts
knrt10: Done, modified it as per k8s documentation

plugins/examples/dynamic-clusters/README.md
knrt10: For plugin we needed a way to directly give the kubeconfig in encoded format so that it can be used at a string. But for future we plan to have a drop down or upload file directly, so that user does not have to encode it manually. I hope this is okay for now.

plugins/examples/dynamic-clusters/src/index.tsx
knrt10: Updated, but please remember, when stateless cluster is enabled, there won't be any response from the backend, because everything is done on the frontend. I did however have added checks to validate kubeconfig.

backend/cmd/headlamp.go
```diff
@@ -937,6 +952,12 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 
 	for _, context := range contexts {
 		context := context
+
+		// Stateless clusters should not be visible to other users.
```
yolossn, line 956: Lets call it Dynamic for consistency.
```suggestion
		// Dynamic clusters should not be visible to other users.
```

yolossn (commented):
Just few small NITs apart from this looks good to me.

backend/cmd/headlamp.go
```diff
@@ -951,10 +972,62 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+// parseClusterFromKubeConfig parses the kubeconfig and returns a list of contexts and errors.
+func (c *HeadlampConfig) parseClusterFromKubeConfig(kubeConfigs []string) ([]Cluster, []error) {
```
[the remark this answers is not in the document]
knrt10: Thanks, make sense

frontend/src/components/App/Layout.tsx
```diff
@@ -91,8 +88,14 @@ export default function Layout({}: LayoutProps) {
...
   const clusters = useTypedSelector(state => state.config.clusters);
   const { t } = useTranslation();
+  const allClusters = useClustersConf();
   const clusterInURL = getCluster();
 
+  /** This fetches the cluster config from the backend and updates the redux store on an interval.
+   * When stateless clusters are enabled, it also parses the stateless cluster config from the
+   * backend and updates the redux store on an interval. The parsed config is fetched from the
+   * indexDB.
+   * */
```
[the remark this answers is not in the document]
knrt10: Aah missed it. Nice catch

yolossn (approved):
Backend is LGTM. Please get approval from @illume before merging.

illume (approved):
👍 Nice work @knrt10

frontend/src/lib/k8s/apiProxy.ts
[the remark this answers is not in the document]
knrt10: Aah, now I understand. I thought you meant removing socket.close(). Updated. Thanks

Issue #1265, opened 2023-07-20
Feature request: Winget support


clarkezone:
Winget is the built in package manager for windows that ships natively in win11.  Would love a winget package for the headlamp desktop app on windows.

https://learn.microsoft.com/en-us/windows/package-manager/winget/

joaquimrocha:
This done BTW. Closing.

PR #1325, opened 2023-08-14
app: Creates winget manifest

vyncent-t:
For issue #1265 

## Description

This PR addresses the request to support `winget`, the built-in package manager for Windows, especially since it's natively available in Windows 11. The goal is to streamline the installation experience for Windows users of the Headlamp desktop app.

## Changes

1. **Winget Manifest**: Introduced a new `winget` manifest folder tailored for the Headlamp desktop app. This manifest will allow users to install Headlamp via the `winget install headlamp` command.

2. Followed the official `winget` documentation to ensure the manifest adheres to recommended guidelines and best practices.

3. Updated installation documentation (if applicable) to include instructions for the `winget` installation method.

## Testing

- [x] Validate the `winget` manifest using the official `winget` CLI tools.

### DIFF
--- a/app/windows/winget/templates/Headlamp.Headlamp.locale.en-US.yaml
+++ b/app/windows/winget/templates/Headlamp.Headlamp.locale.en-US.yaml
@@ -1,0 +1,27 @@
+PackageIdentifier: Headlamp.Headlamp
+PackageVersion: __PACKAGE_VERSION__
+PackageLocale: en-US
+Publisher: Headlamp
+PublisherUrl: https://headlamp.dev
+PackageName: Headlamp
+License: Apache 2.0
+LicenseUrl: https://github.com/headlamp-k8s/headlamp/blob/main/LICENSE
+ShortDescription: Kubernetes web UI.
+Description: A Kubernetes web UI that is fully-featured, user-friendly and extensible.
+Tags:
+  - kubernetes
+  - plugins
+  - kinvolk
+  - headlamp
+  - dashboard
+  - ui
+  - web
+  - monitoring
+  - logging
+  - cncf
+Documentations:
+  - DocumentLabel: headlamp documentation
+    DocumentUrl: https://headlamp.dev/docs/latest
+ReleaseNotesUrl: __RELEASE_NOTES_URL__
+ManifestType: defaultLocale
+ManifestVersion: 1.5.0
--- a/app/windows/winget/winget-create.js
+++ b/app/windows/winget/winget-create.js
@@ -1,0 +1,140 @@
+/**
+ * Usage:
+ *
+ * 1. Ensure that Node.js and npm are installed on your system.
+ * 2. Run this script using the command `node winget-create.js <projectVersion> <directoryPath>`.
+ *
+ * Parameters:
+ * - projectVersion: The version of the project you want to fetch information for (e.g., 0.19.1).
+ * - directoryPath: The path to the directory where the generated YAML files should be stored (e.g., app/windows/winget/manifests).
+ *
+ * Ensure that you replace `<projectVersion>` with the actual project version, do not include the 'v'.
+ *
+ * Note:
+ * - The templates directory should contain the necessary YAML templates and should be located in the same directory as this script.
+ * - The manifests directory will be created if it does not exist, and it will contain the generated YAML files.
+ */
+
+const fetch = require('node-fetch');
+const fs = require('fs');
+const path = require('path');
+
+const args = process.argv.slice(2);
+
+if (!(args.length === 2)) {
+  console.error('Usage: node winget-create.js <projectVersion> <directoryPath>');
+  process.exit(1);
+}
+
+const projectVersion = args[0];
+const directoryPath = args[1];
+
+if (projectVersion.startsWith('v')) {
+  console.error('Please provide the project version without the "v" prefix.');
+  console.error('Example: node winget-create.js 0.19.1 my/path/to/folder');
+  process.exit(1);
+}
+
+async function createYAMLFiles(version, downloadURL, date, notesURL, checksum) {
+  const templates = [
+    'Headlamp.Headlamp.installer.yaml',
+    'Headlamp.Headlamp.locale.en-US.yaml',
+    'Headlamp.Headlamp.yaml',
+  ];
+
+  const templateDir = path.join(__dirname, 'templates/');
+  const manifestDir = path.join(directoryPath, version);
+
+  try {
+    // create manifest version directory if it doesnt exist
+    if (!fs.existsSync(manifestDir)) {
+      fs.mkdirSync(manifestDir, { recursive: true });
+      console.log(`Created directory ${manifestDir}`);
+    }
+  } catch (err) {
+    console.error(`Error creating directory ${manifestDir}: ${err}`);
+    process.exit(1);
+  }
+
+  templates.forEach(template => {
+    const manifestPath = path.join(manifestDir, template);
+    const templatePath = path.join(templateDir, template);
+    const templateContent = fs.readFileSync(templatePath, 'utf8');
+
+    const newContent = templateContent
+      .replaceAll('__PACKAGE_VERSION__', version)
+      .replaceAll('__DISPLAY_VERSION__', version)
+      .replaceAll('__INSTALLER_SHA256__', checksum)
+      .replaceAll('__INSTALLER_URL__', downloadURL)
+      .replaceAll('__RELEASE_DATE__', date)
+      .replaceAll('__RELEASE_NOTES_URL__', notesURL);
+
+    fs.writeFileSync(manifestPath, newContent);
+    console.log(
+      `Created winget manifest file ${template} for version ${version} in ${manifestDir}`
+    );
+  });
+}
+
+// Async function to fetch release information from GitHub API for a given tag
+async function fetchGithubReleaseInfo() {
+  const tag = projectVersion;
+  const res = await fetch(
+    `https://api.github.com/repos/headlamp-k8s/headlamp/releases/tags/v${tag}`
+  );
+
+  if (!res.ok) {
+    console.error(
+      `Error fetching release information for tag ${tag}: ${res.status} ${res.statusText}`
+    );
+    process.exit(1);
+  }
+
+  const resJSON = await res.json();
+
+  await updateReleaseInfo(resJSON, tag);
+}
+
+async function getChecksums(checksumsUrl, fileName) {
+  const res = await fetch(checksumsUrl);
+  const checksums = await res.text();
+
+  for (const line of checksums.split('\n')) {
+    const [checksum, filename] = line.split('  ');
+
+    if (!filename) {
+      continue;
+    }
+
+    if (filename === fileName) {
+      return checksum;
+    }
+  }
+
+  return '';
+}
+
+async function updateReleaseInfo(response, tag) {
+  const assets = response.assets;
+  const checksums = assets.find(v => v.name === 'checksums.txt');
+  const checksumsUrl = checksums.browser_download_url;
+  // asset for windows x64, can add more assets for other archs
+  const winx64Asset = assets.find(asset => asset.name === `Headlamp-${tag}-win-x64.exe`);
+  const browserDownloadName = winx64Asset.name;
+  const browserDownloadUrlInfo = winx64Asset.browser_download_url;
+  const checksum = await getChecksums(checksumsUrl, browserDownloadName);
+  const dateYYYYMMDDLength = 10;
+  const releaseDateInfo = response.published_at.slice(0, dateYYYYMMDDLength);
+  const releaseNotesUrlInfo = `https://github.com/headlamp-k8s/headlamp/releases/tag/v${tag}`;
+  const versionInfo = response.name;
+
+  await createYAMLFiles(
+    versionInfo,
+    browserDownloadUrlInfo,
+    releaseDateInfo,
+    releaseNotesUrlInfo,
+    checksum
+  );
+}
+
+fetchGithubReleaseInfo();
--- a/app/windows/winget/templates/Headlamp.Headlamp.installer.yaml
+++ b/app/windows/winget/templates/Headlamp.Headlamp.installer.yaml
@@ -1,0 +1,23 @@
+PackageIdentifier: Headlamp.Headlamp
+PackageVersion: __PACKAGE_VERSION__
+Installers:
+  - Architecture: x64
+    InstallerLocale: en-US
+    InstallerType: exe
+    InstallerUrl: __INSTALLER_URL__
+    InstallerSha256: __INSTALLER_SHA256__
+    Scope: user
+    InstallModes:
+      - interactive
+      - silent
+    InstallerSwitches:
+      Silent: /S
+      SilentWithProgress: /S
+    AppsAndFeaturesEntries:
+      - DisplayName: Headlamp
+        DisplayVersion: __PACKAGE_VERSION__
+        Publisher: Headlamp
+        InstallerType: exe
+    ReleaseDate: __RELEASE_DATE__
+ManifestType: installer
+ManifestVersion: 1.5.0
--- a/app/windows/winget/templates/Headlamp.Headlamp.yaml
+++ b/app/windows/winget/templates/Headlamp.Headlamp.yaml
@@ -1,0 +1,5 @@
+PackageIdentifier: Headlamp.Headlamp
+PackageVersion: __PACKAGE_VERSION__
+DefaultLocale: en-US
+ManifestType: version
+ManifestVersion: 1.5.0
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,38 +6,111 @@
     "": {
       "devDependencies": {
         "@types/jest": "^29.5.3",
-        "glob": "^10.3.3"
+        "glob": "^10.3.3",
+        "node-fetch": "^3.3.2"
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.22.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.5.tgz",
-      "integrity": "sha512-Xmwn266vad+6DAqEB2A6V/CcZVp62BbwVmcOJc2RPuwih1kw02TjQvWVWlcKGbBPd+8/0V5DEkOcizRGYsspYQ==",
+      "version": "7.22.13",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.13.tgz",
+      "integrity": "sha512-XktuhWlJ5g+3TJXc5upd9Ks1HutSArik6jf2eAjYFyIOf4ej3RN+184cZbzDvbPnuTJIUhPKKJE3cIsYTiAT3w==",
       "dev": true,
       "dependencies": {
-        "@babel/highlight": "^7.22.5"
+        "@babel/highlight": "^7.22.13",
+        "chalk": "^2.4.2"
       },
       "engines": {
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@babel/code-frame/node_modules/ansi-styles": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
+      "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
+      "dev": true,
+      "dependencies": {
+        "color-convert": "^1.9.0"
+      },
+      "engines": {
+        "node": ">=4"
+      }
+    },
+    "node_modules/@babel/code-frame/node_modules/chalk": {
+      "version": "2.4.2",
+      "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.2.tgz",
+      "integrity": "sha512-Mti+f9lpJNcwF4tWV8/OrTTtF1gZi+f8FqlyAdouralcFWFQWF2+NgCHShjkCb+IFBLq9buZwE1xckQU4peSuQ==",
+      "dev": true,
+      "dependencies": {
+        "ansi-styles": "^3.2.1",
+        "escape-string-regexp": "^1.0.5",
+        "supports-color": "^5.3.0"
+      },
+      "engines": {
+        "node": ">=4"
+      }
+    },
+    "node_modules/@babel/code-frame/node_modules/color-convert": {
+      "version": "1.9.3",
+      "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-1.9.3.tgz",
+      "integrity": "sha512-QfAUtd+vFdAtFQcC8CCyYt1fYWxSqAiK2cSD6zDB8N3cpsEBAvRxp9zOGg6G/SHHJYAT88/az/IuDGALsNVbGg==",
+      "dev": true,
+      "dependencies": {
+        "color-name": "1.1.3"
+      }
+    },
+    "node_modules/@babel/code-frame/node_modules/color-name": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/color-name/-/color-name-1.1.3.tgz",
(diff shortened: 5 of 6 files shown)


illume:
The docs for ["Create your package manifest"](https://learn.microsoft.com/en-us/windows/package-manager/package/manifest?tabs=minschema%2Cversion-example) contains docs for the fields.

vyncent-t:
I would like to mention a few fields that need some input on ways to be revised. ref file Headlamp.Headlamp.yaml

The single file manifest does validate with the fields currently written, open to input for additional metadata to include.

- PackageIdentifier
Required field, winget asks this to be the name of the app that matched the downloaded App info in windows add or remove program settings. Currently, it is set to `PackageIdentifier: Headlamp.Headlamp` as the convention for this follows `Publisher.App` but the app info after download is still Kinvolk, also not sure if to brand it under Microsoft or CNCF. 
image

- Publisher
Required field, will update this field once the discussion is made for the Publisher name choice.

app/windows/winget/0.19.0/headlamp.headlamp.installer.yaml
```diff
@@ -0,0 +1,53 @@
...
+        DisplayVersion:         	# Optional version displayed in the ARP entry
+        Publisher:              	# Optional publisher displayed in the ARP entry
+        ProductCode:            	# Optional product code of the installer
+        UpgradeCode:            	# Optional upgrade code of the installer
+        InstallerType:          	# Optional installer type
+    UnsupportedArguments:       	# Optional list of Windows Package Manager Client arguments the installer does not support
+      - UnsupportedArgument:    	# Optional unsupported Windows Package Manager Client argument
+    ReleaseDate: 2023-03-08                	# Optional release date
+ManifestType: installer
+ManifestVersion: 1.5.0
```
knrt10, line 53: Can you please end new line at the end.

joaquimrocha (changes requested):
Finally got to review this.
I think if the goal is to submit it to https://github.com/microsoft/winget-pkgs , then we should make these files work as templates, with placeholders like `__PKG_VERSION__` instead of a hardcoded version. And then we have script to fill the details like we have in `choco-bump.sh` and hopefully we can submit a PR every time Headlamp has a release.

Does this make sense?

vyncent-t:
This push changes the Headlamp.Headlamp.yaml file from singleton to version, also updates the triple file format for manifest

vyncent-t:
This push changes the Headlamp.Headlamp.yaml file from singleton to version, also updates the triple file format for manifest

joaquimrocha (changes requested):
This looks much cleaner now. Left a suggestion of a new tag.
This however needs to be turned into a template file somehow and we need a script to populate the template and generate the files (which should look like what you wrote in the end). See for example:
https://github.com/headlamp-k8s/headlamp/blob/main/app/windows/chocolatey/choco-bump.sh

With this, we can fork the manifests Github repo and create a PR over to the original repo every time we have a release.

vyncent-t:
The manifest was able to validate as a single file, currently working on fixing errors for validation of the new structure

error reads

Manifest validation failed.
Manifest Error: Schema Error. [Schema validation failed.
Error context: <root>[Installers][0][Platform] Description: Value type not permitted by 'type' constraint.
Error context: <root>[Installers][0] Description: Failed to validate against schema associated with property name 'Platform'.
Error context: <root>[Installers] Description: Failed to validate item #0 in array.
Error context: <root> Description: Failed to validate against schema associated with property name 'Installers'.
] File: Headlamp.Headlamp.installer.yaml
Manifest Error: Failed to process field. [Platform]

vyncent-t:
Removed the optional platform field from the installer file, this field would prevent the manifest from validating.

Manifest now validates fine

app/windows/winget/0.19.1/Headlamp.Headlamp.installer.yaml
```diff
@@ -0,0 +1,23 @@
...
+      Silent: /S
+      SilentWithProgress: /S
+    AppsAndFeaturesEntries:
+      - DisplayName: Headlamp
+        DisplayVersion: 0.19.1
+        Publisher: Headlamp
+        InstallerType: exe
+    ReleaseDate: 2023-08-24
+ManifestType: installer
+ManifestVersion: 1.5.0
```
illume, line 23: Does this file and the others need a new line at the end? I would guess so.

illume (commented):
Could you please write the command you used to validate it in the PR description somewhere?

Left a note about the missing new lines that github is warning about at the end of the files.

Note there is one conversation left unresolved. Could you please answer that conversation and close it if you have resolved it?

app/windows/winget/manifests/0.19.1/Headlamp.Headlamp.installer.yaml
joaquimrocha, line 1: This is not needed in this repo.

app/windows/winget/manifests/0.19.1/Headlamp.Headlamp.locale.en-US.yaml
joaquimrocha, line 1: This is not needed in this repo.

app/windows/winget/winget_create.sh
```diff
@@ -0,0 +1,55 @@
...
+#
+# Before running the script, ensure it has execute permissions by running:
+# chmod +x winget_create.sh
+#
+# Example of how to run the script:
+# ./winget_create.sh 0.19.1 "http://url.com" "123abc" "0.19.1" "2023-09-01" "http://release_notes_url.com"
+
+# Check if the correct number of arguments has been provided
+if [ "$#" -ne 6 ]; then
+    echo "Usage: $0 PACKAGE_VERSION INSTALLER_URL INSTALLER_SHA256 DISPLAY_VERSION RELEASE_DATE RELEASE_NOTES_URL"
```
joaquimrocha, line 22: All of these can be determined from the Headlamp windows exe, except for the release date.

joaquimrocha (changes requested):

vyncent-t:
- removed non generated manifest
- created js script for winget and removed sh script

Note: works similar to the old sh script, need to run the file using `node winget_create.js <version> <sha256>` as params

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,114 @@
...
+const fetch = require('node-fetch');
+
+const args = process.argv.slice(2);
+
+if (args.length < 2) {
+  console.error('Usage: node winget_create.js <projectVersion> <sha256>');
+  process.exit(1);
+}
+
+const projectVersion = args[0];
```
joaquimrocha, line 32: We should check if the given version has a `v` prefix or not, and adapt it accordingly or tell the user whether a v is needed.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * Usage:
+ *
+ * 1. Ensure that Node.js and npm are installed on your system.
+ * 2. In the project's root directory, initialize a package.json file with the command `npm init -y`.
+ * 3. Install the node-fetch module with the command `npm install node-fetch`.
+ * 4. Run this script using the command `node winget_create.js <projectVersion> <sha256>`.
+ *
+ * Parameters:
```
joaquimrocha, line 9: We need a destination directory parameter, so we write the new files into that.
I imagine we'll be using it like:
node winget_create.js 0.19.1 /my/path/to/the/wingetrepo/h/headlamp/0.19.1

The given directory should be created if it doesn't exist.

joaquimrocha (changes requested):
I think it's almost there, but it's missing something we had talked about.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,144 @@
...
+ * - The manifests directory will be created if it does not exist, and it will contain the generated YAML files.
+ */
+
+const fetch = require('node-fetch');
+const fs = require('fs');
+const path = require('path');
+
+const args = process.argv.slice(2);
+
+if (args.length > 2) {
```
joaquimrocha, line 24: What you want to check here if whether we have 2 args. I ran the script without arguments for getting some help and I get an exception.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,144 @@
...
+  ];
+
+  // use path to get the template directory
+  const templateDir = path.join(__dirname, 'templates/');
+
+  const manifestDir = path.join(directoryPath, '/', version, '/');
+
+  // create manifest version directory if it doesnt exist
+  if (!fs.existsSync(manifestDir)) {
+    fs.mkdirSync(manifestDir);
```
joaquimrocha, line 54: Use the recursive option of this function, as it creates the missing directory paths for the user.
We should inform the user as well that we've created the directory, to inform of ongoing changes.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,144 @@
...
+
+  return checksumMap;
+}
+
+async function updateReleaseInfo(response, tag) {
+  const assets = response.assets;
+
+  const checksums = assets.find(v => v.name === 'checksums.txt');
+
+  const checksumsUrl = checksums.browser_download_url;
```
joaquimrocha, line 119: Remove the empty newlines please.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,144 @@
...
+  const checksumMap = {};
+
+  for (const line of checksums.split('\n')) {
+    const [checksum, filename] = line.split('  ');
+
+    if (!filename) {
+      continue;
+    }
+
+    if (filename.includes(`.exe`)) {
```
joaquimrocha, line 106: This should not be a string literal (just a plane string).
It will also break if we ever add another arch to the windows installers. So maybe check for the full suffix which will guarantee we don't break: `x64.exe`.
Even better if you just declare the arch as a global constant and then use it where needed, so it's easy to change the script in the future.

app/windows/winget/winget_create.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const checksumMap = await getChecksums(checksumsUrl);
+
+  sha256 = checksumMap['x64'];
+
+  const browserDownloadUrlInfo = assets.find(
+    asset => asset.name === `Headlamp-${tag}-win-x64.exe`
+  ).browser_download_url;
+
+  const releaseDateInfo = response.published_at.slice(0, 10);
```
joaquimrocha, line 129: I am not sure the hardcoded length is robust enough.
Maybe it is, but this line should get a comment explaining why we slice.

app/windows/winget/winget_create.js
joaquimrocha, line 1: Please rename the file to `winget-create.js` in order to match the name format we've been using in e.g. app/scripts .

app/windows/winget/templates/Headlamp.Headlamp.locale.en-US.yaml
```diff
@@ -0,0 +1,27 @@
+PackageIdentifier: Headlamp.Headlamp
+PackageVersion: __PACKAGE_VERSION__
+PackageLocale: en-US
+Publisher: Headlamp
+PublisherUrl: https://headlamp.dev
+PackageName: Headlamp
+License: Apache 2.0
+LicenseUrl: https://github.com/headlamp-k8s/headlamp/blob/main/LICENSE
+ShortDescription: Kubernetes web UI.
+Description: Headlamp was created to be a Kubernetes web UI that has the traditional functionality of other web UIs/dashboards available (i.e. to list and view resources) as well as other features.
```
joaquimrocha, line 10: I know this was here before. I just wonder now if we should have something clearer, like we have in Github: `A Kubernetes web UI that is fully-featured, user-friendly and extensible`

joaquimrocha (approved):
Looking good! I have left a few more comments mainly so the script is more robust.

joaquimrocha (changes requested):
Mistakenly had marked it as approved.

app/windows/winget/winget_create.js
vyncent-t: Removed the slice so it will print the fill ISO formatted timestamp

app/windows/winget/winget_create.js
vyncent-t: added and also placed in a try catch to handle errors on creation

app/windows/winget/winget_create.js
vyncent-t: added!

app/windows/winget/winget_create.js
vyncent-t: fixed!

app/windows/winget/winget_create.js
vyncent-t: Done!

app/windows/winget/winget_create.js
vyncent-t: Done!

app/windows/winget/templates/Headlamp.Headlamp.locale.en-US.yaml
vyncent-t: Added!

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+
+const fetch = require('node-fetch');
+const fs = require('fs');
+const path = require('path');
+
+const args = process.argv.slice(2);
+
+if (!(args.length === 2)) {
+  console.error('Usage: node winget-create.mjs <projectVersion> <directoryPath>');
+  console.error('Please provide the project version without the "v" prefix.');
```
joaquimrocha, line 26: I think this error message is misplaced here.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  console.error('Example: node winget-create.js 0.19.1 my/path/to/folder');
+  process.exit(1);
+}
+
+const projectVersion = args[0];
+
```
joaquimrocha, line 32: Remove empty line.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  console.error('Please provide the project version without the "v" prefix.');
+  console.error('Example: node winget-create.js 0.19.1 my/path/to/folder');
+  process.exit(1);
+}
+
+let sha256;
+
+const arch = {
+  x64: 'x64',
+  x86: 'x86',
```
joaquimrocha, line 45: We only have one arch: 'x64'. You can keep it in an obj or array but let's not add more arches we don't deal with now.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+async function createYAMLFiles(version, downloadURL, date, notesURL) {
+  const templates = [
+    'Headlamp.Headlamp.installer.yaml',
+    'Headlamp.Headlamp.locale.en-US.yaml',
+    'Headlamp.Headlamp.yaml',
+  ];
+
+  // use path to get the template directory
```
joaquimrocha, line 55: This comment is not adding much.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  const templateDir = path.join(__dirname, 'templates/');
+
+  const manifestDir = path.join(directoryPath, '/', version, '/');
```
joaquimrocha, line 58: You don't need '/' with path.join

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+
+  try {
+    // create manifest version directory if it doesnt exist
+    if (!fs.existsSync(manifestDir)) {
+      console.log(`Created directory ${manifestDir}`);
+      fs.mkdirSync(manifestDir, { recursive: true });
```
joaquimrocha, line 64: These lines should be inverted: if mkdirSync fails, then we'd not see "Created directory". Otherwise we may misguide the user.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+    }
+  } catch (err) {
+    console.error(`Error creating directory ${manifestDir}: ${err}`);
+  }
+
+  templates.forEach(template => {
+    // read the template file
+    const templateContent = fs.readFileSync(`${templateDir}${template}`, 'utf8');
```
joaquimrocha, line 72: Ideally you'd join the templateDir and template with path.join .

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+    const newContent = templateContent
+      .replace(/__PACKAGE_VERSION__/g, version)
+      .replace('__DISPLAY_VERSION__', version)
+      .replace('__INSTALLER_SHA256__', sha256)
+      .replace('__INSTALLER_URL__', downloadURL)
+      .replace('__RELEASE_DATE__', date)
+      .replace('__RELEASE_NOTES_URL__', notesURL);
+
+    // write the new content to the manifest directory
+    fs.writeFileSync(`${manifestDir}${template}`, newContent);
```
joaquimrocha, line 84: Ditto for path.join here.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  for (const line of checksums.split('\n')) {
+    const [checksum, filename] = line.split('  ');
+
+    if (!filename) {
+      continue;
+    }
+
+    if (filename.includes(arch.x64)) {
+      checksumMap[arch.x64] = checksum;
+    }
```
joaquimrocha, line 124: There's a potential bug here. This code doesn't check whether you have a windows installer here, so if another platform was sorted after the Windows installer in the checksums.txt file, you'd get that checksum here instead of the window's file one.

I suggest that instead of gathering all this info into the map, you first get the file name as you do below and then use it when looking for the checksum. This way you guarantee you really get the checksum for the file you want.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  return checksumMap;
+}
+
+async function updateReleaseInfo(response, tag) {
+  const assets = response.assets;
+  const checksums = assets.find(v => v.name === 'checksums.txt');
+  const checksumsUrl = checksums.browser_download_url;
+  const checksumMap = await getChecksums(checksumsUrl);
+  const browserDownloadUrlInfo = assets.find(
+    asset => asset.name === `Headlamp-${tag}-win-x64.exe`
```
joaquimrocha, line 136: This generated file name can be used for a `getChecksum(checksumsURL, fileName)` function.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,153 @@
...
+  ).browser_download_url;
+  const releaseDateInfo = response.published_at;
```
joaquimrocha, line 138: So now you no longer slice the date info. How do you get it without the time part?

joaquimrocha (changes requested):
Found some more issues.

app/windows/winget/winget-create.js
vyncent-t: Ah okay, I tried to include it to match the original scripts object but where it could be used in the checksum piece!
Removed it!
image

app/windows/winget/winget-create.js
vyncent-t: Ah! I also copy pasted the top usage command but it was also misspelled for the .js file!

I tried to add more messages to provide more context as to not be too short. I felt the one line of just the usage message didn't explain enough about the correct way to use the command (it would be annoying for someone to have to "unlock" the hidden error comment about dropping the v when they could be told within their first mistake to be sure to include it)

I can remove it if its not practical!

app/windows/winget/winget-create.js
vyncent-t: Done!

app/windows/winget/winget-create.js
vyncent-t: I will remove all the notes I left in while learning how to make the script work from the original, they're basically to do notes anyway.

app/windows/winget/winget-create.js
vyncent-t: Okay, wasn't sure if I needed to keep them in or not, I had ran into a pathing issue while implementing the request for a custom file path parameter and it seemed to have been able to work once I put them both in

app/windows/winget/winget-create.js
vyncent-t: Apologies, I had left it in to see what it would print while the mkdir line is commented out!

app/windows/winget/winget-create.js
vyncent-t: ah okay, it seemed logical to use temps when using the manual '/' in path technique from earlier!

app/windows/winget/winget-create.js
vyncent-t: I thought by the original manual slice not being robust enough you meant it cut out too much information so I left it as the full ISO timestamp but I can add the slice back in if you no longer want the full timestamp or I can use split I think

app/windows/winget/winget-create.js
vyncent-t: Done!

app/windows/winget/winget-create.js
vyncent-t: Removed!

app/windows/winget/winget-create.js
vyncent-t: Added slice back in and const for context

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,151 @@
...
+  const manifestDir = path.join(directoryPath, version);
+
+  try {
+    // create manifest version directory if it doesnt exist
+    if (!fs.existsSync(manifestDir)) {
+      fs.mkdirSync(manifestDir, { recursive: true });
+      console.log(`Created directory ${manifestDir}`);
+    }
+  } catch (err) {
+    console.error(`Error creating directory ${manifestDir}: ${err}`);
```
joaquimrocha, line 61: There should be a `process.exit(1)` if something went wrong with creating the directory.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,151 @@
...
+
+async function updateReleaseInfo(response, tag) {
+  const assets = response.assets;
+  const checksums = assets.find(v => v.name === 'checksums.txt');
+  const checksumsUrl = checksums.browser_download_url;
+  const browserDownloadName = assets.find(
+    asset => asset.name === `Headlamp-${tag}-win-x64.exe`
+  ).name;
+  const browserDownloadUrlInfo = assets.find(
+    asset => asset.name === `Headlamp-${tag}-win-x64.exe`
```
joaquimrocha, line 131: This processes the assets list twice. Please get store the resolved asset into an `asset` variable. So you don't have to repeat the logic.

joaquimrocha (changes requested):

app/windows/winget/winget-create.js
vyncent-t: done!

app/windows/winget/winget-create.js
vyncent-t: done!

illume:
This needs the DCO fixed... `git commit --amend -s`

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,149 @@
...
+    process.exit(1);
+  }
+
+  templates.forEach(template => {
+    const manifestPath = path.join(manifestDir, template);
+    const templatePath = path.join(templateDir, template);
+    const templateContent = fs.readFileSync(templatePath, 'utf8');
+
+    const newContent = templateContent
+      .replace(/__PACKAGE_VERSION__/g, version)
```
joaquimrocha, line 71: This format is inconsistent with the rest. Can you make it `'__PACKAGE_VERSION__'`.

app/windows/winget/winget-create.js
```diff
@@ -0,0 +1,149 @@
...
+
+  for (const line of checksums.split('\n')) {
+    const [checksum, filename] = line.split('  ');
+
+    if (!filename) {
+      continue;
+    }
+
+    if (filename === fileName) {
+      checksumMap[arch.x64] = checksum;
```
joaquimrocha, line 117: There's no point in submitting a map when we only have one value.

joaquimrocha (changes requested):
I have reviewed it again. It's almost there so I will take it from here and update the script.

app/windows/winget/winget-create.js
vyncent-t: done!

joaquimrocha (approved):
It LGTM now and it works great!

PR #1330, opened 2023-08-16
frontend: test for repeat translate keys

### LINKED ISSUE #1300: Add a test for repeated translation keys across namespace files
Our i18n is namespaced. This has the caveat that if we use the wrong namespace for a translation string, we may get duplicated strings, multiplying the translation work.

We can avoid this by having a test for whether our language catalog has duplicated keys.

Acceptance Criteria:
- [ ] We have a test that will fail if there are duplicated keys in different translation files.
- [ ] We have a mechanism to say that certain words can actually be duplicated in certain files (this may be needed for certain nuances of languages)

### LINKED ISSUE #1426: Reduce the number of i18n namespaces (so it's easier to track and translate)
We have a big number of i18n namespaces and no easy way to determine sometimes what's in one or another.
This makes translations very difficult and is prone to errors like duplicated strings in different namespace files.

I propose having just a **glossary** namespace (to hold jargon/k8s references) and a all-the-rest namespace (called **translations**, to match the default i18next namespace that sometimes gets created by mistake).

If this turns out to be a lot for the UI to load in one go, we can split action related strings out of **translations**, into an **actions** namespace (to hold verbs/actions, like "Create", "Delete", "View", etc.).

vyncent-t:
# Testing for Duplicated Keys in Translation Files

fixes #1300
fixes #1426 

#### **Background**:
With our current approach to i18n, using namespaced translations could lead us into accidentally duplicating translation strings. This might increase the translation work unnecessarily.

#### **Solution**:
- Implement a testing mechanism to ensure we don't have duplicated keys across different translation files.
- Develop a way to whitelist specific keys that can be intentionally duplicated, taking into account language nuances.

#### **Changes**:
- [x] Introduced a test script that detects duplicated keys in all translation files.
- [x] Created a whitelist file where allowable duplicated keys can be defined.

#### **Next Steps**:
- Enhance the detection mechanism for more nuanced cases.
- Discuss and finalize keys that might need whitelisting due to language intricacies.


vyncent-t:
WIP - console logs the "original" key and its value and the duplicates
image

vyncent-t:
I wonder if for the mechanism we can utilize the context field for languages that may reuse words?
https://www.i18next.com/translation-function/context

vyncent-t:
WIP
Current commit now includes a way to whitelist words by manually inputting them into the whitelist.json file.

example:
image

when running test:
image

vyncent-t:
Commit note:
- removed example whitelist word
- added counter for test

frontend/src/i18n/locales/keys.test.ts
```diff
@@ -0,0 +1,85 @@
+import { error } from 'console';
+import fs from 'fs';
+import glob from 'glob';
+
+const whitelist = require('./whitelist.json');
```
illume, line 5: Can you please change this to allowlist?

frontend/src/i18n/locales/keys.test.ts
```diff
@@ -0,0 +1,85 @@
+import { error } from 'console';
```
illume, line 1: I feel there should be a description of this test that's a bit more 
- near the top
- descriptive when it fails. Currently the test is called "Translation keys", but it should give people a clue what is happening and what they need to do right away when it fails in the error message given in the failure. `npm run test` should describe the problem and what to do to fix it.

illume (changes requested):
Cool.

I added a few notes. Did you see there is a test failure?

frontend/src/i18n/locales/keys.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  /*
+   * Will for loop over each word in the master word list, if the lang does not exist as a key, create it, pair with an array and push to it
+   */
+  for (let i = 0; i < contents.length; i++) {
+    const lang = contents[i].lang;
+    if (wordMap[lang]) {
+      wordMap[lang].push(contents[i]);
+    } else {
+      wordMap[lang] = [contents[i]];
+    }
```
joaquimrocha, line 66: Nit: Usually, for readability, we check for the inexistence of what we need, and initialize it before proceeding with the regular logic.
```typescript
    if (!wordMap[lang]) {
      wordMap[lang] = [contents[i]];      
    }

    wordMap[lang].push(contents[i]);
```

frontend/src/i18n/locales/keys.test.ts
```diff
@@ -0,0 +1,107 @@
...
+   */
+  const contents: any[] = [];
+
+  for (const file of files) {
+    const content = await fs.readFileSync(file, 'utf-8');
+    const words = JSON.parse(content);
+
+    const filePart = file.split('/');
+    const language = filePart[filePart.length - 2];
+    const fileName = filePart[filePart.length - 1];
```
joaquimrocha, line 36: I think we should try to use:
https://nodejs.org/api/path.html#pathdirnamepath
https://nodejs.org/api/path.html#pathbasenamepath-suffix

frontend/src/i18n/locales/keys.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      if (!inAllowlist && knownWords[wordMap[lang][i].wordKey]) {
+        repeatKeyCount++;
+        error('Error: the key for this translation is already in use');
+        error(wordMap[lang][i]);
+        error('Error: translation key already in use by:');
+        error(knownWords[wordMap[lang][i].wordKey]);
+      } else {
+        knownWords[wordMap[lang][i].wordKey] = wordMap[lang][i];
+      }
```
joaquimrocha, line 95: Replace `wordMap[lang][i]` by `currentWord` for readability.

joaquimrocha (changes requested):
Left a couple of comments, mostly on readability.

joaquimrocha:
> I wonder if for the mechanism we can utilize the context field for languages that may reuse words? https://www.i18next.com/translation-function/context

Yes. This makes sense. Thanks.

frontend/src/i18n/locales/keys.test.ts
joaquimrocha, line 1: This and the allowlist.json files will lead to issues because they're placed in the `locales` folder, which is destined for locale folders only. Run `make i18n` to see what I mean.
Please move these files one level up (to `frontend/src/i18n`) and also I think this test file should be called `index.test.ts` as it's the main test we have for this i18n module.

joaquimrocha (changes requested):
Looking good but found a few issues running it.

joaquimrocha:
I have updated the test to fix a few corner cases (it was not discarding the "old" translations files, and it was not checking whether the same keys were being checked in the same file).
Then I added a commit to reduce the number namespaces, deduplicated translation strings, and added translations.

Maybe @illume can give an extra look.

illume:
Why are all the namespaces gone? This makes it much harder to get context on what the strings are for, and where to see the text in app.

joaquimrocha:
> Why are all the namespaces gone? This makes it much harder to get context on what the strings are for, and where to see the text in app.

It explained in #1426 .
Using multiple namespaces makes it very easy to use the wrong namespace and then we get duplicated strings for the same translation, which multiplies the translation work. It's also less convenient to translate strings in different files.
Besides, if context is what's needed for each string, then adding the context to the translation function makes it more convenient (`t('Desired', {context: 'pods'})`).

I believe maintaining the i18n as it is and trying to get anywhere close to 100% coverage represents a big effort that I find hard to justify. So this PR and the #1429 aim at simplifying the process. My concerns are whether having all these strings together represents a performance issue for the application UX, but even if cases of ambiguous translations arise, having it all together makes for an easier fix for those when they happen.

frontend/src/i18n/ThemeProviderNexti18n.tsx
```diff
@@ -23,7 +23,7 @@ function getLocale(locale: string): typeof enUS {
  */
 const ThemeProviderNexti18n: React.FunctionComponent<{ theme: Theme }> = props => {
   const { i18n, ready: isI18nReady } = useTranslation(
-    ['sidebar', 'frequent', 'glossary', 'auth', 'resource', 'notifications'],
+    ['sidebar', 'translation', 'glossary', 'auth', 'resource', 'notifications'],
```
illume, line 26: Some of these should be removed.

illume (approved):
👍 thanks

joaquimrocha (approved):

PR #1334, opened 2023-08-17
Fix automatic back link

joaquimrocha:
We use the browser's back function but only when we know that the user has navigated to more than one link. However, the logic to check this was off.

How to test:
- [ ] Open Headlamp in a pod's details view; click on a link from there (like its controller), then click the back link in that new page's details view: it should go back to the original pod's details view

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -165,5 +165,5 @@ export function PreviousRouteProvider({ children }: React.PropsWithChildren<{}>)
 
 export function useHasPreviousRoute() {
   const routeLevels = React.useContext(PreviousRouteContext);
-  return routeLevels > 1;
+  return routeLevels >= 1;
 }


vyncent-t:
I used this navigation path and it does route back to the details view

home > minikube link > workloads tab > pods tab > coredns pod link  > control link to replica set (on pod details page) > back link on replica set page > routes back to pod details page

vyncent-t (approved):

joaquimrocha:
> home > minikube link > workloads tab > pods tab > coredns pod link > control link to replica set (on pod details page) > back link on replica set page > routes back to pod details page

Unfortunately, this doesn't reproduce the issue in the broken version. The starting point has to be a pod details view, so the browser's context has no previous backlink, i.e. go to the pods' details but then copy&paste&load the URL again, so that's the first link the browser loads.

vyncent-t:
Was able to recreate the issue on the main branch and see that this branch fixes the issue

main:
coredns pod link > control link to replica set (on pod details page) > back link on replica set page > replica sets tab under workloads

this branch:
coredns pod link > control link to replica set (on pod details page) > back link on replica set page > back to pod detail page

PR #1340, opened 2023-08-21
frontend: Fix alignment for calendar Icon

vyncent-t:
## Description:
When navigating to a timestamp using the date and calendar option, there has been an alignment issue noticed. This fix ensures that the date and calendar display are properly aligned to enhance user readability and aesthetics.

## Changes:

- [x]  Adjusted CSS properties for date element to ensure proper alignment.

**Before**
image

**After**
image

### DIFF
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -161,7 +161,8 @@ export interface HoverInfoLabelProps {
 
 const useHoverInfoLabelStyles = makeStyles({
   display: {
-    display: 'inline-block',
+    display: 'flex',
+    justifyContent: 'flex-end',
   },
   noWrap: {
     whiteSpace: 'nowrap',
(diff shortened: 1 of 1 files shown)


joaquimrocha:
Testing this again, I noticed that it brings the label to the left, whereas it was aligned to the right before these changes. And this slight difference makes that column look misaligned. So this issue needs to be fixed.

joaquimrocha (changes requested):
The label needs to be aligned to right as before.

egomezbpedro:
Hi, 
this is due the property flex expanding all the way to the left. We could use justifyContent/justifySelf: 'flex-end' to bring the label to the original position without changing the element width or retain the inline-block display and use verticalAlign: 'middle' on the icon.

vyncent-t:
> Hi, this is due the property flex expanding all the way to the left. We could use justifyContent/justifySelf: 'flex-end' to bring the label to the original position without changing the element width or retain the inline-block display and use verticalAlign: 'middle' on the icon.

Thank you I applied fixes for this issue to properly align the label

**before**

image

**after**

image

joaquimrocha (approved):
Nice!

Issue #986, opened 2023-02-28
Fix the calendar icon color on dark theme


ashu8912:
In some of the places, the color of the calendar icon is still dark instead it should be white as shown below
How it is:
<img width="18" alt="Screenshot 2023-02-28 at 11 30 55 AM" src="https://user-images.githubusercontent.com/30126128/221767510-5a850dfe-aeab-471c-babd-22b8dcec87d5.png">
How it should be
<img width="26" alt="Screenshot 2023-02-28 at 11 30 41 AM" src="https://user-images.githubusercontent.com/30126128/221767527-e4f0b726-0b5d-4b72-ac1a-5fd5c61ab8e0.png">

xdevore:
Hello! I'm new to open source, but would love to try take on this bug if possible. If that would be alright, do you have any tips for potential solutions or getting started? Thanks :)

egomezbpedro:
Hi @ashu8912,

I've been searching through the frontend codebase for a possible explanation for this issue, but I've not found anything. I'm still getting familiar with the code. I first thought to be a misbehaviour of the theme library but so far is been a dead end. 

I'll appreciate any insights that you may have on where to look for this behavior. 

Cheers

![Dark Theme Wrong](https://github.com/headlamp-k8s/headlamp/assets/57415533/f87ec522-34b0-44b6-b517-6e314d9a7378)
Dark Theme Correct
![Dark Theme Correct](https://github.com/headlamp-k8s/headlamp/assets/57415533/3fe3d57b-2bc8-45c4-8398-f3340ce9d135)
Light Theme Correct
![Light Theme Correct](https://github.com/headlamp-k8s/headlamp/assets/57415533/69932b33-fe8b-42c0-b096-5e47ea0b59b1)
Light Theme Wrong
![Light Theme Wrong](https://github.com/headlamp-k8s/headlamp/assets/57415533/1be3f7b9-2861-4df7-8d6c-07b030b1d075)

ashu8912:
Hello @egomezbpedro first of all welcome to headlamp codebase :D,
Ok so about the issue It's an issue that is there in only some views like if you go to the namespace list view the calendar icon contrast there is not correct in dark theme and then the detail view for it also has the same issue, 
So this issue can be solved in two parts where we first find all the places where this is happening and then adding a pr that fixes this.

If you need any help with this i am more than happy to help.

egomezbpedro:
Hi @ashu8912,

[...]

PR #1343, opened 2023-08-22
Normalize the color of the calendar icon

egomezbpedro:
# Issue Related: #986

# Description: 
In some pages the `mdi:calendar` icon is displayed with a wrong color in both dark and light themes.

# Changes: 
- [x]  Force the unset color property on the icon.

**Before:**
Screenshot 2023-08-22 at 08 45 14

**After:**
Screenshot 2023-08-22 at 08 47 28

### DIFF
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -169,6 +169,7 @@ const useHoverInfoLabelStyles = makeStyles({
   icon: {
     marginRight: '0.2rem',
     marginLeft: '0.2rem',
+    color: 'unset !important',
   },
 });
 


joaquimrocha:
This looks fine, and it also fixes the icon's color in the dark theme.
I am a bit concerned that we're adding that `!important` to it, which probably prevents any further theming, but let's deal with that later as this change right now improves the usability and accessibility of the date label.

Graciñas @egomezbpedro !

egomezbpedro:
Hi @joaquimrocha,

Overall I'm not to happy about the fix, I'm sure that there is another solution without the !important flag. I'll continue to check on this. For the mean time thanks for the lgtm.

Obrigado, ;)

PR #1345, opened 2023-08-22
frontend ingress: Fix showing ingresses with different kinds of backend

### LINKED ISSUE #1323: Headlamp goes blank when viewing Ingresses due to port number being undefined
When I open Ingresses menu, the Headlamp sitesgoes blank and show errors regarding the port number.
It was showing **TypeError: Cannot read properties of undefined (reading 'toString')** for the *port* constant.

```
function RulesDisplay(props: { ingress: Ingress }) {
  const { ingress } = props;

  const rulesText = React.useMemo(() => {
    const rules = ingress.getRules();
    let labels: string[] = [];

    rules.forEach(({ http }) => {
      const text = http.paths.map(({ path, backend }) => {
        const service = backend.service.name;
        const port = backend.service.port.number; <~ This is the problem
        return `${path} 🞂 ${!!service ? service + ':' + port.toString() : port.toString()}`;
      });
      labels = labels.concat(text);
    });

    return labels;
  }, [ingress]);

  return <LabelListItem labels={rulesText} />;
}
```

I assume that it's because almost all of my backend is using **port.name** instead of **port.number**
```
backend:
    service:
      name: headlamp
      port:
        name: http
```

It also shows an error in API stream but I'm not sure if it's the consequence of the preceeding error
```
"/apis/networking.k8s.io/v1/ingresses?watch=1&resourceVersion=5220024"
```

joaquimrocha:
A backend for an Ingress resource may represent a service with a port number, a port name, or it can be a resource reference.

This patch fixes displaying an Ingress for different types of backend and adds also stories to it.

fixes #1323

### DIFF
--- a/frontend/src/components/ingress/Details.stories.tsx
+++ b/frontend/src/components/ingress/Details.stories.tsx
@@ -1,0 +1,42 @@
+import { Meta, Story } from '@storybook/react/types-6-0';
+import Ingress, { KubeIngress } from '../../lib/k8s/ingress';
+import { TestContext } from '../../test';
+import Details from './Details';
+import { PORT_INGRESS, RESOURCE_INGRESS } from './storyHelper';
+
+export default {
+  title: 'Ingress/DetailsView',
+  component: Details,
+  argTypes: {},
+  decorators: [
+    Story => {
+      return (
+        <TestContext>
+          <Story />
+        </TestContext>
+      );
+    },
+  ],
+} as Meta;
+
+interface MockerStory {
+  ingressJson?: KubeIngress;
+}
+
+const Template: Story = (args: MockerStory) => {
+  const { ingressJson } = args;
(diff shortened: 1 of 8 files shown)


illume (approved):
👍 thanks.

Cool... I see there's tests for when port is not there, and name is used instead.

illume:
## How to test?

- follow this tutorial https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/
- change the example-ingress to use `name: http` instead of `port: 8080`.

```yaml
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$1
spec:
  rules:
    - host: hello-world.info
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: web
                port:
                  name: http
```

PR #1355, opened 2023-08-25
load-tests: Add beginning of load testing scripts, document load testing

illume:


load-tests/README.md
joaquimrocha, line 1: Should we call this folder just tests, then an intro section in the README redirecting the user to the docs for other tests in Headlamp?

joaquimrocha: Should this actually be a section in docs, and just a link here to it?

load-tests/scripts/create-events.js
joaquimrocha, line 1: I think we should show the current context of the kubeconfig to the user as ask for confirmation, unless the user passes a `--force` flag. Because running this against a normal k8s can be... a delicate op 🙂

joaquimrocha (commented):
Left a couple of suggestions.

ashu8912 (approved):
I tested the scripts and they work fine, and the code also LGTM, just the comments from @joaquimrocha and i think this is good to go

load-tests/README.md
illume: I added a testing doc, and pointed the readme there.

load-tests/scripts/create-events.js
illume: Good point. I had only a warning in the docs... but to make it stronger I added an explicit check for the context to always be kwok. So it should be safer now by default.

PR #1357, opened 2023-08-25
frontend Notifications/List: Correct status for context buttons

joaquimrocha:
The mark-as-seen and the clear buttons were still enabled when there were no notifications. This patch fixes that and also optimizes how the check for the status is being computed.

How to test:
(make sure you copy the notifications from the browser's local storage so you can repeat the case where you have unseen and visible notifications)
- [ ] Go to the Notifications area (sidebar Notifications); then set all notifications as seen -> it works as expected
- [ ] When all notifications are seen/read, verify that the mark-all-as-seen button is disabled in the menu popup
- [ ] Go to the Notifications area (sidebar Notifications); then clear all notifications -> it works as expected
- [ ] When there are no notifications (all are cleared), verify that the clear-all button is disabled in the menu popup

### DIFF
--- a/frontend/src/components/App/Notifications/List.tsx
+++ b/frontend/src/components/App/Notifications/List.tsx
@@ -1,6 +1,6 @@
 import { Icon } from '@iconify/react';
 import { Box, IconButton, Menu, MenuItem, Tooltip, Typography, useTheme } from '@material-ui/core';
-import { useState } from 'react';
+import React, { useState } from 'react';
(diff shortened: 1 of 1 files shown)


vyncent-t:
Notifications are cleared when clearing all button is clicked, then button is disabled, marked all as read and button is also disabled, works as mentioned!

vyncent-t (approved):

PR #1360, opened 2023-08-28
Use a phony backend token when run from Makefile

joaquimrocha:
We use the make file to run Headlamp in dev mode, and since the backend token was not set up in that case, it was impossible to use this mode for testing any backend features, i.e. running npm run dev-only-app would not work for testing adding clusters, use Helm, etc.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -793,7 +793,10 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	// On dev mode we're loose about where connections come from
 	if config.devMode {
-		headers := handlers.AllowedHeaders([]string{"X-Requested-With", "Content-Type", "Authorization", "Forward-To"})
+		headers := handlers.AllowedHeaders([]string{
+			"X-HEADLAMP_BACKEND-TOKEN", "X-Requested-With", "Content-Type",
+			"Authorization", "Forward-To",
+		})
 		methods := handlers.AllowedMethods([]string{"GET", "POST", "PUT", "HEAD", "DELETE", "PATCH", "OPTIONS"})
 		origins := handlers.AllowedOrigins([]string{"*"})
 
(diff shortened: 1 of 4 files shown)


backend/cmd/headlamp.go
```diff
@@ -793,7 +793,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	// On dev mode we're loose about where connections come from
 	if config.devMode {
-		headers := handlers.AllowedHeaders([]string{"X-Requested-With", "Content-Type", "Authorization", "Forward-To"})
+		headers := handlers.AllowedHeaders([]string{"X-HEADLAMP_BACKEND-TOKEN", "X-Requested-With", "Content-Type", "Authorization", "Forward-To"})
```
illume, line 796: There's a lint issue here.

frontend/src/helpers/index.ts
```diff
@@ -358,7 +358,9 @@ function loadTableSettings(tableId: string): { id: string; show: boolean }[] {
  *
  * The app also passes the token to the headlamp-server via HEADLAMP_BACKEND_TOKEN env var.
  */
-const backendToken = new URLSearchParams(window.location.search).get('backendToken');
+const backendToken =
+  process.env.REACT_APP_HEADLAMP_BACKEND_TOKEN ||
```
illume, line 362: Hrmm.

This puts the token into the compiled javascript,
so maybe it's not so safe to do it this way.

I'm trying to think through all the cases where someone might set that variable. Maybe it's ok.

joaquimrocha: Can you explain how this puts the token into the compiled js?

backend/cmd/headlamp.go
joaquimrocha: Fixed.

illume (approved):
👍 thanks

Issue #1329, opened 2023-08-16
Missing information about Persistent Volume Claims in pod description


eric-intuitem:
in a pod view, Headlamp indicates volumes name and type, e.g. "db-data  persistentVolumeClaim". However, the name of the PVC is not indicated.
It would be useful to show the name as a dynamic link to the corresponding PVC.

PR #1361, opened 2023-08-28
frontend: Improve pod persistent volume detail

vyncent-t:
# Improve information for volumes under pod details page

Fixes issue #1329 

## Description
This update introduces a new section in the Pod Details view dedicated to displaying volume information. Now, dynamic links are provided enabling seamless navigation within the platform wherever applicable. Additionally, storybook elements have been added to visualize the new volume section, aiding in better understanding and further development.

## Changes
- [x] Added a new section in the Pod Details view for Volumes, displaying relevant volume information.
- [x] Enabled dynamic links for names, facilitating navigation to the corresponding information within the platform.
- [x] Introduced storybook elements for the new volume section, aiding in visualization and further refinement.

image


illume (changes requested):
Added a couple of notes.

Would you be able to let me know how I can test this?

joaquimrocha:
I was supposed to clarify here how we are supposed to update this view. Sorry for the delay and here it is:
Let's have a Volumes section where each volume will use a `NameValueTable` whose first row is the name of the volume (similar to `ContainerInfo`).

Each of these tables, besides the 1st row with the name of the volume should have:
* Type: The typeof this volume (ConfigMap, hostPath, projected, ...), see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#volume-v1-core
* Others: Rows depending on the kind

For the "Others" mentioned above, they will be different depending on the type, so we should have likely a table renderer per kind, like: `VolumeConfigMapRenderer`, `VolumeHostPathRenderer`, etc.
Since it's nearly impossible to cover all cases, we can show the plain YAML for the keys within the type value. If these keys are a string, boolean, or a number, just render their value as is. If it's an object/array, then just convert (YAML) to a string (indented/pretty) and display it.

Hope this makes it clearer.

vyncent-t:
WIP NOTE: still deciding if to keep the rendering table to its own component or to use current logic / something similar to render detail data

- currently working on a fix for better YAML prints (ref: VolumeDetailRender component)

- rendering the details through their own component render table could allow easier changes to an Innertable (already experimented with) but shows problems with rendering as either inner table squished or all others data squished to the right

vyncent-t:
WIP NOTE: (continued...)
**Added working YAML print that I forgot, still needs polish**

image

vyncent-t:
commit note:
fixes bug where detail would print as empty with a label when string is set to ""

joaquimrocha:
It's coming up nice!
I think we should change Type to Kind, to avoid the clash with the other "type". WDYT?

vyncent-t:
commit note: 
**- data and yaml render working**
- removed old component file
- changed default label to keyword "Kind"

image

vyncent-t:
Note: will remove the console logs during the next review

illume:
Could you please add a story for the VolumeSection component?

(probably this one will need the signoff as well)

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
     </SectionBox>
   );
 }
+
+export interface VolumeSectionProps {
+  volume?: any | null;
```
illume, line 900: If possible, please don't use `any`. It's in effect turning off the type checker.

vyncent-t:
Commit
- Created a story for volume section

frontend/src/components/common/Resource/Resource.tsx
vyncent-t: I removed some uses of any where I could, but I don't know of a way to include coverage for each kind of volume within the list
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#volume-v1-core

 I did try to make some props based on examples rendered below
image

vyncent-t: Improved the previous bit with a new KubeVolume

vyncent-t:
commit note
- issue fixed where the volume for rendering and volume used for links shared no fields

this bit is still broken and needs to be fixed
image

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+
+  const volumes = resource?.spec?.volumes;
+
+  if (!volumes) {
+    return null;
+  }
+
+  const namespace = resource?.metadata?.namespace;
+
+  const { t } = useTranslation();
```
joaquimrocha, line 927: Hooks cannot be conditionally used. This is the case in this function because depending on the value of volumes, this hook is either used or not. (Please move the hooks to the first lines, maybe after the props expansion)

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+
+    const volumeDetails: any = volumeDetailGrab(currentVolume[volumeType]);
+
+    function yamlPrinter(yamlData: any) {
+      return (
+        <Typography component="pre" variant="body2">
+          {YAML.stringify(yamlData)}
+        </Typography>
+      );
+    }
```
joaquimrocha, line 1018: I think it's a bit hard to read the function having all these sub-functions scattered around the code.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+              };
+            }),
+          ]
+        : []),
+    ];
+  }
+
+  return (
+    <SectionBox title={t('frequent|Volumes')}>
+      {volumes.map((volume: any) => (
```
joaquimrocha, line 1062: Volume should not be any.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+                value: yamlPrinter(detail.volValue),
+              };
+            }),
+          ]
+        : []),
+    ];
+  }
+
+  return (
+    <SectionBox title={t('frequent|Volumes')}>
```
joaquimrocha, line 1061: When you rebase this PR. Please remove the namespace unless this word belongs to the glossary namespace.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+  /*
+   * printVolumeLink will print a working link that is set within the router using fields from the resource as params
+   */
+  function printVolumeLink(volumeName: string, volume: PrintVolumeLinkProps) {
+    const { persistentVolumeClaim, configMap } = volume;
+
+    if (persistentVolumeClaim) {
+      return (
+        <Link
+          routeName="persistentVolumeClaim"
```
joaquimrocha, line 938: The way to get these routeNames, should be to try to correspond the resource type here with the classes in `ResourceClasses` and then getting its detailsRoute. Those classes are give us whether they are namespaced or not.

frontend/src/lib/k8s/pod.ts
```diff
@@ -19,6 +24,7 @@ export interface KubePodSpec {
   readinessGates?: {
     conditionType: string;
   }[];
+  volume: KubeVolume[];
```
joaquimrocha, line 27: This property is optional and it's called `volumes`. After this change, the error with the storyHelper should be fixed.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,175 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+  }
+
+  const namespace = resource?.metadata?.namespace;
+
+  const { t } = useTranslation();
+
+  /*
+   * printVolumeLink will print a working link that is set within the router using fields from the resource as params
+   */
+  function printVolumeLink(volumeName: string, volume: PrintVolumeLinkProps) {
```
illume, line 932: Because these functions return components, they should be components (with capital letters, and taking a props as an argument).

frontend/src/components/common/Resource/Resource.tsx
illume: VolumeSectionProps['volume']

vyncent-t:
refactored the code and implemented fixes from comments

vyncent-t:
I ran the npm i18n to try to fix the error for the new word "volumes" being added but it seems not to want to take it?

vyncent-t:
found a bug where clicking on a secret source link may not work 
it also undone the fix for optional field somehow?
image

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,190 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+        const volumeInstance = new VolumeClass(volume);
+        const volumeRoute = volumeInstance.detailsRoute;
+        const volumeNamespace = volumeInstance.isNamespaced ? namespace : null;
+        // we use the keys of the volume[volumeKind], which is the non class created volume item, to find the name of the volume source
+        const volumeKeys = Object.keys(volume[volumeKind]);
+        let detailName;
+
+        // where the volume[volumeKind] contains a "name" word in the keys, grab that value
+        for (const key of volumeKeys) {
+          if (key.includes('name')) {
```
joaquimrocha, line 947: This is a very implicit and we cannot guarantee that the name you find this way is the right name of the volume kind.
So we should make this explicit: maybe having an object corresponding a volume kind with the name key, like
```typescript
const volumeKindNames = {
  configMap: 'name',
  secret: 'secretName',
};
```

This way we know exactly what is supposed to work and what is not.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,190 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+    }
+
+    let volumeName = currentVolume.name;
+    let volumeKind: string = '';
+
+    const volumeKeys = Object.keys(currentVolume);
+
+    for (const key of volumeKeys) {
+      if (key === 'name') {
+        volumeName = currentVolume[key];
```
joaquimrocha, line 981: You already extracted the volume name on line 974.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,190 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+      } else {
+        volumeKind = key;
```
joaquimrocha, line 983: Assuming that volumes is always `{ name, KIND_WHATEVER_IT_IS: {...} }`
We can simplify this logic by:
```typescript
const { name, ...objWithVolumeKind } = volume;
const volumeKind = Object.keys(objWithVolumeKind)[0] || '';
```

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,190 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+      for (const key in vol) {
+        if (!(vol[key] === '')) {
+          if (
+            typeof vol[key] === 'string' ||
+            typeof vol[key] === 'number' ||
+            typeof vol[key] === 'boolean'
+          ) {
+            directPrint.push({
+              volKey: key,
+              volValue: typeof vol[key] === 'boolean' ? JSON.stringify(vol[key]) : vol[key],
```
joaquimrocha, line 1004: I think that for converting the `vol[key]` to string, it's more correct to do `vol[key].toString()` instead of the `JSON.stringify`.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,190 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+
+    for (const key of volumeKeys) {
+      if (key === 'name') {
+        volumeName = currentVolume[key];
+      } else {
+        volumeKind = key;
+      }
+    }
+
+    function volumeDetailGrab(vol: VolumeSectionProps['volume']): VolumePrints {
```
joaquimrocha, line 987: I was not immediately show what `volumeDetailGrab` did because of its name. Why not `printVolumeDetails`?

joaquimrocha (changes requested):
Gave it another look. Thanks.

vyncent-t:
removed the console logs and edit spacing

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,187 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+                <Link
+                  routeName={volumeRoute}
+                  params={{ namespace: volumeNamespace, name: detailName }}
+                >
+                  {volumeName}
+                </Link>
+              );
+            }
+          }
+        }
```
joaquimrocha, line 969: Unless I misunderstood it, I think the logic would be simpler if you check directly whether the volumeKindName mapped key exists in the volume definition. Something like:
```typescript
let volumeNameKey = volumeKindNames[volumeKind as keyof typeof volumeKindNames];
if (!!volumeNameKey) {
  let detailName = volume[volumeKind][volumeNameKey];
  if (!!detailName) {
    return (
      <Link .... 
    );
  }
}
```

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -894,3 +895,187 @@ export function ConditionsSection(props: { resource: KubeObjectInterface | null
...
+  /*
+   * printVolumeLink will print a working link that is set within the router using fields from the resource as params
+   */
+  function PrintVolumeLink(props: PrintVolumeLinkProps) {
+    const { volumeName, volumeKind, volume } = props;
+    const resourceClasses = ResourceClasses;
+    const classList = Object.keys(ResourceClasses);
+
+    for (const kind of classList) {
+      if (kind.toLowerCase() === volumeKind.toLowerCase()) {
```
joaquimrocha, line 938: You don't need to loop when you can just check if you have the class there:
```typescript
let resourceClass = ResourceClasses[volumeKind];
if (!!resourceClass ) {
  ....
}
```

joaquimrocha (changes requested):

vyncent-t:
I have tried to rebase onto the current main and met a few issues.

There seems to be an issue originating from the cronjob details file (unrelated to the changes in this branch) which this branch does not update for some reason, the commits were made a few days ago but it does not update on rebase?

image

Somehow the issue from the volumes keyword has returned, could be a similar issue?
image

vyncent-t:
I was able to fix the first error where it wouldnt update the changes made to another file, but the volumes keyword error still remains

joaquimrocha (approved):
I have re-run the i18n so it doesn't delete the keys that were translated before + rebased.

PR #1394, opened 2023-09-15
Improve Custom Resources UX

### LINKED ISSUE #1393: Add dedicated list reviews for Custom Resources kinds
As reported in #955, we need to improve the UX around CRDs.
The suggested approach of having a 3rd level in the sidebar has some related UX issues as reported in #1302, but we can definitely provide an alternative that improves the current status:

Since we're showing already the kinds (as Name in the CRD list) and the actual CRD names (as Full Name, in the CRD list). We can rename these columns to Resource, and Definition, respectively. The Definition values (full name of the CRD) will link to the current CRD details view. The values in the Resource column will link to a new Custom Resource (not Definition) list view, where all the CRs of that kind will be listed, together with the additional printer columns.

cc/ @gberche-orange

Acceptance Criteria:
- [ ] Go to Cluster > CRDs: There's a list of the CRDs with the Resource and Definition columns as described above
- [ ] Click on a link in the Resource column: you're taken to the list of resources/objects for that kind (this is a new view)
- [ ] The new Resource list view shows the resources and respective additionalPrinterColumns. Details of how this looks is TDB
- [ ] Click on a resource from the resource list: you're taken to the CR details view (this view is the same we had already, accessible from the objects table in the CRD details view)
- [ ] Click on a link in the Definition column: you're taken to the CRD details view (this view is the same we had already)

joaquimrocha:
fixes #1393

We now have:
* Name CRDs in the sidebar is now Custom Resources
Sidebar with Custom Resources label

* The list view associated with the sidebar entry has now a link to Resources and Definitions:
New CRD list view

* Clicking on a Resource in the view above, will show the list view for those resources (if the CR spec has additional columns, these are shown too)
Screenshot for new CR list

* Clicking on an object in the Resource list view will show that object's details view
Screenshot for CR details view
  * This details view now has a Definition value, allowing to see the CRD association
* In the CRD details view (which existed before), now we have a new Resource value, allowing to access the Resource list view

Screenshot for CRD details view

### DIFF
--- a/frontend/src/components/Sidebar/__snapshots__/Sidebar.stories.storyshot
+++ b/frontend/src/components/Sidebar/__snapshots__/Sidebar.stories.storyshot
@@ -480,7 +480,7 @@ exports[`Storyshots Sidebar/Sidebar In Cluster Sidebar Closed 1`] = `
                               <span
                                 class="MuiTypography-root MuiListItemText-primary MuiTypography-body1 MuiTypography-displayBlock"
                               >
-                                CRDs
+                                Custom Resources
                               </span>
                             </div>
                             <span
@@ -1605,7 +1605,7 @@ exports[`Storyshots Sidebar/Sidebar In Cluster Sidebar Open 1`] = `
                               <span
                                 class="MuiTypography-root MuiListItemText-primary MuiTypography-body1 MuiTypography-displayBlock"
                               >
-                                CRDs
+                                Custom Resources
                               </span>
                             </div>
                             <span
@@ -2777,7 +2777,7 @@ exports[`Storyshots Sidebar/Sidebar Selected Item With Sidebar Omitted 1`] = `
                               <span
                                 class="MuiTypography-root MuiListItemText-primary MuiTypography-body1 MuiTypography-displayBlock"
                               >
-                                CRDs
+                                Custom Resources
                               </span>
                             </div>
                             <span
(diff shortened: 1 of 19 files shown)


ashu8912 (approved):
Tested, LGTM

ashu8912:
@joaquimrocha maybe update the comment with the new screenshot where link for definition is aligned on right

gberche-orange:
great, thanks @joaquimrocha !

Some quick feedback on the UX:
* would be great to surface the current CR in the navigation as to allow easy browsing to other CRs of the same type
   * on the left colum: CRD -> Cronjob -> my-new-cronjob-object
   * on the top of screen (next to the `back` button)
* the [object detail view](https://user-images.githubusercontent.com/1029635/269042402-67cd74fd-2644-40b4-aa7f-78e6cdde85c6.png) should display the fields in the additional columns as well as an option to display the full details (spec and status)

joaquimrocha:
Hi @gberche-orange .
Adding the kind of CRs as entries in the sidebar may mean showing dozens of them and this would make the sidebar look awkward and make navigation difficult. Having them as a 3rd level, and have the group be the sub-entries of the root-level (Custom Resources > org.mygroup.name > MyKind) would scale better but there are other UX issues as I mentioned. Like what to do when you collapse the sidebar.

> the [object detail view](https://user-images.githubusercontent.com/1029635/269042402-67cd74fd-2644-40b4-aa7f-78e6cdde85c6.png) should display the fields in the additional columns as well as an option to display the full details (spec and status)

We do show the additional columns as extra items in that table. This particular CronTab just doesn't have any.
I have to double check about the status.

PR #1396, opened 2023-09-19
frontend common/Label: Do not end-align HoverInfoLabel

joaquimrocha:
The HoverInfoLabel was changed to have its contents aligned to the
end (right, in LTR languages) as the date label used should be aligned
as such and uses this component internally. However, many other values
are represented with this label, so we cannot always align it.

This patch fixes that by making this component have an inline-flex
display, allowing for the labels+icons to be vertically aligned to the
center (reason for which it had a display:flex rule) while respecting
the text-align rule set on a wrapping element. i.e. the date labels get
aligned to the right in most tables as their cell has that text-align
rule, but other uses of the mentioned component don't get affected.

**Before the fix:**
Screenshot showing the issue

**After the fix:**
Screenshot showing the issue being fixed

### How to test:
- [ ] Go to the details view of a Custom Resource (object) that has extra columns (values in the metadata display grid that have an info icon): verify that these are not aligned to the right in the row (screenshots above)
- [ ] Go to any resource list: verify the date column is still aligned to the right

### DIFF
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -161,8 +161,7 @@ export interface HoverInfoLabelProps {
 
 const useHoverInfoLabelStyles = makeStyles({
   display: {
-    display: 'flex',
-    justifyContent: 'flex-end',
+    display: 'inline-flex',
   },
   noWrap: {
     whiteSpace: 'nowrap',


vyncent-t (approved):

vyncent-t:
This fixes both the extra info alignment and the date

PR #1403, opened 2023-09-22
Fix standard name CRDs

joaquimrocha:
When applying a resource, we tried to get a match for a known/default resource (CronJob, Pod, etc.) but this may result in trying to apply a resource using the wrong API when we have a CustomResource that has a name matching any exiting default resource's. This patch always tries to get the API details from the server for the API version of the data we need to apply, thus ensuring we always get the right API.

How to test:
- [ ] Set up a CronTab CRD from the [example](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) in the official k8s docs; then try to apply the YAML for a custom CronTab `*`: it should not work on `main`, but should work with this PR

`*`:
```yaml
apiVersion: "stable.example.com/v1"
kind: CronTab
metadata:
  name: my-new-cron-object
spec:
  cronSpec: "* * * * */5"
  image: my-awesome-cron-image
  replicas: 5
```

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -11,12 +11,14 @@ import Registry, * as registryToExport from './registry';
 
 window.pluginLib = {
   ApiProxy: require('../lib/k8s/apiProxy'),
-  Crd: require('../lib/k8s/crd'),
   ReactMonacoEditor: require('@monaco-editor/react'),
   MonacoEditor: require(process.env.NODE_ENV === 'test'
     ? 'monaco-editor/esm/vs/editor/editor.api.js'
     : 'monaco-editor'),
   K8s: require('../lib/k8s'),
+  // Anything that is part of the lib/k8s/ folder should be imported after the K8s import, to
+  // avoid circular dependencies' issues.
+  Crd: require('../lib/k8s/crd'),
   CommonComponents: require('../components/common'),
   MuiCore: require('@material-ui/core'),
   MuiStyles: require('@material-ui/styles'),
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -14,7 +14,6 @@ import helpers, { getHeadlampAPIHeaders, isDebugVerbose } from '../../helpers';
 import store from '../../redux/stores/store';
 import { getToken, logout, setToken } from '../auth';
 import { getCluster } from '../util';
-import { ResourceClasses } from '.';
 import { KubeMetadata, KubeMetrics, KubeObjectInterface } from './cluster';
 import { KubeToken } from './token';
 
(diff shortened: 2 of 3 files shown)


illume:
Seems these tests are failing...

```js
 FAIL  src/plugin/filterSources.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'className')

      80 |   // Ideally this should just be the class name, but until we ensure the class name is consistent
      81 |   // (in what comes to the capitalization), we use this lazy approach.
    > 82 |   const className: string = cls.className.charAt(0).toUpperCase() + cls.className.slice(1);
         |                                 ^
      83 |   resourceClassesDict[className] = cls;
      84 | });
      85 |

      at forEach (src/lib/k8s/index.ts:82:33)
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (src/lib/k8s/index.ts:79:11)
      at Object.<anonymous> (src/lib/k8s/crd.ts:1:1)
      at Object.<anonymous> (src/plugin/index.ts:14:8)
      at Object.<anonymous> (src/plugin/filterSources.test.ts:2:1)

 FAIL  src/plugin/updateSettingsPackages.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'className')

      80 |   // Ideally this should just be the class name, but until we ensure the class name is consistent
      81 |   // (in what comes to the capitalization), we use this lazy approach.
    > 82 |   const className: string = cls.className.charAt(0).toUpperCase() + cls.className.slice(1);
         |                                 ^
      83 |   resourceClassesDict[className] = cls;
      84 | });
      85 |

      at forEach (src/lib/k8s/index.ts:82:33)
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (src/lib/k8s/index.ts:79:11)
      at Object.<anonymous> (src/lib/k8s/crd.ts:1:1)
      at Object.<anonymous> (src/plugin/index.ts:14:8)
      at Object.<anonymous> (src/plugin/updateSettingsPackages.test.ts:2:1)

```

frontend/src/plugin/index.ts
```diff
@@ -11,12 +11,14 @@ import Registry, * as registryToExport from './registry';
...
   ApiProxy: require('../lib/k8s/apiProxy'),
-  Crd: require('../lib/k8s/crd'),
   ReactMonacoEditor: require('@monaco-editor/react'),
   MonacoEditor: require(process.env.NODE_ENV === 'test'
     ? 'monaco-editor/esm/vs/editor/editor.api.js'
     : 'monaco-editor'),
   K8s: require('../lib/k8s'),
+  // Anything that is part of the lib/k8s/ folder should be imported after the K8s import, to
+  // avoid circular dependencies' issues.
+  Crd: require('../lib/k8s/crd'),
```
illume, line 21: Ah Ok!

illume:
- I re-ran the failing "Build Container" job several times (which can sometimes be flaky). But it keeps failing... so maybe it's a real issue, or is something on the internet currently broken? It's the later. I ran the job on main and it's failing there too.
- I tested as per instructions above, and it works for me.

illume (approved):
thanks 👍

PR #1409, opened 2023-09-27
headlamp-plugin: Add .vscode folder for vscode integration

illume:
This includes some task runners for building and testing.
It also suggests some extensions (prettier and eslint).

### DIFF
--- a/plugins/headlamp-plugin/template/.vscode/extensions.json
+++ b/plugins/headlamp-plugin/template/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
+}
--- a/plugins/headlamp-plugin/template/.vscode/settings.json
+++ b/plugins/headlamp-plugin/template/.vscode/settings.json
@@ -1,0 +1,17 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "editor.formatOnSave": true,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "[javascript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[javascriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
+}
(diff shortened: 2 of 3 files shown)


plugins/headlamp-plugin/template/.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
```
joaquimrocha, line 4: I thought we had agreed to move with a Headlamp extension that declares dependencies or recommended extensions, so we have a single point to be updated and makes the UX nicer for the user?

plugins/headlamp-plugin/template/.vscode/settings.json
```diff
@@ -0,0 +1,26 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "files.exclude": {
+    "build": false // set this to true to hide the "dist" folder with the compiled JS files
+  },
+  "search.exclude": {
+    "build": true // set this to false to include "dist" folder in search results
+  },
+  // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
+  // "typescript.tsc.autoDetect": "off",
```
joaquimrocha, line 10: Should this setting be commented out?

plugins/headlamp-plugin/template/.vscode/extensions.json
illume: Yes, when there is an "extension pack" one, it can be added here.
The `headlamp-plugin upgrade` will handle an update of this file.

I'd like to get the .vscode files added first so I can get some feedback on this part first.

plugins/headlamp-plugin/template/.vscode/settings.json
illume: I've removed it. Doesn't seem to be the case that we need it as off. I wasn't sure before.

joaquimrocha: It's still here, right?

plugins/headlamp-plugin/template/.vscode/settings.json
```diff
@@ -0,0 +1,26 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "files.exclude": {
+    "build": false // set this to true to hide the "dist" folder with the compiled JS files
```
joaquimrocha, line 4: So this hides this folder from the file tree in VSCode? I am not sure I'd like this change for me as a developer. As long as it doesn't get in my way in git (i.e. it's in .gitignore), I'm fine and find it useful to have things available in the file tree. Same logic for node_modules for example.

joaquimrocha (changes requested):

plugins/headlamp-plugin/template/.vscode/settings.json
illume: removed this

plugins/headlamp-plugin/template/.vscode/settings.json
illume: It's removed now. (Seems I didn't push before. )

illume:
@joaquimrocha it's been updated to remove those things you noted.

joaquimrocha (approved):

PR #1412, opened 2023-09-28
frontend: Fix regression when applying resources

joaquimrocha:
We started using the "apis" prefix when applying resources but need also to distinguish when "api" is needed.

How to test:
- [ ] Apply a non-grouped resource like [Pod](https://kubernetes.io/docs/concepts/workloads/pods/): it should allow applying the resource
- [ ] Apply a grouped resource like [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/): it should allow applying the resource
- [ ] Apply a [CR whose name matches a default one:](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) it should allow applying the resource


illume (approved):
👍 good catch

PR #1413, opened 2023-09-28
backend: Decrompress data from external-proxy when needed

joaquimrocha:
Otherwise the frontend gets the compressed data when it assumes it's JSON content.

This resulted in the app details (app-catalog) never being loaded, as it wanted JSON and was given compressed data. Maybe decompressing in the client would be better, but I am not sure if that's feasible.

How to test:
Run Headlamp with the app-catalog plugin and click on an app in the Apps page: it goes to the app's details and should stop loading and show the contents of that app.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"crypto/tls"
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


backend/cmd/headlamp.go
```diff
@@ -491,7 +492,18 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
 			return
 		}
-		respBody, err := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
+
+		// Check that the server actually sent compressed data
+		var reader io.ReadCloser
+		switch resp.Header.Get("Content-Encoding") {
+		case "gzip":
+			reader, err = gzip.NewReader(resp.Body)
```
illume, line 501: Was gonna say... err needs checking. But the linter beat me to it.

illume (approved):
👍

PR #1423, opened 2023-09-29
Fix mutating webhook config list

joaquimrocha:
- frontend: Use MutatingWebhookConfigList for its route
- frontend: Add missing MutatingWebhookConfig details view to the router


ashu8912 (approved):
Tested, Works for me

PR #1425, opened 2023-09-29
Fix resource double requests

### LINKED ISSUE #1398: BUG: Double requests
It seems like after latest headlamp update something related to requests went wrong, since requests are repeated.

image

Steps to reproduce:
Go to any list page and open network tab(disable cache).

joaquimrocha:
fixes #1398 

The double requests were triggered by how `useCluster` was being initialized to `null` when it didn't need to.

@callmevladik, maybe you can test it?


illume (approved):
🎉 🎈

PR #1427, opened 2023-10-02
headlamp-plugin 0.7.1

illume:
- headlamp-plugin: Bump version v0.7.0
- headlamp-plugin: Fix so vscode template files are added on upgrade
- headlamp-plugin: Bump version v0.7.1
- plugins/examples: Bump headlamp-plugin 0.7.1 and upgrade examples

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -575,7 +575,13 @@ function upgrade(packageFolder, skipPackageUpdates) {
    * Assumes we are in the package folder.
    */
   function addMissingTemplateFiles() {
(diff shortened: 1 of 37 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha, line 1: I wonder if we should switch to using hardlinks instead of copying files (if those are supported on Windows).

joaquimrocha (approved):
Left a comment on moving from copying files, but the changes LGTM.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: Something to think about.

PR #1443, opened 2023-10-09
Add action to open PR for updating chart automatically

joaquimrocha:
There's been a few times where we've forgotten to update the Headlamp version for its Helm Chart, so this PR aims at preventing that by opening the PR automatically.

### DIFF
--- a/.github/workflows/pr-to-update-chart.yml
+++ b/.github/workflows/pr-to-update-chart.yml
@@ -1,0 +1,67 @@
+name: PR to update Charts
+
+# This action will run automatically after the workflow "Publish container" is completed
+on:
+  workflow_run:
+    workflows: ["Publish container"]
+    types:
+      - completed
+  workflow_dispatch:
+
+env:
+  LATEST_HEADLAMP_TAG: latest
+jobs:
+  create_pr_to_upgrade_chart:
+    name: Create PR to update Headlamp's version in the Helm Chart
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout headlamp repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Configure Git
+        run: |
+          user=${{github.actor}}
+          if [ -z $user ]; then
+            user=joaquimrocha
+          fi
+          git config --global user.name "$user"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-to-update-chart.yml
```diff
@@ -0,0 +1,65 @@
+name: PR to update Charts
+
+# This action will run after a tag starting with "v" is published
+on:
+  registry_package:
+    types: [published]
```
yolossn, line 6: I am not sure if `registry_package` published is the right trigger as this action can get triggered any time a container image is published. Like even if we push an image with `staging` tag this event will get triggered.

.github/workflows/pr-to-update-chart.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Update Headlamp version in charts
+        run: |
+          if [ -z $user ]; then
+            user=joaquimrocha
+          fi
+          # Extract current version
+          CURRENT_CHART_VERSION=$(grep '^version:' ./charts/headlamp/Chart.yaml | cut -d' ' -f2)
+          CURRENT_CHART_VERSION=$(echo $CURRENT_CHART_VERSION | awk -F. -v OFS=. '{$(NF - 1) += 1 ; print}')
+          # Create branch
+          BRANCH=update_chart_headlamp_$HEADLAMP_VERSION
```
yolossn, line 42: `$HEADLAMP_VERSION` is not set anywhere, perhaps you wanted this to be `$LATEST_HEADLAMP_TAG`?

.github/workflows/pr-to-update-chart.yml
joaquimrocha: So you suggest we do this with manual triggering only?

.github/workflows/pr-to-update-chart.yml
```diff
@@ -0,0 +1,67 @@
+name: PR to update Charts
+
+# This action will run after a tag starting with "v" is published
```
yolossn, line 3: Since we don't check the tag of the container

```suggestion
# This action runs on successful completion of "Publish container" action
```

PR #1446, opened 2023-10-10
frontend: clusterActionSlice: Refactor clusterActions into a slice

illume:
- Moved clusterAction related stuff from reducers/actions/sagas into clusterActionSlice.
- Added some documentation.
- Added some tests.

## How to test?

- Create or edit some yaml, and it should apply.
- Try cancelling an action with the snackbar cancel buttons.

Part of https://github.com/headlamp-k8s/headlamp/issues/1159 epic.


illume:
Rebased against main.

illume:
> Why are sagas removed?

Because redux-toolkit provides a way to do async stuff now, and it seems better to remove the saga dependency and use the one the redux folks decided on.

> I have been thinking we should move request related code into sagas, to ensure they don't delay the UI. I would like to understand how this code is analogous to what sagas provided. i.e. aren't we opening the door to a slower UI given we removed some things from sagas?

I don't think it's slower. The requests are async and don't block the UI.

illume:
Rebased against main.

illume:
Rebased against main.

joaquimrocha (changes requested):
Code LGTM but testing it I have noticed that the Cancel snackbar doesn't go away when I click the cancel button.
How to reproduce:
1. Go to a pod, click the delete button;
2. After a second or two, click the delete button in the snackbar that appears. -> That snackbar should go away and the only the one about the having been cancelled should be on for a few more secs.

illume:
I can't seem to reproduce it... Could you perhaps leave a gif showing how it's done?

This is what I was trying based on your description:
deleting a pod

joaquimrocha:
@illume I think your video does show the issue.
See the [main branch's behavior](https://github.com/headlamp-k8s/headlamp/assets/1029635/a8f3bdae-01e7-4637-bc97-8e1cbbfb11c5) vs this [branch's behavior](https://github.com/headlamp-k8s/headlamp/assets/1029635/54f37421-1b62-4a71-a6ca-2f01461dec07).

illume:
Thanks.

I've updated it so the cancel snackbar leaves right away when the cancel button is pressed, like before.

joaquimrocha (approved):
Works now. LGTM.

Issue #1354, opened 2023-08-25
Open details views for resources in a drawer


joaquimrocha:
We've had requests for the details views to be opened in a drawer (pane sliding in from the right side) instead of being shown as a different view.
This allows the user to keep the context of the list they've been using, instead of driving them away.

I think that we can implement this feature as an option though, to be off by default, so users can keep the default/familiar behavior for Headlamp if they prefer it.

vyncent-t:
moved into blocked - awaiting design feedback

PR #1447, opened 2023-10-10
frontend: Add detail drawer mode

vyncent-t:
# Introduce Detail Drawer Mode in Headlamp

For issue #1354 

## Description
This PR introduces a new 'Detail Drawer' mode to Headlamp, providing users with the option to view details in a side drawer. This enhances the user interface by allowing for a more flexible and efficient way of displaying detailed information without navigating away from the current view.

## Features
- [x] Implemented the Detail Drawer mode.
- [x] Added toggle functionality to enable/disable the Detail Drawer view in settings.
- [x] Ensured seamless integration of the side drawer with existing UI components.

## Changes
- [x] Refactored existing Link component to be drawer compatible.
- [x] Refactored `FetchAndExecutePlugins` into a hook to use for getting plugin data.

## Verification
- Verified that the detail information is accurately displayed in the drawer.
- Checked the responsiveness and consistency of the drawer across different screen sizes and orientations.


vyncent-t:
Commit Notes
- Fixes broken switch in settings, used to only turn enable/disable once due to reading from localStorage and not global state
- Fixed detail drawer enabled mode from setting back to default, improved local storage usage fixed it

vyncent-t:
Note
- fixes the issue where changes made would not run with new UI

vyncent-t:
note
- fixes translation
- need to fix label *

vyncent-t:
Work note

### Progress
- Pro: When the drawer is on open mode, you can now click the button which will minimize it to the closed mode off to the right
- Pro: Able to see behind drawer with no backdrop

### Issues
- Issue: When the drawer is in the Closed Mode, there is no pushing the content to the middle of the screen, it sits on top of the right of the page and not next to it
- Issue: The main content is still unable to be clicked on, which could be good since you can bet focused on clicking links in the main menu?
- Issue: When refreshing the page on the settings view, the drawer mode is set to enabled and open by default, it needs to be taking the state of the last made changes
    - Currently if you turn off the drawer mode it will be turned on automatically when refreshing the settings page
- Issue: The "flat" drawer is not rendering anything inside it 
    - it also does not stay on the screen
    - It leaves when you move pages

vyncent-t:
Work note

### Progress
- [x] Pro: When the drawer is on open mode, you can now click the button which will minimize it to the closed mode off to the right
- [x] Pro: Able to see behind drawer with no backdrop

### Issues
- [ ] Issue: When the drawer is in the Closed Mode, there is no pushing the content to the middle of the screen, it sits on top of the right of the page and not next to it
- [ ] Issue: The main content is still unable to be clicked on, which could be good since you can bet focused on clicking links in the main menu?
- [ ] Issue: When refreshing the page on the settings view, the drawer mode is set to enabled and open by default, it needs to be taking the state of the last made changes
    - [ ] Currently if you turn off the drawer mode it will be turned on automatically when refreshing the settings page
- [x] Issue: The "flat" drawer is not rendering anything inside it 
    - [x] it also does not stay on the screen (fixed to where it will remain on screen when close allowing to navigate to other pages)
    - [ ] It leaves when you move pages

vyncent-t:
Commit fix pushed

- [x] Issue: When clicking the enable button the box is not matching the enable (FIXED, now when clicking the enable drawer mode on or off the drawer render will match)

vyncent-t:
image
- [ ] Issue: When the drawer is in the Closed Mode the top bar content is squished

vyncent-t:
- [x] Placeholder button now implemented, but not keeping the same shape when on settings page

vyncent-t:
- [x] Issue: Drawer state is reset when refreshing the page (FIXED, now pulling and setting open status from local storage for drawer mode and open drawer)

vyncent-t:
### ISSUE

- [ ] Currently the iframes render within another iframe, need actual content within frame

illume:
- we need to get the design considered more I think
  - What should the text be in the settings, and the drawer name? 
  - Should there be a button to leave the drawer open at all? Maybe it should be always open, with no option to toggle it (except in settings).
- probably we need a URL for the view with the drawer open. So if someone copies the URL to someone else they would have the same view.

vyncent-t:
Push note
- Enabled `isDetailDrawerOpen` again in order to get the drawer to open and close when clicking on backdrop
* NOTE: Backdrop must be enabled in order to be clickable as hidebackdrop does not render it thus not allowing the click off

- Fixed the drawer being reset on refresh by implementation of `isDetailDrawerOpen`

- Implemented `useMediaQuery` from material ui to make drawer responsive, now will only render on width with enough screen real-estate

vyncent-t:
Working on new issue:
- Fix list view for list that dont use `ResourceTableProps`
    - currently only the lists with similar structure to Pod list that use `ResourceTableProps` will be easily configured to be drawer compatiable, it seems other reasources use `ResourceListView` that use the `WorkloadDetails` component in the router to render details view

vyncent-t:
Changed the git commit style to include changes made

vyncent-t:
Feedback notes

- [ ] The drawer should have no backdrop and allow users to click on contents of the page behind the drawer
    - This can be done by use of the "persistent" field rather than the default temporary
    - The page view should NOT be squished to fit both the drawer and the main content, the drawer should be appearing and disappearing from the right


- [ ] The drawer should be persistent as a sperate component and not tied to the list render
    - This means users should be able to view a pod details drawer, click to a new resource but still allow the pod details drawer to exist
 
- [ ] When the screen is small the drawer should render in from the bottom
- [ ] The drawer should feature a Close button
- [ ] The drawer should be able to render different views within the drawer while not redirecting pages

vyncent-t:
last push fixes drawer open close on pods with no need for button

vyncent-t:
## Feedback notes

### WIP
- [ ] The drawer should be able to render different views within the drawer while not redirecting pages

### To do
- [ ] When the screen is small the drawer should render in from the bottom

### Solved

 **(SOLVED using persistent)**
- [X] The drawer should have no backdrop and allow users to click on contents of the page behind the drawer 
    - This can be done by use of the "persistent" field rather than the default temporary
    - The page view should NOT be squished to fit both the drawer and the main content, the drawer should be appearing and disappearing from the right

---

 **(SOLVED: detached list view from being the parent of the drawer, list view now provides link information, _you must include the drawerEnabled prop and reformat the name column if you wish to use drawer mode for that link_**
image
- [x] The drawer should be persistent as a sperate component and not tied to the list render
- This means users should be able to view a pod details drawer, click to a new resource but still allow the pod details drawer to exist

---

 **(SOLVED: now features close button)**
- [x] The drawer should feature a Close button

vyncent-t:
Demo video

https://github.com/headlamp-k8s/headlamp/assets/78232183/f8dd40eb-6bfa-425f-8b45-9b32d9e9fb21

vyncent-t:
not sure if I had the same changes as the video made a push just to update

vyncent-t:
last push
- fix logic for drawerEnabled being in PureLink
- clicking drawer links now updates the page URL with no navigation

vyncent-t:
last push
- solves errors in storybook tests that needed the missing providers for redux

vyncent-t:
last push:

I think this fixes the typescript error for kubeObject<any> & void return

image


image

vyncent-t:
@sniok review ready

frontend/src/components/App/Layout.tsx
```diff
@@ -109,6 +110,9 @@ export default function Layout({}: LayoutProps) {
    * indexDB and then sends the backend to parse it and then updates the parsed value into redux
    * store on an interval.
    * */
+  // DETAIL DRAWER MODE
```
sniok, line 113: `// DETAIL DRAWER MODE` comment is not necessary and this new line is inserted between the useEffect and the comment related to useEffect, please move this line or move the comment above

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useDispatch } from 'react-redux';
+import { setDetailDrawerEnabled } from '../../../redux/drawerModeSlice';
+import { useTypedSelector } from '../../../redux/reducers/reducers';
+
+export default function DrawerModeButton() {
+  const dispatch = useDispatch();
+  const { t } = useTranslation('translation');
+  // This will fix the problem of the project refreshing the state away from needed position
```
sniok, line 12: I don't understand this comment

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  // DETAIL DRAWER MODE
```
sniok, line 14: please remove this

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const isDetailDrawerEnabled = useTypedSelector(state => state.drawerMode.isDetailDrawerEnabled);
+  const [isDrawerEnabled, changeDetailDrawerEnabled] = useState<any>(isDetailDrawerEnabled);
```
sniok, line 16: isDrawerEnabled is duplicated state, use only the state from the store instead

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <FormControlLabel
+      control={
+        <Switch
+          checked={isDetailDrawerEnabled}
+          onClick={drawerModeToggle}
+          name="drawerMode"
+          color="primary"
+        />
+      }
+      // will need to replace label
```
sniok, line 45: 👀

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -0,0 +1,39 @@
...
+export default function DetailsDrawer() {
+  const selectedResource = useTypedSelector(state => state.drawerMode.selectedResource);
+
+  const dispatch = useDispatch();
+
+  function closeDrawer() {
+    dispatch(setSelectedResource(undefined));
+  }
+
+  console.log({ selectedResource });
```
sniok, line 17: 👀

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -0,0 +1,39 @@
...
+      {selectedResource && (
+        <Drawer variant="persistent" anchor="right" open onClose={() => closeDrawer()}>
+          <Box width={800}>
+            <Box style={{ marginTop: '5rem', marginBottom: '2rem' }}>
+              <Button variant="outlined" color="primary" onClick={() => closeDrawer()}>
+                Close
+              </Button>
+            </Box>
+            <Box>
+              {/* NOTE: there is an issue where TS would throw an error for using KubeObject<any> along with the return undefined, this seems to fix it */}
```
sniok, line 30: 👀

frontend/src/components/common/Link.tsx
```diff
@@ -77,5 +77,39 @@ export default function Link(props: React.PropsWithChildren<LinkProps | LinkObje
     }
   }
 
+  if ('kubeObject' in props && kubeObject) {
+    {
+      /* NOTE: there is an issue where TS would throw an error for using KubeObject<any> along with the return undefined, this seems to fix it
```
sniok, line 82: is this still relevant?

frontend/src/components/common/Link.tsx
```diff
@@ -77,5 +77,39 @@ export default function Link(props: React.PropsWithChildren<LinkProps | LinkObje
...
+       * if we were to use dispatch(setSelectedResource(kubeObject!)); the error for returning kubeObject<any> & void would appear
+       */
+    }
+
+    const kubeJSON = kubeObject.jsonData;
+    return (
+      <>
```
sniok, line 89: Fragment is not required here

frontend/src/components/common/Link.tsx
```diff
@@ -77,5 +77,39 @@ export default function Link(props: React.PropsWithChildren<LinkProps | LinkObje
...
+        {drawerEnabled === true && props.kubeObject ? (
+          <MuiLink
+            onClick={() => {
+              if (drawerEnabled) {
+                dispatch(setSelectedResource(kubeJSON!));
+                window.history.pushState(
```
sniok, line 95: please use react-router history instead, mutating window.history will not play nicely with our existing router

frontend/src/redux/drawerModeSlice.ts
```diff
@@ -0,0 +1,35 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { KubeObject } from '../lib/k8s/KubeObject';
+
+interface DrawerModeState {
+  isDetailDrawerEnabled: boolean;
+  selectedResource: KubeObject | undefined;
+}
+
+const getLocalDrawerStatus = (key: string) => {
+  const local = localStorage.getItem(key);
```
sniok, line 10: can be replaced with `(key: string) => localStorage.getItem(key) === 'true' `

sniok (changes requested):

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: old comment

frontend/src/components/common/Link.tsx
vyncent-t: I previously used a separate component that would display the "drawer mode" links so that when clicking on a link it would not direct the user to the details page itself but rather the "/drawer" url variant

I recall moving away from that solution because the goal was to be able to send links easily without having it be dependent on having drawer mode enabled

I have tried using history.push and history.replace from the useHistory but both seem to force a page navigation, any other ideas in mind for this situation or should I just remove the URL changes entirely?

frontend/src/redux/drawerModeSlice.ts
vyncent-t: done!

frontend/src/components/common/Link.tsx
vyncent-t: done!

frontend/src/components/common/Link.tsx
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: done!

frontend/src/components/App/Layout.tsx
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
sniok: it's still here :)

frontend/src/i18n/locales/de/translation.json
```diff
@@ -53,8 +53,9 @@
   "Delete Plugin": "",
   "Are you sure you want to delete this plugin?": "",
   "Save": "",
-  "Uh-oh! Something went wrong.": "Oh-oh! Etwas ist schief gelaufen.",
+  "Uh-oh! Something went wrong.": "",
```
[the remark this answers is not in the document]
vyncent-t: running i18n removes it

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  console.log({ selectedResource });
+
+  return (
+    <>
+      {selectedResource && (
+        <Drawer variant="persistent" anchor="right" open onClose={() => closeDrawer()}>
+          <Box width={800}>
+            <Box style={{ marginTop: '5rem', marginBottom: '2rem' }}>
+              <Button variant="outlined" color="primary" onClick={() => closeDrawer()}>
+                Close
```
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }, [isDrawerEnabled]);
+
+  // this function takes in the current changes and updates it, this kicks off the useEffect that is listening for changes to newDrawerEnabled
+  function drawerModeToggle() {
+    changeDetailDrawerEnabled(!isDrawerEnabled);
+  }
+
+  // NOTICE THAT WE DO NOT USE isDrawerEnabled TO DETERMINE HOW THE SWITCH IS RENDERED UNDER THE CHECKED PROP, THIS IS BECAUSE THE USEEFFECT WILL RERENDER THE COMPONENT WITH THE NEW STATE
+  return (
+    <FormControlLabel
```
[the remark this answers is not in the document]
vyncent-t: added a tooltip explanation

vyncent-t:
added description to drawer mode button

fixed drawer not working on app mode (currently only updating the URL for browser mode)

frontend/src/i18n/locales/de/translation.json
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
```diff
@@ -0,0 +1,55 @@
+import { FormControlLabel, Switch } from '@mui/material';
+import React from 'react';
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useDispatch } from 'react-redux';
+import { setDetailDrawerEnabled } from '../../../redux/drawerModeSlice';
+import { useTypedSelector } from '../../../redux/reducers/reducers';
+import { TooltipIcon } from '../../common';
+
+export default function DrawerModeButton() {
```
sniok, line 10: Let's name this DrawerModeSettings. There's more here than a button

frontend/src/components/App/Settings/DrawerModeButton.tsx
sniok: @vyncent-t

sniok:
Could you also try to split the commit into separate ones? For example one with adding the slice, another with settings, another with updates to the Link, etc

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useTypedSelector } from '../../../redux/reducers/reducers';
+import { TooltipIcon } from '../../common';
+
+export default function DrawerModeSettings() {
+  const dispatch = useDispatch();
+  const { t } = useTranslation('translation');
+
+  const isDrawerEnabled = useTypedSelector(state => state.drawerMode.isDetailDrawerEnabled);
+
+  if (isDrawerEnabled) {
```
sniok, line 16: this part of the code takes isDrawerEnabled from the store and then immediately sets it back to the store, it's not needed. the same thing with the useEffect below

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const selectedResource = useTypedSelector(state => state.drawerMode.selectedResource);
+
+  const dispatch = useDispatch();
+
+  function closeDrawer() {
+    dispatch(setSelectedResource(undefined));
+  }
+
+  return (
+    <>
```
sniok, line 21: fragment is not required here

```tsx
if(!selectedResource) return null

return <Drawer...
```

frontend/src/components/common/Link.tsx
```diff
@@ -23,6 +27,7 @@ export interface LinkProps extends LinkBaseProps {
   state?: {
     [prop: string]: any;
   };
+  drawerEnabled?: boolean;
```
sniok, line 30: this prop is not used, unless I'm missing something

frontend/src/redux/drawerModeSlice.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface DrawerModeState {
+  isDetailDrawerEnabled: boolean;
+  selectedResource: KubeObject | undefined;
+}
+
+const getLocalDrawerStatus = (key: string) => localStorage.getItem(key) === 'true';
+
+const initialState: DrawerModeState = {
+  isDetailDrawerEnabled: getLocalDrawerStatus('detailDrawerEnabled'),
```
sniok, line 12: 'detailDrawerEnabled' is repeated, you should put it in a constant

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -0,0 +1,53 @@
...
+  useEffect(() => {
+    dispatch(setDetailDrawerEnabled(isDrawerEnabled));
+  }, [isDrawerEnabled]);
+
+  // this function takes in the current changes and updates it, this kicks off the useEffect that is listening for changes to newDrawerEnabled
+  function drawerModeToggle() {
+    dispatch(setDetailDrawerEnabled(!isDrawerEnabled));
+  }
+
+  // NOTICE THAT WE DO NOT USE isDrawerEnabled TO DETERMINE HOW THE SWITCH IS RENDERED UNDER THE CHECKED PROP, THIS IS BECAUSE THE USEEFFECT WILL RERENDER THE COMPONENT WITH THE NEW STATE
```
sniok, line 32: this comment is not needed anymore? the switch uses isDrawerEnabled

sniok (changes requested):

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -0,0 +1,53 @@
...
+          onClick={drawerModeToggle}
+          name="drawerMode"
+          color="primary"
+        />
+      }
+      label={
+        <>
+          {t('translation|Drawer Mode')}
+          <TooltipIcon>
+            {t('translation|Enable details to render in side drawer window')}
```
sniok, line 47: the wording here is a bit confusing. It's either a drawer or a window.

maybe something like this?
`Displays resource details in a side drawer instead of a full page`

frontend/src/components/App/Settings/DrawerModeSettings.tsx
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
vyncent-t: done

frontend/src/components/common/Link.tsx
vyncent-t: done

frontend/src/redux/drawerModeSlice.ts
vyncent-t: done

frontend/src/components/App/Settings/DrawerModeSettings.tsx
vyncent-t: done

vyncent-t:
issues
- deployments broken > open deployment resource in drawer, go to another resource and open it, open the deployment tab again and notice its broken

vyncent-t:
> issues
> 
> * deployments broken > open deployment resource in drawer, go to another resource and open it, open the deployment tab again and notice its broken

Was able to fix the issue by expanding on the logic for sorting the array (it was trying to sort an immutable array object)

frontend/src/components/common/Link.stories.tsx
```diff
@@ -1,15 +1,23 @@
...
+
 export default {
   title: 'Link',
   component: Link,
   decorators: [
     Story => (
-      <MemoryRouter>
-        <Story />
-      </MemoryRouter>
+      <Provider store={store}>
```
sniok, line 16: TestContext can be used here

vyncent-t:
Note: there seems to be a lot of waterfall failures due to the way links are now structured to make use of drawers from every link on the UI. Right now there are more than 30 storybook failures, not sure if they all need to be changed (maybe something with the store config?) or if its better to split the link logic up further to handle just for old tests

frontend/src/components/App/Settings/DrawerModeButton.tsx
vyncent-t: DrawerModeButton is now DrawerModeSettings and these changes have been made

frontend/src/i18n/locales/de/translation.json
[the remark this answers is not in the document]
vyncent-t: has been done for a while, resolving this

frontend/src/redux/drawerModeSlice.ts
joaquimrocha, line 1: Do we really need this in redux? I don't think we should have one reducer per setting. I think in the near future we should have some centralized settings for the app and a reducer for all of them, but not per setting.
So I think that for this feature, just relying on the local storage should be enough.

joaquimrocha: Okay, I see you are using this value to notify all the links. I still don't think it's needed but we can live with it for now as an internal change and later we do the changes I was suggesting.

frontend/src/components/common/Link.tsx
```diff
@@ -84,7 +89,47 @@ function PureLink(props: React.PropsWithChildren<LinkProps | LinkObjectProps>) {
...
+
+  const { tooltip, kubeObject, ...otherProps } = props as LinkObjectProps;
+  const name = 'kubeObject' in props ? props.kubeObject?.getName() : props.params?.name;
+  const namespace =
+    'kubeObject' in props ? props.kubeObject?.getNamespace() : props.params?.namespace;
+  const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props.routeName;
+
+  let content: any;
+
+  console.log(name, kind, canRenderDetails(kind));
```
joaquimrocha: Remove this please.

frontend/src/components/common/Link.tsx
```diff
@@ -84,7 +89,47 @@ function PureLink(props: React.PropsWithChildren<LinkProps | LinkObjectProps>) {
...
+  const drawerEnabled = useTypedSelector(state => state.drawerMode.isDetailDrawerEnabled);
+  const dispatch = useDispatch();
+
+  const { tooltip, kubeObject, ...otherProps } = props as LinkObjectProps;
+  const name = 'kubeObject' in props ? props.kubeObject?.getName() : props.params?.name;
+  const namespace =
+    'kubeObject' in props ? props.kubeObject?.getNamespace() : props.params?.namespace;
+  const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props.routeName;
+
+  let content: any;
```
joaquimrocha, line 101: Maybe React.ReactNode instead of any?

frontend/src/components/common/Link.tsx
```diff
@@ -84,7 +89,47 @@ function PureLink(props: React.PropsWithChildren<LinkProps | LinkObjectProps>) {
...
+
+  console.log(name, kind, canRenderDetails(kind));
+  if (!drawerEnabled || !canRenderDetails(kind)) {
+    content = <PureLink {...otherProps} />;
+  } else {
+    content = (
+      <MuiLink
+        onClick={() => {
+          if (drawerEnabled) {
+            dispatch(setSelectedResource({ kind, metadata: { name, namespace } }));
```
joaquimrocha, line 110: I wonder if we should use the URL's query params for this, e.g.:
`localhost:3000/c/my-cluster/pods/?details=Pod:my-namespace:my-pod` or similar.
Maybe the only weird case would be when we press the `< Back` button on top of the list after closing the drawer and this gets reset to the URL?

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -24,10 +25,16 @@ export default function DetailsDrawer() {
...
+        {/* Note: the top margin is needed to not clip into the topbar */}
+        <Box
+          sx={{
+            marginTop: '4rem',
+            display: 'flex',
+            padding: '1rem',
+            justifyContent: 'right',
+          }}
+        >
+          <ActionButton onClick={() => closeDrawer()} icon="mdi:close" description={t('Close')} />
```
joaquimrocha, line 37: Maybe you could merge this change with the one where you introduce the details drawer.

joaquimrocha (changes requested):
Good work! It's working now but I think we can improve it. I left a couple of comments.
Some are easy to change, others are more exploratory. For the latter, we may also change them later if that makes sense.

frontend/src/components/App/Layout.tsx
```diff
@@ -235,6 +238,7 @@ export default function Layout({}: LayoutProps) {
             </Container>
           </Box>
         </Main>
+        {isDetailDrawerEnabled && <DetailsDrawer />}
```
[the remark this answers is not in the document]
sniok: yeah this check can be moved inside the drawer component, it will make it a bit nicer

frontend/src/components/common/Link.tsx
vyncent-t: I believe I had tried something similar to this back when I was using routes and string slices for setting links, I remember fetching params and setting them for links and there being issues around using navigation for back and forwards page tabbing and the refresh behavior- especially for app mode. Which is when I started to implement the window.location setting

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
vyncent-t: done!

frontend/src/components/common/Link.tsx
vyncent-t: done!

frontend/src/components/common/Link.tsx
vyncent-t: done!

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <Switch
+          checked={isDrawerEnabled}
+          onClick={drawerModeToggle}
+          name="drawerMode"
+          color="primary"
+        />
+      }
+      label={
+        <>
+          {t('translation|Drawer Mode')}
```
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/components/App/Layout.tsx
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
```diff
@@ -0,0 +1,38 @@
...
+    dispatch(setSelectedResource(undefined));
+  }
+
+  if (!selectedResource) {
+    return null;
+  }
+
+  return (
+    <Drawer variant="persistent" anchor="right" open onClose={() => closeDrawer()}>
+      <Box width={800}>
```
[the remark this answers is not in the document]
vyncent-t: done! uses view width now

frontend/src/components/common/Link.tsx
sniok: currently it updates the url so that it represent the path to the resource in the overlay. adding it to the query params is also an option but for I think it's good as is for now

sniok (changes requested):
I think this is pretty much ready. The only slight visual issue is that the cursor for the link is incorrect, should be "pointer"

One caveat is that this feature doesn't work for Custom Resources but we can work on this in the future

frontend/src/components/common/Link.tsx
joaquimrocha: Still here.

frontend/src/components/common/Link.tsx
joaquimrocha: Ah right. I don't mind that the URL is the same as the regular dedicated-details-view. I have noticed however that there's an issue: when closing the drawer, the URL remains with the resource details.

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
joaquimrocha, line 1: Hmm, should this be under the Resource folder actually?

joaquimrocha (changes requested):
Left a couple more comments including a corner case I hit. BTW, there's also an empty commit in the list, and I'd like CRs and CRDs to be displayed in the drawer if it's not too complicated.

vyncent-t:
it is a bit too complicated to put in the CR details into the drawer as it is now, seems that it would likely need to be its own PR that can follow this one

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
vyncent-t: the current resource folders we have are for `resourceQuota` and the `resourceMap` that holds everything for the resource map tab, while they do use the same `<kubeObjectDetails>` comp to render, I don't think detail drawer falls under the map feature since you can use the drawer without ever opening/using the map tab

frontend/src/components/common/Link.tsx
vyncent-t: done!

frontend/src/components/DetailsDrawer/DetailsDrawer.tsx
sniok: the resource folder is the `src/components/common/Resource`

vyncent-t: done!

joaquimrocha (changes requested):

PR #1455, opened 2023-10-13
frontend: Add a way for plugins to react to Headlamp events

### LINKED ISSUE #1451: RFC: Headlamp event callbacks
I have been looking into how to allow plugins to be aware of certain events in the UI with the idea that they can use that to track the app's usage and get also insights on errors, etc. Partially this goal was described in #1211 .

As I was prototyping just having plugins registered callbacks for being called when certain things happened: error boundary triggered, create button clicked, etc. I realized that this may be also a useful feature beyond usage tracking.

With this feature we are thus able to provide a way for Headlamp vendors to track usage using their own plugin, but also any extra functionality as a response to different events.

joaquimrocha:
Plugins may be interested in reacting to when e.g. all plugins are
loaded, in order to show a notification or perform some different
action. These changes allows plugins to do that by registering
callbacks that are run when some events in the app happen.

fixes #1451


illume:
@ashu8912 walked me through the PR.

Nice.

- Should plugins be able to dispatch events too? Has this been considered. Is the dispatchEvent API exposed?
- Probably needs an example plugin for documentation and testing reasons.
- He queried the resource list event. I was also wondering about the name. Should it match the component name?

frontend/src/components/common/ObjectEventList.tsx
```diff
@@ -13,6 +14,12 @@ export interface ObjectEventListProps {
 export default function ObjectEventList(props: ObjectEventListProps) {
   const [events, setEvents] = useState([]);
 
+  useEffect(() => {
+    if (events) {
+      dispatchHeadlampEvent({ type: HeadlampEventType.OBJECT_EVENT, data: events });
```
illume, line 19: I'm not sure about having "Headlamp" in the name.

frontend/src/lib/util.ts
```diff
@@ -394,6 +395,16 @@ export function useId(prefix = '') {
   return id;
 }
 
+export function dispatchHeadlampEvent(eventInfo: HeadlampEvent | HeadlampEventType) {
```
illume, line 398: I don't think this should be in util. Because it's a grab bag of different things and it makes things complicated with import circles.

illume (changes requested):
Left a couple of notes for your consideration.

frontend/src/components/common/ObjectEventList.tsx
joaquimrocha: The name was included on purpose to not be confused with k8s events 🙂 
Does this make sense?

frontend/src/lib/util.ts
joaquimrocha: I am happy to move it to a more appropriate place. What do you suggest?

illume:
Hi. There were some [notes and questions that were missed](https://github.com/headlamp-k8s/headlamp/pull/1455#issuecomment-1878480764) because they were obscured with the other comment below. I'll repeat them here.

- Should plugins be able to dispatch events too? Has this been considered. Is the dispatchEvent API exposed?
- Probably needs an example plugin for documentation and testing reasons.
- He queried the resource list event. I was also wondering about the name. Should it match the component name?

I have some other feedback, but I'm not sure of the entire context or requirements. So I'm not clear if all of this applies.
- Is there an issue for this somewhere? It would be good to link it to get more context and to see if it solves the problems there.

- I like that there is only one exposed API for registration.
- The API doesn't use the type checker for the data given to the callback. It would be better for it to have types (see below for an option)
- It would be good to consider the number of files that need to be edited for this to add a new one. But I think it's pretty good. Maybe a little comment block inside of eventCallbackSlice explaining how to add one would be helpful.


### On redux...
- I like how it's possible to see these events in the store now. It will help with observability/debugging too inside the dev tools.
- How does this compare to the plugins using the redux state? It seems a hook (typedselector) won't work for these types of events.
- The events won't happen at the callsite, but will go through the redux store. I'm not sure if this matter for the requirements. But the implication of this is that folks won't be able to run code within the context of where the event is sent. Maybe this is a good thing. But I would like to know if you considered this.

### names
- Probably the dispatchHeadlampEvent could live inside the eventCallbackSlice rather than util.
- I feel dispatchEvent is better because it's still a unique name within the headlamp codebase. On hover inside vscode it's easy enough to see where it's from, and also by looking at the import.



### add types to data

```ts
interface ErrorBoundaryEvent {
  type: HeadlampEventType.ERROR_BOUNDARY;
  data?: { error: string };
}

interface DeleteResourceEvent {
  type: HeadlampEventType.DELETE_RESOURCE;
  data?: { resourceId: string };
}

type HeadlampEvent =
  | ErrorBoundaryEvent
  | DeleteResourceEvent
  | /* others */;
```

joaquimrocha:
I updated the PR with a more complex, single commit, but I think it makes sense as it now makes things more structured, type checked, and easier to use. Also added an example plugin but that's still WIP as I haven't finished adding the details in the README, etc.

frontend/src/redux/headlampEventSlice.ts
```diff
@@ -0,0 +1,452 @@
...
+  trackerFuncs: [],
+};
+
+export const eventAction = createAction<HeadlampEvent>('headlamp/event');
+
+export const listenerMiddleware = createListenerMiddleware();
+listenerMiddleware.startListening({
+  actionCreator: eventAction,
+  effect: async (action, listernerApi) => {
+    // @ts-ignore
```
illume, line 291: Could you please add a justification as comment to why this is here? (or fix it).
So in the future someone cleaning up can understand why.

illume:
Some notes for future selves... I remember we spoke verbally, but I wanted to have it written down somewhere.

- How would plugins who want to send events be able to do that? So plugin A can listen to an event from plugin B. There's no user requirement for this yet. However, I could see it possible to extend this to make an API for plugin authors in the future. Under it all it's just a { type, data }, and all the types in here are prefixed with "headlamp.". So a future plugin sending event API would just need to expose that.
-  Having functionality for different modules all in one file was a bit concerning. However, it's not such a great concern because there's not much cyclical importing of things going on. It only imports three internal things so far. But it does mean if there are 10x more things added that file will get quite crowded. I can roughly see a refactoring that might make this worthwhile to do when we get up to 10x more events, so I feel we do that then not now... if required. This current way seems more convenient, and better now.
- For "The events won't happen at the callsite, but will go through the redux store. " there doesn't appear to be a use case for this. If there is later, I could see there being something like a 'sync' attribute added so the callbacks are called at the callsite.

plugins/examples/headlamp-events/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "headlamp-events",
+  "version": "0.0.1",
+  "description": "Your Headlamp plugin",
```
illume, line 4: Needs a description.

plugins/examples/headlamp-events/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "headlamp-events",
```
illume, line 2: Maybe this could be `@headlamp-k8s/headlamp-events`

plugins/examples/headlamp-events/src/index.tsx
```diff
@@ -0,0 +1,83 @@
+import {
+  DefaultHeadlampEvents,
+  registerAppBarAction,
+  registerHeadlampEventCallback,
+} from '@kinvolk/headlamp-plugin/lib';
+import { useSnackbar } from 'notistack';
+import React from 'react';
+
```
illume, line 8: A brief description comment of what the example does would be good up top here.

plugins/examples/headlamp-events/README.md
```diff
@@ -0,0 +1,18 @@
+# Example Plugin: Reacting to Headlamp events
```
illume, line 1: Some of these sentences are marked hard and very hard to read. 

See...
image

illume (commented):
Very very Nice. 

I left a few comments, now I'm testing it locally.

frontend/src/plugin/registry.tsx
```diff
@@ -560,4 +597,20 @@ export function registerGetTokenFunction(override: (cluster: string) => string |
...
 
+/**
+ * Add a callback for headlamp events.
+ * @param callback - The callback to add.
+ *
+ * @example
+ *
+ * ```ts
+ * registerHeadlampEventCallback((event) => {
+ *   console.log('Last Headlamp event:', event)
```
illume, line 608: In the example, can you please show the type and data fields being used?
It would help to include the import here for DefaultHeadlampEvents as well.

frontend/src/redux/headlampEventSlice.ts
```diff
@@ -0,0 +1,452 @@
+import {
```
illume, line 1: Can you please add a headlampEventSlice.test.ts with some minimal tests to cover adding a callback and dispatching an event?

frontend/src/redux/headlampEventSlice.ts
joaquimrocha: This was a c&p left-over. I'll fix it.

plugins/examples/headlamp-events/package.json
joaquimrocha: Maybe we should have a prefix for all example plugins, e.g. `@headlamp-k8s/plugin-examples.headlamp-events`?

frontend/src/plugin/registry.tsx
joaquimrocha: Will do.

frontend/src/redux/headlampEventSlice.ts
joaquimrocha: Yup. Sorry I missed this.

plugins/examples/headlamp-events/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Headlamp has the concept of Headlamp events, which are fired when something relevant happens in Headlamp.
+For example, when a user chooses to apply the changes their changing for a resource, or when a
+list/details view is loaded.
+
+To run the plugin:
```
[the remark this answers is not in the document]
joaquimrocha: Yup. Completely forgot about that. Thanks.

plugins/examples/headlamp-events/README.md
joaquimrocha: I have used this app in the past and to be honest I find it very difficult to write sentences it likes while still keeping a speech that is technical and resembles my own "voice". So I prefer to get yours and other readers' own review/suggestions when sentences sound complex. Maybe this is my problem or maybe I am trying to reach the best state in it and that that's too demanding.

However, I do agree these texts you pointed out sounded weird. I did an effort and used the app for the examples you mentioned and hopefully it's better now (though not perfect, according to the app 🙂 ).

joaquimrocha:
Thanks for the review @illume . I have updated the branch after your suggestions.

illume (approved):
Great job! Thanks.

plugins/examples/headlamp-events/package.json
illume: Hrmm. Yeah, so we can see that it is a plugin example in the name. As apposed to something completely different (like a lint config package.

I like how `cd plugins/examples` is kind of in the name too.

Another option could be: `@headlamp-k8s/headlamp-events-plugin-example`.
But... probably we should have this naming discussion outside of this PR.

joaquimrocha: Right. We'd need to rename all example plugins anyway, and this is only relevant if we decide to publish them. So let's keep a pin on the conversation and get back to it when we decide to publish the packages.

PR #1485, opened 2023-10-20
Show spinner when workloads are loading

### LINKED ISSUE #1483: Add spinner to workloads' overview table
The workload overview table doesn't show a spinner when it's loading and shows the "No data to show"  instead.
In order to improve UX and consistency, we should add that spinner.

**Acceptance Criteria:**
- [ ] Navigate to the workloads page: there should be a spinner until any data is loaded (and the "no data" message should only be shown if there's no actual data)

### LINKED ISSUE #1489: Add a spinner to the tiled chart
When the data is still loading, the titled chart (workloads charts, for example) should show a spinner until the data is loaded.

**Accepted Criteria:**
- [ ] Go to the workloads page and observe the charts: until they are loaded, there would be a spinner

joaquimrocha:
This PR fixes #1483 (check it out for testing instructions) and simplifies the logic for the workloads state as well.
It also adds a spinner to the tiled charts.
fixes #1489

### DIFF
--- a/frontend/src/components/workload/Charts.tsx
+++ b/frontend/src/components/workload/Charts.tsx
@@ -1,4 +1,5 @@
 import { useTheme } from '@material-ui/core/styles';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Workload } from '../../lib/k8s/cluster';
 import { getPercentStr, getReadyReplicas, getTotalReplicas } from '../../lib/util';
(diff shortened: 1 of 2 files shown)


vyncent-t:
The spinners do work but I noticed that they will always spin when there are 0 resources on some cards

this branch
image

main
image

vyncent-t (approved):

joaquimrocha:
> The spinners do work but I noticed that they will always spin when there are 0 resources on some cards, this is with the metrics-server not enabled only
> 

This needs to be fixed then. Thanks.

joaquimrocha:
Thanks a lot for the test, @vyncent-t . I have updated the PR. Please take another look when you can.

vyncent-t:
branch now no longer spins continuously with no metrics server addon, both addon modes are identical!
Looks like that fixed the infinite spin!

ON
image

OFF
image

vyncent-t (approved):

PR #1486, opened 2023-10-20
README: Add openssf best practices badge

illume:
A cncf sandbox requirement, see https://github.com/cncf/toc/issues/1056

It looks like this:
image

We have "passing" status. See below for requirements to get silver or gold:
- passing https://www.bestpractices.dev/en/projects/7551
- silver requirements https://www.bestpractices.dev/en/projects/7551?criteria_level=1
- gold requirements https://www.bestpractices.dev/en/projects/7551?criteria_level=2


joaquimrocha:
Ah! I started this process and didn't follow through because we were not meeting all points. I think now we are but I want to take a closer look (next week) before enabling the badge. Thanks for adding this.

PR #1488, opened 2023-10-23
docker-ext: Fix portforward issue

yolossn:
In docker desktop since the backend is running
inside a container portforwards running in the
container are not accessible from the host machine
to fix this a range of ports from 30000 to
32000 is exposed and a value in this range is used
when starting a portforward.


ashu8912 (commented):
The code overall looks good to me Thanks, but can you please explain in the commit  a bit more about what we are trying to do here and what ports are exposed etc.

ashu8912 (approved):
Tested this with help from @yolossn this LGTM

Issue #1500, opened 2023-10-27
Style Issue for Endpoints Field in Service View


vyncent-t:
# Style Issue for Endpoints Field in Service View

## Description
In the detailed view of a service, a styling issue is observed under the Endpoints section where the table renders a blank box across all viewing sizes. This issue hinders the user's ability to view and interact with endpoint information efficiently.

I will work on this bug if this blank rendering is not intended.

## Expected Behavior
- The table under the Endpoints section should render correctly and display the relevant information without any blank box or misalignment.

## Steps to Reproduce
1. Navigate to the Service View.
2. Select a service to view its details.
3. Scroll down to the Endpoints section to observe the blank box in the table rendering.

## Possible Solution
- Inspect the styling and rendering logic for the Endpoints section to identify and fix the source of the issue.

![image](https://github.com/headlamp-k8s/headlamp/assets/78232183/017ba551-59f5-485e-9623-e0b7550df004)

Src0p:
Can someone , review my PR  related to this issue https://github.com/headlamp-k8s/headlamp/pull/1510

vyncent-t:
@Src0p 
Thank you for your contribution! I'll review your PR shortly.

PR #1510, opened 2023-10-28
Fix: Issue #1500 

Src0p:
Fixed the styling and rendering issue in endpoint section of the Services view.

### DIFF
--- a/frontend/src/components/service/Details.tsx
+++ b/frontend/src/components/service/Details.tsx
@@ -100,8 +100,10 @@ export default function ServiceDetails() {
                       },
                       {
                         label: t('translation|Addresses'),
-                        getter: endpoint => endpoint.getAddressesText(),
-                        cellProps: { style: { width: '40%', maxWidth: '40%' } },
+                        getter: endpoint => {
+                          return endpoint.getAddressesText();
+                        },
+                        cellProps: { style: { width: '100%', maxWidth: '100%' } },
                       },
                     ]}
                     reflectInURL="endpoints"
--- a/frontend/src/lib/k8s/endpoints.ts
+++ b/frontend/src/lib/k8s/endpoints.ts
@@ -44,7 +44,7 @@ class Endpoints extends makeKubeObject<KubeEndpoint>('endpoint') {
   }
 
   getAddressesText() {
-    this.getAddresses().join(', ');
+    return this.getAddresses().join(', ');
   }
 
   getAddresses() {


frontend/src/components/service/Details.tsx
```diff
@@ -100,8 +100,10 @@ export default function ServiceDetails() {
                       },
                       {
                         label: t('translation|Addresses'),
-                        getter: endpoint => endpoint.getAddressesText(),
-                        cellProps: { style: { width: '40%', maxWidth: '40%' } },
+                        getter: endpoint => {
+                          return endpoint.getAddressesText();
+                        },
```
joaquimrocha, line 105: This change is not needed. As the code deleted does the same.

frontend/src/lib/k8s/endpoints.ts
```diff
@@ -44,7 +44,7 @@ class Endpoints extends makeKubeObject<KubeEndpoint>('endpoint') {
   }
 
   getAddressesText() {
-    this.getAddresses().join(', ');
+    return this.getAddresses().join(', ');
```
joaquimrocha, line 47: Ouch! Weird we got this one through.

joaquimrocha (changes requested):
Hi @Src0p . Thanks for your PR!

I have commented on one of the changes you did.
The commit title should also be aligned to the [guidelines](https://github.com/kinvolk/contribution/blob/master/topics/git.md#authoring-prs) we use. So it should be something like:
```
frontend endpoin.ts: Actually return the addresses in getAddressesText

By mistake, a return statement was missing.
```

As part of Headlamp being now a CNCF Sandbox project, we also need every commit to be signed-off. So please add the `-s` option when committing/amending.

frontend/src/components/service/Details.tsx
Src0p: Thank you so much for the feedback , I am really sorry for overlooking the guidelines.

joaquimrocha: No worries. We definitely encourage folks to submit PRs even if they need a bit adjustment to get them merged.
I saw you closed this PR. Of course you can reopen a new one when you have the new changes, but usually you just have to amend your commits and force push. If you need help with this, join our [Slack channel](https://kubernetes.slack.com/messages/headlamp) and I can help.

This is an important fix that I would like to get in the next version. Are you able to send us a new commit after my review?

joaquimrocha: Ah, I see that there's another PR that's related to this and we should use that one. Sorry for the confusion @Src0p . I hope to see more contributions from you!

PR #1514, opened 2023-10-30
frontend: Remove the unused serviceWorker

illume:


joaquimrocha:
@ashu8912 , I think you added this a long time ago. Can you double check whether it's needed?

ashu8912:
I think ServiceWorker was added by default when we ran the CRA script initially to add the frontend base, In this case, though we are not registering the service worker so in a way it is unused I am not sure if we should remove it coz there could be opportunities in future where we could rely on the service worker for optimization purposes.

joaquimrocha:
Let's remove it then.

PR #1526, opened 2023-11-03
frontend: Use the right group and version in KubeObject.getAuthorization

### LINKED ISSUE #1490: Headlamp doesn't send apiGroup parameter to the SelfSubjectAccessReview API
Hello, I was trying to log into Headlamp using the access token for a user who has RoleBinding to the ClusterRole admin that is restricted with a specific namespace (it means it should have full access to the namespace scoped resources).

After this It was found out that my user cannot edit Deployments and other resources that have apiGroups.

It seems like the issue is because Headlmap doesn't send correct apiGroup parameter when requesting for checking access for the current user:

`{"kind":"SelfSubjectAccessReview","apiVersion":"authorization.k8s.io/v1","spec":{"resourceAttributes":{"verb":"update","name":"my-deployment","namespace":"my-namespace","resource":"deployments"}}}` - this is the current request sent by Headlamp to the kube api server and returns false as cluster role admin allow update only for the deployments that are under the apiGroup "apps":

```
- apiGroups:
  - apps
  resources:
  - daemonsets
  - deployments
  - deployments/rollback
  - deployments/scale
  - replicasets
  - replicasets/scale
  - statefulsets
  - statefulsets/scale
  verbs:
  - create
  - delete
  - deletecollection
  - patch
  - update
```

If I add apiGroups parameter to the permission check request then its working fine:
`{"kind":"SelfSubjectAccessReview","apiVersion":"authorization.k8s.io/v1","spec":{"resourceAttributes":{"verb":"update","name":"my-deployment","namespace":"my-namespace","resource":"deployments","group":"apps"}}}`

So it seems like Headlamp should send this apiGroup parameter to correctly check permissions for the resources that are under apiGroups.

joaquimrocha:
The right group and version were not being used when checking for a
resources' authorization which resulted in wrong authorization
assessments.

fixes #1490

### DIFF
--- a/frontend/src/lib/k8s/cluster.ts
+++ b/frontend/src/lib/k8s/cluster.ts
@@ -309,6 +309,9 @@ export interface AuthRequestResourceAttrs {
   resource?: string;
   subresource?: string;
   namespace?: string;
+  version?: string;
+  group?: string;
+  verb?: string;
 }
 
 // @todo: uses of makeKubeObject somehow end up in an 'any' type.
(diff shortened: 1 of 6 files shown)


frontend/src/lib/k8s/cluster.ts
```diff
@@ -667,10 +670,35 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
       return this._class().apiEndpoint.patch(...args);
     }
 
+    private static async fetchAuthorization(reqResourseAttrs?: AuthRequestResourceAttrs) {
```
illume, line 673: Needs docs.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -679,37 +707,49 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+        return this.fetchAuthorization(resourceAttrs);
+      }
+
+      // If we don't have the group, version, and resource, then we have to try all of the
+      // API info versions until we find one that works.
+      const apiInfo = this.apiEndpoint.apiInfo;
+      for (let i = 0; i < apiInfo.length; i++) {
+        const { group, version, resource } = apiInfo[i];
+
+        if (!!resourceAttrs.resource) {
```
illume, line 724: Could this use resourceAttrs from the previous iteration of the loop? If group is set in i=0, and then not in i=1... it will use the group from i=0;. Is this intentional?

illume (commented):
Left a couple of notes...

frontend/src/lib/k8s/cluster.ts
```diff
@@ -679,37 +707,49 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
-          return await post(
-            `/apis/authorization.k8s.io/${version}/selfsubjectaccessreviews`,
-            {
-              kind: 'SelfSubjectAccessReview',
-              apiVersion: `authorization.k8s.io/${version}`,
-              spec,
-            },
-            false
-          );
+          authResult = this.fetchAuthorization(resourceAttrs);
```
illume, line 737: Do you need to await here? Seems it's needed, otherwise the loop will just keep going.

frontend/src/lib/k8s/cluster.ts
joaquimrocha: Hmm, that would be a possible side effect if the apiInfo doesn't have all the values in all iterations. But I think it always does. I will try to refine the logic nonetheless.

illume (approved):
👍 thanks. LGTM

PR #1527, opened 2023-11-03
app: Update notarize package and script for the new notarize tool

joaquimrocha:
Apple now requires that the new notarize tool is used, so these
changes are necessary for our notarize process to keep working.


joaquimrocha:
@illume I have just tested it and it works.

illume (approved):
👍 thanks

PR #1529, opened 2023-11-03
Fix cronjob details

joaquimrocha:
How to test:
- [ ] Set up the prometheus plugin, then go to a CronJob details view: there should be a chart showing before the details for the CronJob


ashu8912 (approved):
Tested

PR #1534, opened 2023-11-08
frontend: Fix start portforward breaking in list view

yolossn:
This patch fixes the start portforward function
is not working issue in the portforward list
view.


frontend/src/components/portforward/index.tsx
```diff
@@ -117,6 +118,11 @@ export default function PortForwardingList() {
     const { id, namespace, cluster, port, targetPort, pod, service, serviceNamespace } =
       portForwardInAction;
 
+    var address = 'localhost';
```
joaquimrocha, line 121: Please use let instead of var.

joaquimrocha (changes requested):
Just one small comment. Otherwise it looks good.

joaquimrocha (approved):

PR #1548, opened 2023-11-10
Show spinners when cluster changes

joaquimrocha:
Headlamp has a convention where lists of items are set as null which means we are still fetching the objects, and a spinner should be shown instead of an empty or filled list.
However, if we seamlessly replace the cluster for which the list of items are being shown, then the list will hold the old items until new ones are prepared. So there's no perception for the user that there's a loading state.

This PR fixes this for the cluster and workloads overview. The first is fixed by resetting the items list in the useList hook, and using this hook in the cluster overview (instead of useApiList). The workloads overview is a bit more elaborate and replacing useApiList with useList is more complicated, so in this case we reset the object holding the workloads when the cluster changes (directly in the overvierw component).

How to test:
-[ ] Go to the cluster overview, then switch to a different cluster (it will show the login and redirect to the cluster view), choose the previous cluster again: this time the login screen should no longer be shown and you'll see the spinners show up for the charts and table (without this PR, it'll not show spinners, the data is just changed)

cc/ @lijianzhi01

### DIFF
--- a/frontend/src/components/cluster/Overview.tsx
+++ b/frontend/src/components/cluster/Overview.tsx
@@ -25,13 +25,11 @@ const useOverviewStyle = makeStyles({
 });
 
 export default function Overview() {
-  const [pods, setPods] = React.useState<Pod[] | null>(null);
-  const [nodes, setNodes] = React.useState<Node[] | null>(null);
   const { t } = useTranslation(['translation']);
   const classes = useOverviewStyle();
 
-  Pod.useApiList(setPods);
-  Node.useApiList(setNodes);
+  const [pods] = Pod.useList();
+  const [nodes] = Node.useList();
 
   const [nodeMetrics, metricsError] = Node.useMetrics();
 
--- a/frontend/src/components/workload/Overview.tsx
+++ b/frontend/src/components/workload/Overview.tsx
@@ -2,6 +2,7 @@ import Grid from '@mui/material/Grid';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router-dom';
+import { useCluster } from '../../lib/k8s';
 import { ApiError } from '../../lib/k8s/apiProxy';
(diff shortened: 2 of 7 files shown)


frontend/src/components/cluster/Overview.tsx
```diff
@@ -26,13 +26,11 @@ const useOverviewStyle = makeStyles({
...
 
 export default function Overview() {
-  const [pods, setPods] = React.useState<Pod[] | null>(null);
-  const [nodes, setNodes] = React.useState<Node[] | null>(null);
   const { t } = useTranslation(['translation']);
   const classes = useOverviewStyle();
 
-  Pod.useApiList(setPods);
-  Node.useApiList(setNodes);
+  const [pods] = Pod.useList();
```
ashu8912, line 32: The useList API also returns a error in case something went wrong while making the request, should we handle it here??

frontend/src/components/cluster/Overview.tsx
```diff
@@ -26,13 +26,11 @@ const useOverviewStyle = makeStyles({
...
+  const [nodes] = Node.useList();
```
ashu8912, line 33: same here

frontend/src/components/workload/Overview.tsx
```diff
@@ -26,6 +27,11 @@ export default function Overview() {
   const location = useLocation();
   const filterFunc = useFilterFunc(['.jsonData.kind']);
   const { t } = useTranslation('glossary');
+  const cluster = useCluster();
+
+  React.useEffect(() => {
```
ashu8912, line 32: How to test the workload overview not showing the spinner on cluster change??

frontend/src/components/cluster/Overview.tsx
joaquimrocha: We could do that in a new PR if needed. This logic should do the same as the one it's changing (minus returning null again if the cluster changes).

ashu8912 (approved):

PR #1549, opened 2023-11-10
github: Fix PR comment issue for forks in backend test

yolossn:
this patch selectively skips commenting the code
coverage change for PRs raised from forks to avoid 
the authorization error that arises when commenting.

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -49,7 +49,6 @@ jobs:
       - name: Get base branch code coverage
         run: |
           cd backend
-          ls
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):
This is fine. I think you also can make the whole job/workflow not run unless the PR comes from an external repo. Do you want to add that?

yolossn:
> This is fine. I think you also can make the whole job/workflow not run unless the PR comes from an external repo. Do you want to add that?

Since the code coverage check and commenting part is added to the backend test PR, completely skipping it will make the backend test not be run in the CI.

PR #1558, opened 2023-11-15
Improve CR permissions and class generation

### LINKED ISSUE #1552: Headlamp doesn't send correct resource name parameter for CRDs to the SelfSubjectAccessReview API
Hello, this issue is similar to https://github.com/headlamp-k8s/headlamp/issues/1490 but a little bit different and related to CustomResourceDefinition objects. I've tried to edit a Namespace scoped CustomResource (like ServiceMonitor or Certificate) with my namespace scoped service account that is bound to the cluster role admin.

Currently headlamp send incorrect resource name for CRD resources in the SelfSubjectAccessReview API call:

`{"kind":"SelfSubjectAccessReview","apiVersion":"authorization.k8s.io/v1","spec":{"resourceAttributes":{"verb":"update","name":"my-servicemonitor","namespace":"my-namespace","group":"monitoring.coreos.com","version":"v1","resource":"crds"}}}`

Response:

`"status": {
    "allowed": false
  }`

This request returns incorrect false response for a user who is admin in the namespace only and should have ability to edit these resources.

Note that resource is **crds** for all requests to all CustomResource objects. Instead it should send correct CustomResource name, like:

`{"kind":"SelfSubjectAccessReview","apiVersion":"authorization.k8s.io/v1","spec":{"resourceAttributes":{"verb":"update","name":"my-servicemonitor","namespace":"my-namespace","group":"monitoring.coreos.com","version":"v1","resource":"servicemonitors"}}}`

This request returns correct response:

`"status": {
    "allowed": true,
    "reason": "RBAC: allowed by RoleBinding \"namespace-admin" of ClusterRole \"admin\" to ServiceAccount \"namespace-admin\""
  }`

joaquimrocha:
This PR simplifies the use of CR classes by adding a new method for generating them out of CRD objects, and also fixes the permissions issues reported in #1552.

fixes #1552 

@Denis220795 , thanks again for reporting another important issue with a great insight that made this fix easier to do.
Can you try this branch and check if the issue if fixed for you?

### DIFF
--- a/frontend/src/components/crd/CustomResourceDetails.tsx
+++ b/frontend/src/components/crd/CustomResourceDetails.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
 import { ResourceClasses } from '../../lib/k8s';
 import { ApiError } from '../../lib/k8s/apiProxy';
-import CustomResourceDefinition, { KubeCRD, makeCustomResourceClass } from '../../lib/k8s/crd';
+import CustomResourceDefinition, { KubeCRD } from '../../lib/k8s/crd';
 import { localeDate } from '../../lib/util';
 import { HoverInfoLabel, Link, NameValueTableRow, ObjectEventList, SectionBox } from '../common';
 import Empty from '../common/EmptyContent';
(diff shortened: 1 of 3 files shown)


Denis220795:
@joaquimrocha sure, I'll try to test this branch tomorrow and provide the results.

Denis220795:
@joaquimrocha just tested, working fine. Now I can edit namespaced custom resource with namespace scoped admin role.

PR #1561, opened 2023-11-17
Use kubeconfig namespace

joaquimrocha:
User are sometimes interested in using the namespace that's set up in their kubeconfig, associated with the context/cluster they're using in Headlamp.
These changes accomplish that.

How to test:
- [ ] Set up a namespace in a context in the kubeconfig, then go to that cluster's settings in Headlamp: the placeholder default namespace should be the same as set in the kubeconfig
- [ ] Apply some resource without specifying the namespace: the default namespace where it's applied should automatically be the one set in the kubeconfig

### DIFF
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -22,7 +22,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const staticTestPath = "headlamp_testdata/static_files/"
+const (
+	staticTestPath = "headlamp_testdata/static_files/"
+	minikubeName   = "minikube"
+)
 
 // Is supposed to return the index.html if there is no static file.
 func TestSpaHandlerMissing(t *testing.T) {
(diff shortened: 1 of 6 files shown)


backend/cmd/headlamp_test.go
```diff
@@ -318,8 +321,31 @@ func TestDynamicClustersKubeConfig(t *testing.T) {
...
+			contextWithoutNamespace = &clusters[i]
+		}
+	}
+
+	assert.NotNil(t, contextWithoutNamespace)
+	assert.Equal(t, "", contextWithoutNamespace.Metadata["namespace"])
+
+	assert.NotNil(t, minikubeCluster)
+	assert.Equal(t, minikubeName, minikubeCluster.Name)
+	assert.Equal(t, "default", minikubeCluster.Metadata["namespace"])
```
yolossn, line 348:
```suggestion
	for i, cluster := range clusters {
	        cluster:=cluster
		if cluster.Name == minikubeName {
	             assert.Equal(t, minikubeName, cluster.Name)
	             assert.Equal(t, "default", cluster.Metadata["namespace"])

		} else if cluster.Name == "docker-desktop" {
	             assert.Equal(t, "", context.Metadata["namespace"])
	  }
	}


	```

yolossn:
on manual testing I am getting an error when creating a simple pod without namespace

image
image

joaquimrocha:
Thanks. I will have to check it out.

joaquimrocha:
Fixed. It was checking the cluster info's metadata which can be null.
Should work now.

PR #1562, opened 2023-11-19
Makefile: Dockerfile: Add support for custom IMAGE_BASE and Debian

illume:
Support for apt-get using Debian-like systems.

- Fixes https://github.com/headlamp-k8s/headlamp/issues/1072
- Adds a section in the documentation about how to specify different base images.

## How to test

1) Try a base image with a Debian base image.
```bash
IMAGE_BASE=debian:latest make image
```

2) See that the default used is alpine.

```bash
make image
```


joaquimrocha (approved):
Nice. Thanks.

PR #1564, opened 2023-11-24
frontend: Add Vertical Pod Autoscalers

yolossn:


frontend/src/components/verticalPodAutoscaler/Details.tsx
```diff
@@ -0,0 +1,199 @@
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import VPA from '../../lib/k8s/vpa';
+import { DateLabel, DetailsGrid, Link, SectionHeader, SimpleTable } from '../common';
+
+export default function VpaDetails() {
+  const { namespace, name } = useParams<{ namespace: string; name: string }>();
+  const { t } = useTranslation();
```
joaquimrocha, line 8: You need to load the namespaces you need to use here.
```typescript
const { t } = useTranslation(['translation', 'glossary']);
```

frontend/src/components/verticalPodAutoscaler/Details.tsx
```diff
@@ -0,0 +1,199 @@
...
+                      label: t('glossary|Lower Bound'),
+                      getter: item => {
+                        let lowerBound = '';
+                        if (item.lowerBound && Object.keys(item.lowerBound).length > 0) {
+                          Object.keys(item.lowerBound).map(key => {
+                            lowerBound += `${key}:${item.lowerBound[key]},`;
+                          });
+                        }
+                        lowerBound = lowerBound.slice(0, -1);
+                        return lowerBound;
```
joaquimrocha, line 148: This is the rest of the similar code could be abstracted into a helper function instead of having the code repeated.

joaquimrocha (changes requested):
Besides my comments, and what we had discussed in our meeting (adding a notice when the VPAs are not available).

frontend/src/components/verticalPodAutoscaler/Details.tsx
```diff
@@ -0,0 +1,174 @@
...
+              </>
+            ),
+          },
+          {
+            id: 'headlamp.vpa-status-conditions',
+            section: (
+              <>
+                <SectionHeader title={t('translation|Conditions')} />
+                <SimpleTable
+                  data={item.status.conditions}
```
joaquimrocha, line 109: This intermediate object needs to be checked.

frontend/src/components/verticalPodAutoscaler/Details.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  ]}
+                />
+              </>
+            ),
+          },
+          {
+            id: 'headlamp.vpa-status-conditions',
+            section: (
+              <>
+                <SectionHeader title={t('translation|Conditions')} />
```
joaquimrocha, line 107: Notice how the alignment on the sections + their titles is not correct. This can be done consistently by using the SectionBox instead of passing the SectionHeader.

joaquimrocha (changes requested):
Left some comments on the visuals.
Also tried to run it with the nginx-vpa example in [this page](https://www.kubecost.com/kubernetes-autoscaling/kubernetes-vpa/) and it crashed both in the list and the details view.
Please do an extra check for intermediate objects/states that may be missing.

frontend/src/components/verticalPodAutoscaler/List.tsx
```diff
@@ -0,0 +1,65 @@
...
+              Vertical Pod Autoscaler is not enabled, To enable it, please{' '}
+              <a
+                href="https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#installation"
+                target="_blank"
+              >
+                follow
+              </a>
+            </p>
+          </Empty>
+        </Paper>
```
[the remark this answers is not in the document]
yolossn: Made it look similar to the other views.
image
image

frontend/src/components/verticalPodAutoscaler/List.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <Empty>
+              <Typography style={{ textAlign: 'center' }}>
+                Vertical Pod Autoscaler is not enabled.{` `}
+                <Link
+                  href="https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#installation"
+                  target="_blank"
+                  rel="noopener"
+                >
+                  Learn More
+                </Link>
```
joaquimrocha, line 71: This needs to be translatable. Maybe using the `Trans` component.

frontend/src/components/verticalPodAutoscaler/List.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, []);
+
+  return (
+    <>
+      {vpaEnabled === null ? (
+        <SectionBox title={t('glossary|Vertical Pod Autoscalers')}>
+          <Paper variant="outlined">
+            <Empty>
+              <Typography style={{ textAlign: 'center' }}>
+                Checking if Vertical Pod Autoscaler is enabled...
```
joaquimrocha, line 28: This needs to be translatable. And the elipsis character should be one single character (not three period characters).

joaquimrocha (changes requested):
Now it works great. I have noticed that the empty strings need translating, but that after that, it should be fine.

joaquimrocha:
In the interest of merging as soon as possible, I implemented the changes I had required (just had to do with translations).

joaquimrocha (approved):

PR #1565, opened 2023-11-25
Add e2e tests

knrt10:
Streamline setup for e2e tests. We do not need to authenticate repetedly now. Added helpers for it.

CI did not ran e2e tests when changes were made to e2e folder, so updated it.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -12,6 +12,7 @@ on:
     - Makefile
     - '.github/**'
     - Dockerfile
+    - 'e2e-tests/**'
 
 jobs:
   build:
@@ -38,7 +39,7 @@ jobs:
     - name: Deploy to cluster
       run:
         kubectl apply -f e2e-tests/kubernetes-headlamp-ci.yaml
-    - name: Test service URLs
+    - name: Run e2e tests
       run: |
         echo "------------------sleeping 3...------------------"
         sleep 3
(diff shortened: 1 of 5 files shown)


e2e-tests/tests/helpers.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { expect } from "@playwright/test";
```
illume, line 1: I've seen these put into Page Object Models in other playwright code bases. This is a fairly minimal doc with an explanation: https://playwright.dev/docs/pom This tutorial is much better (but longer): https://ray.run/blog/mastering-poms

What do you think about follow that convention?

e2e-tests/tests/helpers.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  await page.locator("#token").fill(token);
+
+  // Click on the "Authenticate" button
+  await page.click('button:has-text("Authenticate")');
+  // Wait for the page to load
+  await page.waitForLoadState('load');
+};
+
+const servicesPage = async (page) => {
+  // Click on the "Network" button
```
illume, line 25: For my own interest... I wonder with these was it difficult to use `await page.goto("/");` to other pages? Did that not work, and that is why you are using click? Or are there other advantages to doing it this way?

Are these links are always there? I guess the sidebar is there mostly.

illume (commented):
Very nice. It's working!

I left a couple of comments for your consideration, and a question about page navigation.

Note, the commit message is missing the `e2e-tests:` context part at the beginning.

e2e-tests/tests/helpers.spec.ts
knrt10: > I wonder with these was it difficult to use await page.goto("/"); to other pages?

When we navigate to "/" page after authenticating it takes time for page to load. Also, I wanted to test click functionality of the app so 2 birds in one stone I guess. Nothing much to it other than this. 

> Are these links are always there? I guess the sidebar is there mostly.
 
Yes sidebar is there and I wait for page to load for other things to load.

e2e-tests/tests/helpers.spec.ts
knrt10: Cool, make sense. Will update the PR.

knrt10: Done

e2e-tests/tests/helpers.spec.ts
illume: Thanks for the explanation.

illume (approved):
Thanks 👍

Issue #1567, opened 2023-11-28
[Bug]: CI: All plugins not getting copied to /build/plugins


knrt10:
### Problem

Plugins **[ 'plugins/app-menus', 'plugins/details-view' ]** are only available inside `/build/plugins` folder. I logged them while working on a PR, check [CI](https://github.com/headlamp-k8s/headlamp/actions/runs/7001102134/job/19042702546?pr=1303) for more info.

Also it can be viewed in any CI run `Build Container and test` workflow, see `Build image` run. Also, I am attaching image below for more reference.

<img width="1083" alt="Screenshot 2023-11-28 at 11 42 29 AM" src="https://github.com/headlamp-k8s/headlamp/assets/24803604/0efbb4bc-7d39-4451-8fa8-af1b2431b886">

### Error

The `node_modules` folder is not available for the absent plugins. I also tried locally it's the same error. Attaching the screenshot for the same.

<img width="1335" alt="Screenshot 2023-11-28 at 11 44 35 AM" src="https://github.com/headlamp-k8s/headlamp/assets/24803604/47073f51-60c8-43c6-b899-b06b8a2e4cb4">

PR #1568, opened 2023-11-28
ci: Fix plugins installation

knrt10:
When building docker image for plugins we were using npm latest version of @kinvolk/headlamp-plugin which was resulting in unsynced behaviour if any plugin changes were made. To solve this we are now building example plugins with the latest changes of headlamp-plugin.

Fixes: #1567

### DIFF
--- a/Dockerfile.plugins
+++ b/Dockerfile.plugins
@@ -1,21 +1,37 @@
+# Build the plugin
 FROM node:16.4 as builder
 
 WORKDIR /headlamp-plugins
 
+# Copy the example plugins source code to the working directory
 COPY ./plugins/examples /headlamp-plugins/
 
+# Create a build directory
 RUN mkdir -p /headlamp-plugins/build
 
-# Build the plugin
+# Copy the plugin source code to the working directory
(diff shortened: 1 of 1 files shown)


yolossn:
Built the container image and checked if plugins exist.
image

yolossn (approved):

Dockerfile.plugins
```diff
@@ -1,21 +1,38 @@
...
 RUN mkdir -p /headlamp-plugins/build
 
-# Build the plugin
+# Copy the plugin source code to the working directory
+COPY ./plugins/headlamp-plugin /headlamp-plugins/headlamp-plugin
+
+# Change to the headlamp-plugin directory, install dependencies,
+# and link the local plugin to the global headlamp-plugin, so that
+# we can use the headlamp-plugin to build the example plugins
+# with the latest changes.
```
yolossn, line 18: comment mentions link, please update it otherwise good to go.

knrt10: Edited and updated PR.

Issue #1349, opened 2023-08-23
Add more clarity as to why kubeconfigs are not being loaded correctly


joaquimrocha:
Today a user reported that loading a kube config was broken and shared from YAML that allows to reproduce that.
I looked into it and it seems like initially there were some values in the kubeconfig that were dummy data (redacted) as numbers and the k8s lib we use in the backend failed to parse the values as it expected strings.

Then I replaced those numbers with dummy strings and still things failed because apparently the lib is attempting to decode the `certificate-authority-data` and `client-certificate-data` but I set those strings without being encoded in base 64.

Of course users should try to load valid kubeconfigs, but we can also do a better job at reporting exactly why we cannot load a YAML file. That's what this issue is about.

Here is a YAML that's broken:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: 123
    server: https://0.0.0.0
  name: cluster1
- cluster:
    certificate-authority-data: hello-world
    server: https://0.0.0.0
  name: cluster2
contexts:
- context:
    cluster: cluster1
    user: cluster1
  name: cluster1
- context:
    cluster: cluster2
    user: cluster2
  name: cluster2
kind: Config
users:
- name: cluster1
  user:
    client-certificate-data: 123
    client-key-data: 123
    token: e12345
- name: cluster2
  user:
    client-certificate-data: a123
    client-key-data: dummy
    token: e12345
```

Acceptance Criteria:
- [ ] When a kubeconfig cannot be loaded: show translatable/human-readable errors as much as possible;
- [ ] When a translatable error is not possible (because it's an error we have not accounted for), we should still display the error string as an explanation, even if it's always in English.

knrt10:
Working on this.

PR #1569, opened 2023-11-30
backend: human readable errors for kubeconfig

knrt10:
When user was loading invalid kubeconfig it was difficult to understand what was wrong with the kubeconfig. This addition helps returing human readable errors to some extent.

Fixes: #1349


backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -342,3 +354,52 @@ func LoadAndStoreKubeConfigs(kubeConfigStore ContextStore, kubeConfigs string, s
...
+			Condition: func(e error) bool {
+				return strings.Contains(e.Error(), "cannot unmarshal number into Go struct field") &&
+					(strings.Contains(e.Error(), "client-certificate-data") ||
+						strings.Contains(e.Error(), "client-key-data") ||
+						strings.Contains(e.Error(), "certificate-authority-data"))
+			},
+			Message: "invalid certificate or key data values in kubeconfig",
+		},
+		{
+			Condition: func(e error) bool {
```
yolossn, line 377: when there is an error with `client-certificate-data` both the conditions pass and create an bit ambiguous message
image

knrt10: yes, the purpose was to append all the errors of the kubeconfig rather than the first hit. Users can have multiple user info in kubeconfig. So I appended those errors. WDYT?

knrt10:
Thank you for the suggestion. Yes using `clientcmd.Validate, clientcmd.ConfirmUsable` makes more sense. Will update it

PR #1581, opened 2023-12-04
plugins: dynamic clusters example plugin update

knrt10:
Previously it only had a button with was taking hard coded values and setting those values. Now backend sends information if the cluster has dynamic clusters enabled or not. If it is enabled user are able to store the kubeconfig statlessly.

It adds input options for setting values accordingly by the information received from backend about isDynamicClusterEnabled.

#### If dynamic clusters is enabled from backend

Screenshot 2023-12-04 at 8 02 24 PM

#### If dynamic clusters is disabled from backend

Screenshot 2023-12-04 at 8 04 05 PM


knrt10:
Closing this in favour of https://github.com/headlamp-k8s/headlamp/pull/1303/commits/eec306afd19b3252169c428c0a301a58e45e0489

PR #1583, opened 2023-12-05
Handle frequent update for events in overview section

ashu8912:


frontend/src/lib/k8s/event.ts
```diff
@@ -60,6 +60,12 @@ class Event extends makeKubeObject<KubeEvent>('Event') {
     return this.getValue('message');
   }
 
+  static async listEvents(queryParams: { [key: string]: any }) {
```
joaquimrocha, line 63: We have the useList as a consistent way to list items. If this is needed, it should be very clear what is the reason for that. And knowing that reason maybe shows a way we can fix anything in that function that is not working in this case.

joaquimrocha (changes requested):

frontend/src/lib/k8s/event.ts
ashu8912: The useList hook connects to a web socket stream, When we use limit query to fetch only 2000 events the problem is the stream actually can't limit the no of events itself, either we will have to manually close the connection when the no of events reaches to 2000 but that means special casing the useList hook for just the limit query.

PR #1586, opened 2023-12-06
Ship official plugins in image

joaquimrocha:
We want to ship plugins with the official container image as well, but these will be only plugins that are safe in a non-desktop environment, so for now this means the prometheus plugin.

## How to test

1. Build and run the image from this branch, make sure you can access a cluster, then go to a workload, like a deployment: verify that the prometheus charts are enabled in a workload's details view

### DIFF
--- a/container/build-manifest.json
+++ b/container/build-manifest.json
@@ -1,0 +1,8 @@
+{
+  "plugins": [
+    {
+      "name": "prometheus",
+      "archive": "https://github.com/headlamp-k8s/plugins/releases/download/v0.1.4/prometheus-0.0.1.tgz"
+    }
+  ]
+}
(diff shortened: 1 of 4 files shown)


container/build-manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "plugins": [
+    {
+      "name": "prometheus",
+      "archive": "https://github.com/headlamp-k8s/plugins/releases/download/v0.1.4/prometheus-0.0.1.tgz"
```
illume, line 5: We will need to update this when the PR is merged updating the plugin to mui5 and a new release of it is made.

illume:
I could confirm that plugins are being put into the /headlamp/static-plugins/prometheus folder.

```bash
make image
# get the image id.
docker images -a
docker create --name temp_container 7afde1de366f
docker start temp_container


$ docker exec temp_container ls /headlamp/static-plugins/prometheus
main.js
package.json
```

Issue #1230, opened 2023-07-10
duplicate id: editor-dialog-title


illume:
When editing something... There is a duplicate id: "editor-dialog-title"

```REACT_APP_SKIP_A11Y=false make run-frontend```
http://localhost:3000/c/minikube/storage/classes/standard

![Screenshot 2023-07-10 134601](https://github.com/headlamp-k8s/headlamp/assets/9541/28c03ccc-53b2-432f-88e0-47c08cb36e46)

knrt10:
I can reproduce this, working on it.

PR #1589, opened 2023-12-12
frontend: Add unique IDs to dialog title and tabs

knrt10:
When editing something there were duplicate IDs. The IDs are not generated once when the component mounts and it's used as suffix for each dynamicID.

This should ensure that IDs are unique and conform to valid values.

Fixes: #1230

### Testing (Done)

- Go to page http://localhost:3000/c/minikube/storage/classes/standard and click on edit button


frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -314,6 +314,8 @@ export default function EditorDialog(props: EditorDialogProps) {
       : t('translation|Edit: {{ itemName }}', { itemName });
   }
 
+  const [dialogTitleId] = React.useState(`editor-dialog-title-${generateUniqueID()}`);
```
joaquimrocha, line 317: We have a `useId` in the utils.ts file.

knrt10: Thanks for the help, updated.

joaquimrocha (approved):

PR #1591, opened 2023-12-14
frontend: Add the ability to specify the cluster in apiProxy methods

### LINKED ISSUE #1472: Allow to pass the cluster as an argument to resource management functions
When we list/get resources, the cluster is taken from the context we are in.
However, given we have the new Home view, where we are not in a cluster context and yet maybe plugin devs want to show data for each cluster, we should have a way to pass the cluster as an argument for resource related functions.

A version of this is implemented in the aggregated-view branch already, so we should clean it up and publish it.

fixes #1459

joaquimrocha:
We need ot be able to specify the cluster so we can later call the
apiProxy related methods on any cluster instead of just the currently
set up one.

fixes #1472

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -467,7 +467,14 @@ async function repeatStreamFunc(
     ...args: any[]
   ) {
     const endpoint = apiEndpoints[endpointIndex];
-    return endpoint[funcName as keyof ApiFactoryReturn](...args, errCb);
+    const fullArgs = [...args];
+    let errCbIndex = 2;
+    if (endpoint.isNamespaced) {
+      ++errCbIndex;
+    }
+    fullArgs.splice(errCbIndex, 0, errCb);
+
+    return endpoint[funcName as keyof ApiFactoryReturn](...fullArgs);
   }
 
   let endpointIndex = 0;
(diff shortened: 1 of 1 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -607,23 +616,30 @@ function singleApiFactory(group: string, version: string, resource: string) {
...
+    ) => {
       if (isDebugVerbose('k8s/apiProxy@singleApiFactory list')) {
-        console.debug('k8s/apiProxy@singleApiFactory list', { queryParams });
+        console.debug('k8s/apiProxy@singleApiFactory list', { cluster, queryParams });
       }
 
-      return streamResults(url, cb, errCb, queryParams);
+      return streamResultsForCluster(url, { cb, errCb, cluster }, queryParams);
     },
     get: (name: string, cb: StreamResultsCb, errCb: StreamErrCb, queryParams?: QueryParameters) =>
```
illume, line 631: I wonder why the GET don't get a cluster?

illume (commented):
I'm not sure how to test this really. Just reading the code it looks ok.

We really should have some sort of tests for this code in the future... maybe an example plugin using the cluster name parts. Or using the functionality somehow in the app.

frontend/src/lib/k8s/apiProxy.ts
joaquimrocha: The main motivation was the listing functions. I think it makes sense to have the `get` and other functions receive the cluster too.
Let's do that in a later iteration.

joaquimrocha:
I have added the cluster arg to the missing functions but noticed that getting CRs is broken. Making this WIP for further fixes.

joaquimrocha:
I have found and fixed the issue I had encountered before. Should be good to review.
I really wish this is the last time we have to touch apiProxy and that we very soon implement a saner version of it.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -852,38 +907,52 @@ export function post(
...
-  return request(url, opts, autoLogoutOnAuthError);
+  const cluster = getCluster() || '';
+  const opts = { method: 'POST', body, headers: JSON_HEADERS, cluster, ...requestOptions };
+
+  return clusterRequest(url, { ...opts, autoLogoutOnAuthError });
 }
 
 export function patch(url: string, json: any, autoLogoutOnAuthError = true, requestOptions = {}) {
   const body = JSON.stringify(json);
+  const cluster = getCluster() || '';
```
illume, line 921: This one is not using the clusterName from requestOptions.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -852,38 +907,52 @@ export function post(
...
+    cluster: clusterName || getCluster() || '',
+    ...restOptions,
+  };
+  return clusterRequest(url, opts);
 }
 
 export function remove(url: string, requestOptions = {}) {
-  const opts = { method: 'DELETE', headers: JSON_HEADERS, ...requestOptions };
-  return request(url, opts);
+  const cluster = getCluster() || '';
```
illume, line 953: This one is not using the clusterName from requestOptions.

frontend/src/lib/k8s/apiProxy.ts
illume: ok

illume:
> We really should have some sort of tests for this code in the future... maybe an example plugin using the cluster name parts. Or using the functionality somehow in the app.

@joaquimrocha mentioned he is working on replacing this code in a separate piece of work. We agreed that tests for this can be skipped and we put extra time into manually testing this.

PR #1593, opened 2023-12-14
frontend apiProxy: Add limitting to the streamResults function

joaquimrocha:
If we call a function like "list" with the "limit" request param, that will work in the first call that the listing performs internally, but then it connects through a websocket (to stream any changes to that list), and from that point the limit is no longer respected. This means that a listing with a limit of X can get way more items as it remains streaming.

This patch fixes that by doing a client side limiting on the results coming from the streaming, guaranteeing that only the most recent X items are returned. If no limit has been set, then it behaves as previously (no limits are applied at all).

This fixes the case we found out where the UI gets very slow in the cluster overview when the cluster has loads of events getting fired, since our request for limiting the events to 2000 was not working.

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -974,6 +974,12 @@ export async function streamResults(
     console.debug('k8s/apiProxy@streamResults', { url, queryParams });
   }
 
+  // -1 means unlimited.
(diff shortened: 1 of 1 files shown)


joaquimrocha:
@ashu8912 I think this fixes the case you had opened the #1583 for.
Can you verify this fixes the issues?

PR #1596, opened 2023-12-15
frontend: Export the Resource's ResourceListView and AuthVisible

farodin91:


joaquimrocha (approved):
I think these make sense to be added. Let's wait for CI to run but it LGTM.

joaquimrocha:
Ah, this passed CI to my surprised (we have a test to check if all things in common/ are marked to be exported or intentionally not exported). So we should check why the test is not failing. Nothing wrong with the PR but I'll open an issue to verify it.

Issue #719, opened 2022-09-27
[RFE] Improve SimpleTable's page switcher to allow setting pages


joaquimrocha:
## Current situation

As reported by a user:
> Suppose there are 100+ pages of pods in Pod List Page. Users have to switch page one by one for 50 times to view the pods on page 51. We need to jump to page 51 directly.

## Impact

This impedes users from advancing to different pages in a table using the UI.

## Ideal future situation

The current page is editable, so users can type in the desired page to be rendered.

PR #1598, opened 2023-12-15
rontend SimpleTable: Show first/last page buttons in pagination

farodin91:
Part of #719


joaquimrocha (changes requested):
Thanks @farodin91 . Can you give the commit a more descriptive title, like: "frontend SimpleTable: Show first/last page buttons in pagination"?

joaquimrocha (approved):

PR #1602, opened 2023-12-16
frontend: cleanup pod, workloads, and job

farodin91:
* frontend workload: Show namespace filter 
* frontend pod: Use ActionButton
* frontend job/cronjob: fix namespace handling

Follow up idea:
* move queryData into `TableFromResourceClass`
* move hideColumns into `Table`


frontend/src/components/job/List.tsx
```diff
@@ -55,18 +56,16 @@ export function makePodStatusLabel(job: Job) {
...
+  jobs: Job[] | null;
+  error: string | null;
+  hideColumns?: 'namespace'[];
+  reflectTableInURL?: SimpleTableProps['reflectInURL'];
+  noNamespaceFilter?: boolean;
 }
 
 export function JobsListRenderer(props: JobsListRendererProps) {
-  const { jobs = null, error } = props;
+  const { jobs, error, hideColumns = [], reflectTableInURL = 'pods', noNamespaceFilter } = props;
```
joaquimrocha, line 68: I think the `reflectTableInURL` here should be called `jobs`, rigth?

joaquimrocha (changes requested):
Very good changes. Thank you @farodin91 !
I left a comment about the default name for the jobs pagination in the URL.
Normally we also want to have these changes in atomic commits. The reason is:
1. It's easier to spot the commits if they have descriptive titles (e.g. `frontend workload/Overview: Show namespace filter`) 
2. If there's a bug in one of the changes and we need to revert, we don't have to revert all the changes
3. Easier to fix conflicts if they happen

Let me know if you can separate these changes in atomic commits.

farodin91:
@joaquimrocha Others changes are splitted into 3 different commits.

joaquimrocha (approved):
LGTM. Thank you @farodin91 .

PR #1603, opened 2023-12-16
frontend: add node shell

### LINKED ISSUE #996: Add support for execing into a node
We should support for execing into a node. i.e. running a container in the specific node from a node's details view.

Acceptance Criteria: TBD

farodin91:
image

image

Follow up PR:
* Windows Support
* Improved Error handling

fixes #996


farodin91:
@joaquimrocha What do you think of this?

frontend/src/lib/k8s/node.ts
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
+import { v4 as uuidv4 } from 'uuid';
```
joaquimrocha, line 2: We have already some places where we use unique strings without uuid. See the [spawned jobs creation](https://github.com/headlamp-k8s/headlamp/blob/4d09f00a142842ef677db9fd790a7f429eb63c90/frontend/src/components/cronjob/Details.tsx#L39) for example.

joaquimrocha (changes requested):

frontend/src/components/common/Shell.tsx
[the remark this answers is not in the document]
farodin91: merged

frontend/src/lib/k8s/node.ts
farodin91: dep removed

frontend/src/components/node/Details.tsx
```diff
@@ -194,6 +196,19 @@ export default function NodeDetails() {
...
           },
+          {
+            id: DefaultHeaderAction.NODE_SHELL,
+            action: (
+              <ActionButton
+                description={t('Drop Node Shell')}
+                icon="mdi:console"
+                onClick={() => setShowShell(true)}
+                iconButtonProps={{
+                  disabled: item?.status?.nodeInfo?.operatingSystem !== 'linux',
```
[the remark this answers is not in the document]
farodin91: What do you think of my change?

frontend/src/components/node/Details.tsx
```diff
@@ -194,6 +196,19 @@ export default function NodeDetails() {
               </AuthVisible>
             ),
           },
+          {
+            id: DefaultHeaderAction.NODE_SHELL,
+            action: (
+              <ActionButton
```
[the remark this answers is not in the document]
farodin91: What do you think of my change?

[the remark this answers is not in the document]
joaquimrocha: I will try to take a look tomorrow. Thanks a lot for always following up on your PRs! 🙏

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -295,6 +326,35 @@ export default function SettingsCluster() {
                 </>
               ),
             },
+            {
+              name: t('translation|Drop Node Shell Linux Image'),
```
joaquimrocha, line 330: I think "Drop shell" will be difficult to translate to other languages. But we may ignore the "drop" in those languages I guess.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -295,6 +326,35 @@ export default function SettingsCluster() {
...
+                <TextField
+                  onChange={event => {
+                    let value = event.target.value;
+                    value = value.replace(' ', '');
+                    setDropShellLinuxImage(value);
+                  }}
+                  value={dropShellLinuxImage}
+                  placeholder={DEFAULT_DROP_SHELL_LINUX_IMAGE}
+                  helperText={t(
+                    'translation|The default image is used for dropping a shell into a node (when not specified directly).'
```
joaquimrocha, line 341: What does "when not specified directly" refer to?

frontend/src/components/node/Details.tsx
```diff
@@ -194,6 +201,30 @@ export default function NodeDetails() {
...
+            id: DefaultHeaderAction.NODE_SHELL,
+            action: (
+              <AuthVisible authVerb="create" item={Pod} namespace={'kube-system'}>
+                <AuthVisible item={Pod} namespace={'kube-system'} authVerb="get" subresource="exec">
+                  <ActionButton
+                    description={
+                      isLinux(item)
+                        ? t('Drop Node Shell')
+                        : t(
+                            'Drop Node Shell is not support on this operatingSystem: {{ device }}',
```
joaquimrocha, line 214: Let's shorten/fix it to `Node shell is not supported in this OS: {{ nodeOS }}`...

frontend/src/helpers/index.ts
```diff
@@ -306,9 +306,12 @@ function getProductName(): string | undefined {
   return process.env.REACT_APP_HEADLAMP_PRODUCT_NAME;
 }
 
+export const DEFAULT_DROP_SHELL_LINUX_IMAGE = 'docker.io/alpine:3.19';
```
joaquimrocha, line 309: Should we use the `latest` tag? I guess using the latest tag we risk users having to pull when the image is not around but has the advantage that we don't have to keep updating it?

frontend/src/i18n/locales/de/glossary.json
```diff
@@ -97,6 +97,8 @@
   "Uncordon": "Planungssperre aufheben",
   "Cordon": "Planungssperre setzen",
   "Drain": "Entleeren",
+  "Drop Node Shell": "Drop Node Shell",
+  "Drop Node Shell is not support on this operatingSystem: {{ device }}": "Drop Node Shell is not support on this operatingSystem: {{ device }}",
```
joaquimrocha, line 101: All translations except the English are supposed to be empty. So we can at a later point fetch them and translate them. I know this generates a warning in the dev console, but that's okay.

frontend/src/lib/k8s/node.ts
```diff
@@ -71,6 +109,50 @@ class Node extends makeKubeObject<KubeNode>('node') {
...
+      '((clear && bash) || (clear && zsh) || (clear && ash) || (clear && sh))',
+    ];
+    const tty = true;
+    const stdin = true;
+    const stdout = true;
+    const stderr = true;
+    const commandStr = command.map(item => '&command=' + encodeURIComponent(item)).join('');
+    const url = `/api/v1/namespaces/kube-system/pods/${podName}/exec?container=shell${commandStr}&stdin=${
+      stdin ? 1 : 0
+    }&stderr=${stderr ? 1 : 0}&stdout=${stdout ? 1 : 0}&tty=${tty ? 1 : 0}`;
```
joaquimrocha, line 140: The 1st time I tried your PR, it failed to show any terminal. I think it's because the image was not pulled and it took longer? After re-trying it did work. Maybe you can reproduce it by using an always pull policy in the pod during dev.

I wonder if we can somehow show a spinner and wait till the pod is available, for a better UX.

frontend/src/components/common/Terminal.tsx
```diff
@@ -321,32 +417,34 @@ export default function Terminal(props: TerminalProps) {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
     [container, terminalContainerRef, shells, props.open]
   );
 
-  React.useEffect(
-    () => {
-      if (props.open && container === null) {
-        setContainer(getDefaultContainer());
+  if (isPod(item)) {
+    React.useEffect(
```
joaquimrocha, line 429: Hooks aren't supposed to be run conditionally as this leads to problems.

frontend/src/components/common/Terminal.tsx
joaquimrocha, line 1: I think making the Terminal dependent on a node or pod is not the right direction here. After all we are execing into a pod.
I wonder if instead of passing the node so it executes the shell command there, we could instead override the `exec` method for that pod instance, either directly in that instance, or by creating a new Pod class (NodeShellPod) with that new exec method, and we instantiate it from the json of the created Pod.

I understand the exec is not the only thing you're branching on. But e.g. things like hiding the container selector could be a new property like `showContainerSelector?: boolean`; not circling through the shells could be achieved by overriding the list of shells from a property too, and if this list has only 1 element, there's no point in circling (`shellCommand?: string[] | string`); the title can be a new title property.

I hope we don't end up with 10 more properties but perhaps just 3 or 4. This would make the terminal more flexible without special casing it for the node. These changes should be independent commits.

I know it's yet more work, but do you think this makes sense?

joaquimrocha (changes requested):
Pretty cool to have this feature! I left a few comments. Let me know if you have any questions or need to discuss it (we can also do it in the community slack).

farodin91:
I'm vacation at moment. I will come back in one half week.

frontend/src/components/common/Terminal.tsx
farodin91: I started to work on it. 

I'm currently think of splitting up the terminal any logic such as attach or not or container selector and building a Terminal dialog for pods and nodes. What do you think?

joaquimrocha: Having a Terminal dialog for pods and nodes is fine if that makes the code simpler, as long as they use the same underlying component. Since we are in any case execing into the same type of resource, it's just some options on how to connect and what to show that change, I think it's probably easier to just have those exposed in the Terminal and calling it differently depending on whether it's the Node shell or the Pod one.

joaquimrocha (commented):
I commented on your last reply. I see you updated the branch but the logic still needs to address the review, right?

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -127,10 +128,32 @@ export default function SettingsCluster() {
     };
   }, [userDefaultNamespace]);
 
+  React.useEffect(() => {
+    let timeoutHandle: NodeJS.Timeout | null = null;
+
+    if (isEditingDropShellImage()) {
+      // We store the namespace after a timeout.
```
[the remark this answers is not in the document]
illume: This is done, resolving.

frontend/src/components/App/Settings/SettingsCluster.tsx
illume: This has been addressed. Resolving.

frontend/src/components/node/Details.tsx
illume: This has been addressed. Resolving.

frontend/src/components/common/Terminal.tsx
illume: This was addressed. Resolving.

frontend/src/i18n/locales/de/glossary.json
illume: This has been addressed. Resolving.

frontend/src/helpers/index.ts
illume: Addressed. Resolving.

frontend/src/components/node/Details.tsx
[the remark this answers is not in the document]
illume: This was addressed. Resolving.

illume:
I went through and closed convos that were done.  It should be a bit easier to read the PR now.

farodin91:
@illume  I have rebased my changes and now they pass.

farodin91:
I thought about moving it to plugins. As a Plugin it would be possible to ship only if wanted and develop further. Any thoughts?

farodin91:
@joaquimrocha Moved to plugins: https://github.com/headlamp-k8s/plugins/pull/133

joaquimrocha:
@farodin91 We discussed whether this feature makes sense as a core one or plugin and we all agree it belongs to the core. I know at the time we just had a different idea of how to reflect this in the dialog, where I had mentioned that since the Terminal accepts a pod, we could not have it accept a node but instead have a wrapper that does accept the node and runs the needed pod from it, and after that we should be able to load the default Terminal. And it's totally okay to add properties to the Terminal to override something like the title, etc.
In any case, now that the team is bigger we should be able to get more eyes on it if you update the PR. Cheers.

farodin91:
@joaquimrocha I think i need to fix this https://github.com/headlamp-k8s/headlamp/pull/2776 before can update this PR properly.

farodin91:
@joaquimrocha I based the new implementation based on my plugin code.

farodin91:
Small update to design of the settings view.
image

farodin91:
@illume Would you like to review?

joaquimrocha (changes requested):
Thanks @farodin91 . I tried the PR and the shell works fine. Maybe we should also add similar controls to the regular pod shell.
OTOH, something seems wrong with the settings (see screenshot). Maybe it's an update issue that gets fixed once you rebase the PR.
Screenshot showing 2 entries for the cluster selector in the settings view

farodin91:
@joaquimrocha I rebased and fixed the issue.

joaquimrocha:
Thank you so much @farodin91 ! Can you open issues for the follow up changes: the ones you added to the PR description + making the terminals more consistent between Pod and Node?

/lgtm
/approve

PR #1605, opened 2023-12-17
frontend: Add IngressClass

farodin91:
Ingress List view
image

Ingress Details view
image

IngressClass List view
image

IngressClass Details view
image


frontend/src/components/ingress/ClassDetails.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <DetailsGrid
+      resourceType={IngressClass}
+      name={name}
+      withEvents
+      extraInfo={item =>
+        item && [
+          {
+            name: t('translation|Default'),
+            value: item.isDefault ? 'Default' : '',
```
joaquimrocha, line 19: I see that the Default column is translated, but the value in the columns is not. Maybe in this case we just add a translatable "Yes" and "No" according to the result.

joaquimrocha (approved):
Left a comment about the label for the default. Otherwise, it LGTM! Thank you for another nice change, @farodin91 .

joaquimrocha (approved):

joaquimrocha:
Great. Thanks.

PR #1611, opened 2023-12-18
frontend: ResourceTable support hideColums

farodin91:


farodin91:
I don't think so, as i have no idea to provide infos from the view side. 

Isn't more change columns or add columns on all places for on type.

farodin91:
@joaquimrocha Any feedback? 

As far as i can tell, tableProcessors is should be used by plugins to enchance the columns.

farodin91:
@joaquimrocha the table processor can do all of this stuff. I need one which is not type specific, it must be view specific and this not possible as far as i understand, see https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/common/Resource/ResourceTable.tsx#L130. I could allow to add tableprocessor on definition, i thought that would to much changes.

illume:
> @joaquimrocha 
> The table processor can be used to process the columns, meaning they may insert/delete/change columns. I understand your change seems simple enough but I wonder if this property is too custom for the use-cases we have in a couple of components when it can easily be accomplished by just processing or modifying the columns data itself, as those components are doing.
> 
> Let me bring in a 3rd opinion: @illume WDYT?

Perhaps hideColumns is common enough to benefit from a specific property? Table processors are flexible, but are a lot to use/understand compared to just giving a list of columns to hide. An alternative for making it easier with table processor... would be to improve the docs with an example of implementing hideColumns with a table processor.


> @farodin91 
> @joaquimrocha the table processor can do all of this stuff. I need one which is not type specific, it must be view specific and this not possible as far as i understand, see https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/common/Resource/ResourceTable.tsx#L130. I could allow to add tableprocessor on definition, i thought that would to much changes.

Indeed.

That reminds me of something that might work for this... We do have an id available to a table processor id. This can be used to make the processor only apply to 'pods' table. See example of this:
https://github.com/headlamp-k8s/headlamp/blob/main/plugins/examples/tables/src/index.tsx#L58

PR #1614, opened 2023-12-18
frontend: Include LimitRange and ResourceQuota on Namespace Details

farodin91:
based on #1611 

From workflow perspective, it could be interesting to see ResourceQuotas and LimitRanges per Namespace. As they are mainly for all stuff inside the namespace.

What does anyone think?

image

### DIFF
--- a/frontend/src/components/common/SectionFilterHeader.tsx
+++ b/frontend/src/components/common/SectionFilterHeader.tsx
@@ -110,52 +110,54 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
     actions.push(...preRenderFromFilterActions);
   }
 
-  if (!showFilters.show) {
-    actions.push(
-      <IconButton
-        aria-label={t('Show filter')}
-        onClick={() => setShowFilters({ show: true, userTriggered: true })}
-        size="medium"
-      >
-        <Icon icon="mdi:filter-variant" />
-      </IconButton>
-    );
-  } else {
-    actions.push(
-      <Grid container alignItems="flex-end" justifyContent="flex-end" spacing={1} wrap="nowrap">
-        {!noNamespaceFilter && (
+  if (!noSearch) {
+    if (!showFilters.show) {
+      actions.push(
+        <IconButton
+          aria-label={t('Show filter')}
+          onClick={() => setShowFilters({ show: true, userTriggered: true })}
+          size="medium"
+        >
+          <Icon icon="mdi:filter-variant" />
+        </IconButton>
+      );
+    } else {
+      actions.push(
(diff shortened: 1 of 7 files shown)


joaquimrocha:
@farodin91 Sorry I clicked the Ready for Review button by mistake. Let me know if this is actually ready to be reviewed so I take a look. I agree that having the ResourceQuotas and LimitRanges is good (and I think it's what kubectl does too, rigth?).

farodin91:
It depends on https://github.com/headlamp-k8s/headlamp/pull/1611.

farodin91:
@joaquimrocha This is now rebased.

joaquimrocha:
> @joaquimrocha Any feedback?

Sorry. Completely dropped the ball on this one.

I had checked the code and it looked fine. I have tested now which is what I still had to do, and the only problem is that the table search filter is not designed to happen multiple times in one view. So the result is that if we use the search filter, it changes the state for all the tables in that view:
screenshot showing the issue

We should fix the filter by allowing individual searches per table, instead of the global redux-based one.

farodin91:
@joaquimrocha Should we fix this in different PR?

joaquimrocha:
> @joaquimrocha Should we fix this in different PR?

To be honest, the UX looks pretty broken with it. So if we merge, we wouldn't be able to release like this. We could however merge if we hide the search filter all along (because then we get a missing feature, rather than a broken one 😉 ).

farodin91:
Search will be hidden on this page for now.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -110,52 +110,54 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
...
-        size="medium"
-      >
-        <Icon icon="mdi:filter-variant" />
-      </IconButton>
-    );
-  } else {
-    actions.push(
-      <Grid container alignItems="flex-end" justifyContent="flex-end" spacing={1} wrap="nowrap">
-        {!noNamespaceFilter && (
+  if (!noSearch) {
```
joaquimrocha, line 113: I thought we had already this capability. If this is needed in the end, please split it into its own commit.

joaquimrocha (changes requested):
Just a comment about splitting out the header changes. The rest looks fine.

frontend/src/components/common/SectionFilterHeader.tsx
farodin91: split into two commits

joaquimrocha:
Hey @farodin91 . Sorry it took me a bit to get to review the update. Can you rebase the branch so I can try it again?

joaquimrocha:
I tried the changes locally and they LGTM. Please rebase and we will merge it.

illume (approved):
👍 thanks

Issue #1597, opened 2023-12-15
Verify why common/index.test.ts is not checking all files in common/


joaquimrocha:
I just realized from #1596 that the common/index.test.ts doesn't check all the files (if it did, that PR would have failed).
So we need to verify why it's not checking all files.

vyncent-t:
- when running tests there is a folder 'auth verify', which isnt included in test.ts (test.ts checks every file to determine if it is to be exported)
- test.ts (catches files that are unintendedly added, stills runs even though there are files that are not tracked by the tests

to run:
- write `cd frontend`
- run `npm run test` 
- press `p`
- choose common folder

PR #1616, opened 2023-12-18
frontend: Fix import test checks

vyncent-t:
Fixes issue #1597

## Description
This PR resolves the issue where `common/index.test.ts` was not checking all files within the `common/` directory. An import checker for the `resource` folder has been added, and tests similar to those used in the `common` folder have been implemented to ensure comprehensive coverage.

## Changes
- [x] Added an import checker for the `resource` folder.
- [x] Expanded test coverage to include all files within the `common/` directory.

## Verification
- Ran `npm run test` along with the `p` flag for the index files to confirm it now checks all files in the `common` and 'resource' directory.
- Verified that the new tests are functioning as expected and provide comprehensive coverage.


vyncent-t:
- Added the missing file names to exports, maybe don't need the export within the index?

vyncent-t:
Was able to export missing checks for the toHaveProperty, exported the MainSectionInfo from the index.tsx file I added to the folder since it could not locate the module

All tests run fine now in both common and resource

frontend/src/components/common/Resource/index.tsx
```diff
@@ -6,3 +6,30 @@ export * from './ResourceListView';
 export { default as ResourceListView } from './ResourceListView';
 export * from './AuthVisible';
 export { default as AuthVisible } from './AuthVisible';
+// adding missing imports based on npm run test
```
joaquimrocha, line 9: Please remove this comment, as it makes sense in the case of this PR but we have to think about the code as a whole.

frontend/src/components/common/Resource/index.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  '.stories',
+  '.test',
+  'index',
+  '__snapshots__',
+  // Not exported on purpose
+  'ReleaseNotes',
+  'ActionsNotifier',
+  'AlertNotification',
+  'ConfirmButton',
+  'NamespacesAutocomplete',
```
joaquimrocha, line 14: We could export these 2 as well IMO.

joaquimrocha (changes requested):
Just 2 comments. We can leave the exportation of the mentioned components for a later PR, but the code comment should go away IMO.

frontend/src/components/common/Resource/index.test.ts
vyncent-t: I noticed that these are missing from their test in the common directory, I have moved them to be included there. Unless it would be better fitting for them to be in the Resource test then I shall move them there.
I figured since it would be testing for the export where it exist that would be the go to move?

illume:
@vyncent-t could you please add some of the info from the PR description into the commit message?

vyncent-t:
I believe this should be the rebased one and added more context!

vyncent-t:
Updated the commit message and rebased to current main

illume (approved):
Thanks! Looks good to me.

(ps. The PR title and description could be updated to include recent changes)

joaquimrocha:
@vyncent-t , Please update the commit title to something like:
`frontend common/Resource: Test module exports`

vyncent-t:
Updated the commit message and rebased to the current main!

PR #1619, opened 2023-12-20
frontend: Use NameValueTable for ConfigMap and Secret

farodin91:
* add tests
* simlar look for secrets and confimaps
* secret add type to details view

secret:
image
configmap:
image

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -7,7 +7,7 @@ import Grid, { GridProps, GridSize } from '@mui/material/Grid';
 import IconButton from '@mui/material/IconButton';
 import Input, { InputProps } from '@mui/material/Input';
 import Paper from '@mui/material/Paper';
-import { TextFieldProps } from '@mui/material/TextField';
+import { BaseTextFieldProps } from '@mui/material/TextField';
 import Typography from '@mui/material/Typography';
 import { makeStyles, useTheme } from '@mui/styles';
 import { Location } from 'history';
@@ -374,8 +374,12 @@ export function SectionGrid(props: SectionGridProps) {
   );
 }
 
-export function DataField(props: TextFieldProps) {
-  const { label, value } = props;
+export interface DataFieldProps extends BaseTextFieldProps {
+  disableLabel?: boolean;
+}
+
+export function DataField(props: DataFieldProps) {
+  const { disableLabel, label, value } = props;
   // Make sure we reload after a theme change
   useTheme();
   const themeName = getThemeName();
@@ -385,8 +389,11 @@ export function DataField(props: TextFieldProps) {
     if (!editorElement) {
       return;
     }
+
     const lineCount = editor.getModel()?.getLineCount() || 1;
-    if (lineCount <= 10) {
+    if (lineCount < 2) {
+      editorElement.style.height = '3vh';
+    } else if (lineCount <= 10) {
       editorElement.style.height = '10vh';
     } else {
       editorElement.style.height = '40vh';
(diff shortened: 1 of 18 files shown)


frontend/src/components/secret/Details.tsx
```diff
@@ -14,11 +17,35 @@ export default function SecretDetails() {
...
+      extraSections={item =>
+        item && [
+          {
+            id: 'headlamp.secrets-data',
+            section: () => {
+              const itemData = item?.data || {};
+              const mainRows = Object.entries(itemData).map((item: unknown[]) => ({
+                name: item[0],
+                value: <SecretField value={item[1]} />,
+              })) as NameValueTableRow[];
```
joaquimrocha, line 36: If you declare the mainRows as `const mainRows: NameValueTableRow[]`, that should still get the right type while not masking any possible mistypes in the actual data.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -398,26 +401,15 @@ export function DataField(props: TextFieldProps) {
...
-          <Editor
-            value={value as string}
-            language={language}
-            onMount={handleEditorDidMount}
-            options={{ readOnly: true, lineNumbers: 'off' }}
-            theme={themeName === 'dark' ? 'vs-dark' : 'light'}
-          />
-        </Box>
-      </Box>
-    </>
```
joaquimrocha, line 420: Assuming this function is accessible from plugins, we shouldn't change it to just ignore the `label` property here, otherwise we may break external plugins' UIs.

joaquimrocha (changes requested):
The changes for the secrets LGTM.
The ones for the ConfigMap I am not entirely sure. I think we should probably give it the same heading we have on containers and volumes:
screenshot of the containers in a pod

Of course this style is applied to the NameValueTable but I think we'd use the data name as the title here and then just have the editor within that area, instead of a key->value style as the NameValueTable expects. In fact, the NameValueTable allows to hide the value, so we may be able to accomplish what I mean by passing the editor as the key property and hide the value.

farodin91:
I'm a bit confused why you want to use a different style for secret and configmap, if both data sections are key/values maps.

If you have an issue with large data sets on the value side. Pod -> volumes-> Sources could be already be huge.

I can change the design, but i want to understand why.

joaquimrocha:
I see your point that configmaps and secrets are very similar and thus should have very similar UIs. I just think that the ConfigMaps tend to have larger data but I may be wrong.

I think the UI I suggested looks better/simpler, because it scales things better horizontally and doesn't have so many borders encapsulated within other borders, but actually I think we can do this later and add what you have created, as it looks better than what we have ATM.

joaquimrocha:
We still need to address the comments I had left, especially the one which may break functionality for any other users of it.

farodin91:
I will address the comments.

frontend/src/components/secret/Details.tsx
joaquimrocha: This has not been addressed (same for the similar change in the next commit).

farodin91: Oh it seems i reverted this.

joaquimrocha (approved):

joaquimrocha:
@farodin91 The changes LGTM but DCO is complaining you signed with an unexpected email:
```
Summary
Commit sha: c04fdc0, Author: farodin91, Committer: farodin91; Expected "farodin91 github@jan-jansen.net", but got "Jan Jansen jan.jansen@gdata.de".
Commit sha: e443d6c, Author: farodin91, Committer: farodin91; Expected "farodin91 github@jan-jansen.net", but got "Jan Jansen jan.jansen@gdata.de".
Commit sha: af78fe1, Author: farodin91, Committer: farodin91; Expected "farodin91 github@jan-jansen.net", but got "Jan Jansen jan.jansen@gdata.de".
```

(maybe you need to set up these emails in your Github account)

farodin91:
@joaquimrocha dco fixed

joaquimrocha:
> @joaquimrocha dco fixed

Thank you!

PR #1620, opened 2023-12-20
frontend: cleanup storage section

farodin91:
* reorder lists to look the same
* status label on PV and PVC
* links to storageclass in details
* add basic tests
# Screenshots

PV:
image
image
PVC:
image
image


frontend/src/components/storage/ClaimList.stories.tsx
```diff
@@ -4,45 +4,15 @@ import PersistentVolumeClaim, {
...
+import { BASE_PVC } from './storyHelper';
 
 PersistentVolumeClaim.useList = () => {
-  const noStorageClassNamePVC = _.cloneDeep(basePVC);
+  const noStorageClassNamePVC = _.cloneDeep(BASE_PVC);
   noStorageClassNamePVC.metadata.name = 'no-storage-class-name-pvc';
   noStorageClassNamePVC.spec!.storageClassName = '';
 
-  const noVolumeNamePVC = _.cloneDeep(basePVC);
+  const noVolumeNamePVC = _.cloneDeep(BASE_PVC) as KubePersistentVolumeClaim;
```
joaquimrocha, line 15: Declaring the type of the BASE_PVC as it had for basePVC is better than using `as` since `as` can mask any missing members here.

joaquimrocha (approved):
LGTM though I left a comment to improve the use of types.
For similar PRs, we should keep in mind having atomic commits instead.

joaquimrocha (changes requested):
Meant to request the changes for the typing as mentioned.

farodin91:
I'll try to fix on Friday.

farodin91:
> For similar PRs, we should keep in mind having atomic commits instead.

Just splitting into smaller units, or?

joaquimrocha:
> > For similar PRs, we should keep in mind having atomic commits instead.
> 
> Just splitting into smaller units, or?

Atomic as in: a commit should be related to one fix or one feature that's contained in that commit, rather than mixing different features/fixes in one bigger commit.

farodin91:
@joaquimrocha splitted into multi atomic PRs and typing fixed.

frontend/src/components/storage/VolumeDetails.stories.tsx
joaquimrocha, line 1: Not to be pedantic here, but normally our commit titles are like:
```
frontend storage/VolumeDetails: Add link to StorageClass

Also adds Storybook coverage.
```

This could also be two commits: Add link  + Add storybook coverage, but it's okay.

frontend/src/components/storage/ClaimList.stories.tsx
joaquimrocha, line 1: For future reference, we try to keep commit titles (and body) under 72 chars length. See also my other comment about the commit message referring to the main thing a commit does, and that the body can be used to elaborate on secondary things.

frontend/src/components/storage/storyHelper.ts
joaquimrocha, line 1: Ah, I see you added the fix here.
As a good practice, we don't add commits fixing things introduced in the very same PR. The reasons for these are:
1. Makes reviewing more difficult as we tend to review commit by commit.
2. Adds "wrong" changes to the history, so if we do a git bisect, we may run into problematic commits because they get fixed in a later commit. While this is inevitable throughout the development, we can avoid it in PRs.

Hope it makes sense.

joaquimrocha (changes requested):
Thanks @farodin91 for splitting things up.
I made some comments about the commit format and others. I know these may sound a bit nitty, but I hope they make sense to you as making things closer in format will make for an easier radibility/bug-fix later on.

farodin91:
@joaquimrocha I tried to cleanup commits.

joaquimrocha (approved):
LGTM! Thanks a lot @farodin91 for having changed the commits/format.
@illume , the Mac build is failing on an svgo dep. Weird that it's the only one that has failed.

I am tempted to just merge this and figure out the build later as it seems unrelated, but would like your opinion.

farodin91:
I would say it's a timeout issue or. It could help to restart of the failed job?

farodin91:
@joaquimrocha  rebase did work.

PR #1624, opened 2023-12-20
WIP: backend/pkg/plugins: Add CheckHeadlampPluginVersion

illume:
For checking plugin version compatibility.

WIP: This should only fail when run in-cluster mode, not in app mode. Because those delivering an app with bundled plugins would consider this an error if the in-cluster app ran without plugins.

## How to test?

- [ ] there is a unit test
- [ ] add an old plugin (less than 0.8.0-alpha.10), and the backend should fail to run with an error explaining the problem
- [ ] it should *NOT* exit with an error when run in app mode
- [ ] it *SHOULD* exit with an error when run in-cluster mode

The backend should fail when in-cluster mode.
```
# checkout the previous release and install an old plugin
git checkout 
cd plugins/examples/pod-counter/
npm i
npm start

# running the backend should fail because old plugin
cd ../../../
make run-backend
```

Test the app should not error:

```
# checkout the previous release and install an old plugin
git checkout 
cd plugins/examples/pod-counter/
npm i
npm start

# running the app should NOT fail because of an old plugin
cd ../../../
cd app
npm start
```

### DIFF
--- a/backend/pkg/plugins/plugins.go
+++ b/backend/pkg/plugins/plugins.go
@@ -2,7 +2,9 @@ package plugins
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


joaquimrocha:
I think we moved the plugin compat to the frontend layer, right? So this shouldn't be needed anymore.
cc/ @illume

PR #1627, opened 2023-12-20
Revert "backend: human readable errors for kubeconfig"

joaquimrocha:
This reverts commit c00b184d698e10f4e3c05862018c3b3da8c49ad2. We found out this was causing the clusters not to be sent by the backend at least on Windows and Mac, so we are reverting this for now.


illume:
I guess this requires some certain .kube/config to reproduce.

illume (approved):
👍

PR #1635, opened 2023-12-30
frontend: get rid of most @mui/styles

farodin91:
prepare for react 18

### DIFF
--- a/docs/development/plugins/how-to.md
+++ b/docs/development/plugins/how-to.md
@@ -33,23 +33,20 @@ displays that information in the top bar (i.e. registers an "app bar action").
 ```tsx
 import { K8s, registerAppBarAction } from '@kinvolk/headlamp-plugin/lib';
 import { Typography } from '@mui/material';
-import { makeStyles } from '@mui/styles';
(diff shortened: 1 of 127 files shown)


joaquimrocha:
Hi @farodin91 , can you run the snapshots on this branch and push again?

farodin91:
Updated.

I have to checkout how make test more stable.

joaquimrocha:
@farodin91 Looks like these changes end up with generated css class names that are unique and thus this will break tests, since each run will generate new names and thus not match the snapshots.

I am not sure there's a straight solution to address the unique name generation. I wonder if this is also needed for React 18, but maybe I missed the docs from MUI that indicate that.

farodin91:
https://mui.com/system/styles/api/ here is deprecation notice.

joaquimrocha:
@farodin91  Right. I see that they mention using `sx`. Perhaps this property is more snapshot friendly?

farodin91:
I will test out one with sx.

joaquimrocha:
@farodin91 I guess we should close this in favor of updating the styles using the `sx` way?

Issue #1587, opened 2023-12-08
JWT is not refreshed after it expires


CrimsonFez:
Hello,
I have setup headlamp with keycloak for oidc.
The issue I'm having is that after 5 minutes (my token lifespan) headlamp isn't authenticated anymore.
The ability for headlamp to get a new token with the refresh token should be added.

I don't understand the inner workings of headlamp, but maybe you could use `oidc-ts` to handle oidc tokens.

joaquimrocha:
cc/ @yolossn

Issue #1101, opened 2023-04-22
Auto logout after 5 minutes of inactivity via keycloak login.


jgyy:
I kept receiving these stream of error in console log whenever I login to headlamp via keycloak OIDC, it happens when I am inactive from the webapp for 5 minutes. console logs shown below. Is there a workaround to increase inactivity session duration? I am hosting this headlamp in cluster and it is very inconvenient to keep logging back in so frequently. Thank You!

```
apiProxy.ts:773  Socket closed unexpectedly {path: 'api/v1/persistentvolumes?watch=1&resourceVersion=163140949', args: Array(1)}
g @ apiProxy.ts:773
apiProxy.ts:693 Reconnecting in 3 seconds {url: 'api/v1/persistentvolumes?watch=1&resourceVersion=163140949'}
apiProxy.ts:736  WebSocket connection to 'wss://headlamp.sandbox.k8ns.in/clusters/main/api/v1/events?watch=1&resourceVersion=163140949' failed: 
(anonymous) @ apiProxy.ts:736
d @ apiProxy.ts:686
apiProxy.ts:778  Error in api stream {err: Event, path: 'api/v1/events?watch=1&resourceVersion=163140949'}
```

knrt10:
Fixed via https://github.com/headlamp-k8s/headlamp/pull/1636

PR #1636, opened 2023-12-31
Refresh OIDC token if they are about to expire and use it.

yolossn:
Fixes: #1587 #1101

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


backend/cmd/headlamp.go
```diff
@@ -605,9 +611,168 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+	// check if token is expired
+	exp, ok := payload["exp"].(float64)
+	if !ok {
+		return false
+	}
+
+	// if token is not about to expire, then skip
+	expTime := time.Unix(int64(exp), 0)
+
+	return time.Until(expTime) <= time.Second*10
```
joaquimrocha, line 661: Any particular reason why we're using 10s as the check? Not sure how long the tokens are valid for on average but we probably should try to renew it before such a short period.

backend/cmd/headlamp.go
```diff
@@ -605,9 +611,168 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+}
+
+func refreshAndCacheNewToken(oidcAuthConfig *kubeconfig.OidcConfig,
+	cache cache.Cache[interface{}], token string,
+) (string, error) {
+	const ExtendRefreshTokenTTL = 10 // seconds
```
joaquimrocha, line 667: If this number needs to match the one above, please use a shared const for it.

joaquimrocha (changes requested):

backend/cmd/headlamp.go
yolossn: Not necessary to match the above value, this value is to keep the old refresh token in cache for extended period of time to handle the case were the frontend didn't pickup the new token, and is still using the old token. This way we provide a buffer of 10 seconds for the frontend to use the old token and pickup the new token from the response headers.

joaquimrocha (approved):

PR #1639, opened 2024-01-03
Remove broken format version from the helm repository

joaquimrocha:
By mistake, in version 0.21.0 (Helm version 0.17.0) we pushed the Helm config as having an appVersion of v0.21.0 but this config requires the tag version without the `v`. Since then, ArtifactHub has been giving us alerts that it cannot find the mentioned image.

Since we readily pushed a Helm 0.17.1 version with the right config, I think we should rather just remove the 0.17.0 from the registry to avoid anyone picking it up + fixing the warnings.

Once we merge this, we should remove the Helm 0.17.0 tag/release from the repo.


yolossn (approved):
LGTM

PR #1647, opened 2024-01-09
github-actions: bump all official actions

farodin91:


farodin91:
@joaquimrocha  I could add dependabot support for actions in a follow up PR. What do you think?

illume (approved):
Thanks :) Nice improvement.

Issue #1481, opened 2023-10-19
RFC: Allow to download all logs instead of just the lines currently chosen


joaquimrocha:
Right now the download button for the logs will include only the log lines that are currently selected. However, some users may want to download all the logs.

I think that if the user has chosen a number of lines and has a download button in the same context, it makes sense for the download to include those lines (i.e. download what's being shown). So I think we should offer both options: download X lines, download the full log, but this may be a bit confusing, so maybe just having that download be the full log is the best approach here.

knrt10:
I am working on this.

joaquimrocha:
@lijianzhi01 Please check out how we implemented this. Basically it's just another option for displaying all logs. And the download button should download all logs if that's the option selected.

PR #1658, opened 2024-01-15
frontend: Add option to download all logs

knrt10:
User can now download all logs, there is an option in the dropdown menu of pod's log page.

Fixes: #1481

Screenshot 2024-01-15 at 9 58 46 AM

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -219,6 +219,7 @@ function PodLogViewer(props: PodLogViewerProps) {
                 {i}
               </MenuItem>
             ))}
+            <MenuItem value={-1}>All</MenuItem>
           </Select>
         </FormControl>,
         <LightTooltip
--- a/frontend/src/lib/k8s/pod.ts
+++ b/frontend/src/lib/k8s/pod.ts
@@ -120,8 +120,15 @@ class Pod extends makeKubeObject<KubePod>('Pod') {
       showTimestamps = false,
       follow = true,
     } = logsOptions;
+
     let logs: string[] = [];
-    const url = `/api/v1/namespaces/${this.getNamespace()}/pods/${this.getName()}/log?container=${container}&previous=${showPrevious}&tailLines=${tailLines}&timestamps=${showTimestamps}&follow=${follow}`;
+    let url = `/api/v1/namespaces/${this.getNamespace()}/pods/${this.getName()}/log?container=${container}&previous=${showPrevious}&timestamps=${showTimestamps}&follow=${follow}`;
+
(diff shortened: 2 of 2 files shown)


frontend/src/components/pod/Details.tsx
```diff
@@ -96,38 +97,36 @@ function PodLogViewer(props: PodLogViewerProps) {
   }
   const debouncedSetState = _.debounce(setLogsDebounced, 500, options);
 
-  React.useEffect(
-    () => {
-      let callback: any = null;
+  React.useEffect(() => {
+    let callback: any = null;
```
joaquimrocha, line 101: I see that only one part of this function is effectively changed. Can you keep the format that it had before? (makes the change safer and easier to review)

frontend/src/lib/k8s/pod.ts
```diff
@@ -51,7 +51,7 @@ export interface ExecOptions extends StreamArgs {
 
 export interface LogOptions {
   /** The number of lines to display from the end side of the log */
-  tailLines?: number;
+  tailLines?: number | 'all';
```
joaquimrocha, line 54: I'd prefer to keep it as is and just use a negative number to indicate "all". Yes, it'll need extra comments explaining the special meaning of negative numbers, but I feel like it's safer in the long run to have just one type being treated.

frontend/src/components/pod/Details.tsx
```diff
@@ -96,38 +97,36 @@ function PodLogViewer(props: PodLogViewerProps) {
...
 
   function handleContainerChange(event: any) {
     setContainer(event.target.value);
   }
 
   function handleLinesChange(event: any) {
-    setLines(event.target.value);
+    const value = event.target.value;
+    setShowAll(value === 'all');
+    setLines(showAll ? 0 : parseInt(value, 10) || 100);
```
joaquimrocha, line 129: If we use negative numbers as the way to set up the "all", then we don't need an extra state var and the logic becomes easier IMO.

frontend/src/components/pod/Details.tsx
```diff
@@ -211,14 +210,15 @@ function PodLogViewer(props: PodLogViewerProps) {
...
-            value={lines}
+            value={showAll ? 'all' : lines}
             onChange={handleLinesChange}
           >
             {[100, 1000, 2500].map(i => (
               <MenuItem value={i} key={i}>
                 {i}
               </MenuItem>
             ))}
+            <MenuItem value="all">All</MenuItem>
```
joaquimrocha, line 221: If we use negative numbers as suggested, then this could simply be `-1`.

joaquimrocha (changes requested):

joaquimrocha (approved):
LGTM.

PR #1662, opened 2024-01-15
frontend: Fix fullscreen icon bypassing tests and add missing icon

### LINKED ISSUE #1644: Missing exit full screen in Air Gapped cloud
It seems like exit full screen icon isn't cached properly. 

image

joaquimrocha:
The logic around setting the icon for the fullscreen toggle in the
mentioned component was done by appending the icon name to the "mdi:"
prefix depending on the state. Therefore, this prevented the icon test
from scanning this file correctly, since it looks for a string that
started by "mdi:".

As a cause for this, the fullscreen-exit icon was not added to the
icon list for offline use, which this patch also adds.

Signed-off-by: Joaquim Rocha <joaquim.rocha@microsoft.com>

fixes #1644


vyncent-t (approved):
Ran the icons test on its own with the new changes!

PR #1665, opened 2024-01-15
frontend: Do not apply cluster actions if the cluster has changed

### LINKED ISSUE #1660: Bug: Deletion in the wrong cluster through delay
Reproduce:
1.  click on delete secret
2. switch to different cluster which has a secret with the same namespace and name
3. wait for delay to end
4. secret in switched cluster is gone

joaquimrocha:
This issue fixes #1660 although the right fix is to explicitly set up the cluster for which to call the cluster action on. Till we have that, this change makes things safer.

fixes #1660

### DIFF
--- a/frontend/src/redux/clusterActionSlice.ts
+++ b/frontend/src/redux/clusterActionSlice.ts
@@ -1,6 +1,7 @@
 import { createAsyncThunk, createSlice, PayloadAction } from '@reduxjs/toolkit';
 import i18next from 'i18next';
 import { OptionsObject as SnackbarProps } from 'notistack';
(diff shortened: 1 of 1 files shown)


joaquimrocha:
Hi @farodin91 . Are you able to test this PR and see if it prevents the issue you discovered? (Until we carry the cluster info together with the cluster action)

joaquimrocha:
Update: Now that #1591 is merged. We can fix this issue properly. I will update the PR next week.

farodin91:
@joaquimrocha this pr works fine.

joaquimrocha:
Closing in favor of #1707

PR #1666, opened 2024-01-16
Fix notifications infinite loop

### LINKED ISSUE #1664: Notifications still get into an infinite loop under certain conditions
The notifications are still getting into an infinite loop under certain occasions. This is due to the fact that the way we check that an event needs to go into a notification doesn't mean that the notification will keep/store it in the end. So when the Redux store is updated, we run the logic to check again if a certain event should go into the notification, and since it's not there, we add it, dispatch, and end up in repeating the cycle. i.e. it'll get an infinite loop as long as there are events that are not kept in the notifications.

I will fix this issue.

joaquimrocha:
## How to test
 * Get 200 notifications stored in the local storage. Then go to a cluster with warnings and check the dev console logs to see if there are any indication/warnings about infinite loops.
 * Verify that the latest cluster warning events are shown in the notifications
 * Verify that marking notifications as read and deleting them does work

fixes #1664

### DIFF
--- a/frontend/src/components/App/Notifications/index.tsx
+++ b/frontend/src/components/App/Notifications/index.tsx
@@ -184,7 +184,7 @@ export default function Notifications() {
   const history = useHistory();
 
   useEffect(() => {
-    let notificationsToShow: NotificationIface[] = [];
+    const notificationsToShow: NotificationIface[] = [];
     let currentNotifications = notifications;
     let changed = false;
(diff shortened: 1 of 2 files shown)


farodin91:
@joaquimrocha It looks like this fixes my performance issue with the notification thing.

ashu8912 (approved):

Issue #1667, opened 2024-01-16
Fix layout/styling in Notifications area


joaquimrocha:
The notifications area looks a bit unattended...
![Notifications area screenshot](https://github.com/headlamp-k8s/headlamp/assets/1029635/47612ad8-3947-4342-bcd0-344096dc7f8b)

We can definitely make a better use of that horizontal space to display more of the message from the notification. The cluster should also be a link to the cluster itself.

We should do some small tweaks to make this view UX a bit better.

PR #1674, opened 2024-01-25
frontend: Fix notifications format styles

vyncent-t:
# Enhancements to Notifications List View for Better UX

Solves issue #1667

## Description
This PR focuses on improving the user experience of the Notifications List View. It involves utilizing the available horizontal space more effectively to display a larger portion of the notification message. Additionally, the cluster name in each notification will now be a clickable link, redirecting users directly to the associated cluster.

## Changes
- [x] Expanded the message display area in the Notifications List View to make better use of horizontal space.
- [x] Converted cluster names in notifications into hyperlinks that lead to the respective clusters.

## Verification
- [x] Ensured that more of the notification message is visible in the expanded view.
- [x] Tested the functionality of the new cluster links to confirm they correctly redirect to the appropriate cluster.

## Screenshots
image

### DIFF
--- a/frontend/src/components/App/Notifications/List.tsx
+++ b/frontend/src/components/App/Notifications/List.tsx
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { useTypedSelector } from '../../../redux/reducers/reducers';
-import { DateLabel, SectionBox, SectionFilterHeader, SimpleTable } from '../../common';
+import { DateLabel, Link, SectionBox, SectionFilterHeader, SimpleTable } from '../../common';
 import Empty from '../../common/EmptyContent';
 import { Notification, setNotifications, updateNotifications } from './notificationsSlice';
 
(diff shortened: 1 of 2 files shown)


vyncent-t:
## Dev notes

- _This push is a solid base, I have tweaked the vw for each size to not clip or run into other columns (ignore the colors)_
- _I am using the CellProps and a custom style map to put in custom style changes to achieve this_ 
image
image


- _Currently using useEffect to rerender when any screen size changes, when it changes it grabs a new view size for the message column_ 
image



- The comps are responsive
**Large**
image

**Medium**
image

**Small**
image

vyncent-t:
## Dev note

- [x] Added context links to the notifications view

image

joaquimrocha (changes requested):
The PR approach has an issue with cutting off parts of the table. See the screenshot below:

Screenshot showing the prob

I think we can accomplish a good result with a less engineered effort here. Namely by using `gridTemplate:` for each column as adequate. The text column should just have `auto` so it expands, and the rest should have a `min-content`.
The div around the text column should not have the ` width={'30vw'}`, otherwise it looks artificially cut.

joaquimrocha (approved):
LGTM. The mobile version is not functional but we should fix that in a different occasion anyway.

PR #1676, opened 2024-01-26
load-tests: docs: Add create deployments load test

vyncent-t:
Adds a load test for deployments up to 500
Also fixes typo for load test documents


illume (approved):
👍 thanks

Issue #1670, opened 2024-01-24
Run local commands (desktop only)


joaquimrocha:
It may be useful to get access to running local commands, such as kubectl and others. So we offer a full local terminal inside Headlamp, or allow plugins to implement features based on commands (imagine `minikube`, `az`, etc.).

A nice way to do this without having to imply the backend (which would require us to be careful about not supporting this in a server environment) is to have the app layer (electron) deal with the command running.
I have already a POC from a couple of months ago where this is done, in the [run-local-command](https://github.com/headlamp-k8s/headlamp/tree/run-local-command) branch.

PR #1677, opened 2024-01-29
app: frontend: plugins: Add runCommand for shell commands in Electron app

illume:
So that some local shell commands can be run from inside a plugin.
This communicates the process interaction with the Electron app render process via ipc.

The interface exposed is a subset of the node spawn() which is what is used on the app side. It only implements the parts which are already used by the plugin we have in mind.

The interface supports streaming stdout and stderr, and getting exit codes.

Note, this supports running multiple commands at the same time. The complexities of this are hidden to the user of runCommand.

It limits the commands that can be run to only minikube and az so far. If there is a need for more this can be extended later. This is intended as a way to limit potential misuse.

### How to test?

The app-menu example plugin has been extended to add a minikube menu which displays minikube status.

```bash
cd plugins/headlamp-plugin
npm i
npm run build
npm pack
cd ../examples/app-menus
npm i ../../headlamp-plugin/kinvolk-headlamp-plugin-0.9.0-alpha.2.tgz
npm start
```

Now in a separate terminal

```bash
cd app
npm i
npm start
```

Now with the app-menus example plugin there should be a minikube menu shown which when extended shows minikube status.

image

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -98,7 +98,10 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
  */
 function isWSL(): boolean {
   try {
-    const data = fs.readFileSync('/proc/version', { encoding: 'utf8', flag: 'r' });
+    const data = fs.readFileSync('/proc/version', {
+      encoding: 'utf8',
+      flag: 'r',
+    });
     return data.indexOf('icrosoft') !== -1;
   } catch {
     return false;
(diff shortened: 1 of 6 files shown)


app/electron/main.ts
```diff
@@ -644,6 +647,62 @@ function startElecron() {
...
+        return;
+      }
+
+      const child: ChildProcessWithoutNullStreams = spawn(
+        eventData.command,
+        eventData.args,
+        eventData.options
+      );
+
+      child.stdout.on('data', data => {
```
knrt10, line 691: Instead of `any` type for data, how about `(data: Buffer | string)` and for below 2 as well?

frontend/src/components/App/runCommand.ts
```diff
@@ -0,0 +1,83 @@
...
+  stdout: { on: (event: string, listener: (chunk: any) => void) => void };
+  stderr: { on: (event: string, listener: (chunk: any) => void) => void };
+  on: (event: string, listener: (code: number | null) => void) => void;
+} {
+  if (!window.desktopApi) {
+    throw new Error('runCommand only works in Headlamp app mode.');
+  }
+
+  // Generate a unique ID for the command, so that we can distinguish between
+  //   multiple commands running at the same time.
```
knrt10, line 41:
```suggestion
  // Generate a unique ID for the command, so that we can distinguish between
  // multiple commands running at the same time.
```

knrt10 (commented):
Nice work @illume. Some question/suggestions. Yet to test it.

illume:
Thanks for the review :) I'll make those changes.

app/electron/main.ts
illume: done

app/electron/main.ts
```diff
@@ -644,6 +647,62 @@ function startElecron() {
...
+     *
+     * @param event - The event object.
+     * @param eventData - The data sent from the renderer process.
+     */
+    function handleRunCommand(event: IpcMainEvent, eventData: CommandData): void {
+      // Only allow "minikube", "az", and "ls" commands
+      const validCommands = ['minikube', 'az', 'ls'];
+      if (!validCommands.includes(eventData.command)) {
+        console.log(`Ignoring invalid command: ${eventData.command}`);
+        return;
```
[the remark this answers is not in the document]
illume: Thanks. I added a type "az" | "minikube" to runCommand so they will get notified there of an issue.

I also improved the log here to be an error log, and also with added detail.

illume:
@knrt10 thanks, I addressed your notes.

knrt10 (approved):
Awesome work @illume . Tested it.
Screenshot 2024-01-30 at 12 22 07 PM


**Note**: Not in the scope of this PR, but maybe we can edit the plugin later to show app menu for the current cluster that it is in. i.e. if there are multiple clusters when open it should show its status. 

For example, here **knrt10** cluster is opened it should show it's status

Screenshot 2024-01-30 at 12 28 42 PM

Again, it's not in the scope of this PR, just something we could improve later.

joaquimrocha:
fixes #1670

PR #990, opened 2023-03-01
frontend: Add component for PluginSettings

vyncent-t:
# Plugin management Settings Feature

_Creates a plugin management feature allowing users to control installed plugins from the settings._
_Feature is following the design in issue #385_ 

image

This feature introduces a solution for Kubernetes Container Plugins Management. This feature is designed to help users efficiently manage their Kubernetes container plugins by providing a clean and organized interface for enabling and disabling plugins as needed.

_**Currently this feature is only supported on desktop app**_

### Includes
- **Plugin management feature component** 

The core component of the feature, the plugin management component, is designed to help users easily enable and disable Kubernetes container plugins. This component provides the main functionality for managing plugins, ensuring a smooth and efficient user experience.

- **Dynamic Save Button**

To enhance user experience and ensure changes are saved efficiently, our feature includes a dynamic save button that automatically detects when changes have been made to the enabled or disabled status of plugins. This button will appear on the page whenever changes can be saved, allowing users to quickly and easily save their modifications. The dynamic save button ensures that users can confidently manage their plugins, knowing that their preferred configurations are saved and applied within the Kubernetes cluster.

- **Storybook component**

To enhance understanding this commit includes a Storybook component that showcases various plugin management scenarios and use cases. This component serves as a visual guide and resource for users, allowing them to explore and learn about the plugin management process.

- **Route for / to component**

The feature incorporates a dedicated route to the plugin management component, enabling quick and easy access for users. By navigating to the root path ("/"), users will be directed to the plugin management component, where they can manage their Kubernetes container plugins effortlessly.

- **Plugin management page**

The plugin management feature is complete with a dedicated plugin management page that consolidates all plugin-related information and controls. This page provides users with a centralized location to manage their plugins, making it simple to enable, disable, and access plugin information.

- **Navigation option within settings page**

To further simplify the user experience, this feature includes a navigation option within the settings page. This addition makes it easy for users to locate the plugin management feature from the settings page and ensures seamless navigation within the application.

### Key Features:

1. **Organized list of Kubernetes container plugins:** The feature offers an orderly list of available Kubernetes container plugins, making navigation and selection a breeze.

2. **Essential plugin information:** Each plugin comes with crucial details such as:

-    Name: The plugin's name for straightforward identification.
-    Description: A brief explanation of the plugin's features and advantages.
-    Homepage: A link directing users to the plugin's homepage for more information.

[...]

PR #1678, opened 2024-01-29
Add Plugin Settings Detail View and Plugin Config Store

yolossn:
This PR adds Plugin Settings Details View and Plugin Config Store that can be used by plugin developers to store and retrieve plugin configurations.

To Do:
- [x] Documentation (Pausing this to get first round of reviews to avoid redoing it later)
- [x] Examples
- [x] Debug an import issue related to ConfigStore that occures when using the ConfigStore in Plugins.

Demo:
https://github.com/headlamp-k8s/headlamp/assets/22731013/8a43a75c-014c-4884-b641-ca9d5396b3be

How to test:

1. Run headlamp frontend and backend.
2. Run the `change-logo` and `pod-counter` plugins.
3. Either navigate to [http://localhost:3000/settings/plugins](http://localhost:3000/settings/plugins) or Click on Settings>Plugin in the side bar to view the Plugin Settings page.


backend/pkg/plugins/plugins.go
```diff
@@ -212,3 +212,8 @@
 		}
 	}
 }
+
+// Delete deletes the plugin from the plugin directory.
+func Delete(name string) error {
+	return os.RemoveAll(name)
```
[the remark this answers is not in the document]
yolossn: The caller of this function appends the plugin path to the plugin name to ensure that only plugin related path is sent to this function as argument.

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+    deletePlugin(name)
+      .then(() => {
+        // update the plugin list
+        const dispatch = useDispatch();
+        dispatch(reloadPage());
+      })
+      .finally(() => {
+        // redirect /plugins page
+        // history.push isn't updating the plugins list so we're using window.location.pathname
+        window.location.pathname = '/settings/plugins';
```
joaquimrocha, line 41: Please check if this works both in the app and the web modes. As the router is different.

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  return (
+    <>
+      <SectionBox aria-live="polite" title={plugin.name} backLink={'/settings/plugins'}>
+        {plugin.description}
+        <ScrollableBox style={{ height: '70vh' }} py={0}>
+          <ConfirmDialog
+            open={openDeleteDialog}
+            title={'Delete Plugin'}
+            description={'Are you sure you want to delete this plugin?'}
```
joaquimrocha, line 140: Translations

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+                <Button
+                  variant="contained"
+                  disabled={!enableSaveButton}
+                  style={{ backgroundColor: 'silver', color: 'black' }}
+                  onClick={handleSave}
+                >
+                  Save
+                </Button>
+                <Button style={{ color: 'silver' }} onClick={handleCancel}>
+                  Cancel
```
joaquimrocha, line 167: translations

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+                </Button>
+              </>
+            )}
+          </Stack>
+
+          <Button variant="text" color="error" onClick={handleDelete}>
+            Delete Plugin
```
joaquimrocha, line 174: translations

frontend/src/lib/router.tsx
```diff
@@ -88,7 +89,6 @@ import Deployment from './k8s/deployment';
 import Job from './k8s/job';
 import ReplicaSet from './k8s/replicaSet';
 import { getCluster, getClusterPrefixedPath } from './util';
-
```
joaquimrocha, line 91: Don't remove this.

frontend/src/plugin/registry.tsx
```diff
@@ -559,4 +560,11 @@ export function registerGetTokenFunction(override: (cluster: string) => string |
   store.dispatch(setFunctionsToOverride({ getToken: override }));
 }
 
+export function registerPluginSettings(
+  name: string,
+  component: PluginSettingsComponentType,
+  autoSave: boolean = false
```
joaquimrocha, line 566: I'd make these args as options, so we can easily extend it in the future.
Also, the autoSave is not very explicit as a name. I would actually guess the opposite of what it does.
So calling it `withSaveButton: boolean = true` is more explicit IMO (and we need to explain anyway that for the save button to take effect, the component needs to use the onDataChanged).

joaquimrocha (changes requested):

backend/cmd/headlamp.go
```diff
@@ -245,6 +245,21 @@ func defaultKubeConfigPersistenceFile() (string, error) {
 // It serves plugin static files at “/plugins/” and “/static-plugins/”.
 // It disables caching and reloads plugin list base paths if not in-cluster.
 func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
+	r.HandleFunc("/plugins/{name}", func(w http.ResponseWriter, r *http.Request) {
+		if err := checkHeadlampBackendToken(w, r); err != nil {
+			return
+		}
+		pluginName := mux.Vars(r)["name"]
+		pluginPath := path.Join(config.pluginDir, pluginName)
```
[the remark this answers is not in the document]
yolossn: since the "name" is a route variable it is not possible for the user to pass relative paths (like './../') as value, still I have added extra checks

frontend/src/plugin/pluginsSlice.ts
```diff
@@ -35,6 +60,19 @@ export type PluginInfo = {
   devDependencies?: {
     [key: string]: string;
   };
+
+  /**
+   * Component associated with the plugin's settings.
+   */
+  settingsComponent?: PluginSettingsComponentType;
```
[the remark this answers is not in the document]
yolossn: Thanks

backend/cmd/headlamp.go
```diff
@@ -245,6 +245,21 @@ func defaultKubeConfigPersistenceFile() (string, error) {
 // It serves plugin static files at “/plugins/” and “/static-plugins/”.
 // It disables caching and reloads plugin list base paths if not in-cluster.
 func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
+	r.HandleFunc("/plugins/{name}", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
yolossn: we can use the `inCluster` config for this. Have made sure this is available only in desktop mode.

backend/pkg/plugins/plugins.go
```diff
@@ -212,3 +214,27 @@ func HandlePluginReload(cache cache.Cache[interface{}], w http.ResponseWriter) {
 		}
 	}
 }
+
+// Delete deletes the plugin from the plugin directory.
+func Delete(pluginDir string, filename string) error {
```
knrt10, line 219:
```suggestion
func Delete(pluginDir, filename string) error {
```

backend/pkg/plugins/plugins_test.go
```diff
@@ -319,3 +319,42 @@ func TestPopulatePluginsCache(t *testing.T) {
...
+
+	// Test cases
+	tests := []struct {
+		pluginDir  string
+		pluginName string
+		expectErr  bool
+	}{
+		{pluginDir: tempDir, pluginName: tempFile.Name(), expectErr: false},          // Existing file
+		{pluginDir: tempDir, pluginName: "non-existent-directory", expectErr: false}, // Non-existent file
+		{pluginDir: tempDir, pluginName: "../", expectErr: true},                     // Directory traversal
```
knrt10, line 343: Maybe add another test where we can check for `isSubdirectory`?

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+    overflowY: 'scroll',
+    msOverflowStyle: 'none',
+    scrollbarWidth: 'none',
+    '&::-webkit-scrollbar': {
+      display: 'none',
+    },
+  },
+}))(Box);
+
+export interface PluginSettingsDetailPureProps {
```
knrt10, line 74: Can you please add comment?

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
```diff
@@ -0,0 +1,180 @@
...
+  config?: { [key: string]: any };
+  plugin: PluginInfo;
+  onSave?: (data: { [key: string]: any }) => void;
+  onDelete: () => void;
+}
+
```
knrt10, line 80: Can you please add comment?

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1830,3 +1830,17 @@ function getClusterDefaultNamespace(cluster: string, checkSettings?: boolean): s
 
   return defaultNamespace;
 }
+
+/**
+ * Deletes the plugin with the given name.
+ * @param name The name of the plugin to delete.
```
knrt10, line 1836:
```suggestion
 * @param name: The name of the plugin to delete.
```

frontend/src/plugin/configStore.ts
```diff
@@ -0,0 +1,69 @@
+import { useSelector } from 'react-redux';
+import store from '../redux/stores/store';
+import { setPluginConfig, updatePluginConfig } from './pluginConfigSlice';
+
+/**
+ * A class to manage the configuration state for plugins in a Redux store.
+ *
+ * @template T The type of the configuration object.
```
knrt10, line 8:
```suggestion
 * @template T - The type of the configuration object.
```

frontend/src/plugin/pluginConfigSlice.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The state structure for storing plugin configurations.
+ * Each plugin configuration is stored under a unique key.
+ */
+export interface PluginConfigState {
+  [configKey: string]: { [key: string]: any };
+}
+
+// Key used for local storage to persist plugin configurations.
+const PLUGIN_CONFIG_KEY = 'pluginConfigs';
```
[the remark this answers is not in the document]
knrt10: I think this might require too much of a change, but just a question do you think it would be possible to use indexDB? Since we already have it setup now.

frontend/src/plugin/registry.tsx
```diff
@@ -559,4 +560,11 @@ export function registerGetTokenFunction(override: (cluster: string) => string |
   store.dispatch(setFunctionsToOverride({ getToken: override }));
 }
 
+export function registerPluginSettings(
```
knrt10, line 563: Can you please add comment?

knrt10 (commented):
Nice work @yolossn, added some suggestions.

backend/cmd/headlamp.go
```diff
@@ -245,6 +245,20 @@ func defaultKubeConfigPersistenceFile() (string, error) {
...
```
illume, line 248: Please add a comment for this function.

illume (changes requested):
Added a few notes, will look more later. DCO and conflicts.

backend/pkg/plugins/plugins_test.go
yolossn: We do blackbox testing and since isSubdirectory is not exported it cannot be tested here. Also the test cases handle the sub directory case.

frontend/src/plugin/pluginConfigSlice.ts
```diff
@@ -0,0 +1,64 @@
...
+     */
+    updatePluginConfig(
+      state,
+      action: PayloadAction<{ configKey: string; payload: { [key: string]: any } }>
+    ) {
+      state[action.payload.configKey] = {
+        ...state[action.payload.configKey],
+        ...action.payload.payload,
+      };
+      localStorage.setItem(PLUGIN_CONFIG_KEY, JSON.stringify(state));
```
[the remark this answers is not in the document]
yolossn: AFAIK the localStorage functions are synchronous so this shouldn't be a problem. If the idea is to debounce this for performance reasons we can do but that will require a manual merging step to avoid storing stale data.

frontend/src/plugin/registry.tsx
yolossn: have changed it to `displaySaveButton`.

plugins/examples/change-logo/src/index.tsx
```diff
@@ -3,11 +3,20 @@
 // import { registerAppLogo } from '@kinvolk/headlamp-plugin/lib';
 // registerAppLogo(() => <p>My Logo</p>);
 
-import { AppLogoProps, registerAppLogo } from '@kinvolk/headlamp-plugin/lib';
+import {
+  AppLogoProps,
+  registerAppLogo,
+  registerPluginSettings,
+} from '@kinvolk/headlamp-plugin/lib';
```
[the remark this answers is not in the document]
yolossn: I have moved all settings related logic to a separate file to make it less complex.

frontend/src/plugin/pluginsSlice.ts
```diff
@@ -68,9 +110,33 @@ export const pluginsSlice = createSlice({
     reloadPage() {
       window.location.reload();
     },
+    /**
+     * Set the plugin settings component.
+     */
+    setPluginSettingsComponent(
```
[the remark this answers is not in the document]
yolossn: added tests for this

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
yolossn: @illume I would like to know your thoughts on this. Is it better to use indexDB or use prefix+plugin config key in localStorage.

frontend/src/components/App/PluginSettings/pluginSettingsDetail.tsx
yolossn: Tested this, works in app too.

frontend/src/i18n/locales/en/translation.json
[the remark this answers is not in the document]
yolossn: It is already added.

plugins/examples/change-logo/src/index.tsx
```diff
@@ -18,7 +22,12 @@ import LogoLight from './icon-small-light.svg';
 function SimpleLogo(props: AppLogoProps) {
   const { logoType, className } = props;
 
-  return (
+  const useConf = store.useConfig();
+  const config = useConf();
+
+  return config?.url ? (
+    <img src={config?.url} alt="logo" className={className} />
```
illume, line 29: We shouldn't use html tags, only Mui components.

plugins/examples/change-logo/src/settings.tsx
```diff
@@ -0,0 +1,113 @@
+import { ConfigStore } from '@kinvolk/headlamp-plugin/lib';
+import { NameValueTable } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import Box from '@mui/material/Box';
+import TextField from '@mui/material/TextField';
+import { useEffect, useState } from 'react';
+
```
illume, line 6: Can you please add a comment for the reader which explains what this demonstrates?

Please also add a note to the README.

frontend/src/plugin/registry.tsx
```diff
@@ -623,4 +624,26 @@ export function registerHeadlampEventCallback(callback: HeadlampEventCallback) {
...
+ * @param name - The name of the plugin.
+ * @param component - The component to use for the settings.
+ * @param displaySaveButton - Whether to display the save button.
+ *
+ * @example
+ *
+ * ```tsx
+ * import { registerPluginSettings } from '@kinvolk/headlamp-plugin/lib';
+ *
+ * registerPluginSettings('my-plugin', MyPluginSettingsComponent, true);
```
illume, line 638: Can you please improve this example? Looking at this, I wouldn't know how to use it.

Some ideas...
- Adding a minimal component inline (as an arrow function) if it's small, or a minimal function component.
- Seeing true here, I don't see what it means. Perhaps adding a `const displaySaveButton = true;` and then using that in the function call.

frontend/src/plugin/registry.tsx
```diff
@@ -623,4 +624,26 @@ export function registerHeadlampEventCallback(callback: HeadlampEventCallback) {
...
+ * ```
+ */
+export function registerPluginSettings(
```
illume, line 641: Please add a section to docs/development/plugins/functionality.md following the format of the other pieces in that document.

See the existing ones there for how it should be written. (In short... It links to the API reference, and also the example plugin with the small description of the functionality.)

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
illume: > @illume I would like to know your thoughts on this. Is it better to use indexDB or use prefix+plugin config key in localStorage.

Is there already an indication on the size of items? I expect many users will just use small data. If localStorage is not leaked to the user API, it may be changing it to use indexdb could be done in a future iteration. If so, please open an issue about using indexdb instead.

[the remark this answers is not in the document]
illume: Please address the comment from Joaquim.

illume:
At least one comment was hidden from me on this PR. The only way I could see it was to get a direct link from Santhosh
Screenshot 2024-02-27 135746


Maybe it's happening to others... there are a number of open conversations without response on here.

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
illume: @yolossn What happens in the case of 20 plugins and some data is updated? Is it only 1 write.

frontend/src/plugin/registry.tsx
```diff
@@ -623,4 +624,26 @@ export function registerHeadlampEventCallback(callback: HeadlampEventCallback) {
...
+ * @param component - The component to use for the settings.
+ * @param displaySaveButton - Whether to display the save button.
+ *
+ * @example
+ *
+ * ```tsx
+ * import { registerPluginSettings } from '@kinvolk/headlamp-plugin/lib';
+ *
+ * registerPluginSettings('my-plugin', MyPluginSettingsComponent, true);
+ * ```
```
illume, line 639: Can you please also mention the examples where this function is used?

Similar to this part of the comment for another function in registry.tsx.
```
 * More complete logo example in plugins/examples/change-logo:
 * @see {@link http://github.com/kinvolk/headlamp/plugins/examples/change-logo/ Change Logo Example}
```

illume (changes requested):
Nice work.

I left some notes. The only blocker for me is adding section into functionality.md so we have an entry point for users to look at.

I think some of the things can be deferred for future work. It's more important to get this into a release.

Please finish addressing each comment and rebase to fix the conflicts... then I will do a final test.

plugins/headlamp-plugin/src/stateless/index.ts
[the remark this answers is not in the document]
illume: @yolossn please address this comment.

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
yolossn: @joaquimrocha The size limit on localStorage is 5MB per domain and not per key level so storing one key per plugin wouldn't help much. So I am going ahead and marking this thread as resolved.

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
yolossn: Yes it is one write per call to setPluginConfig/updatePluginConfig. Basically on page load the config from localStorage is loaded in redux, whenever the value is changed in redux through setPluginConfig/updatePluginConfig the current redux state is set in localStorage.

We allow the pluginSettings component to choose between Auto Save and Save on button click.  When using Auto Save there can be a case where the updatePluginConfig function can be called frequently leading to performance issue.

plugins/headlamp-plugin/src/stateless/index.ts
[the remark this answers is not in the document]
yolossn: Removed it as it is not supposed to be part of the PR changes.

yolossn:
@joaquimrocha Should the plugin settings be made available in incluster mode also? In that case ill have to hide the delete plugin button.

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
yolossn: added debounce

illume:
> @joaquimrocha Should the plugin settings be made available in incluster mode also? In that case ill have to hide the delete plugin button.

I assume that yes they should be made available in-cluster.

backend/cmd/headlamp.go
```diff
@@ -245,6 +245,24 @@ func defaultKubeConfigPersistenceFile() (string, error) {
 // It serves plugin static files at “/plugins/” and “/static-plugins/”.
 // It disables caching and reloads plugin list base paths if not in-cluster.
 func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
+	// Delete plugin route
+	// This is only available when running locally.
+	if !config.useInCluster {
```
illume, line 250: This should be protected somehow.

yolossn: Sorry I didn't get you, there is already a check for Headlamp backend token. Is there any other token that I am missing here.

frontend/src/plugin/pluginConfigSlice.ts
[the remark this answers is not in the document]
joaquimrocha: I really thought there was a limit per value. If there isn't, then your @yolossn comment makes sense. I still feel like having the values separated does make sense, but if we are going to add index db in any case, then I am fine with not separating the keys here.

frontend/src/plugin/pluginConfigSlice.ts
```diff
@@ -0,0 +1,71 @@
...
+// Key used for local storage to persist plugin configurations.
+const PLUGIN_CONFIG_KEY = 'pluginConfigs';
+
+// Initial state is loaded from local storage, or an empty object if nothing is stored.
+const initialState: PluginConfigState = JSON.parse(localStorage.getItem(PLUGIN_CONFIG_KEY) || '{}');
+
+const DEBOUNCE_DELAY = 500; // ms
+
+const debouncedSetItemInLocalStorage = _.debounce((key: string, value: string) => {
+  localStorage.setItem(key, value);
```
joaquimrocha, line 21: When we reach the limits of local storage, I think an exception is raised. In the past, this issue rendered Headlamp useless, since once we reached a max storage for notifications, the exception would prevent any use of Headlamp. Please catch it here.

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -6,7 +6,6 @@
  * Copyright © 2020 Eric Herbrandson
  * Copyright © 2020 Kinvolk GmbH
  */
-
```
joaquimrocha, line 9: Mistakenly removed?

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -1,12 +1,16 @@
-import { Switch } from '@mui/material';
+import { Switch, SwitchProps, Typography } from '@mui/material';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
+import Link from '@mui/material/Link';
+import { styled } from '@mui/material/styles';
```
joaquimrocha, line 5: We are trying to get rid of styles. Please use `@mui/system` or use the `sx` props.

frontend/src/components/App/PluginSettings/PluginSettingsDetails.tsx
```diff
@@ -0,0 +1,193 @@
...
+import NotFoundComponent from '../../404';
+import { SectionBox } from '../../common';
+import { ConfirmDialog } from '../../common/Dialog';
+import ErrorBoundary from '../../common/ErrorBoundary';
+
+export default function PluginSettingsDetails() {
+  const pluginSettings = useTypedSelector(state => state.plugins.pluginSettings);
+  const { name } = useParams<{ name: string }>();
+  const decodedName = decodeURIComponent(name);
+  const plugin = pluginSettings.find(plugin => plugin.name === decodedName);
```
joaquimrocha, line 24: This should be a useMemo with a dependency on the pluginSettings + name (you can decode within the memo).

frontend/src/components/App/PluginSettings/PluginSettingsDetails.tsx
```diff
@@ -0,0 +1,193 @@
...
+  if (!plugin) {
+    return <NotFoundComponent />;
```
joaquimrocha, line 26: This is an early return when you have what looks like a hook (store.useConfig) below. This can mess up React.
If it's indeed a hook and you still want to have an early return, you have to create a new component that does those config related inits + returns PluginSettingsDetailsPure itself.

frontend/src/components/App/PluginSettings/PluginSettingsDetails.tsx
```diff
@@ -0,0 +1,193 @@
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import Stack from '@mui/material/Stack';
+import withStyles from '@mui/styles/withStyles';
```
joaquimrocha, line 4: We shouldn't use @mui/styles anymore.

frontend/src/lib/router.tsx
```diff
@@ -723,6 +724,18 @@ const defaultRoutes: {
     disabled: !helpers.isElectron(),
     component: () => <PluginSettings />,
   },
+  pluginDetails: {
+    path: '/settings/plugins/:name',
```
joaquimrocha, line 727: I wonder if we will have clashes on the name among diff vendors (but we can solve it later).

docs/development/plugins/functionality.md
```diff
@@ -166,3 +166,10 @@ React to Headlamp events with [registerHeadlampEventCallback](../api/modules/plu
 
 - Example plugin shows [How to show snackbars for Headlamp events](https://github.com/kinvolk/headlamp/tree/main/plugins/examples/headlamp-events).
 - API reference for [registerHeadlampEventCallback](../api/modules/plugin_registry.md#registerheadlampeventcallback)
+
+
+### Plugin Settings
+
+The plugins can have user configurable settings that can be used to change the behavior of the plugin. The plugin settings can be created using [registerPluginSettings](../api/modules/plugin_registry.md#registerpluginsettings).
+
+- Example plugin shows [How to create plugin settings and use them](https://github.com/kinvolk/headlamp/tree/main/plugins/examples/change-logo)
```
joaquimrocha, line 175: For the sake of consistency, we should have a screenshot.

frontend/src/lib/router.tsx
```diff
@@ -721,7 +721,6 @@ const defaultRoutes: {
     },
     useClusterURL: false,
     noAuthRequired: true,
-    disabled: !helpers.isElectron(),
```
joaquimrocha, line 723: Since this was added in this PR, can you delete it where it was added? Same for the other change. Unless there are complications I haven't noticed.

joaquimrocha (changes requested):

frontend/src/lib/k8s/apiProxy.ts
yolossn: Not sure why the linter isn't handling these.

frontend/src/components/App/PluginSettings/PluginSettingsDetails.tsx
yolossn: wasn't aware of this, may I know why?

frontend/src/lib/router.tsx
yolossn: not only in this instance the clash is possible most of the places where the plugin name is used, would prefer to solve it later as you mentioned.

frontend/src/lib/router.tsx
yolossn: AFAIU this wasn't added in this PR.

yolossn: This was added in #990

joaquimrocha: Ah, ok. Sorry.

joaquimrocha (approved):
I think this LGTM now.
BTW, I thought the "now enabled" routes were initially added here and that was a mistake as you mentioned @yolossn . But making the routes available does mean users will see the "disable" button, and I think disabling plugins shouldn't work unless in the app mode.
This can be added in a second PR, but maybe it's important to have it for the release?
WDYT @illume ?

yolossn:
> @yolossn . But making the routes available does mean users will see the "disable" button, and I think disabling plugins shouldn't work unless in the app mode.

A check is added to make sure the Delete plugin is only visible in the app mode. So the button wont be visible to the user. And similarly check is added in the backend to make sure the delete API route isn't available in the in-cluster mode.

https://github.com/headlamp-k8s/headlamp/pull/1678/files#diff-4803c78d91a225f18ee5bd4c6eda6e00b92e0f7055f9b91f909af685d725a71dR191

illume (approved):
🎉🎈

PR #1681, opened 2024-01-30
frontend: fix calling /parseKubeConfig

knrt10:
This endpoint is used only for parsing stateless clusters kubeconfig. So, it should only be called when dynamic clusters are enabled. When the flag was disabled, the endpoint was unavailable. This fix calling the endpoint unnecessarily when dynamic cluster is disabled.


illume (approved):
👍 thanks

PR #1685, opened 2024-01-31
frontend: Fix default sorting column for the ResourceTable

joaquimrocha:
The ResourceTable by default sorts data by the age column. However,
since the columns can now be hidden, the way to compute the index of
the age column needed to be updated to take hidden columns into
account.


vyncent-t (commented):
LGTM!

vyncent-t (approved):
LGTM!

Issue #1684, opened 2024-01-30
Frontend is breaking in v0.22.0.


yolossn:
Steps to reproduce:
- Install Headlamp incluster using Helm chart
- Expose Headlamp using kubectl port-forward
- The frontend page is a broken white page and the console shows the following error message
<img width="960" alt="image" src="https://github.com/headlamp-k8s/headlamp/assets/22731013/48920931-9137-481e-8b46-626921fd3e9f">

Looks like it must be related to this [commit](https://github.com/headlamp-k8s/headlamp/commit/469d8b328361b463ecab96aadc0be7d0d3b549e9)

joaquimrocha:
@yolossn can you check with the main branch. Probably not related, but checking if #1666 fixes it is probably worth it still.

illume:
@yolossn I couldn't reproduce this locally.

Would you mind showing me what (the redacted) `localStorage.getItem('notifications')` is?

I worked with @vyncent-t on a potential fix in https://github.com/headlamp-k8s/headlamp/pull/1686

yolossn:
```json
[
  {
    "id": "2e02d825-432c-4640-a6ea-2af96eb815ba",
    "seen": false,
    "url": "/c/minikube/?eventsFilter=2e02d825-432c-4640-a6ea-2af96eb815ba",
    "date": 1706461592000,
    "deleted": false,
    "cluster": "minikube",
    "message": "Back-off restarting failed container kube-controller-manager in pod kube-controller-manager-minikube_kube-system(b3702ceb912504d37098b922ccdcfa41)",
  },
  {
    "id": "746dcfc7-2db7-4a36-85f9-5b4d6e52e9de",
    "seen": false,
    "url": "/c/minikube/?eventsFilter=746dcfc7-2db7-4a36-85f9-5b4d6e52e9de",
    "date": 1706461582000,
    "deleted": false,
    "cluster": "minikube",
    "message": "error determining status: rpc error: code = Unknown desc = Error response from daemon: No such container: 8a32326ea70e5ef6f083829ef6753bdf607a73758fc82de63d8e0a4bd20033f9",
  },
  {
    "id": "adacd33a-c36c-4a5f-b866-a240281c1733",
    "seen": false,
    "url": "/c/minikube/?eventsFilter=adacd33a-c36c-4a5f-b866-a240281c1733",
    "date": 1706461560000,
    "deleted": false,
    "cluster": "minikube",
[...]

PR #1686, opened 2024-01-31
frontend: Add fix and test for notifcations

vyncent-t:
Fixes and adds test for issue 1684 where notification logic was not being checking for null array

Fixes issue #1684


illume (approved):
👍

PR #1688, opened 2024-02-01
frontend: fix stream function for websocket conn

knrt10:
When called stream function was calling connect function synchronously and returning cancel and getSocket which were always null. This converts the function to a async function and returns a promise which resolves to the desired values for both getSocket and cancel function.


knrt10:
Make sense. I discussed this with @ashu8912 and we agreed that having a socket ready should be in the scope of user. So closing this PR.

PR #1692, opened 2024-02-02
github: Dockerfile: Fix .plugins copying and usage with test

### LINKED ISSUE #1687: .plugins folder not copied when creating an image
- [x] The .plugins folder used to work as a place where the Dockerfile would look for plugins to copy in. But it doesn't work now. See last few commits on Dockerfile.
- [x] The docs should be improved for building container images with plugins. We have a blogpost https://headlamp.dev/blog/2022/10/get-up-to-speed-deploying-headlamp-with-plugins/ We have some docs: https://headlamp.dev/docs/latest/development/plugins/building/ But probably the blog post should be merged into the docs. Also note .plugin folder functionality mentioned in the docs is currently broken.

cc @ashu8912

illume:
At some point the .plugin package copying stopped working. 
It's documented that a .plugins folder will be copied and used by the container.

There's a test included proving that the files are copied properly now.
Fixes #1687

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -28,6 +28,18 @@ jobs:
       # now you can run kubectl to see the pods in the cluster
     - name: Try the cluster!
       run: kubectl get pods -A
+    - name: Make a .plugins folder for testing later
+      run: |
+        echo "Extract pod-counter plugin into .plugins folder, which will be copied into image later by 'make image'."
+        cd plugins/examples/pod-counter
+        npm install
+        npm run build
+        cd ../../../
+        cd plugins/headlamp-plugin
+        npm install
+        node bin/headlamp-plugin.js extract ../examples/pod-counter ../../.plugins/
+        cd ../../
+        ls -laR .plugins
     - name: Build image
       run: |
         export SHELL=/bin/bash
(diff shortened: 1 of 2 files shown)


.github/workflows/build-container.yml
```diff
@@ -28,6 +28,17 @@ jobs:
       # now you can run kubectl to see the pods in the cluster
     - name: Try the cluster!
       run: kubectl get pods -A
+    - name: Test .plugins folder is copied in the right place
```
knrt10, line 31: I think this is not testing whether the plugins folder is getting copied, it is just copying the folder, so maybe edit the workflow name? WDYT?

illume: You're right, thanks.

Dockerfile
```diff
@@ -53,13 +53,14 @@ RUN cat ./frontend/.env
 # Backwards compatibility, move plugin folder to only copy matching plugins.
 RUN mv plugins plugins-old || true
 
+# Copy a .plugins folder if it is there to ./plugins, otherwise create an empty one.
+COPY ./.plugi*s ./plugins
```
knrt10, line 57: Is there a reason for using `.plugi*s`? I am sorry I don't understand.

.github/workflows/build-container.yml
illume: I moved that .plugins folder making code to where the test is done. Now the name makes sense.

knrt10 (approved):
Apart from one question, rest LGTM. Nice work :-)

Dockerfile
illume: Thanks.

I added further explanation of this... let me know if it makes sense now?

knrt10: Cool. Thanks.

PR #1694, opened 2024-02-04
frontend: use sx instead of styles

farodin91:


farodin91:
@joaquimrocha This would be the start of the migration with sx instead of styles.

frontend/src/components/App/Home/__snapshots__/SquareButton.stories.storyshot
```diff
@@ -8,14 +8,14 @@ exports[`Storyshots common/SquareButton Basic 1`] = `
     type="button"
   >
     <div
-      class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation1 MuiCard-root makeStyles-root makeStyles-root css-bhp9pd-MuiPaper-root-MuiCard-root"
+      class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation1 MuiCard-root css-1o0qjyj-MuiPaper-root-MuiCard-root"
```
ashu8912, line 11: Will these names be generated on each CI run?? If yes then that can be an issue

farodin91: They seems to be static.

frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "@mui/lab": "^5.0.0-alpha.152",
     "@mui/material": "^5.14.17",
     "@mui/styles": "^5.14.17",
+    "@mui/system": "^5.14.17",
```
illume, line 18: Inside plugins/headlamp-plugin/ folder... Please run "npm run update-dependencies" to sync dependencies from frontend/ to plugins/headlamp-plugin/. This will fix the current error in CI.

Because headlamp-plugin needs to keep in sync with the frontend package versions.

farodin91: rebased and updated

joaquimrocha (approved):

Issue #1638, opened 2024-01-03
[Bug]: App crashes on opening Storage volumes.


knrt10:
The console logs say

```
Uncaught TypeError: Expected "namespace" to be defined
    at index.js:179:1
```

The backend calls to fetch persistent volumes is retuning the correct data with 200 status code. 

### Environment

- Run Headlamp on Mac OS or web application. 
- AKS cluster with PV created and PVC bounded to a pod.
- Open storage volumes tab in the Headlamp app to reproduce the error

I am also attaching the screen recording for the error. 

https://github.com/headlamp-k8s/headlamp/assets/24803604/1af16ab7-cdfb-46e3-873a-c7916b789c96

knrt10:
Closed via #1697

PR #1697, opened 2024-02-06
frontend: If no claim available we don't create link for it

ashu8912:


ashu8912:
fixes #1638

knrt10 (approved):
+1

frontend/src/components/storage/VolumeList.tsx
```diff
@@ -67,6 +67,11 @@ export default function VolumeList() {
           getter: volume => {
             const claim = volume.spec.claimRef?.name;
             const claimNamespace = volume.spec.claimRef?.namespace;
+
+            if (!claim) {
+              return null;
+            }
```
joaquimrocha, line 73: We should return before checking the claimNamespace, as it's unused if claim is empty/undefined anyway...

ashu8912: updated

joaquimrocha (approved):

PR #1707, opened 2024-02-09
Fix item crud when changing clusters

### LINKED ISSUE #1660: Bug: Deletion in the wrong cluster through delay

joaquimrocha:
This issue fixes https://github.com/headlamp-k8s/headlamp/issues/1660 by making the cluster name as part of each resource instance and using it for the operations (+ something similar for the resource creation).

fixes #1660

How to test:
- [ ] Create/edit/delete a resource (example, a ConfigMap); during the grace period, move to a different cluster and wait for the resource to be created/edited/deleted: verify that the operation happened on the original cluster and not in the one you moved to

### DIFF
--- a/frontend/src/components/common/Resource/CreateButton.tsx
+++ b/frontend/src/components/common/Resource/CreateButton.tsx
@@ -4,6 +4,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useLocation } from 'react-router-dom';
+import { getCluster } from '../../../lib/cluster';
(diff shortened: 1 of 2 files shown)


joaquimrocha:
@farodin91 , can you test this one too?

farodin91:
@joaquimrocha I will try to check it on monday.

farodin91:
@joaquimrocha Works fine.

Issue #750, opened 2022-10-13
Ensure namespaces are used in list queries when defined


joaquimrocha:
## Current situation

We keep a namespace filter that persists across views, but we use it just for filtering the resources on client side.
i.e. we do requests for all namespaces and then we filter the results locally.

## Impact

Besides the extra processing needed when more results are returned (due to asking from all namespaces), we may also run into issues where users do not have access to such namespaces (this is related to #679 ).

## Ideal future situation

If a namespace (or more) are set for the entire application, we should use them in the queries as well. Maybe this solution should be done together with #679 .

cc/ @callmevladik, @lijianzhi01

This is part of the performance epic https://github.com/headlamp-k8s/headlamp/issues/1161

joaquimrocha:
This is already implemented. Closing now.

PR #1710, opened 2024-02-10
frontend: filter list views by default namespace

tazo90:
This issue fixes [#750](https://github.com/headlamp-k8s/headlamp/issues/750) by allowing to filter cluster resources on the backend by using a globally defined default namespace.

It currently works with one namespace at a time, as that's what the Kubernetes API allows. If we want to support multiple namespaces, extra requests would be needed for each one.

Thanks to that we can limit results from API and reduce response size.

How to test:
- Define default namespace in settings.
- Check e.g. workloads -> pods whether results are filtered to a given namespace.


joaquimrocha:
This was implemented in a different way as part of our ReactQuery usage, so closing now.

Issue #1695, opened 2024-02-05
Dockerfile.plugins needs a test


illume:
Seems the Dockerfile.plugins is kind of broken, and there's no test to verify that the Dockerfile.plugins is creating the files in the container image that are intended.

See full errors here: https://github.com/headlamp-k8s/headlamp/actions/runs/7784225360/job/21224343101?pr=1692#step:7:1318

There's a bunch of errors like this below, but CI does not fail because of this error. 
```
#15 2.951 ModuleNotFoundError: Module not found: Error: Can't resolve 'js-yaml' in '/headlamp-plugins/dynamic-clusters/src'
#15 2.951     at /headlamp-plugins/headlamp-plugin/node_modules/webpack/lib/Compilation.js:2022:28
#15 2.951     at /headlamp-plugins/headlamp-plugin/node_modules/webpack/lib/NormalModuleFactory.js:817:13
#15 2.951     at eval (eval at create (/headlamp-plugins/headlamp-plugin/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:10:1)
#15 2.951     at /headlamp-plugins/headlamp-plugin/node_modules/webpack/lib/NormalModuleFactory.js:275:22
#15 2.951     at eval (eval at create (/headlamp-plugins/headlamp-plugin/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:9:1)
#15 2.951     at /headlamp-plugins/headlamp-plugin/node_modules/webpack/lib/NormalModuleFactory.js:448:22
```

## note

probably node version needs updating

PR #1714, opened 2024-02-13
plugins: Fix copying to docker containers

knrt10:
We were copying plugins but we were not installing dependecies which resulted some example plugins to not get copied.

headlamp-events plugin was also using local dev dependency which resulted in failure of installing it. This fixes it.

Fixes: #1695

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -40,6 +40,10 @@ jobs:
         node bin/headlamp-plugin.js extract ../examples/pod-counter ../../.plugins/
         cd ../../
         ls -laR .plugins
+    - name: Remove unnecessary files
+      run: |
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf "$AGENT_TOOLSDIRECTORY"
     - name: Build image
       run: |
         export SHELL=/bin/bash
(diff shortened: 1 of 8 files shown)


knrt10:
> Can you please add a test to check if (some or at least one) plugin(s) are copied so it doesn't regress in the future? See recent PR for Dockerfile.

There is already an e2e tests, for it, but I will modify it to check exact plugins are there as we have in examples folder.

joaquimrocha (approved):
LGTM.

joaquimrocha:
@knrt10 , the patch LGTM. Can you check if you can answer these questions by Rene:
```
It would be interesting to know why it didn't fail before? The headlamp-plugin build command should fail CI there if it fails itself. Or maybe another command was failing and continuing without error? That command failing should fail CI.

It looks like the build command isn't checking webpack for errors. https://github.com/headlamp-k8s/headlamp/blob/main/plugins/headlamp-plugin/bin/headlamp-plugin.js#L478 See the docs here for how it uses callbacks not throwing an exception to handle errors: https://webpack.js.org/api/node#webpack
```

knrt10:
> It would be interesting to know why it didn't fail before? The headlamp-plugin build command should fail CI there if it fails itself. Or maybe another command was failing and continuing without error? That command failing should fail CI.

It didn't fail earlier because as you suggested we are not handling webpack callback errors, so due to which it marks the builds as successful. It should fail the CI, I am looking more into it.

> It looks like the build command isn't checking webpack for errors. https://github.com/headlamp-k8s/headlamp/blob/main/plugins/headlamp-plugin/bin/headlamp-plugin.js#L478 See the docs here for how it uses callbacks not throwing an exception to handle errors: https://webpack.js.org/api/node#webpack

Thank you, looking more into why webpack issues are there.

We should not merge it now because webpack errors are still there. I am trying to fix it.

joaquimrocha (approved):
LGTM! Thanks!

PR #1715, opened 2024-02-13
chart: Added ability to use environment variables

macintoshme:
Add ability to use environment variables in helm chart


knrt10 (approved):
Tested. Thank you for the PR @macintoshme. LGTM

PR #1717, opened 2024-02-15
frontend: fix performance of notifications

farodin91:


frontend/src/components/App/Notifications/index.tsx
```diff
@@ -225,6 +229,20 @@ export default function Notifications() {
...
+    setAreAllNotificationsInDeleteState(currentAreAllNotificationsInDeleteState);
+    setAreThereUnseenNotifications(
+      notifications.filter(notification => notification.seen !== true).length > 0
+    );
+    if (currentAreAllNotificationsInDeleteState) {
+      setFilteredNotifications([]);
+    } else {
+      setFilteredNotifications(notifications.slice(0, maxNotificationsInPopup));
+    }
+  }, [notifications]);
```
joaquimrocha, line 244: Since we are not yet on React 18, where state setting like these are batched together automatically, I think we can make it slightly more performant if we use a `useMemo` and return all the values we are processing here:
`const [areThereUnseenNotifications, areAllNotificationsInDeleteState, filteredNotifications] = useMemo()...`

joaquimrocha (changes requested):

frontend/src/components/App/Notifications/index.tsx
farodin91: changed to memo

joaquimrocha (approved):

joaquimrocha:
Thanks a lot @farodin91 !

Issue #1645, opened 2024-01-08
Format backend log to easily be parsed by log platform


lijianzhi01:
Headlamp backend server log isn't easy to recognized by log platform. For example. azure app service will take all headlamp backend log as error level. 

I have classified logger to refer:
```
package main

import (
	"fmt"
	"io"
	"log"
)

type Level int

const (
	LevelInfo Level = iota
	LevelWarning
	LevelError
)

type CustomLogger struct {
	logger *log.Logger
	level  Level
}

func (c *CustomLogger) init(out io.Writer, level Level) {
	c.logger = log.New(out, "", 0)
	c.level = level
}

func (c *CustomLogger) output(level Level, format string, v ...interface{}) {
	if level >= c.level {
		levelText := [...]string{"[INFO] ", "[WARNING] ", "[ERROR] "}
		message := fmt.Sprintf("%s%s", levelText[level], fmt.Sprintf(format, v...))
		c.logger.Printf("%s", message)
	}
}

func (c *CustomLogger) Info(format string, v ...interface{}) {
	c.output(LevelInfo, format, v...)
}

func (c *CustomLogger) Warning(format string, v ...interface{}) {
	c.output(LevelWarning, format, v...)
}

func (c *CustomLogger) Error(format string, v ...interface{}) {
	c.output(LevelError, format, v...)
}

var logger = CustomLogger{}

```

use logger:
```
func setupProxyForContext(context *Context, config *HeadlampConfig, source int) error {
	proxy, err := config.createProxyForContext(*context)
	if err != nil {
		logger.Info("Error setting up proxy for context %s: %s\n", context.Name, err)
		return err
	}

	_, isReplacement := config.contextProxies[context.Name]
	if isReplacement {
		logger.Info("Replaced cluster \"%s\" proxy by:\n", context.Name)
	} else {
		logger.Info("Created new cluster proxy:")
	}

	logger.Info("\tlocalhost:%d%s%s/{api...} -> %s\n", config.port, config.baseURL, "/clusters/"+context.Name,
		*context.cluster.getServer())

	config.contextProxies[context.Name] = contextProxy{
		context,
		proxy,
		source,
	}

	return nil
}
```

It will show up like:
![image](https://github.com/headlamp-k8s/headlamp/assets/13809749/7dea5afd-1ca1-4179-b3a5-b060164a895d)

lijianzhi01:
[...]

PR #1723, opened 2024-02-19
backend: format backend logs to use zlog

knrt10:
Earlier we were not using zlog everywhere due to which some logs were not correctly parsed by log platforms. They were all parsed as errors. This fixes that and adds correct format of all logs in headlamp. They are now shown as json which can be parsed by any log platform.

This also adds a wrapper function for zlog which increases code readability and cleanliness.

Fixes: #1645

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -11,7 +11,6 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"log"
 	"net/http"
 	"net/url"
 	"os"
@@ -30,6 +29,7 @@ import (
 	"github.com/headlamp-k8s/headlamp/backend/pkg/cache"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/helm"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/plugins"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/portforward"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -37,7 +37,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 
-	zlog "github.com/rs/zerolog/log"
 	"golang.org/x/oauth2"
 )
 
@@ -104,7 +103,9 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	} else if err != nil {
 		// if we got an error (that wasn't that the file doesn't exist) stating the
 		// file, return a 500 internal server error and stop
+		logger.Log(logger.LevelError, nil, err, "stating file")
 		http.Error(w, err.Error(), http.StatusInternalServerError)
+
 		return
 	}
 
@@ -129,7 +130,9 @@ func copyReplace(src string, dst string,
 ) {
 	data, err := os.ReadFile(src)
 	if err != nil {
-		log.Fatal(err)
+		// Error Reading the file
+		logger.Log(logger.LevelError, nil, err, "reading file")
+		os.Exit(1)
 	}
 
 	data1 := bytes.ReplaceAll(data, search, replace)
@@ -138,7 +141,9 @@ func copyReplace(src string, dst string,
 
 	err = os.WriteFile(dst, data2, fs.FileMode(fileMode))
 	if err != nil {
-		log.Fatal(err)
+		// Error writing the file
+		logger.Log(logger.LevelError, nil, err, "writing file")
+		os.Exit(1)
 	}
 }
 
@@ -252,7 +257,7 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 			pluginsList = []string{}
 		}
 		if err := json.NewEncoder(w).Encode(pluginsList); err != nil {
-			log.Println("Error encoding plugins base paths list", err)
+			logger.Log(logger.LevelError, nil, err, "encoding plugins base paths list")
 		}
 	}).Methods("GET")
 
@@ -279,11 +284,13 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	config.staticPluginDir = os.Getenv("HEADLAMP_STATIC_PLUGINS_DIR")
 
-	log.Printf("static plugin dir: %s\n", config.staticPluginDir)
-	log.Printf("plugins-dir: %s\n", config.pluginDir)
-	log.Printf("dynamic clusters support: %v\n", config.enableDynamicClusters)
-	log.Printf("Helm support: %v\n", config.enableHelm)
-	log.Printf("Proxy URLs: %+v\n", config.proxyURLs)
+	logger.Log(logger.LevelInfo, nil, nil, "Creating Headlamp handler")
+	logger.Log(logger.LevelInfo, nil, nil, "Kubeconfig path: "+kubeConfigPath)
+	logger.Log(logger.LevelInfo, nil, nil, "Static plugin dir: "+config.staticPluginDir)
+	logger.Log(logger.LevelInfo, nil, nil, "Plugins dir: "+config.pluginDir)
+	logger.Log(logger.LevelInfo, nil, nil, "Dynamic clusters support: "+fmt.Sprint(config.enableDynamicClusters))
+	logger.Log(logger.LevelInfo, nil, nil, "Helm support: "+fmt.Sprint(config.enableHelm))
+	logger.Log(logger.LevelInfo, nil, nil, "Proxy URLs: "+fmt.Sprint(config.proxyURLs))
 
 	plugins.PopulatePluginsCache(config.baseURL, config.staticPluginDir, config.pluginDir, config.cache)
 
@@ -302,19 +309,19 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			config.oidcClientID, config.oidcClientSecret,
 			strings.Join(config.oidcScopes, ","))
 		if err != nil {
-			log.Println("Failed to get in-cluster config", err)
+			logger.Log(logger.LevelError, nil, err, "Failed to get in-cluster context")
 		}
 
 		context.Source = kubeconfig.InCluster
 
 		err = context.SetupProxy()
 		if err != nil {
-			log.Println("Failed to setup proxy for in-cluster context", err)
+			logger.Log(logger.LevelError, nil, err, "Failed to setup proxy for in-cluster context")
 		}
 
 		err = config.kubeConfigStore.AddContext(context)
 		if err != nil {
-			log.Println("Failed to add in-cluster context", err)
+			logger.Log(logger.LevelError, nil, err, "Failed to add in-cluster context")
 		}
 	}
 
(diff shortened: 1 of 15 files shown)


backend/pkg/helm/handler.go
```diff
@@ -39,7 +39,10 @@ func NewActionConfig(clientConfig clientcmd.ClientConfig, namespace string) (*ac
 		namespace:    namespace,
 	}
 	logger := func(format string, a ...interface{}) {
-		log.Info().Str("namespace", namespace).Msg(format + "\n" + fmt.Sprintf("%v", a))
+		zlog.Info().
```
yolossn, line 42: I think instead of printing the action manually you can create a helper function which internally uses the golang runtime packages [Caller](https://pkg.go.dev/runtime#Caller) function to print the source of the log. That can be helpful in identifying the source of the log very easily.

yolossn (changes requested):
I think it will be a better idea to create a helper interface or function to print the logs. It will make the code more cleaner and will make it easier for us to migrate to another logging package incase we want to migrate in the future. Also gives us the advantage of using the runtime package to print the source of the log which will make it easier to identify from where the log was created. WDYT?

backend/pkg/helm/handler.go
knrt10: Let me look into it

knrt10:
Thanks for the suggestion @yolossn. I have updated the code as PTAL. Thanks

backend/cmd/headlamp.go
```diff
@@ -361,19 +368,15 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			proxyURL = r.Header.Get("Forward-to")
 		}
 		if proxyURL == "" {
-			zlog.Error().Err(err).
-				Str("action", "externalproxy").
-				Str("proxyURL", proxyURL).
-				Msg("proxy URL is empty")
+			str := map[string]string{"proxyURL": proxyURL}
+			utils.Log(utils.LevelError, str, errors.New("proxy URL is empty"), "proxy URL is empty")
```
yolossn, line 372:
```suggestion
			utils.Log(utils.LevelError, map[string]string{"proxyURL": proxyURL}, errors.New("proxy URL is empty"), "proxy URL is empty")
```

backend/cmd/headlamp.go
```diff
@@ -361,19 +368,15 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
 			return
 		}
 		url, err := url.Parse(proxyURL)
 		if err != nil {
-			zlog.Error().Err(err).
-				Str("action", "externalproxy").
-				Str("proxyURL", proxyURL).
-				Msg("The provided proxy URL is invalid")
+			str := map[string]string{"proxyURL": proxyURL}
+			utils.Log(utils.LevelError, str, err, "The provided proxy URL is invalid")
```
yolossn, line 379:
```suggestion
			utils.Log(utils.LevelError, map[string]string{"proxyURL": proxyURL}, err, "The provided proxy URL is invalid")
```

backend/cmd/headlamp.go
```diff
@@ -463,20 +471,26 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 		kContext, err := config.kubeConfigStore.GetContext(cluster)
 		if err != nil {
-			log.Printf("Error: failed to get context: %s", err)
+			str := map[string]string{"cluster": cluster}
+			utils.Log(utils.LevelError, str, err, "Error: failed to get context")
+
```
yolossn, line 476:
```suggestion
			utils.Log(utils.LevelError, map[string]string{"cluster": cluster}, err, "Error: failed to get context")

```

backend/cmd/headlamp.go
```diff
@@ -463,20 +471,26 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
 			http.NotFound(w, r)
 			return
 		}
 
 		oidcAuthConfig, err := kContext.OidcConfig()
 		if err != nil {
-			log.Printf("Error getting %s cluster oidc config %s", cluster, err)
+			str := map[string]string{"cluster": cluster}
```
yolossn, line 483: similar to other comments no need to create the map as a separate variable since it is used only as an argument to the `utils.Log` function.

backend/cmd/headlamp.go
```diff
@@ -675,17 +709,23 @@ func refreshAndCacheNewToken(oidcAuthConfig *kubeconfig.OidcConfig,
...
 	refreshToken, err := cache.Get(context.Background(), fmt.Sprintf("oidc-token-%s", token))
 	if err != nil || refreshToken == "" {
+		utils.Log(utils.LevelError, nil, err, "Error: failed to get refresh token")
+
 		return "", err
 	}
 
 	rToken, ok := refreshToken.(string)
 	if !ok {
+		utils.Log(utils.LevelError, nil, err, "Error: failed to cast refresh token")
```
yolossn, line 727:
```suggestion
		utils.Log(utils.LevelError, nil, err, "failed to cast refresh token")
```

since it is passed with Error as severity I think it is not necessary to add Error to the message.

backend/cmd/headlamp.go
```diff
@@ -675,17 +709,23 @@ func refreshAndCacheNewToken(oidcAuthConfig *kubeconfig.OidcConfig,
...
 	if err != nil {
+		utils.Log(utils.LevelError, nil, err, "Error: failed to get provider")
+
 		return "", err
 	}
 
 	// get refresh token from cache
 	refreshToken, err := cache.Get(context.Background(), fmt.Sprintf("oidc-token-%s", token))
 	if err != nil || refreshToken == "" {
+		utils.Log(utils.LevelError, nil, err, "Error: failed to get refresh token")
```
yolossn, line 720:
```suggestion
		utils.Log(utils.LevelError, nil, err, "failed to get refresh token")
```

backend/cmd/headlamp.go
```diff
@@ -703,19 +743,25 @@ func refreshAndCacheNewToken(oidcAuthConfig *kubeconfig.OidcConfig,
 
 	tk, err := ts.Token()
 	if err != nil {
+		utils.Log(utils.LevelError, nil, err, "Error: failed to get new token")
```
yolossn, line 746:
```suggestion
		utils.Log(utils.LevelError, nil, err, "failed to get new token")
```

backend/cmd/headlamp.go
```diff
@@ -703,19 +743,25 @@ func refreshAndCacheNewToken(oidcAuthConfig *kubeconfig.OidcConfig,
...
+
 		return "", err
 	}
 
 	idToken, ok := tk.Extra("id_token").(string)
 	if ok {
 		// update cache
 		if err := cache.Set(context.Background(), fmt.Sprintf("oidc-token-%s", idToken), tk.RefreshToken); err != nil {
+			utils.Log(utils.LevelError, nil, err, "Error: failed to cache refresh token")
```
yolossn, line 755:
```suggestion
			utils.Log(utils.LevelError, nil, err, "failed to cache refresh token")
```
Make sure to remove the extra "Error" every where the log level is set to Error.

backend/pkg/utils/utils.go
```diff
@@ -10,3 +17,52 @@ func Contains[T comparable](elems []T, v T) bool {
...
+	}
+
+	if err != nil {
+		// Check if err is a error slice ([]error)
+		if errs, ok := err.([]error); ok {
+			event.Errs("error", errs).Msg(msg)
+		} else {
+			// Handle other error types
+			event.Err(err.(error)).Msg(msg)
+		}
```
yolossn, line 64: The code will reach this state even when a integer is passed as err. Either handle the state properly here or use custom type supported by Golang instead of interface in the function argument.

yolossn (changes requested):
There are few nits apart from that it looks good to me.

yolossn (approved):

backend/pkg/kubeconfig/watcher.go
```diff
@@ -95,7 +101,8 @@ func addFilesToWatcher(watcher *fsnotify.Watcher, paths []string) {
 		// if it isn't, add it to the watcher
 		err := watcher.Add(path)
 		if err != nil {
-			log.Printf("Couldn't add %s to watcher: %v", path, err)
+			utils.Log(utils.LevelError, map[string]string{"path": path},
+				err, "Error adding path to watcher")
```
illume, line 105: I guess there needs to be some consistency in the logs with regards to "Error" being mentioned in the message. 

I expect "Error" should not be used, since it's removed from some of the messages?

backend/pkg/helm/handler.go
```diff
@@ -39,7 +39,8 @@ func NewActionConfig(clientConfig clientcmd.ClientConfig, namespace string) (*ac
 		namespace:    namespace,
 	}
 	logger := func(format string, a ...interface{}) {
-		log.Info().Str("namespace", namespace).Msg(format + "\n" + fmt.Sprintf("%v", a))
+		utils.Log(utils.LevelInfo, map[string]string{"namespace": namespace},
+			nil, format+"\n"+fmt.Sprintf("%v", a))
```
illume, line 43: Should this Sprintf still be there? I think a is passed into the logger already.

illume (changes requested):
Nice one.

I left a few questions/notes.

backend/pkg/utils/utils.go
```diff
@@ -10,3 +17,58 @@ func Contains[T comparable](elems []T, v T) bool {
...
+	// LevelWarn is the warn level.
+	LevelWarn
+	// LevelError is the error level.
+	LevelError
+)
+
+// Log is a wrapper function for logging. It uses zlog package and logs to stdout.
+// It logs the message, source file and line number.
+// It logs the message at the level specified.
+func Log(level uint, str map[string]string, err interface{}, msg string) {
```
illume, line 33: Can you please add test(s) for this function?

illume:
Leaving this topic @joaquimrocha and I discussed about this being a breaking change... It might potentially break someone somewhere, but IMHO I think it's not a wide spread break. We should mention the change to the log format in the release notes.

illume:
Just checking... have you confirmed the logs are compatible with this link Julian left? https://learn.microsoft.com/en-us/azure/developer/java/sdk/logging-overview#default-logger-for-temporary-debugging

I wonder specifically if these are compatible? https://pkg.go.dev/github.com/rs/zerolog#pkg-variables
- field names ( LevelFieldName, MessageFieldName, TimestampFieldName, CallerFieldName )
- time stamp (TimeFieldFormat)

backend/pkg/helm/release.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-playground/validator/v10"
 	"github.com/gorilla/schema"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/utils"
 
 	"github.com/rs/zerolog"
```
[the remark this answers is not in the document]
knrt10: There are some instances in this file were zerolog is used differently than passing as a logging function, that's why it's imported here. Please check https://github.com/headlamp-k8s/headlamp/blob/main/backend/pkg/helm/release.go#L683

backend/pkg/helm/handler.go
knrt10: I think yes, since the last part of Log function is for message so `format+"\n"+fmt.Sprintf("%v", a)` should be kept as it is.

knrt10:
> Just checking... have you confirmed the logs are compatible with this link Julian left? https://learn.microsoft.com/en-us/azure/developer/java/sdk/logging-overview#default-logger-for-temporary-debugging

yes, I pinged him regarding this, we have implemented zlog everywhere which gives JSON data for errors, warning, info and such. It is standard with all the tools since JSON is easily parse able everywhere. Also, if user's usecase is other than JSON logs we have also implemented a wrapper function Log [backend: format backend logs to use zlog by knrt10 · Pull Request #1723 · headlamp-k8s/headlamp (github.com)](https://github.com/headlamp-k8s/headlamp/pull/1723/files#diff-a205007fb04a23a21624acc3bde58fe4d346f7a52fc7ceccf709f3b1dfd2dc6cR33) which can be easily modified as per their needs.

joaquimrocha (approved):

illume (approved):
🎉 🎈

PR #1726, opened 2024-02-19
frontend: get rid of more makeStyles

farodin91:
Down to 32 Files


joaquimrocha (approved):
Interesting. I thought the `styled` would generate unique names which was a problem for the snapshot testing, but I guess it is fine now.

PR #1727, opened 2024-02-19
frontend: Add create namespace ui

### LINKED ISSUE #1705: Add special UI for creating a Namespace
We want to have a + button next to the Namespaces list title in order to open a dialog for creating a new namespace.

Acceptance Criteria:
- [ ] There's an Add button (+ sign) next to the namespaces list view's title
- [ ] When clicking that + button, a new dialog is shown asking the user to input the name of the new namespace.
- [ ] Clicking the create button in the creation dialog will create the new namespace after the typical grace period

### LINKED ISSUE #2055: Provide option to create namespace
When installing helm charts, I need to create a namespace on the fly many times.

Right now, in namespaces there is no option to create a namespace. This forces me to use kubectl just for creating namespace. It would be great if there is an option to create namespace directly.

Screenshot 2024-06-19 at 06 50 56

vyncent-t:
# Add Create Namespace Feature to UI

fixes #1705 
fixes #2055 

## Description
This PR introduces a new feature to the UI that allows users to create namespaces directly from the namespaces list view. A 'Create' button (represented by a + sign) has been added next to the title of the namespaces list view. Upon clicking this button, users are presented with a dialog box where they can input the name for the new namespace. Submitting this by clicking the 'Create' button will create the new namespace.

## Changes
- [x] Added a 'Create' button (+ sign) next to the title of the namespaces list view.
- [x] Implemented a dialog box that prompts the user to enter the name of the new namespace.
- [x] Added functionality to create a new namespace upon clicking the 'Create' button in the dialog box.

## Verification
- [x] Tested the 'Create' button to ensure it triggers the dialog box as expected.
- [x] Verified the dialog box accepts input and allows for the creation of a new namespace.
- [x] Confirmed that the new namespace appears in the namespaces list view upon creation.

## Screenshots
image
image


vyncent-t:
- push to fix build failures and translations

frontend/src/i18n/locales/de/translation.json
```diff
@@ -303,6 +303,7 @@
   "Default Request": "Standardanforderung",
   "Max": "Max",
   "Min": "Min",
+  "Create Namespace": "",
```
knrt10, line 306:
```suggestion
  "Create Namespace": "Namensraum erstellen",
```

frontend/src/i18n/locales/es/translation.json
```diff
@@ -303,6 +303,7 @@
   "Default Request": "Solicitud por defecto",
   "Max": "Máx.",
   "Min": "Mín.",
+  "Create Namespace": "",
```
knrt10, line 306:
```suggestion
  "Create Namespace": "Crear espacio de nombres",
```

frontend/src/i18n/locales/pt/translation.json
```diff
@@ -303,6 +303,7 @@
   "Default Request": "Pedido por defeito",
   "Max": "Máx.",
   "Min": "Mín.",
+  "Create Namespace": "",
```
knrt10, line 306:
```suggestion
  "Create Namespace": "Criar Namespace",
```

knrt10 (commented):
Some suggestions for translations

joaquimrocha (commented):
Thanks Vincent.
Can you move the dialog/button into its own module? So it's more isolated and also in case we want to show it in a different view.

For consistency with the rest of the UX, I think the create button should trigger a cancellable action (like the Editor's apply / Delete / Scale / etc.).

joaquimrocha (changes requested):
Req changes.

vyncent-t:
- last push removes notes

frontend/src/components/namespace/CreateNamespaceDialog.tsx
joaquimrocha, line 1: Call it CreateNamespaceButton.tsx

frontend/src/components/namespace/CreateNamespaceDialog.tsx
```diff
@@ -0,0 +1,92 @@
...
+export function CreateNamespaceButton({ toggleFunc }: CreateNamespaceButtonProps) {
+  const { t } = useTranslation(['glossary', 'translation']);
+
+  return (
+    <ActionButton
+      color="primary"
+      description={t('translation|Create')}
+      icon={'mdi:plus-circle'}
+      onClick={() => {
+        toggleFunc();
```
joaquimrocha, line 29: The logic for toggling the dialog for this button can be done within this button. See the DeleteButton.
We also need it visible only if the user can create namespaces (see use of AuthVisible).

joaquimrocha (changes requested):

vyncent-t:
- last push is still needing some fixes, when creating a namespace item it is almost instant at this current implementation.

maybe artificially adding a timer would allow it to be slow enough to react to a quit action?

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+        dispatch(
+          clusterAction(callback.bind(itemName), {
+            startMessage: t('Creating namespace {{ itemName }}…', { itemName }),
+            cancelledMessage: t('Cancelled creation of {{ itemName }}.', { itemName }),
+            successMessage: t('Created namesapce {{ itemName }}.', { itemName }),
+            errorMessage: t('Error creating namespace {{ itemName }}.', { itemName }),
+            cancelUrl: location.pathname,
+            // startUrl: item!.getListLink(),
+            // errorUrl: item!.getListLink(),
+            // ...options,
```
joaquimrocha, line 67: Please remove this.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    // need a callback so just made one to use
+    const callback = (itemName: string) => {
+      console.log(`Callback called with ${itemName}`);
+    };
+
+    async function createNamesapce(newNamespaceData: any) {
+      try {
+        const response = await post('/api/v1/namespaces', newNamespaceData);
+        console.log('response', response);
```
joaquimrocha, line 55: Please remove this.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const itemName = newNamespaceData.metadata.name;
+
+    // need a callback so just made one to use
+    const callback = (itemName: string) => {
+      console.log(`Callback called with ${itemName}`);
+    };
+
+    async function createNamesapce(newNamespaceData: any) {
+      try {
+        const response = await post('/api/v1/namespaces', newNamespaceData);
```
joaquimrocha, line 54: We need to pass the cluster here. And this function can only be called after the grace period. i.e. it should be passed to the `clusterAction`.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+      },
+    };
+
+    // need this for the item name?
+    const itemName = newNamespaceData.metadata.name;
+
+    // need a callback so just made one to use
+    const callback = (itemName: string) => {
+      console.log(`Callback called with ${itemName}`);
+    };
```
joaquimrocha, line 50: The callback should be the namespace creation (the post).

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const dispatchCreateEvent = useEventCallback(HeadlampEventType.CREATE_RESOURCE);
+
+  const dispatch = useDispatch();
+  const kubeNamespace = Namespace;
+
+  function toggleCreateNamespaceVisibility() {
+    setNamespaceDialogOpen(namespaceDialogOpen => !namespaceDialogOpen);
+  }
+
+  console.log('kubeNamespace', kubeNamespace);
```
joaquimrocha, line 33: Please remove this.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const callback = (itemName: string) => {
+      console.log(`Callback called with ${itemName}`);
+    };
+
+    async function createNamesapce(newNamespaceData: any) {
+      try {
+        const response = await post('/api/v1/namespaces', newNamespaceData);
+        console.log('response', response);
+
+        toggleCreateNamespaceVisibility();
```
joaquimrocha, line 57: I think we can avoid future discrepancies by just explicitly hiding the dialog, instead of toggling the visibility (which I understands hides it, but maybe not in the future).

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+            // errorUrl: item!.getListLink(),
+            // ...options,
+          })
+        );
+
+        return response;
+      } catch (error) {
+        console.error('Error creating namespace', error);
+
+        toggleCreateNamespaceVisibility();
```
joaquimrocha, line 75: Explicitly show the dialog please.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+        return error;
+      }
+    }
+
+    createNamesapce(newNamespaceData);
```
joaquimrocha, line 80: ?

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { clusterAction } from '../../redux/clusterActionSlice';
+import { EventStatus, HeadlampEventType, useEventCallback } from '../../redux/headlampEventSlice';
+import { ActionButton, AuthVisible } from '../common';
+
+export function CreateNamespaceButton() {
+  const { t } = useTranslation(['glossary', 'translation']);
+  const [namespaceName, setNamespaceName] = useState('');
+  const [namespaceDialogOpen, setNamespaceDialogOpen] = useState(false);
+  const dispatchCreateEvent = useEventCallback(HeadlampEventType.CREATE_RESOURCE);
+
```
joaquimrocha, line 25: Remove this newline.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+        toggleCreateNamespaceVisibility();
+        return error;
+      }
+    }
+
+    createNamesapce(newNamespaceData);
+  }
+
+  return (
+    <AuthVisible item={kubeNamespace} authVerb="create" namespace="default">
```
joaquimrocha, line 84: Namespaces aren't namespaced. So the namespace argument is not needed here.
I am not sure the Namespace class can be used here like this either. Maybe AuthVisible needs changes to accommodate what we need.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+        onClick={() => {
+          toggleCreateNamespaceVisibility();
+        }}
+      />
+
+      <Dialog open={namespaceDialogOpen} onClose={() => toggleCreateNamespaceVisibility()}>
+        <DialogTitle>{t('translation|Create Namespace')}</DialogTitle>
+        <DialogContent>
+          <Box component="form">
+            <Typography>Please enter the name of the new namespace</Typography>
```
joaquimrocha, line 98: Needs translation.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,126 @@
...
+            />
+          </Box>
+        </DialogContent>
+        <DialogActions>
+          <Button
+            onClick={() => {
+              toggleCreateNamespaceVisibility();
+              dispatchCreateEvent({
+                status: EventStatus.CONFIRMED,
+              });
```
joaquimrocha, line 116: This is reporting that the namespace creation was confirmed when in fact this button cancels it.

frontend/src/i18n/locales/de/translation.json
```diff
@@ -303,6 +303,7 @@
   "Default Request": "Standardanforderung",
   "Max": "Max",
   "Min": "Min",
+  "Create Namespace": "Namensraum erstellen",
```
joaquimrocha, line 306: Namespace in glossary is translated as "Namespace" since it's a technical term. So in here the translation is not correct.
Please leave it empty or update it to using Namespace.

joaquimrocha (changes requested):

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: would this be correct?
`<Typography>{t('translation|Please enter the name of the new namespace')}</Typography>`

frontend/src/i18n/locales/de/translation.json
vyncent-t: change to `"Create Namespace": ""` ?

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: sorry this was wip and didn't mean to have it reviewed yet!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: sorry this was wip and didn't mean to have it reviewed yet!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: sorry this was wip and didn't mean to have it reviewed yet!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: Fixed it!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: Created a more unique name

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: post logic currently handles setting the cluster

joaquimrocha: Here is the problem: the post handles setting the cluster, but post is called only after the grace period, and by then the user may have changed to another cluster (and that's the one the post picks up). This is why we had to add the cluster as an arg to these functions. In the future we should have the cluster as a mandatory argument to avoid this.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }
+    }
+
+    setNamespaceDialogOpen(false);
+    dispatch(
+      clusterAction(() => namespaceRequest(), {
+        startMessage: t('Creating namespace {{ itemName }}…', { namespaceName }),
+        cancelledMessage: t('Cancelled creation of {{ itemName }}.', { namespaceName }),
+        successMessage: t('Created Namespace {{ itemName }}.', { namespaceName }),
+        errorMessage: t('Error creating namespace {{ itemName }}.', { namespaceName }),
```
joaquimrocha, line 52: The `itemName` doesn't match any key in the passed object. And testing shows:
Screenshot showing the cancel message without the name of the namespace

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+        throw error;
+      }
+    }
+
+    setNamespaceDialogOpen(false);
+    dispatch(
+      clusterAction(() => namespaceRequest(), {
+        startMessage: t('Creating namespace {{ itemName }}…', { namespaceName }),
+        cancelledMessage: t('Cancelled creation of {{ itemName }}.', { namespaceName }),
+        successMessage: t('Created Namespace {{ itemName }}.', { namespaceName }),
```
joaquimrocha, line 51: Do not capitalize the N here. So we are consistent with the rest of the messages.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+        errorMessage: t('Error creating namespace {{ itemName }}.', { namespaceName }),
+        cancelUrl: location.pathname,
```
joaquimrocha, line 53: Since the URL doesn't change. You can use the `cancelCallback` argument instead for bringing up the namespace creation dialog on cancel.

frontend/src/components/namespace/List.tsx
```diff
@@ -86,12 +87,15 @@ export default function NamespacesList() {
...
 
   return (
-    <ResourceListView
-      title={t('Namespaces')}
-      headerProps={{
-        noNamespaceFilter: true,
-      }}
-      {...resourceTableProps}
-    />
+    <>
```
joaquimrocha, line 90: This is not needed AFAIU.

joaquimrocha (changes requested):
Left a few more comments.

vyncent-t:
rebased and changed Namespace to namespace

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: fixed! let me know if this usage is okay, since the variable being used a custom name by the user not sure if it would need a direct translation back to them

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: Fixed! using the cancelCallback to set dialog to open

frontend/src/components/namespace/List.tsx
vyncent-t: The headerProps for titleSideActions is what lets the button render neatly next to the page title

vyncent-t:
Fixed some changes and should be ready for review again

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+        onClick={() => {
+          setNamespaceDialogOpen(true);
+        }}
+      />
+
+      <Dialog open={namespaceDialogOpen} onClose={() => setNamespaceDialogOpen(false)}>
+        <DialogTitle>{t('translation|Create Namespace')}</DialogTitle>
+        <DialogContent>
+          <Box component="form" style={{ width: '20vw', maxWidth: '20vw' }}>
+            <TextField
```
sniok, line 85: It'd be nice to have some validation for the namespace name, like max characters and allowed characters

vyncent-t: I tried using some special chars and it doesnt seem to create the namespace with special chars, the length could be long though I tried that also. What would be a good word length for names?

sniok: According to the docs it should be a valid DNS name, more on that here
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

vyncent-t: Fixed!

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,163 @@
...
+export function CreateNamespaceButton() {
+  const { t } = useTranslation(['glossary', 'translation']);
+  const [namespaceName, setNamespaceName] = useState('');
+  const [namespaceDialogOpen, setNamespaceDialogOpen] = useState(false);
+  const dispatchCreateEvent = useEventCallback(HeadlampEventType.CREATE_RESOURCE);
+  const dispatch = useDispatch();
+
+  function createNewNamespace() {
+    function validateNamespaceName(namespaceName: string) {
+      const namespaceRegex = /^[a-z]([-a-z0-9]*[a-z0-9])?$/;
```
joaquimrocha, line 29: We have a function for this at SettingsCluster.tsx. We should add this as a static method to the Namespace class so we can reuse it when needed.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,163 @@
...
+        alert(
+          t(
+            'translation|Invalid namespace name: name must consist of lower case alphanumeric characters or -'
+          )
+        );
+        return false;
+      }
+
+      return true;
+    }
```
joaquimrocha, line 67: See my comment about the function we have in SettingsCluster.tsx.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,163 @@
...
+      metadata: {
+        name: namespaceName,
+      },
+    };
+    const newNamespaceName = newNamespaceData.metadata.name;
+
+    async function namespaceRequest() {
+      try {
+        const response = await post('/api/v1/namespaces', newNamespaceData, true, {
+          cluster: `${cluster}`,
```
joaquimrocha, line 86: You shouldn't use a string literal here (because if cluster is `null`, it will result in a cluster called `"null"`). You can use `getCluster() || ''` when you declare the var up there.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,163 @@
...
+        });
+        return response;
```
joaquimrocha, line 88: You can return without having to declare a var in between (`return await post...`)

frontend/src/i18n/locales/de/translation.json
```diff
@@ -315,6 +315,15 @@
   "Default Request": "Standardanforderung",
   "Max": "Max",
   "Min": "Min",
+  "Invalid namespace name: name must be 63 characters or shorter": "",
+  "Invalid namespace name: name must start with an alphanumeric character": "",
+  "Invalid namespace name: name must end with an alphanumeric character": "",
+  "Invalid namespace name: name must consist of lower case alphanumeric characters or -": "",
```
joaquimrocha, line 321: Can you reuse the string we already have for SettingsCluster? So we don't have to translate these? (We are missing the 63 chars one, so maybe keep that or adapt the new one)

frontend/src/i18n/locales/fr/translation.json
```diff
@@ -315,6 +315,15 @@
...
   "Min": "Min",
+  "Invalid namespace name: name must be 63 characters or shorter": "",
+  "Invalid namespace name: name must start with an alphanumeric character": "",
+  "Invalid namespace name: name must end with an alphanumeric character": "",
+  "Invalid namespace name: name must consist of lower case alphanumeric characters or -": "",
+  "Creating namespace {{  name  }}…": "",
+  "Cancelled creation of {{  name  }}.": "",
+  "Created namespace {{  name  }}.": "",
+  "Error creating namespace {{  name  }}.": "",
+  "Create Namespace": "",
```
joaquimrocha, line 326: I wonder if in the interest of not having to translate 5 strings per resource, we can reuse the string "Failed to create {{ kind }} {{ name }}." and similar.

joaquimrocha (changes requested):
Left a few notes. Please take a look.

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: fixed!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: fixed!

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: fixed!

frontend/src/i18n/locales/de/translation.json
vyncent-t: fixed!

frontend/src/components/namespace/List.tsx
sniok: I think he meant the Fragment `<></>` that wraps ResourceListView

vyncent-t: removed the `<></>`

vyncent-t:
Fixed start message to the correct use

frontend/src/components/namespace/CreateNamespaceButton.tsx
illume, line 1: Can you please add a CreateNamespaceButton.stories.tsx for this?

frontend/src/lib/k8s/namespace.ts
```diff
@@ -13,6 +13,18 @@ class Namespace extends makeKubeObject<KubeNamespace>('namespace') {
...
+
+  static isValidNamespaceFormat(namespace: string) {
+    // We allow empty strings just because that's the default value in our case.
+    if (!namespace) {
+      return true;
+    }
+
+    // Validates that the namespace is a valid DNS-1123 label and returns a boolean.
+    // https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+    const regex = new RegExp('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$');
```
illume, line 25: There is this rule which the function does not validate:
"contain at most 63 characters"

frontend/src/lib/k8s/namespace.ts
```diff
@@ -13,6 +13,18 @@ class Namespace extends makeKubeObject<KubeNamespace>('namespace') {
   get status() {
     return this.jsonData!.status;
   }
+
+  static isValidNamespaceFormat(namespace: string) {
+    // We allow empty strings just because that's the default value in our case.
+    if (!namespace) {
```
illume, line 19: Can you please add a unit test for this function?

frontend/src/lib/k8s/namespace.ts
```diff
@@ -13,6 +13,18 @@ class Namespace extends makeKubeObject<KubeNamespace>('namespace') {
   get status() {
     return this.jsonData!.status;
   }
+
+  static isValidNamespaceFormat(namespace: string) {
```
illume, line 17: Please add docs for this function?

illume (changes requested):
Nice work.

I left a few notes.

Can you please also rebase this against main? It's been a while.

illume:
The build Container and test job failed so I restarted it in case it was flaky.

frontend/src/lib/k8s/namespace.ts
vyncent-t: fixed!

frontend/src/lib/k8s/namespace.ts
vyncent-t: Fixed!

frontend/src/lib/k8s/namespace.ts
vyncent-t: added a namespace.test.ts for unit testing but I am hitting a blocker with this error

image


The error says the makeKubeObject is not a function, yet it is one and the import for it should be fine on the normal namespace.ts, what is even more weird is that I cannot seem to find how or why its using the Role extends as I do not see anything that Namespace class extends directly off of any Role usage? other than searching project wide for Role or ClusterRole extensions?

vyncent-t:
### Prev push

- fixed (for now) the error when using namespace.test.ts would cause `npm run test` to fail due to "TypeError: makeKubeObject is not a function"

frontend/src/lib/k8s/index.test.ts
```diff
@@ -264,3 +265,24 @@ describe('Test class namespaces', () => {
     expect(classCopy).toEqual({});
   });
 });
+
+/**
+ * This is a function test for the Namespace class.
+ * note: this test will not work correctly if used in its own environment (i.e. namespace.test.ts)
+ */
+describe('Namespace testing', () => {
+  describe('test working isValidNamespaceFormat', () => {
```
joaquimrocha, line 274: You should test more cases:
empty name;
names with capitals;

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+export function CreateNamespaceButton() {
+  const { t } = useTranslation(['glossary', 'translation']);
+  const [namespaceName, setNamespaceName] = useState('');
+  const [namespaceDialogOpen, setNamespaceDialogOpen] = useState(false);
+  const dispatchCreateEvent = useEventCallback(HeadlampEventType.CREATE_RESOURCE);
+  const dispatch = useDispatch();
+
+  function createNewNamespace() {
+    if (!Namespace.isValidNamespaceFormat(namespaceName)) {
+      alert(
```
joaquimrocha, line 29: For telling the user about the error, use the error + helper text functionality of the text field as we do in the cluster settings.

joaquimrocha (changes requested):
Thanks Vincent. I left the comments we had also talked about during our call.

vyncent-t:
### Push
- fixes review comments

- [ ] current push does not allow users to input empty string for a namespace name, need to implement a way to use a default name

vyncent-t:
- [ ] create an error for if namespace already exist

vyncent-t:
Currently trying to work around an issue for the storybook component, it seems that the storybook compile is running into the same circular dependency error that the namespace.test.ts was hitting (in short it would fail if we tried to create a unit test for the namespace functions in its own .test env) 

image

vyncent-t:
- [ ] need to fix being able to use enter key to create

vyncent-t:
### Push note
- [x] Fixed page reloading when pressing enter when writing name
- [ ] story component still broken

vyncent-t:
### push note

- Removed the <AuthVisible> comp around the create button as it is not working correctly with the params (example: this doesn't prevent the button from rendering)
image

- [x] Fixed story comp not rendering

vyncent-t:
### prev push
- [x] Adds working story components

vyncent-t:
The build storybook error is likely a bug due to the storybook vite build on the publisher side https://github.com/storybookjs/storybook/issues/26685

I may just remove the test for now until this is fixed for vite

vyncent-t:
I was able to find a fix, seems the latest update isn't vite compatible so rolled back to the previous version

- storybook/addons-interactions `7.6.20`

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+        return await post('/api/v1/namespaces', newNamespaceData, true, {
+          cluster: clusterData || '',
+        });
+      } catch (error) {
+        console.error('Error creating namespace', error);
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        if (errorMessage.includes('already exists')) {
+          setNamespaceDialogOpen(true);
+          setNamespaceNameValid(false);
+          setNameHelperMessage(t('translation|Namespace name already exists.'));
```
joaquimrocha, line 52: This text sounds a bit odd. Maybe just "Namespace already exists" or "A namespace with this name already exists"

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+    async function namespaceRequest() {
+      try {
+        return await post('/api/v1/namespaces', newNamespaceData, true, {
+          cluster: clusterData || '',
+        });
+      } catch (error) {
+        console.error('Error creating namespace', error);
+        const errorMessage = error instanceof Error ? error.message : String(error);
+        if (errorMessage.includes('already exists')) {
```
joaquimrocha, line 49: Checking for substrings in error messages is not a good practice (as those are not standard and may change).
Can you check for the error code instead?

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+      metadata: {
+        name: namespaceName,
+      },
+    };
+
+    const newNamespaceName = newNamespaceData.metadata.name;
+
+    async function namespaceRequest() {
+      try {
+        return await post('/api/v1/namespaces', newNamespaceData, true, {
```
joaquimrocha, line 43: Please use `Namespace.apiEndpoint.post`.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+              margin="dense"
+              id="name"
+              aria-label="Name"
+              label={t('translation|Name')}
+              type="text"
+              error={!namespaceNameValid && namespaceName.length > 0}
+              helperText={!namespaceNameValid && namespaceName.length > 0 ? nameHelperMessage : ''}
+              fullWidth
+              value={namespaceName}
+              onChange={event => setNamespaceName(event.target.value)}
```
joaquimrocha, line 127: We could lowercase the name of the namespace here. So users would have it automatically done.

joaquimrocha (approved):
Almost there. Left a few comments after trying it out.
Seems also like some storybook snapshots that are unrelated did get updated. So I am not sure if it's a legitimate side-effect or if we're now breaking the snapshot generation with this addon.

joaquimrocha (changes requested):
Approved by mistake. I meant it needs changes.

vyncent-t:
Still can't seem to source that build error?

vyncent-t:
rebased to current main and updated the `"@types/lodash": "^4.17.4",` to avoid compatibility error, its now `"@types/lodash": "4.17.4",`

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+        creationTimestamp: new Date().toISOString(),
+        uid: '',
+      },
+    };
+
+    const newNamespaceName = newNamespaceData.metadata.name;
+
+    async function namespaceRequest() {
+      try {
+        const response = await Namespace.apiEndpoint.post(newNamespaceData, {}, clusterData || '');
```
joaquimrocha, line 44: If this succeeds we should set the namespace as '', so if we press the button again, we don't see the name of the namespace we just created.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { getCluster } from '../../lib/cluster';
+import Namespace from '../../lib/k8s/namespace';
+import { clusterAction } from '../../redux/clusterActionSlice';
+import { EventStatus, HeadlampEventType, useEventCallback } from '../../redux/headlampEventSlice';
+import { ActionButton } from '../common';
+
+export default function CreateNamespaceButton() {
+  const { t } = useTranslation(['glossary', 'translation']);
+  const [namespaceName, setNamespaceName] = useState('');
+  const [namespaceNameValid, setNamespaceNameValid] = useState(false);
```
joaquimrocha, line 22: It's a good practice to prefix boolean variables about state with `is`. `isNamespaceNameValid`.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+          t(
+            "translation|Namespaces must contain only lowercase alphanumeric characters or '-', and must start and end with an alphanumeric character."
+          )
+        );
+      }
+    }
+  }, [namespaceName]);
+
+  return (
+    <>
```
joaquimrocha, line 99: The name needs to be wrapped in a AuthVisible component, because we don't want to show it unless the user can create the namespace.

joaquimrocha (changes requested):
Almost there. Found a few things in the code.

frontend/src/components/namespace/CreateNamespaceButton.tsx
```diff
@@ -0,0 +1,168 @@
...
+          setNamespaceDialogOpen(true);
+        },
+      })
+    );
+  }
+
+  useEffect(() => {
+    setIsValidNamespaceName(Namespace.isValidNamespaceFormat(namespaceName));
+
+    if (!Namespace.isValidNamespaceFormat(namespaceName)) {
```
joaquimrocha, line 84: These 2 lines call the same function, please set the result of isValidNamespaceFormat to a const and reuse it.

joaquimrocha (changes requested):
Just one last change needed.

frontend/src/components/namespace/CreateNamespaceButton.tsx
vyncent-t: done!

joaquimrocha (approved):

PR #1741, opened 2024-02-24
Notarize Mac app using ESRP

joaquimrocha:
We are moving the app in Apple from the legacy Kinvolk developer account to Microsoft's.
For that, we have to use the ESRP tool that is also used to sign the Windows binaries.

- [X] Replace the Mac signing mechanism by ESRP
- [X] Use just one ESRP script for Windows and Mac signing

How to test:
- In the actions, run the Windows and Mac artifacts with the Sign/Notarize option checked: install the resulting binaries and verify that the OS doesn't complain about them being unsigned


app/scripts/esrp.js
```diff
@@ -0,0 +1,238 @@
...
+    throw new Error('Invalid operation. The options are:', SIGN_OPS.join(', '));
+  }
+
+  const policyJson = path.resolve(os.tmpdir(), 'Policy.json');
+  fs.writeFileSync(policyJson, JSON.stringify(POLICY_JSON, undefined, 2));
+  const authJson = path.resolve(os.tmpdir(), 'Auth.json');
+  fs.writeFileSync(authJson, JSON.stringify(AUTH_JSON, undefined, 2));
+
+  try {
+    execSync(`${esrpTool} Sign -l Verbose -a ${authJson} -p ${policyJson} -i ${signInputJson}`);
```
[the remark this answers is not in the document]
joaquimrocha: This is for building the app only and the code is not shipped in Headlamp. So I don't think there's a possibility of attack as it could be if the script was called during the execution of Headlamp.

app/mac/scripts/codeSign.js
```diff
@@ -0,0 +1,25 @@
...
+    console.log('Mac codesign: No Apple Team ID found, skipping codesign');
+    return;
+  }
+
+  const entitlementsPath = path.resolve(path.join(__dirname, '..', 'entitlements.mac.plist'));
+
+  let exitCode = 0;
+  try {
+    execSync(
+      `codesign -s ${teamID} --deep --force --options runtime --entitlements ${entitlementsPath} ${config.app}`
```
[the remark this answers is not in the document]
joaquimrocha: This is for building the app only and the code is not shipped in Headlamp. So I don't think there's a possibility of attack as it could be if the script was called during the execution of Headlamp.

app/package.json
```diff
@@ -148,7 +149,7 @@
     "@babel/preset-typescript": "^7.15.0",
     "@electron/notarize": "^2.1.0",
     "electron": "^27.0.4",
-    "electron-builder": "^24.0.0",
+    "electron-builder": "^24.9.1",
```
illume, line 152: I think package-lock.json needs to be committed too?

.github/workflows/app-artifacts-mac.yml
```diff
@@ -35,22 +41,106 @@ jobs:
...
+      run: |
+        xcrun stapler staple ./dmgs/Headlamp*.dmg
+    - name: Upload Stapled
+      uses: actions/upload-artifact@v4
+      with:
+        name: dmgs
+        path: ./dmgs/Headlamp*.dmg
+        if-no-files-found: error
+        overwrite: true
+        retention-days: 2
```
illume, line 146: Should this be only 2 days? If it's the last step, maybe it should stick around longer.

app/package.json
joaquimrocha: Good catch! Thanks.

.github/workflows/app-artifacts-mac.yml
joaquimrocha: Before we had 1 day 🙂 
This should be fine since this action is mostly used for quick testing or for the generation of the release artifacts, in which case I do upload them to the long term release storage.

illume (approved):
👍 lgtm

note: I didn't test the binary (but I saw Chris did?).

joaquimrocha:
> 👍 lgtm
> 
> note: I didn't test the binary (but I saw Chris did?).

Yup, he has.

PR #1749, opened 2024-02-28
frontend: Add owned pods and jobs events

ashu8912:
Added new events to cover owned pod and owned jobs


joaquimrocha (commented):
What if we instead dispatch something like:
`DETAILS_VIEW_OWNED_RESOURCES`
and then we have:
```typescript
{
  resources: {
     [resourceKind: 'Event' | 'Pod' | 'Job' ...]: KubeObject[]
  },
  owner: KubeObject
}
```

Maybe this is generic enough that we can add it for many things in the future.

illume:
I restarted the failed Build container job, which failed because of a flaky e2e test that's been fixed in main.

PR #1750, opened 2024-02-28
Cluster in hook api functions / Warnings in home view

### LINKED ISSUE #1747: Allow to use the cluster argument in useList/Get functions
We recently added the possibility to specify the cluster in the apiProxy functions, but we don't have it in the more convenient functions/hooks in lib/k8s/cluster.ts and developers have requested this (see [#1459](https://github.com/headlamp-k8s/headlamp/issues/1459)).

This is needed for the new home view, as we need to list pods/events/etc in the home view, which is not a view where a cluster is defined as the context/URL.

fixes #1459

joaquimrocha:
Allow useGet, useList, and apiGet/apiList functions to specify the cluster.
This is needed and it's possible now that we have a way to specify the cluster in apiProxy.

As part of testing whether this was working, I ended up adding the number of warnings for the clusters in the home view. Which is part of the new design when we implement the new multi-cluster view.

fixes #1747


ashu8912 (approved):
LGTM!!

PR #1751, opened 2024-02-28
frontend ClusterChooser: Fix color (make it black again)

joaquimrocha:
After 2bc3979e7c41f264226bfb44e01e03f679b3a370 , the cluster chooser button's color became blue as apparently the sx prop isn't picking up custom theme entries. This patch fixes that by accessing the theme directly in the sx prop.

cc/ @farodin91 This may be important if you add more style replacements (the sx prop doesn't seem to find custom props by string directly, so they need to be referred by theme).


ashu8912 (approved):
Tested, LGTM

Issue #1755, opened 2024-02-29
logger source is always the same


illume:
All the logs have the same `source:` field, which makes it hard to debug. 

{"level":"info","source":"/home/rene/dev/headlamp/backend/pkg/logger/logger.go","line":27,"time":"2024-02-29T16:05:18+01:00","message":"Proxy URLs: []"}

PR #1756, opened 2024-02-29
backend: Fix log file source

knrt10:
This fixes the log source file. Earlier the runtime.Caller function was 1 in the log function. This change ensures that we retrieve the source file from the caller of the Log function, rather than the log function itself.

Fixes: #1755


illume (approved):
👍 thanks

Issue #1744, opened 2024-02-26
Run commands only after user consent


joaquimrocha:
With #1677 we allow plugins to run commands when in the app mode, but we should make this option available only after user consent, for extra security.

Since the command running is done in the app layer, I think we should be able to show a (native/electron) dialog asking the user to run the command. If there was a way to have a list of allowed commands for the user to review, that'd be even better, but maybe that's too complicated.

I can think of a couple of options for the UX, but maybe this works:
 * Plugin calls runCommand, which continues till the request is done to the electron layer
 * The electron app checks the run-command-consent setting:
   * If **true**: continue to running the command
   * If **false**: trigger the ask-for-command-consent dialog; after which the command is run.

In general settings, we should have a setting showing whether commands are allowed to run. When clicking the switch, the dialog runs again (we can **never** set the setting directly/automatically from the frontend, else it can simply be faked without user intervention)

illume:
Done in 
- #1995

PR #1757, opened 2024-03-01
frontend: runCommand: Disable unless env var is set

illume:
For https://github.com/headlamp-k8s/headlamp/issues/1744


ashu8912:
Thanks for the change @illume but is this a temporary solution as of now to not allow running Commands???

illume:
@ashu8912 yeah, this is temporary. We're not using run commands function yet, so disabling it is ok. 

I'll work on adding the UX described in #1744 after the release.

PR #1763, opened 2024-03-04
frontend: Fix redirect on Save/Cancel in Plugin Settings Details View

yolossn:
this patch adds redirect to plugin settings list
view when the user clicks Save/Cancel. When save
is clicked the data is saved and redirected to
the plugin settings list view.


knrt10 (approved):
+1

joaquimrocha (approved):

Issue #1745, opened 2024-02-27
Add k8s API tests


joaquimrocha:
Using kwok, we should be able to set up tests for the common API (in apiProxy.ts) methods. This is important for us to more safely be able to simplify the k8s API files.

Acceptance Criteria:
- [x] We have test coverage for (at least) the most important functions apiProxy file

PR #1765, opened 2024-03-04
frontend: k8s: Add k8s API tests

skoeva:
This gives us test coverage of the most important apiProxy functions. This is important for us to more safely be able to simplify the k8s API files.

Fixes: #1745 

### Testing
- [x] Run `cd frontend && npm test -- --coverage --testPathPattern=apiProxy.test.ts`

### Remaining work
We have full coverage of the exposed functions, though E2E tests could potentially be added for more rigorous testing.

### Difficulties
The stream functions were tricky to test since they depend on other functions which needed testing. The last three groups of functions (port forward, drain node, and delete plugin) all needed custom jest mock implementations for fetch, which may be preferable to introducing a new package since nock does currently allow for this.

### Documentation
Here's the documentation for testing with nock: https://github.com/nock/nock


illume:
For the git comments, we follow the linux kernel style format. First subject line in imperative mood (Add instead of Added) at 50 chars max starting with the context. Then more description if it's needed in the body 72 chars after a blank line.
eg. 
```
frontend: TastyFood: Add component to show delicious food

Body goes here with extra if you need it in 72 chars per line... It allows 
showing tasty food. We needed this because tasty food deserves to be
shown in a special way.
```

Note the context can have "frontend:" or "backend:" at the start, but please provide the context of the module(s) you're working on.

We do atomic commits with one commit containing a self contained thing.  This means when you read each commit by itself, it should be able to stand by itself. It should be able to pass all CI checks by itself. Depending on the previous commits is fine, but it should pass all checks and do one thing.

vyncent-t:
hey! just curious, does the last commit solve the shell bot complaint?

skoeva:
> hey! just curious, does the last commit solve the shell bot complaint?

@vyncent-t I think so, that complaint came up after rebasing so don't think any changes I made caused it

plugins/headlamp-plugin/src/stateless/index.ts
```diff
@@ -0,0 +1,431 @@
+import * as jsyaml from 'js-yaml';
```
illume, line 1: Is this file supposed to be included?

plugins/headlamp-plugin/package.json
```diff
@@ -108,6 +111,7 @@
     "monaco-editor-webpack-plugin": "^4.2.0",
     "msw": "^0.47.4",
     "msw-storybook-addon": "^1.10.0",
+    "nock": "^13.5.4",
```
illume, line 114: Can you please explain why nock is used over other options?

skoeva: it's much simpler to set up than, say, MSW, with comparable if not better performance at scale. but I imagine either would work

plugins/headlamp-plugin/src/stateless/index.ts
skoeva: AH I looked through and didn't find this used anywhere, I believe it may have been created when running `npm audit fix` (I noticed it created some redundant unused files), thanks for picking this up

knrt10 (commented):
Nice work @skoeva. Can you please add a make recipe to run the tests? Also, please add to run it in CI. Thanks

skoeva:
> Nice work @skoeva. Can you please add a make recipe to run the tests? Also, please add to run it in CI. Thanks

@knrt10 it should be running in CI as part of `make frontend-test`, is that what you mean?

knrt10:
Ahh, I thought it was running differently. Thanks for the clarification. Will review the code.

frontend/src/lib/k8s/apiProxy.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+});
+
+describe('request, post, patch, put, delete', () => {
+  const testData = { key: 'value' };
+
+  beforeEach(() => {
+    nock(baseApiUrl)
+      // Successful GET request
+      .persist()
+      .get(`/clusters/${clusterName}` + testPath)
```
illume, line 266: Can you please use template literals for this? (and in other places)

Like this:
```
`/clusters/${clusterName}${testPath}`
```
Rather than:
```
`/clusters/${clusterName}` + testPath
```

illume (changes requested):
Looking good. Thanks for this epic amount of work.

I left a few notes and questions for your consideration.

A couple of things to make it a bit easier to review would be much appreciated:
- Would you be able to break this down into separate commits for each thing you're testing?
- Can you please add into a PR description how to test this PR? Also please add how to run just these tests, and how to see the coverage for them?

Some other questions...
- What parts are left to test for later work?
- Which functions did you notice had bugs or were particularly smelly, hard to understand or hard to test?

plugins/headlamp-plugin/package.json
illume: Ok, cool. Sounds like we should use nock then.

---

I looked into it a bit further and here are some notes in case anyone else is interested about the packages added.

I see we're already sort of msw in the storybook files. But only to mock some of the calls that some components are making. So I'd like to know if there is a good reason to add a new dependency. But also evaluate each dependency.

https://mswjs.io/docs/comparison
https://github.com/nock/nock
https://github.com/mswjs/msw
https://securityscorecards.dev/viewer/?uri=github.com/nock/nock
https://securityscorecards.dev/viewer/?uri=github.com/mswjs/msw
https://www.npmjs.com/package/nock
https://www.npmjs.com/package/msw

The tests are already using nock, so avoiding rewrite work would be a plus. I'm not sure how complicated a rewrite it would be(chuck it into copilot and it's done?).

It appears we couldn't use nock for the in-browser tests, because it only works on node. But saying that it appears our msw setup in storybook is at least partially broken. 

Also we're using a pretty old msw 0.47.4 and it's up to 2.x already. So we'd need to upgrade that. Except https://github.com/mswjs/msw-storybook-addon only supports < 2.0 at the moment. This means we can't use msw here at the moment without taking on the need to rewrite it later.

https://www.npmjs.com/package/ws
https://github.com/websockets/ws
https://securityscorecards.dev/viewer/?uri=github.com/websockets/ws

frontend/src/lib/k8s/apiProxy.test.ts
skoeva: adding in the next push

frontend/src/lib/k8s/apiProxy.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+
+      socket.send(JSON.stringify(errorMessage));
+      await new Promise(resolve => setTimeout(resolve, 100));
+      expect(cb).toHaveBeenNthCalledWith(2, errorMessage.object);
+
+      done();
+    });
+  });
+
+  it('Successfully handles stream cancellation', done => {
```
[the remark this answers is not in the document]
skoeva: adding in the next push

frontend/src/lib/k8s/apiProxy.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+        .streamResultsForCluster(
+          streamResultsUrl,
+          { cb, errCb, cluster: clusterName },
+          { watch: '1' }
+        )
+        .then(cancel => {
+          mockServer.on('connection', async socket => {
+            socket.send(JSON.stringify({ type: 'ADDED', object: mockConfigMap }));
+
+            setTimeout(() => {
```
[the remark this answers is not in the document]
skoeva: I left the timeouts in to be safe since there were some tests where we would not get a callback if we didn't wait, but I see now that some tests still pass without them. working on this now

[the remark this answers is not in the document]
skoeva: oops turns out most of them did not need them, I removed all the unnecessary ones and added comments for the ones that were. in each case, we would either get an empty callback or none at all, so the sleep calls are there to ensure we get proper callbacks

plugins/headlamp-plugin/package.json
skoeva: it would make sense to me if we wanted to stay consistent and use msw throughout (it's more versatile than nock, anyway) and I don't think a refactor in msw would be too difficult like you mention, but performance-wise nock may be better down the road

here's a comparison between the two:
https://www.bam.tech/article/nock-vs-msw-i-tested-both-and-here-is-what-i-learned

illume:
Thanks for the updates @skoeva 

Would you mind adding "frontend:" to the beginning of the subject line in your commits? Something like...

- "frontend: Add nock dependency for testing"
- "frontend: apiProxy.test.ts: Add tests for cluster functions"

skoeva:
> Thanks for the updates @skoeva
> 
> Would you mind adding "frontend:" to the beginning of the subject line in your commits? Something like...
> 
> * "frontend: Add nock dependency for testing"
> * "frontend: apiProxy.test.ts: Add tests for cluster functions"

for sure, added them in

illume (approved):
👍

PR #1769, opened 2024-03-05
Add status filter for KubeCRDs

lalexander07:
This pr is to be able to filter by statuses (partner ask was specifically pod statuses) on WebXT Falcon's clusterfleet explorer headlamp plugin

### How to test

Go to Pods, click on the filter, enter some status in the "Statuses" filter.
image


frontend/src/components/common/StatusFilter.tsx
```diff
@@ -0,0 +1,138 @@
...
+        });
+
+        return (
+          <Typography style={{ overflowWrap: 'anywhere' }}>
+            {statusesToShow.length > maxStatusesChars
+              ? statusesToShow.slice(0, maxStatusesChars) + '…'
+              : statusesToShow}
+            {tags.length > joinnedStatuses && (
+              <>
+                <span>,&nbsp;</span>
```
illume, line 95: If at all possible, we avoid html tags and use Mui components. Could you please switch these ones out for mui components?

frontend/src/components/common/StatusFilter.tsx
```diff
@@ -0,0 +1,138 @@
...
+      autoComplete
+      options={statusNames}
+      onChange={onChange}
+      onInputChange={onInputChange}
+      inputValue={statusInput}
+      // We reverse the namespaces so the last chosen appear as the first in the label. This
+      // is useful since the label is ellipsized and this we get to see it change.
+      value={[...filter.statuses.values()].reverse()}
+      renderOption={(props, option, { selected }) => (
+        <li {...props}>
```
illume, line 59: If at all possible, we avoid html tags and use Mui components. Could you please switch these ones out for mui components?

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -110,54 +114,57 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
...
-            >
-              {t('Clear')}
-            </Button>
+            <NamespacesAutocomplete />
           </Grid>
+        )}
+        <Grid item>
+          <TextField
+            id="standard-search"
+            label={t('Search')}
```
illume, line 143: I think it needs the namespace added... `t('translation|Search")`

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -110,54 +114,57 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
...
-      );
-    }
+        <Grid item>
+          <Button
+            variant="contained"
+            endIcon={<Icon icon="mdi:filter-variant-remove" />}
+            onClick={resetFilters}
+            aria-controls="standard-search"
+          >
+            {t('Clear')}
```
illume, line 163: I think it needs the namespace added... `t('translation|Clear")`

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -110,54 +114,57 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
...
-        <Grid container alignItems="flex-end" justifyContent="flex-end" spacing={1} wrap="nowrap">
-          {!noNamespaceFilter && (
-            <Grid item>
-              <NamespacesAutocomplete />
-            </Grid>
-          )}
+  if (!showFilters.show) {
+    actions.push(
+      <IconButton
+        aria-label={t('Show filter')}
```
illume, line 120: I think it needs the namespace added... `t('translation|Show filter")`

illume (changes requested):
Thank you for this.

I left a few notes about the html tags, and in few spots where the translation might need attending (the old code didn't have a namespace... maybe it was buggy?).

Would you mind rebasing so there are no merge commits?

We do atomic commits with one commit containing a self contained thing. For the git comments, we follow the linux kernel style format. First subject line in imperative mood at 50 chars max starting with the context. Then more description if it's needed in the body 72 chars after a blank line.
eg. 
```
frontend: StatusFilter: Add component to filter by pod statuses

Body goes here with extra if you need it in 72 chars per line... It facilitates 
resource filtering based on statuses with an autocomplete input field.
Works seamlessly with existing filters.
```

I haven't got around to testing it yet... but I will do that next.

frontend/src/components/common/StatusFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { useDispatch } from 'react-redux';
+import { useHistory, useLocation } from 'react-router-dom';
+import { addQuery } from '../../helpers';
+import { setStatusFilter } from '../../redux/filterSlice';
+import { useTypedSelector } from '../../redux/reducers/reducers';
+
+export interface PureStatusesAutocompleteProps {
+  statusNames: string[];
+  onChange: (event: React.ChangeEvent<{}>, newValue: string[]) => void;
+  filter: { statuses: Set<string>; search: string };
```
illume, line 20: I think if you made this `filter?:` and change the rest of the code to account for that some of the tests would pass. Because currently they are sending filter as undefined.

plugins/examples/change-logo/src/index.tsx
```diff
@@ -20,18 +20,19 @@ import Settings, { store } from './settings';
  * it's easier to make it look good with light and dark themes.
  */
 function SimpleLogo(props: AppLogoProps) {
-  const { logoType, className } = props;
+  const { logoType, className, sx } = props;
```
illume, line 23: I guess there was some sort of error during rebasing and this commit snuck into the PR somehow.

frontend/src/redux/filterSlice.ts
```diff
@@ -39,6 +43,14 @@ export function filterResource(
     return false;
   }
 
+  let status: string = '';
+  let phase: string = '';
+  if (isKubeCRD(item)) {
+    const kubeCRD: KubeCRD = item as KubeCRD;
+    status = String(kubeCRD?.status);
```
illume, line 50: I'm not sure if this was intentional, or maybe the isKubeCRD(item) guard covers this case? When it's undefined the result is... `String(undefined) -> 'undefined'`.

If it's a problem, an alternative would be something like:
`status = kubeCRD?.status ? kubeCRD?.status : '';`

illume:
I'm trying it out, and get this message repeated over and over in the console.
```ts
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    at StatusesAutocomplete (http://localhost:3000/static/js/bundle.js:27277:80)
...
    at SectionHeader (http://localhost:3000/static/js/bundle.js:26169:5)
    at SectionFilterHeader (http://localhost:3000/static/js/bundle.js:25843:5)
```

illume:
Note, I see the Statuses drop box is appearing on views without a Status column.

image

joaquimrocha:
Hi @lalexander07 . Thanks for the PR.
I understand the need for filtering on status and we have had a similar request in the past for other columns.
Headlamp's SimpleTable uses the MUI Table within, which is a very simple table, meaning it's not easy to provide e.g. filter by column value.

Due to the space around the filter UI being limited, and given that other partners have asked for different columns to be filtered, I have been pushing back on adding more fields to the SectionFilterHeader.
We do have it in our backlog to replace the tables we use by a more advanced table that provides column filtering for "free". See https://github.com/headlamp-k8s/headlamp/issues/1646 .

Till we replace the table component, maybe we could have an alternative approach here: instead of having a dedicated entry for the status search, how about we use a search prefix? e.g. `status:running` would show only rows whose status column matches the value "running". This approach is also used in other popular apps like Slack.
We could add a button to the search input field (as an "start adornment" in MUI) to select any searchable column, so the users do not have to guess that searching by column requires a prefix, i.e. that extra button in the input shows a dropdown with the options "Search by: Status | Node | ...". This would also allow to combine search column values: `status:running,node:my-super-node`.

What do you think?

illume:
Hrmm. The problem with the status:running approach is that it's not easily discoverable.

joaquimrocha:
> Hrmm. The problem with the status:running approach is that it's not easily discoverable.

Correct, but if we add that dropdown filter in the input, then it should fix that issue.
I have talked to @lalexander07 offline and this solution tackles the problem for them, while it is also a more scalable fix for us, so it feels like a good way forward.

ashu8912:
One thing I noticed while testing this PR is that unlike namespace filter the queryParam status is not persisted on view change but the state still shows the updated values

ashu8912:
Also the tile of this PR is a bit confusing to me, are we adding this to the CRDs, how??? Maybe changing it to something like adding status filter to views where we have the status column would make more sense

joaquimrocha:
@lalexander07 , we now (for a couple of releases) support filtering by any column. Hope these cover your use-cases.
Closing the issue.

PR #1771, opened 2024-03-06
frontend plugins: Fix enable button disappearing

joaquimrocha:
The enable button was not being shown for plugins unless there was at
least one incompatible plugin being processed. This happened because
the code that set the isCompatible part of each plugin info was not
being directly set unless there were incompatible plugins.

Signed-off-by: Joaquim Rocha <joaquim.rocha@microsoft.com>


knrt10 (approved):

illume (approved):
👍

PR #1782, opened 2024-03-11
frontend: Show label as terminated when exitCode is not available

### LINKED ISSUE #1778: Init containers always show status as "Error" when in terminated state
There probably needs to be a conditional that only sets the state to error if the return code is not 0.

https://github.com/headlamp-k8s/headlamp/blob/7d25e012ea6f5dacde0a8ad746ea2d9349eec674/frontend/src/components/common/Resource/Resource.tsx#L679

ashu8912:
fixes #1778


illume (changes requested):
There's some CI failures.

ashu8912:
I think this is resolved

PR #1785, opened 2024-03-12
backend: Update golang to 1.22.0

knrt10:
Update golang version, minimum version is 1.22


illume:
I'm not sure about putting the mod tidy in there. Because it should be tidy.

.github/workflows/backend-test.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           testcoverage="${{ env.coverage }}"
           git fetch origin "$base_branch"
           git checkout "origin/$base_branch"
+          go mod tidy
```
illume, line 56: Can you please explain this one?

I don't think we should do this here. Because the module system should be working for anyone checking it out. Having this here means we don't test that `go mod tidy` was ran.

backend/.golangci.yml
```diff
@@ -3,8 +3,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - deadcode
```
illume, line 6: Were they removed because they don't work anymore or just that they are deprecated? Maybe it can be that we continue using them a bit longer if they still work.

I wonder if there are alternatives for the ones removed?

.github/workflows/backend-test.yml
knrt10: Make sense. When updating Golang to 1.22, what happens is that https://github.com/headlamp-k8s/headlamp/blob/main/.github/workflows/backend-test.yml#L56 is unable to get the base branch coverage because it is still running on 1.20. I discussed this with Santhosh and the logical way is to update Golang version first in a separate PR and the other things in this PR.

backend/.golangci.yml
knrt10: Many of them were deprecated and suggested alternatives which I have added in the PR. For example deadcode is replaced by `unused`

illume (approved):
🎉🎈

Thank you.

PR #1796, opened 2024-03-13
github: build-container: Cache container build for faster testing

illume:
So that e2e tests can be run faster if only the e2e test changes.

Especially useful if there is a test that's flaky on CI, because timing is different there. Then you can try different test changes more quickly, because the container image does not need to be built each time.

# Testing done

- Here's an action run that caches the built headlamp container images: https://github.com/headlamp-k8s/headlamp/actions/runs/8262403292/job/22601643056?pr=1773
- Here's one that uses the restored container images and skips building the container image making the test run 8 minutes quicker: https://github.com/headlamp-k8s/headlamp/actions/runs/8262137512/job/22600860057?pr=1773

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -12,6 +12,7 @@ on:
     - Makefile
     - '.github/**'
     - Dockerfile
+    - Dockerfile.plugins
     - 'e2e-tests/**'
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-container.yml
```diff
@@ -114,6 +133,21 @@ jobs:
         else
           echo "Playwright tests passed successfully"
         fi
+    - name: Save Docker Images to Tar files in image-cache Folder
+      if: steps.cache-image-restore.outputs.cache-hit != 'true'
+      run: |
+        export SHELL=/bin/bash
```
illume, line 139: Probably add `set -o errexit` to these blocks. So if there's an error CI fails.

.github/workflows/build-container.yml
```diff
@@ -28,7 +29,22 @@ jobs:
       # now you can run kubectl to see the pods in the cluster
     - name: Try the cluster!
       run: kubectl get pods -A
+    - name: Restore image-cache Folder
+      id: cache-image-restore
+      uses: actions/cache/restore@v4
+      with:
+        path: ~/image-cache
+        # cache the container image image. All the paths this PR depends on except the e2e-tests folder for the key.
```
knrt10, line 37:
```suggestion
        # cache the container image. All the paths this PR depends on except the e2e-tests folder for the key.
```

knrt10 (approved):
NIT, but can be updated later. Awesome work. This is great TBH

PR #1797, opened 2024-03-13
e2e-tests headlampPage: Fix flaky test to wait for page to load

illume:
Sometimes the token would not be in the URL in time, and it would fail the tests. So we wait for the heading to be there.


knrt10 (approved):
+1. Thank you for the PR :-)

PR #1805, opened 2024-03-14
backend: Update go.mod to 1.22

knrt10:


knrt10:
This is not needed. Ci will always fail closing it in favour of #1785

Issue #1780, opened 2024-03-08
Cluster delete doesn't work


jowharshamshiri:
Clicking delete in the context menu opened when clicking on the 3-dots in All Clusters list doesn't work. 
Context info: the cluster config I am deleting is no longer valid, that might be needed to reproduce the issue.
It seems instead of just removing the config from the list, it is asking minikube to delete the cluster which is weird. Doubly so bc it doesn't even ask for confirmation.

Here's the stacktrace from the developer console:
`

DELETE http://localhost:4466/cluster/minikube 500 (Internal Server Error)
b @ apiProxy.ts:376
y @ apiProxy.ts:301
V @ apiProxy.ts:1621
(anonymous) @ index.tsx:33
onClick @ index.tsx:92
Ve @ react-dom.production.min.js:52
Ke @ react-dom.production.min.js:52
(anonymous) @ react-dom.production.min.js:53
Dr @ react-dom.production.min.js:100
Lr @ react-dom.production.min.js:101
(anonymous) @ react-dom.production.min.js:113
Ie @ react-dom.production.min.js:292
(anonymous) @ react-dom.production.min.js:50
Mr @ react-dom.production.min.js:105
Jt @ react-dom.production.min.js:75
Yt @ react-dom.production.min.js:74
t.unstable_runWithPriority @ scheduler.production.min.js:18
zi @ react-dom.production.min.js:122
Pe @ react-dom.production.min.js:292
Xt @ react-dom.production.min.js:73

Unable to parse error json at url: http://localhost:4466/cluster/minikube {err: SyntaxError: Unexpected token 'r', "removing c"... is not valid JSON} with request data: {signal: AbortSignal, headers: {…}, method: 'DELETE'}
b @ apiProxy.ts:415
await in b (async)
y @ apiProxy.ts:301
V @ apiProxy.ts:1621
(anonymous) @ index.tsx:33
onClick @ index.tsx:92
Ve @ react-dom.production.min.js:52
Ke @ react-dom.production.min.js:52
(anonymous) @ react-dom.production.min.js:53
Dr @ react-dom.production.min.js:100
Lr @ react-dom.production.min.js:101
(anonymous) @ react-dom.production.min.js:113
Ie @ react-dom.production.min.js:292
(anonymous) @ react-dom.production.min.js:50
Mr @ react-dom.production.min.js:105
Jt @ react-dom.production.min.js:75
Yt @ react-dom.production.min.js:74
[...]

Issue #1781, opened 2024-03-11
All delete and update functions should have a Confirmation Dialog


ashu8912:
#1780 
Right now we have a few places like the Home closer view where if we go to delete a cluster we don't show a confirmation, We should investigate if we have other places we are doing the same and add a Confirmation Dialog to all of them

PR #1806, opened 2024-03-14
frontend: Fix deletion of cluster

knrt10:
There was no option to delete stateless clusters. Since they are also dynamic clusters and stored in frontend, the call to backend was useless. They had to be removed from indexDB.

Adds a confirm dialog to delete clusters

Fixes: #1780, #1781

### How to test

- Run frontend `make run-frontend`
- Run backend `make run-backend`
- Run app `npm run dev-only-app`

### Testing

- [x] Test for stateless clusters
- [x] Test for dynamically loaded clusters


frontend/src/components/App/Home/index.tsx
```diff
@@ -89,14 +94,43 @@ function ContextMenu({ cluster }: { cluster: Cluster }) {
...
+              setOpenConfirmDialog(true);
               handleMenuClose();
             }}
           >
             <ListItemText>{t('translation|Delete')}</ListItemText>
           </MenuItem>
         )}
       </Menu>
+
+      <Dialog
```
yolossn, line 106: I think here you can use the `ConfirmDialog` component to make the UI consistent 
https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/common/ConfirmDialog.tsx

frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -1618,6 +1619,14 @@ export async function setCluster(clusterReq: ClusterRequest) {
 }
 
 export async function deleteCluster(cluster: string) {
```
yolossn, line 1621: If we are removing the request to the backend then this function shouldn't be in the `apiProxy.ts` file.

knrt10: We are not removing request to the backend, it handles cases if it's a stateless cluster or not. If it's a stateless cluster it handles the indexDB part, if it's not, it makes request to the backend.

frontend/src/components/App/Home/index.tsx
knrt10: Thank you, make sense. Updated

illume:
note: The Build container and test job failed because of that flaky test that was fixed in main, so the job was restarted.

frontend/src/stateless/index.ts
```diff
@@ -419,13 +419,83 @@ export async function fetchStatelessClusterKubeConfigs(dispatch: any) {
...
+        // The cursor is used to find the kubeconfig by cluster name.
+        store.openCursor().onsuccess = function storeSuccess(event: Event) {
+          // delete the kubeconfig by cluster name
+          const successEvent = event as CursorSuccessEvent;
+          const cursor = successEvent.target.result;
+          if (cursor) {
+            const kubeconfigObject = cursor.value;
+            const kubeconfig = kubeconfigObject.kubeconfig;
+
+            const parsedKubeconfig = jsyaml.load(atob(kubeconfig)) as KubeconfigObject;
```
illume, line 457: This could fail, so it probably needs a try/catch/reject.

illume (changes requested):
Nice one. I left a couple of notes.

frontend/src/stateless/index.ts
knrt10: This whole is inside try catch here https://github.com/headlamp-k8s/headlamp/pull/1806/files#diff-c420684a8dc177e0c1311ea7c11cfd598ec94e9563b1ab6c095f14bdf0d602d8R428. Do you need more specific try catch is needed?

illume (approved):
🎈 🎉 
thanks

yolossn (approved):

PR #1808, opened 2024-03-14
frontend: apiProxy: Fix some types and add review comments

illume:
A summary of things marked for fixing.

- Return types are missing in places.
    - Some types are "any".
- No docs on some functions and interfaces.
- Async is missing on some functions that need to be marked as so.
- Some of the users of the functions are not handling errors.


skoeva:
was there ever an issue open as a reminder to document the apiProxy functions/interfaces? we can add one if not, would be useful

joaquimrocha:
I think we had someone adding the docs to it but I cannot find that PR now. In any case, it'd be nice to have but I think we will be changing the apiProxy a lot after we had the tests in, so not sure if it's worth doing by the team.

skoeva:
another minor fix I noticed when testing apiProxy: lines 915 and 922 have a slight spelling error

'Unkown' can be changed to 'Unknown'

joaquimrocha (approved):

PR #1809, opened 2024-03-14
frontend: e2e-tests: Improve code formatting

illume:
- Code formatting for e2e-tests folder will also be format checked on CI. 
- Some of the e2e test code was also formatted.
- Use new cached option for prettier, format goes from 14seconds to 2 seconds.

### `npm run format` Before

```
real    0m8.543s
user    0m14.382s
sys     0m0.842s
```

### `npm run format` After

```
real    0m1.951s
user    0m2.218s
sys     0m0.356s
```


knrt10 (approved):
💯

PR #1856, opened 2024-03-25
headlamp-plugin: Do not exit process when watching for changes

### LINKED ISSUE #1855: headlamp-plugin command quits immediately after build, instead of watching for files
On Mac the headlamp-plugin command (run from `npm run start` within a plugin) no longer hangs watching for files. Instead it just bundles the plugin and quits. I have experienced this in WSL at some point too but on Mac it seems to be consistent.

joaquimrocha:
Due to a recent change, after webpack compiled or had an error
compiling a plugin's code, the process was being exited, meaning that
the functionality of watching for changes was broken.

fixes #1855 

## How to test
1. Set up a version of headlamp-plugin with these changes (create a package using `npm pack`);
2. Go to an example plugin and install that package from the tarball (`npm i /path/to/your/kinvolk-headlamp-...tgz`);
3. Build the plugin for changes: `npm run start` -> Realize that the command does not quit after copying the files to the plugins folder, and instead it keeps watching for changes in the plugin files.

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -292,19 +292,19 @@ function start() {
   config.mode = 'development';
   process.env['BABEL_ENV'] = 'development';
   copyToPluginsFolder(config);
+  let exitCode = 0;
   webpack(config, (err, stats) => {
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks!

PR #1873, opened 2024-03-26
frontend cluster/Overview: Show only warnings by default

joaquimrocha:
By default we should show only warnings. This wrong default has been
around since we added the related switch.


illume:
test snapshots need updating

illume (approved):
👍 thanks

PR #1882, opened 2024-03-27
frontend: Fix success state when adding kubeconfig

### LINKED ISSUE #1831: Loading kubeconfig seems broken
_Originally posted by @gaalw in https://github.com/headlamp-k8s/headlamp/issues/389#issuecomment-2008874569_

Steps to reproduce.

```brew install --cask headlamp
Running `brew update --auto-update`...
==> Auto-updated Homebrew!
Updated 5 taps (hashicorp/tap, weaveworks/tap, minio/stable, homebrew/core and homebrew/cask).
==> New Formulae
cotp                 git-split-diffs      hashicorp/tap/hcp    tartufo              whisperkit-cli       yo
==> New Casks
nperf                           posture-pal                     proton-mail                     steinberg-activation-manager

You have 79 outdated formulae and 3 outdated casks installed.

==> Downloading https://github.com/kinvolk/headlamp/releases/download/v0.23.0/Headlamp-0.23.0-mac-x64.dmg
==> Downloading from https://objects.githubusercontent.com/github-production-release-asset-2e65be/220546758/eb639fc6-b9ab-4da6-bb
########################################################################################################################## 100.0%
==> Installing Cask headlamp
==> Moving App 'Headlamp.app' to '/Applications/Headlamp.app'
```

Run Headlamp.

Screenshot 2024-03-20 at 08 02 06

The clusters from `~/.kube/config` are loaded properly.

If I click on "Load kubeconfig" button:

Screenshot 2024-03-20 at 08 02 47

Screenshot 2024-03-20 at 08 03 04

Screenshot 2024-03-20 at 08 03 11

So the function of loading external kubeconfig is not working at all. No idea why.

joaquimrocha:
The way we were adding clusters previously (before the stateless
clusters) meant we received the configuration when we added a
cluster. After stateless clusters, that same call to add a cluster
in the frontend doesn't return a response. This triggered an
exception which meant the UI assumed there was an issue with setting
up the cluster but that was false.

fixes #1831

### DIFF
--- a/frontend/src/components/cluster/KubeConfigLoader.tsx
+++ b/frontend/src/components/cluster/KubeConfigLoader.tsx
@@ -11,7 +11,7 @@ import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { setCluster } from '../../lib/k8s/apiProxy';
-import { setConfig } from '../../redux/configSlice';
+import { setStatelessConfig } from '../../redux/configSlice';
 import { DialogTitle } from '../common/Dialog';
 import Loader from '../common/Loader';
 import { ClusterDialog } from './Chooser';
@@ -145,10 +145,10 @@ function KubeConfigLoader() {
         const selectedClusterConfig = configWithSelectedClusters(fileContent, selectedClusters);
         setCluster({ kubeconfig: btoa(yaml.dump(selectedClusterConfig)) })
           .then(res => {
-            if (res.clusters.length > 0) {
-              dispatch(setConfig(res));
-              setState(Step.Success);
+            if (res?.clusters?.length > 0) {
+              dispatch(setStatelessConfig(res));
             }
+            setState(Step.Success);
           })
           .catch(e => {
             console.debug('Error setting up clusters from kubeconfig:', e);
--- a/docker-extension/docker-compose.yml
+++ b/docker-extension/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 services:
   headlamp:
     image: ghcr.io/headlamp-k8s/headlamp:v0.20.0
-    command: ["--kubeconfig","/headlamp/config/config", "--port","64446"]
+    command: ["--kubeconfig","/headlamp/config/config", "--port","64446", "--enable-dynamic-clusters"]
     restart: unless-stopped
     volumes:
       - ~/.kube/:/headlamp/config:ro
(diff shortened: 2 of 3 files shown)


frontend/src/components/cluster/KubeConfigLoader.tsx
```diff
@@ -145,10 +145,12 @@ function KubeConfigLoader() {
...
         setCluster({ kubeconfig: btoa(yaml.dump(selectedClusterConfig)) })
           .then(res => {
-            if (res.clusters.length > 0) {
+            // @todo: The setCluster may return a response or not. This is not consistent, so we
+            // should make sure we always either have a response or we don't.
+            if (res?.clusters?.length > 0) {
               dispatch(setConfig(res));
-              setState(Step.Success);
             }
+            setState(Step.Success);
```
knrt10, line 153: Since kubeconfig is sent in the request. We are considering it as stateless cluster. So we should use `setStatelessConfig` here. Tested it

knrt10 (commented):
Small suggestion. Also instead of another PR, we can have this in `apiProxy.ts`. This should fix the backend todo part as well

```ts
  const kubeconfig = clusterReq.kubeconfig;

  if (kubeconfig) {
    await storeStatelessClusterKubeconfig(kubeconfig);
    // We just send parsed kubeconfig from the backend to the frontend.
    return request(
    '/parseKubeConfig',
      {
        method: 'POST',
        body: JSON.stringify(clusterReq),
        headers: {
          ...JSON_HEADERS,
        },
      },
      false,
      false
    )
  }

  return request(
    '/cluster',
    {
      method: 'POST',
      body: JSON.stringify(clusterReq),
      headers: {
        ...JSON_HEADERS,
        ...getHeadlampAPIHeaders(),
      },
    },
    false,
    false
  );
}
```

frontend/src/components/cluster/KubeConfigLoader.tsx
knrt10: Made the push

yolossn:
Tested and it works, I think the load kubeconfig from file option is breaking in the docker-desktop extension too. For this we need to add `-enable-dynamic-clusters` flag to the backend container in headlamp docker extension [Here](https://github.com/headlamp-k8s/headlamp/blob/main/docker-extension/docker-compose.yml#L5), only then the `/parseKubeConfig` endpoint will be enabled. Please make sure this is fixed for docker-desktop extension too.

knrt10:
Thank you @yolossn. Good catch. Let me fix it too

yolossn (approved):

PR #1884, opened 2024-03-28
backend: Fix plugin paths when base-url is set

yolossn:
this patch removes the base-url that was appended
to the /plugins response as the base path is
added automatically by the frontend


illume (approved):
👍 thanks

Issue #1746, opened 2024-02-27
Replace minikube with k3d when testing in CI


joaquimrocha:
We use minikube for testing changes in-cluster in CI, but minikube can take some time, which means PR checks take longer than they should.

So we should replace minikube by a lighter k8s flavor (except in places where we're actually deliberately testing for minikube functionality). This page shows that to be k3d: https://minikube.sigs.k8s.io/docs/benchmarks/timetok8s/v1.32.0/

joaquimrocha:
Following the discussion in the team, please check how much the _minikube_ step alone is taking in order to verify whether this issue is required.

skoeva:
> Following the discussion in the team, please check how much the _minikube_ step alone is taking in order to verify whether this issue is required.

See the "Build Container and test" runs in the Actions tab, using minikube the "start cluster" step takes around 1m30s with minikube, while the runs in CI using k3d are 1min quicker:
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/d374be2a-0ffb-40af-9b91-b441a7729b8b)
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/c5f9da1f-f114-4800-92f6-720f7240c5dd)

PR #1890, opened 2024-04-03
github: Replace minikube with kind in CI

skoeva:
This change replaces the use of minikube in the `Build Container and test / build discover and deploy` workflow with the use of kind. We estimate 60% decrease in run time on a cold run for the "Start Cluster" step of 1 minute (from ~1m50s to ~45s). Even with a cached image for the minikube run, we estimate a 25% decrease in run time (from ~1m to ~45s). This should speed up CI workflows for future changes.

Fixes: #1746 

### Testing
- [x] Run `Build Container` workflow locally and ensure tests pass
- [x] Build Headlamp images in a local kind cluster
- [x] Apply deployment manifests and configure Headlamp credentials
- [x] Run e2e playwright tests against the local Headlamp instance

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -35,7 +35,9 @@ jobs:
       with:
         node-version: 20.x
     - name: Start cluster
-      uses: medyagh/setup-minikube@d8c0eb871f6f455542491d86a574477bd3894533 # latest
+      uses: helm/kind-action@0025e74a8c7512023d06dc019c617aa3cf561fde # v1.0.0
+      with:
+          cluster_name: test
       # now you can run kubectl to see the pods in the cluster
     - name: Try the cluster!
       run: kubectl get pods -A
@@ -50,7 +52,6 @@ jobs:
       if: steps.cache-image-restore.outputs.cache-hit == 'true'
       run: |
         export SHELL=/bin/bash
-        eval $(minikube -p minikube docker-env)
         docker load -i ~/image-cache/headlamp-plugins-test.tar
         docker load -i ~/image-cache/headlamp.tar
     - name: Make a .plugins folder for testing later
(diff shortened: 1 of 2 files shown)


.github/workflows/build-container.yml
```diff
@@ -105,28 +109,41 @@ jobs:
...
         echo $SERVICE_URL
         curl -L $SERVICE_URL | grep -q "Headlamp: Kubernetes Web UI"        
-        echo "------------------Getting HEADLAMP_TOKEN------------------"
+        echo "--------------------------------Getting HEADLAMP_TOKEN--------------------------------"
         kubectl create serviceaccount headlamp-admin --namespace kube-system
         kubectl create clusterrolebinding headlamp-admin --serviceaccount=kube-system:headlamp-admin --clusterrole=cluster-admin
         export HEADLAMP_TOKEN=$(kubectl create token headlamp-admin --duration 24h -n kube-system)
-        echo "------------------Running playwright e2e tests------------------"
+        echo "---------------------------------Certificate handling---------------------------------"
+        export KUBECONFIG=$HOME/.kube/config
```
illume, line 132: I don't understand this really... Are there any documentation links for this you know of?

e2e-tests/kubernetes-headlamp-ci.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: kube-system
 spec:
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
```
illume, line 9: What is this change for? I don't see it mentioned in the git message.

illume:
Nice work, I see it's faster with starting the cluster. I also see some things are cached.

### kind
image


### minikube

image

.github/workflows/build-container.yml
```diff
@@ -70,16 +71,19 @@ jobs:
...
         export SHELL=/bin/bash
-        eval $(minikube -p minikube docker-env)
         DOCKER_IMAGE_VERSION=latest make image
         DOCKER_IMAGE_VERSION=latest DOCKER_PLUGINS_IMAGE_NAME=headlamp-plugins-test make build-plugins-container
         echo -n "verifying images:"
         docker images
+    - name: Import images to kind
+      run: |
+        export SHELL=/bin/bash
+        kind load docker-image ghcr.io/headlamp-k8s/headlamp-plugins-test:latest --name test
```
illume, line 81: I wonder if you tested the case where a person is only making changes to the e2e tests. In this situation the image should be cached, and not rebuilt.

illume (commented):
🎉 

Awesome.

I left a few questions.

e2e-tests/kubernetes-headlamp-ci.yaml
skoeva: I noticed locally that this fixed an issue where the headlamp deployment timed out, otherwise CI would fail here

.github/workflows/build-container.yml
skoeva: The issue is that the kind kubeconfig does not specify local paths for the certificates, and our e2e test setup tries to read these paths. I copied the certificate handling over from the k3d setup but I suppose there should be a simpler way to authenticate, I'll continue testing

.github/workflows/build-container.yml
skoeva: Glanced over this, thanks for the catch

illume (approved):
🎉🎈 Thanks :)

PR #1902, opened 2024-04-10
github: Switch azure login to OIDC connect

joaquimrocha:
From @jepio:
Github issues every job a token, and a service principal in Azure can be configured to trust that token. This is called "federated credentials" or OpenID Connect. Switch to using the azure/login action and enable write access on the id-token for the jobs that need it.

Preserve the default `contents: read` permission and move the tenant id to a secret.


illume (approved):
👍

illume:
The build container failure is unrelated to this PR and is fixed in main.

PR #1908, opened 2024-04-15
app: Fix for MacOS auth plugins

### LINKED ISSUE #1885: external auth (exec) not working (azure/kubelogin and aws-iam-authenticator)
I tried connecting to our prod clusters using "azure/kubelogin" and aws-iam-authenticator.

None of these Auth Methods worked.

Is there an eta we can expect exec auth to work? We want to switch from the closed source software lens to headlamp, but this is currently blocking.

illume:
On at least MacOS, GUI apps do not take on the shell paths.
However they are needed for kubectl auth plugins,
and perhaps other binaries like minikube

Fixes #1885

Signed-off-by: René Dudfield <renesd@gmail.com>


illume:
The frontend failure is unrelated to this PR, and is fixed in https://github.com/headlamp-k8s/headlamp/pull/1910

app/electron/main.ts
```diff
@@ -6,6 +6,7 @@ import { app, BrowserWindow, dialog, ipcMain, Menu, MenuItem, screen, shell } fr
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import log from 'electron-log';
 import find_process from 'find-process';
+import fixPath from 'fix-path';
```
joaquimrocha, line 9: I was checking and this module + the dependencies from the same author and they have not been updated for 3 years.
I wonder if we should send them a patch for updating the dependencies, or just try to bring in the relevant parts of the module's logic.

illume: I saw that suggestion about the dependencies, but thought the module was 'done'. But yeah, good call out.

The dependencies could be reduced to none. We don't need to strip ansi just to get PATH and SHELL, and in 2024 executing processes is fine with node/electron APIs.

illume:
@ashu8912 to recap where we are up to, and suggestions on next steps...

- [x] the azure/kubelogin isn't working in headlamp (in Headlamp started by terminal, or in Headlamp started by finder). So this is not a good thing to test that this PR is working.
- [ ] aws-iam-authenticator hasn't been tested yet (@ashu8912 can you do this one?)
- [ ] I will outline some steps to test more clearly with the minikube/az binaries as well (which might be on a terminal path but not in the PATH for the user running the app from finder) @illume to do
- [ ] there should also be a check to see if the app is NOT being run by the terminal, and in that case there is no need to modify the PATH. @illume to do
- [ ] Add a logging of the PATH variable, which would allow us to check that it is equivalent to the terminal one easily

illume:
Hi @ashu8912 I made changes to the PR.

Also outlined some much simpler testing steps, to just check the PATH from the terminal is being added.

illume:
@joaquimrocha I've updated the test instructions and added a screen shot showing it works to the PR description.

PR #1912, opened 2024-04-15
frontend: Fix warning shown in npm start

illume:
This gets rid of the yellow warning message that is shown when running `npm start` in the frontend.

### How to test

Run `npm start` and there should be no warning text block about a babel package.


skoeva (approved):
thanks for taking care of this, looks good!

PR #1913, opened 2024-04-16
docs: Fix the link in building plugin page

sniok:
The original intent of that link was to point to the "run the code" section of the development page, but index.md wasn't resolving and currently it shows 404. Also extra # is unnecessary. Verified the fix locally


joaquimrocha (approved):
Thanks!

PR #1915, opened 2024-04-17
github: Fix permissions so they are minimal

illume:
See 
- https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
- https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

Note: `contents: read` at the top turns off all other permissions. It's best practice to do that and specify in each job any extra permissions they have. This is so that it's easier to tell which permission is for which job, and also so that the other permissions are turned off or not.

### DIFF
--- a/.github/workflows/container-publish.yml
+++ b/.github/workflows/container-publish.yml
@@ -13,6 +13,9 @@ on:
         description: 'Container tags (separated by a comma)'
         type: string
 
+permissions:
+  contents: read
+
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: headlamp-k8s/headlamp
@@ -21,8 +24,7 @@ jobs:
     name: Test building container image
     runs-on: ubuntu-latest
     permissions:
-      contents: read
-      packages: write
+      packages: write # needed for publishing the container image
     steps:
     - name: Check out the repo
       uses: actions/checkout@v4
@@ -71,7 +73,7 @@ jobs:
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
+        password: ${{ github.token }}
     - name: Build and push Docker image
       uses: docker/build-push-action@v4
       with:
--- a/.github/workflows/docker-extension-release.yml
+++ b/.github/workflows/docker-extension-release.yml
@@ -6,6 +6,9 @@ on:
       - completed
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 env:
   REGISTRY: docker.io
   IMAGE_NAME: headlamp/headlamp-docker-extension
@@ -13,6 +16,8 @@ jobs:
   build_and_push_docker_extension:
     name: Build docker extension
     runs-on: ubuntu-latest
+    permissions:
+      secrets: read # needed to fetch docker hub creds
     steps:
       - name: Login to DockerHub
         uses: docker/login-action@v1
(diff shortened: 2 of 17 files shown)


.github/workflows/container-publish.yml
```diff
@@ -21,7 +24,6 @@ jobs:
     name: Test building container image
     runs-on: ubuntu-latest
     permissions:
-      contents: read
       packages: write
```
yolossn, line 27:
```suggestion
      packages: write  # needed for publishing the container image.
```

.github/workflows/docker-extension-release.yml
```diff
@@ -6,13 +6,18 @@ on:
...
+
 env:
   REGISTRY: docker.io
   IMAGE_NAME: headlamp/headlamp-docker-extension
 jobs:
   build_and_push_docker_extension:
     name: Build docker extension
     runs-on: ubuntu-latest
+    permissions:
+      secrets: read
```
yolossn, line 20:
```suggestion
      secrets: read  # needed to fetch docker hub creds
```

.github/workflows/draft-release.yml
```diff
@@ -8,17 +8,21 @@ on:
...
         default: 'X.Y.Z'
 
+permissions:
+  contents: read
+
 jobs:
   build:
     permissions:
       contents: write
+      secrets: read
```
yolossn, line 18: since we are using `github.token` permission to read secrets isn't required.

.github/workflows/helm-chart-release.yml
```diff
@@ -26,14 +32,12 @@ jobs:
...
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.2.1
+        uses: helm/chart-releaser-action@v1.6.0
         env:
-          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_TOKEN: "${{ github.token }}"
         with:
           config: .github/cr.yaml
-
+          mark_as_latest: false # only headlamp is set to latest
```
yolossn, line 43: good addition.

.github/workflows/trigger-flatpak-update.yml
```diff
@@ -7,10 +7,17 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
   trigger_flatpak_update:
     name: Trigger Flatpak Headlamp version update
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: secrets
```
yolossn, line 19:
```suggestion
      contents: read
      secrets: read
```

plugins/examples/pod-counter-old/.gitignore
yolossn, line 1: The changes related to renaming the pod-counter plugin folder name can be part of another PR.

yolossn (changes requested):

.github/workflows/draft-release.yml
illume: Thanks, fixed.

.github/workflows/trigger-flatpak-update.yml
illume: Thanks, fixed.

plugins/examples/pod-counter-old/.gitignore
illume: Thanks, fixed.

illume (commented):
Thanks @yolossn, I've applied your fixes.

skoeva (commented):
the github context for the token seems more appropriate, overall looks good!

knrt10 (approved):
Looks good

joaquimrocha (approved):

Issue #1898, opened 2024-04-09
Helm chart - give an option to pass additional command line arguments


gecube:
Hi!

As an user I want to pass an additional command line arguments to the headlamp deployment. Now there is no such an option.

Particularly I am interested in `-insecure-ssl` command line option for debug purposes

Alternative: make all command line options to have an ENV variable alternative and allow user to pass them to the deployment.

knrt10:
working on it

gecube:
@knrt10 thanks for the update.

PR #1919, opened 2024-04-22
charts: Option to pass additional arg to container

knrt10:
User were unable to pass any additional arguments to the headlamp deployment, this adds enhancement for that.

Fixes: #1898

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -91,6 +91,9 @@ spec:
             {{- with .Values.config.baseURL }}
             - "-base-url={{ . }}"
             {{- end }}
+            {{- with .Values.config.extraArgs }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
           ports:
             - name: http
               containerPort: 4466
--- a/charts/headlamp/tests/test_cases/extra-args.yaml
+++ b/charts/headlamp/tests/test_cases/extra-args.yaml
@@ -1,0 +1,3 @@
+config:
+  extraArgs:
+  - -insecure-ssl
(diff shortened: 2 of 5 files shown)


charts/headlamp/templates/deployment.yaml
```diff
@@ -91,6 +91,9 @@ spec:
             {{- with .Values.config.baseURL }}
             - "-base-url={{ . }}"
             {{- end }}
+            {{- range .Values.config.extraArgs }}
+            - "{{ . }}"
```
illume, line 95: What about an extra space here or at the end of the base-url part? 

Otherwise it could end up like `"-base-url=example-otherarg"` ?

knrt10: Adding extra space or keeping it like this, did not effect it. I tried it but the output for the helm template was as follows

```yaml
args:
    - "-in-cluster"
    - "-base-url=/example"
    - "-insecure-ssl"

```

joaquimrocha (commented):
I think we should add some template testing to avoid mistakes like the one I mentioned in my previous case.

knrt10:
@joaquimrocha I think adding template testing is in the scope of a different PR. We don't have it setup currently and it requires complete setup of the helm template testing. I have created a new issue for the same #1954. I have also created a PR for the same

joaquimrocha:
@knrt10  Now that we have the template testing, we could add the test for the extraArgs here, right?

charts/headlamp/tests/test_cases/extra-args.yaml
```diff
@@ -0,0 +1,3 @@
+config:
+  extraArgs:
+  - insecure-ssl
```
joaquimrocha, line 3: I know this is just testing the formatting of the template, but the option should start with `-`.

joaquimrocha (changes requested):

joaquimrocha (approved):

Issue #1897, opened 2024-04-09
Illogical behaviour with oidc


gecube:
Good day Sirs,

I am experimenting with your helm chart extensively.
And found the next thing. I can set up oidc settings with the `config.oidc.secret.create: false` value. 
It leads to broken deployment as it expects the secret anyway. It is opaque behaviour as in fact as user I want to have three options:

- put oidc settings and inject them like ENV variables directly from deployment (like for debug purposes)
- use external secret precreated by means of any possible solution (external-secrets operator, vault, I created the secret in k8s myself etc.) and give me an option to select this secret name
- create the secret inside of helm chart and pass it to the deployment with the option to reload deployment in case of changing the secret (add sha annotation in labels for the deployment)
- make the options exclusive to each other. Now I can set up oidc settings and not create the secret, which does not make sense...

I think good example could be `bitnami` charts for redis/mongo/postgresql, where it is obvious what to put into `values.yaml`

knrt10:
Thank you for the detailed issue @gecube. Fixing this

PR #1920, opened 2024-04-22
charts: Provide 3 options for oidc configuration

knrt10:
Now users have 3 different way they can set oidc configuration

- Directly set values of respective config.oidc.clientID and others which inject them into ENV variable, to be used by args.
- Use external i.e. already created secret with the same keys as args.
- Use config.oidc.secret.create functionality to create secret and have them dynamically loaded into the headlamp deployment.

Fixes: #1897

### Scenario 1
This is when user just want to use ENV variables directly. 

- Set values of `config.oidc.`  `clientID, clientSecret, issuerURL, scopes` in `values.yaml` and set `config.oidc.secret` as `false`
- Run `helm template charts/headlamp`

### Scenario 2
This is when user wants to use external secret created by them

- Create a new secret with above config.oidc values. For reference check `template/secret.yaml`. Then create it in your k8s cluster using kubectl command.
- Set `config.oidc.secret` as `false` and set `config.oidc.externalSecret` as `true` and also the name of secret.
- Run `helm template charts/headlamp`

### Scenario 3
This is when user want to use the secret created by headlamp

- Set values of `config.oidc.`  `clientID, clientSecret, issuerURL, scopes` in `values.yaml` and set `config.oidc.secret` as `true`
- Run `helm template charts/headlamp`

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -1,3 +1,28 @@
+{{- $oidc := .Values.config.oidc }}
+{{- $env := .Values.env }}
+
+{{- $clientID := "" }}
+{{- $clientSecret := "" }}
+{{- $issuerURL := "" }}
+{{- $scopes := "" }}
+
+# This block of code is used to extract the values from the env.
+# This is done to check if the values are non-empty and if they are, they are used in the deployment.yaml.
+{{- range $env }}
+  {{- if eq .name "OIDC_CLIENT_ID" }}
+    {{- $clientID = .value }}
+  {{- end }}
+  {{- if eq .name "OIDC_CLIENT_SECRET" }}
+    {{- $clientSecret = .value }}
+  {{- end }}
+  {{- if eq .name "OIDC_ISSUER_URL" }}
+    {{- $issuerURL = .value }}
+  {{- end }}
+  {{- if eq .name "OIDC_SCOPES" }}
+    {{- $scopes = .value }}
+  {{- end }}
+{{- end }}
+
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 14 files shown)


illume (commented):
Nice one.

Could you please mention what testing was done in the PR description?

I see documentation inline, but perhaps this needs to be documented somewhere in docs/ as well?

Did you look at how the bitnami charts for redis/mongo/postgresql are done? 
I see they use examples and document paramaters in detail https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L1635
Also they have a values.schema.json https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.schema.json

illume:
I added a separate issue for adding a json schema to the chart https://github.com/headlamp-k8s/headlamp/issues/1926 if that is out of scope for this PR.

charts/headlamp/templates/deployment.yaml
```diff
@@ -37,35 +37,57 @@ spec:
...
                   key: scopes
             {{- end }}
+            {{- else if .Values.config.oidc.externalSecret.enabled }}
+            # Check if externalSecret is enabled
+            envFrom:
+              - secretRef:
+                  name: {{ .Values.config.oidc.externalSecret.name }}
+            {{- else }}
+            {{- if .Values.config.oidc.clientID }}
+            - name: OIDC_CLIENT_ID
```
joaquimrocha, line 76: These are not aligned AFAICS.

charts/headlamp/templates/deployment.yaml
```diff
@@ -76,16 +98,27 @@ spec:
             {{- with .Values.config.pluginsDir}}
             - "-plugins-dir={{ . }}"
             {{- end }}
-            {{- if or .Values.config.oidc.clientID (not .Values.config.oidc.secret.create) }}
+            {{- if or (and .Values.config.oidc.clientID (not .Values.config.oidc.secret.create)) (and .Values.config.oidc.secret.create (ne .Values.config.oidc.clientID "")) }}
```
joaquimrocha, line 101: Is it my impression or does this just translates to `{{- if ne .Values.config.oidc.clientID "" -}}`? What am I missing?
(same for the similar logic below)

joaquimrocha (changes requested):
Added a comment about some of the logic used. Can you also add the template tests for this?

charts/headlamp/templates/deployment.yaml
knrt10: While it's true that both expressions check whether the `clientID` value is not empty, the original expression has additional conditions that may affect the outcome.

This expression contains two `and` conditions combined with an `or` condition:

- The first `and` condition:

Checks if `.Values.config.oidc.clientID` is truthy (i.e., not empty or false).
Checks if `.Values.config.oidc.secret.create` is falsy (i.e., evaluates to false or an empty value).

- The second and condition:

Checks if `.Values.config.oidc.secret.create` is truthy (i.e., not empty or false).
Checks if `.Values.config.oidc.clientID` is not equal to an empty string ("").

The `or` condition combines the two `and` conditions, meaning that the if block will execute if either of the `and` conditions evaluates to true.

The expression `{{- if ne .Values.config.oidc.clientID "" -}}` only checks whether `.Values.config.oidc.clientID` is not equal to an empty string (""). It does not consider the `secret.create` condition at all.

charts/headlamp/tests/test_cases/oidc-create-secret.yaml
```diff
@@ -0,0 +1,9 @@
+config:
```
illume, line 1: It would be good if each of these test cases could explain what they are testing. Not sure the best way exactly...

Maybe with a comment in the test cases file? "It should ..." "This is testing that..."

illume (approved):
👍 thanks

joaquimrocha (approved):

Issue #1641, opened 2024-01-08
Enable Pod filter by node name and pod status


lijianzhi01:
In Pod list page, we want to filter pods by node name or pod status.

achauve:
@joaquimrocha I can't find how to add new filters using a plugin. Is it possible or is it planned? or the only solution is to add the feature in headlamp directly?

joaquimrocha:
Hi @achauve , you are rights we need to add more advanced filters. We have a plan for doing that before #1646 (this one will allow filter by any column) and be configurable by each table. Maybe in 3 weeks.

joaquimrocha:
I believe this is already done after we added the new table. Please confirm @lijianzhi01 .

Issue #1640, opened 2024-01-08
Ability to delete multiple Pods


lijianzhi01:
In Pod list page, we want to delete multiple pods at the same time.

lijianzhi01:
This is description from partner:

Request to add the ability to restart (delete all pods at once) the deployment with a click of a button (instead of deleting pod by pod)

joaquimrocha:
This should be done when we get the new table as described in #1646 .

Issue #1006, opened 2023-03-10
Add context menu to different lists


joaquimrocha:
As reported by some users, it may be useful to have a context menu for each list row, in order to display options that the user would otherwise have to go into the respective object's details in order to execute.

An example would be running the header actions of resources without having to open the resources' details views. Due to our philosophy of "not showing controls for actions that are not allowed" (we don't show delete/edit unless those options are allowed for the user), this is complicated/costly to get right as the population of that context menu would depend on verifying with the API server that the user is allowed to do those.
Maybe a solution for this would be to perform that check only when the user clicks the context menu: this would add a small delay before all the items are shown in the menu, but it's likely the best solution to avoid multiple requests per row...

illume:
Azure portal does actions with a *hover*... 
which is a bit clumsy imho.

Also there is a delay before some actions are activated, which again feels a bit clumsy.

![image](https://github.com/headlamp-k8s/headlamp/assets/9541/8ca4edfe-d199-4ae2-9a23-0b9a82383457)

---

This extension to the datatables provides a *context menu* on one or many selected rows. Using the contextmenu action in the browser (right click, or pressing the context menu keyboard button for example).
https://github.com/torrobinson/datatables-contextual-actions

![image](https://github.com/headlamp-k8s/headlamp/assets/9541/0c06a5f3-c86a-485b-b3c9-84742acc2d43)

Note the actions on multiple rows. Is this something we want to do? Would be useful for delete at least.

---

Another option is to just list the actions in a final row. This makes them more discoverable than contextmenu/hover, at the expense of using more column-browser-width-space.

Something like this, but probably with icons instead of 
![image](https://github.com/headlamp-k8s/headlamp/assets/9541/72719099-0bbb-48a9-886f-33c0ee5f58d6)

illume:
Some notes from discussion with @ashu8912 


### How many actions usually?
- 3-4 action button items by default
- plugins can add action buttons

So, there can be any number of them.

### How many API requests?

- potentially thousands when there are 1000s of items
- Is there a way to batch request multiple permission checks into one API call?
- Could tie into virtualization of the table, where it only requests the ones shown (20 per page, 20 requests).
- It could check only when an action menu appears

### Actions that work on multiple items. 

- Delete and scale could work on many items.
- Edit would only work on one item.

### multiple row select?

What happens if 7 of 10 only can be deleted?
Show a dialog and ask them if they want to delete 7 of 10?

[...]

PR #1921, opened 2024-04-22
frontend: Add new Table component and use it in ResourceTable

### LINKED ISSUE #1646: Support advanced use-cases in our table
We want to support use-cases where the MUI Table (that we use to implement our SimpleTable) falls short.

One option is to use e.g. [Material React Table](https://www.material-react-table.com/).
Maybe we also take the chance and instead of reimplementing SimpleTable, we leave it be and create a new component with a less deceiving name 🙂 Maybe just... Table.

Here are the things we need in this new table:
- [ ] Easy context menus (unblocks #1006 )
- [ ] Add column-based filtering (unblocks #1641 )
- [ ] Batch operations on items / rows selection + run action (unblocks #1640 )

sniok:
We want more features in our tables. Stuff like sorting, filtering, selection, actions, etc. This is where the new `Table` component comes in and `SimpleTable` will remain simple.

material-react-table (MRT) was chosen because it uses the same ui framework, looks good, has all the feature we will need, MIT licensed, based on headless tanstack table and very customizable. 

`Table` component is mostly a wrapper around MRT with some sensible defaults and additional app specific behaviour (like storing page state in url). I kept most of the props as aliases to the MRT props so it can be extensible without introducing any plumbing. 

In the scope of this PR I also updated ResourceTable component to use the new table. The only change from the 'outside' is that each column needs to provide getter and renderer functions, first is needed for filtering and sorting, second one is for displaying. Getter needs to return a string and since most of the existing column getters were simple functions there wasn't many changes there. In those cases where existing getter returned JSX node it was renamed to render and a simple getter was added, representing plain value.

While not in the scope of this PR there are features that we can implement now using the new Table: row selection, actions column, multi-select for the filtering, selection actions (like deleting all selected Pods). 

Fixes #1646 #1641 and unblocks #1006 #1640

I'm still checking if I didn't break anything and ported the existing functionality properly but if you want you can leave any comments or suggestions


sniok:
This is how it looks like
Screenshot 2024-04-23 113138

Column filtering is enabled by default for all columns, can be disabled
Screenshot 2024-04-23 113155

Sorting by multiple column works out of the box
Screenshot 2024-04-23 113317

Drop down with values in a filter can be enabled by setting `filterVariant: 'multi-select'`
Screenshot 2024-04-23 113338

joaquimrocha (changes requested):
It's looking good. I left a few comments. The ResourceTable is something we export to plugins, so we need to make sure it's backwards compatible (including the exported `useThrottle` hook that you moved). I know this is complicated to do with so many fields, so if there's another suggestion (new component name that fits right), I'm open to that.
Otherwise, we do try to keep commits atomic, so if there are changes that should be in other isolated commits, let's move them there.

frontend/patches/@mui+material+5.14.17.patch
```diff
@@ -1,11 +0,0 @@
...
-index 56d3b7e..18bf936 100644
---- a/node_modules/@mui/material/ClickAwayListener/index.d.ts
-+++ b/node_modules/@mui/material/ClickAwayListener/index.d.ts
-@@ -1 +1,5 @@
--export { ClickAwayListener as default, type ClickAwayListenerProps, } from '@mui/base/ClickAwayListener';
-+import { ClickAwayListenerProps } from '@mui/base/ClickAwayListener';
-+
-+export { ClickAwayListener as default } from '@mui/base/ClickAwayListener';
-+
-+export type { ClickAwayListenerProps };
```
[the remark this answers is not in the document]
sniok: since adding material-react-table bumped @mui/material version this patch wasn't applying, I removed it since it's not necessary anymore, I think

frontend/src/components/common/Tooltip/TooltipLight.tsx
```diff
@@ -32,5 +34,5 @@ export default function TooltipLight(props: TooltipLightProps) {
     );
   }
 
-  return <StyledTooltip {...props} />;
+  return <StyledTooltip {...props} children={props.children as any} />;
```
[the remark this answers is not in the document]
sniok: added more context as to why casting is needed

sniok:
> It's looking good. I left a few comments. The ResourceTable is something we export to plugins, so we need to make sure it's backwards compatible (including the exported `useThrottle` hook that you moved). I know this is complicated to do with so many fields, so if there's another suggestion (new component name that fits right), I'm open to that. Otherwise, we do try to keep commits atomic, so if there are changes that should be in other isolated commits, let's move them there.

I've reverted some name changes to keep ResourceTable backwards compatible. I missed that it was a common component. The only partly incompatible thing is the `getter` change in the column definition. It is going to work without an issue, even if the JSX is returned. But it will display a type error during the build/development of the plugin in cases where `getter` returns JSX. This will help plugin authors update it to take advantage of the filtering/sorting functionality.

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,198 @@
+import { Paper } from '@mui/material';
+import { useTheme } from '@mui/styles';
+import {
+  MaterialReactTable,
+  MRT_ColumnDef,
+  MRT_TableOptions,
```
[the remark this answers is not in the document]
sniok: Renamed them

frontend/src/components/common/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -52,6 +53,7 @@ export default class ErrorBoundary extends Component<ErrorBoundaryProps, State>
     if (isValidElement(this.props.fallback)) {
       return this.props.fallback;
     }
-    return this.props.fallback ? Children.toArray([<this.props.fallback error={error} />]) : null;
+    const FallbackComponent = this.props.fallback as unknown as ComponentType<{ error: Error }>;
```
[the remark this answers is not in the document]
sniok: Made it a bit cleaner but still with casting. I couldn't figure out a way to go without it, I think newer version of typescript are a bit better with type narrowing for cases like this

frontend/src/components/common/Resource/ResourceListView.tsx
```diff
@@ -1,49 +1,36 @@
...
   const { title, children, headerProps, ...tableProps } = props;
-  const { t } = useTranslation();
-  const [columnChooserAnchorEl, setColumnChooserAnchorEl] = React.useState<null | HTMLElement>(
-    null
-  );
-  const withNamespaceFilter = (props as ResourceListViewWithResourceClassProps).resourceClass
-    ?.isNamespaced;
+  const withNamespaceFilter = (
+    (props as ResourceListViewWithResourceClassProps<KubeObject>).resourceClass as KubeObject
+  )?.isNamespaced;
```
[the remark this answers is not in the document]
sniok: This was casted already, even if we got rid of any casting here it still wouldn't really change much since KubeObject resolves to `any`. It can be improved though, I want to take a look at that after finishing with the table

frontend/src/components/common/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -44,6 +44,7 @@ export default class ErrorBoundary extends Component<ErrorBoundaryProps, State>
   render() {
     const { error } = this.state;
     if (error) {
+      console.error(error);
```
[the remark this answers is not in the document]
sniok: good point

joaquimrocha:
> But it will display a type error during the build/development of the plugin in cases where `getter` returns JSX. This will help plugin authors update it to take advantage of the filtering/sorting functionality.

This will be annoying to developers who need to rebuild but cannot yet update that part of the table (maybe they can be time-pressed).
So I would recommend either doing a type composition where the version with the JSX getter is deprecated. Or (probably easier and less confusing), just deprecating that property altogether and add a new `getValue` for getting the value instead.

sniok:
> This will be annoying to developers who need to rebuild but cannot yet update that part of the table (maybe they can be time-pressed). So I would recommend either doing a type composition where the version with the JSX getter is deprecated. Or (probably easier and less confusing), just deprecating that property altogether and add a new `getValue` for getting the value instead.

Good point, I went with the latter option. Catching places where JSX getter is used is unreliable, even in our codebase jsx is sometimes typed as any, so I went with introducing a new property getValue.

joaquimrocha (changes requested):
I have noticed a few issues with the table replacement:
**Column chooser considers a column that has just a context menu:**
Column chooser showing an empty column

**Long cell values are not ellipsizing or wrapping, just expanding the table:**
Table with a scrollbar due to values in a cell

**The age column has the column controls on the left (probably because we align this columns name to the right but what we really need is for the column to occupy the minimum width it needs)**
Age column with the column controls on the left

**The filters/search in the table shouldn't conflict with the ones in the section header:**
If **not super complicated** I think we should for now keep the section filter but make the search there be used for the table search (which is much better than what we have). The namespace filter should just show the section namespace filtering (we need to study how to move the namespace to a more global way).
Image showing all filters visible

If I find more cases, I will comment them.

sniok:
> I have noticed a few issues with the table replacement: **Column chooser considers a column that has just a context menu:** 

Fixed, columns without a label will not be shown in the column picker.

> **Long cell values are not ellipsizing or wrapping, just expanding the table:**

Fixed, wrapping now behaves like it used to
 
> **The age column has the column controls on the left (probably because we align this columns name to the right but what we really need is for the column to occupy the minimum width it needs)**

Age column is now narrow, I've kept the right alignment, I think it looks better that way

> **The filters/search in the table shouldn't conflict with the ones in the section header:** If **not super complicated** I think we should for now keep the section filter but make the search there be used for the table search (which is much better than what we have). The namespace filter should just show the section namespace filtering (we need to study how to move the namespace to a more global way).

Implemented the changes we discussed. The section header will now only have namespace switcher that is displayed without a toggle, search is now handled only inside the table.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -146,12 +136,14 @@ function EventsSection() {
...
         ],
       }}
+      defaultGlobalFilter={eventsFilter ?? undefined}
       data={events}
       errorMessage={Event.getErrorMessage(eventsError)}
       columns={[
         {
           label: t('Type'),
           getValue: event => event.involvedObject.kind,
+          size: 150,
```
joaquimrocha, line 146: Is it possible to determine non-fixed sizes, like "min-content" or similar, as it is possible to do it with an HTML grid?

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -829,11 +829,11 @@ export function ContainerInfo(props: ContainerInfoProps) {
 export interface OwnedPodsSectionProps {
   resource: KubeObjectInterface;
   hideColumns?: PodListProps['hideColumns'];
-  noSearch?: boolean;
```
joaquimrocha, line 832: We need to mark such changes as deprecated (and add a warning when they are used) rather than removing them. Otherwise we can break things, even if it's at a typing check level.

joaquimrocha (changes requested):
Thanks @sniok . Here are some comments:

translations' *_old.json files are not to be committed.

Age column:
* Even though it's nice for the values to be aligned to the right, the column shows the controls to the left of it:
  Screenshot showing the problem above

Nodes:
* Some columns look incorrectly spaces:
Screenshot showing the problem above

Namespace Header filter:
* Looks unaligned with the table, or maybe it's the table that is missing some padding...
  Screenshot showing the problem above

Column chooser popover:
* Has a "Hide All" option. Not sure if there's a quick way to remove this option, but the option doesn't make a lot of sense IMO.
Screenshot showing the problem above

Please do not remove any options that are exported to plugins, instead mark them as deprecated (we have some in the source code).

frontend/src/components/cluster/Overview.tsx
sniok: The table has two strategies 'semantic' which uses simple th,td,tr and 'grid' which confusingly uses flexbox for each row. I tried doing something like what was before 'minmax(1fr, 150px)' but neither option has a straightforward way to do it. I'll take another look into that

sniok: okay I can just completely override the layout and use grid for everything, just like SimpleTable. I've returned all the gridTemplate parameters in column so now it should behave the same way it used to.

frontend/src/components/common/Resource/Resource.tsx
sniok: kept the name

sniok:
> translations' *_old.json files are not to be committed. 

done

> Age column:
> * Even though it's nice for the values to be aligned to the right, the column shows the controls to the left of it:

now only body cell values aligned to the right, column itself now has min-content width

image


> Nodes:
> * Some columns look incorrectly spaces:

implemented grid layout for the table, it now behaves the same way it used to

> Namespace Header filter:
> * Looks unaligned with the table, or maybe it's the table that is missing some padding...

I didn't really touch paddings there, it was like this:
image

There are some negative margins and paddings that can be cleaned up, I could look into that but in a different pr probably

> Column chooser popover:
> * Has a "Hide All" option. Not sure if there's a quick way to remove this option, but the option doesn't make a lot of sense IMO.

I didn't find a quick way to customize it, but I can see a usecase for it. For example if you want only 2 columns out of 10, you hide all and then click only 2 times to show the ones you need

e2e-tests/package.json
```diff
@@ -9,9 +9,11 @@
   "license": "ISC",
   "devDependencies": {
     "@playwright/test": "^1.42.1",
-    "@types/node": "^20.12.2"
+    "@types/node": "^20.12.2",
```
illume, line 12: This change seems unrelated.

e2e-tests/tests/headlampPage.ts
```diff
@@ -104,7 +104,9 @@ export class HeadlampPage {
     const rowsDisplayed1 = await this.getRowsDisplayed();
 
     // Click on the next page button
-    const nextPageButton = this.page.getByTitle('Next page');
+    const nextPageButton = this.page.getByRole('button', {
+      name: 'Go to next page',
```
illume, line 109: Why was this text changed?

docs/development/api/modules/plugin_registry.md
```diff
@@ -639,7 +639,7 @@ registerResourceTableColumnsProcessor(function ageRemover({ id, columns }) {
   if (id === 'headlamp-pods') {
     columns.push({
       label: 'Init Containers',
-      getter: (pod: Pod) => {
+      getValue: (pod: Pod) => {
```
illume, line 642: This is a better name, but is getter also available for backwards compatibility?

illume (changes requested):
Nice work.

I left a few questions and notes, but haven't tested it yet.

Could you please let me know what testing you have done?

A nit on the commit messages. We use a capital letter for the verb. Rather than just "frontend:" it's good to put the component(s) that are affected. Probably the main components can be given as context, or if things are unrelated they can be moved to other commits.

frontend/src/redux/filterSlice.test.ts
```diff
@@ -19,17 +18,10 @@ describe('filterSlice', () => {
     expect(state.namespaces).toEqual(new Set(namespaces));
   });
 
-  it('should handle setSearchFilter', () => {
```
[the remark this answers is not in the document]
sniok: since the table itself has a search feature we don't really need the search filter in the header and in this slice, after discussion with @joaquimrocha we decided to just leave the namespace filter there

frontend/src/components/common/NamespacesAutocomplete.tsx
```diff
@@ -50,6 +58,7 @@ export function PureNamespacesAutocomplete({
       multiple
       id="namespaces-filter"
       autoComplete
+      openOnFocus
```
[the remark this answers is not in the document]
sniok: all of the column filters are handled inside the table, all except for the namespace filter that we have in the header, to have a bit more consistent experience you can click on the namespace column and then filter by namespace which will focus on the namespace filter in the header. but focusing programmatically on the input didn't open the dropdown with options, so I added it here

e2e-tests/package.json
sniok: while updating e2e tests I noticed that the prettier config wasn't loading in, so I made this change. it is a bit unrelated but nice to have

e2e-tests/tests/headlampPage.ts
sniok: new table uses a bit different wording in its' pagination

docs/development/api/modules/plugin_registry.md
sniok: yes getter is still available and works but marked as deprecated, there was an idea to catch only places where getter returns JSX but it wasn't reliable so a new field getValue was added instead. you can look at [ResourceTable.tsx:54](https://github.com/headlamp-k8s/headlamp/blob/81b5dd8063bf74e2344a3a1d686548c319665206/frontend/src/components/common/Resource/ResourceTable.tsx#L54) for more info

frontend/src/components/common/NamespacesAutocomplete.tsx
```diff
@@ -15,10 +15,18 @@ import Namespace from '../../lib/k8s/namespace';
...
+  setTimeout(() => {
+    inputRef.current?.focus?.();
+  }, 0);
+}
+
 export interface PureNamespacesAutocompleteProps {
   namespaceNames: string[];
   onChange: (event: React.ChangeEvent<{}>, newValue: string[]) => void;
-  filter: { namespaces: Set<string>; search: string };
+  filter: { namespaces: Set<string> };
```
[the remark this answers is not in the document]
sniok: same as above

frontend/patches/@mui+material+5.14.17.patch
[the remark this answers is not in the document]
illume: I would still like this with the mui upgrade to be in it's own commit. It makes debugging later easier if there's an issue in an unrelated part of the app.

frontend/package.json
```diff
@@ -86,7 +87,7 @@
     "spacetime": "^7.4.0",
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
-    "typescript": "4.4.3",
+    "typescript": "^4.5.5",
```
illume, line 90: Could you please separate this into it's own commit as well? For debugging ease later.

I assume the upgrade is necessary for material-react-table to work?

frontend/src/storybook.test.tsx
```diff
@@ -16,6 +16,8 @@ class OurConverter extends Stories2SnapsConverter {
   }
 }
 
+jest.mock('@mui/utils/useId', () => jest.fn().mockReturnValue('mui-test-id'));
```
illume, line 19: Can you please add this to its own commit?

frontend/src/components/common/SimpleTable.stories.tsx
```diff
@@ -223,31 +223,30 @@ ReflectInURLWithPrefix.args = {
 
 // filter Function
 
-type SimpleTableWithFilterProps = SimpleTableProps & { matchCriteria?: string[] };
```
illume, line 226: Is SimpleTable being deprecated? If so, it should be marked so.

If it's being kept for backwards compatibility it should be backwards compatible without an API change.

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,244 @@
+import { Paper } from '@mui/material';
```
illume, line 1: Can you please check if there is a performance regression or not? There's a few scripts in the load tests folder to generate some large tables which might be helpful.

frontend/src/components/common/Table.stories.tsx
```diff
@@ -0,0 +1,453 @@
+import { Box, Typography } from '@mui/material';
```
illume, line 1: These are nice.

I was wondering if you could do some accessibility testing on it? (if you have already, could please let me know what you covered?).

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,244 @@
...
+import { useTheme } from '@mui/styles';
+import {
+  MaterialReactTable,
```
illume, line 4: Can you please test this on mobile? (if you haven't already)

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,244 @@
...
+> & {
+  columns: TableColumn<RowItem>[];
+  emptyMessage?: string;
+  errorMessage?: any;
+  /** Whether to reflect the page/perPage properties in the URL.
+   * If assigned to a string, it will be the prefix for the page/perPage parameters.
+   * If true or '', it'll reflect the parameters without a prefix.
+   * By default, no parameters are reflected in the URL. */
+  reflectInURL?: string | boolean;
+  initialPage?: number;
```
illume, line 42: Can you please add documentation for these fields? It helps on hover, and in the storybook to understand what they are (for).

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * @see https://www.material-react-table.com/docs/api/table-options
+ */
+export type TableProps<RowItem extends Record<string, any>> = Omit<
+  MaterialTableOptions<RowItem>,
+  'columns'
+> & {
+  columns: TableColumn<RowItem>[];
+  emptyMessage?: string;
+  errorMessage?: any;
```
illume, line 36: Can this be a more specific type than any?

illume (commented):
looked through it some more, and left more notes...

e2e-tests/package.json
illume: Could you please put this into a separate commit/pr?

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
illume: It's not clear to me if it matters that there is an API change here.

docs/development/api/modules/plugin_registry.md
illume: Is getValue required now for ordering or something else? Or will getter work as before for everything?

frontend/package.json
sniok: yes its needed for the material-react-table, and I don't think it should be a separate commit, it won't compile on its own

frontend/src/storybook.test.tsx
sniok: same as the other one, without this change it wont pass CI

frontend/src/components/common/SimpleTable.stories.tsx
sniok: SimpleTable itself is not changed and not deprecated, it can still be used for simple cases. api is kept unchanged.
The reason there are changes here is because these stories are also coupled with the filterSlice Context

frontend/src/components/common/Table.tsx
sniok: I tried it with load tests and bunch of objects and it works okay. There are no noticeable slowdowns compared to the previous table. One perf issue that is present is when table is quickly updated, but it's not caused by the table itself and is same with the old table. I looked a bit into it and adding some throttling to the useList functions in KubeObjects would probably improve perf in this case but it's out of scope for this pr

frontend/src/components/common/Table.stories.tsx
sniok: Keyboard navigation works fine, there are aria-labels set on everything important. One interesting thing about table a11y is how it works with screen readers. When screen reader is enabled it provides new ways to navigate table, like arrow keys to go to cells within a table. Screen reader needs to detect semantic elements for table (table,th,tr,rd,etc). There's an [issue](https://github.com/w3c/csswg-drafts/issues/3040) with display: contents (that we use) that kind of breaks a11y tree but I've tested it with NVDA on Windows and it seems to be working great, can jump to table landmark and navigate the cells.

frontend/src/components/common/Table.tsx
sniok: didn't test it too thoroughly but it looks good to me, adapts to small widths and handles touch

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
joaquimrocha: @illume I discussed with @sniok that given that the table search was actually getting more powerful than the redux based search we have + avoided the problem we have now that we can only use one instance of the search per view, otherwise it will reflect in all tables.

docs/development/api/modules/plugin_registry.md
sniok: getValue is there for filtering, sorting the column and expects a plaintext value, render is a new property that controls rendering, getter will work as before but filtering or sorting won't work

joaquimrocha:
One thing I just remembered is that this needs to be overridden by plugins (look for `registerResourceTableColumnsProcessor` in example plugins). 
If you implement this @sniok , @yolossn has an interesting corner case he's facing with the table column processor + sorting.

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
illume: Ok. 

Does the API change matter here? I don't know if anyone is using this from plugins or not.

frontend/package.json
illume: I think there was a misunderstanding. I tried upgrading typescript to 4.5.5 on main branch by itself and it does build, and run. The tests and tsc pass too.

frontend/src/storybook.test.tsx
illume: It can be added as a separate earlier commit to the one it's in now. It would be good to put this in the same commit that upgrades the mui version.

frontend/src/components/common/Table.tsx
illume: Thanks for checking. It's good that there's no regression in performance.

frontend/src/components/common/SimpleTable.stories.tsx
illume: Ok, thanks.

@joaquimrocha is it your understanding that we not deprecate SimpleTable?

frontend/src/components/common/SimpleTable.stories.tsx
```diff
@@ -223,31 +223,30 @@ ReflectInURLWithPrefix.args = {
...
+type SimpleTableWithFilterProps = SimpleTableProps & { matchCriteria?: string[]; search?: string };
 function SimpleTableWithFilter(props: SimpleTableWithFilterProps) {
-  const { matchCriteria, ...otherProps } = props;
+  const { matchCriteria, search, ...otherProps } = props;
   const filterFunc = useFilterFunc(matchCriteria);
-  return <SimpleTable filterFunction={filterFunc} {...otherProps} />;
+  return <SimpleTable filterFunction={item => filterFunc(item, search)} {...otherProps} />;
```
illume, line 230: Is there an API change here with useFilterFunc?

frontend/src/components/common/Table.stories.tsx
illume: Nice one, thanks.

We also have react-axe making warnings in the console if there are any runtime issues it detects. I wonder if you've checked if any new warnings were printed in the console from axe?

Did you also check the lighthouse accessibility audit in the dev tools?

frontend/src/components/common/Table.tsx
illume: Great, thanks.

frontend/src/components/common/Table.stories.tsx
sniok: good idea, just checked and neither axe nor lighthouse report any issues with the table

joaquimrocha:
@sniok I have pushed this style change just for tracking it. Today I will bring this to the design team and check whether we keep this design change or do something else. Let's wait for that before we merge the commit + update snaptshots.

I have found the following problems when looking further:
Node details:
* Hidden cell contents that have hover info icons
Screenshot showing the issue

Maybe if we enable the column resizing feature, it will prevent these issues?

joaquimrocha:
> * Hidden cell contents that have hover info icons

I noticed now that this table may not be the new one.

joaquimrocha (changes requested):
We need to remove the old.json translation files + we cannot add the snapshots to a fixup commit because when it's squashed to where it belongs, the snapshots may include conflicts. So usually we need to squash it and then regenerate the snapshots for each commit.

sniok:
namespace filter was reverted. went through each commit and regenerated snapshots

frontend/package.json
sniok: typescript bump is now a separate commit

e2e-tests/package.json
sniok: removed this from this pr

frontend/package.json
illume: Thanks

frontend/src/components/common/Table.tsx
joaquimrocha, line 1: We could add this into a Table/ folder using this same file but exporting everything from index.tsx as well. So the table and stories are more contained.

frontend/src/components/common/index.ts
```diff
@@ -48,3 +48,4 @@ export * from './ErrorPage';
 export * from './ConfirmButton';
 export { default as ConfirmButton } from './ConfirmButton';
 export * from './NamespacesAutocomplete';
+export * from './Table';
```
joaquimrocha, line 51: We need to export the default component as well from here if we export one.

frontend/src/components/common/Table.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  return [zeroIndexPage, setZeroIndexPage];
+}
+
+/**
+ * Table component based on the Material React Table
+ *
+ * @see https://www.material-react-table.com/docs
+ */
+export function Table<RowItem extends Record<string, any>>({
```
joaquimrocha, line 117: Are we not exporting this as default?

joaquimrocha (approved):
A couple of last minute things to check but otherwise looks good to me.

frontend/src/components/common/SimpleTable.stories.tsx
joaquimrocha: I think eventually we do want to deprecate the `SimpleTable` since we don't want to maintain two tables.
But my understanding was that we are not doing that in this PR because there are still places where `SimpleTable` is used, so this PR has been about introducing the new table + changing the `ResourceTable` implementation.
I am fine with merging this and moving on to deprecate the `SimpleTable`.

frontend/src/components/common/SimpleTable.stories.tsx
joaquimrocha: LGTM (I don't think there's an API break).

illume:
The gh interface is pretty bad with these big PRs. It's super slow, and I don't get notified when a convo is resolved. To help with this, can we please leave convos open until they are resolved by note opener. This way we don't have to search through 90 of them whilst my CPU fans are on 100%.

sniok:
Made Table exported as default and rebased with main branch

frontend/src/components/common/Table/Table.tsx
```diff
@@ -0,0 +1,288 @@
...
+      if (typeof it.gridTemplate === 'number') {
+        return `${it.gridTemplate}fr`;
+      }
+      return it.gridTemplate ?? '1fr';
+    })
+    .join(' ');
+
+  const table = useMaterialReactTable({
+    ...tableProps,
+    columns: tableColumns ?? [],
```
illume, line 173: I haven't tested this, but I don't see i18n/localization hooked up? Does anything need to be done for that?

https://www.material-react-table.com/docs/guides/localization#localization-(i18n)-guide

frontend/src/components/job/List.tsx
```diff
@@ -138,7 +145,12 @@ export function JobsListRenderer(props: JobsListRendererProps) {
         {
           id: 'images',
           label: t('Images'),
-          getter: job => {
+          getValue: job =>
+            job
+              .getContainers()
+              .map(c => c.image)
+              .join(''),
```
illume, line 152: Does this matter that it's `''` instead of `', '`?

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -39,25 +72,89 @@ export interface ResourceTableProps extends Omit<SimpleTableProps, 'columns'> {
   id?: string;
   /** Deny plugins to process this table's columns. */
   noProcessing?: boolean;
-  /** The anchor element for the column chooser popup. */
-  columnChooserAnchor?: HTMLElement | null;
   /** The callback for when the columns chooser is closed. */
   onColumnChooserClose?: () => void;
+  /** Column to be sorted */
+  defaultSortingColumn?: { id: string; desc: boolean };
+  defaultGlobalFilter?: string;
```
illume, line 79: These new fields could use documenting.

illume (changes requested):
Did another big chunk of review, and left some more notes. 
image

frontend/src/components/common/Resource/ResourceTableColumnChooser.tsx
```diff
@@ -8,8 +8,8 @@ import React from 'react';
 import { ResourceTableColumn } from './ResourceTable';
 
 interface ColumnsPopupProps {
-  columns: ResourceTableColumn[];
-  onToggleColumn: (cols: ResourceTableColumn[]) => void;
+  columns: ResourceTableColumn<unknown>[];
```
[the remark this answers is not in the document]
sniok: this component doesn't really care about what kind of columns are passed into it and `unknown` is a bit nicer than `any`. but now thinking about it again, the props should be generic. will fix

frontend/src/components/job/List.tsx
sniok: not really, even if you type `,` it still uses fuzzy search by default

frontend/src/components/common/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -44,14 +44,20 @@ export default class ErrorBoundary extends Component<ErrorBoundaryProps, State>
   render() {
     const { error } = this.state;
     if (error) {
+      console.error('ErrorBoundary:', error);
       store.dispatch(eventAction({ type: HeadlampEventType.ERROR_BOUNDARY, data: error }));
     }
     if (!error) {
       return this.props.children;
     }
+    if (!this.props.fallback) {
```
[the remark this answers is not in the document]
sniok: I'm going to do it in a separate PR. VSCode by default uses the bundled typescript version instead of the one from the workspace and in this case there's an error in the latest ts version but not in the 4.5.x. So it compiles fine but in IDE you see an error.

app/package.json
```diff
@@ -156,7 +156,7 @@
     "mime-types": "^2.1.35",
     "octokit": "^2.0.14",
     "tar": "^6.1.13",
-    "typescript": "^4.4.3"
+    "typescript": "^4.5.5"
```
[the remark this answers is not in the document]
sniok: done

frontend/package.json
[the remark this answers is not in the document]
sniok: good idea, done

frontend/src/components/common/Resource/ResourceTableColumnChooser.tsx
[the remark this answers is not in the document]
sniok: it's generic now

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -9,28 +11,59 @@ import { useTypedSelector } from '../../../redux/reducers/reducers';
...
+   * but you can provide your own sorting function here
+   *
+   * @default true
+   **/
+  sort?: boolean | ((a: RowItem, b: RowItem) => number);
+  /** Change how filtering behaves, by default it will just search the text value */
+  filterVariant?: TableColumn<any>['filterVariant'];
+  disableFiltering?: boolean;
+  /** Column width, if not set the column will grow to fit */
+  // size?: number;
```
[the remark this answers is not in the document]
sniok: fixed

frontend/src/components/common/Table/Table.tsx
sniok: added localization

frontend/patches/@mui+material+5.14.17.patch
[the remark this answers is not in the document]
sniok: whoops, fixed 

mui bump is now explicit and with patch removal in a separate commit
I was considering pinning version 5.15.16 for all mui packages but only "@mui/material" has that version, /styles and /system are .15 and .14, I'd prefer keeping their version the same in package.json so I used ^5.15.14 so all of them can resolve to compatible versions.

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
sniok: typecheck won't fail because {namespace, search} will still be assignable to {namespace}. and search wasn't even used in this component anyway, this is just a namespace selector.

illume (approved):
🎉🎈

PR #1925, opened 2024-04-24
github: frontend: Split steps into jobs, and cache npm deps

illume:
Signed-off-by: René Dudfield <renedudfield@microsoft.com>


knrt10 (approved):
Nice work.

Issue #1917, opened 2024-04-19
Desktop-App: Clicking on hostname in ingress opens URL in the app


micw:
I'm using the desktop app (installed on arch linux with aur package). When I klick on the domain name in any ingress, it open's in the system browser as well as in the desktop app. I need to restart the desktop app after the click to get back to headlamp UI.

joaquimrocha:
This used to happen somewhere else as well and I thought we had fixed it completely but apparently not. Thanks for reporting. We will check it out.

PR #1938, opened 2024-04-29
app: Fix external url opening in app

skoeva:
This fix prevents external URLs from being opened within the Headlamp app.

Fixes: #1917 

### Testing
- [x] Create new cluster and apply an ingress with an external URL
- [x] Click the external URL to ensure that it only opens in the web browser

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -574,6 +574,7 @@ function startElecron() {
       if (url.startsWith(startUrl)) {
         return;
       }
+      event.preventDefault();
       shell.openExternal(url);
     });
 


skoeva:
> package-lock.json change seems unrelated. Can you please remove it or add it in a different commit?

@knrt10 there were updates to a lot of electron related things but I've removed those for now

knrt10 (approved):

app/electron/main.ts
```diff
@@ -571,10 +571,11 @@ function startElecron() {
...
     */
     mainWindow.webContents.on('will-navigate', (event, url) => {
-      if (url.startsWith(startUrl)) {
-        return;
+      if (!url.startsWith(startUrl)) {
+        event.preventDefault();
+        shell.openExternal(url);
       }
-      shell.openExternal(url);
+      return;
```
joaquimrocha, line 578: You inverted the logic here but the return with this logic is not doing anything.
A more understandable change would be to just add the preventDefault, which seems to be the missing bit here.

joaquimrocha (changes requested):
Left a comment about simplifying the PR a bit.

app/electron/main.ts
skoeva: got it, made the change

joaquimrocha (approved):

Issue #1937, opened 2024-04-29
Plugins endlessly reloading on dev only app


skoeva:
running `npm run dev-only-app` locally with the `.plugins` directory in headlamp causes a loop of checks for updates to the plugins. deleting this directory fixes the issue though.
![Screenshot 2024-04-29 111907](https://github.com/headlamp-k8s/headlamp/assets/52797224/2e1a6195-b49c-4223-964c-0f3b2ca46037)
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/3581440d-368b-46c4-9f7d-7d0d97359908)

joaquimrocha:
I don't think we need to watch for changes in the static-plugins folder, so we should just remove that from the watch list.

PR #1939, opened 2024-04-29
backend: config: Change default plugins dir

skoeva:
This fix prevents the development app (when running `npm run dev-only-app`) from watching the .plugins directory locally, which causes an endless loop of refreshes to the app. To do this, we ensure that .plugins is not set as the plugins dir by default.

Fixes: #1937 

### Testing
- [x] Create .plugins folder locally (e.g. by running `make app-linux`)
- [x] Run `make run-frontend`, `make run-backend` and `npm run dev-only-app`
- [x] Ensure that "Plugins are loaded" message is only outputted once by the app
- [x] Ensure that the backend does not send "Sending reload plugins signal to frontend" in a loop

### DIFF
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -172,26 +172,17 @@ func flagset() *flag.FlagSet {
 
 // Gets the default plugins-dir depending on platform.
 func defaultPluginDir() string {
-	// These are the folders we use for the default plugin-dir.
-	//  - the passed in pluginDir if it's not empty.
-	//  - "./.plugins" if it exists.
+	// This is the folder we use for the default plugin-dir:
 	//  - ~/.config/Headlamp/plugins exists or it can be made
-	//  - "./.plugins" if the ~/.config/Headlamp/plugins can't be made.
 	// Windows: %APPDATA%\Headlamp\Config\plugins
 	//   (for example, C:\Users\USERNAME\AppData\Roaming\Headlamp\Config\plugins)
-	pluginDirDefault := "./.plugins"
-
-	if folderExists(pluginDirDefault) {
-		return pluginDirDefault
-	}
-
 	// https://www.npmjs.com/package/env-paths
 	// https://pkg.go.dev/os#UserConfigDir
 	userConfigDir, err := os.UserConfigDir()
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "getting user config dir")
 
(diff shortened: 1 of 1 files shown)


backend/pkg/plugins/plugins.go
```diff
@@ -52,6 +52,10 @@ func periodicallyWatchSubfolders(watcher *fsnotify.Watcher, path string, interva
 	for ; true; <-ticker.C {
 		// Walk the path and add any new directories to the watcher.
 		_ = filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
+			if strings.Contains(path, ".plugins") {
+				return filepath.SkipDir
+			}
```
joaquimrocha, line 57: This function (periodicallyWatchSubfolders) should be fine. The problem is that we are including from a higher level the static plugins dir. The static plugins dir should be included for building Headlamp with the shipped plugins, but it's called static because we don't expect those to be changed, and thus we should not listen to changes in the the static folder (whatever that folder is; nowadays it's by default .plugins, but could be something else).

joaquimrocha (changes requested):

backend/pkg/plugins/plugins.go
skoeva: when running `make run-backend` locally it shows the .plugins folder as the plugins dir rather than the static plugins dir, so would it make sense to only watch for changes to the plugin dir when there are plugins defined in static plugin dir?

joaquimrocha: That's unrelated IIUC. If we are picking up .plugins/ as taking precedence (over the default folder), then that should be changed, but that's a different PR. This one should be only about not watching the static plugins dir for changes.

backend/pkg/config/config.go
```diff
@@ -207,22 +199,12 @@ func defaultPluginDir() string {
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "creating plugins directory")
 
-		return pluginDirDefault
+		return "./.plugins"
```
joaquimrocha, line 202: Actually I thinking maybe we just never return this `.plugins` if we are never using it, and we remove it from the comments above as well.

skoeva: made the change, let me know if there's something we are missing here

joaquimrocha (approved):

Issue #1899, opened 2024-04-09
Make headlamp-plugin faster


joaquimrocha:
headlamp-plugin checks whether there are updates to it every time we run it. While this is nice and useful in a normal use, if we run the command repeatedly, this feature does impact dev experience by taking a long time when we may just be running a simple command, as every check takes usually 1.5-2.0 sec .

So I suggest we only check for updates a limited number of times per day. Maybe once a day is even enough.

If the update is to be skipped, we could print a message telling the user how to force check the updates.

cc/ @illume

illume:
Are you having this issue in just the `npm start` command?

That should usually be only run once, and then the command just sits there watching for changes - with no update check done. I understand it would be more of a problem when the `npm start` was buggy and it had to be run on each change.

joaquimrocha:
During the "watch" bug it was certainly annoying. But also in live demos it usually just seems like it takes too long.
I think it's worth making faster or only running sporadically, but certainly not urgent.

PR #1941, opened 2024-05-01
headlamp-plugin: Make update check asynchronous

skoeva:
This change makes the update check for headlamp-plugin asynchronous to avoid slowing down and blocking the user on start.

Fixes: #1899

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -231,7 +231,7 @@ function extract(pluginPackagesPath, outputPlugins) {
 
 /**
  * Start watching for changes, and build again if there are changes.
- * @returns {0} Exit code, where 0 is success.
+ * @returns {number} - Exit code, where 0 is success.
  */
 function start() {
   /**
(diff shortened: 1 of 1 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -269,24 +270,46 @@ function start() {
...
-    if ('@kinvolk/headlamp-plugin' in outdated) {
-      const url = `https://github.com/kinvolk/headlamp/releases`;
-      console.warn(
-        '    @kinvolk/headlamp-plugin is out of date. Run the following command to upgrade \n' +
-          `    See release notes here: ${url}` +
-          '    npx @kinvolk/headlamp-plugin upgrade'
+  async function informIfOutdated() {
+    const lastCheckTime = process.env.LAST_UPDATE_CHECK
+      ? parseInt(process.env.LAST_UPDATE_CHECK, 10)
+      : null;
```
joaquimrocha, line 278: This logic means lastCheckTime can be set as null, and then on line 282 it's used in arythmetic.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -992,9 +1053,20 @@ yargs(process.argv.slice(2))
...
+    yargs => {
+      yargs.option('force-check', {
+        describe: 'Force the update check regardless of the last check time.',
+        type: 'boolean',
+        default: false,
+      });
+    },
+    async argv => {
+      process.exitCode = await start(argv.forceCheck);
+    }
```
joaquimrocha, line 1068: Will this command still hang while it's started?

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -586,6 +609,44 @@ function getNpmOutdated() {
   return JSON.parse(result);
 }
 
+/**
+ * Use fetch to get the latest package version from the npm registry.
+ *
+ * @returns {Promise<Object>} - promise that resolves to an object containing version information
+ * if the package is outdated, or an empty object.
+ */
+async function getLatestPkgVersion() {
```
joaquimrocha, line 618: You introduced a new way to check for the latest version. But the previous function is still used for the upgrade. Should we unify this?

joaquimrocha (changes requested):
Besides the comments about the logic. I think this PR is doing 2 things: replacing the function that checks for the latest pkg version + limiting the checks.

illume (changes requested):
From this comment in the issue: https://github.com/headlamp-k8s/headlamp/issues/1899#issuecomment-2099965074

> It could be made async, so it doesn't block the user getting work done.
>
> If the check is done async, then it can show the message some moments later. This avoids having state, and has the benefit that it isn't slow once a day. The user is never slowed down in their work.

Currently doing `npm start` with this branch will do the check in an async fashion never block the user. So I think all of the last check stuff is not necessary.

The main goal of not blocking the user is achieved with the code:

```ts
  informIfOutdated().catch(error => {
    console.error('Failed to check for updates:', error);
  });
```

So please keep that `informIfOutdated()` call, and remove all the `lastCheckTime` code and the `force-check` code.

skoeva:
@illume I see, thanks for the fix! I think this makes sense since I noticed issues with performing the `getNpmOutdated()` code async, I can remove the debug lines and squash this now

joaquimrocha (approved):

PR #1961, opened 2024-05-06
WIP: frontend: router: Add lazy loading of routes for code splitting

illume:


sniok:
Splitting on every route might be an overkill imho, usually it's only needed in components that import big libraries (for example EditorDialog that imports monaco would be good case for lazy loading). Otherwise it'll just generate a lot of small chunks which might negatively impact build time

PR #1967, opened 2024-05-09
frontend: Refactor KubeObject class

sniok:
This PR refactors KubeObject class and how we define different Kubernetes resources.

Resource classes will now:
- extend KubeObject class instead of makeKubeObject function
  KubeObject currently resolves to any. So everywhere KubeObject was used there's no type safety. 
- not have `objectName`
- have `kind` `apiName` `isNamespaced` and `apiVersion` fields
- not have to define `apiFactory`, it will automatically made based on the class fields

This PR makes KubeObject a class that's available outside of makeKubeObject (deprecated) function

Summary of the changes

```ts
// Before
class Pod extends makeKubeObject<KubePod>('Pod') {
  static apiEndpoint = apiFactoryWithNamespace('', 'v1', 'pods');
}

// After
class Pod extends KubeObject<KubePod> {
  static kind = 'Pod';
  static apiName = 'pods';
  static apiVersion = 'v1';
  static isNamespaced = true;
}
```

This PR also fixes all the type errors that were surfaced by KubeObject not resolving to `any`


sniok:
This is ready for review. 
While there are a lot of files changed, most of them (with some exceptions) are only type changes. I've tried not to change any implementation to fix type errors (unless it was obvious like a typo)

sniok:
CI fails because CodeQL hangs on some files, I could isolate the issue to the ResourceTable import in one of the files but it's taking forever to track down what exactly trips CodeQL here. Will come back to this later

frontend/src/components/common/Resource/RestartButton.tsx
```diff
@@ -35,7 +36,7 @@ export function RestartButton(props: RestartButtonProps) {
 
   function applyFunc() {
     try {
-      const clonedItem = _.cloneDeep(item);
+      const clonedItem = _.cloneDeep(item) as Deployment;
```
joaquimrocha, line 39: I think _.cloneDeep accepts a <type>

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -127,11 +127,11 @@ function ScaleDialog(props: ScaleDialogProps) {
   const dispatchHeadlampEvent = useEventCallback(HeadlampEventType.SCALE_RESOURCE);
 
   function getNumReplicas() {
-    if (!resource?.spec) {
+    if (!('spec' in resource)) {
       return -1;
     }
 
-    return parseInt(resource.spec.replicas);
+    return parseInt((resource as any).spec.replicas);
```
joaquimrocha, line 134: The scale dialog is only applied to some types of resources, maybe we should set those in the declaration of the resource in the dialog props' interface, rather than adding this bypass here.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -294,9 +295,9 @@ export default function SimpleTable(props: SimpleTableProps) {
     return <Loader title={t('Loading table data')} />;
   }
 
-  let filteredData = displayData;
+  let filteredData = displayData!;
   if (filterFunction) {
-    filteredData = displayData.filter(filterFunction);
+    filteredData = displayData!.filter(filterFunction);
```
joaquimrocha, line 300: Weird, typescript was normally smart enough to understand this is prevalidated, IIRC.

frontend/src/components/crd/CustomResourceDetails.tsx
```diff
@@ -32,7 +33,7 @@ export function CustomResourceDetails(props: CustomResourceDetailsProps) {
   const { t } = useTranslation('glossary');
 
   const namespace = ns === '-' ? undefined : ns;
-  const CRD = ResourceClasses.CustomResourceDefinition as CustomResourceDefinition;
+  const CRD = ResourceClasses.CustomResourceDefinition;
```
joaquimrocha, line 36: Can you try the type declaration instead (const CRD: CustomResourceDefinition = ...).

frontend/src/components/crd/storyHelper.ts
```diff
@@ -100,14 +100,15 @@ const CRDMockMethods = {
...
 };
 
-class CRMockClass extends makeKubeObject<KubeObject>('customresource') {
+class CRMockClass extends KubeObject<KubeObjectInterface> {
+  static objectName = 'customresource';
   static apiEndpoint = apiFactoryWithNamespace(['', '', '']);
 
   static useApiGet(
-    setItem: (item: CRMockClass | null) => void,
+    setItem: (item: any | null) => void, // Update the type of the 'setItem' parameter
```
joaquimrocha, line 108: Did you mean this comment to be a `// @todo: `?

frontend/src/components/endpoints/Details.tsx
```diff
@@ -1,7 +1,7 @@
 import { useTranslation } from 'react-i18next';
 import { useLocation, useParams } from 'react-router-dom';
 import { ResourceClasses } from '../../lib/k8s';
-import Endpoints, { KubeEndpoint } from '../../lib/k8s/endpoints';
+import Endpoint, { KubeEndpoint } from '../../lib/k8s/endpoints';
```
joaquimrocha, line 4: I know it's confusing, but the K8s resource kind is indeed plural! https://kubernetes.io/docs/concepts/services-networking/service/#endpoints

frontend/src/lib/k8s/event.ts
```diff
@@ -21,7 +21,8 @@ export interface KubeEvent {
   [otherProps: string]: any;
 }
 
-class Event extends makeKubeObject<KubeEvent>('Event') {
+class Event extends KubeObject<KubeEvent> {
+  static objectName = 'Event';
```
joaquimrocha, line 25: The point of this objectName was to keep a consistent class name because the class names get mangled and minimized.
So I guess we should call this class name. Would there be a nicer way to define these than having a static var declared (it's fine, just wondering if there's anything alternative that is nicer).

frontend/src/lib/k8s/index.ts
```diff
@@ -39,53 +39,40 @@ import ServiceAccount from './serviceAccount';
...
+  ReplicaSet: ReplicaSet,
+  Role: Role,
+  RoleBinding: RoleBinding,
+  RuntimeClass: RuntimeClass,
+  Secret: Secret,
+  Service: Service,
+  ServiceAccount: ServiceAccount,
+  StatefulSet: StatefulSet,
+  StorageClass: StorageClass,
+};
```
joaquimrocha, line 75: Why all these declared with both key and value?

frontend/src/lib/k8s/index.ts
```diff
@@ -39,53 +39,40 @@ import ServiceAccount from './serviceAccount';
...
-const resourceClassesDict: {
-  [className: string]: KubeObject;
-} = {};
-
-classList.forEach(cls => {
-  // Ideally this should just be the class name, but until we ensure the class name is consistent
-  // (in what comes to the capitalization), we use this lazy approach.
-  const className: string = cls.className.charAt(0).toUpperCase() + cls.className.slice(1);
-  resourceClassesDict[className] = cls;
-});
```
joaquimrocha, line 86: Are you sure these are no longer needed?

frontend/src/lib/k8s/cluster.ts
```diff
@@ -55,7 +40,13 @@ export interface KubeObjectInterface {
...
   apiVersion?: string;
   metadata: KubeMetadata;
-  [otherProps: string]: any;
+  spec?: any;
+  status?: any;
+  items?: any[];
+  actionType?: any;
+  lastTimestamp?: string;
+  key?: any;
+  // [otherProps: string]: any;
```
joaquimrocha, line 49: Why this comment? This is also likely to make ts complain.
The idea here is that instead of having to cover always what k8s API may have, we allow for other unknown props to be defined.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -330,554 +294,25 @@ export interface AuthRequestResourceAttrs {
...
-type JsonPath<T> = T extends object
-  ? {
-      [K in keyof T]: K extends string ? `${K}` | `${K}.${JsonPath<T[K]>}` : never;
-    }[keyof T]
-  : never;
-
-// @todo: uses of makeKubeObject somehow end up in an 'any' type.
 
 /**
+ * @deprecated For backwards compatibility, please extend KubeObject
```
joaquimrocha, line 299: I think you need a period where you added a comma. Otherwise it reads as a the solution for the backwards compatibility.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -330,554 +294,25 @@ export interface AuthRequestResourceAttrs {
...
-  return KubeObject as KubeObjectIface<T>;
+  return KubeObjectInternal;
 }
 
-export type KubeObjectClass = ReturnType<typeof makeKubeObject>;
-export type KubeObject = InstanceType<KubeObjectClass>;
+/**
+ * This type refers to the *class* of a KubeObject.
+ */
+export type KubeObjectClass = typeof KubeObject<any>;
```
joaquimrocha, line 315: Should <unknown> be used instead?

joaquimrocha (changes requested):
Very nice changes! Really looking forward to ditch that hack of returning the class as a result of running the function.
Left some comments.
I mainly noticed that, while there are changes due to now having a lot more typing information and assertion, there are also many changes adding type casting, which bypasses the type checks, so I left some comment about those, but not about all instances of it.

I also think this may not be retro-compatible, i.e. this may break plugins because of how the ResourceClasses are being defined.
If we are to stop supporting the previous ResourceClasses, we then should rectify the objectName (actually className) of the resource classes, to always be upper case and maybe long form (`CustomResourceDefinition` instead of `crd`). But I'd keep a deprecation mechanism before if possible.

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -1141,7 +1143,7 @@ export function VolumeSection(props: VolumeSectionProps) {
   function PrintVolumeLink(props: PrintVolumeLinkProps) {
     const { volumeName, volumeKind, volume } = props;
     const resourceClasses = ResourceClasses;
-    const classList = Object.keys(resourceClasses);
+    const classList = Object.keys(resourceClasses) as Array<keyof typeof resourceClasses>;
```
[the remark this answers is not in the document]
sniok: no, unfortunately Object.keys returns just `string[]`

frontend/src/components/common/Resource/RestartButton.tsx
sniok: yes it accepts a generic but it cannot accept Deployment. casting here is fine, clonedItem exists for just a couple of lines

sniok: going to narrow down item type in props, the same way as in ScaleButton

frontend/src/components/common/SimpleTable.tsx
sniok: there's an additional case where it can be an undefined, but we only check for null. I'll remove undefined from displayData type

frontend/src/components/crd/CustomResourceDetails.tsx
sniok: casting was removed here, not added

frontend/src/lib/k8s/event.ts
sniok: I've kept the objectName to avoid breaking changes

frontend/src/lib/k8s/index.ts
sniok: for better type safety. in other files when using something like `ResourceClasses.Pod` it will be correctly typed as a `Pod`. Otherwise it would just be `Record<string, KubeObject>` which erases all that info

frontend/src/lib/k8s/index.ts
sniok: yeah, it constructs the ResourceClasses object, but now we have the object from the start

frontend/src/lib/k8s/index.ts
sniok: refactored where key and value are the same

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -24,7 +24,7 @@ export interface NameValueTableProps {
 function Value({
   value,
 }: {
-  value: string | JSX.Element | JSX.Element[] | undefined;
+  value: string | null | number | boolean | JSX.Element | JSX.Element[] | undefined;
```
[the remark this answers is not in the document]
sniok: there are some weird type errors, but with react 18 it works better, I'll look into it after we merge react upgrade

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -6,7 +6,7 @@ export interface NameValueTableRow {
   /** The name (key) for this row */
   name: string | JSX.Element;
   /** The value for this row */
-  value?: string | JSX.Element | JSX.Element[];
+  value?: string | number | null | boolean | JSX.Element | JSX.Element[];
```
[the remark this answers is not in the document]
sniok: same as above, will change after react 18

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -39,7 +39,7 @@ function Value({
       </>
     );
   } else {
-    return value;
+    return value as JSX.Element | null;
```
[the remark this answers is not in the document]
sniok: same as above, will change after react 18

frontend/src/lib/k8s/cluster.ts
sniok: ?

frontend/src/components/DetailsViewSection/DetailsViewSection.stories.tsx
```diff
@@ -63,10 +64,10 @@ const Template: Story<DetailsViewSectionProps> = args => {
 
 export const MatchRenderIt = Template.bind({});
 MatchRenderIt.args = {
-  resource: { kind: 'Node' },
+  resource: { kind: 'Node' } as KubeObject,
```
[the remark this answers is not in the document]
sniok: fixed, made a helper function that creates mock kubeobjects

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -272,9 +274,9 @@ export function DetailsGrid(props: DetailsGridProps) {
     if (Array.isArray(extraSections)) {
       actualExtraSections = extraSections;
     } else if (typeof extraSections === 'function') {
-      const extraSectionsResult = extraSections(item) || [];
+      const extraSectionsResult = extraSections(item!) || [];
       if (Array.isArray(extraSectionsResult)) {
-        actualExtraSections = extraSectionsResult;
+        actualExtraSections = extraSectionsResult as any[];
```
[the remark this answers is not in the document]
sniok: fixed

frontend/src/components/common/Resource/ResourceListView.tsx
```diff
@@ -1,28 +1,34 @@
...
 }
 
-export default function ResourceListView<ItemType>(
-  props: ResourceListViewProps<ItemType> | ResourceListViewWithResourceClassProps<ItemType>
+export default function ResourceListView<ItemClass extends KubeObjectClass>(
+  props: ResourceListViewWithResourceClassProps<ItemClass>
+): any;
+export default function ResourceListView<Item extends KubeObject<any>>(
+  props: ResourceListViewProps<Item>
+): any;
```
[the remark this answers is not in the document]
sniok: fixed, set as ReactElement

frontend/src/components/common/Resource/ScaleButton.tsx
sniok: done

frontend/src/components/common/NameValueTable/NameValueTable.tsx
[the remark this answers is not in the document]
sniok: replaced with ReactNode

frontend/src/components/common/NameValueTable/NameValueTable.tsx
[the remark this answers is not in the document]
sniok: replaced with ReactNode

frontend/src/components/common/NameValueTable/NameValueTable.tsx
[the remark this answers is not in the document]
sniok: replaced with ReactNode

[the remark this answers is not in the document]
sniok: removed type casting here, not needed anymore

sniok:
Rebased after react 18 update and updated some type to use ReactNode when needed

joaquimrocha (changes requested):

frontend/src/lib/k8s/cluster.ts
joaquimrocha: I have no idea what I meant anymore 👓

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -0,0 +1,561 @@
...
+  const clusterSettings = helpers.loadClusterSettings(cluster);
+  return clusterSettings.allowedNamespaces || [];
+}
+
+export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
+  jsonData: T;
+  /** Readonly field defined as JSONPath paths */
+  static readOnlyFields: string[] = [];
+  readonly _clusterName: string;
+  cluster?: string;
```
joaquimrocha, line 37: I think these should be the same, right? And we use resource.cluster in the recent multicluster changes, so this `cluster` var needs to be assigned to the cluster name in the constructor or we need to make this a getter (and just return the _clusterName) which is likely easier.

joaquimrocha (changes requested):
Left a comment about something I found. Also needs rebasing now.

frontend/src/lib/k8s/KubeObject.ts
sniok: yep you're right

sniok:
CodeQL issue was with `<U extends KubeObjectClass>` and replacing it with `<U extends typeof KubeObject>` fixed it

sniok:
I've managed to refactor types a bit to avoid CodeQL hanging (it's not as pretty but works)
Opened an issue in codeql repo, but for now this is okay

Issue #1653, opened 2024-01-12
Option to rename a cluster?


waldner:
Sometimes it would be useful to be able to assign a custom name to an imported cluster (for example, because it's the same or too similar to an existing one). It seems to me that currently headlamp just takes the name it finds in the kubeconfig and that cannot be changed afterwards.

joaquimrocha:
Yup. We should have that. I will add it to the backlog.

knrt10:
Working on this.

joaquimrocha:
I think the UI needs a bit of tweaking.
This is how it looked before:
![Error before](https://github.com/headlamp-k8s/headlamp/assets/1029635/f2dc7504-1193-466c-98e5-13d4f395866b)

This is after my patch:
![Error after](https://github.com/headlamp-k8s/headlamp/assets/1029635/997d7681-275e-4ac9-8002-502ff5ea2565)

Also, I added that restriction of only showing the UI if in electron, which was missing.

However, trying the logic again I see an issue: it renames but still is using the old name at some point.
![Error while authenticating to a renamed cluster](https://github.com/headlamp-k8s/headlamp/assets/1029635/b704a364-60b3-4a6c-96f0-6d9a864fb79c)

Renaming back to what it was originally also didn't work (didn't do the renaming back).

So I think this feature still needs some work and therefore it will be added for the next milestone.

PR #1974, opened 2024-05-13
Add enhancement to rename cluster

knrt10:
This adds feature where user can now use renamed cluster name in their UI.

Fixes: #1653 

### Testing

- Remove old kubeconfig `rm ~/.config/Headlamp/kubeconfigs/config`
- Run app `cd app && npm run dev-only-app`
- Run backend with stateless clusters enabled `cd backend && go run ./cmd -dev -enable-dynamic-clusters`
- Go to cluster settings and there is option for user to rename their cluster

### Screenshots

Screenshot from 2024-05-13 12-49-05

Screenshot from 2024-05-13 12-49-20

### TODO

- [x] Add for stateless clusters
- [x] Add tests


frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -39,12 +51,23 @@ export default function SettingsCluster() {
...
+
   const classes = useStyles();
   const theme = useTheme();
 
   const history = useHistory();
   const dispatch = useDispatch();
 
+  const handleUpdateClusterName = () => {
+    try {
+      console.log('came here', newClusterName); // Call the backend function to update the cluster name
```
ashu8912: I think the log came here by mistake, maybe just a comment we can use here.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -147,12 +174,33 @@ export default function SettingsCluster() {
     });
   }
 
+  function storeNewClusterName(name: string) {
+    let actualName = name;
+    if (name === cluster) {
```
ashu8912, line 192: can you please explain the logic here, Why would we set it to empty name if the cluster name has not changed??

frontend/src/components/App/Layout.tsx
```diff
@@ -70,7 +70,9 @@ export default function Layout({}: LayoutProps) {
    * indexDB and then sends the backend to parse it and then updates the parsed value into redux
    * store on an interval.
    * */
+  console.log('Layout.tsx: Layout()');
```
ashu8912, line 73: debug log which shouldn't be here

frontend/src/components/App/Layout.tsx
```diff
@@ -70,7 +70,9 @@ export default function Layout({}: LayoutProps) {
...
   useEffect(() => {
+    console.log('Mounted');
```
ashu8912, line 75: same

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -59,10 +59,23 @@ export default function SettingsCluster() {
...
       storeNewClusterName(newClusterName);
+      renameCluster(cluster || '', newClusterName, source)
+        .then(config => {
+          dispatch(setConfig(config));
+          history.push('/');
+          window.location.reload();
+        })
+        .catch((err: Error) => {
+          if (err.message === 'Not Found') {
+            // TODO: create notification with error message
```
ashu8912, line 76: Why is this a TODO?? Do you wish to do it in a different PR??

backend/cmd/headlamp.go
```diff
@@ -1289,6 +1290,149 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
+	if source == "kubeconfig" {
+		return c.kubeConfigPath, nil
+	}
+
+	return defaultKubeConfigPersistenceFile()
+}
+
+// Handler for renaming a cluster.
+//
+//nolint:funlen
```
ashu8912, line 1304: I think we can prevent this by moving the looping part of config.Contexts into a different function and letting it handle the contextName setting?? Would that be something difficult to achieve??? Does it add more complexity??

backend/cmd/cluster.go
```diff
@@ -28,3 +28,9 @@ type ClusterReq struct {
 type KubeconfigRequest struct {
```
ashu8912, line 28: Shouldn't this commit be before the frontend commits??

frontend/src/components/App/Settings/SettingsCluster.tsx
knrt10: Sure. If they are equal, it means that the user hasn't changed the name of the cluster. In this case:
It sets `actualName` to an empty string. This effectively clears the new name, indicating that the user doesn't want to change the cluster name.

It calls `setNewClusterName` to update the state with the empty string, ensuring that the UI reflects this change (if applicable). 

If the user attempts to rename the cluster but provides the same name as before, it indicates that they don't want to change the name. Setting it to an empty string reflects this intention.

Regardless of whether the cluster name has changed or not, it updates the `clusterSettings` state.

frontend/src/components/App/Settings/SettingsCluster.tsx
knrt10: I think this is not part of this PR, but in scope of notifications. Please refer to https://github.com/headlamp-k8s/headlamp/pull/1974/commits/c5f310e551daa7e4eac6dda5dfebf8018afbe3ac#diff-360436c42cc1c1880e31a8c726c2dc1cd6e52ed471da9443e9956ac8f7210d5aL79

backend/cmd/headlamp.go
```diff
@@ -1056,8 +1056,9 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			Server:   context.Cluster.Server,
 			AuthType: context.AuthType(),
 			Metadata: map[string]interface{}{
-				"source":    context.SourceStr(),
-				"namespace": context.KubeContext.Namespace,
+				"source":     context.SourceStr(),
+				"namespace":  context.KubeContext.Namespace,
+				"extensions": context.KubeContext.Extensions,
```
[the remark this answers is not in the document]
knrt10: We did not create additional extensions. It is part of k8s kubeconfig contexts. We are making direct changes to kubeconfig. Please refer to [k8s docs](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#Context) more context.

backend/cmd/headlamp.go
knrt10: It might not be easy because we have logging which should be part of k8s handler which creates additional lines. Even if we move some logic to other function it is not possible with clear code to have it under 60 with logging involved.

illume: I agree with Ashu and the linter about this function being too long. It will be easier to add tests then as well.

joaquimrocha (changes requested):

joaquimrocha:
So I swear I had sent a message here with some screenshots showing some problems but I don't see the message now...
Anyway, I pushed some changes to make the UI hopefully nicer + restrict it to electron only (this is important!), and I have also verified that some of the problems I had which made me think we shouldn't add this to the 0.24.0 release were in a past iteration of this branch.
Now things seem to work fine, though it looks like if I switch between the default cluster name and the new one, in the settings it shows me the previous name (not sure if this is autocomplete getting in the way).

Anyway, here is how it looks now (notice how the error status no longer extends the input till the end of the space):
Screenshot

joaquimrocha:
@illume What do you think? Should try to add it for 0.24.0 or should we move it (keep it now, since I had changed the milestone when I was in the outdated branch) to 0.25.0?

frontend/src/i18n/locales/pt/translation_old.json
joaquimrocha, line 1: *_old.json files shouldn't be checked into the repo.

joaquimrocha (dismissed):

illume (approved):
👍 thanks

illume:
@knrt10 

Seems the text is wrong after I change it, then change it back.

- click 'minikube' cluster and cluster settings
- change 'minikube' to 'mycluster'
- change back to 'minikube'
- click cluster, and cluster settings... it says 'mycluster' still in the input text.


See in the minikube cluster but the text has the old 'mycluster' there. It should be 'minikube' or empty.
image

illume (dismissed):
I had a look around for the bug a little bit, but couldn't find it.

Probably we should add some tests, and perhaps refactor SettingsCluster a bit.

illume:
Maybe we should add this flag to dev mode? Or to the app?

```
diff --git a/Makefile b/Makefile
index 2d426825..102fd637 100644
--- a/Makefile
+++ b/Makefile
@@ -86,10 +86,10 @@ run-backend:
        @echo "**** Warning: Running with Helm and dynamic-clusters endpoints enabled. ****"

 ifeq ($(UNIXSHELL),true)
-       HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/*
+       HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/* -enable-dynamic-clusters
 else
        @echo "**** Running on Windows without bash or zsh. ****"
-       @cmd /c "set HEADLAMP_BACKEND_TOKEN=headlamp&& set HEADLAMP_CONFIG_ENABLE_HELM=true&& set HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true&& backend\headlamp-server -dev -proxy-urls https://artifacthub.io/*"
+       @cmd /c "set HEADLAMP_BACKEND_TOKEN=headlamp&& set HEADLAMP_CONFIG_ENABLE_HELM=true&& set HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true&& backend\headlamp-server -dev -proxy-urls https://artifacthub.io/*" -enable-dynamic-clusters
 endif

 run-frontend:
```

knrt10:
@illume fixed, can you please test again. Thanks

illume:
I'm having an issue where it's not changing...

image

illume:
Ok... it works after I clear my localstorage. I think when it was broken something bad got in there.

skoeva (approved):
Works great on my end, thanks for this! 🙌

PR #1976, opened 2024-05-13
frontend: Add node/npm version requirements to package.json

illume:
So people will get a hint when they install that they have older versions of node/npm than is required.

image

### How to test

- Use node 18.
- Run `cd frontend && npm install`
- See warning message in screenshot about node version being wrong.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,6 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "productName": "Headlamp",
+  "engines" : {
+    "npm" : ">=10.0.0",
+    "node" : ">=20.9.0"
+  },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -3,6 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "productName": "Headlamp",
+  "engines" : {
+    "npm" : ">=10.0.0",
+    "node" : ">=20.9.0"
```
ashu8912, line 8: This is a nice dx change but i was wandering if this is the limit we have will node 18 not work??

illume: We had some people try it at the hackerthon and fail to get the install working with node 18. We also only test in CI with node 20.x.

This version of node is the first LTS release in the 20.x series.

joaquimrocha (approved):

PR #1981, opened 2024-05-15
frontend: Replace create-react-app with vite and vitest

sniok:
This PR continues work started in https://github.com/headlamp-k8s/headlamp/pull/1696
Thanks @tazo90! 

Replaced cra with vite
Remoted storybook storyshot addon because it is deprecated and doesn't work with vite
Updated all tests to use vitest instead of jest

Two tests don't work because of the bug in vitest, it stuggles with cyclic imports sometimes. 

headlamp-plugin doesn't work yet, looking into it

changes:
- [x] Migrated frontend/ from CRA to vite
- [x] Migrated from jest to vitest
- [x] Migrated from storybook/storyshots 
- [x] See if I can keep the REACT_ prefix for env variables
- [x] Create readme file for the frontend/
- [x] Fix baseUrl replacement in backend
- [x] Update all env variable access to import.meta.env
- [x] Sync dependencies  

testing:
- [x] Tested building headlamp-plugin
- [x] Tested compiling, bundling and running plugin in a frontend/ built by vite 
- [x] Tested app/
- [x] Tested -base-url and PUBLIC_URL options


sniok:
regenerated package-lock and now a lot of things that were working before are broken, very frustrating 🙃

Dockerfile
```diff
@@ -39,7 +39,7 @@ COPY app/package.json /headlamp/app/package.json
 COPY frontend/package*.json /headlamp/frontend/
 COPY frontend/patches/* /headlamp/frontend/patches/
 WORKDIR /headlamp
-RUN cd ./frontend && npm install --only=prod
+RUN cd ./frontend && npm install
```
sniok, line 42: vite is a dev dependency and we need it to build the frontend/
so --only=prod was removed

backend/cmd/headlamp.go
```diff
@@ -166,10 +166,14 @@ func baseURLReplace(staticDir string, baseURL string) {
 
 	copyReplace(indexBaseURL,
 		index,
+		[]byte("headlampBaseUrl = './'"),
```
sniok, line 169: needed to change this since vite outputs a slightly different html

frontend/README.md
```diff
@@ -1,68 +1,50 @@
-This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+# Headlamp Frontend
```
sniok, line 1: replaced a placeholder CRA readme, this duplicates a bit what we have in the docs and we would need to consolidate this somehow but not in scope of this pr

frontend/index.html
```diff
@@ -2,18 +2,15 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="icon" href="/favicon.ico" />
```
sniok, line 5: no need for PUBLIC_URL here as vite handles that nicely
https://vitejs.dev/guide/assets.html
https://vitejs.dev/guide/build.html#public-base-path

frontend/package.json
```diff
@@ -174,54 +176,30 @@
...
     "@types/redux-mock-store": "^1.0.4",
+    "@vitest/coverage-v8": "^1.6.0",
     "fetch-mock": "^9.11.0",
     "http-proxy-middleware": "^2.0.1",
     "husky": "^4.3.8",
-    "i18next-parser": "^4.7.0",
-    "jest-canvas-mock": "^2.4.0",
-    "jest-html-reporter": "^3.10.2",
-    "jest-websocket-mock": "^2.5.0",
+    "i18next-parser": "^8.13.0",
```
sniok, line 197: had a bunch of CommonJS import errors with the old i18next-parser version so I updated it 
I don't see a lot of breaking changes, just some improvements in the way plural is handled, you can see updates to the translation files

frontend/src/components/404/index.tsx
```diff
@@ -5,6 +5,9 @@ import ErrorComponent from '../common/ErrorPage';
 export default function NotFoundComponent() {
   const { t } = useTranslation();
   return (
-    <ErrorComponent graphic={headlampBrokenImage} title={t(`Whoops! This page doesn't exist`)} />
+    <ErrorComponent
+      graphic={headlampBrokenImage as any}
```
sniok, line 9: headlamp-plugin build step complained about the types here, not a nice fix but the type casting is very local and not that important here

frontend/src/components/App/AppLogo.tsx
```diff
@@ -3,10 +3,10 @@ import { SxProps } from '@mui/system';
 import React, { isValidElement, ReactElement } from 'react';
 import { getThemeName } from '../../lib/themes';
 import { useTypedSelector } from '../../redux/reducers/reducers';
-import { ReactComponent as LogoDark } from '../../resources/icon-dark.svg';
-import { ReactComponent as LogoLight } from '../../resources/icon-light.svg';
-import { ReactComponent as LogoWithTextDark } from '../../resources/logo-dark.svg';
-import { ReactComponent as LogoWithTextLight } from '../../resources/logo-light.svg';
+import LogoDark from '../../resources/icon-dark.svg?react';
```
sniok, line 6: I tried to find a way for `vite-plugin-svgr` to behave the same way as CRA but couldn't

frontend/src/stateless/index.ts
```diff
@@ -117,10 +104,7 @@ function handleDataBaseError(event: DatabaseErrorEvent, reject: (reason?: any) =
  */
 export function storeStatelessClusterKubeconfig(kubeconfig: string): Promise<void> {
   return new Promise<void>(async (resolve, reject) => {
-    const request =
-      process.env.NODE_ENV === 'test'
-        ? indexedDBtest.open('kubeconfigs', 1)
-        : indexedDB.open('kubeconfigs', 1);
+    const request = indexedDB.open('kubeconfigs', 1) as any;
```
sniok, line 107: doing mocking here wasn't really necessary, we can mock indexedDB in test files

frontend/src/storybook.test.tsx
```diff
@@ -1,58 +1,93 @@
-import initStoryshots, { Stories2SnapsConverter } from '@storybook/addon-storyshots';
-import * as rtl from '@testing-library/react';
+import 'vitest-canvas-mock';
```
sniok, line 1: this is the replacement for addon-storyshots which doesn't work with vite and also is deprecated.
used this guide: https://storybook.js.org/docs/writing-tests/storyshots-migration-guide

now it generates one file for all snapshots which is not ideal. I've tried to generate individual files but for some reason vitest `toMatchFileSnapshot` function doesn't format the output properly when generating snapshots

sniok:
this is ready for review @joaquimrocha @illume

frontend/index.html
illume: Did you try with a different baseURL?

illume:
Are you going to get storyshots working?

illume:
There should be a thankyou/co-author added from the other vite PR.

illume:
It could be a good idea to do preparation PR(s) of changes. Changes that can be done to prepare for this one that don't break what's already there. This way the size of the PR could be reduced.

frontend/src/components/App/runCommand.ts
```diff
@@ -37,7 +37,7 @@ export function runCommand(
     throw new Error('runCommand only works in Headlamp app mode.');
   }
 
-  if (process.env.REACT_APP_ENABLE_RUN_CMD !== 'true') {
+  if (import.meta.env.REACT_APP_ENABLE_RUN_CMD !== 'true') {
```
illume, line 40: Can support for process.env be kept somehow?

Mostly interested for a future move of the plugins to vite.

sniok:
Improved storyshots test, now it generates separate files for each story, instead of putting it in one big file.
Initially I thought there was something wrong with the formatter but then I figured out that toMatchFilesnapshot would override the same file when there are more than 1 story per file, so now it generates it in separate files. 

Moved the i18next-parser version bump into separate commit. Everything else seems to require vite so keeping it in the other (big) commit. 

Added env plugin to make process.env work

frontend/index.html
sniok: yep, it works

frontend/src/components/App/runCommand.ts
sniok: revisited this topic. I removed the process env plugin because app itself is using import.meta.env and plugins don't have access to this since all the env variables are statically replaced during build time and process.env is not available at all. 
you can check it by printing process in the dev console in the current version of the app

sniok:
Some performance comparison:

## npm run start
measured time from running the command to page finishing loading in the browser

CRA cold start: ~23s
CRA hot start: ~17s

Vite hot/cold start: ~6s
_vite server starts in <200ms but initial page load takes some time_

## Build

CRA build: 41s
vite build: 41s

same time, no surprises here. patiently waiting for [rolldown](https://rolldown.rs/about)

## hot reload

CRA reload: 2s
vite reload: <100ms, too fast to measure 

now this is one of the best quality of life improvement, having quick feedback loop while developing frontend is great

sniok:
Build output comparison is here: https://gist.github.com/sniok/95165478c77f1bf86e752c7d6f1f9741

It's mostly similar, I don't anything concerning.

Unrelated to vite, there is a lot we can do about what's in each chunk. We can improve headlamp startup performance by using more lazy components and improving code splitting

frontend/src/components/App/runCommand.ts
joaquimrocha: We have an issue about how plugins cannot access the process.env anyway, or maybe that's an opportunity for not having the process anymore, if plugins will be able to access the vars differently anyway.

illume: It would be better to keep process.env because it's a known node API, and also known in other tools like in vue and create-react-app / nextjs / etc.

joaquimrocha (approved):
LGTM!

frontend/src/components/App/runCommand.ts
sniok: aliasing import.meta.env to process.env breaks the process polyfill so it's one or the other.
I've tried to keep process.env but it just doesn't work together with polyfill unfortunately.

in addition import.meta is part of the standard [import.meta - JavaScript | MDN (mozilla.org)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta) unlike process which is a node specific thing that we should try to move away from (similar to commonjs)

illume (approved):
👍

illume:
Adding the related issue here: https://github.com/headlamp-k8s/headlamp/issues/1282

Issue #1820, opened 2024-03-15
Make create/editor dialog work with a given YAML/JSON


joaquimrocha:
The `EditorDialog` component receives an item that needs to look like a Kubernetes resource. It takes into account things like the resourceVersion and other aspects of the K8s resources, which is nice, but it is impossible to use this editor with just plain text.
E.g. if I want to pass it a YAML that is a list of resources, it is not possible. The only reason why the editor works for the create button is that it is opened with an empty item and then the user can add the code to it.

We should make sure that we have an editor component that accepts any YAML/JSON, and then only if we open an existing resource, we use the logic currently implemented in `EditorDialog`.

This affects e.g. any plugin who wants to show the default editor with code to the user.

illume:
I see the PR is still open... It looks like this will be for the next release, so I changed the milestone. 

/cc @joaquimrocha fyi

PR #1982, opened 2024-05-15
frontend: Accept given YAML/JSON in EditorDialog

skoeva:
This change aims to allow `EditorDialog` to accept generic YAML/JSON text rather than explicitly expecting an item that looks like a Kubernetes resource. It also adds placeholder text in the editor for improving accessibility.

Fixes: #1820


skoeva:
Closed in favor of #1996

Issue #1968, opened 2024-05-09
Empty X-Forwarded-Host header


adriangonz:
## Problem
In Go, trying to access the `Host` header directly of a request object through `r.Header.Get("Host")` always return an empty string - regardless of whether the header is actually set or not. Instead, you need to use the `r.Host` field (which under the hood will just read the actual header). This leads to the `X-Forwarded-Host` header being left empty when requests are proxied to the cluster API Server:

https://github.com/headlamp-k8s/headlamp/blob/d4f774877ed58ad9eb8d9fa82bd0348355b79f58/backend/cmd/headlamp.go#L1010

This can cause problems when the API Server is fronted by some reverse proxies (e.g. NGINX) which do a strict implementation of the HTTP 1.1 protocol - where the `Host` (or `X-Forwarded-Host`) has to always be present. Otherwise, every request sends back a `400` response with an error along the following lines:

```
400 Bad Request: missing required Host header error
```

## Proposed Solution

To work around the above, it should be enough to just read the `Host` header directly via `r.Host`, i.e. doing something like (note that `r.Host` is already set on the line above):

```golang
r.Header.Set("X-Forwarded-Host", r.Host)
```

joaquimrocha:
@yolossn Can you check this one out? I think this code was last touched by you.

PR #1984, opened 2024-05-20
backend: Fix empty X-Forwared-Host header

knrt10:
The header was not getting set correctly and returning an empty string

Fixes: #1968


joaquimrocha:
Any way we can test this?

knrt10:
We could add a unit test to check for header. Will do that

joaquimrocha (approved):

PR #1940, opened 2024-04-30
frontend: Add crd's categories 

Guilamb:
**Description:**
---
As mentioned in #955, this pull request addresses the need to filter CRDs by categories. 
Here the features implemented:

- **List of Categories**: Added the ability to list all available categories.
- **Category Selection**: Implemented the functionality to select and unselect categories.
- **Filtering**: Enabled filtering CRDs based on selected categories.

**Implementation Details:**
---
- Introduced a new component, `CRDCategoriesList`, to facilitate category selection and filtering.
- Utilized Redux for state management, specifically the `filterSlice`, to store and update the selected categories.
- Integrated with existing CRD functionality to fetch and display categories dynamically.

**Changes Made:**
---
- Added the `CRDCategoriesList` component to handle category selection and filtering.
- Updated Redux state management to accommodate the new category filter.
- Enhanced UI to display available categories and allow user interaction.

**Feedback Request:**
---
- Any suggestions for optimizing loading time for the `CRDCategoriesList` component.
- Review of Redux state management implementation and usage.
- Feedback on UI/UX improvements or any potential issues.

**Known Issues:**
---
- Currently experiencing loading time issues with the `CRDCategoriesList` component. Gather categories when initially loading the CRD may resolve the issue.

PR #1991, opened 2024-05-27
Display categories for CRDs

joaquimrocha:
Adds the categories item to the CRD list and details views.
Based on #1940 by @guilamb.

How to test:
- [ ] Create a [CRD with categories](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#categories) and check if the categories are displayed in the new CRD/list column and the new CRD/details field


illume (approved):
Thanks folks. 👍

PR #1993, opened 2024-05-27
Do not use isNamespaced for adding namespace to API calls

joaquimrocha:
This isNamespaced will not work in the case of CRDs for example.

How to test:
- [ ] Create a CRD; then delete it -> it was not succeeding to delete before this patch.


skoeva:
I can create and delete CRDs on main without any problems, can you reproduce the issue?

skoeva (approved):

PR #1994, opened 2024-05-27
pod/List: Fix IPs being displayed as "undefined" when not set

joaquimrocha:
Just noticed that after the table changes we have pod IPs as being `undefined` (when there's no IP defined) in the pod list.

How to test:
- [ ] Go to a pod that has no IP set to them; check how it shows nothing in the IP column (as opposed to `undefined`)

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -500,11 +500,11 @@ export default function PodDetails(props: PodDetailsProps) {
           },
           {
             name: t('Host IP'),
-            value: item.status.hostIP,
+            value: item.status.hostIP ?? '',
           },
           {
             name: t('Pod IP'),
-            value: item.status.podIP,
+            value: item.status.podIP ?? '',
           },
           {
             name: t('QoS Class'),
--- a/frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
+++ b/frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
@@ -5703,9 +5703,7 @@ exports[`Storyshots Pod/PodDetailsView Successful 1`] = `
                 >
                   <span
                     class="MuiTypography-root MuiTypography-body1 css-1qti38g-MuiTypography-root"
-                  >
-                    0.0.0.1
-                  </span>
+                  />
                 </dd>
                 <dt
                   class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-4 css-trrd7p-MuiGrid-root"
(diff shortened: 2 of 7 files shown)


frontend/src/components/pod/Details.tsx
```diff
@@ -500,11 +500,11 @@ export default function PodDetails(props: PodDetailsProps) {
           },
           {
             name: t('Host IP'),
-            value: item.status.hostIP,
+            value: item.status.hostIP || '',
```
sniok, line 503: "new" recommended way of assigning defaults is to use `??` which handles edge cases better than `||`.
in this case it doesn't matter but it's a good habit 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

sniok (approved):

frontend/src/components/pod/Details.tsx
joaquimrocha: I know that but I tend to always use ||. I will change it though;

frontend/src/components/pod/__snapshots__/PodDetails.stories.storyshot
```diff
@@ -5471,9 +5471,7 @@ exports[`Storyshots Pod/PodDetailsView Successful 1`] = `
                 >
                   <span
                     class="MuiTypography-root MuiTypography-body1 css-1qti38g-MuiTypography-root"
-                  >
-                    0.0.0.1
```
illume, line 5475: Seems odd this IP address isn't being rendered anymore?

joaquimrocha: It's just because I did modify one of the mock pods to remove any IPs from it. So this is expected, even though I understand the confusion.

vyncent-t:
showed undefined for a minute when in the browser (may have been a weird pod I created or was behind on the render) but the storybook comp shows empty, current build error keeps the browser view away

frontend/src/components/pod/Details.tsx
```diff
@@ -500,11 +500,11 @@ export default function PodDetails(props: PodDetailsProps) {
           },
           {
             name: t('Host IP'),
-            value: item.status.hostIP,
+            value: item.status.hostIP ?? '',
           },
           {
             name: t('Pod IP'),
-            value: item.status.podIP,
+            value: item.status.podIP ?? '',
```
illume, line 507: Below it's `pod.status?.podIP || ''` should it be that here too?

illume:
Rerunning the container e2e tests, which failed.

illume:
> showed undefined for a minute when in the browser (may have been a weird pod I created or was behind on the render) but the storybook comp shows empty, current build error keeps the browser view away

I did some digging on this... The API reference says the status field should always be there. Could be a kubernetes bug though where the status field is missing.

illume:
I reran the container job a few times, and it keeps failing. It's passing recently in other PRs, so I guess it's a real issue.

joaquimrocha:
The problem is that if we do not merge, then this is a real issue as it could show undefined for the IP (the state must exist but the IP could be undefined).

illume:
This is the last one that passed the container build from this PR: https://github.com/headlamp-k8s/headlamp/actions/runs/9288417462/job/25559819347

With this commit fc6578ae5885560c574a3b547160c5f13b3b45fe

illume (approved):
👍

PR #1996, opened 2024-05-28
frontend: Add create resource UI

skoeva:
These changes introduce a new UI feature that allows users to create resources from the associated list view. Clicking the 'Create' button opens up the `EditorDialog` used in the generic 'Create / Apply' button, now accepting generic YAML/JSON text rather than explicitly expecting an item that looks like a Kubernetes resource. The dialog box also includes a generic template for each resource. The apply logic for this new feature (as well as the original 'Create / Apply' button) has been consolidated in EditorDialog, with a flag allowing external components to utilize their own dispatch functionality.

Fixes: #1820 

#### Create resource button
- [x] Open a cluster in Headlamp (`cd app & npm start`) and navigate to a resource (e.g. Config Maps)
- [x] Click the + sign in the header and ensure the appropriate template appears in the editor
- [x] Apply changes and ensure the created resource appears in the list view

#### EditorDialog changes
- [x] Open Headlamp (`cd app & npm start`) and click the 'Create / Apply' button
- [x] Pass in a list of resources to the editor (e.g. a list of Config Maps)
- [x] Apply changes and ensure that all the resources in the list appear in the list view

### Screenshots
image
image


ashu8912:
These are some nice set of changes thanks @skoeva I was wandering what all resources we are covering to have the create resource button on?? And what is the reasoning behind adding it to some and not to others if that's the case?

skoeva:
> These are some nice set of changes thanks @skoeva I was wandering what all resources we are covering to have the create resource button on?? And what is the reasoning behind adding it to some and not to others if that's the case?

@ashu8912 Currently covering config maps, secrets, leases, and runtime classes as those I already found with base templates in their associated directories, though if needed, we can add templates for more resources and cover those as well

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { getCluster } from '../../lib/cluster';
+import { apply } from '../../lib/k8s/apiProxy';
+import { KubeObjectInterface } from '../../lib/k8s/cluster';
+import { clusterAction } from '../../redux/clusterActionSlice';
+import { EventStatus, HeadlampEventType, useEventCallback } from '../../redux/headlampEventSlice';
+import { ActionButton, EditorDialog } from '../common';
+import { BASE_EMPTY_CONFIG_MAP } from '../configmap/storyHelper';
+import { LEASE_DUMMY_DATA } from '../lease/storyHelper';
+import { BASE_RC } from '../runtimeClass/storyHelper';
+import { BASE_EMPTY_SECRET } from '../secret/storyHelper';
```
illume, line 13: I guess it might make sense to copy these default structures in here, because their use case is different. I worry someone might change the story in the future not realizing they are not just changing a test... but data that is used to create things.

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+        });
+      }
+    );
+  };
+
+  function handleSave(newItemDefs: KubeObjectInterface[]) {
+    let massagedNewItemDefs = newItemDefs;
+    const cancelUrl = location.pathname;
+
+    // check if all yaml objects are valid
```
illume, line 58: Does k8s give us back error messages?

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -34,8 +34,8 @@ if (process.env.NODE_ENV === 'test') {
 type KubeObjectIsh = Partial<KubeObjectInterface>;
 
 export interface EditorDialogProps extends DialogProps {
-  /** The object to edit, or null to make the dialog be in "loading mode". Pass it an empty object if no contents are to be shown when the dialog is first open. */
-  item: KubeObjectIsh | null;
+  /** The object(s) to edit, or null to make the dialog be in "loading mode". Pass it an empty object if no contents are to be shown when the dialog is first open. */
+  item: KubeObjectIsh | object | object[] | string | null;
```
illume, line 38: Can you please explain why object, and these other types were added?

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -82,36 +85,48 @@ export default function EditorDialog(props: EditorDialogProps) {
     setUseSimpleEditorState(data);
   }
 
+  function isKubeObjectIsh(item: any): item is KubeObjectIsh {
```
illume, line 88: Love the function name.

frontend/src/components/common/CreateResourceButton.tsx
illume, line 1: Can you please add a CreateResourceButton.stories.tsx for this component?

illume (commented):
Nice work.

I left a few notes for your consideration.

As well as a CreateResourceButton.stories.tsx, maybe a few small pieces could be unit tested, and perhaps an e2e test would be useful for the overall feature?

illume:
> > These are some nice set of changes thanks @skoeva I was wandering what all resources we are covering to have the create resource button on?? And what is the reasoning behind adding it to some and not to others if that's the case?
> 
> @ashu8912 Currently covering config maps, secrets, leases, and runtime classes as those I already found with base templates in their associated directories, though if needed, we can add templates for more resources and cover those as well

Probably you know this already... but the k8s website gives example yaml for each resource in their docs.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -96,36 +99,48 @@ export default function EditorDialog(props: EditorDialogProps) {
...
 
-    const originalCode = originalCodeRef.current.code;
-    const itemCode =
-      originalCodeRef.current.format === 'json' ? JSON.stringify(item) : yaml.dump(item);
-    if (itemCode !== originalCodeRef.current.code) {
-      originalCodeRef.current = { code: itemCode, format: originalCodeRef.current.format };
-    }
+    // Determine the format (YAML or JSON) and serialize to string
+    const format =
+      Array.isArray(item) || (typeof item === 'object' && !('kind' in item && 'apiVersion' in item))
```
joaquimrocha, line 117: Why does this determine that the format is `yaml`?

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -296,7 +311,9 @@ export default function EditorDialog(props: EditorDialogProps) {
   const errorLabel = error || errorMessage;
   let dialogTitle = title;
   if (!dialogTitle && item) {
-    const itemName = item.metadata?.name || t('New Object');
+    const itemName = isKubeObjectIsh(item)
+      ? item.metadata?.name || t('New Object')
+      : t('New Object');
```
joaquimrocha, line 316: Is this equivalent to this simplification?
```typescript
const itemName = (isKubeObjectIsh(item) && item.metadata?.name) || t('New Object');
```

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+import { apply } from '../../lib/k8s/apiProxy';
+import { KubeObjectInterface } from '../../lib/k8s/cluster';
+import { KubeConfigMap } from '../../lib/k8s/configMap';
+import { KubeRuntimeClass } from '../../lib/k8s/runtime';
+import { KubeSecret } from '../../lib/k8s/secret';
+import { clusterAction } from '../../redux/clusterActionSlice';
+import { EventStatus, HeadlampEventType, useEventCallback } from '../../redux/headlampEventSlice';
+import { ActionButton, EditorDialog } from '../common';
+
+const creationTimestamp = new Date('2022-01-01').toISOString();
```
joaquimrocha, line 14: A `creationTimestamp` is not needed nor advisable for creating an object (say you take 3 hours to press send, then the creation timestamp is not the actual creation timestamp).

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+const BASE_EMPTY_CONFIG_MAP: KubeConfigMap = {
```
joaquimrocha, line 15: I think these templates should be part of each resource class.
Similar to what we are doing with https://github.com/headlamp-k8s/headlamp/pull/2036 .

We can have a `getBaseObject` static function. So we can do `ConfigMap.getBaseObject()` to get the very basic requirements for setting up such resources.

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+  apiVersion: 'v1',
+  kind: 'ConfigMap',
+  metadata: {
+    creationTimestamp: '2023-04-27T20:31:27Z',
+    name: 'my-pvc',
+    namespace: 'default',
```
joaquimrocha, line 21: The default namespace may not be `default`, and it's automatically set by Headlamp automatically.

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+  },
+  data: {},
+  type: 'bla',
+};
+
+export interface CreateResourceButtonProps {
+  resource: string;
+}
+
+export function CreateResourceButton(props: CreateResourceButtonProps) {
```
joaquimrocha, line 92: AFAIU, the only difference between the CreateButton and this one, is that the this one uses a predefined item. So we are kind of duplicating the apply logic between this button, the CreateButton, and it's also close to the EditorButton.

So, it would be nice to be able to run the creation dialog from any other component. So we could perhaps move the apply logic to the EditorDialog, and then we open it with the code we want, and from where we want.

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,210 @@
...
+  return (
+    <React.Fragment>
+      <ActionButton
+        color="primary"
+        description={t('translation|Create {{ resource }}', { resource })}
+        icon={'mdi:plus-circle'}
+        onClick={() => {
+          setOpenDialog(true);
+        }}
+      />
```
joaquimrocha, line 196: This button needs to be within a AuthVisible component. Because the user may not have the permissions to create the resource.

joaquimrocha (changes requested):

frontend/src/components/configmap/List.tsx
```diff
@@ -1,13 +1,32 @@
...
+  kind: 'ConfigMap',
+  metadata: {
+    creationTimestamp: '2023-04-27T20:31:27Z',
+    name: 'my-pvc',
+    namespace: 'default',
+    resourceVersion: '1234',
+    uid: 'abc-1234',
+  },
+  data: {},
+};
```
joaquimrocha, line 17: I thought I had reviewed it a couple of weeks ago where I mentioned that the create timestamp should be set by us, nor the resource version or UID.
I had also mentioned that this base config should be at the cluster.ts's base object class, using a static var `baseObject`, and then every subclass should implement their own version in their class declaration (in ConfigMap.ts, etc.).

frontend/src/components/configmap/List.tsx
```diff
@@ -1,13 +1,32 @@
...
+
 export default function ConfigMapList() {
   const { t } = useTranslation(['glossary', 'translation']);
 
   return (
     <ResourceListView
       title={t('glossary|Config Maps')}
+      headerProps={{
+        titleSideActions: [
+          <CreateResourceButton resource={BASE_EMPTY_CONFIG_MAP} resourceName="ConfigMap" />,
```
joaquimrocha, line 27: You can add this button as default, and the ResourceListView is already passed the resourceClass, so if we have the base object declared the class as I suggested, then it should be easy to check whether the class has a base object and use a generic create resource button for it if it does.

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+      case 'ConfigMap':
+        return ConfigMap;
+      case 'Lease':
+        return Lease;
+      case 'RuntimeClass':
+        return RuntimeClass;
+      case 'Secret':
+        return Secret;
+    }
+  }
```
joaquimrocha, line 117: Please check my comment regarding passing only the class here and checking whether the class has a baseObject defined.

joaquimrocha (changes requested):

frontend/src/components/configmap/List.tsx
skoeva: these fields we should not set but are necessary to create these objects, would empty strings suffice? these will show up in the dialog as well and the user would have to delete them manually with the current setup

vyncent-t: I ran into a similar issue with the timestamp and the UID the fields would throw an error if I didn't include something in UID or timestamp. Maybe something like the solution from this PR to make those fields non required as input? 

https://github.com/headlamp-k8s/headlamp/pull/2155

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  resourceClass: ClassWithBaseObject<KubeObject>;
+}
+
+export function CreateResourceButton(props: CreateResourceButtonProps) {
+  const { resourceClass } = props;
+  const { t } = useTranslation(['glossary', 'translation']);
+  const [openDialog, setOpenDialog] = React.useState(false);
+  const [errorMessage, setErrorMessage] = React.useState('');
+
+  const baseObject = resourceClass.getBaseObject ? resourceClass.getBaseObject() : {};
```
joaquimrocha, line 16: This getBaseObject should be available on the base class defined in cluster.ts, so we shouldn't have to check if the function exists. If it gives `null` or something similar by default, we could do:
```typescript
const baseObject = resourceClass.getBaseObject() ?? {};
```

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <React.Fragment>
+      <ActionButton
+        color="primary"
+        description={t('translation|Create {{ resourceName }}', { resourceName })}
+        icon={'mdi:plus-circle'}
+        onClick={() => {
+          setOpenDialog(true);
+        }}
+      />
+
```
joaquimrocha, line 29: Nit: Remove this newline.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -62,11 +71,22 @@ export interface EditorDialogProps extends DialogProps {
   errorMessage?: string;
   /** The dialog title. */
   title?: string;
+  /** The flag for dispatching from a different component. */
+  externalDispatch?: boolean;
```
joaquimrocha, line 75: Maybe a more direct name is better: `applyOnSave`.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -36,6 +36,29 @@ export interface Cluster {
...
+    };
+  }
+}
+
+export interface ClassWithBaseObject<T> {
+  new (...args: any[]): T;
+  kind?: string;
+  getBaseObject?(): KubeObjectInterface;
+}
+
```
joaquimrocha, line 61: This shouldn't be needed. The getBaseObject should be added to the KubeObject class in cluster.ts .

frontend/src/lib/k8s/cluster.ts
```diff
@@ -36,6 +36,29 @@ export interface Cluster {
   [propName: string]: any;
 }
 
+export class BaseKubeObject {
+  static getBaseObject(): KubeObjectInterface {
+    return {
+      apiVersion: 'v1',
+      kind: '',
```
joaquimrocha, line 43: Once this is moved to the cluster.ts's KubeObject, you should fill this will the appropriate values from the endpoint.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -36,6 +36,29 @@ export interface Cluster {
...
+      metadata: {
+        name: '',
+        namespace: '',
```
joaquimrocha, line 46: This shouldn't be needed in the template because when the resource needs a namespace and doesn't have one, the default one for the cluster will be used.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -36,6 +36,29 @@ export interface Cluster {
...
+        annotations: {},
+        labels: {},
+        creationTimestamp: '',
+        uid: '',
```
joaquimrocha, line 50: These are not be needed.

frontend/src/lib/k8s/configMap.ts
```diff
@@ -1,12 +1,21 @@
 import { apiFactoryWithNamespace } from './apiProxy';
-import { KubeObjectInterface, makeKubeObject, StringDict } from './cluster';
+import { BaseKubeObject, KubeObjectInterface, makeKubeObject, StringDict } from './cluster';
 
 export interface KubeConfigMap extends KubeObjectInterface {
   data: StringDict;
 }
 
-class ConfigMap extends makeKubeObject<KubeConfigMap>('configMap') {
+class ConfigMap extends makeKubeObject<KubeConfigMap>('configMap') implements BaseKubeObject {
```
joaquimrocha, line 8: This shouldn't be needed when you add the getBaseObject to the KubeConfig class.

joaquimrocha (changes requested):

frontend/src/components/common/CreateResourceButton.tsx
skoeva: Removed

frontend/src/components/common/Resource/EditorDialog.tsx
skoeva: Changed the name and documented this

frontend/src/lib/k8s/cluster.ts
skoeva: Moved the function and removed the unneeded changes here

frontend/src/lib/k8s/cluster.ts
skoeva: Removed

frontend/src/lib/k8s/configMap.ts
skoeva: Removed

frontend/src/lib/k8s/cluster.ts
skoeva: apiVersion is now filled from the endpoint and kind is filled by the resource class since the KubeObject class's `className` is not valid by default

frontend/src/components/common/CreateResourceButton.tsx
skoeva: Removed the additional check here

app/electron/plugin-management.ts
```diff
@@ -52,6 +52,22 @@ interface PluginData {
   artifacthubVersion: string;
 }
 
+/**
+ * Move directories from currentPath to newPath by copying.
```
illume, line 56: I think a merge issue here too. I don't think this should be in this PR.

illume (changes requested):
Looks like there's some unrelated changes I think from a messed up merge in this PR.

app/electron/plugin-management.ts
skoeva: Thanks, removed these changes (I noticed a couple of other ones as well that should not have been in there) and rebased against main.

frontend/src/lib/k8s/lease.ts
```diff
@@ -14,6 +14,22 @@ export interface KubeLease extends KubeObjectInterface {
 
 export class Lease extends makeKubeObject<KubeLease>('Lease') {
   static apiEndpoint = apiFactoryWithNamespace('coordination.k8s.io', 'v1', 'leases');
+  static kind = 'Lease';
```
joaquimrocha, line 17: I don't think we should repeat the kind here. The kind should correspond to the objectName we pass to it. There are a couple of places where this is not the case, but we can rectify those. In those cases where it doesn't match, please use the kind hardcoded in the getBaseObject instead of creating a new static var.

frontend/src/lib/k8s/secret.ts
```diff
@@ -8,6 +8,16 @@ export interface KubeSecret extends KubeObjectInterface {
...
+  static kind = 'Secret';
+
+  static getBaseObject(): KubeSecret {
+    const baseObject = super.getBaseObject() as KubeSecret;
+    return {
+      ...baseObject,
+      kind: Secret.kind,
+      metadata: { ...baseObject.metadata, name: 'base-secret' },
+    };
+  }
```
joaquimrocha, line 20: I was hoping we could make things work without having to override this function everywhere (but only in cases where e.g. the objectName is not matching the kind).

joaquimrocha (changes requested):
Gave it another go. A few things need to be improved but is getting there.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -273,6 +321,36 @@ export default function EditorDialog(props: EditorDialogProps) {
       setError(t("Error parsing the code. Please verify it's valid YAML or JSON!"));
       return;
     }
+
+    const newItemDefs = obj!;
+
+    if (!applyOnSave) {
```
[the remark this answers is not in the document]
skoeva: This is essentially checking whether or not to use an external dispatch (formerly called `externalDispatch`). The name might be a bit confusing so I think it could be changed to something closer to what it was originally, but setting this to true means no dispatch is performed when creating a resource and 2 dispatches are performed when editing

frontend/src/lib/k8s/lease.ts
skoeva: Removed the static vars and introduced a new parameter for `getBaseObject()` in cluster.ts to take the kind from

frontend/src/components/common/Resource/ResourceListView.tsx
```diff
@@ -23,6 +24,10 @@ export default function ResourceListView<ItemType>(
   const { title, children, headerProps, ...tableProps } = props;
   const withNamespaceFilter =
     'resourceClass' in props && (props.resourceClass as KubeObject)?.isNamespaced;
+  const resourceClass =
+    'resourceClass' in props && props.resourceClass.hasOwnProperty('getBaseObject')
+      ? props.resourceClass
+      : null;
```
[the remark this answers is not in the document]
skoeva: Added this change

frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { KubeObject } from '../../lib/k8s/cluster';
+import { ActionButton, EditorDialog } from '../common';
+
+type Class<T> = new (...args: any[]) => T;
+
+export interface CreateResourceButtonProps {
+  resourceClass: Class<KubeObject>;
```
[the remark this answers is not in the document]
skoeva: Done

frontend/src/lib/k8s/secret.ts
joaquimrocha: What I mean is that I think there should be a way for us to have this logic abstract enough in the cluster.ts that it adapts correctly to most of the subclasses. We should definitely not move individual resources' logic into the cluster.ts file.

joaquimrocha:
@skoeva , I added a commit to illustrate what I meant: you don't have to pass the objectName to the getBaseObject, you already have it as a base static variable, and if you use it in the cluster.ts's getBaseObject, but then call the method on e.g. ConfigMap, it'l use the objectName from the latter (OOP). I also added a new static getter kind, since it should be same for the entire class.

As I mentioned, not all of the objectNames match exactly what the kind is, but they should. I just changed the objectName for ConfigMap and Secret in my patch, but maybe we should rather override the kind getter in those classes, to avoid breaking any reliance from plugins on the resource classes' objectNames.

I also moved the spreading of the props for the section header's title to the end of the component, so classes can override the default behavior. Otherwise this was breaking things like the Create button for the Namespace resource, which is a special behavior, rather than opening the editor.

Let me know if this clarifies what I was trying to convey.

skoeva:
@joaquimrocha Makes sense, I get what you mean. Also tweaked the labels to make sure they were all capitalized like how you adjusted Secret and ConfigMap. One thing I noticed: I think it might be good to fix the label size to prevent things like this when using the app:

image

But this can be resolved in a different PR.

illume (approved):
Great! Thanks.

(I'll leave it open to give a chance for Joaquim to review your latest changes)

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -49,12 +59,12 @@ import SimpleEditor from './SimpleEditor';
 type KubeObjectIsh = Partial<KubeObjectInterface>;
 
 export interface EditorDialogProps extends DialogProps {
-  /** The object to edit, or null to make the dialog be in "loading mode". Pass it an empty object if no contents are to be shown when the dialog is first open. */
-  item: KubeObjectIsh | null;
+  /** The object(s) to edit, or null to make the dialog be in "loading mode". Pass it an empty object if no contents are to be shown when the dialog is first open. */
+  item: KubeObjectIsh | object | object[] | string | null;
```
sniok, line 63: minor nit: If the only string value this prop can accept is 'default' then it should be represented in its' type
`((...args: any[]) => void) | "default" | null`

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -49,12 +59,12 @@ import SimpleEditor from './SimpleEditor';
...
   /** Called when the dialog is closed. */
   onClose: () => void;
-  /** Called when the user clicks the save button. */
-  onSave: ((...args: any[]) => void) | null;
+  /** Called by a component for when the user clicks the save button. When set to "default", internal save logic is applied. */
+  onSave?: ((...args: any[]) => void) | string | null;
```
sniok, line 67: minor nit: If the only string value this prop can accept then it should be represented in its' type
`((...args: any[]) => void) | "default" | null`

frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -36,6 +36,7 @@ export default function CreateButton(props: CreateButtonProps) {
   const dispatchCreateEvent = useEventCallback(HeadlampEventType.CREATE_RESOURCE);
   const clusters = useClusterGroup();
   const [targetCluster, setTargetCluster] = React.useState(clusters[0] || '');
+  const [currentItem, setCurrentItem] = React.useState({}); // eslint-disable-line no-unused-vars
```
joaquimrocha, line 39: I don't see the setCurrentItem being used here. Is this why you disabled the eslint? You could work around this by not declaring the setCurrentItem. But if that's the case, is there a need for using the state at all? Could you use a {} instead?

joaquimrocha (changes requested):
I have spotted something simple that needs to be improved in the code, but I have tested the feature again (creating a new item) and there's an inconsistency against the normal Create button:
When there's a failure to create a resource, the creation dialog should be brought back up again, and it's not in the case of this creation UI.

Steps to reproduce:
1. Start creating a Node from the new + button
2. Make sure you add an invalid name like my-super-node-
3. Verify that it silently fails: no notification, and the dialog is not brought up. When then opening the dialog again, the error is there

frontend/src/components/common/Resource/CreateButton.tsx
skoeva: changed this to a plain useRef and works perfectly

joaquimrocha (approved):

PR #1999, opened 2024-05-29
frontend statefulset/list: fix List view broken

yuriscott:
I found there is a error that will crash the StateFulSet/List 
`
          label: t('Containers'),
          getValue: statefulSet => statefulSet.containerNames.join(', '),
          render: statefulSet => {

`
seem like statefulSet object has no property named 'containerNames'

So I change the code to 
`
getValue: statefulSet => statefulSet.getContainers().map(it => it.name).join(', '),
`

### DIFF
--- a/frontend/src/components/statefulset/List.tsx
+++ b/frontend/src/components/statefulset/List.tsx
@@ -35,7 +35,7 @@ export default function StatefulSetList() {
         {
           id: 'containers',
           label: t('Containers'),
-          getValue: statefulSet => statefulSet.containerNames.join(', '),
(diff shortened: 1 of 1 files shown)


sniok:
good catch, thanks!

@joaquimrocha we should include this in the upcoming release

sniok (approved):

joaquimrocha (approved):

joaquimrocha:
Closing in favor of #2002 . It's the same author/changes, just formatted. Thank you so much @yuriscott !

Issue #1971, opened 2024-05-10
Add more info to containers


joaquimrocha:
As pointed out in #1929 , we should add the following info to the pods:

Pod List:
- [x] #2689

Pod Details:
- [x] Add exit code for containers
- [x] Add started/finished times
- [x] Add restart reason

feym78:
@joaquimrocha Hi
for point "Pod Details (for those with completed containers)" I would say not only with complited but also for running, the container could have been killed and restarted, but is currently running so it is also useful to get restart reason

joaquimrocha:
Reopening for tackling the last item.

illume:
Most parts are done, but there's one item left open. Moving this to the next release milestone.

illume:
Last item was done, so closing.

PR #2000, opened 2024-05-29
frontend: Add last state to container info

joaquimrocha:
Users want to know e.g. why a container was restarted.

fixes [#1971](https://github.com/headlamp-k8s/headlamp/issues/1971)


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -645,60 +645,87 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+          statusType = 'error';
+          label = t('translation|Error');
+        }
+      }
+
+      return [stateDetails, label, statusType];
+    }, [state]);
+
+    const tooltipID = React.useMemo(
+      () => 'container-state-message-' + (container?.name ?? ''),
```
sniok, line 710: useMemo is a bit overkill for this, it's not that expensive

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -645,60 +645,87 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
-    if (!status || !container) {
-      return undefined;
-    }
+  function ContainerStatusLabel(props: {
+    state?: KubeContainerStatus['state'] | null;
+    container?: KubeContainer;
+  }) {
+    const { state, container } = props;
+
+    const [stateDetails, label, statusType] = React.useMemo(() => {
```
sniok, line 679: also here useMemo is not necessary

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -709,6 +736,40 @@ export function ContainerInfo(props: ContainerInfoProps) {
...
+    }, [rows]);
+
+    if (rowsToDisplay.length === 0) {
+      return null;
+    }
+
+    return (
+      <Box>
+        {rowsToDisplay.map(({ name, value }) => {
+          const id = useId('status-value-');
```
sniok, line 755: this violates rule of hooks "Do not call Hooks inside conditions or loops.", if number of rows changes then it will complain that component has different amount of hook calls than in previous render

frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -709,6 +736,40 @@ export function ContainerInfo(props: ContainerInfoProps) {
     );
   }
 
+  function StatusValue(props: {
```
sniok, line 739: While I like having related components near each other in a single file, this file is more than 1000 lines and growing, maybe it's a good time to move something out of this file

sniok (changes requested):

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: You are right. I got excited.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: Well, this one uses translation and all. I know it's not a lot of gains, but I feel like we should be processing all this every time there's a slight update.

frontend/src/components/common/Resource/Resource.tsx
joaquimrocha: Good catch. I will move it outside.

frontend/src/components/common/Resource/Resource.tsx
sniok: yeah I'm not sure of the `t` function call cost. in theory we can move it out of component somewhere at the top level but it doesn't matter that much

sniok (approved):

sniok:
congrats on 2000 PRs 🎉

PR #2003, opened 2024-05-29
app: Fix electron opening browser on page reload

sniok:
Currently in the app on windows when application triggers page reload it opens a browser.

The reason is difference in forward and back slashes in the url. 
in main.ts startUrl has forward slashes
but when page reloads, it requests a link with backward slashes. 

Since windows handles forward slashes just fine, I added a replacement that changes to forward slashes.

Fixes https://github.com/headlamp-k8s/headlamp/issues/1990

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -506,7 +506,7 @@ function startElecron() {
       frontendPath = path.join(process.resourcesPath, 'frontend', 'index.html');
     }
 
-    const startUrl =
+    const startUrl = (
       process.env.ELECTRON_START_URL ||
       url.format({
         pathname: frontendPath,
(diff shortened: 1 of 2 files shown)


sniok:
Oh and the code that acutally opened the browser is in main.ts:574
```typescript
      if (url.startsWith(startUrl)) {
        return;
      }
      event.preventDefault();
      shell.openExternal(url);
    });
```

app/electron/main.ts
```diff
@@ -515,7 +515,8 @@ function startElecron() {
         query: {
           backendToken: backendToken,
         },
-      });
+      })
+    ).replace(/\\/g, '/');
```
joaquimrocha, line 519: Can you add a comment explaining why this is needed?

joaquimrocha (changes requested):

app/electron/main.ts
sniok: done

joaquimrocha (approved):

illume (approved):
👍

PR #2009, opened 2024-05-29
Fix listing resources + CRD edition/deletion

joaquimrocha:
One of the patches I added recently fixed problems for the CRDs, but introduced some others for listing any resources. This fixes those issues and does it in a safer way.


vyncent-t:
All tabs are working again

Issue #1926, opened 2024-04-24
charts: Add a values schema


illume:
It was mentioned here https://github.com/headlamp-k8s/headlamp/issues/1897 to take a look at the bitnami charts for postgresql.

The have some docs  https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L1635
Also they have a values.schema.json https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.schema.json

So this issue is about adding a values.schema.json JSON schema file.
https://helm.sh/docs/topics/charts/#schema-files
https://medium.com/@hans.knechtions/6-tips-to-writing-a-great-helm-chart-35899b00a6cc

This would help people use the headlamp chart, by validating the user input.

---

Additionally, on artifact hub one of the quality badges is missing because it does not use a values schema.

![image](https://github.com/headlamp-k8s/headlamp/assets/9541/f28c3f53-70ea-4609-bc2b-181181cee6ab)

knrt10:
Thank you for creating the issue. Working on this

engnatha:
Adding this schema reduced some flexibility that was previously useful. Things like https://github.com/kubernetes-sigs/headlamp/issues/2022#issuecomment-2143334793 are no longer supported. This was useful to avoid having to specify all the details for oidc configuration in an external secret. Most of the items that go into configuring OIDC aren't actual secrets.

knrt10:
Thank you for pointing it out. Will try to figure out how we can avoid it and have the same flexibility as before

PR #2025, opened 2024-06-03
charts: Add schema for values.yaml

knrt10:
Charts did not have defined structure for Headlamp value configurations. Now we have defined schema for the same which will run while linting, installing, upgrading or templating the helm charts.

Fixes: #1926

### DIFF
--- a/charts/headlamp/values.yaml
+++ b/charts/headlamp/values.yaml
@@ -84,7 +84,7 @@ config:
   # -- directory to look for plugins
   pluginsDir: "/headlamp/plugins"
   # Extra arguments that can be given to the container. See charts/headlamp/README.md for more information.
-  extraArgs: {}
+  extraArgs: []
 
 # -- An optional list of environment variables
 # env:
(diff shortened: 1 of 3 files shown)


charts/headlamp/values.yaml
```diff
@@ -84,7 +84,7 @@ config:
   # -- directory to look for plugins
   pluginsDir: "/headlamp/plugins"
   # Extra arguments that can be given to the container. See charts/headlamp/README.md for more information.
-  extraArgs: {}
+  extraArgs: []
```
joaquimrocha, line 87: This doesn't seem related to the patch.

joaquimrocha (changes requested):

charts/headlamp/values.yaml
knrt10: Fixed in a seperate commit. extraArgs should indeed be an array rather than an object. Since array can also contain object values.

joaquimrocha (approved):

Issue #2033, opened 2024-06-10
Setting BackLink to null does not work


skoeva:
When trying to remove the backLink from the `MainInfoSection` component, setting it to null does not work. We can currently set it to the empty string "", but adding the boolean null could make it more obvious that we do not want to use the backLink. We should account for null in this case.

![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/00811a81-4517-46b0-9a93-94d63894bfdc)
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/d29f61e0-5d4c-4560-9dd3-42c4b14a76c4)

PR #2034, opened 2024-06-10
frontend: MainInfoSection: Correctly handle null backLink

skoeva:
This fix explicitly handles the null case in `getBackLink()` to ensure that no back button will be shown, as defined in MainInfoSectionProps.

Fixes: #2033

### Testing
- [ ] Go to `frontend/src/components/crd/Details.tsx` and set backLink to null in the `MainInfoSection` component
- [ ] Open a cluster in Headlamp (`npm start`) and navigate to the Custom Resources page
- [ ] Click on a CRD and ensure that the back button does not appear on the CRD page

### DIFF
--- a/frontend/src/components/common/SectionBox.tsx
+++ b/frontend/src/components/common/SectionBox.tsx
@@ -33,7 +33,7 @@ export function SectionBox(props: SectionBoxProps) {
 
   return (
     <>
-      {actualBackLink !== undefined && <BackLink to={actualBackLink} />}
+      {actualBackLink && <BackLink to={actualBackLink} />}
       <Box py={0} {...outterBoxProps}>
         {title && titleElem}
         <Box
--- a/frontend/src/components/App/Notifications/__snapshots__/List.List.stories.storyshot
+++ b/frontend/src/components/App/Notifications/__snapshots__/List.List.stories.storyshot
@@ -1,22 +1,5 @@
 <div>
-  <button
-    class="MuiButtonBase-root MuiButton-root MuiButton-text MuiButton-textPrimary MuiButton-sizeSmall MuiButton-textSizeSmall MuiButton-colorPrimary MuiButton-root MuiButton-text MuiButton-textPrimary MuiButton-sizeSmall MuiButton-textSizeSmall MuiButton-colorPrimary css-w7fpvq-MuiButtonBase-root-MuiButton-root"
-    tabindex="0"
-    type="button"
-  >
-    <span
-      class="MuiButton-icon MuiButton-startIcon MuiButton-iconSizeSmall css-y6rp3m-MuiButton-startIcon"
-    />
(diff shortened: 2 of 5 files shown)


illume:
I think instead just not adding the `backLink` would work for this(or `backLink=false`. If no backLink is supplied, then no back link is used. 

Can you please explain a bit more why this is necessary?

skoeva:
> I think instead just not adding the `backLink` would work for this(or `backLink=false`). If no backLink is supplied, then no back link is used.
> 
> Can you please explain a bit more why this is necessary?

@illume The backLink value is by default the empty string `''`, and the spec for backLink specifies a difference between passing in the empty string and a null value:

`/** The route or location to go to. If it's an empty string, then the "browser back" function is used. If null, no back button will be shown. */`

Previously passing in a null value did the same thing as passing in the empty string. Unless this is the intended behavior, we are missing the functionality to handle the null string as specified

joaquimrocha:
Yeah. By default we have had the backLink, so we should just fix the null case which is what docs mean. The backLink={false} should also be the same as null.

joaquimrocha (changes requested):
My comment was in the context of MainInfoSection. You need to account for `null` there, but checking SectionBox, where it is then used, we shouldn't update it to receive null but instead just pass false to it from MainInfoSection as that's what null means.

skoeva:
Ran into merge conflict issues...reopening here #2040

joaquimrocha:
We shouldn't close PRs because of merge conflicts. I am reopening this and closing the new one since this one has the context from previous reviews.

frontend/src/components/common/SectionBox.tsx
```diff
@@ -33,7 +33,7 @@ export function SectionBox(props: SectionBoxProps) {
 
   return (
     <>
-      {actualBackLink !== undefined && <BackLink to={actualBackLink} />}
+      {actualBackLink && <BackLink to={actualBackLink} />}
```
joaquimrocha, line 36: The behavior will be different from before if you add this. Is this needed for the latest code to work?

joaquimrocha (changes requested):

frontend/src/components/common/SectionBox.tsx
skoeva: they should be functionally equivalent but it isn't necessary, reverting back

vyncent-t:
you could likely drop the `MainInfoSection` part to make it `frontend: Fixes handle null backLink`

joaquimrocha:
> this is a nit but the `MainInfoSection` part could be left off to make it `frontend: Fixes handle null backLink`

Why? I think the only thing not according to format is the first `:`. Normally we do `frontend MainInfoSection: MESSAGE`

joaquimrocha (approved):

PR #2035, opened 2024-06-11
Remove unnecessary package.json and package-lock from root dir

ashu8912:


skoeva (approved):
These were generated automatically when working on the apiProxy tests, if they aren't necessary to pass CI then it seems they can safely be removed

joaquimrocha (approved):

PR #2036, opened 2024-06-11
Remove unnecessary fields from the editor

### LINKED ISSUE #2032: Edit Resource Default YAML Should Not Include Non-Editable Fields (status, managedFields)
### Problem Description

When editing a Custom Resource Definition (CRD) in Headlamp, we have observed that certain sections in the YAML, such as `status` and `managedFields`, are not editable. 
Including these fields in the YAML editor is not helpful as they are not meant to be modified manually and would generate errors if edited.

### Proposed Solution

To enhance the user experience and prevent potential errors, it would be beneficial to exclude non-editable fields like `status` and `managedFields` from the default YAML editor view when editing a CRD.

### Kubernetes Documentation References

- **Status Field**: The `status` field is used by the Kubernetes control plane to indicate the current state of the resource. It should not be edited manually by users. For more details, refer to the Kubernetes API Conventions.
- **ManagedFields Field**: The `managedFields` field is used to track changes to the resource for server-side apply.
More information can be found in the Server-Side Apply Documentation.

There was an issue on the kubernetes repo, but the change haven’t been made server side, so changes have to be made on the client size :

https://github.com/kubernetes/kubernetes/issues/90066

https://github.com/kubernetes/kubernetes/issues/90066#issuecomment-814461277

Guilamb:
A first draft to resolve #2032. I believe there might be a better way to do this.

I've removed the Status and ManagedFields fields from the code when making the editor dialog.

Fixes: #2032

### DIFF
--- a/frontend/src/lib/k8s/cluster.ts
+++ b/frontend/src/lib/k8s/cluster.ts
@@ -1,4 +1,5 @@
 import { OpPatch } from 'json-patch';
+import { JSONPath } from 'jsonpath-plus';
 import { cloneDeep, unset } from 'lodash';
 import React from 'react';
 import helpers from '../../helpers';
@@ -329,6 +330,11 @@ export interface AuthRequestResourceAttrs {
   group?: string;
   verb?: string;
 }
+type JsonPath<T> = T extends object
+  ? {
+      [K in keyof T]: K extends string ? `${K}` | `${K}.${JsonPath<T[K]>}` : never;
+    }[keyof T]
+  : never;
 
 // @todo: uses of makeKubeObject somehow end up in an 'any' type.
 
@@ -343,6 +349,7 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
   class KubeObject {
     static apiEndpoint: ReturnType<typeof apiFactoryWithNamespace | typeof apiFactory>;
     jsonData: T | null = null;
+    public static readOnlyFields: JsonPath<T>[];
     private readonly _clusterName: string;
 
     constructor(json: T) {
@@ -428,6 +435,26 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
 
     static get isNamespaced() {
       return this.apiEndpoint.isNamespaced;
+    }
+
+    getEditableObject() {
+      const fieldsToRemove = this._class().readOnlyFields;
+      const code = this.jsonData ? cloneDeep(this.jsonData) : {};
+
+      fieldsToRemove?.forEach((path: JsonPath<T>) => {
+        JSONPath({
+          path,
+          json: code,
+          callback: (result, type, fullPayload) => {
+            if (fullPayload.parent && fullPayload.parentProperty) {
+              delete fullPayload.parent[fullPayload.parentProperty];
+            }
+          },
+          resultType: 'all',
+        });
+      });
+
+      return code;
     }
 
     // @todo: apiList has 'any' return type.
--- a/frontend/src/lib/k8s/crd.ts
+++ b/frontend/src/lib/k8s/crd.ts
@@ -52,6 +52,7 @@ class CustomResourceDefinition extends makeKubeObject<KubeCRD>('crd') {
     ['apiextensions.k8s.io', 'v1', 'customresourcedefinitions'],
     ['apiextensions.k8s.io', 'v1beta1', 'customresourcedefinitions']
   );
+  static readOnlyFields = ['metadata.managedFields'];
 
   static get className(): string {
     return 'CustomResourceDefinition';
--- a/frontend/src/components/common/Resource/EditButton.tsx
+++ b/frontend/src/components/common/Resource/EditButton.tsx
@@ -108,7 +108,7 @@ export default function EditButton(props: EditButtonProps) {
       />
       {openDialog && (
         <EditorDialog
-          item={item.jsonData}
+          item={item.getEditableObject()}
           open={openDialog}
           onClose={() => setOpenDialog(false)}
           onSave={handleSave}


joaquimrocha:
Hey @Guilamb . 
I think a more extensible way we could implement this is to have a function called `getEditableObject` in the `KubeObject` class. 
This alone would allow flexibility enough that each class can just implement what the editable version of their objects looks like.

However, we can make this even more powerful by adding a getter called `readOnlyFields` which returns a list of `jsonPath` paths. So we use this list in the base `getEditableObject` to remove any read-only fields. Meaning that k8s classes just have to define the list of readonly fields.

This way, we have enough flexibility to define which fields are readonly and also what the editable object looks like.

Can you implement this?

I am sure that there's a way to extract which of these fields are read-only from the server API, but we can connect something like that later.

Guilamb:
Hi @joaquimrocha,

I'll try to implement this.

To get the read-only fields, the solution I've found so far is to get the OpenAPI schema and check if there is a "read-only" in the description.

I've generated a Python script to test this solution, and it seems to work. However, if someday someone writes in a description that a field is "not read-only," it would still be added to the read-only fields 🫤. I'm not sure how likely this scenario is to happen.https://gist.github.com/Guilamb/aecd34fe815d60a343dfe57fd795fda3

Additionally, metadata.managedFields doesn't appear because the description states:
``` Markdown
ManagedFields maps workflow-id and version to the set of fields that are managed by that workflow. 
This is mostly for internal housekeeping, and `users typically shouldn't need to set or understand this field.`
 A workflow can be the user's name, a controller's name, or the name of a specific apply path like \"ci-cd\". 
The set of fields is always in the version that the workflow used when modifying the object.
```

I've looked at how other tools have handled this problem, and from what I saw, for the managedFields they are simply removed from display. I've looked at Backstage (https://github.com/backstage/backstage/issues/14177) and kubectl (https://github.com/kubernetes/kubernetes/pull/96878).

joaquimrocha:
Thanks. It's fine to go without checking the API server. The solution I proposed with the 2 functions in the base class + overriding them in classes that need it, should be flexible enough.

frontend/src/lib/k8s/crd.ts
```diff
@@ -52,6 +52,7 @@ class CustomResourceDefinition extends makeKubeObject<KubeCRD>('crd') {
     ['apiextensions.k8s.io', 'v1', 'customresourcedefinitions'],
     ['apiextensions.k8s.io', 'v1beta1', 'customresourcedefinitions']
   );
+  readOnlyFields = ['managedFields'];
```
joaquimrocha, line 55: Should be given as JSONPath.

joaquimrocha (changes requested):
Left a few comments.
Feel free also to squash the commits + force push.

Guilamb:
I will squash those commits, if the changes made are correct

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -62,6 +62,7 @@ export default function EditorDialog(props: EditorDialogProps) {
   const themeName = getThemeName();
 
   const originalCodeRef = React.useRef({ code: '', format: item ? 'yaml' : '' });
+
```
joaquimrocha, line 65: Please do not add changes like this that are unrelated.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -81,7 +82,6 @@ export default function EditorDialog(props: EditorDialogProps) {
     localStorage.setItem('useSimpleEditor', JSON.stringify(data));
     setUseSimpleEditorState(data);
   }
-
```
joaquimrocha, line 84: Please do not add changes like this that are unrelated.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -429,6 +443,38 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
     static get isNamespaced() {
       return this.apiEndpoint.isNamespaced;
     }
+    static get readOnlyFields(): JsonPath<T>[] {
```
joaquimrocha, line 446: Add a newline before this function.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -429,6 +443,38 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+      return this._readOnlyFields;
+    }
+    getEditableObject() {
```
joaquimrocha, line 449: Add a newline before this function.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -429,6 +443,38 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+      // Recursively remove fields
+      /*      (function recurse(currentField: any) {
+        for (const key in currentField) {
+          if (fieldsToRemove.includes(key)) {
+            delete currentField[key];
+          } else if (typeof currentField[key] === 'object' && currentField[key] !== null) {
+            recurse(currentField[key]);
+          }
+        }
+      })(code);*/
```
joaquimrocha, line 461: Please remove unused code instead of commenting it out.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -429,6 +443,38 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
     static get isNamespaced() {
       return this.apiEndpoint.isNamespaced;
     }
+    static get readOnlyFields(): JsonPath<T>[] {
+      return this._readOnlyFields;
+    }
+    getEditableObject() {
+      const fieldsToRemove = this._class().readOnlyFields;
+      const code = this.jsonData ? cloneDeep(this.jsonData) : '{}';
```
joaquimrocha, line 451: The type of code will be different depending on whether we have jsonData or not (i.e. the else part should be {}, not '{}').

frontend/src/lib/k8s/cluster.ts
```diff
@@ -329,6 +330,18 @@ export interface AuthRequestResourceAttrs {
...
+  : {
+      [K in keyof T]: T[K] extends never
+        ? never
+        : T[K] extends Record<string | number | symbol, unknown>
+        ? K extends string
+          ? `${K}.${JsonPath<T[K]>}` | K
+          : never
+        : K;
+    }[keyof T] &
+      string;
```
joaquimrocha, line 344: Do we need all this typing? Maybe @sniok can help with simplifying this.

joaquimrocha (changes requested):
Thanks @Guilamb . It's almost there. Please check my review.

Guilamb:
Thank you for the feedback, I've tried another way to define JsonPath type, to make it shorter.
If every things ok, just send me a message by slack or put a comment here, and I'll squash all the commit together 👍

frontend/src/lib/k8s/cluster.ts
```diff
@@ -430,6 +437,30 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
       return this.apiEndpoint.isNamespaced;
     }
 
+    static get readOnlyFields(): JsonPath<T>[] {
+      return this._readOnlyFields;
+    }
+
+    getEditableObject() {
+      const fieldsToRemove = this._class().readOnlyFields;
```
joaquimrocha, line 445: We should check if fieldsToRemove is undefined as this leads to an error when trying to run the foreach later in this block.
If !fieldsToRemove, we should just return the this.jsonData.

frontend/src/lib/k8s/cluster.ts
```diff
@@ -430,6 +437,30 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
       return this.apiEndpoint.isNamespaced;
     }
 
+    static get readOnlyFields(): JsonPath<T>[] {
```
joaquimrocha, line 440: I was wrong about asking for this to be a getter, as overriding this in the subclasses, as you did, requires declaring the _readOnlyFields, which is not a good practice. Let's turn _readOnlyFields into readOnlyFields and make it public.

joaquimrocha (commented):
I found a bug in the code, and a change I requested which actually doesn't make sense. After this I think you can squash all commits as the PR is ready. Please give it a commit message like:
```
frontend: Allow KubeObjects to set which parts of them are editable

This PR adds a new var to KubeObject classes to declare fields that shouldn't be
considered when editing the files. It also sets a new getEditableObject method
that uses these fields and removes them from an object clone and is used with
the Editor.
```

It would be nice if the code that adds the fields to the CRD is a different commit, but it's simple enough that you can keep it the same. If you keep them together, please add a new paragraph after the above and mention that CRD is also getting its list of readonly fields.

joaquimrocha (approved):

PR #2040, opened 2024-06-14
frontend: MainInfoSection: Correctly handle null backLink

skoeva:
This fix explicitly handles the null case in getBackLink() to ensure that no back button will be shown, as defined in MainInfoSectionProps.

Fixes: #2033

### Testing
- [ ] Go to `frontend/src/components/crd/Details.tsx` and set backLink to null in the `MainInfoSection` component
- [ ] Open a cluster in Headlamp (`npm start`) and navigate to the Custom Resources page
- [ ] Click on a CRD and ensure that the back button does not appear on the CRD page


joaquimrocha:
Closing for reopening the previous PR, since we shouldn't close PRs due to git conflicts.

Issue #1933, opened 2024-04-27
"failed to get oidc config" spam in logs


feym78:
Hello.
After start headlamp is constantly spamming with "failed to get oidc config" message, but OIDC is not configured and no OIDC parameters passed:
![image](https://github.com/headlamp-k8s/headlamp/assets/81779189/24b601fb-3bb3-470d-b8eb-02332539bcdf)
![image](https://github.com/headlamp-k8s/headlamp/assets/81779189/a7f2f8f0-1eea-4f36-a8f6-cfe9ccae58f3)

PR #2041, opened 2024-06-14
backend: Prevent OIDC errors when not in use

skoeva:
This change prevents the spam of OIDC config errors in the logs when running the latest version of Headlamp without OIDC. Instead of logging an error when we check for the OIDC config, we simply return since we know we are not using OIDC auth.

Fixes: #1933 

### Testing
- [x] Build a custom headlamp image in minikube and point the `kubernetes-headlamp.yaml` to it
- [x] Apply the YAML (`kubectl apply -f kubernetes-headlamp.yaml`) and configure non-OIDC credentials
- [x] Run headlamp in-cluster and watch the logs for the new headlamp pod: `kubectl logs -n kube-system <pod-name> -f`
- [x] Open the app, click around, and ensure that the OIDC config errors do not appear

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -817,10 +817,7 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
 		// skip if cluster is not using OIDC auth
 		oidcAuthConfig, err := kContext.OidcConfig()
 		if err != nil {
-			logger.Log(logger.LevelError, map[string]string{"cluster": cluster},
-				err, "failed to get oidc config")
 			next.ServeHTTP(w, r)
-
 			return
 		}
 


backend/cmd/headlamp.go
```diff
@@ -815,12 +815,9 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
 		}
 
 		// skip if cluster is not using OIDC auth
-		oidcAuthConfig, err := kContext.OidcConfig()
-		if err != nil {
```
yolossn, line 819: I understand the need for removing the log but any specific reason to change the condition here?

skoeva: @yolossn Theoretically no but I thought it would be safer to confirm directly that the config was null, WDYT?

yolossn: If we are going to check for an error case we should ideally use the err check as it it provided by the function, In cases where the function doesn't return an error the null check makes sense.

skoeva: got it, will revert the condition back. tested once again and works as expected

illume:
I think rather than disabling logging of the error...  instead `OIDCTokenRefreshMiddleware` should not be used if oidc is not configured. If the `config.oidcClientID` is not set, don't add the `OIDCTokenRefreshMiddleware`.

What do you think?

skoeva:
> I think rather than disabling logging of the error... instead `OIDCTokenRefreshMiddleware` should not be used if oidc is not configured. If the `config.oidcClientID` is not set, don't add the `OIDCTokenRefreshMiddleware`.
> 
> What do you think?

@illume the same check is performed on line 499, I think removing logging after 818 makes sense since it is assumed that OIDC auth is not being used by the comment on line 817. I'm not very familiar, but is there a reason to call `OIDCTokenRefreshMiddleware` when OIDC is not configured? line 818 assumes no config so that may need to be changed

illume:
>  I'm not very familiar, but is there a reason to call OIDCTokenRefreshMiddleware when OIDC is not configured? 

I don't think there's a reason to use OIDCTokenRefreshMiddleware when oidc is not configured. /cc @yolossn

yolossn:
> I don't think there's a reason to use OIDCTokenRefreshMiddleware when oidc is not configured. /cc @yolossn

Since the middleware is for the cluster proxy endpoint we cannot conditionally add it, this is because there can be certain clusters with OIDC and other without and all of them connect to the cluster via the cluster proxy endpoint.

illume:
Thanks @yolossn

illume (approved):
🎉🎈

thanks

yolossn (approved):

PR #2042, opened 2024-06-14
frontend: Add story for MainInfoSection

joaquimrocha:
Following #2034 , I am adding this story for testing the null backlink.


joaquimrocha:
@sniok , I added this quick story for the MainInfoSection, mainly for testing the backlink. However, it looks like the story is generating changes in other snapshots. Maybe this is because I am importing the podList which comes from a different section and maybe that messes up some things?
I am mentioning you here just in case this is related to the Vite changes, since I don't remember a case like this before.

sniok:
I'll take a look

sniok:
There was an issue with snapshots, they were not deterministic sometimes, I made a PR with a fix.

skoeva (commented):
Change looks good, snapshot output was a bit confusing so appreciate the changes made in #2045

sniok:
Regenerated snapshots as a fixup commit

PR #2044, opened 2024-06-16
frontend: add default actions to ResourceListView

farodin91:
image

Now the actions can be used from the list views (by clicking the action ... to bring up a menu) as well as on the detail view.


frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -370,6 +377,41 @@ function ResourceTableContent<RowItem>(props: ResourceTableProps<RowItem>) {
...
+      action: ({ item, closeMenu }) => (
+        <EditButton item={item} menuAction afterConfirm={closeMenu} />
+      ),
+    },
+    {
+      id: DefaultHeaderAction.DELETE,
+      action: ({ item, closeMenu }) => (
+        <DeleteButton item={item} menuAction afterConfirm={closeMenu} />
+      ),
+    },
```
joaquimrocha, line 392: For read-only roles, where users cannot delete nor edit resources, this will render the menu empty AFAIU.
One quick way to avoid this is to add a `VIEW` item. So at least we have an option in the menu. It's also something I have seen as an option sometimes, maybe it even improves a11y.

frontend/src/components/common/ActionMenuItem/ActionMenuItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export interface ActionMenuItemProps {
+  /** A short description of the action. */
+  description: string;
+  /** Either a string icon, or imported icon. */
+  icon: string | IconifyIcon;
+  /** The action when it's activated. */
+  onClick: React.MouseEventHandler<HTMLLIElement>;
+  /** A longer description of the action. Used in the tooltip. */
+  longDescription?: string;
```
joaquimrocha, line 13: This overrides the description in the code, which would surprise me.
I'd prefer to have `description` and `tooltip`. And if tooltip is undefined, we use the description for it.
This way we are flexible. We could even make the description optional (and make the item text component optional) if it makes sense to have menu items with icons only.

frontend/src/components/common/ActionMenuItem/ActionMenuItem.tsx
```diff
@@ -0,0 +1,43 @@
+import { Icon, IconifyIcon, IconProps } from '@iconify/react';
+import { ListItemIcon, ListItemText, MenuItem } from '@mui/material';
+import React from 'react';
+
+export interface ActionMenuItemProps {
+  /** A short description of the action. */
+  description: string;
+  /** Either a string icon, or imported icon. */
+  icon: string | IconifyIcon;
```
joaquimrocha, line 9: You could just type this as `Icon['icon']` or `IconifyIcon` solely, in case the latter includes string there.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -136,6 +145,8 @@ export default function Table<RowItem extends Record<string, any>>({
   filterFunction,
   errorMessage,
   loading,
+  noActions,
```
joaquimrocha, line 148: I think we should call it the same as MRT: `enableRowActions` and have the same default: no actions by default.
Otherwise we're also breaking expecations as now all plugins using this table will have to set `noActions={true}` or the table will look different.

joaquimrocha (changes requested):
Incredible changes @farodin91 ! Thanks a lot.
I left some comments which I believe will improve the code. I also noticed that the menu options partially outside the view. Not sure if you can e.g. render the menu with the anchor being the top right to avoid this.
Menu partially hidden

Also. Once the review is approved, I think it would make sense to have the commits separated a bit. At least in 2: 1. Add actions capability to tables; 2. Add the view/edit/delete context actions to resource tables

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -10,17 +10,20 @@ import {
...
 import ActionButton from '../ActionButton';
+import ActionMenuItem from '../ActionMenuItem';
 import { ConfirmDialog } from '../Dialog';
 import AuthVisible from './AuthVisible';
 
 interface DeleteButtonProps {
   item?: KubeObject;
   options?: CallbackActionOptions;
+  menuAction?: Boolean;
+  afterConfirm?: () => void;
```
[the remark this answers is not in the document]
farodin91: I had the issue that delete didn't went through, if i close it with onclick.

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -47,11 +50,11 @@ export default function DeleteButton(props: DeleteButtonProps) {
             startUrl: item!.getListLink(),
             errorUrl: item!.getListLink(),
             ...options,
-          })
+          }),
         );
     },
     // eslint-disable-next-line
-    [item]
+    [item],
```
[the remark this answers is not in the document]
farodin91: i had the wrong prettier version some home.

frontend/src/components/common/Resource/DeleteButton.tsx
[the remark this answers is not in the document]
farodin91: We have to work more dispatching that would be a major refactor in this PR of all actions.

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -10,17 +10,20 @@ import {
...
 } from '../../../redux/headlampEventSlice';
 import ActionButton from '../ActionButton';
+import ActionMenuItem from '../ActionMenuItem';
 import { ConfirmDialog } from '../Dialog';
 import AuthVisible from './AuthVisible';
 
 interface DeleteButtonProps {
   item?: KubeObject;
   options?: CallbackActionOptions;
+  menuAction?: Boolean;
```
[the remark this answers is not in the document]
farodin91: I updated it a bit.

frontend/src/components/common/Resource/ResourceTable.tsx
farodin91: I added VIEW.

farodin91:
> I also noticed that the menu options partially outside the view. Not sure if you can e.g. render the menu with the anchor being the top right to avoid this. 

i used the inhouse method to render the menu.

farodin91:
@joaquimrocha I updated a lot of thinks. What do you think?

farodin91:
@joaquimrocha @illume Would you like to review?

illume (approved):
🎉 Thanks.

I tested it, and it's a really nice improvement.

joaquimrocha:
@farodin91 Another great PR! Thanks a lot, and keep those coming.

About this one, I think it's great as it is, but if we could hide the view button when the edit button is enabled (as its done through the AuthVisible comp in the details view), that would be more consistent.

PR #2045, opened 2024-06-17
frontend: Improve snapshot tests to be more deterministic

sniok:
Main change is removing "waitFor" in the test code. It would wait until rendered stuff matched the snapshot which wasn't ideal. Now it is more deterministic. Second rerender was added because material react table was misbehaving, but rerendering twice for some reason works. Also added "asFragment" as it seems to return rendered content more consistently, just using "container" would sometimes return empty div

fixes an issue with snapshots in #2042, tested by rebasing that branch on this fix


illume (approved):
👍 

thanks

PR #2054, opened 2024-06-18
frontend use action if possible

farodin91:


frontend/src/components/App/Notifications/index.tsx
```diff
@@ -93,15 +93,11 @@ function NotificationsList(props: {
           </Grid>
           {!notification.seen && (
             <Grid item md={1}>
-              <Tooltip title={t('translation|Mark as read')}>
-                <IconButton
-                  onClick={e => notificationSeenUnseenHandler(e, notification)}
-                  aria-label={t('translation|Mark as read')}
-                  size="medium"
-                >
-                  <Icon icon="mdi:circle" color={theme.palette.error.main} height={12} width={12} />
```
joaquimrocha, line 102: I think the color has not been taken into account in the replacement, right?
We do have notifications not visible for a couple of versions till we fix performance issues, but it would be good to have them look the same.

joaquimrocha (changes requested):
Looks mostly good but left a comment about a possible regression.
Also, the commit title should have `frontend:` as a prefix.
Something like this would be more descriptive:
```
frontend: Replace use of Tooltip+IconButton by Action
```

PR #2064, opened 2024-06-20
Fix helm chart ingress

BlindB0:
Fix for values.schema.json in helm chart.
With type "string" it cannot validate paths.path array in values.yaml


joaquimrocha:
Thanks @BlindB0 . I have requested @knrt10 to take a look.
Can you sign your commits so the DCO checks validate it? (`git commit --amend -s` will do the job)

PR #2071, opened 2024-06-20
fix for helm chart ingress

BlindB0:
fix for helm chart ingress


knrt10 (approved):
Thank you for the PR @BlindB0.

PR #2074, opened 2024-06-21
Broken namespace.test.ts usage

vyncent-t:
This is the source code for the failing test.

Note: The code within namespace.test.ts (located in frontend/src/lib/k8s) has been copied and put into the index.test.ts in the same directory, which does not fail when placed there.


vyncent-t:
related to the authvis not fixed

joaquimrocha:
I think this no longer applies, does it, @vyncent-t ?

vyncent-t:
> I think this no longer applies, does it, @vyncent-t ?

yea this seems to be out of date, closing

PR #2082, opened 2024-06-24
frontend: Allow editor to auto-resize in case of text size change

ashu8912:
If you to a config map view which shows data and your system text size setting is increased above 130% the monaco editor collapses, This PR allows it to resize itself on text size change


illume (approved):
Thanks.

PR #2086, opened 2024-06-26
frontend: Fix table search when some values are undefined

sniok:
Right now if you search within a table and some row cell has "undefined" as a value it crashes the page. 
This change will handle null and undefined values and replace them with a string, which table expects


sniok:
I was very surprised the snapshots changed. If you're wondering why some columns have different sorting button label, here's the reason:

https://tanstack.com/table/latest/docs/guide/sorting#sorting-direction
```
By default, the first sorting direction when cycling through the sorting for a column using the toggleSorting APIs is ascending for string columns and descending for number columns.
```

joaquimrocha (approved):

PR #2093, opened 2024-06-27
frontend Layout.tsx: Fix comparing new clusters config

joaquimrocha:
The cluster config processing in the mentioned file was testing stateless clusters as coming from the /config endpoint but that's not the case. Also, it was changing the redux state directly which led to issues.

These changes simplify this comparison and remove the check for stateless clusters.


joaquimrocha:
@knrt10 , please test with stateless clusters.

illume:
@joaquimrocha I guess the milestone for the 0.25.0 release should be removed from this.

frontend/src/components/App/Layout.tsx
```diff
@@ -56,6 +56,46 @@ function ClusterNotFoundPopup() {
...
+/**
+ * Merges the new cluster config with the current cluster config.
+ * It also preserves the useToken property.
+ * @param newConfig - The new cluster config.
+ * @param currentClusters - The current cluster config.
+ * @param statelessClusters - The stateless cluster config.
+ * @returns The merged cluster config.
+ */
+function mergeClusterConfigs(
+  newConfig: Omit<ConfigState, 'clusters'> & { clusters: Record<string, any> },
```
sniok, line 68: The only thing that's used from the `newConfig` is `cluster` property, so I think it should be replaced with just `newClusters: Record<string, any>`, then responsibility of this function is now only to merge clusters.

frontend/src/stateless/index.ts
```diff
@@ -317,44 +317,29 @@ function generateSecureToken(length = 16): string {
...
  */
-export function processClusterComparison(
-  clusters: ConfigState['clusters'],
-  clustersToConfig: ConfigState['clusters'],
-  isStateless: boolean
-) {
-  if (!clusters) {
-    // If no clusters, then add all clusters
-    return false;
+export function isClusterConfigDifferent(
```
sniok, line 320: For boolean values it is usually better to name them positive rather than negative, like `isEnabled` rather than `isDisabled`, it makes it easier to reason later on, for example `if(isEnabled)` is easier to read than `if(!isDisabled)`. 

So in this case I'd sugget naming it `isEqualClusterConfigs`

sniok (dismissed):
Looks good, left some minor suggestions

illume:
Is there an issue to link for this?

knrt10:
I am sorry but no. Shall I create one and attach it to the PR?

illume:
@knrt10 No need to create one. I was just wondering if there was an issue that we could close and notify people of the fix.

illume (approved):
🎉🎈 Thanks!

PR #2098, opened 2024-06-27
github: Adding rc- and testing-rc- branches to be tested

illume:
So we can run our tests on the rc branches as well.


knrt10 (approved):
+1

Issue #1992, opened 2024-05-27
Update deprecated packages


sniok:
When you run `npm install` there are a bunch of packages that are marked as deprecated.

We should look into updating them. It's a security issue, because it's sometimes easier to update to a security fix that might not be available without spending a lot of work upgrading.

### 90% of packages are updated to the most recent version for:

- [x] app/ `npm outdated`
- [ ] frontend/  `npm outdated`
- [x] backend/ `go list -u -m all`
- [ ] plugins/headlamp-plugin/ `npm outdated`

### Additional info
```
cd frontend && npm install
npm WARN deprecated @types/nock@11.1.0: This is a stub types definition. nock provides its own type definitions, so you do not need this installed.
npm WARN deprecated source-map-resolve@0.[6](https://github.com/headlamp-k8s/headlamp/actions/runs/9196281036/job/25294014099#step:4:7).0: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated @babel/plugin-proposal-private-property-in-object@[7](https://github.com/headlamp-k8s/headlamp/actions/runs/9196281036/job/25294014099#step:4:8).21.11: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-private-property-in-object instead.
npm WARN deprecated domexception@2.0.1: Use your platform's native DOMException instead
npm WARN deprecated w3c-hr-time@1.0.2: Use your platform's native performance.now() and performance.timeOrigin.
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated svgo@1.3.2: This SVGO version is no longer supported. Upgrade to v2.x.x.
npm WARN deprecated core-js@2.6.12: core-js@<3.4 is no longer maintained and not recommended for usage due to the number of issues. Because of the V[8](https://github.com/headlamp-k8s/headlamp/actions/runs/9196281036/job/25294014099#step:4:9) engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.
```

sniok:
It'd be nice to also make sure we don't have any unused deps. Also there's a package.json in the root for some reason, which probably shouldn't be there

illume:
I edited the PR description to add the different places we need to update deps.

[...]

PR #2104, opened 2024-06-28
backend: Update dependencies

illume:
This updates the backend dependencies that I could to their latest versions.

There are some indirect dependencies that need updating, but I think I have to wait for upstream to bump those.

Related to:
- #1992 (the update backend task)

### Testing done

Used the app locally, and clicked about.


vyncent-t:
Working on app mode!

vyncent-t:
Works on web also!

knrt10 (approved):
+1, other than the indirect dependencies that needs updating. I think it can be done in a separate PR.

PR #2111, opened 2024-07-01
Fix error boundary Page link

### LINKED ISSUE #2109: Home link in "uh-oh" page doesn't work in app mode
When in the desktop app, if the "Uh-oh" page is shown, the "Get back home" link doesn't work. On Mac it reportedly tries to open the link in Finder, and on WSL it does nothing (likely tries to open the link but fails).

This means that users hitting this problem cannot navigate away from this page.

Let's do two things:
- [ ] Fix home link in app
- [ ] Add an app menu to navigate home
- [ ] Add back/forward app menus to navigate history

ashu8912:
fixes #2109 

- [x] Fix home link in app
- [x] Add an app menu to navigate home
- [x] Add back/forward app menus to navigate history

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -10,6 +10,7 @@ import fs from 'fs';
 import { spawnSync } from 'node:child_process';
 import { userInfo } from 'node:os';
 import open from 'open';
+import { platform } from 'os';
 import path from 'path';
 import url from 'url';
 import yargs from 'yargs';
@@ -20,6 +21,33 @@ dotenv.config({ path: path.join(process.resourcesPath, '.env') });
 
 const pathInfoDebug = false;
 let pathInfo;
+
+const isDev = process.env.ELECTRON_DEV || false;
+let frontendPath = '';
+
+if (isDev) {
+  frontendPath = path.resolve('..', 'frontend', 'build', 'index.html');
+} else {
+  frontendPath = path.join(process.resourcesPath, 'frontend', 'index.html');
+}
+const backendToken = randomBytes(32).toString('hex');
+
+const startUrl = (
+  process.env.ELECTRON_START_URL ||
+  url.format({
+    pathname: frontendPath,
+    protocol: 'file:',
+    slashes: true,
+    query: {
+      backendToken: backendToken,
+    },
+  })
+)
+  // Windows paths use backslashes and for consistency we want to use forward slashes.
+  // For example: when application triggers refresh it requests a URL with forward slashes and
+  // we use startUrl to determine if it's an internal or external URL. So it's easier to
+  // convert everything to forward slashes.
+  .replace(/\\/g, '/');
 
 /**
  * On MacOS apps do not get the same environment variables as the terminal.
@@ -86,9 +114,7 @@ const args = yargs
 const isHeadlessMode = args.headless;
 let disableGPU = args['disable-gpu'];
 const defaultPort = 4466;
-const backendToken = randomBytes(32).toString('hex');
 
-const isDev = process.env.ELECTRON_DEV || false;
 const useExternalServer = process.env.EXTERNAL_SERVER || false;
 const shouldCheckForUpdates = process.env.HEADLAMP_CHECK_FOR_UPDATES !== 'false';
 const manifestDir = isDev ? path.resolve('./') : process.resourcesPath;
@@ -181,6 +207,17 @@ function quitServerProcess() {
   serverProcess.kill();
 
   serverProcess = null;
+}
+
+function getAcceleratorForPlatform(navigation: 'left' | 'right') {
+  switch (platform()) {
+    case 'darwin':
+      return navigation === 'right' ? 'Cmd+]' : 'Cmd+[';
+    case 'win32':
+      return navigation === 'right' ? 'Alt+Right' : 'Alt+Left';
+    default:
+      return navigation === 'right' ? 'Alt+Right' : 'Alt+Left';
+  }
 }
 
 function getDefaultAppMenu(): AppMenu[] {
@@ -315,11 +352,6 @@ function getDefaultAppMenu(): AppMenu[] {
       id: 'original-view',
       submenu: [
         {
-          label: i18n.t('Reload'),
-          role: 'forcereload',
-          id: 'original-force-reload',
-        },
-        {
           label: i18n.t('Toggle Developer Tools'),
           role: 'toggledevtools',
           id: 'original-toggle-dev-tools',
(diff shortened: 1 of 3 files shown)


app/electron/main.ts
```diff
@@ -319,6 +344,14 @@ function getDefaultAppMenu(): AppMenu[] {
           role: 'forcereload',
           id: 'original-force-reload',
         },
+        {
+          label: i18n.t('Go To Home Screen'),
```
joaquimrocha, line 348: Maybe just "Go to Home".

joaquimrocha: Or maybe we could have a new menu, with the options:
```
Navigate
  Reload
  Go to Home
  Go Back
  Go Forward
```

Here is an example on how to add the back/forward:
https://www.electronjs.org/docs/latest/api/browser-window#event-app-command-windows-linux

joaquimrocha (changes requested):

app/electron/main.ts
ashu8912: done

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
...
+          role: 'homescreen',
+          id: 'original-home-screen',
+          click: () => {
+            mainWindow?.loadURL(startUrl);
+          },
+        },
+        {
+          label: i18n.t('Go Back'),
+          role: 'back',
+          id: 'original-back',
```
joaquimrocha, line 358: It would be nice if this and the Go Forward were disabled when there's no way to go back/forward. This has to be done by responding to a webContents event, then fetching the menus by ID and setting them as not enabled.

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
           role: 'forcereload',
           id: 'original-force-reload',
         },
+        {
```
joaquimrocha, line 347: The new menus items should be in a new top menu "Navigate".
We should move the Reload there too. So in the end we get:
```
Navigate
  Reload
  separator
  Go to Home
  Go Back
  Go Forward
```

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
...
+          click: () => {
+            mainWindow?.loadURL(startUrl);
+          },
+        },
+        {
+          label: i18n.t('Go Back'),
+          role: 'back',
+          id: 'original-back',
+          click: () => {
+            mainWindow?.webContents.goBack();
```
joaquimrocha, line 360: Check if the webContents can go back before you call this function.

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
...
+          },
+        },
+        {
+          label: i18n.t('Go Forward'),
```
joaquimrocha, line 364: We should add an accelerator: Alt+Right
(and whatever it is in Mac)

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
...
+        {
+          label: i18n.t('Go To Home'),
+          role: 'homescreen',
+          id: 'original-home-screen',
+          click: () => {
+            mainWindow?.loadURL(startUrl);
+          },
+        },
+        {
+          label: i18n.t('Go Back'),
```
joaquimrocha, line 356: We should add an accelerator: Alt+Left
(and whatever it is in Mac)

joaquimrocha (changes requested):
Please check my comments, which I think improve the UX.

app/electron/main.ts
```diff
@@ -319,6 +344,30 @@ function getDefaultAppMenu(): AppMenu[] {
           role: 'forcereload',
           id: 'original-force-reload',
         },
+        {
+          label: i18n.t('Go To Home'),
```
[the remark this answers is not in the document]
ashu8912: done

app/electron/main.ts
ashu8912: done

app/electron/main.ts
ashu8912: done

app/electron/main.ts
ashu8912: done

app/electron/main.ts
ashu8912: done

app/electron/main.ts
ashu8912: done

frontend/src/components/App/Layout.tsx
```diff
@@ -47,7 +47,7 @@ function ClusterNotFoundPopup() {
       alignItems="center"
     >
       <Box p={0.5}>{t(`Something went wrong with cluster {{ cluster }}`, { cluster })}</Box>
-      <Button variant="contained" size="small" href="/">
+      <Button variant="contained" size="small" href={window.desktopApi ? '#' : '/'}>
```
joaquimrocha, line 50: Hmm, can you add the fixes as a different commit? Sorry I didn't ask for this before.

app/electron/main.ts
```diff
@@ -348,6 +380,46 @@ function getDefaultAppMenu(): AppMenu[] {
...
+          id: 'original-home-screen',
+          click: () => {
+            mainWindow?.loadURL(startUrl);
+          },
+        },
+        {
+          label: i18n.t('Go Back'),
+          role: 'back',
+          id: 'original-back',
+          accelerator: getAcceleratorForPlatform('Alt+Left', 'Cmd+Left', 'Alt+Left'),
```
joaquimrocha, line 405: I'd prefer to call this function with only one param: Left | Right. Then inside the function should decide what to output.
Also, did you verify if Option+Left works on Mac? See:
https://www.electronjs.org/docs/latest/api/accelerator/#platform-notice
https://support.apple.com/guide/safari/keyboard-and-other-shortcuts-cpsh003/mac (it says it's `Command+[`)

joaquimrocha (changes requested):
Works fine but I think we can improve the code a bit. Please check my comments.

frontend/src/components/App/Layout.tsx
ashu8912: but this is a different commit @joaquimrocha

app/electron/main.ts
ashu8912: updated, thanks @joaquimrocha

joaquimrocha (approved):

PR #2117, opened 2024-07-02
github: Remove redundant backend workflow

illume:
We already build and test the backend in backend-test workflow.
(and also in the app ones, and container test)


yolossn (approved):
LGTM

PR #2118, opened 2024-07-02
Deployment fixes

ashu8912:


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -985,7 +997,7 @@ export function patch(
   const opts = {
     method: 'PATCH',
     body,
-    headers: { ...JSON_HEADERS, 'Content-Type': 'application/json-patch+json' },
+    headers: { ...JSON_HEADERS, 'Content-Type': 'application/merge-patch+json' },
```
joaquimrocha, line 1000: Will this work for the other cases where it was working before? i.e. did you check just changing e.g. a label for a resource?

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -163,7 +163,9 @@ function ScaleDialog(props: ScaleDialogProps) {
               <Fab
                 size="small"
                 color="primary"
-                onClick={() => setNumReplicas(numReplicas => numReplicas - 1)}
+                onClick={() =>
+                  setNumReplicas(numReplicas => (numReplicas <= 0 ? 0 : numReplicas - 1))
+                }
```
joaquimrocha, line 168: You should also disable the button if numReplicas <= 0.

joaquimrocha (changes requested):
Looks very close to final, but left a couple of comments.

frontend/src/components/common/Resource/ScaleButton.tsx
joaquimrocha, line 1: Don't use `-ve` for negative. It reads as minusve.

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -163,7 +163,9 @@ function ScaleDialog(props: ScaleDialogProps) {
               <Fab
                 size="small"
                 color="primary"
-                onClick={() => setNumReplicas(numReplicas => numReplicas - 1)}
+                onClick={() =>
+                  setNumReplicas(numReplicas => (numReplicas <= 0 ? 0 : numReplicas - 1))
```
joaquimrocha, line 167: You could use Math.max here.

frontend/src/lib/k8s/apiProxy.ts
ashu8912: Yupp i think in this is just about patching what we send otherwise we would have to add json operations as well if we use the older way which i am not sure if is supported

frontend/src/components/common/Resource/ScaleButton.tsx
ashu8912: -ve???

ashu8912: aah the commit message

frontend/src/components/common/Resource/ScaleButton.tsx
ashu8912: done

frontend/src/components/common/Resource/ScaleButton.tsx
ashu8912: done

frontend/src/lib/k8s/cluster.ts
```diff
@@ -686,18 +686,19 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
 
       type ApiEndpointWithScale = {
         scale: {
-          put: (
-            data: { metadata: KubeMetadata; spec: { replicas: number } },
```
joaquimrocha, line 690: Hmm, I just noticed that this is part of the lib which may be used by plugins. So you should keep the put but just make it do the same thing as patch.

frontend/src/components/common/Resource/ScaleButton.tsx
```diff
@@ -163,16 +163,21 @@ function ScaleDialog(props: ScaleDialogProps) {
...
               </Fab>
               <Input
                 type="number"
                 value={numReplicas}
                 sx={{ marginLeft: '6px', marginRight: '6px' }}
-                onChange={e => setNumReplicas(Number(e.target.value))}
+                onChange={e =>
+                  Number(e.target.value) <= 0
+                    ? setNumReplicas(0)
+                    : setNumReplicas(Number(e.target.value))
```
joaquimrocha, line 179: Nit: Could be simpler by doing `setNumReplicas(Math.max(0, Number(e.target.value)))`

joaquimrocha (changes requested):
Noticed an API break. After that, it should be fine.

joaquimrocha (approved):

PR #2122, opened 2024-07-02
charts: Fix expected templates for new headlamp version

illume:
This fixes the broken CI.

(A more long term fix would be to make it so it ignores version numbers somehow. But this is a quick fix to fix CI).

### How to test

- locally `make helm-template-test`
- Also... the chart test CI Check passes ("Template test Helm charts / lint-test")


knrt10 (approved):
+1.

Will try to fix this for the long run. Thank you for the PR

PR #2124, opened 2024-07-02
plugins/examples: Add .eslintcache file to gitignores

illume:
After running lint the .eslintcache file is made. It should be ignored by git.

### how to test

After this PR, there should be nothing reported as changed by git.

```
npm i
npm run lint
git status
```


sniok:
why not add it to the root .gitignore once?

illume:
> why not add it to the root .gitignore once?

The examples should be self contained for when someone copies the folder to make their own.

sniok (approved):

Issue #2125, opened 2024-07-02
frontend coverage issues


illume:
- [x] It is including node_modules and other irrelevant files when I run it locally.
- [x] The file names can not be read, as they are cut off. ( An example name: "...g.stories.tsx" )
- [x] Is it now a lot higher percentage than before vitest? Which one is correct?
- [x] seems there is no mocking of API requests anymore. Before we had some API requests mocked, so now there are more console error messages when running the tests
- [x] tests run twice for some reason when run locally?

### cut off file names

The coverage report from CI https://github.com/headlamp-k8s/headlamp/actions/runs/9759274465/job/26935598959?pr=2111#step:5:3586

```
 Test Files  21 passed | 1 skipped (22)
      Tests  530 passed | 1 skipped (531)
   Start at  10:26:07
   Duration  31.75s (transform 4.81s, setup 3.44s, collect 25.14s, tests 21.85s, environment 12.25s, prepare 2.92s)

 % Coverage report from v8
-------------------|---------|----------|---------|---------|-------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------|---------|----------|---------|---------|-------------------
All files          |   76.26 |    77.25 |      66 |   76.26 |                   
 frontend          |       0 |        0 |       0 |       0 |                   
  make-env.js      |       0 |        0 |       0 |       0 | 1-26              
 frontend/src      |   14.63 |        0 |       0 |   14.63 |                   
  App.tsx          |    40.9 |      100 |       0 |    40.9 | 16-30,32-42       
  index.tsx        |       0 |        0 |       0 |       0 | 1-69              
  setupProxy.js    |       0 |        0 |       0 |       0 | 1-10              
 ...components/404 |   30.76 |      100 |       0 |   30.76 |                   
  index.tsx        |   30.76 |      100 |       0 |   30.76 | 5-13              
 ...components/App |   67.04 |    81.25 |      44 |   67.04 |                   
  AppContainer.tsx |   26.47 |      100 |       0 |   26.47 | 10-34             
  AppLogo.tsx      |   79.76 |       70 |     100 |   79.76 | 41,43-45,64-76    
  Layout.tsx       |   17.02 |      100 |       0 |   17.02 | 30-55,59-188      
  ...eSwitcher.tsx |   23.66 |      100 |   16.66 |   23.66 | ...04-143,148-164 
[...]

PR #2140, opened 2024-07-05
frontend: Check translations in pre commit hook instead of a test

sniok:
This fixes tests being run twice #2125 
This will also check translation even if you forget to run tests


sniok:
Pre-commit hook is a bit slow. Closing in favor of #2164 2164

PR #2141, opened 2024-07-05
frontend: Improve test coverage collection

sniok:
For coverage task in #2125

Copied coverage collection includes/excludes from previous jest setup so it's more comparable and switched to istanbul coverage provider.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -96,6 +96,7 @@
     "vite": "^5.2.11",
     "vite-plugin-node-polyfills": "^0.22.0",
     "vite-plugin-svgr": "^4.2.0",
+    "vitest": "^1.6.0",
     "xterm": "^4.19.0",
     "xterm-addon-fit": "^0.5.0",
     "xterm-addon-search": "^0.9.0"
@@ -190,7 +191,7 @@
     "@testing-library/user-event": "^14.5.1",
     "@types/nock": "^11.1.0",
     "@types/redux-mock-store": "^1.0.4",
-    "@vitest/coverage-v8": "^1.6.0",
+    "@vitest/coverage-istanbul": "^1.6.0",
     "fetch-mock": "^9.11.0",
     "http-proxy-middleware": "^2.0.1",
     "husky": "^4.3.8",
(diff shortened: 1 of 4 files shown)


sniok:
oh wait v8 coverage collector count comments as lines. Maybe let's stick to istanbul

frontend/vite.config.ts
```diff
@@ -3,6 +3,7 @@ import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vite';
 import { nodePolyfills } from 'vite-plugin-node-polyfills';
 import svgr from 'vite-plugin-svgr';
+import { coverageConfigDefaults } from 'vitest/config';
```
illume, line 6: This would mean vitest and friends would need to move from devDependencies to dependencies? Because now they are required for `npm run build`, and `npm ci` doesn't install devDependencies.

I think that's the cause of the container build failure anyway.

frontend/package.json
```diff
@@ -190,7 +190,7 @@
     "@testing-library/user-event": "^14.5.1",
     "@types/nock": "^11.1.0",
     "@types/redux-mock-store": "^1.0.4",
-    "@vitest/coverage-v8": "^1.6.0",
+    "@vitest/coverage-istanbul": "^1.6.0",
```
illume, line 193: This would need to be added to headlamp-plugin packages as well.

frontend/vite.config.ts
sniok: yeah but it's not really an issue because headlamp is not an npm package

illume: Not sure what that means, but what I'm trying to say is...

We only use `npm ci` for building the container image, that's why it's not installing the devDependencies and erroring in that CI check.

sniok: oh sorry, I meant that I agree that vitest needs to move to normal dependencies and the fact that vitest will be moved is not a big issue since we're using devDeps mostly to speedup CI

illume (approved):
Thanks. 🎉🎈

The coverage numbers are back around where it was before.

Issue #2113, opened 2024-07-01
chart: Image pull secrets not usable anymore


disconn3ct:
The [values schema](https://github.com/headlamp-k8s/headlamp/blame/df4efdd6252d39d9d7afac818f7274acfbc070a9/charts/headlamp/values.schema.json#L34) conflicts with the [k8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret) about image pull secrets.

Specifically, the chart now requires it to be an array of strings, which is still passed directly to 'toyaml'. K8S actually wants an array of secretref objects (`name`) in that slot.

Chart validates this:
```yaml
imagePullSecrets:
  - image-pull-secret
```
K8S wants this:
```yaml
imagePullSecrets:
  - name: image-pull-secret
```

knrt10:
Thank you for opening the issue, will fix it

PR #2145, opened 2024-07-08
charts: Fix imagePullSecrets in value schema

knrt10:
k8s wanted it in a different way, earlier we were validating wrong input.

Fixes: #2113

### DIFF
--- a/charts/headlamp/values.schema.json
+++ b/charts/headlamp/values.schema.json
@@ -33,9 +33,15 @@
     },
     "imagePullSecrets": {
       "type": "array",
-      "description": "Image pull secrets",
+      "description": "ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use.",
       "items": {
-        "type": "string"
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "Name of the image pull secret"
+          }
+        }
       }
     },
     "nameOverride": {


charts/headlamp/values.schema.json
```diff
@@ -35,7 +35,14 @@
       "type": "array",
       "description": "Image pull secrets",
```
illume, line 36: Would you mind fixing this description too?

Currently someone looking at the description for ImagePullSecrets is told imagePullSecrets is...  "Image pull secrets"

The description from the documentation is good
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers

> ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use. More info: https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod

charts/headlamp/values.schema.json
```diff
@@ -35,7 +35,14 @@
...
       "items": {
-        "type": "string"
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "Name of the image pull secret"
+          }
+        },
+        "required": ["name"]
```
illume, line 45: The docs say the name is not required...

See imagePullSecrets here: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers

If you click [LocalObjectReference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/local-object-reference/#LocalObjectReference) from there you see name doesn't say it is required. If a field is required, it always says it's required.

illume (changes requested):
👍

charts/headlamp/values.schema.json
knrt10: Thank you, will fix it

charts/headlamp/values.schema.json
```diff
@@ -35,7 +35,13 @@
       "type": "array",
       "description": "Image pull secrets",
       "items": {
-        "type": "string"
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use."
```
illume, line 42: This is the name field description, but the it reads like it is for the parent. I think this should be moved up.

knrt10: Oops, my bad. Fixed. Thanks again :-)

illume (approved):
Thanks 🎉🎈

Issue #1989, opened 2024-05-22
headlamp helm chart should be signed


illume:
Artifact hub is complaining that the helm chart is not signed.

![image](https://github.com/headlamp-k8s/headlamp/assets/9541/60d81527-e443-47bc-abf2-dc3b84f03b6e)

Note: there's an issue for all of our releases being signed here: https://github.com/headlamp-k8s/headlamp/issues/2092

---

From the [helm documentation](https://helm.sh/docs/helm/helm_package/#helm-package)
> To sign a chart, use the '--sign' flag. In most cases, you should also provide '--keyring path/to/secret/keys' and '--key keyname'.
> 
> $ helm package --sign ./mychart --key mykey --keyring ~/.gnupg/secring.gpg
> 
> If '--keyring' is not specified, Helm usually defaults to the public keyring unless your environment is otherwise configured.

knrt10:
Working on this

PR #2146, opened 2024-07-08
charts: Add signing of Helm charts

knrt10:
This add steps to releasing of Helm charts where it uses the gpg keys to sign and release the charts.

Fixes: #1989

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -34,10 +34,37 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78 # v3.5
 
+      - name: Prepare GPG key
+        run: |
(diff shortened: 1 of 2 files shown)


knrt10:
Tested it [here](https://github.com/knrt10/my-test-chart/releases) and verification tick at ArtifactHub [here](https://artifacthub.io/packages/helm/test-charts/my-test-chart)

knrt10:
For everyone's context, it is required to [save](https://github.com/headlamp-k8s/headlamp/pull/2146/files#diff-3caecf2c7a94138e082e26c2d2d2fe2d5c6e49d3e44e78920aa56a94deab4485R51) passphrase in file, otherwise second option is to pass it as stdin which is not a secure solution in CI. I think since we are already adding it a temp folder `.cr-gpg` which gets cleaned after the CI is stopped, it should not be a problem.

PR #2148, opened 2024-07-08
Plugin catalog extras

yolossn:
This patch adds few extra changes that are required for the plugin catalog plugin.
1. Add extra meta to installed plugins
2. Add user confirmation dialog to plugin Install
3. Add pluginName arg to pluginManager install function.


illume (approved):
👍 thanks

PR #2149, opened 2024-07-09
frontend: Fix reduced motion styles

sniok:
Randomly noticed that reduced motion styles weren't applying, used GlobalStyles component instead


illume (approved):
👍  thanks

PR #2152, opened 2024-07-10
plugin catalog: small improvements

illume:
Some small improvements for headlamp-plugin/plugin-management

- Improve module intro doc
- Fix making temp dir using mkdtempdir which handles a few security issues
- Add validation of pluginName from artifact hub
- Add validation of artifact-url from artifact hub, limit to only artifact hub files for now (and the test plugins on yolossn github too)
- move files into plugin-management/ folder, and rename files to be a bit shorter
- Add #! to top of e2e test to make it easier to un

For https://github.com/headlamp-k8s/headlamp/issues/1654

### How to test

- [ ] Should not be able to put a plugin name like "../../"
- [ ] Should only be allowed to use artifact hub URLs for artifact-url (and the test plugins on yolossn github too)
- [ ] existing tests pass


plugins/headlamp-plugin/plugin-management/plugin-management.js
```diff
@@ -237,6 +242,31 @@ class PluginManager {
...
+}
+
+/**
+ * @param {string} archiveURL - the one to validate
+ * @returns true if the archiveURL looks good.
+ */
+function validateArchiveURL(archiveURL) {
+  return (
+    archiveURL.startsWith('https://artifacthub.io/packages/') ||
+    archiveURL.startsWith('https://github.com/yolossn/headlamp-plugins/')
```
yolossn, line 266: Is this URL required?

plugins/headlamp-plugin/plugin-management/plugin-management.js
```diff
@@ -260,7 +290,15 @@ async function downloadExtractPlugin(URL, headlampVersion, progressCallback, sig
     progressCallback({ type: 'info', message: 'Plugin Metadata Fetched' });
   }
   const pluginName = pluginInfo.name;
+  if (!validatePluginName(pluginName)) {
+    throw new Error('Invalid plugin name');
+  }
+
   const archiveURL = pluginInfo.data['headlamp/plugin/archive-url'];
+  if (!validateArchiveURL(archiveURL)) {
```
yolossn, line 298: archiveURL can be any url pointing to a tarball. this check is failing the install command.
image

plugins/headlamp-plugin/plugin-management/plugin-management.js
illume: I could see the tests were using this URL currently. Probably we should move the test plugins to the plugins repo or the headlamp repo in the future.

plugins/headlamp-plugin/plugin-management/plugin-management.js
yolossn: IMHO I think the developer should have the flexibility to host the plugin tar where they want. The credibility cannot be just associated with the source. @joaquimrocha had a similar opinion of allowing only npm.

illume: I'd like us to lock it down more first, and then expand as time goes on.

yolossn (approved):

PR #2155, opened 2024-07-12
frontend: Adds JSON / object / KubeObjectInterface options

vyncent-t:
Linked to PR: https://github.com/headlamp-k8s/headlamp/pull/1727

## Description

This PR enhances the `singleApiFactory` function in `apiProxy.ts` to improve its flexibility. The `post` function within `singleApiFactory` has been updated to accept parameters of different types: JSON, `object`, or `KubeObjectInterface`. This change ensures that the function can handle various data formats more efficiently, making it more versatile and robust.

## Changes

- [x] Updated the `post` function in `singleApiFactory` to accept parameters of type JSON, `object`, or `KubeObjectInterface`.

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -679,8 +679,11 @@ function singleApiFactory(group: string, version: string, resource: string) {
       queryParams?: QueryParameters,
       cluster?: string
     ) => streamResult(url, name, cb, errCb, queryParams, cluster),
-    post: (body: KubeObjectInterface, queryParams?: QueryParameters, cluster?: string) =>
-      post(url + asQuery(queryParams), body, true, { cluster }),
+    post: (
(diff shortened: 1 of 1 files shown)


sniok:
`object` type is almost like `any`. It will accept anything that's object (for example a function is an object). so this change essentially removes type safety from this function. I'm not sure what was the context for this change but this is probably not the way to do it.

joaquimrocha:
This was my suggestion, after verifying that we had KubeObjectInterface for this post function, but not for the one it ends up calling, which allows JSON | object | KubeObjectInterface . The alternative would be to make many fields in KubeObjectInterface optional, but that's not reflective of a k8s resource, only of the user-sent data.

I think the current PR is fine.

joaquimrocha:
Hmm, I merged it but perhaps I should have waited for your response back, @sniok . Please feel free to propose an alternative that solves both concerns, on top of what we have now.

Issue #2123, opened 2024-07-02
Successful notification when apply/edit operation actually fails


joaquimrocha:
I have noticed that when we apply changes using the create or edit buttons and these fail in the server, there is still a notification showing it's successful, before the editor opens up again with the error being shown.

How to reproduce:
- [x] Create a resource (using the Create button in a cluster view); then apply the same YAML (should fail because the resource exists): notice the green successful notification, before the editor opens up again

PR #2157, opened 2024-07-12
frontend: Prevent success pop-up on invalid operations

skoeva:
This change ensures that invalid resource operations (e.g. changing a resource kind) do not trigger a success notification. Previously, the app would indicate a successful attempt to apply changes even when it is unsuccessful. Now when trying to perform an invalid apply/edit operation, instead the editor will reappear with the error message and a failure notification.

Fixes: #2123

### Testing

- [x] Create/edit a resource (using the Create button in cluster view or Edit button in resource view)
- [x] Perform an invalid operation (e.g. change the kind, add '!' to the name)
- [x] Ensure that the editor pops up with the error message and a failure notification

### DIFF
--- a/frontend/src/redux/clusterActionSlice.ts
+++ b/frontend/src/redux/clusterActionSlice.ts
@@ -242,7 +242,7 @@ export const executeClusterAction = createAsyncThunk(
         if (controller.signal.aborted) {
           return rejectWithValue('Action cancelled');
         }
-        callback();
+        await Promise.resolve(callback());
         dispatchSuccess();
       } catch (err) {
         if ((err as Error).message === 'Action cancelled' || controller.signal.aborted) {


frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -84,8 +84,9 @@ export default function CreateButton(props: CreateButtonProps) {
 
     const clusterName = getCluster() || '';
 
+    await applyFunc(massagedNewItemDefs, clusterName);
     dispatch(
-      clusterAction(() => applyFunc(massagedNewItemDefs, clusterName), {
+      clusterAction(() => {}, {
```
joaquimrocha, line 89: You are waiting for the apply function and then you are calling the cluster action which has the timeout. This does not do what is intended.

frontend/src/components/common/Resource/EditButton.tsx
```diff
@@ -51,14 +51,16 @@ export default function EditButton(props: EditButtonProps) {
...
 
-  function handleSave(items: KubeObjectInterface[]) {
+  async function handleSave(items: KubeObjectInterface[]) {
     const newItemDef = Array.isArray(items) ? items[0] : items;
     const cancelUrl = location.pathname;
     const itemName = item.metadata.name;
 
     setOpenDialog(false);
+
+    await applyFunc(newItemDef);
```
joaquimrocha, line 61: Ditto.

joaquimrocha (changes requested):

frontend/src/redux/clusterActionSlice.ts
```diff
@@ -242,7 +242,7 @@ export const executeClusterAction = createAsyncThunk(
         if (controller.signal.aborted) {
           return rejectWithValue('Action cancelled');
         }
-        callback();
+        await callback();
```
joaquimrocha, line 245: Will this not break if callbacks are not promises? We should double check this.
If this logic is internal so far, we can alter the cases where it's being called and convert those functions to async ones, but we need to double check nonetheless.

joaquimrocha (changes requested):

frontend/src/redux/clusterActionSlice.ts
skoeva: I believe just wrapping `callback()` with a Promise should address this, they seem to be functionally equivalent in how they send failure/success messages

joaquimrocha (approved):

PR #2164, opened 2024-07-15
Makefile: github/workflows/frontend: Add i18n-check to CI

illume:
Fixes tests so they are not run twice #2125

Running it inside the tests makes the translation files be written
when the tests run, causing the tests to run twice.

Running it as a pre-commit hook is a bit slow (3.0s), because
it looks at all the files that have changed, and then writes all
the translation files again.

So instead we run it in CI so it catches the issues.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -78,6 +78,10 @@ jobs:
         run: |
           make frontend-test
 
+      - name: Run frontend-i18n-check
+        run: |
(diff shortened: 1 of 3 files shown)


illume:
Related to https://github.com/headlamp-k8s/headlamp/pull/2140

sniok:
If it's too slow as a pre-commit hook then this is a nice alternative. I'll close my PR

sniok (approved):

PR #2170, opened 2024-07-18
frontend: Mark some modules in pluginLib as esModule

sniok:
To maintain compatibility with plugins that are using webpack, some modules that have default exports need to have a special __esModule property so that plugins can grab the default export

- [x] Tested Kompose plugin with current version 0.24.1 (built with webpack)
- [x] Tested Kompose plugin with main branch (plugin crashes)
- [x] Tested Kompose plugin with this change (plugin works as expected)


illume (approved):
👍  Nice one.

PR #2173, opened 2024-07-18
app: plugins: Fix validateArchiveURL to support github/gitlab/bitbucket

illume:
So release tar.gz files can be put on there.


illume:
For some reason my local `npm run format` is not working the same as CI.

joaquimrocha:
While this adds some extra security that we only allow downloading from managed services like these, theoretically plugins can be hosted anywhere, e.g. GitHub / Cloudflare pages, NPM repos,... So we should understand whether this is short list is really needed.

joaquimrocha (approved):

illume:
Fixed a test failure.

PR #2174, opened 2024-07-19
app: Fix defaultPluginsDir on unix

illume:
Now it gives ~/.config/Headlamp/plugins

Now it uses the same envPaths version so the plugin paths should be consistent.

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -9,3 +9,4 @@
 /electron/windowSize.js
 /electron/plugin-management.js
 electron/windowSize.test.js
+electron/env-paths.js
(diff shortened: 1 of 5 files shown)


yolossn (commented):
We should also update it in https://github.com/headlamp-k8s/headlamp/blob/820caaabe9ac0201fc46a3315ab8f1be1bd099d2/plugins/headlamp-plugin/plugin-management/plugin-management.js#L487

illume:
Fixed lint issue, and removed some debugging console.logs now it seems to be working.

joaquimrocha (approved):

PR #2175, opened 2024-07-19
headlamp-plugin: Add package command

joaquimrocha:
This command creates a tarball in the format that headlamp-plugin expects to be in when we use the install command.

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -4,9 +4,11 @@
 
 const webpack = require('webpack');
 const config = require('../config/webpack.config');
+const crypto = require('crypto');
 const fs = require('fs-extra');
 const FileManagerPlugin = require('filemanager-webpack-plugin');
 const envPaths = require('env-paths');
+const os = require('os');
 const path = require('path');
 const resolve = path.resolve;
 const child_process = require('child_process');
@@ -15,6 +17,7 @@ const yargs = require('yargs/yargs');
 const headlampPluginPkg = require('../package.json');
 const pluginManager = require('../plugin-management/plugin-management');
 const { table } = require('table');
+const tar = require('tar');
 
 /**
  * Creates a new plugin folder.
@@ -153,25 +156,32 @@ function compileMessages(err, stats) {
  *
  * @param {string} pluginPackagesPath - can be a package or a folder of packages.
  * @param {string} outputPlugins - folder where the plugins are placed.
+ * @param {boolean} logSteps - whether to print the steps of the extraction (true by default).
  * @returns {0 | 1} Exit code, where 0 is success, 1 is failure.
  */
-function extract(pluginPackagesPath, outputPlugins) {
+function extract(pluginPackagesPath, outputPlugins, logSteps = true) {
   if (!fs.existsSync(pluginPackagesPath)) {
     console.error(`"${pluginPackagesPath}" does not exist. Not extracting.`);
     return 1;
   }
   if (!fs.existsSync(outputPlugins)) {
-    console.log(`"${outputPlugins}" did not exist, making folder.`);
+    if (logSteps) {
+      console.log(`"${outputPlugins}" did not exist, making folder.`);
+    }
     fs.mkdirSync(outputPlugins);
   }
 
   function copyFiles(plugName, inputMainJs, mainjs) {
     if (!fs.existsSync(plugName)) {
-      console.log(`Making output folder "${plugName}".`);
+      if (logSteps) {
+        console.log(`Making output folder "${plugName}".`);
+      }
       fs.mkdirSync(plugName);
     }
 
-    console.log(`Copying "${inputMainJs}" to "${mainjs}".`);
+    if (logSteps) {
+      console.log(`Copying "${inputMainJs}" to "${mainjs}".`);
+    }
     fs.copyFileSync(inputMainJs, mainjs);
   }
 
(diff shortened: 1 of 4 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1048,6 +1115,27 @@ yargs(process.argv.slice(2))
       process.exitCode = extract(argv.pluginPackages, argv.outputPlugins);
     }
   )
+  .command(
+    'pack <pluginPath> <outputDir>',
```
illume, line 1119: Can you please add docs to these spots?
- headlamp-plugin/README.md
- headlamp/docs/development/plugins/building.md

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -231,6 +233,71 @@ function extract(pluginPackagesPath, outputPlugins) {
...
+ * @returns {0 | 1} Exit code, where 0 is success, 1 is failure.
+ */
+async function createArchive(pluginDir, outputFolderPath) {
+  const pluginPath = path.resolve(pluginDir);
+  if (!fs.existsSync(pluginPath)) {
+    console.error(`"${pluginPath}" does not exist. Not creating archive.`);
+    return 1;
+  }
+
+  if (!fs.existsSync(outputFolderPath)) {
```
illume, line 255: Maybe some of this code could be shared with code in extract(). This could be done later.

illume (changes requested):
Nice one. Left a few comments for your consideration.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1048,6 +1115,27 @@ yargs(process.argv.slice(2))
       process.exitCode = extract(argv.pluginPackages, argv.outputPlugins);
     }
   )
+  .command(
+    'pack <pluginPath> <outputDir>',
+    'Creates a tarball of the plugin package in the format Headlamp expects.',
```
illume, line 1120: I wonder if it should error before overwriting? I guess it would be similar to `npm pack`, and overwrite.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -231,6 +233,71 @@ function extract(pluginPackagesPath, outputPlugins) {
   return 0;
 }
 
+/**
```
illume, line 236: headlamp/plugins/headlamp-plugin/template/package.json is where you'd add it so it's added to peoples packages.

When you do `npm run pack` should it output in the current directory? Like `npm pack` does?

plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha: I will add to README.md but for the building, we need more docs, as pack is tied to install as well.

yolossn:
Not a blocker but It would be nice to have the command also print the checksum that can be used in the `artifacthub-pkg.yaml`.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1048,6 +1115,27 @@ yargs(process.argv.slice(2))
       process.exitCode = extract(argv.pluginPackages, argv.outputPlugins);
     }
   )
+  .command(
+    'pack <pluginPath> <outputDir>',
+    'Creates a tarball of the plugin package in the format Headlamp expects.',
+    yargs => {
+      yargs.positional('pluginPath', {
```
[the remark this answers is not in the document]
joaquimrocha: Let's add it later.

plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha: I will make it err if the destination tarball exists.

plugins/headlamp-plugin/bin/headlamp-plugin.js
joaquimrocha: yeah, I was wondering about what the package.json script should be named. I think 2 packs will be confusing?

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1048,6 +1115,27 @@ yargs(process.argv.slice(2))
       process.exitCode = extract(argv.pluginPackages, argv.outputPlugins);
     }
   )
+  .command(
```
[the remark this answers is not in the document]
joaquimrocha: Done.

joaquimrocha:
Updated the branch with the checksum printing, test, and error when the final tarball already exists.

joaquimrocha:
I have updated it again:
- Now the command is called `package` instead of `pack` so we keep the latter for NPM
- Now the command defaults to the current dir in both options, like build does
- Added the package script to the template package.json

yolossn (commented):
Getting this error message but the tar.gz file is created

image

when i try to extract the tar.gz file I am getting this error
image

joaquimrocha:
@yolossn there was an error in extract as it didn't return the error code when no packages were extracted. I have updated that.

yolossn (approved):
Tested.

image

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -226,11 +236,107 @@ function extract(pluginPackagesPath, outputPlugins) {
...
   }
 
   return 0;
 }
 
+/**
+ * Calculate the checksum of a file.
+ *
+ * @param {*} filePath
+ * @returns
```
illume, line 249: The docs here are missing.

illume (approved):
👍 great!

PR #2178, opened 2024-07-19
plugins/examples/pod-counter: Fix notifications issue

illume:
Was causing a crash when actions is not there.

## How to test

- Load pod-counter
- don't see a pod-counter related issue in the web dev console


sniok:
could you also add that actions might be undefined in `AppBarActionsProcessorArgs` type?

illume:
> could you also add that actions might be undefined in AppBarActionsProcessorArgs type?

Yeah, good point. Rather than the fix in this PR... Probably we should look into why it's giving undefined and stop it from doing that.

sniok (approved):

PR #2179, opened 2024-07-19
app: Fix serialization error in app

illume:
There was a serialization error sending the data webContents.

## how to test

This error was in the console.
image

- "Navigation" menu still works
- app-menus example plugin still works

app-menus plugin

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -989,7 +989,9 @@ function startElecron() {
     });
 
     mainWindow.webContents.on('dom-ready', () => {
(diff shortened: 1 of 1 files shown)


sniok (approved):
I saw that error too, thanks for the fix. Do you think it's related to the electron version? Or why would it stop working suddenly

illume:
> I saw that error too, thanks for the fix. Do you think it's related to the electron version? Or why would it stop working suddenly

I think someone put functions or objects in there that don't serialize. Not sure what they are.

PR #2185, opened 2024-07-22
frontend: Include styles in MuiMaterial in pluginLib

sniok:
When plugins try to use `@mui/x-tree-view/TreeView` it fails because importing `@mui/material/styles` fails. 
Webpack version (0.24.1) also fails 

This fixes that

With this fix plugins can use Tree component like so 

```
import { TreeItem } from '@mui/x-tree-view/TreeItem';
import { TreeView } from '@mui/x-tree-view/TreeView';
```

- [x] Tested by using Tree component in plugins importing the treeview (on main and 0.24.1)


joaquimrocha (approved):
Great finding! Thanks!

PR #2187, opened 2024-07-22
github: build-container: Fix cached image loading

illume:
Before cached images meant that the e2e tests would fail.

Additionally, before changing go dependencies meant that the
image would not rebuild. Now if backend dependencies change
a new image is built and tested.

In this job you'll see that it uses a cached image, and the job fails. https://github.com/headlamp-k8s/headlamp/actions/runs/10042659886/job/27753598525

It seems to be because when things are cached we were skipping loading the images into kind. So we change it to always load the images into kind... no matter if the image is built or restored from cache.

---

There's a test of the caching behavior in the PR here: https://github.com/headlamp-k8s/headlamp/pull/2189
The job that successfully builds even though it is using caching is here: https://github.com/headlamp-k8s/headlamp/actions/runs/10044873649/job/27795185747

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -47,7 +47,7 @@ jobs:
       with:
         path: ~/image-cache
         # cache the container image. All the paths this PR depends on except the e2e-tests folder for the key.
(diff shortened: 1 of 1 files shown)


illume:
kind docs for loading images into a kind cluster https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

joaquimrocha (approved):

PR #2194, opened 2024-07-24
OLD - fixes for app e2e

vyncent-t:


vyncent-t:
This issue seems to be related to the WSL render issue as this error only occurs on machines that are impacted by that issue.

Machines impacted by the render error:
- run 'npm run start'
- look for the error messages: "MESA: error: ZINK: failed to choose pdev" "glx: failed to create drisw screen"

image

For machines impacted when running the 'npx playwright test -g "launch app" the app will not launch and will return the error codes "MESA: error: ZINK: failed to choose pdev" "glx: failed to create drisw screen"

For machines NOT impacted:

- when running the 'npx playwright test -g "launch app" a different error occurs but it seems that playwright for electron works fine
image

vyncent-t:
- disableGPU seems to get the app to launch but the error codes still occur in playwright 
- also the backend error 
image


image

vyncent-t:
Last push works with creating an env for playwright to launch the app in

the env must contain a few specific things before starting

## Instructions

### Readying the playwright env
- Get the env ready for playwright to launch electron by running the commands for both frontend and backend
- In a terminal cd into headlamp and run 'make run-frontend'
- In another terminal cd into headlamp and run 'make backend' and then 'make run-backend'

### current working directory
- change the 'cwd' to match the electron path for your machine 
image

### Launch playwright UI
- cd into app/app-e2e-tests and run 'npx playwright test --ui'

skoeva:
My recent push allows for running the frontend and backend servers (otherwise using `make run-frontend` or `make run-backend`) within the test suite.

## Testing instructions

### Change the current working directory
- Change the `cwd` to match the paths for your local machine (replace `skoeva` with your username)

image

### Launch the test suite and run the test
- Run `cd app/app-e2e-tests/e2e && npx playwright test --ui`
- Run the 'launch app with integrated shell commands' test

image

vyncent-t:
Was working on a side commit for pages, it is not compatible at the moment, that sideline was abandoned- will continue with this

app/app-e2e-tests/e2e/example.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  // await headlampPage.hasNetworkTab();
+  // Direct Electron console to Node terminal.
+  window.on('console', console.log);
+  // Click button.
+  // await window.click('text=Click me');
+});
+
+test('launch app with integrated shell commands', async () => {
+  // Start the frontend server
+  await runCommand('npm', ['start'], { cwd: '/home/skoeva/headlamp/frontend' });
```
vyncent-t, line 89: The automation needs a way to be closed after running the commands for npm start etc, this could be done by running this all before testing using `beforeAll`

The server being ran this also has issues as it does not auto run to the next lines after launching, needing manual input for either continue or quit as it clashes

vyncent-t, line 89: The automation needs a way to be closed after running the commands for npm start etc, this could be done by running this all before testing using beforeAll

The server being ran this also has issues as it does not auto run to the next lines after launching, needing manual input for either continue or quit as it clashes

app/app-e2e-tests/e2e/example.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Detach the process and resolve immediately
+    process.unref();
+    resolve();
+  });
+}
+
+test('launch app', async () => {
+  // Launch Electron app.
+  const electronApp = await electron.launch({
+    cwd: '/home/vtaylor/headlamp/app/electron',
```
vyncent-t, line 48: We will need a better way to use the absolute path as we cannot just run `headlamp/app/electron` we could also maybe use path?

vyncent-t:
working on solution to be able to run all tests at once and not one after another, currently errors when trying to run whole group

app/app-e2e-tests/e2e/example.spec.ts
vyncent-t: Added changes to solve this

vyncent-t:
Issue: when trying to merge all the e2e tests in one place

- The playwright config issues between the app mode and the base project prevent us from using the HeadlampPage methods when either importing and working with the main structure of it

- Because of this, there are also differences in the playwright config when it comes to working on running tests, currently the app mode MUST be ran one test at a time in order for it to not clash with the electron app launch and close. The base project tests do NOT need to be ran with the same configuration thus proving difficult to merge them.

vyncent-t:
Issue: Trying to work and connect to in cluster app mode

- Current there seems to be an issue where the app is running twice when using the headlamp addon
image

This issue does NOT occur when running tests with the headlamp addon disabled

app/app-e2e-tests/e2e/example.spec.ts
vyncent-t: I think we may need to have this be used offline like the current base project for in cluster and web mode

vyncent-t:
working on fixing headlamp addon clash

vyncent-t:
Currently the app mode does connect and we are able to create and delete resources as normal, the only issue now is that when running create or delete tests

- test fails to find delete box and continues to wait
- test fails to find the pop up that reads "XYZ has been deleted"
- test fails to close after delete process finishes (need to add a new way to find if something deletes)

vyncent-t:
I am trying to use scripts that will change the env for when we run the tests in different modes.

WIP: moving the app namespace tests to what we currently use

to do: create a test for namespace create button icon

- the playwright test command runs another test so no longer using `"playwright-test-web": "cross-env PLAYWRIGHT_TEST_MODE=web npx playwright test -g 'create a namespace with the minimal editor then delete it' --workers=1",`
- trying to use test.before to work with the pages does not seem to work (it would pause at the main page)
- working on starting every test from the same point (the main page)
- starting point will also be different when running the script for each playwright mode

joaquimrocha:
@vyncent-t What should we do with this PR?

vyncent-t:
> @vyncent-t What should we do with this PR?

This is the demo branch that held a lot of the foundational work for getting playwright to work in app mode. I keep it around mostly for notes and references. I wouldn't merge it, but I would like to keep it as a archive/sandbox PR if we were to expand more with electron playwright work.

PR #2195, opened 2024-07-24
app setup-plugins.js: Add support for new plugin tarball structure

joaquimrocha:
The new plugin structure is slightly different, so we add support for it in the setup-plugins.js script.

### DIFF
--- a/app/scripts/setup-plugins.js
+++ b/app/scripts/setup-plugins.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 const tar = require('tar');
+const glob = require('glob');
 var zlib = require('zlib');
 const os = require('os');
 const https = require('https');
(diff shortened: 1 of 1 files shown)


app/scripts/setup-plugins.js
```diff
@@ -37,14 +38,32 @@ async function extractArchive(
...
-        fs.copyFileSync(
-          path.join(tmpFolder, 'package', 'dist', 'main.js'),
-          path.join(pluginFolder, 'main.js')
-        );
-        fs.copyFileSync(
-          path.join(tmpFolder, 'package', 'package.json'),
-          path.join(pluginFolder, 'package.json')
-        );
+        const mainLocationExpr = path.join(tmpFolder, '*', 'main.js');
+        const mainLocations = glob.sync(mainLocationExpr);
```
yolossn, line 42: why do we need this pattern match?

joaquimrocha: Because AFAIU the tarballs generated will unpack into PLUGIN_NAME/main.js and PLUGIN_NAME/package.js .
We don't know what PLUGIN_NAME is at this point.

yolossn: Cool got it, makes sense.

Issue #1927, opened 2024-04-24
Settings are a bit confusing


sniok:
Currently the settings icon in the top right corner just says "Settings". For example I was looking for the theme toggle and assumed it would be there but it was actually cluster settings. Then to get the headlamp settings you need to click "General settings" which is even more confusing because what is "general" about it?

Maybe we can put cluster settings in the sidebar on the left? Next to all the existing cluster stuff. Then in the navbar we can have headlamp settings, all the way in the corner unrelated to the cluster.

![image](https://github.com/headlamp-k8s/headlamp/assets/4517681/803ebb85-9aad-46c7-8560-f920aeaa487e)

joaquimrocha:
I have avoided adding settings there because it would show right below Configuration. So you would have Configuration + Settings, which sounds just confusing.
I will bring this question to the design team so we they can probably give a final suggestion.

Issue #2037, opened 2024-06-12
Unify cluster settings with general settings


joaquimrocha:
Right now, as #1927 indicates, accessing settings is very confusing in Headlamp.
To fix that, we want to have just one settings button, which will give access to the settings. The problem with this is that we have generic settings, cluster settings, and plugin settings.

So a quick fix for this for now is to add the cluster settings as part of the settings area.
This means the sidebar would become like:

```
Settings
  General
  Plugins
  Cluster
```

This means that when a user clicks the Settings button at the top, it will do one of two things:
1. If we are in a cluster context, then we're taken to the Settings > Cluster.
2. If we are in a non-cluster context, then the user is taken to the Settings > General.

Now the problem is that if the user navigates into the Cluster settings from a non-cluster context, we need to know which cluster to display there, and that's why this Cluster settings view will have a cluster selector there instead of a cluster name. See the following mockup:

![Cluster settings with a cluster selector](https://github.com/headlamp-k8s/headlamp/assets/1029635/75b49ff1-8ab9-49f5-a613-8ad0a4caf391)

Note: If we only have one cluster set up AND we are in non-desktop mode, then we should show the cluster name as a title, not a selector (as it's unlikely to ever get more than one cluster).

joaquimrocha:
Update for how the sidebar should show (it's using a different theme but you can understand how it should be in our current one):
![Settings mockup](https://github.com/headlamp-k8s/headlamp/assets/1029635/611ddff9-8cb8-4bca-9b19-3577f2755c29)

PR #2205, opened 2024-07-25
frontend: Unify general and cluster settings

skoeva:
Accessing settings in Headlamp is confusing. This change adds the cluster settings to the sidebar along with the general and plugin settings. When we open a cluster, clicking the Settings (gear) icon at the top will now take us to the cluster settings page in the home context, with a cluster selector appearing at the top of the page.

Fixes: #1927 and #2037

#### Unified settings in the sidebar
- [X] Open Headlamp and click on Settings
- [X] Ensure that the sidebar contains general, plugins, and cluster settings
#### Cluster settings in home context
- [X] Open a cluster and click on the Settings (gear) icon at the top
- [X] Ensure that the cluster settings page appears in the home context
#### Cluster settings (1+ clusters set up)
- [X] With at least one cluster set up, open the cluster settings
- [X] Ensure that the cluster settings page displays the name in the title and a cluster selector below

image
#### Cluster settings (change URL to non-existing cluster)
- [X] Open the cluster settings from the cluster context
- [X] Change the cluster name in the URL to a non-existing cluster
- [X] Ensure that the cluster settings for this cluster cannot be accessed

image


frontend/src/lib/router.tsx
```diff
@@ -699,10 +714,14 @@ const defaultRoutes: {
     ),
   },
   settingsCluster: {
-    path: '/settings',
+    path: '/settings/cluster',
```
joaquimrocha, line 717: Instead of replacing this route. Keep the route (but adapt the sidebar), then add a new route that is essentially the one you replaced it with (/settings/cluster). Because we don't know if some plugins out there refer to this route, in which case we'd be breaking their use-cases.

frontend/src/lib/router.tsx
```diff
@@ -673,7 +673,23 @@ const defaultRoutes: {
...
+        <Settings />
+      </PageGrid>
+    ),
+  },
+  settingsGeneral: {
+    path: '/settings/general',
+    exact: true,
+    name: 'General Settings',
+    sidebar: {
+      item: 'general',
```
joaquimrocha, line 692: This one I think we don't need because we already have /settings, right?

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -38,45 +38,44 @@ function isValidClusterNameFormat(name: string) {
 }
 
 export default function SettingsCluster() {
-  const cluster = useCluster();
   const clusterConf = useClustersConf();
+  const clusters = Object.keys(clusterConf || {});
```
joaquimrocha, line 42: Although the cluster keys may equal the cluster name, this should rather be the name of each cluster (i.e. name from the each cluster value).

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -38,45 +38,44 @@ function isValidClusterNameFormat(name: string) {
 }
 
 export default function SettingsCluster() {
-  const cluster = useCluster();
```
joaquimrocha, line 41: If we keep the cluster route for the settings, c/mycluster/settings, we will be able to determine the cluster we should display for the settings directly from the context, and the logic of the component would be much closer to what we had before as well.

joaquimrocha (changes requested):

frontend/src/lib/router.tsx
skoeva: Done

frontend/src/lib/router.tsx
skoeva: Yup, removed

frontend/src/components/App/Settings/SettingsCluster.tsx
skoeva: Added back in

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -38,14 +39,15 @@ function isValidClusterNameFormat(name: string) {
...
   const clusterConf = useClustersConf();
+  const clusters = Object.values(clusterConf || {}).map(cluster => cluster.name);
   const { t } = useTranslation(['translation']);
   const [defaultNamespace, setDefaultNamespace] = React.useState('default');
   const [userDefaultNamespace, setUserDefaultNamespace] = React.useState('');
   const [newAllowedNamespace, setNewAllowedNamespace] = React.useState('');
   const [clusterSettings, setClusterSettings] = React.useState<ClusterSettings | null>(null);
-  const [newClusterName, setNewClusterName] = React.useState(cluster || '');
+  const [newClusterName, setNewClusterName] = React.useState(useCluster() || '');
+  const [cluster, setCluster] = React.useState(useCluster() || '');
```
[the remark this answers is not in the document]
joaquimrocha: Once we have the query parameter, we should modify the settings cog button to point to the /settings/cluster?c=CURRENT_CLUSTER.

joaquimrocha (changes requested):
Gave it a try and the functionality is there, thanks. I think we should still make some changes to make it more unified and less prone to future problems.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -221,201 +218,204 @@ export default function SettingsCluster() {
...
   return (
     <>
       <SectionBox
         title={
-          Object.keys(clusterConf || {}).length > 1
-            ? t('translation|Cluster Settings ({{ clusterName }})', { clusterName: cluster || '' })
+          clusters.length < 2 || isValidCluster
+            ? t('translation|Cluster Settings ({{ clusterName }})', {
+                clusterName: cluster || '',
+              })
```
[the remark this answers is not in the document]
skoeva: There are currently multiple versions of the settings:

### Opening in the cluster context
In this case, we show the cluster name in the title and hide the selector.

### Opening in the home context
In this case, we hide the cluster name and show the selector.

### Invalid cluster
In the case that we open in an invalid cluster (change the cluster in the URL to a random name), we show the cluster name in the title and hide the rest of the settings.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -221,201 +218,204 @@ export default function SettingsCluster() {
...
-                          disabled={!newAllowedNamespace}
-                          size="medium"
-                        >
-                          <InlineIcon icon="mdi:plus-circle" />
-                        </IconButton>
-                      ),
-                      onKeyPress: event => {
-                        if (event.key === 'Enter') {
-                          storeNewAllowedNamespace(newAllowedNamespace);
+        {!location.pathname.includes(cluster) || isValidCluster ? (
```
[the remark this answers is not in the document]
skoeva: These are two separate conditions:

### !location.pathname.includes(cluster)
Let's say we open the cluster settings from the home context, and the URL is `localhost:3000/settings/cluster`. In this case, we are not in a cluster but still want to show the settings (along with the selector). We can use `!location.pathname.includes(cluster)` or another condition to ensure that we are not in a cluster.

### isValidCluster
Let's say open the cluster settings from the cluster context, and the URL is `localhost:3000/c/clustername/settings`. In this case, we are in a (valid) cluster and want to show the settings. We use `isValidCluster` to check that we are in a valid cluster.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -221,201 +218,204 @@ export default function SettingsCluster() {
   const invalidClusterNameMessage = t(
     "translation|Cluster name must contain only lowercase alphanumeric characters or '-', and must start and end with an alphanumeric character."
   );
+  const isValidCluster = clusters.includes(cluster) && location.pathname.includes(cluster);
```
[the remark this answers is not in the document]
skoeva: Removed the condition and created a redirection for the route url creation

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -58,25 +60,23 @@ export default function SettingsCluster() {
...
-                  .then((config: any) => {
-                    storeNewClusterName(newClusterName);
-                    dispatch(setStatelessConfig(config));
-                  })
-                  .catch((err: Error) => {
-                    console.error('Error updating cluster name:', err.message);
-                  });
-              }
-            } else {
-              dispatch(setConfig(config));
```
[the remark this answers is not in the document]
skoeva: Reverted this change

frontend/src/components/App/Settings/SettingsCluster.tsx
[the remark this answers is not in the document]
skoeva: There is now one (real) version of the settings:

### Opening in the cluster context
In this case, we show the cluster name in the title and show the selector.

### Opening in the home context
In this case, we pick the first cluster in the list of clusters and proceed as above.

### Invalid cluster
In this case, we show the cluster name in the title and the `Empty` error component, and hide the rest of the settings as before.

frontend/src/components/App/Settings/SettingsCluster.tsx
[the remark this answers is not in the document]
skoeva: In condensing the versions of the settings into one, there is now only one condition to render the settings:

`isValidCluster = cluster && clusters.includes(cluster)`

frontend/src/components/App/Settings/SettingsCluster.tsx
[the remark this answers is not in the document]
skoeva: Done

joaquimrocha (changes requested):
I think there are still a few things that indicate this is risky to merge. Please try to touch as little code as possible the you can, in the actual settings component.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -222,200 +228,206 @@ export default function SettingsCluster() {
     "translation|Cluster name must contain only lowercase alphanumeric characters or '-', and must start and end with an alphanumeric character."
   );
 
+  const clusterFromUrl = new URLSearchParams(location.search).get('c');
+  const isValidCluster = cluster && clusters.includes(cluster);
```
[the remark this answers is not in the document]
skoeva: Looks to be redundant in testing, removed

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -196,16 +207,11 @@ export default function SettingsCluster() {
   }
 
   function storeNewClusterName(name: string) {
-    let actualName = name;
-    if (name === cluster) {
-      actualName = '';
-      setNewClusterName(actualName);
-    }
-
+    setNewClusterName(name);
```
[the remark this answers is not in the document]
skoeva: Don't remember exactly but reverting this looks to work fine as well

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -222,200 +228,206 @@ export default function SettingsCluster() {
...
-              routeName={'settings'}
-              align="right"
-              style={{ color: theme.palette.text.primary }}
-            >
-              {t('translation|General Settings')}
-            </Link>,
-          ],
-        }}
       >
-        {helpers.isElectron() && (
```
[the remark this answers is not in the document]
skoeva: Actually keeping this condition seems to fail locally. Will revert the change and continue testing

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -222,200 +232,205 @@ export default function SettingsCluster() {
...
-                          disabled={!newAllowedNamespace}
-                          size="medium"
-                        >
-                          <InlineIcon icon="mdi:plus-circle" />
-                        </IconButton>
-                      ),
-                      onKeyPress: event => {
-                        if (event.key === 'Enter') {
-                          storeNewAllowedNamespace(newAllowedNamespace);
+        {cluster ? (
```
joaquimrocha, line 245: Instead of adding this big conditional to the block, you can use what was there before (checking for cluster before) and return the empty component instead of the null that was being returned before. This will allow the changes to be less so.

joaquimrocha (changes requested):
Once these changes are addressed, I will help with the URL redirection.

frontend/src/components/App/Settings/SettingsCluster.tsx
skoeva: Reverted the change

joaquimrocha (dismissed):
I have left a comment and pushed the branch after rebase. Also added a solution for the redirection. @skoeva Please test my changes and squash my commit with yours if it works fine. I would also like to get a way to access the cluster directly from the settings. Maybe we make the cluster name in the title a link, or we add a link to the right of the picker ("Go to cluster").

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -222,27 +233,53 @@ export default function SettingsCluster() {
     "translation|Cluster name must contain only lowercase alphanumeric characters or '-', and must start and end with an alphanumeric character."
   );
 
+  const clusterFromUrl = new URLSearchParams(location.search).get('c');
```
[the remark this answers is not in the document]
skoeva: Making the changes in `settingsCluster` works as expected, though I believe there may be an issue with the redirection. When opening the cluster settings, the screen briefly flickers with the error page before opening the settings. Commit [be04fe2](https://github.com/headlamp-k8s/headlamp/commit/be04fe2b8b5029964f33adf3dd657628d5fc2ebf) does not have this issue, but the latest [5959892](https://github.com/headlamp-k8s/headlamp/commit/5959892503f946185cf9a9290e8afe57981f8624) does. Do you see the same thing on your end?

joaquimrocha:
@skoeva I have added a new fixup commit which should fix the problem you saw. I have also fixed 2 issues I found: no cluster selector when the current cluster doesn't exist; and the empty state when no clusters are configured. Please test these too please.

skoeva:
> @skoeva I have added a new fixup commit which should fix the problem you saw. I have also fixed 2 issues I found: no cluster selector when the current cluster doesn't exist; and the empty state when no clusters are configured. Please test these too please.

@joaquimrocha Works well on my end, thanks for the fix! Made a couple of minor changes to grammar and formatting, should be good now

illume (changes requested):
Can you please check for a11y issues with lighthouse in dev tools?

a11y issues

(Note: the "Elements use prohibited aria attributes" one is not because of this PR and can be ignored in here).

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -37,19 +38,53 @@ function isValidClusterNameFormat(name: string) {
...
+}
+
+function ClusterSelector(props: ClusterSelectorProps) {
+  const { currentCluster = '', clusters } = props;
+  const history = useHistory();
+  const { t } = useTranslation('glossary');
+
+  return (
+    <FormControl variant="outlined" margin="normal" sx={{ minWidth: 250 }}>
+      <Select
```
illume, line 53: A nice to have: This could take an "error={props.error}" attribute to show when it's not valid. Like when the name isn't there.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -222,27 +266,56 @@ export default function SettingsCluster() {
...
+        <Empty color="error">{t('translation|There seem to be no clusters configured…')}</Empty>
+      </>
+    );
+  }
+
+  if (!cluster) {
+    return (
+      <>
+        <SectionBox title={t('translation|Cluster Settings')} backLink>
+          <ClusterSelector clusters={clusters} />
```
illume, line 287: What do you think of placing it as a sub heading?
```ts
        <SectionBox title={t('translation|Cluster Settings')} backLink>
          <Typography color="error" component="h3" variant="h5">
            {t(
              'translation|Cluster {{ clusterName }} does not exist. Please select a valid cluster.',
              {
                clusterName: clusterFromURLRef.current,
              }
            )}
          </Typography>
          <ClusterSelector clusters={clusters} />
        </SectionBox>

```

image

illume: Note: for the error color contrast issue I made a PR https://github.com/headlamp-k8s/headlamp/pull/2284 

So the red is a bit lighter and more readable.

skoeva: I think placing it as a sub-heading might be appealing, especially if we end the message with a colon like this:
#### Please select a valid cluster:

Played around with the colors as well, I think the alpha on `error.light` is a little too low to read on the dark background comfortably, so I think the main error color will suffice for now. Maybe I can revisit that? looking through the lighthouse results as well

frontend/src/components/App/Settings/SettingsCluster.tsx
skoeva: I feel this would clash with the error message, since the selector would turn red even though it only displays valid clusters. would be less confusing IMO to leave this out but if you don't think so I can add this in

illume (approved):
🎉 Nice work. Thanks!

PR #2208, opened 2024-07-27
Update docs format

joaquimrocha:
Use the docs format that docusaurus expects, since now we use this project as our docs viewer/website.


sniok (approved):

illume (approved):
👍 thanks.

(the container job failure is unrelated and fixed in main)

PR #2210, opened 2024-07-29
container fetch-plugins.sh: Update to new plugin tarball format

joaquimrocha:


illume:
I noticed that a Build container job is not running on CI with just this change... so built it locally. 

On main it fails, and on this branch it builds.
```
Dockerfile:73
--------------------
  71 |     WORKDIR /tools
  72 |     RUN mkdir -p /plugins
  73 | >>> RUN ./fetch-plugins.sh /plugins/
  74 |
  75 |     FROM image-base as final
--------------------
ERROR: failed to solve: process "/bin/sh -c ./fetch-plugins.sh /plugins/" did not complete successfully: exit code: 2
make: *** [Makefile:129: image] Error 1
```

illume (approved):
👍 thanks

PR #2219, opened 2024-07-31
app: frontend: plugins: Update typescript version to 5.5.4 and typedoc

sniok:
Typescript version is inconsistent across packages, and the one in the frontend/ is quite old

Having latest typescript version would help with some issues I'm having while refactoring apiProxy and KubeObject type

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -161,7 +161,10 @@
     "jest": "^29.7.0",
     "mime-types": "^2.1.35",
     "octokit": "^4.0.2",
-    "typescript": "5.5.3"
+    "typescript": "5.5.4"
+  },
+  "overrides": {
+    "typescript": "5.5.4"
   },
   "dependencies": {
     "copyfiles": "^2.4.1",
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -31,7 +31,7 @@
         "jest": "^29.7.0",
         "mime-types": "^2.1.35",
         "octokit": "^4.0.2",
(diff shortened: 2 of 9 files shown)


sniok:
Testing done:

- [ ] Generated new API docs and previewed them in the headlamp website

Actions needed after this:

-  Update links in the documentation that point to any page in API (since folder structure is different now)
-  Explore more options/plugins for typedoc to make API reference more approachable

For now I'm keeping the current generated API docs as is

app/package.json
```diff
@@ -161,7 +161,10 @@
     "jest": "^29.7.0",
     "mime-types": "^2.1.35",
     "octokit": "^4.0.2",
-    "typescript": "5.5.3"
+    "typescript": "5.5.4"
+  },
+  "overrides": {
```
illume, line 166: Just out of curiosity, why was this needed?

sniok: @kinvolk/eslint-config depends on typescript@4 which results in two copies of typescript in node_modules. Forcing @5 fixes that and eslint seems to work just fine

illume (approved):
👍 Thanks.

I looked into if we can get away with upgrading to a lesser typedoc version that doesn't break the page structure... and still upgrade typescript. https://github.com/TypeStrong/typedoc/releases But it seems page structure changes came at the same time as TypeScript 5.5 support.

Would you mind creating an issue (with next release milestone) with the things you noticed that need to be fixed later?

PR #2220, opened 2024-07-31
frontend: plugins: Remove unused if-env package

sniok:
It wasn't used and was depending on a bunch of unsupported polyfills


illume (approved):
👍 thanks!

illume:
for 
- https://github.com/headlamp-k8s/headlamp/issues/1992

PR #2222, opened 2024-07-31
frontend: Update typedoc version

sniok:
Required for new typescript version https://github.com/headlamp-k8s/headlamp/pull/2219

The latest typedoc and markdown have some breaking changes resulting in a different file format. It's a bit easier to navigate than the current one but still not the best. There's some work needed to make the API docs more approachable.

Testing done:

- [x] Generated new API docs and previewed them in the headlamp website

Actions needed after this:

- [ ] Update links in the documentation that point to any page in API (since folder structure is different now)
- [ ] Explore more options/plugins for typedoc to make API reference more approachable 

For now I'm keeping the current generated API docs as is

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -127,9 +127,9 @@
         "redux-mock-store": "^1.5.4",
         "resize-observer-polyfill": "^1.5.1",
         "storybook": "^7.6.7",
-        "typedoc": "0.22.10",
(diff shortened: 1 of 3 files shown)


sniok:
This was originally in different PR but It's not building without new typescript version, Moving this back into typescript version bump PR

PR #2223, opened 2024-07-31
app: Replace symlinks in node_modules by their target in Mac builds

joaquimrocha:
Otherwise, Mac will fail to notarize the app.


illume (approved):
👍 lgtm

joaquimrocha:
Just tested and this fixes the issue.

Issue #198, opened 2021-02-15
Implement a graph / observability view


joaquimrocha:
Graph views are useful for a quick, visual representation of the relation between cluster resources. Moreover, we can use information from tools like Inspektor Gadget to implement observability on top of it.

Deps:
* [x] Add a graph view: #2225 
* [x] Add the ability for plugins to change the graph
* [x] Add observability via a plugin

illume:
@sniok I think this issue can be closed?

This observability task is I guess tracked as part of the inspektor gadget plugin? 

> Add observability via a plugin

Is there another issue for this work? I guess we can close this issue and drop the link to it here.

illume:
This is done I think.

PR #2225, opened 2024-08-01
frontend: Add Map

sniok:
This PR introduces a new Map view under /map route
It displays most of the kubernetes resource as a graph

Implements https://github.com/headlamp-k8s/headlamp/issues/198

---

# Description of all the changes in this PR

## frontend: Add Map page and route
    
Introduces new resource Map view under /map route
This is the main commit introducing the map under `src/components/resourceMap`

##  frontend: Allow Routes to take full width, add isFullWidth route parameter

Map needs as much space as possible.
By default Layout Container will have a max width.
New isFullWidth parameter on the route removes max width
    
---

Out of scope for this PR:

- Tests
- Exposing Map to plugins
- Ability to extend map from plugins
- Full coverage of all resources (some resources are not implemented yet)
- Full coverage of all possible connections

---

Testing done

- [x] Tested with different clusters (k3s, azure aks)
- [x] Load testing with kwok
- [x] Listing, filtering, grouping, updating resources


illume:
Related to https://github.com/headlamp-k8s/headlamp/issues/198

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+                  <Chip
+                    label={t('Group By: {{ name }}', { name: t('Namespace') })}
+                    color={groupBy === 'namespace' ? 'primary' : undefined}
+                    variant={groupBy === 'namespace' ? 'filled' : 'outlined'}
+                    icon={groupBy === 'namespace' ? <Icon icon="mdi:check" /> : undefined}
+                    onClick={() => setGroupBy(it => (it === 'namespace' ? undefined : 'namespace'))}
+                    sx={{
+                      lineHeight: '1',
+                    }}
+                  />
```
joaquimrocha, line 235: Should these be abstracted into their own components? Seems like we are repeating similar code.

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+                }}
+                onBackgroundClick={() => {
+                  if (selectedGroup && selectedNodeId === selectedGroup.id) {
+                    const parent = getParentNode(fullGraph, selectedGroup.id);
+                    if (parent) {
+                      setSelectedNodeId(parent.id);
+                    }
+                  } else if (selectedGroup) {
+                    setSelectedNodeId(selectedGroup.id);
+                  }
```
joaquimrocha, line 301: How about?

```typescript
if (selectedGroup) {
  let nodeId = selectedGroup.id;
  if (selectedNodeId === nodeId) {
    const parent = getParentNode(fullGraph, selectedGroup.id);
    if (parent) {
      nodeId = parent.id;
    }
  }
  setSelectedNodeId(nodeId);
}
```

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+        </CustomThemeProvider>
+        {selectedNode && (
+          <GraphNodeDetails
+            node={selectedNode}
+            close={() => {
+              if (selectedGroup) {
+                setSelectedNodeId(selectedGroup.id);
+              } else {
+                setSelectedNodeId(defaultNodeSelection);
+              }
```
joaquimrocha, line 330: How about?

```typescript
setSelectedNodeId(selectedGroup?.id ?? defaultNodeSelection);
```

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+        sx={{
+          position: 'relative',
+          height: height ?? '800px',
+          display: 'flex',
+          flexDirection: 'row',
+          flex: 1,
+        }}
+      >
+        <CustomThemeProvider>
+          <Box
```
joaquimrocha, line 195: This whole header maybe should be abstracted, if you don't have to pass over a 100 props.

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+              }
+            }}
+          />
+        )}
+      </Box>
+    </GraphViewContext.Provider>
+  );
+}
+
+function CustomThemeProvider({ children }: { children: ReactNode }) {
```
joaquimrocha, line 339: Could this be integrated into the main theme?

frontend/src/components/resourceMap/KubeObjectGlance.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const kind = resource.kind;
+
+  const sections = [];
+
+  if (kind === 'Pod') {
+    sections.push(
+      <Box display="flex" gap={1} alignItems="center" mt={2} flexWrap="wrap" key="pod">
+        <Box>{makePodStatusLabel(resource)}</Box>
+        {resource.spec.containers.map((it: any) => (
```
joaquimrocha, line 28: any -> KubeContainer? (same for others)

frontend/src/components/resourceMap/KubeObjectGlance.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const [events, setEvents] = useState<Event[]>([]);
+  useEffect(() => {
+    Event.objectEvents(resource).then(it => setEvents(it));
+  }, []);
+
+  const kind = resource.kind;
+
+  const sections = [];
+
+  if (kind === 'Pod') {
```
joaquimrocha, line 24: Maybe abstracting each resource component into its own functional component would make the code easier?

frontend/src/components/resourceMap/details/GraphNodeDetails.tsx
```diff
@@ -0,0 +1,65 @@
...
+      })}
+    >
+      <Button
+        onClick={() => {
+          close();
+        }}
+        startIcon={<Icon icon="mdi:close" />}
+      >
+        {t('Close')}
+      </Button>
```
joaquimrocha, line 61:
```suggestion
      <Box px={1}>
        <ActionButton
          onClick={() => {
            close();
          }}
          icon="mdi:close"
          description={t('Close')}
        />
      </Box>
```

frontend/src/components/resourceMap/details/KubeNodeDetails.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * Shows details page for a given Kube resource
+ */
+export const KubeObjectDetails = memo(({ resource }: { resource: KubeObject }) => {
+  const kind = resource.kind;
+  const { name, namespace } = resource.metadata;
+
+  return (
+    <Box sx={{ overflow: 'hidden' }}>
+      <Box sx={{ marginTop: '-70px' }}>
+        {kind === 'Pod' && <PodDetails name={name} namespace={namespace} />}
```
joaquimrocha, line 48: Seems like most of these take the name + namespace.
I wonder if we could just have a map like
```typescript
{
  Service: ServiceDetails,
  Pod: PodDetails,
  ...
}
```

I have also been thinking that maybe each KubeObject should have a list view and a details view components assigned to them, so we could extract this info. Maybe in a future PR.

frontend/src/components/resourceMap/graph/graphLayout.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  if ('resource' in node.data) {
+    const res = node.data.resource;
+    if (res.kind === 'Deployment') {
+      layer = 1;
+    }
+    if (res.kind === 'ReplicaSet' || res.kind === 'ServiceAccount' || res.kind === 'CronJob') {
+      layer = 3;
+    }
+
```
joaquimrocha, line 43: I am puzzled for where the layer 2 is!

frontend/src/components/resourceMap/graph/graphLayout.tsx
```diff
@@ -0,0 +1,246 @@
...
+/**
+ * To increase readability of the graph we can sort nodes left-to-right
+ * Where more 'owner' nodes like Deployment or ReplicaSet are on the left
+ */
+function getPartitionLayer(node: KubeObjectNode) {
+  let layer: any = undefined;
+
+  if ('resource' in node.data) {
+    const res = node.data.resource;
+    if (res.kind === 'Deployment') {
```
joaquimrocha, line 37: How about?

```ts
const partitionLayers = [
  [
    'Deployment',
  ],
  [
    'ReplicaSet',
    'ServiceAccount',
  ],
  ...
];

if ('resource' in node.data) {
  const kind = node.data.resource.kind;
  const layer = partitionLayers.findIndex(partition => kind);

  if (layer !== -1) {
    return layer;
  }
}
```
If we could return a negative value to mean it wasn't found, then it'd be even better.

frontend/src/components/resourceMap/graph/graphLayout.tsx
```diff
@@ -0,0 +1,246 @@
...
+    height: 70,
+  },
+};
+
+/**
+ * To increase readability of the graph we can sort nodes left-to-right
+ * Where more 'owner' nodes like Deployment or ReplicaSet are on the left
+ */
+function getPartitionLayer(node: KubeObjectNode) {
+  let layer: any = undefined;
```
joaquimrocha, line 33: `let layer?: number;` ?

frontend/src/components/resourceMap/details/GraphNodeDetails.tsx
```diff
@@ -0,0 +1,65 @@
+import { Icon } from '@iconify/react';
+import { Button, Card } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { GraphNode, KubeObjectNode } from '../graph/graphModel';
+import { KubeObjectDetails } from './KubeNodeDetails';
+
+interface GraphNodeDetailsSection {
+  id: string;
+  nodeType?: string;
+  render: (node: GraphNode) => any;
```
joaquimrocha, line 10: any -> ReactNode ?

joaquimrocha (changes requested):
Thanks @sniok . I left a few comments. Overall, I think we can reduce the number of `any` types we got now.
I found a possible issue with the interaction: if you enable the error filter and you don't have errors (or any filter without results I guess), you end up not seeing anything. Even the breadcrumbs don't show up, so I felt a bit lost.

Another thing to consider, maybe in a future PR: as an alternative to passing the namespace and name (but still supporting both) to each kube resource, I think we could pass the object of the resource itself optionally, and render initially with it, until the item is gathered from the server. This would have the effect of e.g. preloading the last known state of that object, if you already have it from the graph or a list view, and thus it is immediate for the user, instead of them having to wait for the server responses.

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
sniok: > If this is about making the namespaces autocomplete look a certain way for the map view,

Yes, kind of, it fits better this way on the Map but also imho it looks better in the rest of the app as well.
To keep it simple I didn't introduce any new variant and decided to just change it everywhere. 

I don't have any strong opinions about this look though, so if you want to keep the look for the rest of the app the same I can add that

frontend/src/components/resourceMap/GraphView.tsx
sniok: Our current app theme is something I want to work on separately

frontend/src/components/resourceMap/graph/graphLayout.tsx
sniok: a layer can have a negative value, meaning layout algorithm would place all those nodes on the left and we want lack of layer value so it can place it where it fits best.

frontend/src/components/resourceMap/details/KubeNodeDetails.tsx
sniok: > I have also been thinking that maybe each KubeObject should have a list view and a details view components
yeah that'd be nice

frontend/src/components/resourceMap/KubeObjectGlance.tsx
sniok: if this grows larger we can refactor it. for now I think this is simple enough

frontend/src/components/resourceMap/GraphView.tsx
sniok: there's a lot of props that'd need to be passed. I agree that this part would make sense as a separate component but I think for now this is okay

frontend/src/components/resourceMap/GraphView.tsx
sniok: that whole logic block was not necessary, replaced it with something simpler

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,390 @@
...
+      setExpandAll(false);
+    }
+  }, [graphSize]);
+
+  return (
+    <GraphViewContext.Provider
+      value={useMemo(
+        () => ({ nodeSelection: selectedNodeId, highlights, setNodeSelection: setSelectedNodeId }),
+        [selectedNodeId, setSelectedNodeId, highlights]
+      )}
```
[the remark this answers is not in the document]
sniok: this shouldn't be an issue but just in case moved it up

sniok:
> Overall, I think we can reduce the number of `any` types we got now.

Now that the KubeObject refactor is merged a lot of `any` were not needed anymore. Also removed most of the other usages of `any`

> Another thing to consider, maybe in a future PR: as an alternative to passing the namespace and name (but still supporting both) to each kube resource, I think we could pass the object of the resource itself optionally, and render initially with it, until the item is gathered from the server. This would have the effect of e.g. preloading the last known state of that object, if you already have it from the graph or a list view, and thus it is immediate for the user, instead of them having to wait for the server responses.

I think this would be better solved by populating the cache for individual resource query, this way it would work everywhere where `Resource.useGet` is used and not just the Details component. There are also some other ideas how to make UI a bit more snappier (optimistic updates, preloading on link hover, etc). I'd love to work on that but in a different PR

sniok:
> I found a possible issue with the interaction: if you enable the error filter and you don't have errors (or any filter without results I guess), you end up not seeing anything. Even the breadcrumbs don't show up, so I felt a bit lost.

Added a little message for no data

image

frontend/src/components/common/NamespacesAutocomplete.tsx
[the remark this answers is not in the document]
joaquimrocha: I think it does look nicer and is more consistent with the new table fields too.

illume:
I have a suggestion.... Can you please break out the early commits as separate PRs already, so people can start reviewing/merging those? Especially the dependencies.

I would prefer if this could be broken down some more into small commits. Because we can't go back afterwards and add these commits in the future. It also makes it easier to review if they are in smaller PRs so folks can review smaller chunks.

Fields and functions should be documented.

@joaquimrocha  We spoke about adding a Beta label with a link to feedback. Will we do that? I suggest linking to the issue for this. (which needs to be linked in the PR description) Another option is we can start adding a discussion thread for new features?

(Yeah, it would be nice to have some more stories/tests, to understand the states and to allow better testing of components in isolations... But I think smaller commits is something we can't do later if we don't do it now.)

frontend/src/components/resourceMap/useQueryParamsState.tsx
```diff
@@ -0,0 +1,43 @@
+import { Dispatch, SetStateAction, useEffect, useState } from 'react';
+import { useLocation } from 'react-router';
+
+type UseQueryParamsStateReturnType<T> = [T, Dispatch<SetStateAction<T>>];
+export function useQueryParamsState<T extends string | undefined>(
```
illume, line 5: This is a good candidate for extraction.

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
```diff
@@ -0,0 +1,167 @@
+import { Icon } from '@iconify/react';
```
illume, line 1: It would be good to have what these sources are documented somewhere.

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
```diff
@@ -0,0 +1,167 @@
...
+import NetworkPolicy from '../../../../lib/k8s/networkpolicy';
+import Pod from '../../../../lib/k8s/pod';
+import Secret from '../../../../lib/k8s/secret';
+import Service from '../../../../lib/k8s/service';
+import { GraphEdge, GraphSource } from '../../graph/graphModel';
+import { getKindGroupColor, KubeIcon } from '../../kubeIcon/KubeIcon';
+import { makeKubeObjectNode, makeKubeToKubeEdge } from '../GraphSources';
+import { matchesSelector } from './WorkloadSource';
+
+const serviceSource: GraphSource = {
```
illume, line 15: The filenames make me think these are react components, but they are not AFAIK.

frontend/src/components/resourceMap/graph/graphFiltering.ts
```diff
@@ -0,0 +1,90 @@
...
+      type: 'related';
+      id: string;
+    }
+  | {
+      type: 'custom';
+      label: string;
+      filterFn: (node: GraphNode) => boolean;
+    };
+
+export function filterGraph(nodes: GraphNode[], edges: GraphEdge[], filters: GraphFilter[]) {
```
illume, line 26: It seems a lot of the functions in graph/ could be extracted into separate commits. The graphFiltering.ts should be in the same commit as the graphFiltering.test.ts

frontend/src/components/resourceMap/graph/graphGrouping.tsx
```diff
@@ -0,0 +1,318 @@
...
+    const nonGroup = it.id.includes('undefined');
+    const hasOneMember = it.data.nodes.length === 1;
+
+    return nonGroup || (hasOneMember && !allowSingleMemberGroup) ? it.data.nodes : [it];
+  });
+
+  return result;
+};
+
+export function groupGraph(
```
illume, line 126: There's a few functions in here still needing docs.

frontend/src/components/resourceMap/kubeIcon/KubeIcon.tsx
```diff
@@ -0,0 +1,148 @@
...
+  security: `oklch(${lightness} ${chroma} 275.16)`,
+  configuration: `oklch(${lightness} ${chroma} 320.03)`,
+  other: `oklch(${lightness} 0 215.25)`,
+} as const;
+
+const getKindColor = (kind: string) => kindGroupColors[getKindGroup(kind) ?? 'other'];
+export const getKindGroupColor = (group: keyof typeof kindGroupColors) =>
+  kindGroupColors[group] ?? kindGroupColors.other;
+
+export function KubeIcon({
```
illume, line 115: docs needed

frontend/src/components/resourceMap/nodes/KubeObjectStatus.tsx
```diff
@@ -0,0 +1,35 @@
+import { KubeObject, Workload } from '../../../lib/k8s/cluster';
+import Pod from '../../../lib/k8s/pod';
+import { getReadyReplicas, getTotalReplicas } from '../../../lib/util';
+
+type KubeObjectStatus = 'error' | 'success' | 'warning';
+
+function getPodStatus(pod: Pod): KubeObjectStatus {
```
illume, line 7: docs needed

frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -0,0 +1,388 @@
...
+              {graphSize < 50 && (
+                <ChipToggleButton
+                  label={t('Expand All')}
+                  isActive={expandAll}
+                  onClick={() => setExpandAll(it => !it)}
+                />
+              )}
+            </Box>
+
+            <div style={{ flexGrow: 1 }}>
```
illume, line 269: html tags shouldn't be used

frontend/src/components/resourceMap/search/ResourceSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }}
+        onChange={(e, value) => {
+          if (value && typeof value !== 'string') {
+            onSearch(value);
+          }
+        }}
+        options={results}
+        renderOption={(props, it) => {
+          return (
+            <li {...props}>
```
illume, line 62: html

frontend/src/components/resourceMap/search/ResourceSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useTranslation } from 'react-i18next';
+import { KubeObject } from '../../../lib/k8s/cluster';
+import { KubeIcon } from '../kubeIcon/KubeIcon';
+
+const MAX_RESULTS = 8;
+export function ResourceSearch({
+  resources,
+  onSearch,
+}: {
+  resources: KubeObject[];
```
illume, line 12: docs

frontend/src/components/resourceMap/sources/GraphSourcesView.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+function GraphSourceView({
+  source,
+  sourceData,
+  selection,
+  activeItemId,
+  setActiveItemId,
+  toggleSelection,
+}: {
+  source: GraphSource;
```
illume, line 49: docs

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
sniok: I'd like to document things like this in our docs as well, when we get plugin api for this ready

frontend/src/components/resourceMap/GraphView.tsx
sniok: why not? rendering a bunch of Boxes components noticeably affect performance here

frontend/src/components/resourceMap/search/ResourceSearch.tsx
sniok: I don't understand, even in the mui example they use li. It's a perfectly reasonable element to use here
https://mui.com/material-ui/react-autocomplete/#search-as-you-type

frontend/src/components/resourceMap/graph/graphFiltering.ts
illume: In order to help with a good review now, and maintenance in the future.

There's many separate pieces which could be moved into atomic commits. With related changes grouped together - for example the tests in a separate commit mentioned above.

sniok: isn't there a risk of losing context by introducing too much of splitting? having them in one commit makes it easier to understand how they relate directly. since this is an essential part of the map, wouldn't grouping it together provide a better view of the feature as a whole? at a certain point splitting everything into tiny commits might actually be harder to review since you have to switch contexts between different commits/PRs

frontend/src/components/resourceMap/GraphView.tsx
illume: This is a nit, you can ignore if you want.

> why not? 

For consistency.

> rendering a bunch of Box components noticeably affects performance here

There's several react components at the same level, so I'm skeptical if this is true.

frontend/src/components/resourceMap/nodes/KubeObjectNode.tsx
```diff
@@ -0,0 +1,184 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
illume, line 1: Can you please add justifications for disabling eslint and typescript things next to the disable lines?

Also, please check other changes in this PR if that's necessary too?

frontend/src/components/resourceMap/search/ResourceSearch.tsx
illume: nit, feel free to ignore

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
illume: I'd like things documented so I can review more easily.

frontend/src/components/resourceMap/KubeObjectGlance.tsx
illume: Having separate components here would be better for readability. Also better for inspecting in developer tools, and for adding tests.

frontend/src/components/resourceMap/graph/graphFiltering.ts
illume: Having a commit with just a filterGraph and it's related tests in the same commit would be best IMHO.

illume:
As well as the related issue, can you please add a list of what testing has been done to the PR description?

Has anyone done any accessibility testing yet?

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
sniok: I understand, I was talking about additional documentation on the website for plugin authors

sniok: added better docs to the GraphSource interface

frontend/src/components/resourceMap/sources/definitions/NetworkSource.tsx
sniok: changed to lowercase

frontend/src/components/resourceMap/nodes/KubeObjectNode.tsx
sniok: removed it, wasn't needed anymore

frontend/src/components/resourceMap/graph/graphFiltering.ts
sniok: moved it to this PR #2499

frontend/src/components/resourceMap/search/ResourceSearch.tsx
sniok: added

frontend/src/components/resourceMap/useQueryParamsState.tsx
sniok: moved to #2498

illume:
Here's the issue that should be added into the PR description:
- https://github.com/headlamp-k8s/headlamp/issues/198

I added a label "resourceMap" named after the folder name:
- https://github.com/headlamp-k8s/headlamp/labels/resourceMap

btw, Maybe the name could just be "map"? (Not Map, since it's not a single component, and not resourceMap because it's a bit awkward/long).  WDYT?

sniok:
> Here's the issue that should be added into the PR description:
> 
> * [Implement a graph / observability view #198](https://github.com/headlamp-k8s/headlamp/issues/198)

Added to the description

> btw, Maybe the name could just be "map"? (Not Map, since it's not a single component, and not resourceMap because it's a bit awkward/long). WDYT?

I think 'map' is a little too generic, also could clash with Map data structure

frontend/package.json
```diff
@@ -10,6 +10,7 @@
   "type": "module",
   "dependencies": {
     "@apidevtools/swagger-parser": "^10.0.3",
+    "@dagrejs/dagre": "^1.1.2",
```
illume, line 13: Would it be possible to add the dependencies one commit at a time? This is useful for seeing which things are pulled in... but more importantly if it causes a side effect in some other part of the app (as can happen with maybe another dependency we have already being upgraded/downgraded) then we can bisect it more easily. If they are all in one commit this is harder.

frontend/package.json
```diff
@@ -112,7 +116,7 @@
     "start": "cross-env REACT_APP_HEADLAMP_BACKEND_TOKEN=headlamp vite",
     "preview": "vite preview",
     "prebuild": "npm run make-version",
-    "build": "cross-env PUBLIC_URL=./ NODE_OPTIONS=--max-old-space-size=8096 vite build && npx shx rm -f build/frontend/index.baseUrl.html",
+    "build": "cross-env PUBLIC_URL=/ NODE_OPTIONS=--max-old-space-size=8096 vite build && npx shx rm -f build/frontend/index.baseUrl.html",
```
illume, line 119: This change wasn't mentioned in a commit message. I'm pretty sure this is depended on for baseURL functionality... does this still work?

frontend/package.json
sniok: It used to be in a separate commit 

https://github.com/headlamp-k8s/headlamp/pull/2225#discussion_r1816316172

illume: @joaquimrocha see request and reasoning above for adding dependencies in separate commits.

frontend/package.json
sniok: not sure why this was added, reverted

frontend/src/components/resourceMap/sources/definitions/workloadSource.tsx
```diff
@@ -0,0 +1,233 @@
+/* eslint-disable no-unused-vars */
```
illume, line 1: Can you please add justifications next to any places where you disable eslint/tsc?

frontend/src/components/resourceMap/useQueryParamsState.tsx
```diff
@@ -27,7 +27,7 @@ export function useQueryParamsState<T extends string | undefined>(
     const searchParams = new URLSearchParams(search);
     const paramValue = searchParams.get(param);
 
-    return paramValue !== null ? (decodeURIComponent(paramValue) as T) : initialState;
+    return paramValue !== null ? (decodeURIComponent(paramValue) as T) : undefined;
```
illume, line 30: Are the changes in this file intentional? I don't see any commit them for them.

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -251,6 +251,11 @@ function prepareRoutes(
         },
       ],
     },
+    {
+      name: 'map',
+      icon: 'mdi:map',
+      label: t('glossary|Map'),
```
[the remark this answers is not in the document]
illume: thanks, resolving.

frontend/src/components/resourceMap/GraphView.tsx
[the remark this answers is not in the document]
illume: @joaquimrocha can you please confirm the change?

frontend/src/components/resourceMap/GraphView.tsx
illume: It looks like these were moved into a ChipToggleButton, resolving.

frontend/src/components/resourceMap/GraphView.tsx
illume: Not sure where this is... @joaquimrocha can you please confirm the change?

frontend/src/components/resourceMap/GraphView.tsx
illume: looks done

frontend/src/components/resourceMap/GraphView.tsx
illume: Yes, this code seems strange. But let's skip this to avoid delay and argument.

frontend/src/components/resourceMap/details/GraphNodeDetails.tsx
illume: looks done

frontend/src/components/resourceMap/KubeObjectGlance.tsx
illume: This is done

frontend/src/components/resourceMap/GraphView.tsx
illume: Let's skip this.

frontend/src/components/resourceMap/search/ResourceSearch.tsx
illume: let's skip this

illume:
I went through a whole bunch of unresolved conversations and resolved the ones that were obvious to me. There's only a few left open now. Please both look at the remaining ones?

illume:
Can you rebase on main? This one was merged in this morning, and it would be good to test with that:
- https://github.com/headlamp-k8s/headlamp/pull/2486

illume:
It flashes the "No data to be shown" message when data is loading. Instead there should be a loading spinner.

illume:
There's an a11y problem with size.

### original size

Here you see the size of the text with default zoom.  I pasted in the table data for comparison.
image

### Zoomed twice

When I do the zoom 2x test, the map resizes to be smaller. It doesn't get bigger.

Here's how it looks zoomed twice... You will notice that the map gets even smaller than usual when zooming in twice.
image

The original size of the map is smaller than our table data which is a good readable size. There's a lot of people who won't be able to read the default size of the map data. Zooming makes the map even smaller.

sniok:
> The original size of the map is smaller than our table data which is a good readable size. There's a lot of people who won't be able to read the default size of the map data. Zooming makes the map even smaller.

image

The font size is 16px which is accessible. Maps allows you to zoom in (scroll wheel or + - buttons on the bottom left, not the browser zoom in) to 120% size which will result in text that is even bigger than 16px.
By default map fits to view which may result in smaller text but that is expected.

> When I do the zoom 2x test, the map resizes to be smaller. It doesn't get bigger.

Using "browser zoom" behavior wasn't tested and can be improved in next PRs

frontend/src/components/resourceMap/useQueryParamsState.tsx
sniok: Yes, initial state wasn't initialized properly, moved it into a separate commit

frontend/src/components/resourceMap/sources/definitions/workloadSource.tsx
sniok: removed

illume:
When hovering over (Deployment nginx 8) it shows the cursor as if there is a link there. But clicking it does nothing. Links/buttons should do something or not be links IMHO.

image

illume:
> The font size is 16px which is accessible. 
Yes, but the view is zoomed... You could have font-size 1000px there and if it's zoomed out for enough it could be tiny.

sniok:
> When hovering over (Deployment nginx 8) it shows the cursor as if there is a link there. But clicking it does nothing. Links/buttons should do something or not be links IMHO.

It is a button/link, it wasn't reacting due to a small issue, it is now fixed

illume:
I first tried to click on ("Group By" Namespace) because I thought it was a toggle. It didn't do anything.
Then I realized maybe it's linked with the other "Group By:" tags... so I clicked on those and it did nothing. ("Group By" Namespace) stayed selected.

image

sniok:
> I first tried to click on ("Group By" Namespace) because I thought it was a toggle. It didn't do anything. Then I realized maybe it's linked with the other "Group By:" tags... so I clicked on those and it did nothing. ("Group By" Namespace) stayed selected.

same issue as the above, please do a git pull

illume:
On a remote system on first load it jumps about animating for a few seconds until it seems to eventually stabilize. It's not too bad on a local k8s system with not much going on. I guess the data is still loading and changes are happening as it loads. I usually use the reduced animation setting (Animation Effects off it's now called in win 11) so I checked it was still on.

illume:
> > I first tried to click on ("Group By" Namespace) because I thought it was a toggle. It didn't do anything. Then I realized maybe it's linked with the other "Group By:" tags... so I clicked on those and it did nothing. ("Group By" Namespace) stayed selected.
> 
> same issue as the above, please do a git pull

Thanks. Clicking on them works now.

I think there should be some visual link between that toggle button set. For example by using a mui toggle button group. https://mui.com/material-ui/react-toggle-button/#color

image

illume:
The tab focus/highlight is not visible on this (Workloads, Storage, +1) button when tabbing through.
image

frontend/package.json
illume: Discussed with Joaquim. We're going to leave it as is.

frontend/src/components/resourceMap/sources/definitions/workloadSource.tsx
illume: thanks

frontend/src/components/resourceMap/useQueryParamsState.tsx
illume: Cool, thanks.

illume:
I ran into an issue which I can't figure out how to reproduce now unfortunately. But leaving a note here to investigate more later. Click the zoom a few times the content is off screen...
image

illume:
1. Click on a pod, so only a pod in the graph and the details sidebar is visible.
2. Close the Details sidebar, so only the pod in that namespace is shown in the graph.
3. Refresh browser
4. Now the map view shows a bunch of other things, not just the pod.

The URL is http://localhost:3000/c/cluster/map?group=namespace&namespace=XXX&node=Namespace-yyyy before refresh.
After refresh the URL is: http://localhost:3000/c/cluster/map?group=namespace&namespace=XXX

illume:
Sometimes clicking on Home breadcrumb doesn't take you home. It is leaving the namespace filter set to the namespace I was in with no breadcrumbs visible.

I'm not sure how to reproduce this yet.

illume:
Sometimes there is significant CPU usage and my browser freezes, and I can't interact with anything. After some seconds it seems ok again.

Not sure how to reproduce (I just click about for a while and seems to happen at random).

I captured it once with the Edge devtools "performance monitor" active... maybe that gives a hint.
image

frontend/src/components/resourceMap/GraphView.tsx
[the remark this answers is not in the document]
illume: It looks fine to me. J is away today, so I'll resolve.

frontend/src/components/resourceMap/GraphView.tsx
illume: Let's skip this. J is away today, so I'll resolve.

illume:
I've finished reviewing.

The only convo left open I need to check is this one: https://github.com/headlamp-k8s/headlamp/pull/2225#discussion_r1816412822

The only issue blocking this is the text size/zoom issue. https://github.com/headlamp-k8s/headlamp/pull/2225#issuecomment-2451607716

illume:
There's no keyboard control for this component to allow opening the tree below the top level things. Meaning people can't select to see Pods only because they can't toggle open the Workloads sub tree.

image

illume:
In medium browser width, I can't really see anything because everything is very tiny.

image

illume:
Resizing the window from medium to large does not seem to trigger it to layout again, and everything remains tiny. Even though probably the layout should change with a wider browser.

If I refresh the browser it does seem to redo the layout to fit in the bigger window. Also, sometimes when I click the maximise button it does the layout properly with the bigger window.

image

illume:
The nothing selected state for this looks strange:

image

vyncent-t:
works pretty solid for both web and app mode, currently have disabled gpu on so some icons may not be the exact same as current for me in app mode, been adding resources and looking at different sizes, seen a few of the same resize issues rene pointed to

skoeva:
^^ echoing the previous comments, seems to work pretty well but there are some concerns when resizing as well as performance-wise (but this may be unrelated, I have noticed the app can be a CPU hog sometimes lately)

sniok:
Thanks everyone for the review, I've created an issue to keep track of all the non-blocking items here https://github.com/headlamp-k8s/headlamp/issues/2519

As we're nearing the release I'd like to minimize amount of big changes here, so let me know if anyone has objections to shipping this as is

frontend/src/components/resourceMap/GraphView.tsx
sniok: It wasn't moved, it was just replaced with a couple of lines in the same place. GraphView.tsx:283

illume:
> I don't see this as a blocking issue. If the map is zoomed out to fit everything the items will be small, that's expected

It's an accessibility issue where most people can't read it.

frontend/src/components/resourceMap/GraphView.tsx
sniok: @joaquimrocha and I had a call. Changed the logic here to remove "go back when on background click" behavior, now it will only close the side panel on background click

sniok:
By default it now zooms in to 100%, with top-left origin if things don't fit or centered as before if things fit

image

Fit to view button will fit everything into view 
100% button will zoom in to 100%, same as default

image

Max zoom is now also increased to 200%

Here are the changes
https://github.com/headlamp-k8s/headlamp/compare/1791c11944d00a1953fe60104781789588b88863..cb55d2a4bcbb92f117e35a37f29c6ad71b883ba8

illume:
Thanks for addressing that default size of text issue. Appreciate it.

I would like to say this is a really great accomplishment, and excellent work. 🎉

Especially by Olek (but also you Joaquim, plus Wilder, Grant, Ivelisse, and Chris for all the design work, as well as Vincent & Evangelos). With many weeks of reporting problems, it can feel unending (So, also thanks for putting up with all the wishes and issue reports in these last weeks/months).

joaquimrocha (approved):

Issue #1962, opened 2024-05-06
Option to pass Cluster Settings via Config for user Groups


derbauer97:
Hello! We're delighted users of Headlamp!

Our objective is to provide users with the capability to view their Resource Status and Definitions without necessitating the development of a custom UI. It's important to note that our users possess restricted access limited to their respective Namespace.

We've been actively utilizing Custom Resources, and to facilitate their access via Headlamp while maintaining restricted permissions, we need to configure the namespace within the cluster settings under allowed Namespaces otherwise if not possible it would be greate to have an option to filter namespaces in the CustomResource Overview like it is possible with Workloads.

I found this issue which i can confirm is fixed and described a similar problem: https://github.com/headlamp-k8s/headlamp/issues/1553

joaquimrocha:
> Hello! We're delighted users of Headlamp!

Nice! Happy to meet you!

> Our objective is to provide users with the capability to view their Resource Status and Definitions without necessitating the development of a custom UI. It's important to note that our users possess restricted access limited to their respective Namespace.

Just slightly related, but we are improving how users get aware that they need to configure the list of "allowed" namespaces if they don't have permissions to list all namespaces.

> We've been actively utilizing Custom Resources, and to facilitate their access via Headlamp while maintaining restricted permissions, we need to configure the namespace within the cluster settings under allowed Namespaces otherwise if not possible it would be greate to have an option to filter namespaces in the CustomResource Overview like it is possible with Workloads.

Yeah. I think that makes sense. We will get to it.

> I found this issue which i can confirm is fixed and described a similar problem: #1553

Thank you. I will close it.

Guilamb:
so I've made a pull request that adds a new page that displays the CRs and enables Namespace filtering, in order to maintain the namespace flow as in the workloads.

I hope this helps. If you have any suggestions or improvements, please feel free to mention me or send me a message on Slack. 😄

Issue #955, opened 2023-02-13
Improve UX for CRs


gberche-orange:
## Expected behavior

As an headlamp user
In order to list/read/edit/update/delete custom resources
I need headlamp to make it as easy to deal with CR than with built-in K8S resources

See related screenshots of current similar support in lens that leverages the CRDs [additionalPrinterColumns](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns) to display the CR fields to display in the table view

![image](https://user-images.githubusercontent.com/4748380/218525613-da9b8633-0a54-4ac0-a189-16ba0ac83d73.png)
![image](https://user-images.githubusercontent.com/4748380/218526091-e402b0f4-664f-46a6-abb4-100efce262b7.png)
![image](https://user-images.githubusercontent.com/4748380/218526177-c3fc080d-e3f2-4477-be11-2d8fcc5ae063.png)

See related snapshots in k9s
![image](https://user-images.githubusercontent.com/4748380/218530400-366ae2fd-44f9-4e97-a646-ab2e86373fd9.png)
browsing down provides the describe view
![image](https://user-images.githubusercontent.com/4748380/218530658-6ec8b271-6f38-4fb4-bc23-3f3a153bae86.png)
or the full yaml view

See related snapshot of structural schema support for creating/updating CRs in openshift 
![openshift2](https://user-images.githubusercontent.com/4748380/219031916-48d6589c-bb98-40cf-a0b4-b3d96974b67c.gif)

## Current behavior

In version 1.25.3, it is hard to discover CRs are indeed listed:

Once one the CRD list
![image](https://user-images.githubusercontent.com/4748380/218433544-80910280-30d9-48f3-a0dc-b117ca323745.png)
The CRD page displays some details of the CRD
![image](https://user-images.githubusercontent.com/4748380/218528567-2214de51-0b2f-4899-a85e-a33eefb2d21a.png)
and only after scrolling at the bottom of the page, the CR are displayed in the "Objects" section
![image](https://user-images.githubusercontent.com/4748380/218529183-64ec96b0-dcd9-403e-983d-174cdcfc85c4.png)but without honoring the additionalPrinterColumn
![image](https://user-images.githubusercontent.com/4748380/218533343-0e0d6bd7-5e44-4e92-ba28-8a93b3d09e30.png)

[...]

PR #2226, opened 2024-08-01
frontend: Add /crs page to display a list of all instantiated custom resources.

Guilamb:
# Description
This pull request addresses the issue of restricted user access to custom resources and the need to view only the instantiated custom resources (#955 ), by introducing a new /crs page in Headlamp that displays a list of all instantiated custom resources, that can be filtered by category.

The /crs page provides users with the capability to view their Resource Status and Definitions without requiring the development of a custom UI. It also allows users to filter namespaces, similar to the existing functionality for Workloads, ensuring restricted permissions are maintained.

cr

# Changes
To implement this feature, the following changes were made:
* Added a page that list instantiated Custom resources
   * Used a ResourceListView with its namespaceFilter
* Added a route and a sideBarItem

Fixes: #1962

### DIFF
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -71,6 +71,12 @@ function prepareRoutes(
         {
           name: 'crds',
           label: t('glossary|Custom Resources'),
+          subList: [
+            {
+              name: 'crs',
+              label: t('glossary|CR Instances'),
+            },
+          ],
         },
       ],
     },
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -233,10 +233,11 @@ function prepareRoutes(
     {
       name: 'crds',
       label: t('glossary|Custom Resources'),
+      icon: 'mdi:puzzle',
       subList: [
         {
           name: 'crs',
-          label: t('glossary|CR Instances'),
+          label: t('translation|Instances'),
         },
       ],
     },
--- a/frontend/src/components/Sidebar/prepareRoutes.ts
+++ b/frontend/src/components/Sidebar/prepareRoutes.ts
@@ -68,16 +68,6 @@ function prepareRoutes(
           name: 'nodes',
           label: t('glossary|Nodes'),
         },
-        {
-          name: 'crds',
-          label: t('glossary|Custom Resources'),
-          subList: [
-            {
-              name: 'crs',
-              label: t('glossary|CR Instances'),
-            },
-          ],
-        },
       ],
     },
     {
@@ -237,6 +227,16 @@ function prepareRoutes(
         {
           name: 'validatingWebhookConfigurations',
           label: t('glossary|Validating Webhook Configurations'),
+        },
+      ],
+    },
+    {
+      name: 'crds',
+      label: t('glossary|Custom Resources'),
+      subList: [
+        {
+          name: 'crs',
+          label: t('glossary|CR Instances'),
         },
       ],
     },
--- a/frontend/src/components/App/icons.ts
+++ b/frontend/src/components/App/icons.ts
@@ -311,6 +311,9 @@ const mdiIcons = {
     'expand-all': {
       body: '<path fill="currentColor" d="M18 8H8v10H6V8a2 2 0 0 1 2-2h10zm-4-6H4a2 2 0 0 0-2 2v10h2V4h10zm8 10v8a2 2 0 0 1-2 2h-8a2 2 0 0 1-2-2v-8a2 2 0 0 1 2-2h8a2 2 0 0 1 2 2m-2 3h-3v-3h-2v3h-3v2h3v3h2v-3h3z"/>',
     },
+    puzzle: {
+      body: '<path fill="currentColor" d="M20.5 11H19V7a2 2 0 0 0-2-2h-4V3.5A2.5 2.5 0 0 0 10.5 1A2.5 2.5 0 0 0 8 3.5V5H4a2 2 0 0 0-2 2v3.8h1.5c1.5 0 2.7 1.2 2.7 2.7S5 16.2 3.5 16.2H2V20a2 2 0 0 0 2 2h3.8v-1.5c0-1.5 1.2-2.7 2.7-2.7s2.7 1.2 2.7 2.7V22H17a2 2 0 0 0 2-2v-4h1.5a2.5 2.5 0 0 0 2.5-2.5a2.5 2.5 0 0 0-2.5-2.5" />',
+    },
   },
   aliases: {
     'more-vert': {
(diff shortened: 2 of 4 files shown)


Guilamb:
For the Build Frontend / test (20.x, ubuntu-22.04) action, should I edit the snapshots to add the new sidebar item ?

joaquimrocha:
I think that at this point it probably makes more sense to have Custom Resources as a top level sidebar entry. Because Custom Resources + Custom Resource Instances looks a bit odd as separate items. But we should do that in a separate PR.

Guilamb:
> I think that at this point it probably makes more sense to have Custom Resources as a top level sidebar entry. Because Custom Resources + Custom Resource Instances looks a bit odd as separate items. But we should do that in a separate PR.

I see. I've been looking at Lens to get an idea of how they organized their CRD section:
image


So basically, they have a CRD top-level entry with a definition page, and for each API group, they have an entry with the CRDs of the same group. 
I feel like this would be difficult to use when you have a large number of CRDs. I think that having a definition entry and an instances entry is a good solution.

I've also looked at how it is on the OpenShift console, but there is only a CRD entry, similar to how it currently is in Headlamp.
image

joaquimrocha:
I want to understand better why the instances view is needed. What's the value it brings that you don't get with the current Custom Resources page in Headlamp. IMO, just like you have to navigate between resource kinds across the UI (Pod, ConfigMaps, etc.) before you see the list of definitions, it's similar when going to the list of Resource kinds in the current CR page in Headlamp (where the first column there has the kinds).

Guilamb:
> I want to understand better why the instances view is needed. What's the value it brings that you don't get with the current Custom Resources page in Headlamp. IMO, just like you have to navigate between resource kinds across the UI (Pod, ConfigMaps, etc.) before you see the list of definitions, it's similar when going to the list of Resource kinds in the current CR page in Headlamp (where the first column there has the kinds).

Okay, I'll try to expose what I've noticed using Headlamp to manage crds t'ill now.
The current CRD view is excellent for discovering CRDs. I can easily  search for a name, a category or a group and easily found out what the CRD does.
However, when you try to manage custom resources created with these CRDs, it quickly becomes complicated to find your way around, especially if you are not the only one using the cluster, and even more if not all users have the same roles. 

The objective of the CR instance page is to ease the management, navigation, and discovery of custom resources.

I have in mind two scenarios for which the CR instances view would be useful. 
I'll try to illustrate this as much as possible.

### Scenario 1: Two developers with the same roles in the same cluster
---

- **Dev 1**: Instantiates a MongoDB database using a CRD.
- **Dev 2**: Learns that his colleague has instantiated a database but does not know what kind of database was created.

Currently, to find it, Dev 2 can sort the CRDs by categories, and let's say there is a "database" category, he will still have to go through each CRD to find the right one. 
A CR instances page would let him know exactly what databases are currently instantiated. 
It would also simplify management since he could directly see what resources have been instantiated.


### Scenario 2: One administrator who has the right to create resources and one developer who can only view resources
---
- **Admin**: Instantiates CRs for a project.
- **Dev**: Wants to know what has been instantiated for him. He must go through the CRDs one by one to see which ones he has access rights to.

In a CR Instance view, the developer would only see the CRs that he has the rights to see. If his project is partitioned to a specific namespace, he cannot sort by namespace to see which resources are in his namespace either. By clicking on "Namespaces" and then on his namespace, he only obtains the Resource Quotas, Limit Range, Pods, and Events.

The vision by CR would suddenly add the possibility of navigating more easily in the CRs, having direct visibility on the available resources, and being able to sort and filter the resources by namespace, facilitating navigation in compartmentalized projects.

I've tried to make a short video to illustrate the 2 scenarios. I hope it helps 😄 
Use_case1

joaquimrocha:
@Guilamb Okay. I think this makes sense. Let's mix this together with a higher level Custom Resources sidebar item as we talked about.

Guilamb:
> @Guilamb Okay. I think this makes sense. Let's mix this together with a higher level Custom Resources sidebar item as we talked about.

Should I close https://github.com/headlamp-k8s/headlamp/pull/2234 and transfer what have been done there https://github.com/headlamp-k8s/headlamp/pull/2234 into this PR ?
Or do you prefer to have 2 distinct PR ?

joaquimrocha:
> Should I close #2234 and transfer what have been done there #2234 into this PR ? Or do you prefer to have 2 distinct PR ?

Hi @Guilamb . Yes please. Transfer it here.
Today in out internal call we discussed this PR and would like to have this feature for the next release, hopefully by next Monday. Let me know if you are able to add it by then. We will help land this.

Guilamb:
> > Should I close #2234 and transfer what have been done there #2234 into this PR ? Or do you prefer to have 2 distinct PR ?
> 
> Hi @Guilamb . Yes please. Transfer it here. Today in out internal call we discussed this PR and would like to have this feature for the next release, hopefully by next Monday. Let me know if you are able to add it by then. We will help land this.

Hi, Should be ok, I've added the sidebar as required and I'm closing the other PR

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -240,6 +230,16 @@ function prepareRoutes(
         },
       ],
     },
+    {
+      name: 'crds',
+      label: t('glossary|Custom Resources'),
```
joaquimrocha, line 235: Please add icon: `'mdi:puzzle',` here so we get a nice icon for CRs. The icon is this one: https://icon-sets.iconify.design/mdi/puzzle/
The contents of the SVG should be copied into the icon cache at `frontend/src/components/App/icons.ts`, with the format similar to those already there. So this icon works fine offline.

frontend/src/components/Sidebar/prepareRoutes.ts
```diff
@@ -240,6 +230,16 @@ function prepareRoutes(
...
+      subList: [
+        {
+          name: 'crs',
+          label: t('glossary|CR Instances'),
```
joaquimrocha, line 239: Just call these `translation|Instances`. The context of being CR instances is already given by the parent item.

joaquimrocha (changes requested):
Thank you @Guilamb . I have reviewed the sidebar. I will review the rest later.

Guilamb:
> Thank you @Guilamb . I have reviewed the sidebar. I will review the rest later.

Thanks, I'll correct this. Also when showing the changes to @gberche, he spotted a problem with the filtering in the CR page.
I have correct it and pushed the changes. 
Also for the build-frontend test, do I need to do something ?

illume:
For "Build Frontend / test"... You might need to rebase against main, `npm i` to get the latest packages, and then regenerate your snapshots again.

The failing "Build Frontend / test plugins" which just hangs on CI is unrelated to this. Ignore this one for now, I am currently working on a fix.

joaquimrocha:
@Guilamb I think some snapshots are failing. Are you able to update them?

Guilamb:
> @Guilamb I think some snapshots are failing. Are you able to update them?

Hello, I couldn't finf out how to update them.

joaquimrocha:
Closing in favor of #2356 .

PR #2228, opened 2024-08-05
Fix workload charts infinite spinning

joaquimrocha:
This PR fixes an issue where it could lead to infinite spinning in the workload overview charts.
It also makes it more performant by not joining the workload items on every render, which could be a problem when having thousands of items.

This was reported by @lijianzhi01 .

**How to test:**
- [ ] Go to the Workloads area of Headlamp -> verify that the charts do not keep spinning forever

### DIFF
--- a/frontend/src/components/workload/Overview.tsx
+++ b/frontend/src/components/workload/Overview.tsx
@@ -104,9 +104,9 @@ export default function Overview() {
       <SectionBox py={2} mt={1}>
         <Grid container justifyContent="flex-start" alignItems="flex-start" spacing={2}>
           {workloads.map(workload => (
-            <Grid item lg={3} md={4} xs={6} key={workload.name}>
+            <Grid item lg={3} md={4} xs={6} key={workload.className}>
               <WorkloadCircleChart
-                workloadData={workloadsData[workload.name] || null}
+                workloadData={workloadsData[workload.className] || null}
                 // @todo: Use a plural from from the class itself when we have it
                 title={ChartLink(workload)}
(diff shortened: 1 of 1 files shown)


frontend/src/components/workload/Overview.tsx
```diff
@@ -70,7 +70,7 @@ export default function Overview() {
       joint = joint.concat(items);
     }
     return joint;
-  }
+  }, [workloadsData]);
```
illume, line 73: I don't think this is right... Dependencies only trigger an update if the reference to workloadsData changes. It does not do a deep comparison. From what I see the key/values are updated, and workloadsData keeps its reference. So the reference won't change - meaning this is a stale data bug. Suggest removing the useMemo change.

joaquimrocha: > I don't think this is right... Dependencies only trigger an update if the reference to workloadsData changes. It does not do a deep comparison. From what I see the key/values are updated, and workloadsData keeps its reference. So the reference won't change - meaning this is a stale data bug. Suggest removing the useMemo change.

Hmm, workloadsData is a state. We set it with the useState's setter. We never update only its key/values, from what I can see. What am I missing?

illume: Ok, I was mistaken. 

I see workloads is updated with a new reference in here:
```
  function setWorkloads(newWorkloads: WorkloadDict) {
    setWorkloadsData(workloads => ({
      ...workloads,
      ...newWorkloads,
    }));
  }
```

illume (approved):
👍 thanks!

PR #2231, opened 2024-08-06
frontend: Introduce react-query and v2 API

### LINKED ISSUE #1700: Centralize the calls for listing/getting resources from the k8s server
We have nice hooks for listing or getting resources (useList/useGet). However, if we have these hooks with the same options (e.g. `Pod.useList()` and `Pod.useGet(name, namespace)`) in N components, we are still making N calls to the API server when we should just have at maximum 1.

~There's an attempt at this at #415 but it is now obsolete and maybe investigating also https://tanstack.com/query/v3/ to see if it helps here is a good idea.~

sniok:
fixes #1700 

This PR introduces react-query as the library to perform and coordinate requests to the backend.
It brings nice quality of life improvements like caching, deduplicating requests, error handling, convenient react hooks.

For KubeObject classes, new methods were added: useQuery (alternative to useGet), useListQuery (alternative to useList). The old methods are left as-is (marked as deprecated) for compatibility with plugins. 

Some requests were not converted to limit the scope of this PR. Things like health and config requests for example, where caching might introduce problems, are left to do in next PRs. 

## Testing done

- [x] Manually checked the application
- [x] Ensured that the apiProxy exports stay the same after refactor (all the exported functions/types are present and unchanged)
- [x] Manually checked plugins/examples
- [x] Manually checked plugins repo plugins
- [x] Manually checked app compiled with `make app-linux`
- [x] Checked headlamp-plugin page reloaded when plugin changed.
- [x] Load tested with load-test scripts in Edge browser


illume:
Very nice work. I'm looking forward to these improvements.

I don't have time for an in depth review right now... But my general questions of all PRs: are there docs, tests for new code, atomic commits? Can anything be broken out into separate PRs? How does it affect plugins? Adopting a new technology... were alternatives evaluated and is the decision reversable/or leaking to users? ( I guess @reduxjs/toolkit/query is the alternative here ). Is there something broken/deprecated and how is that going to be communicated?

joaquimrocha (commented):
Definitely a great step in the direction of simplifying and optimizing what we have. Left some comments. I think that understanding where we place the API/lib versioning barrier is crucial for it to be more future proof.

frontend/src/lib/k8s/apiProxy/pluginsApi.ts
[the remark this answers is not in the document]
sniok: apiProxy also had a lot of stuff related to just any backend requests (functions like backendRequest, refreshToken, etc) which didn't really fit in k8s/ folder anyway. For compatibility it would make sense to keep it here as is, as it would brake imports

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -114,7 +114,7 @@ export default function ResourceTable<RowItem>(
 
 function TableFromResourceClass<RowItem>(props: ResourceTableFromResourceClassProps<RowItem>) {
   const { resourceClass, id, ...otherProps } = props;
-  const [items, error] = resourceClass.useList();
+  const { items, error } = resourceClass.useListQuery();
```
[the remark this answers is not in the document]
sniok: That would be a very nice improvement. Query also returns a `status` string which I think is even better than `isLoading`. But currently in the code we rely on null checks a lot and it would require a big refactor which I think should be done separately

frontend/src/components/crd/CustomResourceDefinition.stories.tsx
```diff
@@ -23,14 +21,12 @@ export default {
 } as Meta;
 
 const Template: StoryFn<MockerStory> = args => {
-  const { useGet, useList, useApiGet, useApiList, name, namespace, viewType = 'list' } = args;
+  const { useListQuery, useApiGet, name, namespace, viewType = 'list' } = args;
```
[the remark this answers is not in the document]
sniok: There are still some places, I think only for CRDs, where this kind of overriding is still in use

frontend/src/components/workload/Overview.tsx
```diff
@@ -24,21 +22,29 @@ interface WorkloadDict {
...
+  const { items: deployments } = Deployment.useListQuery();
+  const { items: statefulSets } = StatefulSet.useListQuery();
+  const { items: daemonSets } = DaemonSet.useListQuery();
+  const { items: replicaSets } = ReplicaSet.useListQuery();
+  const { items: jobs } = Job.useListQuery();
+  const { items: cronJobs } = CronJob.useListQuery();
 
-  React.useEffect(() => {
-    setWorkloadsData({});
-  }, [cluster]);
```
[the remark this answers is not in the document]
sniok: `cluster` value is part of the key for queries, so react-query will handle this for us

frontend/src/lib/k8s/cluster.ts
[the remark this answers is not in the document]
sniok: KubeObject refactor will be compatible, so only api stuff is changing

sniok:
@illume 

> are there docs

yes some functions are documented, where appropriate. One thing I'd want to do is to use the new `useListQuery` in the plugin examples but that can be done later

> tests for new code

unit tests would be nice for the v2/ folder, I'll make an issue for this

> atomic commits? Can anything be broken out into separate PRs? 

apiProxy refactor is in a separate commit but I don't think it's worth doing it in a separate PR


> How does it affect plugins? 

Plugins can use new `useQuery` `useQueryList` functions or keep using the old ones, they're unchanged.

> Adopting a new technology... were alternatives evaluated and is the decision reversable/or leaking to users? ( I guess @reduxjs/toolkit/query is the alternative here ).

I've had a really good experience with react-query before. There's also `swr` but they're pretty similar. 
I'd say that react-query is a safe choice 


> Is there something broken/deprecated and how is that going to be communicated?

All the current functions still work, we could mark them as deprecated later. Probably once we remove all the usages from our app first

frontend/src/lib/k8s/cluster.ts
```diff
@@ -567,6 +578,30 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
+    ) {
+      return useKubeObjectList({
+        kubeObjectClass: this,
+        cluster: options.cluster,
+        queryParams: options.queryParams as any,
+        namespace: options.namespace,
+      });
+    }
+
+    static useQuery<U extends KubeObjectClass>(
```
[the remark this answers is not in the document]
sniok: I think `get` implies that it'll only do HTTP GET request, but it also watches via websockets, I think `useQuery` is good enough. Tbh maybe even just `use` would be better. 

Something like 

```ts
const myPod = Pod.use({ name: "my-pod" })
```

frontend/src/components/workload/Overview.tsx
```diff
@@ -13,7 +11,7 @@ import Pod from '../../lib/k8s/pod';
 import ReplicaSet from '../../lib/k8s/replicaSet';
 import StatefulSet from '../../lib/k8s/statefulSet';
 import { getReadyReplicas, getTotalReplicas } from '../../lib/util';
-import Link from '../common/Link';
+import { Link } from '../common';
```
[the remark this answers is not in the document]
sniok: not intentional, reverting

frontend/src/lib/k8s/apiProxy/scaleApi.ts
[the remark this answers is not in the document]
sniok: moved

illume:
@sniok did you check the plugins still work? (examples, plugins repo)

frontend/src/lib/k8s/cluster.ts
```diff
@@ -732,13 +768,14 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
       const patchMethod = this._class().apiEndpoint.patch;
-      const args: Parameters<typeof patchMethod> = [body];
+      const args: Partial<Parameters<typeof patchMethod>> = [body];
 
       if (this.isNamespaced) {
         args.push(this.getNamespace());
       }
 
       args.push(this.getName());
+      // @ts-expect-error
```
illume, line 778: What's this about? `ts-expect-error` is usually for testing invalid values.

If it's supposed to be ignored, can you please add a comment why?

illume (commented):
Left a few notes.

The commit "frontend: Refactor apiProxy" needs a more descriptive commit message.

frontend/src/lib/k8s/cluster.ts
sniok: the way the arguments are passed into a variadic function makes it hard to make ts happy, this should just be ts-ignore, I'll change it

frontend/src/lib/k8s/apiProxy/index.ts
```diff
@@ -0,0 +1,67 @@
+/*
```
illume, line 1: I wonder about git history for apiProxy. Sometimes tracking doesn't work well if lots of the file has changed. If it's not working, doing a move in a separate commit can get around it.

frontend/src/lib/k8s/api/v2/utils.ts
```diff
@@ -0,0 +1,12 @@
+export function makeUrl(urlParts: any[], query: Record<string, string> = {}) {
```
illume, line 1: Is there a better place this can go than "utils"?
(which is a bit of an antipattern)

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,128 @@
...
+import { getCluster } from '../../../cluster';
+import { KubeObjectClass, KubeObjectInterface } from '../../cluster';
+import { BASE_HTTP_URL } from './fetch';
+import { KubeList, KubeListUpdateEvent } from './KubeList';
+import { KubeObjectEndpoint } from './KubeObjectEndpoint';
+import { makeUrl } from './utils';
+
+const BASE_WS_URL = BASE_HTTP_URL.replace('http', 'ws');
+
+export async function openWebSocket<T>(
```
illume, line 12: needs docs

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,128 @@
...
+    onMessage(data);
+  });
+  socket.addEventListener('error', error => {
+    console.error('WebSocket error:', error);
+  });
+
+  return socket;
+}
+
+export async function watchList<T extends KubeObjectInterface>({
```
illume, line 63: needs docs

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,128 @@
...
+  const socket = await openWebSocket<KubeListUpdateEvent<T>>(watchUrl, {
+    type: 'json',
+    cluster,
+    onMessage: update,
+  });
+
+  return socket;
+}
+
+export async function watchObject<T extends KubeObjectInterface>({
```
illume, line 97: needs docs

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -0,0 +1,192 @@
...
+      if (!it.ok) {
+        throw new Error('error');
+      }
+      return endpoint;
+    });
+  });
+  return Promise.any(promises);
+};
+
+const useEndpoints = (endpoints: KubeObjectEndpoint[]) => {
```
illume, line 97: needs docs

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -0,0 +1,192 @@
...
+    queryKey: ['endpoints', endpoints],
+    queryFn: () => getWorkingEndpoint(endpoints),
+  });
+
+  if (endpoints.length === 1) return endpoints[0];
+
+  return endpoint;
+};
+
+export function useKubeObjectList<T extends KubeObjectClass>({
```
illume, line 109: needs docs

frontend/src/components/common/Resource/ResourceTable.tsx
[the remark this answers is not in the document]
joaquimrocha: Status is fine too. Now that we return an object, it's fine to have these added later.

frontend/src/lib/k8s/api/v2/hooks.ts
illume, line 97: This one returns a useQuery type. Can you please consider abstracting it? Perhaps creating a new type name from the useQuery one if that's the only option.

(same for any other ones in this file I may have missed)

illume (commented):
... a few more notes.

frontend/src/lib/k8s/cluster.ts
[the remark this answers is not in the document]
joaquimrocha: Definitely let's not use `use`. I think you may get a point, but this get is about the object, so as we kept List in `useListQuery` and it lists + watches, I'd keep the Get. In fact, I think the watch is a param that the caller can turn off if they want, right?

frontend/src/components/common/Resource/ResourceTable.tsx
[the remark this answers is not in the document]
joaquimrocha: BTW, if we are returning anything that are implementation details from react-query? (Maybe that `status`) I think we should probably not do that ATM and just stick with some basics till we figure out what makes sense to return, unless you tell me we are limiting a lot the possibilities because of that.

frontend/src/lib/k8s/api/v2/hooks.ts
joaquimrocha: I agree here. Unless we are limiting too much what can be done, I'd keep the exposure to react-query to a minimum so we don't have to compromise on an implementation detail.

illume:
fyi: I'm testing with plugins and in the compiled app.

sniok:
Abstracted hook responses so it doesn't return react-query types. Added documentation to some functions

frontend/package.json
```diff
@@ -21,9 +21,12 @@
     "@mui/lab": "^5.0.0-alpha.152",
     "@mui/material": "^5.15.14",
     "@mui/system": "^5.15.14",
+    "@mui/x-date-pickers": "^7.15.0",
```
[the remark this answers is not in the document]
sniok: it was actually required for the material-react-table, but it didn't cause any troubles until now. Added this into a separate commit with an explanation

frontend/package.json
```diff
@@ -109,6 +112,11 @@
     "vitest": "^2.0.5",
     "cheerio": "1.0.0-rc.12"
   },
+  "optionalDependencies": {
```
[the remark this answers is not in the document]
sniok: done

sniok:
also added isFetching as it is slightly different from isLoading and will be useful

frontend/src/lib/k8s/api/v2/utils.ts
sniok: other stuff in the file wasn't used and only makeUrl left, renamed to makeUrl

frontend/src/components/common/Resource/ResourceTable.tsx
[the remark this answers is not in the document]
sniok: added an interface, so we don't have to rely on react-query

frontend/src/lib/k8s/cluster.ts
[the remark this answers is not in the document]
sniok: renamed to useGetQuery. There's no parameter to disable watching yet but can be added later

frontend/src/lib/k8s/apiProxy/index.ts
sniok: moved apiProxy file to apiProxy/index.ts in a separate commit and history seems to be working now

frontend/src/lib/k8s/api/v2/hooks.ts
sniok: done

illume:
This error keeps popping up bottom right of the screen. Every few minutes maybe.
image

illume:
Going to CRDs page:
image

illume:
It's fetching the /pods URL and getting a 404 on this home page:
image

sniok:
> Going to CRDs page:
> compiled app traceback:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'pluralName')
>     at xe (index-B1wInVTN.js:1128:19944)
>     at index-B1wInVTN.js:1128:20388
>     at Array.map (<anonymous>)
>     at Overview (index-B1wInVTN.js:1128:20235)
>     at um (vendor-mui-BWZhF6T-.js:38:17018)
>     at nx (vendor-mui-BWZhF6T-.js:40:44058)
>     at Jb (vendor-mui-BWZhF6T-.js:40:39790)
>     at YR (vendor-mui-BWZhF6T-.js:40:39718)
>     at Hc (vendor-mui-BWZhF6T-.js:40:39570)
>     at Zp (vendor-mui-BWZhF6T-.js:40:35934)
> jp @ vendor-mui-BWZhF6T-.js:40
> ```

I've seen that before but it was lost in rebase I suppose, I'll fix it

sniok:
> It's fetching the /pods URL and getting a 404 on this home page

Isn't that pod-counter plugin?

sniok:
Tested Kompose, opencost, plugin-catalog, app-plugin, prometheus plugins. A bug with prometheus was fixed (#2290) and was caused by react 18. Found unrelated bug in openscost plugin but other than that they work fine.

illume (commented):
> Tested Kompose, opencost, plugin-catalog, app-plugin, prometheus plugins.

Nice one.

I left a few notes. But could you please move the first 3 commits to separate PRs? (Note the comment about mui/x-date-pickers in headlamp-plugin/package.json is in the wrong commit first).

I'm still seeing history isn't there for some bits of code moved. It would be better to break down the larger "Refactor" commit into smaller pieces anyway.

illume:
Re: [opencost plugin issue](https://github.com/headlamp-k8s/plugins/issues/83) I think `const config = store.get();` can be used instead of the hook. Santhosh is going to look into it.

frontend/src/lib/k8s/event.ts
```diff
@@ -164,6 +164,46 @@ class Event extends makeKubeObject<KubeEvent>('Event') {
     return objInstance;
   }
 
+  static useListQueryForClusters(
```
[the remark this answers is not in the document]
sniok: It means that this hook will fail if the number of clusters change between renders.

To avoid that, the parent component should set the `key` for the component that uses this hook.
Using the `key` will remount the component when clusters change and hook wont fail.

You can see this in the `src/component/App/Home/index.tsx:170`

```tsx
  return <HomeComponent clusters={clusters} key={Object.keys(clusters).join('')} />;
```

HomeComponent uses `useWarningListQuery` hook and when clusters change, the key will also change, meaning the previous state of the HomeComponent will be discarded and it will be created from scratch.

frontend/src/components/cluster/Overview.tsx
```diff
@@ -138,7 +140,7 @@ function EventsSection() {
       }}
       defaultGlobalFilter={eventsFilter ?? undefined}
       data={events}
-      errorMessage={Event.getErrorMessage(eventsError)}
+      errorMessage={Event.getErrorMessage(eventsError as any)}
```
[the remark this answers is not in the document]
sniok: yes casting is not needed here anymore, good catch, thanks

frontend/src/lib/k8s/cluster.ts
```diff
@@ -732,13 +768,14 @@ export function makeKubeObject<T extends KubeObjectInterface | KubeEvent>(
...
       const patchMethod = this._class().apiEndpoint.patch;
-      const args: Parameters<typeof patchMethod> = [body];
+      const args: Partial<Parameters<typeof patchMethod>> = [body];
 
       if (this.isNamespaced) {
         args.push(this.getNamespace());
       }
 
       args.push(this.getName());
+      // @ts-ignore Dynamic arguments spread
```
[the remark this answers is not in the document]
sniok: we're spreading args array into function arguments, and ts is not happy about spreading `string[]` as arguments

frontend/src/lib/k8s/event.ts
```diff
@@ -164,6 +164,50 @@ class Event extends makeKubeObject<KubeEvent>('Event') {
...
+   * Important! Make sure to have the parent component have clusters as a key
+   * so that component remounts when clusters change, instead of rerendering
+   * with different number of clusters
+   */
+  static useListQueryForClusters(
+    clusterNames: string[],
+    options: { queryParams?: QueryParameters } = {}
+  ) {
+    // Calling hooks in a loop is usually forbidden
+    // But if we make sure that clusters don't change between renders it's fine
```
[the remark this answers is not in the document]
sniok: `key` property of the parent component will remount instead of rerendering. responded in more detail in the comment above

sniok:
First 4 commits are now in the following PRs: #2309 #2310 #2311 #2312

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,159 @@
...
+    type: 'json',
+    cluster,
+    onMessage: update,
+  });
+
+  return socket;
+}
+
+/**
+ * Listen for changes to a particular Kubernetes resource object
```
illume, line 126: Can you please add `@param` and `@returns` here?

illume (commented):
Can you please add `@param` and `@returns` to the newly documented functions?

frontend/src/lib/k8s/api/v2/watching.ts
sniok: done

frontend/src/lib/k8s/api/v2/fetch.ts
```diff
@@ -0,0 +1,84 @@
...
+import { makeUrl } from './makeUrl';
+
+export const BASE_HTTP_URL = helpers.getAppUrl();
+
+/**
+ * Simple wrapper around Fetch function
+ * Sends a request to the backend
+ *
+ * @param url - URL path
+ * @param init - same as second parameter of the Fetch function
```
skoeva, line 15:
```suggestion
 * @param init - options parameter for the Fetch function
```
I think this could be a little more descriptive, just a nit though. Feel free to ignore

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -0,0 +1,282 @@
...
+    isSuccess: query.isSuccess,
+    status: query.status,
+  };
+}
+
+/**
+ * Test different endpoints to see which one is working.
+ *
+ * @params endpoints - List of possible endpoints
+ * @returns Returns endpoint that works
```
skoeva, line 145:
```suggestion
 * @returns Endpoint that works
```

skoeva (approved):
Test coverage looks good, docs are concise and readable, and big thanks for splitting the commits across the different PRs. Monumental task, great work!

illume (approved):
🎉 thanks

(please fix those issues @skoeva mentioned, then we can merge)

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,178 @@
...
+import { KubeObjectEndpoint } from './KubeObjectEndpoint';
+import { makeUrl } from './makeUrl';
+
+const BASE_WS_URL = BASE_HTTP_URL.replace('http', 'ws');
+
+/**
+ * Create new WebSocket connection to the backend
+ *
+ * @param url - WebSocket URL
+ * @param options - Connection options
```
illume, line 16: These options need docs

frontend/src/lib/k8s/api/v2/watching.ts
```diff
@@ -0,0 +1,178 @@
...
+    console.error('WebSocket error:', error);
+  });
+
+  return socket;
+}
+
+/**
+ * Listen for changes in a List of Kubernetes resources
+ *
+ * @param options - watch options
```
illume, line 86: The options need documenting.

illume:
I went through and checked the remaining open convos were addressed and resolved them. There's only the two notes left.

illume:
> marked as deprecated

What is the plan for deprecation and migration?  How long are we keeping these around?

I think users should have a "`@see newfunction` and an example of how to change to the new function. So when they hover over their striked out deprecated function they can see how to change it very easily.

It would make sense to update the examples/ plugins in this PR so we can more easily test them.

---

Looking at the uses of useListQuery... why does the function signature have to be different?


Old use of useList:
```ts
  const [pods, podsFetchError] = Pod.useList({
    namespace,
    labelSelector: getPodsSelectorFilter(service),
  });
```

New use of useListQuery.
```ts
  const { items: pods, error: podsFetchError } = Pod.useListQuery({
    namespace,
    queryParams: {
      labelSelector: getPodsSelectorFilter(service),
    },
  });
```

- I don't see benefit for this function signature change as it's used. Theoretically if we want to extend the returned values it's a bit nicer. But in terms of how it's used... it doesn't seem to be a gain. So can you please explain that?
- Can useList be implemented in terms of useListQuery? Allowing us to drop that old code?

sniok:
> What is the plan for deprecation and migration? How long are we keeping these around?

The old functions are probably used a lot, first we can mark them deprecated in the comment, then probably console warning, then removal. But I think we'll keep them for quite a while

> Looking at the uses of useListQuery... why does the function signature have to be different?
> 
> Old use of useList:
> 
> ```ts
>   const [pods, podsFetchError] = Pod.useList({
>     namespace,
>     labelSelector: getPodsSelectorFilter(service),
>   });
> ```
> 
> New use of useListQuery.
> 
> ```ts
>   const { items: pods, error: podsFetchError } = Pod.useListQuery({
>     namespace,
>     queryParams: {
>       labelSelector: getPodsSelectorFilter(service),
>     },
>   });
> ```
> 
> * I don't see benefit for this function signature change as it's used. Theoretically if we want to extend the returned values it's a bit nicer. But in terms of how it's used... it doesn't seem to be a gain. So can you please explain that?

There are already more returned values than `items` and `error`. Things like `status` or `refetch` or `data` that contains `KubeList` (if you want to keep track of resourceVersion of the list). Returning tuple makes sense if we're sure that nothing else will be added but it's not the case here. 
The queryParams parameter is more explicit now, which I think is easier to understand

> * Can useList be implemented in terms of useListQuery? Allowing us to drop that old code?

I considered it, but the behavior of the functions is slightly different, `useList` will always send a request while useListQuery will sometimes use cached data. Together with the function signature change I think introducing a new function is justified.

sniok:
Pushed the suggested changes and created an issue for updating the plugin examples #2322

illume:
> > I don't see benefit for this function signature change as it's used. Theoretically if we want to extend the returned values it's a bit nicer. But in terms of how it's used... it doesn't seem to be a gain. So can you please explain that?

> There are already more returned values than items and error. Things like status or refetch or data that contains KubeList (if you want to keep track of resourceVersion of the list). Returning tuple makes sense if we're sure that nothing else will be added but it's not the case here.


At the moment I can't see any of the extra pieces of data used. The most common usage is like this:
```ts
  const [pods, podsFetchError] = Pod.useList({
    namespace,
    labelSelector: getPodsSelectorFilter(service),
  });
```

Which is simpler to use than this:
```ts
  const { items: pods, error: podsFetchError } = Pod.useListQuery({
    namespace,
    queryParams: {
      labelSelector: getPodsSelectorFilter(service),
    },
  });
```

> The queryParams parameter is more explicit now, which I think is easier to understand

The queryParams is redundant here, because "Selector" is already signifying that it's a selector.

In the future if users of these functions do use the extra information returned... at that point a different input could allow returning that extra info in an object. Which would keep compatibility with existing call sites.


> > Can useList be implemented in terms of useListQuery? Allowing us to drop that old code?

> I considered it, but the behavior of the functions is slightly different, useList will always send a request while useListQuery will sometimes use cached data. Together with the function signature change I think introducing a new function is justified.

The major issue with this is that we have to keep maintaining two versions of code. For maintenance it would be better to drop that old code now.

The problem with having a cached and non-cached view of the data is that it's inconsistent. This is why I think it's better to have them both use the cached version. I hope that the watches update that cached data more quickly than a new request would make (or at least just as quickly).

illume:
What is the idea with adding /api/ inside lib/k8s?

```
frontend/src/lib/k8s/apiProxy/portForward.ts → frontend/src/lib/k8s/api/v1/portForward.ts
```

Could you also please explain the idea and the plan with having a v1 and v2 folder?

sniok:
> At the moment I can't see any of the extra pieces of data used. The most common usage is like this:
> 
> ```ts
>   const [pods, podsFetchError] = Pod.useList({
>     namespace,
>     labelSelector: getPodsSelectorFilter(service),
>   });
> ```
> 
> Which is simpler to use 

Yes, right now in most places we're just using result and error. 
But now we use react-query and there's more information that we can (and should) use.

For example, now that we have caching, a common feature that improves UX is some kind of loading indicator that appears when data is refetched (while cached data is displayed). 

There's no way to represent this with just `[data,error]` tuple. Currently we use `null` to represent that data is loading but that is not a good practice (while appearing simple, sometimes `null` might be a valid response from backend). A much better solution is having a `status` variable. 

If you look at the current type definition of useList you'll see 

```
[any[], ApiError | null, (items: any[]) => void, (err: ApiError | null) => void];
```

which is not always obvious, for example guessing from just the type what the last item does is not clear.
Having a well documented object with clearly named properties is much easier to understand and to work with.

> The queryParams is redundant here, because "Selector" is already signifying that it's a selector.

Query parameters is a set of properties that is separate from the other parameters. Just like it is separated in Kubernetes documentation, [PodList example](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#podlist-v1-core). People that are already familiar with Kubernetes API would probably find it easier to work with.

> In the future if users of these functions do use the extra information returned... at that point a different input could allow returning that extra info in an object. Which would keep compatibility with existing call sites.

Functions which type definition changes based on parameters passed are usually hard to work with, I'd prefer avoiding doing that.

> The major issue with this is that we have to keep maintaining two versions of code. For maintenance it would be better to drop that old code now.
> 
> The problem with having a cached and non-cached view of the data is that it's inconsistent. This is why I think it's better to have them both use the cached version. I hope that the watches update that cached data more quickly than a new request would make (or at least just as quickly).

I could replace old functions completely, preserving function definitions. That would require some additional effort

sniok:
> What is the idea with adding /api/ inside lib/k8s?

that's where apiProxy was, then it was refactored `apiProxy` -> `api/{factories,requests,etc..}/`

> ```
> frontend/src/lib/k8s/apiProxy/portForward.ts → frontend/src/lib/k8s/api/v1/portForward.ts
> ```
> 
> Could you also please explain the idea and the plan with having a v1 and v2 folder?

v1 is what we currently have, v2 is new version of doing requests to the backend and react-query stuff
after a while I'd expect to move all functionality from v1 to v2 and then remove v1

sniok:
Pushed a big change, instead of new function, they're now replacing old useGet and useList. But with some iterator trickery it is now possible to do both `const [items,error] = useList` and `const { data, items, status, ..etc } = useList` at the same time

illume:
hi.

I’ll leave this to @joaquimrocha to give his sign off and merge, because he was reviewing it before.

Looks good to me. The only slight concerns I have left are about the “api” name, and introducing v1/v2, but will leave that up to you both to consider/decide.

thanks!!

frontend/src/components/cronjob/Details.tsx
```diff
@@ -137,7 +137,7 @@ export default function CronJobDetails() {
   const { t, i18n } = useTranslation('glossary');
 
   const [jobs, jobsError] = Job.useList({ namespace });
-  const [cronJob, setCronJob] = useState<CronJob | null>(null);
+  const [cronJob] = CronJob.useGet(name, namespace);
```
joaquimrocha, line 140: Supposedly this was done that way because the DetailsGrid already fetches the CronJob. I know that with react-query, we are not re-fetching things multiple times, but I wonder if this was needed in this commit.

frontend/src/components/App/Home/index.tsx
```diff
@@ -183,7 +183,7 @@ function HomeComponent(props: HomeComponentProps) {
   const { t } = useTranslation(['translation', 'glossary']);
   const [versions, errors] = useClustersVersion(Object.values(clusters));
   const maxWarnings = 50;
-  const warningsMap = Event.useWarningList(Object.values(customNameClusters).map(c => c.name));
+  const warningsMap = Event.useWarningListQuery(Object.values(customNameClusters).map(c => c.name));
```
joaquimrocha, line 186: In this case we are adding the new query suffix. Why don't we keep the original name, as we did for the other functions?

joaquimrocha (commented):
Left a comment. I think this PR would be easier to review if we had the move into v1 as a previous, separate commit.
About the versioned folders. I am totally in favor of having them so we can eventually move to a new version that has little to do with the previous one. And I am also happy to drop the previous version of apiProxy that was doing things like sequentially trying different k8s APIs without remembering any. However, in this case, unless I misunderstood it, it looks like we keep using most of the v1 (apply function, scale API, etc.) but we have the v2 being used for replacing the internals of the useGet/useList (and others), but it reads more like we are extending the previous API rather than deprecating it.

So I want to understand better the justification for having the 2 versions. Is v1 going to be deprecated? When do we get a replacement API for things like scale, apply, ...?

sniok:
>  it looks like we keep using most of the v1 (apply function, scale API, etc.) but we have the v2 being used for replacing the internals of the useGet/useList (and others), but it reads more like we are extending the previous API rather than deprecating it.

Here's the state of v1 and v2. The plan is to have everything in v2 eventually. Getting/watching list/objects is the most impactful for UX and it is now using react-query.

| | v1 | v2 |
| --- | --- | --- |
| Requests to backend/cluster | ✅ | ✅ |
| Get object/list | ✅[0] | ✅[1] |
| Watch object/list | ✅[0] | ✅[2] |
| Apply | ✅ | ❌ |
| Drain node |  ✅ | ❌ |
| Metrics | ✅ | ❌ |
| Plugin | ✅ | ❌ |
| Port Forward | ✅ | ❌ |
| Scale |  ✅ | ❌ |
| Token |  ✅ | ❌ |

[0] - This is still present in the code and available via apiProxy/index.ts file for compatibility, but it's not used for requests in our codebase anymore
[1] - These requests were greatly simplified by refactoring away apiFactories and using named arguments (objects) instead of positional arguments
[1][2] - react-query stuff

Now those API that are not yet in v2 are quite simple and can be included in v2 with minimal changes.

frontend/src/components/cronjob/Details.tsx
sniok: yeah it's not required, reverted

frontend/src/components/App/Home/index.tsx
sniok: good point, I overlooked this. it's now using the original name

sniok:
Pushed an update that moves v1 into a new commit, updates websocket logic

joaquimrocha (approved):

joaquimrocha:
@sniok Okay. I still think it's a bit weird we are branching the version when we are not marking anything as deprecated and we don't have 100% coverage for what we had in v1, but I can see a way where we add those along the way, so I have merged the changes.

PR #2234, opened 2024-08-06
Add custom resources navbar top lvl entry

Guilamb:
Add custom resources on top level of the navbar as discussed in https://github.com/headlamp-k8s/headlamp/pull/2226

image


joaquimrocha:
We should do this together with the full moving of CR/Ds to the top level, not just the sidebar.

PR #2235, opened 2024-08-06
Dockerfile: Update + pin alpine version (to 3.20.2)

joaquimrocha:


illume:
I wonder if we should use the "index digest" in the future (I think this is the sha256 for the am64 version of the image)?
(we are using amd64 architecture specific ones for other images, so probably this PR we should merge)

https://hub.docker.com/layers/library/alpine/3.20.2/images/sha256-eddacbc7e24bf8799a4ed3cdcfa50d4b88a323695ad80f317b6629883b2c2a78?context=explore

You can pull the image and get the "index digest" like this too.
```
docker inspect --format='{{index .RepoDigests 0}}' alpine:3.20.2
```

The eddacbc7e24bf8799a4ed3cdcfa50d4b88a323695ad80f317b6629883b2c2a78 is for amd64, and is the sha256 for that architecture.

Interestingly when you pull with the version and the sha256 in there it ignores the version and gets the sha256. It does no checking that the version matches the sha256.

illume (approved):
👍 thanks!

Issue #1716, opened 2024-02-14
Cannot connect to AWS EKS with SSO


bentatham:
I am having issues getting headlamp to connect to my AWS EKS cluster.

It works fine in OpenLens and kubectl. Not sure how headlamp connects, but perhaps it needs additional related AWS libraries loaded if using something in AWS SDK directly?

We are using AWS Identity Center (formerly AWS SSO) to authenticate to the cluster, so maybe that's the issue?  The UI just says Bad Gaterway and the logs have this. No idea which certificate it can't verify?

```
09:43:59.460 › server process stderr: 2024/02/14 09:43:59 http: proxy error: tls: failed to verify certificate: x509: certificate signed by unknown authority

09:44:08.935 › server process stderr: 2024/02/14 09:44:08 Error: failed to get context: key not found

09:44:12.280 › server process stderr: 2024/02/14 09:44:12 http: proxy error: tls: failed to verify certificate: x509: certificate signed by unknown authority

09:44:12.286 › server process stderr: 2024/02/14 09:44:12 http: proxy error: tls: failed to verify certificate: x509: certificate signed by unknown authority
```

elrondvega:
From my experience this seems to happen because the CA cert being used to connect to the API endpoint is the one being mounted in with the service account in the container at  /var/run/secrets/kubernetes.io/serviceaccount. If your cluster is signed by a different CA then the one used for your api endpoint then you'll see this. In my case the cluster CA was self-signed and the api endpoint was a letsencrypt signed.

joaquimrocha:
cc/ @yolossn .

kdeyko:
I have a similar issue with EKS (but without SSO): there are "Bad Gateway" messages in the UI.
Surprisingly, if I open the app via `open /Applications/Headlamp.app` in the terminal, it connects just fine 🤔 

Just in case: version 0.23.1 on macOS.

lindblombr:
Can confirm what @kdeyko says.  I have a similar issue with a custom OIDC binary we invoke from exec.  If we open the app through point-and-click (macOS), we just get bad gateway and lots of errors in the dev console connecting to localhost.  If I open up in a terminal via `/Applications/Headlamp.app/Contents/MacOS/Headlamp`, everything works perfectly.

lindblombr:
This looks like it can be related to #1885

illume:
I hope a fix for this is in the latest release.

Is anyone able to confirm if the latest release works for them or not?

[...]

PR #2246, opened 2024-08-16
app: Fix adding extra paths to PATH env var on Mac

joaquimrocha:
The command we were using to get the env vars was not yielding any different results than what we had for the PATH already, so this patch adds the /usr/local/bin path directly when it's missing, as fix-path does too: https://github.com/sindresorhus/fix-path/blob/main/index.js

Related: #1716

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -14,9 +14,7 @@ import {
 } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import find_process from 'find-process';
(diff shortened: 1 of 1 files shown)


joaquimrocha:
Here is the Mac build for this PR: https://github.com/headlamp-k8s/headlamp/actions/runs/10427218797
(You have to be logged in to Github to be able to download the DMGs at the bottom of that action run)

joaquimrocha:
Closing in favor of #2248 .

PR #2248, opened 2024-08-20
app: Use shell to run headlamp-server

illume:
It runs the headlamp-server via a shell. But also this takes care to try extra hard to kill headlamp-server.

On MacOS this means environment variables are passed into it, some of which may be needed for authentication.

### testing

These were tested:
- The headlamp-server is not running after the app closes.
- Run the server, and start the packaged app, then press Quit button, the app should close and the headlamp-server still be running
- tested with the compiled app, and development app
- check /tmp/envfiles.txt has the correct PATH and other env vars which match the terminals ones

Platforms tested:

- [x] linux WSL Ubuntu 22.04 with bash
- [x] linux WSL Ubuntu 22.04 with zsh
- [x] windows
- [x] mac with zsh running from Finder
- [x] mac with bash running from Finder

Related to: 
- https://github.com/headlamp-k8s/headlamp/issues/1716

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -14,7 +14,6 @@ import {
 } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import find_process from 'find-process';
-import { spawnSync } from 'node:child_process';
 import fs from 'node:fs';
 import { userInfo } from 'node:os';
 import { platform } from 'os';
@@ -56,51 +55,6 @@ const startUrl = (
   // we use startUrl to determine if it's an internal or external URL. So it's easier to
   // convert everything to forward slashes.
   .replace(/\\/g, '/');
-
-/**
- * On MacOS apps do not get the same environment variables as the terminal.
- *
- * However we want the same PATH as the shell to run the users terminal programs.
- */
-function addPathFromShellToEnvOnMac() {
-  if (process.platform !== 'darwin') {
-    // only necessary on MacOS
-    return;
-  }
-  if (process.env.TERM_PROGRAM) {
-    // if we are running in a terminal then we already have the correct PATH
-    return;
-  }
-
-  let defaultShell;
-  try {
-    defaultShell = userInfo().shell || '/bin/zsh';
-  } catch (error) {
-    defaultShell = '/bin/zsh';
-  }
-
-  // login interactive shell
-  // f option is to prevent menu on zshell when user has no config.
-  // DISABLE_AUTO_UPDATE is to prevent the shell from updating.
-  const env = { ...process.env, DISABLE_AUTO_UPDATE: 'true' };
-  const result = spawnSync(defaultShell, ['--login', '-fic', 'echo $PATH'], {
-    env: env,
-    encoding: 'utf-8',
-    timeout: 8000, // in case it's stuck
-  });
-
-  if (result.status === 0) {
-    const path = result.stdout.toString();
-    pathInfo = {
-      previousPath: process.env.PATH,
-      newPath: path,
-    };
-    process.env.PATH = path;
-  } else {
-    console.error('Failed to get shell PATH, just using process.env.PATH');
-  }
-}
-addPathFromShellToEnvOnMac();
 
 const args = yargs
   .command('$0 [kubeconfig]', '', yargs => {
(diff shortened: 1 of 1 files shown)


joaquimrocha:
I have tested this PR and there is a regression with these changes.
How to reproduce:
1. Run a headlamp-server process (`make run-backend`)
2. Build the app in unpacked mode: `cd app && npm run package -- --dir`
3. Run the app (`cd app; ./dist/linux-unpacked/headlamp`)
4. When prompted about what to do since a server is already running, choose `Quit` -> observe the log coming from an infinite loop and the app not getting closed.

I think this may be because of doing the closing logic on before-quit rather than quit but I am not sure. On main this doesn't happen.

joaquimrocha (changes requested):
Asking for changes given the regression found. Maybe if the signal is changed back, we should document that it needs to be that signal and how to test it.

backend/cmd/headlamp.go
```diff
@@ -306,6 +306,12 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	kubeConfigPath := config.kubeConfigPath
 
+	// Log all environment variables
```
illume, line 309: This is just for debugging, I'll remove before merge.

illume:
I added logging to headlamp-server so you can check the PATH and other env vars are there. It creates an `/tmp/envfiles.txt` with the environment variables in there.

Can you please run this and confirm the env vars are passed through now? Then I will remove the `/tmp/envfiles.txt` writing code from the PR.
(it works for me on my macos)

joaquimrocha:
@illume I have tested it again and the issue with quitting the app is gone.
The env vars declared in the terminal (`MY_ENV_VAR=test npm run dev`) do indeed pass over to the generated file. As expected, running from Finder will not have anything close to the env vars we have when running from terminal.

illume:
> As expected, running from Finder will not have anything close to the env vars we have when running from terminal.

Are you sure? On my system the PATH output to `/tmp/envfiles.txt` is the same as the PATH in my terminal. Additionally, the other env variables are there.

illume:
@joaquimrocha Now it is working for me with a MacOS user that is using zsh.

(Note, use .zshrc instead of .profile because zsh doesn't look at .profile)

illume:
fixed formatting

app/electron/main.ts
```diff
@@ -488,11 +464,23 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
   serverArgs = serverArgs.concat(flags);
   console.log('arguments passed to backend server', serverArgs);
 
+  const shell = getSpawnShell();
+
+  if (typeof shell === 'string') {
```
illume, line 469: Test zsh on Ubuntu.

app/electron/main.ts
```diff
@@ -488,11 +464,23 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
...
+    return spawn(shell, ['--login', '-ic', '-c', serverFilePath].concat(serverArgs), {
```
illume, line 470: "-ic" has redundant "c".

illume, line 470: Wrap in a try/catch and if it fails try to start normally.

backend/cmd/headlamp.go
```diff
@@ -302,10 +302,30 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 	}
 }
 
+// WriteEnvToFile writes all environment variables to a temporary file.
```
illume, line 305: This is just for testing, it will be removed after testing is complete.

illume:
@joaquimrocha changes we talked about where done.
I tested this again with MacOS with bash and zsh, and also Ubuntu with zsh.

Also added another fix for zsh showing an interactive shell if zsh is not installed properly (like if the user has no .zh files).

illume:
The `kill ESRCH` is because the process is already terminated. I should catch that error and prevent the error message being shown.

(I'm assuming the headlamp-server was not running after quit, and that this was on MacOS?)

joaquimrocha (approved):
WFM now. Please remove the env file creation code and merge.

yolossn:
I tested and this works for me but I was able to get it to work just by adding the proper env to the spawn options without having to use the shell to run the backend, IMHO using the shell when spawning is complicating the termination process.

Providing a git diff below for reference.
```
diff --git a/app/electron/main.ts b/app/electron/main.ts
index 0991c753..5b7ccf55 100644
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
   }
 }
 
+function getShellEnv(): NodeJS.ProcessEnv {
+  if (process.platform === 'win32') {
+    return process.env;
+  }
+
+  let shell = process.env.SHELL || '/bin/sh';
+  if (process.platform === 'darwin') {
+    try {
+      shell = userInfo().shell || '/bin/zsh';
+    } catch (error) {
+      shell = '/bin/zsh';
+    }
+  }
+
+  try {
+    const env = { ...process.env, DISABLE_AUTO_UPDATE: 'true' };
+    const result = execSync(`${shell} -ilc 'env'`, {
+      encoding: 'utf8',
+      timeout: 10000,
+      env,
+    });
+
+    const envVars = result.split('\n').reduce((acc, line) => {
+      const [key, value] = line.split('=', 2);
+      if (key && value) acc[key.trim()] = value.trim();
+      return acc;
+    }, {} as NodeJS.ProcessEnv);
+
+    console.log('Shell environment variables:', envVars);
+    return { ...process.env, ...envVars };
+  } catch (error) {
+    console.error('Failed to get shell environment:', error);
+    return process.env;
+  }
+}
+
 function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
   const serverFilePath = isDev
     ? path.resolve('../backend/headlamp-server')
@@ -488,11 +524,14 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
   serverArgs = serverArgs.concat(flags);
   console.log('arguments passed to backend server', serverArgs);
 
-  // We run detached but not in shell, otherwise it's hard to make sure the
-  // server process gets killed. When changing these options, please make sure
-  // to test quitting the app in the different platforms and making sure the
-  // server process has been correctly quit.
-  const options = { detached: true };
+  const shellEnv = getShellEnv();
+  const options = {
+    detached: true,
+    env: shellEnv,
+    shell: true,
+  };
+
+  console.log('Server process options:', options);
 
   return spawn(serverFilePath, serverArgs, options);
 }
```

illume:
@yolossn after running into trouble parsing environment variables I decided to go this way. There are a number of edge cases with environment variables. Additionally there are differences with different shells.

illume:
Removed `WriteEnvToFile("envfiles.txt")` which was just added for testing.

illume:
@yolossn Hey hey. Can you please see review notes I left in [app: Improve shell env capture for better cross-platform support by yolossn · Pull Request #2271 · headlamp-k8s/headlamp (github.com)](https://github.com/headlamp-k8s/headlamp/pull/2271)
 
I think it's right that it's better to not use a shell to launch. Let me know if you want to continue to fix these things in your PR, or I can also switch back to reading env vars in this PR?

Issue #1872, opened 2024-03-26
Create Flux plugin


joaquimrocha:
A Flux plugin would be very useful for all users of flux who would like a UI for it and who could also benefit from having it integrated in a generic K8s UI like Headlamp.

ashu8912:
The progress on this is tracked here https://github.com/headlamp-k8s/plugins/pull/75

The plugin should look for Flux installation on the Cluster and show different views related to Flux.

- [x] Application List View is Done
- [x] Application Detail View Lists basic detail about a Kustomization or a HelmRelease
- [x] Application Detail View has Sync With Source and Sync Without Source action buttons
- [x] Application Detail View has Suspend action button 
- [x] Application Detail View also lists The inventory and links them to the correct K8s resource view on Headlamp
- [x] Source List View is Done
- [x] Source Detail View shows basic detail About a GitRepository or  OCIRepository or Buckets or HelmRepos and HelmCharts
- [x] Source Detail View Allows suspending and resuming reconciliation
- [x] Notification List View is done
- [] Notification Detail View is done
- [x] Image Automation List View is done
- [x] Image Automation Detail View shows basic details for Image Repo or Image Policies or Image Update Automations
- [x] Flux Runtime View Lists all the flux related controllers and CRDs
- [ ] Flux Runtime shows the running version of flux

kingdonb:
I've got this running in a cluster now:

<img width="1399" alt="Screenshot 2024-08-22 at 10 02 02 AM" src="https://github.com/user-attachments/assets/9af7b24d-2387-4122-a787-3783a468ff83">

looking good! I need to add some flux artifacts, there may be one or more possibilities for how to detect the Flux Runtime version. I am seeing a label on each of my controllers, for example:

```
  labels:
    app.kubernetes.io/component: helm-controller
    app.kubernetes.io/instance: cozy-fluxcd
    app.kubernetes.io/managed-by: flux-operator
    app.kubernetes.io/part-of: flux
    app.kubernetes.io/version: v2.3.0
    control-plane: controller
    fluxcd.controlplane.io/name: flux
    fluxcd.controlplane.io/namespace: cozy-fluxcd
```

This is a managed flux, via flux-operator, so this may differ (?) from how it looks on a flux installed via `flux boostrap` or `flux install` - need to double check

kingdonb:
So I have a declarative (HelmRepository/HelmRelease) YAML to install Headlamp with the Flux plugin now, and it's working, but:

[...]

PR #2256, opened 2024-08-22
docs: fix bad link in plugin dev docs

kingdonb:
This is a broken reference in the docs.

(Can anyone let me know if this doc is outdated before I invest a bunch of time and find out it doesn't work?)

I won't hold this PR up by marking it as a draft, because the link is definitely broken, but if the doc is outdated and that's why the reference is bad, then maybe we shouldn't merge it. (I'll get around to building headlamp today with flux plugin that's in development, if the documents are good, but ...)

### DIFF
--- a/docs/development/plugins/building.md
+++ b/docs/development/plugins/building.md
@@ -154,7 +154,7 @@ For more on how to extract files into there see "Shipping and Deploying Plugins"
 ### More on making a headlamp container image including plugins
 
 See the blog post
-"[Get up to speed deploying Headlamp with plugins](https://headlamp.dev/blog/2022/10/20/get-up-to-speed-deploying-headlamp-with-plugins/)"
+"[Get up to speed deploying Headlamp with plugins](https://headlamp.dev/blog/2022/10/20/best-practices-for-deploying-headlamp-with-plugins)"
 for more information on building a container image with your plugins.
 
 ## Writing storybook stories


joaquimrocha (approved):

kingdonb:
FWIW, I am stuck, trying to build a plugin locally or in a dockerfile results in a failure.

Lots of ModuleNotFound errors:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'elkjs'
ModuleNotFoundError: Module not found: Error: Can't resolve 'react-i18next'
...
```

I am on a NodeJS 20 runtime locally, NodeJS 18 in the dockerfile (adapted from the example, that uses NodeJS 16) - I think the result is the same on either one.

I can try with a simpler plugin to start, but if someone is available to talk me through it - or if a more recent walkthrough of plugin building is already published somewhere, I'd be very appreciative of a link!

joaquimrocha:
@kingdonb Thanks. For testing a plugin, the fastest way is just to have it locally and test it with the desktop app. This involves having the plugin repo cloned, then running `npm i` and `npm start`, similar to creating a plugin:
https://headlamp.dev/docs/latest/development/plugins/building#creating-a-new-plugin

joaquimrocha:
> ```
> ModuleNotFoundError: Module not found: Error: Can't resolve 'elkjs'
> ModuleNotFoundError: Module not found: Error: Can't resolve 'react-i18next'
> ...
> ```

Sorry for the obvious question, but did you run `npm install`? cc/ @ashu8912 maybe you can follow up.

kingdonb:
I can't run it locally, I am setting up a collaborative environment for me, @stefanprodan, and @ashu8912 - it has to run in the cluster. No, I didn't run `npm install` that will probably help 😅 and that might have been in the very short list of things that Ashu mentioned, but I got lost somewhere between "you should run it locally" and the blog post about how to build it into a container.

I was going by the Dockerfile in the blog, we should update the blog so it works! :)

kingdonb:
following those steps, including npm start I was able to get the plugin building for production locally at least.

I'll bang on it a bit longer until I figure out what's missing from the blog, there are only a few places where "npm install" could go so maybe it will be an easy fix

joaquimrocha:
@kingdonb thanks. I've just opened https://github.com/headlamp-k8s/headlamp/pull/2257

kingdonb:
I landed on:

```
FROM docker.io/library/node:18@sha256:d0bbfdbad0bff8253e6159dcbee42141db4fc309365d5b8bcfce46ed71569078 as builder

WORKDIR /headlamp-plugins

COPY ./flux-plugin /headlamp-plugins/flux-plugin

RUN mkdir -p /headlamp-plugins/build

RUN cd /headlamp-plugins/flux-plugin && npm install

# Build the plugin
RUN npx @kinvolk/headlamp-plugin build /headlamp-plugins

# Extract the built plugin files to the build directory
RUN npx @kinvolk/headlamp-plugin extract /headlamp-plugins/ /headlamp-plugins/build


FROM alpine:latest

# Copy the built plugin files from the base image to /plugins directory
COPY --from=builder /headlamp-plugins/build/ /plugins/

CMD ["/bin/sh -c 'mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/'"]
```

I think there's a better formula because it seems like you will need to run npm install for each plugin this way, but my skepticism about if this would even work seems to have been overblown, I was expecting the resulting image to be hundreds of megabytes, and it's only 10.8MB 👍 

Will spend a bit more time testing this and see what I can come up with. Next step for now is to push the image somewhere and see if I can get this running in the cluster.

joaquimrocha:
FWIW, now that our `headlamp-plugin` command can install plugins from ArtifactHub, in the future we will improve how the images can be built when the plugins are published to ArtifactHub (not the case with this one, but still).

kingdonb:
Will continue the discussion in #1872

Issue #2261, opened 2024-08-23
cannot show ingresses in headlamp 0.25.0


intrand:
# Problem

When navigating to Network -> Ingresses, I'm greeted with an "Uh-oh! Something went wrong." Upon opening the console, I see more than a few of

```
vendor-mui-C3ciNw1R.js:39 TypeError: Cannot read properties of undefined (reading 'paths')
    at index-CCAjWvBk.js:849:42225
    at Array.forEach (<anonymous>)
    at Ai.getRules (index-CCAjWvBk.js:849:42185)
    at index-CCAjWvBk.js:1123:9724
    at Object.useMemo (vendor-mui-C3ciNw1R.js:37:58775)
    at He.useMemo (vendor-mui-C3ciNw1R.js:12:5905)
    at RulesDisplay (index-CCAjWvBk.js:1123:9701)
    at Wf (vendor-mui-C3ciNw1R.js:37:52830)
    at Rb (vendor-mui-C3ciNw1R.js:41:8734)
    at $b (vendor-mui-C3ciNw1R.js:41:950)
```

Which appears to me that `.spec.rules[].http` is coming back as null for some reason.

# Info

- Kubernetes 1.30.3
- Headlamp 0.25.0 / 06314eeb4ba7ea45d85441166fb9b696a0ac7a48
- A service account token bound to the ClusterRole `view`
- An `ingress` object which specifies `.spec.tls` for a wildcard subdomain
    ```yaml
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      name: wildcard
      namespace: ingressController
    spec:
      ingressClassName: whateverIngressClass
      rules:
        - host: '*.one.domain.tld'
        - host: '*.two.domain.tld'
      tls:
        - hosts:
            - '*.one.domain.tld'
            - '*.two.domain.tld'
          secretName: wildcard-cert
    ```

- `ingress` objects which don't specify `.spec.tls`
    ```yaml
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      name: headlamp
      namespace: headlamp
    spec:
      ingressClassName: whateverIngressClass
      rules:
      - host: headlamp.one.domain.tld
        http:
          paths:
          - backend:
              service:
                name: headlamp
                port:
                  number: 80
            path: /
            pathType: ImplementationSpecific
      - host: headlamp.two.domain.tld
        http:
          paths:
          - backend:
              service:
                name: headlamp
                port:
[...]

PR #2262, opened 2024-08-26
frontend: ingress: Handle missing HTTP paths

skoeva:
This change addresses a regression where Headlamp would crash when encountering an ingress that does not specify HTTP paths. This is particularly relevant for ingresses that only specify TLS settings for wildcard subdomains.

Fixes: #2261 

### Testing
- [X] Open Headlamp and apply an ingress which specifies `.spec.tls` for a wildcard subdomain, such as:
```yaml
kind: Ingress
metadata:
  name: wildcard
spec:
  rules:
    - host: '*.one.domain.tld'
    - host: '*.two.domain.tld'
  tls:
    - hosts:
        - '*.one.domain.tld'
        - '*.two.domain.tld'
      secretName: wildcard-cert
```

- [X] Navigate to the Ingresses page in the Network tab and ensure that the page does not crash

### DIFF
--- a/frontend/src/lib/k8s/ingress.ts
+++ b/frontend/src/lib/k8s/ingress.ts
@@ -92,31 +92,38 @@ class Ingress extends makeKubeObject<KubeIngress>('ingress') {
     const rules: IngressRule[] = [];
 
     this.spec!.rules?.forEach(({ http, host }) => {
-      const paths = http.paths.map(({ backend, path }) => {
-        if (!!(backend as LegacyIngressBackend).serviceName) {
-          return {
-            path,
-            backend: {
-              service: {
-                name: (backend as LegacyIngressBackend).serviceName,
-                port: {
-                  number: parseInt((backend as LegacyIngressBackend).servicePort, 10),
+      if (http) {
+        const paths = http.paths.map(({ backend, path }) => {
+          if (!!(backend as LegacyIngressBackend).serviceName) {
+            return {
+              path,
+              backend: {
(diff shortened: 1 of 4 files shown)


illume:
Could you please add this case to a story? 

Maybe this one? frontend/src/components/ingress/Details.stories.tsx

frontend/src/lib/k8s/ingress.ts
```diff
@@ -92,31 +92,35 @@ class Ingress extends makeKubeObject<KubeIngress>('ingress') {
...
-      const paths = http.paths.map(({ backend, path }) => {
-        if (!!(backend as LegacyIngressBackend).serviceName) {
-          return {
-            path,
-            backend: {
-              service: {
-                name: (backend as LegacyIngressBackend).serviceName,
-                port: {
-                  number: parseInt((backend as LegacyIngressBackend).servicePort, 10),
+      if (http) {
```
illume, line 95: I think it should still display the rule if it doesn't have http paths. This would currently just not display the whole rule.

skoeva: Set the rule to display with empty paths list, as well as the `console.debug` line to keep track. Also added the case without HTTP paths to a story, I believe the other stories should cover the other case

vyncent-t:
just reviewed this works on current branch

illume (approved):
Thanks 🎉 

Looks good to me.

(I added a note about removing the console.debug)

PR #2269, opened 2024-08-27
frontend: Fix refreshToken

yolossn:
the token expiry check in refreshToken was failing when the decodeToken returned null.this patch adds a check to see if the decodedToken exists before accessing the .exp property

Fixes: https://github.com/headlamp-k8s/headlamp/issues/2268

### DIFF
--- a/frontend/src/lib/k8s/apiProxy.ts
+++ b/frontend/src/lib/k8s/apiProxy.ts
@@ -208,7 +208,7 @@ async function refreshToken(token: string | null): Promise<void> {
   const decodedToken: any = decodeToken(token);
 
   // return if the token doesn't have an expiry time
-  if (!decodedToken.exp) {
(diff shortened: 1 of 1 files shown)


frontend/src/lib/k8s/apiProxy.ts
```diff
@@ -208,7 +208,7 @@ async function refreshToken(token: string | null): Promise<void> {
   const decodedToken: any = decodeToken(token);
 
   // return if the token doesn't have an expiry time
-  if (!decodedToken.exp) {
+  if (!decodedToken || !decodedToken.exp) {
```
sniok, line 211: minor, this could be written as 
```ts
if(!decodedToken?.exp)
```

sniok (approved):

skoeva (approved):
thanks for the fix

PR #2270, opened 2024-08-27
backend: Make sure context name is DNS compatible

yolossn:
this patch makes sure the context name is DNS
compatible.

### DIFF
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -261,6 +261,9 @@ func LoadContextsFromAPIConfig(config *api.Config, skipProxySetup bool) ([]Conte
 		// Note: nil authInfo is valid as authInfo can be provided by token.
 		authInfo := config.AuthInfos[context.AuthInfo]
 
+		// Make contextName DNS friendly.
+		contextName = makeDNSFriendly(contextName)
+
 		context := Context{
 			Name:        contextName,
 			KubeContext: context,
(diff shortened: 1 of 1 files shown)


knrt10 (commented):
apart from CI fail. looks good

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -408,3 +411,11 @@ func LoadAndStoreKubeConfigs(kubeConfigStore ContextStore, kubeConfigs string, s
 
 	return nil
 }
+
+// makeDNSFriendly converts a string to a DNS-friendly format.
+func makeDNSFriendly(name string) string {
+	name = strings.ReplaceAll(name, "/", "-")
```
illume, line 417: Probably reducing the likelihood of a clash could be done with some less common marker? 

For example "--" would be less likely to clash than "-"?

"front end" -> "front-end"
"front end" -> "front--end"

In this example "front--end" is less likely to exist than "front-end".

joaquimrocha: uI agree. Also, I think that having the same replacement for 2 different characters (`/` and ` `) is probably going to end up in problems to decode. Having spaces is unlikely to happen [AFAIU](https://stackoverflow.com/questions/74516648/eks-aks-cluster-name-convention), but I agree with keeping that as a defensive measure.
How about:
`my/cluster` -> `my--cluster`
`my cluster` -> `my__cluster` (double underscore)

I am assuming that double underscore and double dash are not used by any convention out there.
BTW, this all happens in memory, or are we storing the encoded cluster names somewhere? Because if we are storing them, the need for having something with minimal clash chances is of course very high.

joaquimrocha (approved):
LGTM, though the commit message has a period (very big nit, I know...), which is not common.

PR #2271, opened 2024-08-27
app: Improve shell env capture for better cross-platform support

yolossn:
This patch improves the approach for capturing and using the shell environment vars, particularly for non-Windows platforms.

 
### testing

- [x] The headlamp-server is not running after the app closes.
- [x] ~~Run the server, and start the packaged app, then press Quit button, the app should close and the headlamp-server will not be running~~ (Not applicable for this solution since the server is not killed based on PID).
- [x] tested with the compiled app (`make app-mac` etc), and development app (`cd app; npm start` etc)
- [x] check /tmp/envfiles.txt has the correct PATH and other env vars which match the terminals ones

Platforms tested:

- [x] linux WSL Ubuntu 22.04 with bash
- [x] linux WSL Ubuntu 22.04 with zsh
- [x] windows started from start bar
- [x] windows started from cmd
- [x] mac with zsh running from Finder
- [x] mac with bash running from Finder

Related to: 
- https://github.com/headlamp-k8s/headlamp/issues/1716

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1,4 +1,4 @@
-import { ChildProcessWithoutNullStreams, execSync, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams, exec, execSync, spawn } from 'child_process';
 import { randomBytes } from 'crypto';
 import dotenv from 'dotenv';
 import {
@@ -14,9 +14,10 @@ import {
 } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
 import find_process from 'find-process';
-import { spawnSync } from 'node:child_process';
+import * as fsPromises from 'fs/promises';
 import fs from 'node:fs';
 import { userInfo } from 'node:os';
+import { promisify } from 'node:util';
 import { platform } from 'os';
 import path from 'path';
 import url from 'url';
@@ -57,51 +58,6 @@ const startUrl = (
   // we use startUrl to determine if it's an internal or external URL. So it's easier to
   // convert everything to forward slashes.
   .replace(/\\/g, '/');
-
-/**
- * On MacOS apps do not get the same environment variables as the terminal.
- *
- * However we want the same PATH as the shell to run the users terminal programs.
- */
-function addPathFromShellToEnvOnMac() {
-  if (process.platform !== 'darwin') {
-    // only necessary on MacOS
-    return;
-  }
-  if (process.env.TERM_PROGRAM) {
-    // if we are running in a terminal then we already have the correct PATH
-    return;
-  }
-
-  let defaultShell;
-  try {
-    defaultShell = userInfo().shell || '/bin/zsh';
-  } catch (error) {
-    defaultShell = '/bin/zsh';
-  }
-
-  // login interactive shell
-  // f option is to prevent menu on zshell when user has no config.
-  // DISABLE_AUTO_UPDATE is to prevent the shell from updating.
-  const env = { ...process.env, DISABLE_AUTO_UPDATE: 'true' };
-  const result = spawnSync(defaultShell, ['--login', '-fic', 'echo $PATH'], {
-    env: env,
-    encoding: 'utf-8',
-    timeout: 8000, // in case it's stuck
-  });
-
-  if (result.status === 0) {
-    const path = result.stdout.toString();
-    pathInfo = {
-      previousPath: process.env.PATH,
-      newPath: path,
-    };
-    process.env.PATH = path;
-  } else {
-    console.error('Failed to get shell PATH, just using process.env.PATH');
-  }
-}
-addPathFromShellToEnvOnMac();
 
 const args = yargs(hideBin(process.argv))
   .options({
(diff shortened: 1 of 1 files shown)


app/electron/main.ts
```diff
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
...
+    try {
+      shell = userInfo().shell || '/bin/zsh';
+    } catch (error) {
+      shell = '/bin/zsh';
+    }
+  }
+
+  try {
+    const env = { ...process.env, DISABLE_AUTO_UPDATE: 'true' };
+    const result = execSync(`${shell} -ilc 'env'`, {
```
illume, line 472: Can you please expand the shortcuts to long form so they are more readable? `--interactive --login -c`

app/electron/main.ts
```diff
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
...
+      env,
+    });
+
+    const envVars = result.split('\n').reduce((acc, line) => {
+      const [key, value] = line.split('=', 2);
+      if (key && value) acc[key.trim()] = value.trim();
+      return acc;
+    }, {} as NodeJS.ProcessEnv);
+
+    console.log('Shell environment variables:', envVars);
```
illume, line 484: This should be removed before merge.

illume (changes requested):
Nice one.

I left notes about things I found out writing the other PR https://github.com/headlamp-k8s/headlamp/pull/2248

app/electron/main.ts
```diff
@@ -488,11 +524,12 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
...
-  // We run detached but not in shell, otherwise it's hard to make sure the
-  // server process gets killed. When changing these options, please make sure
-  // to test quitting the app in the different platforms and making sure the
-  // server process has been correctly quit.
-  const options = { detached: true };
+  const shellEnv = getShellEnv();
+  const options = {
+    detached: true,
+    env: shellEnv,
+    shell: true,
```
illume, line 531: Note, this causes issues with quiting the app. Please remove shell=true here.

app/electron/main.ts
illume, line 472: When running zsh, in the unlikely but possible case the user has no config... then zsh can put up an interactive mode telling the user which causes zsh to hang.
To disable this the `--no-rcs` can be used (on zsh only).

```
          .concat(shell === '/bin/zsh' ? [] : ['--no-rcs'])
```

app/electron/main.ts
```diff
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
   }
 }
 
+function getShellEnv(): NodeJS.ProcessEnv {
+  if (process.platform === 'win32') {
```
illume, line 457: btw, I think the only reason we should read env vars from a shell is if zsh is being used. For other shells they already have env vars set.

Since zsh can be used on linux and even windows I guess.

illume:
Please also remove `addPathFromShellToEnvOnMac`, since it is not needed anymore.

app/electron/main.ts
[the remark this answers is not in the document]
yolossn: this is not a blocker as long as the required env is present and headlamp backend binary is able to access the required binaries required for exec auth.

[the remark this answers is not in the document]
yolossn: since we are getting the env vars from the shell that is used by the user and using it in the spawn options, this wont be a blocker right.

app/electron/main.ts
illume, line 472: Please do not use sync function here. We need to not block the main thread.

app/electron/main.ts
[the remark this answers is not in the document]
illume: Right, for here the shell should not be used anyway. It's important to use the users shell though, not the default `shell: true` way of using the shell.

app/electron/main.ts
yolossn: even for bash the env setup can be present in `bashrc` right? Ill handle the case for windows also

app/electron/main.ts
yolossn: adding `--no-rcs` is breaking the solution, in my case i have the required configurations setup in the .zshrc file.

yolossn: I checked it by removing my .zshrc and running the command. zsh doesn't get into informing the user to setup the config.
image

yolossn: Even if we change getShellEnv to be asynchronous, the startServer function that calls it is still synchronous. This means the main thread will block until getShellEnv completes, regardless of whether it's sync or async.

illume: Not so.

An async function doesn't block if it's called in a sync function. Additionally, higher level functions can be changed.

app/electron/main.ts
```diff
@@ -453,6 +453,78 @@ class PluginManagerEventListeners {
...
+  }
+
+  // Fallback chain: user's shell > SHELL env > zsh > bash > sh
+  if (!shell || shell.trim() === '') {
+    shell = '/bin/zsh';
+  }
+  if (!fs.existsSync(shell)) {
+    // Fallback to bash if zsh is not found
+    console.error('Shell not found:', shell);
+    shell = '/bin/bash';
```
illume, line 478: A better default would be `/bin/sh`. Because that's a default others use, and is more likely to be there because bash is optional.

app/electron/main.ts
```diff
@@ -488,11 +560,18 @@ function startServer(flags: string[] = []): ChildProcessWithoutNullStreams {
...
-  // server process gets killed. When changing these options, please make sure
-  // to test quitting the app in the different platforms and making sure the
-  // server process has been correctly quit.
-  const options = { detached: true };
+  const extendedEnv = getShellEnv();
+  const options = {
+    detached: true,
+    env: {
+      ...extendedEnv,
+      HEADLAMP_CONFIG_ENABLE_HELM: 'true',
```
illume, line 568: Can you please explain why these environment variables are added?

They are supposed to be user supplied, so they shouldn't be set here.

yolossn: Since we ship the app store plugin in the app we already set it up [here](https://github.com/headlamp-k8s/headlamp/blob/c26c9137f21312a925d85a63f41808d573cb4a0f/app/electron/main.ts#L480). I was adding it for debugging the `envfiles.txt`. I will remove this.

app/electron/main.ts
```diff
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
   }
 }
 
+function getShellEnv(): NodeJS.ProcessEnv {
+  if (process.platform === 'win32') {
+    return process.env;
+  }
+
+  let shell = process.env.SHELL || '/bin/sh';
+  if (process.platform === 'darwin') {
```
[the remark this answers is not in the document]
yolossn: for all non windows environment I check for the shell in the following order `zsh>bash>sh`.

app/electron/main.ts
```diff
@@ -453,6 +453,42 @@ class PluginManagerEventListeners {
...
+
+  try {
+    const env = { ...process.env, DISABLE_AUTO_UPDATE: 'true' };
+    const result = execSync(`${shell} -ilc 'env'`, {
+      encoding: 'utf8',
+      timeout: 10000,
+      env,
+    });
+
+    const envVars = result.split('\n').reduce((acc, line) => {
```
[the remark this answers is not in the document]
yolossn: I changed the logic from `env` to `env -0` which outputs the environment variables separated by null characters (\0) instead of newlines.

app/electron/main.ts
yolossn: removed.

illume:
> Not applicable for this solution since the server is not killed based on PID

This still needs to be tested on each platform and package type. Because it's tricky to get right.

app/electron/main.ts
```diff
@@ -456,7 +412,80 @@ class PluginManagerEventListeners {
...
+
+  // For Windows, just return the current environment
+  if (isWindows) {
+    return { ...process.env };
+  }
+
+  // For Unix-like systems
+  const isZsh = shell.includes('zsh');
+  // interactive is supported only on zsh
+  const shellArgs = isZsh ? ['--login', '--interactive', '-c'] : ['--login', '-c'];
```
illume, line 461: Why not interactive here for other shells?

app/electron/main.ts
illume: Windows with bash or other shells can be handled in another future PR if it's an issue.

app/electron/main.ts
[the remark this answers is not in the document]
illume: great!

illume (approved):
Great work. Thank you!

(I'm going to do a final test of all the different platforms and package formats).

app/electron/main.ts
yolossn: image

yolossn:
> This still needs to be tested on each platform and package type. Because it's tricky to get right.

Agree, the server getting killed when the app is closed is already in the list so I removed it.

illume (changes requested):
Doesn't work on MacOS with user using bash.

The correct PATH isn't being picked up.

illume:
Also not picking up env from my zsh macos user. On zsh I put an environment variable in .zshrc and it's not in /tmp/envfiles.txt.

Can you confirm if environment variables put in .zshrc are showing up in your macos zsh user /tmp/envfiles.txt, when running the compiled app from Finder?

illume:
```
App starting...
Check for updates:  true
arguments passed to backend server [ '--proxy-urls', 'https://artifacthub.io/*' ]
Failed to get shell environment: Error: Command failed: /usr/local/bin/bash --login -c 'env -0'
env: illegal option -- 0
usage: env [-iv] [-P utilpath] [-S string] [-u name]
           [name=value ...] [utility [argument ...]]

    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at ChildProcess.exithandler (node:child_process:422:12)
    at ChildProcess.emit (node:events:519:28)
    at maybeClose (node:internal/child_process:1105:16)
    at Socket.<anonymous> (node:internal/child_process:457:11)
    at Socket.emit (node:events:519:28)
    at Pipe.<anonymous> (node:net:338:12) {
  code: 1,
  killed: false,
  signal: null,
  cmd: "/usr/local/bin/bash --login -c 'env -0'",
  stdout: '',
  stderr: 'env: illegal option -- 0\n' +
    'usage: env [-iv] [-P utilpath] [-S string] [-u name]\n' +
    '           [name=value ...] [utility [argument ...]]\n'
}
```

yolossn:
what is the bash version that you are on ? `env -0` is working for me in bash 3.2

image

illume:
env is a the macos env binary, not bash.

I think this works on MacOS...
```
/usr/bin/perl -le 'print "$_=$ENV{$_}\0" for keys %ENV'
```

yolossn:
> ```
> /usr/bin/perl -le 'print "$_=$ENV{$_}\0" for keys %ENV'
> ```
This will again print the env with \n separator and since \n is a valid env value I used `env -0` to print it null separated.

illume:
Ok... this perl on macOS works for me.
```
perl -e 'print "$_=$ENV{$_}\0" for keys %ENV'
```

illume:
I guess we might have some freebsd, or alpine users one day without gnu env. So maybe see if /usr/bin/perl is there, and fallback to env if perl isn't there?

illume:
I guess we could switch back to just env and parsing new lines? And fix this issue later. The newline issue is unlikely to affect folks much.

illume (approved):
👍 thanks

illume (approved):
👍

Issue #2095, opened 2024-06-27
Improve error message for unsupported kubeconfig encoding


omniproc:
Currently when a _kubeconfig_ file is loaded that has a unsupported encoding, e.g. Windows Powershell's "Out-File" default UTF-16 LE BOM, Headlamp will return the very vague error message _"Load a valid kubeconfig"_.
This makes it hard to understand what's wrong with the file given that the syntax of the file itself will be perfectly fine. 

Headlamp should either support different encodings or at very least report back that an encoding error occurred.

Steps to reproduce:

1. Take a working kubeconfig `my-working-kubeconfig.yaml` and load it in Headlamp making sure it is accepted as valid kubeconfig
2. Run the Powershell snippet `Get-Content my-working-kubeconfig.yaml -Raw | Out-File my-failing-kubeconfig.yaml`
3. Try to load `my-failing-kubeconfig.yaml`

omniproc:
P.S.: I noticed that other encoding formats do get loaded but then the connection to the cluster will fail. The headlamp dev console shows an "non whitespace character" error for the kubeconfig json passed to the headlamp server.

So there seem to be multiple issues with the encoding, not just the one I explicitly pointed out above with UTF-16- LE BOM. Might be worth re-designing the hole kubeconfig loading processes: if the file is loaded it should be possible to use it. If it's unable to be loaded a clear error message what's wrong with the file should be returned to the user.

PR #2273, opened 2024-08-27
frontend: KubeConfigLoader: Improve error message

skoeva:
Currently, loading a kubeconfig with unsupported encoding prompts a very vague error message. This change improves the error message to explicitly describe the issue.

Fixes: #2095

### Testing

- [X] Take a working kubeconfig (i.e. run `kubectl config view` and save it locally as `valid-kubeconfig.yaml`) and load it in Headlamp to make sure it is valid
- [X] Run the PowerShell snippet `Get-Content valid-kubeconfig.yaml -Raw | Out-File invalid-kubeconfig.yaml`
- [X] Try to load `invalid-kubebconfig.yaml` and ensure that a descriptive error message appears

image


frontend/src/components/cluster/KubeConfigLoader.tsx
```diff
@@ -181,7 +181,7 @@ function KubeConfigLoader() {
         }
         setFileContent(doc);
       } catch (err) {
-        setError(t('translation|Load a valid kubeconfig'));
+        setError(t('translation|Unsupported encoding: Please use UTF-8 encoding'));
```
joaquimrocha, line 184: There is the possibility that the error is related to encoding, but there are other possibilities, including the one we are directly throwing, on like 180 above this catch block.
I think we should do two things:
1. replace the `Invalid kubeconfig file` message from the error by something translatable. We can even throw two different messages, since we are checking for two different things there (existence of `clusters` and existence of `contexts`).
2. Use the `err.message` to set the error. This way, no matter what the error is, we know it will be more related to what actually happened.

joaquimrocha (changes requested):
Left a comment, as this fix should not be restricted to the possibility of an encoding error. Please rebase the branch when you update.

frontend/src/components/cluster/KubeConfigLoader.tsx
skoeva: Made the changes

frontend/src/components/cluster/KubeConfigLoader.tsx
```diff
@@ -176,12 +176,15 @@ function KubeConfigLoader() {
           ...new Uint8Array(reader.result as ArrayBuffer),
         ]);
         const doc = yaml.load(data) as kubeconfig;
-        if (!doc.clusters || !doc.contexts) {
-          throw new Error('Invalid kubeconfig file');
+        if (!doc.clusters) {
+          throw new Error('Invalid kubeconfig file: No clusters found');
```
joaquimrocha, line 180: We can make these strings translatable, is what I meant. `t('tanslation|No clusters found!')`. Then we can use this in the setError. Something like:
```ts
setError(t('translation|Invalid kubeconfig file: {{ errorMessage }}', {errorMessage: error.message}))
```

joaquimrocha (changes requested):
Thanks for the changes. I think we can still improve how the translation is working, please see my comments.

frontend/src/components/cluster/KubeConfigLoader.tsx
skoeva: Makes sense, pushed changes and included a screenshot of the new error message in the PR description

sniok:
I know this PR is about message content, but is it possible to change the style as well? White text on bright red background is very hard to read and not accessible

skoeva:
> I know this PR is about message content, but is it possible to change the style as well? White text on bright red background is very hard to read and not accessible

@sniok We are using this color scheme elsewhere in the project, maybe we can look into a11y in a different PR to cover those as well?

I ran this through a contrast checker for anyone else interested, this seems worth addressing from an a11y perspective:

image

sniok:
> We are using this color scheme elsewhere in the project, maybe we can look into a11y in a different PR to cover those as well?

Yeah going through all cases systematically sounds good. I've opened an issue #2280 to keep track 

> I ran this through a contrast checker for anyone else interested, this seems worth addressing from an a11y perspective:

While contrast ratio is important we should also not forget about other considerations for text a11y, like font size/thickness/color combinations. For example reading thin text on really bright background (even if the contrast ratio passes typical checks) with astigmatism can be hard

joaquimrocha (approved):

PR #2275, opened 2024-08-28
eslint-config: Add eslint-config and update app, frontend, and headlamp-plugin to use it

### LINKED ISSUE #2276: test plugins is hanging on CI

illume:
Updating dependencies, and a new @headlamp-k8s/eslint-config

- Fixes #2276
- new `@headlamp-k8s/eslint-config` from `@kinvolk/eslint-config`
- Fixes broken prettier with frontend/ (only one npm install needed now)

There is still more work to be done to update headlamp-plugin, but that will come soon in future PRs.
This has value now because it fixes the CI hang(and prettier issue).


sniok:
There's still an issue with react and react-hooks plugins. 
To reproduce create a file/paste this code anywhere in frontend/

```js
import { useEffect, useState } from "react";

export function Component() {
  const [test] = useState();

  useEffect(() => {
    console.log(test);
  }, []);

  return null;
}

```

react-hooks plugin should warn about missing dependency but it passes eslint check.
plugins are loaded correctly but the rules from those plugins are not used at all. 

adding this fixes that, but then there's a lot of warnings/errors to fix
```
extends:
- plugin:react/recommended
- plugin:react-hooks/recommended
```

sniok:
In the code prettier "trailingComma" is set to all but in published prettier config it's not defined

eslint-config/prettier-config/index.js
```
module.exports = {
  arrowParens: 'avoid',
  singleQuote: true,
  semi: true,
  printWidth: 100,
  trailingComma: "all",
};
```
```
cat node_modules/@headlamp-k8s/eslint-config/prettier-config/index.js
module.exports = {
  arrowParens: 'avoid',
  singleQuote: true,
  semi: true,
  printWidth: 100,
};
```

illume:
Thanks. Removed trailingComma.

sniok (approved):

PR #2277, opened 2024-08-28
frontend KubeConfigLoader: Do not override Next button's color

joaquimrocha:
The button was supposed to be black but the text was also being set as black. Since primary buttons are now blue from the theme, we shouldn't hardcode their color like that when it comes to normal buttons. This patch removes that custom styling from that "next" button, making it use the default primary color.

### DIFF
--- a/frontend/src/components/cluster/KubeConfigLoader.tsx
+++ b/frontend/src/components/cluster/KubeConfigLoader.tsx
@@ -102,15 +102,6 @@ const WideButton = styled(Button)({
   maxWidth: '300px',
 });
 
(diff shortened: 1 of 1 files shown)


skoeva (commented):
LGTM, I think you could throw the Finish button on this PR as well (lines 335-337)

joaquimrocha:
> LGTM, I think you could throw the Finish button on this PR as well (lines 335-337)

Good catch. Thanks!

skoeva (approved):

illume (approved):
👍

PR #2284, opened 2024-08-30
frontend: Fix dark theme error colors to be lighter

illume:
So they contrast more.


skoeva:
New color shows up here too:

image

The color looks good for error text on the dark background (i.e. from #2205), but we should also take care of managing the error colors across different error messages. I think the color used for the banner at the top can be a darker red, here's the contrast ratio:

image

illume:
I changed this to use the lighter red color. But maybe it needs more testing with other components.

Issue #2236, opened 2024-08-07
Not possible to add multiple cluster with same context


dark0ni:
Hello,

I found the following bug: when I want to add an additional cluster where the context has the namespace "default" like the first added cluster, it does nothing.  I can't also change the name of the first added cluster from "default" to another name.  

Here is an example:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: XXXXXXXXXX....
    server: https://XXXXXXXXX.k8s.ondigitalocean.com
  name: clustername-01
contexts:
- context:
    cluster: clustername-01
    namespace: custom-namespace-01-prod
    user: team-xxxxx
  name: default
current-context: default
kind: Config
preferences: {}
users:
- name: team-xxxxx
  user:
    client-key-data: XXXXXXXXX....
    token: XXXXXXXXXXX....

```

All my kubeconfig files look the same with different cluster names and providers. Also, the program requires the token to open the cluster, but there are already existing bug tickets for that :)  

Thank you in advance.

joaquimrocha:
Hi @dark0ni . I would like more details:
1. Are you running Headlamp desktop or web?
2. What are the steps you do to reproduce this issue? (it may seem obvious from your description, but I want to make sure we're thinking about the same when I read "adding a cluster")

Also, do you see more clusters/contexts if you add them to the kubeconfig that you know is being loaded?

dark0ni:
Sure

Also i cant remove the cluster anymore


1. I run the Desktop Version on Windows 11 Pro 23H2
2. Steps to reproduce:

1. Add the kubeconfig to headlamp
2. added as "default"
3. try to change name to another name with numeric and - for example "m6-dev"
4. Dont change the cluster name
5. Try to add an additional kubeconfig with default as name
6. No reaction from GUI
7. Try to remove current cluster in the list, also not possible

--> some json parsing error occure in the console

Here additional logs from web console:
```
index-BhWvm8rb.js:849 Unable to parse error json at url: http://localhost:4466/clusters/default/version {err: SyntaxError: Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)} with request data: {signal: AbortSignal, headers: {…}}
clusterRequest @ index-BhWvm8rb.js:849
```

![image](https://github.com/user-attachments/assets/f6c4f574-9918-48cc-a627-032d889b004d)

dark0ni:
when i try JSON.stringify(ke) i get a empty json '{}' as result
[...]

PR #2288, opened 2024-09-02
Add check for multiple clusters with the same context name

knrt10:
Earlier users were able to submit clusters with same name to the backend,
due to which it was getting overwritten on the proxy. Even if the
cluster was shown in frontend the backend did not knew about the
duplicate cluster.
    
`kubectl` throws the error of adding duplicate context name. We are
adding feature for the same.
    
If now user try to add a context which has the same name as earlier
context the app with throw an error and ask them to change the context
name.
    
Fixes: #2236


joaquimrocha (approved):
LGTM.

Issue #1628, opened 2023-12-20
Load Valid Contexts from Kubeconfig Despite Errors


joaquimrocha:
In the testing for the latest release I have noticed that Headlamp was not loading any of the clusters in Mac and Windows (somehow in WSL it worked, maybe because the kubeconfig is different or because I didn't test correctly).

So we ended up having to revert the related commit: https://github.com/headlamp-k8s/headlamp/pull/1627

We need to check what is wrong in that code. I will share my (sanitized) kubeconfig privately.

knrt10:
Adding more information on this, currently the Headlamp returns no context even if there is one error with kubeconfig. As discussed with the team it is concluded we should not renter Headlamp useless if there if only 1 issue in multi cluster setup.

We should show working clusters which are valid in Kubeconfig

knrt10:
Started work on this.

illume:
No one got to this yet, so moving to next release milestone.

PR #2289, opened 2024-09-03
backend: send working list of contexts

knrt10:
If there is a problem with the kubeconfig in any of the context or cluster, headlamp won't be rendered useless. It would throw human readable message to the backend and show the list of working clusters.

- Fixes: #1628
- The original issue was here: https://github.com/headlamp-k8s/headlamp/issues/1349

### Testing

- In your kubeconfig add some context with partial validity. For reference please check [here](https://github.com/headlamp-k8s/headlamp/blob/e13376a49d4191032926373e11d30a1a6c5ff4f7/backend/pkg/kubeconfig/test_data/kubeconfig_partialcontextvalid)
- Run backend `cd backend && go run ./cmd -dev`
- Run frontend `cd frontend && npm start`
- You should see errors in backend and only valid clusters in frontend.
- You can try with different kubeconfig and check.

### DIFF
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -7,10 +7,11 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
-	"path/filepath"
+	"os"
 	"runtime"
 	"strings"
 
+	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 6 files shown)


joaquimrocha:
CI is failing. And these changes need to be thoroughly tested.

knrt10:
Yes, fixing the CI error. Probably because of dynamic cluster. Tested it on various use cases. Would request others to test as well

illume:
@knrt10 could you please add instructions on how to test this manually to the PR description?

backend/pkg/portforward/handler_test.go
```diff
@@ -75,14 +74,16 @@ func TestStartPortForward(t *testing.T) {
...
+			break
 		}
 	}
 
 	require.NotEmpty(t, podName)
 	require.NotEmpty(t, targetPort)
 
 	// start portforward
-	req := &http.Request{}
+	req := httptest.NewRequest(http.MethodPost, "/start-portforward", nil)
```
illume, line 86: Looks like there's some port forward related changes in here too?

I guess it makes sense to move these portforward related things to a separate commit, and mention them in the PR description too?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -419,3 +752,205 @@ func makeDNSFriendly(name string) string {
...
+	switch fieldName {
+	case "server":
+		cluster.Server = strValue
+	case "certificate-authority":
+		cluster.CertificateAuthority = strValue
+	case "proxy-url":
+		cluster.ProxyURL = strValue
+	case "tls-server-name":
+		cluster.TLSServerName = strValue
+	}
```
illume, line 789: What happens to things not covered by these cases?

backend/pkg/portforward/handler_test.go
knrt10: Aaah, I added that path for testing, forgot to remove it. There is no change made to port forward functionality just some changes in err handling due to `LoadContextsFromFile`, so I think it can be in the same commit. If we move it to separate commit the file will fail due to err handling. Fixing it. Thanks for catching it

illume (approved):
🎉 Thanks!

yolossn:
@illume @knrt10 
This PR reverts to custom code for loading kubeconfig contexts, which can introduce issue related to unhandled cases and we will have to play catchup with new Kubernetes updates. A more robust solution would be to split the kubeconfig file into separate files, one per context. This approach would:
1. Isolate errors to individual contexts
2. Allow continued use of client-go functions for kubeconfig loading

I've created a proof-of-concept demonstrating this method. For reference, please see the attached [gist](https://gist.github.com/yolossn/204de669e12559bdea0e3b5edd1402f2).

knrt10:
This makes more sense, will try it out and implement in our current code base. The above implementation would be a better solution. Thanks

PR #2296, opened 2024-09-03
docs/contributing.md: Add section about the monthly call

joaquimrocha:
This is a good bookmark till we have the solution for the meeting link.


illume (approved):
👍 thanks

PR #2297, opened 2024-09-04
ci: fix commenting of test coverage

knrt10:
currently due to insufficient permission of write permission for gh cli,
it was unable to comment on the PR. This fixes that.

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -10,7 +10,7 @@ on:
       - main
       - rc-*
       - testing-rc-*
-    
+
 permissions:
   contents: read
 
@@ -21,7 +21,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
-      issues: write # needed for commenting on PRs for coverage changes
+      contents: read
+      pull-requests: write # needed for commenting on PRs for coverage changes
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/backend-test.yml
```diff
@@ -10,9 +10,10 @@ on:
       - main
       - rc-*
       - testing-rc-*
-    
+
 permissions:
   contents: read
+  pull-requests: write
```
illume, line 16: btw. This permission should only be in the specific job below rather than at the top. Otherwise it gives permissions to all of them. Even if there aren't others now, there could be later. This is checked by the openssf scorecard https://scorecard.dev/viewer/?uri=github.com/headlamp-k8s/headlamp

More details in the docs here: https://github.com/ossf/scorecard/blob/7ce8609469289d5f3b1bf5ee3122f42b4e3054fb/docs/checks.md#token-permissions

.github/workflows/backend-test.yml
```diff
@@ -21,7 +22,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
-      issues: write # needed for commenting on PRs for coverage changes
+      contents: read
+      pull-requests: write
```
illume, line 26: Can you please add the comment for what this permission is for?

illume (commented):
Left a few notes about some openssf scorecard checks regarding token permissions.

yolossn (approved):

illume (commented):
thanks

PR #2302, opened 2024-09-06
backend: Add tests

knrt10:
This increases code coverage for backend by adding unit tests for functions.

### DIFF
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


backend/cmd/headlamp.go
```diff
@@ -912,10 +912,12 @@ func checkHeadlampBackendToken(w http.ResponseWriter, r *http.Request) error {
 //nolint:gocognit,funlen
 func handleClusterHelm(c *HeadlampConfig, router *mux.Router) {
 	router.PathPrefix("/clusters/{clusterName}/helm/{.*}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if err := checkHeadlampBackendToken(w, r); err != nil {
-			logger.Log(logger.LevelError, nil, err, "failed to check headlamp backend token")
-
-			return
+		// Skip token check if we're in a test environment
+		if os.Getenv("HEADLAMP_TEST_ENV") != "true" {
```
illume, line 916: Is there a way to disable the token check in each test? I think this disables it for all the tests?

Perhaps adding the default token in the test? I think it's valuable to show in the test that something requires this token.

illume (commented):
Wow, that's a nice improvement to the test coverage.

Thank you.

I think separate commits would be a nice addition...
```
backend/cmd/stateless_test: Add happy path test for websocketConnContextKey
```

PR #2306, opened 2024-09-06
frontend: Update logo

sniok:
This update also gets rid off those pesky warnings about using viewbox="0 0 auto 32" 

image

image


illume (approved):
🎉 thanks

PR #2307, opened 2024-09-06
frontend: Fix Chooser popup header visibility

sniok:
Currently the header blends in with the background and in dark mode makes Info button not visible

Current, Dark mode:

image

Current, Light Mode:

image

After the change, Dark mode:

image

After the change, Light mode:

image


skoeva (approved):
Nice catch, looks good! Appreciate this fix

illume (approved):
👍 thanks

PR #2309, opened 2024-09-06
frontend: Fix missing optional dependencies

sniok:
Due to bug in the NPM (https://github.com/npm/cli/issues/4828) it generates inconsistent package-lock and app may fail on other platforms

Needed for #2231


illume (approved):
🎉 thanks

PR #2310, opened 2024-09-06
frontend: Explicitly add mui date pickers

sniok:
This library is required for material react table
Adding it explicitly to fix inconsistencies with package-lock

Needed for #2231


illume (approved):
👍 thanks

PR #2312, opened 2024-09-06
frontend: Refactor apiProxy

sniok:
Mostly split apiProxy into multiple files

Needed for #2231


frontend/src/lib/k8s/apiProxy/apiTypes.ts
```diff
@@ -0,0 +1,309 @@
+import { OpPatch } from 'json-patch';
+import { KubeMetadata, KubeObjectInterface } from '../cluster';
```
illume, line 2: This brings a circular import with cluster. I wonder if we can remove in this refactor?

Maybe a cluster.types.ts would allow this?

frontend/src/lib/k8s/apiProxy/apiTypes.ts
```diff
@@ -0,0 +1,309 @@
...
+  isNamespaced: boolean;
+  apiInfo: {
+    group: string;
+    version: string;
+    resource: string;
+  }[];
+  scale?: ScaleApi;
+}
+
+export interface ScaleApi {
```
illume, line 126: I wonder if we should break types out into .types.ts files? This is something mui does.

The benefit of having the types next to their implementation is that things are easier to find. scaleApi.types.ts

frontend/src/lib/k8s/apiProxy/apiUtils.ts
```diff
@@ -0,0 +1,67 @@
+import { omit } from 'lodash';
```
illume, line 1: Can we please move these into files named without utils?

frontend/src/components/common/Resource/PortForward.tsx
```diff
@@ -11,6 +11,7 @@ import {
   startPortForward,
   stopOrDeletePortForward,
 } from '../../../lib/k8s/apiProxy';
+import { PortForward as PortForwardState } from '../../../lib/k8s/apiProxy/portForward';
```
illume, line 14: It would be good to expand on the commit message explaining in more detail what's happening.

illume (commented):
Left some notes on things discussed on our call...

frontend/src/lib/k8s/apiProxy/apiTypes.ts
sniok: types were moved next to places where they're used, queryParameters moved to a new file

frontend/src/lib/k8s/apiProxy/apiTypes.ts
sniok: we can consider it. I think current type colocation is good enough

frontend/src/lib/k8s/apiProxy/apiUtils.ts
sniok: moved some functions around, only url formatting functions were left, renamed it to formatUrl.ts

frontend/src/components/common/Resource/PortForward.tsx
sniok: good idea. added a better description to the commit

illume (approved):
🎉 Thanks.

Issue #1493, opened 2023-10-25
a11y: Check and improve documentation accessibility


illume:
Check our READMEs and other documentation for accessibility issues.

- markup
- language (plain language hemmingway app)

To track the different files that have been checked.

- [ ] /README.md
- [ ] app/README.md
- [ ] backend/README.md
- [ ] frontend/README.md
- [ ] plugins/README.md
- [ ] docs/...
  - [ ] docs/

PR #2313, opened 2024-09-06
docs: Improve documentation a11y

skoeva:
These changes bring accessibility improvements to the docs, as well as minor improvements to wording, clarity, and formatting. The language in the docs was run through the Hemingway app to improve readability for a technical audience, and markdownlint was used for markdown formatting.

Fixes: #1493

### DIFF
--- a/docs/development/i18n/index.md
+++ b/docs/development/i18n/index.md
@@ -9,10 +9,10 @@ react-i18next libraries.
 
 ## Default language, and locales
 
-We started with an international English, and that will be the fallback language.
+We started with an international English and that will be the fallback language.
 
-Now we're starting with locales familiar, and will accept
-translations through github.
+Now we're starting with locales familiar to us and will accept
+translations through GitHub.
 
 ## Browser based language detection
 
@@ -20,7 +20,7 @@ We use
 [i18next-browser-languagedetector](https://github.com/i18next/i18next-browser-languageDetector#readme)
 
 This can select the browser language through various means like through
-cookies, the html language tag and other ways.
+cookies or the html language tag.
 
 One way to change the locale is to use `?lng=en` in the URL.
 
--- a/docs/development/plugins/index.md
+++ b/docs/development/plugins/index.md
@@ -3,7 +3,7 @@ title: Plugins
 sidebar_position: 6
 ---
 
-Plugins are one of the key features of Headlamp. They allow you to change how and what information is displayed, as well as other functionality. The ultimate goal of the plugins system is to allow vendors to build and deploy Headlamp with extra functionality without having to maintain a fork of the project.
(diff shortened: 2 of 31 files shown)


docs/development/api/classes/plugin_registry.PluginManager.md
```diff
@@ -88,7 +88,7 @@ Sends a request to install a plugin from the specified ArtifactHub URL.
 **`static`**
 
 **`example`**
-PluginManager.install('pluginID', ' https://artifacthub.io/packages/headlamp/<repo_name>/<plugin_name>');
+PluginManager.install('pluginID', ' <https://artifacthub.io/packages/headlamp/><repo_name>/<plugin_name>');
```
illume, line 91: This is a generated file. Also, I don't think this is correct because it's a source code example?

Note: `docs/development/api` is generated, so those changes should be made in the source code instead.  However, I know the doc generation is having issues at the moment, so if you change these ones in the source code, please don't regenerate the docs in this PR.

docs/development/i18n/index.md
```diff
@@ -9,18 +9,18 @@ react-i18next libraries.
 
 ## Default language, and locales
 
-We started with an international English, and that will be the fallback language.
+We started with an international English and that will be the fallback language.
 
-Now we're starting with locales familiar, and will accept
-translations through github.
+Now we're starting with locales familiar and will accept
```
illume, line 14: I guess this should be:
"Now we're starting with locales familiar **_to us_** and will accept"

docs/development/plugins/index.md
```diff
@@ -32,9 +32,14 @@ for more details.
 
 ## Developing Plugins
 
-Plugins are supposed to be built and shipped out-of-tree, i.e. instead of managing the plugins'
+Plugins are supposed to be built and shipped out-of-tree, i.e., instead of managing the plugins'
 code within the Headlamp project or a Headlamp fork (which would require
 always rebuilding/maintaining Headlamp when changing a plugin), they're
 supposed to be built outside of the project and loaded by Headlamp.
 
+Plugins are meant to be build and shipped out-of-tree (i.e., outside of the project and loaded by
```
illume, line 40: Looks like there's an extra paragraph here.

illume (changes requested):
Very nice work.  Thanks.

I read through it all and left a few small notes.

docs/development/plugins/index.md
skoeva: Ah! Thanks for catching this

docs/development/api/classes/plugin_registry.PluginManager.md
skoeva: Makes sense, I'll go ahead and roll back the changes in `docs/development/api` then

illume (approved):
🎉 thanks!

Issue #2316, opened 2024-09-09
Repeat translation key ERROR


vyncent-t:
There seems to be an issue when running the duplicate keys test

![image](https://github.com/user-attachments/assets/b6b2eeaf-0663-4033-8f7b-5588c239dd23)

illume:
This was fixed in 
- #2318

PR #2318, opened 2024-09-10
frontend: Fix duplicate translation keys pathing for windows

vyncent-t:
## Description

This PR resolves the issue where the duplicate keys test for translations fails on Windows environments due to path inconsistencies. Previously, the test passed without issues on Linux machines but failed on Windows due to differences in file path handling. The error has been fixed by using `path.basename`, ensuring that the test works correctly on both Linux and Windows environments.

Fixes issue #2316 

## Changes

- Updated the duplicate keys test to use `path.basename` for handling file paths.
- Ensured cross-platform compatibility for both Linux and Windows by standardizing the way paths are processed in the test.
- The test now successfully passes on both operating systems without raising a `TypeError`.

## Verification

- [x] Ran the duplicate keys test on Windows and confirmed that the `TypeError` no longer occurs.
- [ ] Tested the same on Linux to ensure that the test still passes without issues.
- [ ]  Verified that the translation keys are correctly processed on both systems using `path.basename`.

### DIFF
--- a/frontend/src/i18n/index.test.ts
+++ b/frontend/src/i18n/index.test.ts
@@ -26,7 +26,8 @@ async function checkKeys() {
    * Allowlist function scans the json and return if the word is within the allowlist
    */
   function allowlistScan(word: string, lang: string) {
-    return allowlist[lang].some((item: any) => item.wordKey === word);
+    const allowListLang: any[] = allowlist[lang];
+    return allowListLang.some((item: any) => item.wordKey === word);
   }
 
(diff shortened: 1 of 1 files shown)


sniok (approved):
nice!

sniok:
Ran `npm run test` locally (ubuntu wsl) and it passes

skoeva (approved):
Works for me locally too (Ubuntu in WSL), great work! LGTM

Issue #2161, opened 2024-07-15
Have an option to list plugins


joaquimrocha:
Headlamp should have a CLI option in the backend + desktop to list all plugins, showing a division between static/shipped plugins, and user added ones.

E.g. `headlamp --list-plugins`
or
E.g. `headlamp-server -list-plugins`

PR #2319, opened 2024-09-10
Add option to list plugins

skoeva:
This change creates a CLI subcommand in the backend + desktop to list all plugins, with a distinction between static/shipped plugins and user-added ones.

Fixes: #2161

#### Backend
- [x] Run `make backend`
- [x] Run `cd backend && ./headlamp-server list-plugins` and ensure that the list of plugins shows up

#### Desktop
- [x] Run `cd app && npm run build`
- [x] Run `cd dist/linux-unpacked && ./headlamp list-plugins` and ensure that the list of plugins shows up

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -60,6 +60,22 @@ const startUrl = (
   .replace(/\\/g, '/');
 
 const args = yargs(hideBin(process.argv))
+  .command(
+    'list-plugins',
+    'List all static and user-added plugins.',
+    () => {},
+    () => {
+      try {
+        const backendPath = path.join(process.resourcesPath, 'headlamp-server');
+        const stdout = execSync(`${backendPath} list-plugins`);
+        process.stdout.write(stdout);
+        process.exit(0);
+      } catch (error) {
+        console.error(`Error listing plugins: ${error}`);
+        process.exit(1);
+      }
+    }
+  )
   .options({
     headless: {
       describe: 'Open Headlamp in the default web browser instead of its app window',
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -8,9 +8,24 @@ import (
 	"github.com/headlamp-k8s/headlamp/backend/pkg/config"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/plugins"
 )
 
 func main() {
+	if len(os.Args) == 2 && os.Args[1] == "list-plugins" {
+		conf, err := config.Parse(os.Args[2:])
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "fetching config:%v")
+			os.Exit(1)
+		}
+
+		if err := plugins.ListPlugins(conf.StaticDir, conf.PluginsDir); err != nil {
+			logger.Log(logger.LevelError, nil, err, "listing plugins")
+		}
+
+		return
+	}
+
 	conf, err := config.Parse(os.Args)
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "fetching config:%v")
--- a/backend/pkg/plugins/plugins.go
+++ b/backend/pkg/plugins/plugins.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/fs"
 	"net/http"
@@ -77,8 +78,8 @@ func periodicallyWatchSubfolders(watcher *fsnotify.Watcher, path string, interva
 	}
 }
 
-// GeneratePluginPaths takes the staticPluginDir and pluginDir and returns a list of plugin paths.
-func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, error) {
+// generateSeparatePluginPaths takes the staticPluginDir and pluginDir and returns separate lists of plugin paths.
+func generateSeparatePluginPaths(staticPluginDir, pluginDir string) ([]string, []string, error) {
 	var pluginListURLStatic []string
 
 	if staticPluginDir != "" {
@@ -86,11 +87,21 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
 
 		pluginListURLStatic, err = pluginBasePathListForDir(staticPluginDir, "static-plugins")
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 	}
 
 	pluginListURL, err := pluginBasePathListForDir(pluginDir, "plugins")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return pluginListURLStatic, pluginListURL, nil
+}
+
+// GeneratePluginPaths generates a concatenated list of plugin paths from the staticPluginDir and pluginDir.
+func GeneratePluginPaths(staticPluginDir, pluginDir string) ([]string, error) {
+	pluginListURLStatic, pluginListURL, err := generateSeparatePluginPaths(staticPluginDir, pluginDir)
 	if err != nil {
 		return nil, err
 	}
@@ -101,6 +112,59 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
 	}
 
 	return pluginListURL, nil
+}
+
+// ListPlugins lists the plugins in the static and user-added plugin directories.
+func ListPlugins(staticPluginDir, pluginDir string) error {
+	staticPlugins, userPlugins, err := generateSeparatePluginPaths(staticPluginDir, pluginDir)
+	if err != nil {
+		logger.Log(logger.LevelError, nil, err, "listing plugins")
+		return fmt.Errorf("listing plugins: %w", err)
+	}
+
+	getPluginName := func(pluginDir string) string {
+		packageJSONPath := filepath.Join(pluginDir, "package.json")
+
+		content, err := os.ReadFile(packageJSONPath)
+		if err != nil {
+			// If there's an error reading package.json, just return the folder name as fallback.
+			return filepath.Base(pluginDir)
+		}
+
+		var packageData struct {
+			Name string `json:"name"`
+		}
+
+		// Parse the JSON and extract the name. If it fails, return the folder name.
+		if err := json.Unmarshal(content, &packageData); err != nil || packageData.Name == "" {
+			return strings.TrimPrefix(filepath.Base(pluginDir), "plugins/")
+		}
+
+		return packageData.Name
+	}
+
+	if len(staticPlugins) > 0 {
+		fmt.Printf("Static Plugins (%s):\n", staticPluginDir)
+
+		for _, plugin := range staticPlugins {
+			fmt.Println(" -", getPluginName(plugin))
+		}
+	} else {
+		fmt.Println("No static plugins found.")
+	}
+
+	if len(userPlugins) > 0 {
+		fmt.Printf("\nUser-added Plugins (%s):\n", pluginDir)
+
+		for _, plugin := range userPlugins {
+			pluginName := getPluginName(filepath.Join(pluginDir, plugin))
+			fmt.Println(" -", pluginName)
+		}
+	} else {
+		fmt.Printf("No user-added plugins found.")
+	}
+
+	return nil
 }
 
 // pluginBasePathListForDir returns a list of valid plugin paths for the given directory.
--- a/backend/pkg/plugins/plugins_test.go
+++ b/backend/pkg/plugins/plugins_test.go
@@ -2,6 +2,7 @@ package plugins_test
 
 import (
 	"context"
+	"io"
 	"net/http/httptest"
 	"os"
 	"path"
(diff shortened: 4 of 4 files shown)


backend/pkg/plugins/plugins.go
```diff
@@ -77,20 +77,30 @@ func periodicallyWatchSubfolders(watcher *fsnotify.Watcher, path string, interva
 	}
 }
 
-// GeneratePluginPaths takes the staticPluginDir and pluginDir and returns a list of plugin paths.
-func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, error) {
+// GenerateSeparatePluginPaths takes the staticPluginDir and pluginDir and returns a list of plugin paths.
```
illume, line 80: This says different to what it returns.

backend/pkg/config/config.go
```diff
@@ -86,6 +87,32 @@ func Parse(args []string) (*Config, error) {
 		return nil, fmt.Errorf("error parsing flags: %w", err)
 	}
 
+	// Handle list-plugins arg
```
illume, line 90: Instead of in the config module, this code should probably set the config.ListPlugins true here, and a ListPlugins function could be called inside `server.go@main`.

All the code related to plugins should generally go in the plugins folder. If it needs to be called from elsewhere then call the function from the plugins module. So that we can easily find where plugins related code is.

illume (commented):
Thanks.

I wrote a few notes about how the backend/ code is organized.

btw, Probably this option could be documented somewhere? We don't have a page for all options. The headlamp-plugin tool has them listed in its README.md. If you don't have any inspiration for this at the moment, we can document the command line arguments in a separate PR.

backend/pkg/plugins/plugins.go
skoeva: Made this more descriptive to reflect the two lists

backend/pkg/config/config.go
skoeva: Done

illume:
Looks good to me. But I think we can wait for Joaquim to see if it meets his requirements before merging.

ps. To make the code coverage bot use a happy face emoji you might consider adding a test for ListPlugins(possibly using a mock for GenerateSeparatePluginPaths).

vyncent-t:
was looking into this from our sync earlier this week so far so good

joaquimrocha (changes requested):
Tested and it works. Thanks!
Left a couple of comments to improve things.

app/electron/main.ts
```diff
@@ -78,6 +82,28 @@ const args = yargs(hideBin(process.argv))
...
+
+// If the user wants to list plugins, we don't need to start the app.
+if (listPlugins) {
+  const staticDir = path.join(process.resourcesPath, '.plugins');
+  const pluginsDir = path.join(app.getPath('userData'), 'plugins');
+
+  const staticPlugins = fs.existsSync(staticDir) ? fs.readdirSync(staticDir) : [];
+  const userPlugins = fs.existsSync(pluginsDir) ? fs.readdirSync(pluginsDir) : [];
+
+  console.log('  Static Plugins:');
```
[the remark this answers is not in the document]
skoeva: Done

app/electron/main.ts
```diff
@@ -78,6 +82,28 @@ const args = yargs(hideBin(process.argv))
   .help()
   .parseSync();
 
+const listPlugins = args['list-plugins'] === true;
+
+// If the user wants to list plugins, we don't need to start the app.
+if (listPlugins) {
+  const staticDir = path.join(process.resourcesPath, '.plugins');
+  const pluginsDir = path.join(app.getPath('userData'), 'plugins');
```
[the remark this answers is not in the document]
skoeva: Tried testing using PluginManager's `list` function, but it would break and return empty lists. I tried testing changes to `checkValidPluginFolder` to see if that would fix it but it caused new JS issues. I've made the other change and can continue testing with PluginManager but I seem to be hitting a wall with this

backend/cmd/server.go
```diff
@@ -17,6 +18,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if conf.ListPlugins {
+		if err := plugins.ListPlugins(conf.StaticDir, conf.PluginsDir); err != nil {
+			logger.Log(logger.LevelError, nil, err, "error listing plugins")
```
knrt10, line 23: Small NIT. Since this is already an error, adding error keywords adds redundancy. 

```suggestion
			logger.Log(logger.LevelError, nil, err, "listing plugins")
```

backend/cmd/server.go
```diff
@@ -17,6 +18,15 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		os.Exit(0)
```
knrt10, line 27: Why are we exiting here? This can be removed otherwise server will close.

backend/pkg/plugins/plugins.go
```diff
@@ -77,20 +77,30 @@ func periodicallyWatchSubfolders(watcher *fsnotify.Watcher, path string, interva
 	}
 }
 
-// GeneratePluginPaths takes the staticPluginDir and pluginDir and returns a list of plugin paths.
-func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, error) {
+// GenerateSeparatePluginPaths takes the staticPluginDir and pluginDir and returns separate lists of plugin paths.
+func GenerateSeparatePluginPaths(staticPluginDir string, pluginDir string) ([]string, []string, error) {
```
knrt10, line 81: We do not need to export it if this is used internally only

```suggestion
func generateSeparatePluginPaths(staticPluginDir, pluginDir string) ([]string, []string, error) {
```

backend/pkg/plugins/plugins.go
```diff
@@ -77,20 +77,30 @@ func periodicallyWatchSubfolders(watcher *fsnotify.Watcher, path string, interva
...
 	pluginListURL, err := pluginBasePathListForDir(pluginDir, "plugins")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return pluginListURLStatic, pluginListURL, nil
+}
+
+// GeneratePluginPaths generates a concatenated list of plugin paths from the staticPluginDir and pluginDir.
+func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, error) {
```
knrt10, line 102:
```suggestion
func GeneratePluginPaths(staticPluginDir, pluginDir string) ([]string, error) {
```

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,31 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
 }
 
+var GenerateSeparatePluginPathsFunc = GenerateSeparatePluginPaths
+
+// ListPlugins lists the plugins in the static and user-added plugin directories.
+func ListPlugins(staticPluginDir, pluginDir string) error {
+	staticPlugins, userPlugins, err := GenerateSeparatePluginPathsFunc(staticPluginDir, pluginDir)
+	if err != nil {
+		logger.Log(logger.LevelError, nil, err, "listing plugins")
+		return fmt.Errorf("error listing plugins: %w", err)
```
knrt10, line 123:
```suggestion
		return fmt.Errorf("listing plugins: %w", err)
```

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,31 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+	}
+
+	fmt.Println("  Static Plugins:")
```
knrt10, line 126: Can you please use `logger.Log` with Info field on this?

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,31 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+
+	for _, plugin := range staticPlugins {
+		fmt.Println("  -", plugin)
+	}
+
+	fmt.Println("  User-added Plugins:")
```
knrt10, line 132: Same here, please use `logger.Log`

knrt10 (changes requested):
Some suggestions.

backend/cmd/server.go
skoeva: That is the idea, to have the server terminate after running the command. It's simply to list plugins without running the server

backend/pkg/plugins/plugins.go
skoeva: I think using Print would be a lot easier to read in general as parsing the logs for the list of plugins seems tedious, WDYT?

backend/cmd/server.go
skoeva: Great catch! Removed these redundancies

backend/pkg/plugins/plugins.go
knrt10: Hmm, as you mentioned earlier we just need to list plugins and since we are not running a server here. Folks won’t need to parse the logs so I think fmt can work fine for this case. Cool

joaquimrocha (changes requested):

backend/pkg/plugins/plugins_test.go
```diff
@@ -178,6 +178,22 @@ func TestGeneratePluginPaths(t *testing.T) { //nolint:funlen
...
+	// mock the GenerateSeparatePluginPaths function to create lists of plugins
+	plugins.GenerateSeparatePluginPathsFunc = func(staticPluginDir string, pluginDir string) ([]string, []string, error) {
+		staticPlugins := []string{"static-plugin-1", "static-plugin-2"}
+		userPlugins := []string{"user-plugin-1", "user-plugin-2"}
+
+		return staticPlugins, userPlugins, nil
+	}
+
+	err := plugins.ListPlugins("", "")
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
skoeva: The original idea was keeping it concise without adding the extra steps, but for coverage's sake I added the extra steps

app/electron/main.ts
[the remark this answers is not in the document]
skoeva: Changed this to call execSync using the backend logic

joaquimrocha (changes requested):
Nice changes. It's almost there.
I left a few comments.
About koanf picking list-plugins from config (e.g. `HEADLAMP_CONFIG_LIST_PLUGINS=1 ./backend/headlamp-server`), @yolossn do you think this could be of any user? I think it's just a bit weird to have a command be triggered on a flag like this.
If we should remove this, we have to remove the koanf tag from the declaration but then we have to do something extra so the flag ends up with the right value in the conf struct.

app/electron/main.ts
```diff
@@ -78,6 +82,20 @@ const args = yargs(hideBin(process.argv))
   .help()
   .parseSync();
 
+const listPlugins = args['list-plugins'] === true;
+
+if (listPlugins) {
+  try {
+    const backendPath = path.join(process.resourcesPath, 'headlamp-server');
+    const stdout = execSync(`${backendPath} --list-plugins`);
+    console.log(stdout.toString().trim());
```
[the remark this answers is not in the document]
skoeva: I noticed a discrepancy in formatting between the outputs from the app and from the backend, turns out the fix is using `process.stdout.write` rather than `console.log`

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,39 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+
+		for _, plugin := range staticPlugins {
+			fmt.Println("  -", plugin)
+		}
+	} else {
+		fmt.Printf("  No static plugins found.\n")
+	}
+
+	if len(userPlugins) > 0 {
+		fmt.Printf("  User-added Plugins (%s):\n", pluginDir)
```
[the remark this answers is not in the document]
skoeva: Done

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,39 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+// ListPlugins lists the plugins in the static and user-added plugin directories.
+func ListPlugins(staticPluginDir, pluginDir string) error {
+	staticPlugins, userPlugins, err := GenerateSeparatePluginPathsFunc(staticPluginDir, pluginDir)
+	if err != nil {
+		logger.Log(logger.LevelError, nil, err, "listing plugins")
+		return fmt.Errorf("listing plugins: %w", err)
+	}
+
+	if len(staticPlugins) > 0 {
+		fmt.Printf("  Static Plugins (%s):\n", staticPluginDir)
```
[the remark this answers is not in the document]
skoeva: Done

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,39 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+		}
+	} else {
+		fmt.Printf("  No static plugins found.\n")
+	}
+
+	if len(userPlugins) > 0 {
+		fmt.Printf("  User-added Plugins (%s):\n", pluginDir)
+
+		for _, plugin := range userPlugins {
+			fmt.Println("  -", plugin)
```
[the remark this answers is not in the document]
skoeva: Done

backend/pkg/config/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	InsecureSsl           bool   `koanf:"insecure-ssl"`
 	EnableHelm            bool   `koanf:"enable-helm"`
 	EnableDynamicClusters bool   `koanf:"enable-dynamic-clusters"`
+	ListPlugins           bool   `koanf:"list-plugins"`
```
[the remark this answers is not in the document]
illume: After the changes since, is this question still relevant? /cc @yolossn

joaquimrocha (approved):

joaquimrocha:
Let's just wait for @yolossn 's comments on the koanf thing. Otherwise LGTM.

yolossn:
> Let's just wait for @yolossn 's comments on the koanf thing. Otherwise LGTM.

@joaquimrocha It is not clear if we want to handle listing plugins using a subcommand or using an argument. And the output is just printing the path and not the actual plugins names in the package.json so that is also a blocker.

joaquimrocha (changes requested):
After talking to @yolossn I think we should make this a subcommand. Being a subcommand we also can then use options for that subcommand, like for listing names/paths/etc.

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,40 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
 	return pluginListURL, nil
 }
 
+var GenerateSeparatePluginPathsFunc = generateSeparatePluginPaths
```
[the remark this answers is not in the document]
skoeva: Thanks for catching this, must have been remnants of trying to test by mocking that function

backend/cmd/server.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/headlamp-k8s/headlamp/backend/pkg/config"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/plugins"
 )
 
 func main() {
```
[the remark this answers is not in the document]
skoeva: Done

backend/pkg/plugins/plugins.go
```diff
@@ -103,6 +113,40 @@ func GeneratePluginPaths(staticPluginDir string, pluginDir string) ([]string, er
...
+
+// ListPlugins lists the plugins in the static and user-added plugin directories.
+func ListPlugins(staticPluginDir, pluginDir string) error {
```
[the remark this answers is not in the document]
skoeva: Changed to use the package.json plugin name, with the path name as fallback

backend/cmd/server.go
```diff
@@ -17,6 +18,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	if conf.ListPlugins {
+		if err := plugins.ListPlugins(conf.StaticDir, conf.PluginsDir); err != nil {
+			logger.Log(logger.LevelError, nil, err, "listing plugins")
+			os.Exit(1)
+		}
+
+		os.Exit(0)
```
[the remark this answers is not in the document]
skoeva: Done

backend/pkg/config/config.go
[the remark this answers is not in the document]
skoeva: Confirmed and changed

backend/cmd/server.go
```diff
@@ -8,9 +8,24 @@ import (
 	"github.com/headlamp-k8s/headlamp/backend/pkg/config"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/plugins"
 )
 
 func main() {
+	if len(os.Args) == 2 && os.Args[1] == "--list-plugins" {
```
yolossn, line 15:
```suggestion
	if len(os.Args) == 2 && os.Args[1] == "list-plugins" {
```

skoeva: This would change the option from `./headlamp-server --list-plugins` to `/headlamp-server list-plugins`, is that the intention? Do we want to make this distinct from other options @joaquimrocha

joaquimrocha: Yes. This way, at a later stage we can add options to this subcommand (`./headlamp-server list-plugins --paths` or `./headlamp-server list-plugins --static` or `./headlamp-server list-plugins --user-added`)

skoeva: Done

joaquimrocha:
Thank you @skoeva !

PR #2323, opened 2024-09-11
backend/cmd/headlamp: Fix issue in cluster looking for kube config

### LINKED ISSUE #1826: In-cluster deployment looks for a kubeconfig
Trying to deploy headlamp using the kubernetes-headlamp.yaml in to a k8s cluster.
Does the in cluster deployment suppose to look for the kubeconfig?

```
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":305,"time":"2024-03-18T22:51:07Z","message":"Creating Headlamp handler"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":306,"time":"2024-03-18T22:51:07Z","message":"Kubeconfig path: "}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":307,"time":"2024-03-18T22:51:07Z","message":"Static plugin dir: /headlamp/static-plugins"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":308,"time":"2024-03-18T22:51:07Z","message":"Plugins dir: /headlamp/plugins"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":309,"time":"2024-03-18T22:51:07Z","message":"Dynamic clusters support: false"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":310,"time":"2024-03-18T22:51:07Z","message":"Helm support: false"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":311,"time":"2024-03-18T22:51:07Z","message":"Proxy URLs: []"}
{"level":"info","context":"main","clusterURL":"https://10.96.0.1:443","source":"/headlamp/backend/pkg/kubeconfig/kubeconfig.go","line":172,"time":"2024-03-18T22:51:07Z","message":"Proxy setup"}
*** Headlamp Server ***
  API Routers:
{"level":"error","source":"/headlamp/backend/cmd/headlamp.go","line":365,"error":"read /: is a directory","time":"2024-03-18T22:51:07Z","message":"loading kubeconfig"}
{"level":"error","source":"/headlamp/backend/cmd/headlamp.go","line":376,"error":"open /home/headlamp/.config/Headlamp/kubeconfigs/config: no such file or directory","time":"2024-03-18T22:51:07Z","message":"loading dynamic kubeconfig"}
```

illume:
Fixes #1826

I'm not sure what's best? The first commit disables it.

There is a use case of using kube config in cluster here: https://github.com/headlamp-k8s/headlamp/issues/1826#issuecomment-2343595252

But are the other statements needed in-cluster too?

I guess if we do support kube config loading in cluster then they should be warnings and not error logs?

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -365,21 +365,33 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	fmt.Println("*** Headlamp Server ***")
 	fmt.Println("  API Routers:")
 
-	if !config.useInCluster {
-		// load kubeConfig clusters
-		err := kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPath, kubeconfig.KubeConfig)
-		if err != nil {
+	// load kubeConfig clusters
+	err := kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPath, kubeconfig.KubeConfig)
+	if err != nil {
+		if config.useInCluster {
+			logger.Log(logger.LevelInfo, nil, err, "loading kubeconfig")
+		} else {
 			logger.Log(logger.LevelError, nil, err, "loading kubeconfig")
 		}
+	}
 
-		// load dynamic clusters
-		kubeConfigPersistenceFile, err := defaultKubeConfigPersistenceFile()
-		if err != nil {
+	// load dynamic clusters
+	// @todo: is this needed in-cluster at all?
+	kubeConfigPersistenceFile, err := defaultKubeConfigPersistenceFile()
+	if err != nil {
+		if config.useInCluster {
+			logger.Log(logger.LevelInfo, nil, err, "getting default kubeconfig persistence file")
+		} else {
 			logger.Log(logger.LevelError, nil, err, "getting default kubeconfig persistence file")
 		}
+	}
 
-		err = kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPersistenceFile, kubeconfig.DynamicCluster)
-		if err != nil {
+	// @todo: is this needed in-cluster at all?
+	err = kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPersistenceFile, kubeconfig.DynamicCluster)
+	if err != nil {
+		if config.useInCluster {
+			logger.Log(logger.LevelInfo, nil, err, "loading dynamic kubeconfig")
+		} else {
 			logger.Log(logger.LevelError, nil, err, "loading dynamic kubeconfig")
 		}
 	}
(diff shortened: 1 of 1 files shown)


backend/cmd/headlamp.go
```diff
@@ -365,21 +365,23 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	fmt.Println("*** Headlamp Server ***")
 	fmt.Println("  API Routers:")
 
-	// load kubeConfig clusters
-	err := kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPath, kubeconfig.KubeConfig)
-	if err != nil {
-		logger.Log(logger.LevelError, nil, err, "loading kubeconfig")
-	}
+	if !config.useInCluster {
```
joaquimrocha: Many users have asked how they could add more clusters when headlamp is running in a cluster. I find the mixed use-case a bit cornerish, but don't see a reason why we should restrict mixing loading kubeconfigs with in-cluster. So maybe what we need to do is not log issues about loading kubeconfigs if we are running in-cluster.

backend/cmd/headlamp.go
```diff
@@ -365,21 +365,33 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
 
-	if !config.useInCluster {
-		// load kubeConfig clusters
-		err := kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPath, kubeconfig.KubeConfig)
-		if err != nil {
+	// load kubeConfig clusters
+	err := kubeconfig.LoadAndStoreKubeConfigs(config.kubeConfigStore, kubeConfigPath, kubeconfig.KubeConfig)
+	if err != nil {
+		if config.useInCluster {
+			logger.Log(logger.LevelInfo, nil, err, "loading kubeconfig")
```
joaquimrocha, line 372: maybe we could define the log into a variable and then only have one logger.Log call? I guess the code would be more easily understood.

joaquimrocha (changes requested):
Left a comment in the 1st patch that then I realize you had already implemented in the 2nd patch.
I think the 2nd patch is the right approach. Let's keep the possibility of mixing loading clusters from different places, but I don't think we need the new todo comments.

illume:
There's a bit more discussion since in this issue:
> There is a use case of using kube config in cluster here: https://github.com/headlamp-k8s/headlamp/issues/1826#issuecomment-2343595252

joaquimrocha:
> There's a bit more discussion since in this issue:
> 
> > There is a use case of using kube config in cluster here: [#1826 (comment)](https://github.com/headlamp-k8s/headlamp/issues/1826#issuecomment-2343595252)

Either way, I think we need to keep supporting this. By default we don't look for a kubeconfig if `-in-cluster` is passed, so I don't see an inconvenient with keeping support for this "misfeature".
In terms of making it work, just not aborting if `-in-cluster` is also used should be sufficient.

joaquimrocha:
Maybe @skoeva can revamp this if it's still valid.

Issue #2324, opened 2024-09-11
make plugins-test in CI fails


skoeva:
Error output here:
![image](https://github.com/user-attachments/assets/dab66f23-769a-474b-acc2-71c0f38ee4b8)

PR #2325, opened 2024-09-11
headlamp-plugin: Install tar and update dependencies-sync

skoeva:
This change addresses the plugins test failure in the CI.

Fixes: #2324

#### Testing
Tested on issue #1996 in [this commit](https://github.com/headlamp-k8s/headlamp/commit/ddd6c4a3cc49b67878bdfe05af285a37c9c6eb06).
- [x] Run `make plugins-test` and ensure that it passes

### DIFF
--- a/plugins/headlamp-plugin/dependencies-sync.js
+++ b/plugins/headlamp-plugin/dependencies-sync.js
@@ -54,6 +54,7 @@ const dependenciesFrontDoesNotHave = new Set([
   'monaco-editor-webpack-plugin',
   'webpack',
   'table',
+  'tar',
 ]);
 
 // Dependencies from frontend/package.json that aren't wanted by headlamp-plugin.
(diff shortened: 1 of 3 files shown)


vyncent-t:
it would be good to detail out the reproduction steps, I'm assuming its meant to be ran at the root level of the project.

I also seem to be hitting a different error than the one you have in the original issue

image

vyncent-t:
also, I would try

- pull from main for the most recent changes
- remove the node modules folders in the frontend and plugins
- run npm i for frontend and then npm i for the plugins
- run make plugins-test again

frontend/package.json
```diff
@@ -92,6 +92,7 @@
     "spacetime": "^7.4.0",
     "stream-browserify": "^3.0.0",
     "stream-http": "^3.2.0",
+    "tar": "^7.4.3",
```
sniok, line 95: tar is only missing from the plugins/headlamp-plugin, no need to add to the frontend

sniok (changes requested):
nice fix, thanks! just remove tar from frontend and we can merge this

sniok (approved):

PR #2326, opened 2024-09-12
backend: refactor listing all contexts and error

knrt10:
This removes the dependacy of manually parsing of k8s resources. It was a problem as we had to be in sync with k8s API changes, which is not a good way. This change removes that and it splits kubeconfig into seperate contexts and validates them.

### Testing

- In your kubeconfig add some context with partial validity. For reference please check [here](https://github.com/headlamp-k8s/headlamp/blob/e13376a49d4191032926373e11d30a1a6c5ff4f7/backend/pkg/kubeconfig/test_data/kubeconfig_partialcontextvalid)
- Run backend `cd backend && go run ./cmd -dev`
- Run frontend `cd frontend && npm start`
- You should see errors in backend and all clusters in frontend.
- You can try with different kubeconfig and check.

### Reference

https://github.com/headlamp-k8s/headlamp/pull/2289#issuecomment-2343545934

### DIFF
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -14,12 +14,12 @@ import (
 	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/tools/clientcmd/api"
 )
 
 // TODO: Use a different way to avoid name clashes with other clusters.
@@ -41,6 +41,7 @@ type Context struct {
 	OidcConf    *OidcConfig            `json:"oidcConfig"`
 	proxy       *httputil.ReverseProxy `json:"-"`
 	Internal    bool                   `json:"internal"`
+	Error       string                 `json:"error"`
 }
 
 type OidcConfig struct {
@@ -118,6 +119,26 @@ type DataError struct {
 // Error returns a string representation of the error.
 func (e DataError) Error() string {
 	return fmt.Sprintf("Error in field '%s': %s", e.Field, e.Reason)
+}
+
+// Base64Error is an error that occurs when decoding base64 data.
+type Base64Error struct {
+	ContextName string
+	ClusterName string
+	UserName    string
+	Errors      []error
+}
+
+// Error returns a string representation of the error.
+func (e Base64Error) Error() string {
+	var messages []string //nolint:prealloc
+
+	for _, err := range e.Errors {
+		messages = append(messages, err.Error())
+	}
+
+	return fmt.Sprintf("Base64 decoding errors in context '%s', cluster '%s', user '%s':\n%s",
+		e.ContextName, e.ClusterName, e.UserName, strings.Join(messages, "\n"))
 }
 
 // ClientConfig returns a clientcmd.ClientConfig for the context.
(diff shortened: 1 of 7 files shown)


backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -316,299 +335,362 @@ func loadContextsFromData(data []byte, source int) ([]Context, []error) {
...
+// createAndValidateConfig creates and validates the config.
+func createAndValidateConfig(
+	contextName string,
+	contextMap,
+	cluster,
+	user map[interface{}]interface{},
+	kubeconfig map[string]interface{},
+) (*api.Config, error) {
+	singleConfig := createKubeConfig(contextName,
+		toStringKeyMap(contextMap),
```
yolossn, line 474: why is the `toStringKeyMap` conversion used here?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -316,299 +335,362 @@ func loadContextsFromData(data []byte, source int) ([]Context, []error) {
...
-			}
-		}
+	authInfo, exists := clientConfig.AuthInfos[context.AuthInfo]
+	if !exists {
+		return Context{}, fmt.Errorf("user not found in loaded config")
 	}
 
-	return kubeContext
+	return Context{
+		Name:        contextName,
```
yolossn, line 688: the contextName is not made `dnsFriendly`

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -316,299 +335,362 @@ func loadContextsFromData(data []byte, source int) ([]Context, []error) {
...
-	if cluster, ok := contextData["cluster"].(string); ok {
-		kubeContext.Cluster = cluster
-	}
+	return kubeconfig
+}
 
-	if namespace, ok := contextData["namespace"].(string); ok {
-		kubeContext.Namespace = namespace
+// convertToContext converts the config to a context.
+func convertToContext(contextName string, clientConfig *api.Config, source int) (Context, error) {
```
yolossn, line 671: SetupProxy is not called for the context in this function.

backend/pkg/kubeconfig/kubeconfig.go
knrt10: `toStringKeyMap` is used to ensure that all map keys are strings, which is a requirement for proper YAML marshaling and compatibility with the client-go library.

It ensures that all the key-value pairs in `contextMap, cluster, and user` have string keys before they're used to create the new config.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -316,299 +335,381 @@ func loadContextsFromData(data []byte, source int) ([]Context, []error) {
...
+	err := newContext.SetupProxy()
+	if err != nil {
+		return Context{}, ContextError{ContextName: contextName, Reason: fmt.Sprintf("couldn't setup proxy: %v", err)}
 	}
 
-	return kubeContext
+	return newContext, nil
 }
 
 // LoadContextsFromAPIConfig loads contexts from the given api.Config.
```
yolossn, line 715: The `LoadContextsFromAPIConfig` function is not used anywhere in the code, remove it if it is not used. If it is replaced by another function then make sure that it also supports the `skipProxySetup` functionality.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,33 +309,40 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
 
-	return loadContextsFromData(kubeConfigByte, source)
+	var skipProxySetup bool
+
+	if source == KubeConfig {
+		skipProxySetup = false
+	} else {
+		skipProxySetup = true
+	}
+
```
joaquimrocha, line 319: Same comment as before.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,33 +309,40 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
 
 	var errs []error
 
-	// Parse and validate the config.
-	rawConfig, _, err := parseAndValidateConfig(data)
+	// Unmarshal the kubeconfig data
+	kubeconfig, err := UnmarshalKubeconfig(data)
 	if err != nil {
-		// Log the error but continue processing
-		errs = append(errs, fmt.Errorf("error parsing config: %v", err))
```
joaquimrocha, line 300: We don't need to format the error anymore?

knrt10: It is now handled inside `UnmarshalKubeconfig` function

backend/pkg/kubeconfig/kubeconfig_test.go
```diff
@@ -76,7 +77,7 @@ func TestLoadContextsFromKubeConfigFile(t *testing.T) {
 		contexts, errs := kubeconfig.LoadContextsFromFile(kubeConfigFile, kubeconfig.KubeConfig)
 		require.NotEmpty(t, errs, "Expected some errors for partially valid file")
 		require.NotEmpty(t, contexts, "Expected some valid contexts from partially valid file")
-		require.Equal(t, 1, len(contexts), "Expected one context from the partially valid file")
+		require.Equal(t, 3, len(contexts), "Expected one context from the partially valid file")
```
illume, line 80: Seems there's 3 now, but the comment says one?

It looks like there's only 1 valid context, and 2 invalid ones in the test file kubeconfig_partialcontextvalid. So the test should check that, and it doesn't at the moment.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,334 +304,412 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
 }
 
-// loadContextsFromData is a helper function that loads contexts from a byte slice.
-// It parses the byte slice as a YAML file and validates the parsed data.
-// It then creates contexts from the parsed data and returns them.
-// It returns any errors that occurred during parsing or validation.
-func loadContextsFromData(data []byte, source int) ([]Context, []error) {
+// loadContextsFromData loads contexts from a kubeconfig data.
+// It unmarshals the kubeconfig data, extracts the contexts, and processes each context.
+// It returns the processed contexts and any errors that occurred during the process.
```
illume, line 314: Can you please add some documentation on how to discover if a context is valid, and that it returns all contexts even if they are invalid?

It looks to me there's a bug, because errors are only conditionally added. So there could be 3 contexts and 2 errors. But are the errors for the 1,2, or 3rd contexts? There's no way to tell.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,334 +304,412 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
-// It returns the created context and any errors that occurred.
-// It returns an error if the context cannot be created.
-// It returns a valid context if the context is created.
-// It returns any errors that occurred during context creation.
-func createContext(rawContext interface{}, rawClusters, rawUsers []interface{}, source int) (Context, error) {
-	contextMap, ok := rawContext.(map[interface{}]interface{})
-	if !ok {
-		return Context{}, ContextError{ContextName: "unknown", Reason: "invalid context format"}
-	}
+// ProcessContext processes a context from the kubeconfig.
```
illume, line 367: I'm not sure what some of these arguments are. Can you please document them along with the return?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,334 +304,412 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
 
-	if cluster, ok := contextData["cluster"].(string); ok {
-		kubeContext.Cluster = cluster
-	}
+	return kubeconfig
+}
 
-	if namespace, ok := contextData["namespace"].(string); ok {
-		kubeContext.Namespace = namespace
+// convertToContext converts the config to a context.
```
illume, line 674: Can you please add docs for the args here as well?

backend/pkg/kubeconfig/kubeconfig_test.go
knrt10: The scope of the PR has been changed, it now doesn't return partial valid contexts, it returns all contexts. Will edit this, thanks

backend/pkg/kubeconfig/kubeconfig.go
knrt10: We are now not checking which context it valid or not, we are now just returning all contexts and errors

backend/pkg/kubeconfig/kubeconfig_test.go
```diff
@@ -67,7 +68,7 @@ func TestLoadContextsFromKubeConfigFile(t *testing.T) {
 
 		contexts, errs := kubeconfig.LoadContextsFromFile(kubeConfigFile, kubeconfig.KubeConfig)
 		require.NotEmpty(t, errs, "Expected errors for invalid auth file")
-		require.Empty(t, contexts, "Expected no contexts from invalid auth file")
+		require.NotEmpty(t, contexts, "Expected contexts from invalid auth file")
```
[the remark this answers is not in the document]
joaquimrocha: I think that if the context here is missing a user or something, then we could check for that in the tests.
We do have to return all clusters to the frontend but maybe in a different PR we can add a key/value to the context to show that there is a problem.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -281,334 +304,422 @@ func LoadContextsFromBase64String(kubeConfig string, source int) ([]Context, []e
...
-func createContext(rawContext interface{}, rawClusters, rawUsers []interface{}, source int) (Context, error) {
-	contextMap, ok := rawContext.(map[interface{}]interface{})
-	if !ok {
-		return Context{}, ContextError{ContextName: "unknown", Reason: "invalid context format"}
-	}
+// ProcessContext processes a context from the kubeconfig.
+// It returns errors for invalid contexts, but will return all contexts.
+// rawContext can be a single context or a list of contexts.
+// kubeconfig is the kubeconfig data.
+// source is the source of the kubeconfig.
```
illume, line 372: Sorry, can you please expand on source further? It's an index into something?

backend/pkg/kubeconfig/kubeconfig.go
illume: @yolossn does this make sense for you?

backend/pkg/kubeconfig/kubeconfig_test.go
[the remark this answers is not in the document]
illume: I reopened this, because I'm still confused about this. I still think there could be a bug/undefined issue with this function. Because the error array returned is not matched up with the returned contexts.

How do the callers know which contexts failed, and which contexts are invalid?

All the LoadContexts (and LoadContextsFromMultipleFiles, LoadContextsFromBase64String) users would also need to be changed so that all the errors are returned, and all the errors are matched up.

loadContextsFromData for example could return contexts[] and errors[] with when there is no error a nil is stored in the errors[] array. In this case the caller would need to check each error element for nil or error. Maybe it should return (contexts[], contextErrors[], error) to make the error handling clear.

backend/pkg/kubeconfig/kubeconfig.go
knrt10: Sure, let me clarify it more.

backend/pkg/kubeconfig/kubeconfig_test.go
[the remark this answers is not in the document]
knrt10: You're right, and that's a great observation. The current implementation doesn't provide a clear way to know which contexts failed or are invalid, it only logs it with the data and context name. Let me fix it

backend/pkg/kubeconfig/kubeconfig.go
yolossn: Yes it does

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -258,357 +279,492 @@ func (c *Context) AuthType() string {
...
+}
 
-		for k, v := range config {
-			if key, ok := k.(string); ok {
-				if val, ok := v.(string); ok {
-					authProvider.Config[key] = val
-				}
-			}
+// getNameOrUnknown returns the name of the cluster or user if it exists, otherwise it returns "<unknown>".
+func getNameOrUnknown(data map[interface{}]interface{}, key string) string {
```
illume, line 541: I'm not sure about using "<unknown>" here. Why was that chosen over an empty string "" ?

I also wonder about the uniqueness of the name if there are multiple ones with out a name?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -258,357 +279,492 @@ func (c *Context) AuthType() string {
...
 	}
 
-	friendlyName := makeDNSFriendly(name)
-	kubeContext := createKubeContext(contextData)
+	// Create and validate the config
+	singleConfig, err := createAndValidateConfig(contextName, contextMap, cluster, user, kubeconfig)
+	if err != nil {
+		errs = append(errs, err)
+		context = Context{
+			Name:  contextName,
```
illume, line 436: I think Name here is not unique. So where there's non-unique names finding which error goes with which context won't be possible.

backend/pkg/kubeconfig/kubeconfig.go
illume: Ultimately we'd want to ask the user if they can set a name?

Additionally, what about using the other two fields? User name and cluster name?

Maybe we can also handle this case outside of this PR.

The errors and context could be returned as a parallel array, so that at least in our code the error can be matched up with the context no matter the name.

illume (approved):
👍 thanks for this

illume:
I think we can merge this because it fixes some issues, and generally improves things.

However, in the future I think we need to look into using context Name or not. It's not really unique or even required to be there strictly (however for 99% of users it probably is there). We need something which is unique across clusters, and is stable after a restart. Currently much of the frontend assumes this value is non-empty and unique - which isn't true strictly.


Leaving my comments here for future work...
---

I'm not sure about using "<unknown>" here. Why was that chosen over an empty string "" or nil?

kubectl just gives nothing if a context doesn't have a name.
```
kubectl config get-contexts
```

kubectl can work without a name, because it uses the current context.

I also wonder about the uniqueness of the name if there are multiple ones with out a name?



Ultimately we'd want to ask the user if they can set a name?

Additionally, what about using the other two fields? User name and cluster name?

Maybe we can also handle this case outside of this PR.

The errors and context could be returned as a parallel array, so that at least in our code the error can be matched up with the context no matter the name.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -258,357 +279,492 @@ func (c *Context) AuthType() string {
...
+// getNameOrUnknown returns the name of the cluster or user if it exists, otherwise it returns "<unknown>".
+func getNameOrUnknown(data map[interface{}]interface{}, key string) string {
+	if nameInterface, ok := data[key]; ok {
+		if nameString, ok := nameInterface.(string); ok {
+			return nameString
 		}
 	}
 
-	return authProvider
+	return "<unknown>"
```
illume, line 548: Maybe just "unknown"? I don't know if "<>" are valid. 

Or for context name could possible use (cluster name and user name?)

PR #2331, opened 2024-09-13
Add option to disable plugins

skoeva:
This change creates a CLI option in the backend + desktop to disable plugins.

Fixes: #2162

### How it works
- `--disable-plugins` disables all plugins
- `--disable-plugins=plugin1,plugin2` disables plugins plugin1 and plugin2


vyncent-t:
lets sync on this soon

skoeva:
#### Current status: 
- `--disable-plugins=plugin1,plugin2` removes plugin1 and plugin2 from the plugin list
- `--disable-plugins` does not make any changes to the plugin list
- `--disable-plugins=all` removes all plugins from the plugin list

#### Recent testing:
- Tried to change logic from removing plugins in the list to disabling them within the list, but this seems to be having issues
  - Disconnect between frontend and backend. If we change the output type in `plugins.go` from `[]string` to a list of plugin objects with field `isEnabled` (similar to `PluginInfo` from pluginsSlice), we see that changes to the backend are not reflected in the frontend and vice versa
- Tried to allow for default value using `NoOptDefVal` but seems to not be working. Could be because the flag arguments are strings? This is unclear atm
  - Also tried to handle this in the `Parse()` function from `config.go` by manually setting the empty string argument to "all", but this was not working either

#### To be done:
- Figure out how to allow for flag with no argument (`./headlamp --disable-plugins`) to disable all plugins
- Figure out how to adjust logic from removing plugins to disabling them without creating synchronization issues with the frontend and backend

Issue #2314, opened 2024-09-06
"Something went wrong" alert needs some work


sniok:
Tested on 0.25.1, disabled the network and this is how it looks like 

On wide screens it's way too short
![image](https://github.com/user-attachments/assets/e2384e9c-b64a-412f-acfb-19dfc3c93ecc)

Medium screens, covers the logo
![image](https://github.com/user-attachments/assets/41d3ec84-dbb0-4720-adb5-ff68b4cf5da7)

On narrow screens, padding pushes the content all the way to the right
![image](https://github.com/user-attachments/assets/a446d3a7-aefe-4df7-9a66-f1181a93f210)
![image](https://github.com/user-attachments/assets/1ef47ae5-42b9-47a0-8e44-984fadb6abcf)

This seems like a good component to use 

https://mui.com/material-ui/react-alert/

sniok:
Also I think the message should be "Lost connection to the cluster" because it checks if the browser is online or if healthz endpoint of a cluster responds.

skoeva:
+1 this banner has been really buggy and the message isn't at all visibly traceable to the issue

skoeva:
Got around to checking out the alert component, looks promising. We can keep the width at 100% as before, but I think it would be cleaner and more convenient if the alert was a bit smaller (maybe something closer to this):

![image](https://github.com/user-attachments/assets/f571841a-f0b4-40e2-9904-931185660e40)

@sniok WDYT?

PR #2336, opened 2024-09-17
frontend: AlertNotification: Improve alert component

skoeva:
This change addresses issues with accessibility and clarity for the `AlertNotification` component. This now uses the React Alert component from Material UI and displays a more descriptive error message: "Lost connection to the cluster".

Fixes: #2314 

#### Desktop/web view
image

#### Mobile view
image

### DIFF
--- a/frontend/src/components/common/AlertNotification.tsx
+++ b/frontend/src/components/common/AlertNotification.tsx
@@ -1,11 +1,10 @@
-import { Box, Button } from '@mui/material';
+import { Alert, Button, Typography } from '@mui/material';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { matchPath, useLocation } from 'react-router-dom';
 import { testClusterHealth } from '../../lib/k8s/apiProxy';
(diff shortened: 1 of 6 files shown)


frontend/src/components/common/AlertNotification.tsx
```diff
@@ -99,32 +97,32 @@ export function PureAlertNotification({ checkerFunction }: PureAlertNotification
...
+      variant="filled"
+      severity="error"
       sx={theme => ({
         color: theme.palette.common.white,
+        background: theme.palette.error.main,
         textAlign: 'center',
         display: 'flex',
-        paddingTop: theme.spacing(1),
-        paddingBottom: theme.spacing(0.5),
+        paddingTop: theme.spacing(0.5),
```
vyncent-t, line 108: looks nice from out sync earlier, I would maybe try adding some break points for the smaller view so that it is stacked and not so squished

skoeva: Tweaked the padding on the right, should look more proportionate now

illume (approved):
👍 thanks

PR #2345, opened 2024-09-19
backend: Use token only if present in PortForward

### LINKED ISSUE #1834: Issue with portforward on Mac
A user at KubeCon just reported to me that the port forward feature has never worked for him in Mac (he's upgraded to the latest version and still has problems).
Headlamp was installed from homebrew.

yolossn:
this patch uses the token in startPortForward only if the token from the API request is not empty.

fixes #1834 

before:

https://github.com/user-attachments/assets/e40a708b-7ccc-4fd9-bc4c-c92b470a0dcb

after:

https://github.com/user-attachments/assets/398c7aba-958f-4ef0-90cf-d6df83624a45


joaquimrocha (approved):
Nice fix! Thanks!

PR #2356, opened 2024-09-26
Add custom resource instances page

joaquimrocha:
(taken from #2226 as the author @Guilamb has no availability to work on it anymore)

This pull request addresses the issue of restricted user access to custom resources and the need to view only the instantiated custom resources (https://github.com/headlamp-k8s/headlamp/issues/955 ), by introducing a new CR instances page in Headlamp that displays a list of all instantiated custom resources, that can be filtered by category.

The CR instances page provides users with the capability to view their Resource Status and Definitions without requiring the development of a custom UI. It also allows users to filter namespaces, similar to the existing functionality for Workloads, ensuring restricted permissions are maintained.

Fixes: #1962


frontend/src/components/crd/CustomResourceInstancesList.tsx
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  useEffect(() => {
+    const fetchCRs = async () => {
+      const allCrs: KubeCRD[] = [];
+      const newCrDictionary = new Map<string, KubeCRD>();
+
+      for (const crd of crds) {
+        const crClass = crd.makeCRClass();
+        const [crItems, crError] = await new Promise<[KubeCRD[] | null, any | null]>(resolve => {
```
sniok, line 30: these requests are made one after another, `Promise.all` would allow doing them concurrently. Or maybe even using react-query?

joaquimrocha:
@sniok I have done an attempt on adding react-query but I am not sure this is performing more fetches than needed (maybe triggering an infinite loop in some cases). Can you double check?

PR #2367, opened 2024-10-01
frontend: Update vitest and related, move vitest to devDependencies so it's not needed for prod only builds

illume:
- Move vitest to a dev dependency, add vitest.config.ts. Moving the vitest config to a separate file from the vite.config.ts so that prod only builds can work without vitest being installed.
- Upgrade vitest to 2.1.1
- Upgrade vitest-websocket-mock to 0.4.0
- Upgrade @vitest/coverage-istanbul to 2.1.1

### release notes

- [vitest releases](https://github.com/vitest-dev/vitest/releases)
- [vitest-websocket-mock releases](https://github.com/akiomik/vitest-websocket-mock/releases)
- [coverage-istanbul commits](https://github.com/vitest-dev/vitest/commits/main/packages/coverage-istanbul)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -119,7 +119,7 @@
         "@testing-library/user-event": "^14.5.2",
         "@types/nock": "^11.1.0",
         "@types/redux-mock-store": "^1.0.4",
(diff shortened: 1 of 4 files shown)


sniok (approved):
nice

PR #2383, opened 2024-10-03
frontend: Handle createRouteURL gracefully

ashu8912:


joaquimrocha:
Can you tell me what use cases this helps? Seems like it would silently fail and move to the home?

PR #2390, opened 2024-10-07
chart: Add change ClusterRole option

knrt10:
This adds custom configuration of clusterRole. User can now modify the settings by changing the default value.


illume (approved):
👍 thanks

PR #2402, opened 2024-10-07
frontend: headlamp-plugin: Update xterm 5.5.0 and related deps

illume:
Move to new package names:
@xterm/xterm @xterm/addon-search @xterm/addon-fit

https://github.com/xtermjs/xterm.js/releases

Also fixes a bug with the storybook.test runner where timers were not being cleared, as they are required. Causing an error to be triggered with one of the components that uses xterm related to the timers not being faked properly.

### Testing done

- smoke test with the log viewer


skoeva (approved):
Nice catching the storybook timer bug

Issue #2404, opened 2024-10-07
Broken URL for settings from unreachable clusters


joaquimrocha:
If you have a cluster that is not reachable and you see the "Failed to Connect" dialog, the cluster settings button is not working from there:
![Screenshot for the dialog](https://github.com/user-attachments/assets/a89914ad-b02e-4d58-8201-48901d1a95e8)

Probably a regression from unifying the cluster/app settings.

PR #2405, opened 2024-10-07
frontend: AuthChooser: Fix broken cluster settings URL

skoeva:
This change addresses a regression from unifying the general and cluster settings. Now, when trying to access a missing cluster, clicking on the "Cluster Settings" link should open the cluster settings in the home context.

Fixes: #2404 

### Testing
- [x] Start Headlamp with two clusters running
- [x] Stop one of the clusters and then try to access it in Headlamp
- [x] When it fails to open, click the "Cluster Settings" link and ensure the cluster settings open

image


sniok (approved):
nice, thanks!

PR #2406, opened 2024-10-08
backend: Return after attempt to serve a path using ..

joaquimrocha:
For security reasons we don't allow serving paths with .. in them, so we were setting an error but not returning.


illume (approved):
👍 thanks!

illume:
Looks like the tests are failing...

skoeva:
Test failures in `headlamp_test.go`: `TestSpaHandlerMissing`, `TestSpaHandlerBaseURL`, and `TestSpaHandlerOtherFiles`

PR #2410, opened 2024-10-09
Dockerfile: Use npm ci so lockfile is used

illume:
Npm install uses package.json, and npm ci uses package-lock.json.
Fixes a security issue where packages were not pinned.


skoeva (approved):
LGTM

PR #2415, opened 2024-10-14
frontend: Update vite to 5.4.9

illume:


skoeva (approved):
Thanks for this update

PR #2422, opened 2024-10-14
frontend: util: Add flattenClusterListItems for cluster objects

illume:
Sometimes we need to convert the objects per cluster into a flat list
of objects.

This is for multi cluster, it's not used yet.

### DIFF
--- a/frontend/src/lib/util.ts
+++ b/frontend/src/lib/util.ts
@@ -183,6 +183,22 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
   return [error, setError as any];
 }
 
+/**
+ * This function joins a list of items per cluster into a single list of items.
+ *
+ * @param args The list of objects per cluster to join.
+ * @returns The joined list of items, or null if there are no items.
+ */
+export function flattenClusterListItems<T>(
+  ...args: ({ [cluster: string]: T[] | null } | null)[]
+): T[] | null {
+  const flatItems = args
+    .filter(Boolean)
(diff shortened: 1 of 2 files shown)


frontend/src/lib/util.ts
```diff
@@ -183,6 +183,39 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
...
+/**
+ * This function joins a list of items per cluster into a single list of items.
+ *
+ * @param args The list of objects per cluster to join.
+ * @returns The joined list of items, or null if there are no items.
+ */
+export function flattenClusterListItems<T = any>(
+  ...args: ({ [cluster: string]: T[] | null } | null)[]
+): T[] | null {
+  const flatItems: T[] = [];
```
sniok, line 195: I think with it could be a bit simplified using flatMap

```
args.filter(Boolean).flatMap(it => Object.values(it)).filter(Boolean)
```

frontend/src/lib/util.ts
```diff
@@ -183,6 +183,39 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
...
+  let hasItems = false;
+  for (const clusterItems of Object.values(args)) {
+    if (clusterItems === null) {
+      continue;
+    }
+    for (const items of Object.values(clusterItems)) {
+      if (items === null || items.length === 0) {
+        continue;
+      }
+      console.log('meow', items);
```
sniok, line 206: 🐈

frontend/src/lib/util.ts
illume: Thanks. Fixed.

sniok (approved):

PR #2423, opened 2024-10-14
frontend: util: Add combineClusterListErrors

illume:
For errors, we may need to combine several cluster->errors
objects into a single object.

This is in preparation for multi cluster.

### DIFF
--- a/frontend/src/lib/util.ts
+++ b/frontend/src/lib/util.ts
@@ -1,4 +1,5 @@
 import humanizeDuration from 'humanize-duration';
+import merge from 'lodash/merge';
 import React from 'react';
 import { useHistory } from 'react-router';
 import { filterGeneric, filterResource } from '../redux/filterSlice';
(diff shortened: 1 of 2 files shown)


frontend/src/lib/util.ts
```diff
@@ -183,6 +183,36 @@ export function useErrorState(dependentSetter?: (...args: any) => void) {
   return [error, setError as any];
 }
 
+/**
+ * Combines errors per cluster.
+ *
+ * @param args The list of errors per cluster to join.
+ * @returns The joint list of errors, or null if there are no errors.
+ */
+export function combineClusterListErrors(
```
sniok, line 192: it seems like this function acts very similar to lodash's merge, can it be used instead?

illume: I used merge to reduce it a little bit. But because null handling is different it's not a perfect match.

sniok (approved):

illume:
(rebased against main to fix conflicts)

illume:
The Build Container CI failure is an unrelated problem with github/ubuntu.

PR #2424, opened 2024-10-15
frontend: Add ClusterGroupErrorMessage for displaying clusters with errors

illume:
This is for displaying clusters with errors.

image

Look for ClusterGroupErrorMessage in the storybook.
`npm run storybook`

### DIFF
--- a/frontend/src/components/cluster/ClusterGroupErrorMessage.tsx
+++ b/frontend/src/components/cluster/ClusterGroupErrorMessage.tsx
@@ -1,0 +1,63 @@
+import { InlineIcon } from '@iconify/react';
+import { Box, Typography, useTheme } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { ApiError } from '../../lib/k8s/apiProxy';
+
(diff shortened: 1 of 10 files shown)


frontend/src/components/cluster/ClusterGroupErrorMessage.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return cleanedErrors;
+}
+
+export function ClusterGroupErrorMessage({
+  clusterErrors: clusters,
+  message,
+}: ClusterGroupErrorMessageProps) {
+  const { t } = useTranslation();
+  const theme = useTheme();
+  const clusterObj = cleanNullErrors(typeof clusters === 'object' ? clusters : {});
```
sniok, line 42: type for `clusters`(`clusterErrors`) is an object or an array, but I think in case of array the `cleanNullErrors` won't properly work, so maybe restricting the type to only an object would be enough

illume: Thanks. Fixed.

sniok (approved):

PR #2427, opened 2024-10-15
ci: Use Ubuntu 22 for build-container workflow

sniok:
ubuntu-latest now resolves to 24 
playwright install fails on 24
so in this PR pinning the version to 22


skoeva (approved):
LGTM

Issue #2429, opened 2024-10-15
Restarting deployments and statefulsets fail without create permission


jda258:
When restarting a deployment or statefulset, a user only needs update or patch permission. The apply function tries a POST first, which does a create operation and fails if users don't have create permission. The put should be tried if a 403 error is returned.

PR #2430, opened 2024-10-15
frontend: allow apply with update permissions

jda258:
Fixes issue #2429.


joaquimrocha (approved):

joaquimrocha (changes requested):
Thanks @jda258 ! The linter is complaining about the format. Can you run `make frontend-fixlint` from the root source folder?
Since you need to change these, can you also update the `// We had a conflict. Try a PUT` like comment to better reflect things? Like `// We had a conflict or cannot create. Try a PUT in case the resource already exists.`

jda258:
> Thanks @jda258 ! The linter is complaining about the format. Can you run `make frontend-fixlint` from the root source folder? Since you need to change these, can you also update the `// We had a conflict. Try a PUT` like comment to better reflect things? Like `// We had a conflict or cannot create. Try a PUT in case the resource already exists.`

Thanks @joaquimrocha - I've pushed the changes that hopefully resolve those two items.

PR #2431, opened 2024-10-16
frontend: Resource/CreateButton: Add cluster selector

illume:
So users can select which cluster to apply the resource to.

Only shows the cluster selector if there is more than one
cluster being viewed.

For single cluster mode there is no change.

~~Note: depends on this being merged:~~
- ~~https://github.com/headlamp-k8s/headlamp/pull/2428~~

### DIFF
--- a/frontend/src/components/common/Resource/CreateButton.tsx
+++ b/frontend/src/components/common/Resource/CreateButton.tsx
@@ -1,12 +1,16 @@
 import { InlineIcon } from '@iconify/react';
 import Button from '@mui/material/Button';
+import FormControl from '@mui/material/FormControl';
+import InputLabel from '@mui/material/InputLabel';
+import MenuItem from '@mui/material/MenuItem';
+import Select from '@mui/material/Select';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useLocation } from 'react-router-dom';
-import { getCluster } from '../../../lib/cluster';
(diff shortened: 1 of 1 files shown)


sniok (approved):

joaquimrocha (changes requested):
I think the cluster selector makes sense to be in the extraActions you created the other PR for, right?
As it's more discoverable for the user? Also, for now we may keep it single cluster selection, but it's be pretty cool to be able to select either the full cluster group or multiple individual clusters, so we could batch applies. But this can be added later.

illume:
> I think the cluster selector makes sense to be in the extraActions you created the other PR for, right?

Yeah [https://github.com/headlamp-k8s/headlamp/pull/2428/files#diff-bd17d08e06e757d92d123c96d6fe154071e41f2dd53bfd1a096743023d66e8d2R67
](in this "frontend: EditorDialog: Add optional actions property" PR), the `actions` prop was added which is used in this PR.

> As it's more discoverable for the user? Also, for now we may keep it single cluster selection, but it's be pretty cool to be able to select either the full cluster group or multiple individual clusters, so we could batch applies. But this can be added later.

Ok, I added a task here for that: https://github.com/headlamp-k8s/headlamp/issues/1699#issuecomment-2416239082

illume:
(rebased against main now that the #2428 is merged)

joaquimrocha:
I did misunderstand where the actions in this PR would end up. Clarified in DM. All good.

joaquimrocha (approved):

PR #2434, opened 2024-10-16
frontend: Add cluster column to List views

illume:
These are not shown for single clusters.


joaquimrocha (commented):
Should we have the cluster column right after the namespace? Right now, normally, the namespace col is after the resource name. So it reads like `resource R in namespace N`; having the cluster after the namespace would read like `resource R in namespace N in cluster C`.

PR #2435, opened 2024-10-16
frontend: PodList: Make multi cluster aware

illume:


frontend/src/components/pod/List.tsx
```diff
@@ -207,13 +216,15 @@ export function PodListRenderer(props: PodListProps) {
       ]}
       data={pods}
       reflectInURL={reflectTableInURL}
+      clusterErrors={clusterErrors}
```
sniok, line 219: in what PR `clusterErrors` prop is implemented for ResourceListView?

illume: It was implemented for ResourceTable in this PR (ResourceListView extends ResourceTable so gets the prop from there): 
- https://github.com/headlamp-k8s/headlamp/pull/2436

sniok (approved):

PR #2437, opened 2024-10-16
frontend: RoleList: Make multi cluster aware

illume:
In single cluster mode the RoleList is the same.

The ColumnType is exported needed here to keep the typechecker happy.

### How to test
- [x] this fix needs to be merged first https://github.com/headlamp-k8s/headlamp/pull/2479
- [x] check Role list is the same as before (without a cluster column)
- [x] Add extra cluster to the URL and see Role List has a cluster column. Example of adding an extra cluster: `minikube+minikube2`

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -76,7 +76,7 @@ export type ResourceTableColumn<RowItem> = {
     }
 );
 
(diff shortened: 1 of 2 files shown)


joaquimrocha (approved):
The ResourceTable change could be a different commit.

illume:
(Rebased against main, and moved the ResourceTable change into a separate commit.)

PR #2443, opened 2024-10-17
frontend: cluster: Add useListPerCluster

illume:
List resource items or errors for each cluster being viewed.


joaquimrocha (commented):
I think this should use useKubeObjectList and it could receive a new param/arg to get whether it should return the results per cluster.
Or maybe it just returns another item in the results, grouped by cluster.

illume:
Closing in favour of https://github.com/headlamp-k8s/headlamp/pull/2448

PR #2444, opened 2024-10-17
frontend: Add cluster getter to KubeObject class

joaquimrocha:
Because we are accessing it from the ResourceTable and we have that data but didn't have a way to get it.
See the ResourceTable changes at https://github.com/headlamp-k8s/headlamp/commit/1e78de8ca6df0f9d6fde03743712f4aa323a5cf1 .


illume (approved):
👍 thanks

PR #2448, opened 2024-10-18
frontend: cluster: hooks: Make useList multi cluster aware

illume:
Adds these keyed by cluster name to the return of useList():
- `clusterResults`, valued by result of useList on a single one.
- `clusterErrors`, valued by the error if any, otherwise null.

Adds the cluster field for the results from each cluster. So that the tables can display the cluster if needed.

The new useList is backwards compatible and works with one cluster too.

People can now also do this to get the results for cluster groups:
```ts
{ clusterErrors, clusterResults } = useList()
const error = clusterErrors[cluster]
clusterResults[cluster].items[0].cluster // this cluster field is now in each row
```

Another example of usage:
```ts
const { 
  items, // in multiple clusters mode items has .cluster field.
  error, // in single cluster mode this is a single error or null. In multi cluster it's the first error.
  clusterErrors // this is null if no errors, or a Record keyed by cluster name, with a value of the error.
} = resourceClass.useList();
``
```

### Testing

Note: for single clusters it uses the same function as before.

- [ ] in single mode the cluster list views work
- [ ] in multi mode the cluster list views show results from multiple clusters and there is a 'cluster' column in the results.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.ts
+++ b/frontend/src/lib/k8s/api/v2/hooks.ts
@@ -55,6 +55,14 @@ export interface QueryResponse<DataType, ErrorType> {
 export interface QueryListResponse<DataType, ItemType, ErrorType>
   extends QueryResponse<DataType, ErrorType> {
   items: Array<ItemType> | null;
+  /**
+   * Results from individual clusters. Keyed by cluster name.
+   */
+  clusterResults?: Record<string, QueryListResponse<DataType, ItemType, ErrorType>>;
+  /**
+   * Errors from individual clusters. Keyed by cluster name.
+   */
+  clusterErrors?: Record<string, ApiError | null> | null;
 }
 
 /**
@@ -189,8 +197,10 @@ const useEndpoints = (endpoints: KubeObjectEndpoint[]) => {
 
 /**
  * Returns a list of Kubernetes objects and watches for changes
+ *
+ * @private please use useKubeObjectList.
  */
-export function useKubeObjectList<T extends KubeObjectClass>({
+function _useKubeObjectList<T extends KubeObjectClass>({
   kubeObjectClass,
   namespace,
   cluster: maybeCluster,
@@ -231,9 +241,11 @@ export function useKubeObjectList<T extends KubeObjectClass>({
           cluster,
         }
       ).then(it => it.json());
-      list.items = list.items.map(
-        item => new kubeObjectClass({ ...item, kind: list.kind.replace('List', '') })
-      );
+      list.items = list.items.map(item => {
+        const itm = new kubeObjectClass({ ...item, kind: list.kind.replace('List', '') });
+        itm.cluster = cluster;
+        return itm;
+      });
 
       return list;
     },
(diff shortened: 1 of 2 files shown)


frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -275,3 +283,77 @@ export function useKubeObjectList<T extends KubeObjectClass>({
...
+  kubeObjectClass: T;
+  /** Object list namespace */
+  namespace?: string;
+  /** Object list clusters */
+  clusters: string[];
+  queryParams?: QueryParameters;
+}): [Array<InstanceType<T>> | null, ApiError | null] &
+  QueryListResponse<KubeList<InstanceType<T>>, InstanceType<T>, ApiError> {
+  const clusterResults = clusters.reduce((acc, cluster) => {
+    acc[cluster] = useKubeObjectList<T>({
```
joaquimrocha, line 306: I wonder if we will have problems with this given we are not using the hook in the function body. But we do use some like this already anyway...

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -275,3 +283,77 @@ export function useKubeObjectList<T extends KubeObjectClass>({
...
+    error,
+    isError,
+    isLoading,
+    isFetching,
+    isSuccess,
+    status,
+    *[Symbol.iterator](): ArrayIterator<ApiError | InstanceType<T>[] | null> {
+      yield items;
+      yield error;
+    },
```
joaquimrocha, line 355: I think this is only for the return to work also as a list. Since there's no backward compat need for this function, we can remove this.

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -275,3 +283,77 @@ export function useKubeObjectList<T extends KubeObjectClass>({
     },
   };
 }
+
+/**
+ * Returns a combined list of Kubernetes objects and watches for changes from the clusters given.
+ */
+export function useKubeObjectLists<T extends KubeObjectClass>({
```
joaquimrocha, line 290: I had actually thought that we could adapt the useKubeObjectList and pass a new `clusters` arg. But then we'd have to use `useQueries` if we want to maintain a cache of each cluster's results independently and maybe that adds complexity. So maybe we can use this new function.
I'd like @sniok 's opinion here.

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -55,6 +55,14 @@ export interface QueryResponse<DataType, ErrorType> {
...
   extends QueryResponse<DataType, ErrorType> {
   items: Array<ItemType> | null;
+  /**
+   * Results from individual clusters. Keyed by cluster name.
+   */
+  clusterResults?: Record<string, QueryListResponse<DataType, ItemType, ErrorType>>;
+  /**
+   * Errors from individual clusters. Keyed by cluster name.
+   */
+  clusterErrors?: Record<string, ApiError | null> | null;
```
joaquimrocha, line 65: Maybe we could extend the previous interface, rather than adding more members to this one? Since it's a new function, having a new interface doesn't seem inadequate.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: plugins may need backward compatibility

frontend/src/lib/k8s/api/v2/hooks.ts
illume: I wrapped it in useMemo to ensure consistent hook order and such when clusters change.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: I removed useKubeObjectLists, and now just useKubeObjectList is exported. The multi cluster version is internal.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: I removed useKubeObjectLists, and now just useKubeObjectList is exported. The multi cluster version is internal.

illume:
I removed useKubeObjectLists, and now just useKubeObjectList is exported. The multi cluster version is internal.

The tests are now passing in CI.

illume:
todo: There's a better way to add the cluster field.

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -55,6 +55,14 @@ export interface QueryResponse<DataType, ErrorType> {
 export interface QueryListResponse<DataType, ItemType, ErrorType>
   extends QueryResponse<DataType, ErrorType> {
   items: Array<ItemType> | null;
+  /**
+   * Results from individual clusters. Keyed by cluster name.
+   */
+  clusterResults?: Record<string, QueryListResponse<DataType, ItemType, ErrorType>>;
```
sniok, line 61: is clusterResults ever used? I'd think that items list with all the responses is enough? 

this is a bit hard to review without seeing other changes

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -275,3 +287,122 @@ export function useKubeObjectList<T extends KubeObjectClass>({
...
+  namespace?: string;
+  /** Object list clusters */
+  clusters: string[];
+  queryParams?: QueryParameters;
+}): [Array<InstanceType<T>> | null, ApiError | null] &
+  QueryListResponse<KubeList<InstanceType<T>>, InstanceType<T>, ApiError> {
+  const clusterResults: Record<string, ReturnType<typeof useKubeObjectList>> = {};
+
+  for (const cluster of clusters) {
+    clusterResults[cluster] = useKubeObjectList({
```
sniok, line 351: Having hooks in a loop is a bit sketchy (although we already have loops like that in just a couple of places where we also ensure that the parent component will remount) the useList is used pretty much everywhere. Ideally we'd use useQueries to fetch multiple queries at once but I think this is okay for now

illume: When clusters change top level components remount. Which means the order is fixed.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: > this is a bit hard to review without seeing other changes to the ListViews of resources

You can see clusterErrors used here in this PR from the other day:
https://github.com/headlamp-k8s/headlamp/pull/2436

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -231,9 +241,11 @@ export function useKubeObjectList<T extends KubeObjectClass>({
           cluster,
         }
       ).then(it => it.json());
-      list.items = list.items.map(
-        item => new kubeObjectClass({ ...item, kind: list.kind.replace('List', '') })
-      );
+      list.items = list.items.map(item => {
+        const itm = new kubeObjectClass({ ...item, kind: list.kind.replace('List', '') });
+        itm.cluster = cluster;
```
joaquimrocha, line 246: Why is the cluster not set in the constructor? Is this the reason you added a cluster setter? I am asking because it feels like the cluster should be a readonly in that class.

joaquimrocha (approved):
It LGTM even if we should keep in mind to change the logic to use `useQueries` later on.
I left a comment about the cluster setter, but if there's a good reason to keep that, please feel free to move on.

sniok (approved):

frontend/src/lib/k8s/api/v2/hooks.ts
illume: Yeah... I had to do this because the typing is broken for that code unfortunately. It doesn't know that there is a constructor that takes an extra argument and complains.

I guess we can look back at this once the other typing fixes are merged.

Issue #2450, opened 2024-10-18
storybook brokenness


illume:
## Describe the bug

List views are broken in storybook.

![image](https://github.com/user-attachments/assets/a01aaaee-bc8c-4b64-ae36-f4c19ac45318)

## To Reproduce

`npm start storybook`

Go to node list.

## Environment (please provide info about your environment):
 - Headlamp Version: main branch

## Are you able to fix this issue?
maybe

## Additional Context

Related to:
- https://github.com/headlamp-k8s/headlamp/issues/1282

With the switch to vite it seems some of these started failing.

illume:
More info from @skoeva about this here: https://github.com/headlamp-k8s/headlamp/pull/2451#issuecomment-2423087309

> - Loading the storybook home (http://localhost:6006) usually does not cause any components to fail. Even in main, once the home view opens, you can click around, and even the List test does not fail.
> - Loading a component in storybook (i.e. http://localhost:6006/?path=/story/resource-maininfosection--null-backlink) and refreshing on this page will eventually cause storybook to fail with similar messages to what you've posted, both in main and the node-story branch. This isn't unique to List though, I've attached some sample screenshots below.

sniok:
This error is cause by circular imports, temp workaround is to always start from the home

Issue #1282, opened 2023-07-31
Replace CRA (create-react-app)


joaquimrocha:
CRA seems to be inactive/unmaintained. So we should move to another solution in order to keep getting updates.
Maybe [Vite](https://vitejs.dev/) is the solution but not at all a drop-in replacement. So anything that doesn't require rewiring the whole app would be good.

Acceptance Criteria:
- [x] We have a chosen solution to replace CRA
- [x] Running Headlamp in dev mode is possible with the new solution
- [x] Building Headlamp for production is possible with the new solution

Open issues/PRs
- [x] https://github.com/headlamp-k8s/headlamp/issues/2125 (mocking of API calls in test runner still open)
- [x] https://github.com/headlamp-k8s/headlamp/pull/2001
- [x] https://github.com/headlamp-k8s/headlamp/issues/2466

illume:
- https://github.com/dilanx/craco is also not getting updated now.
- https://turbo.build/ used by next isn't ready to be used outside of next apparently.

sniok:
Honestly I think vite is the only real options that is both mature and performant. And vite's development experience is really great.
Migration should be straightforward, this project doesn't really use any complicated webpack features, all the plugins are simple js/css/ts/svg loaders that vite will handle without a problem.

sniok:
In addition we can also take advantage of [vitest](https://vitest.dev/), which is mostly compatible with jest. It uses the same resolution and bundling as vite (and the same viteconfig.ts file) so we can reduce amount of configuration and unify building and testing.

illume:
Opening because the second vite PR is still ongoing: https://github.com/headlamp-k8s/headlamp/pull/2001

illume:
Moved to next release milestone. The new headlamp plugin can be done in the next release cycle. The only item left in the coverage issue still open is about the mocks in the frontend/ tests.

sniok:
create react app is officially deprecated now 

https://react.dev/blog/2025/02/14/sunsetting-create-react-app

PR #2451, opened 2024-10-18
frontend: Remove broken node List.stories.tsx

illume:
This story is broken, so removing it for now.

- It is always failing in the storybook.
- It is intermittently failing for me locally in the test runner.

Related to 
- #2450
- #1282 

image


skoeva:
How were you able to pinpoint the issue to List? I tried to test storybook in both branches to break it, and this is what I found:
- Loading the storybook home (`http://localhost:6006`) usually does not cause any components to fail. Even in main, once the home view opens, you can click around, and even the List test does not fail.
- Loading a component in storybook (i.e. `http://localhost:6006/?path=/story/resource-maininfosection--null-backlink`) MAY cause storybook to fail with similar messages to what you've posted. This isn't unique to List though, I've attached some sample screenshots below.

Is this something we can look to change? Or is this the nature of how storybook works? (as in, storybook will basically be unusable if a single story isn't able to render)

image
image

illume:
@skoeva thanks for sharing that. I added your notes into here:
- https://github.com/headlamp-k8s/headlamp/issues/2450

> How were you able to pinpoint the issue to List?

For me it's just this one thing that intermittently fails locally with the test runner. I can see other ones failing like you mentioned.

> Is this something we can look to change? Or is this the nature of how storybook works? (as in, storybook will basically be unusable if a single story isn't able to render)

Yeah we should fix it. They should be working when one component fails. I haven't bisected it yet... but some months ago those stories were working, so it's something since then.

illume:
Since now that one isn't failing for me either.

Still have this issue open for the storybook issues: https://github.com/headlamp-k8s/headlamp/issues/2450

PR #2452, opened 2024-10-18
frontend: authchooser: Fix testAuth function to include a cluster

illume:
So we may check any clusters directly instead of the currently
viewed one.


skoeva (approved):
Looks good

PR #2453, opened 2024-10-18
frontend: stateless: Fix error when clusters is null

illume:
There was no checking in the case where clusters is null.

This is the traceback in the dev console. You need to have no "stateless clusters" (which is the default).
image


skoeva (approved):
Tested in the UI and works. LGTM

knrt10 (approved):
Thanks

PR #2457, opened 2024-10-21
chart: support deploying additional kubernetes manifests

guydomb:
This change allows the Helm chart's user to include additional Kubernetes manifests to be deployed along with the chart. The incentive for this change was to allow overriding the default use of cluster-admin cluster role and also deploy a different cluster role and binding. 

This is achievable by including the additional manifests as an array of multiline strings in field `extraManifests`. For example:
```
# values.yaml
extraManifests:
  - |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: dummy-configmap
    data:
      key1: value1
      key2: value2
  - |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: configmap-with-templated-data
    data:
      injectedKey: {{ .Release.Name }}
```

### DIFF
--- a/charts/headlamp/templates/extra-manifests.yaml
+++ b/charts/headlamp/templates/extra-manifests.yaml
@@ -1,0 +1,6 @@
+{{- if .Values.extraManifests }}
+{{- range $manifest := .Values.extraManifests }}
+---
+{{- tpl $manifest $ | nindent 0 }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


illume:
@knrt10 implemented the helm chart testing, and might be able to give some pointers on what's going wrong with the test failure.

guydomb:
@illume @knrt10 Apologies for the noise, test is fixed and all checks are passing. Are we OK to proceed? Please let me know if anything's missing.

knrt10 (approved):
+1. Thank you

Issue #1802, opened 2024-03-13
multi-plexing single browser websocket to multiple backend websockets


illume:
In https://github.com/headlamp-k8s/headlamp/issues/1373 we investigated websocket pool max limit that browsers have. Which we quickly encounter.

One solution to this problem is to have a backend service which takes one websocket from the browser and makes multiple websocket connections via the backend to kubernetes API server.

We already use the backend server to proxy K8s API requests.

## Architecture diagram:

```mermaid
graph LR
    A[Browser] <-->|WebSocket| B[headlamp_server]
    B <-->|WebSocket| C[Kubernetes API Server]
    B <-->|WebSocket| D[Kubernetes API Server]
    B <-->|WebSocket| E[Kubernetes API Server]
```

## Some related websocket links to multi plexing and proxying

- https://dev.to/hgsgtk/reverse-http-proxy-over-websocket-in-go-part-1-13n4
- https://medium.com/@prithuadhikary/go-api-gateway-using-gos-reverse-proxy-21f3fe7b88e3
- https://ckousik.github.io/gsoc/2017/06/16/WebSocket-Multiplexer-Overview.html
- https://dev.to/justlorain/how-to-apply-reverse-proxy-over-websocket-27ml

knrt10:
So far, my findings on the issue have been this. 

1. **Create a Function to Open a New Websocket Connection to the Cluster from the Backend:**
 We can use the `gorilla/websocket` package to handle websocket connections. We can create a function like `openWebsocketConnectionToCluster` that utilizes this package to establish a websocket connection to the Kubernetes cluster. This function would typically involve upgrading an HTTP connection to a websocket connection using the `Upgrade` method provided by the `gorilla/websocket` package.

   ```go
   import "github.com/gorilla/websocket"

   func openWebsocketConnectionToCluster(clusterURL string) (*websocket.Conn, error) {
       // Dial websocket connection to the cluster URL
       conn, _, err := websocket.DefaultDialer.Dial(clusterURL, nil)
       if err != nil {
           return nil, err
       }
       return conn, nil
   }
   ```

2. **Create a Map to Store Active Websocket Connections:**
   We can use a map to store active websocket connections, where the keys are the URLs or anything unique like userID + clustername + url of the clusters and the values are the websocket connections.

   ```go
   var activeConnections = make(map[string]*websocket.Conn)
   ```

3. **Create a New Websocket Endpoint in the Backend (/websocket):**
[...]

PR #2459, opened 2024-10-22
backend: Add websocket multiplexer

knrt10:
This adds websocket multiplexer to the backend. Frontend now need to make a single websocket call to the backend. Once that connection is established with the backend it will send message to the backend with appropriate data. Backend will open multiple websockets and act as a proxy to the frontend. It will make request to k8s server and return data to frontend.

This also adds retry logic if the connection is broken between frontend and backend.

Fixes: #1802

## Testing

- Run backend `cd backend && go run ./cmd -dev -enable-dynamic-clusters`
- Run frontend `cd frontend && npm start` (we need this for userID of the session).
- Open postman application and use the API [here](https://www.postman.com/knrt1017/workspace/headlamp/ws-raw-request/6717608832943442e3cc99a7)
- It has message payload which you need to edit 
- 

```json
{
  "clusterID": "minikube", // update it with the cluster you have running
  "path": "/api/v1/pods",
  "query": "watch=true",
  "userId": "9872e4f36eaadb0e" // go to localstorage and copy `headlamp-userID` and paste here
}
```

- Click on the `Connect` button
Screenshot 2024-10-22 at 2 08 34 PM

- Once connected hit the send button.
Screenshot 2024-10-22 at 2 09 41 PM

- You will see response from your k8s cluster

Screenshot 2024-10-24 at 12 42 28 PM

- Create an nginx pod `kubectl run nginx --image=nginx` and try other things like logs and exec

#### Logs

```json
{
  "clusterID": "minikube",
  "path": "/api/v1/namespaces/default/pods/nginx/log",
  "query": "follow=true&tailLines=100",
  "userId": "9872e4f36eaadb0e"
}
```

Screenshot 2024-10-24 at 12 45 18 PM

#### Exec

```json
{
  "clusterID": "minikube",
  "path": "/api/v1/namespaces/default/pods/nginx/exec",
  "query": "command=ls&command=-l&stdout=true&stderr=true",
  "userId": "9872e4f36eaadb0e"
}
```

Screenshot 2024-10-24 at 12 43 45 PM

**The data is sent as binary and it is encoded to bas64 to parse as JSON response.**

#### Close connection

```json
{
  "clusterID": "minikube",
  "path": "/api/v1/pods",
  "userId": "9872e4f36eaadb0e",
  "data": "close"
}
```

Screenshot 2024-10-24 at 12 40 07 PM

You can edit `path` as per your needs and check different output.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -59,6 +59,7 @@ type HeadlampConfig struct {
 	proxyURLs             []string
 	cache                 cache.Cache[interface{}]
 	kubeConfigStore       kubeconfig.ContextStore
+	multiplexer           *Multiplexer
 }
 
 const DrainNodeCacheTTL = 20 // seconds
@@ -1599,6 +1600,9 @@ func (c *HeadlampConfig) addClusterSetupRoute(r *mux.Router) {
 
 	// Rename a cluster
 	r.HandleFunc("/cluster/{name}", c.renameCluster).Methods("PUT")
+
+	// Websocket connections
+	r.HandleFunc("/wsMutliplexer", c.multiplexer.HandleClientWebSocket)
 }
 
 /*
--- a/backend/cmd/multiplexer.go
+++ b/backend/cmd/multiplexer.go
@@ -1,0 +1,632 @@
+package main
+
+import (
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
+	"github.com/headlamp-k8s/headlamp/backend/pkg/logger"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	// StateConnecting is the state when the connection is being established.
+	StateConnecting ConnectionState = "connecting"
+	// StateConnected is the state when the connection is established.
+	StateConnected ConnectionState = "connected"
+	// StateError is the state when the connection has an error.
+	StateError ConnectionState = "error"
+	// StateClosed is the state when the connection is closed.
+	StateClosed ConnectionState = "closed"
+)
+
+const (
+	// HeartbeatInterval is the interval at which the multiplexer sends heartbeat messages to the client.
+	HeartbeatInterval = 30 * time.Second
+	// HandshakeTimeout is the timeout for the handshake with the client.
+	HandshakeTimeout = 45 * time.Second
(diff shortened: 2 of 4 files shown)


sniok:
Nice, tested it and it seems to work

There are a couple of crucial things missing though:

- The data coming in doesn't contain information about which request it is 
- There's no way to close the request (e.g. stop watching `/pods`)

knrt10:
Thank you for the feedback @sniok. Awesome suggestions. I did forget to add closing the connection on request. Will add 

Regarding attaching, I did not try it. Will check why it’s not working.

illume:
I think we can delay any work on attaching to the container if there is a problem. Because we mainly want this to help with watching.

illume:
> Backend Code coverage changed from 61.6% to 61.6%. Change: 0% 😃.

Hrmm. Not sure why this is at 0, since there are tests... lol.

knrt10:
@sniok @illume updated the `code + PR description + postman calls`. You can now do all k8s operation. Please try and let me know. Also added the state handling of webhook

sniok:
Thanks for the changes.

- I'm not sure why "type" is required? Isn't path enough to uniquely identify the operation? 

- The messages still don't have information about which request they belong to.

image

Imagine we started watching two /pods from two different clusters (or two different ReplicaSets,etc) and we need to figure out where to put this update that we received from websocket. Currently we just get a Pod without any info where it came from

- Also I don't see how to close the connection?

knrt10:
@sniok thank you for the detailed message. I have updated the description, screenshot, postman API and code. You will now have detailed information from where request is coming from. `type` as mentioned was not needed so has been removed. Not you can also close connection. Added usage in the description. Thanks

illume:
Thanks @knrt10 

(It looks like close can be used, and now a clusterID can be used to identify the response.)

Does it make sense to start implementing the frontend/ integration now? Will you do that?

Note I started an issue here about multi cluster e2e tests here:
- https://github.com/headlamp-k8s/headlamp/issues/2460

illume:
Thanks for the explanation @knrt10 

So the response returns (clusterID, path, query, and userID) as well as the data. This can be used to know from which websocket it comes from.

We discussed the possibility that multiple ones could be made with the same of those... in this case it will not work. I don't think we need to cover this now though. It could be solved in the future by adding a separate unique ID for each one. Again, I don't think this is necessary for our use cases... we can use (clusterID, path, query, and userID) as the unique key.

illume:
If @sniok you don't have any more comments or concerns, then we can merge this?

@knrt10 will begin frontend integration probably tomorrow, but we can merge this in first.  (please correct me if I'm wrong)

sniok (approved):

sniok:
looks good, thanks!

Issue #2382, opened 2024-10-02
Bug with duplicated notifications


sniok:
### Steps to reproduce
1. go to any Deployment
2. Scale it 
3. Restart (right away)

### Result
3 notifications are shown

### Expected 
2 notifications, one for scaling, one for restart

### Demo

https://github.com/user-attachments/assets/65f6019f-539b-4e4c-866a-bc7cf5cdd50b

same behavior on 0.25.1 and main

PR #2462, opened 2024-10-22
frontend: Add state for notification handling

skoeva:
This change introduces a state to track cluster actions and improve notification handling. This prevents duplicate notifications when performing scale/restart events to deployments in quick succession.

Fixes: #2382 

### Testing
- [X] Go to any deployment in Headlamp
- [X] Scale the deployment, and then immediately restart it
- [X] Ensure that only one notification gets displayed for each event in progress and each completed event

### DIFF
--- a/frontend/src/components/common/ActionsNotifier.tsx
+++ b/frontend/src/components/common/ActionsNotifier.tsx
@@ -1,7 +1,7 @@
 import Button from '@mui/material/Button';
 import _ from 'lodash';
 import { useSnackbar } from 'notistack';
-import React from 'react';
+import React, { useRef } from 'react';
 import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { CLUSTER_ACTION_GRACE_PERIOD, ClusterAction } from '../../redux/clusterActionSlice';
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/ActionsNotifier.tsx
```diff
@@ -49,13 +50,28 @@ function PureActionsNotifier({ dispatch, clusterActions }: PureActionsNotifierPr
...
+
     if (clusterAction.message) {
-      enqueueSnackbar(clusterAction.message, {
-        key: clusterAction.key,
-        autoHideDuration: clusterAction.autoHideDuration || CLUSTER_ACTION_GRACE_PERIOD,
-        action,
-        ...clusterAction.snackbarProps,
-      });
+      // Check for completed actions
+      const refKey = !clusterAction.message.includes('ing')
```
sniok, line 62: the `.includes('ing')` seems way too hacky, is there other way to achieve this?

skoeva: It covers all the current cases, but you are right that it is not fool proof. The alternative would be to add a new state to ClusterAction to keep track of what stage it is in since this is currently not implemented, can make this change though

sniok (approved):
nice

illume (approved):
🎉🎈

PR #2463, opened 2024-10-22
WIP CodeQL test

sniok:


sniok:
okay so problem is somewhere in cluster.ts and KubeObject.ts

PR #2465, opened 2024-10-23
backend: Prevent sending refresh signal before the app is ready

### LINKED ISSUE #2182: app reloads when starting
When starting the app, I see it load, and then reload about 2 seconds after. It's much quicker when not in developer mode, but it does still reload.

I tried this with current main, and also a previous release 0.24.1. I'm not sure how long it's been happening, and haven't bisected back further.

It's not such a problem because in release mode it's very quick and barely noticeable. In dev mode it's just slightly annoying (to me anyway).

joaquimrocha:
The backend was sending a refresh signal to the frontend very early due to reading the plugins folders, and this resulted in a refresh right after the UI was loaded (desktop only).

It is only interesting for the app to refresh after a plugin change if it already has the list of plugins, so these changes prevent that refresh until the app requests the plugins list.

fixes #2182

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -303,6 +303,12 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 		}
 		if err := json.NewEncoder(w).Encode(pluginsList); err != nil {
 			logger.Log(logger.LevelError, nil, err, "encoding plugins base paths list")
+		} else {
+			// Notify that the client has requested the plugins list. So we can start sending
+			// refresh requests.
+			if err := config.cache.Set(context.Background(), plugins.PluginCanSendRefreshKey, true); err != nil {
+				logger.Log(logger.LevelError, nil, err, "setting plugin-can-send-refresh key")
+			}
 		}
 	}).Methods("GET")
 
--- a/backend/pkg/plugins/plugins.go
+++ b/backend/pkg/plugins/plugins.go
@@ -3,6 +3,7 @@ package plugins
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/fs"
 	"net/http"
@@ -19,9 +20,10 @@ import (
 )
 
 const (
-	PluginRefreshKey       = "PLUGIN_REFRESH"
-	PluginListKey          = "PLUGIN_LIST"
-	subFolderWatchInterval = 5 * time.Second
+	PluginRefreshKey        = "PLUGIN_REFRESH"
+	PluginListKey           = "PLUGIN_LIST"
+	PluginCanSendRefreshKey = "PLUGIN_CAN_SEND_REFRESH"
+	subFolderWatchInterval  = 5 * time.Second
 )
 
 // Watch watches the given path for changes and sends the events to the notify channel.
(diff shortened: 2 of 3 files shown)


backend/cmd/headlamp.go
```diff
@@ -303,6 +303,11 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 		}
 		if err := json.NewEncoder(w).Encode(pluginsList); err != nil {
 			logger.Log(logger.LevelError, nil, err, "encoding plugins base paths list")
+		} else {
+			// Notify that the client has requested the plugins list. So from
+			if err := config.cache.Set(context.Background(), plugins.PluginCanSendRefreshKey, true); err != nil {
+				logger.Log(logger.LevelError, nil, err, "setting pplugin-can-send-refresh key")
```
skoeva, line 309:
```suggestion
				logger.Log(logger.LevelError, nil, err, "setting plugin-can-send-refresh key")
```

backend/pkg/plugins/plugins.go
```diff
@@ -214,14 +216,42 @@ func pluginBasePathListForDir(pluginDir string, baseURL string) ([]string, error
...
 // and plugin refresh key in the cache.
 func HandlePluginEvents(staticPluginDir, pluginDir string,
 	notify <-chan string, cache cache.Cache[interface{}],
 ) {
 	for range notify {
-		// set the plugin refresh key to true
-		err := cache.Set(context.Background(), PluginRefreshKey, true)
+		// Do not set the refresh signal if we cannot send it. We prevent it here
+		// because we only want to send refresh signals that happen after we are
+		// allowed to send them).
```
skoeva, line 246:
```suggestion
		// allowed to send them.
```

backend/pkg/plugins/plugins_test.go
```diff
@@ -370,10 +400,29 @@ func TestHandlePluginReload(t *testing.T) {
 	err := ch.Set(context.Background(), plugins.PluginRefreshKey, true)
 	require.NoError(t, err)
 
+	// err := ch.Set(context.Background(), plugins.PluginRefreshKey, false)
+	// require.NoError(t, err)
```
skoeva, line 404: Do we want to keep this portion commented out?

joaquimrocha: Good catch.

joaquimrocha:
Thanks for the review, @skoeva . I have updated the branch.

backend/pkg/plugins/plugins.go
```diff
@@ -214,14 +216,42 @@ func pluginBasePathListForDir(pluginDir string, baseURL string) ([]string, error
...
 	notify <-chan string, cache cache.Cache[interface{}],
 ) {
 	for range notify {
-		// set the plugin refresh key to true
-		err := cache.Set(context.Background(), PluginRefreshKey, true)
+		// Do not set the refresh signal if we cannot send it. We prevent it here
+		// because we only want to send refresh signals that happen after we are
+		// allowed to send them.
+		var err error
+		if canSendRefresh(cache) {
```
sniok, line 248: err := cache.Set(context.Background(), PluginRefreshKey, canSendRefresh(cache))

joaquimrocha: I didn't do it like this because the comment was more clear in this version of the code. But I will update the code and do this.

illume (approved):
🎉 thanks

PR #2473, opened 2024-10-24
frontend: Add snapshot test for window.pluginLib

sniok:
This will make it easier to maintain compatibility with plugins by keeping track of what is in pluginLib object


joaquimrocha (approved):
Awesome! This allows us to better check for unintended backward compatibility breaks!

PR #2474, opened 2024-10-24
frontend: util filterFunc: Fix for when item is null

illume:
This was causing a crash sometimes.

### DIFF
--- a/frontend/src/lib/util.ts
+++ b/frontend/src/lib/util.ts
@@ -160,7 +160,7 @@ export function useFilterFunc<
   const filter = useTypedSelector(state => state.filter);
 
   return (item: T, search?: string) => {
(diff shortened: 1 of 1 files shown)


frontend/src/lib/util.ts
```diff
@@ -160,7 +160,7 @@ export function useFilterFunc<
   const filter = useTypedSelector(state => state.filter);
 
   return (item: T, search?: string) => {
-    if (!!item.metadata) {
+    if (item && item.metadata) {
```
sniok, line 163: `if(item?.metadata) {`
would be a bit simpler

illume: Thanks. Fixed.

skoeva (approved):
LGTM

PR #2479, opened 2024-10-25
frontend: k8s: hooks: Fix useKubeObjectList items could contain null

illume:
When I removed the ! at the end, the type checker complained and helped me find this bug.

If one cluster doesn't have results for something this could result in one null entering the items list.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.ts
+++ b/frontend/src/lib/k8s/api/v2/hooks.ts
@@ -356,7 +356,7 @@ function _useKubeObjectLists<K extends KubeObject>({
     if (items === null) {
       items = clusterResults[cluster].items;
     } else {
-      items = items.concat(clusterResults[cluster].items!);
+      items = items.concat(clusterResults[cluster].items ?? []);
     }
   }
 


frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -356,7 +356,9 @@ function _useKubeObjectLists<K extends KubeObject>({
     if (items === null) {
```
sniok, line 356: this block can be simplified a bit

```ts
const items = clusterResults.flatMap(result => result.items ?? [])
```

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -356,7 +356,9 @@ function _useKubeObjectLists<K extends KubeObject>({
...
       items = clusterResults[cluster].items;
     } else {
```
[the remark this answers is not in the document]
illume: Thanks. Fixed.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: No, that doesn't quite work. Because nullability of that data, and it also needs to iterate over clusters.

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -355,8 +355,8 @@ function _useKubeObjectLists<K extends KubeObject>({
   for (const cluster of clusters) {
     if (items === null) {
       items = clusterResults[cluster].items;
-    } else {
-      items = items.concat(clusterResults[cluster].items!);
+    } else if (clusterResults[cluster].items !== null) {
+      items = items.concat(clusterResults[cluster].items);
```
joaquimrocha, line 359: Maybe `items.concat(clusterResults[cluster].items ?? []);` would be more readable, but either way is fine.

joaquimrocha (approved):
Left a nit.

frontend/src/lib/k8s/api/v2/hooks.ts
illume: Thanks. Fixed.

PR #2484, opened 2024-10-28
Error boundary in routes

joaquimrocha:
This PR adds an error boundary to the router components + adds info about the error to the error page.

New error boundary
New error boundary with errors

## How to reproduce
1. Change the members in a table that a list view is rendering (for example, in the Job list) to access a non-existing (e.g. `item.somethingThatDoesntExist`);
2. Verify that the error page is displayed just in the main area of Headlamp and not the entire page

## Done
- [X] Error boundary for route
- [X] Error details in error page
- [x] Story for Storybook


frontend/src/components/common/ErrorPage/ErrorPage.tsx
```diff
@@ -63,6 +72,34 @@ export default function ErrorComponent(props: ErrorComponentProps) {
...
+                <Box textAlign="right">
+                  <Button
+                    onClick={() => {
+                      navigator.clipboard.writeText(error.stack!);
+                    }}
+                  >
+                    {t('translation|Copy')}
+                  </Button>
+                </Box>
+                <Typography variant="body2">{error.stack}</Typography>
```
sniok, line 97: Using a `pre` tag would render text a bit nicer
`<Typography variant="body2" component="pre" align="left" overflow="auto">`

image

sniok (approved):

frontend/src/components/common/ErrorPage/ErrorPage.tsx
joaquimrocha: Good point. I added the pre and alignment, but made it wrap instead of the overflow.

sniok (approved):

PR #2489, opened 2024-10-29
frontend Home: Refresh clusters versions/errors on interval

### LINKED ISSUE #2478: Headlamp app is not moving minikube from stopped to started on home
## Describe the bug

- minikube is stopped
- headlamp app is started and is on the home screen showing minikube is stopped
- minikube is started
- headlamp app doesn't show minikube as started (you have to reload to see it started)

## To Reproduce
as above

 - Installation type: WSL ubuntu 22
 - Headlamp Version: main branch with cd app; npm start

I'm not sure if it's a regression.

sniok:
Fixes #2478


skoeva (approved):
Tested by deleting a cluster (changed to Bad Gateway on home) and then by restarting the cluster (changed back to Active on home), looks like this works. Thanks!

PR #2490, opened 2024-10-29
headlamp-plugin: Fix extract/package to copy all in dist/

illume:
All files made in the dist/ folder are copied in.

So plugins can add extra files if they need in the build step.

For example:

```
  "scripts": {
    "build": "headlamp-plugin build && npx shx cp main.wasm dist/main.wasm",
```

Needed for:
- https://github.com/headlamp-k8s/plugins/issues/13

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -173,25 +173,12 @@ function extract(pluginPackagesPath, outputPlugins, logSteps = true) {
     fs.mkdirSync(outputPlugins);
   }
 
-  function copyFiles(plugName, inputMainJs, mainjs) {
-    if (!fs.existsSync(plugName)) {
-      if (logSteps) {
-        console.log(`Making output folder "${plugName}".`);
-      }
-      fs.mkdirSync(plugName);
-    }
-
-    if (logSteps) {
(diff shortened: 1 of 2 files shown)


illume:
cc: @ashu8912 this lets main.wasm work in ig plugin. It will need this merged, and a new headlamp-plugin release first though.

sniok:
nice changes!

Correct me if I'm wrong though but I think this is not enough for plugins to use other files in the `dist/`.
From what I understand, main.js won't be able to access any of those additional files because it doesn't know the path for them 

let's say I have a plugin

`some-plugin/main.js`
`some-plugin/somefile`

Plugin is loaded and executed in the headlamp like this 

```
code = fetch('http://headlamp-server/plugins/some-plugin/main.js');
eval(code)
```

then from inside `main.js` there's no way to access other files without knowing the full path

main.js
```
fetch('./somefile') // wrong path, resolves headlamp-client/c/default/somefile for example
fetch('dist/somefile') // wrong path, resolves headlamp-client/dist/somefile
fetch('http://headlamp-server/plugins/some-plugin/somefile') // this where the file is but plugin doesn't know this path
```

illume:
The IG plugin uses this code to fetch:
```
fetch("/plugins/headlamp-ig/main.wasm")
```

joaquimrocha (approved):
Great stuff!

PR #2492, opened 2024-10-30
frontend: Improve handling of an edgecase in useKubeObject useKubeObjectList hooks

joaquimrocha:
So we don't have a different key when a function uses the namespace as `undefined` and another as `''`.


illume:
btw, speaking of namespaces... Did you check about that useList used to take namespace[]?

illume (approved):
👍 thanks

PR #2494, opened 2024-10-30
frontend: Add Suspense boundary for the route Switcher

joaquimrocha:
This allows routes to suspend and lazy load. This is useful for routes that have a lot of dependencies.

We need to test also the #2484 .


illume (approved):
👍 thanks

joaquimrocha:
> We need to test also the https://github.com/headlamp-k8s/headlamp/pull/2484 .

It's working.

PR #2495, opened 2024-10-30
frontend: Add optional parameters to all Details components

joaquimrocha:
Allows for rendering Details components without the router params. This is useful for the map view but also when we add the drawer opening of details as it's planned.

### DIFF
--- a/frontend/src/components/workload/Details.tsx
+++ b/frontend/src/components/workload/Details.tsx
@@ -12,10 +12,13 @@ import {
 
 interface WorkloadDetailsProps<T extends WorkloadClass> {
   workloadKind: T;
+  name?: string;
+  namespace?: string;
 }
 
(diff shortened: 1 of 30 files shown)


frontend/src/components/workload/Details.tsx
```diff
@@ -12,10 +12,13 @@ import {
 
 interface WorkloadDetailsProps<T extends WorkloadClass> {
   workloadKind: T;
+  name?: string;
+  namespace?: string;
```
illume, line 16: Do we need to support `namespace: string[]` in these?

cc @sniok

joaquimrocha: No. This is a details view. So we just show one resource in it, which can only be in one namespace.

illume (approved):
👍 thanks

PR #2497, opened 2024-10-30
headlamp-plugin: Bump version v0.10.0-alpha.7

illume:


skoeva (commented):
Been a while but I'm guessing this can be closed

PR #2501, opened 2024-10-30
Add Orange as an adopter

gberche-orange:
Thanks again for the support you gave to @Guilamb in previous contributions. This small PR to share usage we have of headlamp at orange.


joaquimrocha:
Thanks @gberche-orange !

PR #2503, opened 2024-10-30
frontend KubeObject: Fix missing scale api in apiEndpoint getter

### LINKED ISSUE #2500: Scaling Deployments doesn't work
Scaling doesn't work

How to reproduce

1. Go to any deployment
2. Scale to any number
3. Nothing happens

This was most likely caused by changes to the apiEndpoint of KubeObject 

I'm trying to find a fix now

sniok:
Fixes #2500

apiEndpoint.scale was missing in Deployment/ReplicaSet/StatefulSet because factory that handles multiple endpoints doesn't create scale API.

Getter was passing array in every case, even the length was 1, but that resulted in multipleApiFactory always being called (the one that doesn't handle scale api)

### DIFF
--- a/frontend/src/lib/k8s/KubeObject.ts
+++ b/frontend/src/lib/k8s/KubeObject.ts
@@ -48,14 +48,23 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
     const factory = this.isNamespaced ? apiFactoryWithNamespace : apiFactory;
     const versions = Array.isArray(this.apiVersion) ? this.apiVersion : [this.apiVersion];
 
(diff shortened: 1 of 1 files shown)


joaquimrocha (dismissed):
I think this fix is a bit cryptic. I think we won't remember why this was made after a while. Can we check more directly? Feels like the problem is the side effect of being a single or multiple factory depending on the arguments being a 1 elem list or not.

sniok:
refactored it a bit so it's clearer what and why we're doing it

illume (approved):
🎉 thanks.

This fixes it for me.

PR #2512, opened 2024-11-03
frontend: ClusterNotFoundPopup: Fix to work with multi clusters

illume:
Now when there's a wrong cluster name in the URL when using
multiple clusters it also gives a cluster not found alert
message for each cluster not found.

### How to test

- [x] Add two clusters with + between then in the URL, one should be wrong, then see cluster not found message

### DIFF
--- a/frontend/src/components/App/Layout.tsx
+++ b/frontend/src/components/App/Layout.tsx
@@ -10,7 +10,7 @@ import { useDispatch } from 'react-redux';
 import { useClustersConf } from '../../lib/k8s';
 import { request } from '../../lib/k8s/apiProxy';
 import { Cluster } from '../../lib/k8s/cluster';
-import { getCluster } from '../../lib/util';
+import { getCluster, getClusterGroup } from '../../lib/util';
 import { setConfig } from '../../redux/configSlice';
 import { ConfigState } from '../../redux/configSlice';
 import { useTypedSelector } from '../../redux/reducers/reducers';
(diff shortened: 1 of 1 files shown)


joaquimrocha (changes requested):
Does this show a popup per cluster? I think this UX could be unnerving for the user. I had actually disabled the popup when in multi-cluster.
https://github.com/headlamp-k8s/headlamp/commit/c42f7b20d78bb7ae443b9ce5e953d6ee66c28b08

Can't remember exactly but I think I had tried to add the cluster problems in the main view, to be expanded at will.

illume:
@joaquimrocha please respond to this comment? https://github.com/headlamp-k8s/headlamp/pull/2512#issuecomment-2464301865

joaquimrocha:
> I don't think completely disabling the alert is a good idea.
> 
> It's most likely every single cluster won't break at the same time.
> 
> How about:
> 
> * There is already a note in the main view cluster list about the clusters being down.
> * I limit the amount of popups for max 2 clusters?

Okay. Let's try this. If it gets unnerving, we can act on that.

joaquimrocha:
Solved a conflict it had.

illume:
> > I don't think completely disabling the alert is a good idea.
> > It's most likely every single cluster won't break at the same time.
> > How about:
> > 
> > * There is already a note in the main view cluster list about the clusters being down.
> > * I limit the amount of popups for max 2 clusters?
> 
> Okay. Let's try this. If it gets unnerving, we can act on that.

I made it so only a maximum of only two alerts will be shown.

illume:
rebased against main

knrt10 (approved):
Tested. Lgtm. Thanks

PR #2513, opened 2024-11-03
frontend: RecentClusters: Add multi cluster selection feature

illume:
If there are multiple clusters then this allows selecting up to three of them from the recent clusters list.

If there is a single cluster it acts as before.

This feature can be enabled by using:
- `REACT_APP_MULTI_ENABLED=true make run-frontend`

### Screenshot

Here we see a toggle button group of the three most recently used clusters.

image

As a bonus this fixes a bug:
- text is not cut off (important when the shared resource/subscription name is the start of the cluster name)

### How to test

- [x] have at least two clusters enabled, and select the clusters you want to use at once, it takes you to the correct URL with a + separating the clusters
- [x] when you don't select any clusters it doesn't let you select the view button

### DIFF
--- a/frontend/src/components/App/Home/RecentClusters.tsx
+++ b/frontend/src/components/App/Home/RecentClusters.tsx
@@ -1,4 +1,5 @@
-import Grid from '@mui/material/Grid';
+import { Button, Grid, ToggleButton as MuiToggledButton, ToggleButtonGroup } from '@mui/material';
+import { styled } from '@mui/system';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { generatePath, useHistory } from 'react-router-dom';
@@ -7,6 +8,13 @@ import { Cluster } from '../../../lib/k8s/cluster';
 import { createRouteURL } from '../../../lib/router';
 import { getClusterPrefixedPath } from '../../../lib/util';
 import SquareButton from './SquareButton';
+
+const ToggleButton = styled(MuiToggledButton)({
+  textTransform: 'none',
+});
+
+/** Allow selecting multiple recent clusters. */
+const MULTI_ENABLED = import.meta.env.REACT_APP_MULTI_ENABLED === 'true' || true;
 
 interface ClusterButtonProps extends React.PropsWithChildren<{}> {
   /** The cluster to display this button for. */
@@ -49,6 +57,8 @@ export default function RecentClusters(props: RecentClustersProps) {
     }
   }, []);
   const { t } = useTranslation();
+  const [selectedClusters, setSelectedClusters] = React.useState<Cluster[]>([]);
+
   const recentClustersLabelId = 'recent-clusters-label';
   const maxRecentClusters = 3;
   // We slice it here for the maximum recent clusters just for extra safety, since this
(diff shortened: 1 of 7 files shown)


sniok:
The button group showing recent clusters doesn't seem to be very intuitive, I think it would be hard to understand
To be honest I personally never use recent cluster buttons, I always select from the list

What if we did something like this? 

image

illume:
Yeah, we can add multiple select with checkboxes like that. But I'll explain why I didn't...

> The button group showing recent clusters doesn't seem to be very intuitive, I think it would be hard to understand
To be honest I personally never use recent cluster buttons, I always select from the list

For what it's worth I tried it with two people without prompting "Select multiple clusters" and they figured it out. The third person I just showed the video and they suggested changing it to use the existing Recent cluster cards instead of just the words.

Mainly this code is a temporary step to allow something now and until both 
1) the full 'new home' design is implemented and can be iterated on/user tested
2) and the websocket multiplexer is integrated and tested.

The reasons are:
- The idea is to limit it to 3 clusters until the multiplexer is in properly
- It lets people test the rest of the multi cluster more easily without waiting for the rest of the stuff
- We can use it in a demo in Kubecon in a few days
- This can be enabled/disabled trivially

illume:
> This should be replaced by the commits in https://github.com/headlamp-k8s/headlamp/commits/aggregated-view-original-rebased/ (lookout for fixups) related to [e9b7f46](https://github.com/headlamp-k8s/headlamp/commit/e9b7f46d5bd1befb3ea7df1d3ba79d81d275867c) .

@joaquimrocha we discussed this is a temporary measure for [reasons mentioned here](https://github.com/headlamp-k8s/headlamp/pull/2513#issuecomment-2464776151). Additionally it was discussed and agreed to take this approach with Chris. First a minimal version that is disabled, and then the full version.

sniok:
If this is temporary and was user tested a bit then it's okay, we can do checkboxes later

sniok (approved):

sniok:
> For the checkboxes is what I proposed to handle maxium of 3 sound ok? Or if you have any other ideas for that I'd appreciate it.

Yeah that's reasonable. Although I'm a bit worried about allowed namespaces feature. If we have 2 allowed namespaces for each of 3 selected clusters and go to Workloads page then we'll get (7 workloads)*(3 cluster)*(2 namespaces) websockets. But it's and edgecase.

> Maximum of 3 selected, and then it won't let you add any more, with an additional message. "Selected 3 of 3. Currently only three clusters can be selected at once"

I think it's better to disable the "connect" (or "view") button (keep the message, but highlight in red "Selected 4 of 3. Currently only...") but still allow selecting checkboxes, I think it'll be less frustrating

illume:
Thanks @sniok

farodin91:
@illume i started to work on multi delete feature. You table and my table changes should be merged. What do you think of my design by hiding filter during the selection? #2594

joaquimrocha:
ATM we should rather add the named groups feature instead of a radio button based solution.
We should still keep the multiple selection of clusters so we create a group (named) from them.

frontend/src/components/App/Home/RecentClusters.tsx
```diff
@@ -8,6 +9,13 @@ import { createRouteURL } from '../../../lib/router';
 import { getClusterPrefixedPath } from '../../../lib/util';
 import SquareButton from './SquareButton';
 
+const ToggleButton = styled(MuiToggledButton)({
+  textTransform: 'none',
+});
+
+/** Allow selecting multiple recent clusters. */
+const MULTI_ENABLED = import.meta.env.REACT_APP_MULTI_ENABLED === 'true' || true;
```
joaquimrocha, line 17: I am willing to merge this PR as a temporary way to test this, but seems like this is enabled all the time.

illume:
- to be prepared to be merged at the start of the next release cycle
- change to checkboxes to view, not the current toggle box thing
- leave on by default, and then before the release it can be disabled if it's not working
- This will let us devs test during the release cycle

illume:
Closing in favour of https://github.com/kubernetes-sigs/headlamp/pull/3119

PR #2514, opened 2024-11-03
frontend: Home: Fix so it does not rerender a lot

illume:
Some of the hooks were causing a rerender several times every 5 seconds or so. Even when things have not changed. Some of the types have been added to make it more clear what data is being passed around.

Also now there is a HomePure presentational component.

Event.useWarningList and useClustersVersion needed a deep equality comparison to prevent rerendering properly.

### how to test

- [x] You can either use dev tools or add a console.log into the HomePure component to see how often it is rendered. With this change it should only rerender when something changes.
- [x] stop a cluster, and start a cluster... the home page should update to show the correct cluster active status

### DIFF
--- a/frontend/src/components/App/Home/index.tsx
+++ b/frontend/src/components/App/Home/index.tsx
@@ -6,7 +6,8 @@ import ListItemText from '@mui/material/ListItemText';
 import Menu from '@mui/material/Menu';
 import MenuItem from '@mui/material/MenuItem';
 import Typography from '@mui/material/Typography';
-import React from 'react';
+import isEqual from 'lodash/isEqual';
+import React, { useMemo, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
@@ -160,13 +161,15 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
 
 export default function Home() {
   const history = useHistory();
-  const clusters = useClustersConf() || {};
+  const clustersConf = useClustersConf() || {};
+  const clusters = useMemo(() => {
+    return clustersConf;
+  }, [clustersConf]);
 
   if (!helpers.isElectron() && Object.keys(clusters).length === 1) {
     history.push(createRouteURL('cluster', { cluster: Object.keys(clusters)[0] }));
     return null;
   }
-
   return <HomeComponent clusters={clusters} key={Object.keys(clusters).join('')} />;
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/App/Home/index.tsx
```diff
@@ -160,30 +161,106 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
...
+  const [versions, errors] = useDeepCompareMemo(versionsErrors, [versionsErrors]);
+  const warningsList = Event.useWarningList(customClusterNames);
+  const warningsMap = useDeepCompareMemo(warningsList, [warningsList]);
+
+  console.log('versions', versions);
+  console.log('clusterList', clusterList);
+  console.log('warningsList', warningsList);
+  console.log('customClusterNames.join', customClusterNames.join('-'));
+
+  const rendered = useMemo(() => {
```
sniok, line 230: Please use React.memo with _.isEqual for the pure Home component, then most of the useMemos won't be necessary

frontend/src/components/App/Home/index.tsx
```diff
@@ -160,30 +161,106 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
 
 export default function Home() {
   const history = useHistory();
-  const clusters = useClustersConf() || {};
+  const clustersConf = useClustersConf() || {};
+  const clusters = useMemo(() => {
+    return clustersConf;
+  }, [clustersConf]);
```
joaquimrocha, line 167: I thought memo didn't compare objects. If that's the case, this should not do much.

frontend/src/components/App/Home/index.tsx
```diff
@@ -160,30 +161,106 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
...
 
   if (!helpers.isElectron() && Object.keys(clusters).length === 1) {
     history.push(createRouteURL('cluster', { cluster: Object.keys(clusters)[0] }));
     return null;
   }
-
```
joaquimrocha, line 169: Not needed.

frontend/src/components/App/Home/index.tsx
```diff
@@ -160,30 +161,106 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
...
+  }, [customNameClusters]);
+
+  const clusterList = useMemo(() => {
+    return Object.values(clusters);
+  }, [clusters]);
+
+  const versionsErrors = useClustersVersion(clusterList);
+  const [versions, errors] = useDeepCompareMemo(versionsErrors, [versionsErrors]);
+  const warningsList = Event.useWarningList(customClusterNames);
+  const warningsMap = useDeepCompareMemo(warningsList, [warningsList]);
```
joaquimrocha, line 223: Assuming the warnings are only updated when the actual warning objects in them change, I don't think this is making a lot of difference.

frontend/src/components/App/Home/index.tsx
```diff
@@ -160,30 +161,106 @@ function ClusterStatus({ error }: { error?: ApiError | null }) {
...
+    return (
+      <HomePure
+        customNameClusters={customNameClusters}
+        customClusterNames={customClusterNames}
+        versions={versions}
+        errors={errors}
+        warningsMap={warningsMap}
+      />
+    );
+  }, [customClusterNames, customClusterNames, versions, errors, warningsMap]);
```
joaquimrocha, line 240: I don't understand how the memo helps here. The component shouldn't refresh anyway unless one of its state or props changes, and AFAICS, the properties are now all extracted and being passed to them but also added here as deps, so this is having the same effect as if we just pass the properties.

joaquimrocha (changes requested):
I think a much more effective and simpler change may be to fix useWarningList and useClustersVersion so it doesn't update unless something in the end result is changed.

PR #2518, opened 2024-11-04
frontend: CustomResourceInstancesList: Fix to work with multi cluster

illume:


frontend/src/components/crd/CustomResourceInstancesList.tsx
```diff
@@ -10,8 +10,14 @@ import { ResourceListView } from '../common/Resource';
 function CrInstancesView({ crds }: { crds: CRD[]; key: string }) {
   const { t } = useTranslation(['glossary', 'translation']);
 
-  const crdInfo = useMemo(() => crds.map(crd => crd.makeCRClass()), [crds]);
-  const queries = crdInfo.map(cr => cr.useList());
+  const dataClassCrds = crds.map(crd => {
+    const crdClass = crd.makeCRClass();
+    const data = crdClass.useList({ cluster: crd.cluster });
+    return { data, crdClass, crd };
+  });
```
joaquimrocha, line 17: Should we keep a useMemo for at least generating the classes here?

joaquimrocha (changes requested):
Also, I don't understand what the issue was from the commit description.

illume:
> Also, I don't understand what the issue was from the commit description.

The issue is the view completely failed to render.

frontend/src/components/crd/CustomResourceInstancesList.tsx
illume: I don't think so. I had tried that, but generating the classes doesn't take very much time.

If it's a problem for you, and not a nit... I can have a bit more of a think.

sniok (approved):
nice

skoeva (approved):

frontend/src/components/crd/CustomResourceInstancesList.tsx
illume: I've expanded on the commit message.

illume:
@joaquimrocha please review https://github.com/headlamp-k8s/headlamp/pull/2518#discussion_r1834016798

frontend/src/components/crd/CustomResourceInstancesList.tsx
joaquimrocha: > I don't think so. I had tried that, but generating the classes doesn't take very much time.
> 
> If it's a problem for you, and not a nit... I can have a bit more of a think.

Alright. It was mostly a nit. We can keep it like this and optimize later if needed.

PR #2520, opened 2024-11-04
frontend: Details: Make multi cluster aware

illume:
This makes Details views work with multiple cluster. 

The ?cluster is passed via a URL query parameter. Because in the normal cluster part of the URL is the group of clusters, and the details view needs to know which of the clusters the resource is in.

Here you can see details views being clicked on and the work (even though multiple clusters are active).

https://github.com/user-attachments/assets/3a840c4d-7e73-4dd4-8ef9-4b81d171cb2f

- **frontend: DetailsGrid: Fix for using a specific cluster**
- **frontend: router: Fix createRouteURL to take query params**
- **frontend: KubeObject: Add cluster query param to getDetailsLink**
- **frontend: k8s: Add useClusterFromURLVar to get cluster**
- **frontend: Details: Add support for cluster via URL var**

### DIFF
--- a/frontend/src/lib/k8s/index.ts
+++ b/frontend/src/lib/k8s/index.ts
@@ -140,6 +140,17 @@ export function useClusterGroup(): string[] {
   return clusterGroup;
 }
 
+/**
+ * Use the cluster name from the URL query parameters if it's there.
+ *
+ * @returns the cluster name from the URL. If no cluster is defined in the URL, undefined is returned.
+ */
+export function useDetailsSingleCluster(): string | undefined {
+  const location = useLocation();
+  const searchParams = new URLSearchParams(location.search);
+  return searchParams.get('cluster') || undefined;
+}
+
 export function getVersion(clusterName: string = ''): Promise<StringDict> {
   return clusterRequest('/version', { cluster: clusterName || getCluster() });
 }
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -17,7 +17,7 @@ import React, { PropsWithChildren, ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 import { generatePath, NavLinkProps, useLocation } from 'react-router-dom';
 import YAML from 'yaml';
-import { labelSelectorToQuery, ResourceClasses } from '../../../lib/k8s';
+import { labelSelectorToQuery, ResourceClasses, useClusterFromURLVar } from '../../../lib/k8s';
 import { ApiError } from '../../../lib/k8s/apiProxy';
 import { KubeCondition, KubeContainer, KubeContainerStatus } from '../../../lib/k8s/cluster';
 import { KubeEvent } from '../../../lib/k8s/event';
@@ -113,7 +113,6 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
     resourceType,
     name,
     namespace,
-    cluster,
     children,
     withEvents,
     extraSections,
@@ -128,6 +127,8 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
     state => state.detailsViewSection.detailsViewSectionsProcessors
   );
   const dispatchHeadlampEvent = useEventCallback();
+  const clusterFromURLVar = useClusterFromURLVar();
+  const cluster = props.cluster || clusterFromURLVar;
 
   // This component used to have a MainInfoSection with all these props passed to it, so we're
   // using them to accomplish the same behavior.
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -89,6 +89,8 @@ export interface DetailsGridProps<T extends KubeObjectClass>
   name: string;
   /** Namespace of the resource. If not provided, it's assumed the resource is not namespaced. */
   namespace?: string;
+  /** Cluster of the resource. If not provided, it's assumed single cluster mode. */
+  cluster?: string;
   /** Sections to show in the details grid (besides the default ones). */
   extraSections?:
     | ((item: InstanceType<T>) => boolean | DetailsViewSection[] | ReactNode[])
@@ -111,6 +113,7 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
     resourceType,
     name,
     namespace,
+    cluster,
     children,
     withEvents,
     extraSections,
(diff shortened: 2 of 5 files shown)


joaquimrocha:
Please look into this branch, since I think I had also added a way to llink to resources details without having to create a new URL query param:
https://github.com/headlamp-k8s/headlamp/commits/aggregated-view/

joaquimrocha (changes requested):
I think (haven't double checked) that in https://github.com/headlamp-k8s/headlamp/tree/aggregated-view-original-rebased , I had computed the route directly from the resource in ResourceLink, thus using the known URLs without a cluster param.

illume:
>frontend: Automatically add the cluster param to ResourceLink 0e2f0f3
> IIUC, this replaces that 1st PR from the list you shared earlier
>> That's included already. It adds to the cluster param AFAIK. Which is /c/<cluster>/
>> Thanks for the tip though. Let me try running that branch and see if I can find it that way.
>
> Yes, and that's okay because when you go into a resource details view, you want to see only one cluster.
> When you press back, it goes back to whatever the URL was, including the c/SOMEGROUP

The problem with only relying on this is that:
- the URL has only one cluster in it, which means sharing/using links loses the cluster group
- if a person clicks on another link in the detail view then they lose the cluster group

This PR makes these things work.

joaquimrocha:
I see, and it's a good idea to keep the group context. But for that, I think it makes more sense to do it the other way around: keep the URL for the details view matching the normal details URL (i.e. /c/SINGLE_CLUSTER/resource-kind/my-resource) and use the query param for a group association: `/c/SINGLE_CLUSTER/resource-kind/my-resource?cluster-group=MyGroup`. This way, we accomplish what you want (keep a reference of the group) we don't have 2 different URLs for the same view, and we can eventually lose the query param and the view still works.
Does this make sense?

illume:
All other links would then need to special case where /c/<cluster> means something different?

I don't think your proposal makes sense. Because why would we change the URL to '/c/<clusterGroup>' to just a single cluster just for this?

frontend/src/lib/k8s/index.ts
```diff
@@ -140,6 +140,17 @@ export function useClusterGroup(): string[] {
   return clusterGroup;
 }
 
+/**
+ * Use the cluster name from the URL query parameters if it's there.
+ *
+ * @returns the cluster name from the URL. If no cluster is defined in the URL, undefined is returned.
+ */
+export function useClusterFromURLVar(): string | undefined {
```
sniok, line 148: I think we should be very explicit about what this means, because we already have a cluster 'variable' in the url /c/cluster/. To disambiguate I think this should be named something like `useClusterFromQueryParam`

illume: What do you mean, never mind?

illume: I was thinking “query” is maybe overloaded with react query stuff.  But maybe it’s ok.

useClusterFromSearchParam is also not quite right making it sound like it’s from a search.

useClusterFromURLQuery?
useURLQueryCluster?

sniok: well there's a separate discussion about the url format, so this comment may not be relevant

sniok: useURLQueryCluster sounds good
hm or maybe an approach of naming it for what it is instead of where it's stored, something like `useDetailsSingleCluster`?

illume: Yeah maybe useDetailsSingleCluster is the least misdirecting/confusing.

illume: > well there's a separate discussion about the url format, so this comment may not be relevant, let me rephrase it better

Can you please let us know which proposal you’d prefer and why?

sniok:
URL is supposed to be a unique resource identifier
if we think in terms of purely displaying details of a resource then `/c/SINGLE_CLUSTER/resource-kind/my-resource` makes sense
but the app also has selected clusters in the sidebar for example, so url has to represent that too

with the previous multi cluster changes we already changed the meaning of the /c/cluster/ param, it now represents all the selected clusters and not the cluster of a given resource. 

so now we need to store information about a given resource cluster:

we could repurpose the `/c/cluster/` param for it, but now it means different thing depending where we are in the app (selected clusters or cluster of a resource when showing details) and IMHO it's a bit confusing

we could put it in the query params `/c/clusetA+clusterB/pods/my-pod?cluster=clusterA` but query parameters are not supposed to be used for unique identification of a resource

a logical solution would be to introduce a new param in the url, like
`/clusterB+clusterA/details/clusterA/pods/my-pod` where `/{selectedclusters}/details/{resourceCluster}/{resourceKind}/{resourceName}`
(details can be ommited, variable order can be changed, it doesn't matter)

but this kind of solution seems like a big breaking change. so I think doing `?cluster=clusterA` is the simplest solution here?

frontend/src/lib/k8s/index.ts
sniok: posted it in a separate comment

illume:
> a logical solution would be to introduce a new param in the url, like
`/clusterB+clusterA/details/clusterA/pods/my-pod` where `/{selectedclusters}/details/{resourceCluster}/{resourceKind}/{resourceName}`
(details can be ommited, variable order can be changed, it doesn't matter)
> 
>  but this kind of solution seems like a big breaking change. so I think doing ?cluster=clusterA is the simplest solution here?

Adding `{resourceCluster}` into the URL could work too... I feel like it might even be the more correct choice?

How is it a breaking change though?

illume:
> Detail links from current version wouldn't work in new version (if that is something we're concerned about)

I guess we can keep the previous route? So it will keep working as well.
- new `/{selectedclusters}/{resourceCluster}/details/{resourceKind}/{resourceName}`
- prev `/{selectedclusters}/details/{resourceKind}/{resourceName}`

One other con against ?cluster is that if you remove it.
- if you have `/{selectedclusters}/details/{resourceKind}/{resourceName}?cluster={resourceCluster}`
- then remove `?cluster?` the URL is broken: `/{selectedclusters}/details/{resourceKind}/{resourceName}`

So yeah... for these reasons I think it's worth trying out:
- new `/{selectedclusters}/{resourceCluster}/details/{resourceKind}/{resourceName}`
- prev `/{selectedclusters}/details/{resourceKind}/{resourceName}`

What do you think?

illume:
@joaquimrocha can you please respond to this comment? https://github.com/headlamp-k8s/headlamp/pull/2520#issuecomment-2464610293

illume:
Closing in favour of 
- https://github.com/kubernetes-sigs/headlamp/pull/3120

PR #2521, opened 2024-11-04
frontend: Overview: Add cluster field for multi cluster

illume:
This is only visible if multiple clusters are being used.

image


sniok (approved):

skoeva (approved):
LGTM

PR #2522, opened 2024-11-04
Fix home constant update

joaquimrocha:
This PR fixes the useWarningsList and useClustersVersion so they do not update whenever there's new data from the server but instead when the data differs from what we returned the last time. (Same goal as #2514)

how to test
- [ ] You can either use dev tools or add a console.log into the HomePure component to see how often it is rendered. With this change it should only rerender when something changes.
- [x] stop a cluster, and start a cluster... the home page should update to show the correct cluster active status

### DIFF
--- a/frontend/src/components/App/Home/index.tsx
+++ b/frontend/src/components/App/Home/index.tsx
@@ -175,18 +175,14 @@ interface HomeComponentProps {
   clusters: { [name: string]: Cluster };
 }
 
-function HomeComponent(props: HomeComponentProps) {
-  const { clusters } = props;
-  const [customNameClusters, setCustomNameClusters] = React.useState(getClusterNames());
-  const { t } = useTranslation(['translation', 'glossary']);
-  const [versions, errors] = useClustersVersion(Object.values(clusters));
+function useWarningSettingsPerCluster(clusterNames: string[]) {
+  const warningsMap = Event.useWarningList(clusterNames);
+  const [warningLabels, setWarningLabels] = React.useState<{ [cluster: string]: string }>({});
   const maxWarnings = 50;
-  const warningsMap = Event.useWarningList(Object.values(customNameClusters).map(c => c.name));
 
-  function renderWarningsText(clusterName: string) {
+  function renderWarningsText(warnings: typeof warningsMap, clusterName: string) {
     const numWarnings =
-      (!!warningsMap[clusterName]?.error && -1) ||
-      (warningsMap[clusterName]?.warnings?.length ?? -1);
+      (!!warnings[clusterName]?.error && -1) || (warnings[clusterName]?.warnings?.length ?? -1);
 
     if (numWarnings === -1) {
       return '⋯';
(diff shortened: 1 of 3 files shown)


frontend/src/lib/k8s/event.ts
```diff
@@ -1,4 +1,5 @@
-import { useMemo } from 'react';
+import _ from 'lodash';
```
illume, line 1: nit: It's better to import just what you need, for shaking.

frontend/src/lib/k8s/event.ts
```diff
@@ -238,7 +239,19 @@ class Event extends KubeObject<KubeEvent> {
       options?.queryParams ?? {}
     );
 
-    return this.useListForClusters(clusters, { queryParams: queryParameters });
+    const newWarningsList = this.useListForClusters(clusters, { queryParams: queryParameters });
```
illume, line 242: For later... we should probably investigate doing this change inside of "useListForClusters"? Or I thought this had a caching layer built in?

illume (commented):
Hrmm. This doesn't fix the issue of the home rendering several times every five seconds.

image

frontend/src/lib/k8s/index.ts
```diff
@@ -297,8 +297,11 @@ export function useClustersVersion(clusters: Cluster[]) {
   const lastUpdateRef = React.useRef(0);
 
   React.useEffect(() => {
-    const newClusterNames = Object.values(clusters).map(c => c.name);
-    if (_.isEqual(newClusterNames, clusterNames)) {
+    const newClusterNames = Object.values(clusters)
```
illume, line 300: Can you please add a comment for why this is here?

frontend/src/components/App/Home/index.tsx
```diff
@@ -174,84 +175,125 @@ interface HomeComponentProps {
...
-              getValue: () => '',
-              cellProps: {
-                align: 'right',
+  React.useEffect(() => {
+    setWarningLabels(currentWarningLabels => {
+      const newWarningLabels: { [cluster: string]: string } = {};
+      for (const cluster of clusterNames) {
+        newWarningLabels[cluster] = renderWarningsText(warningsMap, cluster);
+      }
+      if (!_.isEqual(newWarningLabels, currentWarningLabels)) {
```
sniok, line 202: _ import is missing, replace with `if(isEqual(...`

frontend/src/lib/k8s/event.ts
```diff
@@ -238,7 +239,19 @@ class Event extends KubeObject<KubeEvent> {
...
+    const [warningList, setWarningList] = React.useState<typeof newWarningsList>(newWarningsList);
+
+    // Only update the warnings if they actually differ
+    React.useEffect(() => {
+      if (_.isEqual(warningList, newWarningsList)) {
```
sniok, line 247: this might be slow when comparing a lot of warnings, but we can profile and look into it later

sniok (approved):

PR #2523, opened 2024-11-05
frontend AuthChooser: Fix incorrect UI flashing

sniok:
AuthChooser would sometimes initially flash incorrect state requiring token before clusters config is loaded


vyncent-t:
we just reviewed this and the use token button is gone so its fixed!

sniok (commented):
lgtm

vyncent-t (approved):
This fixes the use token button lgtm

Issue #2526, opened 2024-11-05
No label for the actions menu in home view


skoeva:
## Describe the bug
The actions menu does not have a label

## To Reproduce
Steps to reproduce the bug:
1. Open Headlamp
2. Hover over actions menu on a cluster
3. Notice that it has no label

## Are you able to fix this issue?
Yes

## Additional Context
The issue is shown here (hovering over the ... on the right):

![image](https://github.com/user-attachments/assets/b6d6d6fe-4374-48a5-94ce-0f80966a30c1)

PR #2527, opened 2024-11-05
frontend: Home: Add label for actions menu

skoeva:
This change adds a label for the actions menu for each cluster on the home page, where there previous was no label.

Fixes: #2526

### Testing
- [X] Open Headlamp with 2+ clusters (not in-cluster)
- [X] Hover over the actions menu for a cluster and ensure that it has a label titled 'Actions'


illume (approved):
🎉 Awesome! Thanks.

PR #2532, opened 2024-11-06
Fix push release assets

joaquimrocha:
The push-release-assets script stopped working a while ago due to some node related plumbering.


sniok:
it's a small thing but I had trouble finding this script: could you move it outside of app into new `scripts/` folder in the root? imho `app/` should only be about electron related stuff

joaquimrocha:
I think we can rearrange things in another PR. We now have a tools folder there, so we could add these things there.

illume (approved):
👍 thanks

Issue #1632, opened 2023-12-21
Show the origin of each cluster in the home view


joaquimrocha:
In the home view, we should have a new column in the clusters list called _origin_ which shows the user whether the clusters are coming from a kubeconfig (and which path), in-cluster, or dynamic (if we could have the plugin name here or a string identifying who added it'd be great).

knrt10:
Working on this

knrt10:
This issue needs design, removing myself right now

PR #2536, opened 2024-11-06
frontend: Home: Show the origin of each cluster

skoeva:
This change introduces a new "Origin" column for the cluster list on the home page, displaying whether a cluster came from a kubeconfig, Headlamp, or a dynamic cluster, with references when available.

Fixes: #1632 

### Testing
- [X] Open Headlamp with 2+ active clusters
- [X] Ensure that the cluster list on the Home page includes the Origin column with respective origins

image


frontend/src/components/App/Home/index.tsx
```diff
@@ -209,6 +209,17 @@ function useWarningSettingsPerCluster(clusterNames: string[]) {
   return warningLabels;
 }
 
+function getOrigin(cluster: Cluster) {
```
illume, line 212: Please add function documentation?

frontend/src/components/App/Home/index.tsx
```diff
@@ -209,6 +209,17 @@ function useWarningSettingsPerCluster(clusterNames: string[]) {
...
+  if (cluster.meta_data?.source === 'kubeconfig') {
+    return `Kubeconfig (${process.env.KUBECONFIG ?? '~/.kube/config'})`;
+  } else if (cluster.meta_data?.source === 'dynamic_cluster') {
+    return `Dynamic cluster`;
+  } else if (cluster.meta_data?.source === 'in_cluster') {
+    return 'Headlamp';
```
illume, line 218: It could be Headlamp is branded as something else, so we shouldn't use Headlamp here I think.

I think "In-Cluster" is maybe better?

illume (commented):
👍 thanks

Left a couple of notes...

frontend/src/components/App/Home/index.tsx
skoeva: done

frontend/src/components/App/Home/index.tsx
```diff
@@ -264,6 +281,13 @@ function HomeComponent(props: HomeComponentProps) {
                   </Link>
                 ),
               },
+              {
+                label: t('Origin'),
+                getValue: cluster => getOrigin(cluster),
+                render: ({ name }) => (
+                  <Typography variant="body2">{getOrigin(clusters[name])}</Typography>
```
illume, line 288: Oh... I guess these will need translations.

frontend/src/components/App/Home/index.tsx
illume: Maybe the return type could be added `function getOrigin(cluster: Cluster): string {` instead of in the returns.

`@returns the description of where the cluster is picked up from. Dynamic, in-cluster or from a kubeconfig file`

frontend/src/components/App/Home/index.tsx
```diff
@@ -209,6 +209,23 @@ function useWarningSettingsPerCluster(clusterNames: string[]) {
...
 
+/**
+ * Gets the origin of a cluster.
+ *
+ * @param {Cluster} cluster
+ * @returns {string}
+ */
+function getOrigin(cluster: Cluster) {
+  if (cluster.meta_data?.source === 'kubeconfig') {
+    return `Kubeconfig (${process.env.KUBECONFIG ?? '~/.kube/config'})`;
```
illume, line 220: But maybe the translations could be added in this function, so they can be detected by that i18n script.

illume (approved):
👍 thanks

(probably just the translation change should be done, the docs change is just a nit)

frontend/src/components/App/Home/index.tsx
skoeva: done

frontend/src/components/App/Home/index.tsx
skoeva: added translations

frontend/src/components/App/Home/index.tsx
```diff
@@ -236,6 +236,24 @@ function HomeComponent(props: HomeComponentProps) {
...
+  /**
+   * Gets the origin of a cluster.
+   *
+   * @param cluster
+   * @returns A description of where the cluster is picked up from: dynamic, in-cluster, or from a kubeconfig file.
+   */
+  function getOrigin(cluster: Cluster): string {
+    if (cluster.meta_data?.source === 'kubeconfig') {
+      const kubeconfigPath = process.env.KUBECONFIG ?? '~/.kube/config';
+      return t('translation|Kubeconfig') + ` (${kubeconfigPath})`;
```
joaquimrocha, line 248: We don't need to translate Kubeconfig. And the value here should be: `kubeconfig:PATH_TO_FILE`
Can you make another PR to fix this, since this one is merged?

frontend/src/components/App/Home/index.tsx
```diff
@@ -236,6 +236,24 @@ function HomeComponent(props: HomeComponentProps) {
...
+    } else if (cluster.meta_data?.source === 'dynamic_cluster') {
+      return t('translation|Dynamic cluster');
```
joaquimrocha, line 250: I am not sure this will give a lot of info to the user. I'd prefer the still poor but more accurate `Plugin` for now.

joaquimrocha (commented):
I think a few things can be improved.

frontend/src/components/App/Home/index.tsx
skoeva: Got it

frontend/src/components/App/Home/index.tsx
skoeva: This change as well as the other one mentioned are done here: #2543

Issue #2534, opened 2024-11-06
Map crahsing: RangeError: Maximum call stack size exceeded


h0jeZvgoxFepBQ2C:
## Describe the bug
Map is not working

## To Reproduce
Steps to reproduce the bug:
Go to map

## Environment (please provide info about your environment):
 - Installation type: Downloaded OSX ARM binary
 - Headlamp Version: 0.26.0

## Are you able to fix this issue?
No

## Additional Context

We have a big cluster, around 500 namespaces and 1200 pods / 500 deployments

```error
RangeError: Maximum call stack size exceeded
    at file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/GraphView-C5xOIooh.js:43:226311
    at Set.forEach (<anonymous>)
    at file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/GraphView-C5xOIooh.js:43:226235
    at UJn (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/GraphView-C5xOIooh.js:43:225328)
    at VJn (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/GraphView-C5xOIooh.js:43:226199)
    at pm (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:38:17018)
    at Vp (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:3139)
    at rx (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:44833)
    at tx (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:39790)
    at WR (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:39718)
    at Kc (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:39570)
    at tf (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:35934)
    at Zb (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-DuPX71-Q.js:40:34883)
[...]

PR #2538, opened 2024-11-07
frontend: Improve Map performance

sniok:
This fixes a stack overflow crash #2534 and improves grouping and filtering performance 

There's still a bunch of stuff that can be optimized so this is only first step

Testing done:

- [x] Tested on KWOK cluster with 20,000 pods

After fixes:
Map loading time: 5s
Filter by error: <1s
Group by node: <1s

Before:
Map loading time: ~3.5min
Filter by error: 50s
Group by node: 9s

### DIFF
--- a/frontend/src/components/resourceMap/GraphView.tsx
+++ b/frontend/src/components/resourceMap/GraphView.tsx
@@ -129,7 +129,7 @@ function GraphViewContent({
     if (hasErrorsFilter) {
       filters.push({ type: 'hasErrors' });
     }
-    if (namespaces) {
+    if (namespaces?.size > 0) {
       filters.push({ type: 'namespace', namespaces });
     }
     return filterGraph(nodes, edges, filters);
--- a/frontend/src/components/resourceMap/graph/graphLookup.ts
+++ b/frontend/src/components/resourceMap/graph/graphLookup.ts
@@ -1,0 +1,56 @@
+import { GraphEdge, GraphNode } from './graphModel';
+
+/**
+ * Constant time lookup of graph elements
+ */
+export interface GraphLookup<N, E> {
+  /** Get list of outgoing edges from the given node */
+  getOutgoingEdges(nodeId: string): E[] | undefined;
+  /** Get list of incoming edges to the given node */
+  getIncomingEdges(nodeId: string): E[] | undefined;
+  /** Get Node by its' ID */
+  getNode(nodeId: string): N | undefined;
+}
+
+/**
+ * Creates a utility for constant time lookup of graph elements
+ *
+ * @param nodes - list of graph Nodes
+ * @param edges - list of graph Edges
+ * @returns lookup {@link GraphLookup}
+ */
+export function makeGraphLookup<N extends GraphNode, E extends GraphEdge>(
+  nodes: N[],
+  edges: E[]
+): GraphLookup<N, E> {
+  const nodeMap = new Map<string, N>();
+  nodes.forEach(n => {
+    nodeMap.set(n.id, n);
+  });
+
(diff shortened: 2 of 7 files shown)


frontend/src/components/resourceMap/GraphView.tsx
```diff
@@ -129,7 +129,7 @@ function GraphViewContent({
     if (hasErrorsFilter) {
       filters.push({ type: 'hasErrors' });
     }
-    if (namespaces) {
+    if (namespaces && namespaces.size > 0) {
```
skoeva, line 132: This can be simplified:
```suggestion
    if (namespaces?.size > 0) {
```

frontend/src/components/resourceMap/graph/graphLookup.ts
```diff
@@ -0,0 +1,56 @@
...
+  /** Get list of outgoing edges from the given node */
+  getOutgoingEdges(nodeId: string): E[] | undefined;
+  /** Get list of incoming edges to the given node */
+  getIncomingEdges(nodeId: string): E[] | undefined;
+  /** Get Node by its' ID */
+  getNode(nodeId: string): N | undefined;
+}
+
+/**
+ * Creates an utility for constant time lookup of graph elements
```
skoeva, line 16: minor nit
```suggestion
 * Creates a utility for constant time lookup of graph elements
```

frontend/src/components/resourceMap/sources/GraphSources.tsx
```diff
@@ -232,13 +232,17 @@ export function GraphSourceManager({ sources, children }: GraphSourceManagerProp
...
       selectedSources.forEach(id => {
         const data = sourceData.get(id);
-        nodes.push(...(data?.nodes ?? []));
-        edges.push(...(data?.edges ?? []));
+        if (data?.nodes) {
+          nodes = nodes.concat(data.nodes);
+        }
+        if (data?.edges) {
+          edges = edges.concat(data.edges);
+        }
```
skoeva, line 245: I think push is faster than concat, is there another reason to make this change?

skoeva (commented):
Huge perf improvement, thanks for this!

frontend/src/components/resourceMap/sources/GraphSources.tsx
sniok: push(...arr) crashes if array is larger than the call stack can hold, and this fixes that. 
concat works but you're right it's not the fastest, although this is not in a hot path. I'll check other solutions

illume:
Very nice.

Did you look into how it reacts with a fairly high frequency of change? I mean with pods/events being added, rather than with a static set of resources.
(If not, probably something to look at in another PR.)

frontend/src/components/resourceMap/sources/GraphSources.tsx
sniok: I did some simple tests and it looks like concat is consistently faster, maybe this case is optimized by v8 or something. keeping it like this

sniok:
> Did you look into how it reacts with a fairly high frequency of change? I mean with pods/events being added, rather than with a static set of resources. (If not, probably something to look at in another PR.)

Yeah it struggles a bit with frequent changes, but that is another issue

sniok:
> Doh. I think I broke the test plugins CI job with the release of the new headlamp-k8s/eslint-config. It is trying to upgrade to the "latest" `headlamp-k8s/eslint-config`, where it should be getting the version related to that version of headlamp-plugin.
> 
> So you can ignore that test plugins job failure in this PR.

Got it, thanks for the explanation

illume (approved):
👍

PR #2543, opened 2024-11-07
frontend: Home: Improve cluster origin labels

skoeva:
This change improves on #2536 by adjusting the labels in the cluster origin column of the Home page for clarity and accuracy.


frontend/src/components/App/Home/index.tsx
```diff
@@ -248,9 +248,9 @@ function HomeComponent(props: HomeComponentProps) {
   function getOrigin(cluster: Cluster): string {
     if (cluster.meta_data?.source === 'kubeconfig') {
       const kubeconfigPath = process.env.KUBECONFIG ?? '~/.kube/config';
-      return t('translation|Kubeconfig') + ` (${kubeconfigPath})`;
+      return `Kubeconfig: ${kubeconfigPath}`;
     } else if (cluster.meta_data?.source === 'dynamic_cluster') {
-      return t('translation|Dynamic cluster');
+      return t('translation|Plugin');
```
sniok, line 253: When I manually add a cluster in a desktop app, isn't that a dynamic cluster and it shouldn't be called plugin? Or am I misunderstanding

skoeva: I think you'll be able to see the difference if you try logging the metadata for the cluster you create. Using the dynamic-clusters plugin to create a cluster should give the cluster a different source than the generic add cluster button I believe, correct me if I'm wrong

skoeva: I think both? I can try this soon

sniok: The use case I'm interested in: Launch App -> Add cluster with "Add cluster" button -> Look at the table of clusters -> What will be the origin?

skoeva: I know when using headlamp with stateless clusters enabled, adding using the add cluster button from dynamic-clusters plugin will show "dynamic cluster" as the origin. Need to test for in-cluster case

skoeva: Unsure but don't think this should be blocking here, plugin was specified in #1632. If there is an issue with the cluster source in the metadata, I can look into that as well in a separate PR

sniok: From #1632 
> or dynamic (if we could have the plugin name here or a string identifying who added it'd be great).

Meaning it shows something like "Dynamic (Added by plugin: plugin-name)" if we have information about who added it, but by default I think it should show "Dynamic"

skoeva: The issue is that the plugin name unfortunately isn't stored anywhere, and dynamic is not descriptive enough so [this is the alternative for now](https://github.com/headlamp-k8s/headlamp/pull/2536#discussion_r1832909760)

sniok: ah okay, got it

joaquimrocha (approved):

PR #2549, opened 2024-11-07
frontend: eslint-config: headlamp-plugin: eslint related updates

illume:
- **eslint-config: Bump typescript to 5.6.3**
- **eslint-config: Bump yaml to 2.6.0**
- **eslint-config: Bump @typescript-eslint/parser to 8.13.0**
- **eslint-config: Bump eslint and related packages**
- **eslint-config: Bump version to 0.7.0**
- **frontend: headlamp-plugin: Bump headlamp-k8s/eslint-config and related**


illume:
> react-hooks eslint plugin is not used, does it make sense to keep/bump it?

Yeah, I will remove it. I guess we should turn it on in a future version. Because I guess it will uncover issues in Headlamp and plugins... so just turning it on right now would not be a good idea.

illume:
This release uncovered a bug in the upgrading of the headlamp-k8s/eslint-config package. It should upgrade to the version related to headlamp-plugin not to the "latest" version always. Because now the CI test job is broken in `main` because it is trying to get the latest headlamp-k8s/eslint-config and that is not compatible with that older version of headlamp-plugin.

So I will fix that also in this PR too.

skoeva:
Just saw the update, feel free to link #2560 to this PR in the case that the issue gets resolved

sniok:
any updates on this?

PR #2550, opened 2024-11-07
WIP: add a11y to playwright tests

vyncent-t:
rework of lost branch for adding a11y tests to main playwright tests


vyncent-t:
note: this branch is built off the app playwright work, the app parts will need to be removed later

I will work on only working on the main e2e work

vyncent-t:
we now have a newer way of doing this, closing

PR #2555, opened 2024-11-11
docs: Add publishing guide

joaquimrocha:
Here are the docs for the plugin publishing guide.
After putting these together, I have 2 remarks:
1. I still feel like the guide would make more sense if we add the repo in ArtifactHub and only then add the files to the github page, same with the package. Just would feel more natural to me: I create a repo in AH, then enable it.
2. The guide is for publishing on ArtifactHub which has little to do with Headlamp and more to do with the Plugin Catalog, but it's important enough  that I think it's okay to have it here. I added some text explaining that it's not about discoverability on Headlamp core but rather in the plugin.

cc/ @yolossn


docs/development/plugins/publishing.md
```diff
@@ -0,0 +1,100 @@
...
+```
+It will create a tarball of the plugin in the format that is accepted by headlamp and also print the checksum of the tarball that we will use in the next step.
+2. Create a new release on GitHub for your plugin.
+3. Tag the release with a semantic version number (e.g., v1.0.0).
+4. Upload the built plugin files as assets to this release.
+
+For detailed instructions on creating a new release, refer to the [GitHub documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository).
+
+:::info
+For security reasons, Headlamp allows the download of plugins exclusively from GitHub, GitLab, and BitBucket. While this may change in the future, if you need your plugins to be downloaded from somewhere else, please [let us know](https://github.com/headlamp-k8s/headlamp/issues).
```
yolossn, line 43: image

Formatting is bit off here.

docs/development/plugins/publishing.md
```diff
@@ -0,0 +1,100 @@
...
+
+Also, by default the Plugin Catalog shows only plugins marked as official in ArtifactHub, or allow-listed by Headlamp. Users can override this behavior and make Headlamp show all the plugins by switching off the "Only official plugins" settings for the Plugin Catalog. Read more at the [Plugin Catalog README](https://github.com/headlamp-k8s/plugins/tree/main/plugin-catalog#readme).
```
skoeva, line 45:
```suggestion
Also, by default the Plugin Catalog only shows plugins marked as official in ArtifactHub, or allow-listed by Headlamp. Users can override this behavior and make Headlamp show all the plugins by switching off the "Only official plugins" setting for the Plugin Catalog. Read more at the [Plugin Catalog README](https://github.com/headlamp-k8s/plugins/tree/main/plugin-catalog#readme).
```

docs/development/plugins/publishing.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Step 2: Build and Release Your Plugin on GitHub
+
+1. Move to your plugin’s folder, then run the following command:
+  ```bash
+  npm install
+  npm run build
+  healdamp-plugin package
+```
+It will create a tarball of the plugin in the format that is accepted by headlamp and also print the checksum of the tarball that we will use in the next step.
```
skoeva, line 35:
```suggestion
It will create a tarball of the plugin in a Headlamp-accepted format and also print the checksum of the tarball that we will use in the next step.
```

docs/development/plugins/publishing.md
```diff
@@ -0,0 +1,100 @@
+---
+title: Publishing Plugins
+sidebar_label: Publishing
+---
+
+Headlamp in its core does not have a way to make plugins discoverable or to list them to the user. However, the desktop version of Headlamp ships with the [Plugin Catalog](URL_FOR_PLUGIN) plugin which does provide a way for users to list Headlamp plugins from [ArtifactHub](https://artifacthub.io), install and remove them.
```
skoeva, line 6:
```suggestion
Headlamp at its core does not have a way to make plugins discoverable or to list them to the user. However, the desktop version of Headlamp ships with the [Plugin Catalog](URL_FOR_PLUGIN) plugin, which does provide a way for users to list Headlamp plugins from [ArtifactHub](https://artifacthub.io), install and remove them.
```

Issue #2564, opened 2024-11-13
Sorting by duration incorrectly sorts seconds vs minutes


rukas:
## Describe the bug
When sorting jobs by duration the UI incorrectly sorts a larger number of seconds to be greater than a smaller number of minutes. 

Ex. 59s is larger than 19m

## To Reproduce
Steps to reproduce the bug:
1. Go to Jobs
2. Run jobs that run various durations, including some that run for seconds and some that run for minutes
3. Sort by Duration (either ascending or descending)
4. See that it is incorrectly sorting the duration

## Environment (please provide info about your environment):
 - Installation type: Windows
 - Headlamp Version: 0.26.0
 - Other: n/a

## Are you able to fix this issue?
No.
> Do you want to help fix this issue? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional Context
![image](https://github.com/user-attachments/assets/1530e73d-f370-48e2-bb61-c7b44df366a0)

skoeva:
Thanks for submitting this, will take a look

PR #2566, opened 2024-11-13
frontend: Job: Add sort function for job durations

skoeva:
This change adds a sort function for job duration times. Previously, sorting by duration would show the list of jobs out of order, but this should no longer be the case.

Fixes: #2564 

### Testing
- [X] Open Headlamp with multiple jobs of varying durations
- [X] Sort by duration and ensure that the jobs are properly sorted

image

### DIFF
--- a/frontend/src/components/job/List.tsx
+++ b/frontend/src/components/job/List.tsx
@@ -111,14 +111,13 @@ export function JobsListRenderer(props: JobsListRendererProps) {
           id: 'duration',
           label: t('translation|Duration'),
           getValue: job => {
-            const startTime = job.status?.startTime;
-            const completionTime = job.status?.completionTime;
-            if (!!startTime && !!completionTime) {
-              const duration = new Date(completionTime).getTime() - new Date(startTime).getTime();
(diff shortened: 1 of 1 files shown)


frontend/src/components/job/List.tsx
```diff
@@ -111,14 +120,13 @@ export function JobsListRenderer(props: JobsListRendererProps) {
           id: 'duration',
           label: t('translation|Duration'),
           getValue: job => {
-            const startTime = job.status?.startTime;
-            const completionTime = job.status?.completionTime;
-            if (!!startTime && !!completionTime) {
-              const duration = new Date(completionTime).getTime() - new Date(startTime).getTime();
+            const duration = getJobDuration(job);
```
sniok, line 123: you could use a getDuration method in the job class.

```ts
const duration = job.getDuration();
```

skoeva: Thanks for pointing this out, didn't realize the logic was being duplicated in this file

sniok (approved):
Nice

PR #2567, opened 2024-11-13
[OLD] app: home: Add delete button for clusters

### LINKED ISSUE #1418: Allow to delete any clusters when using Headlamp (desktop version)
A few users have been confused about why they cannot delete clusters from Headlamp when in fact that's possible but we've restricted it to only clusters added from Headlamp itself. The reasons for this were that we didn't want to eventually let the user delete a cluster from their kubeconfig when thinking they're deleting it just from Headlamp. However, the reason why they cannot delete such clusters is not clear to users and not even practical according to some.

The way to fix this, IMO, is to allow users to delete any clusters, but making what they are doing very clear to them.
This would be:
- [ ] Show the origin of each cluster in the UI (with options like: `~/.kube/config`, `Headlamp`, `MyPluginName`)
- [ ] When deleting a cluster from Headlamp, we have a confirmation dialog that informs the user about where the cluster came from. And in the case of it coming from non-Headlamp origin (like the kubeconfig), we should reinforce that it will delete it from there too.

Alternatively, we could ask the user whether they want to delete non-Headlamp-origin clusters just from Headlamp (i.e. hide them), or also from their config file.

vyncent-t:
## Description

Fixes #1418 

This PR updates the cluster deletion functionality in Headlamp to allow users to delete non-dynamic clusters directly from the home page. Previously, only clusters added through Headlamp could be deleted, which led to confusion for users attempting to delete clusters originating from other sources (e.g., kubeconfig). 

To address this, the deletion process has been updated to clearly inform users about the impact of deleting a cluster, particularly when the cluster originates from non-Headlamp sources.

---

## Images
image

## Changes Made

1. Updated the cluster deletion functionality to allow deletion of non-dynamic clusters, including those originating from kubeconfig.
2. Enhanced the deletion confirmation dialog to inform users about the origin of the cluster and to clarify that deleting non-dynamic clusters will also remove them from the kubeconfig file.

## Steps to Test

1. Open the home view in Headlamp.
2. Add clusters from both Headlamp and kubeconfig.
3. Attempt to delete clusters:
   - Confirm the dialog accurately informs users about the cluster’s origin.
   - Verify that deleting non-dynamic clusters removes them from kubeconfig.
   - Verify that deleting Headlamp-added clusters functions as expected.

## Notes

- This PR ensures that users have a clear understanding of the impact of deleting clusters.
- It improves usability by addressing feedback received about the cluster deletion restrictions in earlier versions.


vyncent-t:
last push:

re worked the buttons components so that the app does not fail at main screen, repushed current WIP removing the clusterAPI changes that caused failures

working on better way to use delete cluster that deletes kubeconfig parts

backend/cmd/headlamp.go
```diff
@@ -1384,6 +1384,22 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	removeKubeConfig := r.URL.Query().Get("removeKubeConfig") == "true"
+	// removeKubeConfig := true 
+
+
+	if removeKubeConfig {  
+		// delete context from actual deafult kubecofig file  
+		// to do : replace the hard coding of this line with an actual path
+		err = kubeconfig.RemoveContextFromFile(name, filepath.Join("/home/vynty/.kube/config"))
```
vyncent-t, line 1394: This is the part here @yolossn

vyncent-t: was able to solve this

illume:
We have to be careful here with calling this "Delete cluster"

What it is doing more specifically is "deleting the cluster entry in the .kube/config file, and then any context entries that are related to that cluster". Not sure about the second part... but if it isn't it probably should do that. Because those contexts won't make sense anymore with the cluster gone that they were using.

joaquimrocha:
> We have to be careful here with calling this "Delete cluster"

Maybe we call it "remove"? Would that make a difference?

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -30,6 +35,44 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
...
+              </DialogContentText>
+              <Checkbox onChange={() => setCheckboxClicked(!checkboxClicked)} />
+            </Box>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={handleClose} color="primary">
+              {t('Cancel')}
+            </Button>
+            <Button disabled={!checkboxClicked} onClick={onConfirmationClicked} color="primary">
+              {t('I Agree')}
```
joaquimrocha, line 68: I know I asked for "I agree", but I didn't remember this was in the context of an existing ConfirmDialog.
The ConfirmDialog should work similar whether you have a checkbox or not. I think it's better to add the checkbox (when existing) to the original code, rather than having two different ways of rendering the dialog. And the buttons should remain with the original labels, unless we want to provide properties for those too.

joaquimrocha (changes requested):
please check also the CI checks.

frontend/src/components/common/ConfirmDialog.tsx
vyncent-t: Done! made some changes to the logic for the dialog which should allow us to just provide a description and then the smaller section of that description and the checkbox should render

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -9,7 +11,8 @@ import { DialogTitle } from './Dialog';
 
 export interface ConfirmDialogProps extends MuiDialogProps {
   title: string;
-  description: string;
+  description: string | JSX.Element;
```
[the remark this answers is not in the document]
vyncent-t: done!

vyncent-t:
last push fixes format issue

vyncent-t:
had to manually resolve merge conflict in this tab as the pushes I made after already resolving locally the normal way would not get read even after rebasing

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -34,21 +47,46 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
...
           <DialogContentText id="alert-dialog-description">{description}</DialogContentText>
+          {props.checkboxDescription && (
+            <Box
+              sx={{
+                display: 'flex',
+                alignItems: 'center',
+                marginTop: '10px',
+              }}
+            >
+              <DialogContentText id="alert-dialog-description">
```
illume, line 65: This id is not unique. Is it used for anything?

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,35 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
 	return setupErrors
 }
 
+// removeContextFromFile removes the context from the kubeconfig file.
+func removeContextFromFile(w http.ResponseWriter, contextName string) error {
```
illume, line 1381: Please add a test for this?

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -1,28 +1,41 @@
...
-import React, { ReactNode } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { DialogTitle } from './Dialog';
 
 export interface ConfirmDialogProps extends MuiDialogProps {
   title: string;
-  description: ReactNode;
+  description: string | React.ReactNode;
+  checkboxDescription?: string;
```
illume, line 15: Please document this field?

(If you can document the other ones whilst you're here... that would be nice to have).

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -34,21 +47,46 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
...
         open={open}
-        onClose={handleClose}
+        onClose={closeDialog}
         aria-labelledby="alert-dialog-title"
         aria-describedby="alert-dialog-description"
       >
         <DialogTitle id="alert-dialog-title">{title}</DialogTitle>
         <DialogContent ref={focusedRef}>
           <DialogContentText id="alert-dialog-description">{description}</DialogContentText>
+          {props.checkboxDescription && (
```
illume, line 57: Can you please add a story for the new states?

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -78,7 +78,8 @@ export async function setCluster(clusterReq: ClusterRequest) {
 // @todo: needs documenting.
 
 export async function deleteCluster(
-  cluster: string
+  cluster: string,
+  removeKubeConfig?: boolean
```
illume, line 82: Can you please document this?

backend/cmd/headlamp.go
illume, line 1381: removeContextFromFile could be removeContextFromDefaultKubeConfig? To be more descriptive. With just File it wasn't obvious that it was acting on the defaultKubeConfigPersistenceFile.

illume, line 1381: What happens when people have multiple kubeconfig files?

illume (commented):
Left some notes.

vyncent-t:
still very new to Go, let me know if is anything that I need to change

for context the linter did not like how long the deleteCluster function was starting to get, especially after I tried reworking it to include the multiple config file check (try to delete the right cluster by name)

the removeContextFromConfigs is meant to cover the multi config user scenario, which it should mostly just always run and find the one config file and then delete it from there as normal (if there are more than one it loops through them to find where the cluster context matches)

I did have some trouble trying to get the test to work for TestRemoveContextFromDefaultKubeConfig with changing where the defaultKubeConfigPersistenceFile was being used (I was originally trying to create a test that would allow us to see the non dynamic kube being deleted live but soon found it cannot be ran like that for the workflow on github so I went for a mock environment approach, this is done to kind of aim to recreate my local linux setup) as it only worked if I changed some env var to change the new dir for kubeconfigs but that also caused waterfall issues with other tests when running make backend-test which is why I am trying to mock it now

I am still trying to create a reasonable test for the removeContextFromConfigs function since it would also need another mock env created using the temp dir method so it can be a multi config use scenario, if I can make it small enough I will change the current kubeconfig_remove to just hold the config info we want to delete (the delta one) and have the kubeconfig_rename and kubeconfig_remove act as the multiple config set up in that temp test env)

- would also likely need to have that test find the name of the path where the contextName is found and check it again to see if that has been removed

also, side question for anyone who can answer, should I rewrite the multi config use test in a way to account for configs that are nearly identical? for example:

configA: has contexts named examplekube, deltakube

configB: has contexts named otherexamplekube, deltakube

should I further refactor the deleteCluster function to try to delete it by a more unique identifier? I want to avoid having a scenario where I want to delete deltakube from configB but since configA is ran first it deletes it there (unless its not possible to have the same name even on different config files) any thoughts? @illume

backend/cmd/headlamp.go
```diff
@@ -234,6 +234,18 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
 	}
 }
 
+// defaultKubeConfigFile returns the default path to the kubeconfig file.
+func defaultKubeConfigFile() (string, error) {
```
joaquimrocha, line 238: AFAIU, we have GetDefaultKubeConfigPath() in the config.go, we also have a way to retrive this from k8s client-go.

backend/cmd/headlamp.go
```diff
@@ -1384,6 +1396,30 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	removeKubeConfig := r.URL.Query().Get("removeKubeConfig") == "true"
+
+	if removeKubeConfig {
+		// delete context from actual default kubecofig file
+		kubeConfigFile, err := defaultKubeConfigFile()
```
joaquimrocha, line 1522: I think clusters can be loaded from multiple kubeconfig files, yet we only have a flag telling us whether the clusters from kubeconfig or dynamic. This is a limitation in our backend code (not your PR), but one that we have to handle before we consider this PR IMO.
Tied to this is the notion of what happens when we load a cluster with the same name from a different (or same) file.

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -1,28 +1,41 @@
...
 import DialogContentText from '@mui/material/DialogContentText';
-import React, { ReactNode } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { DialogTitle } from './Dialog';
 
 export interface ConfirmDialogProps extends MuiDialogProps {
   title: string;
-  description: ReactNode;
+  description: string | React.ReactNode;
```
joaquimrocha, line 20: You don't need to explicitly make it a string, as it's a type already covered by ReactNode.

frontend/src/components/common/ConfirmDialog.tsx
joaquimrocha: @vyncent-t I know we talked about adding this prop but in the end I don't think this is too specific a behavior to have in the component. Maybe we can pass a fragment that includes the description and the checkbox as part of the description. I will comment below why we don't need to make Agree insensitive depending on the check mark.

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -34,21 +47,46 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
...
+          >
             {t('No')}
           </Button>
-          <Button onClick={onConfirmationClicked} color="primary">
-            {t('Yes')}
-          </Button>
+          {props.checkboxDescription ? (
+            <Button disabled={!checkedChoice} onClick={onConfirmationClicked} color="primary">
+              {t('I Agree')}
+            </Button>
```
joaquimrocha, line 94: This behavior is leading to the following problem: if I want to remove the cluster from Headlamp but not delete it from the kubeconfig (which was the point of the checkbox), then I can not even remove the cluster Headlamp (as the agree button is disabled).

frontend/src/components/App/Home/index.tsx
```diff
@@ -105,18 +124,28 @@ function ContextMenu({ cluster }: { cluster: Cluster }) {
...
+          if (cluster.meta_data?.source !== 'dynamic_cluster') {
+            removeCluster(cluster, true);
+          } else {
+            removeCluster(cluster);
+          }
         }}
         title={t('translation|Delete Cluster')}
         description={t(
-          'translation|Are you sure you want to remove the cluster "{{ clusterName }}"?',
+          'translation|This action will delete cluster "{{ clusterName }}" from {{ source }}.',
```
joaquimrocha, line 140: Let's make this about the kubeconfig. If a plugin added the cluster, or we loaded it from a file directly, we just want to confirm that the user wants to remove the cluster. The removal from kubeconfig (i.e. non-Headlamp kubeconfigs) should be a configurable option for the user as a checkbox, and that's a very particular case.

backend/cmd/headlamp.go
```diff
@@ -1377,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
 	return setupErrors
 }
 
+// collectMultiConfigPaths looks at the default dynamic directory
```
joaquimrocha, line 1380: This commit should be merged with the previous one AFAIU.

frontend/src/i18n/locales/de/translation.json
```diff
@@ -146,8 +146,8 @@
   "Last Seen": "Zuletzt gesehen",
   "Offline": "Offline",
   "Lost connection to the cluster.": "",
-  "I Agree": "",
   "No": "Nein",
+  "I Agree": "",
```
joaquimrocha, line 150: Please merge with the commit that adds the strings.

joaquimrocha (changes requested):
Left some comments. I think that before this PR, for to make it much simpler, the config for the clusters in our backend needs to contain the kubeconfig if they are loaded from one. I know it's a different case/PR, but without that, these changes are complex and can contain corner cases that are unexpected.

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+// (e.g. ~/.config/Headlamp/kubeconfigs) and returns any files found there.
+// This is called from the 'else' block in deleteCluster().
+func (c *HeadlampConfig) collectMultiConfigPaths() ([]string, error) {
```
knrt10, line 1383: This should be moved to `backend/pkg/config/config.go`

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+	dynamicDir, err := defaultKubeConfigPersistenceDir()
+	if err != nil {
+		return nil, fmt.Errorf("getting default kubeconfig persistence dir: %w", err)
+	}
+
+	entries, err := os.ReadDir(dynamicDir)
```
knrt10, line 1389: Currently, the function reads all files from the directory without checking their extensions or format. Can you please add validation for known kubeconfig extensions?

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+	if err != nil {
+		return nil, fmt.Errorf("reading dynamic kubeconfig directory: %w", err)
+	}
+
+	var configPaths []string //nolint:prealloc
```
knrt10, line 1394: please remove this linter and allocate it earlier

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+
+		filePath := filepath.Join(dynamicDir, entry.Name())
+
+		configPaths = append(configPaths, filePath)
+	}
+
+	return configPaths, nil
+}
+
+func removeContextFromDefaultKubeConfig(
```
knrt10, line 1410: please move it to `backend/pkg/config/config.go`

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+	if len(configPaths) == 0 {
+		discoveredPath, err := defaultKubeConfigPersistenceFile()
+		if err != nil {
+			logger.Log(
+				logger.LevelError,
+				map[string]string{"cluster": contextName},
+				err,
+				"getting default kubeconfig persistence file",
+			)
+			http.Error(w, "getting default kubeconfig persistence file", http.StatusInternalServerError)
```
knrt10, line 1425: please do not `http.Error` here. This should be done at the top level function. This function should return only error message. This could lead to double error handling if the caller also tries to handle the error. Please remove this.

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+	}
+
+	// Hand off to a small helper function that handles multi-file iteration.
+	return removeContextFromConfigs(w, contextName, configPaths)
+}
+
+// removeContextFromConfigs does the real iteration over the configPaths.
+func removeContextFromConfigs(w http.ResponseWriter, contextName string, configPaths []string) error {
+	var removed bool
+
```
knrt10, line 1440: Please check whether contextName is empty or if configPath exists

```go
if contextName == "" {
    return fmt.Errorf("context name cannot be empty")
}
if len(configPaths) == 0 {
    return fmt.Errorf("no config paths provided")
}
```

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+			continue
+		}
+
+		logger.Log(logger.LevelError,
+			map[string]string{"cluster": contextName},
+			err, "removing cluster from kubeconfig",
+		)
+
+		http.Error(w, "removing cluster from kubeconfig", http.StatusInternalServerError)
```
knrt10, line 1478: again please remove this

backend/cmd/headlamp.go
```diff
@@ -1365,6 +1377,126 @@ func (c *HeadlampConfig) addContextsToStore(contexts []kubeconfig.Context, setup
...
+	if !removed {
+		e := fmt.Errorf("context %q not found in any provided kubeconfig file(s)", contextName)
+
+		logger.Log(
+			logger.LevelError,
+			map[string]string{"cluster": contextName},
+			e,
+			"context not found in any file",
+		)
+		http.Error(w, e.Error(), http.StatusBadRequest)
```
knrt10, line 1492: same

backend/cmd/headlamp_test.go
```diff
@@ -665,6 +665,34 @@ func TestRenameCluster(t *testing.T) {
 	}
 }
 
+func TestRemoveContextFromDefaultKubeConfig(t *testing.T) {
```
knrt10, line 668: This function should also be moved. Thanks

knrt10 (changes requested):
Nice work @vyncent-t. Some suggestions. Also, the code coverage is decreasing for backend, please add more tests for functions added.

backend/cmd/headlamp.go
vyncent-t: moved it to the `RemoveContextFromDefaultKubeConfig` where the function is being called

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp_test.go
vyncent-t: done!

backend/cmd/headlamp_testdata/kubeconfig_remove
```diff
@@ -0,0 +1,45 @@
...
...RGdnRVBBRENDQVFvQ2dnRUJBTnk3Ci9kREMxV0w3TXNSWGV2Z2tUQXkzcFZHMVVLa1VQeXd4cS9ETHBPdmRzQmloQjZoVmN1bWNZUTkzYUxLbERzSXMKR0Q0QUJkUFM4cEFPMzhMb3RBWWVDeDIwcDFPem9LYVMvVkp6ZlJKQWVUSStCY3dzRjh2U1VXYU0reWZ4STBPUgpnalE0OVR0eUppYURyS2tzbnd4R3Y0K0U3aWFhZUVPMG55U01EcnpON1RvYkVyb1pObHRzNkdMN2tpTDB0TG5ZCkorNnNtSHlhSGh6WThaR0JZMFdWUXpzNENFMnJ0Q1k5eTV4N2F3bUlDUWE2anBXVFVQazNqa0RMcU93bEQyRmMKcHNkeXI4a1Z3UUhTUUVnRkg2Yzgwdnp3Ny9RSUVDdGRYNlZRRnE1bzYzOWlvc3hQcXVKV3ZtMGVjdkx5dC81cApxNXZpNzMxWThEb0VDMjFtS2NzQ0F3RUFBYU5aTUZjd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0hRWURWUjBPQkJZRUZJNFFkU2FSRFVodi8wWjk0ZzV5RmlVdWlMZHBNQlVHQTFVZEVRUU8KTUF5Q0NtdDFZbVZ5Ym1WMFpYTXdEUVlKS29aSWh2Y05BUUVMQlFBRGdnRUJBQ3pWWUpBUzQ1UFBOSFVSaDJKWQpKWDFycmFMdGNTbzVuNG1DVy9oeE5YdHpCMlIzWkhPU0hnNmF2R3JNeFY4ZlpCdmtBdEJFaUYzM2JvRThzZzVhCjhhWHRFTjR5bzlZQ2FZc2ZXK2tNNlZDRUdtVWd5bm13aXltYTBzSW5USlZ1R3ZVbDVucVhjUHJJdW9OTVVrTUwKdCsrckxCb0NwY2xrN09VSTA0dXZvanpxc2hsQ0JiMURSOXRwT0s0Kys0UGdPait6OXZ0N3g0dzhMYlhvQmtvegozOEJyVEoyQ3NqbU0xS2ZqZXlpNWdHVmFjeE9YSXRjbXprNzRpQzZ0SjdqVm1MVmNacEc5ZElvcFk5WTBaTkQ0CmQzZjlmOGdCWkJzaXA0a3gxMmFxMlJ5dzFYNGVOaFY2dW5OaCtHVHNhNlFDSlJ0Zk9FK1Q4Njd2ZHlPbjZMb2wKYWQ4PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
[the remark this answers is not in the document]
vyncent-t: removed and using the other test data

backend/cmd/headlamp_testdata/kubeconfig_remove
```diff
@@ -0,0 +1,45 @@
...
...RUJBUVVBQTRJQkR3QXdnZ0VLQW9JQkFRQzRLUmwrS0lsN0NJYVgKbzIwYjdBOVEvaURDbUN6dWFXMSs3WEJyelhiZHNmNkNaRzhVMWZYWTdVWXl3bXVhYkZldUFrUHRBT1hyWVg0YQpMTGZtWTRvdkZYc1RQWmtPUktJeWRFUmNnLy9hOStPd3d2c1ZCUUp4NFplbUtrN1NzaFYxcjl3WGVqVnJIUkFOCm5xQ3JIQVhFNHA5bmFKZHNkTXIyQWdDa0VIK01tTFNqTExNL1lWcnExdmJpRWRtUVFSWHduVnFwcmNyRXBIQzUKWWJJenl4cVZRWWZIZVdWc2N0SUxFeVdPMFQwMS9tYkZ2RVY4QW9BL3phekIycjF3Y0VaeUNSRXFXbExrS2RXTwpNYmU1WnlwMDNhQzlBOSs4cThQNFBEOUNnVXlrcVovN0xydGlja2k0TVBsK2VmaGFlUk9YSEJMSURuQmplTHJkCmJGdHpaOVhKQWdNQkFBR2pWakJVTUE0R0ExVWREd0VCL3dRRUF3SUZvREFUQmdOVkhTVUVEREFLQmdnckJnRUYKQlFjREFqQU1CZ05WSFJNQkFmOEVBakFBTUI4R0ExVWRJd1FZTUJhQUZJNFFkU2FSRFVodi8wWjk0ZzV5RmlVdQppTGRwTUEwR0NTcUdTSWIzRFFFQkN3VUFBNElCQVFBRjNvaTFvNVlNM1UvOWxPRElhaUpmaGllNzdieG1pN3NwCitiL0NEOGRCTXhIdWpPVnBSaTFNaHRJa2U3N2U1RVVuZEFGRzYvQTQwK3c2TGtCYXJFUEl5R2daRlBvZkttcSsKRGlIMGxPZHBYY0hFd3laTjhWSmdRd0JKUkhKcDhBc0p3TGFYWGplU1FQdmZyeHhLdUFGenRzeXNaYlBMUkxoYQpjeXZmeDNwTE91ZVJ4MDJqQVZUUlNJUGNPZEV4SERPa0FGWFFCdDV4TFo2eGFKTU1VQjZXNUYwcVpPelFuVUZsCk80QUNNOEhnOEdKc2xqLzFqZnpZaGlneWdwL2psQ0Jkd1Izb2c1ZXFqaC9ZRzlxWHVsU2Z0WUNhMURaOEp2QnAKaGRSYzZxOVM0ZFdtRW9zMmkxTDA1WUs3ZFBaQk5JVHRLNkVzQS9CRCs0VlVWRHczZldkNQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/components/common/ConfirmDialog.tsx
vyncent-t: added some changes that allow for us to delete the non dynamic cluster from headlamp by default

the changes also include the option to check the box for the non dynamic cluster to be deleted from the source

frontend/src/components/common/ConfirmDialog.tsx
vyncent-t: done!

backend/cmd/headlamp.go

illume:
@vyncent-t This is marked as OLD in the PR title, and I see another PR linked to the issue. 

So I'll close this one, but please re-open it if you plan to continue it?

PR #2570, opened 2024-11-14
backend: Fix reloading of clusters in kubeconfig

### LINKED ISSUE #2541: Contexts removed from kubeconfig is not updated
## Describe the bug
Headlamp watches for changes in the kubeconfigs to keep the contexts in sync with the ones in the kubeconfig files. When a context is added it is reflecting in the UI but when a cluster is removed it is still showing up.

## To Reproduce
Steps to reproduce the bug:
1. Start Headlamp
2. Add a new context to the kubeconfig file and refresh headlamp
3. The context shows up in the UI
4. Remove out a context from the kubeconfig file and refresh headlamp
5. The removed context still shows up in the UI

knrt10:
When a context was being added and removed from KUBECONFIG file, it was still visible in UI. This removes the cluster from store and returns correct list of clusters.

Fixes #2541

### DIFF
--- a/backend/pkg/kubeconfig/watcher.go
+++ b/backend/pkg/kubeconfig/watcher.go
@@ -1,6 +1,7 @@
 package kubeconfig
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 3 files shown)


backend/pkg/kubeconfig/watcher.go
```diff
@@ -106,3 +106,46 @@ func addFilesToWatcher(watcher *fsnotify.Watcher, paths []string) {
...
+			if existingCtx.Name == newCtx.Name {
+				found = true
+
+				break
+			}
+		}
+
+		if !found {
+			err := kubeConfigStore.RemoveContext(existingCtx.Name)
+			if err != nil {
```
yolossn, line 138: This solution assumes that all contexts are loaded from the file that changed now only. There can be contexts from multiple files and different sources.

knrt10: I think even if they are from multiple store they would still be present in `kubeConfigStore` right? Line `119` does exactly the same. Do you think that even if there are other sources, they won't be present in the store?

yolossn: 119 is listing all the contexts. If you see the code below there are three different types of sources from which we load the clusters from. Here you are listing(line 119) and removing all the contexts(line 137). We should remove only the clusters from kubeconfig source because that is what is triggering the sync here and they are only loaded in line 145.

https://github.com/headlamp-k8s/headlamp/blob/8fb46f63c2a51344478be29656ed36c928962f6d/backend/pkg/kubeconfig/kubeconfig.go#L28-L32

knrt10: Thank you for the explanation, updated it.

yolossn (approved):
LGTM

illume (approved):
👍 thanks!!

PR #2573, opened 2024-11-15
headlamp-build: Add package.json with scripts from Makefile

illume:
This is to simplify using Headlamp so that people do not need to
have installed gnu Make. Which isn't on Windows, and isn't on
some Mac/unix systems.

This just starts with backend and frontend targets because
those are the ones that will be used in Contrib fest session.

This package is not intended to be published, and so the name is
not important at this stage.


sniok:
nice! since this is just a beginning I'd like to suggest a convention for package script names which I've seen in other projects
where ":" character separates groups and "-" separates rest, like this:

```
frontend:run
frontend:install-ci
frontend:coverage
backend:run
backend:coverage-html
etc..
```

another suggestion is to use `npm run --prefix frontend ...` instead of `cd frontend && npm run ...` 

also I'm not sure if it's just me but I'd like to avoid duplicated commands like `run backend` and `backend run`. I know this might be a convenient shortcut for people that already used to one way or the other but I think having just one variant might be cleaner

docs/development/backend.md
```diff
@@ -18,33 +18,33 @@ redirects the requests to the defined proxies.
 The backend (Headlamp's server) can be quickly built using:
 
 ```bash
-make backend
+npm run backend
```
joaquimrocha, line 21: I think the `run backend`, even though matching what we had is more confusing, since this is not running backend but building it. Maybe `npm run build:backend` and `npm run backend` would actually try to run the backend. WDYT?

joaquimrocha (changes requested):
I thought it was simpler but I think we need to rethink what the scripts should be called rather than having a match to the make targets we had before.

docs/development/backend.md
illume: I will change `npm run backend` so it runs the backend. So it is the same as `npm run backend-run`. Same thing with `npm run frontend`.

I don't think `npm run build:backend` is a good idea to do or discuss now.

PR #2574, opened 2024-11-15
frontend: Add global search

sniok:
Global search in top bar

Features:

- Find resources
- Find app pages
- Find clusters
- Custom actions (switching theme)
- Remembers recently used results
- Highlights query text in the output
- Lazy loads component code and resources
- Minimal performance impact when not used (when not used it will display just a placeholder input)

demo

https://github.com/user-attachments/assets/a03eb7b2-8162-4dea-896a-b0850d108a38

testing done

 - a11y: axe and edge a11y quick checks
 - narrow screens
 - slow network


joaquimrocha:
This looks and feels (I tested it) pretty awesome! Thanks for implementing it.
At KubeCon, Rene and I talked to some folks from the k8s sig-ui about how we implement things and they were surprised we hold most things in the frontend, and that for dealing with many resources + search, they would recommend having some backend caching. I am mentioning this so we keep it in mind for the search too.

A couple of things I noticed:
1. The search suggests showing pages like the Pods (maybe I have been to them), even when we are at the home view (we need to turn those off if we are not in a cluster view)
2. Changing routes takes too long (at least in dev mode) that it may feel like the jump was not activated. Maybe we could add a spinner to the search or somewhere else (besides speeding up the routes changes).
3. No matching with the Home view (so we can go there without having to use the mouse)

In the future we should also make sure that the plugin defined routes are indexed; and add the ability to specify how to search resources (by doing the json-path matching, or an easier subset of it).

Good job!

sniok:
> The search suggests showing pages like the Pods (maybe I have been to them), even when we are at the home view (we need to turn those off if we are not in a cluster view)

That was fixed

> Changing routes takes too long (at least in dev mode) that it may feel like the jump was not activated. Maybe we could add a spinner to the search or somewhere else (besides speeding up the routes changes).

It's a bit unrelated to search but we can definitely improve navigation perf, although it's not as bad in prod builds.

> No matching with the Home view (so we can go there without having to use the mouse)

That page is called "Choose a cluster". I'd change it to Home but in a different PR.

> In the future we should also make sure that the plugin defined routes are indexed;

It now uses routes from the redux slice as well, so plugin routes will show up

> and add the ability to specify how to search resources (by doing the json-path matching, or an easier subset of it).

I've looked into it, for this feature we need an expression parser/runner. This would be good place to use same expressions as kubernetes does https://kubernetes.io/docs/reference/using-api/cel/ https://cel.dev/. But it lacks javascript implementation, the go implementation compiled to wasm is huge and passing objects to it is inefficient (I tried a small poc). Maybe there's another way to use it, ideally in apiserver itself. But for now keeping it simple.

---

Updated the look a bit, after @joaquimrocha suggestion it now uses forward slash as a shortcut (just like a certain source version control hosting website)

<details>

image

</details>

---

Worked on support for narrow screens, it switches to button and then on activation takes up the whole top bar

<details>

image

image

</details>

---

For slow networks it will display a spinner when component is being lazy loaded.
Also it will pass the value from the placeholder to the actual search for smooth experience.
Icons are now also lazy loaded.

<details>

https://github.com/user-attachments/assets/b7c3c0d9-ebb7-4d41-b98f-0a33c60dd296

In this demo I type "pods" in the placeholder input while the search component is loading. Once it has loaded it immediately shows results for "pods"

</details>

skoeva (commented):
Looks good overall, big thanks for this!

Just want to echo the small screen issue, it seems that on mobile the search bar is completely hidden. Maybe a search icon in the top bar to trigger the search bar would work here?

image

and maybe for screens that are a bit wider but not wide enough, I think the size of bar could maybe be tweaked. Either something a bit longer, or like with the smaller screen, trigger the search bar with an icon

image

joaquimrocha (changes requested):
Working great and code LGTM but I think we need to add the search also in mobile views.
I'd add an action button to the top bar to the left of the tripple dot menu, with a magnifying glass + when clicked it opens the popover to most of the width + height so the user can easily search and scroll through results.

sniok:
Added search to the mobile top bar, also adjusted how max width works so it fits narrow screens.
also removed an unwanted behavior where tapping an empty focused input would close it (it was coming from mui's useAutocomplete)

https://github.com/user-attachments/assets/63bf5983-522a-4c7e-8b83-7d04d170d8db

sniok:
Updated styles a bit so there's some spacing between search on all screens. 

> might be an issue with detecting when to switch between the expanded search bar and a condensed one:

I'm using an 'md' breakpoint here, I'd rather not add another breakpoint, and also it won't be perfect anyway, some plugins add stuff to the top bar, we'll probably also change items up there so I think this is okay

joaquimrocha (approved):

joaquimrocha (approved):

Issue #2535, opened 2024-11-06
Plugins version in the UI is not matching the actual versions


joaquimrocha:
## Describe the bug
I have noticed that the plugins' versions in the UI do not match the versions coming from their package.json .
See screenshot:
![Screenshot showing the prob](https://github.com/user-attachments/assets/c6d1c8ea-9de4-4896-9db6-62a50a598610)

## To Reproduce
Steps to reproduce the bug:
1. Go to Settings > Plugins
2. Notice all the shipped plugins versions there
3. Match these versions against the ones in the package.json files (filter by them in the Network tab of devtools) -> Sometimes the versions do not match

## Environment (please provide info about your environment):
 - Installation type: Linux/Mac/Windows
 - Headlamp Version: 0.26.0

## Additional Context
Reproduced in different platforms and the non-match is not consistent (i.e. it's the wrong version but not the same number across my machines)

feym78:
Same for in-cluster installation, inside pod there is version 2.4.0 for prometheus in /headlamp/static-plugins/prometheus/package.json, but in UI it's 0.1.0.

feym78:
> Same for in-cluster installation, inside pod (v0.26.0) there is version 2.4.0 for prometheus in /headlamp/static-plugins/prometheus/package.json, but in UI it's 0.1.0.

Strange thing.
After celaring cookies plugin version become 0.2.0, but stil empty settings screen fot it, BUT in private mode plugin version shows as 0.2.4 and settings page work.

joaquimrocha:
Reopening for exploring the reason behind storing plugins' package.json, see my comment at: https://github.com/headlamp-k8s/headlamp/pull/2578#issuecomment-2491732968

viceice:
Another issue is that the `package.json` and `main.js` plugion files are cached in browser, so the browser loads the old plugins even if there is a newer version on disk.

viceice:
still an issue 😞

viceice:
it's still an issue with cluster integrated mode

PR #2578, opened 2024-11-15
frontend: plugin: Sync plugin version with backend

skoeva:
This change allows the frontend to update plugin fields from the backend, avoiding the case where a plugin is updated but the version number does not reflect this.

Fixes: #2535 

#### Using the prometheus plugin as an example:
- [X] Open Headlamp and navigate to the plugin settings. Ensure the latest prometheus plugin version is installed (currently 0.2.2)
- [X] Downgrade the prometheus plugin version by changing the version in `package.json` to 0.0.1 and run `rm -rf node_modules package-lock.json dist` in the prometheus folder
- [X] Install the prometheus plugin at this downgraded version: `npm install prometheus@0.0.1`
- [X] Refresh the page and ensure that the prometheus plugin version now reads as `0.0.1` without being disabled

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -221,7 +221,10 @@ export function updateSettingsPackages(
         isEnabled: true,
       };
     }
-    return settingsPlugins[index];
+    return {
+      ...settingsPlugins[index],
+      ...plugin,
+    };
   });
 }
 


illume (approved):
👍 RAD

joaquimrocha:
I think this PR doesn't address why the versions were being wrongly displayed. The cause is that we are storing the package.json's contents in the local storage, but I don't understand why we need to do so.
If we have the plugins' info being loaded from the backend, why do we need to store it too in the local storage?

PR #2590, opened 2024-11-21
frontend: Small performance improvements

sniok:
This PR adds some useMemo, memo, useCallbacks to avoid unnecessary rerenders. Also hides a create resource popup when it's not open


frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -112,7 +112,7 @@ function SidebarToggleButton() {
   );
 }
 
-function DefaultLinkArea(props: { sidebarName: string; isOpen: boolean }) {
+const DefaultLinkArea = memo((props: { sidebarName: string; isOpen: boolean }) => {
```
joaquimrocha, line 115: Is this really needed? I thought React already only rendered components only if the props change, and there's no complex work in the body of the component AFAICS.

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -186,6 +186,15 @@ export default function Sidebar() {
...
   const search = namespaces.size !== 0 ? `?namespace=${[...namespaces].join('+')}` : '';
 
+  const handleToggleOpen = useCallback(() => {
+    dispatch(setWhetherSidebarOpen(!sidebar.isSidebarOpen));
+  }, [sidebar.isSidebarOpen]);
+
+  const linkArea = useMemo(
+    () => <DefaultLinkArea sidebarName={sidebar.selected.sidebar || ''} isOpen={isOpen} />,
+    [sidebar.selected.sidebar, isOpen]
+  );
```
joaquimrocha, line 196: If we are using a memo here, do we need the memo around the DefaultLinkArea?

joaquimrocha (commented):
I am a bit surprised we need `memo(...)` surrounding a bunch of components. I thought this was only needed if we were doing a lot of complex work in components' bodies, and even those could be abstracted into useCallback + useEffect to avoid this.

frontend/src/components/Sidebar/Sidebar.tsx
sniok: React rerenders component if the parent updates, even if the props are the same. https://react.dev/reference/react/memo#skipping-re-rendering-when-props-are-unchanged

Since the props here are super simple, comparing if props are changed is not an expensive operation

frontend/src/components/Sidebar/Sidebar.tsx
sniok: linkArea is passed as a prop, without this it triggers rerender for PureSidebar 

image

And the DefaultLinkArea component is memoized to avoid rerenders inside PureSidebar for unrelated rerenders
(on the screenshot DefaultLinkArea is _c3 and memoized)
image

frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -147,39 +147,41 @@ export default function CreateButton(props: CreateButtonProps) {
...
-                      </MenuItem>
-                    ))}
-                  </Select>
-                </FormControl>,
-              ]
-            : []
-        }
-      />
+      {openDialog && (
+        <EditorDialog
```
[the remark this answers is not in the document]
sniok: yes we should, but let's handle it in a different PR
there are other places already that use the same pattern as here (EditButton.tsx:117 for example) 
for now this is the simplest fix but if we're going to modify the component itself then I'd want to make EditorDialog lazy loaded as well

sniok:
> I am a bit surprised we need memo(...) surrounding a bunch of components. I thought this was only needed if we were doing a lot of complex work in components' bodies, and even those could be abstracted into useCallback + useEffect to avoid this.

Let's look at what is happening on the main branch, without optimizations. This PR is mostly focused on the Sidebar and navigating between pages.

The test case is clicking on Pods from Deployments page. I've commented out the Table so it's clearer.

image

What we see here is the first rerender that happens after a click. The trigger is the change of current route and the whole sidebar rerenders. What user sees is Pods button turns yellow and Deployment button turns back to black but we're rerendering every item.

You might think that rerendering the same thing in react is not a big deal and in most cases it is not. But what we see is that single rerender takes 70ms (to maintain 60 fps you need to have less than 16ms per frame)

Let's see what code actually runs during this without the react profiler overhead. Here I'm using the Chrome Profiler.

image

On the top is the flame graph that represents CPU work over time and on the bottom is the list of functions that take the most time.

6 out of 10 slowest functions are from mui/emotion styling packages dealing with generating styles in runtime. This can be somewhat improved by getting rid of sx usage everywhere and using separate `styled(Box)` component for example. 

Style recalculation is taking a big chunk here because certain mui components use functions like `getBoundingClientRect` during render which makes it slow.

So this is not React's fault. Aggresive memoization will help us deal with runtime style generation and heavy mui components.

joaquimrocha:
Thanks a lot for the explanation. Appreciate that. I am willing to merge once the conflicts are solved.

sniok:
Merge conflicts were fixed

sniok:
> I have rebased these changes but noticed there's a bug related to my previous concerns: if you add content to the creation dialog (from the create button), then you apply and it fails to apply, or you close and reopen that dialog, it will no longer retain the contents that the user has added. This is a UX problem.
> 
> So to fix this while keeping your goal of not rendering unless necessary, I added a new commit which only renders the dialog after it's first opened (unless the keepMounted prop is passed, as maybe some plugins do want to pre-render it). I have reverted your commit for the CreateButton, so please test and if it makes sense we just remove the original commit.

There's a simpler fix for that, we can preserve the state of the EditorDialog but also avoid rendering the Dialog by reusing the `open` prop and returning only inside EditorDialog (previously it would not render EditorDialog completely which would cause the state to be lost). I've tried the scenario you've described and it keeps the content of that popup. 

I've dropped the initial commit and added a new commit with this fix.

PR #2593, opened 2024-11-21
frontend GraphView: Fix duplicate edges on the map and add unit test

sniok:
This is a regression caused by some previous optimizations. Added a unit test that would fail without this fix


skoeva (approved):
nice

knrt10 (approved):
+1

Issue #2595, opened 2024-11-21
Plugin settings info handling rework


vyncent-t:
## Describe the bug
Currently there are some information issues between the storage of the plugin settings and the package.json data. 

I.e. when the plugin is not running but there has been changes (either description or the version) the information does not update

I am working on a rework of the local storage and redux slice usage to work with this PR https://github.com/headlamp-k8s/plugins/pull/117

and these related issues:

#2578 

## To Reproduce
Steps to reproduce the bug:
1. Go to settings
2. Open the package.json of a plugin that is installed and change a text string
3. View the field changed and it should update (it does not without it running)

## Are you able to fix this issue?
I cam currently working on a fix

## Additional Context
[If needed, add any other info or screenshots about the bug here.]
[Is it a regression? What version of Headlamp did not have this bug?]

PR #2596, opened 2024-11-21
OLD - frontend: PluginSettings: Rework plugin settings local storage usage

vyncent-t:
## Description

resolves issue #2595

This PR refactors the way Headlamp handles plugin settings stored in local storage, focusing on simplifying and improving the management of plugin states. Previously, the project stored the entire JSON data of plugin information in `headlampPluginSettings`, which included unnecessary details. This refactor reduces complexity by saving only the essential state information, improving performance and maintainability.

### Key Changes

1. **Simplified Storage**:
   - Replaced the old `headlampPluginSettings` structure with `enabledPluginsList` in local storage.
   - The new structure saves only the plugin name and its `on/off` state, eliminating the need to store complete JSON data.

2. **State Management Updates**:
   - Introduced the `pluginData` slice in the plugin state to store plugin information.
   - Added `enabledPlugins` to maintain a list of plugins currently in use.

3. **Backward Compatibility**:
   - The project now detects the existence of the old `headlampPluginSettings` in local storage.
   - If detected, it converts and migrates the old settings into the new `enabledPluginsList` format automatically upon visiting the plugin settings page.

---

## Steps to Test

To verify the changes and ensure backward compatibility:

1. **Setup**:
   - Start on the `main` branch with multiple plugins installed.
   - Disable one plugin while keeping others enabled, and save the settings.
   - Open the browser’s local storage and confirm that the disabled plugin’s state is saved in the `headlampPluginSettings` JSON.

2. **Switch to this Branch**:
   - Checkout this branch and navigate to the plugin settings page.

3. **Observe Migration**:
   - Confirm that the page detects the `headlampPluginSettings` in local storage, migrates the settings into a new `enabledPluginsList` object, deletes the old `headlampPluginSettings`, and reloads the page.
   - Verify that the plugin states from the previous settings are preserved in the new format.

4. **Post-Migration**:
   - Ensure that plugin settings updates are correctly saved in the new `enabledPluginsList` structure.
   - Confirm that enabling/disabling plugins reflects accurately on subsequent reloads.

---

## Notes

- It improves the performance and clarity of plugin state management by reducing storage overhead.
- The migration process ensures a seamless transition for users upgrading from previous versions.

### DIFF
--- a/frontend/src/components/App/PluginSettings/PluginSettings.stories.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettings.stories.tsx
@@ -1,4 +1,5 @@
 import { Meta, StoryFn } from '@storybook/react';
+import { PluginInfo } from '../../../plugin/pluginsSlice';
 import { TestContext } from '../../../test';
 import { PluginSettingsPure, PluginSettingsPureProps } from './PluginSettings';
 
@@ -41,6 +42,18 @@ function createDemoData(arrSize: number, useHomepage?: boolean) {
 }
 
 /**
+ * createDemoEnabledList function will create a list of plugin objects with a boolean value to enable/disable the plugin.
+ * The function will return an object of plugin names with a boolean value.
+ */
+function createDemoEnabledList(arr: PluginInfo[]): Record<string, boolean> {
+  const enabledList = arr.reduce((acc, p) => {
+    acc[p.name] = !!p.isEnabled;
+    return acc;
+  }, {} as Record<string, boolean>);
+  return enabledList;
+}
+
+/**
  * Creation of data arrays ranging from 0 to 50 to demo state of empty, few, many, and large numbers of data objects.
  * NOTE: The numbers used are up to the users preference.
  */
@@ -55,6 +68,7 @@ const demoEmpty = createDemoData(0);
 export const FewItems = Template.bind({});
 FewItems.args = {
   plugins: demoFew,
+  pluginsEnabledMap: createDemoEnabledList(demoFew),
   onSave: plugins => {
     console.log('demo few', plugins);
   },
@@ -63,12 +77,14 @@ FewItems.args = {
 export const Empty = Template.bind({});
 Empty.args = {
   plugins: demoEmpty,
+  pluginsEnabledMap: createDemoEnabledList(demoEmpty),
 };
 
 /** NOTE: The save button will load by default on plugin page regardless of data */
 export const DefaultSaveEnable = Template.bind({});
 DefaultSaveEnable.args = {
   plugins: demoFewSaveEnable,
+  pluginsEnabledMap: createDemoEnabledList(demoFewSaveEnable),
   onSave: plugins => {
     console.log('demo few', plugins);
   },
@@ -78,6 +94,7 @@ DefaultSaveEnable.args = {
 export const ManyItems = Template.bind({});
 ManyItems.args = {
   plugins: demoMany,
+  pluginsEnabledMap: createDemoEnabledList(demoMany),
   onSave: plugins => {
     console.log('demo many', plugins);
   },
@@ -86,6 +103,7 @@ ManyItems.args = {
 export const MoreItems = Template.bind({});
 MoreItems.args = {
   plugins: demoMore,
+  pluginsEnabledMap: createDemoEnabledList(demoMore),
   onSave: plugins => {
     console.log('demo more', plugins);
   },
@@ -94,6 +112,7 @@ MoreItems.args = {
 export const EmptyHomepageItems = Template.bind({});
 EmptyHomepageItems.args = {
   plugins: demoHomepageEmpty,
+  pluginsEnabledMap: createDemoEnabledList(demoHomepageEmpty),
   onSave: (plugins: any) => {
     console.log('Empty Homepage', plugins);
   },
--- a/frontend/src/components/App/PluginSettings/PluginSettings.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettings.tsx
@@ -3,12 +3,12 @@ import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import Link from '@mui/material/Link';
 import { MRT_Row } from 'material-react-table';
-import { useEffect, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import helpers from '../../../helpers';
 import { useFilterFunc } from '../../../lib/util';
-import { PluginInfo, reloadPage, setPluginSettings } from '../../../plugin/pluginsSlice';
+import { PluginInfo, reloadPage, setEnablePlugin } from '../../../plugin/pluginsSlice';
 import { useTypedSelector } from '../../../redux/reducers/reducers';
 import { Link as HeadlampLink, SectionBox, Table } from '../../common';
 import SectionFilterHeader from '../../common/SectionFilterHeader';
@@ -24,6 +24,7 @@ import SectionFilterHeader from '../../common/SectionFilterHeader';
  */
 export interface PluginSettingsPureProps {
   plugins: PluginInfo[];
+  pluginsEnabledMap: Record<string, boolean>;
   onSave: (plugins: PluginInfo[]) => void;
   saveAlwaysEnable?: boolean;
 }
@@ -92,20 +93,15 @@ const EnableSwitch = (props: SwitchProps) => {
 
 /** PluginSettingsPure is the main component to where we render the plugin data. */
 export function PluginSettingsPure(props: PluginSettingsPureProps) {
+  const dispatch = useDispatch();
+
   const { t } = useTranslation(['translation']);
 
-  /** Plugin arr to be rendered to the page from prop data */
-  const pluginArr: any = props.plugins ? props.plugins : [];
-
-  /** enableSave state enables the save button when changes are made to the plugin list */
-  const [enableSave, setEnableSave] = useState(false);
-
   /**
-   * pluginChanges state is the array of plugin data and any current changes made by the user to a plugin's "Enable" field via toggler.
-   * The name and origin fields are split for consistency.
+   * Copy of plugins with pending changes
    */
-  const [pluginChanges, setPluginChanges] = useState(() =>
-    pluginArr.map((plugin: PluginInfo) => {
+  const [pluginChanges, setPluginChanges] = useState<PluginInfo[]>(
+    props.plugins.map((plugin: PluginInfo) => {
       const [author, name] = plugin.name.includes('@')
         ? plugin.name.split(/\/(.+)/)
         : [null, plugin.name];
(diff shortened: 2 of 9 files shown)


vyncent-t:
last push

- allows old storage settings to be migrated to the new settings

vyncent-t:
last push fixes merge conflicts

vyncent-t:
last push rebases and fixes storybook

joaquimrocha (commented):
I am a bit surprised this needs so many changes. Wasn't the idea just to remove the saving of the whole package.json from the settings and use just the plugin name instead? Even if we change the name of the key in local storage, or move it to index db, why are we touching the plugin settings this much?

I would also like to understand: are plugins now supposed to be disabled by default? I understood that shipped plugins need to be enabled by default, and that the other ones were also like that for keeping a consistent behavior with the previous versions.

vyncent-t:
noticed some weird behavior when not using any of the old `headlampPluginSettings` or `enabledPluginsList` in local storage, working on a fix again

vyncent-t:
some meeting notes

- can keep the enabledPluginsList to handle toggles
- may need to handle default loading state of a plugin if they come disabled or enabled but most come enabled as of current
- handled the backwards compatibility for users on pervious vers that allow migration of old settings to new settings

vyncent-t:
attempting to use the npm link to try to work on the slide plugin changes for updated state

frontend/src/plugin/pluginsSlice.ts
```diff
@@ -90,14 +90,17 @@ export type PluginInfo = {
 export interface PluginsState {
   /** Have plugins finished executing? */
   loaded: boolean;
+  /** Information for the plugin's enable */
```
sniok, line 93: a better comment would be something like `Map where key is plugin's name and value is whethere it is enabled`

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +278,69 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
...
-  const pluginSettings = useTypedSelector(state => state.plugins.pluginSettings);
+  const pluginData = useTypedSelector(state => state.plugins.pluginData);
+
+  const oldLocalEnabledList = localStorage.getItem('headlampPluginSettings');
+  const localEnabledList = localStorage.getItem('enabledPluginsList');
+
+  let pluginsEnabledList;
+
+  /**
+   * Need to improve later, this is a temporary fix to handle the migration of the old plugin settings.
```
sniok, line 289: what exactly needs to be improved here? please mark it as todo and add information on when this needs to be done (if you can)

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +278,69 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
 export default function PluginSettings() {
   const dispatch = useDispatch();
 
-  const pluginSettings = useTypedSelector(state => state.plugins.pluginSettings);
+  const pluginData = useTypedSelector(state => state.plugins.pluginData);
```
sniok, line 281: all this logic should probably be moved to a useEffect

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +278,69 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
...
+  let pluginsEnabledList;
+
+  /**
+   * Need to improve later, this is a temporary fix to handle the migration of the old plugin settings.
+   */
+
+  if (oldLocalEnabledList) {
+    const oldSettings = JSON.parse(oldLocalEnabledList);
+
+    const newSettings = oldSettings.map((setting: { name: string; isEnabled: boolean }) => {
```
sniok, line 295: this `.map` doesn't seem to do that much? you could use oldSettings in the reduce directly

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +278,69 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
...
+      return {
+        name: setting.name,
+        isEnabled: setting.isEnabled,
+      };
+    });
+
+    pluginsEnabledList = newSettings.reduce(
```
sniok, line 302: this name is a bit misleading, it's not really a list

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -92,20 +93,17 @@ const EnableSwitch = (props: SwitchProps) => {
...
-   * pluginChanges state is the array of plugin data and any current changes made by the user to a plugin's "Enable" field via toggler.
-   * The name and origin fields are split for consistency.
+  /** Plugin arr to be rendered to the page from prop data
+   * Note:
+   * The pluginArr is the main array that will be used to render the plugin data to the page.
+   * Any changes made here is what will be rendered to the page.
    */
-  const [pluginChanges, setPluginChanges] = useState(() =>
-    pluginArr.map((plugin: PluginInfo) => {
+  const [pluginArr, setPluginArr] = useState<PluginInfo[]>(
```
sniok, line 105: `plugins` would be more readable imho

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -92,20 +93,17 @@ const EnableSwitch = (props: SwitchProps) => {
...
-  const [enableSave, setEnableSave] = useState(false);
+  const { t } = useTranslation(['translation']);
 
-  /**
-   * pluginChanges state is the array of plugin data and any current changes made by the user to a plugin's "Enable" field via toggler.
-   * The name and origin fields are split for consistency.
+  /** Plugin arr to be rendered to the page from prop data
+   * Note:
+   * The pluginArr is the main array that will be used to render the plugin data to the page.
+   * Any changes made here is what will be rendered to the page.
```
sniok, line 103: this comment repeats the same thing three times :D

sniok (changes requested):
Left some comments. Let me know if you need any help with testing using `npm link`

frontend/src/components/App/PluginSettings/PluginSettings.tsx
vyncent-t: Is the wording too non descriptive? This is kind of the most important part of the settings list so I want to make sure its known that this is the main array of the component but what is being rendered is just the copy of the pluginData prop

I was trying not to make it too lengthy but I think it needs this context to be a starting point for anyone looking at this for the first time (I know I had some old knowledge from settings it up that I had to relearn) , I'm trying to communicate that there are three things about this array part, that it:

- comes from the prop data so its a copy and not the original from the state
- this is the main array the component uses both to render the info to the menu but also for the switchChangeHandler
- any changes we make to the array here (like how we split the name for plugins with the @ symbol or changing of the enable switch) are what will be rendered and isnt entirely a 1 to 1 in the original 

changing it to something like 
```  /** 
   * Plugin array derived from prop data (copy, not original state).
   * This array is the main array used for rendering plugin info and handling switch changes.
   * Modifications done here (e.g., splitting names, toggling enable switch) are what will be rendered.
   */

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 6 above)
Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:105]: this comp already uses plugins as a prop which is always the redux plugins list from the state and since what we are using here is actually just a copy of that array I'm trying to keep it distinct

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 4 above)
Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:295]: fixed!

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 2 above)
Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:289]: this is somewhat of a meta comment from a meeting I had with joaquim where essentially we should having some backwards compatibility for now but eventually phasing it out to just use the new way

File: frontend/src/plugin/pluginsSlice.ts (hunk 1 above)
Comment (vyncent-t) [reply frontend/src/plugin/pluginsSlice.ts:93]: fixed!

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 2 above)
Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:289]: Removing wording of 'to do' tone and making it a note

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 7 above)
Comment (sniok) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx]: How about `// Copy of plugins with pending changes`? The fact that the state is used for rendering is not necessary to point out since pretty much all react state is used for displaying things

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 6 above)
Comment (sniok) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:105]: maybe then `pendingPlugins` or `unsavedPlugins` or something like that, `arr` doesn't add anything to the name, it's an information about the type

Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx:105]: using the `pluginChanges` from before since its reworked now

File: frontend/src/components/App/PluginSettings/PluginSettings.tsx (hunk 7 above)
Comment (vyncent-t) [reply frontend/src/components/App/PluginSettings/PluginSettings.tsx]: done!

File: frontend/src/components/App/PluginSettings/PluginSettings.stories.tsx [hunk 8]
```diff
@@ -40,6 +41,18 @@ function createDemoData(arrSize: number, useHomepage?: boolean) {
...
+/**
+ * createDemoEnabledList function will create a list of plugin objects with a boolean value to enable/disable the plugin.
+ * The function will return an object of plugin names with a boolean value.
+ */
+function createDemoEnabledList(arr: PluginInfo[]): Record<string, boolean> {
+  const enabledList = arr.reduce((acc, p) => {
+    acc[p.name] = !!p.isEnabled;
+    return acc;
+  }, {} as Record<string, boolean>);
+  return enabledList;
```
joaquimrocha, line 53: This seems overly complex for its purpose. Also it says it creates a list but it creates a record.

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +275,60 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
...
+    );
+
+    localStorage.setItem('enabledPluginsList', JSON.stringify(pluginsEnabledMap));
+    localStorage.removeItem('headlampPluginSettings');
+
+    dispatch(setEnablePlugin(pluginsEnabledMap));
+    dispatch(reloadPage());
+  } else {
+    /**
+     * NOTE: For compatibility with old settings, we need to check if the `localEnabledList` exists.
```
joaquimrocha, line 306: So we have 2 deprecated items? headlampPluginSettings + enabledPluginsList?
I thought this was just about fixing the former, where we were storing the package.json contents for no reason, when what we were apparently using that list for was just to check if the plugins were enabled.

frontend/src/components/App/PluginSettings/PluginSettings.tsx
```diff
@@ -284,15 +275,60 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
...
   return (
     <PluginSettingsPure
-      plugins={pluginSettings}
-      onSave={plugins => {
-        dispatch(setPluginSettings(plugins));
-        dispatch(reloadPage());
-      }}
+      plugins={pluginData}
+      pluginsEnabledMap={pluginsEnabledMap}
+      onSave={() => {}}
```
joaquimrocha, line 331: So on save we do nothing? I am confused with what this PR is supposed to fix now.

joaquimrocha (changes requested):
Left some comments. I think this PR looks too risky to me as it's doing many things when the original reasons why we were touching this part of the code was so we 1. didn't have the plugins version/data coming from a client-side stored version of it; 2. avoid having to store all that data when what we are interested in is checking whether the plugins are enabled on the user's side.

I suggest the following: turn this PR into draft so you keep any relevant changes for later.
And for now let's focus on:
1. Not having the data for the plugin be populated from the client-side stored version (but instead use whatever comes from the backend)
2. Keep just the plugin name + isEnabled stored in the headlampPluginSettings record for now. Don't rename it or change the format as I had suggested for now.

Later on we can optimize these changes based on pieces you have here, but I really want to fix the other issues in an isolated way first.

illume:
Seems there’s a new PR for this, so closing. Please reopen if I’m mistaken.

PR #2600, opened 2024-11-22
frontend: Rewrite useQueryParamsState logic and add unit tests 

sniok:
How to test:

Go to map
Select a node
Refresh the page
Node should still be selected

Before this fix it wouldn't work

### DIFF
--- a/frontend/src/components/resourceMap/useQueryParamsState.tsx
+++ b/frontend/src/components/resourceMap/useQueryParamsState.tsx
@@ -1,7 +1,10 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 import { useHistory, useLocation } from 'react-router';
 
-type UseQueryParamsStateReturnType<T> = [T | undefined, (newValue: T | undefined) => void];
+type UseQueryParamsStateReturnType<T> = [
+  T | undefined,
+  (newValue: T | undefined, params?: { replace?: boolean }) => void
+];
 
 /**
  * Custom hook to manage a state synchronized with a URL query parameter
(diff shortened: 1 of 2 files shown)


frontend/src/components/resourceMap/useQueryParamsState.tsx
```diff
@@ -23,6 +23,9 @@ export function useQueryParamsState<T extends string | undefined>(
 
   // State for managing the value derived from the query parameter
   const [value, setValue] = useState<T | undefined>(() => {
+    if (initialState) {
+      return initialState;
+    }
```
joaquimrocha, line 28: AFAIU, if the `initialState` is valid, this logic will never actually get the param from the URL if it has an initialState. Is this intended?
Shouldn't we keep the whole logic but return the `initialState` instead of undefined at the end of this function?

sniok: rewrote the logic to get rid of useState and it's so much simpler now. also added unit tests

frontend/src/components/resourceMap/useQueryParamsState.tsx
```diff
@@ -18,64 +21,43 @@ export function useQueryParamsState<T extends string | undefined>(
...
+  const { search } = useLocation();
   const history = useHistory();
 
-  // State for managing the value derived from the query parameter
-  const [value, setValue] = useState<T | undefined>(() => {
-    const { search } = location;
-    const searchParams = new URLSearchParams(search);
-    const paramValue = searchParams.get(param);
+  const params = new URLSearchParams(search);
+  const value = (params.get(param) ?? undefined) as T | undefined;
```
joaquimrocha, line 28: We should memo the value, since this logic should only run if the search changes.

joaquimrocha (changes requested):
Just one more thing to memo and then I think it's fine.

frontend/src/components/resourceMap/useQueryParamsState.tsx
sniok: done, although this is not a very expensive operation

joaquimrocha (approved):

PR #2602, opened 2024-11-22
frontend Map: Allow plugins to customize the map

sniok:
## Basic Map extension

This PR introduces 2 new functions that allow for the simplest case of extending the Map.

You can read the full docs page over here https://github.com/headlamp-k8s/headlamp/blob/7cbc2ef226bab1f65ad6d8ddccaf07b43c093f95/docs/development/plugins/functionality/extending-the-map.md
You can add your own nodes and edges, customize the icon for the nodes and show a details page for the node.

1. `registerMapSource`
registers a "Source" which provides Nodes and Edges to the map

2. `registerKindIcon`
adds an icon for a specific resource "kind", which is displayed on the node and the map search (not the global one)


joaquimrocha (commented):
This looks very reasonable. I have 2 questions, however:
1. Does it makes sense somehow to unify the details page for the map and for other places?Right now we have the notion of a details route, associated with a KubeObject, but a component is registered for this just as a normal route by using registerRoute.
2. I think the "framework" you are proposing matches a lot the case where we want to display kubernetes resources, but I wonder if it makes sense to abstract it to something that is not Kubernetes specific. Say e.g. we want to display users? Or we want to display clusters?

sniok:
> 1. Does it makes sense somehow to unify the details page for the map and for other places?Right now we have the notion of a details route, associated with a KubeObject, but a component is registered for this just as a normal route by using registerRoute.

Absolutely, I think would be perfect for adding details page, especially for custom resources. Even now we could already replace all routes to render `KubeObjectDetails` component.

> 2. I think the "framework" you are proposing matches a lot the case where we want to display kubernetes resources, but I wonder if it makes sense to abstract it to something that is not Kubernetes specific. Say e.g. we want to display users? Or we want to display clusters?

Yes it can support other objects as well, when we register a source it currently only returns nodes with the type "kubeObject" but it can return any type. The only missing piece for now is registering custom node type.

So it is going to look something like this:
```tsx
// register customUser type
registerNodeType('customUser', { renderNode: (nodeData) => <div>{nodeData.name}</div> } ))

registerMapSource({
   useData() {
      // specify type and pass any data
      return {  nodes: [{ id: 'some-user', type: 'customUser', data: { name: 'user-name' } }] }
   }
})

```

sniok:
Removed the `registerKindDetailsPage` method from this PR as it needs some more investigation and thought

faebr:
Hi @sniok, this feature does exactly what we hacked into headlamp without plugins, very nice! When using the sample code from the latest doc linked above and running the latest version of this branch, the node content is not visible. Would you know what the issue could be?

Thanks! Fabian from Swisscom

Screenshot 2025-02-27 at 09 54 06

faebr:
Hi @sniok, with these changes it works as expected now, this looks great! I have some follow-up questions / ideas:

- Using this new plugin api, the CRD's could be added using a plugin to the map instead of in the base code as we demo'ed to you ( our code implementing CRD in the map is now outdated due to the changes you have done since ). Would you still prefer this feature to be in the core code itself, then I could look into adapting it or would you be more interested in a CRD plugin?
- In the dropdown to select the kind of resources to show in the map, it cuts off the text, what would you think of making the box adjustable using a fit-content() ? (https://github.com/headlamp-k8s/headlamp/commit/ae687ce2fd8bac9c570cc284287a278b3161663c)
- Is there a specific reason you remove loops in the graph? We could think of Scenario's where it would be nice to allow loops in case we would like to have edges not only represent owner-references but other dependencies. Since by default edges show owner-references only, there should not be any loops anyhow or am I wrong? ( example: https://github.com/headlamp-k8s/headlamp/commit/9cb6e50377653d4b86fad14263a9ab5e88a412ae )

sniok:
> Hi @sniok, with the new example it works as expected now, this looks great! I have some follow-up questions / ideas:
> 
> * Using this new plugin api, the CRD's could be added using a plugin to the map instead of in the base code as we demo'ed to you ( our code implementing CRD in the map is now outdated due to the changes you have done since ). Would you still prefer this feature to be in the core code itself, then I could look into adapting it or would you be more interested in a CRD plugin?

+1 on what @joaquimrocha said, Custom Resources should be supported in core. and plugins then can provide edges and additional details component.

> * In the dropdown to select the kind of resources to show in the map, it cuts off the text, what would you think of making the box adjustable using a fit-content() ? ([ae687ce](https://github.com/headlamp-k8s/headlamp/commit/ae687ce2fd8bac9c570cc284287a278b3161663c))

yeah that sounds good!

> * Is there a specific reason you remove loops in the graph? We could think of Scenario's where it would be nice to allow loops in case we would like to have edges not only represent owner-references but other dependencies. Since by default edges show owner-references only, there should not be any loops anyhow or am I wrong? ( example: [9cb6e50](https://github.com/headlamp-k8s/headlamp/commit/9cb6e50377653d4b86fad14263a9ab5e88a412ae) )

I don't think loops were intentionally removed, it just was overlooked since there weren't any loops during development, but map should support any graph, with or without loops. That fix for grouping you proposed looks good. 

Edges can represent any kind of relationship, not just ownership, we have an option to introduce labels and custom styles for edges, there was no need for it yet but it's definitely something we'd want to do.

faebr:
> > * In the dropdown to select the kind of resources to show in the map, it cuts off the text, what would you think of making the box adjustable using a fit-content() ? ([ae687ce](https://github.com/headlamp-k8s/headlamp/commit/ae687ce2fd8bac9c570cc284287a278b3161663c))
> 
> yeah that sounds good!
> 
> > * Is there a specific reason you remove loops in the graph? We could think of Scenario's where it would be nice to allow loops in case we would like to have edges not only represent owner-references but other dependencies. Since by default edges show owner-references only, there should not be any loops anyhow or am I wrong? ( example: [9cb6e50](https://github.com/headlamp-k8s/headlamp/commit/9cb6e50377653d4b86fad14263a9ab5e88a412ae) )
> 
> I don't think loops were intentionally removed, it just was overlooked since there weren't any loops during development, but map should support any graph, with or without loops. That fix for grouping you proposed looks good.
> 
> Edges can represent any kind of relationship, not just ownership, we have an option to introduce labels and custom styles for edges, there was no need for it yet but it's definitely something we'd want to do.

Labels and custom styles for edges would be a nice feature!

I created an mr with these small changes in case this helps: https://github.com/headlamp-k8s/headlamp/pull/2963

For the CRD on the map within the core code I talked to my team but we can not commit for a timeline right now, I will update you when this changes.

joaquimrocha (approved):

Issue #2066, opened 2024-06-20
Empty menu on home page on mobile viewport


sniok:
![image](https://github.com/headlamp-k8s/headlamp/assets/4517681/9f6f8c87-6277-4e11-ae96-f489a2f3947b)

the three dot menu on the right opens an empty menu. on desktop it's not visible

skoeva:
Is this issue the same as #1719?

illume:
Yeah, this was the same issue and it's fixed now. AFAIK

sniok:
This is still an issue, the ticket mentioned above is unrelated. The menu that's not working is in top right

![image](https://github.com/user-attachments/assets/1fa20752-a83e-4d81-a26d-bd3352f97070)

sniok:
On the home view it's empty. On other views (like on your screenshot) if you click not directly on the icon it will do a ripple effect but nothing happens

skoeva:
Ah I see. Will look into this

PR #2603, opened 2024-11-22
frontend: TopBar: Hide menu on mobile when empty

skoeva:
This change hides the top bar menu on mobile when there are no actions to display.

Fixes: #2066

### Testing
- [X] Open Headlamp in the mobile view
- [X] Ensure that the top bar menu shows up only when there are available actions

image

### DIFF
--- a/frontend/src/components/App/TopBar.tsx
+++ b/frontend/src/components/App/TopBar.tsx
@@ -399,6 +399,12 @@ export const PureTopBar = memo(
         ),
       },
     ];
+
+    const visibleMobileActions = processAppBarActions(
+      allAppBarActionsMobile,
+      appBarActionsProcessors
+    ).filter(action => React.isValidElement(action.action) || typeof action === 'function');
+
     return (
       <>
         <AppBar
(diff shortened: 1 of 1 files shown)


frontend/src/components/App/TopBar.tsx
```diff
@@ -395,6 +409,12 @@ export function PureTopBar({
       ),
     },
   ];
+
+  const visibleMobileActions = processAppBarActions(
+    allAppBarActionsMobile,
+    appBarActionsProcessors
+  ).filter(action => React.isValidElement(action) || typeof action === 'function');
```
sniok, line 416: I'm always getting empty array here. It's checking if action is valid element but it has to check an action property inside action object so either `React.isValidElement(action.action)` or 

```tsx
filter(({action}) => React.isValidElement(action) || typeof action === 'function');
```

sniok (changes requested):
Thanks for taking a look at it

I tried it and the menu was always missing, I left a comment on the place that has an issue.

Also I think this PR is a good place to also fix the look of the menu on mobile 
It's a bit awkward with just an icon in the middle, I think adding label to buttons and aligning them to the left would look a lot better. Similar to how other popup menus look in the app

image

image

frontend/src/components/App/TopBar.tsx
skoeva: made the change, looks like this now:

image

can you confirm if the home context still shows the three dots? I'm trying to see if it's worth keeping the second condition mainly for things like the dynamic cluster plugin, although it's being buggy for me atm:

image

frontend/src/components/App/TopBar.tsx
```diff
@@ -139,9 +139,21 @@ function AppBarActionsMenu({
...
+              <MenuItem
+                onClick={(action as AppBarAction)?.onClick || undefined}
+                sx={{
+                  justifyContent: 'center',
+                  gap: '12px',
+                }}
+              >
+                <ListItemIcon sx={{ minWidth: '40px', justifyContent: 'center' }}>
+                  {Action}
+                </ListItemIcon>
```
joaquimrocha, line 154: I am a bit puzzled. What are rendering here?

joaquimrocha (commented):
I am a bit puzzled with this one, since it looks like we are changing the functionality that plugins may be overriding. @sniok can you check/confirm this?
It at least looks like we should have a couple of commits, since the original issue is a menu that is rendering as empty.

frontend/src/components/App/TopBar.tsx
skoeva: With this code:

image

With the code commented out:

image

skoeva:
@sniok Sounds good. Will move the additional changes into a new PR and adjust this one accordingly

joaquimrocha (approved):

Issue #2100, opened 2024-06-27
nodes table looks cut off


illume:
![image](https://github.com/headlamp-k8s/headlamp/assets/9541/5859db0c-dc94-4e6e-90fe-1694219ba08c)

Note the info icon cut off and also all fields are not displayed.

PR #2604, opened 2024-11-22
frontend: Node: Fix table columns and empty values

skoeva:
This change adjusts the width of the table columns to prevent UI issues and fills in empty values where applicable.

Fixes: #2100

### Testing
- [X] Open Headlamp in cluster and navigate to the Nodes page
- [X] Ensure that all empty fields are filled with "None" and columns with the info icon have sufficient spacing.

image

### DIFF
--- a/frontend/src/components/node/List.tsx
+++ b/frontend/src/components/node/List.tsx
@@ -26,6 +26,7 @@ export default function NodeList() {
         {
           id: 'cpu',
           label: t('CPU'),
+          gridTemplate: 'min-content',
           getValue: node => {
             const [used] = getResourceMetrics(node, nodeMetrics || [], 'cpu');
             return used;
@@ -58,7 +59,6 @@ export default function NodeList() {
         {
           id: 'ready',
           label: t('translation|Ready'),
-          gridTemplate: 'minmax(150px, .3fr)',
           getValue: node => {
             const isReady = !!node.status.conditions?.find(
               condition => condition.type === 'Ready' && condition.status === 'True'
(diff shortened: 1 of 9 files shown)


frontend/src/components/node/List.tsx
```diff
@@ -26,6 +26,7 @@ export default function NodeList() {
         {
           id: 'cpu',
           label: t('CPU'),
+          gridTemplate: 'minmax(120px, .5fr)',
```
joaquimrocha, line 29: Why is this needed?

frontend/src/components/node/List.tsx
```diff
@@ -58,7 +59,7 @@ export default function NodeList() {
         {
           id: 'ready',
           label: t('translation|Ready'),
-          gridTemplate: 'minmax(150px, .3fr)',
+          gridTemplate: 'minmax(120px, .3fr)',
```
joaquimrocha, line 62: maybe we could just delete this line?

frontend/src/components/node/List.tsx
skoeva: Otherwise we get this in the UI (described in the issue):

image

frontend/src/components/node/List.tsx
skoeva: your suggestion actually looks cleaner, removed the line here

frontend/src/components/node/List.tsx
joaquimrocha: Maybe try `gridTremplate: 'min-content'`?

skoeva: seems to look very similar, made the change here for simplicity

joaquimrocha (approved):

Issue #2605, opened 2024-11-22
Create namespace button appearing under title


skoeva:
Showing up on the main branch, both in the app and in the browser
![image](https://github.com/user-attachments/assets/e5239d25-43f0-4d8b-adcf-6f7d359142f4)

PR #2606, opened 2024-11-22
frontend: SectionHeader: Fix prop alignment

skoeva:
This change addresses a regression introduced by #2572 since the SectionHeader components were aligned in a column, causing title side actions to appear under the title. Now, subtitles should appear under the title, and title side actions should appear next to the title.

Fixes: #2605

#### Subtitles
- [X] Open Headlamp with an enabled plugin of the format `@org-name/plugin-name` (i.e. `@kinvolk/dynamic-clusters`
- [X] Click to view the plugin details and ensure the org name appears under the plugin name

image

#### Title side actions
- [X] Open a cluster and navigate to the namespaces page
- [X] Ensure that the Create Namespace button appears next to the title

image

### DIFF
--- a/frontend/src/components/common/SectionHeader.tsx
+++ b/frontend/src/components/common/SectionHeader.tsx
@@ -39,25 +39,18 @@ export default function SectionHeader(props: SectionHeaderProps) {
     >
       <Grid item>
         {(!!props.title || titleSideActions.length > 0) && (
-          <Box display="flex" flexDirection="column" alignItems="left">
+          <Box display="flex" alignItems="center">
             {!!props.title && (
-              <>
-                <Typography
-                  variant={titleVariants[headerStyle]}
-                  noWrap
-                  sx={theme => ({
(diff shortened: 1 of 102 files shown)


frontend/src/components/common/SectionHeader.tsx
```diff
@@ -66,6 +59,11 @@ export default function SectionHeader(props: SectionHeaderProps) {
             )}
           </Box>
         )}
+        {!!props.subtitle && (
+          <Typography variant="h6" component="h2" sx={{ fontStyle: 'italic' }}>
```
sniok, line 63: It looks like both title and subtitle are using h2, I'd change subtitle to 'p'. I know this was like this before but it'd be nice to change it here as well 

image

skoeva: done

sniok (approved):
nice!

Issue #2147, opened 2024-07-08
AuthVisible authenticates with any auth verb


skoeva:
Running into a bug where the AuthVisible component runs `getAuthorization` with an invalid authVerb and still returns an allowed status of true. Added some helpful screenshots.

### Setting the authVerb:
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/7d4d1f4d-bdad-4cfc-83bd-703baa0022f1)
### Logging within authVisible:
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/6e805e94-fa27-4e7b-a8b7-f47d4355e8a9)
### Log output:
![image](https://github.com/headlamp-k8s/headlamp/assets/52797224/f5cfc7f8-49f6-4e12-9b32-28d847ed0120)

illume:
Are you able to leave instructions on how to reproduce this?

PR #2610, opened 2024-11-25
frontend: AuthVisible: Validate verb before auth check

skoeva:
This change validates the authVerb passed into the AuthVisible component before calling `getAuthorization()`, which prevents components from rendering with an invalid verb.

Fixes: #2147 

### Testing
- [X] Open a cluster in Headlamp and navigate to the Namespaces page
- [X] Open `frontend/src/components/namespace/CreateNamespaceButton.tsx` and set the authVerb to something random (e.g. '1234')
- [X] Reload the page and ensure that the create button does not show up
- [X] Ensure that you see the following warning in the console:

image

### DIFF
--- a/frontend/src/components/common/Resource/AuthVisible.tsx
+++ b/frontend/src/components/common/Resource/AuthVisible.tsx
@@ -3,6 +3,18 @@ import React, { useEffect } from 'react';
 import { KubeObject } from '../../../lib/k8s/KubeObject';
 import { KubeObjectClass } from '../../../lib/k8s/KubeObject';
 
+/** List of valid request verbs. See https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb. */
+const VALID_AUTH_VERBS = [
+  'create',
+  'get',
+  'list',
+  'watch',
+  'update',
+  'patch',
+  'delete',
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/Resource/AuthVisible.tsx
```diff
@@ -3,10 +3,22 @@ import React, { useEffect } from 'react';
...
+  'patch',
+  'delete',
+  'deletecollection',
+];
+
 export interface AuthVisibleProps extends React.PropsWithChildren<{}> {
   /** The item for which auth will be checked or a resource class (e.g. Job). */
   item: KubeObject | KubeObjectClass | null;
-  /** The verb associated with the permissions being verifying. See https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb . */
+  /** The verb associated with the permissions being verifying. */
```
joaquimrocha, line 21: No need to remove this reference just because it's mentioned above.

skoeva:
@joaquimrocha the `this.fetchAuthorization()` call returns `allowed: true` for the invalid verb here. Maybe the rules permit all verbs without actually validating them.

image

joaquimrocha (approved):

Issue #2601, opened 2024-11-22
Add Cluster button is not visible in dark mode


sniok:
## Describe the bug
Add Cluster button is not visible in dark mode

## To Reproduce
Open app, select cluster dropdown in top right, Add Cluster has black text on black background

![image](https://github.com/user-attachments/assets/108f2711-1dcc-4036-84b3-23cee2a506c6)

(while hovering)
![image](https://github.com/user-attachments/assets/555f4a9d-695b-4564-8366-bdfdc63c4ac6)

windows app

vyncent-t:
I'll take a look 😎

skoeva:
took a quick look, can wrap this up rq before the holiday

PR #2616, opened 2024-11-27
frontend: ClusterChooserPopUp: Fix dark mode visibility

skoeva:
This change fixes a visibility issue with the 'Add Cluster' button in the cluster chooser pop up.

Fixes: #2601 

### Testing
- [X] Open Headlamp in dark mode and go into a cluster
- [X] Open the cluster chooser pop up and ensure that the 'Add Cluster' button is visible

image


sniok (approved):
nice!

PR #2621, opened 2024-11-30
cleanup actions

farodin91:


frontend/src/components/common/Resource/index.test.ts
```diff
@@ -27,6 +27,7 @@ const checkExports = [
   'ResourceListView',
   'ResourceTable',
   'resourceTableSlice',
+  'ResourceTableSingleActions',
```
joaquimrocha, line 30: I wonder if this should actually be exported.

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -84,7 +80,7 @@ export interface ResourceTableProps<RowItem> {
   columns: (ResourceTableColumn<RowItem> | ColumnType)[];
   /** Show or hide row actions @default false*/
   enableRowActions?: boolean;
-  actions?: null | RowAction[];
+  actions?: null | HeaderAction[];
```
joaquimrocha, line 83: I think we need to leave this with the previous type because plugins may be using it and it would mean a type break for them.

joaquimrocha (changes requested):
Left a comment about backward compatibility. Also check out the unsorted imports.

farodin91:
@joaquimrocha I merged header and row action to make the equal. What do you think?

joaquimrocha:
> @joaquimrocha I merged header and row action to make the equal. What do you think?

Hi @farodin91 . I am not sure those 2 types are equivalent in this PR ATM. Maybe @illume can help us understand if I am seeing this correctly.

PR #2626, opened 2024-12-02
frontend: Avoid keeping dialogs mounted

sniok:
Dialog/Popup components have "keepMounted" prop that makes all dialogs render even when they're closed which impacts performance. This PR removes those props


joaquimrocha (approved):
LGTM. We have to be careful just with things like dialogs where we want the contents not to be nuked when the dialog is reopened. I thought we were expecting this for the Terminal, but seems like the contents go away in the `main` branch too....

PR #2627, opened 2024-12-02
frontend: Populate query cache when kubeObject is passed into a Link

sniok:
Every time Link accepts kubeObject we can populate query cache with that item so the navigation is instant and resource isn't loaded again.
This is done on "pointerenter" even to only populate the item user is about to click

### DIFF
--- a/frontend/src/components/common/Link.tsx
+++ b/frontend/src/components/common/Link.tsx
@@ -1,6 +1,8 @@
 import MuiLink from '@mui/material/Link';
+import { useQueryClient } from '@tanstack/react-query';
 import React from 'react';
 import { Link as RouterLink } from 'react-router-dom';
+import { kubeObjectQueryKey, useEndpoints } from '../../lib/k8s/api/v2/hooks';
 import { KubeObject } from '../../lib/k8s/KubeObject';
 import { createRouteURL, RouteURLProps } from '../../lib/router';
 import { LightTooltip } from './Tooltip';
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/Link.tsx
```diff
@@ -28,14 +30,39 @@ export interface LinkObjectProps extends LinkBaseProps {
...
+      onPointerEnter={() => {
+        client.setQueryData(
+          kubeObjectQueryKey({
+            cluster: kubeObject.cluster,
+            endpoint,
+            namespace,
+            name,
+          }),
+          kubeObject
+        );
```
joaquimrocha, line 51: I need to clarify something here. Is this resulting in pulling data from the server on "mouse-enter"? If so, I am not sure this is a great idea since we may hover through several Links till we click one.
Regardless, would it work if we set the data on-click instead? Or is this not compatible with the `to=` attribute here?

joaquimrocha (approved):

joaquimrocha:
@sniok I found a bug: when clicking on the related pods in the events from the cluster overview, the view never loads:
screenshot showing the problem

joaquimrocha:
Ah wait, let me confirm with the most recent changes

joaquimrocha:
> Ah wait, let me confirm with the most recent changes

Sill happening.

sniok:
> when clicking on the related pods in the events from the cluster overview, the view never loads:

this is a case when a partial object is passed to the link. 
added immediate cache invalidation to the item so it always fetches a resource (in the background)

joaquimrocha:
Tested again and it's working fine. LGTM.

Issue #2591, opened 2024-11-21
Auth failed dialog has no option to go home when there's only 1 cluster


joaquimrocha:
## Describe the bug

This is a reminiscence of when we didn't show the home if only one cluster was configured (because there's not much sense for pure in-cluster deployments): the auth dialog is not showing an option to go back home if there's only one cluster configured.

## To Reproduce
Steps to reproduce the bug:
1. Have only one cluster configured for Headlamp; make sure the cluster is unnaccessible.
2. Open Headlamp try to access that cluster -> it should bring you to the auth failed dialog and there's no option to go back to the home view

## Environment (please provide info about your environment):
 - Installation type: any
 - Headlamp Version: 0.26.0

skoeva:
@joaquimrocha when this happens, we can just navigate to home using the alert at the top (assuming it shows up):

![image](https://github.com/user-attachments/assets/9cbe4c9d-e404-41d1-b20a-698bd975be92)

but if it doesn't, we can have the back button persist as shown to navigate to home. maybe the notification can be more descriptive though

PR #2631, opened 2024-12-03
frontend: AuthChooser: Add home navigation on auth fail

skoeva:
This change makes the back button on the auth fail pop-up persist to allow the user to navigate back to home from an inaccessible single-cluster context.

Fixes: #2591 

### Testing
- [X] First, open Headlamp in-cluster (only one cluster, use minikube for simplicity here)
- [X] In the browser, navigate to an inaccessible url (e.g., `https://localhost:3000/c/fake-cluster`)
- [X] Run `minikube stop` in the terminal and navigate to `https://localhost:3000/c/minikube`
- [X] When the auth fails, click the back button in the pop-up and ensure that you are taken to the home (`https://localhost:3000`) instead of the actual previous link (`https://localhost:3000/c/fake-cluster`)

image

### DIFF
--- a/frontend/src/components/authchooser/AuthChooser.stories.tsx
+++ b/frontend/src/components/authchooser/AuthChooser.stories.tsx
@@ -32,7 +32,6 @@ const argFixture = {
   error: null,
   oauthUrl: 'http://example.com/',
   clusterAuthType: '',
-  haveClusters: false,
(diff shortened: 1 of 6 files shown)


joaquimrocha (approved):
LGTM. Thanks

PR #2636, opened 2024-12-04
frontend: fix nullable node status

farodin91:


frontend/src/components/cluster/Charts.tsx
```diff
@@ -19,7 +19,7 @@ export function MemoryCircularChart(props: ResourceCircularChartProps) {
   }
 
   function memoryAvailableGetter(item: Node | Pod) {
-    return parseRam(item.status!.capacity.memory) / TO_GB;
+    return parseRam(item.status!.capacity?.memory) / TO_GB;
```
sniok, line 22: might as well remove the `!` here `item.status?.capacity?.memory`

sniok:
nice, thanks for adding this @farodin91. Was this change prompted by some issue you've experienced?

farodin91:
I did a node rotation and the node list was broken during after a new node was created.

sniok (approved):
Thanks! Really appreciate updated test cases and type definitions

frontend/src/components/cluster/Charts.tsx
farodin91: I will change this.

farodin91:
@sniok updated.

PR #2639, opened 2024-12-05
charts: fix: allow env vars to be set when reading others from secrets

devjoes:
There is currently no way of setting arbitrary environment variables when reading OIDC config from secrets, this PR allows it.


knrt10 (approved):
Thank you for the PR @devjoes. LGTM

PR #2644, opened 2024-12-06
backend: Disable websocket multiplexer

illume:
This should not be enabled in a release.

I'd like this disabled now, so it isn't enabled in the next release accidentally before it's ready.

related:
- https://github.com/headlamp-k8s/headlamp/pull/2563


joaquimrocha:
Is there any concern here? Assuming nothing in the UI is using it?

illume:
Nothing is using it, but it should not be turned on for a release.

PR #2649, opened 2024-12-09
frontend Map: Remember zoom mode. Move viewport logic to useGraphViewport hook

sniok:
Remembers your last selected view mode

For example:
Click on fit to view on the map
Navigate around
It will now always fit to view (used to be fit to 100% by default)


joaquimrocha (changes requested):
Works perfectly but I think these are two different commits:
Move viewport logic to useGraphView
Remember zoom mode

joaquimrocha (approved):

Issue #2640, opened 2024-12-05
After clicking on button toggle full screen the logs do not adjust to the screen width


Andrew1988123123:
## Describe the bug
After clicking on toggle full screen the logs do not adjust to the screen width.

## To Reproduce
1. Deploy an application https://github.com/spring-projects/spring-petclinic/tree/main/k8s
2. Click on pod
3. Click on button Show Logs
4. Click on button Toggle full screen
5. You should see logs do not adjust to the screen width

## Environment (please provide info about your environment):
 - Installation type: browser, Windows-Winget
 - Headlamp Version: 0.26.0

## Are you able to fix this issue?
No

## Additional Context
before clicking on button toggle full screen
![image](https://github.com/user-attachments/assets/4788b182-32f2-4156-bb18-f3a4cf9c74f3)

after clicking on button toggle full screen
![image](https://github.com/user-attachments/assets/67d538ad-2f68-4839-b672-3ff2670d2b9a)

joaquimrocha:
Thanks @Andrew1988123123 , does this happen only to the lines that were already printed? Or also to newly written (after fullscreen) ones?

Andrew1988123123:
Newly added lines are also not displayed correctly.

PR #2650, opened 2024-12-09
frontend: LogViewer: Fix log not expanding on full screen

vyncent-t:
fixes issue #2640

## Description

This PR fixes an issue with the `LogViewer` component where log lines did not expand to the full width when in full-screen mode. This issue impacts user experience, especially for users relying on the full-screen view to read long log lines.

### Key Changes

1. **Full-Width Expansion**:
   - Updated the `LogViewer` component's logic allowing render handlers to run resizes to ensure log lines expand to utilize the full width in full-screen mode.

---

## Steps to Test

1. Open the `LogViewer` component in the pods details view.
2. Switch to full-screen mode.
3. Verify:
   - Log lines expand to the full width of the screen in full-screen mode.
   - Log lines remain readable and aligned correctly in both standard and full-screen modes.

4. Test on various screen sizes to ensure responsiveness and compatibility.

### DIFF
--- a/frontend/src/components/common/LogViewer.tsx
+++ b/frontend/src/components/common/LogViewer.tsx
@@ -129,7 +129,17 @@ export function LogViewer(props: LogViewerProps) {
   }
 
   return (
-    <Dialog title={title} withFullScreen onClose={onClose} {...other}>
+    <Dialog
+      title={title}
+      onFullScreenToggled={() => {
+        setTimeout(() => {
+          fitAddonRef.current!.fit();
+        }, 1);
+      }}
+      withFullScreen
+      onClose={onClose}
+      {...other}
+    >
       <DialogContent
         sx={theme => ({
           height: '80%',


vyncent-t:
note: I have tried setting different widths to take up the full size of the container, this would only resize on save after the full screen mode is already expanded and would not work if starting from the beginning (goes to details, opens log viewer, click full screen, would still be half size)

the xterm `fitAddonRef.current!.fit();` is the main fix for setting the sizes, I also tried looking into ResizeObserver but I think that would be too much for performance.

sniok:
would calling `fitAddonRef.current!.fit();` inside `onFullScreenToggled` callback on Dialog component work? This current change seems overly complicated, there's a new state variable with an unusual setter that seemingly triggers the effect that sets up the whole terminal from scratch.

vyncent-t:
@sniok @joaquimrocha 

I have tried the following and it does not seem to not fix the full screen issue, the console log prints but the fit method does not resize anything, I tried to call it inside the onFullScreenToggled entirely and I also tried creating a stand alone function and passing that in and had the same result, all three ways still print but the calling of the fitAddonRef.current?.fit() does not work, I saw this trend and thought maybe it just needed to rerender from the useEffect 

`    <Dialog title={title} withFullScreen onClose={onClose} onFullScreenToggled={(isFullScreen: boolean) => {
      if (isFullScreen) {
        fitAddonRef.current?.fit();
        console.log("REFITS")
      }
    }} {...other}>`

--- 

image

---

image

--

also tried inserting some new instances to maybe try that too but nothing 

```  function fullScreenResize(isFullScreen: boolean) {

    fitAddonRef.current = new FitAddon();

    if (isFullScreen) {
      fitAddonRef.current?.fit();
      console.log("REFITS")
    }
  }

vyncent-t:
using the timer fix `<Dialog
      title={title}
      onFullScreenToggled={() => {
        setTimeout(() => {
          fitAddonRef.current!.fit();
        }, 1);
      }}
      withFullScreen
      onClose={onClose}
      {...other}
    >`

@sniok

vyncent-t:
need i18n missing words from #2660

sniok (approved):

PR #2653, opened 2024-12-10
frontend: Add Pod list metrics

### LINKED ISSUE #1: Pods: Add resource metrics to the list view
We should add the charts for memory/CPU requested/actual/limit used by pods.

sniok:
Fixes #1 

How to test:
Go to Pods
Check new columns CPU and Memory

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -25,6 +25,7 @@ import { KubeObject } from '../../../lib/k8s/KubeObject';
 import { KubeObjectInterface } from '../../../lib/k8s/KubeObject';
 import { KubeObjectClass } from '../../../lib/k8s/KubeObject';
 import Pod, { KubePod, KubeVolume } from '../../../lib/k8s/pod';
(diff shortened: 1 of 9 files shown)


skoeva (approved):
Tested and works, thanks for this

joaquimrocha (changes requested):
Code looks fine but it's not accounting for when the metrics server is not installed. We do adapt the cluster overview's charts when there's no metrics-server, so here we should do something similar.

illume:
@sniok this has some conflicts with main branch now.

illume:
Tested it without rebasing to main...

- Does not fail without metrics server enabled (`minikube addons enable metrics-server`)
- Works with metrics server enabled. (`minikube addons disable metrics-server`) No cpu/mem columns.


image

illume (changes requested):
Please fix conflicts with main.

sniok:
> Note there are some gaps in the table?
> 

Fixed an edge case when usage is exactly 0, it will now display 0.

image

 For pods that don't have metrics (like a completed pod) it'll stay empty

image

joaquimrocha (approved):

Issue #2633, opened 2024-12-04
Headlamp does not show restart/edit/scale buttons on resources with sufficient permissions


Andrew1988123123:
## Describe the bug
The bug title says it all. I gave permissions to restart, scale, edit the deployment but the button with the option to restart and scale the deployment does not appear. These are sufficient permissions because I tested using kubectl.
**kubectl rollout restart deployment gotify -n default**

According to the documentation the buttons should appear
https://headlamp.dev/docs/latest/faq/#headlamp-is-not-showing-deleteeditscale-buttons-in-a-resource-why-is-that

## To Reproduce
Create Deployment
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: gotify
  namespace: default
spec:
  selector:
    matchLabels:
      app: gotify
  replicas: 1
  template:
    metadata:
      labels:
        app: gotify
    spec:
      containers:
        - name: gotify
          image: gotify/server
          env:
            - name: GOTIFY_SERVER_PORT
              value: "81"
          ports:
            - containerPort: 81
          readinessProbe:
            httpGet:
              path: /health
              port: 81
            initialDelaySeconds: 5
            periodSeconds: 10
          livenessProbe:
            httpGet:
              path: /health
              port: 81
            initialDelaySeconds: 10
            periodSeconds: 20
```
Create SA with secret
```
apiVersion: v1
kind: Secret
metadata:
  name: cluster-developer
  namespace: default
  annotations:
    kubernetes.io/service-account.name: cluster-developer
type: kubernetes.io/service-account-token
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cluster-developer
  namespace: default
```
Create custom RBAC with sufficient permissions.
```apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: custom-role
rules:
  - apiGroups: [ "" ]
    resources: [ "pods" ]
    verbs: [ "get", "delete" ]
  - apiGroups: [ "" ]
    resources: [ "pods/portforward" ]
    verbs: [ "get", "create", "delete" ]
  - apiGroups: [ "" ]
    resources: [ "pods/exec" ]
    verbs: [ "get", "create" ]
  - apiGroups: [ "" ]
    resources: [ "configmaps" ]
    verbs: [ "get" ]
  - apiGroups: [ "apps" ]
    resources: [ "deployments/scale" ]
    verbs: [ "get", "patch" ]
[...]

PR #2654, opened 2024-12-10
frontend: Fix table row action bugs

skoeva:
These changes address two issues with the row actions, specifically seen in the deployments table. Firstly, we remove a duplicate ViewButton icon from the row actions that was previously visible. Secondly, we set the apiVersion in the KubeObject constructor to ensure that row actions for the given KubeObject are visible.

Fixes: #2633

### Testing
- [X] Apply the YAML for the deployment, SA, and RBAC given in #2633 
- [X] Generate a new kubeconfig from this and apply it in the Headlamp app
- [X] Open the new context and navigate to the "Deployments" page
- [X] Ensure that the row actions in the table and actions on the deployment page match the permissions
- [X] Ensure that no duplicate ViewButton icon is visible in the row actions


frontend/src/components/common/Resource/EditButton.tsx
```diff
@@ -86,7 +87,7 @@ export default function EditButton(props: EditButtonProps) {
     return null;
   }
 
-  if (isReadOnly) {
+  if (isReadOnly && isDetailPage) {
```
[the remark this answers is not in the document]
skoeva: now that I think about it, makes sense to remove it from the actions list...

I can drop this commit then and take a look at your PR

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -74,7 +74,10 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
   }
 
   constructor(json: T, cluster?: string) {
-    this.jsonData = json;
+    this.jsonData = {
+      ...json,
+      apiVersion: json?.apiVersion ?? (this.constructor as typeof KubeObject).apiVersion,
```
sniok, line 79: jsondata is supposed to be the exact thing we receive from the backend and I think we should try to avoid modifying it. 
could you explain where in the code this field is needed and expected?

skoeva: The problem with the row actions is that AuthVisible requires a KubeObject and the jsondata provides an object without an apiVersion, which is needed for auth. Without the apiVersion, the row actions do not appear, so this is supposed to fix that. Will add this to the PR description

sniok (approved):

joaquimrocha (commented):
@skoeva Do we always need to set the group + version + resource when checking for permissions? I think the original logic was "check if we are checking permissions with e.g. the resource, if so, use this API version's one instead", in these changes it does: "even if we were not testing permissions using the resource, will just set it".

Are we sure this is the right approach? We ought to test variants of these calls.

skoeva:
> @skoeva Do we always need to set the group + version + resource when checking for permissions? I think the original logic was "check if we are checking permissions with e.g. the resource, if so, use this API version's one instead", in these changes it does: "even if we were not testing permissions using the resource, will just set it".
> 
> Are we sure this is the right approach? We ought to test variants of these calls.

@joaquimrocha I think it would be fitting to have a complete resource to evaluate permissions for, no? The issue before was that the object was incomplete and therefore we were unable to check for its permissions. In what case would we prefer an incomplete object?

joaquimrocha:
> @joaquimrocha I think it would be fitting to have a complete resource to evaluate permissions for, no? The issue before was that the object was incomplete and therefore we were unable to check for its permissions. In what case would we prefer an incomplete object?

I just saying the nature of how we were checking things was more of `if we use a resource name -> now try the resource name from this other version`, and right now we're saying `if a resource name is not set -> try the resource name` from this version. So I believe this is incorrect to what is the point of that logic. It may be that the logic is wrong in some cases, but if we are missing fields, we should verify why that is instead of adding them from a new version, or by not replacing them with the current testing version, which is what these changes **also** do.

skoeva:
> I just saying the nature of how we were checking things was more of `if we use a resource name -> now try the resource name from this other version`, and right now we're saying `if a resource name is not set -> try the resource name` from this version. So I believe this is incorrect to what is the point of that logic. It may be that the logic is wrong in some cases, but if we are missing fields, we should verify why that is instead of adding them from a new version, or by not replacing them with the current testing version, which is what these changes **also** do.

@joaquimrocha The code comments suggest that the changes made in this PR are more in line with the intended logic here:

image

Not sure if the original logic makes much sense in this case

joaquimrocha (commented):
I think you may have a point, now that look at this with fresher eyes. I think if that's the case, then the previous logic is indeed wrong, but seems like we're possibly mixing things (mixing the `group` from an API version with `version` from another), so maybe what we need to do is to use the full apiName + apiVersion here instead.
In either case, I'd kindly ask you to test this function and understand what happens when e.g. you add newer+fictitious apiVersion to a resource, and test this function (and the instances' getAuthorization), to simulate falling back to a working version.

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -513,31 +513,22 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
...
     // API info versions until we find one that works.
     const apiInfo = this.apiEndpoint.apiInfo;
     for (let i = 0; i < apiInfo.length; i++) {
-      const { group, version, resource } = apiInfo[i];
-      // We only take from the details from the apiInfo if they're missing from the resourceAttrs.
+      const { group, version } = apiInfo[i];
+      // We only take the details from the apiInfo if they're missing from the resourceAttrs.
       // The idea is that, since this function may also be called from the instance's getAuthorization,
       // it may already have the details from the instance's API version.
-      const attrs = { ...resourceAttrs };
```
joaquimrocha, line 530: The comments that existed already before this line need to go. (// The idea is that, ....)

joaquimrocha (changes requested):
LGTM. Only a comment that should be deleted as it no longer makes sense.

frontend/src/lib/k8s/KubeObject.ts
skoeva: done, thanks for pointing this out!

joaquimrocha (approved):

PR #2661, opened 2024-12-11
frontend: Small table fixes and improvements

sniok:
1. Fix name column value in Role list
Search by Role name now works

2. Fix searching error when some items don't have a namespace
Go to roles, search, it now works and doesn't silently print errors in console

3. Fix per column filtering not working
This is a regression. This button now works
image

4. Make checkboxes smaller and fixed width
Just a small cosmetic improvement


joaquimrocha (approved):
LGTM.

joaquimrocha:
Merging since the failed tests are not introduced by this PR.

Issue #2560, opened 2024-11-11
test plugins failing


skoeva:
## Describe the bug
The 'test plugins' workflow is failing locally and in the CI

## Additional Context
Here's a screenshot of the error output:

![image](https://github.com/user-attachments/assets/8f17c515-3d2a-42a8-aa62-638872ebd893)

vyncent-t:
noticed this also

PR #2662, opened 2024-12-11
CI: disable test plugins steps

sniok:
Since our CI has been broken for more than a month this PR disables the step that fails

I personally missed a couple of actual ci errors by assuming it's just 'test plugin' step that's failing 

Related #2549 #2560


skoeva (approved):
LGTM. I addressed the i18n test failure in a different PR so this should otherwise be good

Issue #2666, opened 2024-12-11
Plugin Settings switch buttons not changing


vyncent-t:
## Describe the bug

When in app mode and changing the setting of a plugin enable/disable the switches do not display the correct positioning. They always remains in the same position the page was loaded on whenever changing, the UI will update to the correct enable / disable positions when the save button is clicked and the page reloads. This seems to be a UI only issue as the save button still continues to enable / disable plugins on reload.

![image](https://github.com/user-attachments/assets/b3b27e4a-81ed-4cd6-b944-34c99931e70d)

## To Reproduce
Steps to reproduce the bug:
1. Open headlamp in app mode and navigate to settings then plugins settings page
2. Have a plugin installed and click on the enable switch
3. Observe that the status does change but the button does not

## Are you able to fix this issue?
Looking into this now

joaquimrocha:
Fixed

PR #2667, opened 2024-12-11
frontend: PluginSettings: Fix non responsive enable toggle

vyncent-t:
## Description

fixes issue #2666 

This PR fixes a UI issue in the plugin settings page when running Headlamp in `app mode`. Previously, toggling the enable/disable switches for plugins did not visually reflect the updated state. The switches remained in their original positions until the save button was clicked and the page reloaded. This issue was limited to the UI, as the save functionality correctly updated the plugin state on reload.

### Key Changes

1. **Real-Time Switch Updates**:
   - Fixed the UI behavior to ensure that the enable/disable switches visually reflect the updated state immediately after being toggled.
   - The switches now display the correct position without requiring a page reload.

2. **Enhanced State Handling**:
   - Updated the component logic to synchronize the UI state with the underlying plugin state upon toggle.

---

## Steps to Test

1. Open Headlamp in `app mode` and navigate to the plugin settings page.
2. Ensure you have at least one plugin installed.
3. Toggle the enable/disable switch for a plugin and verify:
   - The switch visually updates to the correct position immediately after toggling.
   - The save functionality still works as expected, correctly persisting the state across reloads.

4. Reload the page and confirm that the plugin state matches the toggled positions.

### DIFF
--- a/frontend/src/components/App/PluginSettings/PluginSettings.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettings.tsx
@@ -240,6 +240,7 @@ export function PluginSettingsPure(props: PluginSettingsPureProps) {
             },
             {
               header: t('translation|Enable'),
+              accessorFn: (plugin: PluginInfo) => plugin.isEnabled,
               Cell: ({ row: { original: plugin } }: { row: MRT_Row<PluginInfo> }) => {
                 if (!plugin.isCompatible || !helpers.isElectron()) {
                   return null;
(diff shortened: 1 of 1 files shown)


vyncent-t:
I think the issues was coming from changes in the memo from the table a while ago [here](https://github.com/headlamp-k8s/headlamp/commit/dea38d8fbe49b488de9ef146e3eede38f79059f5) I could not get the page to follow the re-render changes that the `switchChangeHandler` was previously able to control but I noticed the Status col was updating dynamically still so I moved the render logic into the accessorFn

joaquimrocha (approved):
Nice. Thanks!

PR #2671, opened 2024-12-12
[OLD] frontend: PluginSettings: Refactor local storage and plugin data

vyncent-t:
Fixes Issue #2595

## Description

- This PR shrinks the saved JSON data from the plugins information and saves only the `name` and `isEnabled` status in local storage. This ensures plugin settings persist when closing and reopening the app.
- This PR introduces the `usePlugins` hook into `lib/k8s/api/v2`
- It also handles backward compatibility, allowing previously saved settings to be used in this new format.
- Additionally, it no longer takes old information from the JSON in local storage. Instead, it retrieves the necessary plugin data from the backend when the main Plugin component is called.

## Changes

1. **Introduces usePlugins hook**: Hook used to fetch the plugins from the backend and return the plugins with their settings
2. **Reduced Storage Data**: Only stores essential plugin info (`name` and `isEnabled`) in local storage.
3. **Backward Compatibility**: Converts old-format data into the new structure on first run.
4. **Data Source Shift**: Eliminates use of outdated local storage JSON data, and instead fetches plugin details from the backend when the Plugin component mounts.

## How to Test

1. **Setup**:
   - Pull this branch and run the application.
2. **Verify Plugin Settings**:
   - Open headlamp in app mode
   - Navigate to the settings page then to plugins settings tab
   - Check local storage item `headlampPluginSettings`
3. **Check Backward Compatibility**:
   - Start with an older version of the app or start from main.
   - Navigate to the settings page then to plugins settings tab
   - Check local storage item `headlampPluginSettings`
   - Switch to this branch and run the app.
   - Confirm that previously stored plugin settings are recognized and converted correctly.
5. **Backend Data**:
   - Verify that after loading, the plugin details (other than `name` and `isEnabled`) are fetched from the backend and displayed correctly in the UI.
   - Switch plugins off and on and observe the changes in local storage and in app

## Notes

- If you have preexisting local storage data, this PR will convert it automatically. You can reset the local storage if you return to main and go to the same plugin settings paage.
- Double-check the logs to ensure no errors appear during the migration from old data to new data.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/plugins.ts
+++ b/frontend/src/lib/k8s/api/v2/plugins.ts
@@ -1,0 +1,101 @@
+import '@tanstack/react-query';
+import { useQuery } from '@tanstack/react-query';
+import semver from 'semver';
+import helpers from '../../../../helpers';
+import { PluginInfo, PluginSettings } from '../../../../plugin/pluginsSlice';
+import { backendFetch } from './fetch';
+
+/*
+ * this function is used to check the compatibility of the plugin version with the headlamp version
+ *
+ * @param compatibleVersion headlamp-plugin version this build is compatible with.
+ *     If the plugin engine version is not compatible, the plugin will not be loaded.
+ *     Can be set to a semver range, e.g. '>= 0.6.0' or '0.6.0 - 0.7.0'.
+ *     If set to an empty string, all plugin versions will be loaded.
+ */
+export function checkCompatibleVersion(packageInfo: PluginInfo, checkAllVersions: boolean = false) {
+  /*
+   * this is the compatible version of the plugin with the headlamp version
+   */
+  const compatibleVersion = !checkAllVersions ? '>=0.8.0-alpha.3' : '';
+
+  // Can set this to a semver version range like '>=0.8.0-alpha.3'.
+  // '' means all versions.
+  const isCompatible = semver.satisfies(
+    semver.coerce(packageInfo.devDependencies?.['@kinvolk/headlamp-plugin']) || '',
+    compatibleVersion
+  );
+
+  return isCompatible;
+}
+
+export async function getPluginPaths() {
+  const pluginPaths = (await fetch(`${helpers.getAppUrl()}plugins`).then(resp =>
+    resp.json()
+  )) as string[];
+
+  return pluginPaths;
+}
+
+/*
+ * this function is used for the plugin settings and is extended from the original `fetchAndExecutePlugins` function in the Plugins.tsx file
+ * - the function is used to fetch the plugins from the backend and return the plugins with their settings
+ * - it will also do a compatibility check for the plugins and return the plugins with their compatibility status,
+ * - the compatibility check is needed to render the plugin switches
+ */
+export async function getPlugins(pluginSettings: PluginSettings[]) {
+  const pluginPaths = await getPluginPaths();
+
+  const packageInfosPromise = await Promise.all<PluginInfo>(
+    pluginPaths.map(path =>
+      backendFetch(`${path}/package.json`).then(resp => {
+        if (!resp.ok) {
+          if (resp.status !== 404) {
+            return Promise.reject(resp);
+          }
+
+          console.warn(
+            'Missing package.json. ' +
+              `Please upgrade the plugin ${path}` +
+              ' by running "headlamp-plugin extract" again.' +
+              ' Please use headlamp-plugin >= 0.8.0'
+          );
+
+          return {
+            name: path.split('/').slice(-1)[0],
+            version: '0.0.0',
+            author: 'unknown',
+            description: '',
+          };
+        }
+        return resp.json();
+      })
+    )
+  );
+
+  const packageInfos = await packageInfosPromise;
+
+  const pluginsWithIsEnabled = packageInfos.map(plugin => {
+    const matchedSetting = pluginSettings.find(p => plugin.name === p.name);
+    if (matchedSetting) {
+      const isCompatible = checkCompatibleVersion(plugin);
+
+      return {
+        ...plugin,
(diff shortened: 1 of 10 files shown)


vyncent-t:
Last push adds backwards comp so that all settings wont be set to off or on

vyncent-t:
I have tried to keep the changes as small as possible to accomplish everything it needs to, I am not using the map object to save information anymore, although to hit these targets I still had to use most of what I had reworked previously

@joaquimrocha @sniok 

for this PR the targets it needs to hit were:
- not have the plugins version/data coming from a client-side stored version of it

The plugin data comes from the backend, reused the same method of reaching it from the previous rework

- avoid having to store all that data when what we are interested in is checking whether the plugins are enabled on the user's side

The data is now trimmed down to just being the name and isEnabled, it is now being used the same way as the original settings where this local saved item is how new app start ups save plugin settings

- need to be backwards compatible with the previous settings method

The local storage handling logic checks for the old format and changes it to the new format

sniok:
Some things to do to avoid duplicating of logic:

 - refactor fetching logic into a single place that both PluginSettings and fetchAndExecutePlugins use
 - use react-query for fetching to have caching and avoid refetching the same stuff

vyncent-t:
pushing rebase of current main

e2e-tests/tests/pluginSetting.spec.ts
```diff
@@ -18,13 +18,17 @@ test('plugin settings page should have a table', async ({ page }) => {
   await headlampPage.tableHasHeaders('table', expectedHeaders);
 });
 
-test('pod counter plugin should have setting option', async ({ page }) => {
-  const headlampPage = new HeadlampPage(page);
-  const pluginName = 'headlamp-pod-counter';
+// to do: fix this or remove this test all together to make compatiable with new plugin name print format (not including @kinvolk in name)
```
sniok, line 21: could you try to modify this test so it passes (in CI) on this branch?

frontend/src/lib/k8s/api/v2/plugins.ts
```diff
@@ -0,0 +1,71 @@
...
+    resp.json()
+  )) as string[];
+  const packageInfosPromise = await Promise.all<PluginInfo>(
+    pluginPaths.map(path =>
+      fetch(`${helpers.getAppUrl()}${path}/package.json`).then(resp => {
+        if (!resp.ok) {
+          if (resp.status !== 404) {
+            return Promise.reject(resp);
+          }
+          {
```
sniok, line 24: curly brackets are not needed here

frontend/src/lib/k8s/api/v2/plugins.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+    )
+  );
+
+  const packageInfos = await packageInfosPromise;
+
+  const pluginsWithIsEnabled = packageInfos.map(plugin => {
+    const matchedSetting = pluginSettings.find(p => plugin.name === p.name);
+    if (matchedSetting) {
+      const compatibleVersion = '>=0.8.0-alpha.3';
```
sniok, line 49: this constant is repeated in 'fetchAndExecutePlugins' function. you could reuse this function there to avoid duplication

frontend/src/plugin/index.ts
```diff
@@ -199,17 +199,20 @@ export function filterSources(
...
-  settingsPlugins: PluginInfo[]
+  settingsPlugins: { name: string; isEnabled: boolean }[]
 ): PluginInfo[] {
   if (backendPlugins.length === 0) return [];
 
   const pluginsChanged =
     backendPlugins.length !== settingsPlugins.length ||
-    backendPlugins.map(p => p.name + p.version).join('') !==
-      settingsPlugins.map(p => p.name + p.version).join('');
+    backendPlugins.map(p => p.name) !== settingsPlugins.map(p => p.name);
```
sniok, line 208: you can't compare arrays like that, this will always be false because they are compared by reference and not by their content

sniok (changes requested):
thanks for the update, left a couple of comments

frontend/src/plugin/index.ts
vyncent-t: done!

frontend/src/lib/k8s/api/v2/plugins.ts
vyncent-t: done!

frontend/src/lib/k8s/api/v2/plugins.ts
```diff
@@ -0,0 +1,71 @@
...
+ * - it will also do a compatibility check for the plugins and return the plugins with their compatibility status,
+ * - the compatibility check is needed to render the plugin switches
+ */
+async function getPlugins(pluginSettings: { name: string; isEnabled?: boolean }[]) {
+  const pluginPaths = (await fetch(`${helpers.getAppUrl()}plugins`).then(resp =>
+    resp.json()
+  )) as string[];
+  const packageInfosPromise = await Promise.all<PluginInfo>(
+    pluginPaths.map(path =>
+      fetch(`${helpers.getAppUrl()}${path}/package.json`).then(resp => {
```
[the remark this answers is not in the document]
vyncent-t: done!

frontend/src/lib/k8s/api/v2/plugins.ts
vyncent-t: done!

sniok (changes requested):
Plugins setting stopped working, check the pod-counter plugin settings

illume:
@vyncent-t It seems there’s conflicts now. Can you please fix them?

illume:
Closing because new PR is here: https://github.com/kubernetes-sigs/headlamp/pull/3123

PR #2678, opened 2024-12-13
frontend: Fix memoization for Head cell when selecting all rows

### LINKED ISSUE #2669: Table select all from partial not working as expected
## Describe the bug

In a table, if you select a couple of rows and then click the `-` (select/deselect all) button, you'll see that it selects all rows but doesn't turn into a checkmark nor does it do anything when clicked again.

## To Reproduce
Steps to reproduce the bug:
1. In the main branch, go to pods + select 2 pods
2. Click the `-` on the top check mark (checkmark column) -> all rows are selected 
3. Click the `-` again -> nothing happens

I expected the `-` checkbox to deselect the 2 selected rows on step 2 instead.

 - Installation type: any
 - Headlamp Version: main

sniok:
Fixes #2669

This bug only occurs when the table has more than one page.

Steps to reproduce

1. Go to the page with a bunch of items (>1 page)
2. Select one item
3. Then click on select all checkbox

### DIFF
--- a/frontend/src/components/common/Table/Table.tsx
+++ b/frontend/src/components/common/Table/Table.tsx
@@ -369,13 +369,7 @@ export default function Table<RowItem extends Record<string, any>>({
                 isFiltered={header.column.getIsFiltered()}
                 sorting={header.column.getIsSorted()}
                 showColumnFilters={table.getState().showColumnFilters}
-                selected={
-                  table.getIsAllRowsSelected()
-                    ? 'all'
-                    : table.getIsSomeRowsSelected()
-                    ? 'some'
-                    : 'none'
-                }
+                selected={table.getSelectedRowModel().flatRows.length}
               />
             ))}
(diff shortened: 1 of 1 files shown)


skoeva:
Have you tested the changes? I switched to this feature branch and seems to me that the table selection acts functionally identical to the main branch, which is the same as in the [MRT docs](https://www.material-react-table.com/docs/guides/row-selection) (correct me if I'm wrong but is there an issue to fix here?)

sniok:
@skoeva 

Currently when you select one item, the top "select all" checkbox has a "-" icon as expected.
image
But then if you click the "select all" checkbox it selects everything but the icon stays the same, but should turn into a check
image
And then clicking it again should deselect everything but nothing happens.
So this PR fixes that

skoeva (approved):
Tested and works as expected 🎉

PR #2680, opened 2024-12-13
frontend: Add CssBaseline component for improved default styles

sniok:
It's a "css reset" that comes from mui library.

Makes scrollbars match the theme for example.

Before

image

After

image


skoeva (approved):
LGTM

sniok:
Didn't notice that we already have CssBaseline in Layout.tsx, I'll update the property there

PR #2690, opened 2024-12-13
docs(platforms): Add support for Red Hat OpenShift

pichuang:
After actually installing headlamp according to the in-cluster installation file, Red Hat OpenShift can run headlamp without any modification.


pichuang:
Headlamp view
20241214-023141@2x

erikgb:
@pichuang How did you manage to get the image to run on OpenShift? https://github.com/headlamp-k8s/headlamp/issues/2342 is still open, and I cannot see that the image has changed to support the random UID assigned to all OpenShift workloads. Would you mind adding some lines to the documentation about this, since Headlamp claims to support OpenShift now (after this PR)?

pichuang:
@erikgb I just use `oc apply -f <deployment>`, and most of the information was as I expected, and after all displaying the information via Kubernetes API was a no brainer for me.

Issue #2656, opened 2024-12-10
Sort namespaces from k3s/k3d


skoeva:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
The namespace list provided by k3s/k3d is unsorted, unlike with other platforms like kind and k0s

## Describe the solution you'd like
It would be nice for this list to be alphabetically sorted

## What users will benefit from this feature?
All users using k3s/k3d

## Are you able to implement this feature?
I can help with this

## Additional context
Slack thread describing the issue [here](https://kubernetes.slack.com/archives/C01FXB5E8ER/p1733765806831109)

PR #2692, opened 2024-12-13
frontend: k8s: Sort namespaces in UI

skoeva:
This change alphabetically sorts namespaces in the UI, addressing the case where namespace lists from k3d/k3s clusters are not sorted by default.

Fixes: #2656 

### Testing
- [X] Create a k3d cluster: `k3d cluster create`
- [X] Open the cluster in Headlamp and navigate to the "Namespaces" page
- [X] Ensure that all the namespaces are sorted alphabetically

image

- [X] Go back to the cluster home and click on the Namespace filter dropdown menu
- [X] Ensure that all the namespaces are sorted alphabetically

image

### DIFF
--- a/frontend/src/components/common/NamespacesAutocomplete.tsx
+++ b/frontend/src/components/common/NamespacesAutocomplete.tsx
@@ -157,7 +157,11 @@ function NamespacesFromClusterAutocomplete(
 ) {
   const [namespacesList] = Namespace.useList();
   const namespaceNames = useMemo(
-    () => namespacesList?.map(namespace => namespace.metadata.name) ?? [],
+    () =>
+      namespacesList
+        ?.map(namespace => namespace.metadata.name)
+        .slice()
+        .sort((a, b) => a.localeCompare(b)) ?? [],
     [namespacesList]
   );
 


frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
```diff
@@ -313,17 +313,29 @@ export function useKubeObjectList<K extends KubeObject>({
     queryParams: cleanedUpQueryParams,
   });
 
+  const sortedItems = useMemo(() => {
```
joaquimrocha, line 316: @sniok , I wonder if it's a good idea to special case Namespace here, instead of just where it's needed (UI) or overriding useList in KubeNamespace.

sniok (changes requested):
could you please move this logic to NamespaceAutocomplete component

frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
skoeva: it's also in the list on the namespace page but since there is no need beyond the UI for sorting, adjustments to the UI code should be fine

sniok (approved):
nice, thanks!

PR #2694, opened 2024-12-16
frontend: fix location import on mutli delete button

farodin91:


sniok:
could you describe what issue this fixes and how to test it please?

farodin91:
@sniok If you delete something with the multi if got redirect to a 404. tested in flatpak. 

As, I've implemented this I compared it with the single delete and here is my fix.

sniok (approved):

Issue #2679, opened 2024-12-13
Scrollbar size issue


sniok:
The sidebar scrollbar extends all the way up and is obscured by the top bar. Demo on the video:

https://github.com/user-attachments/assets/5dd36fa9-f482-423f-a79d-faa18745e82e

Also the 'main' scrollbar on the right could be shortened to not include the top bar as well but that's not critical.

And the narrow version gets squished by the scrollbar

![image](https://github.com/user-attachments/assets/c542924b-773c-4815-98a9-304be3883d4e)

## Environment (please provide info about your environment):
any

## Are you able to fix this issue?
maybe

skoeva:
looking into this now

PR #2715, opened 2024-12-19
frontend: Improve scroll bars

skoeva:
These changes display the scroll bars for the main content and sidebar under the top bar. The width of the sidebar and the height of the top bar are also adjusted to prevent inconsistencies.

Fixes: #2679

### Testing
- [X] Open Headlamp and resize the window height to display the scroll bars
- [X] Ensure that the icons in the side bar are not squished like before, and that the scroll bars are displayed under the top bar

image

### DIFF
--- a/frontend/src/components/App/Layout.tsx
+++ b/frontend/src/components/App/Layout.tsx
@@ -2,7 +2,6 @@ import { Box, Button } from '@mui/material';
 import Container from '@mui/material/Container';
 import CssBaseline from '@mui/material/CssBaseline';
 import Link from '@mui/material/Link';
-import { useTheme } from '@mui/material/styles';
 import { styled } from '@mui/material/styles';
 import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -103,7 +102,6 @@ export default function Layout({}: LayoutProps) {
   const isFullWidth = useTypedSelector(state => state.ui.isFullWidth);
   const { t } = useTranslation();
   const allClusters = useClustersConf();
-  const theme = useTheme();
 
   /** This fetches the cluster config from the backend and updates the redux store on an interval.
    * When stateless clusters are enabled, it also fetches the stateless cluster config from the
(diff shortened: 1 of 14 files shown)


frontend/src/components/App/Layout.tsx
```diff
@@ -202,12 +202,27 @@ export default function Layout({}: LayoutProps) {
...
         <TopBar />
         <Sidebar />
-        <Main id="main" sx={{ flexGrow: 1, marginLeft: 'initial' }}>
+        <Main
+          id="main"
+          sx={{
+            flexGrow: 1,
+            marginLeft: 'initial',
+            overflow: 'auto',
+            [theme.breakpoints.down('sm')]: { marginTop: { xs: '-12px', sm: '-8px' } },
```
sniok, line 223: that's a lot of new magic values (64px, 72px, -12px, -8px). maybe we can do this change without introducing them? maybe making topbar not positioned absolute or sticky. 
or if that's not possible at least introduce some constants and plan for a possibility that topbar size might be changed in the future

frontend/src/components/App/Layout.tsx
```diff
@@ -202,12 +202,27 @@ export default function Layout({}: LayoutProps) {
       >
         {t('Skip to main content')}
       </Link>
-      <Box sx={{ display: 'flex', [theme.breakpoints.down('sm')]: { display: 'block' } }}>
+      <Box
+        sx={{
+          display: 'flex',
+          overflow: 'auto',
+          paddingTop: { sm: '64px', xs: '72px' },
+          height: '100vh',
```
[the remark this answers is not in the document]
skoeva: idk if I see a difference but I suppose it's more appropriate to include here

frontend/src/components/App/Layout.tsx
skoeva: if the topbar isn't set to fixed then that defeats the purpose of it right? I think the new values are a good workaround (temporary if we can fix the topbar somehow) so I can abstract them into constants for now

sniok: if the actual content container is scrollable then topbar and sidebar can be positioned without 'fixed' or 'absolute' and will still stay in the same place. then we won't need to have any hardcoded values for positioning 

image

skoeva: good point. changed the topbar to 'static' which removes the need for the new vars in layout

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -275,6 +275,9 @@ export const PureSidebar = memo(
     linkArea,
   }: PureSidebarProps) => {
     const { t } = useTranslation();
+    const listContainerRef = React.useRef<HTMLDivElement | null>(null);
+    const [isOverflowing, setIsOverflowing] = React.useState(false);
+    const adjustedDrawerWidthClosed = isOverflowing ? 84 : 72;
```
sniok, line 280: this assumes that scrollbar is 12px wide but it depends on the platform

skoeva: you're right, now using a state to keep track of the scrollbar width and seems to work well across the electron app and browser

sniok: if you set width to the inner element that doesn't have overflow then you won't need it at all.
Here's a simplified example https://jsfiddle.net/syhp36qr/7/
in this example you can set width to `sidebar-inner` and have sidebar width to automatically include sidebar

illume (approved):
I rebased on main and tested it... nice improvement.

Thanks! 🎉🎈

PR #2724, opened 2024-12-24
frontend: Put apiVersion for individual items from list when fetching

### LINKED ISSUE #2719: Restart button doesn't work
## Describe the bug
After clicking the restart button, information appears that the Deployment/Stateful Set/Daemon Set has been restarted, but nothing happens. An error appears in the browser console.

## To Reproduce
Steps to reproduce the bug:
1. Go to Deployment/Stateful Set/Daemon
2. Click on restart button
3. You will get an error in the console

 - Installation type: Windows and In-Cluster
 - Headlamp Version: 0.27.0

No

## Additional Context
image

sniok:
Fixes #2719 
This adds apiVersion field for the individual items in the list.

When fetching a list apiVersion is only present on the root object but when fetching a single item apiVersion is present in the item itself, this discrepancy caused some issues.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
@@ -73,7 +73,11 @@ export function kubeObjectListQuery<K extends KubeObject>(
           }
         ).then(it => it.json());
         list.items = list.items.map(item => {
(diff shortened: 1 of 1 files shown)


skoeva (approved):
Works after testing, LGTM

PR #2727, opened 2025-01-03
frontend: Add TimeAgo component that displays and auto refreshes time

### LINKED ISSUE #2726: The "Age" column does not refresh
## Describe the bug
The "Age" column does not refresh. I have to manually refresh the browser for the time counter to change.

## To Reproduce
Steps to reproduce the bug:
1. Go to Pods
2. Delete some pod

 - Installation type: Windows-Winget/In-Cluster
 - Headlamp Version: 0.27.0

No

## Additional Context

https://github.com/user-attachments/assets/92c994b8-fa32-450e-9271-7f54b71a2589

sniok:
Fixes #2726 

This PR adds TimeAgo component that updates every second to display time since the given date. 
Only the date itself updates and the table is not re-rendered so there's pretty much no impact on performance.


joaquimrocha (approved):
LGTM. We may also be able to optimize this by understanding what is the right period to update this, based on the format + abbreviation.

PR #2730, opened 2025-01-06
auto-hide menubar (can still be recalled by holding Alt key)

jonassvatos:
The default Electron menu bar is visible even when there is nothing custom, eg. settings or so on.
This PR effectively hides it, with the option to recall it again using Alt key.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1040,6 +1040,7 @@ function startElecron() {
         contextIsolation: true,
         preload: `${__dirname}/preload.js`,
       },
+      autoHideMenuBar: true,
     });
 
     setMenu(mainWindow, currentMenu);


app/electron/main.ts
```diff
@@ -1040,6 +1040,7 @@ function startElecron() {
         contextIsolation: true,
         preload: `${__dirname}/preload.js`,
       },
+      autoHideMenuBar: true,
```
joaquimrocha, line 1043: This should be a setting. I imagine not all users want this behavior.

joaquimrocha (changes requested):
Thanks @jonassvatos . I think this behavior may not be what all users want, so we should add a menu to set it.
Also, please check out the git commit guidelines: https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines

skoeva (commented):
Hi! Are you still planning to continue work on this issue?

joaquimrocha:
This cannot be the default behavior. It would require some additional thought as to how to integrate this option and a setting to control. We may also soon change how the menu is done, and this would be influenced by that. Please open an issue if there is still interest in having this feature.

PR #2731, opened 2025-01-06
backend: prevent panic in WebSocket multiplexer

knrt10:
This fixes a race condition in the WebSocket multiplexer where sending a COMPLETE message could panic if the connection was closed. Added proper connection state checking and error handling to safely handle closed connections.

- Add connection state check before writing messages
- Return early if connection is already closed
- Add error logging for failed writes
- Properly handle WriteJSON errors

Ref: https://github.com/headlamp-k8s/headlamp/pull/2563#issuecomment-2572713349

### DIFF
--- a/backend/cmd/multiplexer.go
+++ b/backend/cmd/multiplexer.go
@@ -583,7 +583,6 @@ func (m *Multiplexer) sendIfNewResourceVersion(
 // sendCompleteMessage sends a COMPLETE message to the client.
 func (m *Multiplexer) sendCompleteMessage(conn *Connection, clientConn *websocket.Conn) error {
 	conn.mu.RLock()
-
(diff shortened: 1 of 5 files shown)


farodin91:
@knrt10 got the same issues again with this patch applied.

I hope the multiplexer patch another bug. i have but no.

knrt10:
@farodin91 Can you please provide more info? It would be great if you please help me reproducing this issue. Can you please help with your setup and what led to this error. Anything will help. Thanks

farodin91:
@knrt10 Could we call later?

knrt10:
Pinged you on k8s slack

illume (approved):
👍 thanks

PR #2739, opened 2025-01-10
frontend: Fix createRouteURL to match routes by name

yolossn:
The route lookup for storedRoutes in createRouteURL was matching based on path key, now correctly matches using the route's name property.

### DIFF
--- a/frontend/src/lib/router.tsx
+++ b/frontend/src/lib/router.tsx
@@ -866,7 +866,27 @@ export interface RouteURLProps {
 
 export function createRouteURL(routeName: string, params: RouteURLProps = {}) {
   const storeRoutes = store.getState().routes.routes;
-  const route = (storeRoutes && storeRoutes[routeName]) || getRoute(routeName);
+
+  // First try to find by name
+  const matchingStoredRouteByName =
+    storeRoutes &&
+    Object.entries(storeRoutes).find(
+      ([, route]) => route.name?.toLowerCase() === routeName.toLowerCase()
+    )?.[1];
+
+  // Then try to find by path
+  const matchingStoredRouteByPath =
+    storeRoutes &&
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):
That's a great catch! I am surprised we didn't run into problems before.
Did you test some of the links that are using createRouteURL? I wonder if this fix (as legitimate as it is) ends up actually breaking anyone's plugins... I will ping on Slack.

skoeva (approved):
thanks for catching this!

yolossn:
> I thought this was intentional? Can you confirm that this change doesn't break any existing links?

This change is effective only for the storedRoutes (ie routes that are added by the plugins) and it doesn't affect any existing links. The [KubeObjectLink](https://github.com/headlamp-k8s/headlamp/blob/e44b7fc31874177c99a22a90062019d3f38cabb6/frontend/src/components/common/Link.tsx#L33) that is used in the [ResourceListView](https://github.com/headlamp-k8s/headlamp/blob/e44b7fc31874177c99a22a90062019d3f38cabb6/frontend/src/components/common/Resource/ResourceListView.tsx#L28) component for linking the List item to the detail page of the item is not working for the custom objects that extend the [KubeObject](https://github.com/headlamp-k8s/headlamp/blob/e44b7fc31874177c99a22a90062019d3f38cabb6/frontend/src/lib/k8s/KubeObject.ts#L25) in plugins alone. 

This issue is leading to weird workarounds in plugins [example](https://github.com/headlamp-k8s/plugins/blob/b8ebea987a4777186bee7aefca1f3d66a73b7fdf/cert-manager/src/types/certificate.ts#L43).

joaquimrocha:
Let's please add a 2nd search on the previous logic that compared the path and print a deprecation notice when it matches. So we don't break any existing workarounds.

yolossn:
Update to add second match and print a deprecation warning.
image

Issue #2736, opened 2025-01-08
frontend: websocket fails to connect - wrong URL


vlasov-y:
## Describe the bug
I have added a console.log to check what is tried to be parsed.

![image](https://github.com/user-attachments/assets/b9d0e7e9-2234-49c3-a073-bd99f86d2ea3)

As you can see, URL does not contain a scheme and domain.

joaquimrocha:
@knrt10 / @illume could this be related to recent changes?

knrt10:
Thank you for opening the issue @vlasov-y. I can reproduce the issue but websocket works and it does not affect with data be in sync. I tried creating secret and deleting it and it works. But will look more why URL is constructed wrong.

Can you confirm if it is working for you or not?

vlasov-y:
It works, just spotted the error

knrt10:
Thanks, will fix the URL issue

PR #2744, opened 2025-01-13
frontend: Add missing base url in useWebSocket hook

sniok:
This fixes watching individual kubernetes objects.
 
How to fix:

1. Go to any resource details (Pods > any pod, for example)
2. Make sure there are no errors in console and websocket connection is established

Fixes: #2736


knrt10 (approved):
Tested. Thanks @sniok

PR #2745, opened 2025-01-13
frontend: CronJob details refactor

sniok:
Main point of this refactor is to use patch function when using suspend/resume button. This will only send a `{ spec: { suspend: true/false } }` update to the server instead of sending the whole object. Which will make it simpler and avoid any resourceVersion conflitcs. 

How to test:

 - Open any CronJob
 - Click on Suspend/Resume
 - Click on Spawn Job
 - Edit the CronJob

There shouldn't be any errors and should behave as expected

<details>
<summary>Sample CronJob yaml</summary>

```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: yearly-cronjob
  namespace: default
spec:
  schedule: "@yearly"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: yearly-job
            image: busybox:latest
            command:
            - /bin/sh
            - -c
            - "echo Happy New Year!; date"
          restartPolicy: OnFailure
```

</details>

### DIFF
--- a/frontend/src/components/cronjob/Details.tsx
+++ b/frontend/src/components/cronjob/Details.tsx
@@ -10,14 +10,13 @@ import {
   InputLabel,
 } from '@mui/material';
 import _ from 'lodash';
-import { useEffect, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useParams } from 'react-router-dom';
 import { apply } from '../../lib/k8s/apiProxy';
 import CronJob from '../../lib/k8s/cronJob';
 import Job from '../../lib/k8s/job';
-import { KubeObjectInterface } from '../../lib/k8s/KubeObject';
 import { clusterAction } from '../../redux/clusterActionSlice';
 import { AppDispatch } from '../../redux/stores/store';
 import { ActionButton } from '../common';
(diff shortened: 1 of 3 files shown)


joaquimrocha (changes requested):
This branch is actually introducing issues now. `main` works fine now with the recent fixes, but with this branch I notice:
- The suspend/resume button is not disabled when the request is being sent;
- The suspend/resume button and the suspend state in the table doesn't reflect the actual state of the resource

So it seems like the resource is not getting updated.

sniok:
Fixed disabling suspend/resume button. 
The state in the table now works, just needed a rebase on main

frontend/src/components/cronjob/Details.tsx
```diff
@@ -137,138 +120,83 @@ function SpawnJobDialog(props: {
...
+  const [jobs, jobsError] = Job.useList({ namespace });
+  const [cronJob] = CronJob.useGet(name, namespace);
+  const [isSpawnDialogOpen, setIsSpawnDialogOpen] = useState(false);
+  const [isPendingSuspend, setIsPendingSuspend] = useState(false);
+  const isCronSuspended = cronJob?.spec.suspend;
+
+  const ownedJobs =
+    jobs?.filter(job =>
+      job.metadata.ownerReferences?.find(ref => ref.kind === 'CronJob' && ref.name === name)
+    ) ?? [];
```
joaquimrocha, line 136: This one could be set with useMemo

joaquimrocha (approved):
Approved but there's something I think we should memoize.

Issue #2681, opened 2024-12-13
Set a fixed application size


Andrew1988123123:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
After starting application, I have to set application size again with Zoom in/out buttons. Reset Zoom button is not an option for me because application is still not properly adjusted.  Problem occurs only on desktop application.

## Describe the solution you'd like
Add a button that save application size.

## What users will benefit from this feature?
User will not have to set application size every time. 

## Are you able to implement this feature?
No

## Additional context
Add button here
![image](https://github.com/user-attachments/assets/cd08383a-95ac-48df-ae71-a777524e93d5)

joaquimrocha:
Rather than a button, we should remember the zoom after restarts.

joaquimrocha:
@Andrew1988123123 I have just tried this on our main branch on WSL and it did remember the zoom after I changed it and restarted the app. Can you give us more info about your environment? What version of Headlamp and what platform (Mac, Windows, Flatpak, etc.)?

joaquimrocha:
@Andrew1988123123 Nevermind my previous comment. I updated it.

skoeva:
Tested with the dev app, the zoom is kept because it's linked to devtools. The zoom persisted after restarting my entire machine and reset when I hit `Ctrl + 0`. Not sure if the main app can be synchronized with this but noticing some bugs with the changes to the app code, looking into this

skoeva:
Hi, working on this and testing different cases. Have you tried refreshing the app before restarting?

Also, what is the time difference between you changing the zoom level and closing/reopening the app?

Andrew1988123123:
Hi, 
1. Didn't help.
2. Just a few seconds. Always the same result.

The application returns to the default size most often when I restart or minimize it.

I forgot to add I'm using Windows 11 with increased scaling to 150%

![Image](https://github.com/user-attachments/assets/0e5922f4-50e7-4c4b-a5ab-32e57b639152)

When opening the application I use such settings.

![Image](https://github.com/user-attachments/assets/45c34b8b-cce7-428f-bb38-e9cf287ddd46)

skoeva:
Do you still see the issue if you wait ~10 seconds before closing the app? I noticed that the zoom level is stored locally but that it takes some time to automatically be saved. Depending on how the app is being run this may be a temp workaround but lmk if that does not work

Andrew1988123123:
[...]

PR #2746, opened 2025-01-13
app: Remember zoom level after restarts

skoeva:
This change refactors the zoom functionality in the Electron app to remember the zoom level in the Electron app after restarts.

Fixes: #2681 

### Testing
- [X] Open the app and change the zoom (i.e. max zoom)
- [X] Close and restart the app
- [X] Ensure that the zoom persists from the previous session

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -783,18 +783,21 @@ function getDefaultAppMenu(): AppMenu[] {
         sep,
         {
           label: i18n.t('Reset Zoom'),
-          role: 'resetzoom',
           id: 'original-reset-zoom',
+          accelerator: 'CmdOrCtrl+0',
+          click: () => setZoom(1.0),
         },
         {
           label: i18n.t('Zoom In'),
-          role: 'zoomin',
           id: 'original-zoom-in',
+          accelerator: 'CmdOrCtrl+Plus',
+          click: () => adjustZoom(0.1),
         },
         {
           label: i18n.t('Zoom Out'),
-          role: 'zoomout',
           id: 'original-zoom-out',
+          accelerator: 'CmdOrCtrl+-',
+          click: () => adjustZoom(-0.1),
         },
         sep,
         {
@@ -1030,6 +1033,42 @@ function killProcess(pid: number) {
   }
 }
 
+const ZOOM_FILE_PATH = path.join(app.getPath('userData'), 'headlamp-config.json');
+let cachedZoom: number = 1.0;
+
+function saveZoomFactor(factor: number) {
+  try {
+    fs.writeFileSync(ZOOM_FILE_PATH, JSON.stringify({ zoomFactor: factor }), 'utf-8');
+  } catch (err) {
+    console.error('Failed to save zoom factor:', err);
+  }
+}
+
+function loadZoomFactor() {
+  try {
+    const { zoomFactor = 1.0 } = JSON.parse(fs.readFileSync(ZOOM_FILE_PATH, 'utf-8'));
+    return zoomFactor;
+  } catch (err) {
+    console.error('Failed to load zoom factor, defaulting to 1.0:', err);
+    return 1.0;
+  }
+}
+
+// The zoom factor should respect the fixed limits set by Electron.
+function clampZoom(factor: number) {
+  return Math.min(5.0, Math.max(0.25, factor));
+}
+
+function setZoom(factor: number) {
+  cachedZoom = factor;
+  mainWindow?.webContents.setZoomFactor(cachedZoom);
+}
+
+function adjustZoom(delta: number) {
+  const newZoom = clampZoom(cachedZoom + delta);
+  setZoom(newZoom);
+}
+
 function startElecron() {
   console.info('App starting...');
 
@@ -1081,6 +1120,11 @@ function startElecron() {
       if (!!goForwardMenu) {
         goForwardMenu.enabled = mainWindow?.webContents.canGoForward() || false;
       }
+    });
+
+    mainWindow.webContents.on('did-finish-load', () => {
+      const startZoom = loadZoomFactor();
+      setZoom(startZoom);
     });
 
     mainWindow.webContents.on('dom-ready', () => {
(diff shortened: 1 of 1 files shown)


joaquimrocha:
@skoeva I just tried reproducing the original issue on WSL and I couldn't. After trying a couple of things, turns out the zoom is kept if we use the `npm run dev-only-app`, which doesn't load the HTML as with the built app but uses the running processes. Still I am puzzled that it initializes the web view in that new process anyway but apparently keeps the zoom... Anyway, maybe worth investigating in case the fix has a less manual solution.

skoeva:
Looks like we cannot modify the zoom functionality with the existing setup, refactored the logic within the app file to address this

sniok:
Electron support saving and restoring zoom level by default.

Electron stores Zoom level in a file 'Preferences'. On windows it's in the folder `C:\Users\<user>\AppData\Roaming\Headlamp`. 

Example:
```
{"per_host_zoom_levels":
  {"7164555560649993444":
   {"file:///C:/Users/<user>/AppData/Local/Programs/Headlamp/resources/frontend/index.html?backendToken=707d388ccf5c3a2b49c33047393346e0ee5b783d4fdcc55b234a27d8800b6855#/":1.5, ...
```

the issue is that it is being stored per url and backendToken is a part of that url. Since backendToken is unique and generated each time the app is started then it's technically a different 'page' and zoom level is 1.0 by default.

joaquimrocha:
Thanks for the check @sniok ! This issue then has turned into "how do we make sure we use the backend token while not making it part of the URL"

skoeva:
Dropped previous commit, now removing backendToken from the url to fix this issue

app/package.json
```diff
@@ -170,6 +170,7 @@
     "copyfiles": "^2.4.1",
     "cross-env": "^7.0.3",
     "dotenv": "^16.4.5",
+    "electron-browser-storage": "^1.0.7",
```
joaquimrocha, line 173: I'd prefer not to add another dependency for this. I prefer in the end to save it using a file as you had suggested than introducing another dependency. My main problem with a preferences file is that then we have 2 files related to preferences. Maybe we actually remove the default one as a clean up action when we quit the app (assuming Electron has no way to prevent this by default)...

joaquimrocha (changes requested):

app/electron/main.ts
```diff
@@ -1030,6 +1033,43 @@ function killProcess(pid: number) {
...
+    fs.writeFileSync(ZOOM_FILE_PATH, JSON.stringify({ zoomFactor: factor }), 'utf-8');
+  } catch (err) {
+    console.error('Failed to save zoom factor:', err);
+  }
+}
+
+function loadZoomFactor() {
+  try {
+    const data = fs.readFileSync(ZOOM_FILE_PATH, 'utf-8');
+    return JSON.parse(data) || 1.0;
```
joaquimrocha, line 1049: This is not correctly returning the values. There's even an error when loading the app because this returns `{zoomFactor: SOME_NUMBER}` instead of a number.

app/electron/main.ts
```diff
@@ -1030,6 +1033,43 @@ function killProcess(pid: number) {
...
+  } catch (err) {
+    console.error('Failed to load zoom factor, defaulting to 1.0:', err);
+    return 1.0;
+  }
+}
+
+function clampZoom(factor: number) {
+  return Math.min(5.0, Math.max(0.25, factor));
```
joaquimrocha, line 1057: If these limits are due to some electron convention or its own limits, please add a comment explaining.

app/electron/main.ts
```diff
@@ -1030,6 +1033,43 @@ function killProcess(pid: number) {
   }
 }
 
+const ZOOM_FILE_PATH = path.join(app.getPath('userData'), 'zoomconfig.json');
```
joaquimrocha, line 1036: Let's make this a headlamp-config.json so in the future we may choose to save more preferences there without having to rename + migrate.

joaquimrocha (changes requested):
There's a bug and the file should be renamed, but it's close now.,

app/electron/main.ts
skoeva: done

app/electron/main.ts
skoeva: done

app/electron/main.ts
skoeva: pretty sure I understand what you mean now, ran `make frontend` and `npm run dev` and there are no popups on my end

app/electron/main.ts
```diff
@@ -783,18 +783,21 @@ function getDefaultAppMenu(): AppMenu[] {
         sep,
         {
           label: i18n.t('Reset Zoom'),
-          role: 'resetzoom',
           id: 'original-reset-zoom',
+          accelerator: 'CmdOrCtrl+0',
+          click: () => saveZoomFactor(1.0),
```
joaquimrocha, line 788: This is causing a regression: the zoom doesn't change because it's being saved but not adjusted.

joaquimrocha (changes requested):
Found a regression on resetting the zoom. But otherwise it's working fine! Thanks.

app/electron/main.ts
skoeva: Right, thanks for catching this. Pushing up the fix now

joaquimrocha (changes requested):
Apparently I had reviewed yesterday but failed to send...
Just a couple more changes so it's solid.

app/electron/main.ts
```diff
@@ -783,18 +783,24 @@ function getDefaultAppMenu(): AppMenu[] {
         sep,
         {
           label: i18n.t('Reset Zoom'),
-          role: 'resetzoom',
           id: 'original-reset-zoom',
+          accelerator: 'CmdOrCtrl+0',
+          click: () => {
+            saveZoomFactor(1.0);
+            mainWindow?.webContents.setZoomFactor(1.0);
```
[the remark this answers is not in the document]
skoeva: Removed the additional write here, no need to create a new function

joaquimrocha (approved):

PR #2747, opened 2025-01-13
fix the cluster info for headlamp_info to inlcude originalName

vyncent-t:
Fixes issue #2750 

### Description

This PR aims to expands the current implementation of the `headlamp_info` extension that is created to provide additional information about a cluster within the kubeconfig. 

### Changes
- the `headlamp_info` extension is now created at the home view of the app (it used to only be created if we rename a cluster)
- the `headlamp_info` extension now includes a field for storing the original name of a cluster `originalName`
- for clusters that have been renamed before these changes were introduced, the `headlamp_info` is now updated to include the `originalName` field and correct val
- prepares the `headlamp_info` field for future cluster renaming by introducing the `customName` field along with `originalName`

### How to test

⚠️  IMPORTANT ⚠️ :  please be aware that this PR changes config data within the kubeconfig, I recommend copying any local kubeconfig files or have the undo button ready with your kubeconfig pulled up in your IDE

- locate kubeconfig file
- open headlamp app and navigate to the home view where all clusters are visible
- in kubeconfig, observe that all existing clusters now have `headlamp_info` field and `originalName` field containing the clusters name
- create a new cluster
- let app load new cluster to home view
- view changes for the new cluster

Context notes:

**Non dynamic clusters**
- I think its best to keep the changes in the home view where we use getCluster func from the backend (this is called on the home view of the app) since this would then populate the headlamp_info field and append the original name field, this would make it so that we can use the original name field along with the custom name field to delete a cluster from the home view

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1170,6 +1170,16 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		contextSouce := context.SourceStr()
+
+		// this function handles the non dynamic clusters that need to have headlamp_info created or updated
+		if contextSouce == "kubeconfig" {
+			if err := c.handleHeadlampInfo(context); err != nil {
+				logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+					err, "handling headlamp_info")
+			}
+		}
+
 		clusters = append(clusters, Cluster{
 			Name:     context.Name,
 			Server:   context.Cluster.Server,
@@ -1183,6 +1193,72 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	}
 
 	return clusters
+}
+
+// handleHeadlampInfo this function handles the clusters that need to have headlamp_info created or updated.
+func (c *HeadlampConfig) handleHeadlampInfo(context *kubeconfig.Context) error {
+	// the headlamp_info is stored in the extensions of the kubeconfig context
+	headlampInfo := context.KubeContext.Extensions["headlamp_info"]
+
+	// if the headlamp_info is not present we create it and keep the original cluster name
+	// else the headlamp_info extension exists we must update it with the original name
+	if headlampInfo == nil {
+		err := c.handleMissingHeadlampInfo(context)
+		if err != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+				err, "handle missing headlamp_info")
+
+			return err
+		}
+	} else {
+		err := c.handleUpdateHeadlampInfo(context)
+		if err != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+				err, "handle update headlamp_info")
+
+			return err
+		}
+	}
+
+	return nil
+}
+
+// handleUpdateHeadlampInfo this function handles the clusters that need to have headlamp_info updated.
+func (c *HeadlampConfig) handleUpdateHeadlampInfo(context *kubeconfig.Context) error {
+	// the headlamp_info is stored in the extensions of the kubeconfig context
+	headlampInfo := context.KubeContext.Extensions["headlamp_info"]
+
+	existingObj, marshalErr := MarshalCustomObject(headlampInfo, context.Name)
+	if marshalErr != nil {
+		logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+			marshalErr, "unmarshal headlamp_info")
+		return marshalErr
+	}
+
+	// if original name is not present and the custom name is, update the headlamp_info with the original name.
+	if existingObj.OriginalName == "" && existingObj.CustomName != "" {
+		keepErr := c.keepOriginalClusterName(context.Name, context.SourceStr(), context, UpdateLegacy)
+		if keepErr != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+				keepErr, "update original cluster name and custom name on headlamp_info")
+			return keepErr
+		}
+	}
+
+	return nil
+}
+
+// handleMissingHeadlampInfo function handles the clusters that need to have headlamp_info created.
+func (c *HeadlampConfig) handleMissingHeadlampInfo(context *kubeconfig.Context) error {
+	// if the headlamp_info is not present we create it and keep the original cluster name
+	err := c.keepOriginalClusterName(context.Name, context.SourceStr(), context, CreateNew)
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+			err, "keep original cluster name")
+		return err
+	}
+
+	return nil
 }
 
 // parseCustomNameClusters parses the custom name clusters from the kubeconfig.
@@ -1499,6 +1575,54 @@ func (c *HeadlampConfig) handleStatelessClusterRename(w http.ResponseWriter, r *
 	c.getConfig(w, r)
 }
 
+/*
+* nameToExtensions writes or updates the original name to the Extensions map in the kubeconfig.
+ */
+func nameToExtensions(context *kubeconfig.Context, contextName string, path string, updateMode infoUpdateMode) error {
+	// get the context with the given cluster name
+	headlampInfo := context.KubeContext.Extensions["headlamp_info"]
+
+	switch updateMode {
+	case CreateNew:
+		if headlampInfo == nil {
+			customObj := &kubeconfig.CustomObject{
+				TypeMeta: v1.TypeMeta{
+					Kind:       "HeadlampInfo",
+					APIVersion: "v1",
+				},
+				ObjectMeta:   v1.ObjectMeta{},
+				OriginalName: contextName,
+				SourcePath:   path,
+			}
+
+			context.KubeContext.Extensions["headlamp_info"] = customObj
+
+			return nil
+		}
+	case UpdateLegacy:
+		if headlampInfo != nil {
+			existingObj, err := MarshalCustomObject(headlampInfo, contextName)
+			if err != nil {
+				logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+					err, "unmarshaling headlamp_info")
+				return err
+			}
+
+			existingObj.OriginalName = contextName
+			existingObj.SourcePath = path
+
+			context.KubeContext.Extensions["headlamp_info"] = &existingObj
+
+			return nil
+		}
+
+	default:
+		return nil
+	}
+
+	return nil
+}
+
 // customNameToExtenstions writes the custom name to the Extensions map in the kubeconfig.
 func customNameToExtenstions(config *api.Config, contextName, newClusterName, path string) error {
 	var err error
@@ -1586,6 +1710,42 @@ func (c *HeadlampConfig) getPathAndLoadKubeconfig(source, clusterName string) (s
 	}
 
 	return path, config, nil
+}
+
+// UpdateMode specifies how the headlamp_info extension should be handled.
+type infoUpdateMode int
+
+const (
+	// CreateNew creates a new headlamp_info extension if none exists.
+	CreateNew infoUpdateMode = iota
+	// UpdateLegacy updates existing headlamp_info for legacy clusters.
+	UpdateLegacy
+)
+
+/* handler to keep the original name for a cluster.*/
+func (c *HeadlampConfig) keepOriginalClusterName(
+	contextName string,
+	source string,
+	context *kubeconfig.Context,
+	updateMode infoUpdateMode,
+) error {
+	// // get path of kubeconfig from source
+	path, err := c.getKubeConfigPath(source)
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+			err, "getting kubeconfig file")
+
+		return err
+	}
+
+	if err := nameToExtensions(context, contextName, path, updateMode); err != nil {
+		logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+			err, fmt.Sprintf("writing custom extension to kubeconfig %s: %v", contextName, context.Extensions["headlamp_info"]))
+
+		return err
+	}
+
+	return nil
 }
 
 // Handler for renaming a cluster.
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -16,6 +16,8 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"github.com/headlamp-k8s/headlamp/backend/pkg/cache"
@@ -23,6 +25,7 @@ import (
 	"github.com/headlamp-k8s/headlamp/backend/pkg/kubeconfig"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 )
 
@@ -633,6 +636,8 @@ func TestHandleClusterAPI_XForwardedHost(t *testing.T) {
 	assert.Equal(t, "OK", rr.Body.String())
 }
 
+// test for keeping the original name of a cluster
+
 func TestRenameCluster(t *testing.T) {
 	kubeConfigByte, err := os.ReadFile("./headlamp_testdata/kubeconfig")
 	require.NoError(t, err)
@@ -1060,6 +1065,130 @@ func TestHandleClusterHelm(t *testing.T) {
 			router.ServeHTTP(w, req)
 
 			assert.Equal(t, tc.expectedStatus, w.Code)
+		})
+	}
+}
+
+// test for handleHeadlampInfo, currently checks for headlamp_info field and headlamp_info.originalName field.
+//
+//nolint:funlen
+func TestHandleHeadlampInfo(t *testing.T) {
+	// Define test cases
+	tests := []struct {
+		name          string
+		configFile    string
+		expectedError bool
+		validateFunc  func(*testing.T, *kubeconfig.Context)
+	}{
+		{
+			name:       "Empty headlamp_info",
+			configFile: "kubeconfig_headlampinfo",
+			validateFunc: func(t *testing.T, ctx *kubeconfig.Context) {
+				assert.NotNil(t, ctx.KubeContext.Extensions["headlamp_info"], "headlamp_info field should be present")
+			},
+		},
+		{
+			name:       "Existing valid headlamp_info original name",
+			configFile: "kubeconfig_headlampinfo",
+			validateFunc: func(t *testing.T, ctx *kubeconfig.Context) {
+				ext, ok := ctx.KubeContext.Extensions["headlamp_info"]
+				require.True(t, ok, "headlamp_info not found in Extensions")
+
+				// Convert the runtime.Object to a map
+				unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ext)
+				require.NoError(t, err, "failed to convert runtime.Object to map")
+
+				headlampInfo := unstructuredMap
+
+				// Check the originalName field
+				originalName, ok := headlampInfo["originalName"].(string)
+				require.True(t, ok, "headlamp_info.originalName missing or invalid")
+				assert.NotEmpty(t, originalName, "headlamp_info.originalName should not be empty")
+			},
+		},
+		{
+			name:       "Existing valid headlamp_info source path",
+			configFile: "kubeconfig_headlampinfo",
+			validateFunc: func(t *testing.T, ctx *kubeconfig.Context) {
+				ext, ok := ctx.KubeContext.Extensions["headlamp_info"]
+				require.True(t, ok, "headlamp_info not found in Extensions")
+
+				// Convert the runtime.Object to a map
+				unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ext)
+				require.NoError(t, err, "failed to convert runtime.Object to map")
+
+				headlampInfo := unstructuredMap
+
+				// Check the sourcePath field
+				sourcePath, ok := headlampInfo["sourcePath"].(string)
+				require.True(t, ok, "headlamp_info.sourcePath missing or invalid")
+				assert.NotEmpty(t, sourcePath, "headlamp_info.sourcePath should not be empty")
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// create a temp directory for our test kubeconfig
+			tmpDir := t.TempDir()
+			mockConfigFile := filepath.Join(tmpDir, "config")
+
+			// copy test data to the temp directory
+			testDataPath := filepath.Join("headlamp_testdata", tt.configFile)
+			testData, err := os.ReadFile(testDataPath)
+			require.NoError(t, err, "failed to read test data for '%s'", tt.configFile)
+
+			err = os.WriteFile(mockConfigFile, testData, 0o600)
+			require.NoError(t, err, "failed to write test kubeconfig")
+
+			// load the kubeconfig file
+			config, err := clientcmd.LoadFromFile(mockConfigFile)
+			require.NoError(t, err, "failed to load kubeconfig")
+
+			// Create a HeadlampConfig object
+			c := &HeadlampConfig{
+				useInCluster:    false,
+				cache:           cache.New[interface{}](),
+				kubeConfigStore: kubeconfig.NewContextStore(),
+				kubeConfigPath:  mockConfigFile,
+			}
+
+			// create a slice to hold the kubeconfig.Context objects
+			kubeContexts := make([]*kubeconfig.Context, 0, len(config.Contexts))
+
+			// loop through each context and push the kubeContext object into the slice
+			for contextName, context := range config.Contexts {
+				kubeContext := &kubeconfig.Context{
+					Name:        contextName,
+					KubeContext: context,
+					Source:      1,
+					Cluster:     config.Clusters[context.Cluster],
+					AuthInfo:    config.AuthInfos[context.AuthInfo],
+				}
+				kubeContexts = append(kubeContexts, kubeContext)
+			}
+
+			// use the kubeContexts slice for further processing
+			for _, ctx := range kubeContexts {
+				// Call handleHeadlampInfo
+				err = c.handleHeadlampInfo(ctx)
+				if err != nil {
+					fmt.Println("Error:", err)
+				}
+			}
+
+			// use the kubeContexts slice for further processing
+			for _, ctx := range kubeContexts {
+				if tt.expectedError {
+					require.Error(t, err, "expected an error but got none")
+				} else {
+					require.NoError(t, err, "failed to handle headlamp_info")
+
+					if tt.validateFunc != nil {
+						tt.validateFunc(t, ctx)
+					}
+				}
+			}
 		})
 	}
 }
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -59,6 +59,10 @@ type CustomObject struct {
 	metav1.TypeMeta
 	metav1.ObjectMeta
 	CustomName string `json:"customName"`
+	// originalName is the name the cluster was created with.
+	OriginalName string `json:"originalName"`
+	// sourcePath is the path to the source file.
+	SourcePath string `json:"sourcePath"`
 }
 
 // DeepCopyObject returns a copy of the CustomObject.
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -45,6 +45,7 @@ type Context struct {
 	proxy       *httputil.ReverseProxy `json:"-"`
 	Internal    bool                   `json:"internal"`
 	Error       string                 `json:"error"`
+	Extensions  map[string]interface{} `json:"extensions"`
 }
 
 type OidcConfig struct {
--- a/backend/cmd/headlamp_testdata/kubeconfig_headlampinfo
+++ b/backend/cmd/headlamp_testdata/kubeconfig_headlampinfo
@@ -1,0 +1,114 @@
+apiVersion: v1
+clusters:
+  - cluster:
+      certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCg== # Example base64-encoded data
+      extensions:
+        - extension:
+            last-update: Mon, 26 Dec 2022 20:33:03 IST
+            provider: testkubex.sigs.k8s.io
+            version: v1.28.0
+          name: cluster_info
+      server: PLACEHOLDER_SERVER_URL_X
+    name: testkubex
+  - cluster:
+      certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCg== # Example base64-encoded data
+      extensions:
+        - extension:
+            last-update: Tue, 27 Dec 2022 20:33:03 IST
+            provider: testkubey.sigs.k8s.io
+            version: v1.28.0
+          name: cluster_info
+      server: PLACEHOLDER_SERVER_URL_Y
+    name: testkubey
+  - cluster:
+      certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCg== # Example base64-encoded data
+      extensions:
+        - extension:
+            last-update: Wed, 28 Dec 2022 20:33:03 IST
+            provider: testkubez.sigs.k8s.io
+            version: v1.28.0
+          name: cluster_info
+      server: PLACEHOLDER_SERVER_URL_Z
+    name: testkubez
+contexts:
+  - context:
+      cluster: testkubex
+      extensions:
+        - extension:
+            last-update: Mon, 26 Dec 2022 20:33:03 IST
+            provider: testkubex.sigs.k8s.io
+            version: v1.28.0
+          name: context_info
+        - extension:
+            creationTimestamp: null
+            customName: testkubexworks
+          name: headlamp_info
+      namespace: default
+      user: testkubex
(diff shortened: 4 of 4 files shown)


vyncent-t:
progress: able now to have the cluster generate a headlamp_info field when it is created and gathered via the `getCluster` changes 

need to fix:
- unable to keep the previous name in the custom headlamp_info field after renaming it

vyncent-t:
note:

done
- [x] non dynamic clusters now have `headlamp_info` extension created at home view
- [x] dynamic clusters now have `headlamp_info` extension created at home view

to do

- [ ] create tests for new functions after reviewed

vyncent-t:
let me know what you guys think for these changes before I move on to create some tests 😎  @knrt10 @yolossn

vyncent-t:
note: previous push fixes logic for clusters that were renamed already (with the headlamp_info already there) to update the headlamp_info to have their original name saved also

knrt10:
> Currently, on main if you create a cluster and rename it and then you console log the clusters that are being rendered at the home page, you will see that the "name" field populates the changed name while the kubeconfig still holds the old name

@vyncent-t Thank you for the PR. I am not sure what this would solve. This was the intended change. Our approach initially was we do want to change the kubeconfig name of user because if user change the name in headlamp does not mean they want to change their kubeconfig. For example if they get their clusters using kubeclt they should still see the same name. 

This is the reason we do not change the name field in kubeconfig, rather than we edit the headlmap_info extension in the metadata. So when user renames the cluster only that thing is changed.

Regarding clusters that use the same name. That I think should not be possible because we automatically throw an error if there are same context name. Also, if there are same name in kubeconfig that is automatically rejected by kubectl as well

vyncent-t:
> > Currently, on main if you create a cluster and rename it and then you console log the clusters that are being rendered at the home page, you will see that the "name" field populates the changed name while the kubeconfig still holds the old name
> 
> @vyncent-t Thank you for the PR. I am not sure what this would solve. This was the intended change. Our approach initially was we do want to change the kubeconfig name of user because if user change the name in headlamp does not mean they want to change their kubeconfig. For example if they get their clusters using kubeclt they should still see the same name.
> 
> This is the reason we do not change the name field in kubeconfig, rather than we edit the headlmap_info extension in the metadata. So when user renames the cluster only that thing is changed.
> 
> Regarding clusters that use the same name. That I think should not be possible because we automatically throw an error if there are same context name. Also, if there are same name in kubeconfig that is automatically rejected by kubectl as well

@knrt10 thanks for getting back to me 😎 

allow me to clarify a few things and feel free to let me know if we should sync up to cover this

This PR is a prerequisite for continuing the progress of the [delete non dynamic clusters PR](https://github.com/headlamp-k8s/headlamp/pull/2567#discussion_r1907712414) that you had reviewed for me earlier (along with two other issues that stem from this that I will explain later). 

In the delete non dynamic clusters PR's current state, the delete function needs a name in which to find and delete that cluster, however for clusters that have been renamed, that new customName is what is being being used in the app and thus when we go to delete that cluster on the app the customName is what is being passed into the delete function. 

The goal of this PR is to expand the `headlamp_info` extension so that it will preserve the original name of the cluster, while also having the `headlamp_info` extension be created up front as opposed to only being created if we rename a cluster.

This PR does not change the existing main data of a kubeconfig context, it only creates the `headlamp_info` extension and saves the clusters name to it. I have also added logic that would check if a cluster has already been renamed in the app, which means `headlamp_info` already exists on that cluster, and thus that section is just updated to include `originalName: [ognamehere]`

I will include more information about what this PR does and how to test it in the description

As for the clusters using the same name and renaming- I am also working on two issues that spawn from the current name handling for clusters. Currently we are able to create a cluster and rename it to the name of another cluster, causing the renamed cluster link to 'consume' the og cluster link. There is also the issue of creating a cluster with the og name of an older renamed cluster causing duplicate links. These were able to be done through renaming use and no errors stop this. I am also working on fixes for both of these but they too require this `headlamp_info` improvement

---

TLDR: I have been tasked with expanding the `headlamp_info` extension to also include the original name of a cluster so that we can access that information and use it readily for both the backend and the UI. It is also needed in PRs that are tied to this one. Currently on main the `headlamp_info` extension is only created if we rename a cluster, this PR just creates the `headlamp_info` extension at the home view and stores the original name under originalName

knrt10:
Thank you for explaining. It makes sense now. Will review it now.

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1126,35 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+func (c *HeadlampConfig) handleHeadlampInfo(context *kubeconfig.Context) error {
```
knrt10, line 1129: can you please add comment what this do?

backend/cmd/headlamp.go
```diff
@@ -1526,6 +1635,44 @@ func (c *HeadlampConfig) getPathAndLoadKubeconfig(source, clusterName string) (s
 	return path, config, nil
 }
 
+// handler to keep the original name for a cluster.
+func (c *HeadlampConfig) keepOriginalClusterName(source string, contextName string, isUpdateNeeded bool) error {
```
knrt10, line 1639:
```suggestion
func (c *HeadlampConfig) keepOriginalClusterName(source, contextName string, isUpdateNeeded bool) error {
```

knrt10: The current name `isUpdateNeeded` is vague. WDYT about `shouldUpdateExistingExtension`. Also instead of a boolean flag, we could use an enum approach:

```go
// UpdateMode specifies how the headlamp_info extension should be handled
type HeadlampInfoUpdateMode int

const (
    // CreateNew creates a new headlamp_info extension if none exists
    CreateNew HeadlampInfoUpdateMode = iota
    // UpdateLegacy updates existing headlamp_info for legacy clusters
    UpdateLegacy
)

// keepOriginalClusterName updates or creates the headlamp_info extension with the original cluster name
func (c *HeadlampConfig) keepOriginalClusterName(source, contextName string, updateMode HeadlampInfoUpdateMode) error {
    // rest code
}
```

You could use it like

```go
c.keepOriginalClusterName(context.SourceStr(), context.Name, CreateNew)     // new extension
c.keepOriginalClusterName(context.SourceStr(), context.Name, UpdateLegacy)  // update legacy
```

backend/cmd/headlamp.go
```diff
@@ -1526,6 +1635,44 @@ func (c *HeadlampConfig) getPathAndLoadKubeconfig(source, clusterName string) (s
...
+	// establish the headlamp info field if it does not exist
+	for key, val := range config.Contexts {
+		originalClusterName = contextName
+
+		if err := originalNameToExtensions(config, originalClusterName, path, isUpdateNeeded); err != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+				err, "writing custom extension to kubeconfig "+key)
+
+			logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+				err, fmt.Sprintf("writing custom extension to kubeconfig %s", val.Extensions["headlamp_info"]))
```
knrt10, line 1667: 2 loggers here.

backend/cmd/headlamp.go
```diff
@@ -1526,6 +1635,44 @@ func (c *HeadlampConfig) getPathAndLoadKubeconfig(source, clusterName string) (s
...
+		logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+			err, "getting kubeconfig file")
+
+		return err
+	}
+
+	if isUpdateNeeded {
+		// log the current isUpdatedNeeded
+		logger.Log(logger.LevelInfo, map[string]string{"cluster": contextName},
+			nil, fmt.Sprintf("isUpdateNeeded zzzzzzzzzzzzzzzzzzzzzz: %t", isUpdateNeeded))
```
knrt10, line 1652: hehe. I hope this is removed in final iteration :-)

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1126,35 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
+		err := c.keepOriginalClusterName(context.SourceStr(), context.Name, false)
+		if err != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+				err, "failed to keep original cluster name")
+			return err
+		}
+	}
+
+	// this handles the clusters that have been renamed before the headlamp_info was introduced with original name
+	if headlampInfo != nil {
```
knrt10, line 1143: This check is redundant because we already know headlampInfo is not nil here (due to the previous if-block). We could restructure this as an else-if block.

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1126,35 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
+		existingObj, marshalErr := MarshalCustomObject(headlampInfo, context.Name)
```
knrt10, line 1144: The marshalErr is silently ignored if there's an error. We should log and handle this error.

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1126,35 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
+		if marshalErr == nil && existingObj.OriginalName == "" && existingObj.CustomName != "" {
```
knrt10, line 1145: This condition checks for a specific state but isn't clearly documented. Can you please add comment for this?

backend/cmd/headlamp.go
knrt10, line 1129: A better suggestion to improve this function

```suggestion
func (c *HeadlampConfig) handleHeadlampInfo(context *kubeconfig.Context) error {
    headlampInfo := context.KubeContext.Extensions["headlamp_info"]

    // Create new headlamp_info if it doesn't exist
    if headlampInfo == nil {
        return c.keepOriginalClusterName(context.SourceStr(), context.Name, CreateNew)
    }

    // Parse existing headlamp_info
    existingInfo, err := MarshalCustomObject(headlampInfo, context.Name)
    if err != nil {
        logger.Log(logger.LevelError, 
            map[string]string{"cluster": context.Name},
            err, 
            "failed to parse existing headlamp_info")
        return fmt.Errorf("parsing headlamp_info: %w", err)
    }

    // Handle legacy clusters that were renamed before originalName was introduced
    if isLegacyRenamedCluster(existingInfo) {
        return c.keepOriginalClusterName(context.SourceStr(), context.Name, UpdateLegacy)
    }

    return nil
}

// isLegacyRenamedCluster checks if this is a cluster that was renamed
// before the originalName field was introduced
func isLegacyRenamedCluster(info *HeadlampInfo) bool {
    return info.OriginalName == "" && info.CustomName != ""
}
```

backend/cmd/headlamp.go
```diff
@@ -1437,6 +1469,69 @@ func (c *HeadlampConfig) handleStatelessClusterRename(w http.ResponseWriter, r *
...
+// originalNameToExtensions writes the original name to the "headlamp_info" Extensions map in the kubeconfig.
+func originalNameToExtensions(config *api.Config, contextName string, path string, isNeedUpdate bool) error {
+	var err error
+	// get the context with the given cluster name
+	contextConfig, ok := config.Contexts[contextName]
+	if !ok {
+		logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
+			err, "ERROR setting originalName - failed getting context from kubeconfig for original name.")
+
+		return err
```
knrt10, line 1481: There seems to be error handling issue. err is nil here!

backend/cmd/headlamp.go
```diff
@@ -1437,6 +1469,69 @@ func (c *HeadlampConfig) handleStatelessClusterRename(w http.ResponseWriter, r *
 	c.getConfig(w, r)
 }
 
+// originalNameToExtensions writes the original name to the "headlamp_info" Extensions map in the kubeconfig.
+func originalNameToExtensions(config *api.Config, contextName string, path string, isNeedUpdate bool) error {
```
knrt10, line 1473:
```suggestion
// UpdateHeadlampInfoExtension manages the headlamp_info extension in the kubeconfig context
func UpdateHeadlampInfoExtension(config *api.Config, contextName string, path string, mode HeadlampInfoUpdateMode) error {
    // Get context configuration
    contextConfig, exists := config.Contexts[contextName]
    if !exists {
        return fmt.Errorf("context %q not found in kubeconfig", contextName)
    }

    // Get existing headlamp_info
    headlampInfo := contextConfig.Extensions["headlamp_info"]

    var customObj *kubeconfig.CustomObject
    var err error

    switch {
    case headlampInfo == nil:
        // Create new headlamp_info
        customObj = &kubeconfig.CustomObject{
            TypeMeta: v1.TypeMeta{
                Kind:       "HeadlampInfo",
                APIVersion: "v1",
            },
            ObjectMeta:   v1.ObjectMeta{},
            OriginalName: contextName,
        }
    case mode == UpdateLegacy:
        // Update existing headlamp_info
        customObj, err = MarshalCustomObject(headlampInfo)
        if err != nil {
            return fmt.Errorf("parsing existing headlamp_info: %w", err)
        }
        
        // Only update if originalName is empty
        if customObj.OriginalName == "" {
            customObj.OriginalName = contextName
            logger.Log(logger.LevelInfo, 
                map[string]string{"cluster": contextName},
                nil, 
                "updating legacy cluster with original name")
        }
    default:
        // No update needed
        return nil
    }

    // Update the extension
    contextConfig.Extensions["headlamp_info"] = customObj

    // Write changes to file
    if err := writeKubeConfig(config, path); err != nil {
        return fmt.Errorf("writing kubeconfig: %w", err)
    }

    return nil
}

// writeKubeConfig safely writes the config to file with proper error handling
func writeKubeConfig(config *api.Config, path string) error {
    if err := clientcmd.WriteToFile(*config, path); err != nil {
        logger.Log(logger.LevelError, nil, err, "writing kubeconfig file")
        return fmt.Errorf("writing to file %s: %w", path, err)
    }
    return nil
}
```

knrt10 (changes requested):
Some suggestions.

backend/cmd/headlamp.go
vyncent-t: haha! pay no attention to the human reader lines! 😵

backend/cmd/headlamp.go
vyncent-t: added some changes to make use of the enum, I think the inline I added for if the customName is there and the originalName is not should work rather than making another function just for that line

backend/cmd/headlamp.go

backend/cmd/headlamp.go
vyncent-t: added some changes to make use of the enum

backend/cmd/headlamp.go

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go

backend/cmd/headlamp.go

backend/cmd/headlamp.go

backend/cmd/headlamp.go
vyncent-t: done!

backend/cmd/headlamp.go
```diff
@@ -1108,6 +1108,9 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		// this function handles the clusters that need to have headlamp_info created or updated
+		c.handleHeadlampInfo(context)
```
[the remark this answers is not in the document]

vyncent-t:
@knrt10 made some changes from what you suggested and also added a test, I am using a mock env approach that seems to work when its ran at the end of the test, still new to tests so let me know what you think! I was also thinking of maybe using the format that you have for the TestRenameCluster if what I have here isn't the best way

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1130,45 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+// this function handles the clusters that need to have headlamp_info created or updated.
```
knrt10, line 1133: In Go you always start with the function name
```suggestion
// handleHeadlampInfo handles the clusters that need to have headlamp_info created or updated.
```

backend/cmd/headlamp.go
```diff
@@ -1108,6 +1108,13 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		// this function handles the clusters that need to have headlamp_info created or updated
+		err := c.handleHeadlampInfo(context)
+		if err != nil {
+			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
+				err, "failed to handle headlamp_info")
+		}
```
knrt10, line 1116: Since level of error is already `LevelError` we do not need to be rendundant with the log message. Can you please update it everywhere? Thanks
```suggestion
		if err := c.handleHeadlampInfo(context); err != nil {
			logger.Log(logger.LevelError, map[string]string{"cluster": context.Name},
				err, "handling headlamp_info")
		}
```

backend/cmd/headlamp.go
knrt10: Again, this is not removed. Can you please remove it?

backend/cmd/headlamp.go
```diff
@@ -1437,6 +1483,67 @@ func (c *HeadlampConfig) handleStatelessClusterRename(w http.ResponseWriter, r *
 	c.getConfig(w, r)
 }
 
+/*
+* this function writes or updates the original name to the Extensions map in the kubeconfig.
+ */
+//nolint:lll
```
knrt10, line 1489: Please make the function small, we avoid linter unless urgent

backend/cmd/headlamp.go
```diff
@@ -1437,6 +1483,67 @@ func (c *HeadlampConfig) handleStatelessClusterRename(w http.ResponseWriter, r *
 	c.getConfig(w, r)
 }
 
+/*
+* this function writes or updates the original name to the Extensions map in the kubeconfig.
```
knrt10, line 1487: Again comment should start with function name

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -56,9 +57,14 @@ type OidcConfig struct {
 
 // CustomObject represents the custom object that holds the HeadlampInfo regarding custom name.
 type CustomObject struct {
+	// ???
```
knrt10, line 60: ?? hehe, nice comment

backend/cmd/headlamp_test.go
```diff
@@ -1025,3 +1028,58 @@ func TestHandleClusterHelm(t *testing.T) {
 		})
 	}
 }
+
+// test for handleHeadlampInfo, currently checks for headlamp_info field and headlamp_info.originalName field.
+func (c *HeadlampConfig) TestHandleHeadlampInfo(t *testing.T) {
```
knrt10, line 1033: This function should call the existing function to check the functionality. This is the reason the test coverage does not increase. Can you please rewrite this? Thanks

backend/cmd/headlamp.go
```diff
@@ -1123,6 +1130,45 @@ func (c *HeadlampConfig) getClusters() []Cluster {
...
 }
 
+// this function handles the clusters that need to have headlamp_info created or updated.
+func (c *HeadlampConfig) handleHeadlampInfo(context *kubeconfig.Context) error {
+	// the headlamp_info is stored in the extensions of the kubeconfig context
+	headlampInfo := context.KubeContext.Extensions["headlamp_info"]
+
+	// if the headlamp_info is not present we create it and keep the original cluster name
+	if headlampInfo == nil {
+		err := c.keepOriginalClusterName(context.SourceStr(), context.Name, CreateNew)
```
knrt10, line 1140: same error can be modified as mentioned above. Can you please do that everywhere err is checked?

knrt10 (changes requested):
Thanks for the changes, added some suggestions

backend/pkg/kubeconfig/kubeconfig.go
vyncent-t: Oh! I forgot about these, I meant to ask if you could provide context here on what these params are/do?

backend/cmd/headlamp.go
vyncent-t: fixed the string!

backend/cmd/headlamp.go
vyncent-t: fixed the string!

backend/cmd/headlamp.go
vyncent-t: fixed the string!

backend/cmd/headlamp.go
vyncent-t: I run into linter issues when trying to use else if (too many nest) and the other alternative I can think of is adding more functions and just calling those, but this PR already adds 3 lengthy ones and alters an existing one.. should we stick with the ifs or create more functions here you think?

backend/cmd/headlamp.go
vyncent-t: fixed the string!

backend/cmd/headlamp.go
vyncent-t: done! the names are too long so I shortened it

backend/cmd/headlamp_test.go
vyncent-t: this test requires a bit of set up for it to work, mainly creating a temp mock env and then creating mock data within that env

I then convert the mock data from a api.Config structure into a kubeconfig.Context structure since that what handleHeadlampInfo test requires

I think this is necessary due to the placement/use of handleHeadlampInfo being within the getClusters function which I have mocked its functionality here (see the code leading up to handleHeadlampInfo at the top of getClusters)

I call the handleHeadlampInfo function, have it change the mock data, then check if the mock data has the headlamp_info fields and if the originalName is populated within headlamp_info

I will tag you where it's called

backend/cmd/headlamp_test.go
```diff
@@ -1025,3 +1028,58 @@ func TestHandleClusterHelm(t *testing.T) {
...
+			Cluster:     config.Clusters[context.Cluster],
+			AuthInfo:    config.AuthInfos[context.AuthInfo],
+		}
+		kubeContexts = append(kubeContexts, kubeContext)
+	}
+
+	// use the kubeContexts slice for further processing
+	for _, ctx := range kubeContexts {
+		// Call handleHeadlampInfo
+		err = c.handleHeadlampInfo(ctx)
```
vyncent-t, line 1067: @knrt10 I call the func here but for it to be ran I first have to create a mock env and then take the test data and turn it into kubeconfig.Context since I can't use config directly (its type is api.Config and the handleHeadlampInfo takes kubeconfig.Context)

In the two for loops under it I check if the headlamp_info field exists on all the contexts and I also check if all the originalName field exists within headlamp_info in each context with another loop

backend/cmd/headlamp.go
knrt10: if linter is saying too many nest, it does make sense to break the function, adding more function is ok since it solves complexity and is easier to understand for future modifications. Also it helps in adding unit tests for them as well.

backend/pkg/kubeconfig/kubeconfig.go
knrt10:
```suggestion
type CustomObject struct {
    // TypeMeta describes the type of the object and its API schema version.
    // It should have "Kind" set to "HeadlampInfo" and "APIVersion" set to "v1".
    // +k8s:deepcopy-gen=false
    metav1.TypeMeta

    // ObjectMeta contains metadata about the custom object, such as name and labels.
    // This is used to store additional metadata about the headlamp_info extension.
    // +optional
```

backend/cmd/headlamp_test.go
knrt10: The test function declaration is not correct. In Go, test functions should not be methods on a type. They should be standalone functions in the test file.


```suggestion
func TestHandleHeadlampInfo(t *testing.T) {
    // Create config instance for testing
    config := &HeadlampConfig{
        // initialize necessary fields for testing
    }

    // Rest of the test code using 'config' instead of 'c'
    ...
}

Comment (knrt10) [reply backend/cmd/headlamp_test.go:1033]: Currently it is lacking 3 main things

- Test Cases Organization: The test lacks clear separation of test cases. It should test different scenarios:

```go
tests := []struct {
        name           string
        configFile     string
        expectedError  bool
        validateFunc   func(*testing.T, *kubeconfig.Context)
    }{
        {
            name:       "new headlamp_info creation",
            configFile: "kubeconfig_headlampinfo_empty",
            validateFunc: func(t *testing.T, ctx *kubeconfig.Context) {
                info := ctx.KubeContext.Extensions["headlamp_info"]
                require.NotNil(t, info, "headlamp_info should be created")
                customObj, err := ParseHeadlampInfo(info)
                require.NoError(t, err)
                assert.Equal(t, ctx.Name, customObj.OriginalName)
            },
        },
        {
            name:       "update legacy headlamp_info",
            configFile: "kubeconfig_headlampinfo_legacy",
            validateFunc: func(t *testing.T, ctx *kubeconfig.Context) {
                info := ctx.KubeContext.Extensions["headlamp_info"]
                require.NotNil(t, info, "headlamp_info should exist")
                customObj, err := ParseHeadlampInfo(info)
                require.NoError(t, err)
                assert.NotEmpty(t, customObj.OriginalName, "originalName should be set")
                assert.NotEmpty(t, customObj.CustomName, "customName should be preserved")
            },
        },
    }
```

- Type Safety Improvements: Instead of using type assertion for `map[string]interface{}`, use the CustomObject type:
- Helper Functions: Extract common setup code. Just sample name are below

```go
func TestHandleHeadlampInfo(t *testing.T) {
    // IMPROVEMENT: Consider creating a newTestConfig() helper function instead
    // of inline configuration for better reusability across tests
    config := &HeadlampConfig{
        kubeConfigStore: kubeconfig.NewContextStore(),
    }

    // IMPROVEMENT: Add more fields to the test struct:
    // - description string: for detailed test documentation
    // - setupFunc: for test-specific setup if needed
    tests := []struct {
        name          string
        configFile    string
        expectedError bool
        validateFunc  func(*testing.T, *kubeconfig.Context)
    }{
        // MISSING TEST CASES:
        // 1. Empty headlamp_info
        // 2. Existing valid headlamp_info
        // 3. Invalid/malformed headlamp_info
        // 4. Missing required fields
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            // GOOD: Setup is separated into helper functions
            // SUGGESTION: Add cleanup in defer if needed
            mockConfigFile, config := setupTestKubeconfig(t, tt.configFile)
            contexts := contextFromConfig(config)

            // IMPROVEMENT: Consider combining Execute and Validate loops
            // to reduce iterations over contexts
            for _, ctx := range contexts {
                // MAIN TEST: This is where handleHeadlampInfo is called
                // It tests:
                // 1. Creation of new headlamp_info for empty extensions
                // 2. Update of existing headlamp_info for legacy contexts
                // 3. Error handling for invalid configurations
                err := config.handleHeadlampInfo(ctx)
                
                // ERROR CHECK: Verify expected error behavior
                // If expectedError is true, we should get an error
                // If false, we should not get an error
                if tt.expectedError {
                    require.Error(t, err)
                    return
                }
                require.NoError(t, err)

                // VALIDATION: Run the test-case specific validation
                // This checks the actual content and structure
                // of the headlamp_info extension
                tt.validateFunc(t, ctx)
            }
            
            // IMPROVEMENT: Add timeout context for long-running tests

            // PERSISTENCE CHECK: Verify changes were actually saved
            // SUGGESTION: Add more specific checks for the content
            // of headlamp_info, not just its presence
            persistedConfig, err := clientcmd.LoadFromFile(mockConfigFile)
            require.NoError(t, err)
            
            // FINAL VALIDATION: Ensure changes persisted to disk
            // Verify each context has the expected headlamp_info
            for name, ctx := range persistedConfig.Contexts {
                info := ctx.Extensions["headlamp_info"]
                require.NotNil(t, info, "headlamp_info should be persisted for context %s", name)
            }
        })
    }
}
```

knrt10 (changes requested):
Added some comments

backend/cmd/headlamp.go
vyncent-t: done! they're in smaller functions now

backend/cmd/headlamp_test.go
vyncent-t: reworked the test

backend/cmd/headlamp_test.go
vyncent-t: - Added the different scenarios 
- Test needed to be ran in such a way that I need to check changes in a file, so the test mocks a file, extracts from it, runs the function, then rechecks the file's data by extracting it again and checking (this is similar to how it behaves when being used in the project)

vyncent-t:
pushing rebase

vyncent-t:
> Think I noticed a bug, moving to draft until I'm sure

Never mind this works fine

I had forgot to close the app and start it up again after working on a different branch, running npm i and npm run start fresh fixed it, should be fine now

knrt10:
Hi, apologies for misssing it early but clusters added through app are all stored in indexDB and are dynamic clusters. This PR only adds the `headlamp_info` to the kubeconfig. It still does nothing to dynamic clusters. For that you need to update indexDB as well. You can use [updateStatelessClusterKubeconfig](https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/stateless/index.ts#L470) for do that.

I would request not to merge this PR without it as this is very important.

### Why this is a priority?

Every cluster that is added after cluster starts using `ADD CLUSTER` is stored in indexDB not in kubeconfig. kubeconfig clusters are only used if someone starts their cluster with that.

I just tested it and the indexDB kubeconfig still does not have the `headlamp_info` on cluster addition.

### Steps

- Open cluster and add a cluster using `ADD CLUSTER` button. You will see the cluster is stored in indexDB not backend kubeconfig. 
- Open indexDB and copy your endcoded value and decode it easily at [base64decode](https://www.base64decode.org/)
- You will see it still doesn't have `headlamp_info`

vyncent-t:
rebased to main
- dynamic clusters keep headlamp_info now

backend/cmd/headlamp.go
```diff
@@ -1539,6 +1671,46 @@ func (c *HeadlampConfig) getPathAndLoadKubeconfig(source, clusterName string) (s
...
+
+const (
+	// CreateNew creates a new headlamp_info extension if none exists.
+	CreateNew HeadlampInfoUpdateMode = iota
+	// UpdateLegacy updates existing headlamp_info for legacy clusters.
+	UpdateLegacy
+)
+
+/* handler to keep the original name for a cluster.*/
+//nolint:lll
```
joaquimrocha, line 1699: Can you avoid this?

backend/cmd/headlamp.go
```diff
@@ -1136,6 +1142,72 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 	return clusters
 }
 
+// handleHeadlampInfo this function handles the clusters that need to have headlamp_info created or updated.
```
joaquimrocha, line 1145: I don't understand exactly what this is doing from this comment. What does "handles the clusters" mean?

joaquimrocha: I am a bit confused as to what exactly this patch is trying to achieve and why.

backend/pkg/kubeconfig/kubeconfig.go
joaquimrocha: Still not updated here.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -56,9 +56,14 @@ type OidcConfig struct {
...
 	metav1.TypeMeta
+	// ???
```
joaquimrocha: And here.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -56,9 +57,13 @@ type OidcConfig struct {
...
 type CustomObject struct {
-	// ???
+	// TypeMeta describes the type of the object and its API schema version.
+	// It should have "Kind" set to "HeadlampInfo" and "APIVersion" set to "v1".
+	// +k8s:deepcopy-gen=false
 	metav1.TypeMeta
-	// ???
+	// ObjectMeta contains metadata about the custom object, such as name and labels.
+	// This is used to store additional metadata about the headlamp_info extension.
+	// +optional
```
joaquimrocha: Doesn't belong to this commit.

frontend/src/components/App/Home/index.tsx
```diff
@@ -238,6 +239,10 @@ function HomeComponent(props: HomeComponentProps) {
       .sort();
   }
 
+  async function handleDynamicClusterName(cluster: string) {
+    findKubeconfigByClusterName(cluster);
+  }
```
joaquimrocha, line 244: Why do we need to have a handleDynamicClusterName function wrapping the new one? (and in general "handle*" is not a great name unless it's for something that's just a response, e.g. `handleConfigRouteRequest`).

frontend/src/stateless/index.ts
```diff
@@ -252,7 +252,45 @@ export function findKubeconfigByClusterName(clusterName: string): Promise<string
             );
 
             if (matchingKubeconfig || matchingContext) {
-              resolve(kubeconfig);
+              const contextToUpdate = matchingKubeconfig || matchingContext;
+              const extensions = contextToUpdate?.context.extensions || [];
+              let headlampExtension = extensions.find(
+                extension => extension.name === 'headlamp_info'
```
joaquimrocha, line 258: Why do stateless clusters need the `headlamp_info` extension? This extension's purpose was to avoid modifying the cluster name of clusters coming from "external" (to headlamp) kubeconfigs, and we were even talking the other day about whether we should even do this at all for those. But for anything that "belongs" or is defined in Headlamp, like stateless clusters, we should just change the actual context name when the user chooses to update it.

joaquimrocha (changes requested):
I am not sure why we need these changes, especially if they are to accommodate stateless clusters, which shouldn't need the headlamp_info extension.

vyncent-t:
NOTE:

- rebased to current main
- WIP for the tests as they are still working off old test logic
- CURRENT BRANCH NOW ALLOWS FOR STORING THE ORIGINAL NAME AND SOURCE PATH IN MEMEORY (console log avail)

vyncent-t:
- rebased to main
- fixed tests to show in memory changes 
- removed console logs
- branch now stores headlamp_info in memory

illume:
@vyncent-t Closing this since you mentioned the new PR replaces it. Please reopen of course if I’m mistaken :)

PR #2751, opened 2025-01-16
WIP frontend: Update Map graph model 

sniok:
This is still in progress, still need to split into more commits and add explanation about what is happening here


joaquimrocha:
Nice changes! Small nit: do we need the checkbox on the Group by button to show it's selected? I thought the styling of the button would do that.
I was also a bit puzzled that the subtitle shows up on top of the title, but I guess later we can add style related props to choose where to place it.

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -146,17 +146,19 @@ export function GlobalSearchContent({
   // Resource search items
   const resources = useSearchResources();
   const loading = resources.filter(it => it.isLoading).map(it => it.kind);
+  const isMap = history.location.pathname.includes('/map');
```
joaquimrocha, line 149: I think this may yield false positives. Maybe you have to compare the current map route using the react-route-dom functions.

joaquimrocha (changes requested):
Left just a couple of comments. Otherwise it's looking great!

frontend/src/components/globalSearch/GlobalSearchContent.tsx
sniok: replaced with useRouteMatch and also got the path from the route definition

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -146,17 +146,22 @@ export function GlobalSearchContent({
...
-        history.push(
-          createRouteURL(item.kind, {
-            name: item.metadata.name,
-            namespace: item.metadata.namespace,
-          })
-        )
-      ),
-    [resources]
+      makeKubeObjectResults(resources, item => {
+        const search = new URLSearchParams(location.search);
```
joaquimrocha, line 154: Do you need to use the `location` as a dependency for this memo?

joaquimrocha (commented):
Just left one comment. Otherwise it looks fine.

frontend/src/components/globalSearch/GlobalSearchContent.tsx
sniok: yep, good catch. (we have to get those react eslint rules configured properly)

Issue #2693, opened 2024-12-15
Monaco DiffEditor does not apply coloring to changed text


mgalesloot:
## Describe the bug
Monaco DiffEditor does not apply coloring to changed text (regression in Headlamp v0.27) 

## To Reproduce
Steps to reproduce the bug:
1. Prepare dev environment 
2. Add the following lines to a Headlamp page.
```
import { DiffEditor } from '@monaco-editor/react';
 <DiffEditor language="yaml" original={'original'} modified={'fixedYAML'} height="10vh" />
```
3. Run the frontend (with make run-frontend)
4. Navigate to the changed page
5. Notice that the coloring for different text is not applied. The text is default (black and white). 
Expected was green and red coloring as in the example https://codesandbox.io/p/sandbox/get-values-diffeditor-c6xrg?file=%2Fsrc%2FApp.js

This issue is new in Headlamp v0.27. The previous version v0.26 was OK. Running the DiffEditor on the code of the v0.26 tag ( `git checkout headlamp-helm-0.26.0,  make run-frontend`) shows a proper working DiffEditor. 

This issue was first detected in the Kubescape plugin. 

## Environment 
Headlamp v0.27 

## Are you able to fix this issue?
No

joaquimrocha:
@skoeva  Maybe you can take a look

PR #2752, opened 2025-01-16
frontend: EditorDialog: Remove unneeded Monaco override

skoeva:
This change addresses a regression where a local MonacoEnvironment override caused conflicts with Monaco-based components like DiffEditor. The Monaco override logic in EditorDialog is now removed to restore the default environment.

Fixes: #2693 

### Testing
- [X] Perform the steps outlined in the issue
- [X] Ensure that the colors are applied as intended

image

### DIFF
--- a/frontend/src/components/common/Resource/EditorDialog.tsx
+++ b/frontend/src/components/common/Resource/EditorDialog.tsx
@@ -11,11 +11,6 @@ import Switch from '@mui/material/Switch';
 import Typography from '@mui/material/Typography';
 import * as yaml from 'js-yaml';
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
-import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
-import cssWorker from 'monaco-editor/esm/vs/language/css/css.worker?worker';
-import htmlWorker from 'monaco-editor/esm/vs/language/html/html.worker?worker';
-import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker';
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks

sniok:
The original issue only mentions DiffEditor from '@monaco-editor/react' package. How does showing Editor Dialog impacts rendering the DiffEditor? They seem to be separate

skoeva:
@sniok I went through the commit history and narrowed down the issue to that. I think it has to do with Monaco not loading properly if the editor isn't rendered, but feel free to test

sniok:
@skoeva The example codesandbox in the original issue renders a DiffEditor by itself without any additional logic. Which means that there's something in EditorDialog that breaks monaco editor for the rest of the app unless EditorDialog is being rendered. If that's true then if we just merge this without understanding underlying cause we might encounter some additional problems with this later or in other spots. I think it's important to understand what is causing the issue in the first place.

skoeva:
@sniok I suppose it's related to the MonacoEnvironment override, I don't see any specific reason to be redefining it here. I believe the default environment should be fine: I commented it out and it also fixed the issue. Let me know if I'm missing something though

illume:
> I suppose it's related to the MonacoEnvironment override, I don't see any specific reason to be redefining it here. I believe the default environment should be fine: I commented it out and it also fixed the issue. Let me know if I'm missing something though

Was it a workaround for old jest+vite? Seems the tests pass now without it.

skoeva:
@illume It was a remnant from that migration, yeah. If it isn't used anywhere and fixes this issue I think it can safely be removed

joaquimrocha (approved):

PR #2758, opened 2025-01-20
feat(helm): add extended service values

oliverbaehler:
Adds more exposure options for plain services, hope it's okay i did not go all out on the schema validation for the IPs

### DIFF
--- a/charts/headlamp/values.schema.json
+++ b/charts/headlamp/values.schema.json
@@ -306,7 +306,10 @@
           "description": "Kubernetes Service loadBalancerIP"
         },
         "loadBalancerSourceRanges": {
-          "type": "string",
+"type": "array",
+    "items": {
+      "type": "string"
+}
           "description": "Kubernetes Service loadBalancerSourceRanges"
         },
         "nodePort": {
--- a/charts/headlamp/values.schema.json
+++ b/charts/headlamp/values.schema.json
@@ -313,7 +313,7 @@
           "description": "Kubernetes Service loadBalancerSourceRanges"
         },
         "nodePort": {
-          "type": "integer",
+"type": ["integer", "null"],
           "description": "Kubernetes Service Nodeport"
         }
       }
--- a/charts/headlamp/values.schema.json
+++ b/charts/headlamp/values.schema.json
@@ -296,6 +296,22 @@
         "port": {
           "type": "integer",
           "description": "Kubernetes Service port"
+        },
+        "clusterIP": {
+          "type": "string",
+          "description": "Kubernetes Service clusterIP"
+        },
+        "loadBalancerIP": {
+          "type": "string",
+          "description": "Kubernetes Service loadBalancerIP"
+        },
+        "loadBalancerSourceRanges": {
+          "type": "string",
+          "description": "Kubernetes Service loadBalancerSourceRanges"
+        },
+        "nodePort": {
+          "type": "integer",
+          "description": "Kubernetes Service Nodeport"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


charts/headlamp/values.schema.json
```diff
@@ -296,6 +296,22 @@
...
+        "clusterIP": {
+          "type": "string",
+          "description": "Kubernetes Service clusterIP"
+        },
+        "loadBalancerIP": {
+          "type": "string",
+          "description": "Kubernetes Service loadBalancerIP"
+        },
+        "loadBalancerSourceRanges": {
+          "type": "string",
```
knrt10, line 309: This should be array.


```suggestion
"type": "array",
    "items": {
      "type": "string"
}
```

charts/headlamp/values.schema.json
```diff
@@ -296,6 +296,22 @@
...
+          "description": "Kubernetes Service loadBalancerSourceRanges"
+        },
+        "nodePort": {
+          "type": "integer",
```
knrt10, line 313: Sometimes nodePort would not be required, when we run `helm template` it fails. So it would be better to have null as well.

```suggestion
"type": ["integer", "null"],
```

charts/headlamp/values.yaml
```diff
@@ -134,6 +134,14 @@ service:
   type: ClusterIP
   # -- Kubernetes Service port
   port: 80
+  # -- Kubernetes Service clusterIP
+  clusterIP:
```
knrt10, line 138: Can we have empty string here?

charts/headlamp/values.yaml
```diff
@@ -134,6 +134,14 @@ service:
...
+  # -- Kubernetes Service loadBalancerIP
+  loadBalancerIP:
```
knrt10, line 140: same

charts/headlamp/values.yaml
```diff
@@ -134,6 +134,14 @@ service:
...
+  # -- Kubernetes Service loadBalancerSourceRanges
+  loadBalancerSourceRanges: []
+  # -- Kubernetes Service Nodeport
+  nodePort:
```
knrt10, line 144: please add null here

knrt10 (changes requested):
Thank you so much for the PR @oliverbaehler. Added some suggestions. Also, can you please add the docs for the same in https://github.com/headlamp-k8s/headlamp/tree/main/charts/headlamp#network-configuration? Thanks

oliverbaehler:
@knrt10 done

knrt10:
Awesome, everything is great. Thank you for addressing the change. Since this adds new values to value.yaml. It would be great if you could add the documentation for the same in https://github.com/headlamp-k8s/headlamp/tree/main/charts/headlamp#network-configuration

oliverbaehler:
@knrt10 Why are you separating these values so weirdly? If that would be grouped by prefix we could use a helm-docs template for the README:

eg.

```
### Network Configuration

| Key | Type | Default | Description |
|-----|------|---------|-------------|
{{- range .Values }}
  {{- if or (hasPrefix "ingress" .Key) (hasPrefix "service" .Key)  }}
| {{ .Key }} | {{ .Type }} | {{ if .Default }}{{ .Default }}{{ else }}{{ .AutoDefault }}{{ end }} | {{ if .Description }}{{ .Description }}{{ else }}{{ .AutoDescription }}{{ end }} |
  {{- end }}
{{- end }}
```

But thats complicated when you group pvc values and mounts eg.

knrt10 (approved):
Thank you @oliverbaehler, your suggestions makes a lot of sense for several good reasons. I will improve our documentation in a seperate PR. Thanks for the changes, this PR LGTM.

PR #2762, opened 2025-01-20
docs: Add Flatpak to platforms with comment about external tools

### LINKED ISSUE #1885: external auth (exec) not working (azure/kubelogin and aws-iam-authenticator)

joaquimrocha:
The external tools running from Flatpak have been reported to be working, but there's a prior command that needs to be run by the user. So this PR adds a mention of that to the docs:

fixes #1885

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -47,6 +47,7 @@ We test on MacOS, various flavours of Linux, and Windows. Headlamp runs in the b
 | MacOS (arm, x86)                   |  ✔️   |
 | Ubuntu 20.04, 22.04, 22.10         |  ✔️   |
 | Fedora                             |  ✔️   |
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks

gaeljw:
Shouldn't this be documented in https://headlamp.dev/docs/latest/installation/desktop/linux-installation as well?

joaquimrocha:
Definitely. Thanks for pointing it out. I will add it.

joaquimrocha:
> Shouldn't this be documented in https://headlamp.dev/docs/latest/installation/desktop/linux-installation as well?

Reads much better: https://github.com/headlamp-k8s/headlamp/pull/2768 . Thanks.

PR #2766, opened 2025-01-21
frontend: Quick fix for a case when no endpoints are available for the resource

sniok:
This is the smallest change that can fix the https://github.com/headlamp-k8s/headlamp/issues/2738 

There are other areas where error handling is not the best or not working at all and will be handled in a separate PR

Opened an issue for it here https://github.com/headlamp-k8s/headlamp/issues/2767


sniok:
This PR also fixes mocks in gateway stories, because it would depend on mocks from different stories and some mocks were defined twice

joaquimrocha (approved):

PR #2771, opened 2025-01-22
frontend: Improve error handling in list pages

sniok:
This PR combines and replaces different fields we have for error handling, currently we have 'error' 'errorMessage' 'clusterErrors' props in various places.

`error` is not enough since we can have more than one error
`errorMessage` is a string and loses important context like which cluster the error is coming from and limits how we can render it
`clusterErrors` prop has a limitation of one error per cluster but we can have more (see Role page example)

New field is
`errors` an array of `ApiError`

ApiError is also updated to be a class instead of an interface and now contains all relevant information (if available) like cluster name, namespace of a resource and status code.

ApiError message is now also parsed from backend response (if available)

To display the error the `ClusterGroupErrorMessage` component was updated since it is already can display error.
I would also change the name to something more generic like `ErrorAlerts` or something but it's not a priority.

---

### Testing done / Examples

All error messages by default are collapsed, in screenshots I'll expand them

When resource is not available, endpoint returns 404

image

image

When user doesn't have access 403

image

image

When user doesn't have access and List page loads multiple resource
Role page loads Roles and ClusterRoles

image

image

When two clusters are selected and one of them doesn't have access

image

When two clusters are selected and one of them doesn't have access for a List page with multiple resources

image

Cluster overview page with no access
image


sniok:
Updated the error message to use Alert component from mui, and made 403 error display as blue since it's probably expected that user with limited access won't be able to see some things

frontend/src/components/horizontalPodAutoscaler/__snapshots__/HPADetails.Error.stories.storyshot
```diff
@@ -79,7 +79,7 @@
               <p
                 class="MuiTypography-root MuiTypography-body1 MuiTypography-alignCenter css-9ldkza-MuiTypography-root"
               >
-                Error: Unreachable
+                TypeError: Failed to fetch
```
joaquimrocha, line 82: Is this expected?

frontend/src/components/cluster/ClusterGroupErrorMessage.tsx
```diff
@@ -1,63 +1,66 @@
...
   const { t } = useTranslation();
-  const theme = useTheme();
-  const clusterObj = cleanNullErrors(typeof clusterErrors === 'object' ? clusterErrors : {});
+  const showClusterName = useClusterGroup().length > 1;
+  const [showMessage, setShowMessage] = useState(false);
 
-  if ((!clusterErrors && !message) || Object.keys(clusterObj).length === 0) {
-    return null;
+  const defaultTitle = t('Failed to load resources');
+  const forbiddenTitle = t("You don't have permission to view this resource");
```
joaquimrocha, line 28: (Should be "permission**s**")
If we used the resource kind here, would it make it more clear in the Roles and ClusterRoles case?
`You don't have permissions to view {{ resource.kind }} resources`
I know in this case we have to make sure we distinguish between a list view or a details view, but maybe that's a good thing (besides solving the weird case of repeated error messages because its referring to ClusterRoles and Roles)?

joaquimrocha (commented):
Pretty cool. I have left a couple of comments.

frontend/src/components/horizontalPodAutoscaler/__snapshots__/HPADetails.Error.stories.storyshot
sniok: Yes, "Unreachable" message was hard coded, now it has an actual original error message

frontend/src/components/cluster/ClusterGroupErrorMessage.tsx
sniok: I thought about it a bit and this message is vague on purpose, there's more information about the error in the details. Which shows actually which permissions are missing, what role, etc. 

I think this component is generic enough to use for all kinds of places, not just for listing resources

joaquimrocha (approved):

PR #2773, opened 2025-01-23
frontend: Enhance Table component with column resizing functionality

### LINKED ISSUE #2652: Allow users to change column widths
Add ability to increase the width of the columns yourself because sometimes "name" column contains two rows and it is difficult to read.

Allow users to change column widths

It's easier to read.

No

## Additional context
Change the deployment name to something longer (no more than 63 characters)
Sample application https://github.com/spring-projects/spring-petclinic/tree/main/k8s
image

zcyc:
fix #2652 
Dear maintainers, this is my first time contributing code to this project, please forgive me if there are any problems. Thank you for your work.


sniok (changes requested):
Hi, thanks for opening a PR! 

When I tried it, it behaves a bit weird, there's something wrong with the resizing

https://github.com/user-attachments/assets/dd03a3cd-5b36-4f2f-9483-56e3d0b28452

I'm guessing it's because of the custom table layout we're using so it might be a lot harder to get it right

zcyc:
Hi, I have tried several different solutions. The reason why the columns next to each other change with the adjustment is that the width of the table is fixed. If we want to ensure that the adjustment does not affect other columns, i cannot fix the width of the table. Adjusting the column width will cause the page to stretch. We can discuss which solution is better.

https://github.com/user-attachments/assets/762a414c-1d2f-44e4-827e-780de2f8c2d1

zcyc:
I pushed the code for the second solution to pr, but I can't fix the width of the page. Maybe all the places where it is used need to be changed, which will be a big change. I'm not sure whether to continue.

sniok:
Thanks for taking a look into this and trying to find a solution. Column resizing appears to be quite tricky to implement without a major refactor of our current table layout.

illume:
Closing because I think this was addressed in:
- #2802

Thanks again for investigating this.

PR #2786, opened 2025-01-24
frontend: Fix a circular dependency in KubeObject

sniok:
This fixes hot module reloading during development and reenables App.test.tsx

This is a fix for a problem I've been trying to fix for a long time. You might have noticed that during development if you're making any changes the whole page reloads. This is because Hot Module Reload is failing due to a circular dependency. Not every circular dependency causes this kind of failure.

This PR fixes circular dependency: KubeObject -> Router -> All Routes -> The whole app. By reimplementing getDetailsLink and getListLink methods without requiring router. Since routes follow the convention to use apiName, we can take a shortcut and avoid importing and assume the routes follow this convention. The only resources that weren't following this were Storage resources which were updated.

This unfortunately doesn't fix the storybook issue, so that workaround is still there


frontend/src/lib/router.tsx
```diff
@@ -182,41 +182,41 @@ const defaultRoutes: {
     component: () => <NodeDetails />,
   },
   storageClasses: {
-    path: '/storage/classes',
+    path: '/storageclasses',
```
joaquimrocha, line 185: What exactly are these fixing? I think changing a long standing URL may represent a breaking change.

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -126,12 +125,16 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
       namespace: this.getNamespace(),
       name: this.getName(),
     };
-    const link = createRouteURL(this.detailsRoute, params);
-    return link;
+    const cluster = this.cluster ?? getCluster();
+    if (this.isNamespaced) {
+      return `/c/${cluster}/${this._class().apiName}/${params.namespace}/${params.name}`;
+    }
+    return `/c/${cluster}/${this._class().apiName}/${params.name}`;
```
joaquimrocha, line 132: Is replacing a function that produces a full URL by a repetition of the different URL parts a good idea? I'd say these changes make things harder to maintain and more error prone.

frontend/src/lib/router.tsx
sniok: it's inconsistent with the rest of the paths, every other route is using /apiName/:namespace/:name

frontend/src/lib/k8s/KubeObject.ts
sniok: yeah in the current form it will make it a bit worse. 
but if we take this convention a step further and replace all the routes with:
```js
{
details: {
   path: '/:apiName/:?namespace/:name',
   component: KubeObjectDetails,
},
list: {
  path: '/:apiName/',
  component: KubeObjectList
}
}
```

I think it will make it less error prone in the long term

sniok:
Maybe revisit this later. Hot reload is nice to have but not critical.

PR #2794, opened 2025-01-27
frontend: Fallback to default namespace found in the kubeconfig when all namespaces fail to load

### LINKED ISSUE #1522: Resort to showing the default namespace when namespaces fail to be listed
A couple of users have reported now what they think is an auth issue in Headlamp, since it reports as having "no permissions" when listing resources while other solutions like kubectl work fine.

In a recent [Slack conversation](https://kubernetes.slack.com/archives/C01FXB5E8ER/p1698873279757819), it's become (almost) clear that this has to do with the fact that Headlamp tries to list all namespaces while those other tools use the kubeconfig's declared default namespace. In that conversation, the reporter has shared a [terraform config](https://gitlab.com/flashpixx/kubernetes-access-test) to initialize a cluster and reproduce this.

IMO, a way to fix this would be to still attempt to list resources from all namespaces, and in case of failure resort to showing the default namespace (if specified). We could also that default namespace by default immediately, but I do like to have the global view that all namespaces give us as a default view.

**Acceptance Criteria:**
- [x] Pass the default namespace together with the cluster info (from kubeconfig)
- [x] Resort to selecting the default namespace if listing all namespaces fails
- [x] **Update: I think this one is covered with the allowed namespaces feature:** Remember the namespaces selected when loading Headlamp (i.e. store this setting in the session storage)

sniok:
Fixes #1522

Adds a hooks that will to try fallback to a default namespace (if any set in the kubeconfig) when we failed to load namespaces

Testing done:

configured default namespace in the kubeconfig and tried using service account that has access only to pods in default namespace

### DIFF
--- a/frontend/src/components/common/NamespacesAutocomplete.tsx
+++ b/frontend/src/components/common/NamespacesAutocomplete.tsx
@@ -5,12 +5,12 @@ import Checkbox from '@mui/material/Checkbox';
 import { useTheme } from '@mui/material/styles';
 import TextField from '@mui/material/TextField';
 import Typography from '@mui/material/Typography';
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useHistory, useLocation } from 'react-router-dom';
 import helpers, { addQuery } from '../../helpers';
-import { useCluster } from '../../lib/k8s';
+import { useCluster, useClustersConf } from '../../lib/k8s';
 import Namespace from '../../lib/k8s/namespace';
 import { setNamespaceFilter } from '../../redux/filterSlice';
 import { useTypedSelector } from '../../redux/reducers/reducers';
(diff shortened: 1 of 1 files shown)


frontend/src/App.tsx
```diff
@@ -36,6 +36,7 @@ const queryClient = new QueryClient({
     queries: {
       staleTime: 3 * 60_000,
       refetchOnWindowFocus: false,
+      retry: false,
```
joaquimrocha, line 39: Will this affect the rest of the application? AFAIU, if something fails, it will not be attempted to be fetched again. Won't this turn off the "live" UI when there's an error?
Maybe this is actually desirable in cases where things fail, but we need to have a way for the users to retry without having to actually refresh the application.

frontend/src/components/common/NamespacesAutocomplete.tsx
```diff
@@ -152,10 +152,47 @@ export function NamespacesAutocomplete() {
   );
 }
 
+/**
+ * This hook will try to select a namespace in a specific case
+ *
+ * If we failed to load namespaces it might be because the user
+ * doesn't have acces to list all the namespaces but still has an
```
joaquimrocha, line 159: typo: access

and "still has ~an~"

joaquimrocha (approved):
Found a typo in the comment.

frontend/src/App.tsx
sniok: by default react-query will retry 3 times and then gives up and returns an error. 
I disabled retries here so that the fallback is executed faster (without having to wait for 3 retries)
But this might not be a good idea, I'll remove this commit

PR #2797, opened 2025-01-28
Update platforms.md with K0s

wrkode:
I've thoroughly tested with k0s and I'm happy to report it working well and with no workarounds.


illume (approved):
👍

illume:
Thanks!

PR #2808, opened 2025-01-29
frontend: plugins: Remove unused @types/nock pkg

skoeva:
This package isn't necessary to have installed as per the message in #1992.


vyncent-t (commented):
test look good 😎

sniok (approved):
nice!

Issue #266, opened 2021-06-12
EKS ? yes


bciceron:
works like a charm on 2 clusters i tried.
using Version 0.3.0 (0.3.0) :)
much faster than Lens

rbudiharso:
How to authenticate with eks? it ask for token, while Lens can do without asking for a token.

joebowbeer:
The inability to easily use kubeconfig w/EKS AWS_PROFILE is a showstopper for us (unlike Octant, Lens, OpenLens, Aptakube, etc.)

goutham-sabapathy:
So whats the solution for this ?? how do I connect to the clusters which are listed in my kube config file ?

pmcdonald-metropolis:
bump

mo-saeed:
is there anybody who was able to access EKS cluster using the token generated from [aws eks get-token](https://docs.aws.amazon.com/cli/latest/reference/eks/get-token.html)

joebowbeer:
When the aws CLI adds a cluster to my kubeconfig, it adds an eks get-token command exec to the config.

Headlamp reads the kubeconfig and lists the cluster.

However, when I try to connect then headlamp asks me for a token. This is odd because the command to fetch a token is right there in the kubeconfig. Furthermore, since the eks get-token expires relatively quickly, there's no use in me entering a token manually. (Furthermore, Lens and Octant work as I expect.)

joaquimrocha:
FTR, in the case of @joebowbeer this has to do with running when kubeconfig is set to use the exec plugin for getting the token from `aws`. We do need to make sure Headlamp is playing well with this.

joaquimrocha:
Hi @joaocc , what version and flavor of Headlamp are you using? (Mac/Linux/Windows)

joaocc:
Desktop-wise I am having issues with macOS and latest headlam.
We are also having some issues on k8s deployment (via helm) with keycloak SSO and multicluster (but it may be keycloak specific, not sure yet)

joaocc:
Hi. Any updates on this? Thanks

gabrielpetry:
It worked for me in aws eks using the AppImage version, the flatpak one didn't worked, probably it's due to how flatpak containers work ;(

aviau:
Same issue here, its asking me for the token instead of running the command itself. I am on macOS

joebowbeer:
I suggest closing this in favor of one of the newer issues. 

This was fixed for me in #1582

[...]

PR #2814, opened 2025-01-30
backend: oidc: impersonate instead of forwarding oidc token to k8s api

iSE-stefan-may:
By default, the OIDC JWT is forwarded to the Kubernetes API server. In some cases it might not be possible to have the Kubernetes API server configured with OIDC, in these cases the OIDC JWT should be evaluated by Headlamp and the request should use the service account of the pod together with impersonating the user in the token.


zivcex (approved):
lg2m

backend/cmd/headlamp.go
```diff
@@ -883,10 +885,76 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
...
+	}
+	return oauth2Token, nil
+}
+
 func StartHeadlampServer(config *HeadlampConfig) {
 	handler := createHeadlampHandler(config)
 
 	handler = config.OIDCTokenRefreshMiddleware(handler)
+	if config.oidcImpersonate {
+		handler = config.OIDCImpersonateMiddleware(handler)
```
illume, line 955: Can you please explain why this middleware is after the OIDCTokenRefreshMiddleware?

backend/pkg/config/config.go
```diff
@@ -166,6 +168,10 @@ func flagset() *flag.FlagSet {
 	f.String("oidc-idp-issuer-url", "", "Identity provider issuer URL for OIDC")
 	f.String("oidc-scopes", "profile,email",
 		"A comma separated list of scopes needed from the OIDC provider")
+	f.Bool("oidc-impersonate", false,
```
illume, line 171: @knrt10 would this need to be added to the helm chart as well?

illume (changes requested):
Hello. Thanks for this!

I left a few questions/notes.

Can you please add some steps in the PR description on how we can test this?

Would it be possible for you to add some test cases?

illume:
Some lint issues in CI job: https://github.com/headlamp-k8s/headlamp/actions/runs/13056969725/job/36711540140?pr=2814#step:7:27

backend/cmd/headlamp.go
```diff
@@ -883,10 +885,76 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
 	})
 }
 
+func (c *HeadlampConfig) OIDCImpersonateMiddleware(next http.Handler) http.Handler {
```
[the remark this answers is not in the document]
iSE-stefan-may: Added some documentation

backend/cmd/headlamp.go
```diff
@@ -883,10 +885,76 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
...
+
+		// parse cluster and token
+		cluster, token := parseClusterAndToken(r)
+		if cluster == "" || token == "" {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		idToken, err := c.getValidToken(r.Context(), token)
+		if err != nil {
```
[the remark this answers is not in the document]
iSE-stefan-may: Logging added, but I'm not really sure, whether it is really the best case to log always when the token validation fails. It may also be, that OIDC impersonation is active, but the user still used a serviceaccount token, in those cases a warning in the log would be irritating.

backend/cmd/headlamp.go
iSE-stefan-may: It was the most fitting position I thought of, but not a particular reason. Do you have better recommendations?

backend/pkg/config/config.go
iSE-stefan-may: Added to the helm chart, however I'm not convinced that I did it in a good way, since the chart is a little bit confusing for me. So please recheck.

iSE-stefan-may:
Have updated some code regarding the comments, but not fully tested yet and not added a test description. Will work on that tomorrow, but perhaps you can already review the answers to the comments.

audunsolemdal:
Hi, not an expert here. I am wondering if this implementation will support impersonation based on groups listed in the `groups` claim found in the token returned from the OIDC provider?

illume:
> Have updated some code regarding the comments, but not fully tested yet and not added a test description. Will work on that tomorrow, but perhaps you can already review the answers to the comments.

Thanks. Reading through it now :)

charts/headlamp/templates/deployment.yaml
```diff
@@ -147,11 +177,21 @@ spec:
             # Check if scopes are non empty either from env or oidc.config
             - "-oidc-scopes=$(OIDC_SCOPES)"
             {{- end }}
+            {{- if or ($oidc.impersonate) ($impersonate) }}
```
zivcex, line 180: Hey great PR!! 

I'm wondering would it be too heavy if you can provide the groups as well here within the chart and deployment? 

more details are here:
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation:~:text=Impersonate%2DGroup%3A%20A%20group%20name%20to%20act%20as.%20Can%20be%20provided%20multiple%20times%20to%20set%20multiple%20groups.%20Optional.%20Requires%20%22Impersonate%2DUser%22.


Convention here is that groups claim in JWT hold the user group coming in from IDP. It would be good if it was configurable as some IDPs send the groups membership under other claim name.

illume:
It looks like the backend/lint CI job is failing.
You can run `make backend-lint` locally.

illume:
The helm template job is having an issue: https://github.com/headlamp-k8s/headlamp/actions/runs/13197046279/job/36885634312?pr=2814

```
Template test failed for ./charts/headlamp/tests/test_cases/oidc-create-secret.yaml against ./charts/headlamp/tests/expected_templates/oidc-create-secret.yaml:
--- ./charts/headlamp/tests/oidc-create-secret.yaml_output/rendered_templates.yaml	2025-02-08 02:48:20.528327104 +0000
+++ ./charts/headlamp/tests/expected_templates/oidc-create-secret.yaml	2025-02-08 02:48:20.536327[13](https://github.com/headlamp-k8s/headlamp/actions/runs/13197046279/job/36885634312?pr=2814#step:5:14)6 +0000
@@ -126,11 +126,6 @@
                 secretKeyRef:
                   name: oidc
                   key: scopes
-            - name: OIDC_IMPERSONATE_CLAIM
-              valueFrom:
-                secretKeyRef:
-                  name: oidc
-                  key: impersonateClaim
make: *** [Makefile:[18](https://github.com/headlamp-k8s/headlamp/actions/runs/13197046279/job/36885634312?pr=2814#step:5:19)5: helm-template-test] Error 1
           args:
             - "-in-cluster"
Error: Process completed with exit code 2.
```

backend/cmd/headlamp.go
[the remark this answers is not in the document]
illume: Great! Thanks.

backend/cmd/headlamp.go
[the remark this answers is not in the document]
illume: Ok. That sounds like it makes sense to remove.

Is something like this possible? Otherwise remove that logging?
```
if ! OIDCImpersonationEnaled and serviceAccount {
    log()
}
```

backend/cmd/headlamp.go
illume: I was wondering how it would interact with the other middleware? Wondering if it would make sense to go before OIDCTokenRefreshMiddleware or not?

Because would the token refresh need to use the impersonated stuff? If that's the case I was wondering if the OIDCImpersonateMiddleware should go first?

backend/pkg/config/config.go
illume: Our main helm chart expert is on holidays at the moment. So this will take a bit longer to review. Bear with me :)

backend/cmd/headlamp.go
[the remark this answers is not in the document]
iSE-stefan-may: Wether the user provided token is a kubernetes serviceaccount token can only be evaluated from the Kubernetes API, therefore this check won't be possible. Only option where this would be possible would be, if we disable the option to login in the ui via a serviceaccount token altogether. In this case we could give a log warning here, but as far as I can see, that's currently not possible.
I'll cleanup the logging for the cases, where it may not be an error case.

backend/cmd/headlamp.go
iSE-stefan-may: You are right, the OIDCTokenRefreshMiddleware needs the original token, therefore it must be evaluated before OIDCImpersonateMiddleware. I'll change the order of those two and add a comment.

mlbiam:
> By default, the OIDC JWT is forwarded to the Kubernetes API server. In some cases it might not be possible to have the Kubernetes API server configured with OIDC, in these cases the OIDC JWT should be evaluated by Headlamp and the request should use the service account of the pod together with impersonating the user in the token.

This is not a good idea.  Dashboards should never be configured with privileged token.  it's too easy to expose a dashbaord by accident.  Even when impersonating the user, headlamp should rely on the token from the proxy, not its own.  This way someone accidentally exposes the dashbaord outside of the proxy, they haven't created a vector to access the API server with a privleged account.

illume:
>This is not a good idea. Dashboards should never be configured with privileged token. it's too easy to expose a dashbaord by accident. Even when impersonating the user, headlamp should rely on the token from the proxy, not its own. This way someone accidentally exposes the dashbaord outside of the proxy, they haven't created a vector to access the API server with a privleged account.

It is a good idea to try and help people prevent making mistakes with configuration accidentally opening things up. 

Also, when hearing about other peoples use cases it is often the situation that they are best placed to evaluate and decide why they want to do things in a particular way.

@iSE-stefan-may maybe you could give us some ideas about this? Could you elaborate on your use case a bit? Also, what do you think about this possibility of opening things up accidentally and how we could address this?

joaquimrocha:
> > By default, the OIDC JWT is forwarded to the Kubernetes API server. In some cases it might not be possible to have the Kubernetes API server configured with OIDC, in these cases the OIDC JWT should be evaluated by Headlamp and the request should use the service account of the pod together with impersonating the user in the token.
> 
> This is not a good idea. Dashboards should never be configured with privileged token. it's too easy to expose a dashbaord by accident. Even when impersonating the user, headlamp should rely on the token from the proxy, not its own. This way someone accidentally exposes the dashbaord outside of the proxy, they haven't created a vector to access the API server with a privleged account.

So far we have avoided requiring or encouraging running Headlamp with its own token/role. We did have some users asking for this and we resorted to advising to use a reverse proxy. We even have thought of maybe having a side project to facilitate this for Headlamp; would this bring any advantage here?

@mlbiam I'd like to ask your opinion whether this is still a security concern if we add the feature as optional and somehow make it clear, via CLI options' names, extra options, and documentation what the implications of this are. What would that look like?

zivcex:
Hi @iSE-stefan-may will you close this one given the new circumstances? Or leave this setting optional?

backend/cmd/headlamp.go
```diff
@@ -896,9 +900,112 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
...
+		// Since we are doing requests with the serviceaccount, make sure there are no
+		// other Impersonate-* headers added
+		for key, _ := range r.Header {
+			if strings.HasPrefix(strings.ToLower(key), "impersonate-") {
+				r.Header.Set(key, "")
+			}
+		}
+		r.Header.Set("Impersonate-User", user)
+		for _, group := range groups {
+			r.Header.Set("Impersonate-Group", group)
```
aslafy-z, line 980: I believe this should use Add instead of Set.

```suggestion
			r.Header.Add("Impersonate-Group", group)
```

mlbiam:
To @engnatha 's point about using RBAC to limit impersonation - There are two threats:

1.  if someone gets a hold of the container's token.  Either via an RCE against the headlamp container or sideways from inside the cluster (as in the Tesla breach on the Kubernetes Dashboard).  Forcing headlamp users to use a purpose built proxy with the privileged account cuts down on the attach surface (if the frontend only exposes APIs, there's less to attack) and making it easier to upgrade, keep up to date, test against, etc.
2. RBAC impersonation is very limited for impersonation.  You can limit which groups are impersonated, and you can limit which users are impersonated, but you can't limit which users can impersonate which groups.  These leads to an over permissioned token that if stolen, allows an attacker incredible flexibility.

Using the dashbaord's service account as privileged was pretty big security issue with the kubernetes dashboard and they removed that capability correctly (hell, i wrote an entire chapter of the Kube: an enterprise guide around how to attack the dashboard running a privileged account until they removed the capability in 7.x)

While I think the lessons learned from the kubernetes dashboard project advise against making the headlamp service account privileged, if you really want to go this route it shouldn't be the default.  The headlamp account shouldn't be used as part of impersonation, but if you want it to be there should be a switch that has to be turned on.

@joaquimrocha including a reverse proxy in the helm charts to do this would probably be not as straight forward as you'd hope.  This biggest issues we run into helping customers and users run OpenUnison is certificate management and ingress integration.  Assuming you're encrypting e2e (and given these tokens are privileged, i certainly would hope that would be the requirement) you're talking about either requiring cert-manager or building your own cert-manager like system.  Then there's testing with each ingress controller/gateway.  You're going to get much better results ivesting efforts into documenting how to deploy headlamp with impersonation then trying to engineer the issue.

Also, who "owns" that proxy you're including?  Who's responsible for continuous testing?  Not just happy-path testing, but failure mode testing to protect against mallicious use?  If that component fails, they're going to blame the headlamp project the same way that the kubernetes dashboard still has the assumption of being "insecure" post tesla breach.

engnatha:
I think impersonation works better for the weaveworks UI because the operations exposed are so much more limited. You're really only interested in viewing resources and patching existing resources of a certain kind. While you could do some harm with that, I think the blast radius minimized. In the headlamp system, there would be a much richer ecosystem of permissions to give more access to the cluster for admins of different components. 

I agree that it will never not be a threat. I'm mostly just trying to advocate as a potential user that the current system is forcing users away from the product. With no guides on how to set up the system on managed clusters (which is becoming the norm rather than the exception) and no willingness to provide alternate mechanisms like proxies, impersonation, or otherwise, we aren't left with many options. I've tried hard to search through existing issues and documentation on how to do this with modern systems and it just isn't there. Even looking at the [platforms page](https://headlamp.dev/docs/latest/platforms), the big cloud provider mentioned for support is AWS and the rationale is [someone in 2021 saying it worked with no mention of how](https://github.com/kubernetes-sigs/headlamp/issues/266). I do really respect the push for security here and if this PR does get blocked, I hope it provides motivation to teach users how to utilize this system properly. The alternative is everyone reinventing the wheel and likely introducing their own vulnerabilities with proxies and impersonations of their own or more widely sharing long lived service account tokens.

engnatha:
Looking into this more, I actually think the current authentication strategy imposed by headlamp is impossible to accommodate in modern GKE. Based on the guidance in https://cloud.google.com/kubernetes-engine/docs/how-to/oidc#external-idp-authentication-methods, we see the following
> In GKE Standard clusters only, GKE also supports Identity Service for GKE. Identity Service for GKE is limited to OIDC IdPs and installs additional components in your cluster. GKE strongly recommends using Workforce Identity Federation instead of Identity Service for GKE.

They are migrating everyone to workforce identity federation which would require headlamp to route through that system to get the proper short-lived google service account tokens. I definitely don't know as much about this system, so I'd really like to be wrong about this if anyone has more experience :grin:

joaquimrocha:
@yolossn , can you double check this in light of your recent updates?

oliverbaehler:
I mean these options only come in handy if there's a proxy anyway in front of the kubernetes API. I am not a security export but if the proxy handles all the impersonation attributes via  [SubjectAccessReview](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#request-attributes-used-in-authorization). This tanks performance somewhat but is resonable safe. I would also argue that headlamp itself should  only just send the Impersonation Headers, the rest is the admins responsability. I think Teleport works the same way, it just sends the authenticated user data via impersonation to the kubernetes API. Obviously the point is valid, that the given serviceAccount has high privileges. But that's the only option really.

PR #2815, opened 2025-01-30
docs: document running the app while backend, frontend are running externally

### LINKED ISSUE #1794: Document running the app while the backend/frontend are running externally
The instructions for running Headlamp in development mode indicate that we do a [couple of make commands](https://headlamp.dev/docs/latest/development/#run-the-code). This will run the backend + frontend, allowing developers to test Headlamp in the browser. However, this doesn't test the app module, and if developers want to run the app, then it will trigger the backend again.

We do have an npm command for running only the app module which will use the already running backend/frontend: `npm run dev-only-app`

This issue is about adding the [documentation](https://github.com/headlamp-k8s/headlamp/blob/main/docs/development/_index.md) for running the app in development mode without triggering the backend (the command above).

## Notes

- See the file docs/development/_index.md at https://github.com/headlamp-k8s/headlamp/blob/main/docs/development/_index.md#build-the-app for where to add the documentation
- The command is to run the app when headlamp is being run in development mode with `make run-backend` and `make run-frontend`. The benefit of it is you can develop the frontend version of headlamp at the same time as the app version (without having to restart any commands).

adwait-godbole:
Fixes #1794

### DIFF
--- a/docs/development/index.md
+++ b/docs/development/index.md
@@ -95,6 +95,23 @@ make app-win-msi
 
 See the generated app files in app/dist/ .
 
+### Running the app
+
+If you already have **BOTH** the `backend` and `frontend` up and running, the quickest way to 
+get the `app` running for development is the following:
+
+```bash
+make run-only-app
+```
+
+or else you can simply do
+
+```bash
+make run-app
+```
+
+which runs everything including the `backend`, `frontend` and `app` in parallel.
+
 ### Running the app on Ubuntu WSL
 
 Headlamp on WSL requires some packages installed (maybe it requires more) to run the app.
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,12 @@ else
 	cd frontend && npm start
 endif
 
+run-app:
+	cd app && npm install && node ./scripts/setup-plugins.js && npm run start
+
+run-only-app:
+	cd app && npm install && node ./scripts/setup-plugins.js && npm run dev-only-app
+
 frontend-lint:
 	cd frontend && npm run lint -- --max-warnings 0 && npm run format-check
 


docs/development/index.md
```diff
@@ -101,6 +101,23 @@ make app-win-msi
...
+make run-only-app
+```
+
+or else you can simply do
+
+```bash
+make run-app
+```
+
+which runs everything including the `backend`, `frontend` and `app` in parallel.
```
joaquimrocha, line 119: Let's add something like: Running the app this way assumes you want to run a pre-built backend and frontend, so make sure you run `make` before `make run-app`.

joaquimrocha (changes requested):

docs/development/index.md
adwait-godbole: Hi @joaquimrocha thanks for the quick feedback! 

Over [here](https://github.com/headlamp-k8s/headlamp/blob/f58c530accfdbb222d35c3e036a0f5e12e841de4/Makefile#L113), I made it so that if the user calls `make run-app` (given he has no backend & frontend already up and running), then we would eventually only call `npm run start` as shown in the Makefile changes of this PR. And `npm run start` over [here](https://github.com/headlamp-k8s/headlamp/blob/f979bf5e11680455aaddf86d51d34623b2a29ee7/app/package.json#L23) already triggers the [start.js](https://github.com/headlamp-k8s/headlamp/blob/main/app/scripts/start.js) script which starts everything in parallel. 

So should we really be asking the user to do `make` before running `make run-app`  given that he/she can simply do `make run-app` and everything starts in parallel?

adwait-godbole:
Hi @joaquimrocha. Do you have any other feedback over here? I have left a comment to address your last review. Please let me know if anything else is pending over here. Thanks.

docs/development/index.md
illume: Sounds good to me @adwait-godbole

illume:
@adwait-godbole the frontend/test CI job failure is unrelated to this PR. There's a fix here:
- https://github.com/headlamp-k8s/headlamp/pull/2862

adwait-godbole:
> @adwait-godbole the frontend/test CI job failure is unrelated to this PR. There's a fix here:

> https://github.com/headlamp-k8s/headlamp/pull/2862

Ohh. I see. Thanks for letting me know. I will wait for it to get resolved.

illume:
The frontend/test is fixed in main branch now.

adwait-godbole:
> The frontend/test is fixed in main branch now.

Hi @illume. I have rebased onto main. Thanks.

illume (approved):
🎉🎈 Thanks!

PR #2817, opened 2025-01-30
Show plugin package URL for user confirmation

joaquimrocha:
When installing a plugin, we are showing the plugin's ArtifactHub URL to the user for confirmation and this is not very interesting since all plugins are listed from ArtifactHub, in fact, what the user should confirm is the origin of the plugin package, because it's from that package that we will deploy the plugin.

## How to test:
- [ ] Go to the plugin catalog and select a plugin
- [ ] Click the install button and verify that the message shows the plugin's package URL instead of the ArtifactHub one
- [ ] Click No -> Verify the process is successfully aborted (there's an existing bug which shows the "reload now" button in the notification even when there's an error or cancellation)
- [ ] Click Yes -> The plugin is successfully installed

### DIFF
--- a/app/electron/locales/de/app.json
+++ b/app/electron/locales/de/app.json
@@ -1,4 +1,11 @@
 {
+  "Failed to fetch plugin info": "",
+  "An error occurred while fetching plugin info from {{  URL }}.": "",
+  "Yes": "",
+  "No": "",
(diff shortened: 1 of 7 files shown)


yolossn (approved):
Tested, works. Looks good to me.

Issue #931, opened 2023-02-02
Stories for components


illume:
Each of the components inside Headlamp need stories made for the different states.

- for easier manual UI testing, where the components are separate from the backend and each other.
- for automated regression testing of the presentational parts of the components

## Components that need stories

There's a number of components that still don't have any stories written for them.

Please write in the issue if you want to start on one, and then tick it off here when it's merged in.

- [ ] common/ReleaseNotes
- [ ] components/App/Notifications
- [ ] components/App/Layout
- [ ] components/App/AppContainer
- [ ] components/cluster/Charts
- [ ] components/cluster/Chooser
- [ ] components/cluster/ClusterChooser
- [ ] components/cluster/Overview

## Components that need stories improving

Some stories test only one possible state. These stories need improving.

- [ ] components/common/LogViewer

## Where should the stories go?

A good example to look at is `src/components/common/ActionButton`.

The preferred layout is the folder layout like this:

```
Component/Component.tsx, main place where the component goes. Named after the component.
Component/index.tsx, This should export things from Component.tsx
Component/Component.stories.tsx, contains the component shown in different states.
Component/__snapshots__ , where the story shot files are.
```

## Presentational VS container (logic) components

See [documentation here](link-to-story-js). In short, the logic should be separated so mainly the component can be tested without dependencies on other things. An example of logic that could be separated is API calls.

Some refactoring is probably needed to implement stories:

- to place components in a folder, with filenames the same name as the component
- to separate the presentational "pure" parts

## viewing the stories, and running the tests

To view the storybook:

```bash
cd frontend
npm install
npm run storybook
```

To run the tests, and generate the snapshot files (rendered html):

```bash
cd frontend
npm install # not needed if you've done it already.
npm run test -- --coverage
```

You can also run the tests without the interactive test runner like this:
```bash
CI=true npm run test -- --coverage
```

EshaanAgg:
Hi! I'd like to work on the stories for `components/cluster/ClusterChooser` if that is available for contribution!

IITI-tushar:
@illume 
sir i guess this issue is still open...
I would like to work on the components/App/Layout

1012Charan:
Hi! I’d like to work on this issue. I’m applying for the LFX mentorship and have been exploring the project — happy to give this a try.

1012Charan:
I’d like to work on components/App/Notifications

SriramKa:
[...]

PR #2821, opened 2025-01-31
Frontend: Add stories for chart componenets

Faakhir30:
References #931 

Brief overview:
- `Charts.tsx` originally had 4 chart components, from which `CpuCircularChart` and `MemoryCirclularChart` components had similar props, while `PodStatusCircularChart` and `NodeStatusCircularChart` had similar props.
- So, refactored chart.tsx into `ResourceCharts` and `StatusCharts`
- Created `.Stories.tsx` for both.
- Created index file and updated imports (if any).

@illume PR is currently ready for review, needed your opinion weather we should have done this separation of files or not? 
IMO it was too messy to deal this in single story file so refatrored a bit. Thanks !

Here's storybook video of current state:
[Screencast from 2025-01-31 23-57-28.webm](https://github.com/user-attachments/assets/b1e2f548-17e0-4ea9-87fc-bff2c9a47998)

### DIFF
--- a/frontend/src/components/cluster/__snapshots__/Overview.Events.stories.storyshot
+++ b/frontend/src/components/cluster/__snapshots__/Overview.Events.stories.storyshot
@@ -91,7 +91,7 @@
                                 <desc />
                                 <defs>
                                   <clippath
-                                    id="recharts1-clip"
+                                    id="recharts-id"
                                   >
                                     <rect
                                       height="102"
@@ -191,7 +191,7 @@
                                 <desc />
                                 <defs>
                                   <clippath
-                                    id="recharts3-clip"
+                                    id="recharts-id"
                                   >
                                     <rect
                                       height="102"
@@ -293,7 +293,7 @@
                                 <desc />
                                 <defs>
                                   <clippath
-                                    id="recharts5-clip"
+                                    id="recharts-id"
                                   >
                                     <rect
                                       height="102"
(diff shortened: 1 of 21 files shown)


frontend/src/components/cluster/__snapshots__/Overview.Events.stories.storyshot
```diff
@@ -91,7 +91,7 @@
                                 <desc />
                                 <defs>
                                   <clippath
-                                    id="recharts1-clip"
+                                    id="recharts17-clip"
```
Faakhir30, line 94: Not sure though, why is this changing ID on running `npm run test -- --coverage`

sniok: it looks like recharts library just increments id for every chart it renders, so the more charts we have in stories it will increase (and change) the id.

we have a `replaceUseId` function in storybook.test.ts that handles other non-deterministic ids. could you try to add this case there too? 
something like `if the attribute is 'id' and includes 'recharts' then replace with just 'recharts-id'` 

let me know if that makes sense

Faakhir30: Yup, makes sense, thanks for the hint :)

Added this `recharts` check in `storybook.test.tsx`.

sniok (approved):
looks good, thanks!

PR #2822, opened 2025-01-31
frontend: Add tests for `@apidevtools/swagger-parser`

### LINKED ISSUE #2440: Add test for @apidevtools/swagger-parser
We have a dependency @apidevtools/swagger-parser for which we have no tests that cover it.

This makes updating it time consuming and risky, as we need to manually test it with every update.

See headlamp/frontend/src/lib/docs.ts

adwait-godbole:
Fixes #2440 

Screenshot from 2025-02-03 17-38-13

### DIFF
--- a/frontend/src/lib/swagger-parser.k8s.test.ts
+++ b/frontend/src/lib/swagger-parser.k8s.test.ts
@@ -1,0 +1,132 @@
+import Swagger from '@apidevtools/swagger-parser';
+import { OpenAPIV2 } from 'openapi-types';
+
+describe('@apidevtools/swagger-parser', () => {
+  it('should resolve $ref references', async () => {
+    const doc: OpenAPIV2.Document = {
+      swagger: '2.0',
+      info: { title: 'Test API', version: '1.0.0' },
+      paths: {},
+      definitions: {
+        'io.k8s.api.apps.v1.Deployment': {
+          type: 'object',
+          properties: {
+            metadata: { $ref: '#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta' },
+          },
+        },
+        'io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta': {
+          type: 'object',
+          properties: {
+            name: { type: 'string' },
+            namespace: { type: 'string' },
+          },
+        },
+      },
+    };
+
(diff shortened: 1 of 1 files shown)


frontend/src/lib/docs.test.ts
```diff
@@ -0,0 +1,130 @@
+import Swagger from '@apidevtools/swagger-parser';
+import { Mock, vi } from 'vitest';
+import { getDocDefinitions, getDocs } from './docs';
+import { request } from './k8s/apiProxy';
+
+vi.mock('@apidevtools/swagger-parser', () => ({
+  default: { dereference: vi.fn() },
```
sniok, line 7: The original issue talks about testing the library, but these tests mock the library, which is the opposite

frontend/src/lib/docs.ts
```diff
@@ -6,19 +6,14 @@ import Swagger from '@apidevtools/swagger-parser';
...
 }
 
-export default async function getDocDefinitions(apiVersion: string, kind: string) {
-  if (!docsPromise) {
-    docsPromise = getDocs(); // Don't wait here. Just kick off the request
-  }
-
-  const { definitions = {} } = (await docsPromise) as OpenAPIV2.Document;
+export async function getDocDefinitions(apiVersion: string, kind: string) {
+  // We want a fresh call to getDocs() each time to ensure we get the latest mock data during testing
```
sniok, line 15: this changes the behavior of the 'getDocDefinitions' function, have you tested that the app feature still works the same?
I think this was done to avoid making duplicate requests

sniok (changes requested):
Thanks for opening a PR! I think these tests are not testing the thing that was mentioned in the original issue. Could you update the tests so they check "@apidevtools/swagger-parser" output?

adwait-godbole:
Hi @sniok, Thanks for the review! I apologize for misunderstanding the issue. 

I have now updated the tests to **actually** test the output of `@apidevtools/swagger-parser` library. 

We are primarily using this library for dereferencing Kubernetes OpenAPI Spec Documents and now I am testing this behavior under various dereferencing conditions. This is the screenshot. Let me know what you think. Thanks.

Screenshot from 2025-02-03 17-38-13

sniok (approved):
nice, thanks!

illume (approved):
🎉🎈

PR #2823, opened 2025-01-31
frontend: Add clusterChooser componnent stories

Faakhir30:
References #931

Added `clusterChooser` components stories.

Here's storybook video of current state:
[Screencast from 2025-02-01 03-55-16.webm](https://github.com/user-attachments/assets/2cb22695-0d97-4878-83e3-f110ed5408c8)

### DIFF
--- a/frontend/src/components/cluster/ClusterChooser.stories.tsx
+++ b/frontend/src/components/cluster/ClusterChooser.stories.tsx
@@ -1,0 +1,53 @@
+import { Meta, StoryFn } from '@storybook/react';
+import ClusterChooser from './ClusterChooser';
+
+export default {
+  title: 'cluster/ClusterChooser',
+  component: ClusterChooser,
+  parameters: {
+    docs: {
+      description: {
+        component:
+          'A button component for selecting clusters. Shows the current cluster name and handles click events.',
+      },
+    },
+  },
+  argTypes: {
+    cluster: {
+      control: 'text',
+      description: 'The name of the current cluster',
+    },
+    clickHandler: {
+      action: 'clicked',
+      description: 'Callback fired when the button is clicked',
+    },
+  },
+} as Meta;
+
+const Template: StoryFn = args => {
+  const defaultProps = {
+    clickHandler: () => {},
+    ...args,
+  };
+  return <ClusterChooser {...defaultProps} />;
+};
+
+export const Normal = Template.bind({});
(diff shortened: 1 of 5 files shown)


frontend/src/components/cluster/ClusterChooser.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export const Normal = Template.bind({});
+Normal.args = {
+  cluster: 'my-cluster',
+};
+
+export const LongClusterName = Template.bind({});
+LongClusterName.args = {
+  cluster: 'very-long-kubernetes-cluster-name-that-should-be-truncated',
```
sniok, line 43: ClusterChooser doesn't really truncate or transform the name so I don't think these cases are needed. I'd keep only the normal case

frontend/src/components/cluster/ClusterChooser.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import { Meta, StoryFn } from '@storybook/react';
+import ClusterChooser from './ClusterChooser';
+
+export default {
+  title: 'cluster/ClusterChooser',
+  component: ClusterChooser,
+  decorators: [Story => <Story />],
```
sniok, line 7: this decorator doesn't add anything so it can be removed

sniok (changes requested):
Hi, thanks for opening a PR for this! This looks good, just needs a couple of small changes

frontend/src/components/cluster/ClusterChooser.stories.tsx
Faakhir30: @sniok It does I guess, could you please confirm this once again, would update acordingly:
https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/cluster/ClusterChooser.tsx#L15

```
const SpanClusterName = styled('span')({
  textOverflow: 'ellipsis',
  whiteSpace: 'nowrap',
  overflow: 'hidden',
  display: 'block',
});
```

sniok: ah okay, well just from the html output in the snapshot it's hard to tell

sniok (approved):
nice, thanks!

illume (approved):
🎉 🎈

PR #2824, opened 2025-02-01
frontend: add Chooser componnent stories

Faakhir30:
References #931

Added clusterChooser components stories.

Heres snapshots video of current state:
[Screencast from 2025-02-01 04-10-54.webm](https://github.com/user-attachments/assets/8c02c7de-1652-4d93-b775-35854776124e)


frontend/src/components/cluster/__snapshots__/Chooser.LoadingClusters.stories.storyshot
```diff
@@ -0,0 +1,249 @@
...
+                      class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation1 MuiCard-root css-t50pmu-MuiPaper-root-MuiCard-root"
+                    >
+                      <div
+                        class="MuiCardContent-root css-d64700-MuiCardContent-root"
+                      >
+                        <p
+                          class="MuiTypography-root MuiTypography-body1 MuiTypography-gutterBottom css-18vfeyy-MuiTypography-root"
+                          title="cluster1"
+                        >
+                          cluster1
```
sniok, line 175: this story says that the clusters are loading, but I can see some clusters here

sniok (changes requested):
nice changes, thanks! could you take a look into LoadingClusters story and why there are clusters in the output?

frontend/src/components/cluster/__snapshots__/Chooser.LoadingClusters.stories.storyshot
Faakhir30: Thanks for pointing out, handling `null` was missing added to `useClustersConf`, now working fine:
image

frontend/src/lib/k8s/index.ts
```diff
@@ -92,8 +92,12 @@ export const ResourceClasses = {
  * */
 export function useClustersConf(): ConfigState['allClusters'] {
   const state = useTypedSelector(state => state.config);
-  const clusters = _.cloneDeep(state.clusters || {});
+  const clusters = _.cloneDeep(state.clusters);
   const allClusters = _.cloneDeep(state.allClusters || {});
+
+  if (clusters === null) {
```
sniok, line 98: this changes the behavior of this function and breaks logic for statelessClusters branch. could you revert the change to this file please

sniok (changes requested):

frontend/src/lib/k8s/index.ts
Faakhir30: Yup, i was expecting something might break due to this
@sniok  I did this because this function `useClustersConf` would always returns `object || {}` and never `null`, therefore 
` loading state` would never be achieved IMO, refering these 2 lines:
https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/cluster/Chooser.tsx#L417


Therefore, Removing loading state story for now

illume (approved):
🎉🎈

PR #2827, opened 2025-02-02
frontend: Add restart bulk action for resources list

### LINKED ISSUE #2156: Add bulk actions for resources lists
Now that we have a more powerful table, we should add bulk actions for user-selected resources, i.e. users can for example choose more than one pod and delete them in one command.

Actions:
- [ ] [Delete pods](https://github.com/headlamp-k8s/headlamp/issues/1640)
- [ ] Restart deployments

adwait-godbole:
Fixes #2156  (ability to delete multiple pods was already done in #1640)

Deployment Demo:

https://github.com/user-attachments/assets/8a23c88f-acb0-4641-898a-f5ae5d335f63

StatefulSet Demo:

https://github.com/user-attachments/assets/622c003f-b0d7-47b4-80e3-06bfa2e7a132

DaemonSet Demo:

https://github.com/user-attachments/assets/bf2cc3a4-5e9b-4622-95f6-39df65bf3442


adwait-godbole:
> It looks like the i18n check failed. Run npm run i18n to check.

Hi @illume ran this on my local machine and attached the corresponding generated translation.json files as well. But still the build seems to be failing. Running `make frontend-test` on my local is not giving any failures.

adwait-godbole:
> It looks like the i18n check failed. Run npm run i18n to check.

Fixed it! Thanks for the review.

adwait-godbole:
Hi @illume. I have fixed the CI. Please let me know if any other issue is pending to be addressed. Thanks for you review. Looking forward to hear back from you.

adwait-godbole:
> There's a slight regression in the Restart popup, it used to say "Cancel" or "Restart" in the actions. Now it just says "Yes" or "No". In popups like this it's better to be as explicit as possible and "cancel" "restart" clearly state what is going to happen, unlike yes/no.

> In console there are some validation errors

Hi @sniok. Thanks for the review. I have addressed the above points in the latest commit. 

> I've noticed that bulk restart is not implemented in Pods, it would be nice to have it there too

Regarding this, currently I have also noticed that Headlamp doesn't have a "Restart" button available for Pods in the Actions drop down menu. I mean we are also not supporting single restarts as of now.

so,  when you say "bulk restarting Pods", the Pods could either be standalone or managed by high-level controllers like `Deployment`, `StatefulSet`, etc. When these pods are running as standalone, then the "restart" functionality is pretty straightforward where in we first Delete the pod and then Re-Create/Re-Apply it once again (both done programmatically). So no doubt over there, but if they are running as managed by these high-level controllers, then what should exactly happen when clicking on "Restart"? Even if the user simply clicks on "Delete" in this case, then that would be sufficient right? since the controllers would then handle rest of the restarting part? So do we need to show a "Restart" option in this case?

So, what I had in mind regarding this request of yours is that first we should enable a single "Restart" option for pods that are running in standalone mode in the Actions drop down. So, is there currently any way of knowing this in the frontend code if the pod is running as standalone or not? And once we are able to introduce this functionality of single restart for a pod running as standalone, then we can further extend it to a bulk action (but again, the bulk action will only apply on standalone pods, cause it makes much sense in that case).

Lmk what are your thoughts regarding this. Once again, thanks for the review.

illume:
> I've noticed that bulk restart is not implemented in Pods, it would be nice to have it there too

It sounds like this extra stuff could be done outside of this PR? If so this discussion could be moved to an issue.

I guess there should be some sort of test for this functionality.

sniok:
Sorry, for some reason I thought we had a 'Restart' action for pods, disregard my comment about it.

adwait-godbole:
> I guess there should be some sort of test for this functionality.

Hi @illume, do you mean e2e-tests for this functionality?

illume:
> do you mean e2e-tests for this functionality?

Yeah, I think ultimately an e2e test for this would be good. But probably we shouldn't block this being merged if it doesn't have one.

adwait-godbole:
> Yeah, I think ultimately an e2e test for this would be good. But probably we shouldn't block this being merged if it doesn't have one.

Whatever you feel is right.

adwait-godbole:
I will open a separate PR for the e2e-test then.

illume:
I selected two deployments, and clicked the restart button.

However it gives me a failed message.
image


Seems to do a POST on the first one, and a PUT on the second one. The first fails.
image

illume:
It works for me with restarting two daemonsets.

```
kubectl apply -f - <<EOF
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: daemonset-one
spec:
  selector:
    matchLabels:
      app: daemonset-one
  template:
    metadata:
      labels:
        app: daemonset-one
    spec:
      containers:
      - name: container-one
        image: nginx:latest
        ports:
        - containerPort: 80
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: daemonset-two
spec:
  selector:
    matchLabels:
      app: daemonset-two
  template:
    metadata:
      labels:
        app: daemonset-two
    spec:
      containers:
      - name: container-two
        image: httpd:latest
        ports:
        - containerPort: 80
EOF
```

image

adwait-godbole:
@illume. That's weird. I don't seem to be getting such an error on my local machine. Here's a demo.

https://github.com/user-attachments/assets/c3e541db-461f-425f-be52-b3f9843747ce

illume:
Hrmm. Maybe it’s a timing issue, or something related to Minikube?

adwait-godbole:
> Hrmm. Maybe it’s a timing issue, or something related to Minikube?

I guess then adding e2e-test here itself would not be a bad idea then? @illume ?

sniok:
One way to avoid Conflict issues is to use `patch` with just the parts that need to updated instead of `apply`. We had a similar issue with Suspend/Resume of CronJobs, here's a PR with a fix for that https://github.com/headlamp-k8s/headlamp/pull/2745/files

I'm not 100% sure if this applies here though

adwait-godbole:
> One way to avoid Conflict issues is to use patch with just the parts that need to be updated instead of apply. We had a similar issue with Suspend/Resume of CronJobs, here's a PR with a fix for that https://github.com/headlamp-k8s/headlamp/pull/2745/files

Hi @sniok, Thanks a lot for that guidance! Using `patch` instead of `apply` worked like a charm!

@illume here's a demo of each Resource using the updated `patch`  logic. Lmk if it works for you. Thanks.

Deployment:

https://github.com/user-attachments/assets/a445fd1c-93cd-438a-a328-7de3ef547f22

StatefulSet:

https://github.com/user-attachments/assets/22586030-fbb6-42d9-853d-32a149550e77

DaemonSet:

https://github.com/user-attachments/assets/935a3666-24ed-440d-9262-292e6518e7e6

illume (approved):
Works well. Thanks!

🎉🎈

illume:
I'll wait for @sniok to review again and merge.

PR #2828, opened 2025-02-02
frontend: OauthPopup: fixed deprecated type

DhairyaMajmudar:
## Description

Fixed deprecated type `ReactChild` to `ReactNode`

## Screenshots of tests passing 

All tests are passing correctly 

Screenshot from 2025-02-02 14-08-10

### DIFF
--- a/frontend/src/components/oidcauth/OauthPopup.tsx
+++ b/frontend/src/components/oidcauth/OauthPopup.tsx
@@ -1,5 +1,5 @@
 import Button from '@mui/material/Button';
-import React, { ReactChild } from 'react';
+import React, { ReactNode } from 'react';
 
 interface OauthPopupProps {
   width?: number;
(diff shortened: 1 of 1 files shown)


illume:
Great, thank you! 

A small nitpick about the commit message. We use more like the linux kernel way of doing commit messages. So it would star with "Fix" instead of "fixed".

`frontend: OauthPopup: Fix deprecated type`

DhairyaMajmudar:
Thank you for the review.

Sure thingy I'll make sure to follow Linux based commit styles.

skoeva (changes requested):
Looks good, but it would be nice if you could use an atomic commit. Thanks for this!

DhairyaMajmudar:
> Looks good, but it would be nice if you could use an atomic commit. Thanks for this!

Thank you for appreciation, rebased my PR with an atomic commit

skoeva (approved):

illume (approved):
👍 appreciated.

PR #2829, opened 2025-02-02
Frontend: Add Storybook Stories for Notifications Component

Faakhir30:
Refers #931
Added Storybook stories to showcase and test different states of the Notifications component. 

- Refactored `Notifications` components to seperate file and updated imports from `index.ts`
- Added test utility function `createTestNotification` to generate sample notifications
- Set up Redux store configuration for testing with `configureStore`
- Included `MemoryRouter` wrapper for proper routing context
- Added decorators to provide necessary context providers
- Added Storybook stories to test different states of the `Notifications` component.

- [x] Verified tests passing
image


illume:
Thanks so much for this!

We have this convention for git commits with the context at the beginning. 

Something like: `frontend: Notifications: Add Storybook Stories`.

I guess you could also merge these two commits into one.

frontend/src/components/App/Notifications/index.tsx
```diff
@@ -1,350 +1,4 @@
...
-          }}
-          style={{ textTransform: 'none' }}
-        >
-          {t('translation|View all notifications')}
-        </Button>
-      </Popover>
-    </>
-  );
-}
+export { Notifications, NotificationList };
```
illume, line 4: We need to keep backwards compatibility here and export the NotificationList as default.

illume (changes requested):
Looks good to me.

There's just that default export to fix for backwards compatibility. 
(Sorry I didn't spot it the first time I reviewed)

frontend/src/components/App/Notifications/index.tsx
Faakhir30: `default export NotificationList;` form this index.tsx wont provide backward compatibility IMO, i'd have to refactor list to its own barrel export folder, it would be `frontend/src/components/App/Notifications/List/index.tsx` to ensure backward compatibility, 
Since `NotificationList` was being exported from `frontend/src/components/App/Notifications/List.tsx` originally.

illume (approved):
🎈 🎉

Thank you. Much appreciated.

PR #2830, opened 2025-02-02
frontend: OauthPopup: Added stories

DhairyaMajmudar:
## Description

Adding OauthPopup component stories and snapshots

### Passing tests 

Screenshot from 2025-02-02 20-03-29

### Story Preview

image

## Screencasts

[Screencast from 2025-02-02 20-07-00.webm](https://github.com/user-attachments/assets/824e8116-918b-422b-b447-d3c95ac7b0cb)


skoeva (changes requested):
could you condense these changes into an atomic commit? looks solid otherwise!

skoeva (approved):

illume (approved):
👍 excellent. Thanks!

PR #2832, opened 2025-02-02
Plugins: added Resource CircularChart example plugin

### LINKED ISSUE #586: [RFE] Add example plugin showing how to use the Resource CircularChart
## Current situation

We give access to the Resource's CircularChart in plugins but we have no good examples/docs showing how to use this component.

## Ideal situation

We have an example plugin showing some data with the circular chart. A quick idea can be to display the failed/non-failed pods chart.

Faakhir30:
Closes #586 **[READY FOR REVIEW]**

Implements a Resource CircularChart example plugin showing pod failure stats

This PR adds a new example plugin that demonstrates how to use Headlamp's CircularChart components to visualize cluster resource data. The implementation process involved:

1. Created a new resource-charts plugin showing failed vs total pods
2. Initially used `CircularChart` component, then simplified to use `TileChart` directly:
   - Removed unnecessary metrics-related functionality
   - Added theme-based error coloring for better visualization
   - Improved chart labeling and percentage calculations

3. Added documentation in README.md explaining:
   - Plugin purpose and features
   - Setup instructions
   - Implementation details and key concepts

The example plugin is:
- Using `K8s.ResourceClasses` for data fetching
- Implementing chart `processors` for `Overview ` page as an example page.
- Using `redux` slice for plugin charts, then processing on frontend code.

#### Pod Failed Chart on Overview Page:
image

Plugin Details:
image

#### Screencast of failing live Pod:
[Screencast from 2025-02-03 02-45-13.webm](https://github.com/user-attachments/assets/0b35d444-f1e0-4f71-a9a5-5815c803ab17)

#### Tests Passing (YES :) 
image


illume:
Nice one.

I noticed there's an issue when building docs on CI: https://github.com/headlamp-k8s/headlamp/actions/runs/13102964037/job/36701030284?pr=2832#step:4:28

Probably you can `make docs` to reproduce it locally.

illume:
btw, we don't use the `feat()` style commit messages.

Faakhir30:
Thanks @illume ,
Reproduced and fixed the failing check.

Rebased with upstream `main` and cleared commit history to single commit with specified style of commits in headlamp.

plugins/examples/resource-charts/src/index.tsx
```diff
@@ -0,0 +1,52 @@
+import { K8s, registerOverviewChartsProcessor } from '@kinvolk/headlamp-plugin/lib';
+import { TileChart } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import { Box, Paper } from '@mui/material';
+import { useTheme } from '@mui/material/styles';
+import React from 'react';
+
```
illume, line 6: I think a short comment here explaining what this plugin is showing would be good. Something like: 

> This shows how to add a custom overview chart on the overview page.

plugins/examples/resource-charts/src/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+function PodFailureChart() {
```
illume, line 7: A short component tsdoc string would help here.

> PodFailureChart shows the percentage of failing pods.

illume (approved):
Thanks! :tada:

I tested it, and it works nicely.

ps. I left a couple of optional suggestions for small comments in the index.tsx file which might be a small improvement.

illume:
image

illume:
@joaquimrocha I requested a review from you since it was you who wrote the original issue.
- https://github.com/headlamp-k8s/headlamp/issues/586

Faakhir30:
> ps. I left a couple of optional suggestions for small comments in the index.tsx file which might be a small improvement.

Thanks for the review, added comments/tsdoc string in last separate single commit for easy review (if needed).

illume:
> Thanks for the review, added comments/tsdoc string in last separate single commit for easy review (if needed).

Awesome, thanks. Those look good to me.

A small nit: In the linux style git commit messages there is no "." at the end of the title line.

yolossn (approved):
Great work @Faakhir30, thanks for contributing.

joaquimrocha (dismissed):
Nice feature @Faakhir30 ! The only thing I think is needed is that the second commit should be actually squashed with the 1st and then the actual functionality of adding a way to process the charts should be split from the example plugin. Because right now the commit says it's about the example plugin but it's actually also adding the functionality.
So I think the 2 final commits should be named something like:
`frontend: Add a way for plugins to process the cluster overview's charts`
`frontend plugins/examples: Add a resource charts example plugin`

Faakhir30:
> So I think the 2 final commits should be named something like: `frontend: Add a way for plugins to process the cluster overview's charts` `frontend plugins/examples: Add a resource charts example plugin`

Updated commits to be atomic as described, thanks.
@illume , finally, PR is ready to merge now ig.

PR #2839, opened 2025-02-03
frontend: added tests for lib/units.ts and lib/cluster.ts files

DhairyaMajmudar:
## Description 

Added unit tests for below files:

1. `lib/units.ts`
2. `lib/cluster.ts`

### DIFF
--- a/frontend/src/lib/cluster.ts
+++ b/frontend/src/lib/cluster.ts
@@ -20,7 +20,7 @@ export function getClusterPrefixedPath(path?: string | null) {
 export function getCluster(): string | null {
   const prefix = helpers.getBaseUrl();
   const urlPath = helpers.isElectron()
-    ? window.location.hash.substr(1)
+    ? window.location.hash.substring(1)
     : window.location.pathname.slice(prefix.length);
 
   const clusterURLMatch = matchPath<{ cluster?: string }>(urlPath, {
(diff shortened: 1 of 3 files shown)


frontend/src/lib/cluster.ts
```diff
@@ -20,7 +20,7 @@ export function getClusterPrefixedPath(path?: string | null) {
 export function getCluster(): string | null {
   const prefix = helpers.getBaseUrl();
   const urlPath = helpers.isElectron()
-    ? window.location.hash.substr(1)
+    ? window.location.hash.substring(1)
```
DhairyaMajmudar, line 23: `substr` is deprecated that's why replaced it with `substring`

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

illume (changes requested):
Thank you for this. Very good!

A small note on the git messages... We use the linux kernel style commit messages, where it starts with a capitalized imperative verb. For example:
`frontend: lib/cluster.ts: Add test for getCluster` instead of "added".

DhairyaMajmudar:
> Thank you for this. Very good!
> 
> A small note on the git messages... We use the linux kernel style commit messages, where it starts with a capitalized imperative verb. For example: `frontend: lib/cluster.ts: Add test for getCluster` instead of "added".

Thank you for appreciation, will take care of git commit message : )

joaquimrocha (approved):
Thanks @DhairyaMajmudar

PR #2776, opened 2025-01-23
Revert "Merge pull request #2615 from farodin91/refactor-settings-cluster"

sniok:
This reverts commit 39d8f7295b5517e35cb75e44b870d7c99de31529, reversing changes made to f5dc9fdfd8865cf73f0785b5e72bee2b8e038248.

- #2615

This fixes cluster settings

### How to test

Type something in default namespace, make sure page isn't replaced by a spinner

Select default namespace, switch to a different cluster, make sure the default namespace values updates accordingly

PR #2841, opened 2025-02-03
frontend: refactor settings cluster

farodin91:
@sniok I tried to fix the issue from #2776.


sniok:
hi @farodin91, I've tried this branch and it still has the same issue.

try setting 'allowed namespace' and then switching between different clusters
it overrides settings for every cluster


https://github.com/user-attachments/assets/a68a3e21-494f-4c98-a362-f349930a7838

sniok (changes requested):
I appreciate the effort and thanks for opening the PR but in the current state there's something broken after refactor

Issue #2846, opened 2025-02-04
Backend token creates new instances with unique URL


skoeva:
## Describe the bug
Every time we run the app (`npm run dev`), the backend token is stored in the URL, which means that each run creates a new instance with a unique URL.

## To Reproduce
Steps to reproduce the bug:
1. Run `npm run dev`
2. Open the Preferences file (i.e. `~/.config/Headlamp && cat Preferences`
3. Notice all the different urls (you might have more or fewer URLs in your list than this):

![Image](https://github.com/user-attachments/assets/e80eea66-f465-4f05-a3ed-49df76a98787)

## Are you able to fix this issue?
Yes

PR #2847, opened 2025-02-04
frontend: app: Pass backendToken via IPC

skoeva:
This change removes the backendToken from the URL and instead passes it via IPC. This ensures that restarting the app does not create new instances with unique URLs.

Fixes: #2846 

### Testing
- [X] Expose parseKubeConfig at the bottom of `clusterApi.ts`:
```ts
(window as any).parseKubeConfig = parseKubeConfig
```
- [X] Start Headlamp using `npm run dev` and proxy fetch to print a stack trace with headers for each call by entering the following in the console:
```ts
window.fetch = new Proxy(window.fetch, {
  apply: (target, thisArg, args) => {
    console.groupCollapsed("+++ Intercepted Network Request");
    console.log("URL:", args[0]);
    console.log("Options:", args[1]);
    console.trace("+++ Request Stack Trace:");
    console.groupEnd();
    return target.apply(thisArg, args);
  }
});
```
- [X] Manually call parseKubeConfig by entering the following in the console:
```ts
window.parseKubeConfig({ kubeconfig: "test-kubeconfig-data" });
```
- [X] Inspect the call headers and ensure that backendToken is passed as a header with the key `X-HEADLAMP_BACKEND-TOKEN`

### DIFF
--- a/frontend/src/components/App/AppContainer.tsx
+++ b/frontend/src/components/App/AppContainer.tsx
@@ -2,11 +2,16 @@ import { GlobalStyles } from '@mui/material';
 import { SnackbarProvider } from 'notistack';
 import React from 'react';
 import { BrowserRouter, HashRouter } from 'react-router-dom';
-import helpers from '../../helpers';
+import helpers, { setBackendToken } from '../../helpers';
 import Plugins from '../../plugin/Plugins';
 import ReleaseNotes from '../common/ReleaseNotes/ReleaseNotes';
 import Layout from './Layout';
 import { PreviousRouteProvider } from './RouteSwitcher';
+
+window.desktopApi?.send('request-backend-token');
+window.desktopApi?.receive('backend-token', (token: string) => {
+  setBackendToken(token);
+});
 
 export default function AppContainer() {
   const Router = ({ children }: React.PropsWithChildren<{}>) =>
--- a/frontend/src/helpers/index.ts
+++ b/frontend/src/helpers/index.ts
@@ -362,14 +362,17 @@ export function getWebsocketMultiplexerEnabled(): boolean {
 
 /**
  * The backend token to use when making API calls from Headlamp when running as an app.
- * The app opens the index.html?backendToken=... and passes the token to the frontend
- * in this way. The token is then used in the getHeadlampAPIHeaders function below.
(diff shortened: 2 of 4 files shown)


frontend/src/helpers/index.ts
```diff
@@ -362,14 +362,17 @@ export function getWebsocketMultiplexerEnabled(): boolean {
...
- * in this way. The token is then used in the getHeadlampAPIHeaders function below.
+ * The token is requested from the main process via IPC once the renderer is ready,
+ * and stored for use in the getHeadlampAPIHeaders function below.
  *
- * The app also passes the token to the headlamp-server via HEADLAMP_BACKEND_TOKEN env var.
+ * The app also sets HEADLAMP_BACKEND_TOKEN in the headlamp-server environment,
+ * which the server checks to validate requests containing this same token.
  */
-const backendToken =
-  import.meta.env.REACT_APP_HEADLAMP_BACKEND_TOKEN ||
```
joaquimrocha, line 371: I think this was here for testing. Should we keep it?

skoeva: I can add this back in, and if we decide we still need this, we can open a new issue for it

frontend/src/components/App/AppContainer.tsx
```diff
@@ -2,13 +2,20 @@ import { GlobalStyles } from '@mui/material';
...
 import { BrowserRouter, HashRouter } from 'react-router-dom';
-import helpers from '../../helpers';
+import helpers, { setBackendToken } from '../../helpers';
 import Plugins from '../../plugin/Plugins';
 import ReleaseNotes from '../common/ReleaseNotes/ReleaseNotes';
 import Layout from './Layout';
 import { PreviousRouteProvider } from './RouteSwitcher';
 
 export default function AppContainer() {
+  React.useEffect(() => {
```
sniok, line 12: I think this be placed outside the component? Then it'll request and send backend token right away when the file is loaded

joaquimrocha (dismissed):

joaquimrocha:
Wait, looks like you haven't checked my changes which move the `backendToken` message as `backend-token`, since this way it's consistent with the rest of the messages

joaquimrocha (approved):

Issue #2854, opened 2025-02-05
Documentation tab in editor dialog breaks on default value


skoeva:
## Describe the bug
When opening the editor with the default text

![Image](https://github.com/user-attachments/assets/f2c97e2f-271b-4eeb-86af-9ee2a0766515)

navigating to the documentation tab causes the app to crash.

## Are you able to fix this issue?
[Yes (I will propose a PR) / No.]
Yes

## Additional Context
~~This may be a regression from #1996~~
Not a regression, just how the logic in the component typically works

PR #2856, opened 2025-02-05
frontend: EditorDialog: Handle null YAML in docs tab

skoeva:
This change gracefully handles comments from the editor dialog in the documentation tab, preventing the app from crashing.

Fixes: #2854 

### Testing
- [X] Open a cluster in Headlamp and click the 'Create' button in the sidebar
- [X] Click on the 'Documentation' tab with the default comment
- [X] Ensure the app does not crash and that the tab displays "No documentation available."

image

### DIFF
--- a/frontend/src/components/common/Resource/EditorDialog.tsx
+++ b/frontend/src/components/common/Resource/EditorDialog.tsx
@@ -234,6 +234,7 @@ export default function EditorDialog(props: EditorDialogProps) {
       res.format = 'yaml';
       try {
         res.obj = yaml.loadAll(code) as KubeObjectInterface[];
+        res.obj = res.obj.filter(obj => !!obj);
         return res;
       } catch (e) {
         res.error = new Error((e as Error).message || t('Invalid YAML'));
--- a/frontend/src/components/common/Resource/DocsViewer.tsx
+++ b/frontend/src/components/common/Resource/DocsViewer.tsx
@@ -98,6 +98,8 @@ function DocsViewer(props: DocsViewerProps) {
     <>
       {docsLoading ? (
         <Loader title={t('Loading documentation')} />
+      ) : docs.length === 0 ? (
+        <Empty>{t('No documentation available.')}</Empty>
       ) : (
         docs.map((docSpec: any, idx: number) => {
           if (!docSpec.error && !docSpec.data) {
--- a/frontend/src/components/common/Resource/__snapshots__/EditorDialog.EditorDialogWithResource.stories.storyshot
+++ b/frontend/src/components/common/Resource/__snapshots__/EditorDialog.EditorDialogWithResource.stories.storyshot
@@ -204,7 +204,17 @@
           >
             <div
               class="MuiBox-root css-94bk7g"
-            />
+            >
+              <div
+                class="MuiBox-root css-19midj6"
+              >
+                <p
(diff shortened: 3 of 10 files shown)


vyncent-t (commented):
the tab does load empty, not sure if this is the expected behavior? should it be rendering the empty field here? 

this is from DocsViewer comp

image

skoeva:
@vyncent-t good point, adding a placeholder for when there is no item to generate docs for

vyncent-t:
I think you might need to run i18n again here, or reuse the existing no docs line

skoeva:
There are some loose translations, will open a separate PR to address those lmao

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -202,18 +202,18 @@ export default function EditorDialog(props: EditorDialogProps) {
   }
 
   function getObjectsFromCode(codeInfo: typeof originalCodeRef.current): {
-    obj: KubeObjectInterface[] | null;
+    obj: KubeObjectInterface[];
```
sniok, line 205: is this function refactor related to the crash? if not please put it in a separate commit

skoeva: the refactor is what ensures the page doesn't crash, the change to the other file is just placeholder text because the refactor causes the page to go blank

skoeva:
> How about adding a story state which triggers this?

@illume Seems like we can't create a template with a comment in plain text, and we also don't have stories for DocsViewer (maybe one of the new contributors can work on that)

illume:
@skoeva there's a story now added in:
- https://github.com/headlamp-k8s/headlamp/pull/2872

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -230,18 +230,12 @@ export default function EditorDialog(props: EditorDialogProps) {
...
-    }
-
-    if (!!res.obj) {
-      res.error = null;
+    res.format = 'yaml';
+    try {
+      const parsedYaml = yaml.loadAll(code) as KubeObjectInterface[];
+      res.obj = (parsedYaml || []).filter(item => item !== null);
+    } catch (e) {
+      res.error = new Error((e as Error).message || t('Invalid YAML'));
```
joaquimrocha, line 238: This code is not equivalent to what we had before: if you fail to JSON convert it before, then res.error is not null.
So this block now converts using YAML, but if it succeeds, the error will still remain set because nothing it resetting it.

So I think a cleaner fix for the issue just adding `res.obj = res.obj.filter(obj => !!obj);` after the res.obj is set (line 236 of main).

Then if you want to refactor some of the code (though keeping it equivalent) and add the new empty string to the docs viewer, please do it in different commits.

joaquimrocha (changes requested):

frontend/src/components/common/Resource/EditorDialog.tsx
skoeva: I think what I was aiming for was ensuring that the app never crashed because of wrong input in the dialog (seems inconvenient and unnecessary), but if we only want to handle things like comments then I can make this change

joaquimrocha (approved):
Feel free to add the refactor if you want, but take into account the issue with the code not being equivalent as I mentioned. I will merge this one.

PR #2857, opened 2025-02-05
frontend: i18n: Add loose translations

skoeva:
This change fixes the loose translations causing the i18n failure in the CI.


illume (approved):
👍 thanks!!

PR #2861, opened 2025-02-06
frontend: Add e2e-tests for multiple clusters

### LINKED ISSUE #2460: e2e tests for multiple clusters

More confidence in multi cluster work if there are end to end tests to test it as well.

At least one end to end test that tests multiple clusters.

Developers making multi cluster features, and users using them.

I would help with it yes.

adwait-godbole:
Fixes #2460 

Screenshot from 2025-02-07 01-15-57

Once this gets merged, #2831 can also be unblocked.

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,4 +1,3 @@
-
 # "Setup minikube as CI step in GitHub Actions"
 # https://minikube.sigs.k8s.io/docs/tutorials/setup_minikube_in_github_actions/
 # https://github.com/marketplace/actions/setup-minikube
@@ -19,7 +18,7 @@ on:
       - main
       - rc-*
       - testing-rc-*
-  
+
 permissions:
   contents: read
 
@@ -34,10 +33,20 @@ jobs:
     - uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7 # v3.8.2
       with:
         node-version: 20.x
-    - name: Start cluster
+    - name: Start Cluster 1
       uses: helm/kind-action@0025e74a8c7512023d06dc019c617aa3cf561fde # v1.0.0
       with:
-          cluster_name: test
+        cluster_name: test
+    - name: Rename Cluster 1 kind context
+      run: kubectl config rename-context kind-test test
+    - name: Start Cluster 2
+      uses: helm/kind-action@0025e74a8c7512023d06dc019c617aa3cf561fde # v1.0.0
+      with:
+        cluster_name: test2
+    - name: Rename Cluster 2 kind context
+      run: kubectl config rename-context kind-test2 test2
+    - name: Use Cluster 1 context
+      run: kubectl config use-context test
       # now you can run kubectl to see the pods in the cluster
     - name: Try the cluster!
       run: kubectl get pods -A
@@ -113,12 +122,30 @@ jobs:
           echo "Examples content: $examples_content"
           exit 1
         fi
-    - name: Deploy to cluster
-      run: kubectl apply -f e2e-tests/kubernetes-headlamp-ci.yaml
+    - name: Multi-Cluster Setup
+      run: |
+        # Use Cluster 1 context
+        kubectl config use-context test
+        export TEST_CA_DATA=$(kubectl config view --raw --minify -o jsonpath='{.clusters[0].cluster.certificate-authority-data}')
+        export TEST_SERVER="https://$(kubectl get nodes -o=jsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}'):6443"
+        kubectl create serviceaccount headlamp-admin --namespace kube-system
+        kubectl create clusterrolebinding headlamp-admin --serviceaccount=kube-system:headlamp-admin --clusterrole=cluster-admin
+        echo "HEADLAMP_TEST_TOKEN=$(kubectl create token headlamp-admin --duration 24h -n kube-system)" >> $GITHUB_ENV
+
+        # Use Cluster 2 context
+        kubectl config use-context test2
+        export TEST2_CA_DATA=$(kubectl config view --raw --minify -o jsonpath='{.clusters[0].cluster.certificate-authority-data}')
+        export TEST2_SERVER="https://$(kubectl get nodes -o=jsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}'):6443"
+        kubectl create serviceaccount headlamp-admin --namespace kube-system
+        kubectl create clusterrolebinding headlamp-admin --serviceaccount=kube-system:headlamp-admin --clusterrole=cluster-admin
+        echo "HEADLAMP_TEST2_TOKEN=$(kubectl create token headlamp-admin --duration 24h -n kube-system)" >> $GITHUB_ENV
+
+        envsubst < e2e-tests/kubernetes-headlamp-ci.yaml | kubectl --context=test apply -f -
     - name: Run e2e tests
       run: |
         echo "------------------------------------sleeping 12...------------------------------------"
         sleep 12
+        kubectl config use-context test
         kubectl get services --all-namespaces
         kubectl get deployments -n kube-system
         echo "------------------Waiting for headlamp deployment to be available...------------------"
@@ -129,10 +156,9 @@ jobs:
         export SERVICE_URL="http://${IP_ADDRESS}:${SERVICE_PORT}"
         echo $SERVICE_URL
         curl -L $SERVICE_URL | grep -q "Headlamp: Kubernetes Web UI"        
-        echo "--------------------------------Getting HEADLAMP_TOKEN--------------------------------"
-        kubectl create serviceaccount headlamp-admin --namespace kube-system
-        kubectl create clusterrolebinding headlamp-admin --serviceaccount=kube-system:headlamp-admin --clusterrole=cluster-admin
-        export HEADLAMP_TOKEN=$(kubectl create token headlamp-admin --duration 24h -n kube-system)
+        echo "--------------------------------Export Tokens Received From Previous Step--------------------------------"
+        export HEADLAMP_TEST_TOKEN=${HEADLAMP_TEST_TOKEN}
+        export HEADLAMP_TEST2_TOKEN=${HEADLAMP_TEST2_TOKEN}
         echo "---------------------------------Certificate handling---------------------------------"
         export KUBECONFIG=$HOME/.kube/config
         ca_data=$(yq e '.clusters[0].cluster."certificate-authority-data"' $KUBECONFIG | base64 --decode)
--- a/e2e-tests/tests/headlampPage.ts
+++ b/e2e-tests/tests/headlampPage.ts
@@ -2,21 +2,21 @@
 import { expect, Page } from '@playwright/test';
 
 export class HeadlampPage {
-  constructor(private page: Page) {}
+  private testURL: string;
 
-  async authenticate() {
-    // Go to the authentication page
-    const url = process.env.HEADLAMP_TEST_URL;
-    await this.page.goto(url || '/');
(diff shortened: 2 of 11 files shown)


e2e-tests/tests/headlampPage.ts
```diff
@@ -74,8 +84,8 @@ export class HeadlampPage {
     await this.page.click('a.MuiMenuItem-root:has-text("Log out")');
     await this.page.waitForLoadState('load');
 
-    // Expects the URL to contain c/main/token
-    await this.hasURLContaining(/.*token/);
+    // Expects the URL to contain c/test/token
+    // await this.hasURLContaining(/.*token/);
```
illume, line 88: I guess this doesn't work now? Do you know why this doesn't work?
Is it because there are two clusters?

I think it's fine to leave it commented out if it's not working and it's too hard to fix.

.github/workflows/build-container.yml
```diff
@@ -34,10 +33,20 @@ jobs:
     - uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7 # v3.8.2
       with:
         node-version: 20.x
-    - name: Start cluster
+    - name: Start Cluster 1
+      uses: helm/kind-action@0025e74a8c7512023d06dc019c617aa3cf561fde # v1.0.0
+      with:
+        cluster_name: test
```
illume, line 39: I guess we would need to update the instructions for running the tests locally. In the e2e-tests/README.md file.

illume (changes requested):
This is great, thank you. It will really help test Headlamp.

I left a request to add some updated instructions for creating two clusters locally for testing in the README.md file. Apart from that, looks good to me.

Also, I asked for a review from @skoeva or @vyncent-t because I know they touched these tests more recently than me and they might have an opinion.

e2e-tests/tests/headlampPage.ts
adwait-godbole: > I guess this doesn't work now? Do you know why this doesn't work?
> Is it because there are two clusters?

Yes it isn't working. When I was testing this multi-cluster setup on my local machine, on logout, this expected URL was not appearing in the browser. I am not really sure about why this might be happening so can't really comment on the underlying reason :(.

.github/workflows/build-container.yml
adwait-godbole: Yes. Working on it.

adwait-godbole: Hi @illume, while working on changing the docs at e2e-tests/README.md, I noticed that currently the docs try to deploy Headlamp inside of a minikube cluster using the addon method as shown [here](https://github.com/headlamp-k8s/headlamp/blob/a725949050866d07309eb035c332b4b4620ca96c/e2e-tests/README.md?plain=1#L14). 

And the headlamp addon tries to deploy the YAML manifests that are defined at https://github.com/kubernetes/minikube/blob/86ca9841adfc02f44977a287cd2248f33e689d9c/deploy/addons/headlamp. 

However, in my current PR in order for us to write multi-cluster e2e-tests, I had to slightly modify the deployment.yaml file to use a non-default kube config as guided in the docs over [here](https://headlamp.dev/docs/latest/installation/in-cluster/#use-a-non-default-kube-config-file). You can see the [deployment.yaml](https://github.com/headlamp-k8s/headlamp/blob/ae0c94bb2e8d845700bed26dbd7e8d9315366175/e2e-tests/kubernetes-headlamp-ci.yaml#L58-#L125) changes in this PR. 

So the instructions for running e2e-tests on local might have to slightly change as per the new YAML file and we might not be able to use the addon method for deploying Headlamp in-cluster, rather we will have to use this [kubernetes-headlamp-ci.yaml]() file for deployment.

So, is it fine if I make the documentation changes accordingly?

illume: Yes I think what you propose is a good idea.

Probably using minikube addon also means that it's only testing the 'old' released version of headlamp... and not the main branch.

e2e-tests/tests/headlampPage.ts
illume: Cool. Thanks for the explanation.

(mainly I wanted to make sure it wasn't commented out by accident)

.github/workflows/build-container.yml
adwait-godbole: > Probably using minikube addon also means that it's only testing the 'old' released version of headlamp... and not the main branch.

@illume So should we drop the idea of using minikube addon method for setup before e2e-tests and do a manual setup instead? The e2e-tests docs will then have multiple steps in order to achieve this mostly following things that are happening in our [build-container.yml](https://github.com/headlamp-k8s/headlamp/blob/main/.github/workflows/build-container.yml) file.

illume: >So you are fine with dropping the idea of using minikube addon method for setup before e2e-tests and writing a manual setup instead, right?

Yeah. But I'm thinking this could be done as future work in another PR (perhaps by someone else).

adwait-godbole: hi @illume, I was trying to write the docs and test my changes once again on local. I was primarily referring [build-container.yml](https://github.com/headlamp-k8s/headlamp/blob/0358518679e87c8b01a133ad42db5a6e9651a360/.github/workflows/build-container.yml) workflow file for the setup alongside [kubernetes-headlamp-ci.yaml](https://github.com/headlamp-k8s/headlamp/blob/0358518679e87c8b01a133ad42db5a6e9651a360/e2e-tests/kubernetes-headlamp-ci.yaml) manifest. However, after completing the setup, I noticed that the Headlamp UI does not seem to show me the following things. 

1) No Namespaces option under "Cluster" Tab. (Also visiting "/c/test/namespaces" is giving me a 404 page)
2) No "Workloads" tab in the side bar for accessing "Pod" button to click. However visiting "/c/test/pods" is working fine).

Here's a screenshot of what I am seeing.

Screenshot from 2025-02-18 09-17-41

Can you help me out why this might be the case?

I tried doing this setup **without any plugins** and then I didn't see any such discrepancies. Are plugins causing this behavior? Was that the reason why in the namespaces test over [here](https://github.com/headlamp-k8s/headlamp/blob/29ff50e13f1f6c989fbc88a0229843b341b62fa8/e2e-tests/tests/namespaces.spec.ts#L10) and in the pods test over [here](https://github.com/headlamp-k8s/headlamp/blob/29ff50e13f1f6c989fbc88a0229843b341b62fa8/e2e-tests/tests/podsPage.spec.ts#L19), these comments were added --> `// If there's no namespaces permission, then we return` and `// if no pod permission, return` ?

adwait-godbole: It would be helpful if you could tell me what UI you see on performing the e2e-tests setup :)

skoeva: Is that what you're seeing after manually running the e2e test steps?

illume: Ah, it looks like you have some plugins installed which remove stuff

Delete your plugins and try again.

The backend prints out the location for the plugin path if you can’t find it.

e2e-tests/tests/headlampPage.ts
```diff
@@ -2,21 +2,33 @@
...
+      await this.page.goto(`${this.testURL}/c/${cluster}`, {
+        waitUntil: 'networkidle',
+      });
+      await this.page.waitForLoadState('networkidle');
+    }
 
-  async authenticate() {
-    // Go to the authentication page
-    const url = process.env.HEADLAMP_TEST_URL;
-    await this.page.goto(url || '/');
```
vyncent-t, line 10: This would could be a different authentication function for your use case of skipping navigation

Currently this is hard corded to false and a hard coded cluster name of 'test'
it also sends the test to the test cluster overview page by default- skipping the home page, ideally these tests should run the auth each time and navigation should be a different step.

e2e-tests/tests/headlampPage.ts
```diff
@@ -2,21 +2,33 @@
...
+    token = process.env.HEADLAMP_TEST_TOKEN,
+    skipNavigation = false
+  ) {
+    if (!skipNavigation) {
+      // Go to the authentication page
+      await this.page.goto(`${this.testURL}/c/${cluster}`, {
+        waitUntil: 'networkidle',
+      });
+      await this.page.waitForLoadState('networkidle');
+    }
```
vyncent-t, line 22: I would rec reverting the original auth func and moving these changes into its own auth func, maybe something like `authenticateToCluster` for your tests

e2e-tests/tests/headlampPage.ts
adwait-godbole: > Currently this is hard corded to false and a hard coded cluster name of 'test'

I won't call it "hardcoding" per say 😅. I was just trying to provide "default" values for those parameters incase they weren't passed when calling it.

e2e-tests/tests/headlampPage.ts
adwait-godbole: Hi @vyncent-t . Thanks for the review.

> I would rec reverting the original auth func and moving these changes into its own auth func, maybe something like authenticateToCluster for your tests and use where needed 

so by this do you mean we should keep both `authenticate` and `authenticateToCluster` functions? Won't that be a little bit confusing?  Or we could make `authenticate` function private to `HeadlampPage` class and expect all other tests to call `authenticateToCluster` everywhere and internally it will call `authenticate`,   the function signature for `authenticateToCluster` would then look something like this -->  

```typescript
  async authenticateToCluster(name: string, token: string) {
    await this.navigateTopage(`/c/${name}`);
    await this.authenticate(token);
  }
```

but then again, the other tests will always have to pass **both the cluster name and the corresponding token** when calling it. Is that fine? (to avoid this hassle i was using the "default" parameters" way that I wrote about [here](https://github.com/headlamp-k8s/headlamp/pull/2861/files#r1965012252). we could still use that way in `authenticateToCluster` function as well).

Lmk your thoughts.

vyncent-t: Similar to what you have here for this logic, just something more along the lines of

```typescript
  async navigateToCluster(name: string, token: string) {
    await this.authenticate(token);
    await this.navigateTopage(`/c/${name}`);
  }
```

That way we do not attach too many navigation steps into the authenticate function, the CI can be flakey when it comes to the auth step, especially with multiple auth token contexts.

adwait-godbole:
Hi @vyncent-t, can you please re-review this once? Thanks.

illume:
It looks like there is a test failing when I look at the logs, but the CI job is passing.

https://github.com/headlamp-k8s/headlamp/actions/runs/13517271859/job/37768475888?pr=2861#step:18:1154
```
Running 24 tests using 1 worker
·······················×±

  1) [chromium] › podsPage.spec.ts:5:5 › multi tab create delete pod ───────────────────────────────

    Test timeout of 60000ms exceeded.

    Error: page.click: Test ended.
    Call log:
      - waiting for locator('span:has-text("Workloads")')


       at podsPage.ts:7

       5 |
       6 |   async navigateToPods() {
    >  7 |     await this.page.click('span:has-text("Workloads")');
         |                     ^
       8 |     await this.page.waitForLoadState('load');
       9 |     await this.page.waitForSelector('span:has-text("Pods")');
      10 |     await this.page.waitForLoadState('load');

        at podsPage.navigateToPods (/home/runner/work/headlamp/headlamp/e2e-tests/tests/podsPage.ts:7:21)
        at /home/runner/work/headlamp/headlamp/e2e-tests/tests/podsPage.spec.ts:34:25

  1 flaky
    [chromium] › podsPage.spec.ts:5:5 › multi tab create delete pod ────────────────────────────────
  23 passed (1.9m)

```

illume (approved):
🎉 🎈 thank you

illume:
@adwait-godbole 
- https://github.com/headlamp-k8s/headlamp/pull/2960

PR #2867, opened 2025-02-07
github: Chocolatey needs to be updated always on the main branch

joaquimrocha:
Tested already from branch and it works.


illume (approved):
👍 thanks.

(the failed container build is unrelated)

PR #2875, opened 2025-02-09
frontend: remove obsolete API call for CronJobs

### LINKED ISSUE #2868: Obsolete API For CronJobs
## Describe the bug

Removed k8s api `/apis/batch/v1beta1/cronjobs` is being called by headlamp.

`In the last 30 days, API clients in your cluster attempted to call APIs that were removed in Kubernetes 1.25. These APIs are no longer available. Functionality that depends on these APIs might be broken.`

## To Reproduce

List or view detail of CronJobs.

Cluster is reporting calls to `/apis/batch/v1beta1/cronjobs` which was removed is K8s 1.25.  This flags by GKE automated scanners.

 - Installation type: Windows
 - Headlamp Version: 0.28.1
 - Other: GKE Autopilot Cluster 1.31

No

adwait-godbole:
Fixes #2868

### DIFF
--- a/frontend/src/lib/k8s/cronJob.ts
+++ b/frontend/src/lib/k8s/cronJob.ts
@@ -38,7 +38,7 @@ export interface KubeCronJob extends KubeObjectInterface {
 class CronJob extends KubeObject<KubeCronJob> {
   static kind = 'CronJob';
   static apiName = 'cronjobs';
-  static apiVersion = ['batch/v1', 'batch/v1beta1'];
+  static apiVersion = 'batch/v1';
   static isNamespaced = true;
 
   get spec() {


sniok:
Thanks for taking a look at it but I'm afraid it's not that easy. We still need to support that version. 

We have multiple apiVersions for resources because headlamp doesn't know which one will be available in the cluster, so we do a check during runtime and test all the versions to determine which one is working.

I think to fix that issue we need to modify the `getWorkingEndpoint` function that currently checks all the versions in parallel, maybe sequential check of the versions would be better.

illume (approved):
The bug with parallel checks is a serious issue, but is separate to the reported issue here.

Old versions of Kubernetes that only have `batch/v1beta1` are not supported anymore.

This solves the users reported problem. It's an improvement over what we have.

skoeva:
@illume The [installation docs](https://headlamp.dev/docs/latest/installation/#:~:text=Authentication%20/%20Log%2Din%E2%80%8B%20*%20Kubernetes%201.24+%20*%20Older%20versions.) still point to supporting Kubernetes 1.24 which has the deprecated `batch/v1beta1`, we should probably change the docs (and maybe advertise) if we don't plan on supporting it anymore

illume:
>@illume The [installation docs](https://headlamp.dev/docs/latest/installation/#:~:text=Authentication%20/%20Log%2Din%E2%80%8B%20*%20Kubernetes%201.24+%20*%20Older%20versions.) still point to supporting Kubernetes 1.24 which has the deprecated batch/v1beta1, we should probably change the docs (and maybe advertise) if we don't plan on supporting it anymore

We don't need the beta API if the stable one is available.

Cronjob API stabilized in 1.21 and Kubernetes 1.21 was released on April 8, 2021. Which is quite some years EOL, and also an older version which we state that we support.

illume:
> It'll break the app for people running older versions so it's not an improvement for everyone.

If there were people running this years old and unsupported Kubernetes version, this would be true.

No cloud provider has supported it for years, Kubernetes EOL'd that version years ago and doesn't support it, and we don't support that version of Kubernetes either.

That last part bears repeating: we don't support that 2020 version of Kubernetes 1.20 (last version that doesn't have cronjob stable API).

> The parallel check will test all the endpoints and pick the one that works. The fact that it tries all of them is not a bug. It can be improved by running checks sequentially from the most stable to beta so it doesn't trigger automated scanners

There are two issues I'm interested in:
- The bug where it generates warnings where it doesn't need to. Probably we will run into this issue in the future with more endpoints being retired. But this PR fixes this issue people are experiencing now. I think we need to look at retiring other beta/alpha endpoints where it makes sense. Also as you suggest looking at that code to perhaps act sequentially, or to detect k8s versions or some other many of behaving better.
- Does it only do one extra request per session, or many? Any unnecessary load on the control plane is a bug, because that part is harder to scale and the consequences of it being overloaded are high. If it's just one extra request per session per beta endpoint that's not much of an issue of course.

illume:
I'd like to get a fix for this in...

@skoeva @sniok ?

skoeva:
I suppose we can open a new issue for sequential checks later, and keep this as a temp workaround? Assuming we need to keep supporting the old version, though seems that we probably don't at this point

skoeva (approved):

sniok (approved):
As long as we address the core issue, I'm okay with this fix for now

illume:
I'll leave opening an issue for updating the rest of the obsolete beta APIs, and fixing API selector to use feature detection up to anyone interested.

PR #2876, opened 2025-02-09
backend: Add app mode flag to fix network exposure

### LINKED ISSUE #2852: Headlamp-server process is exposed to network in app mode
headlamp-server process that is started when you launch headlamp as a desktop app is accessible to network because it listens to "*:4466"

With in-cluster mode it's expected but in app mode it should only listen to local connection from the electron (loopback 127.0.0.1)

## To Reproduce
Steps to reproduce the bug:
1. Start the desktop app
2. Open IP:4466/ from other device on the same network

---

I think to fix this issue, we need to be able to tell headlamp-server that it runs in app mode and then based on that it will listen on 127.0.0.1

adwait-godbole:
Fixes #2852

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,15 +91,16 @@ frontend-build:
 frontend-build-storybook:
 	cd frontend && npm run build-storybook
 
+APP_MODE ?= false
 
 run-backend:
 	@echo "**** Warning: Running with Helm and dynamic-clusters endpoints enabled. ****"
 
 ifeq ($(UNIXSHELL),true)
-	HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/* -enable-dynamic-clusters
+	HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/* $(if $(filter $(APP_MODE), true), -app)
 else
 	@echo "**** Running on Windows without bash or zsh. ****"
-	@cmd /c "set HEADLAMP_BACKEND_TOKEN=headlamp&& set HEADLAMP_CONFIG_ENABLE_HELM=true&& set HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true&& backend\headlamp-server -dev -proxy-urls https://artifacthub.io/*" -enable-dynamic-clusters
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -91,15 +91,16 @@ frontend-build:
 frontend-build-storybook:
 	cd frontend && npm run build-storybook
 
+APP_MODE ?= false
 
 run-backend:
 	@echo "**** Warning: Running with Helm and dynamic-clusters endpoints enabled. ****"
 
 ifeq ($(UNIXSHELL),true)
-	HEADLAMP_BACKEND_TOKEN=headlamp HEADLAMP_CONFIG_ENABLE_HELM=true HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true ./backend/headlamp-server -dev -proxy-urls https://artifacthub.io/* -enable-dynamic-clusters
```
adwait-godbole, line 99: here I removed `-enable-dynamic-clusters` since we were already setting HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true as an env var

adwait-godbole:
Hi @sniok. Appreciate your review and quick feedback.

> so please add the -app flag inside startServer function in app/electron/main.ts file

I have made the above changes.

> I think setting -app inside makefile is nice but not necessary

What if someone runs `cd app && npm run start`? The call to run backend in `start.js` script also needs to tell the backend process that we are running in app mode since `EXTERNAL_SERVER=true` env will be set because of `dev-only-app` later in the `start.js` script and `startServer` won't be called in `main.ts`. Am I getting this right or missing something internal?

app/scripts/start.js
```diff
@@ -6,7 +6,7 @@
  */
 const { spawn } = require('child_process');
 
-const serverProcess = spawn('cd ../ && make backend && make run-backend', [], {
+const serverProcess = spawn('cd ../ && make backend && make run-backend APP_MODE=true', [], {
```
illume, line 9: Did you confirm this works? I think you need to add "APP_MODE=true'" before make run-backend?

illume:
> What if someone runs cd app && npm run start? The manual call to run backend in start.js script also needs to tell the backend process that we are running in app mode right? since EXTERNAL_SERVER=true env will be set because of dev-only-app later in this start.js script and startServer won't be called in main.ts. Am I getting this right or missing something internal?

Probably it would be better inside run-backend. Because we almost always want to only listen on localhost. Someone wanting to do otherwise can do it manually.

Now I'm not sure if app mode is really appropriate. We always want it more secure by default (and to listen only on localhost). But we don't always want app mode. If something later is added to app mode I guess this could cause a problem. This is why I think instead of app mode, it should just set the localhost address by default in run-backend.

app/scripts/start.js
adwait-godbole: @illume yes I have tried it myself. Either way works. Passing before make run-backend also works and passing after it also works. Should I pass before it?

joaquimrocha:
I have merged the changes to a different branch so we can improve upon it as a base.

PR #2717, opened 2024-12-20
frontend: Refactor prepareSidebarItems into a hook

sniok:
Fixes #2714 
This hook will also support arbitrary depth for the sidebar

Testing done:

Added unit tests
Tested with different plugins to make sure the sidebar works

PR #2879, opened 2025-02-10
frontend: NavigationTabs: Move early return after all hooks to prevent a crash

sniok:
Fixes regressions introduced in #2717

First is a page crash when sidebar is collapsed - early return in a react component before a hook was called caused a crash
Second is duplicated entries in NavigationTabs - mutation of the sidebar entries inside NavigationTabs was causing issues

How to test:
Open/close sidebar
Navigate around


sniok:
Noticed another regression and will add a fix here for it too

illume (approved):
👍 thanks.

Works well for me.

PR #130, opened 2020-11-25
Allow custom certificates in oidc mode

ashu8912:
Allow oidc flow with custom certificates when run in insecure mode

# Testing done
Done!! tested with setting up oidc auth flow with dex followed by untrusted cert

fixes #127

Issue #127, opened 2020-11-20
Doesn't like cert signed by our local CA, but docs don't seem to provide method to give it our root certificate


cprivitere:
**Description**

I get an SSL error when trying to use on-prem Dex with a cert signed by our local cert authority. Is there a way to make this work? 

**Impact**

Prevents me from using OIDC

**Environment and steps to reproduce**

1. **Set-up**: Follow the in-cluster directions.
2. **Task**: Try to sign in.
3. **Error**: From the headlamp logs:
```
2020/11/19 23:58:21 Error while fetching the provider from https://dex.tst-k8s.epic.com/callback error Get https://dex.tst-k8s.epic.com/callback/.well-known/openid-configuration: x509: certificate signed by unknown authority
2020/11/19 23:58:21 http: panic serving 10.96.77.112:41566: runtime error: invalid memory address or nil pointer dereference
```
This results in a pop up window that says 502 bad gateway.

**Expected behavior**
Be directed to dex to login

**Additional information**
I'm already passing -insecure-ssl and it did not help.

joaquimrocha:
@cprivite I will try to look into this next week. Thank you for reporting.

gecube:
Hi!

> I'm already passing -insecure-ssl and it did not help.

The same issue. Is there any good documentation how to pass a custom CA bundle to headlamp? I am using the same scheme as described here https://github.com/headlamp-k8s/headlamp/issues/1192#issuecomment-1672526456 and still no success with getting it running.

```
kubectl logs -n headlamp headlamp-5cbbd759f6-6jc86 
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":305,"time":"2024-04-09T13:57:58Z","message":"Creating Headlamp handler"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":306,"time":"2024-04-09T13:57:58Z","message":"Kubeconfig path: "}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":307,"time":"2024-04-09T13:57:58Z","message":"Static plugin dir: /headlamp/static-plugins"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":308,"time":"2024-04-09T13:57:58Z","message":"Plugins dir: /headlamp/plugins"}
[...]

PR #2880, opened 2025-02-10
backend(oidc): Add support for custom CA and skipping TLS verification

yolossn:
this patch adds support for custom CA and skipping TLS verification for OIDC provider verification. This improves flexibility for users working with self-signed OIDC providers.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -54,6 +55,8 @@ type HeadlampConfig struct {
 	oidcClientID          string
 	oidcClientSecret      string
 	oidcIdpIssuerURL      string
+	oidcSkipTLSVerify     bool
+	oidcCACert            string
 	baseURL               string
 	oidcScopes            []string
 	proxyURLs             []string
@@ -366,7 +369,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	if config.useInCluster {
 		context, err := kubeconfig.GetInClusterContext(config.oidcIdpIssuerURL,
 			config.oidcClientID, config.oidcClientSecret,
-			strings.Join(config.oidcScopes, ","))
+			strings.Join(config.oidcScopes, ","), config.oidcSkipTLSVerify, config.oidcCACert)
 		if err != nil {
 			logger.Log(logger.LevelError, nil, err, "Failed to get in-cluster context")
 		}
@@ -545,6 +548,32 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+
+		if oidcAuthConfig.SkipTLSVerify {
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
+			}
+			insecureClient := &http.Client{Transport: tr}
+			ctx = oidc.ClientContext(ctx, insecureClient)
+		}
+
+		if oidcAuthConfig.CACert != "" {
+			caCertPool := x509.NewCertPool()
+			if !caCertPool.AppendCertsFromPEM([]byte(oidcAuthConfig.CACert)) {
+				logger.Log(logger.LevelError, map[string]string{"oidc-ca-file": oidcAuthConfig.CACert},
+					errors.New("failed to append oidc-ca-file to cert pool"), "failed to append oidc-ca-file to cert pool")
+				http.Error(w, "invalid oidc-ca-file", http.StatusBadRequest)
+				return
+			}
+
+			secureTLSClient := &http.Client{
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{RootCAs: caCertPool}, //nolint:gosec
+				},
+			}
+			ctx = oidc.ClientContext(ctx, secureTLSClient)
+		}
+
 		provider, err := oidc.NewProvider(ctx, oidcAuthConfig.IdpIssuerURL)
 		if err != nil {
 			logger.Log(logger.LevelError, map[string]string{"idpIssuerURL": oidcAuthConfig.IdpIssuerURL},
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -36,7 +36,7 @@ func main() {
 	kubeConfigStore := kubeconfig.NewContextStore()
 	multiplexer := NewMultiplexer(kubeConfigStore)
 
-	StartHeadlampServer(&HeadlampConfig{
+	headlampConfig := &HeadlampConfig{
 		useInCluster:          conf.InCluster,
 		kubeConfigPath:        conf.KubeConfigPath,
 		port:                  conf.Port,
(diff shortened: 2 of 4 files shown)


yolossn:
I’ve added support for both skipping TLS verification (`oidc-skip-tls-verify`) and specifying a CA file (`oidc-ca-file`) for OIDC.

Although TLS skip verification was introduced in [#130](https://github.com/headlamp-k8s/headlamp/pull/130), it applies to all clusters, not just in-cluster. This update ensures that TLS settings are applied specifically for in-cluster OIDC configuration.

@illume, should we deprecate the existing `insecure` flag in favor of `oidc-skip-tls-verify`, or keep it for backward compatibility?

yolossn:
@gecube @mlbiam Self signed CA support is added. Can you check if this fix works for you?

https://github.com/headlamp-k8s/headlamp/issues/127#issuecomment-2045380698
https://github.com/headlamp-k8s/headlamp/issues/2207

illume:
> keep it for backward compatibility

We definitely need to keep it for backwards compatibility.

backend/pkg/config/config.go
```diff
@@ -141,6 +144,22 @@ func Parse(args []string) (*Config, error) {
...
+	}
+
+	if config.OidcCAFile != "" {
+		caFile, err := os.ReadFile(config.OidcCAFile)
+		if err != nil {
+			return nil, fmt.Errorf("error reading oidc-ca-file: %w", err)
+		}
+		// check if the file is a valid PEM file
+		caCertPool := x509.NewCertPool()
+		if !caCertPool.AppendCertsFromPEM(caFile) {
```
illume, line 158: Can you please rename the caFile to caFileContents? Currently it shares the same convention as the config.OidcCAFile... which lead me to believe from just reading this part that it was a filename.

It might also be worth renaming config.OidcCAFile to config.OidcCAFilePath to be more explicit.

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -48,10 +48,12 @@ type Context struct {
...
-	ClientID     string
-	ClientSecret string
-	IdpIssuerURL string
-	Scopes       []string
+	ClientID      string
+	ClientSecret  string
+	IdpIssuerURL  string
+	Scopes        []string
+	SkipTLSVerify bool
+	CACert        string
```
illume, line 56: Can you please document this field?

backend/cmd/headlamp.go
```diff
@@ -545,6 +548,32 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+
+		if oidcAuthConfig.SkipTLSVerify {
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
```
illume, line 554: Please leave a justification in a comment for why gosec was disabled here?

backend/cmd/headlamp.go
```diff
@@ -545,6 +548,32 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+			if !caCertPool.AppendCertsFromPEM([]byte(oidcAuthConfig.CACert)) {
+				logger.Log(logger.LevelError, map[string]string{"oidc-ca-file": oidcAuthConfig.CACert},
+					errors.New("failed to append oidc-ca-file to cert pool"), "failed to append oidc-ca-file to cert pool")
+				http.Error(w, "invalid oidc-ca-file", http.StatusBadRequest)
+				return
+			}
+
+			secureTLSClient := &http.Client{
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{RootCAs: caCertPool}, //nolint:gosec
```
illume, line 571: Please leave a justification in a comment for why gosec was disabled here?

backend/cmd/server.go
```diff
@@ -48,12 +48,25 @@ func main() {
...
+	}
+
+	if conf.OidcCAFile != "" {
+		caFile, err := os.ReadFile(conf.OidcCAFile)
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "reading oidc-ca-file")
+			os.Exit(1)
+		}
+
+		headlampConfig.oidcCACert = string(caFile)
```
illume, line 68: I wonder if there is a string - []byte issue here? Why not leave it as []byte?

The functions that use it use []byte, and I see at least one case where it's not being cast back to []byte?
> func (s *[CertPool](https://pkg.go.dev/crypto/x509#CertPool)) AppendCertsFromPEM(pemCerts [][byte](https://pkg.go.dev/builtin#byte)) (ok [bool](https://pkg.go.dev/builtin#bool))

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -243,11 +245,25 @@ func (c *Context) OidcConfig() (*OidcConfig, error) {
 		return nil, errors.New("authProvider is nil")
 	}
 
+	var caCert string
+
+	caFilePath, ok := c.AuthInfo.AuthProvider.Config["idp-certificate-authority"]
```
illume, line 250: Can you please explain what this is in a comment?

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -243,11 +245,25 @@ func (c *Context) OidcConfig() (*OidcConfig, error) {
...
+	if ok {
+		caFile, err := os.ReadFile(caFilePath)
```
illume, line 252: I wonder where caFilePath comes from and if the path needs to be validated?

illume (changes requested):
Thanks for this.

I left some notes inline.

- Can you please add tests and a description of how to test manually in the PR?
- There will need to be support added to the helm chart.
- docs in oidc.md ?

feym78:
Hi! 
@yolossn Any news on this?
For now I tried to use insecure option from https://github.com/kubernetes-sigs/headlamp/pull/130
It worked for in cluster installation v0.34.0, but looks like it does not disable verification in refresh token process.
I can normally login into headlamp using keycloak, but after token expiration I got logged out and in headlamp's logs i see error:
error: failed to refresh token 
getting provider: Get "https://somehost/realms/somerealm/.well-known/openid-configuration": tls: failed to verify certificate: x509: certificate signed by unknown authority

Is there any workaround for this until this PR will be released?

feym78:
> Is there any workaround for this until this PR will be released?

Answering myself.
Thanks to this comment https://github.com/kubernetes-sigs/headlamp/issues/127#issuecomment-2046838571 worked passing ca cert in GO env variable using these settings in values:
```
env:
  - name: SSL_CERT_FILE
    value: "/etc/ssl/certs/custom_selfsign_CA.crt"

volumeMounts:
  - name: ca-cert-volume
    mountPath: /etc/ssl/certs/custom_selfsign_CA.crt
    subPath: custom_selfsign_CA.crt
    readOnly: true

volumes:
  - name: ca-cert-volume
    secret:
      secretName: custom-ca-secret
```

No need to use insecure setting with this.

yolossn:
>  It worked for in cluster installation v0.34.0, but looks like it does not disable verification in refresh token process.

Thanks for things finding. Ill include this case and update the PR.

>  No need to use insecure setting with this

Good to know that this fix solved the problem for you. Ill address the comments and get this merged soon.

yolossn:
closing in favor of https://github.com/kubernetes-sigs/headlamp/pull/3826

PR #2882, opened 2025-02-10
app: runCmd: Extract handleRunCommand functionality from main.ts

illume:
Move functionality for this feature into its own module.


sniok (approved):
nice, thanks!

PR #2888, opened 2025-02-11
headlamp-plugin: Add fuse.js dependency

illume:
To match frontend dependencies, and so
`npm run check-dependencies` passes again.


sniok (approved):
nice, thanks

PR #2893, opened 2025-02-12
app: fix headless mode not running

### LINKED ISSUE #2892: Headless mode not running
## Describe the bug
Whenever I try to open headlamp in headless mode it prints the arguments passed to the backend server and then shows an exception. It doesn't start listening on any port so it is rather useless as headless.

## To Reproduce
Steps to reproduce the bug:
1. Run `headlamp --headless`

 - Installation type: Both Arch Linux AUR and AppImage (running on WSL2)
 - Headlamp Version: 0.28.1

Yes

## Additional Context
Here is the thrown exception
```
arguments passed to backend server [
  '--proxy-urls',
  'https://artifacthub.io/*',
  '-html-static-dir',
  '/usr/share/headlamp/resources/frontend'
]
A JavaScript error occurred in the main process
Uncaught Exception:
TypeError: serverProcess.on is not a function
    at attachServerEventHandlers (/usr/share/headlamp/resources/app/electron/main.js:1203:17)
    at Object.<anonymous> (/usr/share/headlamp/resources/app/electron/main.js:1231:3)
    at Module._compile (node:internal/modules/cjs/loader:1373:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1432:10)
    at Module.load (node:internal/modules/cjs/loader:1215:32)
    at Module._load (node:internal/modules/cjs/loader:1031:12)
    at c._load (node:electron/js2c/node_init:2:17025)
    at node:electron/js2c/browser_init:2:126369
    at node:electron/js2c/browser_init:2:126578
    at node:electron/js2c/browser_init:2:126582
```

**Edit**:  after fixing the above exception in my fork of the repo, I got a different error:
```
server process stderr: {"level":"error","source":"/tmp/headlamp/backend/cmd/headlamp.go","line":176,"error":"open /tmp/.mount_HeadlaSQk2Gk/resources/frontend/index.baseUrl.html: read-only file system","time":"2025-02-11T23:33:42Z","message":"writing file"}
```

DMaxter:
When I tried running both the version of headlamp in the Arch AUR (not official) and the AppImage in this repo I would get an error about a function not existing. This complete makes headlamp useless for my usecase in WSL2 (in one of the computers I tested).

This PR address these issues and performs the following changes:
## 1. Correctly uses promise when on headless mode

In https://github.com/headlamp-k8s/headlamp/blob/v0.28.1/app/electron/main.ts#L1413 , we are not awaiting the Promise returned on the previous line, so it caused the exception I first mentioned in the issue. It happened because `serverProcess` was `null` when reaching the `attachServerEventHandlers` function. Solved by chaining a `then` at the end of it

## 2. Create a temporary directory for the static assets

When using on Arch, after fixing the first exception, I got a permission denied with the AUR package. With the AppImage, it said the system was readonly. It tries to patch the baseURL in place, which is not possible for squashFS on AppImages.
The fix I provided creates a temporary folder in the system temp location, copies the assets there and uses that temporary directory as the new static assets directory, that is allowed to be written by the process

## 3. Add a timeout to starting the browser

I felt the need to add a timer to starting the frontend on the browser as it always opened before the backend was serving requests. This is just a simple timer before trying to open the browser with the link

----

Feel free to scrutinize my code as I am not a Go developer and also not very familiar with Electron

Fixes #2892

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -930,6 +930,23 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
 }
 
 func StartHeadlampServer(config *HeadlampConfig) {
+	// Copy static files as squashFS is read-only (AppImage)
+	if config.staticDir != "" {
+		dir, err := os.MkdirTemp(os.TempDir(), ".headlamp")
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "Failed to create static dir")
+			os.Exit(1)
+		}
+
+		err = os.CopyFS(dir, os.DirFS(config.staticDir))
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "Failed to copy files from static dir")
+			os.Exit(1)
+		}
+
+		config.staticDir = dir
+	}
+
 	handler := createHeadlampHandler(config)
 
 	handler = config.OIDCTokenRefreshMiddleware(handler)
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1398,9 +1398,14 @@ function attachServerEventHandlers(serverProcess: ChildProcessWithoutNullStreams
 }
 
 if (isHeadlessMode) {
-  serverProcess = startServer(['-html-static-dir', path.join(process.resourcesPath, './frontend')]);
-  attachServerEventHandlers(serverProcess);
-  shell.openExternal(`http://localhost:${defaultPort}`);
+  startServer(['-html-static-dir', path.join(process.resourcesPath, './frontend')]).then(
+    serverProcess => {
+      attachServerEventHandlers(serverProcess);
+
+      // Give 1s for backend to start
+      setTimeout(() => shell.openExternal(`http://localhost:${defaultPort}`), 1000);
+    }
+  );
 } else {
   startElecron();
 }


illume:
Hello.

Thanks for reporting the issue, and for the PR!

The backend job is failing with a lint issue. You can run the linter locally for the backend with:
`make backend-lint`

I noticed this error in the App CI build jobs: 
`Error: cmd\headlamp.go:911:12: undefined: os.CopyFS`

DMaxter:
I ran that command locally and it worked. What I think might be happening is that I used a more recent version of go than what you are using. The function `os.CopyFS` was introduced in go 1.23 (https://pkg.go.dev/os#CopyFS) and the CI is using 1.22 (https://github.com/headlamp-k8s/headlamp/blob/1d9c8326cca8baf5739c3b4e58c1b6fabc720949/.github/workflows/app-artifacts-linux.yml#L30)

How should we proceed? Should the go version be updated or does this CopyFS function need to be implemented other way?

illume:
@DMaxter: ah ok, thanks.

Yeah, I've been meaning to update Headlamp to 1.23.x so I'll do that in a separate PR.

DMaxter:
Alright. Let me know if there is something I can support with

illume:
PR for new golang (still needs to be reviewed/merged)
- https://github.com/headlamp-k8s/headlamp/pull/2910

illume:
@DMaxter that golang update PR is merged now.

backend/cmd/headlamp.go
```diff
@@ -897,6 +898,25 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
...
 func StartHeadlampServer(config *HeadlampConfig) {
+	// Copy static files as squashFS is read-only (AppImage)
+	if config.staticDir != "" {
+		dir, err := ioutil.TempDir(os.TempDir(), ".headlamp")
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "Failed to create static dir")
+			os.Exit(1)
+		}
+
+		defer os.RemoveAll(dir)
```
illume, line 909: I don't understand this part.

Why is dir being removed when later it is being assigned to config.staticDir? (line 917)

DMaxter: From the very limited knowledge I have of Go, defer seems to be an instruction to make whatever is in front to be called after the function returns.

So, what happens here is:
1. Create temporary folder
2. **Schedule** the temporary folder to be removed
3. Copy the contents to it
4. Serve contents of temporary folder

When the function returns, it should call this function

DMaxter: According to the error on CI seems like this code never runs, so better remove it

DMaxter:
I hope it is good now ;)

illume:
Looking good. It's almost ready to merge IMHO, just one little thing...

Would you mind rebasing it into 2-3 commits? 

We use this linux kernel style of commits with the module as context in front like this, and with a capitalized verb at the start of the subject line:

For example:
```
app: Fix headless mode

We are not awaiting the Promise returned on the previous line, so it caused the exception I first mentioned in the issue. It happened because serverProcess was null when reaching the attachServerEventHandlers function. Solved by chaining a then at the end of it.
```

DMaxter:
Alright, let me know if something seems off

illume (approved):
Great, thanks for that.

🎉

PR #2907, opened 2025-02-15
frontend: global searchbar: Fix multiple clusterSettings items

### LINKED ISSUE #2902: Multiple 'cluster settings' in search bar
## Describe the bug
The search results do not have any way to distinguish which cluster the settings are for. We can either only display it once (this is probably more appropriate given the context in-cluster), or distinguish them (with the cluster name)

## To Reproduce
Steps to reproduce the bug:
1. Open a cluster in Headlamp (with multiple running clusters)
2. Type "cluster settings" in the search bar
3. Note how there are multiple "cluster settings" results

Image

Faakhir30:
Fixes #2902 

Fixed by only displaying the `cluster settings` page once.

### DIFF
--- a/frontend/src/components/globalSearch/GlobalSearchContent.tsx
+++ b/frontend/src/components/globalSearch/GlobalSearchContent.tsx
@@ -199,9 +199,12 @@ export function GlobalSearchContent({
     () =>
       filteredRoutes
         .filter(([, route]) => route.name && !route.path.includes(':'))
-        .filter(([, route]) => {
+        .filter(([key, route]) => {
           const clusterRoute = route.useClusterURL ?? true;
-
+          // settingsCluster is an old route that is just a redirect and shouldn't be included in the search results
+          if (key === 'settingsCluster') {
+            return false;
+          }
           return clusterRoute ? selectedClusters.length > 0 : true;
         })
         .map(([name, route]) => ({


skoeva (changes requested):
Pretty sure the cluster settings items are ordered alphabetically by context name, maybe keep the option related to the open context rather than just the first in the list? That probably makes the most sense here

sniok (changes requested):
I think there's a little confusion about why this is happnening. 
The duplicated items are not duplicated per cluster. 
The reason there are two search items called "Cluster Settings" is because we have two routes with the label "Cluster Settings". Check out router.tsx:781

There's one "Cluster Settings" with path `/settings`
And there's another "Cluster Settings" with path `/settings/cluster`

The first one is an old one and just redirects to the second one. 
We should only show the second  "Cluster Settings" (`settingsClusterHomeContext` ) route in the search results. 

I'd suggest filtering it out by the key `settingsCluster`, the same way it's done in router.tsx:977

Faakhir30:
Thanks for the reviews,
> I think there's a little confusion about why this is happening.
> The duplicated items are not duplicated per cluster.

Yup verified by creating 3 clusters, it was only showing 2 `cluster settings` items, cluster names are being passed as params, so not considered different routes in `defaultRoutes`. 

>There's one "Cluster Settings" with path /settings
>And there's another "Cluster Settings" with path /settings/cluster
>The first one is an old one and just redirects to the second one.

Regarding the old path, yes I've filtered that as suggested, and now it is showing only 1 cluster setting item referencing key `settingsClusterHomeContext`.

skoeva (approved):
My apologies for the confusion, LGTM

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -199,9 +199,11 @@ export function GlobalSearchContent({
     () =>
       filteredRoutes
         .filter(([, route]) => route.name && !route.path.includes(':'))
-        .filter(([, route]) => {
+        .filter(([key, route]) => {
           const clusterRoute = route.useClusterURL ?? true;
-
+          if (key === 'settingsCluster') {
```
sniok, line 204: please add a little comment about this, cause it's not very clear why it's being filtered out 
something like `// settingsCluster is an old route that is just a redirect and shouldn't be included in the search results`

sniok (approved):
looks good, thanks!

joaquimrocha:
Thanks @Faakhir30 !

PR #2913, opened 2025-02-17
frontend: Update frontend dependencies

sniok:
Updates 

- Vite 6
- Vitest 3
- Storybook 8.5
- Updates caniuse db
- i18next-parser (to avoid duplicates esbuild deps)

Removes

- @axe-core/react & axe-core - MPL License


joaquimrocha (commented):
Needs rebase.

illume (changes requested):
Unsure of major version updates close to a release.

Would prefer if this is merged in near the start of a release cycle, unless there's a very very good reason.

illume (approved):
👍 thanks!

PR #2915, opened 2025-02-17
backend: Do not include clusters with errors when loading them

joaquimrocha:
We were loading clusters and checking if they had errors, but we'd still include them in the final list.

This PR also changes the tests since it seeems like some parts of the code were expecting the list of contexts to be always returned regardless of whether they had invalid contexts or not (this is how the tests were written), while other parts were assuming only [valid contexts](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci9hNTgzYjliNTNhOWU3MTY0NTM2MDMwNGQzNjYyNzhhNjEwMTg0Y2ZlL2YvYmFja2VuZC9wa2cva3ViZWNvbmZpZy9rdWJlY29uZmlnLmdvP3VybD1odHRwcyUzQSUyRiUyRmdpdGh1Yi5jb20lMkZoZWFkbGFtcC1rOHMlMkZoZWFkbGFtcCZsaW5lcz0zMzk%3D?origin=gitlens) were [returned](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci9hNTgzYjliNTNhOWU3MTY0NTM2MDMwNGQzNjYyNzhhNjEwMTg0Y2ZlL2YvYmFja2VuZC9wa2cva3ViZWNvbmZpZy9rdWJlY29uZmlnLmdvP3VybD1odHRwcyUzQSUyRiUyRmdpdGh1Yi5jb20lMkZoZWFkbGFtcC1rOHMlMkZoZWFkbGFtcCZsaW5lcz0zNDg%3D?origin=gitlens).

Hopefully this fixes the mysterious crashes we have reported. I will add tests if I can.

Thanks to [Per0x1de-1337](https://github.com/Per0x1de-1337) for the help in debugging this issue.

## How to test:
1. Add a new context to your kube config with a name, an invalid server, and an invalid user.
2. Verify that Headlamp crashes before this patch
3. Verify that Headlamp does not crash after this patch

Test also adding clusters dynamically (you can use the dynamic cluster plugin), both valid and invalid ones. Make sure the valid ones work, make sure the invalid ones show an error.
Test the same but adding the clusters in desktop mode from the Add Cluster button.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1160,6 +1160,13 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		// This should not happen, but it's a defensive check.
+		if context.KubeContext == nil {
+			logger.Log(logger.LevelError, map[string]string{"context": context.Name},
+				errors.New("context.KubeContext is nil"), "error adding context")
+			continue
+		}
+
 		clusters = append(clusters, Cluster{
 			Name:     context.Name,
 			Server:   context.Cluster.Server,
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -361,6 +361,29 @@ func TestDynamicClustersKubeConfig(t *testing.T) {
 	assert.Equal(t, "default", minikubeCluster.Metadata["namespace"])
 }
 
+func TestInvalidKubeConfig(t *testing.T) {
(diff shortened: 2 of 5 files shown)


illume (approved):
👍 nice work. Thanks!
(also thanks @Per0x1de-1337)

Works well for me.

PR #2916, opened 2025-02-17
backend: Do not include clusters with errors when loading them

Per0x1de-1337:
We were loading clusters and checking if they had errors, but we'd still include them in the final list.

Hopefully this fixes the mysterious crashes we have reported. I will add tests if I can.

Solved with @joaquimrocha


joaquimrocha:
I have a PR opened for this.

Issue #2911, opened 2025-02-17
Enable strictNullChecks for frontend and plugins


sniok:
We should enable typescript `strictNullChecks` option to reduce number of issues that can be avoidable. Especially since we have a pretty good type coverage 

Example bug that could have been avoided with strict null checks https://github.com/headlamp-k8s/plugins/pull/151

This would also require us to fix all the existing issues surfaced by this option, so I expect a pretty big refactor to enable this
For plugins this will have to be opt-in

DhairyaMajmudar:
This fix would be helpful in reducing the bugs. Would love to help in this refactor 🤝🏻

Jineshbansal:
@sniok To resolve this issue, do I just need to add "strictNullChecks": true in the tsconfig.json file in the frontend directory and then fix any issues that come up during npm run build? Am I correct?

sniok:
@Jineshbansal yeah, basically
enable strictNullChecks, then run `npm run tsc`. I'm expecting a lot of errors there so to keep it manageable we should fix only some of them at a time (per pull request) and then turn strictNullChecks off. Once all of them are fixed we can enable strictNullChecks for real to prevent any new issues from being added.

DhairyaMajmudar:
@sniok When I ran the enable the `strictNullChecks` option in /frontend I got below errors 

![Image](https://github.com/user-attachments/assets/441e7440-ee3b-4546-a08f-3868b431e3cd)

 

However, these errors do not seem to be related to null checks because when I disable the `strictNullChecks` then also such errors appear. I tested null checks by running `npm run tsc` command.

Also although these errors are not related to `strictNullChecks` but they are matter to be worked on and resolved.

sniok:
make sure to run `npm install`, if that doesn't help try removing node_modules and running npm install again

DhairyaMajmudar:
> make sure to run `npm install`, if that doesn't help try removing node_modules and running npm install again

It worked, Thank you so much!

![Image](https://github.com/user-attachments/assets/f9301fe4-e4a8-4d48-80c0-589e21bea72a)

PR #2919, opened 2025-02-18
plugins: tsconfig: add strictNullChecks

DhairyaMajmudar:
## Description

1. Adding `"strictNullChecks": true` in typescript config file for plugins
2. After adding so there were some build failures which are now fixed.

Fixes: #2911

## Testing

1. Run `npm run build`: Package is building fine w/o any warnings or errors

### DIFF
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.test.tsx
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.test.tsx
@@ -1,6 +1,5 @@
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { renderHook } from '@testing-library/react';
-import { describe, expect, it, vi } from 'vitest';
 import {
   kubeObjectListQuery,
   ListResponse,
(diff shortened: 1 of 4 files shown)


sniok (changes requested):
thanks for taking a look into this!

plugins/headlamp-plugin project doesn't really have it's own typescript files, it only copies some from `frontend/` project so setting strictNullChecks isn't very beneficial

I'd recommend starting from individual plugins in https://github.com/headlamp-k8s/plugins repo

or if you want to try working on the `frontend/` part, you could try enabling `strictNullChecks` fixing some errors and then disabling it, so it's easier to do bit by bit and not all at once

DhairyaMajmudar:
> thanks for taking a look into this!
> 
> plugins/headlamp-plugin project doesn't really have it's own typescript files, it only copies some from `frontend/` project so setting strictNullChecks isn't very beneficial
> 
> I'd recommend starting from individual plugins in https://github.com/headlamp-k8s/plugins repo
> 
> or if you want to try working on the `frontend/` part, you could try enabling `strictNullChecks` fixing some errors and then disabling it, so it's easier to do bit by bit and not all at once

Thank you for letting me know, let me first start with the `frontend/` parts by parts.

DhairyaMajmudar:
I've added the `strictNullChecks` to /frontend and ran  the script `npm run tsc` 

image


Things are working fine 👍🏻

sniok:
sorry for the confusion

DhairyaMajmudar:
> sorry for the confusion

No worries 🤝🏻

PR #2929, opened 2025-02-19
frontend: headlamp-plugin: Update dependencies

skoeva:
These changes update dependencies in frontend and headlamp-plugin to the desired versions, addressing vulnerabilities.


illume (approved):
Thanks :)

PR #2930, opened 2025-02-19
frontend: Make sidebar display the cluster name again

joaquimrocha:
The sidebar is not being updated when moving clusters. This is due to
the sidebar being memoized without the cluster as a dependency.

## How to test
1. Go to a Headlamp home where you have multiple clusters
2. Click on one of the clusters -> Realize that the cluster name doesn't show up under the Cluster sidebar item
3. Change clusters using the top button and verify that the cluster name is not updated.


frontend/src/components/Sidebar/useSidebarItems.ts
```diff
@@ -12,6 +13,7 @@ export const useSidebarItems = (sidebarName: string = DefaultSidebars.IN_CLUSTER
   const customSidebarEntries = useTypedSelector(state => state.sidebar.entries);
   const customSidebarFilters = useTypedSelector(state => state.sidebar.filters);
   const shouldShowHomeItem = helpers.isElectron() || Object.keys(clusters).length !== 1;
+  const cluster = useCluster();
```
sniok, line 16: then on line 72 it should also use the cluster from this hook

```diff
-subtitle: getCluster() || undefined,
+subtitle: cluster || undefined,
```

sniok (approved):

skoeva (approved):

PR #2935, opened 2025-02-21
frontend: Fix incorrect links in sidebar after switching clusters

sniok:
How to reproduce:

 - Go to Security, click in each sub-item of the sidebar under it
 - Switch to another cluster using the top bar button
 - Go to Secrets and click the sub-items again -> the URL will be the one from the previous cluster

This was caused by the stale `cluster` value that wasn't being updated. Using `useCluster` hook fixes it

Also updated useCluster hook so it doesn't trigger a rerender on every navigation

### DIFF
--- a/frontend/src/components/App/TopBar.tsx
+++ b/frontend/src/components/App/TopBar.tsx
@@ -13,7 +13,7 @@ import { has } from 'lodash';
 import React, { memo, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
-import { useHistory } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import helpers from '../../helpers';
 import { getToken, setToken } from '../../lib/auth';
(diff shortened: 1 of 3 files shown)


frontend/src/components/App/TopBar.tsx
```diff
@@ -71,6 +71,10 @@ export default function TopBar({}: TopBarProps) {
   const history = useHistory();
   const { appBarActions, appBarActionsProcessors } = useAppBarActionsProcessed();
 
+  // getToken may return stale value so we need to rerender on navigation
+  // TODO: create a useToken hook that always returns latest token
+  // eslint-disable-next-line no-unused-vars
+  const _location = useLocation();
```
joaquimrocha, line 77: Why not make the hasToken a const set by a useMemo, with the cluster being a dependency? Would that fix this?

joaquimrocha: Got it. Thanks for elaborating.

joaquimrocha:
Meh, clicked Update by mistake. I will rebase and push as I don't want that merge branch in like that.

PR #2938, opened 2025-02-21
backend: frontend: Pass authentication token to websocket multiplexer

sniok:


backend/cmd/multiplexer.go
```diff
@@ -339,6 +350,7 @@ func (m *Multiplexer) dialWebSocket(wsURL string, tlsConfig *tls.Config, host st
 	)
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "dialing WebSocket")
+		logger.Log(logger.LevelError, nil, resp, "WebSocker response")
```
joaquimrocha, line 353: When I was a kid, I always wanted this: https://en.wikipedia.org/wiki/Super_Soaker

backend/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/headlamp-k8s/headlamp/backend
 
-go 1.24
+go 1.24.0
```
joaquimrocha, line 3: hmm, why?

joaquimrocha (approved):
LGTM. Maybe @yolossn could take a look as well. I think that since the connections are kept in relation with the userID, and this is unique per client, there's no chance of a client eventually getting connected to another's client "token powered" ws connection, right?

backend/go.mod
sniok: without the minor version it fails to build/work in vscode

```
go: downloading go1.24 (linux/amd64)
go: download go1.24 for linux/amd64: toolchain not available
```

backend/cmd/multiplexer.go
sniok: 🕸️ ⚽

frontend/src/lib/k8s/api/v2/webSocket.ts
```diff
@@ -210,6 +213,7 @@ export const WebSocketManager = {
       query,
       userId: userId || '',
       type: 'REQUEST',
+      token: getToken(clusterId),
```
illume, line 216: I think if there is no token the token field should not be added to the message?
Because the check in goland compares token to nil. 


Something like this (untested):
```ts
const token = getToken(clusterId)
if (token) {
  requestMsg.token = token;
}
```

(Maybe runs fine as is... but this might be a bit better to not include the token field if there's not one for debugging too.)

illume (changes requested):
Left a comment about only adding the token field when there's a token.

The PR is missing a description.

frontend/src/lib/k8s/api/v2/webSocket.ts
sniok: wouldn't go parse undefined or null as nil?

illume: JSON.stringify removes fields with undefined values.

So it’s fine as is.

illume (approved):
👍 thanks

illume:
> I think that since the connections are kept in relation with the userID, and this is unique per client, there's no chance of a client eventually getting connected to another's client "token powered" ws connection, right?

It’s fine yes. This uses the secret userID. Even if someone guessed it they could not get the token because it’s sent through a websocket message. If they did guess it however, then they could run commands as that user (but the chance of guessing is very low).

illume:
By the way, I also checked it was not logging the token anywhere in an error message on the backend side.

illume:
Multiplexer enabled again in this PR:
- #2962

backend/cmd/multiplexer.go
```diff
@@ -393,6 +413,7 @@ func (m *Multiplexer) reconnect(conn *Connection) (*Connection, error) {
 		conn.Path,
 		conn.Query,
 		conn.Client,
+		conn.Token,
```
yolossn, line 416: There is a case that the token can be expired when reconnecting

Issue #2894, opened 2025-02-12
Add flag to disable plugins watch


adriangonz:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
[A clear and concise description of what the problem or need is. E.g. I need to...]

We don't run in Headlamp in cluster, but we do build a Headlamp image with a few pre-configured Kubeconfigs which we then host and expose to give access to several clusters. The plugins are baked into the Headlamp image and are not expected to change at run-time.

However, if we add anything to the `/plugins` folder (instead of the `/static-plugins` folder), Headlamp's frontend will constantly keep refreshing (every ~5 secs).

There seems to be a "watch" running over the files in the `/plugins` folder which gets disabled when Headlamp is running in-cluster:

https://github.com/headlamp-k8s/headlamp/blob/979d187d389f1c936135bb05708d48ff1ae19081/backend/cmd/headlamp.go#L356-L363

It seems like the same would be useful on other contexts when Headlamp is not necessarily running in-cluster. Would it be possible to add a new flag to control whether the plugins watch is disabled?

## Describe the solution you'd like
[A clear and concise description of what you want to happen.]

A new env var or CLI flag to disable the plugins watch.

## What users will benefit from this feature?
[What type of users may be interested in this feature? What kind of Headlamp setup is likely used. (e.g. WSL users / Helm chart users / In-Cluster users / Plugin developers)]

Users neither running Headlamp in-cluster, nor locally within a Desktop app. 

## Are you able to implement this feature?
[Yes (I will propose a PR) / No.]
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

If we agree on an approach, we could help out. 

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

skoeva:
I can look into this since I worked on the plugins CLI flags a little while ago

illume:
This was done by @skoeva and is now in main branch, and will be available in the next release.

PR #2939, opened 2025-02-21
app: backend: Add flag for watching plugins changes

skoeva:
This change creates a CLI flag in the backend + desktop for watching changes to the plugins or their directory.

Fixes: #2894

#### Backend
- [X] Run `make backend`
- [X] Run `cd backend && ./headlamp-server -watch-plugins-changes false`

#### Desktop
- [X] Run `cd app && npm run build`
- [X] Run `cd dist/linux-unpacked && ./headlamp -watch-plugins-changes  false` and ensure that the flag is passed in

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -83,6 +83,10 @@ const args = yargs(hideBin(process.argv))
       describe: 'Disable use of GPU. For people who may have buggy graphics drivers',
       type: 'boolean',
     },
+    'watch-plugins-changes': {
+      describe: 'Reloads plugins when there are changes to them or their directory',
+      type: 'boolean',
+    },
   })
   .positional('kubeconfig', {
     describe:
@@ -550,6 +554,9 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
   const proxyUrls = !!buildManifest && buildManifest['proxy-urls'];
   if (!!proxyUrls && proxyUrls.length > 0) {
     serverArgs = serverArgs.concat(['--proxy-urls', proxyUrls.join(',')]);
+  }
+  if (args.watchPluginsChanges !== undefined) {
+    serverArgs.push(`--watch-plugins-changes=${args.watchPluginsChanges}`);
   }
 
   const bundledPlugins = path.join(process.resourcesPath, '.plugins');
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -47,6 +47,7 @@ type HeadlampConfig struct {
 	insecure              bool
 	enableHelm            bool
 	enableDynamicClusters bool
+	watchPluginsChanges   bool
 	port                  uint
 	kubeConfigPath        string
 	staticDir             string
@@ -355,7 +356,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	plugins.PopulatePluginsCache(config.staticPluginDir, config.pluginDir, config.cache)
 
-	if !config.useInCluster {
+	if !config.useInCluster || config.watchPluginsChanges {
 		// in-cluster mode is unlikely to want reloading plugins.
 		pluginEventChan := make(chan string)
 		go plugins.Watch(config.pluginDir, pluginEventChan)
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -26,6 +26,7 @@ type Config struct {
 	EnableHelm            bool   `koanf:"enable-helm"`
 	EnableDynamicClusters bool   `koanf:"enable-dynamic-clusters"`
 	ListenAddr            string `koanf:"listen-addr"`
+	WatchPluginsChanges   bool   `koanf:"watch-plugins-changes"`
 	Port                  uint   `koanf:"port"`
 	KubeConfigPath        string `koanf:"kubeconfig"`
 	StaticDir             string `koanf:"html-static-dir"`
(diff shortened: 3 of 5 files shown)


backend/pkg/config/config.go
```diff
@@ -153,6 +154,7 @@ func flagset() *flag.FlagSet {
 	f.Bool("dev", false, "Allow connections from other origins")
 	f.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	f.Bool("enable-dynamic-clusters", false, "Enable dynamic clusters, which stores stateless clusters in the frontend.")
+	f.Bool("disable-plugins-watch", false, "Disable watching for changes in the plugins directory")
```
joaquimrocha, line 157: Since we already have an `enable` setting here, maybe we call this one also `enable-plugins-watch`, or even `watch-plugins-changes`, with the default being true. Also, the explanation should be more accurate; something like `Reloads plugins when there are changes to them or their directory`.

joaquimrocha (changes requested):

backend/pkg/config/config.go
skoeva: Updated

illume:
A few small changes...

- Rebased against main to fix conflict
- Changed to `-watch-plugins-changes` in PR description away from old flag
- Used -watch-plugins-changes false in the container image inside Dockerfile
- Added a usage of -watch-plugins-changes  to doc

Dockerfile
```diff
@@ -95,4 +95,4 @@ USER headlamp
 EXPOSE 4466
 
 ENV HEADLAMP_STATIC_PLUGINS_DIR=/headlamp/static-plugins
-ENTRYPOINT ["/headlamp/headlamp-server", "-html-static-dir", "/headlamp/frontend", "-plugins-dir", "/headlamp/plugins"]
+ENTRYPOINT ["/headlamp/headlamp-server", "-html-static-dir", "/headlamp/frontend", "-plugins-dir", "/headlamp/plugins", "-watch-plugins-changes", "false"]
```
illume, line 98: Added this. @joaquimrocha please review?

illume (approved):
👍 thanks

Dockerfile
illume: Hrmm… Adding this seems to stop the container tests from passing.

illume (changes requested):
The container tests are failing now that I added the watch false option into the container and I’m not sure why yet.

backend/cmd/headlamp.go
```diff
@@ -355,7 +356,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	plugins.PopulatePluginsCache(config.staticPluginDir, config.pluginDir, config.cache)
 
-	if !config.useInCluster {
+	if !config.useInCluster && config.watchPluginsChanges {
```
joaquimrocha, line 359: I think this is a problem. IIRC there was someone asking for a way to watch plugins while -in-cluster, right? In that case, this would never be possible.
So we should likely have a way to detect if watch-plugins-changes has been overridden by the user directly, and if not, then it takes the !useInCluster value.
Then in this line all we need to check if the value of watchPluginsChanges.

joaquimrocha (changes requested):
Found a possible issue, as commented in line.
Also would it make sense to have it as false by default? It would break the behavior we currently have (where it's on unless -in-cluster), but I feel like the use cases for having it on by default are less than otherwise. I am fine if we don't make it false by default on the grounds of keeping the behavior we had.

backend/cmd/headlamp.go
illume: It should be `!config.useInCluster || config.watchPluginsChanges`  I think.

I believe the request in the issue was using the container, but running it out of cluster?

Dockerfile
skoeva: Is there a good reason for keeping this here? If it's being problematic I suppose we could remove it

app/electron/main.ts
```diff
@@ -551,6 +555,9 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
   if (!!proxyUrls && proxyUrls.length > 0) {
     serverArgs = serverArgs.concat(['--proxy-urls', proxyUrls.join(',')]);
   }
+  if (args.watchPluginsChanges) {
+    serverArgs.push('--watch-plugins-changes');
```
joaquimrocha, line 559: This is not setting watch off when the user provides a false value.

backend/cmd/headlamp.go
```diff
@@ -355,7 +356,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	plugins.PopulatePluginsCache(config.staticPluginDir, config.pluginDir, config.cache)
 
-	if !config.useInCluster {
+	if !config.useInCluster || config.watchPluginsChanges {
```
joaquimrocha, line 359: We need to force the watch here if the user explicitly wants to set it as true. This should be set in the **default** value of watchPluginsChanges, which needs to be false if in-cluster, and true otherwise.

backend/pkg/config/config.go
```diff
@@ -154,6 +155,7 @@ func flagset() *flag.FlagSet {
 	f.Bool("dev", false, "Allow connections from other origins")
 	f.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	f.Bool("enable-dynamic-clusters", false, "Enable dynamic clusters, which stores stateless clusters in the frontend.")
+	f.Bool("watch-plugins-changes", true, "Reloads plugins when there are changes to them or their directory")
```
joaquimrocha, line 158: Make it explicit here that the value will be false by default if -in-cluster is true.

joaquimrocha (changes requested):
We need to make sure that the default value is false IFF we are in-cluster + the user has not specified a value to watch-plugins-changes.

app/electron/main.ts
skoeva: updated to push all values

backend/cmd/headlamp.go
skoeva: added additional logic in config.go

illume:
A new PR needs to be open for this one. cc @skoeva

PR #2941, opened 2025-02-24
frontend: Update useSidebarItems hooks to support extending default sidebar items

sniok:
Fixes the regression to be able to handle custom entries that extend default items

for example 

```js
{
  name: "custom-entry",
  label: "Custom entry",
  parent: "network"
}
```

Should add it to the network item in "IN_CLUSTER" sidebar

Added a test case for this

### DIFF
--- a/frontend/src/components/Sidebar/useSidebarItems.ts
+++ b/frontend/src/components/Sidebar/useSidebarItems.ts
@@ -5,7 +5,17 @@ import helpers from '../../helpers';
 import { useCluster } from '../../lib/k8s';
 import { createRouteURL } from '../../lib/router';
 import { useTypedSelector } from '../../redux/reducers/reducers';
-import { DefaultSidebars, SidebarEntry, SidebarItemProps } from '.';
+import { DefaultSidebars, SidebarItemProps } from '.';
+
+/** Iterates over every entry in the list, including children */
+const forEachEntry = (items: SidebarItemProps[], cb: (item: SidebarItemProps) => void) => {
+  items.forEach(it => {
+    cb(it);
+    if (it.subList) {
+      forEachEntry(it.subList, cb);
+    }
+  });
+};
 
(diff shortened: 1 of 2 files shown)


illume:
cc @farodin91

frontend/src/components/Sidebar/useSidebarItems.ts
```diff
@@ -5,7 +5,17 @@ import helpers from '../../helpers';
...
-import { DefaultSidebars, SidebarEntry, SidebarItemProps } from '.';
+import { DefaultSidebars, SidebarItemProps } from '.';
+
+/** Iterates over every entry in the list, including children */
+const forEachEntry = (items: SidebarItemProps[], cb: (item: SidebarItemProps) => void) => {
+  items.forEach(it => {
+    cb(it);
+    if (it.subList) {
+      forEachEntry(it.subList, cb);
+    }
```
joaquimrocha, line 16: Hopefully we don't stack overflow.

joaquimrocha (commented):
I think the changes make sense. There is possibly an issue but likely one that existed before:
if an entry has a name that clashes with any other name, even if that name is in a different sidebar, they will be replaced AFAIU.
While maybe plugins should prefix their names, or we should add a prefix automatically for them, maybe we could make sense that the entryLookup is per sidebar?

frontend/src/components/Sidebar/useSidebarItems.ts
sniok: I'd be very impressed if someone constructs such a deep sidebar tree, especially since we only really support 2 levels

illume (approved):
Thanks!

I’ll leave it for @joaquimrocha to review your last comments / merge.

👍

illume:
Btw, is this use in an example plugin somewhere?

joaquimrocha:
> Btw, is this use in an example plugin somewhere?

There's a sidebar plugin. Not sure it adds corner cases covered in the PR, but we can also add those later.

PR #2943, opened 2025-02-24
Restrict port listened to

### LINKED ISSUE #2852: Headlamp-server process is exposed to network in app mode

joaquimrocha:
Allows to set what address the server listens to + sets the address to localhost when it's run from make or from the app module.
The container image is still listening to everything because I am not sure we would disrupt current deployments by changing this. Maybe adding a warning if the address is not limited could also be a good idea?

## How to test:
Not really sure unless one opens the localhost:4466 to e.g. the LAN and tries to make a request from another machine, or something more clever, maybe using containers and the networking...

For now, we should also test whether the regular listing, etc. works since a failure in this would affect that.
And verifying what the listen address reports is can be also checked from the logs.

Based on a previous PR by @adwait-godbole . Thanks!

fixes #2852

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -543,7 +543,7 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
     ? path.resolve('../backend/headlamp-server')
     : path.join(process.resourcesPath, './headlamp-server');
 
-  let serverArgs: string[] = [];
+  let serverArgs: string[] = ['--listen-addr=localhost'];
   if (!!args.kubeconfig) {
     serverArgs = serverArgs.concat(['--kubeconfig', args.kubeconfig]);
   }


illume (approved):
Works for me. Thanks.

```
$ sudo netstat -nap  | grep LISTEN | grep 4466
tcp        0      0 127.0.0.1:4466          0.0.0.0:*               LISTEN      1205938/./backend/h
```

PR #2944, opened 2025-02-24
[OLD] Fix rename cluster duplicate name issue / improve cluster metadata for UI

vyncent-t:
This branch replaces the need for https://github.com/headlamp-k8s/headlamp/pull/2747

## Description
This PR is needed as a part to fix the delete non dynamic cluster PR

- This PR aims to improve the cluster information available for clusters via the metadata tag. 
- introduces changes to the metadata field that contain the `originalName`   `origin.kubeconfig` and `pathID`  
- introduces fixes to the `source` field in the home view to have an accurate path for kubeconfigs
- introduces an original name label within cluster settings to help keep track of clusters
---
This PR also fixes an issue with duplicate name clusters.

Currently, users are able to 'reuse' cluster names, causing the existing cluster to be replaced with the renamed one. This UI redirection provides a way to prevent this from happening from within headlamp only.

- This PR introduces logic and UI to stop users from entering a name already in use when trying to rename a cluster
- Reworks the current `renameCluster` functionality in the backend to cover both non dynamic and dynamic clusters

### Cluster info

- metadata field now includes `cluster.metadata.origin.kubeconfig` and `cluster.metadata.originalName` and  `cluster.metadata.pathID`

### Home View

### Cluster Settings
- `OriginalName` now displayed in non-dynamic cluster settings

image

- Introduced UI dialog for preventing names already in use from being used. 

image
image

### Test the source path
- Run headlamp in App mode
- At the home view: notice the specific Kubeconfig path

### Test the original name label
- Run headlamp in App mode
- Create a non-dynamic cluster and let it load at home view
- Click on the settings icon in the cluster list and navigate to settings page
- Notice the original name displayed under the `Name` section

### Test the duplicate rename blocker UI
- Run headlamp in App mode
- Create a two non-dynamic clusters and let them load at home view
- Click on the settings icon in the cluster list and navigate to settings page
- Rename one of the non dynamic cluster to a name already in use by another cluster
- Notice the "Invalid custom name" UI
- Choose a name that is valid and rename it
- Navigate back to the settings page if needed and note the new custom name 
- Click on the settings tab of the renamed cluster and notice and original name displayed


frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -314,6 +314,24 @@ export default function SettingsCluster() {
     );
   }
 
+  // Display the original name of the cluster if it was loaded from a kubeconfig file.
+  function ClusterName() {
+    const originalName = clusterInfo?.meta_data?.originalName;
+    const source = clusterInfo?.meta_data?.source;
+    const isNonDynamic = source === 'kubeconfig';
```
skoeva, line 321: Only using the source as a conditional also renders the typography component when there is no originalName. I think it would be better to have something like this:
```suggestion
    // Display original name in the UI when defined
    const displayOriginalName = source === 'kubeconfig' && originalName;
```

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -314,6 +314,24 @@ export default function SettingsCluster() {
...
+
+    return (
+      <>
+        <Typography>{t('translation|Name')}</Typography>
+        {isNonDynamic && (
```
skoeva, line 326: and then this could be replaced with
```suggestion
        {displayOriginalName && (
```

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: good idea!

quick question though, were you able to get this to render without an originalName? or is this just a potential issue you mean? 

If you were able to see the typography component render without an original name can you tell me the steps to recreate it, that means the meta_data is either failing or not being completed somewhere and I'd need to know where

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: nice!

frontend/src/components/App/Settings/SettingsCluster.tsx
skoeva: It was a bug that I noticed but seems like it went away 🤔 unsure how to reproduce

I guess it's also good to make sure there's something to display before rendering it

illume:
@vyncent-t can you please add How to test to the PR description?

frontend/src/components/App/Home/index.tsx
```diff
@@ -246,7 +246,8 @@ function HomeComponent(props: HomeComponentProps) {
    */
   function getOrigin(cluster: Cluster): string {
     if (cluster.meta_data?.source === 'kubeconfig') {
-      const kubeconfigPath = process.env.KUBECONFIG ?? '~/.kube/config';
+      const sourcePath = cluster.meta_data?.origin?.kubeconfig;
+      const kubeconfigPath = process.env.KUBECONFIG ?? sourcePath;
```
joaquimrocha, line 250: I think it should always be the meta_data case here. No the KUBECONFIG env.

backend/cmd/headlamp.go
```diff
@@ -1178,6 +1178,13 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 				"source":     context.SourceStr(),
 				"namespace":  context.KubeContext.Namespace,
 				"extensions": context.KubeContext.Extensions,
+				"origin": map[string]interface{}{
+					"kubeconfig": func() string {
+						path, _ := c.getKubeConfigPath(context.SourceStr())
+						return path
+					}(),
+				},
+				"originalName": context.Name,
```
joaquimrocha, line 1187: I see in the /config requests from the frontend that the original name is the one being sent as the name above, so we already have the original name + custom name (in the extensions) in the frontend AFAIU.

joaquimrocha (changes requested):
Left a couple of comments.

backend/cmd/headlamp.go
```diff
@@ -1178,6 +1178,13 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 				"source":     context.SourceStr(),
 				"namespace":  context.KubeContext.Namespace,
 				"extensions": context.KubeContext.Extensions,
+				"origin": map[string]interface{}{
+					"kubeconfig": func() string {
+						path, _ := c.getKubeConfigPath(context.SourceStr())
+						return path
```
joaquimrocha, line 1184: I think the path should be set from the methods that read the cluster in the first place rather than trying to find it again here.

backend/cmd/headlamp.go
vyncent-t: in the frontend when I log the end result of a cluster that has been fetched, if it has been renamed then the cluster.name is set to the customName

looking at the /config response looks like the top name is the original name but after its processed and finished for the the frontend the cluster log information is different

image

so I think we do have access to the original name as long as it is never renamed

backend/cmd/headlamp.go
vyncent-t: moved it so that we aren't trying to find it right inside of where it is being initialized

frontend/src/components/App/Home/index.tsx
vyncent-t: done!

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -314,6 +314,27 @@ export default function SettingsCluster() {
...
+    const source = clusterInfo?.meta_data?.source;
+    // Note: display original name is currently only supported for kubeconfig sources.
+    const displayOriginalName = source === 'kubeconfig' && originalName;
+
+    return (
+      <>
+        <Typography>{t('translation|Name')}</Typography>
+        {displayOriginalName && (
+          <Typography variant="body2" color="textSecondary">
+            {t('translation|Original name of cluster: {{ originalName }}', {
```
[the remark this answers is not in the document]
vyncent-t: done!

backend/cmd/headlamp.go
```diff
@@ -1191,6 +1191,11 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		kubeconfigPath, err := c.getKubeConfigPath(context.SourceStr())
```
joaquimrocha, line 1194: The actual path of the kubeconfig needs to be corrected. This will always give the kubeconfig we loaded headlamp with, but if we have more than one, it's not clear this will yield the correct value. So I think that the way to add it is in the methods where we load the actual kubeconfig: we should add a reference to the file's path for each cluster.

backend/cmd/headlamp.go
joaquimrocha: I need to understand what "after its processed and finished for the the frontend " means in this case.
Seems like we likely need to always keep the original name as the key (till we also add the kubeconfig path) for both the backend and frontend, and use the renamed name just for the display.

joaquimrocha (changes requested):

backend/cmd/headlamp.go
vyncent-t: so to confirm, we find where we load the cluster information from the kube config, create a middle func that adds the file path from that kubeconfig, and attach it to the cluster information before sending it off again? 

should the same be done for the original name then? and just save the name under `clusterID` or something similar before sending it off?

vyncent-t:
updated the branch
- split the commits
- added logic to get source path from where we read kubeconfigs
- added new pathID to metadata

console log clusters in the home view now looks like

image

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -45,6 +45,8 @@ type Context struct {
 	proxy       *httputil.ReverseProxy `json:"-"`
 	Internal    bool                   `json:"internal"`
 	Error       string                 `json:"error"`
+	PathName    string                 `json:"pathName"`
+	PathID      string                 `json:"pathID"`
```
joaquimrocha, line 49: Instead of path name and path ID, which are confusing, why not kubeConfigPath and ID or ClusterID?

backend/cmd/headlamp.go
```diff
@@ -1508,9 +1508,15 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
 
-// Get path of kubeconfig from source.
+// Gets the path of the kubeconfig we load headlamp with from source.
+// Note for non dynamic clusters:
+//   - This will always give the kubeconfig we loaded headlamp with, but if more than one,
+//     it's not clear this will yield the correct value.
 func (c *HeadlampConfig) getKubeConfigPath(source string) (string, error) {
-	if source == "kubeconfig" {
+	// note: we need to have it in a const for the linting to pass
+	const sourceKubeconfig = "kubeconfig"
```
joaquimrocha, line 1517: This existed already before, right? I am surprised this change is needed.

joaquimrocha (changes requested):

backend/pkg/kubeconfig/kubeconfig.go
vyncent-t: will change these to `kubeConfigPath` and `ClusterID`

vyncent-t: done!

PR #2947, opened 2025-02-25
frontend: Fix DetailsDrawer to not mutate url when in app mode

sniok:
Fixes a bug in details drawer in app mode

How to test/reproduce:

 - Go to ConfigMaps (or any other resource)
 - Click on one resource
 - Close the details drawer
 - Click on the same resource again

The spinner will show, check the `window.location.href` it will have url path set instead of the hash router


skoeva (approved):
tested and works, thanks for this

PR #2954, opened 2025-02-26
github: Add a new workflow to attach the release assets to a release

joaquimrocha:
Couldn't test it yet because workflows need to be in main to show in the actions area first, apparently.


sniok (approved):
looks good!
maybe a suggestion for the workflow name, "push release assets" almost sounds like it's pushing it to git, maybe "upload release assets"?

PR #2955, opened 2025-02-26
i18n: Add Italian locale

M0Rf30:
As in the title,
Regards


skoeva (commented):
Hi! If/when you are done making changes, it would be great if you could use an atomic commit (squash the changes) and sign it

M0Rf30:
> Hi! If/when you are done making changes, it would be great if you could use an atomic commit (squash the changes) and sign it

done

skoeva:
Looking solid, just run one last `git rebase main` with a force push to the branch and should be good to go :)

skoeva (approved):

illume (approved):
🎉🎈 thanks

PR #2959, opened 2025-02-27
charts: make in-cluster arg configurable

### LINKED ISSUE #2958: Make `in-cluster` Configurable in Helm Chart Values
# Feature Request: Make `in-cluster` Configurable in Helm Chart Values

Currently, the `in-cluster` parameter is hardcoded in the `charts/headlamp/templates/deployment.yaml` file. This means that users cannot disable or configure this value dynamically from `values.yaml`. Instead, they must modify the Helm chart template directly, which is not ideal for maintainability and upgrades.

## Solution 
Add an option in `values.yaml` to control the `in-cluster` setting, allowing users to enable or disable `in-cluster` mode through a Helm value.

**Current Behavior:**
In `charts/headlamp/templates/deployment.yaml`, the argument is set as:
```yaml
args:
  - "-in-cluster"
```
This is always enabled, regardless of user preference.

**Proposed Change:**
Modify `deployment.yaml` to check for a new `inCluster` value in `values.yaml`:
```yaml
args:
{{- if .Values.config.inCluster }}
  - "-in-cluster"
{{- end }}
```

Then, in `charts/headlamp/values.yaml`, introduce:
```yaml
config:
  inCluster: true # Default to true for backward compatibility
```

Helm chart users who need to run Headlamp without `in-cluster` mode, particularly those with external authentication setups.

Faakhir30:
Fixes #2958 

- [x] Added suggested changes.(Read the `in-cluster` argument from `values.yml` dynamically.)
- [x] Updated docs

### DIFF
--- a/charts/headlamp/README.md
+++ b/charts/headlamp/README.md
@@ -69,6 +69,7 @@ $ helm install my-headlamp headlamp/headlamp \
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| config.inCluster | bool | `true` | Run Headlamp in-cluster |
 | config.baseURL | string | `""` | Base URL path for Headlamp UI |
 | config.pluginsDir | string | `"/headlamp/plugins"` | Directory to load Headlamp plugins from |
 | config.extraArgs | array | `[]` | Additional arguments for Headlamp server |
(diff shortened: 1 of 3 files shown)


illume (approved):
👍 cheers

PR #2960, opened 2025-02-28
e2e-tests: Fail CI job if e2e-tests fail

illume:
- **e2e-tests podsPage: Disable broken flaky test**
- **github: build-container: Make e2e job fail if tests fail**

It looks like there is a test failing when I look at the logs, but the CI job is passing.

### How to test

View the container job and look at the logs

Here's an **old failing job** (this one should not fail like this): https://github.com/headlamp-k8s/headlamp/actions/runs/13517271859/job/37768475888?pr=2861#step:18:1154
```
Running 24 tests using 1 worker
·······················×±

  1) [chromium] › podsPage.spec.ts:5:5 › multi tab create delete pod ───────────────────────────────

    Test timeout of 60000ms exceeded.

    Error: page.click: Test ended.
    Call log:
      - waiting for locator('span:has-text("Workloads")')


       at podsPage.ts:7

       5 |
       6 |   async navigateToPods() {
    >  7 |     await this.page.click('span:has-text("Workloads")');
         |                     ^
       8 |     await this.page.waitForLoadState('load');
       9 |     await this.page.waitForSelector('span:has-text("Pods")');
      10 |     await this.page.waitForLoadState('load');

        at podsPage.navigateToPods (/home/runner/work/headlamp/headlamp/e2e-tests/tests/podsPage.ts:7:21)
        at /home/runner/work/headlamp/headlamp/e2e-tests/tests/podsPage.spec.ts:34:25

  1 flaky
    [chromium] › podsPage.spec.ts:5:5 › multi tab create delete pod ────────────────────────────────
  23 passed (1.9m)

```

I think we need to add this to our run: parts. Because by default github actions don't care if one thing fails in the run script.

```yaml
- name: Run e2e tests
  run: |
    set -e
```

You can see the new job passing: https://github.com/headlamp-k8s/headlamp/actions/runs/13585222154/job/37978600041?pr=2960#step:18:1168

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -179,14 +179,7 @@ jobs:
         cd e2e-tests
         npm ci
         npx playwright install --with-deps
-        HEADLAMP_TEST_URL=$SERVICE_URL npx playwright test
-        exit_code=$?
-        if [ $exit_code -ne 0 ]; then
-          echo "Playwright tests failed with exit code $exit_code"
-          exit 1
-        else
-          echo "Playwright tests passed successfully"
-        fi
+        HEADLAMP_TEST_URL=$SERVICE_URL npx playwright test || exit 1
     - name: Save Docker Images to Tar files in image-cache Folder
       if: steps.cache-image-restore.outputs.cache-hit != 'true'
       run: |
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -143,6 +143,7 @@ jobs:
         envsubst < e2e-tests/kubernetes-headlamp-ci.yaml | kubectl --context=test apply -f -
     - name: Run e2e tests
       run: |
+        set -e
         echo "------------------------------------sleeping 12...------------------------------------"
         sleep 12
         kubectl config use-context test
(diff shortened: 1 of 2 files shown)


adwait-godbole:
I am pretty sure that if you download and check the e2e tests report for this PR, we will find the same false positive report for namespaces test.

illume:
Ok... there's still issues here:

There is a failing e2e test here, and the job still passes: https://github.com/headlamp-k8s/headlamp/actions/runs/13585494710/job/37979413059#step:18:1168

illume:
> I am pretty sure that if we download and check the e2e tests report for this PR, we will find the same false positive report for namespaces test.

The main thing I want to fix in this PR is... if there is a new e2e test failure I want the CI job to fail.

I suggest adding a todo note to this namespaces test for fixing in a later PR.

sniok:
'flaky' means that test failed but then passed after a retry, so I don't think it should fail the CI. 
but if you want to have the CI fail, you could disable the retries

https://playwright.dev/docs/test-retries
https://playwright.dev/docs/api/class-testconfig#test-config-retries

illume:
I was mistaken.

PR #2963, opened 2025-02-28
plugin/maps enable loops in graph

faebr:
This PR adapts the algorithm which creates the tree for the map. After these changes, edges which form a loop are not removed anymore. This enables users using the map plugin api to create more complex relations using the graph.

Sample plugin code to create a loop:
```
import { registerMapSource } from '@kinvolk/headlamp-plugin/lib';
import { KubeObject } from '@kinvolk/headlamp-plugin/lib/K8s/cluster';
import { useMemo } from 'react';

let source =
    {
        id: 'my-source', // ID of the source should be unique
        label: 'My Source', // label will be displayed in source picker
        useData() {
          return useMemo(() => {
            // Hardcoded kubernetes object as an example
            const myResource = {
              kind: 'MyResourceKind',
              metadata: {
                uid: '1234',
                name: 'my-test-resource-1',
                namespace: 'test-namespace',
                creationTimestamp: '1234',
              },
            };
            const myResource2 = {
              kind: 'MyResourceKind',
              metadata: {
                uid: '12345',
                name: 'my-test-resource-2',
                namespace: 'test-namespace',
                creationTimestamp: '12345',
              },
            };
            const myResource3 = {
                kind: 'MyResourceKind',
                metadata: {
                  uid: '123456',
                  name: 'my-test-resource-3',
                  namespace: 'test-namespace',
                  creationTimestamp: '123456',
                },
              };
            return {
              nodes: [
                {
                  id: myResource.metadata.uid, // ID should be unique
                  kubeObject: new KubeObject(myResource),
                },
                {
                  id: myResource2.metadata.uid, // ID should be unique
                  kubeObject: new KubeObject(myResource2),
                },
                {
                  id: myResource3.metadata.uid, // ID should be unique
                  kubeObject: new KubeObject(myResource3),
                },
              ],
              edges: [
                  {   
                      source: myResource.metadata.uid,
                      target: myResource2.metadata.uid,
                      id: 'edge2',
                  },
                  {   
                    source: myResource2.metadata.uid,
                    target: myResource3.metadata.uid,
                    id: 'edge3',
                },
                  {   
                    source: myResource.metadata.uid,
                    target: myResource3.metadata.uid,
                    id: 'edge4',
                }
              ]
            };
          }, []);
        },
      };

registerMapSource(source);
```

### DIFF
--- a/frontend/src/components/resourceMap/graph/graphGrouping.tsx
+++ b/frontend/src/components/resourceMap/graph/graphGrouping.tsx
@@ -54,31 +54,37 @@ const getConnectedComponents = (nodes: GraphNode[], edges: GraphEdge[]): GraphNo
     visitedNodes.add(node.id);
     componentNodes.push(node);
 
+    // Outgoing edges
     graphLookup.getOutgoingEdges(node.id)?.forEach(edge => {
-      if (visitedNodes.has(edge.target)) return;
-
+      // Always collect the edge if we haven't yet
       if (!visitedEdges.has(edge.id)) {
         visitedEdges.add(edge.id);
         componentEdges.push(edge);
       }
 
-      const targetNode = graphLookup.getNode(edge.target);
-      if (targetNode) {
-        findConnectedComponent(targetNode, componentNodes, componentEdges);
+      // Only recurse further if we haven't visited the target node
+      if (!visitedNodes.has(edge.target)) {
+        const targetNode = graphLookup.getNode(edge.target);
+        if (targetNode) {
+          findConnectedComponent(targetNode, componentNodes, componentEdges);
+        }
       }
     });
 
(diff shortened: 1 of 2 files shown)


sniok (changes requested):
Tested a simple case like this `a -> b -> c -> d -> back to a` and it works great, thanks for the contribution!

image

while visually it might be a bit messy now I think we can improve it later based on some real cases

Could you update the commits to follow our [guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines)?

sniok (approved):
thanks! added a little change to fix the css invalid value `fit-content(300px)` -> `width: fit-content; minWidth: 300px;`

PR #2966, opened 2025-03-01
frontend: Add e2e-tests for react-hotkeys dependency

### LINKED ISSUE #2380: Add test(s) for react-hotkeys dependency
We have a dependency react-hotkeys for which we have no tests that cover it.

This makes updating it time consuming and risky, as we need to manually test it with every update.

adwait-godbole:
Fixes #2380 
- #2380 

This will help us update the react-hotkeys dependency with an automated check helping us catch issues.

### DIFF
--- a/e2e-tests/tests/headlamp.spec.ts
+++ b/e2e-tests/tests/headlamp.spec.ts
@@ -51,6 +51,28 @@ test('main page should have Network tab', async () => {
   await headlampPage.hasNetworkTab();
 });
 
+test('main page should have global search along with react-hotkey hint text', async () => {
+  const globalSearch = await headlampPage.hasGlobalSearch();
+
+  const searchHintContainer = globalSearch.locator('xpath=following-sibling::div');
+  const pressTextExists = await searchHintContainer.getByText(/^Press/).isVisible();
+  const slashHotKeyExists = await searchHintContainer
+    .locator('div')
+    .filter({ hasText: /^\/$/ })
+    .isVisible();
+  const toSearchTextExists = await searchHintContainer.getByText(/to search$/).isVisible();
+
+  expect(pressTextExists && slashHotKeyExists && toSearchTextExists).toBeTruthy();
+});
+
+test('react-hotkey for global search', async ({ page }) => {
+  await page.keyboard.press('/');
+
+  const focusedSearch = page.getByPlaceholder(/^Search resources, pages, clusters by name$/);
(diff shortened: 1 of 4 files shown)


adwait-godbole:
I have noticed here that on the below page,

Screenshot from 2025-02-18 09-17-41

the cluster chooser hotkey i.e.

https://github.com/headlamp-k8s/headlamp/blob/25372f6e6a7e393e1e60fb72e75c462306875242/frontend/src/components/cluster/Chooser.tsx#L55

is not working when deployed with all the additional plugins.

adwait-godbole:
> I have noticed here that on the below page,
> 
> Screenshot from 2025-02-18 09-17-41
> 
> the cluster chooser hotkey that is:
> 
> https://github.com/headlamp-k8s/headlamp/blob/main/frontend/src/components/cluster/Chooser.tsx#L54-#L60
> 
> is not working when deployed with all the additional plugins but clicking on the button manually is opening the popup.

(This is not the case when headlamp is deployed without any additional plugins. Without any plugins, the hotkey is working as expected.)

skoeva (commented):
Have you looked at debugging the multi-cluster test? seems to be failing the CI

adwait-godbole:
> Have you looked at debugging the multi-cluster test? seems to be failing the CI

Hi @skoeva, the test is failing due to the same reason that I have mentioned in the [this](https://github.com/headlamp-k8s/headlamp/pull/2966#issuecomment-2692147420)

joaquimrocha:
Hi @adwait-godbole . Do you have time to rebase this? I was hoping we could take a fresher look soon.

illume:
I'm working on rebasing this one.

illume:
I had to change a couple of things, and it passes on CI now.

- One of the example plugins changes the cluster chooser, which interferes with the test. Just disabled this test for now. I do see the value of deploying all our example plugins and running the e2e tests... but it does make it complicated to write e2e tests using functionality that plugins remove or change. One main point of these tests is to check that the hotkey dependency works when we upgrade it, so disabling this one test does not lose that. (There are other e2e tests in this PR that test react hot key functionality).
- Headlamp has a regression where URLs for details views are broken. The test was originally written when detail view URLs were set when viewing a detail view. So I had to change one test to not assume headlamp worked with URLs for detail view any more. 
  - We already have and issue for URL thing: https://github.com/kubernetes-sigs/headlamp/issues/3592

Issue #2932, opened 2025-02-20
Clusters list table is visible even if all properties are hidden


DhairyaMajmudar:
## Describe the bug
I've obeserved an edge case where a user if switches oof all the properties manually then after the table frame is still visislbe. Other than this if all the field are made hidden using hide all button  the whole table get hidden. 

## To Reproduce
Steps to reproduce the bug:

- Visit headlamp home page
- hide all fields manually and then  after hide them  all using hide all button
- see the difference.

## Environment (please provide info about your environment):
 - Installation type: Linux (Debian)
 - Headlamp Version: Web
 - Other: N/A

## Are you able to fix this issue?
Yes

## Additional Context

https://github.com/user-attachments/assets/3567a392-3c25-4f33-b50e-031b9042e99c

skoeva:
I can push a fix up for this

PR #2968, opened 2025-03-03
frontend: Hide row actions with hidden columns

skoeva:
This change ensures that when all of the non-action columns are hidden, the action column is also hidden.

Fixes: #2932 

### Testing
- [X] Navigate to the Headlamp home page
- [X] Manually hide all the cluster fields in the table
- [X] Ensure that the row actions are also hidden
- [X] Manually unhide one field and ensure that the row actions appear again

### DIFF
--- a/frontend/src/components/App/Home/index.tsx
+++ b/frontend/src/components/App/Home/index.tsx
@@ -306,6 +306,7 @@ function HomeComponent(props: HomeComponentProps) {
                 getValue: ({ name }) => versions[name]?.gitVersion || '⋯',
               },
               {
+                id: 'actions',
                 label: '',
                 getValue: () => '',
                 cellProps: {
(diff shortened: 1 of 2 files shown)


sniok:
Could you also remove the sorting from the actions column? I've just noticed it
image

skoeva:
Turns out the issue was isolated to the actions column as you pointed out @sniok, but it looks like this behavior is expected [according to the MRT docs](https://www.material-react-table.com/docs/guides/row-actions). 

Still think this might be a UX improvement, so PR has been updated

sniok:
> Turns out the issue was isolated to the actions column as you pointed out @sniok, but it looks like this behavior is expected [according to the MRT docs](https://www.material-react-table.com/docs/guides/row-actions). 
> 
> Still think this might be a UX improvement, so PR has been updated

I was just referring to the sorting button on the actions column. To be honest I don't think we should worry too much about the case when all the columns are hidden because what's the point of not showing any column

skoeva:
rebasing against main

illume:
@skoeva can you please move this to a new PR?

skoeva:
reopened at #3201

PR #2982, opened 2025-03-12
frontend: Open Custom Resources and Custom Resource Definition in a drawer

sniok:
Adds support for the CR and CRD in drawer

### DIFF
--- a/frontend/src/components/common/Link.tsx
+++ b/frontend/src/components/common/Link.tsx
@@ -3,7 +3,6 @@ import { useQueryClient } from '@tanstack/react-query';
 import React from 'react';
 import { useDispatch } from 'react-redux';
 import { Link as RouterLink } from 'react-router-dom';
-import helpers from '../../helpers';
 import { kubeObjectQueryKey, useEndpoints } from '../../lib/k8s/api/v2/hooks';
 import { KubeObject } from '../../lib/k8s/KubeObject';
 import { createRouteURL, RouteURLProps } from '../../lib/router';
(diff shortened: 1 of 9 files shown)


frontend/src/components/resourceMap/details/KubeNodeDetails.tsx
```diff
@@ -99,6 +109,18 @@ export const KubeObjectDetails = memo(
         ([key]) => key.toLowerCase() === kind?.toLowerCase()
       )?.[1] ?? DetailsNotFound;
 
+    let content = <Component name={name} namespace={namespace} />;
+
+    if (customResourceDefinition) {
```
joaquimrocha, line 114: Should we set the content for the non-CRD only on else here? To avoid calling the function and then rewriting it.

frontend/src/components/common/Link.tsx
```diff
@@ -76,11 +88,15 @@ function PureLink(props: React.PropsWithChildren<LinkProps | LinkObjectProps>) {
...
-        state,
-      }}
+      to={
+        props.onClick
+          ? '#'
+          : {
+              pathname: createRouteURL(routeName, params),
+              search,
+              state,
+            }
```
joaquimrocha, line 98: I am not seeing the URL being affected by this.

joaquimrocha (changes requested):
Left a comment in one thing in the code.
Also noticed that now we no are longer changing the URL as before, and this doesn't match with what the Link seems to be updated to.

frontend/src/components/common/Link.tsx
sniok: slightly adjusted logic here, URL now updates and also the middle mouse click works (opens a full details page in a new page)

frontend/src/components/resourceMap/details/KubeNodeDetails.tsx
sniok: replaced with a ternary op
but also creating jsx element is pretty much free, it doesn't call the component function, just creates an object

sniok:
this is a side effect of us putting the drawer state in the url 
if we don't update the URL then back button will behave as you'd expect
if we put sidebar in a query param like `?d=...` then back button will close the drawer, and also navigation will close the drawer

sniok:
this PR is about custom resources, let's get this in and then discuss URL and back button behavior in a separate PR

joaquimrocha:
I prefer then to have no URL reflection than having the back button behaving like this. My original comment about the URL not reflecting it 2 reviews ago was just because the PR seemed to change the URL but I was seeing no changes.
If we have the drawer working for CRs and the back button consistent without having the reflection in the URL, I am happy to merge.

joaquimrocha (approved):

PR #2985, opened 2025-03-13
Add a way to manage external tools' binaries

joaquimrocha:
Since using external tools like minikube, kubectl, etc. is becoming more and more important, we need a way for the user to actually download/manage those external dependencies, and for plugins to request them too.

This PR adds a new settings section "External Tools" which shows minikube + kubectl for now, whether they have been installed by Headlamp or found in the system and also ways to download/remove them.

It also adds a new example plugin which runs minikube status or informs the user when it needs consent, etc.

The code needs to be thoroughly reviewed.


illume:
We need to bundle minikube, because air-gapped.

app/electron/binaries.ts
```diff
@@ -0,0 +1,812 @@
...
+export const BINARY_MANIFESTS: { [key: string]: BinaryManifest } = {
+  minikube: {
+    name: 'minikube',
+    displayName: 'Minikube',
+    version: 'v1.32.0',
+    urls: {
+      win: {
+        x64: {
+          url: 'https://github.com/kubernetes/minikube/releases/download/{version}/minikube-windows-amd64.exe',
+          checksum: '25650839f34526ee0f85fd4faa2d5b327f6127caac3de82e1f905adc3aad29df',
```
sniok, line 115: checksum failed for me, checked in the [release page](https://github.com/kubernetes/minikube/releases/tag/v1.32.0) and the checksum for this file is 
`d4060da824524df744ba85fa91394cabfab15f110c03b1d8b7f1d309116fed15`

joaquimrocha: Maybe I messed up copying the binaries, or AI got in the way 😛 
Please update as appropriate.

PR #2990, opened 2025-03-16
docs: Add guidance for probe modifications in baser-url

Chiwency:
When running Headlamp with a base URL in a Kubernetes deployment, simply adding `-base-url /headlamp` to the container's args is not enough. If the livenessProbe and readinessProbe paths remain unchanged, they will fail with a 404 error. Therefore, additional guidance is needed to update these probe paths accordingly.


skoeva (approved):
thanks for this

illume:
Hi @Chiwency we moved into the Kubernetes org a little while ago, and they require this CLA mentioned by the bot. Let me know if you are able to sign the CLA or not. No stress if not. Thanks :)

illume (changes requested):
Needs CLA.

Chiwency:
> Needs CLA.

Hi, i have signed

illume (approved):
Great. Thank you!

PR #2991, opened 2025-03-17
Install platform-specific tarballs for plugins

joaquimrocha:
These changes allow specifying (in the ArtifactHub package file) extra files to be downloaded and extracted depeding on the platform we are on.

It also adds new tests for the plugin-management.ts.

## How to test

1) Have plugin-catalog plugin installed.

2) Make sure you have your minikube binary backed up ( `sudo mv /usr/local/bin/minikube /usr/local/bin/minikube.bak` ). This PR will download minikube for you into the plugin folder. (You can install minikube again after if you like, this is to just test it works to use the downloaded minikube)

3) Start the app
```shell
git checkout extra-artifact-hub-tarballs
cd app
npm i
npm start
```

4) In app, Go to plugin catalog, and have non official plugins showing

5) install minikubeprerelease plugin from plugin catalog.

6) From the home screen see if you can start a minikube cluster

## Regression test
1. Install a plugin for the plugin catalog
2. Verify the plugin is working


app/electron/__tests__/test-data/artifacthub-pkg.yml
```diff
@@ -0,0 +1,32 @@
...
+description: A Headlamp plugin for managing Minikube clusters
+annotations:
+  headlamp/plugin/archive-url: 'https://github.com/headlamp-k8s/plugins/releases/download/example-change-logo-0.0.1/change-logo-0.0.1.tar.gz'
+  headlamp/plugin/archive-checksum: 'SHA256:8a2a23d116fe175b4350ffc6cc259781ed98d943f932e30fc5c8fe4e07b86e89'
+  headlamp/plugin/version-compat: '>=0.30'
+  headlamp/plugin/distro-compat: 'in-cluster,web,docker-desktop,desktop'
+
+  # Platform-specific archives contain the minikube binary for each platform
+  # Linux x64
+  headlamp/plugin/archive-url:linux:x64: 'https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64.tar.gz'
```
illume, line 15: "https://storage.googleapis.com/minikube" should be added to allowed archive urls?

joaquimrocha: You are right. However, I think the tests were checking this and passing, so now I am puzzled.

joaquimrocha:
> Is it right that this will only work with one extra download per plugin? I guess that's probably ok. Probably we'd have an az download in a separate plugin for example.

True. But maybe we can bundle the binaries (if we are allowed to distribute) ourselves in one single tarball for the minikube workflow at least?

app/electron/__tests__/test-data/artifacthub-pkg.yml
joaquimrocha: Let's allow-list the URLs once we have a real use case for them. Another option is to allow all of them, since we are showing them to the user...

illume: Ok, so this isn't needed after all.

From the minikube installation documentation they serve their binaries from github releases:
> https://github.com/kubernetes/minikube/releases/latest/download/minikube-windows-amd64.exe

app/electron/plugin-management.ts
```diff
@@ -446,20 +457,111 @@ async function downloadExtractArchive(
...
+    if (progressCallback) {
+      progressCallback({
+        type: 'info',
+        message: `Downloading platform-specific archive for ${platform}:${arch}`,
+      });
+    }
+
+    await downloadAndExtractSingleArchive(
+      pluginInfo[platformArchKey],
+      pluginInfo[platformArchChecksumKey],
```
yolossn, line 496: what if the users arch isn't supported by the plugin? We have that before accessing the key

yolossn:
The changes look good, but I am not sure of using the same key (`headlamp/plugin/archive-url`) for the extra archive file to be downloaded. Maybe we should look into supporting a config like

```yaml
version: 0.1.3
name: headlamp_opencost
displayName: OpenCost
createdAt: '2024-07-22T05:35:50Z'
logoURL: 'https://raw.githubusercontent.com/headlamp-k8s/plugins/main/opencost/logo.png'
description: Adds information from OpenCost to resources' lists and details views.
annotations:
  headlamp/plugin/archive-url: 'https://github.com/headlamp-k8s/plugins/releases/download/opencost-0.1.3/headlamp-k8s-opencost-0.1.3.tar.gz'
  headlamp/plugin/archive-checksum: 'SHA256:cb3e67aa5bb8cc4869037b9024f58c95acfc1bc23ab6cbc80aa36e0c37ec21bd'
  headlamp/plugin/version-compat: '>=0.22'
  headlamp/plugin/distro-compat: 'in-cluster,web,docker-desktop,desktop'
  headlamp/plugin/extra-files:
      - url: ....
        checksum: .....
        type: <archive/binary> etc

```

joaquimrocha:
> The changes look good, but I am not sure of using the same key (`headlamp/plugin/archive-url`) for the extra archive file to be downloaded. Maybe we should look into supporting a config like
> 
> ```yaml
> version: 0.1.3
> name: headlamp_opencost
> displayName: OpenCost
> createdAt: '2024-07-22T05:35:50Z'
> logoURL: 'https://raw.githubusercontent.com/headlamp-k8s/plugins/main/opencost/logo.png'
> description: Adds information from OpenCost to resources' lists and details views.
> annotations:
>   headlamp/plugin/archive-url: 'https://github.com/headlamp-k8s/plugins/releases/download/opencost-0.1.3/headlamp-k8s-opencost-0.1.3.tar.gz'
>   headlamp/plugin/archive-checksum: 'SHA256:cb3e67aa5bb8cc4869037b9024f58c95acfc1bc23ab6cbc80aa36e0c37ec21bd'
>   headlamp/plugin/version-compat: '>=0.22'
>   headlamp/plugin/distro-compat: 'in-cluster,web,docker-desktop,desktop'
>   headlamp/plugin/extra-files:
>       - url: ....
>         checksum: .....
>         type: <archive/binary> etc
>         arch: <linux:64/...>
> ```

I think this is a good idea. I also realized that maybe we should by default add the `PLUGIN_FOLDER/PLUGIN_A/bin` to the $PATH as a way to always make sure that whatever extra binaries are installed should be there, WDYT?

illume:
@joaquimrocha the changes look nice to me. Thanks.

@yolossn can you please check these latest changes?

I would feel more comfortable if we limit the downloads (for now) to the kubernetes/minikube github. Then in the next release cycle we could open it up to more. As a precaution.
(What do you think?)

app/electron/plugin-management.ts
```diff
@@ -378,6 +384,8 @@ function validateArchiveURL(archiveURL) {
   const githubRegex = /^https:\/\/github\.com\/[^/]+\/[^/]+\/(releases|archive)\/.*$/;
   const bitbucketRegex = /^https:\/\/bitbucket\.org\/[^/]+\/[^/]+\/(downloads|get)\/.*$/;
   const gitlabRegex = /^https:\/\/gitlab\.com\/[^/]+\/[^/]+\/(-\/archive|releases)\/.*$/;
+  // For testing purposes, we allow localhost URLs.
```
illume, line 387: Could this be enabled under test only?

`if (process.env.NODE_ENV === 'test') { `

illume:
@joaquimrocha we need this to handle output paths. Since each os /architecture has a different output filename.

```shell
$ wget https://github.com/kubernetes/minikube/releases/download/v1.35.0/minikube-linux-amd64.tar.gz
$ tar xvf minikube-linux-amd64.tar.gz
out/minikube-linux-amd64
out/docker-machine-driver-kvm2
```

Or they also supply just the binary. This one is named minikube-linux-amd64, not "minikube". But we want the binary to be called minikube, not minikube-linux-amd64
https://github.com/kubernetes/minikube/releases/download/v1.35.0/minikube-linux-amd64

Perhaps a structure like this for tar.gz? So we can specify input and output file names.
```yaml

  headlamp/plugin/extra-files:
    - url: 'https://github.com/kubernetes/minikube/releases/download/v1.35.0/minikube-linux-amd64.tar.gz'
      checksum: 'SHA256:'
      arch: 'linux/x64'
      output:
        minikube:
          output: minikube
          input: out/minikube-linux-amd64
```

app/electron/main.ts
```diff
@@ -443,6 +443,65 @@ class PluginManagerEventListeners {
...
+ * Adds directories to the PATH environment variable
+ * @param {string[]} dirs - Directories to add to PATH
+ * @param {string} description - Description for logging (e.g., "plugin", "bundled plugin")
+ */
+function addToPath(dirs: string[], description: string): void {
+  if (dirs.length === 0) return;
+
+  const pathSeparator = process.platform === 'win32' ? ';' : ':';
+  const existingPath = process.env.PATH || '';
+  process.env.PATH = [...dirs, existingPath].join(pathSeparator);
```
illume, line 501: Do we want to use their minikube binary, or the one that comes with the plugin?

There are trade offs for both.

But if someone already has minikube, they may have already configured it, and it would work. Or they might be using a specific version for a reason (bugs in most recent version for example).

On the other hand, they may have downloaded it 4 years ago and never updated it.

illume:
Made a copy of this branch as it is now in extra-artifact-hub-tarballs-bak branch.

Going to take it over and change a few things:
- allow setting input paths and output filenames to handle minikube different tar ball names, with different binary names for each platform/arch combo. As per this comment: https://github.com/headlamp-k8s/headlamp/pull/2991#issuecomment-2750702473
- limit it to kubernetes/minikube repo for downloads for now
- instead of using PATH, download binary into a folder and then have runCommand only look into a hard coded set of folders for each one. So evil-plugin can't download it's own minikube binary.

app/electron/__tests__/test-data/artifacthub-pkg.yml
```diff
@@ -0,0 +1,40 @@
...
+  headlamp/plugin/extra-files/1/output/minikube/input: 'out/minikube-linux-arm64'
+  headlamp/plugin/extra-files/2/url: 'https://github.com/kubernetes/minikube/releases/download/v1.35.0/minikube-darwin-amd64.tar.gz'
+  headlamp/plugin/extra-files/2/checksum: 'SHA256:e2eadd14acaeb9876b70ed2b9f1ba3931125ec4b63e166b1088abece119494ea'
+  headlamp/plugin/extra-files/2/arch: 'darwin/x64'
+  headlamp/plugin/extra-files/2/output/minikube/output: 'minikube'
+  headlamp/plugin/extra-files/2/output/minikube/input: 'out/minikube-darwin-amd64'
+  headlamp/plugin/extra-files/2/output/docker-machine-driver-hyperkit/output: 'docker-machine-driver-hyperkit'
+  headlamp/plugin/extra-files/2/output/docker-machine-driver-hyperkit/input: 'out/docker-machine-driver-hyperkit'
+  headlamp/plugin/extra-files/3/url: 'https://github.com/kubernetes/minikube/releases/download/v1.35.0/minikube-darwin-arm64.tar.gz'
+  headlamp/plugin/extra-files/3/checksum: 'SHA256:05880f5b488215d865f6a391c5b52da4b84b220e53b47b333f290ce1edaa4189'
```
yolossn, line 32: IMHO We should use yaml arrays instead of numbers

illume: Artifacthub only allows strings here, not arrays.

yolossn (changes requested):
Left a couple of comments, rest looks good. Yet to test it by running the app.

app/electron/__tests__/test-data/artifacthub-pkg.yml
yolossn: other projects have used arrays and objects too

https://github.com/artifacthub/hub/blob/master/docs/olm_annotations.md
https://github.com/artifacthub/hub/blob/master/docs/krew_annotations.md

app/electron/plugin-management.ts
```diff
@@ -446,20 +484,211 @@ async function downloadExtractArchive(
...
+  matchingExtraFiles: ExtraFile[];
+} {
+  const currentPlatform = os.platform();
+  const currentArch = os.arch();
+  const currentArchString = `${currentPlatform}/${currentArch}`;
+
+  return {
+    currentArchString: currentArchString,
+    matchingExtraFiles: Object.values(extraFiles || {}).filter(
+      file => file.arch === currentArchString
```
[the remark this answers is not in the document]
illume: Thanks. I added the lowercase comparison.

illume:
@yolossn thanks, I addressed your notes and pushed.

app/electron/__tests__/test-data/artifacthub-pkg.yml
illume: Those do not use annotations? Note, they are not under "annotations".

app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "headlamp",
-  "version": "0.29.0",
+  "version": "0.30.0",
```
illume, line 3: This is just to make it easier to test installing the minikube plugin, because it requires version 0.30.0.

vyncent-t:
Review comments:

- after downloading via the plugins catalog, nav to home and noticed that the Minikube tab on side not rendering
- cannot reach the Minikube page to enter a name for a new minikube cluster
- plugin is enabled in general settings, I can still start and stop minikube clusters using the settings icon
- * Settings icon has double delete options if cluster shares a name with non minikube cluster

illume (approved):
👍 

(I'll wait for one more test, then remove the app version commit, and finally another approval before merge)

app/package.json
illume: This commit has been removed now.

illume:
Thanks @skoeva 

I removed the app version update commit. I'll wait for CI to finish before merge.

Issue #2689, opened 2024-12-13
Add quick status for each container (design TBD)


joaquimrocha:
In order for the user to quickly understand the status of containers in a pod, we could add a visual indicator for them in each row of the pods' table, maybe mixed with the status col.

PR #3008, opened 2025-03-21
frontend: pod: Add quick status for containers

skoeva:
This change adds a new icon displaying the status of each container in the pods list. Hovering over the icon gives important info about the container at a glance: name, status, reason, exit code, and start/finish times.

Fixes: #2689 

### Testing
- [X] Navigate to the Pods page in Headlamp
- [X] Note the container status icon, and hover over to see additional info about the container

-----
image
------
image

### DIFF
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -3,6 +3,7 @@ import { Box } from '@mui/material';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { ApiError } from '../../lib/k8s/apiProxy';
+import { KubeContainerStatus } from '../../lib/k8s/cluster';
 import Pod from '../../lib/k8s/pod';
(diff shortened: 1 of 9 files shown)


sniok:
There are some cases where status column and new containers column have different colors for the same status which can be a bit confusing

image

Also maybe we can move container status indicators into Status column?

frontend/src/components/pod/List.tsx
```diff
@@ -60,6 +60,59 @@ function getReadinessGatesStatus(pods: Pod) {
   return readinessGatesMap;
 }
 
+function getContainerDisplayStatus(container: any) {
```
sniok, line 63: We have a nice KubeContainer interface that can be used here instead of any

skoeva: oop thanks for catching this

PR #3010, opened 2025-03-22
frontend: hide pro option button

farodin91:


skoeva (commented):
don't forget to run `make frontend-test` and update the snapshots to get the CI to pass

skoeva (approved):

sniok:
we don't have the pro subscription and while the library is licensed with MIT and technically we're allowed to disable the attribution, I think the right thing to do is to keep it. Can I ask why you want to hide it? @farodin91

farodin91:
The buttom looks a bit broken. 
image


Wouldn't be more helpful to provide headlamp as showcase? https://reactflow.dev/learn/troubleshooting/remove-attribution

PR #3015, opened 2025-03-24
app: Allow commands to be run as elevated

joaquimrocha:
Certain commands may need to be run as admin, e.g. creating VMs.
This patch adds a new option `elevated` to the run-command IPC call.

Tests in platforms:
- [ ] Linux (it's failing on WSL)
- [ ] Mac
- [ ] Windows

I added a temp way to test by adding a button to run an elevated command to the change-logo settings.

cc/ @sniok

### DIFF
--- a/app/electron/runCmd.ts
+++ b/app/electron/runCmd.ts
@@ -1,7 +1,8 @@
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams,spawn } from 'child_process';
 import { app, BrowserWindow, dialog } from 'electron';
 import { IpcMainEvent } from 'electron/main';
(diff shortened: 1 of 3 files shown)


app/electron/runCmd.ts
```diff
@@ -132,6 +267,29 @@ export function handleRunCommand(
   if (!checkCommandConsent(eventData.command, mainWindow)) {
     return;
   }
+console.log('>>>>>>>>>>>>>>>>>>>>', eventData.command, eventData.args, eventData.options);
+  // Handle elevated command requests with a dedicated confirmation dialog
+  if (eventData.elevated) {
```
sniok, line 272: this should be `eventData.options.elevated`

with this change it worked on windows, although no logs were passed to the frontend

PR #3019, opened 2025-03-25
plugins: update-deps: Use npm ci to pin deps

skoeva:
This change addresses a security issue where dependencies are not pinned.


illume (approved):
🎉 thanks

PR #3022, opened 2025-03-26
docs: publishing: Remove install of headlamp-plugin globally

illume:
It is not needed, and installing things globally can be harmful. The package command is inside the package.json scripts section, so use that instead.


sniok (approved):
nice

skoeva (approved):
Thanks for this

PR #3025, opened 2025-03-26
docs: installation: Fix formatting

skoeva:
This change fixes some formatting bugs in the instructions for creating a service account token.


illume (approved):
👍 thanks!

PR #3027, opened 2025-03-26
[OLD] frontend: Add toggle wrap for logViewer and fix log viewer styles

vyncent-t:
requested feature from https://github.com/headlamp-k8s/headlamp/issues/2657

This PR adds a toggle button for the Logs Viewer window that allows the text to be formatted for word wrap. There are also style formatting issues that I have noticed and made an issue for here https://github.com/headlamp-k8s/headlamp/issues/3052  this PR also adds changes that style error while also. 

## Changes
image
- **Introduces new Wrap option for log viewer component to allow toggle between default text formatting and word wrap**

image

- **Introduces new format and tooltip context for log viewer buttons** 
  - At current, adding additional switches or buttons to the log viewer components for both pod and non pod log viewer introduces styling errors (this PR makes it so that we can continue to add more features by reworking the button layout)

- **FIxes rendering for small view**
  - At current when viewing log viewer on smaller windows the buttons are both misaligned and not styled correctly https://github.com/headlamp-k8s/headlamp/issues/3052

### How to test

- Navigate to pods
- Click on any pod
- Open the "View logs" button
- Click the toggle "Wrap" button


vyncent-t:
wip: need to fix window needing more than one click

vyncent-t:
This is more for word break than it is for toggle line wrap, line wrap would require a long scroll bar, not sure if addonfit will accommodate this

skoeva (commented):
after toggling once, looks like you can't toggle back to the original view

vyncent-t:
> after toggling once, looks like you can't toggle back to the original view

yea I had noticed still trying a few things with this

vyncent-t:
i think the pod view alignment may be unique, could be a different UI issue

PR #3029, opened 2025-03-27
Makefile: Fix run-backend on windows

illume:
Otherwise enable dynamic clusters gets "true " and golang
does not think that is valid.


sniok (approved):
tested and it works, thanks

PR #3038, opened 2025-03-27
headlamp-plugin: Allow to copy extra dist files into dist folder

joaquimrocha:
This is needed for adding extra dist files like the wasm file for the IG plugin.

cc/ @ashu8912 

Used like: 
```
{
  "name": "my-headlamp-plugin",
  "version": "1.0.0",
  "headlamp": {
    "extraDist": {
      "configs/defaults.json": "./src/configs/defaults.json",
      "assets/": "./assets/",
      "README.md": "./README.md"
    }
  }
}
```

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -230,6 +230,63 @@ async function calculateChecksum(filePath) {
 }
 
 /**
+ * Copy extra files specified in package.json to the dist folder
+ *
+ * @param {string} [packagePath='.'] - Path to the package root containing package.json
+ * @returns {Promise<void>}
+ */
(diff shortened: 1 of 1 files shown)


sniok:
typically for this kind of extra files the convention is to use "public" folder, like the one we have in the frontend https://github.com/headlamp-k8s/headlamp/tree/main/frontend/public

it would make the code a bit easier too, since all we need to do is copy public/ -> dist/

but this is okay too, we can revisit this later

sniok (approved):

joaquimrocha:
The reason for not using the public folder from vite config is that the vite config is contained/abstracted by headlamp-plugin, so AFAIU we don't have the vite config in the plugin itself and I think that instead of trying to have it and merge both configs, maybe having our own implementation is more future proof.

PR #3039, opened 2025-03-27
frontend: logViewer: Add toggle button for word break

vyncent-t:
This PR adds a toggle button for the Logs Viewer window that allows the text to be formatted for word break.

similar to requested feature from https://github.com/headlamp-k8s/headlamp/issues/2657

WIP:

- not sure if this should be enabled by default

### How to test

- Navigate to pods
- Click on any pod
- Open the "View logs" button
- Click the toggle "Word break" button


skoeva (commented):
this might be good to enable by default, and then have the changes from the other PR as a toggle potentially

illume:
> this might be good to enable by default, and then have the changes from the other PR as a toggle potentially

@vyncent-t can you please move this to a new PR if it needs changes? Sounds like it does. cc @skoeva

illume:
@vyncent-t this needs to be moved to a new PR because it's on a branch that can't be modified (after the move to k8s).

PR #3048, opened 2025-03-28
Makefile: adds `backend-lint-fix` command

DhairyaMajmudar:
## Description

-> Adds a new command `backend-lint-fix` 
-> This command is the updated version of the existing command `backend-lint` 
-> The difference is it has `--fix` flag which fixes the line-space errors, indentation, and missing full-stop errors itself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,9 @@ tools/golangci-lint: backend/go.mod backend/go.sum
 backend-lint: tools/golangci-lint
 	cd backend && ./tools/golangci-lint run
 
+backend-lint-fix: tools/golangci-lint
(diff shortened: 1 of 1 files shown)


illume:
Nice one, thanks!

(A small nitpick on the git commit. We use a capitalized imperative mood verb. So Add, Update, Remove instead of adds, updates, removes. Also we don't use chore/feature/etc)

`Makefile: Add backend-lint-fix target`

DhairyaMajmudar:
Thank you! I've changed the commit message 👍🏻

skoeva (approved):
Thanks for this!

illume (approved):
👍

PR #3050, opened 2025-03-28
headlamp-plugin: Ensure the dist files are all copied on npm run start

joaquimrocha:
Only the index.js and package.json were being copied to the plugin folder. This was because the copy of the extra dist files was happening on "close bundle" rather than on "build end", i.e. it was happening only after copying the static files.

This patch fixes that and also moves the setting up of the static copy to after the setting up of the extra-dist-files logic. This has no real effect but makes the code more readable since it's ordered like it actually is processed.


illume (approved):
👍

sniok (approved):
tested and it works

Issue #3047, opened 2025-03-28
Show failure reason when apply operation fails


AlleM43:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
I've been trying to create objects using the "Create" button, but the generic failure message doesn't show what the issue is, so I have to copy whatever I'm doing to a file and try to apply it using kubectl to get an error message that actually tells me what's wrong, slowing me down significantly.

## Describe the solution you'd like
The failure reason shown by kubectl apply should also be shown by Headlamp when an apply operation fails.

## What users will benefit from this feature?
Anyone who's writing YAML in Headlamp that doesn't work on the first try.

## Are you able to implement this feature?
No.

skoeva:
I can push up a fix for this

illume:
@skoeva fixed this, and it's been merged into main. It will be in the next Headlamp release.

Thanks for the report @AlleM43

PR #3051, opened 2025-03-28
frontend: EditorDialog: Show error reason on failure

skoeva:
This change gives additional information to the user about resource creation failures, using the reason provided in the value object. The message in the UI will fall back to the original message when a reason is not provided in the object.

Fixes: #3047 

### Testing

- [X] Click on the 'Create' button in Headlamp and try to create a resource with an invalid name (i.e. config-map!)
- [X] Note the failure message includes the error reason as expected

image


illume (approved):
Nice fix

PR #3054, opened 2025-03-31
Improve WebSocket compatibility

### LINKED ISSUE #2970: cannot execute shell or view logs of pods that run on rancher cluster
## Describe the bug
cannot execute shell or view logs of pods that run on rancher cluster

## To Reproduce
Steps to reproduce the bug:
1. Go To a Pod
2. [Click on terminal/shell button or show log  button
3. "Failed to connect…Press the enter key to reconnect." is printed

 - Installation type: Windows executable
 - Headlamp Version: version 0.29
 - Other: k8s cluster is rancher
 - kubectl logs with same kubeconfig works as expected

sniok:
Fixes #2970 

Certain k8s vendors don't support tokens via Sec-Websocket-Protocol value, so we need to use the Authorization header instead.
Certain k8s vendors will perform additional cors check when browser-like User-Agent is present, which we currently pass but now will just remove.

How to test:
 - Create rancher cluster
 - Open headlamp and check for websocket connections (live updates to lists, logs and attach features)

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1132,6 +1132,7 @@ func handleClusterAPI(c *HeadlampConfig, router *mux.Router) {
 
 		r.Host = clusterURL.Host
 		r.Header.Set("X-Forwarded-Host", r.Host)
+		r.Header.Del("User-Agent")
 		r.URL.Host = clusterURL.Host
 		r.URL.Path = mux.Vars(r)["api"]
 		r.URL.Scheme = clusterURL.Scheme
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this.

- There's various lint issues.
- Needs tests.

(I guess because our repo is still broken we can not push to branches, it will need to be done in a new PR.)

sniok:
Moved to https://github.com/kubernetes-sigs/headlamp/pull/3142

PR #3072, opened 2025-04-03
github: backend: charts: Fix repo link

skoeva:
The repo moved to kubernetes sig, and the website does not redirect.


joaquimrocha:
Not merging because AFAIU Kubernetes has its own OCI registry. I'll need to investigate that next week, till then, we do have that notice in the readme.
If you adapt the PR to not include the ghcr changes for now, I think we can merge.

skoeva:
Been playing around with the CI failures as well, I can move this to draft and reopen when it's in a better state

charts/headlamp/README.md
joaquimrocha, line 1: This one shouldn't change, because we are still using the image in headlamp-k8s/packages.

backend/pkg/helm/repository_test.go
```diff
@@ -130,6 +131,8 @@ func TestRemoveRepository(t *testing.T) {
 func TestUpdateRepo(t *testing.T) {
 	helmHandler := newHelmHandler(t)
 
+	logger.Log(uint(1), map[string]string{"test": "TestUpdateRepo"}, nil, "Starting test for updating helm repository")
```
joaquimrocha, line 134: Is this needed?

joaquimrocha (dismissed):

backend/pkg/helm/repository_test.go
skoeva: This was just to see why there was a difference between what was pushed and what appeared in the PR. somehow this file got automatically updated

illume:
The build container and test job failed... I reran it (in case it is just a flaky intermittent error).

PR #3080, opened 2025-04-07
frontend: Custom App themes

sniok:
This PR introduces AppTheme definition that contains limited number of tokens that then used to make full material-ui theme. 
Plugins then can provide their own custom themes using the AppTheme definition

```ts
export interface AppTheme {
  name: string;
  /** Base theme to extend */
  base?: 'light' | 'dark';
  /** Primary theme color */
  primary?: string;
  /** Secondary theme color */
  secondary?: string;
  text?: {
    /** Primary text color */
    primary?: string;
  };
  link?: {
    /** Link text color */
    color?: string;
  };
  background?: {
    /** Background color of the page */
    default?: string;
    /** Background color of popups and menus */
    paper?: string;
    /** Shaded background color */
    muted?: string;
  };
  sidebar?: {
    /** Background color of the sidebar */
    background?: string;
    /** Text and icon color of the sidebar */
    color?: string;
    /** Background color for the selected item */
    selectedBackground?: string;
    /** Text color for the selected item */
    selectedColor?: string;
    /** Background color of sidebar action button */
    actionBackground?: string;
  };
  navbar?: {
    /** Background color of the navbar */
    background?: string;
    /** Text and icon color of the navbar */
    color?: string;
  };
  /** General shape radius (things like buttons, popups, etc) */
  radius?: number;
  /** Text style in buttons */
  buttonTextTransform?: 'uppercase' | 'none';
}
```


joaquimrocha:
Tested the PR and found a couple of issues (possibly there are more but I haven't gone yet commit by commit):
1. The theme switch button is not dark, and makes it barely noticeable:
Screenshot showing the point above

2. I think the disabled buttons are not accessible:
Screenshot showing the point above

3. The text/icons color looks a bit too dull on the new background color. I think we intended it to have more contrast.

sniok:
> Tested the PR and found a couple of issues (possibly there are more but I haven't gone yet commit by commit):
> 
> 1. The theme switch button is not dark, and makes it barely noticeable:
> 3. The text/icons color looks a bit too dull on the new background color. I think we intended it to have more contrast.

Colors have been adjusted to match the previously agreed theme colors

> 2. I think the disabled buttons are not accessible:
>    Screenshot showing the point above

Disabled button don't have to be accessible [Source](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

> User Interface Components that are not available for user interaction (e.g., a disabled control in HTML) are not required to meet contrast requirements.

PR #3086, opened 2025-04-08
frontend/src/App.tsx: enable devtools according to environment

warjiang:
I mentioned that headlamp project already made an integration with react-query devtools, but it will display the devtools by setting the `queryDevtoolsEnabled=true` in `frontend/App.tsx`, it will not be convenient for developer. And I found that there some useful [build-in environment var in the vite](https://vite.dev/guide/env-and-mode), like `import.meta.env.PROD`, so I think we can set `queryDevtoolsEnabled` according to env automatically


joaquimrocha:
@sniok can you check this one out?

sniok:
Hi, thanks for opening a PR! The react-query devtools can be useful but most of the time we don't really need it. When enabled it adds a big button to the page that may be confusing or distracting to others so I think we should keep the devtools disabled even in development mode by default.

I agree that setting a variable to true is not very convenient but given how rarely we need the devtools I think it's a good enough compromise. But maybe we can use something like REACT_QUERY_DEVTOOLS env variable for this, what do you think?

joaquimrocha:
> But maybe we can use something like REACT_QUERY_DEVTOOLS env variable for this, what do you think?

I think this is a good idea (and documentation for it).

warjiang:
> REACT_QUERY_DEVTOOLS

Agree with your idea except the env name, it should be named as `REACT_APP_ENABLE_REACT_QUERY_DEVTOOLS` because the env prefix in `vite.config.ts` file:
image


here is the screenshot
20250409215222_rec_

PR #3089, opened 2025-04-09
app, frontend: Unify locales

joaquimrocha:
Instead of having two separate config and dirs for locales, these changes unify the locales under the frontend directory. This will reduce the configuration burden and being less error prone, making it easier for i18n contributors to see all needed translations in one place.

## How to test
1. Build the app and run it
2. Verify that changing the locale in Settings does change the menus + frontend (menus are missing from zh-tw and the sidebar has a bug where it doesn't get updated on translation change, but otherwise it should work)

1. Add a translation change to some string inside the app/ module, do the some for the frontend module
2. Run make i18n and verify that it's updating 2 locale files for all languages.


sniok (approved):
This is a nice improvement! 
I wonder if we should extract this into a separate package so both app and frontend can import it, maybe we can do then if/when we do monorepo setup

PR #3093, opened 2025-04-10
Add registerAppTheme function for plugins

sniok:
Adds ability to register custom themes


sniok:
> @sniok , I thought you had mentioned we could have the theme plugins be just the package.json and have the theme defined in there. This would prevent dynamically setting themes' variables and stuff but maybe that's also not yet supported. The way this PR adds support it's consistent with the rest of the functionality, but I wonder if we doing the package.json-based ingestion is a safer alternative and also more future proof.

Yeah it was just easier to go with a current registerX way, it's consistent with all the other stuff and also provides a bit better developer experience since you get type suggestions and documentation. 

I do think that declaring themes in a package.json is safer but we can add it later, maybe also combine with some other things that can be added declaratively in a json (like logo for example)

joaquimrocha:
There's still possibly something else to consider: before plugins are loaded, we do show some UI elements. If those plugins register a theme dynamically, we cannot understand whether we should hold on showing any themed UI elements till those plugins are loaded.
Having the theme in a package.json won't fix this automatically but it will allow us to know whether we should hold on showing those. Maybe the same can be accomplished by a tag in package.json for now, but I wanted this thought to be registered here.

illume (changes requested):
Looks good.

Can you please add an example plugin?

sniok:
Thanks for taking a look! Marking this as a draft for now since I want to address a case where plugins aren't fully loaded yet to avoid showing different theme for a split second and I'll add an example plugin for this

plugins/examples/app-theme/src/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { registerAppTheme } from '@kinvolk/headlamp-plugin/lib';
+
```
illume, line 2: How about a comment here something like:
> // To see this theme, go into Headlamp settings and change the theme to "my custom theme"

docs/development/plugins/functionality/index.md
```diff
@@ -173,3 +173,10 @@ The plugins can have user-configurable settings that can be used to change the b
 - Example plugin shows [How to create plugin settings and use them](https://github.com/kubernetes-sigs/headlamp/tree/main/plugins/examples/change-logo)
 
 ![screenshot of the plugin settings](../images/plugin-settings.png)
+
+### App Theme
+
+Add a new custom Headlamp theme using [registerAppTheme](../../api/plugin/registry/functions/registerapptheme).
+Theme definition is an object, name must be unique and you must specify 'light' or 'dark' base for the theme.
```
illume, line 180: I don't see in the example plugin where it specifies the base theme?

docs/development/plugins/functionality/index.md
```diff
@@ -173,3 +173,10 @@ The plugins can have user-configurable settings that can be used to change the b
...
+
+Check the full [AppTheme](../../api/lib/AppTheme/interfaces/AppTheme.md) definition for all of the properties you can customize.
```
illume, line 182: A screen shot is needed here. One of selecting the custom theme from the theme select list would work.

plugins/examples/app-theme/src/index.tsx
sniok: done

docs/development/plugins/functionality/index.md
sniok: added it to the example and updated this sentence because 'base' property is not actually mandatory

docs/development/plugins/functionality/index.md
sniok: good point, added a screenshot

plugins/examples/custom-theme/README.md
```diff
@@ -0,0 +1,11 @@
+# Custom Theme Example
+
+This plugin demonstrates how to add a custom theme
+
+```bash
+cd plugins/examples/app-theme
```
illume, line 6: has old name

sniok: fixed

PR #3095, opened 2025-04-10
eslint: documentaton: fix dead link to headlamp eslint-config

Dumbaz:
Due to restructuring the current present link is resulting in a 404.

No changes apart from correcting link pointing to correct README file


illume (approved):
👍 thanks

/approve

PR #3104, opened 2025-04-11
github/backend-test: Fix coverage so it skips when main fails

illume:
When there is an error in the main branch backend tests we skip doing the coverage check.

Because currently there is a problem where other PRs fail the backend CI test job, because main is failing for it. (even if the PR fixes the issue).

### testing done

- I commented out the previous step so that, and ran this code when there was an issue in main with the backend test
- I also ran the workflow injecting a fault into the base coverage step.


skoeva (approved):
Thanks for this 🎉

Issue #2897, opened 2025-02-13
Join Kubernetes SIG-UI


joaquimrocha:
We would like to join the SIG-UI and this involves moving the repo into 

From https://github.com/kubernetes/community/blob/master/github-management/kubernetes-repositories.md#rules-for-donated-repositories , here are the tasks we have to do for this to happen:

From the new repos list:
- [ ] For now all repos will live in `github.com/kubernetes-sigs/<project-name>`
- [ ] For project names: see [project-naming](https://github.com/kubernetes/community/blob/master/committee-steering/governance/project-naming.md)
- [ ] Must contain the topic for the sponsoring SIG - e.g. `k8s-sig-api-machinery` (Added through the *Manage topics* link on the repo page)
- [ ] Must adopt the Kubernetes Code of Conduct
- [x] All code projects use the Apache License version 2.0. Documentation repositories must use the Creative Commons License version 4.0
- [ ] Must adopt the CNCF CLA bot, merge bot and Kubernetes PR commands/bots
- [ ] All OWNERS of the project must also be active SIG members
- [ ] Must be approved by the process spelled out in the SIG's charter and a publicly linkable written decision should be available for the same
- [ ] SIG must already have identified all of their existing subprojects and code, with valid OWNERS files, in `sigs.yaml`

From the donated repos list:
- [ ] All contributors must have signed the [CNCF Individual CLA](https://github.com/cncf/cla/blob/master/individual-cla.pdf) or CNCF Corporate CLA
- [ ] If (a) contributor(s) have not signed the CLA and could not be reached, a NOTICE file should be added referencing section 7 of the CLA with a list of the developers who could not be reached
- [x] Licenses of dependencies are acceptable; please review the [allowed-third-party-license-policy.md](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md) and [exceptions](https://github.com/cncf/foundation/tree/main/license-exceptions). If your dependencies are not covered, then please open a `License Exception Request` issue in cncf/foundation repository
- [ ] Boilerplate text across all files should attribute copyright as follows: `"Copyright <Project Authors>"` if no CLA was in place prior to donation
- [ ] Additions of the standard Kubernetes header to code created by the contributors can occur post-transfer, but should ideally occur shortly thereafter
- [ ] Should contain template files as per the [kubernetes-template-project](https://github.com/kubernetes/kubernetes-template-project)

Changes related to this should go into a `sig-move` branch.

[...]

PR #3106, opened 2025-04-11
Add kubernetes copyright

joaquimrocha:
As part of moving the project under the Kubernetes SIG UI, we should make sure that every source file should have the Kubernetes Authors copyright header.


skoeva (approved):

illume:
Do you have a link to the requirement?

Also, have you checked this is ok with legal?

joaquimrocha:
> Do you have a link to the requirement?
> 
> Also, have you checked this is ok with legal?

This is one of the requirements of projects under Kubernetes. We're following the procedure.

joaquimrocha:
Reference: https://github.com/kubernetes-sigs/headlamp/issues/2897

PR #3109, opened 2025-04-14
frontend: Include pod status in column text value to avoid stale status label state

### LINKED ISSUE #3108: pod list status color not refreshing
## Describe the bug
the status of a pod is transitioned to `Runnig` but the color is still yellow and not refresing to green until you changed the view.

Image

## To Reproduce
Steps to reproduce the bug:
1. Go to pod list
2. Delete a pod
3. Wait for recreation
4. See wrong status color when pod is Running

 - Installation type: In-Cluster /Windows Chocolatey
 - Headlamp Version: 0.30.0
 - Other: k3s v1.31.7+k3s1

MAybe

## Additional Context
none

sniok:
Fixes #3108 

Table cell wasn't re-rendering because the "reason" label stayed the same but status (which impacts color and the icon) changed but wasn't taken into account in the getValue function

How to test (steps from reported issue):

1. Go to pod list
2. Delete a pod
3. Wait for recreation
4. Check status color

### DIFF
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -13,11 +13,9 @@ import { LightTooltip, Link, SimpleTableProps } from '../common';
 import { StatusLabel, StatusLabelProps } from '../common/Label';
 import ResourceListView from '../common/Resource/ResourceListView';
 
-export function makePodStatusLabel(pod: Pod) {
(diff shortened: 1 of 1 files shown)


sniok:
Moved to https://github.com/kubernetes-sigs/headlamp/pull/3141

PR #3113, opened 2025-04-14
frontend: plugins: Bump vite to 6.2.6

skoeva:
These changes bump vite in frontend and the plugins examples to v6.2.6.


illume (approved):
👍 thanks

PR #3121, opened 2025-04-15
Fix rename cluster duplicate name issue / improve cluster metadata for UI 

vyncent-t:
original - https://github.com/kubernetes-sigs/headlamp/pull/2944

This branch replaces the need for https://github.com/headlamp-k8s/headlamp/pull/2747

This PR introduces `clusterID` for non dynamic cluster use. It also includes a fix for issue with duplicate name clusters via blocker UI.

## Cluster metadata info
- Introduces compatible find cluster by name or by clusterID logic to `findClusterByName`
Note: `ClusterID` is currently used for non dynamic cluster actions only. (Cluster renaming, deleting coming soon.)

- Expands metadata field for clusters, metadata field now includes:
  - Source  `cluster.metadata.origin.kubeconfig` : File path of the kubeconfig
  - OriginalName `cluster.metadata.originalName` : Original name the cluster was created with
  - ClusterID  `cluster.metadata.clusterID` : Unique ID for finding cluster, source + originalName

## Home View

image

### How to test actual origin in app mode
- Run headlamp in App mode
- At the home view: notice the specific Kubeconfig path under `Origin` on the table

### Original name label
image

- `OriginalName` now displayed in cluster settings view

### How to test the original name label
- Run headlamp in App mode
- Create a non-dynamic cluster and let it load at home view
- Click on the settings icon in the cluster list and navigate to settings page
- Notice the original name displayed under the `Name` section

---

### Name in use blocker UI
image

- Introduced UI dialog for preventing names already in use from being used. 

### How to test the duplicate rename blocker UI
- Run headlamp in App mode
- Create a two non-dynamic clusters and let them load at home view
- Click on the settings icon in the cluster list and navigate to settings page
- Rename one of the non dynamic cluster to a name already in use by another cluster
- Notice the "Invalid custom name" UI
- Choose a name that is valid and rename it
- Navigate back to the settings page if needed and note the new custom name 
- Click on the settings tab of the renamed cluster and notice and original name displayed


backend/cmd/headlamp.go
```diff
@@ -1582,9 +1582,15 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 	c.getConfig(w, r)
 }
 
-// Get path of kubeconfig from source.
+// Gets the path of the kubeconfig we load headlamp with from source.
+// Note for non dynamic clusters:
+//   - This will always give the kubeconfig we loaded headlamp with, but if more than one,
+//     it's not clear this will yield the correct value.
```
joaquimrocha, line 1588: I think comments like this should be avoided. We should understand whether this yields correct or incorrect values.

backend/cmd/headlamp.go
```diff
@@ -1582,9 +1582,15 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 	c.getConfig(w, r)
 }
 
-// Get path of kubeconfig from source.
+// Gets the path of the kubeconfig we load headlamp with from source.
```
joaquimrocha, line 1585: This new comment is not clearer than before.

backend/cmd/headlamp.go
```diff
@@ -1582,9 +1582,15 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
+// Note for non dynamic clusters:
+//   - This will always give the kubeconfig we loaded headlamp with, but if more than one,
+//     it's not clear this will yield the correct value.
 func (c *HeadlampConfig) getKubeConfigPath(source string) (string, error) {
-	if source == "kubeconfig" {
+	// note: we need to have it in a const for the linting to pass
```
joaquimrocha, line 1590: So right now the linting is failing in main?

backend/cmd/headlamp.go
```diff
@@ -1672,32 +1678,65 @@ func (c *HeadlampConfig) updateCustomContextToCache(config *api.Config, clusterN
...
 
-	return path, config, nil
+	return usePath, kubeConfig, nil
 }
 
 // Handler for renaming a cluster.
 func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	clusterName := vars["name"]
+	parts := strings.SplitN(r.URL.Query().Get("ClusterID"), ":", 2)
```
joaquimrocha, line 1737: I think if we have the ClusterID in our copy of the clusters we have loaded, then we should find the cluster by ID, without deriving the actual kubeconfig from the ID itself but from the object we have the context, cluster ID, etc. for.
If we do need to get the kubeconfig, then maybe we change the request to include the kubeconfig from the client side.

backend/cmd/headlamp.go
```diff
@@ -1715,47 +1754,111 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	var loadPath string
+	if reqBody.Source == "kubeconfig" {
+		loadPath = configPath
+	}
+
 	// Get path of kubeconfig from source
-	path, config, err := c.getPathAndLoadKubeconfig(reqBody.Source, clusterName)
+	path, config, err := c.getPathAndLoadKubeconfig(reqBody.Source, clusterName, loadPath)
```
joaquimrocha, line 1763: It's confusing to have a source and a load path IMO. Maybe we get the correct kubeconfig file here and pass it to the getPathAndLoadKubeconfig.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -97,25 +101,67 @@ export default function SettingsCluster() {
...
+
+    const allNames = [...clusterNames, ...originalNames];
+
+    const nameInUse = allNames.includes(name);
+
+    return nameInUse;
+  }
 
   const handleUpdateClusterName = (source: string) => {
+    const nameInUse = checkNameInUse(newClusterName);
```
joaquimrocha, line 142: Instead of validating here, you can validate in onChange, so the user sees right away that this name is taken, without having to see the error dialog (just seeing the entry validation).

joaquimrocha (changes requested):
Thanks. Overall it's working well but I left some comments about the code and the way we can validate the name change in the client side a bit more dynamically.

backend/cmd/headlamp.go
vyncent-t: this is a note from this comment I thought would be relevant to include, will remove it
https://github.com/kubernetes-sigs/headlamp/pull/2944#discussion_r1985436189

backend/cmd/headlamp.go
vyncent-t: also thought it would be relevant to include, will revert

backend/cmd/headlamp.go
vyncent-t: not in main, but with these new changes while running the linter locally, the linter did not like just using an if block this way and would fail.

image
image

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -337,15 +339,31 @@ type ContextLoadError struct {
 // It reads the kubeconfig file from the given path and loads the contexts from the file.
 // It returns an error if the file cannot be read.
 // It will return valid contexts, ContextLoadError and errors if there are any errors in the file.
-func LoadContextsFromFile(kubeConfigPath string, source int) ([]Context, []ContextLoadError, error) {
-	data, err := os.ReadFile(kubeConfigPath)
+func LoadContextsFromFile(configPath string, source int) ([]Context, []ContextLoadError, error) {
+	data, err := os.ReadFile(configPath)
```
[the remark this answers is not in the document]
vyncent-t: not needed functionally but visually, as we do use `kubeConfigPath` pretty often and wanted a clearer difference between the param `kubeConfigPath` small k and the context field prop of `.KubeConfigPath` big k

will revert for uniform

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -36,15 +36,17 @@ const (
...
+	KubeContext    *api.Context           `json:"kubeContext"`
+	Cluster        *api.Cluster           `json:"cluster"`
+	AuthInfo       *api.AuthInfo          `json:"authInfo"`
+	Source         int                    `json:"source"`
+	OidcConf       *OidcConfig            `json:"oidcConfig"`
+	proxy          *httputil.ReverseProxy `json:"-"`
+	Internal       bool                   `json:"internal"`
+	Error          string                 `json:"error"`
+	KubeConfigPath string                 `json:"kubeConfigPath"`
+	ClusterID      string                 `json:"clusterID"`
```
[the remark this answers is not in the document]
vyncent-t: just to be sure- when you say "inside KubeContext", do you mean this very item through location of origin?

image

```
// Context is a tuple of references to a cluster (how do I communicate with a kubernetes cluster), a user (how do I identify myself), and a namespace (what subset of resources do I want to work with)
type Context struct {
	// LocationOfOrigin indicates where this object came from.  It is used for round tripping config post-merge, but never serialized.
	// +k8s:conversion-gen=false
	LocationOfOrigin string `json:"-"`
	// Cluster is the name of the cluster for this context
	Cluster string `json:"cluster"`
	// AuthInfo is the name of the authInfo for this context
	AuthInfo string `json:"user"`
	// Namespace is the default namespace to use on unspecified requests
	// +optional
	Namespace string `json:"namespace,omitempty"`
	// Extensions holds additional information. This is useful for extenders so that reads and writes don't clobber unknown fields
	// +optional
	Extensions map[string]runtime.Object `json:"extensions,omitempty"`
}

```

backend/cmd/headlamp.go
vyncent-t: I made changes into the way the current rename logic works, where we get the cluster by ID in the frontend and then just pass the kubeConfigPath from the clusterInfo into the url to use it there

it no longer breaks down the cluster ID to get the path after finding the cluster by ID

backend/cmd/headlamp.go
vyncent-t: done

now gets the accurate kube config path then loads it

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: done

backend/cmd/headlamp.go
joaquimrocha: I don't think this linting can refer to this function, where there's only one occurrence of the string in question.

joaquimrocha (changes requested):

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: done

backend/cmd/headlamp.go
```diff
@@ -1738,40 +1742,99 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
...
 				logger.Log(logger.LevelError, map[string]string{"cluster": contextName},
 					err, "marshaling custom object")
-
-				return
+				continue
 			}
 
-			// Check if the CustomName field matches the cluster name
-			if customObj.CustomName != "" && customObj.CustomName == clusterName {
+			if customObj.CustomName == clusterName && customObj.CustomName != "" {
```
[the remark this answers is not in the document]
vyncent-t: reverted

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -337,20 +389,38 @@ export default function SettingsCluster() {
...
+        }}
+        disableNoButton
+        open={clusterErrorDialogOpen}
+        title={t('translation|Invalid custom name')}
+        description={t(
+          'translation|Custom names must be unique and follow valid format for cluster rename. '
+        )}
+        confirmLabel={t('translation|Okay')}
+      >
+        {t('translation|Close')}
```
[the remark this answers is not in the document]
vyncent-t: removed

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -30,10 +30,28 @@ export interface ConfirmDialogProps extends MuiDialogProps {
...
-  const { onConfirm, open, handleClose, title, description, cancelLabel, confirmLabel } = props;
+  const {
+    onConfirm,
+    open,
+    handleClose,
+    title,
+    description,
+    cancelLabel,
+    confirmLabel,
+    disableNoButton = false,
```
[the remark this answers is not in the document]
vyncent-t: done

backend/cmd/headlamp.go
vyncent-t: done

backend/cmd/headlamp.go
```diff
@@ -1738,40 +1739,99 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "custom name already in use", http.StatusBadRequest)
+		logger.Log(logger.LevelError, map[string]string{"cluster": clusterName},
+			err, "cluster name already exists in the kubeconfig")
+
+		return
+	}
+
+	contextName := findMatchingContextName(config, clusterName)
+
+	if err := customNameToExtenstions(config, contextName, reqBody.NewClusterName, path); err != nil {
```
joaquimrocha, line 1753: Nit: Typo but it was here before. We could still fix it in a new commit.

joaquimrocha (approved):
Left a nit but we don't have to fix it in this commit.

joaquimrocha:
Thanks @vyncent-t . I think it's good now. The test failure is unrelated AFAICS.
/lgtm
/approve

illume:
Please rebase against main? Then we can see if the test failure is related to this LR?

frontend/src/lib/k8s/kubeconfig.ts
```diff
@@ -154,6 +154,8 @@ export interface KubeconfigObject {
       user: string;
       /** namespace is the default namespace. */
       namespace?: string;
+      // this is needed for attaching clusterID to the cluster for accurate cluster actions
```
illume, line 157: Please use the same style comment as the surrounding code?

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -344,24 +433,19 @@ export default function SettingsCluster() {
...
-                    error={!isValidCurrentName}
-                    helperText={
-                      isValidCurrentName
-                        ? t(
-                            'translation|The current name of the cluster. You can define a custom name.'
-                          )
-                        : invalidClusterNameMessage
-                    }
+                    error={!isValidCurrentName || customNameInUse}
+                    helperText={handleHelperText()}
```
illume, line 448: You could still keep this inline if you want by using an anon function here. I guess it's easier to read inline, rather than having to find the function elsewhere.

```tsx
helperText={()=> {
  // code goes here
}()}
```

backend/cmd/headlamp_test.go
```diff
@@ -649,6 +649,58 @@ func TestHandleClusterAPI_XForwardedHost(t *testing.T) {
 	assert.Equal(t, "OK", rr.Body.String())
 }
 
+// Handle a cluster rename request.
```
illume, line 652: Please follow golang style comments, and follow the style of the other comments in the file? Start with the function name.

backend/cmd/headlamp_test.go
```diff
@@ -649,6 +649,58 @@ func TestHandleClusterAPI_XForwardedHost(t *testing.T) {
...
+		assert.Equal(t, tc.expectedState, r.Code)
+	} else {
+		url := "/cluster/minikubetest?ClusterID=minikubetest"
+		r, err = getResponseFromRestrictedEndpoint(handler, "PUT", url, tc.clusterReq)
+		require.NoError(t, err)
+		assert.Equal(t, tc.expectedState, r.Code)
+	}
+}
+
+// Used to run the cluster rename tests.
```
illume, line 679: Please follow golang style comments, and follow the style of the other comments in the file? Start with the function name.

backend/pkg/kubeconfig/kubeconfig.go
illume, line 49: Please add documentation for this field?

(All the fields should be documented, but feel free to not do that here... and just document the new ones)

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -345,7 +347,23 @@ func LoadContextsFromFile(kubeConfigPath string, source int) ([]Context, []Conte
 
 	skipProxySetup := source != KubeConfig
 
-	return loadContextsFromData(data, source, skipProxySetup)
+	contexts, contextErrors, err := loadContextsFromData(data, source, skipProxySetup)
```
illume, line 350: Please add test(s) for these changes to kubeconfig_test.go?

backend/cmd/headlamp.go
```diff
@@ -1744,40 +1756,99 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Find the context with the given cluster name
+	isUnique := checkUniqueName(config, clusterName, reqBody.NewClusterName)
```
illume, line 1759: Can you think of a better place to put these changes than headlamp.go? 

(We want to try and move code from headlamp.go into packages when possible. So if there's already a more suitable package, it would be better to move it there.)

backend/cmd/headlamp.go
```diff
@@ -1744,40 +1756,99 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
...
 
-	if err := customNameToExtenstions(config, contextName, reqBody.NewClusterName, path); err != nil {
-		http.Error(w, "writing custom extension to kubeconfig", http.StatusInternalServerError)
-		return
+	return contextName
+}
+
+// checkUniqueName returns false if 'newName' is already in 'names' otherwise returns true.
+// used for checking cluster names.
+func checkUniqueName(config *api.Config, contextName string, newName string) bool {
```
illume, line 1810: I can't tell if this function is well covered by the tests? Can you please check the code coverage and let me know?

illume, line 1810: Rather than passing in all of config... it would be clearer if it took contexts as an argument. Then it's not tightly coupled to the whole of the config object.

Please document the parameters as well? From the current documentation it's not clear what role contextName plays?

illume (changes requested):
Hi. I left some notes for some improvements that could be done.

frontend/src/stateless/index.ts
```diff
@@ -224,11 +224,15 @@ export function getStatelessClusterKubeConfigs(): Promise<string[]> {
 /**
  * Finds a kubeconfig by cluster name.
  * @param clusterName
+ * @param clusterID
```
illume, line 227: Please document this field. Please see note about refactoring this function to add source as well.

Also... maybe the name should not be findKubeconfigByClusterName when it can find it by clusterID now as well. Just `findKubeconfig`?

backend/cmd/headlamp.go
illume, line 1810: > used for checking cluster names.

I think it's important that we use precise terms here.

Clusters and contexts are different things. We mix the terms up in the frontend UI, but in the backend code the terms used should be precise.

This looks at the contexts names not the cluster names.

backend/cmd/headlamp.go
```diff
@@ -1744,40 +1756,99 @@ func (c *HeadlampConfig) renameCluster(w http.ResponseWriter, r *http.Request) {
...
+				contextNames = append(contextNames, customObj.CustomName)
+			}
+		}
 	}
 
-	w.WriteHeader(http.StatusCreated)
-	c.getConfig(w, r)
+	for _, current := range contextNames {
+		if current == newName {
+			logger.Log(
```
illume, line 1838: Is this log necessary? I'm not sure what value it would add to the operator.

Logs should usually log exceptional circumstances, not expected circumstances. Here checking for a unique context name is normal... so no log required.

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -152,17 +152,34 @@ export function getClusterDefaultNamespace(cluster: string, checkSettings?: bool
...
+  source: string,
+  clusterID?: string
+) {
   let stateless = false;
+  let kubeconfig;
+  let renameURL = `/cluster/${cluster}`;
+
   if (cluster) {
-    const kubeconfig = await findKubeconfigByClusterName(cluster);
+    if (source === 'kubeconfig') {
```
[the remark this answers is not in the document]
vyncent-t: I agree it could be smaller, but this is something I may need to come back to later as this PR is meant for clusterID + non dynamic clusters first (in regards to fixing the duplicate name issue for non-dynamic clusters, and setting up the foundation for clusterID to be used for the non-dynamic cluster delete function PR)

Right now, I am only changing the needed parts for the non-dynamic cluster logic to work while leaving the default code as is, since it already handles dynamic/stateless clusters as seen in main currently. 

I think it is too early for refactoring the original findKubeconfigByClusterName to now _**always**_ use a clusterID, even if undefined (as opposed to how it is now: don't use it if we don't have it, stay in this lane). 

I would like to have a working version of this new clusterID logic for renaming and deleting non dynamic clusters finished, then later we refactor this to also make all clusters use clusterID for their delete / rename actions.

frontend/src/stateless/index.ts
```diff
@@ -257,15 +261,28 @@ export function findKubeconfigByClusterName(clusterName: string): Promise<string
...
-            const matchingContext = parsedKubeconfig.contexts.find(
-              context =>
-                context.context.extensions?.find(extension => extension.name === 'headlamp_info')
-                  ?.extension.customName === clusterName
-            );
 
-            const matchingKubeconfig = parsedKubeconfig.contexts.find(
-              context => context.name === clusterName
-            );
+            let matchingContext;
```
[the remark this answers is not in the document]
vyncent-t: `All users of findKubeconfigByClusterName first check source, so I think that logic should be centralized.`
- This is intentional as these changes bring `source` and `clusterID` from the metadata but only after it is created via interactions in the frontend (having a getClusters call work) 

The current idea here is for dynamic clusters and non dynamic clusters to have their own lane, essentially, "if working with non-dynamic clusters = we use clusterID, for all others = we don't" 

We check to see if that metadata has been created, and then use the source to determine if we are working with a non dynamic cluster, then we continue with the clusterID as a param if we are. We do not want to use clusterID if we don't have it and if we don't have it, assume its dynamic.

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: I have it by itself so we can add more helper texts later

backend/cmd/headlamp.go
vyncent-t: I don't think so, I have tried moving it to kubeconfig pkg but ran into issues with importing the marshal object func since its part of cmd as main apparently

backend/cmd/headlamp.go
vyncent-t: test improvements are being addressed in a different PR for this tool func as its tied to a fuzz test I am working on simultaneously https://github.com/kubernetes-sigs/headlamp/pull/3213

vyncent-t: test improvements and refactors are being addressed in a different PR for this tool func as its tied to a fuzz test I am working on simultaneously https://github.com/kubernetes-sigs/headlamp/pull/3213

backend/cmd/headlamp.go
vyncent-t: this was useful during testing but can remove

backend/cmd/headlamp.go
vyncent-t: fixed

backend/cmd/headlamp_test.go
vyncent-t: fixed

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -103,25 +103,61 @@ export default function SettingsCluster() {
...
-  const source = clusterInfo?.meta_data?.source || '';
+  const originalName = clusterInfo?.meta_data?.originalName;
+  const displayName = originalName || (clusterInfo ? clusterInfo.name : '');
+  const source = clusterInfo?.meta_data?.source;
+  // Note: display original name is currently only supported for non dynamic clusters from kubeconfig sources.
+  const clusterID = clusterInfo?.meta_data?.clusterID || '';
+
+  // This function is part of a double check, this is meant to check all the cluster names currently in use as display names
+  // Note: if the metadata is not available or does not load, another check is done in the backend to ensure the name is unique in its own config
+  function checkNameInUse(name: string) {
```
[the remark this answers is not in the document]
vyncent-t: fixed

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -289,6 +327,10 @@ export default function SettingsCluster() {
     "translation|Cluster name must contain only lowercase alphanumeric characters or '-', and must start and end with an alphanumeric character."
   );
 
+  const clusterNameInUseMessage = t(
```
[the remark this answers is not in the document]
vyncent-t: fixed

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -30,10 +30,28 @@ export interface ConfirmDialogProps extends MuiDialogProps {
...
   cancelLabel?: string;
   confirmLabel?: string;
+  /**
+   * Disables the Yes button, defaults to false
+   */
+  hideYesButton?: boolean;
+  /**
+   * Disables the No button, defaults to false
+   */
+  hideNoButton?: boolean;
```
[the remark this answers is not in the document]
vyncent-t: done

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -152,17 +152,34 @@ export function getClusterDefaultNamespace(cluster: string, checkSettings?: bool
  * is the custom name of the cluster used by the user.
  * @param cluster
  */
-export async function renameCluster(cluster: string, newClusterName: string, source: string) {
+export async function renameCluster(
+  cluster: string,
+  newClusterName: string,
+  source: string,
+  clusterID?: string
```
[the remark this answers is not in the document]
vyncent-t: done

frontend/src/stateless/index.ts
vyncent-t: added notes

function renaming this will be done in the refactor when clusterID logic is ready for all clusters (only non dynamic for now)

backend/pkg/kubeconfig/kubeconfig.go
vyncent-t: done

backend/pkg/kubeconfig/kubeconfig.go
vyncent-t: done

frontend/src/components/App/Settings/SettingsCluster.tsx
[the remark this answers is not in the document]
illume: This one isn't fixed... it's still assigned to a variable.

backend/pkg/kubeconfig/kubeconfig.go
illume: After reading this doc for the field I just noticed it looks like it is separated by ":". 

Did you know ":" is valid in files/paths in linux? (It's invalid in windows). Can we perhaps use some other separator or two fields?

backend/pkg/kubeconfig/kubeconfig_test.go
```diff
@@ -86,6 +87,27 @@ func TestLoadContextsFromKubeConfigFile(t *testing.T) {
 	})
 }
 
+// TestLoadContextFromFile validates the behavior of the LoadContextsFromFile function.
+//
+// This test ensures that the function correctly processes a valid kubeconfig file
+// and produces the expected metadata results without errors.
+func TestLoadContextFromFile(t *testing.T) {
```
illume, line 94: It looks like the second commit should go with the first one.

I mean this one with the test change "backend: kubeconfig_test.go: Add TestLoadContextFromFile test for new changes made to LoadContextFromFile" with the TestLoadContextFromFile test should just be in with the code that was changed. If you make a change to the code, the test change should go in the same commit.

backend/cmd/headlamp.go
```diff
@@ -1298,14 +1298,25 @@ func (c *HeadlampConfig) getClusters() []Cluster {
 			continue
 		}
 
+		kubeconfigPath := context.KubeConfigPath
```
illume, line 1301: The commit message is pretty nice now.

But I think this one would be a bit better when reading it with `git log --oneline`:
`backend: headlamp.go: getCluster: Return origin, originalName, and clusterID`

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -327,6 +327,27 @@ export default function SettingsCluster() {
     );
   }
 
+  // Display the original name of the cluster if it was loaded from a kubeconfig file.
```
illume, line 395: The commit message is pretty good.

I think adding the component into the context part will make it easier to understand when reading with `git log --oneline`.

`frontend:  SettingsCluster: Add original name to cluster settings view for non dynamic clusters`

backend/pkg/kubeconfig/kubeconfig.go
vyncent-t: ah good idea, I will find a new symbol to work around

backend/cmd/headlamp.go
illume: Ok.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -327,6 +327,27 @@ export default function SettingsCluster() {
...
+    const originalName = clusterInfo?.meta_data?.originalName;
+    const source = clusterInfo?.meta_data?.source;
+    // Note: display original name is currently only supported for non dynamic clusters from kubeconfig sources.
+    const displayOriginalName = source === 'kubeconfig' && originalName;
+
+    return (
+      <>
+        <Typography>{t('translation|Name')}</Typography>
+        {displayOriginalName && (
+          <Typography variant="body2" color="textSecondary">
```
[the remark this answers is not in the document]
vyncent-t: ND clusters should be the only clusters with original names, so if we have one we should render it.

but the other clusters change their name differently, so there is no kept original name for them.

backend/cmd/headlamp_test.go
```diff
@@ -649,6 +649,58 @@ func TestHandleClusterAPI_XForwardedHost(t *testing.T) {
...
+	resetConfigByte, err := os.ReadFile("./headlamp_testdata/kubeconfig_rename")
+	require.NoError(t, err)
+
+	for _, tc := range tests {
+		handleClusterRenameRequest(t, handler, tc)
+	}
+
+	// this is a reset line that will reset the kubeconfig file to its original state
+	// without it the tests will fail as it makes changes to the kubeconfig
+	err = os.WriteFile("./headlamp_testdata/kubeconfig_rename", resetConfigByte, 0o600)
```
illume, line 698: I don't understand this yet... I'll come back to it.

backend/cmd/headlamp.go
illume: I can't see a test specifically for checkUniqueName in there either.

Can you please add a unit test for it here in this commit?

backend/cmd/headlamp.go
```diff
@@ -1615,7 +1615,7 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 	c.getConfig(w, r)
 }
 
-// Get path of kubeconfig from source.
+// Get path of kubeconfig we load headlamp with from source.
```
illume, line 1618: The commit message for this could be clearer when read with `git log --oneline`:

Something like this?
`backend: headlamp: Simplify cluster rename logic with unique name check`

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -40,8 +40,8 @@ import { isElectron } from '../../../helpers/isElectron';
 import { useCluster, useClustersConf } from '../../../lib/k8s';
 import { deleteCluster, parseKubeConfig, renameCluster } from '../../../lib/k8s/apiProxy';
 import { setConfig, setStatelessConfig } from '../../../redux/configSlice';
-import { findKubeconfigByClusterName, updateStatelessClusterKubeconfig } from '../../../stateless/';
-import { Link, Loader, NameValueTable, SectionBox } from '../../common';
+import { findKubeconfigByClusterName, updateStatelessClusterKubeconfig } from '../../../stateless';
```
illume, line 43: The commit message is very detailed, thank you.

However, I think the commit message makes it clear that this could be broken up into separate commits.

frontend/src/components/App/Settings/SettingsCluster.tsx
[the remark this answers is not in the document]
illume: Ok. Thanks.

frontend/src/components/App/Settings/SettingsCluster.tsx
illume, line 395: The commit message is pretty good.

I think adding the component into the context part will make it easier to understand when reading with git log --oneline.

frontend:  SettingsCluster: Add original name to cluster settings view for non dynamic clusters

frontend/src/components/common/ConfirmDialog.tsx
```diff
@@ -30,10 +30,23 @@ export interface ConfirmDialogProps extends MuiDialogProps {
   handleClose: () => void;
   cancelLabel?: string;
   confirmLabel?: string;
+  /**
+   * Disables the No button, defaults to false
```
illume, line 34: > Disables the No button, defaults to false

Should be:

> Disables the Cancel button, defaults to false

frontend/src/lib/k8s/kubeconfig.ts
illume: thanks

backend/cmd/headlamp_test.go
illume: thanks

backend/cmd/headlamp_test.go
illume: thanks

backend/cmd/headlamp.go
illume: thanks

frontend/src/stateless/index.ts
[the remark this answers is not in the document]
illume: What I mean is that if this logic is moved inside the function, then the multiple callers of the function don't need to take care of this.

Putting the logic inside the function makes the call sites of the function easier to read. It also makes the code easier to reason about and refactor. Because each caller doesn't need to take of that logic in multiple places.

frontend/src/stateless/index.ts
illume: It looks like this field wasn't documented after all?

> `* @param clusterID`

illume (changes requested):
Thanks. You made a lot of improvements.

- I resolved all the notes which were done. A few weren't done after all.
- A few you pushed back and I resolved, other ones I left further notes onto why I think it's important to do the requested change.
- Left notes on some improvements that can be done on the commit messages. Please view commit by commit so those notes make sense.

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: removed

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: done

frontend/src/stateless/index.ts
vyncent-t: fixed this time

frontend/src/stateless/index.ts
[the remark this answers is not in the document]
vyncent-t: Not sure about using the source of a context before actually finding it in `findMatchingContexts({ clusterID, source, clusterName, contexts})` but I have tried making some changes that will try to fit what you're looking for while keeping the clusterID logic in its own lane

backend/pkg/kubeconfig/kubeconfig_test.go
vyncent-t: done

backend/cmd/headlamp.go
vyncent-t: done

backend/cmd/headlamp_test.go
vyncent-t: added note

backend/cmd/headlamp.go
vyncent-t: done

backend/cmd/headlamp.go
vyncent-t: added test

frontend/src/components/App/Settings/SettingsCluster.tsx
vyncent-t: split them

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -103,25 +103,61 @@ export default function SettingsCluster() {
...
+  }
 
   const handleUpdateClusterName = (source: string) => {
     try {
-      renameCluster(cluster || '', newClusterName, source)
+      renameCluster(cluster || '', newClusterName, source, clusterID)
         .then(async config => {
           if (cluster) {
-            const kubeconfig = await findKubeconfigByClusterName(cluster);
+            const kubeconfig =
```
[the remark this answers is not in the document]
vyncent-t: found another way for source handling that fits similar workflow

frontend/src/lib/k8s/api/v1/clusterApi.ts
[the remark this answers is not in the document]
vyncent-t: I was under the impression this comment was leaning towards using clusterID being used for all clusters and not just ND clusters. ( as in a refactor **_and_** reimplementation for how every `findKubeconfigByClusterName` func is being used, hence my reluctance )

I have come up with something that cleans up the `source` related logic while keeping the same flow of using clusterID for non dynamic clusters only.

illume:
@vyncent-t there's a conflict now with some recent changes merged into main.

illume (approved):
Thanks.

@skoeva would you mind doing a final review?

skoeva (approved):
looks good! could use one last rebase

vyncent-t:
asked to split in multi PRs
- [ ] metadata improvements + clusterID PR
- [ ] display actual origin PR
- [ ] fix rename issue PR

vyncent-t:
> asked to split in multi PRs
> 
>     * [x]  metadata improvements + clusterID PR [Backend: Improve cluster metadata and add clusterID #3304](https://github.com/kubernetes-sigs/headlamp/pull/3304)
> 
>     * [x]  display actual origin PR
> 
>     * [x]  fix rename issue PR

done closing old branch

PR #3122, opened 2025-04-15
app: home: Add delete button for non dynamic clusters

### LINKED ISSUE #1418: Allow to delete any clusters when using Headlamp (desktop version)

vyncent-t:
original - https://github.com/kubernetes-sigs/headlamp/pull/2567

## Description

Fixes #1418 

This PR updates the cluster deletion functionality in Headlamp to allow users to delete non-dynamic clusters directly from the home page. Previously, only clusters added through Headlamp could be deleted, which led to confusion for users attempting to delete clusters originating from other sources (e.g., kubeconfig). 

To address this, the deletion process has been updated to clearly inform users about the impact of deleting a cluster, particularly when the cluster originates from non-Headlamp sources.

---

## Images
image

## Changes Made

1. Updated the cluster deletion functionality to allow deletion of non-dynamic clusters, including those originating from kubeconfig.
2. Enhanced the deletion confirmation dialog to inform users about the origin of the cluster and to clarify that deleting non-dynamic clusters will also remove them from the kubeconfig file.

## Steps to Test

1. Open the home view in Headlamp.
2. Add clusters from both Headlamp and kubeconfig.
3. Attempt to delete clusters:
   - Confirm the dialog accurately informs users about the cluster’s origin.
   - Verify that deleting non-dynamic clusters removes them from kubeconfig.
   - Verify that deleting Headlamp-added clusters functions as expected.

## Notes

- This PR ensures that users have a clear understanding of the impact of deleting clusters.
- It improves usability by addressing feedback received about the cluster deletion restrictions in earlier versions.


skoeva (commented):
Is this PR still necessary?

vyncent-t:
> Is this PR still necessary?

it was merged this week, closing this

Issue #3052, opened 2025-03-28
Log viewer buttons are not aligned


vyncent-t:
## Describe the bug
Currently, the top buttons for the log viewer for both pods and other resources are misaligned.

It is more noticeable on log viewers that have more options, and also when resizing the window to a smaller view.

Deployments log viewer - normal
- the switches are not aligned
![Image](https://github.com/user-attachments/assets/a1da70ee-1c85-470f-aae3-db18c0056c10)

Deployment log viewer - small
- the buttons are not dynamic and the side buttons eat into the toggles
![Image](https://github.com/user-attachments/assets/081bab00-0be8-4460-876a-ab10b6d86316)

Pods log viewer - normal

![Image](https://github.com/user-attachments/assets/ad8b0751-958b-4b6a-8fac-cf6c7ea9cc20)

Pods log viewer - small
- buttons are not in formatted
![Image](https://github.com/user-attachments/assets/ef17730e-dc44-414f-8f93-f9ea112201ae)

## To Reproduce
Steps to reproduce the bug:
1. Go to pods list and click into a pod details view
2. Open the view logs and resize

Steps to reproduce the bug:
1. Go to deployments list and click into a deployments details view
2. Open the view logs and resize

## Are you able to fix this issue?
Yes, currently working on a fix for this in this PR https://github.com/headlamp-k8s/headlamp/pull/3027

PR #3124, opened 2025-04-15
[NEW] frontend: Add toggle wrap for logViewer and fix log viewer styles

vyncent-t:
Fork PR from original: https://github.com/kubernetes-sigs/headlamp/pull/3027

requested feature from https://github.com/headlamp-k8s/headlamp/issues/2657

This PR adds a toggle button for the Logs Viewer window that allows the text to be formatted for word wrap. There are also style formatting issues that I have noticed and made an issue for here https://github.com/headlamp-k8s/headlamp/issues/3052  this PR also adds changes that style error while also. 

## Changes
image
- **Introduces new Wrap option for log viewer component to allow toggle between default text formatting and word wrap**

image

- **Introduces new format and tooltip context for log viewer buttons** 
  - At current, adding additional switches or buttons to the log viewer components for both pod and non pod log viewer introduces styling errors (this PR makes it so that we can continue to add more features by reworking the button layout)

- **FIxes rendering for small view**
  - At current when viewing log viewer on smaller windows the buttons are both misaligned and not styled correctly https://github.com/headlamp-k8s/headlamp/issues/3052

### How to test

- Navigate to pods
- Click on any pod
- Open the "View logs" button
- Click the toggle "Wrap" button


frontend/src/components/common/LogViewer.tsx
```diff
@@ -26,6 +26,38 @@ export interface LogViewerProps extends DialogProps {
    * @description This is a boolean that determines whether the reconnect button should be shown or not.
    */
   showReconnectButton?: boolean;
+  isWrappedTextEnabled?: boolean;
+}
+
+// // this function manually formats the logs to a single line.
```
joaquimrocha, line 32: Why two // ?

frontend/src/components/common/LogViewer.tsx
```diff
@@ -26,6 +26,38 @@ export interface LogViewerProps extends DialogProps {
...
+// // currently we must manually format the texts as addonFit does not work with multiline texts.
```
joaquimrocha, line 33: Can you elaborate on this? I don't understand exactly why this function is needed. Are you sure there's no way to delegate this to the xterm component?

e2e-tests/tests/headlampPage.ts
joaquimrocha, line 1: Is this commit related to the logs?

joaquimrocha (changes requested):
Besides my comments, I noticed that when we wrap the text, the vertical scroll bar is now pushed beyond the limits of the dialog, meaning we lose the ability to understand where we are vertically.

frontend/src/components/common/LogViewer.tsx
vyncent-t: I've moved on from using landmarks like "notez" for project wide search to find notes to self, forgot to clean this up, need to make this a /* comment with a note:

frontend/src/components/common/LogViewer.tsx
vyncent-t: I have tried using options for single line printing but it had a few issues.

the format to single line is needed since the log "lines" are not actually single line text strings but rather a bundle of log lines, which I found the terminal has trouble formatting it correctly. It also dynamically sets the width depending on the longest singular line, which that value I use to fix the width sizing.

image

(while the split new line \n is read properly by the console log, in the xterm it was not)
logs w/no timestamp

image

e2e-tests/tests/headlampPage.ts
vyncent-t: Not directly, it was needed for the e2e tests to pass initially but I am working on a fix for this in a e2e test related PR
https://github.com/kubernetes-sigs/headlamp/pull/3144 

this branch (along with a few others) have been seeing some e2e test failures mostly from the new multi cluster changes.

I was work shopping that potential fix and testing to see if the test failures were due to changes made for the logs (they were not, I introduced a fix into that PR, once that is merged in I can rebase and drop it from this branch)

vyncent-t: actually, I will drop it now, just need to see if the e2e pass again

frontend/src/components/common/LogViewer.tsx
vyncent-t: side note: I was responding in the context of you meaning "delegate" to the xterm comp as in using the built in options to handle the single line format printing (changing width, using addonFit, etc.) but if you mean in terms of moving this function from a stand alone into the main log viewer component I moved it there as well.

vyncent-t:
e2e tests are passing, none of the log changes cause it to fail. going to drop e2e test changes here and rebase when https://github.com/kubernetes-sigs/headlamp/pull/3144 is merged

frontend/src/components/common/LogViewer.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it('ignores literal "\\n" inside a line', () => {
+    const logLines = ['foo\\nbar', 'longerstring'];
+    expect(findLongestLine(logLines)).toBe('longerstring'.length + 1);
+  });
+
+  it('returns 0 for empty input', () => {
+    expect(findLongestLine([])).toBe(0);
+  });
```
sniok, line 23: please add a test case like `["abcde", "egf\nabc"]`, I think the current implementation doesn't handle a case like that

vyncent-t: I think 'splits on real LF (\\n) for a line' covers it since the new version of the findLongestLine uses `.split(/\n|\n|\r/)` so should split as needed but I added an another one with ["abcde", "egf\nabc"]

frontend/src/components/common/LogViewer.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  it('handles simple lines', () => {
+    expect(findLongestLine(['one', 'three'])).toBe('three'.length + 1);
+  });
+
+  it('splits on real LF (\\n) for a line', () => {
+    expect(findLongestLine(['a\nbbb\nc'])).toBe(3 + 1);
+  });
+
+  it('splits on real LF (\\n) for a line - longer', () => {
+    expect(findLongestLine(['abcde', 'egf\nabc'])).toBe(5 + 1);
```
sniok, line 13: the longest line in this case is `abcdeefg` and is 7 characters long

vyncent-t: there isn't any joining after the function finds a new line or return syntax, it treats every "main string" (bundle of logs in live) as a collection that needs to be split up into parts and substrings and we count off those smaller parts

the inputs of ['abcde', 'egf\nabc'] are broken into ['abcde', 'egf' abc']

we then add a +1 on every line for the `longest` to make up for splitting the \n (it also makes sure the width is set to exactly the number of chars +1 to ensure a fit every time)

so the function should read 
`abcde` as 6 long
`efg` as 4 long
`abc` as 4 long

sniok: yes I understand that currently the function works this way, but what I'm saying is that the function needs to be adjusted so it passes this test since it something we might encounter and it should properly count the line length

sniok (changes requested):
Wrap toggle is inverted, it wraps lines when the toggle is off

image

There's a lot of extra space at the bottom 
image

All the toggles can should be placed here on wide screens
image

vyncent-t:
> Wrap toggle is inverted, it wraps lines when the toggle is off

> All the toggles can should be placed here on wide screens image

- [x] fixed toggle inverse, since the logs currently load in in wrapped format I have fixed it to start as "Wrap lines" enabled
- [ ] WIP: fixing the extra space
- [x] Need feedback: All toggles and drop downs on one line in wide view
  - I have addressed a few issues in the current format where all toggles and drop downs are all on the same line in this issue https://github.com/headlamp-k8s/headlamp/issues/3052 . The changes I have currently not only fix the rendering issues, but also make things more future proof for adding more inputs later. I noticed that pods and deployments log viewer are different so by separating the drop downs above the toggles we can keep the look uniform.
  
**Pod view logs**
image

**Other view logs**
image

Notice that by putting them all on one line here, we start to collide into the side icon buttons, preventing us from adding another button on the same line in the future.

vyncent-t:
new issues
- drawer mode: background does not flex fit when clicking one log, turning off wrap mode, then clicking another pod log to view it

vyncent-t:
> blocked here due to new issues:
 
* [ ]  scroll issues, default Y scroll works on some pod logs and not others, some pod logs scrolling also no longer accurate? (was working before, works as intended but only when wrap lines is toggled off?)
 * [ ]  cannot format the dead space while also having both scroll bars visible

new issues
 - [ ] drawer mode: background does not flex fit when clicking one log, turning off wrap mode, then clicking another pod log to view it

vyncent-t:
> > Wrap toggle is inverted, it wraps lines when the toggle is off
> 
> > All the toggles can should be placed here on wide screens image
> 
> * [x]  fixed toggle inverse, since the logs currently load in in wrapped format I have fixed it to start as "Wrap lines" enabled
> * [ ]  WIP: fixing the extra space
> * [x]  Need feedback: All toggles and drop downs on one line in wide view
>   
>   * I have addressed a few issues in the current format where all toggles and drop downs are all on the same line in this issue [Log viewer buttons are not aligned #3052](https://github.com/kubernetes-sigs/headlamp/issues/3052) . The changes I have currently not only fix the rendering issues, but also make things more future proof for adding more inputs later. I noticed that pods and deployments log viewer are different so by separating the drop downs above the toggles we can keep the look uniform.
> 
> **Pod view logs** image
> 
> **Other view logs** image
> 
> Notice that by putting them all on one line here, we start to collide into the side icon buttons, preventing us from adding another button on the same line in the future.



made some changes and I am now using grid for the space filling, will keep the non pod log viewer the same style of stacking since it does not have the space for it

ref: just adding as a visual ref for

image

vyncent-t:
last push:

- [x] moved to mix of grid and box for new display

vyncent-t:
May need to settle with the bottom spacing imperfections when not using wrap texts (toggling wrap text off) for some log windows.

This is only an issue for logs that do not contain enough lines to fill out the rest of the spacing when reformatting to unwrap.

I have also tried experimenting with creating a custom scroll bar just for this view and that introduces too many issues for what it is trying to accomplish (having to sync it to the 'real' scroll bar, having format issues with other switches due to needing nested boxes, etc.) and will be abandoning that idea. 

Although I have been able to succeed in making the unwrap view work along with the Prettify option where it preserves the JSON readable format but with unwrapped lines.

sniok:
xterm seems to be way limited, I think we should consider some alternative instead, that's more geared towards displaying logs instead of terminal emulator

PR #3125, opened 2025-04-15
frontend: Sidebar: Use theme color for Add Cluster Button

skoeva:
This change updates the color of the Add Cluster button based on the currently used theme.

### Testing
- [X] Open the General Settings in Headlamp and change the theme
- [X] Notice that the 'Add Cluster' button in the side is visible and the color matches the other items


sniok (approved):
Good catch, thanks!

PR #3126, opened 2025-04-16
Review Japanese Translations

shu-mutou:
Also, added remaining few translations.


skoeva (commented):
Looks good, don't forget to run the linter (`make frontend-test`) and make sure the rest of the translations are up to date (`make i18n`)

skoeva (commented):
Squash into an atomic commit and should be good to go 😄

PR #3128, opened 2025-04-16
backend: add support for skip context

warjiang:
for the default use-case, headlamp will load kubeconfig with `kubeconfig` options, and all context in the `kubeconfig` files will be automatically loaded, but sometimes there are some needs to hide some context across the kubeconfig file. So I want to add the feature to skip the contexts according to `skipped-kube-contexts`, it will be used like 
```
headlamp -kubeconfig=${kubeconfig1}:${kubeconfig2}
-html-static-dir=${path/to/static/bundles}
-plugins-dir=${path/to/plugins}
-dev
-enable-dynamic-clusters
-skipped-kube-contexts=ctx1,ctx2,ctx3
```

### DIFF
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -910,26 +910,27 @@ func GetInClusterContext(oidcIssuerURL string,
 	}, nil
 }
 
-// func type for context filter
-// if the func return true, the context will be skipped
-// otherwise the context will be used
+// Func type for context filter, if the func return true,
+// the context will be skipped otherwise the context will be used.
 type shouldBeSkippedFunc = func(kubeContext Context) bool
 
-// AllowAllKubeContext will keep all contexts
+// AllowAllKubeContext will keep all contexts.
 func AllowAllKubeContext(_ Context) bool {
 	return false
 }
 
-// SkipKubeContextInCommaSeperatedString will skip the contexts
+// SkipKubeContextInCommaSeparatedString will skip the contexts
 // whose names are in the given comma-separated string.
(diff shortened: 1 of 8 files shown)


illume (changes requested):
Thank you! This is quite useful.

One nit:
For the git commit message subject line would you mind changing them to this format? We follow the kernel style messages with the module as context and a capitalized verb at the beginning.

`backend: kubeconfig: Implement context filter`
`backend: Use context filter`

I hope to test it out and give it an in detail review tomorrow. Thanks again.

illume (approved):
👍 thanks

backend/pkg/kubeconfig/kubeconfig.go
```diff
@@ -910,20 +910,58 @@ func GetInClusterContext(oidcIssuerURL string,
...
+// AllowAllKubeContext will keep all contexts.
+func AllowAllKubeContext(_ Context) bool {
+	return false
+}
+
+// SkipKubeContextInCommaSeperatedString will skip the contexts
+// whose names are in the given comma-separated string.
+// For example, if pass the "a,b" for blackKubeContextNameStr,
+// the contexts named "a" and "b" will be skipped.
+func SkipKubeContextInCommaSeperatedString(blackKubeContextNameStr string) shouldBeSkippedFunc {
```
illume, line 926: There's a typo in SkipKubeContextInCommaSeperatedString. It should be Separated not Seperated.

warjiang: thanks for your carefulness, updated ~

illume (approved):

illume:
thanks!

PR #3129, opened 2025-04-16
frontend: helpers: Extract into own modules to reduce cycles

illume:
This separates all the functionality out of the helpers/index.ts file. Into separate files, so that each user only imports the part they want, rather than importing helpers.

The helpers module also did a mistake when it was made doing exports in a really weird way... I think because it somehow prevented issues with import cycles.

Import cycles:
- Makes everything a bit fragile in headlamp. Change something, and it breaks somewhere else.
- Theoretically makes everything a little bit quicker (but didn't measure it yet).
- Non deterministic loading order can cause code to run in a different order potentially triggering bugs.

---

Note: The helpers/index.ts is still there for backwards compatibility with plugins, but I'm 90% sure it could be removed. So the file will remain for now, and after some more investigation will be removed in a later PR.

### How to test

It should work exactly as before.

### Testing done

- Smoke testing in dev mode. Used the app, tried different things, clicked about.
- built the app, and ran that


frontend/src/helpers/clusterSettings.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Cluster settings
+ * This file contains functions to store and load cluster settings in local storage.
+ */
+export interface ClusterSettings {
```
knrt10, line 5: This looks like a comment for the interface, maybe add different comment or just add a new line before it? WDYT?

frontend/src/helpers/clusterSettings.ts
```diff
@@ -0,0 +1,39 @@
...
+  /** Optional default namespace to be used */
+  defaultNamespace?: string;
+  /** Only allow namespaces in this list to be selected */
+  allowedNamespaces?: string[];
+  currentName?: string;
```
knrt10, line 10: Can you please add for this also? Thanks

frontend/src/helpers/isDockerDesktop.ts
```diff
@@ -0,0 +1,18 @@
+declare global {
```
knrt10, line 1: Can you please add comment for this? Thanks

frontend/src/helpers/getBaseUrl.ts
```diff
@@ -0,0 +1,29 @@
+import { isElectron } from './isElectron';
+
+declare global {
```
knrt10, line 3: Same, can you please add comment?

knrt10 (commented):
Awesome work @illume 🥳 . Small NITs, rest LGTM. Tested as well, everything seems to work just fine.

frontend/src/helpers/clusterSettings.ts
illume: Fixed.

frontend/src/helpers/clusterSettings.ts
illume: done

frontend/src/helpers/isDockerDesktop.ts
illume: done

frontend/src/helpers/getBaseUrl.ts
illume: done

knrt10 (approved):

frontend/src/plugin/__snapshots__/pluginLib.snapshot
```diff
@@ -147,7 +147,9 @@
     "ValueLabel": [Function],
     "ViewButton": [Function],
     "ViewDialog": [Function],
+    "addQuery": [Function],
     "addResourceTableColumnsProcessor": [Function],
+    "getFilterValueByNameFromURL": [Function],
```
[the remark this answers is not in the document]
illume: fixed, thanks.

knrt10 (approved):
+1

PR #3131, opened 2025-04-16
app: frontend: plugins: Modify exec and run commands

skoeva:
This change prevents the shell from interpreting special characters in the arguments.

Follow-up from #3020


illume (approved):
🎉 thanks

Issue #3110, opened 2025-04-14
Query params are not kept on refresh for Map page in cluster deployment


cian-sheehy:
## Describe the bug
If you have a namespace selected in headlamp map page and if you refresh the page your namespace selection will be lost. This is not an issue on other pages like pods, workloads etc.

## To Reproduce
Steps to reproduce the bug:
1. Install headlamp in cluster
2. Navigate to Map 
3. Select namespace
4. Refresh page
5. Namespace is not kept from query param

## Environment (please provide info about your environment):
 - Installation type: In-Cluster
 - Headlamp Version: 0.30.1

## Are you able to fix this issue?
No
> Do you want to help fix this issue? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional Context

![Image](https://github.com/user-attachments/assets/ee5077f7-27f1-4004-bbfe-827546b12eaf)

![Image](https://github.com/user-attachments/assets/c08e915c-89f9-4c65-b419-b0b45c1a15ae)

skoeva:
I can take a look at this

PR #3138, opened 2025-04-17
frontend: Map: Persist query params on refresh

skoeva:
This change adds a queryParamsState for the namespace filter, allowing the query parameters to persist on refresh when they previously did not.

Fixes: #3110 

### Testing
- [X] Open the map in Headlamp
- [X] Filter by 2+ namespaces
- [X] Refresh and ensure that the query params persist


illume (approved):
👍 thanks

PR #3141, opened 2025-04-18
frontend: Include pod status in column text value to avoid stale status label state

sniok:
Fixes https://github.com/kubernetes-sigs/headlamp/issues/3108

Table cell wasn't re-rendering because the "reason" label stayed the same but status (which impacts color and the icon) changed but wasn't taken into account in the getValue function

How to test (steps from reported issue):

Go to pod list
Delete a pod
Wait for recreation
Check status color


skoeva (approved):
Looks good! could maybe use one last rebase

PR #3142, opened 2025-04-18
Improve WebSocket compatibility

sniok:
Fixes https://github.com/kubernetes-sigs/headlamp/issues/2970

Certain k8s vendors don't support tokens via Sec-Websocket-Protocol value, so we need to use the Authorization header instead.
Certain k8s vendors will perform additional cors check when browser-like User-Agent is present, which we currently pass but now will just remove.

How to test:

Create rancher cluster
Open headlamp and check for websocket connections (live updates to lists, logs and attach features)


knrt10 (approved):
Awesome

illume (approved):
🎉 nice work. Thanks

PR #3123, opened 2025-04-15
frontend: PluginSettings: Refactor local storage and plugin data

vyncent-t:
original - https://github.com/kubernetes-sigs/headlamp/pull/2671

Fixes Issue #2595

## Description

- This PR shrinks the saved JSON data from the plugins information and saves only the `name` and `isEnabled` status in local storage. This ensures plugin settings persist when closing and reopening the app.
- This PR introduces the `usePlugins` hook into `lib/k8s/api/v2`
- It also handles backward compatibility, allowing previously saved settings to be used in this new format.
- Additionally, it no longer takes old information from the JSON in local storage. Instead, it retrieves the necessary plugin data from the backend when the main Plugin component is called.

## Changes

1. **Introduces usePlugins hook**: Hook used to fetch the plugins from the backend and return the plugins with their settings
2. **Reduced Storage Data**: Only stores essential plugin info (`name` and `isEnabled`) in local storage.
3. **Backward Compatibility**: Converts old-format data into the new structure on first run.
4. **Data Source Shift**: Eliminates use of outdated local storage JSON data, and instead fetches plugin details from the backend when the Plugin component mounts.

## How to Test

1. **Setup**:
   - Pull this branch and run the application.
2. **Verify Plugin Settings**:
   - Open headlamp in app mode
   - Navigate to the settings page then to plugins settings tab
   - Check local storage item `headlampPluginSettings`
3. **Check Backward Compatibility**:
   - Start with an older version of the app or start from main.
   - Navigate to the settings page then to plugins settings tab
   - Check local storage item `headlampPluginSettings`
   - Switch to this branch and run the app.
   - Confirm that previously stored plugin settings are recognized and converted correctly.
5. **Backend Data**:
   - Verify that after loading, the plugin details (other than `name` and `isEnabled`) are fetched from the backend and displayed correctly in the UI.
   - Switch plugins off and on and observe the changes in local storage and in app

## Notes

- If you have preexisting local storage data, this PR will convert it automatically. You can reset the local storage if you return to main and go to the same plugin settings paage.
- Double-check the logs to ensure no errors appear during the migration from old data to new data.

PR #3144, opened 2025-04-18
Fix e2e auth logic + add namespace a11y playwright tests

vyncent-t:
### Description

This PR improves accessibility testing coverage and reliability of Playwright E2E tests in the project.

Note: this also fixes some e2e test failures for my other branches, once this is merged in I will drop the duplicates in the other branches.

- https://github.com/kubernetes-sigs/headlamp/pull/3124
- https://github.com/kubernetes-sigs/headlamp/pull/3123

#### Accessibility Enhancements

- Adds `@axe-core/playwright` integration to the Namespaces Playwright test to perform automated **A11Y checks**.
- Introduces accessibility assertions to verify that no critical violations occur before and after creating a namespace.
- Fixes an issue detected by these tests: the **"Create" button lacked an accessible label**. This was resolved by adding an `aria-label` to ensure compliance with screen reader standards.

#### Navigation & Auth Logic Improvements

- Updates the **navigation and authentication logic** used in our Playwright E2E test setup.
(Previously, navigation during multi-cluster testing was **unreliable and prone to hanging or timing out** due to race conditions and cluster token loading delays.)
- The logic has been refactored to **work consistently in both CI and local testing environments**, with better handling of cluster contexts and token usage.

### Notes

- No issue was filed for this work — these improvements were identified during test development.
- More accessibility and stability improvements are planned in follow-up PRs to expand coverage across additional areas of the UI.

### DIFF
--- a/frontend/src/components/common/CreateResourceButton.tsx
+++ b/frontend/src/components/common/CreateResourceButton.tsx
@@ -36,6 +36,7 @@ export function CreateResourceButton(props: CreateResourceButtonProps) {
         errorMessage={errorMessage}
         onEditorChanged={() => setErrorMessage('')}
         title={t('translation|Create {{ name }}', { name })}
+        aria-label={t('translation|Create {{ name }}', { name })}
       />
     </AuthVisible>
   );
--- a/e2e-tests/package-lock.json
+++ b/e2e-tests/package-lock.json
@@ -9,11 +9,24 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
+        "@axe-core/playwright": "^4.10.1",
         "yaml": "^2.3.4"
       },
       "devDependencies": {
         "@playwright/test": "^1.42.1",
         "@types/node": "^20.12.2"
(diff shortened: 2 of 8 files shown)


frontend/src/components/common/CreateResourceButton.tsx
```diff
@@ -36,6 +36,7 @@ export function CreateResourceButton(props: CreateResourceButtonProps) {
         errorMessage={errorMessage}
         onEditorChanged={() => setErrorMessage('')}
         title={t('translation|Create {{ name }}', { name })}
+        aria-label={t('translation|Create resource button')}
```
illume, line 39: - Probably should be fixed inside EditorDialog to use the title.
- Using the same name as the title would be better, because it's more correct and we already have translations for it.

illume (changes requested):
Nice work.

I just requested one change to the aria-label to use the existing title translation.

illume (approved):
This is great. Thank you!

Issue #3136, opened 2025-04-17
Allow Users to Favorite Menu Items for Easier Access


nrv-96:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
Right now, it takes a bit too much time to get to the sections I use most in Headlamp. I find myself scrolling through the side menu a lot just to access the same components like Workloads, Service Accounts, and Services. It slows things down and feels a bit repetitive, especially when I’m switching between things often.

## Describe the solution you'd like
It would be super helpful to have a **Favorites menu** —something where I can pin or mark specific sections I use frequently. That way, I could jump to them instantly without having to scroll through the whole list every time. Even a small, collapsible section at the top of the sidebar would make a big difference.

## What users will benefit from this feature?
This would be useful for pretty much anyone using Headlamp regularly—developers, admins, or folks managing multiple clusters. Especially those of us using Docker Desktop, WSL setups, or doing lots of plugin dev work where we’re constantly jumping between the same views.

A favorites menu would be helpful for all Headlamp users, regardless of their specific setup. Whether someone is managing a single cluster or multiple environments, being able to quickly access frequently used views can significantly improve workflow and reduce time spent navigating.

## Are you able to implement this feature?
No, not at the moment.

> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context
Here's a screenshot of the current menu. A Favorites section somewhere near the top would save a ton of clicks and make the UI feel more personalized.

![Image](https://github.com/user-attachments/assets/f1c6f87f-00b3-421f-a0c0-897c0794c77f)

nrv-96:
@illume @warjiang @sniok @skoeva 

I kindly request that you consider adding a Favorites menu feature in Headlamp. This small improvement could have a huge impact on daily usability, making it easier and faster for users to access the components they work with the most.

I genuinely believe this would benefit the entire community, whether you're a developer, an admin, or just someone who appreciates efficiency.

Please consider it 🙏 – and thank you for all the great work you're doing with Headlamp!

warjiang:
maybe I can try it. I think I will use custom hooks with localstorage api to implement this feature, any suggestions?

warjiang:
[...]

PR #3150, opened 2025-04-22
frontend: sidebar: support favourite menu

warjiang:
In #3136 , I think there is need to allow people to favourite sidebar menu, avoding duplicate click, so this PR is about add feature for favouriting menu. 
Following is a gif of screen:
20250422085617_rec_

### DIFF
--- a/frontend/src/components/Sidebar/FavouriteSidebarItem.tsx
+++ b/frontend/src/components/Sidebar/FavouriteSidebarItem.tsx
@@ -1,0 +1,110 @@
+import Collapse from '@mui/material/Collapse';
+import List from '@mui/material/List';
+import ListItem from '@mui/material/ListItem';
+import React, { memo, useState } from 'react';
+import { generatePath } from 'react-router';
+import { Link as RouterLink, LinkProps as RouterLinkProps } from 'react-router-dom';
+import { getClusterPrefixedPath } from '../../lib/cluster';
+import { useCluster } from '../../lib/k8s';
+import { createRouteURL, getRoute } from '../../lib/router';
+import ListItemLink from './ListItemLink';
+import SidebarItem, { SidebarItemProps } from './SidebarItem';
+
+export const FavouriteSidebarItem = memo((props: SidebarItemProps) => {
+  const {
+    label,
+    name,
+    subtitle,
+    url = null,
+    search,
+    useClusterURL = false,
+    subList = [],
+    hasParent = false,
+    icon,
+    fullWidth = true,
+    hide,
+    ...other
+  } = props;
+  // isSelected
+  const cluster = useCluster();
+
+  let fullURL = url;
+  if (fullURL && useClusterURL && cluster) {
+    fullURL = generatePath(getClusterPrefixedPath(url), { cluster });
+  }
+
+  if (!fullURL) {
+    let routeName = name;
+    if (!getRoute(name)) {
+      routeName = subList.length > 0 ? subList[0].name : '';
+    }
+    fullURL = createRouteURL(routeName);
+  }
+
+  const [open, setOpen] = useState(false);
+
+  const renderLink = React.forwardRef<any, Omit<RouterLinkProps, 'to'>>((itemProps, ref) => (
+    <RouterLink
+      to={{ pathname: fullURL, search: search }}
+      ref={ref}
+      {...itemProps}
+      onClick={() => {
+        console.log('click me');
+        setOpen(!open);
+      }}
+    />
+  ));
+  console.log('open', open);
+  return hide ? null : (
+    <React.Fragment>
+      <ListItemLink
+        selected={open}
+        pathname={''}
+        primary={fullWidth ? label : ''}
+        icon={icon}
+        name={label}
+        sidebarName={name}
+        subtitle={subtitle}
+        search={search}
+        iconOnly={!fullWidth}
+        hasParent={hasParent}
+        fullWidth={fullWidth}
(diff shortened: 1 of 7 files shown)


warjiang:
all CI problems were finally solved ~

joaquimrocha:
Since this is a big design change, we want to evaluate this with the design team that works with us, so bear with us while this may take longer.

warjiang:
> Since this is a big design change, we want to evaluate this with the design team that works with us, so bear with us while this may take longer.

It's ok, if any changes need to discuss, maybe we can make a discusssion in form of online meeting or in github issue~

illume:
You get my +1. I really like the addition of favourites.

One concern raised was about "buttons on buttons" with the heart icon. The problem is that it's easy to make a mistake tapping or clicking on the heart... and activating the wrong button. The consequence in this case is not disastrous IMHO, but I think we can avoid it anyway...

Proposal: One possible fix for this issue could be to remove the heart from the sidebar. Move it into the main view.

image


@nrv-96 since you proposed the original issue, can you also share some thoughts?

nrv-96:
@illume I would like to suggest adding this feature to the sidebar. As you can see, Microsoft Azure itself has this functionality (please see the video), and many other sites also include similar features—like a "favorites" button—in the sidebar instead of on the main page.

For users like me, it's easier and more intuitive to access it from the sidebar. However, I don’t have any issue if you prefer to place it on the main page as well.

Either way, thank you so much for considering this useful feature and working on it. I really appreciate team's efforts.


https://github.com/user-attachments/assets/748e9714-b246-4024-9a47-b8484633c164

cc: @warjiang

nrv-96:
Hey All,

@illume @sniok @skoeva

Could you please check this PR with the team? It's been open for almost 15 days. I would really appreciate it if you could work on it, as I need this feature. I've been using Headlamp since I installed it and absolutely love it.

joaquimrocha:
Hi @nrv-96 , we are discussing this with design experts and should have an idea on how to move forward before the end of the week.

nrv-96:
Thanks @joaquimrocha, for the update. I hope this feature makes it into the next release.

warjiang:
> Proposal: One possible fix for this issue could be to remove the heart from the sidebar. Move it into the main view.

Maybe we can implement these two kinds of UI. In fact, favouriting button in the main content area is useful when the sidebar hidden, also favouriting button on sidebar button is more intuitive. 

> One concern raised was about "buttons on buttons" with the heart icon being an accessibility concern. The problem is that it's easy to make a mistake tapping or clicking on the heart... and activating the wrong button. The consequence in this case is not disastrous IMHO, but I think we can avoid it anyway. I can imagine some people won't be able to easily use the button-on-button heart on top of the sidebar link as it is at all.

For your concern, I think we can make it configurable, some options like:
- allow favourite on sidebar and main page
- only allow favourite on sidebar
- only allow favourte on main page
- disable favourite
(BTW, the option name maybe not very good, if you have better idea, plz let me know that).


If we still have a lot concern about the feature, we can make favourite function disabled for default case, and if end-users want to enable it, just enable the favourite function.

nrv-96:
Hi @joaquimrocha,

Do we have any update on this?

joaquimrocha:
Hi @nrv-96 . Sorry, I wanted to update but it's been a busy week.
We talked to Design and cleared out the approach: we'd like to use your proposal and the current PR is very close to it but it does require deeper changes because of how the sidebar works:
1. We need to refactor the sidebar so we do not get the original + favorite entries opening at the same time; This is something we can work on in the core team;
2. The favorites' icon should be a star rather than a heart;

So this means we cannot get the feature into the next release, but we're definitely interested in moving this forward.

illume:
@warjiang Are you interested in iterating on the items mentioned here: https://github.com/kubernetes-sigs/headlamp/pull/3150#issuecomment-2880172832 ?

warjiang:
> @warjiang Are you interested in iterating on the items mentioned here: https://github.com/kubernetes-sigs/headlamp/pull/3150#issuecomment-2880172832 ?
> 
> 
> 
> What do you think of these ideas?

I'd like to iterate the feature, agree with change of favorite icon. For the case of original + favorite menu opening at the same time, I think it doesn't matters, if we still think we cannot open original + favourite  at the same time，maybe we can prepare one single
 page

nrv-96:
Hi @joaquimrocha @warjiang,

Do we have any update on this?

warjiang:
> Hi @joaquimrocha @warjiang,
> 
> Do we have any update on this?

sorry late for reply, a little busy recently, I'll move on this feature
- first, as described in discussion, I'll update the icon for favouriting action
- second, I'll provide UX in way of figma for deep discussion(once the ux be decides, the feature will be moved on faster)

nrv-96:
@warjiang thanks for the update.

nrv-96:
Hi @warjiang ,

Do we have any update on this?

nrv-96:
Hi @joaquimrocha @warjiang,

Do we have any update on this?

joaquimrocha:
Hi. @sniok in the Headlamp team has created a POC for a new sidebar completely which changes it's behavior and includes favorites. Right now it's paused while @sniok is busy with a higher priority task, but we'll get back to it, likely in August.
We know this is a very desirable requirement but we want to make sure we do it correctly and unfortunately we need the sidebar to be improved dor this to happen.

nrv-96:
Hi @joaquimrocha @warjiang,

Do we have any update on this?

PR #3151, opened 2025-04-22
frontend: Table: Add canSelect prop to row and cell components to fix memoization

sniok:
Table wouldn't render rows that can't be selected with a disabled checkbox. This pr fixes this issue.

How to test:

Add `enableRowSelection` prop to a Table
Verify that checkbox for those rows is disabled


illume (approved):
🎉 thanks!

PR #3153, opened 2025-04-22
plugins/examples/dynamic-clusters: Use theme color for buttons

skoeva:
This change applies the color of the button text for the dynamic-clusters plugin from the currently used theme.

### Testing
- [X] Run the dynamic-clusters plugin:

```
cd plugins/examples/dynamic-clusters && npm install
npm run build
npm run start
```
- [X] Change the theme and ensure that the text for the "New cluster" button and the "Submit" button in the dialog is visible


sniok (approved):
nice, thanks!

PR #3161, opened 2025-04-23
plugins/examples/custom-theme: Bump vite to 6.2.6

skoeva:
This change bumps vite in the custom-theme example plugin to v6.2.6.


illume (approved):
👍

Issue #3155, opened 2025-04-23
Improve theme picker in Settings


sniok:
Currently we just have a dropdown but we want to introduce something more like in the following sketch

1st row of themes are the default ones (we have now)
2nd row are registered by plugins

![Image](https://github.com/user-attachments/assets/b6bedda5-f79c-47a0-93d9-c981ca9e4a36)

AllanSlomski:
Hello! I'd like to work on this issue as my first contribution. Is it still available?

sniok:
> Hello! I'd like to work on this issue as my first contribution. Is it still available?

Yes, this is available, thanks for taking a look!

Exar04:
hey @AllanSlomski , are you still working on this?

Exar04:
<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/4ab03472-8a19-4295-92f9-dbc5f92a010c" />

something like this? @sniok

Exar04:
hey @AllanSlomski are you still working on this?

Exar04:
<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/9313a940-9a76-4c2e-a1a3-7a499cac9da6" />
@sniok you mean something like this?

AllanSlomski:
Hello! Yes, I'm currently working in this issue.

PR #3164, opened 2025-04-23
frontend: Settings: Replace theme dropdown menu with grid-based selector

AllanSlomski:
This PR replaces the theme <Select> dropdown with a grid-based selector.

The grid uses appThemes to render all themes visually, allowing users to click and apply them.

I noticed the base themes are defined in themeSlice.ts, but I couldn’t clearly identify how to distinguish them from plugin-registered themes. If there’s a recommended way to separate base and plugin themes visually, I’d love some guidance to follow up with a second PR or improvement.

headlamp

#3155

### DIFF
--- a/frontend/src/components/App/Settings/Settings.tsx
+++ b/frontend/src/components/App/Settings/Settings.tsx
@@ -15,9 +15,8 @@
  */
 
 import Box from '@mui/material/Box';
-import MenuItem from '@mui/material/MenuItem';
-import Select from '@mui/material/Select';
 import Switch from '@mui/material/Switch';
+import Typography from '@mui/material/Typography';
 import { capitalize } from 'lodash';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -29,6 +28,7 @@ import { useTypedSelector } from '../../../redux/reducers/reducers';
 import { uiSlice } from '../../../redux/uiSlice';
 import { ActionButton, NameValueTable, SectionBox } from '../../common';
 import TimezoneSelect from '../../common/TimezoneSelect';
+import { theme } from '../../TestHelpers/theme';
 import { setTheme, useAppThemes } from '../themeSlice';
 import DrawerModeSettings from './DrawerModeSettings';
 import { useSettings } from './hook';
@@ -99,29 +99,6 @@ export default function Settings() {
             value: <LocaleSelect showFullNames formControlProps={{ className: '' }} />,
           },
           {
-            name: t('translation|Theme'),
-            value: (
-              <Select
-                variant="outlined"
-                size="small"
-                defaultValue={themeName}
-                onChange={e => {
-                  dispatch(setTheme(e.target.value as string));
-                  console.log(e, e.target.value);
-                }}
-              >
-                {appThemes.map(it => (
-                  <MenuItem key={it.name} value={it.name}>
-                    <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
-                      <ThemePreview theme={it} />
-                      {capitalize(it.name)}
-                    </Box>
-                  </MenuItem>
-                ))}
-              </Select>
-            ),
-          },
-          {
             name: t('translation|Resource details view'),
             value: <DrawerModeSettings />,
           },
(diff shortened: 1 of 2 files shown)


frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -72,24 +72,36 @@ export default function Settings() {
...
                 }}
               >
                 {appThemes.map(it => (
-                  <MenuItem key={it.name} value={it.name}>
-                    <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
-                      <ThemePreview theme={it} />
-                      {capitalize(it.name)}
-                    </Box>
-                  </MenuItem>
+                  <Box
```
sniok, line 83: Using Box (which renders to a div) for an interactive element is not accessible, a better option for this element would be a Button

frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -72,24 +72,36 @@ export default function Settings() {
...
+                    key={it.name}
+                    sx={{
+                      cursor: 'pointer',
+                      border: themeName === it.name ? '2px solid' : '1px solid #ccc',
```
sniok, line 87: using hardcoded colors (#ccc) is not recommended, for a border color you could use `theme.palette.divider `

frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -72,24 +72,36 @@ export default function Settings() {
...
+                      borderRadius: 2,
+                      p: 1,
+                      display: 'flex',
+                      flexDirection: 'column',
+                      alignItems: 'center',
+                      transition: '0.2 ease',
+                      '&:hover': {
```
sniok, line 94: In headlamp we don't really use shadows for hover status, to be more consistent with the rest of the app I'd suggest instead changing background and/or border color

frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -72,24 +72,36 @@ export default function Settings() {
...
+                        boxShadow: '0 0 8px rgba(0,0,0,0.2)',
+                      },
+                    }}
+                    onClick={() => dispatch(setTheme(it.name))}
+                  >
+                    <ThemePreview theme={it} />
```
sniok, line 100: ThemePreview should be a bit bigger, you can use the size prop

sniok (changes requested):
Thanks for opening a PR! This is a good start! 
We had in mind to have a theme selector to be on the bottom of the settings page and take full width, something similar to a github theme selector

image

I've also left a couple of comments regarding some small issues

Appreciate you taking a look into it

AllanSlomski:
Thank you! I'm working in these changes now. I'll update the PR as soon as possible.

frontend/src/components/App/Settings/Settings.tsx
AllanSlomski: I initially tried using Button, but its default properties were overriding the custom styles and configurations I needed. Because of that, I opted to use Box with the role="button" instead. Do you think this is a reasonable approach, or would you recommend another way to handle this?

AllanSlomski:
I made all the changes you requested:

I used role="button" on the Box tag because using the Button component directly would override my styles and configurations.

I used the divider color for the border.

I removed the boxShadow and added a background color using divider as well.

I added a prop to change the ThemePreview's size.

The core change is the design: I moved the Theme Selector to the bottom of the page, as you suggested. To make it look better, I placed it outside the table — otherwise, I wouldn't be able to properly customize all the properties I wanted, especially to make the subtitle and theme cards look similar to GitHub's appearance page.

If there's anything else that needs adjustment, please let me know. Thank you!

frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -127,6 +104,83 @@ export default function Settings() {
...
+              <Box
+                key={it.name}
+                role="button"
+                tabIndex={0}
+                onKeyDown={e => {
+                  if (e.key === 'Enter' || e.key === ' ') dispatch(setTheme(it.name));
+                }}
+                sx={{
+                  cursor: 'pointer',
+                  border: themeName === it.name ? '3px solid' : '1px solid',
```
sniok, line 163: for "active" state let's use a primary color (theme.palette.primary) for the 1px border so it provides more contrast. In the current version it is a bit hard to see which item is selected.

sniok (changes requested):
Nice, thanks for the updates! I think this is very close to being ready. One thing that I think would improve the look a bit is if the items are aligned to the left instead of being stretched.

Also a snapshot test is failing but I think there's a bug that makes it produce different test results based on the OS. If you reverse changes to pluginLib.snapshot I think the tests will pass in CI.

frontend/src/components/App/Settings/Settings.tsx
sniok: yeah this is okay

AllanSlomski:
Thanks for the feedback! I'm already working on the changes you suggested, but I'm not sure if I understood the snapshot error. The snapshots are updated automatically when I run the tests. Should I replace it with the old one, from before my changes, and dont update before commiting?

sniok:
> Thanks for the feedback! I'm already working on the changes you suggested, but I'm not sure if I understood the snapshot error. The snapshots are updated automatically when I run the tests. Should I replace it with the old one, from before my changes, and dont update before commiting?

Since there's an issue with pluginLib.snapshot generation you should exclude changes to that file from the PR (replace it with the old one). All the other snapshots updates are okay

AllanSlomski:
Hello!
As you suggested, I changed the border colors and aligned the items to the left.
I made different breakpoints for smaller screens.
For the snapshot error, I copied the file from the original repository and replaced it, so it's now using the original version.
Additionally, there are some issues that I've noticed, and I'm putting screenshots of them here. 

Captura de Tela (135)
Captura de Tela (137)

sniok:
Looks great! @AllanSlomski thank you for the updates!
Just a couple of small things left, could you rebase your branch and make sure that license header is still present in frontend/src/components/App/Settings/Settings.tsx?
There's an empty package-lock in the root directory that got accidentally committed
And could you also squash all the changes into a single commit
There are still some changes in the pluginLib.snapshot, tests are not passing

sniok:
> About the snapshot errors: I tried copying the original file back into my code, but the tests are still failing. Could you please clarify what I should do to resolve this issue properly?

You could use `git checkout` to get the snapshot file from main branch before creating a commit. 
Since this feature is something that we want to include in the release pretty soon, would you mind if I took over this PR since there are just a couple of small changes left?

AllanSlomski:
Thanks!
I'll be commiting with the changes as soon as I can.

AllanSlomski:
Do you mind waiting for a few hours or just until tomorrow?

AllanSlomski:
Hello!

I believe I accidentally squashed a commit that wasn't mine — the one that updates the translation files. I'm really sorry about that.

Should I try to fix it, or would you prefer to handle it directly?

Thanks again for your help and feedback!

illume:
Thanks @AllanSlomski 🎉

PR #3168, opened 2025-04-24
frontend: Enable the multiplexer in the frontend

illume:
A bunch of issues were fixed, so we are trying again.


sniok (approved):

sniok:
I wonder if we should add an ability to turn it off in general settings. It would be easier to debug if any issues are related to websocket in general or specific to multiplexer

illume:
> I wonder if we should add an ability to turn it off in general settings. It would be easier to debug if any issues are related to websocket in general or specific to multiplexer

Yeah, could be!
(I guess we can see how it goes over the next week in the main branch.)

PR #3172, opened 2025-04-24
e2e: Fix authenticate token hang up and cluster navigation

vyncent-t:
since we now use a multi cluster structure to run our tests, the cluster navigation step must be its own full nav step, different from navigateToPage, if it is tied to an auth step.


vyncent-t:
not needed, closing

PR #3174, opened 2025-04-25
backend: Resolve issues with websocket token encoding

daviddob:
This resolves an issue introduced in #3142 when decoding the bearer token from the websocket Sec-Websocket-Protoccol it assumes the token is padded.

When testing with tokens generated for Azure AKS and GKE, I noticed headlamp started to fail all websocket connections and upon further investigation, the tokens that did not have padding (==) were not decoded properly and the authorization header was malformed when sent to K8s.

This attempts to decode the token additionally with RawStdEncoding to catch these edge cases.

How to test:
- Included test case for tokens without base64 padding
- Test using Azure AKS or GKE using token obtained via OIDC integration or `kubelogin` (depends on the token whether or not padding is necessary, results were intermittent prior to this patch).

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1209,8 +1209,15 @@ func processTokenProtocol(r *http.Request, protocol, tokenPrefix string) {
 	}
 
 	// Try to decode token from base64
-	if decodedBytes, err := base64.URLEncoding.DecodeString(token); err == nil {
+	decodedBytes, err := base64.URLEncoding.DecodeString(token)
+	if err == nil {
 		token = string(decodedBytes)
+	} else {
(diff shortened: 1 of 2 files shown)


illume:
Thank you very much for this!

The `make backend-lint` is failing in CI.
```go
  Error: backend/cmd/headlamp_test.go:1067:6: Function 'TestProcessWebSocketProtocolHeader' is too long (64 > 60) (funlen)
  func TestProcessWebSocketProtocolHeader(t *testing.T) {
       ^
```


Feel free to disable the issue if shortening it doesn't make sense.
```go
  //nolint:funlen
  func TestProcessWebSocketProtocolHeader(t *testing.T) {
```

illume (approved):
🎉 thanks!

PR #3176, opened 2025-04-25
github/workflows/push-release-assets: Limit token permissions

skoeva:
This change limits the permissions of the GITHUB_TOKEN as desired.


illume (approved):
I see the upload script uses a stored secret for uploading artifacts to the release.
```
      - name: Upload assets to release
        env:
          GITHUB_TOKEN: ${{ secrets.RELEASE_UPDATE_TOKEN }}
```

So I feel this change is an improvement.

Issue #2480, opened 2024-10-25
How to authenticate Headlamp with Azure AKS? Also tried using DEX as middleware.


Pree-byte:
I am able to login via Dex in Headlamp, but kubeconfig file is not getting generated, hence the cluster in not being loaded.

Can anyone please help me?

Sharing my config here: 

dex-values.yaml

image:
  tag: ""

config:
  issuer: 'https://dex.poc.xxxxx.net'
  storage:
    type: kubernetes
    config:
      inCluster: true
  web:
    http: 0.0.0.0:5556
    allowedOrigins: ['*']
  connectors:
    - type: microsoft
      id: microsoft
      name: Microsoft
      config:
        clientID: XXXXXXX
        clientSecret: XXXXXXXX
        tenant: organizations
        redirectURI: "https://dex.poc.XXXXt.net/callback"
        scopes:
          - openid
          - profile
          - email
  staticClients:
    - id: headlamp
      redirectURIs:
        - 'https://headlamp.poc.xxxx.net/oidc-callback'
      name: Headlamp Dashboard
      secret: XXXXXXXXXXXXX
  enablePasswordDB: true    
  oauth2:
    responseTypes: ["code", "token", "id_token"]
    skipApprovalScreen: true

headlamp-values.yaml

config:
  oidc:
    clientID: headlamp
    clientSecret: XXXXXXXXXXXXXXX
    issuerURL: 'https://dex.poc.xxxxx.net'
    scopes: "openid email profile groups"
    
    
    
![image](https://github.com/user-attachments/assets/e53aed35-8d25-49e0-ac09-4cc6d1cdd844) --> DEX dialog box

![image](https://github.com/user-attachments/assets/5959d4a1-65ca-402c-9765-d1b1e9f47585) --> 401 unauthorized, not loading anything.

![image](https://github.com/user-attachments/assets/56465b22-fbdf-4b10-a494-b22e4e2d81ba) -->headlamp pod logs

Please help me figure this out.

joaquimrocha:
Maybe @yolossn knows.

Pree-byte:
@yolossn, please help me here..

yolossn:
@Pree-byte  Can you check these two things and share what you see?
1. Open Network tab in Headlamp and copy the Bearer Token that is used in the failed requests. Use any JWT debugger like jwt.io to check if the scopes are properly added to the token that is used.
[...]

Issue #2848, opened 2025-02-05
I get redirected to OIDC auth pop up each time I click on any link/button inside headlamp UI. I'm using Azure AD for OIDC config. Headlamp installed in the cluster via helm


andreistefanzx:
## Describe the bug
I get redirected to OIDC auth pop up each time I click on any link/button inside headlamp UI. I'm using Azure AD for OIDC config. Headlamp installed in the Google GKE cluster via helm

## To Reproduce
Steps to reproduce the bug:
1. Go to main page, click Sign In and go through OIDC auth
2. After authentication succeeds you can see the headlamp UI but if you click anything there, you'll get redirected to the OIDC auth screen again

Note: 

- I don't get any logs while going through the auth step or after that
- Logs from headlamp pod:
```
k logs -l app.kubernetes.io/name=headlamp -f           
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":342,"time":"2025-02-05T05:59:17Z","message":"Plugins dir: /headlamp/plugins"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":343,"time":"2025-02-05T05:59:17Z","message":"Dynamic clusters support: false"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":344,"time":"2025-02-05T05:59:17Z","message":"Helm support: false"}
{"level":"info","source":"/headlamp/backend/cmd/headlamp.go","line":345,"time":"2025-02-05T05:59:17Z","message":"Proxy URLs: []"}
{"level":"info","pluginPath":"/headlamp/plugins/lost+found/main.js","source":"/headlamp/backend/pkg/plugins/plugins.go","line":197,"error":"stat /headlamp/plugins/lost+found/main.js: permission denied","time":"2025-02-05T05:59:17Z","message":"Not including plugin path, main.js not found"}
{"level":"info","context":"main","clusterURL":"https://K8S_API_IP:443","source":"/headlamp/backend/pkg/kubeconfig/kubeconfig.go","line":315,"time":"2025-02-05T05:59:17Z","message":"Proxy setup"}
*** Headlamp Server ***
  API Routers:
{"level":"error","source":"/headlamp/backend/cmd/headlamp.go","line":399,"error":"error loading kubeconfig files: error reading kubeconfig file: open : no such file or directory","time":"2025-02-05T05:59:17Z","message":"loading kubeconfig"}
[...]

Issue #2358, opened 2024-09-27
OIDC token times out too quickly on web client


sarg3nt:
We have Headlamp working with OIDC and Azure but are running into the issue of the token expiring far too quickly.  My Azure team tells me this is usually a setting in the app initiating the request.  

I also found closed ticket https://github.com/headlamp-k8s/headlamp/issues/209 where @joaquimrocha talks a little about this.

>  if you have access with your token throughout the navigation in Headlamp but the problem is apparently a timeout when using the token, maybe increasing the tokens' lifespan will help. In KeyCloak this is done in the settings of the client you are using, under "Advanced Settings > Access Token Lifespan".

But I do not know what he means by "the client"  Maybe the windows version of Headlamp? 

Is there a way to set this in the Helm chart?

joaquimrocha:
I think that comment was about KeyCloak. If you are using Azure, I think something different needs to be done.
Adding @ashu8912 and @yolossn should they know more here.

sarg3nt:
Hi all, any update on this?

sarg3nt:
@ashu8912, @yolossn and @joaquimrocha any update on this.  Is this something we need to investigate in our Azure setup or is this an issue in Headlamp?

yolossn:
@sarg3nt sorry for the late response. May I know if you are using Azure AKS OIDC setup working with Entra ID or AAD?

sarg3nt:
@yolossn Sorry for the late reply.
Not sure, here's the config.

`values.yaml`
```yaml
config:
  oidc:
    secret:
      create: false
    externalSecret:
      enabled: true
      name: "oidc"
```

`oidc_secrets.tf`

```tf
resource "kubernetes_secret" "headlamp_oidc_secret" {
  metadata {
    name      = "oidc"
    namespace = kubernetes_namespace.headlamp.metadata.0.name
  }

  data = {
    OIDC_CLIENT_ID     = var.azure_oidc.client_id
    OIDC_CLIENT_SECRET = var.azure_oidc_client_secret
    OIDC_ISSUER_URL    = var.azure_oidc_issuer_url
    OIDC_SCOPES        = "email profile"
    HTTP_PROXY         = var.proxy
    HTTPS_PROXY        = var.proxy
    http_proxy         = var.proxy
    https_proxy        = var.proxy
    NO_PROXY           = var.proxy_no
    no_proxy           = var.proxy_no
  }

  type = "Opaque"
}
```

I found this.
[...]

Issue #1498, opened 2023-10-27
docs: OIDC with AKS that is using Azure RBAC


illume:
A question was asked...
> I'm unclear how to set up OIDC with AKS that is using Azure RBAC. If I just create a application in AAD and it will automatically be able to tie users to the Azure RBAC in AKS?

We should make headlamp or the docs better to address this.

illume:
No one got to this yet, so moving to next release milestone.

illume:
Closing old issue, covered in newer OIDC issue.

PR #3177, opened 2025-04-25
backend: Implement support for Azure AKS managed OIDC auth provider

daviddob:
This allows for enabling configuration for AKS managed OIDC and is also more flexible for other auth providers in the following ways:

- Allows overriding the `clientID` and `issuerURL` for validation phases
- Allows for grabbing the `access_token` instead of `id_token`.

All three of these changes are required in the case of AKS due to its nonstandard OIDC flow where it returns an access token from a single MSFT controlled enterprise application with the ID `6dae42f8-4368-4678-94ff-3960e28e3630` [msft docs source](https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks). This is also apparent with other auth providers having nonstandard flows and token issuance this could potentially resolve as well.

The `access_token` returned is also using the older v1 token as opposed to the newer v2 tokens. This is not possible to override even when setting up your enterprise app to only issue v2 tokens, if the scope includes `6dae42f8-4368-4678-94ff-3960e28e3630/user.read`, then it will be issues from a v1 endpoint. This means that the issuer returned in the token does not match the one obtained in discovery as it transitions from `https://login.microsoftonline.com/<tenant-uuid>/v2.0` to `https://sts.windows.net/<tenant-uuid>/`  [msft source](https://learn.microsoft.com/en-us/answers/questions/1664801/azure-ad-token-issuer)

The above can be confirmed with any [AKS managed OIDC cluster](https://learn.microsoft.com/en-us/azure/aks/enable-authentication-microsoft-entra-id) and `kubelogin` flow followed by decoding the token returned to you.

This implements three new flags `oidc-validator-idp-issuer-url`, `oidc-validator-client-id`, and `oidc-use-access-token` to permit the above required alterations to token validation

Closes: #2480, #1498 (will provide tests and docs for the above)

Related to: 
- #2358 [Older AAD login flows are deprecated for this new manged aks oidc](https://learn.microsoft.com/en-us/azure/aks/azure-ad-integration-cli)
- #2848 (Was experiencing the same behavior when trying to log in with headlamp using the `id_token`

## Testing & Docs

This can be tested by building the image from source, and deploying it to a K8s cluster with the following additional args defined in the helm chart (in addition to the standard OIDC config pointing at your azure ad enterprise app)

```
extraArgs:
  - "-oidc-validator-client-id=6dae42f8-4368-4678-94ff-3960e28e3630"
  - "-oidc-validator-idp-issuer-url=https://sts.windows.net/<your-tenant-uuid>/"
  - "-oidc-use-access-token"
```

Before writing out all of the tests and docs (I have this working and tested against both this approach and the one discussed below), I wanted to get some initial feedback on the changes and which method is preferred. From there id be happy to button up the tests, helm chart, and docs accordingly to get users able to auth with AKS managed OIDC implementations.

## Alternate Options

Alternatively this can be implemented with a single flag specifically for Azure AKS managed OIDC `oidc-use-aks-managed`. Implementation is here and adds only a single configuration value but is also then specific for AKS only which may not be desired. https://github.com/kubernetes-sigs/headlamp/compare/main...daviddob:headlamp:azure-aks-managed-oidc-support-simplified?expand=1


illume:
Thanks for all this.

I'm starting to understand access_token is for authorization, and id_token is for authentication. It seems other providers use access_token from a bit of searching about. So I think this way is probably better, than the AKS specific flag.

It would be helpful to have some guidance of how to set this up for testing? Setting up AKS (following the link you posted I guess?) 

Then where to get this? "https://sts.windows.net/<your-tenant-uuid>/" I imagine this needs something to be set up first too.

ashu8912:
> Thanks for all this.
> 
> I'm starting to understand access_token is for authorization, and id_token is for authentication. It seems other providers use access_token from a bit of searching about. So I think this way is probably better, than the AKS specific flag.
> 
> It would be helpful to have some guidance of how to set this up for testing? Setting up AKS (following the link you posted I guess?)
> 
> Then where to get this? `"https://sts.windows.net/<your-tenant-uuid>/"` I imagine this needs something to be set up first too.

I think "https://sts.windows.net/<your-tenant-uuid>/" the tenant-uuid is your entra id app tenant id

daviddob:
Yupp that's exactly right, I'll try to get the documentation and tests implemented shortly. Will include links to the MSFT docs for setting up entra OIDC with AKS and then steps for setting up the enterprise app + headlamp with the new parameters

daviddob:
Planning on writing up the documentation tonight - I updated the helm chart accordingly for the 3 methods of setting the config properties (env, inline, or external secret) as well. Let me know if anything should be adjusted there, technically these are all parameters that could be set via the `extraArgs` field in the chart as follows with indirect references to env vars/secrets or with hardcoded values.

```yaml
- "-oidc-validator-client-id=$(OIDC_VALIDATOR_CLIENT_ID)"
- "-oidc-validator-idp-issuer-url=$(OIDC_VALIDATOR_ISSUER_URL)"
- "-oidc-use-access-token=$(OIDC_USE_ACCESS_TOKEN)"
```

Absolutely down to drop the helm chart changes it it introduces too much clutter in favor of just these parameters called out in the documentation (whichever you believe is the best approach)

ashu8912:
Thanks @daviddob i am in the process of reviewing and testing this, will get back to you with more updates

joaquimrocha:
I just want to second how awesome this is! Thank you @daviddob !

illume (approved):
🎉 thanks!

I didn’t test the chart changes. They look ok to me, but I’ll ask our local chart expert @knrt10 to have a quick look before we merge.

ashu8912 (approved):
Tested this with AKS and EntraID, and this works like charm, and the docs as well are on point, thanks, lets wait for @knrt10 final approval and then we are good to go

knrt10 (approved):
Awesome. Thanks for the change @daviddob

viceice:
it seems this has broken oidc auth: https://github.com/kubernetes-sigs/headlamp/issues/2643#issuecomment-2924800621

We're getting this error now:
```
Failed to verify ID Token: oidc: id token issued by a different provider, expected "$(OIDC_VALIDATOR_ISSUER_URL)" got "https://codeberg.org/"
```

joaquimrocha:
Thanks @viceice . We are looking into it. Thanks!

daviddob:
Replied in issue https://github.com/kubernetes-sigs/headlamp/issues/2643#issuecomment-2940035200, can write up a quick patch for this approach as well as the doc version confusion here https://github.com/kubernetes-sigs/headlamp/issues/3380 if desired and no one has triaged this yet.

Invincibear:
It seems that this doesn't handle the edge case where an access token for an Entra ID user has already been exchanged, and redeploying Headlamp loses the refresh token.
```
Failed to exchange token: oauth2: "invalid_grant" "AADSTS54005: 
OAuth2 Authorization code was already redeemed, please retry with a new valid code or use an existing refresh token. 
Trace ID: 00000000-0000-0000-0000-000000000000 
Correlation ID: 00000000-0000-0000-0000-000000000000 
Timestamp: 2025-08-02 20:22:22Z"
```

I've tried the following, they all result in the same `AADSTS54005` error:
- Nuking the Entra ID App Registration & Enterprise App
- Removing Headlamp from https://myapps.microsoft.com
- Signing out from all Microsoft locations
- Private browsing tab
- Logging in with an entirely different browser to rule out caching/cookies/local storage entirely
- Nuking Headlamp from Kubernetes, and redeploying all of the above with a new App Client Id & Client Secret
- Appending `&conset=prompt` to the Microsoft login url to force re-granting consent

PR #3185, opened 2025-04-28
frontend: update material ui

farodin91:


farodin91:
@sniok What do you think? I tried to fix most of the bugs.

illume:
When it's ready, I think we should wait until the start of a release cycle before merging.

Apart from adding code mods and migration notes to headlamp-plugin... we need to check if dependencies have updated to 7 (it's quite fresh).

illume:
Maybe some of these changes can be made without first upgrading from mui 5?

farodin91:
I want to checkout some filter features in mrt but document state this is a mui 6 feature. 

My upgrade path idea:
- Do upgrade to mui 6 (this should be straightforward)
- switch grid to grid2
- do upgrade to mui 7

joaquimrocha:
Hey @farodin91 . Is this something you want to keep working on some time?

farodin91:
Hey @joaquimrocha if someone wants to work on it go for it.

PR #3186, opened 2025-04-29
frontend: i18n: Table: Add missing locale mappings

illume:
This is so Material React Table, and Material UI use their internal locales when we switch to our locales.

So various UI elements in Table/Mui are rendered with the correct locales.

Before:
image

After:
image

image


sniok (approved):

illume:
(I just rebased against main to fix the conflicts)

PR #3190, opened 2025-04-29
e2e-tests: Add a11y check with Axe in many tests

illume:
This should let us find a11y issues in the parts of the app that are tested by our playwright tests.

Thanks @vyncent-t

### When it fails

This is what it looks like when it fails. (Note the a11y violations are there pretty printed with the failing test).

image


skoeva (commented):
thanks for adding this! just needs a rebase against main

skoeva (approved):

Issue #3198, opened 2025-05-01
frontend test failing


skoeva:
## Describe the bug
CI failing on frontend test (triggered by update to vite in #3196)

## To Reproduce
Steps to reproduce the bug:
1. Update vite and run `make frontend-test` 

## Environment (please provide info about your environment):
 - Installation type: WSL

## Screenshot
![Image](https://github.com/user-attachments/assets/09d376c3-7b1d-4779-97fc-d31b0987463e)

skoeva:
seems like a circular dependency issue, I can push up a fix for this

PR #3200, opened 2025-05-02
frontend: themeSlice: Use useSelector to fix import cycle

skoeva:
This change replaces useTypedSelector in the themeSlice file with react-redux's useSelector address a circular import between the themeSlice and reducers files.

Fixes: #3198

### DIFF
--- a/frontend/src/components/App/themeSlice.ts
+++ b/frontend/src/components/App/themeSlice.ts
@@ -15,9 +15,9 @@
  */
 
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { useSelector } from 'react-redux';
 import { AppTheme } from '../../lib/AppTheme';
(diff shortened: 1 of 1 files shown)


skoeva:
> what if we used "useSelector" from redux package instead of the useTypedSelector, then we won't need any other changes and there won't be an import from reducers?

wasn't sure if useTypedSelector was preferred for some other reason but replacing it with useSelector seems to work, will make the change

frontend/src/components/App/themeSlice.ts
```diff
@@ -149,19 +149,19 @@ const themeSlice = createSlice({
 });
 
 export const useAppThemes = () => {
```
sniok, line 151:
```suggestion
export const useAppThemes = (): AppTheme[] => {
```

there's an AppTheme type in scope you can use here for the return type, then we won't need `any` in other files

PR #3201, opened 2025-05-02
frontend: Table: Hide row actions with hidden columns

skoeva:
This change ensures that when all of the non-action columns are hidden, the action column is also hidden.

Fixes: #2932, follow-up from #2968

### Testing
- [X] Navigate to the Headlamp home page
- [X] Manually hide all the cluster fields in the table
- [X] Ensure that the row actions are also hidden
- [X] Manually unhide one field and ensure that the row actions appear again

### DIFF
--- a/frontend/src/components/common/Table/Table.tsx
+++ b/frontend/src/components/common/Table/Table.tsx
@@ -343,6 +343,21 @@ export default function Table<RowItem extends Record<string, any>>({
     },
   });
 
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Please add the area this changes into the commit message?
`frontend: Table: Hide row actions with hidden columns`

illume (approved):
🎉 🎈 thank you.

PR #3210, opened 2025-05-06
plugins: examples: ui-panels: add aria-label attributes for improved accessibility

sniok:
another fix for e2e a11y tests


skoeva (approved):
awesome, thanks for taking care of this

PR #3219, opened 2025-05-10
frontend: Add events for home page load

ashu8912:


illume (changes requested):
- needs PR description, and an issue linked
- needs tests to pass
- new event should be documented

skoeva (commented):
echoing what was said above, rebase with updated snapshots as well as info on the issue/fix would be very helpful

illume:
> I think working on a forked branch would also be good since branch protections prevent force pushes, meaning a new PR would likely need to be opened to make changes

Oh yes, good point. We can't work on branches anymore, so this will have to be done on your personal fork @ashu8912 and resubmitted as a new PR.

PR #3220, opened 2025-05-11
frontend: hide managed fields

farodin91:


farodin91:
I had this idea from argo-cd.

ashu8912 (approved):
Tested, LGTM!!

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -109,6 +112,15 @@ export default function EditorDialog(props: EditorDialogProps) {
 
   const theme = useCurrentAppTheme();
 
+  const [hideManagedFields, setHideManagedFieldsState] = React.useState(() => {
```
sniok, line 115: we have a `useLocalStorageState` hook that you could use here instead

farodin91: used.

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -444,6 +463,20 @@ export default function EditorDialog(props: EditorDialogProps) {
...
+                          <Switch
+                            checked={hideManagedFields}
+                            onChange={() => setHideManagedFields(!hideManagedFields)}
+                            name="hideManagedFields"
+                          />
+                        }
+                        label={t('Hide Managed Fields')}
+                      />
+                    ) : (
+                      <div />
```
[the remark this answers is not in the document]
farodin91: removed

frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -109,34 +113,41 @@ export default function EditorDialog(props: EditorDialogProps) {
 
   const theme = useCurrentAppTheme();
 
-  const [useSimpleEditor, setUseSimpleEditorState] = React.useState(() => {
-    const localData = localStorage.getItem('useSimpleEditor');
-    return localData ? JSON.parse(localData) : false;
-  });
+  const [hideManagedFields, setHideManagedFields] = useLocalStorageState<boolean>(
+    'hideManagedFields',
+    false
```
sniok, line 118: I'd make the default setting here "true", that's what kubectl does by default and I think we should too. But this is not too critical since you only have to check it once and it'll remember it.
What do you think @farodin91 @joaquimrocha ?

joaquimrocha: Yeah, let's turn it on by default please.

farodin91:
Rebased and fixed. 

Another idea:
I tried shortly to use folding section in monaco but didn't get it to work. Basic always folding the managedfields section in the editor. If anyone thinks it is a better idea and get it work go for it.

sniok (approved):
nice, thanks! I think folding would be super cool but from what I remember it's quite tricky to implement so let's keep it simple

illume:
Very cool. Thanks.

btw. We don’t need Signed-off-by anymore now that we’re using the EasyCLA thing.

PR #3229, opened 2025-05-12
frontend: Add preview of rolldown-vite

sniok:
You can read more about rolldown [here](https://vite.dev/guide/rolldown)

I've been trying rolldown-vite previews lately and now it works out of the box without any additional changes.
rolldown-vite is in the preview stage so we might want to hold off from upgrading yet but this looks promising

Based on my local tests the performance of `npm run build` is around 3x faster
vite-rolldown 7-11s
vite 29-31s

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -95,7 +95,7 @@
         "typescript": "5.6.2",
         "url": "^0.11.0",
         "util": "^0.12.4",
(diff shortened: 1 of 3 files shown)


sniok:
Since rolldown vite has been properly announced, I think we should try to use this soon
https://voidzero.dev/posts/announcing-rolldown-vite

PR #3235, opened 2025-05-13
frontend: NamespacesAutocomplete: Show list of selected namespaces while hovering

### LINKED ISSUE #3232: Show a list of selected namespaces while hovering.
In HeadlampUI, when we select more than one namespaces from a filter it only shows +N .  If the user want to see what options they have selected they won't able to see those. unless they open the ``nampespaces`` options. 

Image

Options can be shown by using a list. so when the user hover to ``+`` sign the list will appear. 

[What type of users may be interested in this feature? What kind of Headlamp setup is likely used. (e.g. WSL users / Helm chart users / In-Cluster users / Plugin developers)]

Yes I will propose a PR. 

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

upsaurav12:
Fixes #3232 ,
In PR user will able to see the list of other selected namespaces options while hovering to ``+`` sign.
To achieve this i have used ``useState`` hook for adding state for checking if the user is hovering or not. I have also used ``onMouseEnter`` and ``onMouseLeave`` for toggling hovering effect. 

When the User hovers to ``+`` sign it will trigger ``onMouseEnter`` and ``isHovered = true``. List will only show when ``isHovered`` is true.


frontend/src/components/common/NamespacesAutocomplete.tsx
```diff
@@ -153,7 +155,31 @@ export function PureNamespacesAutocomplete({
...
+                <b
+                  style={{
+                    cursor: 'pointer',
+                  }}
+                  onMouseEnter={() => setIsHovered(true)}
+                  onMouseLeave={() => setIsHovered(false)}
+                >{`+${tags.length - joinnedNamespaces}`}</b>
+
+                {isHovered && (
+                  <ul
```
sniok, line 167: For a more consistent UI I'd suggest using [Tooltip component](https://v5.mui.com/material-ui/react-tooltip/) that we also use in other places

skoeva (commented):
Thanks for contributing! I think these commits can be squashed into one, also don't forget to run `make frontend-test` and update any outdated snapshots 🔥

illume:
Thanks @upsaurav12

Would you mind adding the component to the context part of the commit message? Also the first letter of the verb should be capitalized.
`frontend: NamespacesAutocomplete: Show list of selected namespaces while hovering`


Also please remove this from the commit message?
```
frontend: tooltip: used Tooltip component for hovering and showing list items
fix: fixed failing tests
```

upsaurav12:
it was added in description when i squashed commits.

illume (approved):
🎉 Thanks!

PR #3240, opened 2025-05-15
frontend: Add events for home page load

ashu8912:
When we are on the home page we list all the clusters there, we need to pass this as an event as it was missing


illume (changes requested):
Please add component changed to context (see other commit messages).

A CI check is failing.

Needs rebasing because there's a conflict now.

PR #3241, opened 2025-05-15
frontend: Add hostIPs and podIPs to pod

### LINKED ISSUE #3083: Show pod ip's instead of single ip when cluster is in dualstack mode
I've multiple kubernetes cluster in dualstack mode, so every pod has two ip adresses and i need to edit a resource to see the IPv6 adresses under the status field.

 Show the `podIPs` and `hostIPs` fields instead of only `podIP` and `hostIP`.
Also services can have two adressses.

All users which have dualstack kubernetes clusters

Probably yes, but time is very limited. 😅 

## Additional context

Image

Image

Sample of lens app

Image

ashu8912:
In the pod Details view some pods could have multiple host and pod IPs so when rendering the pod Details view we should show these IPs as well.
fixes #3083

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -529,8 +529,21 @@ export default function PodDetails(props: PodDetailsProps) {
             value: item.status.hostIP ?? '',
           },
           {
+            name: t('Host IPs'),
+            value: item.status.hostIPs
+              ? item.status.hostIPs.map(ipObj => ipObj.ip).join(', ')
+              : '',
+            hideLabel: item.status.hostIPs?.length === 0,
+          },
+          {
             name: t('Pod IP'),
             value: item.status.podIP ?? '',
+          },
+          {
+            name: t('Pod IPs'),
+            value: item.status.podIPs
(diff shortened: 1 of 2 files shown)


frontend/src/components/pod/Details.tsx
```diff
@@ -528,10 +528,23 @@ export default function PodDetails(props: PodDetailsProps) {
...
+          },
           {
             name: t('Pod IP'),
             value: item.status.podIP ?? '',
           },
+          {
+            name: t('Pod IPs'),
+            value: item.status.podIPs
+              ? item.status.podIPs.map(ipObj => ipObj.ip).join(', ')
+              : ''
```
sniok, line 546: this entry should also be hidden if there are no pod ips right?

ashu8912: fixed

sniok:
After this change it shows both values, maybe we should hide plural version if it's the same as singular? (In this case for example only show HostIP and Pod IP, hide Host IPs and Pod IPs, because they're the same)

image

illume (changes requested):
@ashu8912 please see note by @sniok

PR #3242, opened 2025-05-15
frontend: Export helper cluster methods from utils

ashu8912:
We were not exposing all the cluster helper methods, this PR adds all the cluster related methods in utils


sniok:
the functions added in this PR are actually new so we can export them from a different place, we don't have to add them to utils.
we should also consider enabling this eslint rule in the future to prevent regressions https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md

PR #3249, opened 2025-05-15
frontend: headlamp-plugin: Update dependencies

skoeva:
These changes involve patch updates to dependencies in the frontend and headlamp-plugin using `npm outdated`.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -125,7 +125,7 @@
         "prettier": "^2.7.1",
         "resize-observer-polyfill": "^1.5.1",
         "storybook": "^8.5.6",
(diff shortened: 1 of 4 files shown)


illume:
I think this is too much at once.

Because we might be releasing soon, I think a big change like this should wait until the next release cycle.
Can we please do this in a smaller amount of changes at a time?
Can it be limited to patch updates only? Not minor/major?

illume (changes requested):
It's too much to change at once I think. It's lots of testing work to do, and lots of release notes to check.

Would you mind only updating one or a few related things at once?

illume:
This is some months old now with conflicts and such, so closing this.

PR #3252, opened 2025-05-15
frontend: Refactoring mui named imports from top level to individual component imports

illume:
For mui. This has some benefits for faster loading, and tree shaking.

## Testing done

- tried all the pages manually locally
- ran as app on WSL / Ubuntu, and clicked about

### DIFF
--- a/frontend/src/components/App/AppContainer.tsx
+++ b/frontend/src/components/App/AppContainer.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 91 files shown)


skoeva (approved):
LGTM

illume:
Thanks @skoeva

sniok:
we're including whole @mui/material in pluginLib when app first loads so I'm not sure if this will improve load times or bundle size

Issue #2554, opened 2024-11-11
Columns filter state preservation


veyselsahin:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
I need column filters to preserve the last state forever.

## Describe the solution you'd like
There might be either a setting or an app can decide to show filters based on the last desired state.

## What users will benefit from this feature?
Users will save significant time and contextual focus if the filters are shown as they desire.

## Are you able to implement this feature?
[Yes (I will propose a PR) / No.]
> Do you want to help implement this feature? Please see our [contribution docs]
Yes, It's been discussed in the slack channel.

sreisich:
I don't know if this already is included here, but also the column sorting should be preserved.

veyselsahin:
I was overloaded for a while, but this could also be a suitable implementation. I will continue to work on these features.  Thanks @sreisich.

7908837174:
Hey everyone, 

I'd love to take on this issue and contribute to implementing the feature for preserving column filter states. Given its relevance for improving user experience and efficiency, I'm excited to work on it. 

Please assign me to this issue so I can start diving in. Looking forward to collaborating! 

Thanks!

7908837174:
Sure, no problem.

skoeva:
This issue has been resolved with the recent details drawer feature. Let us know if it persists, will be closing for now

kallal79:
No, it has not.

PR #3258, opened 2025-05-16
#2554-Kallal-Columns filter state preservation

7908837174:
#2554-Kallal-Columns filter state preservation

### DIFF
--- a/frontend/src/components/App/Notifications/List/List.stories.tsx
+++ b/frontend/src/components/App/Notifications/List/List.stories.tsx
@@ -17,6 +17,7 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { Meta, StoryFn } from '@storybook/react';
 import { initialState as CONFIG_INITIAL_STATE } from '../../../../redux/configSlice';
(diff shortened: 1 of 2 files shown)


7908837174:
Awesome, thank you so much!

frontend/package.json
```diff
@@ -17,10 +17,10 @@
     "@iconify/icons-mdi": "^1.2.9",
     "@iconify/react": "^3.2.1",
     "@monaco-editor/react": "^4.3.1",
-    "@mui/icons-material": "^5.16.7",
+    "@mui/icons-material": "^5.17.1",
```
illume, line 20: I wonder why the dependencies are changed here?

Are they required to get this to work? If not, can you please remove the dependency changes from the PR?

illume:
It looks like the EasyCLA check is failing.

Something like this:
`frontend: Table: Fix column filters to preserve their state`

7908837174:
Ok, I'll check it out.

7908837174:
Why?

PR #3261, opened 2025-05-16
headlamp-plugin: Bump undici to 6.21.3

skoeva:
This change bumps the undici dependency in the headlamp-plugin to the desired version.


illume (approved):
👍

Issue #3253, opened 2025-05-15
Add German Language


eyeaadil:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
[Add Germsn languages so that people who know Russian languages can use it easily.]

## Describe the solution you'd like
I propose adding support for German languages to Headlamp's UI to make it accessible to a wider audience
. The solution includes the following steps:

Add New Languages like:
German

## Are you able to implement this feature?
Yes ,i am able to implement this feature.

fromsaurav:
@k8s-ci-robot 
 I’ve worked on this issue and would like to submit a PR. Let me know if the current assignee is still working on it or if I can proceed.

eyeaadil:
@fromsaurav , I am working on this issue.

PR #3262, opened 2025-05-17
Arabic language is added

eyeaadil:
Russian language support has been added to the Headlamp website to make it easier for Russian-speaking users to navigate and interact with the site.
This resolves issue no:- #3253 
image


eyeaadil:
@illume , Can you please review this PR?

illume:
مرحبًا.

illume:
Amazing, thank you.

My Arabic is extremely basic, so I'll have to find someone else to review those parts of it.

This will be the first right to left language we use, so I imagine testing it will take longer. Did you spot any parts of the app that need attention or that is completely broken? We don't necessarily need to fix these issues in this PR, but it might be nice to note them somewhere.

For Kubernetes jargon like Pods, kubectl and Kubernetes there's usually not a good word to translate it to. So we tend to keep them in English unless there's a well known word for it in the language.

eyeaadil:
@illume Updated the Kubernetes Jargons, keep it in same English.

eyeaadil:
> Amazing, thank you.
> 
> My Arabic is extremely basic, so I'll have to find someone else to review those parts of it.
> 
> This will be the first right to left language we use, so I imagine testing it will take longer. Did you spot any parts of the app that need attention or that is completely broken? We don't necessarily need to fix these issues in this PR, but it might be nice to note them somewhere.
> 
> For Kubernetes jargon like Pods, kubectl and Kubernetes there's usually not a good word to translate it to. So we tend to keep them in English unless there's a well known word for it in the language.

I do not find any broken part, if you found let me know, isuerly work on that to fix that.

illume:
Oh by the way...

Would you mind changing the commit to something like this?

`frontend: app: i18n: Add Arabic language`

illume:
I think it’s that it needs to be rebased with the main branch first. Then run the check.

illume (changes requested):
Are you able to rebase against the latest main branch?
There's a conflict that needs to be fixed.


Also I think you need to do this if the check fails:
`npm run i18n`

eyeaadil:
@illume Can you please review it?I eagerly waiting for your review

illume:
Hi @eyeaadil

I think the next step is waiting for another person to review your work who understands Arabic.

joaquimrocha:
Yeah, I second @illume point about the Kubernetes jargon. I see a very complete translation in Arabic, which is a language I do not speak, but I found some articles where it seems to mentioned jargon with the original words, e.g. https://academy.hsoub.com/devops/cloud-computing/%D9%85%D8%AF%D8%AE%D9%84-%D8%A5%D9%84%D9%89-kubernetes-r467/

joaquimrocha:
Hi @eyeaadil . Are you able to update/double-check the PR after my previous comments. We would love to get arabic as a locale, but want to make sure this is in line with what's expected by the broader community.

Issue #3199, opened 2025-05-02
Container port update in In-cluster Installtion Docs


prudhvikrishnap:
# Issue

In the documentation, the `kubectl port-forwarding` command mentions the container port as `80` [here](https://headlamp.dev/docs/latest/installation/in-cluster/#exposing-headlamp-with-port-forwarding) 

Dockerfile uses 4466 in the main branch [here](https://github.com/kubernetes-sigs/headlamp/blob/main/Dockerfile#L95)

I tired running it on ports 80 and 8080, port-forwarding failed.

# Expected
Use port-forwarding to open headlamp on host port 8080 and container port 80.

# Fix
This needs to be updated in the above documentation. The container port to use port-forwarding is 4466.

fromsaurav:
I've verified this issue by checking the Dockerfile in the main branch:
$ curl -s https://raw.githubusercontent.com/kubernetes-sigs/headlamp/main/Dockerfile | grep EXPOSE
EXPOSE 4466

![Image](https://github.com/user-attachments/assets/1484e0c3-d73d-45c7-b06e-c99e3703f710)

This confirms that the container exposes port 4466, not port 80 as currently documented in the in-cluster installation guide.
The documentation at /docs/installation/in-cluster/index.md should be updated to use the correct port in the port-forwarding example, Should be:
bashkubectl port-forward -n kube-system service/headlamp 8080:4466

yolossn:
AFAIU the docs uses the [kubernetes-headlamp.yaml](https://github.com/kubernetes-sigs/headlamp/blob/1d2e68d710fccbe3688d93dbf9e3657ae3a2f514/kubernetes-headlamp.yaml#L8) file and it defines the `service` to use port 80. Since the port-forward command targets the service directly the command should work.

fromsaurav:
> AFAIU the docs uses the [kubernetes-headlamp.yaml](https://github.com/kubernetes-sigs/headlamp/blob/1d2e68d710fccbe3688d93dbf9e3657ae3a2f514/kubernetes-headlamp.yaml#L8) file and it defines the `service` to use port 80. Since the port-forward command targets the service directly the command should work.

Got it, thank you for inforamation!

fromsaurav:
[...]

PR #3264, opened 2025-05-17
Fixed container port in port-forwarding docs

fromsaurav:
This PR fixes issue #3199 where the container port in the port-forwarding documentation was incorrectly specified as 80, while the Dockerfile actually exposes port 4466.

The fix updates the port-forwarding command in the in-cluster installation documentation to use the correct container port 4466 instead of 80.

Previous:
kubectl port-forward -n kube-system service/headlamp 8080:80
Currenet:
kubectl port-forward -n kube-system service/headlamp 8080:4466

Refer this attachment for better understanding of fix,
Screenshot from 2025-05-17 10-17-34

### DIFF
--- a/docs/installation/in-cluster/index.md
+++ b/docs/installation/in-cluster/index.md
@@ -83,7 +83,7 @@ If you want to quickly access Headlamp (after having its service running) and
 don't want to set up an ingress for it, you can run use port-forwarding as follows:
 
 ```bash
-kubectl port-forward -n kube-system service/headlamp 8080:80
+kubectl port-forward -n kube-system service/headlamp 8080:4466
 ```
 
 and then you can access `localhost:8080` in your browser.


illume:
Thanks @fromsaurav good fix.

We follow a linux kernel style of commit messages. With the context at the front followed by an action word and then the rest.

Would you mind changing the commit message to something like this?
`docs: in-cluster: Fix port forwarding command port used`

fromsaurav:
> Thanks @fromsaurav good fix.
> 
> We follow a linux kernel style of commit messages. With the context at the front followed by an action word and then the rest.
> 
> Would you mind changing the commit message to something like this? `docs: in-cluster: Fix port forwarding command port used`

Updated, thanks !

yolossn:
Thanks for the update! Just to clarify, the [kubernetes-headlamp.yaml](https://github.com/kubernetes-sigs/headlamp/blob/1d2e68d710fccbe3688d93dbf9e3657ae3a2f514/kubernetes-headlamp.yaml#L8) file defines the `service` to use port 80. Since the port-forward command targets the service directly the command should work.

fromsaurav:
> Thanks for the update! Just to clarify, the [kubernetes-headlamp.yaml](https://github.com/kubernetes-sigs/headlamp/blob/1d2e68d710fccbe3688d93dbf9e3657ae3a2f514/kubernetes-headlamp.yaml#L8) file defines the `service` to use port 80. Since the port-forward command targets the service directly the command should work.

Informative, thank you for clarifying!

PR #3265, opened 2025-05-17
fix: import fix on ./frontend/storybook/main.ts

adityadeshlahre:
### Import fix on main.ts of storybook 

So, I was facing an error while running first time the storybook using command directly 

Error Reproducing

- On a fresh repo clone, go to ./fontend/ and run Storybook
```
cd ./frontedn
npm install
npm run storybook
```
- You will see an error

#### Error Image

image

#### Fixed using import as type in main.ts 

- check file change in PR

Thank you :slightly_smiling_face: 

let me know if any changes required

### DIFF
--- a/frontend/.storybook/main.ts
+++ b/frontend/.storybook/main.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { StorybookConfig } from '@storybook/react-vite';
+import type { StorybookConfig } from '@storybook/react-vite';
 
 export default {
   framework: '@storybook/react-vite',


illume (approved):
Thanks @adityadeshlahre 

I can't reproduce the error locally, but your change makes sense. Since the imported thing is only used as a type. Also the change works for me locally.

adityadeshlahre:
hey @illume, 
quick query
what components are still pending to have a storybook written ? can you point me some!
Thank You : )

illume:
@adityadeshlahre Oh... one more thing.

We follow a linux kernel style of commit messages. With the context at the front followed by an action word and then the rest.

Would you mind changing the commit message to something like this?
`frontend: storybook: Fix npm run storybook crash`

illume:
If you look at the components, the ones without .stories.tsx are ones that are missing.

You might want to also look at the test coverage to see which ones are not covered with: `make frontend-test`. This prints out a report with the components that lack tests. We use stories to make snapshots, so we can see what parts of stories do not have all states.

Note, some stories do not cover all the states... so an alternative to creating a new story could be to add a new state.

illume:
Thanks again @adityadeshlahre

Issue #3023, opened 2025-03-26
headlamp-plugin: Add command to validate artifacthub-pkg.yml


illume:
Waiting 25 minutes per try to get the artifacthub validator to work by waiting for artifacthub to rescan is slow.

## Describe the solution you'd like

A headlamp-plugin command that validates artifacthub-pkg.yml files

## What users will benefit from this feature?

Being able to publish more easily.

BHAVISHYA2005:
Working on it

PR #3269, opened 2025-05-18
feat: Add validate-artifacthub CLI command to Headlamp plugin toolkit (#3023)

BHAVISHYA2005:
This PR adds a new CLI command to the Headlamp plugin toolkit:
**`validate-artifacthub [file]`**

This command allows users to validate an `artifacthub-pkg.yml` file against the official ArtifactHub schema locally, so plugin authors can check their metadata before publishing or waiting for ArtifactHub rescans.

- **New command:**  
  `validate-artifacthub [file]` (defaults to `artifacthub-pkg.yml` if no file is provided)
## Testing

- Added automated tests in `test-headlamp-plugin.js`:
  - Checks that the command succeeds for a valid file.
  - Checks that the command fails for an invalid file (missing required fields).
- Manual testing also performed via CLI.

This feature was requested in [#3023](https://github.com/kubernetes-sigs/headlamp/issues/3023) to help plugin authors catch metadata issues early and improve the development workflow for Headlamp plugins.
---

Closes: #3023

/cc @illume

### DIFF
--- a/plugins/headlamp-plugin/artifacthub-pkg.schema.json
+++ b/plugins/headlamp-plugin/artifacthub-pkg.schema.json
@@ -1,0 +1,55 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "ArtifactHub Package",
+  "type": "object",
+  "required": [
+    "name",
+    "version",
+    "description",
+    "keywords",
+    "links",
+    "maintainers"
+  ],
+  "properties": {
+    "name": { "type": "string" },
+    "version": { "type": "string" },
+    "description": { "type": "string" },
+    "keywords": {
+      "type": "array",
+      "items": { "type": "string" }
+    },
+    "links": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "url"],
+        "properties": {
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Amazing, thanks for this.

I left some notes for some improvements.


Something like this perhaps?
> headlamp-plugin: Add validate-artifacthub CLI command and tests

To see other commit message examples please see git log.

illume:
By the way, you can run the function plugin tests locally (they take a while to run).

```shell
cd headlamp/plugins/headlamp-plugins
test-headlamp-plugin.js
```

If you comment out some parts of the cleanup function it makes it slightly easier to debug sometimes... because it leaves the temporary folder there at the end of the test (instead of cleaning it up when the test finishes)
image

skoeva:
Hi, are you planning on continuing on with this PR? Or would it be better for us to take over this issue?

BHAVISHYA2005:
> Hi, are you planning on continuing on with this PR? Or would it be better for us to take over this issue?

Hi @skoeva 👍🏻, I will update you soon as I am a little busy bz of exams, Sorry for delays @illume. My bad

Issue #3276, opened 2025-05-18
Sidebar expansion description doesn't update with state


Pratik-050:
## Describe the bug
The tooltip/description of the expand icon on sidebar always say collapse sidebar either the sidebar is open or not

## To Reproduce
Steps to reproduce the bug:
1. [Go to home]
2. [Hover on the right chevron icon see the description says Collapse Sidebar]

## Environment:
 - Installation type: [In-Cluster: minikube]
 - Headlamp Version: [0.30.0]
 
## Are you able to fix this issue?
Yes

## Additional Context

![Image](https://github.com/user-attachments/assets/129f2648-8d14-41a9-a649-7f82a5b0f22a)

![Image](https://github.com/user-attachments/assets/59ea051b-eadb-4dd6-9c35-d3c1f6b34de0)

PR #3277, opened 2025-05-18
frontend: Sidebar: Fix conditional expansion description text

Pratik-050:
## Note for reviewers:
Changed the description of the left chevron and right chevron button in the sidebar. 

fixes: #3276 

image
image

Renamed Collapse sidebar to Shrink sidebar as the i18n support for Collapse Sidebar was there with a block 'S' but not the same for Expand Sidebar. Attached a screenshot of the translation.json for more context:
image
This is why some tests expecting Collapse Sidebar are failing as well.

### DIFF
--- a/frontend/src/components/Sidebar/Sidebar.tsx
+++ b/frontend/src/components/Sidebar/Sidebar.tsx
@@ -120,7 +120,7 @@ function SidebarToggleButton() {
           dispatch(setWhetherSidebarOpen(!isOpen));
         }}
         icon={isOpen ? 'mdi:chevron-left-box-outline' : 'mdi:chevron-right-box-outline'}
-        description={t('translation|Collapse Sidebar')}
+        description={isOpen ? t('translation|Shrink sidebar') : t('translation|Expand sidebar')}
       />
     </Box>
   );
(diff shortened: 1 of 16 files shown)


illume:
Very cool.  Thank you.

You can update the snapshot test with:
`npm run test -- -u`

Would you mind changing them to something like this?
`frontend: Sidebar: Fix conditional expansion description text`

Pratik-050:
Thanks @illume I've updated the commit message and the tests as well. Thank you for your guidance!

illume:
The check is failing...

Can you please run `make frontend-i18n-check` in the root of the repo and see what the problem is?

I see we already have  `t('Shrink sidebar') : t('Expand sidebar')` in HeadlampButton.tsx... and in the translation.json files.

eyeaadil (commented):
Looking Good

Pratik-050:
Hello @illume yes we have `t('Shrink sidebar') : t('Expand sidebar'`) in HeadlampButton.tsx but that only changes a aria-label, I think that is for test purpose only. 
This error is coming when I'm trying to run make `make frontend-i18n-check`
Screenshot from 2025-05-19 18-08-45

illume:
After you run that command, then run

git status

… to see what files it has changed.

it might also be a good idea to git rebase against the latest main branch.

Pratik-050:
I took a pull from the latest main just now and ran this command but no file changes are there. @illume

illume:
Hrmm. Ok. Let me have a closer look in a little bit. If I find a fix I’ll push it to your branch.

Pratik-050:
Thanks! @illume

illume (approved):
🎉 🎈 thank you.

PR #3285, opened 2025-05-18
backend: Refactor portforward/handler_test.go TestStartPortForward to be shorter

AyushSriv06:
**Addressed Function Length Issue:**

- Refactored TestStartPortForward to reduce statement count (from 97 to well below the 40 limit)
- Removed the //nolint:funlen comment as it's no longer needed

**Created Helper Functions for Test Segments:**

- testStartPortForward: Starts port forwarding
- testVerifyPortForward: Verifies connection works
- testStopPortForward: Tests stopping functionality
- testListPortForwards: Tests listing port forwards
- testGetPortForwardByID: Tests retrieving specific port forward
- testDeletePortForward: Tests deletion functionality

**Added Utility Helper Functions:**

- Created findPodWithPort: Extracts pod selection logic
- Created makeJSONRequest: Simplifies JSON request creation

Fixes: #3280


backend/pkg/portforward/handler_test.go
```diff
@@ -51,21 +52,46 @@ func getDefaultKubeConfigPath(t *testing.T) string {
...
+	require.NotEmpty(t, podList.Items)
+
+	// Select the pod and make sure it has a port to forward to
+	for _, pod := range podList.Items {
+		if len(pod.Spec.Containers) > 0 && len(pod.Spec.Containers[0].Ports) > 0 {
+			return pod.Name, fmt.Sprint(pod.Spec.Containers[0].Ports[0].ContainerPort)
+		}
+	}
+
+	return "", ""
```
BHAVISHYA2005, line 69: Returning empty strings on failure ? This hides the errors and makes debugging hard. Just return an error if the function can fail.

backend/pkg/portforward/handler_test.go
```diff
@@ -76,221 +102,201 @@ func TestStartPortForward(t *testing.T) {
...
+	require.NotEmpty(t, podName)
+	require.NotEmpty(t, targetPort)
 
-	podName := ""
-	targetPort := ""
-	// select the pod and make sure it has a port to forward to.
-	for _, pod := range podList.Items {
-		if len(pod.Spec.Containers) > 0 && len(pod.Spec.Containers[0].Ports) > 0 {
-			podName = pod.Name
-			targetPort = fmt.Sprint(pod.Spec.Containers[0].Ports[0].ContainerPort)
```
BHAVISHYA2005, line 94: I think using strings like Minikube and headlamp directly in the test makes it difficult to run tests against a different cluster or namespace. There we will have to edit the code. Suggestion: You can use Environment variables here…

backend/pkg/portforward/handler_test.go
```diff
@@ -76,221 +102,201 @@ func TestStartPortForward(t *testing.T) {
...
+	const clusterName = "minikube"
 
-			break
-		}
-	}
+	const namespace = "headlamp"
```
BHAVISHYA2005, line 117: I think using strings like Minikube and headlamp directly in the test makes it difficult to run tests against a different cluster or namespace. There we will have to edit the code. Suggestion: You can use Environment variables here…

backend/pkg/portforward/handler_test.go
AyushSriv06: Thank you for reviewing my PR. I guess the tests use empty strings, and earlier implementations also had the same.

backend/pkg/portforward/handler_test.go
AyushSriv06: okay i will be looking into this!

illume:
@AyushSriv06 would you mind squashing the commits down?

It would be good if commits with merges and lint fixes are not there. 
Please use rebase against the latest main branch

If you want to have multiple commits, they should be atomic commits where each commit makes sense on its own (and passes all tests on its own).

illume:
@AyushSriv06 ok, thanks :)

AyushSriv06:
@illume Hey, I was having some issues with this branch, so I made a new PR. Please have a look. PR #3329

PR #3286, opened 2025-05-18
Backend : portforward : Make TestStartPortForward shorter and cleaner 

meethereum:
…ter and cleaner

Rationale behind the changes : 
1. The function was unusually too long at first, so abstracting some requests and tasks could help reduce a lot of lines, rather  than adjusting a couple of lines here and there.
2. Naturally, I wanted to have more "requires" and "asserts" in the tests, so I picked up chunks which were independent and made a function out of them.
3. At first I thought this would be considered as a naive approach, but I went through one of the CNCF projects I have contributed to earlier, and  I was happy to validated that they also have followed the exact same approach(mainly for cleanups and subtasks as well) 
4. It adds to better readability as well

Having said that, I will be actively waiting for the reviews and discuss more how we can reduce this huge test  even more : )

### DIFF
--- a/backend/pkg/portforward/handler_test.go
+++ b/backend/pkg/portforward/handler_test.go
@@ -51,6 +51,78 @@ func getDefaultKubeConfigPath(t *testing.T) string {
 	return filepath.Join(homeDirectory, ".kube", "config")
 }
 
+func findPortToForward(t *testing.T, kc *kubeconfig.Context) (string, string) {
+	clientSet, err := kc.ClientSetWithToken("")
+	require.NoError(t, err)
+	require.NotNil(t, clientSet)
+
+	podList, err := clientSet.CoreV1().Pods("headlamp").List(context.Background(), metav1.ListOptions{})
+	require.NoError(t, err)
+	require.NotEmpty(t, podList.Items)
+
+	podName := ""
+	targetPort := ""
+	// select the pod and make sure it has a port to forward to.
+	for _, pod := range podList.Items {
+		if len(pod.Spec.Containers) > 0 && len(pod.Spec.Containers[0].Ports) > 0 {
+			podName = pod.Name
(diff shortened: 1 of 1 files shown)


backend/pkg/portforward/handler_test.go
```diff
@@ -51,6 +51,78 @@ func getDefaultKubeConfigPath(t *testing.T) string {
...
+		if len(pod.Spec.Containers) > 0 && len(pod.Spec.Containers[0].Ports) > 0 {
+			podName = pod.Name
+			targetPort = fmt.Sprint(pod.Spec.Containers[0].Ports[0].ContainerPort)
+
+			break
+		}
+	}
+
+	return podName, targetPort
+}
```
BHAVISHYA2005, line 76: Consider returning an error if no suitable pod is found. It currently returns empty strings, which may cause test failures

meethereum: Thanks for reviewing! Actually the earlier implementation used empty strings as well. If you look at the entire code, you will notice that tests are based on the strings being empty / not empty and all possible errors are also being handled by require.NoError .
I have not changed the implementations, just some moved chunks out into functions.
If I am missing something, please let me know 👍

backend/pkg/portforward/handler_test.go
```diff
@@ -51,6 +51,78 @@ func getDefaultKubeConfigPath(t *testing.T) string {
 	return filepath.Join(homeDirectory, ".kube", "config")
 }
 
+func findPortToForward(t *testing.T, kc *kubeconfig.Context) (string, string) {
```
illume, line 54: This one would need a comment too.

meethereum: Done :rocket:

illume:
Can you please rebase this against main? There's a conflict because of a change that was merged in recently.

meethereum:
Sure. Working to solve the conflicts now

PR #3299, opened 2025-05-20
Makefile: Add better failure message for make frontend-i18n-check

illume:
So that when it fails the people know what to do.

Often the check fails, but it does not write the changed files. So people need to run `make i18n`. 

People don't see the warning message at the top of several pages of output, so it's at the bottom.

They are running a make file, so they are told the make command to run (not an npm one).


illume:
There's issues with this approach.

PR #3302, opened 2025-05-20
SECURITY.md: Add security policy specific text

skoeva:
This change adds security policy specific text in the SECURITY file to meet the desired specification.


illume (approved):
👍 thanks!

PR #3305, opened 2025-05-20
frontend: i18n: Update Hindi translations

DhairyaMajmudar:
Fix for workflow failures here https://github.com/kubernetes-sigs/headlamp/actions/runs/15141624043/job/42566938507?pr=3281


skoeva (commented):
Thanks for taking care of this! Minor nit but I think the commit title could be more appropriately changed to something like this:
`frontend: i18n: Update Hindi translations`

DhairyaMajmudar:
> Thanks for taking care of this! Minor nit but I think the commit title could be more appropriately changed to something like this: `frontend: i18n: Update Hindi translations`

Done, changed the commit message 👍🏻

illume (approved):
👍 thanks

PR #3308, opened 2025-05-20
docs: add essential step to build local Docker image for setup

ANIRUDH-SJ:
Included a necessary command in the setup documentation to build the Headlamp Docker image locally. This step ensures the image is available for local Kubernetes environments like Minikube, where the imagePullPolicy is set to Never. It is crucial for successfully deploying Headlamp during development.

### DIFF
--- a/docs/development/index.md
+++ b/docs/development/index.md
@@ -192,6 +192,7 @@ ones made in the local docker environment.
 
 ```bash
 eval $(minikube docker-env)
+docker build -t headlamp-k8s/headlamp:development
 DOCKER_IMAGE_VERSION=development make image
 ```
 


illume:
I wonder why “make image” doesn’t do this already?

I thought it did.

docs/development/index.md
```diff
@@ -192,6 +192,7 @@ ones made in the local docker environment.
 
 ```bash
 eval $(minikube docker-env)
+docker build -t headlamp-k8s/headlamp:development
```
sniok, line 195: `make image` command already builds an image

check Makefile:146

sniok (changes requested):
Thanks for contributing to headlamp.
The step that has been added is not needed since `make image` already builds in image

ANIRUDH-SJ:
Thanks will change this

PR #3314, opened 2025-05-21
plugins/examples/resource-charts: Add .gitignore

illume:


skoeva (approved):
thanks for catching this

PR #3315, opened 2025-05-21
test(Loader): add unit tests for Loader component

ANIRUDH-SJ:
- Added Loader.test.tsx to improve test coverage for the Loader component

### DIFF
--- a/frontend/src/components/common/Loader.test.tsx
+++ b/frontend/src/components/common/Loader.test.tsx
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/Loader.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </TestContext>
+    );
+
+    // Check if the container Box is present
+    const container = screen.getByRole('progressbar').parentElement;
+    expect(container).toHaveClass('MuiBox-root');
+
+    // Check if CircularProgress is rendered
+    const progress = screen.getByRole('progressbar');
+    expect(progress).toHaveClass('MuiCircularProgress-root');
```
sniok, line 36: These tests heavily rely on class names of the CircularProgress component from mui and that's a implementation detail of the mui library. Since there's not much logic in the Loader component itself these tests end up just testing the mui component which we shouldn't do. I don't think this component requires any additional tests other than a snapshot.

ANIRUDH-SJ: Thanks for the feedback !!

PR #3316, opened 2025-05-21
frontend: common: Refactoring common/ imports in frontend directory

upsaurav12:
in this PR  i have refactored ``/common`` imports to their direct address. 
However there any many other ``/common`` imports that are still there i will make further commits.


illume (approved):
🎉 🎈 Thanks!

skoeva (approved):

Issue #3250, opened 2025-05-15
ADD More Different Languages that uses most like Russian,German,Arabic,Turkish


eyeaadil:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
[A clear and concise description of what the problem or need is. E.g. I need to...]

## Describe the solution you'd like
[A clear and concise description of what you want to happen.]

## What users will benefit from this feature?
[What type of users may be interested in this feature? What kind of Headlamp setup is likely used. (e.g. WSL users / Helm chart users / In-Cluster users / Plugin developers)]

## Are you able to implement this feature?
[Yes (I will propose a PR) / No.]
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

PR #3318, opened 2025-05-21
frontend: app: i18n: Add Russian language

eyeaadil:
Russian language support has been added to the Headlamp website to make it easier for Russian-speaking users to navigate and interact with the site.
This resolves issue no:- #3250 

Output:-
image


frontend/src/i18n/locales/hi/translation.json
```diff
@@ -75,12 +75,12 @@
   "General Settings": "सामान्य सेटिंग्स",
   "Version": "संस्करण",
   "Language": "भाषा",
-  "Theme": "थीम",
   "Resource details view": "संसाधन विवरण दृश्य",
   "Number of rows for tables": "तालिकाओं के लिए पंक्तियों की संख्या",
   "Timezone to display for dates": "तिथियों के लिए प्रदर्शित करने का समय क्षेत्र",
   "Sort sidebar items alphabetically": "साइडबार आइटमों को वर्णानुक्रम में सॉर्ट करें",
   "Use evict for pod deletion": "Pod हटाने के लिए निष्कासन का उपयोग करें",
+  "Theme": "थीम",
```
skoeva, line 83: Just want to point out that you should remove the changes to the Hindi translations from this PR, these should be taken care of separately (don't worry if they cause tests to fail here)

eyeaadil: Hi @skoeva ! I've removed the unintended Hindi translation changes from this PR as suggested and rebased the branch against main. Please have a look and let me know if anything else is needed. Thank you!

frontend/src/i18n/locales/ru/app.json
```diff
@@ -0,0 +1,51 @@
+{
+  "Failed to fetch plugin info": "",
+  "An error occurred while fetching plugin info from {{ URL }}.": "",
+  "Yes": "",
```
skoeva, line 4: There are some missing translations here and some key terms are translated in other files but not in others ("Yes" specifically is already translated in a different file but not here)

PR #3325, opened 2025-05-22
frontend: NavigationTabs: Fix to use cloneDeep instead of structuredClone

illume:
There's something structuredClone doesn't handle...

> vendor-mui-Djh-99hi.js:40  DataCloneError: Failed to execute 'structuredClone' on 'Window': Symbol(react.element) could not be cloned.
>   at NavigationTabs (index-CgoiQPYG.js:1119:10657)


sniok (approved):
LGTM. Ideally though we shouldn't mutate anything during render but that's a different topic

PR #3304, opened 2025-05-20
Backend: Improve cluster metadata and add clusterID

vyncent-t:
This PR introduces `clusterID` to be later used for non dynamic cluster actions (rename / delete) only. At current, cluster actions like rename and delete are done by locating the cluster by its `name` field. ClusterID is needed in order to separate non dynamic clusters from being impacted by overlapping action side effects of all other clusters. 

- Expands metadata field for clusters, metadata field now includes:
  - Source  `cluster.metadata.origin.kubeconfig` : File path of the kubeconfig
  - OriginalName `cluster.metadata.originalName` : Original name the cluster was created with
  - ClusterID  `cluster.metadata.clusterID` : Unique ID for finding cluster, source + originalName

---

part 1/3 from - https://github.com/kubernetes-sigs/headlamp/pull/3121

PR #3330, opened 2025-05-22
frontend: Home: Display true kubeconfig origin

vyncent-t:
Displays the actual kubeconfig path for cluster in the `origin` field when on the home view. 

## Home View

image

### How to test actual origin in app mode
- Run headlamp in App mode
- At the home view: notice the specific Kubeconfig path under `Origin` on the table

--- 

part 2/3 from - https://github.com/kubernetes-sigs/headlamp/pull/3121

### DIFF
--- a/frontend/src/components/App/Home/ClusterTable.tsx
+++ b/frontend/src/components/App/Home/ClusterTable.tsx
@@ -112,8 +112,8 @@ export default function ClusterTable({
    */
   function getOrigin(cluster: Cluster): string {
     if (cluster.meta_data?.source === 'kubeconfig') {
-      const kubeconfigPath = process.env.KUBECONFIG ?? '~/.kube/config';
-      return `Kubeconfig: ${kubeconfigPath}`;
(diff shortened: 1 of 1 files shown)


illume:
#3304 has been merged

illume:
@vyncent-t  Does this still work with process.env.KUBECONFIG set before running the backend? Can you please confirm if you've tested this?

I don't know if the default hardcoded path is necessary or not? Maybe it's better to show it as nothing if it's reporting nothing?

skoeva:
Theoretically the hardcoded default could be wrong, I think this change should suffice

skoeva (approved):

illume:
@vyncent-t  I take that back about the process.env.KUBECONFIG part. That's just a bit wrong I think. That would happen at frontend build time... so it's quite wrong. It's good that process.env.KUBECONFIG is gone, and it's taking the information from the backend.

illume (approved):
👍

PR #3351, opened 2025-05-25
Add a way to get the resource kind from its resource classAdd resource kind method

7908837174:
This PR adds a static getKind() method to the KubeObject class to allow accessing the resource kind from resource classes without instantiation.

Fixes the issue where there was no consistent way to get the kind from a resource class.

Changes:
- Added static getKind() method to KubeObject class  ..
- Added tests to verify the method works correctly...


skoeva:
I don't actually think this PR is necessary since every KubeObject already exposes its kind as a static property. Closing for now

kallal79:
Gotcha.

PR #3357, opened 2025-05-25
frontend: ServiceDetails: Adding story for Details

upsaurav12:
This PR adds new story for service/Details with test case :
Default: In which it will show ``Serviceinfo`` , ``ports`` , ``endpoints``
Error: Will show error message if the API endpoint is wrong.

### DIFF
--- a/frontend/src/components/service/ServiceDetails.stories.tsx
+++ b/frontend/src/components/service/ServiceDetails.stories.tsx
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


upsaurav12:
@illume should i consider more test cases ?

illume:
For the commit message, I think this should be the context part: `frontend: ServiceDetails:`

illume:
If you run `make frontend-test` from the root of the repo you'll see it prints out some code coverage.

`cd frontend && npm run test -- --coverage`

From that you could see if there's any more major code paths that might need covering with more story states.

upsaurav12:
Yes i am seeing there are plenty of files that even have 0% of coverage. will make further PR's for coverage improvements.

PR #3362, opened 2025-05-25
frontend: storybook: Fix circular dependency issue

illume:
- **frontend: App: Extract defaultAppThemes to its own file**
- **frontend: .storybook: Fix circular dependency issue**


illume:
@sniok  Added App back here: https://github.com/kubernetes-sigs/headlamp/pull/3378

Issue #2159, opened 2024-07-15
app: "npm start" doesn't kill background server on quit anymore


illume:
```
cd app
npm start

# quit app by clicking on close icon
# notice that the background server is still running, and need to ctrl+c it
```

I guess it's related to vite upgrade, but haven't bisected yet.

PR #3336, opened 2025-05-23
plugin bundling scripts need to check checksum when downloading #2539Add checksum verification

7908837174:
Related files:

build-manifest.json
app-build-manifest.json
app/scripts/setup-plugins.js
container/fetch-plugins.sh

PR #3363, opened 2025-05-25
Fix server termination- app: "npm start" doesn't kill background server on quit anymore #2159

7908837174:
Added process.on('exit') handler in start.js to kill all child processes when the parent process exits.
Added process.on('uncaughtException') handler in start.js to kill all child processes on errors.
Added quitServerProcess() call to the window 'closed' event handler in main.ts.
Added app.on('before-quit', quitServerProcess) event handler in main.ts.
Enhanced quitServerProcess() function with platform-specific termination signals (SIGINT/SIGTERM).
Added timeout-based force kill (SIGKILL) if graceful termination fails after 2 seconds.


illume:
Hello again.

thank you for this.

However, as mentioned before... we can't accept any of your pull requests if there are lots of unrelated commits in them, and the CLA isn't signed. As per my comment here:
https://github.com/kubernetes-sigs/headlamp/pull/3336#issuecomment-2903644119

skoeva (commented):
Unclear why these changes are necessary, PR is stale and we are unable to accept changes without a signed CLA

Closing for now

Issue #1821, opened 2024-03-15
Clustering of CRDS


henrysachs:
In the Current CRD View all CRDs are listed but with a lot of CRDs this can get a bit messy. Would It be possible to have a collapseable table like in openlens? 

![Bildschirmfoto 2024-03-15 um 12 00 49](https://github.com/headlamp-k8s/headlamp/assets/17173951/fec77d06-50fe-40c4-823e-2e9d07f92770)

I have like 150 CRDs in my Cluster and they dont fit a single page, but this would help me navigating through the cluster a lot!

illume:
@henrysachs 
What do you think? https://github.com/kubernetes-sigs/headlamp/pull/3366
There's a screen shot in the pr by @farodin91

illume:
This was done in v0.35.0

PR #3366, opened 2025-05-26
frontend: expand crs on sidebar

farodin91:
Idea to make CRs more accessible:

image

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -149,10 +149,16 @@ interface AuthRouteProps {
 }
 
 function AuthRoute(props: AuthRouteProps) {
-  const { children, sidebar, requiresAuth = true, requiresCluster = true, ...other } = props;
-
+  const {
+    children,
+    sidebar,
+    requiresAuth = true,
+    requiresCluster = true,
+    computedMatch = {},
+    ...other
+  } = props;
   const redirectRoute = getCluster() ? 'login' : 'chooser';
-  useSidebarItem(sidebar);
+  useSidebarItem(sidebar, computedMatch);
   const cluster = useCluster();
   const query = useQuery({
     queryKey: ['auth', cluster],
--- a/frontend/src/components/Sidebar/ListItemLink.tsx
+++ b/frontend/src/components/Sidebar/ListItemLink.tsx
@@ -37,6 +37,7 @@ interface ListItemLinkProps {
   icon?: IconProps['icon'];
   iconOnly?: boolean;
   hasParent?: boolean;
+  level?: number;
   fullWidth?: boolean;
   divider?: boolean;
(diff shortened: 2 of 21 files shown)


ashu8912:
Hii @farodin91 thanks for these changes, I have tried the PR and would like to add a  comment.
1. Why do we need the CRs on the sidebar?? Is it because of the friction to go to the instance page, I ask this because having all the instances like this in the sidebar doesn't look very good.

illume:
A fix for this has been requested by a few people... One issue is here:
- https://github.com/kubernetes-sigs/headlamp/issues/1821

farodin91:
> The difference in the issue is, they are asking for a collapsible menu in the page.

I don't see a difference to the issue. Where do you see the difference?

illume:
@farodin91 I thought they were suggesting putting the table on the page, vs the sidebar?

farodin91:
Okay. In Openlens is it in sidebar.

farodin91:
@sniok I will fix it for this view.

farodin91:
@sniok Multi layer navigation tabs are added.

farodin91:
image

farodin91:
@illume would you like to review?

farodin91:
I will extract two changes into extract PRs -> #3560 & https://github.com/kubernetes-sigs/headlamp/pull/3561

sniok:
multiple tab layers with horizontal scrolling is not a good UX, it was considered before but decided against it

farodin91:
@sniok Do you have a better idea?

farodin91:
Rebased and removed both included changes from #3560 and #3561.

illume:
We're going to release early next week now, and would like try to get this into v0.35.0.

joaquimrocha (changes requested):
I think the functionality is fine and we can handle the UX quirkiness of having 2 levels of tabs later.
I think the CR details view changes should however be added to its own commit, and do not forget to update the snapshots (make frontend-test , then there'll be an option to update the snapshots when it fails) so CI passes.

farodin91:
>  I think the CR details view changes should however be added to its own commit, and do not forget to update the snapshots (make frontend-test , then there'll be an option to update the snapshots when it fails) so CI passes.

@joaquimrocha https://github.com/kubernetes-sigs/headlamp/pull/3855 here is the PR to fix it.

joaquimrocha:
> > I think the CR details view changes should however be added to its own commit, and do not forget to update the snapshots (make frontend-test , then there'll be an option to update the snapshots when it fails) so CI passes.
> 
> @joaquimrocha #3855 here is the PR to fix it.

It's merged now. Thanks.

farodin91:
@joaquimrocha Updated.

joaquimrocha:
/lgtm
/approve

Thank you @farodin91 for another great contribution!

PR #3377, opened 2025-05-27
fix(codeSign) shell command built from environment values

odaysec:
https://github.com/kubernetes-sigs/headlamp/blob/5a334017cc02bb6aec597a2aef2ae66f0b7c6590/app/mac/scripts/codeSign.js#L2-L2

https://github.com/kubernetes-sigs/headlamp/blob/5a334017cc02bb6aec597a2aef2ae66f0b7c6590/app/mac/scripts/codeSign.js#L17-L18

Fix the issue will replace the use of `execSync` with `execFileSync`, which allows us to pass the command and its arguments separately. This approach avoids shell interpretation of the arguments, mitigating the risk of command injection. Specifically:
1. The `codesign` command will be passed as the first argument to `execFileSync`.
2. The dynamic components (`teamID`, `entitlementsPath`, and `config.app`) will be passed as separate arguments in an array.
3. This change ensures that special characters in the inputs are treated as literal values and not interpreted by the shell.

### DIFF
--- a/app/mac/scripts/codeSign.js
+++ b/app/mac/scripts/codeSign.js
@@ -1,5 +1,5 @@
 require('dotenv').config();
-const { execSync } = require('child_process');
+const { execFileSync } = require('child_process');
 const path = require('path');
 
 exports.default = async function codeSign(config) {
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉🎈 thanks!

Issue #976, opened 2023-02-21
documentation for 'frontend' functions and data structures


illume:
Parts of `frontend/` are undocumented. The benefits are better quality (less bugs) and it makes code easier to understand.

See:
- [tsdoc](https://tsdoc.org/) 

## A list of some functions and data structures that need documenting

- [ ] headlamp/frontend/src/lib/k8s/cluster.ts@KubeMetadata
- [ ] headlamp/frontend/src/i18n/ThemeProviderNexti18n.tsx@ThemeProviderNexti18n
- [ ] headlamp/frontend/src/i18n/LocaleSelect/LocaleSelect.tsx@LocaleSelectProps
- [ ] headlamp/frontend/src/components/common/Tabs.tsx@TabsProps

There are more functions and fields than is represented in the list above. Feel free to add or improve docs for other functions. Additionally, some of the existing comments could be improved.

RISHIKESHk07:
@illume  would like to contribute

harshita9104:
Hello @illume ,
I am harshita from India, currently working as Technical Writer at Atlan.
I would like to contribute to this issue
Please assign it to me

Jineshbansal:
Hi @illume, I have raised the PR continuing on the basis of commit [bb02e63](https://github.com/headlamp-k8s/headlamp/commit/bb02e631c27dc8b6040a31a879bcf952e8ab86cc). It incorporates all the files needed. Kindly review.

IITI-tushar:
@illume 
Sir i will write documentation for `headlamp/frontend/src/lib/k8s/cluster.ts@KubeMetadata`
Could you let me know if there any other functions or Data Structures needs documenting??

IITI-tushar:
looks like cluster.ts is well documented.
gonna find something else to work on

PR #3379, opened 2025-05-28
frontend:documentation: for 18n and lib/auth.ts

IITI-tushar:
refers: #976

### DIFF
--- a/frontend/src/lib/auth.ts
+++ b/frontend/src/lib/auth.ts
@@ -42,20 +42,9 @@ export function getToken(cluster: string) {
  * Assumes the token is a JWT and extracts the payload.
  * @param cluster - The name of the cluster for which to get user information.
  * @returns The parsed user information object from the token's payload.
- * @throws Error if the token is not a valid JWT or cannot be parsed.
  */
 export function getUserInfo(cluster: string) {
-  const token = getToken(cluster);
-  if (!token) {
-    // Or handle this case as per your application's requirements,
-    // e.g., return null or throw a more specific error.
-    console.warn(`No token found for cluster: ${cluster}`);
-    return null;
-  }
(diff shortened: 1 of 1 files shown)


frontend/src/i18n/LocaleSelect/LocaleSelect.tsx
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+/*
+ *This component likely renders a dropdown menu or a list that allows the user to choose their preferred language for the application.
```
illume, line 18: I'm not sure about this one. 

Maybe it is better placed above the LocaleSelect component? The component already has a comment... though whilst sparse, I think covers it. But maybe it could be improved a little?

frontend/src/i18n/i18next-parser.config.js
```diff
@@ -14,34 +14,55 @@
...
-    // missing values.
+    // For English, use the key itself as the translation.
     if (locale === 'en') {
       const contextSepIdx = key.indexOf(sharedConfig.contextSeparator);
-      if (contextSepIdx >= 0) {
-        return key.substring(0, contextSepIdx);
-      }
-      return key;
+      // If there's a context (e.g. key_plural), use only the base key.
+      return contextSepIdx >= 0 ? key.substring(0, contextSepIdx) : key;
```
illume, line 63: Can you please use a separate commit for this code change?

illume (changes requested):
Hello. Thanks for this.

I left a couple of notes. Would you mind splitting this into a commit for each of the different files? Plus a separate one for the code change made.

IITI-tushar:
@illume 
Sure.

skoeva:
It appears these files have already been updated, closing this now

PR #3384, opened 2025-05-29
docs: eks: Clarify OIDC setup for non-Cognito and ingress users

illume:
Add guidance for users configuring Headlamp with non-Cognito OIDC providers or behind ingress controllers. Covers redirect URI construction, required headers, and EKS trust setup.

This change aims to prevent common misconfigurations and improve the onboarding experience for users deploying Headlamp.

Here's a screen shot of the extra info block I added to the fifth point:

image


illume:
Thanks @ashu8912

PR #3385, opened 2025-05-29
docs: oidc: Add guidance for large JWTs and ingress header limits

### LINKED ISSUE #1832: Document ingress header setting required for nginx ingress
Hi @illume, yes you were right, we have researched it a little bit deeper and figured out that the issue was with the Ingress Nginx controller that we have in front of the Headlamp. We have figured out that after we have increased the large_client_header_buffers (default was 4 buffers for 8K) property of the Headlamp Ingress through server-snippet annotation, the issue no longer happens. So maybe it would be useful to add a note to the docs for the headlamp that in case JWT token is long (>8K) you most probably need to increase this parameter:

```
nginx.ingress.kubernetes.io/server-snippet: |-
      large_client_header_buffers 4 64k;
```

Though its still interesting why regular HTTP requests were working fine, but WebSockets not, maybe during WebSocket connections Headers are larger then for HTTP. Anyway I think we have figured out how that could be resolved in case of Ingress Nginx Controller in front of Headlamp, so we could close this issue I believe.

Thanks!

_Originally posted by @dioniseo in https://github.com/headlamp-k8s/headlamp/issues/1827#issuecomment-2008171963_

illume:
Add a troubleshooting section to the generic OIDC documentation covering issues with large JWT tokens when using Headlamp behind an Ingress NGINX controller. Includes a recommended configuration to increase `large_client_header_buffers` to prevent WebSocket failures caused by oversized headers.

Also expands the callback URL section with a note about forwarding the `X-Forwarded-Proto` header to avoid redirect URI mismatches when using HTTPS behind a proxy.

These changes aim to help users avoid common ingress-related authentication issues when deploying Headlamp with OIDC.

image

The first issue was added to the EKS doc, but it's general advice:
- https://github.com/kubernetes-sigs/headlamp/pull/3384

This fixes #1832 from another user discussion:
- https://github.com/kubernetes-sigs/headlamp/issues/1832

### DIFF
--- a/docs/installation/in-cluster/oidc.md
+++ b/docs/installation/in-cluster/oidc.md
@@ -17,6 +17,38 @@ To use OIDC, Headlamp needs to know how to configure it, so you have to provide 
 and you have to tell the OIDC provider about the callback URL, which in Headlamp it is your URL + the `/oidc-callback` path, e.g.:
 `https://YOUR_URL/oidc-callback`.
 
+### Callback URL
+
+You must tell your OIDC provider the callback URL that Headlamp will use after login. This is your Headlamp URL plus `/oidc-callback`, for example:
+
+```
+https://YOUR_URL/oidc-callback
+```
+
+> **ℹ️ Note:** If you're running Headlamp behind an ingress or load balancer (e.g., NGINX, AWS ALB/NLB), make sure it forwards the `X-Forwarded-Proto` header. Otherwise, Headlamp may generate the callback URL using `http` instead of `https`, which can cause a mismatch with your OIDC provider.
+>
+> For [NGINX ingress](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/), you can add:
+>
+> ```yaml
+> nginx.ingress.kubernetes.io/configuration-snippet: |
+>   proxy_set_header X-Forwarded-Proto $scheme;
+> ```
+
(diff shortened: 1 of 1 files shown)


illume:
@yolossn and @ashu8912 Please review?

docs/installation/in-cluster/oidc.md
```diff
@@ -17,6 +17,38 @@ To use OIDC, Headlamp needs to know how to configure it, so you have to provide
...
+
+You must tell your OIDC provider the callback URL that Headlamp will use after login. This is your Headlamp URL plus `/oidc-callback`, for example:
+
+```
+https://YOUR_URL/oidc-callback
+```
+
+> **ℹ️ Note:** If you're running Headlamp behind an ingress or load balancer (e.g., NGINX, AWS ALB/NLB), make sure it forwards the `X-Forwarded-Proto` header. Otherwise, Headlamp may generate the callback URL using `http` instead of `https`, which can cause a mismatch with your OIDC provider.
+>
+> For NGINX ingress, you can add:
```
ashu8912: maybe we can add a pointer to the nginx documentation here like

See [NGINX Ingress Annotations Documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/) for more details.

illume: Thanks. Added the link.

PR #3389, opened 2025-05-29
frontend: Map: Add weights to GraphNodes

### LINKED ISSUE #3369: Map: Add ability for plugins to customize main node behavior in node group
Currently it's not possible to pick which node is the 'main' node. Right now it's hardcoded to put more 'important' nodes on the top. It's defined in the getMainNode() function graphGrouping.tsx:132. We should add some kind of 'weight' property (kind of like z-index in css) to determine which node is on top.

More context here https://github.com/headlamp-k8s/plugins/pull/254

Yes but if anyone else wants to take a look feel free to leave a comment

adwait-godbole:
Fixes #3369 

1) Before Weights:

    
Screenshot from 2025-05-30 01-13-41

After Weights:

Screenshot from 2025-05-30 01-15-15

      
2) Before Weights:
  

[Screencast from 29-05-25 08:25:38 PM IST.webm](https://github.com/user-attachments/assets/7dfff547-dab8-40f2-9fc7-7ffb68c5ffa6)

After Weights:

[Screencast from 29-05-25 08:27:19 PM IST.webm](https://github.com/user-attachments/assets/c98fc3e3-449a-42da-960d-c2e69314c19a)

cc @sniok @yolossn

### DIFF
--- a/frontend/src/components/resourceMap/graph/graphGrouping.tsx
+++ b/frontend/src/components/resourceMap/graph/graphGrouping.tsx
@@ -19,7 +19,7 @@ import Namespace from '../../../lib/k8s/namespace';
 import Node from '../../../lib/k8s/node';
 import Pod from '../../../lib/k8s/pod';
 import { makeGraphLookup } from './graphLookup';
-import { forEachNode, GraphEdge, GraphNode } from './graphModel';
+import { forEachNode, getNodeWeight, GraphEdge, GraphNode } from './graphModel';
 
 export type GroupBy = 'node' | 'namespace' | 'instance';
 
@@ -113,7 +113,7 @@ const getConnectedComponents = (nodes: GraphNode[], edges: GraphEdge[]): GraphNo
       findConnectedComponent(node, componentNodes, componentEdges);
       const mainNode = getMainNode(componentNodes);
 
-      const id = 'group-' + mainNode.id;
+      const id = 'group-' + (mainNode?.id ?? 'unknown');
       components.push({
         id: id,
         nodes: componentNodes,
@@ -126,18 +126,34 @@ const getConnectedComponents = (nodes: GraphNode[], edges: GraphEdge[]): GraphNo
 };
 
 /**
- * Try to find a "main" node in the workload group
- * If can't find anything return the first node
+ * Try to find a "main" node in the group based on weight.
+ * Higher weight nodes are considered more important.
+ *
+ * @param nodes - Array of nodes to find the main node from
+ * @returns The node with the highest weight (most important), or undefined if array is empty
  */
-export const getMainNode = (nodes: GraphNode[]) => {
-  const deployment = nodes.find(it => it.kubeObject?.kind === 'Deployment');
-  const replicaSet = nodes.find(it => it.kubeObject?.kind === 'ReplicaSet');
-  const daemonSet = nodes.find(it => it.kubeObject?.kind === 'DaemonSet');
-  const statefulSet = nodes.find(it => it.kubeObject?.kind === 'StatefulSet');
-  const job = nodes.find(it => it.kubeObject?.kind === 'Job');
-  const cronJob = nodes.find(it => it.kubeObject?.kind === 'CronJob');
+export const getMainNode = (nodes: GraphNode[]): GraphNode | undefined => {
+  if (nodes.length === 0) {
+    return undefined;
+  }
 
-  return deployment ?? replicaSet ?? daemonSet ?? statefulSet ?? cronJob ?? job ?? nodes[0];
+  if (nodes.length === 1) {
+    return nodes[0];
+  }
+
+  // Find node with the highest weight
+  let mainNode = nodes[0];
+  let maxWeight = getNodeWeight(mainNode);
+
+  for (let i = 1; i < nodes.length; i++) {
+    const currentWeight = getNodeWeight(nodes[i]);
+    if (currentWeight > maxWeight) {
+      maxWeight = currentWeight;
+      mainNode = nodes[i];
+    }
+  }
+
+  return mainNode;
 };
 
 /**
@@ -189,7 +205,7 @@ const groupByProperty = (
 
 /**
  * Groups the graph into separate 'group' Nodes
- * Nodes within groups are sorted by size
+ * Nodes within groups are sorted by weight and size
  *
  * @param nodes - List of nodes
  * @param edges - List of edge
@@ -273,7 +289,7 @@ export function groupGraph(
       node => {
         if (node.nodes) {
           const mainNode = getMainNode(node.nodes.filter(node => !node.nodes) as GraphNode[]);
-          return mainNode.kubeObject?.metadata?.labels?.['app.kubernetes.io/instance'];
+          return mainNode?.kubeObject?.metadata?.labels?.['app.kubernetes.io/instance'];
         }
         return node.kubeObject?.metadata?.labels?.['app.kubernetes.io/instance'];
       },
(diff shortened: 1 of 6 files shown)


frontend/src/components/resourceMap/graph/graphGrouping.tsx
```diff
@@ -126,18 +126,49 @@ const getConnectedComponents = (nodes: GraphNode[], edges: GraphEdge[]): GraphNo
...
+  if (nodes.length === 0) {
+    return undefined;
+  }
+
+  if (nodes.length === 1) {
+    return nodes[0];
+  }
+
+  // sort nodes by weight (descending) and return the highest weighted node
+  const sortedByWeight = [...nodes].sort((a, b) => {
```
sniok, line 146: Could you rewrite this function so that it doesn't create a copy and doesn't sort them all? It's possible to find a node with the biggest weight in linear time without allocating a new array

frontend/src/components/resourceMap/graph/graphModel.tsx
```diff
@@ -116,3 +122,66 @@ export interface Relation {
...
+
+  // Configuration (lower priority - supporting resources)
+  ConfigMap: 250,
+  Secret: 240,
+  MutatingWebhookConfiguration: 150,
+  ValidatingWebhookConfiguration: 140,
+
+  // Default weight for unknown types
+  default: 50,
+} as const;
```
adwait-godbole, line 165: @sniok, any comments on these weights? any changes you feel like?

sniok:
I've noticed that the order is different, for example replicaset is now to the right of pods sometimes

image

adwait-godbole:
@sniok, can you please share your YAML spec of this scenario?

sniok:
It's the headlamp deployment from here
https://github.com/kubernetes-sigs/headlamp/blob/main/kubernetes-headlamp.yaml

adwait-godbole:
hi @sniok, can you review once again? I have addressed the issue

frontend/src/components/resourceMap/graph/graphLayout.tsx
```diff
@@ -41,33 +41,83 @@ const layoutOptions = {
...
+ * @param node The graph node to determine the layer for
+ * @param partitionThresholds pre-calculated partition thresholds
+ * @returns The ELK partition number (higher number = further left)
  */
-function getPartitionLayer(node: GraphNode) {
-  if (!('kubeObject' in node)) return;
-  const kind = node.kubeObject?.kind;
-  const partitionLayer = partitionLayers.findIndex(layer => layer.includes(kind));
-  return partitionLayer > -1 ? partitionLayer : undefined;
+function getPartitionLayer(node: GraphNode, partitionThresholds: number[]): number {
```
sniok: would using weight directly as a layer work? I think it would simplify this a lot (if it works of course)

frontend/src/components/resourceMap/graph/graphLayout.tsx
```diff
@@ -122,7 +172,7 @@ function convertToElkNode(node: GraphNode, aspectRatio: number): ElkNodeWithData
       convertedEdges.length > 0
         ? {
             'partitioning.activate': 'true',
-            'elk.direction': 'UNDEFINED', // ELK will automatically pick direction
+            'elk.direction': 'LEFT',
```
sniok: we want to keep the direction as undefined because it will order it from top to bottom on narrow screens

frontend/src/components/resourceMap/graph/graphModel.tsx
```diff
@@ -116,3 +122,82 @@ export interface Relation {
...
+ * Default node weight assignments for different Kubernetes resource types.
+ * Higher weight = higher priority/importance in graph layout.
+ */
+const DEFAULT_NODE_WEIGHTS = {
+  // Tier 1: Top-Level Orchestration & Scaling
+  HorizontalPodAutoscaler: 1000,
+
+  // Tier 2: Primary Workload Controllers
+  Deployment: 970,
+  StatefulSet: 970,
```
sniok: stateful set and daemon set should have lower weight, deployment is more 'important'

sniok: please double check the order of the resources here, so the order from 'main' branch is preserved

```tsx
const partitionLayers = [
  ['Deployment'],
  ['ReplicaSet', 'ServiceAccount', 'CronJob', 'DaemonSet', 'StatefulSet'],
  ['Job'],
  ['Pod', 'RoleBinding'],
  ['Service', 'NetworkPolicy', 'Role'],
  ['Endpoints'],
];
```

sniok (changes requested):
thanks! left a couple of comments

frontend/src/components/resourceMap/graph/graphLayout.tsx
adwait-godbole: @sniok, the below function that I added over [here](https://github.com/kubernetes-sigs/headlamp/blob/ccd0bcb3b80e2ca34fa23d78813e53f0ea22bdef/frontend/src/components/resourceMap/graph/graphLayout.tsx#L44-#L65) essentially does the same thing that you're mentioning of "constructing" and using weights directly. Just that I have called this function as `createPartitionThresholds` in this file. We need to [construct the weights/partition thresholds once before creating the partition layer dynamically](https://github.com/kubernetes-sigs/headlamp/blob/ccd0bcb3b80e2ca34fa23d78813e53f0ea22bdef/frontend/src/components/resourceMap/graph/graphLayout.tsx#L108-#L122) because if someone provides custom weights for nodes which are not present in our `DEFAULT_NODE_WEIGHTS` then the partition layer logic does not place them correctly on the layout. This was an observation when I was trying to add custom weights for KEDA's ScaledObject resource

```typescript
/**
 * Creates partition thresholds by merging default + custom weights for graph layering.
 *
 * Example result for graph with KEDA ScaledObject:
 * - Input: Graph contains ScaledObject (custom weight: 2000), Deployment (default weight: 970)
 * - Output: [2000, 1000, 970, 940, 920, 850, ...]
 * - ScaledObject gets partition 11, Deployment gets partition 9
 *
 * @param node The root graph node to analyze for default + custom weights
 * @returns Merged partition thresholds array sorted descending (higher weight = higher priority)
 */
function createPartitionThresholds(rootNode: GraphNode): number[] {
  const weights = new Set<number>();

  // Collect weights of all nodes
  forEachNode(rootNode, node => weights.add(getNodeWeight(node)));

  // Merge: default + custom weights, sorted descending
  const partitionThresholds = Array.from(weights).sort((a, b) => b - a);

  return partitionThresholds;
}
```

adwait-godbole: I can maybe change this function's name to `getAllNodeWeights`? and then get rid of the variable `partitionThresholds` throughout the file? If that is what is making it confusing?

frontend/src/components/resourceMap/graph/graphLayout.tsx
adwait-godbole: fixed this

frontend/src/components/resourceMap/graph/graphModel.tsx
adwait-godbole: fixed this

frontend/src/components/resourceMap/graph/graphLayout.tsx
adwait-godbole: @sniok lmk about this one

sniok: I was thinking about something like this 

```tsx
function getPartitionLayer(node: GraphNode): number {
  return -getNodeWeight(node);
}
```

I've quickly tried it out and it appears to be working fine

image

and if a node has a custom weight, it should also work fine?
I hope it makes sense, unless I'm missing something

adwait-godbole: Yes this is indeed a much much simpler way of doing this. I have made this fix. Thanks.

adwait-godbole:
@sniok, anything else left to address? can we merge?

sniok (approved):
lgtm, thanks!

sniok:
Thanks again for the contribution, it works great!

PR #3392, opened 2025-05-29
frontend: KubeObjectGlance: Introduce registerKubeObjectGlance

### LINKED ISSUE #3368: Map: Add ability for plugins to customize Node glance view
We should add a way of configuring the on hover display text for the nodes.
Adding a `registerKubeObjectGlance(glance)` where glance is something like 
```
id: string;
component: ({ kubeObject }) => ReactNode;
```

Requested in https://github.com/headlamp-k8s/plugins/pull/254

Yes but if anyone else wants to take a look feel free to leave a comment

userAdityaa:
This PR introduces `registerKubeObjectGlance` to allow dynamic registration of hover preview components for Kubernetes and non-Kubernetes nodes in Headlamp.

One Can Create a KubeObjectGlance like this: 

```tsx
const glance: Glance = {
  id: 'default-glance',
  component: (): React.ReactNode => (
    <div>Default Glance</div>
  ),
};
```

Fixes #3368


frontend/src/redux/glanceSlice.ts
```diff
@@ -0,0 +1,44 @@
...
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { ReactNode } from 'react';
+
+export interface Glance {
+  id: string;
+  component: (props: { resource: any }) => ReactNode;
+}
+
+export interface GlanceState {
+  glances: { [id: string]: Glance };
```
sniok, line 26: this state should be put in the existing `graphViewSlice`

frontend/src/components/resourceMap/KubeObjectGlance/KubeObjectGlance.tsx
```diff
@@ -35,44 +37,52 @@ import { ReplicaSetGlance } from './ReplicaSetGlance';
 import { ServiceGlance } from './ServiceGlance';
 
 /**
- * Little Popup preview of a Kube object
+ * Little Popup preview of a node (Kubernetes object or other)
  */
-export const KubeObjectGlance = memo(({ resource }: { resource: KubeObject }) => {
+export const KubeObjectGlance = memo(({ resource }: { resource: any }) => {
```
sniok, line 42: KubeObjectGlance will only render when it is a KubeObject, please check out `KubeObjectNode` component where it's used. I would suggest introducing a generic `NodeGlance` that will either render a KubeObjectGlance or a plugin provided glance from the redux state. And adding it to the KubeObjectNode component

frontend/src/components/resourceMap/KubeObjectGlance/KubeObjectGlance.tsx
```diff
@@ -35,44 +37,52 @@ import { ReplicaSetGlance } from './ReplicaSetGlance';
...
+      sections.push(<ServiceGlance service={resource as Service} key={kind} />);
+    } else if (kind === 'Endpoints') {
+      sections.push(<EndpointsGlance endpoints={resource as Endpoints} key={kind} />);
+    } else if (kind === 'ReplicaSet' || kind === 'StatefulSet') {
+      sections.push(<ReplicaSetGlance set={resource as ReplicaSet} key={kind} />);
+    } else if (kind === 'HorizontalPodAutoscaler') {
+      sections.push(<HorizontalPodAutoscalerGlance hpa={resource as HPA} />);
+    }
+  } else {
+    sections.push(
```
sniok, line 77: this section is not necessary

frontend/src/plugin/registry.tsx
```diff
@@ -194,6 +195,10 @@ export default class Registry {
     store.dispatch(setDetailsViewHeaderAction(actionFunc));
   }
 
+  registerKubeObjectGlance(glance: Glance) {
```
sniok, line 198: adding a register method here is not required, it's deprecated. Having a register function that you've added on line 305 is enough

frontend/src/plugin/registry.tsx
```diff
@@ -297,6 +302,10 @@ export function registerSidebarEntry({
   );
 }
 
+export function registerKubeObjectGlance(glance: Glance) {
```
sniok, line 305: These kind of functions are used by plugin authors and so require a thorough documentation. Please add a description with an example. Check out how other reigster functions are documented

sniok: In addition to this, we also have a website with the documentation, a markdown file `extending-the-map.md` contains explanation on how to extend the map and this function should be added there.
https://headlamp.dev/docs/latest/development/plugins/functionality/extending-the-map
https://github.com/kubernetes-sigs/headlamp/blob/main/docs/development/plugins/functionality/extending-the-map.md

frontend/src/redux/glanceSlice.ts
userAdityaa: I have removed `glanceSlice.ts` and it's test and added the GlanceState in `graphViewSlice`.

frontend/src/components/resourceMap/KubeObjectGlance/KubeObjectGlance.tsx
userAdityaa: Remove the fallback for the unknown case.

userAdityaa:
This change only affects `registerKubeObjectGlance`. I've grouped all related modifications into a single commit for simplicity. Let me know if you'd prefer these changes split into multiple commits.

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Generic glance component for rendering node previews in Headlamp's graph view.
+ * Renders KubeObjectGlance for Kubernetes objects or a plugin-provided glance for other nodes.
+ */
+export const NodeGlance = ({ resource }: { resource: any }) => {
+  const glances = useSelector((state: RootState) => state.graphView.glances);
+  const kind = resource?.kind || resource?.type;
+
+  // Check for a registered glance component
+  const RegisteredGlance = glances[kind]?.component;
```
sniok, line 31: this doesn't seem right, glances is a record where the key is id, but here you're using objects' `kind` for a lookup

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useSelector } from 'react-redux';
+import { KubeObject } from '../../../lib/k8s/cluster';
+import { RootState } from '../../../redux/stores/store';
+import { KubeObjectGlance } from './KubeObjectGlance';
+
+/**
+ * Generic glance component for rendering node previews in Headlamp's graph view.
+ * Renders KubeObjectGlance for Kubernetes objects or a plugin-provided glance for other nodes.
+ */
+export const NodeGlance = ({ resource }: { resource: any }) => {
```
sniok, line 26: instead of resources, it should accept a `node` property. And it should have a `GraphNode` type instead of `any`

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const glances = useSelector((state: RootState) => state.graphView.glances);
+  const kind = resource?.kind || resource?.type;
```
sniok, line 28: we don't have `resource.type` field

frontend/src/components/resourceMap/graphViewSlice.tsx
```diff
@@ -35,14 +35,21 @@ export interface IconDefinition {
...
 
+export interface Glance {
+  id: string;
+  component: (props: { resource: any }) => ReactNode;
+}
+
 export interface GraphViewSliceState {
   graphSources: GraphSource[];
   kindIcons: Record<string, IconDefinition>;
+  glances: { [id: string]: Glance };
```
sniok, line 46: please use `Record` type like in kindIcons, for consistency

frontend/src/components/resourceMap/graphViewSlice.tsx
```diff
@@ -59,7 +66,11 @@ export const graphViewSlice = createSlice({
     addKindIcon(state, action: PayloadAction<{ kind: string; definition: IconDefinition }>) {
       state.kindIcons[action.payload.kind] = action.payload.definition;
     },
+    setGlance(state, action: PayloadAction<Glance>) {
+      state.glances[action.payload.id] = action.payload;
```
sniok, line 70: please consider the case when plugin registers multiple glances for the same kind or a node that does not represent a kubeObject at all, maybe we shouldn't have a lookup by kind at all and render custom glances in all nodes, and then based on the node data plugins can decide for themselves if they should render anything or just return null

sniok (changes requested):
Hi, I've left some more comments

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
userAdityaa: Apologies, that was an oversight on my part.

frontend/src/components/resourceMap/graphViewSlice.tsx
```diff
@@ -35,14 +35,21 @@ export interface IconDefinition {
   color?: string;
 }
 
+export interface Glance {
```
sniok, line 41: please add documentation for this interface and its' fields

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * Generic glance component for rendering node previews in Headlamp's graph view.
+ * Renders KubeObjectGlance for Kubernetes objects or a plugin-provided glance for other nodes.
+ */
+export const NodeGlance = ({ node }: { node: GraphNode }) => {
+  const glances = useSelector((state: RootState) => state.graphView.glances);
+
+  const glanceResults = Object.values(glances).map(glance => {
+    const GlanceComponent = glance.component;
+    return (
+      <Fragment key={glance.id}>
```
sniok: there's no need for a fragment here, `key` can be put in the GlanceComponent

frontend/src/components/resourceMap/KubeObjectGlance/NodeGlance.tsx
```diff
@@ -0,0 +1,52 @@
...
+    );
+  });
+
+  const validGlanceResults = glanceResults.filter(result => result !== null);
+  if (validGlanceResults.length > 0) {
+    return <>{validGlanceResults}</>;
+  }
+
+  // Render KubeObjectGlance for Kubernetes objects
+  if (node.kubeObject instanceof KubeObject) {
```
sniok, line 38: I think we should display default KubeObejctGlance alongside custom glances

docs/development/plugins/functionality/extending-the-map.md
```diff
@@ -160,3 +160,35 @@ This is useful when you want to:
 - Show custom visualizations for your resources
 - Display data from external sources alongside Kubernetes resources
 - Create interactive detail views specific to your use case
+
+## Custom Glance Component
+
+When hovering over a node in the map, a preview (or "glance") is displayed. For Kubernetes resources, Headlamp provides default glance components. You can register a custom glance component for any node type (Kubernetes or custom) using `registerKubeObjectGlance`.
```
sniok, line 166: It would be nice to have a screenshot here, to better explain what is "glance"

docs/development/plugins/functionality/extending-the-map.md
```diff
@@ -160,3 +160,35 @@ This is useful when you want to:
 - Show custom visualizations for your resources
 - Display data from external sources alongside Kubernetes resources
 - Create interactive detail views specific to your use case
+
+## Custom Glance Component
```
sniok, line 164: documentation changes should be in a separate commit

sniok (changes requested):
hi, thanks for the update! I've left a couple of additional comments

userAdityaa:
I have pushed the changes, the summary of the Changes: 

1. The `Fragment` component has been removed, and the `key` is now directly assigned to `GlanceComponent`.
2. Added documentation for the `Glance` interface and its fields.
3. Both `Custom` Glances and `KubeObject` Glances are now pushed into a result list, which is then returned for display, if result is empty then null is returned. 
4. Extending the map docs now includes image for glance and has been committed separately.

sniok:
oh one more thing that I forgot to mention, could you also add an example how to use this in the `plugins/examples/customizing-map`? It could be the same code you've used in the documentation

userAdityaa:
> oh one more thing that I forgot to mention, could you also add an example how to use this in the `plugins/examples/customizing-map`? It could be the same code you've used in the documentation

I have now added the documentation code in `plugins/examples/customizing-map`.

PR #3394, opened 2025-05-29
docs: oidc: Add guidance for large JWTs and ingress header limits

### LINKED ISSUE #1832: Document ingress header setting required for nginx ingress

7908837174:
Fixes #1832

This commit adds comprehensive documentation for resolving WebSocket connection issues when using OIDC with large JWT tokens (>8K) behind nginx ingress controller.

Changes:
- Add troubleshooting section to OIDC documentation with symptoms, solution, and alternative configurations
- Include nginx annotation example for large_client_header_buffers
- Add cross-references in Helm chart README and installation guide
- Update sample ingress file with helpful comments

The issue occurs because nginx's default header buffer size (4x8K) is insufficient for large JWT tokens in WebSocket connections, while regular HTTP requests work fine. The solution increases buffer size to 4x64K using nginx.ingress.kubernetes.io/server-snippet annotation.

### DIFF
--- a/charts/headlamp/README.md
+++ b/charts/headlamp/README.md
@@ -198,6 +198,18 @@ ingress:
         - headlamp.example.com
 ```
 
+**Note for nginx ingress users:** If you're using OIDC with large JWT tokens (>8K), you may need to add the following annotation to prevent WebSocket connection issues:
+
+```yaml
+ingress:
+  enabled: true
+  annotations:
+    nginx.ingress.kubernetes.io/server-snippet: |
+      large_client_header_buffers 4 64k;
+```
+
+For more details, see the [OIDC troubleshooting documentation](https://headlamp.dev/docs/latest/installation/in-cluster/oidc/#large-jwt-tokens-and-websocket-connection-issues).
+
 ### Resource Management
 
 | Key | Type | Default | Description |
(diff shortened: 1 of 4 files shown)


7908837174:
Why?

kallal79:
thanks for the update!

PR #3412, opened 2025-05-31
frontend: cluster: Add Overview stories

1012Charan:
Adds 3 Storybook stories for the Overview component:
- EmptyState (shows the UI when there's no data available)
- LoadingState (shows a spinner while data is being fetched)
- ErrorState (simulates when data fails to load)

Screenshots for new stories :
**EmptyState**
Image 20-06-25 at 3 20 PM
**LoadingState**
Image 20-06-25 at 3 20 PM (1)
**ErrorState**
Image 20-06-25 at 3 21 PM

Note: I noticed that the Empty, Loading, and Error states all look quite similar in Storybook — they seem to show the same fallback message or UI. I'm assuming this is expected because the Overview component handles these cases in a similar way with mock data, but please let me know if I might have missed something!

### DIFF
--- a/frontend/src/components/cluster/Overview.stories.tsx
+++ b/frontend/src/components/cluster/Overview.stories.tsx
@@ -192,3 +192,53 @@ const Template: StoryFn = () => {
 };
 
 export const Events = Template.bind({});
+
+export const EmptyState = Template.bind({});
+EmptyState.parameters = {
+  msw: {
+    handlers: {
+      story: [
+        http.get('http://localhost:4466/api/v1/events', () =>
+          HttpResponse.json({
+            kind: 'EventsList',
+            items: [],
+            metadata: {},
+          })
+        ),
+        http.get('http://localhost:4466/api/v1/pods', () =>
+          HttpResponse.json({
+            kind: 'PodList',
+            apiVersion: 'v1',
+            metadata: {},
+            items: [],
+          })
+        ),
+      ],
+    },
+  },
+};
+
+export const LoadingState = Template.bind({});
+LoadingState.parameters = {
+  storyshots: { disable: true },
+  msw: {
+    handlers: {
+      story: [
+        http.get('http://localhost:4466/api/v1/events', () => new Promise(() => {})),
+        http.get('http://localhost:4466/api/v1/pods', () => new Promise(() => {})),
+      ],
+    },
+  },
+};
+
+export const ErrorState = Template.bind({});
+ErrorState.parameters = {
+  msw: {
+    handlers: {
+      story: [
+        http.get('http://localhost:4466/api/v1/events', () => HttpResponse.error()),
+        http.get('http://localhost:4466/api/v1/pods', () => HttpResponse.error()),
+      ],
+    },
+  },
+};
--- a/frontend/src/components/cluster/__snapshots__/Overview.EmptyState.stories.storyshot
+++ b/frontend/src/components/cluster/__snapshots__/Overview.EmptyState.stories.storyshot
@@ -1,0 +1,698 @@
+<body>
+  <div>
+    <div
+      class="MuiContainer-root MuiContainer-maxWidthXl css-4hqp1a-MuiContainer-root"
+    >
+      <div
+        class="MuiGrid-root MuiGrid-container MuiGrid-spacing-xs-1 css-1ub3t8b-MuiGrid-root"
+      >
+        <div
+          class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 css-49904w-MuiGrid-root"
+        >
+          <div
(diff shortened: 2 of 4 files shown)


frontend/src/components/cluster/Overview.stories.tsx
joaquimrocha, line 1: Thanks @1012Charan . Can you adapt the commit title to be more consistent with our current format:
`frontend: Expand Overview stories with empty, loading, and error states`

joaquimrocha (changes requested):

frontend/src/components/cluster/Overview.stories.tsx
1012Charan: Thanks @joaquimrocha, I've updated the commit message as suggested. Please let me know if anything else is needed.

skoeva (commented):
Looks solid, would be nice to have screenshots in the description for the new stories

Also good to rebase against main to pull the latest changes

1012Charan:
Thanks for the suggestion @skoeva! I've updated this branch added the screenshots, but I noticed that all three states — Empty, Loading, and Error — look almost the same in Storybook. I’m not sure if that’s okay or if I might have missed something, so please let me know if any changes are needed!

skoeva (dismissed):
The similar states should be fine, I would just squash the commits once you're done rebasing. Make sure to update the snapshots too: run `make frontend-test` locally and you can see which are failing

1012Charan:
@skoeva I've rebased the branch, squashed the commits, and updated the snapshots as requested.

Since this was my first time working with things like squash and rebase, it was an exciting learning experience — so if I missed anything or made mistakes, please feel free to point them out.

Also, while running the tests, I noticed a failure in the LoadingState due to it simulating an endless request. If you'd like, I can update it to use a timeout-based delay instead.

skoeva:
sure, no need to mock an endless request. the frontend test still seems to be failing, are you seeing something different locally?

1012Charan:
@skoeva You're right — I rechecked and the LoadingState test is still failing on my side too.

I think it's because the mocked API requests never resolve (to simulate loading), so the test waits indefinitely and throws a waiting error. Would it be okay to skip this story in the tests using:

LoadingState.parameters = {
  ...LoadingState.parameters,
  chromatic: { disableSnapshot: true },
};

Also, another issue was simple-eval causing the frontend tests to fail is now fixed — I had missed installing it initially, but after adding it, the other tests are passing fine.

Let me know if this approach for the loading story makes sense or if you'd prefer a different fix!

joaquimrocha:
@skoeva Can you check this one?

package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
```
skoeva: I think you ran `npm install` in the root directory, this file along with `package-lock.json` should be removed

frontend/package-lock.json
```diff
@@ -15405,7 +15405,7 @@
       "version": "2.7.0",
       "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.7.0.tgz",
       "integrity": "sha512-+hSoy/QHluxmC9kCIJyL/uyFmLmc+e5CFR5Wa+bpIhIj85LVb9ZH2nVnqrHoSvKogwODv0ClqZkmiSSaIH5LTA==",
-      "devOptional": true,
```
skoeva: this change does not need to be made

skoeva:
@1012Charan Fixed the CI failures and removed the unnecessary changes, feel free to test locally and lmk if you notice anything

1012Charan:
Hi @skoeva ,
  I pulled the latest changes and ran the frontend tests locally, and I'm happy to say they're all passing now! So, it looks like those CI issues you mentioned are all cleared up from my end.
I also took a look at the new Overview stories in Storybook (EmptyState, LoadingState, and ErrorState). As we talked about, they present a pretty similar visual output, but I totally get that's the intended behavior for these states.
Everything seems to be in order on my side. Just let me know if there's anything else you'd like me to verify or if you need any more help with this PR.
Thanks!

frontend/src/components/cluster/Overview.stories.tsx
```diff
@@ -192,3 +192,67 @@ const Template: StoryFn = () => {
...
+  },
+};
+
+export const LoadingState = Template.bind({});
+LoadingState.parameters = {
+  msw: {
+    handlers: {
+      story: [
+        http.get('http://localhost:4466/api/v1/events', async () => {
+          await delay(200);
```
sniok: this is not recommended, since we take snapshots of these stories they should be deterministic.
adding such delay may cause inconsistencies in the snapshots

skoeva: this can probably just be removed then, the last setup caused the CI failures

PR #3423, opened 2025-06-02
frontend: Add delete cluster button for non dynamic clusters

vyncent-t:
## Desciption
This PR introduces changes to enable deleting non dynamic clusters while at the Headlamp home view. Users now have the option to either remove the cluster from their default kubeconfig (similar to how it is now for other clusters), or to remove the cluster from the actual source kubeconfig file on their machine.

image

## How to test non default kubeconfig delete
- Start headlamp in app mode
- Create a non dynamic cluster
- Click the menu options for said dynamic cluster
image
- Notice the `Delete` button is now available for dynamic clusters

- Locate your kubeconfig for your nondynamic clusters on your machine
- Click delete and view pop up dialog for the delete cluster button
- Click yes and notice the cluster being removed from the kubeconfig file

## How to test default kubeconfig delete
- Start headlamp in app mode
- Create a non dynamic cluster
- Click the menu options for said dynamic cluster
image
- Notice the `Delete` button is now available for dynamic clusters

- Locate your kubeconfig for your nondynamic clusters on your machine
- Click delete and view pop up dialog for the delete cluster button
- Click yes and notice the cluster being removed from the the home view and not the kubeconfig file

---

Reworked from - https://github.com/kubernetes-sigs/headlamp/pull/2567


vyncent-t:
reviewed deleting non dynamic clusters from different config locations with @skoeva 

works as intended 💯

vyncent-t:
was asked to keep this in one PR and not split it by backend/frontend, mind running through the tests one more time like we did before?

@skoeva

backend/pkg/config/config.go
```diff
@@ -193,6 +193,38 @@ func Parse(args []string) (*Config, error) {
...
+		if runtime.GOOS == "windows" {
+			// golang is wrong for config folder on windows.
+			// This matches env-paths and headlamp-plugin.
+			kubeConfigDir = filepath.Join(userConfigDir, "Headlamp", "Config", "kubeconfigs")
+		}
+
+		// Create the directory if it doesn't exist.
+		fileMode := 0o755
+
+		err = os.MkdirAll(kubeConfigDir, fs.FileMode(fileMode))
```
joaquimrocha, line 212: As a good practice, functions which are about getting something (this one seems to be about getting the default kubeconfig persistent dir) shouldn't create something. Maybe you want to call this MakeHeadlampKubeConfigsDir (kubeconfigs persistent dir seems like an external thing somehow), which is what it does (if the dir exists, there's no reason to create it).

backend/pkg/config/config.go
```diff
@@ -193,6 +193,38 @@ func Parse(args []string) (*Config, error) {
...
+		if err == nil {
+			return kubeConfigDir, nil
+		}
+	}
+
+	// if any error occurred, fallback to the current directory.
+	ex, err := os.Executable()
```
joaquimrocha, line 219: Then this function is no longer returning the default kube config persistent dir?
Is this fallback a good idea?

backend/pkg/config/config.go
```diff
@@ -224,6 +224,14 @@ func DefaultKubeConfigPersistenceDir() (string, error) {
 	return "", fmt.Errorf("failed to get default kubeconfig persistence directory: %v", err)
 }
 
+func DefaultKubeConfigPersistenceFile() (string, error) {
```
joaquimrocha: Maybe DefaultHeadlampKubeConfigFile ?

backend/pkg/config/config.go
```diff
@@ -224,6 +224,14 @@ func DefaultKubeConfigPersistenceDir() (string, error) {
...
+	kubeConfigDir, err := DefaultKubeConfigPersistenceDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(kubeConfigDir, "config"), nil
```
joaquimrocha, line 233: Is this what we are currently expecting? i.e. where is this default being used?

backend/pkg/config/config.go
```diff
@@ -233,6 +233,43 @@ func DefaultKubeConfigPersistenceFile() (string, error) {
 	return filepath.Join(kubeConfigDir, "config"), nil
 }
 
+// collectMultiConfigPaths looks at the default dynamic directory
```
joaquimrocha: looks at Headlamp's kubeconfigs directory.

backend/pkg/config/config.go
```diff
@@ -233,6 +233,43 @@ func DefaultKubeConfigPersistenceFile() (string, error) {
...
+// (e.g. ~/.config/Headlamp/kubeconfigs) and returns any files found there.
```
joaquimrocha: While it's useful to have the paths in the docs, this creates a possibility of the docs becoming wrong if we change the default kubeconfigs dir.

backend/pkg/config/config.go
```diff
@@ -233,6 +233,43 @@ func DefaultKubeConfigPersistenceFile() (string, error) {
...
+// This is called from the 'else' block in deleteCluster().
+//
+//nolint:prealloc
+func CollectMultiConfigPaths() ([]string, error) {
```
joaquimrocha: How about HeadlampKubeConfigPaths ?

backend/pkg/config/config.go
```diff
@@ -270,6 +271,66 @@ func CollectMultiConfigPaths() ([]string, error) {
 	return configPaths, nil
 }
 
+// RemoveContextFromConfigs does the real iteration over the configPaths.
+func RemoveContextFromConfigs(contextName string, configPaths []string) error {
+	var removed bool
```
joaquimrocha: maybe `removed := false` is more readable.

backend/pkg/config/config.go
```diff
@@ -270,6 +271,66 @@ func CollectMultiConfigPaths() ([]string, error) {
 	return configPaths, nil
 }
 
+// RemoveContextFromConfigs does the real iteration over the configPaths.
```
joaquimrocha: Is this description accurate/useful?

backend/pkg/config/config.go
```diff
@@ -270,6 +271,66 @@ func CollectMultiConfigPaths() ([]string, error) {
...
+func RemoveContextFromConfigs(contextName string, configPaths []string) error {
+	var removed bool
+
+	for _, filePath := range configPaths {
```
joaquimrocha: BTW, I thought we now had the actual kubeconfig from which we read the cluster, so we shouldn't do this attempt to match?

backend/pkg/config/config.go
```diff
@@ -331,6 +331,41 @@ func RemoveContextFromConfigs(contextName string, configPaths []string) error {
...
+	configPaths ...string,
+) error {
+	// Check if contextName is empty
+	if contextName == "" {
+		return fmt.Errorf("context name cannot be empty")
+	}
+
+	// If no specific paths passed, fallback to the default.
+	if len(configPaths) == 0 {
+		discoveredPath, err := DefaultKubeConfigPersistenceFile()
```
joaquimrocha: In which case is this happening?
Do we allow this?

backend/cmd/headlamp.go
```diff
@@ -1917,17 +1919,76 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
+	r *http.Request,
+	ctx context.Context,
+	span trace.Span,
+	name string,
+) {
+	configPath := r.URL.Query().Get("configPath")
+	originalName := r.URL.Query().Get("originalName")
+	source := r.URL.Query().Get("source")
+
+	const kubeConfigSource = "kubeconfig"
```
joaquimrocha: Do you have to assign this to a const here?

backend/cmd/headlamp.go
```diff
@@ -1917,17 +1919,76 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
+
+	var err error
+	if source == kubeConfigSource {
+		err = kubeconfig.RemoveContextFromFile(originalName, configPath)
+	} else {
+		err = kubeconfig.RemoveContextFromFile(name, configPath)
+	}
```
joaquimrocha: We could instead assign a variable with the actual name we are using for deleting and then have only one function call with it (more readable).

frontend/src/stateless/index.ts
```diff
@@ -484,15 +488,13 @@ export async function deleteClusterKubeconfig(clusterName: string): Promise<stri
...
-                context.context.extensions?.find(extension => extension.name === 'headlamp_info')
-                  ?.extension.customName === clusterName
+            const { matchingKubeconfig, matchingContext } = findMatchingContexts(
+              clusterName,
+              parsedKubeconfig,
+              clusterID
             );
 
-            if (matchingKubeconfig) {
+            if (matchingKubeconfig || matchingContext) {
```
joaquimrocha: We should make sure we find an existing kubeconfig. Why do we resort to the matching context?

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -51,9 +52,24 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
...
+    const originalName = clusterSource === 'kubeconfig' ? cluster.meta_data?.originalName : '';
+    const clusterName = cluster.name;
+
+    deleteCluster(
+      clusterName,
+      clusterSource,
+      deleteFromKubeconfig,
+      clusterID,
+      kubeconfigOrigin,
+      originalName
```
joaquimrocha: Do we really need all these?

joaquimrocha (changes requested):
I know I asked questions or gave suggestions about previously existing code, but they are good to think about in the context of this PR. Please check them out.
About the i18n, please do not translate new strings unless you know the language (because it'll be difficult to spot by speakers later).

backend/cmd/headlamp.go
vyncent-t: It keeps running into the linter issue when not putting a const

image

backend/cmd/headlamp.go

backend/pkg/config/config.go
vyncent-t: done!

backend/pkg/config/config.go
vyncent-t: done!

backend/pkg/config/config.go
vyncent-t: done!

backend/pkg/config/config.go
vyncent-t: noted

backend/pkg/config/config.go
vyncent-t: Good point, after experimenting and removing a few parts I noticed I can get the deleting to work without all the extra support functions from my old work before the metadata changes.

Now with the direct kubeconfig path, the function can delete the cluster right where it needs to. This old work was part of a solution to answer "what happens if the cluster name we are looking for is only in one of many kubeconfigs", an issue that is no longer a problem with the new metadata changes.

will remove and rework the old support code

backend/pkg/config/config.go
vyncent-t: this has been removed for rework

backend/pkg/config/config.go
vyncent-t: this has been removed for rework

backend/pkg/config/config.go
vyncent-t: this has been removed for rework

backend/pkg/config/config.go
vyncent-t: this has been removed for rework

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: I use each of these pieces to accurately find the cluster then navigate a delete choice tree

clusterName
- used in the name search for `findKubeconfigByClusterName`
- also used to determine if the cluster name to be removed when using the `deleteCluster` func on backend

clusterID
- used in the name search for `findKubeconfigByClusterName` for accuracy

removeKubeConfig
- needed for choosing if we remove cluster from kubeconfig on users machine via `handleDeleteCluster` func on backend

source
- needed to ensure only non-dynamic clusters are removed cluster from kubeconfig on users machine via `handleRemoveKubeConfig` func on backend

kubeconfigOrigin
- used for the accurate file path of the cluster which is used for the file path param when deleting it from the users machine via `handleRemoveKubeConfig` 

originalName
- non dynamic clusters must be deleted by their original name, if the cluster has been renamed passing in the `clusterName` param from the front end will try to locate the custom name in the kubeconfig (can cause issues like deleting the wrong cluster or not finding it when deleting)

frontend/src/stateless/index.ts
vyncent-t: fixed!

backend/cmd/headlamp.go
joaquimrocha: I see. But cannot be related to the function scope, since you only have one string occurrence in the function. Probably the solution it to check for other places where we are comparing "kubeconfig" in the context of a _source_ like this one, and thus we should assign a file-scoped const for that effect.

backend/cmd/headlamp.go
```diff
@@ -274,43 +274,14 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
...
-	}
-
-	// if any error occurred, fallback to the current directory.
-	ex, err := os.Executable()
-	if err == nil {
-		return filepath.Dir(ex), nil
-	}
-
-	return "", fmt.Errorf("failed to get default kubeconfig persistence directory: %v", err)
+func makeHeadlampKubeConfigsDir() (string, error) {
```
joaquimrocha: Do you use this in more than one place? If you don't, maybe just call the MakeHeadlampKubeConfigsDir directly instead.

backend/cmd/headlamp.go
```diff
@@ -274,43 +274,14 @@ func serveWithNoCacheHeader(fs http.Handler) http.HandlerFunc {
...
 }
 
-func defaultKubeConfigPersistenceFile() (string, error) {
-	kubeConfigDir, err := defaultKubeConfigPersistenceDir()
-	if err != nil {
-		return "", err
-	}
-
-	return filepath.Join(kubeConfigDir, "config"), nil
+func defaultHeadlampKubeConfigFile() (string, error) {
```
joaquimrocha, line 280: Similar comment as above.

backend/pkg/config/config.go
joaquimrocha: I am fine keeping this as a defensive measure.

backend/cmd/headlamp.go
```diff
@@ -1904,6 +1904,8 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	c.handleDeleteCluster(w, r, ctx, span, name)
+
```
joaquimrocha: Is the code after this line in the function needed now?

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -92,22 +92,56 @@ export async function setCluster(clusterReq: ClusterRequest) {
...
+  removeKubeConfig?: boolean,
+  /** The ID for a cluster, composed of the kubeconfig path and cluster name */
+  clusterID?: string,
+  // /** The origin of the cluster, e.g., kubeconfig path */
+  kubeconfigOrigin?: string,
+  // /** The original name of the cluster, used for kubeconfig clusters */
+  originalName?: string
 ): Promise<{ clusters: ConfigState['clusters'] }> {
+  let deleteURL;
+  const removeFromKubeConfig = removeKubeConfig ? 'true' : 'false';
```
joaquimrocha: You can just use the `${!!removeKubeConfig}` inside the string literal, as it will automatically use true or false accordingly.

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -92,22 +92,56 @@ export async function setCluster(clusterReq: ClusterRequest) {
...
+
+  // If the source is kubeconfig and the originalName is provided, and removeKubeConfig is true,
+  // we need to construct the URL differently to ensure the correct parameters are passed.
+  if (source === 'kubeconfig' && originalName && removeKubeConfig) {
+    // for non dynamic clusters, we need to use the original name as a query parameter to find the actual context in the kubeconfig
+    // and remove it from the kubeconfig file.
+    deleteURL = `/cluster/${cluster}?removeKubeConfig=${removeFromKubeConfig}&source=${source}&configPath=${kubeconfigOrigin}&originalName=${originalName}`;
```
joaquimrocha: If we are sending the configPath, doesn't the source="kubeconfig" get redundant?
i.e. can we decide in the backend if we should do the same logic by just checking if configPath is sent in the request?
(honest question)

frontend/src/components/App/Home/ClusterContextMenu.tsx
joaquimrocha: I asked a question about source in the previous commit which is related.

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -67,6 +83,50 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
       });
   }
 
+  function removeClusterDescription(cluster: Cluster) {
+    function handleChoiceToggle() {
+      setDeleteFromKubeconfig(!deleteFromKubeconfig);
+    }
+
+    const description = t(
+      'translation|This action will remove cluster "{{ clusterName }}" from Headlamp.',
```
joaquimrocha: Do not use Headlamp in strings if it's avoidable, since we have a way for developers to rename the program.

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -67,6 +83,50 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
...
+      {
+        clusterName: cluster.name,
+      }
+    );
+
+    const removeFromKubeconfigDes =
+      cluster.meta_data?.source === 'kubeconfig' ? (
+        <Typography>
+          {t('translation|Remove from source {{ source }}', {
+            clusterName: cluster.name,
```
joaquimrocha: This is not being used in the translation string.

joaquimrocha (changes requested):
Much better now but still room to improve.

backend/cmd/headlamp.go
vyncent-t: okay

it is only used once so will just use `cfg.MakeHeadlampKubeConfigsdir()`

backend/cmd/headlamp.go
vyncent-t: although this one is used by `getKubeConfigPath` func and `createHeadlampHandler` (keeping as is)

backend/cmd/headlamp.go
vyncent-t: removed

frontend/src/lib/k8s/api/v1/clusterApi.ts
vyncent-t: fixed

frontend/src/lib/k8s/api/v1/clusterApi.ts
vyncent-t: I see what you're saying here, but having something as destructive as this delete function I went for a more double requirement approach. 

I have been experimenting with this logic and finding ways to extend the functionality for permanent deleting clusters for plugins also (I will open a PR for this later after this one is merged, if you think its a worth while feature to have. I have the implementation ready, I just want to land it for non dynamic first)

We are technically getting the config path (under `kubeconfigOrigin`) for every delete request when setting the check box to true, so having just a check for a location to delete from I feel isn't narrow enough.

On the other hand, this does open up possibilities of deleting beyond the non dynamic range here (as I mentioned for removing clusters from default headlamp dir) so I think having another check to make sure the kubeconfig source is nondynamic would be needed for this scope, as it is easily turned to delete in other places.

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: fixed

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: fixed

backend/cmd/headlamp.go
vyncent-t: looks like there were some places doing the same thing, added it to the top scope

vyncent-t:
added a disable option for the confirm dialog to work with checkboxes

also added confirm step to deleting clusters

backend/cmd/headlamp.go
```diff
@@ -1908,30 +1908,54 @@ func (c *HeadlampConfig) deleteCluster(w http.ResponseWriter, r *http.Request) {
...
+	r *http.Request,
+	ctx context.Context,
+	span trace.Span,
+	name string,
+) {
+	configPath := r.URL.Query().Get("configPath")
+	originalName := r.URL.Query().Get("originalName")
+	source := r.URL.Query().Get("source")
+
+	const kubeConfigSource = "kubeconfig"
```
joaquimrocha: If this is the commit introducing the lint issue, you should add the fix into this one and not on a different commit.

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -67,6 +83,51 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
...
 
+  function removeClusterDescription(cluster: Cluster) {
+    function handleChoiceToggle() {
+      setDeleteFromKubeconfig(!deleteFromKubeconfig);
+    }
+
+    const description =
+      cluster.meta_data?.source === 'kubeconfig'
+        ? t(
+            'translation|This action will delete cluster {{ clusterName }} from source {{ source }}',
```
joaquimrocha: I think there's still some confusion in what this message is about:
If users can only either delete or not delete the cluster (including removing it from the kubeconfig), then we don't need a check box to confirm they want to remove the cluster from the kubeconfig.

Adding the checkbox would be only useful in a scenario where we can give the option for the cluster to be removed from Headlamp's listing but not from the kubeconfig. Since this is not possible as the clusters are read from the kubeconfig, we can just tell the user that the action will remove the cluster/context from the kubeconfig file (no need for a checkbox).

And the user choices/buttons should be "Cancel" "Delete".

Hope this makes sense.

joaquimrocha (changes requested):

backend/cmd/headlamp.go
vyncent-t: fixed!

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: removed the checkbox logic

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -51,9 +52,17 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
...
   const menuItems = useTypedSelector(state => state.clusterProvider.menuItems);
+  const [deleteFromKubeconfig, setDeleteFromKubeconfig] = React.useState(false);
+
+  const kubeconfigOrigin = cluster.meta_data?.origin?.kubeconfig;
 
   function removeCluster(cluster: Cluster) {
-    deleteCluster(cluster.name || '')
+    const clusterID = cluster.meta_data?.clusterID;
+    const clusterSource = cluster.meta_data?.source;
+    const originalName = clusterSource === 'kubeconfig' ? cluster.meta_data?.originalName : '';
```
joaquimrocha: Could it be?
```ts
const originalName = cluster.meta_data?.originalName ?? '';
```

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -67,6 +76,53 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
...
   }
 
+  useEffect(() => {
+    // If the cluster is from kubeconfig, we want to set the delete to kubeconfig.
+    if (cluster?.meta_data?.source === 'kubeconfig') {
+      setDeleteFromKubeconfig(true);
+    } else {
+      setDeleteFromKubeconfig(false);
+    }
+  }, [deleteFromKubeconfig]);
```
joaquimrocha: Why do you need this state?
You could just have `const deleteFromKubeconfig = cluster?.meta_data?.source === 'kubeconfig'`.

joaquimrocha (changes requested):
Thanks for the update. I think now it's very close.

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: Had react render issues when trying to use something close to this but inside `removeClusterDescription`

Moved it under kubeconfigOrigin and seems to work the same now

frontend/src/components/App/Home/ClusterContextMenu.tsx
vyncent-t: added this, works the same

vyncent-t:
The splitting rename issue for the dynamic clusters seems to come from the current main and not these changes. 
https://github.com/kubernetes-sigs/headlamp/issues/3502


these changes work as intended w/ @sniok

frontend/src/i18n/locales/de/translation.json
```diff
@@ -39,11 +39,14 @@
...
   "Providers": "",
+  "This action will delete cluster \"{{ clusterName }}\" from \"{{ source }}\"": "",
+  "This action will remove cluster \"{{ clusterName }}\".": "",
+  "This action cannot be undone! Do you want to proceed?": "",
+  "Remove this cluster?": "",
   "View": "Ansicht",
   "Settings": "Einstellungen",
   "Delete": "Löschen",
   "Delete Cluster": "",
-  "Are you sure you want to remove the cluster \"{{ clusterName }}\"?": "Sind Sie sicher, dass Sie den Cluster \"{{ clusterName }}\" entfernen möchten?",
```
illume, line 46: This translation was removed for no reason.

PR #3431, opened 2025-06-03
frontend: AuthChooser: Update buttons for better visibility

sniok:
Fixes color for "log in" and "use cluster" buttons

Reported bug:
image

Now:
image


skoeva (approved):
thanks for this 👍

PR #3432, opened 2025-06-03
frontend: Improve Delete dialog

sniok:
- Better labels Delete/Cancel instead of Yes/No
 - Contained buttons with clear primary and secondary versions
 - Checkbox instead of a switch for better visibility

Before:

image

After:

image

### DIFF
--- a/frontend/src/components/common/ConfirmDialog.tsx
+++ b/frontend/src/components/common/ConfirmDialog.tsx
@@ -77,11 +77,21 @@ export function ConfirmDialog(props: ConfirmDialogProps) {
         </DialogContent>
         <DialogActions>
           {!hideCancelButton && (
-            <Button onClick={handleClose} aria-label="cancel-button" color="primary">
+            <Button
+              onClick={handleClose}
+              aria-label="cancel-button"
+              color="secondary"
+              variant="contained"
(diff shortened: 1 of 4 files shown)


skoeva (approved):
LGTM

illume (dismissed):
Is delete the default now?

Maybe I'm not understanding the screenshots, is force the default? Is the Delete option the default button focused?

sniok:
> Is delete the default now?
> 
> Maybe I'm not understanding the screenshots, is force the default? Is the Delete option the default button focused?

@illume  None of the buttons are focused by default

joaquimrocha:
I think we should have the cancel as default in destructive actions. It should probably be configurable as well. Do you want to update it in this PR or we merge it and leave that for another PR?

sniok:
> I think we should have the cancel as default in destructive actions. It should probably be configurable as well. Do you want to update it in this PR or we merge it and leave that for another PR?

Unless there are any comments about code in this PR in current form, I'd merge it, since it was already reviewed

joaquimrocha:
I think we need to review all confirm dialogs and make the least destructive action the default. But let's merge this in the meanwhile.

/lgtm
/approve

PR #3448, opened 2025-06-06
frontend: PodDetails: Reset Logs state on item change

### LINKED ISSUE #3446: Logs not working on drawer overlay
## Describe the bug
The log viewer for pods is broken when using drawer mode. With drawer mode, the log viewer component does not refresh the logs and seems to disconnect. The pod logs are still working as normal when the drawer mode is not enabled and we are at the pod details page when we open the log viewer.

## To Reproduce

Issue exists in both desktop and browser

Steps to reproduce the bug:
1. Enable drawer mode in headlamp settings
2. Navigate to the Pods tab under Workloads
3. Click the first pod available, then within the drawer click the "Show Logs" button
4. **Scroll to the bottom of the logs, make a note of the last few printed lines**
5. Click another pod that is available in the pod list
6. Click on the "Show Logs" button for the new pod
7. **Notice that the logs from the previous pod are displayed**
8. Click on a third pod in the pod list
9. **Notice that the logs are blank and the disconnect button renders**

Image

Note: Visiting a new cluster's pod list for the or closing the drawer does seem to sometimes fix the issue and logs are rendered as expected. However doing so just resets the chain of `working log` -> `duplicate log` -> `empty logs`

This could likely be an issue with the rendering of the drawer not updating correctly.

 - Headlamp Version: 0.31.1

I am currently working on a slight rework for the logs viewer component which may fix this. If further reworks are required / we move to an alternative terminal, I recommend disabling the log viewer when in drawer mode for the time being.

## Additional Context
This may be an issue for users if drawer mode is on by default.

sniok:
fixes #3446 

PodLogViewer component seems to be struggling with rerendering with a different Pod (item)
To fix this we can completely reset the state of the component when Pod changes

Here's React docs about this if you want to understand how this works https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -640,7 +640,7 @@ export default function PodDetails(props: PodDetailsProps) {
             id: 'headlamp.pod-logs',
             section: (
               <PodLogViewer
-                key="logs"
+                key={'logs-' + item.metadata.uid}
                 open={showLogs}
                 item={item}
                 onClose={() => {


vyncent-t (approved):
It works! thanks for the fix! LGTM

cumcum8197:
b743aa3c01e8aa91caf6a4fcfa8b05a38c7d7afd

PR #3452, opened 2025-06-06
docs: update ingress example

dorser:
Updated the in-cluster installation docs to use the correct GitHub repo URL (`kubernetes-sigs/headlamp` instead of `kinvolk/headlamp`). Also updated the ingress manifest to use the `ingressClassName` field instead of the deprecated `kubernetes.io/ingress.class` annotation, for better compatibility with newer Kubernetes versions.

### DIFF
--- a/docs/installation/in-cluster/index.md
+++ b/docs/installation/in-cluster/index.md
@@ -68,7 +68,7 @@ Assuming your URL is `headlamp.mydeployment.io`, getting the sample ingress
 file and changing the URL can quickly be done by:
 
 ```bash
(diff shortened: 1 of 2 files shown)


skoeva (commented):
Thanks for the changes, I suppose they could be separated into two different commits? Since they aren't directly related

dorser:
@skoeva sure! I splitted it into two commits. Thanks for the review!

joaquimrocha:
Thanks @dorser !

/lgtm
/approve

skoeva (approved):
perfect, thanks!

PR #3454, opened 2025-06-06
fix(codeSign) esrp shell command built from environment values

odaysec:
https://github.com/kubernetes-sigs/headlamp/blob/ff50a9e85855ded3eb476db27487206abfa1936f/app/scripts/esrp.js#L9-L9
https://github.com/kubernetes-sigs/headlamp/blob/ff50a9e85855ded3eb476db27487206abfa1936f/app/scripts/esrp.js#L221-L221

Fix the issue the dynamically constructed shell command should be replaced with a safer approach that avoids shell interpretation. Specifically, the `execSync` call should be replaced with `execFileSync`, which allows passing arguments as an array. This ensures that special characters in the input are treated as literal values and do not alter the command's behavior.

**implement the fix:**
1. Replace the `execSync` call on line 221 with `execFileSync`.
2. Pass the command (`Sign`) and its arguments (`-l Verbose`, `-a`, `authJson`, `-p`, `policyJson`, `-i`, `signInputJson`) as separate elements in an array.
3. Ensure that all paths (`authJson`, `policyJson`, `signInputJson`) are validated or sanitized before use.

### DIFF
--- a/app/scripts/esrp.js
+++ b/app/scripts/esrp.js
@@ -6,7 +6,7 @@
  **/
 
 const crypto = require('crypto');
-const { execSync } = require('child_process');
+const { execFileSync } = require('child_process');
 const path = require('path');
 const os = require('os');
 const fs = require('fs');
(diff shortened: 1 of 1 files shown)


skoeva (changes requested):
Hi, thanks for looking into this! Be sure to fix any frontend linting errors with `make frontend-fixlint`.

It would be also nice if you could tweak the commit & PR titles to match our [commit guidelines](https://headlamp.dev/docs/latest/contributing/#2-follow-commit-guidelines).

joaquimrocha:
Ah, yeah, we need the linter complaint to be fixed. You should run `make frontend-fixlint`, @odaysec .

PR #3460, opened 2025-06-10
app: frontend: Improve startup performance

### LINKED ISSUE #2551: Startup is very slow
## Describe the bug
The startup of Headlamp is very slow, especially on Windows. We should investigate why it is so slow and try speeding it up.
At the same time, we could add some techniques to make the startup perceived as faster (like a splash screen if necessary).

## To Reproduce
Steps to reproduce the bug:
1. Open Headlamp from cold start on Windows -> measure how long it takes till the logo shows up in the top bar

 - Installation type: all desktop version but especially on Windows
 - Headlamp Version: 0.26.0 (but also many more before this one)

## Additional Context
Let's use this section to add any sub-items related to this.

sniok:
Improves startup performance by up to 3x. Fixes #2551

Currently the biggest impact to the load performance is when it checks for existing headlamp-server process. More details [here](https://github.com/kubernetes-sigs/headlamp/issues/2551#issuecomment-2939557291).
This PR improves this by only performing this check when server start fails with an appropriate exit code. 

Also to improve startup, this PR adds loader to the index.html which is instantly visible, and adds background color to avoid flashing screen with dark themes.

To better handle the case when there's an existing headlamp-server and frontend loads early, I've used react-query for loading the cluster config for faster retries when cluster config fails to load. (Otherwise it would refetch every 10s)

Startup time comparison (measured from launching app to visible cluster list)
Before: ~4.5s
After: ~1.6s (loader visible within 0.5s)

https://github.com/user-attachments/assets/8f867382-04a3-4778-bbd8-eb72df9273e2

### DIFF
--- a/frontend/src/components/App/Layout.tsx
+++ b/frontend/src/components/App/Layout.tsx
@@ -20,6 +20,8 @@ import Container from '@mui/material/Container';
 import CssBaseline from '@mui/material/CssBaseline';
 import Link from '@mui/material/Link';
 import { styled } from '@mui/material/styles';
+import { Dispatch, UnknownAction } from '@reduxjs/toolkit';
+import { useQuery } from '@tanstack/react-query';
 import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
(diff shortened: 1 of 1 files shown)


illume:
Very cool. 

Can you please confirm what testing you've done in the PR description? I'm interested to know which platforms, and which ways of running/quitting the app have been tested.

The bot added 4 reviewers, so I removed a couple and left @skoeva and @ashu8912 on as reviewers.

frontend/src/components/App/Layout.tsx
```diff
@@ -150,6 +150,11 @@ export default function Layout({}: LayoutProps) {
     };
   }, []);
 
+  // Remove any styles from the body
```
joaquimrocha: Might be worth explaining why.

joaquimrocha (changes requested):
Haven't tested it but it looks nice. Left a comment and there are some more from Copilot.

frontend/src/components/App/Layout.tsx
```diff
@@ -121,6 +123,57 @@ declare global {
...
 
+/**
+ * Fetches the cluster config from the backend and updates the redux store
+ * if the present stored config is different from the fetched one.
+ */
+const fetchConfig = (dispatch: Dispatch<UnknownAction>) => {
+  const clusters = store.getState().config.clusters;
+  const statelessClusters = store.getState().config.statelessClusters;
+
+  return request('/config', {}, false, false).then(config => {
```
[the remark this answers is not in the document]
sniok: not true

frontend/src/components/App/Layout.tsx
sniok: added some more context

skoeva (approved):
looks like a huge improvement, thanks for this!

joaquimrocha:
Smoke tested and looks like everything is working. The new ellipsis loader that is briefly shown helps thinking the UI is not frozen.
The code also LGTM.

/approve
/lgtm

PR #3461, opened 2025-06-10
frontend: ResourceTable: Preserve column filter state

skoeva:
This change saves the table column filters to localStorage so that they may persist through navigation and reloads.

Fixes: #2554 

### Testing
- [X] Navigate to any resource page in Headlamp (i.e. Pods)
- [X] Set a filter in the table
- [X] Restart the page or navigate away and back to the page and ensure that the filter state persists

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -32,6 +32,7 @@ import { HeadlampEventType, useEventCallback } from '../../../redux/headlampEven
 import { useTypedSelector } from '../../../redux/hooks';
 import { useSettings } from '../../App/Settings/hook';
 import { ClusterGroupErrorMessage } from '../../cluster/ClusterGroupErrorMessage';
+import { useLocalStorageState } from '../../globalSearch/useLocalStorageState';
 import { DateLabel } from '../Label';
 import Link from '../Link';
 import Table, { TableColumn } from '../Table';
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -315,6 +333,15 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
   const [columnVisibility, setColumnVisibility] = useState(() =>
     initColumnVisibilityState(columns, id)
   );
+  const [columnFilters, setColumnFilters] = useState(() => loadTableFilters(id || ''));
```
sniok: please use `useLocalStorageState` hook instead, this logic is already implemented there

sniok:
There's something wrong here, when trying to type some characters are ignored or repeated

Here I'm trying to type 123456789 into the column filter:

https://github.com/user-attachments/assets/dc4cedf9-bf7e-4aee-8637-e1203375ff40

I think if you put `columnFilters` into `initialState` instead of `state` it should work

sniok (changes requested):
thanks for updating the PR, left one comment above

skoeva:
@sniok The issue with putting it in `initialState` is that the UI won't update in real time, only after refreshing the page; is that better?

sniok:
> The issue with putting it in `initialState` is that the UI won't update in real time, only after refreshing the page; is that better?

yeah you're right, disregard my suggestion

skoeva:
Setting to draft, will close if there is no more activity on the issue

PR #3479, opened 2025-06-13
frontend: scripts: Add lint-fix and test-coverage scripts

DhairyaMajmudar:
# Description

1. using lint-fix npm command developers can fix the lining issues
2. test-coverage command gives a test report, it would be helpful to see areas where more tests are required.


sniok:
correct me if I'm wrong but the same can be achieved by using existing commands
`npm run test -- --coverage`
`npm run lint -- --fix`

DhairyaMajmudar:
> correct me if I'm wrong but the same can be achieved by using existing commands
> `npm run test -- --coverage`
> `npm run lint -- --fix`

You're right that we can achieve similar results using above commands. But having dedicated scripts isn't a bit cleaner and more convenient approach?

sniok:
There are `make frontend-fixlint` and `make frontend-test` commands available already
I don't think these new scripts are necessary to be honest

DhairyaMajmudar:
Oh yeah correct and make commands have those flags, Thanks! let me close this PR.

Issue #3482, opened 2025-06-16
backend: Add auth package from headlamp.go OIDC functions


illume:
There's a lot of tangled up auth code inside the headlamp.go package. Some of it without any tests or docs.

## basic plan

Look at OIDC specifications, add tests, refactor code... one function at a time.

### Step 1: Move standalone helper functions (no external dependencies)
- [x] `decodePayload(payload string) (map[string]interface{}, error)` https://github.com/kubernetes-sigs/headlamp/pull/3483
- [x] `getExpiryTime(payload map[string]interface{}) (time.Time, error)` https://github.com/kubernetes-sigs/headlamp/pull/3487
- [x] `isTokenAboutToExpire(token string) bool` https://github.com/kubernetes-sigs/headlamp/pull/3895
- [x] `parseClusterAndToken(r *http.Request) (string, string)` https://github.com/kubernetes-sigs/headlamp/pull/3494

These functions are simple, self-contained, and have no external dependencies.

### Step 2: Move token refresh logic (minimal external dependencies)
- [x] `getNewToken(clientID, clientSecret string, cache cache.Cache[interface{}], tokenType string, token string, tokenURL string) (*oauth2.Token, error)` https://github.com/kubernetes-sigs/headlamp/pull/3917
- [x] `cacheRefreshedToken(token *oauth2.Token, tokenType string, oldToken string, oldRefreshToken string, cache cache.Cache[interface{}]) error` https://github.com/kubernetes-sigs/headlamp/pull/3891
- [x] `refreshAndCacheNewToken(clientID, clientSecret string, cache cache.Cache[interface{}], tokenType, token, issuerURL string) (*oauth2.Token, error)` https://github.com/kubernetes-sigs/headlamp/pull/3998

These functions depend on OAuth2 and caching, but are still relatively isolated.

### Step 3: Move middleware and handlers (more dependencies)
- [ ] `(c *HeadlampConfig) refreshAndSetToken(...)` https://github.com/kubernetes-sigs/headlamp/pull/4230
- [ ] `(c *HeadlampConfig) incrementRequestCounter(...)` #5877 
- [ ] `(c *HeadlampConfig) shouldSkipOIDCRefresh(...)`
- [ ] `(c *HeadlampConfig) shouldBypassOIDCRefresh(...)`
- [ ] `(c *HeadlampConfig) handleGetContextError(...)`
- [ ] `(c *HeadlampConfig) handleOIDCAuthConfigError(...)`
[...]

PR #3494, opened 2025-06-17
backend: auth: Extract ParseClusterAndToken from headlamp.go

skoeva:
This change extracts the ParseClusterAndToken from headlamp.go into the new auth package.

Part of:
- #3482

### Updates
- Handles case-insensitive `Bearer` and trims spaces
- Reuses the compiled regex
- Validates cluster contexts and tokens
- Rejects multiple tokens and handles paths with no cluster

### Testing
```shell
cd backend
go test -v -p 1 ./... -run TestParseClusterAndToken
```

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -32,7 +32,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"syscall"
@@ -824,30 +823,6 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 	return r
 }
 
-func parseClusterAndToken(r *http.Request) (string, string) {
-	cluster := ""
-	re := regexp.MustCompile(`^/clusters/([^/]+)/.*`)
-	urlString := r.URL.RequestURI()
-
-	matches := re.FindStringSubmatch(urlString)
-	if len(matches) > 1 {
-		cluster = matches[1]
-	}
-
-	// Try Authorization header first (for backward compatibility)
-	token := r.Header.Get("Authorization")
-	token = strings.TrimPrefix(token, "Bearer ")
-
-	// If no auth header, try cookie
-	if token == "" && cluster != "" {
-		if cookieToken, err := auth.GetTokenFromCookie(r, cluster); err == nil {
-			token = cookieToken
-		}
-	}
-
-	return cluster, token
-}
-
 func getExpiryTime(payload map[string]interface{}) (time.Time, error) {
 	exp, ok := payload["exp"].(float64)
 	if !ok {
@@ -1098,7 +1073,7 @@ func (c *HeadlampConfig) OIDCTokenRefreshMiddleware(next http.Handler) http.Hand
 		}
 
 		// parse cluster and token
-		cluster, token := parseClusterAndToken(r)
+		cluster, token := auth.ParseClusterAndToken(r)
 		if c.shouldBypassOIDCRefresh(cluster, token, w, r, span, ctx, start, next) {
 			return
 		}
--- a/backend/pkg/auth/auth.go
+++ b/backend/pkg/auth/auth.go
@@ -19,6 +19,9 @@ package auth
 import (
 	"encoding/base64"
 	"encoding/json"
+	"net/http"
+	"regexp"
+	"strings"
 )
 
 // DecodeBase64JSON decodes a base64 URL-encoded JSON string into a map.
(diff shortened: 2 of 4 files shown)


backend/pkg/auth/auth_test.go
```diff
@@ -89,3 +93,65 @@ func TestDecodeBase64JSON(t *testing.T) {
...
+			wantToken:   "spaced-token",
+		},
+		{
+			name:        "not a clusters path",
+			url:         "/no-clusters-prefix/api",
+			authHeader:  "Bearer test-token",
+			wantCluster: "",
+			wantToken:   "test-token",
+		},
+	}
```
illume: How about a test case for "/clusters/". What should the user of this function (OIDCTokenRefreshMiddleware) do in this case?

Is "/clusters/" a valid URL we should handle? I think wantCluster should be "", because there is no cluster and therefore no OIDC needed.

backend/pkg/auth/auth_test.go
```diff
@@ -89,3 +93,65 @@ func TestDecodeBase64JSON(t *testing.T) {
...
+			wantCluster: "extra",
+			wantToken:   "spaced-token",
+		},
+		{
+			name:        "not a clusters path",
+			url:         "/no-clusters-prefix/api",
+			authHeader:  "Bearer test-token",
+			wantCluster: "",
+			wantToken:   "test-token",
+		},
```
illume: I know that multiple Bearer tokens in one request are not supported by OIDC or OAuth 2.0. However headers are allowed multiple values. 

So, I wonder if we should be defensive here and make it return ""? I haven't seen any evidence that this is being used in the wild, and it's against the spec. So my guess is this is always an error/misconfiguration.

The case "Authorization: Bearer xxx, Bearer yyy" I think wantToken should be "". wdyt?

backend/go.mod
```diff
@@ -45,6 +46,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.35.0
 	golang.org/x/term v0.30.0
 	gopkg.in/yaml.v2 v2.4.0
+	gotest.tools/v3 v3.4.0
```
illume: Can you please remove this dependency? I think your test can be written similar to the TestDecodeBase64JSON test in auth_test.go which doesn't require any extra third party dependencies (only uses stdlib things for testing).

backend/go.mod
```diff
@@ -32,6 +32,7 @@ require (
 require (
 	github.com/coreos/go-oidc/v3 v3.11.0
 	github.com/gorilla/websocket v1.5.3
+	github.com/prometheus/client_golang v1.19.1
```
illume: Can this PR be done without this? What is this dependency needed for?

illume (commented):
I like how you moved the regex compilation out of the function.

Looking at how shouldBypassOIDCRefresh uses the return values of this, I noticed above there was a check that the URL is a /clusters/ URL or not, but it's good to check in this function and a regex is being used anyway.

I found the spec for the "Authorization: Bearer" token is here: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1 (Probably worth adding this link into the function documentation).

I wonder if we should add validation of the token characters in here? (see valid ones in the spec above). I feel like they should be validated... but is here the right spot?


Technically this is a cluster context we are taking in from the URL. Although there is ambiguity in the frontend UI between cluster names and context names, I think we should be specific in the backend and mention that we are returning a cluster context.

The other thing I was wondering if we should validate is the cluster context name? We do validate that in other parts of our code, so maybe here too?

backend/pkg/auth/auth_test.go
skoeva: better to be defensive I suppose, can add this in

backend/pkg/auth/auth_test.go
skoeva: agree, the middleware will skip the OIDC logic since the cluster will be empty. adding this test in as well

backend/go.mod
skoeva: keeps popping up when I run `go mod tidy` in the backend dir, maybe it's just a local bug? removed

backend/pkg/auth/auth.go
```diff
@@ -35,3 +38,40 @@ func DecodeBase64JSON(base64JSON string) (map[string]interface{}, error) {
 
 	return payloadMap, nil
 }
+
+// clusterPathRegex matches /clusters/<cluster>/...
+var clusterPathRegex = regexp.MustCompile(`^/clusters/([^/]+)/.*`)
+
+// kubeContextNameRegex matches valid context names following the DNS-1123 standard.
```
illume: Hrmm. Maybe it's not DNS-1123? I can't find any specification for it.

Here's some real world-ish examples:

Provider | Example Context Name
-- | --
AKS | myResourceGroup_myAKSCluster_eastus
EKS | arn:aws:eks:us-west-2:123456789012:cluster/my-eks-cluster
GKE | gke_my-gcp-project_us-central1-a_my-gke-cluster

Maybe it's not a good idea to validate it after all? If we can't find a specification for it?


See this: 
- https://github.com/kubernetes-sigs/headlamp/pull/2270

This discussion includes some real world context names: https://stackoverflow.com/questions/74516648/eks-aks-cluster-name-convention


Also here (https://developers.eksworkshop.com/docs/python/eks/manage-contexts/) includes these two examples: 

- `arn:aws:eks:us-east-1:012345678901:cluster/fargate-quickstart`
- `admin@fargate-quickstart.us-east-1.eksctl.io`

I wonder if headlamp has a bug using cluster context names with things with a / in them?

skoeva: It's the DNS standard in RFC-1123, which I suppose should be met by default given the linked PR. there shouldn't be an issue with slashes (/) because they're handled in that PR too

illume:
Note: we should check all the different formats manually. https://github.com/kubernetes-sigs/headlamp/pull/3494#discussion_r2159140586

Also, makeDNSFriendly doesn't handle @ characters, and it should? Plus some other things ($ signs etc). These work in the URL, so we should let these characters validate.

In kubeconfig.go it loads all of kubeconfigs and replaces the context names with dns friendly ones. So we assume the cluster from the URL is dns friendly. (That is, the / and ' ' are converted to something that works in URLs).

backend/pkg/auth/auth.go
```diff
@@ -35,3 +38,40 @@ func DecodeBase64JSON(base64JSON string) (map[string]interface{}, error) {
 
 	return payloadMap, nil
 }
+
+// clusterPathRegex matches /clusters/<cluster>/...
+var clusterPathRegex = regexp.MustCompile(`^/clusters/([^/]+)/.*`)
+
+// kubeContextNameRegex matches valid context names following the RFC 1123 standard.
+var kubeContextNameRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
```
illume: This should validate less strongly. This should allow @$: in side it.

skoeva: removed the regex (it should already be validated)

illume (approved):
🎉 thanks!

backend/cmd/headlamp.go
```diff
@@ -824,30 +823,6 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
-	matches := re.FindStringSubmatch(urlString)
-	if len(matches) > 1 {
-		cluster = matches[1]
-	}
-
-	// Try Authorization header first (for backward compatibility)
-	token := r.Header.Get("Authorization")
-	token = strings.TrimPrefix(token, "Bearer ")
-
-	// If no auth header, try cookie
```
sniok, line 841: the logic of fetching token from the cookie was removed in this PR. Could you check and revert this @skoeva ? I don't think this was intentional and I suspect this is causing issues with oidc token refresh

PR #3499, opened 2025-06-20
backend: caching: implemented caching for single-cluster

upsaurav12:
**Implementation**

I have implemented caching for the single-cluster. For multi-cluster i am trying to check by creating user by changing ``use-context``. 

Here how i have implemented:
1. I am capturing response during ``ProxyRequest`` and storing it in ``cache`` in string.
2. Then I am changing the response from ``gzip`` to byte and then to string.
3. If the user is making request for the first time it will first first ``selfsubjectrulesreviews``  and if it returns ``Failure`` then it is not authenticated. If doesn't then process to acquire from ``cache``. 
4. A new key will be generated for new ``kind`` and ``namespace`` and ``clusters`` and it will be same if the user is making the same requests. 
5. If key is not present in ``cache`` then it will go for actual request to ``k8's`` and then store in cache.

However more things i have to add like cache should be ``synced`` with the latest k8's value but i am finding solution whether i should go for ``countinuously make requests to k8's after 5s or fixed time`` or``use watch events from k8's``. 

**Test Results**

```
=== RUN   TestInitialize
=== RUN   TestInitialize/initializes_responseCapture_with_defaults
--- PASS: TestInitialize/initializes_responseCapture_with_defaults (0.00s)
--- PASS: TestInitialize (0.00s)
=== RUN   TestExtractNamespace
=== RUN   TestExtractNamespace/valid_url_with_namespace
--- PASS: TestExtractNamespace/valid_url_with_namespace (0.00s)
=== RUN   TestExtractNamespace/empty_url
--- PASS: TestExtractNamespace/empty_url (0.00s)
=== RUN   TestExtractNamespace/invalid_url_format
--- PASS: TestExtractNamespace/invalid_url_format (0.00s)
--- PASS: TestExtractNamespace (0.00s)
=== RUN   TestGetResponseBody
=== RUN   TestGetResponseBody/valid_response
--- PASS: TestGetResponseBody/valid_response (0.00s)
=== RUN   TestGetResponseBody/empty_Response
--- PASS: TestGetResponseBody/empty_Response (0.00s)
=== RUN   TestGetResponseBody/empty_contentType
--- PASS: TestGetResponseBody/empty_contentType (0.00s)
--- PASS: TestGetResponseBody (0.00s)
=== RUN   TestGenerateKey
=== RUN   TestGenerateKey/url_was_valid_
--- PASS: TestGenerateKey/url_was_valid_ (0.00s)
=== RUN   TestGenerateKey/empty_cluster
--- PASS: TestGenerateKey/empty_cluster (0.00s)
--- PASS: TestGenerateKey (0.00s)
```


upsaurav12:
For multiple user's i am trying to extract the ``Authentication Bearer`` token in the backend which will help to differentiate the users and help to identify the user's permissions for the resources. However i tried to ``log`` to check if it returning but it is printing only empty string with the ``clusterName``.

illume:
Nice one.

ps. See [commit guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines), and run `make backend-lint` to see lint issues localy.

illume (changes requested):
Nice start. Thanks.

Would you mind separating the code out from headlamp.go? Code can be put into a separate module/package, and then headlamp.go can import it and use it. We are in the process of moving as much code out of headlamp.go as possible into package modules. It's all a bit too much code tangled up in there.

Could you please add tests? As well, please add documentation to your functions.

I left a comment about commit message guidelines and running backend-lint.

iamundefinedguy:
@illume  I have added extracted the cache implementation from ``headlamp.go``. I am sorry  i wasn't able to write tests because my system is running slow will write the code and push them 


I am not sure that this implementation might work for multiple users however i am finding way to test it with services accounts but in development mode it is directly fetching from kubeconfig file and it loads clusters directly without asking for bearer token

illume:
How about making an own package for pkg/k8cache? I think maybe it should _not_ go inside pkg/cache, since it's topic is slightly different than what pkg/cache already does.

iamundefinedguy:
> How about making an own package for pkg/k8cache? I think maybe it should _not_ go inside pkg/cache, since it's topic is slightly different than what pkg/cache already does.

I was thinking the same but eventually had to impelment in the same package because of cache initialization and ease of importing functions.

I will. modify it.

iamundefinedguy:
I have moved the code to ``pkg/k8cache`` as package.

backend/cmd/headlamp.go
```diff
@@ -45,6 +45,7 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/cache"
 	cfg "github.com/kubernetes-sigs/headlamp/backend/pkg/config"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/helm"
+	kcache "github.com/kubernetes-sigs/headlamp/backend/pkg/k8cache"
```
illume: Is there a reason to call it kcache instead? I think this is best left as k8cache, so the reader can more easily match it with the package name.

backend/cmd/headlamp.go
```diff
@@ -1434,18 +1436,51 @@ func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen
 		r.URL.Path = mux.Vars(r)["api"]
 		r.URL.Scheme = clusterURL.Scheme
 
+		_, token := parseClusterAndToken(r)
```
illume: It would be good to be able to easily configure if caching is on or off. So that we can test it more easily, and even have it turned off by default until the feature has had a longer amount of testing.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,284 @@
...
+// Initialize responseCapture with a http.ResponseWriter and empty bytes.Buffer for the body.
+func Initialize(w http.ResponseWriter) *responseCapture {
+	return &responseCapture{
+		ResponseWriter: w,
+		Body:           &bytes.Buffer{},
+		StatusCode:     http.StatusOK,
+	}
+}
+
+// The function extracts the namespace from the parameter from the given raw URL. This is used to make
```
illume: Function comments in golang start with the function name.

backend/pkg/k8cache/key.go
```diff
@@ -0,0 +1,47 @@
...
+	"encoding/base32"
+	"encoding/json"
+)
+
+// type ResourceKind string
+
+// Key struct used to store different values to make Key unique
+// for different requests.
+type Key struct {
+	Kind      string
```
illume: Can you please add comments to these fields?

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,284 @@
...
+		dcmpBody = bodyBytes
+	}
+
+	return string(dcmpBody), nil
+}
+
+// This function generates a unique cache key based on the requested URL's path (kind of resource),
+// Namespace, and the kubernetes context. this ensures that the cached response are specific to the exact
+// resource, namespace ,and cluster being requested.
+func GenerateKey(url *url.URL, contextKey string, token string) (string, error) {
```
illume: It's not clear to me what contextKey is. Can you please expand on what this is in the documentation?

backend/pkg/k8cache/key.go
```diff
@@ -0,0 +1,47 @@
...
+	"crypto/sha256"
+	"encoding/base32"
+	"encoding/json"
+)
+
+// type ResourceKind string
+
+// Key struct used to store different values to make Key unique
+// for different requests.
+type Key struct {
```
illume: Maybe CacheKey would be a bit more descriptive? So that when I'm reading it being used it's a bit more clear.

backend/pkg/k8cache/key.go
```diff
@@ -0,0 +1,47 @@
...
+	Kind      string
+	Namespace string
+	Context   string
+	Token     string
```
illume: I wonder about using Token, and clients that refresh tokens quickly. It means that the cache length will be tied to the token expiry time.

Additionally, it will mean that the cache is only for a single client. As each client has separate tokens. Can you think of a way to create a key that means the cache can be used by multiple clients safely?

iamundefinedguy:
Hi I have just tested this cache implementation with different ``serviceaccounts`` as a users. for example service accounts have no-access to cluster resources, user having limited access and user having full access. From my end it is working fine like showing the data that the user is authorized. 

https://github.com/user-attachments/assets/66d4b03d-763a-4372-9268-07abe4f787b8

However still there are still some points that is need to be fixed/address.

1.  Initialization of Cache:- Currently i am initializing the cache as a global variable which i am not sure that it would be ideal approach for this project specifically. it is fine to initialize as it? 
2. Synchronization of Cache data :- We can have two options Watch API and Polling to K8;s at a fixed time interval. which approach should be the ideal for the project ? 
3. Use of token :- Would it be good approach to use token in the ``key`` generation because here i am already using SSAR request to authorize the user but in K8's dashboard PR they had used it.

illume (commented):
I left a few notes.

backend/cmd/headlamp.go
yolossn: @upsaurav12 Let's move the cache logic to a middleware as discussed in our 1:1. To start with, make it optional (opt-in). Since this touches the core proxy layer, let's proceed cautiously.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if strings.Contains(dcmp, "Failure") {
+			return false
+		}
+	}
+
+	return true
+}
+
```
yolossn: As discussed in our 1:1,the current `IsAllowed` logic only forwards `selfsubjectrulesreviews` requests and defaults to true for all others, skipping actual permission checks with the K8s API server.

iamundefinedguy: Hi 
I have tried to improve ``isAllowed`` function logic where i am using SSAR request to get ``SelfSubjectAccessReview`` which will contain {denied , allowed} object from where we can know that the user is Authorized or not. This would be used after knowing that key is present. 

If key is not present in the cache then we can't store that error response, due to this if the user doesn't have access to it, we need to make calls to K8's api via ``ProxyRequest``. Since IsAllowed will be used after knowing Key is present we won't test IsAllowed if user hasn't access to it. 


would this workflow correct ? 

EDIT: if i don't have given permissions to ``healthz`` and ``/version`` it closes the connection but when i am trying  it through ``context`` which has all the permissions it working correctly. is it expected behaviour

iamundefinedguy: Currently i am using ``ContextKey`` for differentiating Multi-Clusters, because of it i won't able to implement workflow where i have to check the ``user's`` permissions after checking if the key is present . 
I tried to implement Key generation without using ``ContextKey`` but this is not working correctly in headlamp UI.

iamundefinedguy:
> I left a few notes.

I am sorry i could get these replies. i don't know why but they weren't visible to me?
thanks for suggestion i have made some changes for your requested changes.

backend/cmd/headlamp.go
iamundefinedguy: Done 👍

iamundefinedguy: in K8's Dashboard they have ``args.go``  where they have added various flags for caching purpose. are you trying to say to configure like this ?

backend/pkg/k8cache/key.go
iamundefinedguy: Yes you are right that adding token in CacheKey is not right approach.
Here what we can do is that since i am using ``ContextKey`` in Key so we can use it as unique identifier so we can differentiate in multi context.  So if the user is trying to access we should check first if the user is Authorized or not then check if the key is present or not.

backend/cmd/server.go
```diff
@@ -83,6 +89,68 @@ func main() {
 	})
 }
 
+// CacheMiddleWare is Middleware for Caching purpose.
+func CacheMiddleWare(c *HeadlampConfig) mux.MiddlewareFunc {
```
illume: I think this should be in the k8scache package?

iamundefinedguy: i tried it in ``k8scache`` package but i was unable to import ``HeadalmpConfig``  from ``Headlamp,go`` because i wanted to use it in cache middleware function for using ``handleError`` , and ``KContext``

illume:
> I am sorry i could get these replies. i don't know why but they weren't visible to me?

No worries. I think I wrote them, and then few days came back and noticed they were still pending and submitted the review. I'm not sure why they kept the old date on them, I think it's a github bug. That's why they weren't visible to you - I didn't submit the review yet.

backend/cmd/headlamp.go
illume: thanks

backend/cmd/server.go
illume: It seems like HeadalmpConfig should be moved into a separate package/module so that it can be imported by both headlamp.go and other modules using it. But this move should be done in a separate PR outside of this one, and then this one can then be rebased and import it from the new place.

backend/pkg/k8cache/key.go
illume: Thanks

backend/pkg/k8cache/key.go
illume: thanks

backend/pkg/k8cache/key.go
illume: I don't quite follow, but if that solves the problem of being able to share cached data amongst separate clients that's great! Are there any drawbacks to this you have thought about?

backend/cmd/headlamp.go
illume: thanks

illume: Some command line arguments to turn it on/off would be good (also an environment variable could be handy?). Before we merge caching should be default off, which will allow us to merge it into main so that it can get some wider testing by more people.

illume:
Hi again.

I marked a number of conversations as resolved for all those ones you had addressed. A few conversations are still open, and I responded in a few where you wrote to me. The IsAllowed conversation I don't follow, so I'll leave that to you and yolossn.

backend/pkg/k8cache/key.go
iamundefinedguy: in Kubernetes dashboard's cache implementation they are using ``token`` and ``context`` but for only for multi-cluster setup. Here in headlamp we have multi-cluster already there. I don't see any setup for single cluster in headlamp unless we have only context which is the cluster (admin).  
When i tried to remove ``Context-ID`` from key and check in headlamp it was showing me All ``Active`` or ``Bad Request``, that why we must include ``Context-ID``.
Also they are using exchange token for context key but for that they are using ``exchange-token-endpoint`` . do you think we can implement like this. because they are get this by using ``args``.  

would that approach would be good for this problem ?

backend/cmd/headlamp.go
iamundefinedguy: ok it seems better idea. i was worried that if anything goes wrong and  how i will handle. 
thank for suggestion

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,308 @@
...
+		configOverrides,
+	).ClientConfig()
+	if err != nil {
+		return false
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return false
+	}
```
yolossn: Creating a new clientset everytime there is a cache hit will make the request very slow. It is better to create the clientset once and reuse it.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,308 @@
...
+		},
+	}
+
+	result, err := clientset.AuthorizationV1().SelfSubjectAccessReviews().Create(
+		context.TODO(),
+		review,
+		metav1.CreateOptions{},
+	)
+	if err != nil {
+		return false
```
yolossn: If there's an error while checking user access, we shouldn't assume the user is unauthorized. Instead, we should bypass the cache and allow the proxy request to handle the access control. We shouldn't block legitimate users due to a cache related error.

backend/cmd/server.go
iamundefinedguy: Hi, @illume I was thinking for working on PR to move the ``HeadlampConfig`` to saparate module/package. where do you think i should place this ? I think making a saparate package named as ``headlampConfig`` could be fine?. 
Also the property inside ``HeadlampConfig`` can't be exported( lowecase or unexported ). so for this thing should be consider using saparate methods for each and every properties in it. for example for  ``useInCluster`` we could make like ``headlampConfig.useInCluster()``.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @yolossn  understood. should i consider returning ``err`` from ``IsAllowed`` so that if any err found we can use next.ServeHttp(w , r) to bypass Caching and forward to Proxy ?

backend/pkg/k8cache/k8cache.go
iamundefinedguy: Hi, I have used ``sync.Once`` to create clientset once pls take a look.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: done :+1:

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,337 @@
...
+// Otherwise, it returns false if authorization fails.
+func IsAllowed(url *url.URL,
+	k *kubeconfig.Context,
+	w http.ResponseWriter,
+	r *http.Request,
+) (bool, error) {
+	clientset, err = getClientMD(k, clientset, err, &once)
+	if err != nil {
+		return false, err
+	}
```
yolossn: You can use the `ClientSetWithToken` function that already takes care of creating the clientset
https://github.com/kubernetes-sigs/headlamp/blob/main/backend/pkg/kubeconfig/kubeconfig.go#L273

iamundefinedguy: yes, but it requires token so if the context-id doesn't have token then it might not work.

backend/cmd/server.go
illume: This was merged
- https://github.com/kubernetes-sigs/headlamp/pull/3602

iamundefinedguy: yes i am moving caching middleware to ``k8cache.go``.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: done :+1:

backend/pkg/k8cache/k8cache.go
iamundefinedguy: done

backend/pkg/k8cache/key.go
```diff
@@ -0,0 +1,47 @@
...
+
+// type ResourceKind string
+
+// Key struct used to store different values to make Key unique
+// for different requests.
+type CacheKey struct {
+	Kind      string // Kind is the string object which is the user is requesting
+	Namespace string // Namespace is string object which tells what Namespace is the user trying to access
+	Context   string // Context is the unique Id which helps to differentiate multi-context
+	Token     string
```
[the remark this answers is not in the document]
illume: @upsaurav12 does this comment make any sense?

backend/pkg/k8cache/key.go
illume: @upsaurav12 let me know if you've made changes to address this already?

illume:
@upsaurav12 @yolossn I closed a number of conversations that looked like they were resolved.

backend/pkg/k8cache/key.go
iamundefinedguy: Thanks, 
I was trying to implement the way that Kubernetes UI  had implemented  but it looks like they were using args ``token-exchange-endpoint``,which is the endpoint that will the returns the ``ContextKey`` by exchanging token. In headlamp we can get ``ContextKey`` without using token. 
should we need to add ``token`` to get ``ContextKey`` in ``CacheKey`` ? Also there could be scenerio where we context doesn't use token (maybe client-certificates or something else ).

backend/pkg/k8cache/key.go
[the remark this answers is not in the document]
iamundefinedguy: I have removed token from ``CacheKey``

backend/cmd/headlamp.go
```diff
@@ -1355,7 +1356,9 @@ func (c *HeadlampConfig) handleError(w http.ResponseWriter, ctx context.Context,
 // It parses the request and creates a proxy request to the cluster.
 // That proxy is saved in the cache with the context key.
 func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen
-	router.PathPrefix("/clusters/{clusterName}/{api:.*}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	sub := router.PathPrefix("/clusters/{clusterName}/{api:.*}").Subrouter()
```
yolossn: I think we don't need a subrouter here because we are applying this middleware only to this specific route.

iamundefinedguy: Yes i tried to add middleware without adding ``subrouter`` but it is wasn't consistent.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+	r.ResponseWriter.WriteHeader(code)
+}
+
+func (r *responseCapture) Write(b []byte) (int, error) {
+	r.Body.Write(b)
+	return r.ResponseWriter.Write(b)
+}
+
+// Initialize responseCapture with a http.ResponseWriter and empty bytes.Buffer for the body.
+func Initialize(w http.ResponseWriter) *responseCapture {
```
yolossn: The function name is a bit confusing, when this is used in other packages it will be used like `k8cache.Initialize()` which will give other devs an assumption that the cache is being initialised. Lets rename it to `createResponseCapture/newResponseCapture` or something.

backend/cmd/server.go
```diff
@@ -86,6 +92,70 @@ func main() {
...
+			if err != nil {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			served, err := k8cache.LoadfromCache(k8scache, isAllowed, key, w)
+			if err != nil {
+				c.handleError(w, ctx, span, errors.New(kContext.Error), "failed to load from cache", http.StatusServiceUnavailable)
+				return
+			}
```
yolossn: Error that occurred due to loading from cache shouldn't interrupt the request.

backend/cmd/server.go
```diff
@@ -86,6 +92,70 @@ func main() {
...
+			if err != nil {
+				c.handleError(w, ctx, span, err, "failed to generate key ", http.StatusBadRequest)
+				return
+			}
+
+			isAllowed, err := k8cache.IsAllowed(r.URL, kContext, w, r)
+			if err != nil {
+				next.ServeHTTP(w, r)
+				return
+			}
```
yolossn: If the request is forwarded due to error the response for the request is not cached here for subsequent requests.

backend/cmd/server.go
```diff
@@ -86,6 +92,70 @@ func main() {
...
+
+			rcw := k8cache.Initialize(w)
+
+			key, err := k8cache.GenerateKey(r.URL, contextKey)
+			if err != nil {
+				c.handleError(w, ctx, span, err, "failed to generate key ", http.StatusBadRequest)
+				return
+			}
+
+			isAllowed, err := k8cache.IsAllowed(r.URL, kContext, w, r)
```
yolossn: if `isAllowed` is false then a proper response has to be returned.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+	}
+
+	return false, nil
+}
+
+// If the key was not found inside the cache then this will make actual call to k8's
+// and this will capture the response body and convert the captured response to string.
+// After converting it will store the response with the key and TTL of 10*min.
+func RequestToK8sAndStore(k8scache cache.Cache[string], k *kubeconfig.Context,
```
yolossn: This function does make a request to the k8s api server so lets improve/update the name for better understanding.

yolossn: also please remove unused args from this function.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+
+var (
+	clientsetCache = make(map[string]*kubernetes.Clientset)
+	mu             sync.Mutex
+)
+
+// getClientMD is used to get a clientset for the given context and token.
+// It will reuse clientsets if a matching one is already cached.
+func getClientMD(k *kubeconfig.Context, token string) (*kubernetes.Clientset, error) {
+	contextKey := strings.Split(k.ClusterID, "+")
```
yolossn, line 237: what is the context behind this string split?

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+	return cs, nil
+}
+
+func hashToken(token string) string {
+	if len(token) > 32 {
+		return token[:32]
+	}
+
+	return token
+}
```
yolossn: What is the context behind using the first 32 chars?

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+	apiPath := mux.Vars(r)["api"]
+	parts := strings.Split(apiPath, "/")
+
+	last := parts[len(parts)-1]
+
+	review := &authorizationv1.SelfSubjectAccessReview{
+		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Resource: last,
+				Verb:     "get",
```
yolossn: why is the verb hardcoded to `get` here? the request can also be for listing/watching the resource.

backend/cmd/server.go
```diff
@@ -86,6 +92,70 @@ func main() {
 	})
 }
 
+// CacheMiddleWare is Middleware for Caching purpose.
+func CacheMiddleWare(c *HeadlampConfig) mux.MiddlewareFunc {
```
yolossn: The middleware should check if the request is for updating/creating/deleting a resource and chose not to use cache for those actions. I remember that we discussed this once earlier.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+	for idx, header := range responseHeaders {
+		if strings.EqualFold(idx, "Content-Encoding") && encoding == gzipEncoding {
+			continue
+		}
+
+		cacheHeader[idx] = append(cacheHeader[idx], header...)
+	}
+
+	return cacheHeader
+}
```
yolossn: the function name `SetHeadersToCache` doesn't reflect what is happening in the function. Please update it.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,329 @@
...
+
+var (
+	clientsetCache = make(map[string]*kubernetes.Clientset)
+	mu             sync.Mutex
+)
```
yolossn, line 232: Since we use the token in cache key of `clientsetCache`, the tokens get expired over time and  there is a possibility that the cache is filled with `clientsets` that are expired. We should store the last use time and update the cache to prune unused clientsets in the cache.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: currently, I don't have direct way to get ``contextID``. I can get contextID by using ``k.ClusterID`` but this contains to many things ``/home/saurav/.kubeconfig+<contextID>`` to get contexKey i have to ``split`` ``k.ClusterID``.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: tokens are too long, i thought using 32 char should be unique enough to differentiate tokens. But i think we should use the whole token

backend/pkg/k8cache/k8cache.go
iamundefinedguy: done

illume: @upsaurav12 is this done now?

iamundefinedguy: @illume i have changed the name of function from ``Initialize`` to ``CreateResponseCapture`` , also changed in documentation also

backend/cmd/server.go
iamundefinedguy: @illume here i have removed the return condition , to avoid interupting

backend/cmd/server.go
iamundefinedguy: @illume i have added new function ``StoreAfterAuthError`` in k8cache package that will handle storing and returning to client if there is an error while authorizing user's permissions

backend/cmd/server.go
iamundefinedguy: I have added a new function ``ReturnAuthErrorResponse`` which returns the ``AuthErrResponse`` JSON object if the user is not authorized to access the resources.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: I have removed ``kuberconfig.Context`` arg and changed name from ``RequestToK8sAndStore`` to ``RequestK8ClusterAPIAndStore``.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @illume  I have added function ``GetKindAndVerb``  in ``k8cache`` package to able to return ``Kind`` and ``Verb``

backend/cmd/server.go
iamundefinedguy: In ``LoadFromCache`` i have added a condition which will prevent to go in cache if the request is no-readable (eg. POST , CREATE , DELETE , PUT ) . and return false and error

backend/pkg/k8cache/k8cache.go
iamundefinedguy: I have changed name from SetHeadersToCache to ``FilterHeaderForCache``

backend/pkg/k8cache/k8cache.go
iamundefinedguy: I am currently working in this part and will push the changes

backend/pkg/k8cache/k8cache.go
illume: Looks good, thanks.

backend/cmd/server.go
```diff
@@ -86,6 +94,95 @@ func main() {
...
+
+	return ctx, span, contextKey, kContext, nil
+}
+
+// CacheMiddleWare is Middleware for Caching purpose. It involves generating key for a request,
+// authorizing user , store resource data in cache and returns data if key is present.
+func CacheMiddleWare(c *HeadlampConfig) mux.MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.Method == "POST" || r.Method == "UPDATE" || r.Method == "DELETE" || r.Method == "PUT" {
```
illume: I think this should be `if != "GET"`, so it's an allow list of things we should support. It's usually better to do allow lists in case things change. If things change when it's just GET then it should still work. If it's the opposite way if we've missed something or something else is added this condition will be broken.

What about OPTIONS and HEAD? How should they be handled here?

backend/cmd/server.go
illume: I left a new comment on that line... I think it should be just the methods we support (not the opposite way around).

backend/cmd/headlamp.go
illume: I think the sub router here is a tiny bit cleaner than the alternative.

Can we leave it as is @yolossn ? (I don't feel strongly about it.)

backend/pkg/k8cache/k8cache.go
illume: If you can't think of a cleaner way to do this, can you please leave a comment there explaining what it is about?

backend/pkg/k8cache/key.go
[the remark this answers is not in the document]
illume: thanks

backend/pkg/k8cache/k8cache.go
illume: Cool. That's a better name. Thanks.

backend/cmd/server.go
illume: Cool, thanks. I confirmed the return is gone now.

backend/cmd/server.go
illume: Cool, thanks. Is there a test case that covers this?

I think it's important that we test the authentication parts. I don't see a test for ReturnAuthErrorResponse, which is why I'm wondering if there are other test cases that check this part is correct?

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,417 @@
...
+		_, writeErr := w.Write([]byte(cachedData.Body))
+		if writeErr == nil {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// If the key was not found inside the cache then this will make actual call to k8's
```
illume: Check the comment here follows golang guidelines (function name first).

backend/cmd/server.go
iamundefinedguy: I haven't added test cases for new functions i have added after moving cache logic to ``middleware`` so that why you seeing any test for them. 
If it is necessary i will write those tests once i have completed the cache implementation (ie make cache optional , and many areas to improve).

backend/pkg/k8cache/k8cache.go
iamundefinedguy: I think i missed out many function's documentations to write docs correctly

backend/pkg/k8cache/k8cache.go
illume: @yolossn what do you think of the new name?

(To me the old name RequestToK8sAndStore was fine, so I'm not sure about this one)

backend/pkg/k8cache/k8cache.go
illume: Tokens can be quite long... 8KB for example. But yeah, just using 32 bytes would not make them unique.

I can't see this function anymore? I guess it was removed.

backend/pkg/k8cache/k8cache.go
illume: Thanks

backend/pkg/k8cache/k8cache.go
iamundefinedguy: Yes i removed that function

backend/cmd/server.go
illume: thanks

backend/pkg/k8cache/key.go
illume: Sorry, I'm not clear on this.

Can you please let me know if it's still using a token as part of the cache key?

I feel it shouldn't be so that multiple clients can make use of the cache.

@yolossn what do you think?

iamundefinedguy: @illume Hi i have removed ``token`` from cacheKey

backend/pkg/k8cache/k8cache.go
illume: btw, I don't think we should use the token as part of a cache key. So that multiple clients can use the single cache.

iamundefinedguy: @illume i am using token for creating clientset should it be right approach to use it like this ?

backend/pkg/k8cache/k8cache.go
yolossn: @upsaurav12 you can just pass the cluster name from the request here or use k.ContextName.

backend/cmd/headlamp.go
yolossn: I am suggesting a simpler solution like this which can be easy to read and add cache middleware optionally

```go
cacheMiddleware := CacheMiddleWare(c)
handler := clusterRequestHandler

if c.UseCache() {
    handler = cacheMiddleware(handler)
}

router.PathPrefix("/clusters/{clusterName}/{api:.*}").
    Handler(handler)
```

backend/pkg/k8cache/k8cache.go
iamundefinedguy: will update the code. thanks

backend/cmd/headlamp.go
iamundefinedguy: will push the code by tonight

backend/pkg/k8cache/k8cache.go
yolossn: the client should have the token only then the request for `SelfSubjectAccessReview` will be for the token scoped user

iamundefinedguy: ok I will let this as it is 👍

backend/cmd/headlamp.go
iamundefinedguy: Hi @yolossn @illume  i have implemented this part and made the saparate ``handler`` function
```
func clusterRequestHandler(c *HeadlampConfig) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		start := time.Now()
		ctx := r.Context()

		ctx, span := telemetry.CreateSpan(ctx, r, "cluster-api", "handleClusterAPI",
			attribute.String("cluster", mux.Vars(r)["clusterName"]),
		)
		defer span.End()

		c.telemetryHandler.RecordRequestCount(ctx, r, attribute.String("cluster", mux.Vars(r)["clusterName"]))
		c.telemetryHandler.RecordEvent(span, "Cluster API request started")

		// A deferred function to record duration metrics & log the request completion
		defer recordRequestCompletion(c, ctx, start, r)

		contextKey, err := c.getContextKeyForRequest(r)
		if err != nil {
			c.handleError(w, ctx, span, err, "failed to get context key", http.StatusBadRequest)
			return
		}

		kContext, err := c.KubeConfigStore.GetContext(contextKey)
		if err != nil {
			c.handleError(w, ctx, span, err, "failed to get context", http.StatusNotFound)
			return
		}

		if kContext.Error != "" {
			c.handleError(w, ctx, span, errors.New(kContext.Error), "context has error", http.StatusBadRequest)
			return
		}

		clusterURL, err := url.Parse(kContext.Cluster.Server)
		if err != nil {
			c.handleError(w, ctx, span, err, "failed to parse cluster URL", http.StatusNotFound)
			return
		}

		// Record attributes about the proxy request
		span.SetAttributes(
			attribute.String("cluster.server", kContext.Cluster.Server),
			attribute.String("cluster.api_path", mux.Vars(r)["api"]),
		)
		c.telemetryHandler.RecordClusterProxyRequestsCount(ctx, attribute.String("cluster", contextKey),
			attribute.String("http.method", r.Method))

		r.Host = clusterURL.Host
		r.Header.Set("X-Forwarded-Host", r.Host)
		r.Header.Del("User-Agent")
		r.URL.Host = clusterURL.Host
		r.URL.Path = mux.Vars(r)["api"]
		r.URL.Scheme = clusterURL.Scheme

		// Process WebSocket protocol headers if present
		processWebSocketProtocolHeader(r)
		plugins.HandlePluginReload(c.cache, w)

		if err = kContext.ProxyRequest(w, r); err != nil {
			c.telemetryHandler.RecordErrorCount(ctx, attribute.String("error.type", "proxy_error"),
				attribute.String("cluster", contextKey))
			c.handleError(w, ctx, span, err, "failed to proxy request", http.StatusInternalServerError)

			return
		}

		if c.Telemetry != nil {
			span.SetStatus(codes.Ok, "")
			c.telemetryHandler.RecordEvent(span, "Cluster API request completed")
		}
	})
}

// handleClusterAPI handles cluster API requests. It is responsible for
// all the requests made to /clusters/{clusterName}/{api:.*} endpoint.
// It parses the request and creates a proxy request to the cluster.
// That proxy is saved in the cache with the context key.
func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen

	clusterAPIrequest := clusterRequestHandler(c)

	handler := clusterAPIrequest

	cacheMiddleware := CacheMiddleWare(c)

	if c.CacheEnabled {
		handler = cacheMiddleware(handler)
	}

	router.PathPrefix("/clusters/{clusterName}/{api:.*}").Handler(handler)
}
```
PTAL

backend/pkg/k8cache/key.go
illume: Thanks for the clarification.

backend/cmd/server.go
```diff
@@ -17,15 +17,23 @@ limitations under the License.
...
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/cache"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/config"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/headlampconfig"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/k8cache"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/plugins"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/telemetry"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
```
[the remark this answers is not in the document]
iamundefinedguy: ``attribute`` and ``trace`` should be there for creating ctx and span in ``CacheMiddleware`` function

backend/cmd/headlamp.go
```diff
@@ -1350,36 +1351,11 @@ func (c *HeadlampConfig) handleError(w http.ResponseWriter, ctx context.Context,
...
-		}
-
-		kContext, err := c.KubeConfigStore.GetContext(contextKey)
+func clusterRequestHandler(c *HeadlampConfig) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, span, contextKey, kContext, err := GetContextKeyAndKContext(w, r, c)
 		if err != nil {
-			c.handleError(w, ctx, span, err, "failed to get context", http.StatusNotFound)
-			return
+			c.handleError(w, ctx, span, err, "error while retrieving context and kContext", http.StatusBadRequest)
```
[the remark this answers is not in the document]
illume: It would be good to leave a comment here for why there's no return.

backend/cmd/server.go
illume: Sounds like a good plan.

backend/cmd/server.go
iamundefinedguy: I have updated the code and moved the logic to ``LoadFromCache``. because the request methods like this should not go to cache instead forward to actual request. 

Also there was discussion for cache invalidation where we need to remove the stale data in the cache. but i am not sure what should be the workflow. 
For example if there was a request like ``POST`` on ``secrets`` resource. if that same resource data is present in cache (using key) we can stale the older data in the cache, and store the new data in cache. 
for this we need to check after making request to ``K8's`` (``POST``/``DELETE``) then it will check that the key was in cache or not if present, remove it or else proceed to store that data. 
@yolossn @illume WDYT?

backend/pkg/k8cache/k8cache.go
iamundefinedguy: done

illume: thanks

backend/pkg/k8cache/k8cache.go
illume: Ok. Thanks

backend/cmd/headlamp.go
illume: @yolossn this ok with you?

backend/cmd/server.go
illume: Sounds good. Please mark this resolved when you're done?

iamundefinedguy: will push code once done completely

illume:
@yolossn @upsaurav12 I went through an marked a few convos as resolved that were finished. Note there's a number of convos that are waiting for a response.

What's the plan to work on next? (what's left to do?)

Is it getting near a point where the code should be cleaned up and merged? Maybe now it's a good time to start adding tests, and breaking it up into clean commits?

iamundefinedguy:
yes , @illume i am pushing tests for remaining functions, and resolving the convo's.

backend/cmd/server.go
iamundefinedguy: Hi @illume i have added more tests and increased tests coverage to ``83.2%`` i think it should be enough as a initial point. the remaining are the one's which left over ``error`` tests cases

backend/cmd/server.go
iamundefinedguy: HI @illume i have updated the code, could be please look to the changes ?

backend/pkg/k8cache/k8cache.go
iamundefinedguy: I have tried to change the code to get contextKey to mux.Var(r) but the ``TestHandleClusterAPI_XForwardedHost`` is failing. And kubeconfig.Context doesn't provide any property like ``k.ContextName``. However i found something like k.Name which returns ``ContextKey`` but ``TestHandleClusterAPI_XForwardedHost`` failing with this also.


image

IMO we should use ``k.ClusterId`` and split it to get ``ContextKey``
@illume @yolossn

backend/cmd/headlamp.go
```diff
@@ -1427,6 +1404,24 @@ func handleClusterAPI(c *HeadlampConfig, router *mux.Router) { //nolint:funlen
...
+func handleClusterAPI(c *HeadlampConfig, router *mux.Router) {
+	clusterAPIrequest := clusterRequestHandler(c)
+
+	handler := clusterAPIrequest
+
+	cacheMiddleware := CacheMiddleWare(c)
+
+	if c.CacheEnabled {
+		handler = cacheMiddleware(handler)
+	}
```
yolossn:
```suggestion
	if c.CacheEnabled {
		cacheMiddleware := CacheMiddleWare(c)

		handler = cacheMiddleware(handler)
	}
```

backend/cmd/headlamp.go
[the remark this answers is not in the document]
yolossn: return is added, so closing this.

backend/cmd/server.go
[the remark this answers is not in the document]
yolossn: @upsaurav12 can you format the file using gofumpt ? I think that should resolve this.

backend/cmd/server.go
```diff
@@ -86,6 +94,98 @@ func main() {
...
+func GetContextKeyAndKContext(w http.ResponseWriter,
+	r *http.Request, c *HeadlampConfig) (context.Context,
+	trace.Span, string, *kubeconfig.Context, error,
+) {
+	ctx := r.Context()
+	ctx, span := telemetry.CreateSpan(ctx, r, "cluster-api", "handleClusterAPI",
+		attribute.String("cluster", mux.Vars(r)["clusterName"]),
+	)
+
+	defer span.End()
```
yolossn, line 109: I don't think we need this span to be instrumented. @illume wdyt? this function just gets the context key and context.

backend/pkg/headlampconfig/headlampConfig.go
```diff
@@ -9,6 +9,7 @@ type HeadlampCFG struct {
 	UseInCluster          bool
 	ListenAddr            string
 	DevMode               bool
+	CacheEnabled          bool
```
yolossn, line 12: Thought this variable is added to the struct it is not initialised from the config in the `cmd/server.go`

backend/cmd/server.go
iamundefinedguy: i have added because i was getting linting error of length of ``CacheMiddleware`` function

backend/pkg/headlampconfig/headlampConfig.go
iamundefinedguy: how can i fix this ?

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,544 @@
...
+		}
+	}
+
+	return false, nil
+}
+
+// RequestK8ClusterAPIAndStore ensures if the key was not found inside the cache then this will make actual call to k8's
+// and this will capture the response body and convert the captured response to string.
+// After converting it will store the response with the key and TTL of 10*min.
+func RequestK8ClusterAPIAndStore(k8scache cache.Cache[string],
```
yolossn, line 361: This function doesn't make a request to K8ClusterAPI, please improve the function name here.

iamundefinedguy: @illume said that previous one was good, so can we change the name to previous one

backend/cmd/server.go
```diff
@@ -86,6 +94,98 @@ func main() {
...
+
+			key, err := k8cache.GenerateKey(r.URL, contextKey)
+			if err != nil {
+				c.handleError(w, ctx, span, err, "failed to generate key ", http.StatusBadRequest)
+				return
+			}
+
+			isAllowed, authErr := k8cache.IsAllowed(r.URL, kContext, w, r)
+			if authErr != nil {
+				k8cache.StoreAfterAuthError(k8scache, isAllowed, next, key, w, r, rcw)
```
yolossn, line 152: by default when `authErr` occurs isAllowed is always false. why again attempt `LoadFromCache` in the `StoreAfterAuthError` function?

backend/cmd/server.go
```diff
@@ -86,6 +94,98 @@ func main() {
...
+				c.telemetryHandler.RecordEvent(span, "Served from cache")
+				return
+			}
+
+			next.ServeHTTP(rcw, r)
+
+			err = k8cache.CheckAndPurge(w, r, k8scache, next, rcw, isAllowed)
+			if err != nil {
+				c.handleError(w, ctx, span, err, "error while purging data", http.StatusInternalServerError)
+			}
```
yolossn: Please add a comment here, It is not clear why the Check is done here.

backend/pkg/k8cache/k8cache.go
```diff
@@ -0,0 +1,544 @@
...
+		return nil
+	}
+
+	err := PurgeDataForPostRequest(w, r, k8scache, next, rcw, isAllowed)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
yolossn: Can you please explain the context behind purging post request alone?

iamundefinedguy: I was adding for more methods but from the last meet you asked for only POST

backend/cmd/server.go
iamundefinedguy: because i tried to add it without adding LoadFromCache but it wasn't storing in cache that's why I has added LoadFromCache explicitly.

backend/pkg/k8cache/k8cache.go
yolossn: @upsaurav12 the kubeContext is already there in this functions arg. you can use the kubeContext.name. The reason why I am against splitting the clusterID and using the first part is that clusters of same name can be loaded from different sources then it can cause a same cacheKey to be overwritten and used in different context.

iamundefinedguy: yes I have tried it will k.Name but after adding it is failing ``TestHandleClusterAPI_XForwardedHost``

backend/cmd/server.go
yolossn: AFAIU telemetry doesn't have anything to do with linting issue. Can you share the exact error that you faced?

backend/pkg/headlampconfig/headlampConfig.go
yolossn: In cmd/server.go you have to add something like `CacheEnabled: conf.CachedEnabled`. Did you test by setting the flag on and off? 

https://github.com/kubernetes-sigs/headlamp/blob/d4bcdb2348bccaf9263b9d826b75317d73cdbf43/backend/cmd/server.go#L55

backend/cmd/server.go
iamundefinedguy: no, i am talking about ``funlen`` linting issue, it was saying that the function length is >60 lines of code, that's why i made a saparate function that will return ``contextKey`` , ``KContext``, ``span`` and ``ctx``

backend/pkg/headlampconfig/headlampConfig.go
iamundefinedguy: yes i checked with ``!CacheEnabled`` and ``CacheEnabled`` it was working correctly (like caching things) but i will try to initialize ``CacheEnabled: conf.CachedEnabled``

backend/cmd/server.go
yolossn: Why is storing the response to cache dependent on checking if the cache already has it. Please correct me if I am wrong here. Since in this case the `k8cache.IsAllowed` returns an error isAllowed will always return false right? And using the isAllowed(false) value in LoadFromCache will always return false. This step doesn't do anything.

backend/pkg/k8cache/k8cache.go
yolossn: Looks like there is a misunderstanding here, all kind of update request that can make the cache stale should be handled here. ie put, post, delete

backend/cmd/server.go
iamundefinedguy: IMO if there was an error but the user is authorized( ie IsAllowed: true) then we can't assume that isAllowed: false, that's why 
```
if err != nil {
		if result.Status.Allowed {
			return true, err
		}

		return false, err
	}
```

backend/pkg/k8cache/k8cache.go
iamundefinedguy: Now i am pushing the changes

yolossn: Also synchronously looping through every request in cache and deleting the request at the end of each update request will make it slower. Also the method type check in the cacheMiddleware is removed. Almost all of the update request (put,post,delete) will be unique and we need not have to check them in the cache.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: would ``RequestToK8AndStore`` be good for you @yolossn

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @yolossn @illume 
would this work ? 
```
// PurgeDataForPostRequest remove the stale data which is related to POST request.
// It uses UnmarshalCacheData to filter the response for purging the resource data.
// It also uses GetAll method from cache.Cache to retrieve all the data that was stored.
func PurgeDataForPostRequest(w http.ResponseWriter, r *http.Request, k8scache cache.Cache[string],
	next http.Handler, rcw *responseCapture, isAllowed bool,
) error {
	lasts, _ := GetKindAndVerb(r)

	last := Singular(lasts)

	go func() error {

		// start := time.Now()
		val, err := k8scache.GetAll(context.Background(), nil)
		if err != nil {
			return err
		}

		for key, v := range val {
			var cachedData CachedResponseData

			_, err := UnmarshalCacheData(v, cachedData)
			if err != nil {
				return err
			}

			err = ProcessCachedItem(key, v, last, k8scache, next, w, r, rcw, isAllowed)
			if err != nil {
				return err
			}
		}

		// elapsed := time.Since(start)
		// fmt.Println("elasped : ", elapsed)

		return nil
	}()

	return nil
}
```

backend/cmd/server.go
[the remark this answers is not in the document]
iamundefinedguy: yes doing it.

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @yolossn can we use ``k.kubeContext.AuthInfo`` value in this area as it providing better output of different user's.
it returns ``user`` name details not context ?

backend/pkg/headlampconfig/headlampConfig.go
iamundefinedguy: I have added it in ``CaceEnabled: conf.CacheEnabled``

backend/pkg/k8cache/k8cache.go
iamundefinedguy: i have made some changes in the code

backend/cmd/server.go
yolossn: separate function is okay, not every new function needs to have a telemetry span.

iamundefinedguy: okay understood, should i need to remove telemetry and span for some new functions such as ``ReturnAuthErrRespone`` , ``ReturnAfterAuthError`` etc

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @yolossn here ?

backend/pkg/k8cache/k8cache.go
iamundefinedguy: @yolossn i have use this AuthInfo in clientSet Key but it i don't know why it is failing e2e test and but using ``k.ClusterID`` not .
@illume @yolossn ?

backend/cmd/server.go
yolossn: for now, lets remove this.

PR #3504, opened 2025-06-20
Move cluster chooser to the sidebar

### LINKED ISSUE #3189: Irrelevant cluster selector button
## Describe the bug

The cluster selector button UX is such that the button may not be as used as it could. It's placed away from where there's a reference to the cluster name (cluster name on the left in the sidebar, cluster selection button on the top right) and with the recent theme changes, it looks rather dull and unnoticeable.

We need to find a way to make the cluster selector button more highlighted, relocate it, integrate it into the search bar, or some other solution. Maybe it's not even needed now that we have the search bar and the home view.

Current way:

Cluster button screenshot

Cluster button with selection popover screenshot

 - Installation type: any

joaquimrocha:
This does the following:
1. Adds support for sidebar items to support an action (like the favorites [PR](https://github.com/kubernetes-sigs/headlamp/pull/3150) does)
2. Moves the cluster chooser to the cluster sidebar item
3. Removes the topbar cluster chooser (leaves the logic for the plugins who override it though)

Screenshot showing the cluster chooser action
Screenshot showing the cluster chooser opened

## What's missing
- [ ] Support for plugins to override the cluster chooser action
- [ ] Check if/how plugins can register these actions
- [ ] Make the cluster chooser work while the sidebar is collapsed (due to the actions not being shown, the chooser is not displayed)

## How to test
* Go to a cluster view -> Verify the top cluster chooser button is no longer visible
* Verify there's a new button in the cluster sidebar item, click it (or use Ctrl+Shift+L) -> The cluster chooser popup is shown

Since I may not be able to advance much more on it, I'd like @sniok to check it out and eventually add the missing parts (the branch is in the headlamp-k8s org).

fixes #3189

### DIFF
--- a/frontend/src/components/App/__snapshots__/TopBar.TwoCluster.stories.storyshot
+++ b/frontend/src/components/App/__snapshots__/TopBar.TwoCluster.stories.storyshot
@@ -118,26 +118,7 @@
         </div>
         <div
           class="MuiBox-root css-0"
-        >
-          <button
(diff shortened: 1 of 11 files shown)


joaquimrocha:
I believe @sniok did a POC of a solution that works nicer than this.

PR #3521, opened 2025-06-24
tools/releaser: Bump octokit/rest to 22.0.0

skoeva:
This change updates the @octokit/rest dependency to 22.0.0, which consequently updates the following dependencies as desired:

- @octokit/request to 10.0.3
- @octokit/request-error to 7.0.0
- @octokit/plugin-paginate-rest to 13.1.0


joaquimrocha:
@skoeva Did you test it after the deps were updated?

skoeva:
@joaquimrocha smoke tested locally and seems to be working as expected on my end

PR #3534, opened 2025-06-26
Makefile: Update frontend-lint-fix command

skoeva:
This change updates the frontend lint fix make command from `make frontend-fixlint` to `make frontend-lint-fix`, in line with the backend lint commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ run-only-app:
 frontend-lint:
 	cd frontend && npm run lint -- --max-warnings 0 && npm run format-check
 
-frontend-fixlint:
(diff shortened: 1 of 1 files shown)


sniok:
nice!

/lgtm

mboersma:
Also, it would be nice to have a `lint-fix` target with `backend-lint-fix` and `frontend-lint-fix` as subtargets.

skoeva:
> 👍🏻 Much better.
> 
> Should we leave the `frontend-fixlint` target as well, so we don't break anyone who is using it?

@mboersma That probably isn't necessary, I think the other idea could be useful though

PR #3541, opened 2025-06-27
frontend: ingress: Display missing port for default backend

skoeva:
This change displays the proper port number for the default backend, which was previously not shown. When no port is given, it falls back to the port name.


illume (approved):
Thanks heaps. 🎉

PR #3542, opened 2025-06-27
Accessibility [FastPass]: Fix settings general missing aria input labels 

vyncent-t:
## Summary

This PR fixes accessibility related issues for the settings page, mainly just the issues found by FastPass on the `General` tab. 

These changes are part of a larger accessibility compliance goal from ADO (not sure if linking is allowed).

Issues are found using FastPass automated checks, feel free to download the tool and doublecheck if interested.

## Changes

These aria labels are now added to the settings view

- Add aria labelled by for `Sort sidebar items alphabetically` menu switch
- Add aria labelled by for `Use evict for pod deletion` menu switch
- Add aria label for `Number of rows for tables` source component
- Add aria label for `Timezone to display for dates` source component
- Add aria label for drawer mode switch
- Updates `Language` to use typography (was plain text before)

## Note
I have ran into issues with applying some aria attributes to the MUI components directly. I found that just attaching `aira-label` / `aira-labelledby` etc would only attach to the wrapper for the MUI component and not the input element itself. By adding the `inputProps` field we can attach aria labels to the actual input element.

### Issue for `Number of rows for tables`
image
image

### Issue for `Timezone to display for dates`
image
image

### issue for `Sort sidebar items alphabetically`
image
image

### issue for `Use evict for pod deletion`
image
image

### DIFF
--- a/frontend/src/components/App/Settings/NumRowsInput.tsx
+++ b/frontend/src/components/App/Settings/NumRowsInput.tsx
@@ -31,9 +31,9 @@ import { useDispatch } from 'react-redux';
 import { getTablesRowsPerPage, setTablesRowsPerPage } from '../../../helpers/tablesRowsPerPage';
 import { defaultTableRowsPerPageOptions, setAppSettings } from '../../../redux/configSlice';
 
-export default function NumRowsInput(props: { defaultValue: number[] }) {
+export default function NumRowsInput(props: { defaultValue: number[]; nameLabelID?: string }) {
   const { t } = useTranslation(['frequent']);
-  const { defaultValue } = props;
+  const { defaultValue, nameLabelID } = props;
   const [isSelectOpen, setIsSelectOpen] = useState(false);
   const [options, setOptions] = useState(defaultValue);
   const focusedRef = useCallback((node: HTMLElement) => {
@@ -157,6 +157,7 @@ export default function NumRowsInput(props: { defaultValue: number[] }) {
           renderValue={value => `${value}`}
           size="small"
           variant="outlined"
+          labelId={nameLabelID}
         >
           {options.map(option => {
             const isCustom = !defaultTableRowsPerPageOptions.includes(option);
--- a/frontend/src/components/App/Settings/Settings.tsx
+++ b/frontend/src/components/App/Settings/Settings.tsx
@@ -78,6 +78,7 @@ export default function Settings() {
   }, [useEvict]);
 
   const sidebarLabelID = 'sort-sidebar-label';
+  const evictLabelID = 'use-evict-label';
 
   return (
     <SectionBox
@@ -146,8 +147,12 @@ export default function Settings() {
                 color="primary"
                 checked={useEvict}
                 onChange={e => setUseEvict(e.target.checked)}
+                inputProps={{
+                  'aria-labelledby': evictLabelID,
+                }}
               />
             ),
+            nameID: evictLabelID,
           },
         ]}
       />
--- a/frontend/src/components/App/Settings/Settings.tsx
+++ b/frontend/src/components/App/Settings/Settings.tsx
@@ -80,6 +80,7 @@ export default function Settings() {
   const sidebarLabelID = 'sort-sidebar-label';
   const evictLabelID = 'use-evict-label';
   const tableRowsLabelID = 'rows-per-page-label';
+  const timezoneLabelID = 'timezone-label';
 
   return (
     <SectionBox
@@ -125,9 +126,11 @@ export default function Settings() {
                 <TimezoneSelect
                   initialTimezone={selectedTimezone}
                   onChange={name => setSelectedTimezone(name)}
+                  nameLabelID={timezoneLabelID}
                 />
               </Box>
             ),
+            nameID: timezoneLabelID,
           },
           {
             name: t('translation|Sort sidebar items alphabetically'),
--- a/frontend/src/components/App/Settings/Settings.tsx
+++ b/frontend/src/components/App/Settings/Settings.tsx
@@ -79,6 +79,7 @@ export default function Settings() {
 
   const sidebarLabelID = 'sort-sidebar-label';
   const evictLabelID = 'use-evict-label';
+  const tableRowsLabelID = 'rows-per-page-label';
 
   return (
     <SectionBox
@@ -112,8 +113,10 @@ export default function Settings() {
             value: (
               <NumRowsInput
                 defaultValue={storedRowsPerPageOptions || defaultTableRowsPerPageOptions}
+                nameLabelID={tableRowsLabelID}
               />
             ),
+            nameID: tableRowsLabelID,
           },
           {
             name: t('translation|Timezone to display for dates'),
--- a/frontend/src/components/App/Settings/Settings.tsx
+++ b/frontend/src/components/App/Settings/Settings.tsx
@@ -77,6 +77,8 @@ export default function Settings() {
     );
   }, [useEvict]);
 
+  const sidebarLabelID = 'sort-sidebar-label';
+
   return (
     <SectionBox
       title={t('translation|General Settings')}
@@ -130,8 +132,12 @@ export default function Settings() {
                 color="primary"
                 checked={sortSidebar}
                 onChange={e => setSortSidebar(e.target.checked)}
+                inputProps={{
+                  'aria-labelledby': sidebarLabelID,
+                }}
               />
             ),
+            nameID: sidebarLabelID,
           },
           {
             name: t('translation|Use evict for pod deletion'),
--- a/frontend/src/components/common/NameValueTable/NameValueTable.tsx
+++ b/frontend/src/components/common/NameValueTable/NameValueTable.tsx
@@ -32,6 +32,8 @@ export interface NameValueTableRow {
   valueCellProps?: GridProps;
   /** Whether to highlight the row (used for titles, separators, etc.). */
   withHighlightStyle?: boolean;
+  /** The ID to use for the name element, useful for accessibility */
+  nameID?: string;
 }
 
 export interface NameValueTableProps {
@@ -87,7 +89,10 @@ export default function NameValueTable(props: NameValueTableProps) {
       })}
     >
       {visibleRows.flatMap(
-        ({ name, value, hide = false, withHighlightStyle = false, valueCellProps = {} }, i) => {
+        (
+          { name, nameID, value, hide = false, withHighlightStyle = false, valueCellProps = {} },
+          i
+        ) => {
           let shouldHide = false;
           if (typeof hide === 'function') {
             shouldHide = hide(value);
@@ -112,6 +117,7 @@ export default function NameValueTable(props: NameValueTableProps) {
               sm={hideValueGridItem ? 12 : 4}
               component="dt"
               className={className}
+              id={nameID}
               sx={theme => {
                 const extra = withHighlightStyle
                   ? {
--- a/frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
+++ b/frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
@@ -23,10 +23,12 @@ import spacetime from 'spacetime';
 export interface TimezoneSelectorProps {
   initialTimezone?: string;
   onChange: (timezone: string) => void;
+  /** The custom ID to be used when this component is inside NameValueTable for ARIA labelledby */
+  nameLabelID?: string;
 }
 
 export default function TimezoneSelect(props: TimezoneSelectorProps) {
-  const { onChange, initialTimezone } = props;
+  const { onChange, initialTimezone, nameLabelID } = props;
   const { i18n, t } = useTranslation();
   const timezoneOptions = React.useMemo(() => {
     const timezoneNames = spacetime.timezones();
@@ -51,7 +53,16 @@ export default function TimezoneSelect(props: TimezoneSelectorProps) {
       includeInputInList
       openOnFocus
       renderInput={params => (
-        <TextField {...params} helperText={t('Timezone')} size="small" variant="outlined" />
+        <TextField
+          {...params}
+          helperText={t('Timezone')}
+          size="small"
+          variant="outlined"
+          inputProps={{
+            ...params.inputProps,
+            'aria-labelledby': nameLabelID,
+          }}
+        />
       )}
       onChange={(_ev, value) => onChange(value.name)}
       value={timezoneOptions.find(option => option.name === initialTimezone)}
--- a/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
+++ b/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
@@ -379,6 +379,7 @@
           </dd>
           <dt
             class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-4 css-iqixpy-MuiGrid-root"
+            id="use-evict-label"
           >
             Use evict for pod deletion
           </dt>
@@ -392,6 +393,7 @@
                 class="MuiButtonBase-root MuiSwitch-switchBase MuiSwitch-colorPrimary Mui-checked PrivateSwitchBase-root MuiSwitch-switchBase MuiSwitch-colorPrimary Mui-checked Mui-checked css-1emuodk-MuiButtonBase-root-MuiSwitch-switchBase"
               >
                 <input
+                  aria-labelledby="use-evict-label"
                   checked=""
                   class="PrivateSwitchBase-input MuiSwitch-input css-1m9pwf3"
                   type="checkbox"
--- a/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
+++ b/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
@@ -202,6 +202,7 @@
           </dd>
           <dt
             class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-4 css-1iczkge-MuiGrid-root"
+            id="rows-per-page-label"
           >
             Number of rows for tables
           </dt>
@@ -219,6 +220,7 @@
                   aria-controls=":mock-test-id:"
                   aria-expanded="false"
                   aria-haspopup="listbox"
+                  aria-labelledby="rows-per-page-label"
                   class="MuiSelect-select MuiSelect-outlined MuiInputBase-input MuiOutlinedInput-input MuiInputBase-inputSizeSmall css-oswssk-MuiSelect-select-MuiInputBase-input-MuiOutlinedInput-input"
                   role="combobox"
                   tabindex="0"
--- a/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
+++ b/frontend/src/components/App/Settings/__snapshots__/Settings.General.stories.storyshot
@@ -264,6 +264,7 @@
           </dd>
           <dt
             class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-4 css-1iczkge-MuiGrid-root"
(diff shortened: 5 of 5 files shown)


frontend/src/components/App/Settings/NumRowsInput.tsx
```diff
@@ -157,6 +157,9 @@ export default function NumRowsInput(props: { defaultValue: number[] }) {
           renderValue={value => `${value}`}
           size="small"
           variant="outlined"
+          inputProps={{
+            'aria-label': `option ${selectedValue}`,
+          }}
```
joaquimrocha: I am surprised this is not set by default. Can you elaborate what is the test failing here?
In any case, a solution would need the aria-label value to be translatable. Not sure if "option" is a good idea here either. AFAIU, screen readers already should understand this is part of a select.

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
```diff
@@ -51,7 +51,16 @@ export default function TimezoneSelect(props: TimezoneSelectorProps) {
...
       renderInput={params => (
-        <TextField {...params} helperText={t('Timezone')} size="small" variant="outlined" />
+        <TextField
+          {...params}
+          helperText={t('Timezone')}
+          size="small"
+          variant="outlined"
+          inputProps={{
+            ...params.inputProps,
+            'aria-label': t('Timezone'),
```
joaquimrocha: Same comment as to why this is needed. Can you double check if MUI is not setting this and we're just missing some param?

joaquimrocha (changes requested):

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
vyncent-t: For the timezone select component the FastPass error prints as follows:

image

It does not look like MUI has any aria labels or labelled by here.

I have tried adding just the `aira-label` prop to `<Autocomplete>` and `<TextField>` individually, which still returns the same check error

I have also tried using `aria-labelledby="cluster-selector-autocomplete"` within `<TextField>`, since `<Autocomplete>` wraps it via `renderInput` prop, this also returns the same check error.

I opted for adding the aria label to the spread within `renderInput` method as per the docs https://mui.com/material-ui/api/autocomplete/#props

frontend/src/components/App/Settings/NumRowsInput.tsx
vyncent-t: The FastPass error check for the NumRowsInput: 

image

This error says it is missing an `aria-label` or `aria-labelledby`.

Also tried adding an `aria-label` prop here, still returns that same error above and additionally introduces need for "role" semantics 

image

since adding them as props normally seems to just add to a MUI wrapper and not the input element itself I use `inputProps`

vyncent-t: >Not sure if "option" is a good idea here either. 

~~I agree, could use some ideas for a more fitting word than 'option'.~~

~~Option just seemed to be the a fitting enough catch all word for this select of preset numbers or a custom input (was between using "option" or "choice")~~

~~maybe something like~~
```
          inputProps={{
            'aria-label': t('translation|{{ value }} number of rows', { value: selectedValue }),
          }}
```

> AFAIU, screen readers already should understand this is part of a select.

~~An alternative method would be to try to wrap `<MenuItem>` inside some label and link it via an id= and aria-labelledby.~~

~~let me know if that is a good fit~~

vyncent-t: ^^^

**update: I was able to use the same method of adding an id to the name label that I am using elsewhere for the settings table. This now links and references the name label directly without need for "options" keyword or the like. These changes also pass the check**

illume:
Nice work @vyncent-t 

Maybe try enabling the a11y test on the settings e2e test? I guess it will pass now.
https://github.com/kubernetes-sigs/headlamp/blob/1a0cf56a633216107648e8841e59b12be13d9442/e2e-tests/tests/pluginSetting.spec.ts#L50

vyncent-t:
> Nice work @vyncent-t
> 
> Maybe try enabling the a11y test on the settings e2e test? I guess it will pass now.
> 
> https://github.com/kubernetes-sigs/headlamp/blob/1a0cf56a633216107648e8841e59b12be13d9442/e2e-tests/tests/pluginSetting.spec.ts#L50

These fixes are for the /settings/general view (which I dont think we have playwright tests for yet?) so I think this may still be an issue for the /settings/plugins view

currently the FastPass accessibility checker has no issues with the elements on /settings/plugins but I will run an experimental push to see if it is still a dragons nest 🤔 

we should create playwright tests for settings/general view though, will add that to my to do list later

vyncent-t:
rebased push

vyncent-t:
CI passes with normal

trying experimental now

vyncent-t:
update: trying to enable the a11y tests in pluginSettings still fails, Fast Pass shows no issues on /settings/plugins.

https://github.com/kubernetes-sigs/headlamp/blob/1a0cf56a633216107648e8841e59b12be13d9442/e2e-tests/tests/pluginSetting.spec.ts#L50

build failure: https://github.com/kubernetes-sigs/headlamp/actions/runs/16199733738/job/45735435622?pr=3542

This means that the checks that FastPass runs do not overlap much with the axeBuilder checks we use in our e2e tets. We should look into having some e2e tests for the /settings/general view in the future

vyncent-t:
reverting to normal aria related fixes for /settings/general without e2e changes

illume:
> update: trying to enable the a11y tests in pluginSettings still fails, Fast Pass shows no issues on /settings/plugins.

The spirit of the request is that there are no issues found with automated tools before humans take a look. There's a rule of thumb based on studies that only 30% of a11y issues are found with automated tools.

vyncent-t:
This PR is XXL due to the stories update. 

I am changing the <NameValueTable> items to now take a nameID optional param for the new <Typography> wrapped names. The names are now wrapped in Typography comps so that we may attach IDs and use aria labelledby for paired inputs. 

The nameID being optional does not prevent any of the current usage of NameValueTable elsewhere in the project, it just allows us to use IDs when we want to pair aria labels to things.

frontend/src/components/App/Settings/Settings.tsx
```diff
@@ -77,6 +77,12 @@ export default function Settings() {
     );
   }, [useEvict]);
 
+  const sidebarLabel = 'sort-sidebar-label';
+  const evictLabel = 'use-evict-label';
+  const tableRowsLabel = 'rows-per-page-label';
+  const timezoneLabel = 'timezone-label';
+  const overviewLabel = 'overview-label';
```
vyncent-t: using consts here to make sure the labels being passed are consistent and aren't prone to typos

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -143,7 +149,7 @@ export default function NameValueTable(props: NameValueTableProps) {
                 };
               }}
             >
-              {name}
+              <Typography {...(nameID && { id: nameID })}>{name}</Typography>
```
vyncent-t: wrapping the name label within a Typography here so we can add an `id` via nameID (it is currently optional as to not break anything elsewhere in the project).

This will be used in most situations where we use input UI within NameValueTable, the grid format doesn't automatically relate each element and therefore we need to link them via `id` (for the name) and `aria-labelledby` (within the direct MUI element, usually by `inputProps`)

vyncent-t:
had to revert changes for the drawer mode settings, was passing in props that weren't optional

vyncent-t:
@joaquimrocha not urgent

Reworked the way I was assigning typography names to NameValueTable at only one view (settings).

NameValueTable names now wrap every name in a Typography so that an `id` can be passed to it (via `nameID`), making it easier to assign `aria-labelledby`.

vyncent-t:
@joaquimrocha fixed the spelling and should be good to go

vyncent-t:
> @joaquimrocha fixed the spelling and should be good to go

noticed an issue with the timezones after rebasing need to fix this first

sniok:
The timezone autocomplete stopped working, empty input and clicking on it crashes the page
image

Also several inputs don't have id's anymore 
image

I think autoselect input should be labeled by the label on the left
image

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
```diff
@@ -23,10 +23,12 @@ import spacetime from 'spacetime';
 export interface TimezoneSelectorProps {
   initialTimezone?: string;
   onChange: (timezone: string) => void;
+  /** The ID to be used when inside a table grid for ARIA labelledby */
+  tableID?: string;
```
sniok: I'm not sure why all the props are called tableID, there's no table on this page and also I think this refers to the label element right? Please give it a more descriptive name

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -140,13 +140,14 @@ const OptionButton = ({
...
-export default function DrawerModeSettings() {
+export default function DrawerModeSettings(props: { tableID?: string }) {
   const dispatch = useDispatch();
+  const { tableID } = props;
 
   const isDrawerEnabled = useTypedSelector(state => state.drawerMode.isDetailDrawerEnabled);
 
   return (
-    <Box sx={{ display: 'flex' }}>
+    <Box sx={{ display: 'flex' }} {...(tableID && { 'aria-labelledby': tableID })}>
```
sniok: this Box is not an interactive element and shouldn't have a aria-labelledby

frontend/src/components/App/Settings/NumRowsInput.tsx
```diff
@@ -157,6 +157,7 @@ export default function NumRowsInput(props: { defaultValue: number[] }) {
           renderValue={value => `${value}`}
           size="small"
           variant="outlined"
+          {...(tableID && { labelId: tableID, 'aria-labelledby': tableID })}
```
sniok: aria-labelledby is not required, https://mui.com/material-ui/react-select/#accessibility

vyncent-t:
> The timezone autocomplete stopped working, empty input and clicking on it crashes the page image
> 

This one I need to fix, had it working before a rebase


> Also several inputs don't have id's anymore image
> 

they use aria labelledby since they are in the NameValueTable, I believe using aria-label on its own is for non visible text

> I think autoselect input should be labeled by the label on the left image

I think the FastPass check needs it to be on the last input from MUI in this, it doesn't let us not add something to the text field

the error when moving it to the autoselect
image

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
vyncent-t: I went with tableID as that is the name of the id we want to put in for the NameValueTable we have in settings page, would something like `labelID` be better?

frontend/src/components/App/Settings/DrawerModeSettings.tsx
vyncent-t: since we can only attach one id to one input and this box is a group of two buttons, I put the container of them to have the lablledby.

I read that box should be treated as groups semantically, maybe adding `role=group` here is the way to go?

frontend/src/components/App/Settings/NumRowsInput.tsx
vyncent-t: removing

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
vyncent-t: named to `nameLabelID`

frontend/src/components/App/Settings/DrawerModeSettings.tsx
vyncent-t: > this Box is not an interactive element and shouldn't have a aria-labelledby

adding role="group" to the box for semantic grouping, this way FastPass still passes and screen readers will read all options per one id

sniok: with group role it's okay.
"screen readers will read all options per one id" what do you mean by this? which screen reader are you using to test this?

vyncent-t: I use the built in windows screen reader (crtl + win + enter)

vyncent-t:
@joaquimrocha 

if the changes for NameValueTable are good here I can remove the duplicate work in https://github.com/kubernetes-sigs/headlamp/pull/3598 and it should be ready for final review

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -116,14 +116,17 @@ const OptionButton = ({
...
 }: {
   children: ReactNode;
   active: boolean;
   onClick: React.MouseEventHandler<HTMLButtonElement>;
+  labelId: string;
 }) => (
   <Button
     aria-pressed={active}
     onClick={onClick}
+    aria-label={labelId}
```
sniok: aria-label is used for a readable label not an id, the `windowLabelID` is not used as an id on any element, and since the label is already inside the button you don't really need to do any of this. Are you sure there's an issue here?

frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx
```diff
@@ -51,7 +53,18 @@ export default function TimezoneSelect(props: TimezoneSelectorProps) {
       includeInputInList
       openOnFocus
       renderInput={params => (
-        <TextField {...params} helperText={t('Timezone')} size="small" variant="outlined" />
+        <TextField
+          {...params}
+          helperText={t('Timezone')}
+          size="small"
+          variant="outlined"
+          {...(nameLabelID && {
```
sniok: no need to complicate things here
```tsx
inputProps={{
  ...params.inputProps,
  'aria-labelledby': nameLabelID,
}}
```

frontend/src/components/common/NameValueTable/NameValueTable.tsx
sniok: no need for complicated prop passing
`id={nameID}` is enough

and you should put `id` on the Grid element above, no need for extra Typography

frontend/src/components/App/Settings/DrawerModeSettings.tsx
```diff
@@ -145,14 +148,22 @@ export default function DrawerModeSettings() {
 
   const isDrawerEnabled = useTypedSelector(state => state.drawerMode.isDetailDrawerEnabled);
 
+  const windowLabelID = `resource-details-view-option-window`;
```
sniok: this is not needed

frontend/src/components/App/Settings/NumRowsInput.tsx
```diff
@@ -157,6 +157,7 @@ export default function NumRowsInput(props: { defaultValue: number[] }) {
           renderValue={value => `${value}`}
           size="small"
           variant="outlined"
+          {...(nameLabelID && { labelId: nameLabelID })}
```
sniok: just `labelId={nameLabelID}` is enough

sniok (changes requested):

frontend/src/components/common/NameValueTable/NameValueTable.tsx
vyncent-t: I have this to be an optional usage so we don't break anything since we use an id in every place use a name in NameValueTable

I can move id to the grid but I think it may need to be conditional still to avoid empty id=""

frontend/src/components/App/Settings/DrawerModeSettings.tsx
vyncent-t: I noticed that the options were not being read well enough in relation to the settings name. When using the screen reader, it would just read off "window button" or "full screen button" without mention of the overlay or resource view context

Here I am putting in more descriptive labels to be passed into each of the buttons which the screen reader does read  via
```
  const windowLabelID = `resource-details-view-option-window`;
  const fullPageLabelID = `resource-details-view-option-full-page`;
  ```
  
  ```
    return (
    <Box sx={{ display: 'flex' }}>
      <OptionButton
        active={isDrawerEnabled}
        labelId={windowLabelID}
        onClick={() => dispatch(setDetailDrawerEnabled(true))}
      >
        <OverlayPreview variant="overlay" />
        <Trans>Window</Trans>
      </OptionButton>
      <OptionButton
        active={!isDrawerEnabled}
        labelId={fullPageLabelID}
        onClick={() => dispatch(setDetailDrawerEnabled(false))}
      >
        <OverlayPreview variant="full-page" />
        <Trans>Full page</Trans>
      </OptionButton>
    </Box>
  );
  ```
  

I can turn the `windowLabelID` and `fullPageLabelID` into just `windowLabel` and `fullPageLabel` if that makes it cleaner

File: frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx (hunk 9 above)
Comment (vyncent-t) [reply frontend/src/components/common/TimezoneSelect/TimezoneSelect.tsx]: fixed

File: frontend/src/components/App/Settings/DrawerModeSettings.tsx (hunk 10 above)
Comment (vyncent-t) [reply frontend/src/components/App/Settings/DrawerModeSettings.tsx]: it only reads "window button" on screen readers without it, is that okay without mention of the resource details view?

File: frontend/src/components/App/Settings/NumRowsInput.tsx (hunk 11 above)
Comment (vyncent-t) [reply frontend/src/components/App/Settings/NumRowsInput.tsx]: fixed

File: frontend/src/components/App/Settings/DrawerModeSettings.tsx (hunk 8 above)
Comment (sniok) [reply frontend/src/components/App/Settings/DrawerModeSettings.tsx]: You're currently passing an ID into aria-label which is incorrect. If your intention is to add a better label to the buttons, it should be a human readable label passed as aria-label OR id of a separate label element passed as aria-labelledby

File: frontend/src/components/common/NameValueTable/NameValueTable.tsx (hunk 4 above)
Comment (sniok) [reply frontend/src/components/common/NameValueTable/NameValueTable.tsx]: nameID is optional. id can be undefined

File: frontend/src/components/App/Settings/NumRowsInput.tsx (hunk 11 above)
Comment (sniok) [reply frontend/src/components/App/Settings/NumRowsInput.tsx]: I don't think you've pushed your changes yet

File: frontend/src/components/App/Settings/DrawerModeSettings.tsx (hunk 8 above)
Comment (vyncent-t) [reply frontend/src/components/App/Settings/DrawerModeSettings.tsx]: I received some additional feedback about these changes with concerns to translations also. I will just remove these changes since they aren't in the FastPass failures anyway.

File: frontend/src/components/common/NameValueTable/NameValueTable.tsx (hunk 4 above)
Comment (vyncent-t) [reply frontend/src/components/common/NameValueTable/NameValueTable.tsx]: fixed

File: frontend/src/components/App/Settings/DrawerModeSettings.tsx (hunk 10 above)
Comment (vyncent-t) [reply frontend/src/components/App/Settings/DrawerModeSettings.tsx]: removed

File: frontend/src/components/App/Settings/NumRowsInput.tsx (hunk 11 above)
Comment (vyncent-t) [reply frontend/src/components/App/Settings/NumRowsInput.tsx]: pushing

File: frontend/src/components/common/NameValueTable/NameValueTable.tsx (hunk 4 above)
Comment (sniok) [reply frontend/src/components/common/NameValueTable/NameValueTable.tsx]: no need for Typography

File: frontend/src/components/App/Settings/Settings.tsx [hunk 12]
```diff
@@ -77,6 +77,11 @@ export default function Settings() {
     );
   }, [useEvict]);
 
+  const sidebarLabel = 'sort-sidebar-label';
```
sniok: should be named something like `sidebarLabelId`

vyncent-t: fixed

frontend/src/components/common/NameValueTable/NameValueTable.tsx
vyncent-t: fixed

PR #3543, opened 2025-06-30
frontend: CronJob: Spawn new job using cronjob's namespace

skoeva:
This change sets a newly spawn job's namespace to the namespace of its cron job, rather than defaulting to the `default` namespace.

### Testing
- [X] Open a cronjob in Headlamp
- [X] Spawn a new job and inspect the network requests
- [X] Ensure that the cronjob's namespace, rather than the default one, is displayed in the POST request headers

image


sniok (approved):
/lgtm

PR #3568, opened 2025-07-04
Use shx

joaquimrocha:
This replaces rimraf by shx rm -rf like we have in the frontend and adds the dev-dep to the frontend because in systems that build without internet, the list of all deps is needed but this one was not in the list.


sniok:
headlamp-plugin dependency check fails

Issue #3419, opened 2025-06-02
View multiple logs, Terminal tabs, App tabs?


illume:
Being able to open a terminal to more than one thing at a time would be useful.

Especially in the App where it's not possible to do another browser tab. (or is it possible without Headless?)

## Describe the solution you'd like

To be able to connect to different things with multiple terminals.

Either:
- Multiple tabs in the app
- Multiple terminal tabs
- other?

## What users will benefit from this feature?

Anyone who wants to look at multiple things at once. For example with multiple services making up a bigger service (microservices or backend/frontend separation for example). It's very common to want to look at multiple pods or services to debug issues.

## Additional context

We've talked about this quite a few times... but I didn't see an issue.

A recent discussion about it is here: 
- https://github.com/kubernetes-sigs/headlamp/discussions/3237

tuminzee:
this is very cool idea

UnasZole:
I'm very interested in this feature as well.

I just installed Headlamp as an attempt to move away from OpenLens which is no longer maintained, and while the experience with Headlamp is pretty good so far, this is a big missing feature for me.

Typical usage can be : 
- Opening the logs of several pods at the same time to watch them and compare them on the fly.
- Opening several terminals to execute sequences of commands across different pods in parallel.

For this, a system of tabs, or of floating windows (or ideally both !) would be great.

joaquimrocha:
This is fixed now with the activities feature AFAIU. Closing.

cschar:
> This is fixed now with the activities feature AFAIU. Closing.

where is this documented ?
<img width="1494" height="567" alt="Image" src="https://github.com/user-attachments/assets/b8c924fc-2871-4b78-bf6c-6f8ac37a84b4" />

here? https://headlamp.dev/blog/2025/11/13/headlamp-in-2025#navigation--activities

is there a video showing viewing 3 pods logs at the same time?

PR #3583, opened 2025-07-07
frontend: Implement new Activity feature for tabs and windows

### LINKED ISSUE #3486: Manage terminals/logs/tasks outside of a details view

As described a long time in https://github.com/kubernetes-sigs/headlamp/issues/523 , we need to make sure that shells/logs do not depend on the details view of the item they refer to. i.e. when opening a shell dialog from a pod's details view, we should be able to minimize/close the dialog, navigate to a different view, and be able to bring back the shell dialog.

We have design for a task bar which would allow us to manage the shell and logs windows, keeping them around when needed:

Mockup showing a task bar

Users who need to keep several logs or shell windows opened at the same time, while navigating to different views.
See: https://github.com/kubernetes-sigs/headlamp/issues/3419

## Additional context

Instead of making this very related to the logs/terminal, I think we should have a notion of Tasks in Headlamp. Tasks would be processes that need to be around (connected logs stream, shell exec, external command execution, etc.) and a view for them. This way we can also add a way to visualize long running tasks like e.g. deploying a minikube cluster, etc.

sniok:
This PR adds new Activity concept, allowing users to manage terminals, logs, and tasks outside of a details view.
It's a generic system for opening new tabs/window, resizing them, minimizing.
This replaces drawer mode.

You create a new activity by calling 

```tsx
Activity.launch({
  id: "some-id",
  content: <div>content</div>,
  location: "split-right",
});
```

All activities are stored in the activitySlice redux store.

Settings were updated 

image

## Related Issue

Fixes #3486 #3419

## Steps to Test

1. Open resource details
2. Open terminals
3. Open logs
4. Open Edit dialogs
5. Open Create new resource dialog

## Testing done

1. Manual testing
2. a11y Axe check
3. Fastpass a11y check


frontend/src/components/common/Resource/CreateButton.tsx
```diff
@@ -53,12 +53,57 @@ export default function CreateButton(props: CreateButtonProps) {
     }
   }, [clusters]);
 
+  const openActivity = () => {
+    Activity.launch({
+      id: 'create-button',
```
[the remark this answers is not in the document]
sniok: that's intentional, we only want one instance of the create button dialog

joaquimrocha:
Oops, clicked submit too early. One sec.

frontend/src/redux/reducers/reducers.tsx
```diff
@@ -15,6 +15,7 @@
  */
 
 import { combineReducers } from 'redux';
+import { activityReducer } from '../../components/activity/Activity';
```
joaquimrocha, line 18: Should this commit be squashed to the one introducing the reducer?

frontend/src/components/common/Resource/EditButton.tsx
```diff
@@ -42,12 +44,12 @@ interface EditButtonProps {
 export default function EditButton(props: EditButtonProps) {
   const dispatch: AppDispatch = useDispatch();
   const { item, options = {}, buttonStyle, afterConfirm } = props;
-  const [openDialog, setOpenDialog] = React.useState(false);
   const [isReadOnly, setIsReadOnly] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState<string>('');
   const location = useLocation();
   const { t } = useTranslation(['translation', 'resource']);
   const dispatchHeadlampEditEvent = useEventCallback(HeadlampEventType.EDIT_RESOURCE);
+  const activityId = 'edit-' + item.metadata.uid;
```
joaquimrocha, line 52: Is there a possibility that the item is updated here? Asking because the UID will change if someone modifies the item or the item is re-created.

frontend/src/components/common/Resource/EditButton.tsx
```diff
@@ -122,25 +125,32 @@ export default function EditButton(props: EditButtonProps) {
         description={t('translation|Edit')}
         buttonStyle={buttonStyle}
         onClick={() => {
-          setOpenDialog(true);
+          Activity.launch({
+            id: 'edit-' + item.metadata.uid,
+            title: 'Edit: ' + item.metadata.name,
```
joaquimrocha: Needs translation?

frontend/src/components/common/Resource/LogsButton.tsx
```diff
@@ -156,6 +158,31 @@ export function LogsButton({ item }: LogsButtonProps) {
             autoHideDuration: 3000,
           });
         }
+
+        Activity.launch({
+          id: 'logs-' + item.metadata.uid,
```
joaquimrocha, line 163: Same comment about the UID for ids.

frontend/src/components/pod/Details.tsx
```diff
@@ -504,7 +503,15 @@ export default function PodDetails(props: PodDetailsProps) {
                   aria-label={t('logs')}
                   icon="mdi:file-document-box-outline"
                   onClick={() => {
-                    setShowLogs(true);
+                    Activity.launch({
+                      id: 'logs-' + item.metadata.uid,
+                      title: 'Logs: ' + item.metadata.name,
+                      icon: (
+                        <Icon icon="mdi:file-document-box-outline" width="100%" height="100%" />
+                      ),
```
joaquimrocha, line 601: If the icon could also be a string, we could check for the type and make the component automatically in the back. For readability.

frontend/src/components/pod/Details.tsx
```diff
@@ -504,7 +503,15 @@ export default function PodDetails(props: PodDetailsProps) {
                   aria-label={t('logs')}
                   icon="mdi:file-document-box-outline"
                   onClick={() => {
-                    setShowLogs(true);
+                    Activity.launch({
+                      id: 'logs-' + item.metadata.uid,
+                      title: 'Logs: ' + item.metadata.name,
```
joaquimrocha: Needs to be translatable?

frontend/src/components/pod/Details.tsx
```diff
@@ -519,48 +526,58 @@ export default function PodDetails(props: PodDetailsProps) {
...
-                </AuthVisible>
+              <AuthVisible item={item} authVerb="create" subresource="exec">
+                <ActionButton
+                  description={t('Terminal / Exec')}
+                  aria-label={t('terminal')}
+                  icon="mdi:console"
+                  onClick={() => {
+                    Activity.launch({
+                      id: 'terminal-' + item.metadata.uid,
+                      title: '' + item.metadata.name,
```
joaquimrocha: ?

frontend/src/components/pod/Details.tsx
```diff
@@ -519,48 +526,58 @@ export default function PodDetails(props: PodDetailsProps) {
...
-                          resource: item,
-                          status: EventStatus.OPENED,
-                        },
-                      });
-                    }}
-                  />
-                </AuthVisible>
+                <ActionButton
+                  description={t('Attach')}
+                  aria-label={t('attach')}
```
joaquimrocha: I know this was there before, but I think this is automatically added:
https://github.com/kubernetes-sigs/headlamp/blob/9b3b0519a04a777c1ee6594b644da49b5e48e9e3/frontend/src/components/common/ActionButton/ActionButton.tsx#L81

yolossn:
Hey, this looks great, really intuitive and smooth overall.I wanted to share a few issues and thoughts I encountered during testing:

It will be good have an option to close the tabs from the overview 
image

I noticed a few rendering issues when multiple tabs were open:

1. When switching between two tabs the activity bar moved up and down based on the tab that was selected

https://github.com/user-attachments/assets/9417a352-8d28-4fe9-ac67-29c25d48c7e7

2. When multiple logs, exec and detail tabs were open the logs on the right side started bleeding out of the tab.

https://github.com/user-attachments/assets/ca83ab67-bcb6-4a96-9376-504a29adce28

The tabs are missing Cluster contexts

1. When tabs from multiple clusters are open the tabs dont display which cluster they belong to.

https://github.com/user-attachments/assets/9eecb45a-0a5f-45c7-869b-a2217886bdab

2. Also, if a user is viewing two tabs from different clusters (say A and B) and is currently in the context of cluster B, all links, even from the tab that belongs to cluster A, direct to cluster B.


https://github.com/user-attachments/assets/7c748f11-f115-42d7-8a5e-6419cfd967d0

joaquimrocha:
Wow! Great findings @yolossn ! Thank you!

frontend/src/components/common/Resource/EditButton.tsx
sniok: uid stays the same when you update a resource. if it's recreated then uid will be different and it be a different activity

frontend/src/components/pod/Details.tsx
sniok: I think ReactNode type here is better, allows for more flexibility, for example putting a spinner component as an icon

frontend/src/components/pod/Details.tsx
sniok: oops

sniok:
Updates since the last review:

> The Node Shell doesn't use the windowed rendering

Fixed

> In the map view, the taskbar covers the map zoom lower side buttons

Fixed

> I wonder how discoverable the snapping left/right is. Would it work if we added snap-left/snap-right/maximize buttons to the right of the drag-::: handle? Leaving only the minizime + close button to the right? (maybe it looks bad, just throwing out ideas)

There are now buttons to snap to left/right, and on small screen the :::: handle is disabled (so dragging and popup window mode are disabled) and activities are opened in full screen. 

> When switching between two tabs the activity bar moved up and down based on the tab that was selected

Fixed this by simplifying minimization animation

> When multiple logs, exec and detail tabs were open the logs on the right side started bleeding out of the tab.

There's some improvements we can do to the log viewer that can be addressed in a different PR

> When tabs from multiple clusters are open the tabs don't display which cluster they belong to.

Currently we only show cluster field when more than one cluster is selected, maybe we should just always show cluster field. I'm not sure how to solve this properly right now but this can be addressed in future PRs

> Also, if a user is viewing two tabs from different clusters (say A and B) and is currently in the context of cluster B, all links, even from the tab that belongs to cluster A, direct to cluster B. 

I've updated those links so that it now has proper cluster parameter

joaquimrocha:
Did some new testing. Here are my impressions:
1. Can you remove the Overview label? Just leave the icon + tooltip
2. I think that going from maximize button to unmaximize should go back to the previous state that the window was before, so if snapped-right > maximize > unmaximize = snapped-right. Yes, I do want the free window thing to be more difficult to discover as I am not sure of its utility.
3. Make the task bar a bit narrower, it has unnecessarily big vertical padding. // **UPDATE:** Maybe we keep the big tabs as they are but we add the cluster as a pill/label under the name of the task.
4. Overview: I think it's nice to see all opened tasks, to quickly locate one but I think the current preview looks bad. Maybe it could just show a more prominent icon + name? If we lose the contents, I don't think it's a big deal because it's not super useful now: I can click the actions icons from the preview, that's a bit weird. // **UPDATE:** I think maybe even just having the Icon + Kind + Name showing without the windows preview is better than now.

PR #3587, opened 2025-07-07
frontend: Chart: Fix UI bugs

skoeva:
These changes do the following:

- Adjust the colors of the bar chart so the bar is more visible in dark mode
- Render tooltips in a portal to prevent them from being clipped by the table row

### Testing
- [X] Navigate to the Nodes page in a cluster
- [X] Hover over the CPU and memory bars for a node
- [X] Notice that the bars are easily visible and tooltips aren't clipped by the table

image


sniok (approved):
nice change, thanks!

PR #3590, opened 2025-07-08
    docs: aks-cluster-oauth: enhance with clearer steps and helm-based deployment

yolossn:
A user reported that the existing documentation was minimal and didn't work for them. This update provides clearer, more in-depth step-by-step instructions and replaces custom configurations with proper helm chart deployments for better maintainability.


illume (approved):
🎉 thanks!

PR #3593, opened 2025-07-08
fix the kind at cr creation template

### LINKED ISSUE #3591: [BUG]: Wrong `kind` Value Generated When Creating Custom Resource
## Describe the bug

When creating a Custom Resource (CR) from the **Create** page in Headlamp, the auto-generated example/template incorrectly sets the `kind` field to the **singular name** of the resource, instead of the **actual kind** as defined in the CustomResourceDefinition (CRD). This results in failed or invalid resource creation.

###  Expected behavior

The `kind` field in the CR creation template should match the `spec.names.kind` from the CRD (e.g., ``Book), not the singular name (`book`).

---

## To Reproduce

Steps to reproduce the bug:

1. Go to **Custom Resources → CRDs** in Headlamp.
2. Select any CRD with a defined `kind` (e.g., `MyCustomResource`).
3. Click on **Create** to add a new CR.
4. Observe the `kind` field in the YAML template—it uses the singular form instead of the expected `kind`.
5. Attempting to apply this CR results in a validation or creation error.

---

## Environment

- **Installation type**: local-setup
- **Headlamp Version**: [e.g., 1.32.2]

---

Yes, I will propose a PR.

ChayanDass:
## Summary

This PR fixes an issue with the auto-generated `kind` field in the Custom Resource (CR) creation template.

## Related Issue

Fixes #3591

## Changes

* Previously, the `kind` value in the generated CR template was being set using the `singular` field from the CRD.
* This behavior has been corrected to use the actual `kind` defined in the CRD.

## Steps to Test

1. Create a CustomResourceDefinition (CRD) where the `kind` and `singular` fields are different.
2. Open the CR creation form for that CRD.
3. Verify that the `kind` value in the YAML template uses the `kind` field from the CRD, not the `singular` one.

## Screenshots

**Before:**

```yaml
apiVersion: school.com/v1
kind: student  # Incorrect: used the 'singular' field
metadata:
  name: ''
```

**After:**

```yaml
apiVersion: school.com/v1
kind: Student  # Correct: uses the actual 'kind' field
metadata:
  name: ''
```

### DIFF
--- a/frontend/src/components/crd/__snapshots__/CustomResourceDefinition.Details.stories.storyshot
+++ b/frontend/src/components/crd/__snapshots__/CustomResourceDefinition.Details.stories.storyshot
@@ -488,7 +488,7 @@
                     class="MuiBox-root css-ldp2l3"
                   >
                     <button
-                      aria-label="Create mycustomresource"
+                      aria-label="Create MyCustomResource"
                       class="MuiButtonBase-root MuiIconButton-root MuiIconButton-sizeMedium css-whz9ym-MuiButtonBase-root-MuiIconButton-root"
                       data-mui-internal-clone-element="true"
                       tabindex="0"
--- a/frontend/src/components/crd/__snapshots__/CustomResourceList.List.stories.storyshot
+++ b/frontend/src/components/crd/__snapshots__/CustomResourceList.List.stories.storyshot
@@ -39,7 +39,7 @@
               class="MuiBox-root css-ldp2l3"
             >
               <button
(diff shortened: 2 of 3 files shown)


sniok:
thanks for opening PR! could you please update your commits to follow [our guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines)?

ChayanDass:
done!

ChayanDass:
@skoeva    done! could you check you ?

PR #3594, opened 2025-07-08
frontend: Cookie based token storage

### LINKED ISSUE #3562: Store tokens in the backend
Currently authentication tokens are stored in the localStorage on the frontend
Storing tokens in a cookie is not an option since there are users with tokens that more than 4KB in size (cookie size limit).

Instead we should store the token on the backend in the filesystem and store session cookie on the frontend. We can leverage https://github.com/gorilla/sessions package for that. This will require a big rewrite of how tokens and auth is handled on the frontend and on the backend

Some test cases to consider:
 - Default token flow where user pastes a token created by kubectl
 - OIDC auth flow
 - Electron app flow (backend that restarts frequently)
 - Websocket Multiplexer has its own token flow

sniok:
## Summary

Fixes #3562

Removes token storage from localStorage

Instead tokens are now stored in cookies.
There's a new endpoint in the backend /clusters/CLUSTER/set-token that in the response will set a secure, http-only cookie, that won't be available in js. 
Also OIDC logic was updated to directly set the token in the cookie instead of passing it in the callback URL, which eliminates any frontend code for token handling.
Websocket multiplexer will use token from the cookie as well, removing the need for sending a token in the connect message.
For tokens that exceed cookie size limit (4KB) they will be split across multiple cookies.

For user provided token the flow used to be:
User pastes token -> Token is saved in localstorage -> Token is added in Authorization Bearer header

But now it will be:
User pastes token -> Token is set into a cookie via /set-token endpoint -> Cookies are included in each request for that cluster

## Consequences of this change

 - Frontend doesn't know if token is set at all, so log out button is now always active. Without a token clicking on logout will just redirect to home page.
 - Plugins can no longer access tokens, getToken and setToken will only work if plugins provide their own overridden functions.

## Testing done

Local cluster, AKS cluster
OIDC configured cluster
Token with >4kb length
Built app, tested port forwarding


frontend/src/plugin/__snapshots__/pluginLib.snapshot
```diff
@@ -16100,10 +16100,6 @@
   "Utils": {
     "CLUSTER_ACTION_GRACE_PERIOD": 5000,
     "auth": {
-      "deleteTokens": [Function],
-      "getToken": [Function],
```
illume: Some plugins use token related functions.

backend/pkg/auth/cookies.go
```diff
@@ -0,0 +1,189 @@
...
+	// if token is larger than maxCookieSize, split it into multiple cookies
+	chunks := splitToken(token, chunkSize)
+	for i, chunk := range chunks {
+		cookie := &http.Cookie{
+			Name:     fmt.Sprintf("headlamp-auth-%s.%d", sanitizedCluster, i),
+			Value:    chunk,
+			HttpOnly: true,
+			Secure:   secure,
+			SameSite: http.SameSiteStrictMode,
+			Path:     "/",
```
illume: The Path should be per cluster right? So the cookies only go to that cluster and not other ones, and not other services.

frontend/src/plugin/__snapshots__/pluginLib.snapshot
sniok: they'll keep working, but only if plugins provide their own get/setToken functions via registerGetTokenFunction, registerSetTokenFunction

sniok: otherwise, it'll stop working since js can't access tokens anymore, which is the point

frontend/src/lib/k8s/api/v2/fetch.ts
```diff
@@ -82,23 +81,13 @@ export async function clusterFetch(url: string | URL, init: RequestInit & { clus
 
   // Refresh service account token only if the cluster auth type is not OIDC
   if (getClusterAuthType(init.cluster) !== 'oidc') {
-    await refreshToken(token);
-  }
-
-  if (token) {
-    init.headers.set('Authorization', `Bearer ${token}`);
+    // await refreshToken(token);
```
joaquimrocha: Why is this commented out?

frontend/src/lib/k8s/api/v1/tokenApi.ts
```diff
@@ -99,20 +98,15 @@ export async function refreshToken(token: string | null): Promise<void> {
...
       method: 'POST',
       headers,
       body: JSON.stringify(tokenData),
     });
 
     if (response.status === 201) {
-      const token: KubeToken = await response.json();
-      setToken(cluster, token.status.token);
+      // const token: KubeToken = await response.json();
+      // setToken(cluster, token.status.token);
```
joaquimrocha: Why is this commented out?

joaquimrocha (commented):
Complex PR, but Github Copilot helped me understand it. I just left some comments about the commented out code. Maybe adding a new comment next to it explaining the code is kept commented out to make it clearer when we get to implement token refreshing would be good.

frontend/src/lib/k8s/api/v2/fetch.ts
sniok: refreshing token from the frontend is no longer possible since we don't have access to it, so this should've just been removed instead of commented out

frontend/src/lib/k8s/api/v1/tokenApi.ts
sniok: same reason

frontend/src/lib/k8s/api/v2/fetch.ts
sniok: instead of leaving a comment in the frontend code I've opened an issue to keep track of this 
https://github.com/kubernetes-sigs/headlamp/issues/3763

PR #3602, opened 2025-07-09
backend: headlampConfig: move headlampConfig and import in modules

iamundefinedguy:
## Summary

I have moved some propoerties to new package called as ``headlampConfig`` were we can import different properties of HeadlampConfig struct in various package also.  Some properties related to ``multiplexer`` ,  ``cache`` and ``oidc`` still in headlampConfig.


illume:
> Some properties related to multiplexer , cache and oidc still in headlampConfig.

I wonder why these ones were left here? What's the plan with these? Is it to leave them there, or eventually they should be moved somewhere too?

iamundefinedguy:
> > Some properties related to multiplexer , cache and oidc still in headlampConfig.
> 
> I wonder why these ones were left here? What's the plan with these? Is it to leave them there, or eventually they should be moved somewhere too?

Because these are the one are mostly used in this ``cmd/package`` ( headlamp , server , stateless etc ) so moving to different package might require changing importing in all modules. 

Also I wasn't sure that making properties exportable accross the application would be good option. 

I will move all the properties to ``pkg/headlampConfig`` if this would is the correct implementation to make ``HeadlampConfig`` exportable.

illume:
> > > Some properties related to multiplexer , cache and oidc still in headlampConfig.
> > 
> > 
> > I wonder why these ones were left here? What's the plan with these? Is it to leave them there, or eventually they should be moved somewhere too?
> 
> Because these are the one are mostly used in this `cmd/package` ( headlamp , server , stateless etc ) so moving to different package might require changing importing in all modules.
> 
> Also I wasn't sure that making properties exportable accross the application would be good option.
> 

Ok, cool. I think those can be moved separately later then (if needed).

illume (approved):
👍 thanks

iamundefinedguy:
@illume i think we need to moved ``HeadlampConfig`` to this package. but i am not sure whether it is fine to make methods exportable.

PR #3603, opened 2025-07-09
ci: Fix credentials for nightly build workflow

sniok:
Uses automatically generated GITHUB_TOKEN secret for the workflow. Tested with a manual trigger [here](https://github.com/sniok/headlamp-o/actions/runs/16138765486).

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
-          password: ${{ secrets.GHCR_IMAGES }}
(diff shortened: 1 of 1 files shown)


joaquimrocha:
TBH, I think the other secret should work but maybe the key has expired. I can test this later.

joaquimrocha:
Also, I don't think the GITHUB_TOKEN will work in this case because we are pushing to a different org's package, which is something that the mentioned token has a limitation on IIRC.

joaquimrocha:
I have updated the token and pushed again. It worked.
https://github.com/kubernetes-sigs/headlamp/actions/runs/16200214670/job/45737048559
https://github.com/orgs/headlamp-k8s/packages/container/headlamp/458672010?tag=nightly

Issue #3606, opened 2025-07-09
backend: Add default service account token authentication option for in-cluster deployments


yolossn:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
Currently, when deploying headlamp in-cluster the users have to create a service account token and paste in the frontend. But when users want to use an OIDC proxy infront of headlamp for authentication this step extra pasting  service account token is redundant. At the same time making incluster authenticated by default can lead to security issues where users have headlamp open to internet without proper authentication.

## Describe the solution you'd like

Add a configuration option that allows users to explicitly opt-in to using Kubernetes service account tokens for authentication when running Headlamp in-cluster. This provides:

1. New configuration flags:
- `--use-service-account-token`: Enable service account token authentication (opt-in)
- `--service-account-token-path`: Customize the path to the service account token (defaults to /var/run/secrets/kubernetes.io/serviceaccount/token)
2. Security-first approach:
- Users must explicitly opt-in to use service account authentication
- This prevents accidental exposure of Headlamp without proper authentication

This enables users to deploy OAuth2Proxy in front of Headlamp for OIDC based authentication.

## What users will benefit from this feature?
In-cluster users who want to use OIDC based authentication when the kubernetes cluster itself doesn't have OIDC authentication.

## Are you able to implement this feature?
Yes

> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context
Related to https://github.com/kubernetes-sigs/headlamp/issues/2614

sstarcher:
This blocks our usage of headlamp where we could use a common gateway api that uses it's own auth and offload that from headlamp.

PR #3607, opened 2025-07-09
Add support for optin service account auth for incluster deployments.

yolossn:
The backend now supports two flags that can be used to make the incluster deployment authenticated by default, this allows the users to use OIDC proxies in front of headlamp without having to deal with service account token. For more details refer the issue

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/3606

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -183,6 +183,12 @@ spec:
             {{- with .Values.config.pluginsDir}}
             - "-plugins-dir={{ . }}"
             {{- end }}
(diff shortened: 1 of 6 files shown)


jcpunk:
This would be helpful for my workflow.

andyjhall:
@yolossn do you plan on fixing the conflicts or would you like help doing so? On the face of it they look relatively simple to fix.

Mik4sa:
I also would like to have this. This meets my planed workflow perfectly :)

mhwtx:
+1 - this would be great

Issue #3337, opened 2025-05-23
Support/verify plugins' signatures


joaquimrocha:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

To increase security, we should promote plugin signing and implement signature verification for the user to understand whether a plugin payload is what's intended.

## Describe the solution you'd like

TBD

## What users will benefit from this feature?

Extra assurance/security.

illume:
@gambtho can you please /assign yourself for issue refinement?

illume:
This one doesn't have an estimate.

PR #3608, opened 2025-07-10
permissions for runCmd

illume:
How to give each plugin permission secrets for runCmd (and in the future other Electron calls). 

Additionally this:
- checks consent with arguments not just the command. Eg, it checks "minikube delete" not just "minikube".
- validates more input
- allows permitted plugins to run scripts

### Demo

- The minikube plugin with "Run" button has permissions to runCmd "manage-minikube"
- The pod-counter plugin with "Uh, Pods" button is denied access to runCmd minikube status. (It is not given permission secrets to do so).

https://github.com/user-attachments/assets/bfa7f226-3448-4420-8229-1191c9f725ed

### Simplified example of the context hiding

Function can be used to eval code whilst not giving that code access to the current context.
This way permission secrets can be hidden from each plugin.

This code snippet you can run in node (or browser dev console).

```ts
function runPlugin() {
  const permissionSecrets = { 'permission': 9827598192755 };
  // try and access permissionSecrets from the plugin
  // This should fail, because the plugin is not allowed to access it.
  const sourceStr = "console.log(permissionSecrets)"; 

  // This would not fail. The plugin has access to its own pluginPermissionSecrets
  //const sourceStr = "console.log(pluginPermissionSecrets)"; 

  // This makes a function that can not access this context
  //   so it can't access permissionSecrets because it's a variable in this context.
  const executePlugin = new Function("pluginPermissionSecrets", sourceStr);

  if (executePlugin) {
    try {
      executePlugin({'permission': permissionSecrets['permission']});
    } catch (e) {
      console.log("Error during plugin execution:", e);
      console.log("Plugin tried to access hidden secrets and failed.");
    }
  }
}
runPlugin()
```

If you run the above snippet in node you get this.
```
Error during plugin execution: ReferenceError: permissionSecrets is not defined
    at eval (eval at runPlugin (REPL22:12:25), <anonymous>:3:13)
    at runPlugin (REPL22:16:7)
    at REPL23:1:1
    at ContextifyScript.runInThisContext (node:vm:137:12)
    at REPLServer.defaultEval (node:repl:593:22)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:922:10)
    at REPLServer.emit (node:events:530:35)
    at REPLServer.emit (node:domain:489:12)
Plugin tried to access hidden secrets and failed.
undefined
```

### Data flow

Here's the basic data flow.

- frontend loads
- frontend asks for permission secrets
- Electron app/ gives frontend permission secret numbers
- frontend loads each plugin giving each one it's permission secrets
- plugin sends with window.desktopApi call the secret number for that permission
- Electron app/ checks the secret number is correct, and either runs the command or not

---

### How to test

- Add minikube prerelease plugin from within the app plugin catalog. (Enable non-official plugins, uninstall any other minikube plugins first)
- Open dev console in app.
- Press the "script" button top right (see screenie below)
- See output of script (see screenie below)

image

Related to:
- https://github.com/headlamp-k8s/plugins/issues/269

### DIFF
--- a/app/electron/runCmd.ts
+++ b/app/electron/runCmd.ts
@@ -127,6 +127,76 @@ function checkCommandConsent(command: string, args: string[], mainWindow: Browse
   return true;
 }
 
+const COMMANDS_WITH_CONSENT = {
+  headlamp_minikube: [
+    'minikube start',
+    'minikube stop',
+    'minikube delete',
+    'minikube status',
+    'minikube service',
+    'minikube logs',
+    'minikube addons',
+    'minikube ssh',
+    'scriptjs headlamp_minikubeprerelease/manage-minikube.js',
+    'scriptjs headlamp_minikube/manage-minikube.js',
+    'scriptjs minikube/manage-minikube.js',
+  ],
+};
+/**
+ * Adds the runCmd consent for the plugin.
+ *
+ * This is used to give consent to the plugin to run commands when the plugin is installed.
+ * So the user is not presented with many consent requests.
+ *
+ * @param pluginInfo artifacthub plugin info
+ */
+export function addRunCmdConsent(pluginInfo: { name: string }): void {
+  const settings = loadSettings();
+  if (!settings.confirmedCommands) {
+    settings.confirmedCommands = {};
+  }
+  let commands: string[] = [];
+  const pluginIsMinikube =
+    pluginInfo.name === 'headlamp_minikube' ||
+    pluginInfo.name === 'headlamp_minikubeprerelease' ||
+    (process.env.NODE_ENV === 'development' && pluginInfo.name === 'minikube');
+
+  if (pluginIsMinikube) {
+    commands = COMMANDS_WITH_CONSENT.headlamp_minikube;
+  }
+  for (const command of commands) {
+    if (!settings.confirmedCommands[command]) {
+      settings.confirmedCommands[command] = true;
+    }
+  }
+
+  saveSettings(settings);
+}
+
+/**
+ * Adds the runCmd consent for the plugin.
+ *
+ * @param pluginName The package.json name of the plugin.
+ */
+export function removeRunCmdConsent(pluginName: string): void {
+  const settings = loadSettings();
+  if (!settings.confirmedCommands) {
+    return;
+  }
+  let commands: string[] = [];
+  if (
+    pluginName === '@headlamp-k8s/minikubeprerelease' ||
+    pluginName === '@headlamp-k8s/minikube'
+  ) {
+    commands = COMMANDS_WITH_CONSENT.headlamp_minikube;
+  }
+  for (const command of commands) {
+    delete settings.confirmedCommands[command];
+  }
+
+  saveSettings(settings);
+}
+
 /**
  * Check if the command has the correct permission secret.
  * If the command is 'scriptjs', it checks for a specific script path.
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -48,7 +48,7 @@ import {
   getPluginBinDirectories,
   PluginManager,
 } from './plugin-management';
-import { runScript, setupRunCmdHandlers } from './runCmd';
+import { addRunCmdConsent, removeRunCmdConsent, runScript, setupRunCmdHandlers } from './runCmd';
 import windowSize from './windowSize';
 
 if (process.env.HEADLAMP_RUN_SCRIPT) {
@@ -333,6 +333,8 @@ class PluginManagerEventListeners {
       controller,
     };
 
+    addRunCmdConsent(pluginInfo);
+
     PluginManager.installFromPluginPkg(
       pluginInfo,
       destinationFolder,
@@ -402,6 +404,8 @@ class PluginManagerEventListeners {
       action: 'UNINSTALL',
       progress: { type: 'info', message: 'uninstalling plugin' },
     };
+
+    removeRunCmdConsent(pluginName);
 
     PluginManager.uninstall(pluginName, destinationFolder, progress => {
       updateCache(progress);


illume:
I've changed it so that instead of sending in permissionSecrets it sends to the plugins a function which internally uses permission secrets, and it's own private copy of the send receive functions instead of using window.desktopAPI.

So now permissionSecrets is not passed to plugins:
- plugin authors will not need to know about the permissions secrets in their app.
- plugins will instead get sent the API they need which internally handles using secrets (eg runCommand).
- plugins can not access the permission secrets at all

The other change is that instead of runCommand accessing window.desktopAPI.send /.receive a reference is stored inside the function sent to the plugin (`privateSend = window.desktopAPI.send`). So now a plugin can't alter the send or receive functions that the runCommand uses to snoop on the permission secrets.

If someone uses runCommand directly mistakenly it shows them this bit nicer error.
Screenshot 2025-07-22 at 19 50 20

Thanks @joaquimrocha and @sniok

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,79 @@ export async function fetchAndExecutePlugins(
...
-              error: e,
+              pluginInfo: { name: packageName, version: packageVersion },
+              error,
             },
           })
         );
-      }
-    }).call({}, source);
+      },
+      getAllowedPermissions: (pluginName, pluginPath, secrets): Record<string, number> => {
```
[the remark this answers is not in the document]
illume: Yes, this would come in a future PR.

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,79 @@ export async function fetchAndExecutePlugins(
...
+              secrets['runCmd-scriptjs-minikube/manage-minikube.js'],
+            'runCmd-scriptjs-headlamp_minikube/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikube/manage-minikube.js'],
+            'runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js'],
+          };
+        }
+        return {};
+      },
+      getArgValues: (pluginName, pluginPath, allowedPermissions) => {
```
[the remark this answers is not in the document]
illume: Yes, this would come in a future PR.

app/electron/runCmd.ts
```diff
@@ -123,46 +159,163 @@ function checkCommandConsent(command: string, mainWindow: BrowserWindow): boolea
...
+  if (mainWindow === null) {
+    console.error('Main window is null, cannot set up run command handlers');
+    return;
+  }
+
+  // We only send the plugin permission secrets once. So any code can't just request them again.
+  // This means that if the secrets are requested before the plugins are loaded, then
+  // they will not be sent until the next time the app is reloaded.
+  let pluginPermissionSecretsSent = false;
+  const permissionSecrets = {
```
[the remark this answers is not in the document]
illume: Yes, this would come in a future PR.

frontend/src/components/App/runCommand.ts
```diff
@@ -52,38 +74,44 @@ export function runCommand(
   if (!window.desktopApi) {
     throw new Error('runCommand only works in Headlamp app mode.');
   }
+  if (!desktopApiSend || !desktopApiReceive || !permissionSecrets) {
```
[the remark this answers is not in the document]
illume: There is a message for how to use it.

app/electron/runCmd.ts
```diff
@@ -123,46 +159,163 @@ function checkCommandConsent(command: string, mainWindow: BrowserWindow): boolea
...
+}
+
+/**
+ * Runs a script, using the compiled app, or Electron in dev mode.
+ *
+ * This is needed to run the "scriptjs" commands, as a way of running
+ * node js scripts without requiring node to also be installed.
+ */
+export function runScript() {
+  import(pathToFileURL(process.argv[1]).href);
```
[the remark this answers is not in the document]
illume: This is fixed.

app/electron/runCmd.ts
```diff
@@ -123,46 +159,163 @@ function checkCommandConsent(command: string, mainWindow: BrowserWindow): boolea
...
+  // This means that if the secrets are requested before the plugins are loaded, then
+  // they will not be sent until the next time the app is reloaded.
+  let pluginPermissionSecretsSent = false;
+  const permissionSecrets = {
+    'runCmd-minikube': Math.random(),
+    // 'runCmd-scriptjs-headlamp-pod-counter/bin/manage-minikube.js': Math.random(),
+    'runCmd-scriptjs-headlamp-k8s-minikube/bin/manage-minikube.js': Math.random(),
+    'runCmd-scriptjs-minikube/manage-minikube.js': Math.random(),
+    'runCmd-scriptjs-headlamp_minikube/manage-minikube.js': Math.random(),
+    'runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js': Math.random(),
```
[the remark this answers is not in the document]
illume: this is fixed

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,82 @@ export async function fetchAndExecutePlugins(
...
+          return {
+            'runCmd-minikube': secrets['runCmd-minikube'],
+            'runCmd-scriptjs-minikube/manage-minikube.js':
+              secrets['runCmd-scriptjs-minikube/manage-minikube.js'],
+            'runCmd-scriptjs-headlamp_minikube/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikube/manage-minikube.js'],
+            'runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js'],
+          };
+        }
```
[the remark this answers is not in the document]
illume: Yes, this will be moved in a separate PR.

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,82 @@ export async function fetchAndExecutePlugins(
...
+        return {};
+      },
+      getArgValues: (pluginName, pluginPath, allowedPermissions) => {
```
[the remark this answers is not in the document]
illume: It's more clear duplicated.

app/electron/runCmd.ts
```diff
@@ -123,46 +160,192 @@ function checkCommandConsent(command: string, mainWindow: BrowserWindow): boolea
...
+  if (mainWindow === null) {
+    console.error('Main window is null, cannot set up run command handlers');
+    return;
+  }
+
+  // We only send the plugin permission secrets once. So any code can't just request them again.
+  // This means that if the secrets are requested before the plugins are loaded, then
+  // they will not be sent until the next time the app is reloaded.
+  let pluginPermissionSecretsSent = false;
+  const permissionSecrets = {
```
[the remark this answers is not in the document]
illume: Yes, in a separate PR.

frontend/src/plugin/index.ts
```diff
@@ -371,3 +430,83 @@ export async function fetchAndExecutePlugins(
...
+  ) => Record<string, number>;
+  getArgValues: (
+    pluginName: string,
+    pluginPath: string,
+    allowedPermissions: Record<string, number>
+  ) => [string[], unknown[]];
+}) {
+  if (!pluginPath || !packageName || !packageVersion) {
+    console.error(`Either pluginPath, packageName or packageVersion is missing for ${pluginPath}`);
+    return;
```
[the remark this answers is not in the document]
illume: It's fine logging an error. The exception would just be caught and logged later anyway because runPlugin is wrapped in a try catch log.

illume:
Changed:
- valid path checks for script argument, so it has to be within plugin folder
- use crypto.webcrypto instead of Math.random
- rebased against main

To make it easier to test, I've published a minikube pre release (a separate plugin found in the non-official toggle of plugin catalog). Added testing instructions to the PR description.

illume:
Added checking of consent with arguments not just the command. Eg, it checks consent for "minikube delete" not just "minikube".

Also cleaned up some of the commits.

app/electron/runCmd.ts
```diff
@@ -231,6 +242,27 @@ export function setupRunCmdHandlers(mainWindow: BrowserWindow | null, ipcMain: E
...
   }
+
+  // We only send the plugin permission secrets once. So any code can't just request them again.
+  // This means that if the secrets are requested before the plugins are loaded, then
+  // they will not be sent until the next time the app is reloaded.
+  let pluginPermissionSecretsSent = false;
+  const permissionSecrets = {
+    'runCmd-minikube': cryptoRandom(),
+    // 'runCmd-scriptjs-headlamp-pod-counter/bin/manage-minikube.js': cryptoRandom(),
+    'runCmd-scriptjs-headlamp-k8s-minikube/bin/manage-minikube.js': cryptoRandom(),
```
illume: remove

app/electron/runCmd.ts
```diff
@@ -231,6 +242,27 @@ export function setupRunCmdHandlers(mainWindow: BrowserWindow | null, ipcMain: E
...
+    'runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js': cryptoRandom(),
+  };
+
+  ipcMain.on('request-plugin-permission-secrets', () => {
+    if (!pluginPermissionSecretsSent) {
+      pluginPermissionSecretsSent = true;
+      mainWindow?.webContents.send('plugin-permission-secrets', permissionSecrets);
+    }
+  });
+
```
illume, line 373: Add code for when main frame of electron has been reloaded.

frontend/src/components/App/runCommand.ts
```diff
@@ -24,11 +24,18 @@
  * This function uses the desktopApi.send and desktopApi.receive methods to communicate with the main process.
  * @param command - The command to run.
  * @param args - An array of arguments to pass to the command.
+ * @param options - Additional options for the command.
```
illume, line 27: Improve commit message.

frontend/src/components/App/runCommand.ts
```diff
@@ -41,9 +48,24 @@
  * ```
  */
 export function runCommand(
-  command: 'minikube' | 'az',
+  command: 'minikube' | 'az' | 'scriptjs',
   args: string[],
-  options: {}
+  options: {},
+  permissionSecrets?: Record<string, number>,
```
illume, line 54: consider adding these three args as an option

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,82 @@ export async function fetchAndExecutePlugins(
   );
   onSettingsChange(packagesIncompatibleSet);
 
+  const pluginDesktopApiSend = window.desktopApi.send;
```
illume: Leave a comment here about, "...this could get modified later on..."

app/electron/runCmd.ts
```diff
@@ -96,24 +96,35 @@ function saveSettings(settings: Record<string, any>) {
...
  * @param command - The command to check.
+ * @param args - The arguments to the command.
  * @returns true if the user has consented to running the command, false otherwise.
  */
-function checkCommandConsent(command: string, mainWindow: BrowserWindow): boolean {
+function checkCommandConsent(command: string, args: string[], mainWindow: BrowserWindow): boolean {
   const settings = loadSettings();
   const confirmedCommands = settings?.confirmedCommands;
+
+  console.log(JSON.stringify(settings, null, 2));
```
illume: remove

app/electron/runCmd.ts
illume: done

app/electron/runCmd.ts
illume: done

frontend/src/components/App/runCommand.ts
illume: improved the commit message, and added an extra comment about desktopApiSend.

frontend/src/components/App/runCommand.ts
illume: We already have an options object, and the user is given a runCommand wrapper to use which only uses the command, args, and options params as outlined in the examples.

So the extra arguments will stay as is.

frontend/src/plugin/index.ts
illume: this is done

illume:
@joaquimrocha I addressed the notes we left in yesterday's review.

Also, we didn't make a note about this, but we discussed it... 

I added that thing where it gives consent to running commands at install time. So there's only the consent dialog presented to the user when they install the plugin. Not as previously where it was at install time, and when trying to use it for the first time. This means, by default the commands like "minikube status" and other commands used by the plugin are given consent at install time. But a hypothetical and unexpected "minikube dobad" command would still require consent from the user to run.

app/electron/runCmd.ts
```diff
@@ -130,6 +131,47 @@ function checkCommandConsent(command: string, args: string[], mainWindow: Browse
...
+      'minikube ssh',
+      'scriptjs headlamp_minikubeprerelease/manage-minikube.js',
+      'scriptjs headlamp_minikube/manage-minikube.js',
+      'scriptjs minikube/manage-minikube.js',
+    ];
+  }
+  for (const command of commands) {
+    if (!settings.confirmedCommands[command]) {
+      settings.confirmedCommands[command] = true;
+    }
```
joaquimrocha, line 170: Shouldn't this be per plugin?

```js
{
  consentedCommands: {
    minikube: {
      commands: [
        "minikube logs",
        "..."
      ]
    }
  }
}
```

frontend/src/plugin/index.ts
```diff
@@ -329,30 +336,85 @@ export async function fetchAndExecutePlugins(
...
+            'runCmd-scriptjs-headlamp_minikube/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikube/manage-minikube.js'],
+            'runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js':
+              secrets['runCmd-scriptjs-headlamp_minikubeprerelease/manage-minikube.js'],
+          };
+        }
+        return {};
+      },
+      getArgValues: (pluginName, pluginPath, allowedPermissions) => {
+        if (
```
illume: explain where allowedPermissions comes from in comment

app/electron/runCmd.ts
```diff
@@ -130,6 +131,75 @@ function checkCommandConsent(command: string, args: string[], mainWindow: Browse
...
+ * @param pluginInfo artifacthub plugin info
+ */
+export function addRunCmdConsent(pluginInfo: { name: string }): void {
+  const settings = loadSettings();
+  if (!settings.confirmedCommands) {
+    settings.confirmedCommands = {};
+  }
+  let commands: string[] = [];
+  if (
+    pluginInfo.name === 'headlamp_minikube' ||
```
illume, line 160: consider npm start name

sniok:
a plugin that runs before minikube will be able to run commands using the following snippet 

by overriding Function constructor different plugins can intercept arguments passed into plugin or even return completely different code to execute

```tsx
const original = Function;

Function = function(...args) {
  const func = new original(...args);

  return (...params) => {
    console.log('got arguments from other plugins', params)

    if(typeof params[0] === "function" && params[0].name === "pluginRunCommand") {
        // can run command as a minikube plugin
        params[0]('minikube', ['delete'], {})
    }
    return func(...params)
  } 
};

Function.prototype = original.prototype;
```

--- 

Or like this 

```tsx
const realArrayIterator = Array.prototype[Symbol.iterator];
Array.prototype[Symbol.iterator] = function* () {
  if(this[0]?.[0] === 'pluginRunCommand') {
    console.log('Found pluginRunCommand function');
    console.log(this[0]?.[0])
  }
  yield* realArrayIterator.call(this);
};
```

since arguments are destructured like this in src/plugin/index.ts:499
```js
const [args, values] = getArgValues(packageName, pluginPath, allowedPermissions);
```

sniok:
There should be a way to change your choice on the allow/deny question. I've missclicked "Deny" and then couldn't change this later and had to delete settings.json

app/electron/runCmd.ts
illume: Added the npm start one

frontend/src/plugin/index.ts
illume: This is done.

illume:
Thanks @yolossn I addressed those convos we made during our meeting.

illume:
> a plugin that runs before minikube will be able to run commands using the following snippet

Thanks @sniok I addressed those and made some tests to show they don't work anymore.

joaquimrocha:
@illume What should we do about @sniok 's findings? I do not think we can allow a possible abuse if we tell users that they're enabling running commands only for one plugin and then maybe another abuses it.

illume:
> @illume What should we do about sniok 's findings? I do not think we can allow a possible abuse if we tell users that they're enabling running commands only for one plugin and then maybe another abuses it.

@joaquimrocha See the comment above yours...

> Thanks sniok I addressed those and made some tests to show they don't work anymore.

joaquimrocha:
> @joaquimrocha See the comment above yours...
> 
Ah, I had read them but didn't associate with the exploiting ones.
I will let @sniok re-review it.

illume:
> BTW, I think we need to fix something in the installation of plugins: right now I believe that a plugin can call itself "z-hello-world" in their ArtifactHub file, but then in the package.json they may call themselves "minikube", and if they ship with the scripts they want, this mechanism would allow them to abuse the system.
> To avoid this, we need to make sure that we somehow force a saving of the package.json with the plugin name from artifacthub. Maybe this is easily done by the electron layer just parsing the package.json and adding by itself new name-value headlamp.plugin-name or something similar, then we should use this name for the rest of the plugin related ops. Either this or verifying/enforcing that package.json's current name value matches the artifacthub one (but this would likely break existing plugins).

The package name from artifact hub is used as the folder name, which is unique. So no other plugin installed from artifact hub can be called headlamp_minikube or headlamp_minikubeprerelease.

I've removed "minikube" except when in development mode.

I think we can improve this area with another layer separately of this PR with the signature/integrity work that we need to do.
- https://github.com/kubernetes-sigs/headlamp/issues/3337

The unique artifacthub name does stop this issue you wrote about though.

app/electron/runCmd.ts
illume: hrmm. Been thinking about this for a while, and I can't think of a good reason to do this... Can you please expand on what you were thinking?

joaquimrocha:
> The package name from artifact hub is used as the folder name, which is unique. So no other plugin installed from artifact hub can be called headlamp_minikube or headlamp_minikubeprerelease.
> I've removed "minikube" except when in development mode.

Are we using the ArtifactHub/folder name when we ask the user that plugin X wants to run command CMD? If not, a plugin can still pretend to be another plugin the user trusts.

illume:
> Are we using the ArtifactHub/folder name when we ask the user that plugin X wants to run command CMD? If not, a plugin can still pretend to be another plugin the user trusts.

There is no plugin name or path presented to the user when they are asked for consent to run a command. 

At install time of the minikube plugin the consent is saved.

If no consent has been given (because they didn't install the minikube plugin for example) and evilPlugin tries to run a command there is no plugin name given when they are asked for consent.

This is the dialog presented:
```ts
  const resp = dialog.showMessageBoxSync(mainWindow, {
    title: i18n.t('Consent to command being run'),
    message: i18n.t('Allow this local command to be executed? Your choice will be saved.'),
    detail: command,
    type: 'question',
    buttons: [i18n.t('Allow'), i18n.t('Deny')],
  });
```

app/electron/runCmd.ts
illume: removed

app/electron/runCmd.ts
```diff
@@ -166,3 +168,32 @@ export function handleRunCommand(
...
+ * This is needed to run the "scriptjs" commands, as a way of running
+ * node js scripts without requiring node to also be installed.
+ */
+export function runScript() {
+  // If '..' in process.argv[1] or it starts with / or \, then it is not a valid path.
+  if (
+    !process.argv[1] ||
+    process.argv[1].includes('..') ||
+    process.argv[1].startsWith('/') ||
+    process.argv[1].startsWith('\\')
```
joaquimrocha, line 316: If plugins can ship their own script, I am not sure what this will prevent.

app/electron/runCmd.ts
```diff
@@ -237,3 +233,50 @@ export function setupRunCmdHandlers(mainWindow: BrowserWindow | null, ipcMain: E
...
+      `Invalid permission secrets, it is not an object: ${typeof eventData.permissionSecrets}`,
+    ];
+  }
+  for (const [key, value] of Object.entries(eventData.permissionSecrets)) {
+    if (typeof value !== 'number') {
+      return [false, `Invalid permission secret for ${key}: ${typeof value}`];
+    }
+  }
+
+  const validCommands = ['minikube', 'az', 'scriptjs'];
```
joaquimrocha, line 419: If we validate commands, but inside scriptjs we can run other commands, will this check prevent anything? (or maybe I am missing some details of how the scripts are run)

frontend/src/plugin/runPlugin.test.ts
```diff
@@ -0,0 +1,323 @@
...
+
+function runPluginInner(info: runPluginProps) {
+  const source = info[0];
+  const packageName = info[1];
+  const packageVersion = info[2];
+  const handleError = info[3];
+  const PrivateFunction = info[4];
+  const args = info[5];
+  const values = info[6];
+  const privateRunPlugin = info[7];
```
joaquimrocha, line 27: would this be more readable?
```typescript
const [
  source,
  packageName,
  ...
] = info;
```

app/electron/runCmd.ts
```diff
@@ -173,9 +185,27 @@ export function handleRunCommand(
...
+    commandData.command === 'scriptjs'
+      ? [getPluginsScriptPath(commandData.args[0]), ...commandData.args.slice(1)]
+      : commandData.args;
+
+  const child: ChildProcessWithoutNullStreams = spawn(command, args, {
     ...commandData.options,
     shell: false,
+    env: {
+      ...process.env,
+      ...(commandData.command === 'scriptjs' ? { HEADLAMP_RUN_SCRIPT: 'true' } : {}),
```
joaquimrocha, line 287: So this works even if we do not have the RUN_SCRIPT as true? Maybe leave a comment explaining why this behavior.

frontend/src/plugin/index.ts
```diff
@@ -244,10 +246,57 @@ export function updateSettingsPackages(
...
+ */
+function runPluginInner(info: runPluginProps) {
+  const source = info[0];
+  const packageName = info[1];
+  const packageVersion = info[2];
+  const handleError = info[3];
+  const PrivateFunction = info[4];
+  const args = info[5];
+  const values = info[6];
+  const privateRunPlugin = info[7];
```
joaquimrocha, line 264: Same comment about expanding a const list declaration.

app/electron/runCmd.ts
joaquimrocha: My comment was inline with the idea that we would always ask the user for command consent for any plugins, rather than the special-cased minikube one. So we'd record the consent together with "what plugin we allowed to run it". Unless I am mistake (can very well be), you are storing the commands in the settings without a notion of which plugin requested it because you are only considering minikube to use the commands in the settings.

frontend/src/plugin/runPlugin.ts
```diff
@@ -0,0 +1,209 @@
...
+  /** The values for the arguments to be given to the plugin */
+  values: unknown[],
+  /** A private copy of runPlugin */
+  internalRunPlugin: typeof runPlugin,
+  /** A private copy of console.error */
+  consoleError: typeof console.error
+];
+
+/**
+ * Runs a plugin by executing the source code in the global scope.
```
joaquimrocha: This doc is not accurate or not in the right function I believe.

frontend/src/plugin/runPlugin.ts
```diff
@@ -0,0 +1,209 @@
...
+    source + sourceMapPathForDebugging,
+    packageName,
+    packageVersion,
+    handleError,
+    PrivateFunction,
+    args,
+    values,
+    internalRunPlugin,
+    consoleError,
+  ];
```
joaquimrocha, line 127: You probably have a good reason to process this as a list rather than an object?

joaquimrocha (commented):
Left some comments. I think the code is complex and therefore hopefully you'll excuse/understand if some of my questions do not apply, I left mostly questions in order to get a better understanding.

frontend/src/plugin/runPlugin.test.ts
illume: This is to avoid calling the iterator, which could have the prototype overridden.

frontend/src/plugin/index.ts
illume: This is to avoid calling the iterator, which could have the prototype overridden. I’ll leave comments to describe why it’s being done.

app/electron/runCmd.ts
illume: This limits it to running ones inside the plugins folder as a defensive measure. Following the principal that inputs should be validated, and that if other parts of the system are bypassed perhaps this limit would be useful. It could also be useful if the user writes the wrong path, as they'd get an error message.

app/electron/runCmd.ts
illume: Validation here is still useful in case one of the other parts does not work to prevent them. Inputs should be validated anyway.

frontend/src/plugin/runPlugin.ts
illume: fixed

frontend/src/plugin/runPlugin.ts
illume: Added some comments about this.

frontend/src/plugin/index.ts
illume: Added a comment here about this too.

frontend/src/plugin/runPlugin.test.ts
illume: Left comments about this.

app/electron/runCmd.ts
illume: I added some comments explaining that this passes the HEADLAMP_RUN_SCRIPT env var, and why it does that.

  // If the command is 'scriptjs', we pass the HEADLAMP_RUN_SCRIPT=true 
  // env var so that the Headlamp or Electron process runs the script.

illume:
>There should be a way to change your choice on the allow/deny question. I've missclicked > "Deny" and then couldn't change this later and had to delete settings.json

I don't think this can happen for users of the plugin as this is now. Installing the plugin through artifact hub plugin catalog adds and removes the consent - so the user through this path won't encounter this issue. It could happen for developers, and if there's an evilPlugin. So a settings screen for these is not a priority, since users won't run into this yet.

app/electron/runCmd.ts
illume: It's just that these consents are given for the command, rather than being coupled to the command that gave them.

I don't see any problem that would be solved by changing this, or any value in changing it.

illume:
@joaquimrocha I responded to all your comments, and pushed a number of code comment corrections and additions explaining things.

PR #3611, opened 2025-07-10
Accessibility [FastPass]: Fix Pie Charts labels and contrast

vyncent-t:
This PR aims to fix the failure checks found by FastPass for the Chart component, mainly around ARIA labels and color contrast.

This PR is part of the accessibility compliance tasks which require the FastPass checks to be ran and resolved.|

### Added missing ARIA labels 
- Added logic for ARIA labels, since this component is used multiple times the issue repeats here

image

image

### Added acceptable contrast color to chart links
- There were contrast issues only for the themes that use light mode, added a theme for links within the light mode that uses a color that fits the FastPass range

image

image

### DIFF
--- a/frontend/src/components/App/PluginSettings/__snapshots__/PluginSettings.DefaultSaveEnable.stories.storyshot
+++ b/frontend/src/components/App/PluginSettings/__snapshots__/PluginSettings.DefaultSaveEnable.stories.storyshot
@@ -364,7 +364,7 @@
                   class="MuiTypography-root MuiTypography-subtitle1 css-1t3dwd5-MuiTypography-root"
                 >
                   <a
-                    class="MuiTypography-root MuiTypography-inherit MuiTypography-alignRight MuiLink-root MuiLink-underlineHover css-1rz8yre-MuiTypography-root-MuiLink-root"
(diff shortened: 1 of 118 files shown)


vyncent-t:
note to self: currently found a fix that fixes the issues for the pie chart in both overview and workloads, which is good.. but now its broken when we don't have a label for the chart? 

image
image

- also realized that using a generic id for each data piece may not work and could cause screen readers to just read the same thing multiple times, so maybe using useId would fix this
- also tried using dynamic id names by using `${title}-chart-title` but this does not print 'CPU-chart-title' but instead renders an empty string there since the props could be undefined. the current WIP fixes that with useId but now the fastpass is broken for no % in the circle... 🤔 
- maybe just not use it at all? if we dont have it? or change the actual props to make then needed to continue??? (maybe not since there is no way of knowing if we don't want to use it or the shape could change later??? need to think more on this later...)

vyncent-t:
noticed an issue w/ screen reader, working on a fix.

sniok:
aria-label on Pie element should be enough for this

sniok (approved):

PR #3612, opened 2025-07-10
frontend: Pod: Add tooltip for requests and limits

skoeva:
This change adds a tooltip to display requests and limits values for pods, if provided. This will improve resource tracking for individual pods in real time.

### Testing
- [X] Go to the "Pods" page in Headlamp
- [X] Navigate to a pod with resource requests and/or limits specified
- [X] Ensure that a tooltip displaying the values appears

image

### DIFF
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -31,6 +31,7 @@ import { StatusLabel, StatusLabelProps } from '../common/Label';
 import Link from '../common/Link';
 import ResourceListView from '../common/Resource/ResourceListView';
 import { SimpleTableProps } from '../common/SimpleTable';
+import { TooltipIcon } from '../common/Tooltip';
 import LightTooltip from '../common/Tooltip/TooltipLight';
 
 function getPodStatus(pod: Pod) {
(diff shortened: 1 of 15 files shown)


sniok:
Nice change, thanks!
Could you explain the percentage values? I'm not sure what 18% and 9% means in the screenshot

skoeva:
@sniok Let me know if this should be clearer, but percentages are
- actual (11.5 Mi in the table) / request (64 Mi) so 11.5/64 = 18%
- actual (11.5 Mi in the table) / limit (128 Mi) so 11.5/128 = 9%

sniok:
> @sniok Let me know if this should be clearer, but percentages are
> 
> * actual (11.5 Mi in the table) / request (64 Mi) so 11.5/64 = 18%
> * actual (11.5 Mi in the table) / limit (128 Mi) so 11.5/128 = 9%

that makes a lot more sense, maybe we can use a similar format in the tooltip?

klemen-df:
What's the status of this? Super awesome feature

frontend/src/components/pod/List.tsx
```diff
@@ -257,9 +303,54 @@ export function PodListRenderer(props: PodListProps) {
...
+                  const limit = pod.spec.containers
+                    .map(c => parseRam(c.resources?.limits?.memory || '0'))
+                    .reduce((a, b) => a + b, 0);
+
+                  const tooltipLines = [];
+                  if (request > 0) {
+                    const { value: rValue, unit: rUnit } = unparseRam(request);
+                    const percentOfRequest = ((memory / request) * 100).toFixed(1);
+                    tooltipLines.push(
+                      t('Request: {{percent}}% ({{aValue}} {{aUnit}}/{{rValue}} {{rUnit}})', {
```
sniok: for this we could probably do `t('Request')` and `t('Limit')`, and we already have translations for that

skoeva: done

illume:
@skoeva this is failing: `make frontend-i18n-check`

illume (approved):
🎉 Thanks!

PR #3632, opened 2025-07-16
Revert "tools/releaser: Bump octokit/rest to 22.0.0"

joaquimrocha:
This reverts commit 2dce4a970e48b8cd5982896ffb5bc3f29b1e251f.

This doesn't work when doing npm install + npm run start, as the import style is not supported apparently.


ashu8912 (approved):

sniok (approved):
npm i and npm run start works

PR #3642, opened 2025-07-17
frontend ResourceTable: Prevent eventual use of undefined clusters

joaquimrocha:


joaquimrocha:
Closing since this is not needed after all.

PR #3643, opened 2025-07-17
frontend: Add missing onClose handlers for EditDialog activities

sniok:
I missed a couple of onClose handlers.
Without `open` state and the Popup they were generally not needed but I forgot about the close button on the bottom right.

### How to test

Click "Create Resource" on the bottom left
Then close it using the close button on the bottom right


joaquimrocha:
Code looks fine and simple.

/lgtm
/approve

vyncent-t (approved):
tried the fix, closing works now 👍

PR #3650, opened 2025-07-18
app: .github: Run tests and type checker plus fix broken things

illume:
- Runs the type checker and tests for the app/ on win/linux/mac in a GitHub action.
- Fixes some type/test failures.
- Adds `make app-test` and `make app-tsc`

### DIFF
--- a/app/electron/plugin-management.ts
+++ b/app/electron/plugin-management.ts
@@ -148,9 +148,9 @@ export class PluginManager {
    * @returns {Promise<void>} A promise that resolves when the installation is complete.
    */
   static async install(
-    URL,
-    destinationFolder = defaultPluginsDir(),
-    headlampVersion = '',
+    URL: string,
+    destinationFolder: string = defaultPluginsDir(),
+    headlampVersion: string = '',
     progressCallback: null | ProgressCallback = null,
     signal: AbortSignal | null = null
   ) {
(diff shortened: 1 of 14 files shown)


app/electron/plugin-management.ts
```diff
@@ -633,8 +633,15 @@ async function downloadExtraFiles(
       if (!value.output || !value.input || value.input === value.output) {
         continue;
       }
-      const outputFile = path.join(binDir, value.output);
+      let outputFile = path.join(binDir, value.output);
+      if (os.platform() === 'win32' && !value.output.endsWith('.js')) {
```
ashu8912, line 642: I don't understand this change why is this required???

ashu8912 (approved):
Tested this locally and it works fine. Thanks for these changes @illume just a small comment/question

app/electron/plugin-management.ts
illume: This was so the test pass on windows.

I added some comments that I hope explain it better...

```
      // If on Windows, ensure that the output file ends with .exe
      // For example, minikube should be minikube.exe
      // If the extra file is a .js file, we do not add .exe
```

illume:
Thanks @ashu8912 

Added a comment to explain that piece of code, and rebased against main.

sniok (approved):
nice, thanks!

Issue #3655, opened 2025-07-18
Headlamp 0.33.0 on Mac ARM is broken


joaquimrocha:
## Describe the bug

A couple of people have reported that the last version, 0.33.0, is broken on Mac ARM.
The OS says the file is corrupted.

## Environment (please provide info about your environment):
 - Installation type: Mac / Brew or direct installation

joaquimrocha:
cc/ @illume for HW availability.

kierr:
I am on a Mac Mini M4, Sequoia 15.5, installed via brew. Having upgraded via brew all was fine.

I just removed and reinstalled with brew to test this and you are correct - corruption error shows.

It can be fixed with `xattr -dr com.apple.quarantine /Applications/Headlamp.app` as per the [docs](https://headlamp.dev/docs/latest/installation/desktop/).

```bash
user@mini ~ % codesign --verify --deep --strict --verbose=2 /Applications/Headlamp.app
/Applications/Headlamp.app: code has no resources but signature indicates they must be present
user@mini ~ % spctl --assess --type execute --verbose /Applications/Headlamp.app
/Applications/Headlamp.app: code has no resources but signature indicates they must be present
user@mini ~ % codesign -dvv /Applications/Headlamp.app
Executable=/Applications/Headlamp.app/Contents/MacOS/Headlamp
Identifier=Electron
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20400 size=513 flags=0x20002(adhoc,linker-signed) hashes=13+0 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements=none
```

The [release notes](https://github.com/kubernetes-sigs/headlamp/releases/tag/v0.33.0) for 0.33.0 show no warning about unsigned binaries as previous releases did - was this an oversight?

Nothing about the signature is different compared to the previous release (`.dmg` from brew cache):

```bash
user@mini downloads % codesign -dvv /Volumes/Headlamp\ 0.32.0-arm64/Headlamp.app
Executable=/Volumes/Headlamp 0.32.0-arm64/Headlamp.app/Contents/MacOS/Headlamp
Identifier=Electron
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20400 size=513 flags=0x20002(adhoc,linker-signed) hashes=13+0 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
[...]

PR #3658, opened 2025-07-20
chore(docs): brew `--no-quarantine` kludge to resolve bad sig/corrupt binary error

kierr:
## Summary

Adds `--no-quarantine` to the brew install command so that new users aren't confounded by corrupt binary errors (due to bad signature).

## Related Issue

#3655


illume (approved):
🎉 thanks for this!

PR #3662, opened 2025-07-21
frontend: Activity UI improvements

### LINKED ISSUE #3646: Taskbar UX improvement

Here are some UX suggestions to improve the taskbar:
- [ ] The overview button should be static
  - It makes no sense having to scroll in order to find the overview button
- [ ] The overview button should act like a status button: turn on/off overview
  - Right now it's only showing the overview
- [ ] New tasks should be added to the beginning of the taskbar not to the end
  - Because otherwise we may have to scroll to find it...
- [ ] The Overview should have a close all button. Similar to Android's all open apps view.
- [ ] Clicking on a task should also toggle minimize/unminimize (bring to front)
- [ ] Each Window should have the cluster name as well at the top

We can discuss each item individually if needed.

All users.

sniok:
Fixed #3646 
 
## Changes

The overview button is sticky 
New tabs area added to the beginning of the taskbar
Added Close All button to the overview
Clicking on a tab toggles minimize/unminimize
Each Activity has a cluster name now

### DIFF
--- a/frontend/src/components/activity/__snapshots__/Activity.Basic.stories.storyshot
+++ b/frontend/src/components/activity/__snapshots__/Activity.Basic.stories.storyshot
@@ -165,8 +165,44 @@
         </div>
       </div>
       <div
-        class="MuiBox-root css-1cbmz9o"
+        class="MuiBox-root css-3ykv34"
       >
+        <div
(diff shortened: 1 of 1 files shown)


joaquimrocha:
Tested and works great! Thanks!
/approve
/lgtm

PR #3674, opened 2025-07-22
fix: update extraManifests example

xfrancois:
## Summary

extraManifests example was not working properly, it failed with this error

Error: values don't meet the specifications of the schema(s) in the following chart(s): headlamp:
- extraManifests.0: Invalid type. Expected: string, given: object


illume (approved):
🎉 thanks

PR #3692, opened 2025-07-26
backend: headlamp: Add OIDC PKCE support

### LINKED ISSUE #3137: Enable Headlamp to use Public OIDC Client
## Current implementation enforces a oidc client secret

As part of our user authentication strategy, we want to setup a [_public_ Dex static Client](https://dexidp.io/docs/configuration/custom-scopes-claims-clients/#public-clients) and use this client with Headlamp.

However, as Headlamp is currently implemented, _not providing an `oidcClientSecret`_ will result in a failed login attempt:

```
if oidcClientID != "" && oidcClientSecret != "" && oidcIssuerURL != "" && oidcScopes != "" {
	oidcConf = &OidcConfig{
		ClientID:     oidcClientID,
		ClientSecret: oidcClientSecret,
		IdpIssuerURL: oidcIssuerURL,
		Scopes:       strings.Split(oidcScopes, ","),
	}
}
```

https://github.com/kubernetes-sigs/headlamp/blob/4598276d9a569202e0d16707044d37e05fb1094b/backend/pkg/kubeconfig/kubeconfig.go#L871

My proposal is to add a configuration option, which can be provided during the startup of Headlamp, allowing the use of an OIDC Client without a secret.

Everyone who wants to use [public clients](https://dexidp.io/docs/configuration/custom-scopes-claims-clients/#public-clients) in Dex together with Headlamp.

I think so. We have already started discussing this here: https://kubernetes.slack.com/archives/C01FXB5E8ER/p1744348416232779
 and I will reach out in the thread for more advice.

k-airos:
## Summary

Add PKCE support for OIDC authentication flow

## Related Issue

Fixes #3137

## Changes

- Added PKCE configuration: Newflag (default: false) to enable/disable PKCE
- Updated Config struct: Added OidcUsePKCE field to support PKCE configuration
- Implemented PKCE cryptographic functions
- Enhanced OauthConfig struct: Added field to store PKCE verifier

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/rand"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
@@ -34,6 +35,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -79,6 +81,7 @@ type HeadlampConfig struct {
 	oidcUseAccessToken        bool
 	oidcSkipTLSVerify         bool
 	oidcCACert                string
+	oidcUsePKCE               bool
 	cache                     cache.Cache[interface{}]
 	multiplexer               *Multiplexer
 	telemetryConfig           cfg.Config
@@ -117,9 +120,11 @@ type clientConfig struct {
 }
 
 type OauthConfig struct {
-	Config   *oauth2.Config
-	Verifier *oidc.IDTokenVerifier
-	Ctx      context.Context
+	Config       *oauth2.Config
+	Verifier     *oidc.IDTokenVerifier
+	Ctx          context.Context
+	CodeVerifier string // PKCE code verifier
+	Cluster      string // cluster context name this is associated with
 }
 
 // returns True if a file exists.
@@ -615,7 +620,10 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	config.addClusterSetupRoute(r)
 
-	oauthRequestMap := make(map[string]*OauthConfig)
+	var (
+		oauthRequestMap = make(map[string]*OauthConfig)
+		oauthMu         sync.Mutex
+	)
 
 	r.HandleFunc("/oidc", func(w http.ResponseWriter, r *http.Request) {
 		ctx := context.Background()
@@ -680,12 +688,38 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			RedirectURL:  getOidcCallbackURL(r, config),
 			Scopes:       append([]string{oidc.ScopeOpenID}, oidcAuthConfig.Scopes...),
 		}
-		/* we encode the cluster to base64 and set it as state so that when getting redirected
-		by oidc we can use this state value to get cluster name
-		*/
-		state := base64.StdEncoding.EncodeToString([]byte(cluster))
-		oauthRequestMap[state] = &OauthConfig{Config: oauthConfig, Verifier: verifier, Ctx: ctx}
-		http.Redirect(w, r, oauthConfig.AuthCodeURL(state), http.StatusFound)
+
+		// state should be unique per request, cryptographically secure random, url safe
+		state := func() string {
+			b := make([]byte, 32)
+			if _, err := rand.Read(b); err != nil {
+				panic(err)
+			}
+			return base64.RawURLEncoding.EncodeToString(b)
+		}()
+
+		entry := &OauthConfig{
+			Config:   oauthConfig,
+			Verifier: verifier,
+			Ctx:      ctx,
+			Cluster:  cluster,
+		}
+
+		var authURL string
+
+		if config.oidcUsePKCE {
+			entry.CodeVerifier = oauth2.GenerateVerifier()
+			authURL = oauthConfig.AuthCodeURL(state, oauth2.S256ChallengeOption(entry.CodeVerifier))
+		} else {
+			authURL = oauthConfig.AuthCodeURL(state)
+		}
+
+		// Store the request config keyed by state for callback handling
+		oauthMu.Lock()
+		oauthRequestMap[state] = entry
+		oauthMu.Unlock()
+
+		http.Redirect(w, r, authURL, http.StatusFound)
 	}).Queries("cluster", "{cluster}")
 
 	r.HandleFunc("/drain-node", config.handleNodeDrain).Methods("POST")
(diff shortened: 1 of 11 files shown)


illume:
Thanks @k-airos !

I asked in the issue for feedback, and ran the CI checks.

We'd need to documentation into the headlamp helm chart at some point (README.md, values.yaml and values.schema. Let me know if you'd like to add it in this PR, or we can do it after this is merged.
- https://github.com/kubernetes-sigs/headlamp/blob/main/charts/headlamp/values.yaml
- https://github.com/kubernetes-sigs/headlamp/blob/main/charts/headlamp/values.schema.json
- https://github.com/kubernetes-sigs/headlamp/blob/main/charts/headlamp/README.md

k-airos:
Thanks! I've added the documentation to README.md, updated values.yaml and values.schema.json accordingly.
Let me know if anything else should be adjusted!

illume:
@k-airos much appreciated.

I see Erik from the issue gave a thumbs up. Let's wait a little bit to see if anyone else from that issue has feedback.

I see there's an error in CI:
```
Testing Helm chart templates against expected output...
Error: parse error at (headlamp/templates/deployment.yaml:227): undefined variable "$usePKCE"

Use --debug flag to render out invalid YAML
```

To test this locally you can run:
`make helm-template-test`

illume:
@yolossn @ashu8912 When you get a chance, can you please review?

yolossn (commented):
I tried testing this in in-cluster and the app doesn't work without the client-secret because of this client-secret check. We have to make client-secret optional in the kubconfig.OidcConfig for the user to make headlamp work without client-secret.

https://github.com/kubernetes-sigs/headlamp/blob/68e74773825f910c28ff3f4401c1dd5ee6683646/backend/pkg/kubeconfig/kubeconfig.go#L911

illume:
Rebased it against main.

illume:
> I tried testing this in in-cluster and the app doesn't work without the client-secret because of this client-secret check. We have to make client-secret optional in the kubconfig.OidcConfig for the user to make headlamp work without client-secret.

I made the client-secret optional in the kubconfig.OidcConfig in fb511913100476751b9ef89d1e8e9fcaa628e50e

illume:
I added a test case for the PKCE helm config, and fixed the error copilot pointed out.

backend/cmd/headlamp.go
```diff
@@ -111,9 +114,29 @@ type clientConfig struct {
...
-	Verifier *oidc.IDTokenVerifier
-	Ctx      context.Context
+	Config       *oauth2.Config
+	Verifier     *oidc.IDTokenVerifier
+	Ctx          context.Context
+	CodeVerifier string // PKCE code verifier
+}
+
+// generateCodeVerifier generates a PKCE code verifier (high-entropy cryptographic random string).
+func generateCodeVerifier() (string, error) {
```
illume: generateCodeVerifier needs tests, and should probably be moved to auth package.

backend/cmd/headlamp.go
```diff
@@ -111,9 +114,29 @@ type clientConfig struct {
...
+	bytes := make([]byte, 32)
+	if _, err := rand.Read(bytes); err != nil {
+		return "", err
+	}
+	// Use base64url encoding without padding
+	return base64.RawURLEncoding.EncodeToString(bytes), nil
+}
+
+// generateCodeChallenge generates a PKCE code challenge from the verifier using SHA256.
+func generateCodeChallenge(verifier string) string {
```
illume: generateCodeChallenge needs tests, and should probably be moved to auth package.

backend/cmd/headlamp.go
```diff
@@ -711,7 +775,26 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+
+			var err error
+
+			// Exchange authorization code for token, with or without PKCE
+			if config.oidcUsePKCE && oauthConfig.CodeVerifier != "" {
+				// Use PKCE code verifier for token exchange
+				oauth2Token, err = oauthConfig.Config.Exchange(
+					oauthConfig.Ctx,
+					r.URL.Query().Get("code"),
+					oauth2.SetAuthURLParam("code_verifier", oauthConfig.CodeVerifier),
```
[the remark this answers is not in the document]
illume: I think this is already being done...

> oauth2Token, err = oauthConfig.Config.Exchange(
					oauthConfig.Ctx,
					r.URL.Query().Get("code"),
					oauth2.SetAuthURLParam("code_verifier", oauthConfig.CodeVerifier),

backend/cmd/headlamp.go
```diff
@@ -762,6 +845,10 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			auth.SetTokenCookie(w, r, string(decodedState), rawUserToken, config.BaseURL)
 
 			redirectURL += fmt.Sprintf("auth?cluster=%1s", decodedState)
+
+			// Clean up the OAuth state to prevent memory leaks
+			delete(oauthRequestMap, state)
```
illume: I think this should be in a defer? 

What if this code returns in the error paths above? That means this cleanup will not be done.

illume:
- Added the funlen lint ignore for the lint error in the big config function
- Did some more review, and made a couple more notes about tests requires and about a memory leak

charts/headlamp/values.yaml
```diff
@@ -78,6 +78,8 @@ config:
     validatorIssuerURL: ""
     # -- Use 'access_token' instead of 'id_token' when authenticating using OIDC
     useAccessToken: false
+    # -- Use PKCE (Proof Key for Code Exchange) for enhanced security in OIDC flow
+    usePKCE: false
```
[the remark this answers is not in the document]
illume: Updated the PR description.

illume:
Collecting some PKCE documentation:
- https://datatracker.ietf.org/doc/html/rfc7636
- https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.32.0:pkce.go
- https://pkg.go.dev/golang.org/x/oauth2#GenerateVerifier
- https://pkg.go.dev/golang.org/x/oauth2#S256ChallengeFromVerifier
- https://pkg.go.dev/golang.org/x/oauth2#Config.AuthCodeURL
- https://developer.constantcontact.com/api_guide/pkce_flow.html
- https://www.authlete.com/developers/pkce/

backend/cmd/headlamp.go
illume: This function is now included in Golang. See GenerateVerifier
https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.32.0:pkce.go;l=27

backend/cmd/headlamp.go
illume: Included in Golang now as S256ChallengeFromVerifier

https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.32.0:pkce.go;l=49

backend/cmd/headlamp.go
```diff
@@ -665,12 +688,53 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
...
+			by oidc we can use this state value to get cluster name
+			*/
+			state := base64.StdEncoding.EncodeToString([]byte(cluster))
+			oauthRequestMap[state] = &OauthConfig{
+				Config:       oauthConfig,
+				Verifier:     verifier,
+				Ctx:          ctx,
+				CodeVerifier: codeVerifier,
+			}
+
```
illume: TODO: I think S256ChallengeOption could be used. https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.32.0:pkce.go;l=57

check if it can

illume:
Made it use the golang oauth2 PKCE functions.

illume:
Added a mutex around where the auth config is stored to stop problems if two requests access it at the same time.

illume:
Rebased against main, which has that CI fix in it. 
(So that "Build Container and test" check should pass.)

illume (approved):
🎉 thanks!

PR #3700, opened 2025-07-28
Accessibility [FastPass]: Fix ARIA label and tab navigation for notifications menu

vyncent-t:
This PR fixes the FastPass check for the notifications menu button when on the notifications page. 

Added a translated string for screen readers for the menu icon button, it should now read out "Notifcations menu"

I also included a change that moves the on click so that tab navigation works as intended. (Before you could not tab then press enter to open the menu, now you can.)

## FastPass issues

image

image

### DIFF
--- a/frontend/src/components/App/Notifications/List/List.tsx
+++ b/frontend/src/components/App/Notifications/List/List.tsx
@@ -100,8 +100,8 @@ export default function NotificationList() {
 
     return (
       <>
-        <IconButton size="medium">
-          <Icon icon="mdi:dots-vertical" onClick={handleClick} />
+        <IconButton aria-label={t('translation|Menu')} size="medium" onClick={handleClick}>
+          <Icon icon="mdi:dots-vertical" />
         </IconButton>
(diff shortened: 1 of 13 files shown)


frontend/src/components/App/Notifications/List/List.tsx
```diff
@@ -100,8 +100,12 @@ export default function NotificationList() {
 
     return (
       <>
-        <IconButton size="medium">
-          <Icon icon="mdi:dots-vertical" onClick={handleClick} />
+        <IconButton
+          aria-label={t('translation|Notifications menu')}
```
joaquimrocha: If we use "Menu" and it's under the Notifications section maybe that works fine and we do not have to special case a translation that may be needed for other cases?

joaquimrocha (changes requested):

frontend/src/components/App/Notifications/List/List.tsx
vyncent-t: okay will change from `Notifications menu` > `Menu`

vyncent-t: fixed

PR #3706, opened 2025-07-29
frontend: Table: Enhance click handling for table rows

sniok:
This makes shift-selecting only work by clicking on checkbox instead of the whole row.
Also avoids selecting text by doing e.preventDefault

## Steps to Test

1. Go to any resource table
2. Select rows by clicking on checkboxes
3. Shift-select multiple rows


joaquimrocha:
tested.
/lgtm
/approve

PR #3711, opened 2025-07-29
Update npm deps

joaquimrocha:
Just running npm audit fix in the frontend and app modules.


skoeva (commented):
looks like the dependencies were updated with #3703, so this PR can be closed

PR #3712, opened 2025-07-29
fix: remove the unnecessary  assignment

LinPr:
## Summary
Remove the unnecessary assignment in for loop since we use over 1.24 go version, there is no need to assign a new variable any more


LinPr:
> Thanks for catching this, see [this comment](https://github.com/kubernetes-sigs/headlamp/pull/3709#pullrequestreview-3075091572) for feedback

ok I create the new PR #3726 
and I will close this one

PR #3714, opened 2025-07-30
frontend: Add token only if it's available

### LINKED ISSUE #3605: Port forwarding not working "portforward stopped before becoming ready"
## Describe the bug
Using the local Headlamp, when trying to open a port forward on a pod, I instead get a spinner that stays forever.
The developer console -> network tab shows the response "portforward stopped before becoming ready"
Image

## To Reproduce
Steps to reproduce the bug:
1. Go to 'Workloads -> Pods'
2. Click a pod with defined ports
3. Hit the port forward button
4. See error above

 - Installation type: Linux-AppImage and also tested with Mac-Homebrew
 - Headlamp Version: 0.32.0
 - Other: We are using EKS with a config like this (and AWS SSO if relevant):
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ABCDEF...
    server: https://ABCD123.gr7.eu-central-1.eks.amazonaws.com
  name: arn:aws:eks:eu-central-1:12345:cluster/dev
- cluster:
    certificate-authority-data: ABCDEF...
    server: https://QWERT123.gr7.eu-central-1.eks.amazonaws.com
  name: arn:aws:eks:eu-central-1:12345:cluster/test
contexts:
- context:
    cluster: arn:aws:eks:eu-central-1:12345:cluster/dev
    namespace: abc
    user: dev
  name: dev
- context:
    cluster: arn:aws:eks:eu-central-1:12345:cluster/test
    namespace: qwe
    user: test
  name: test
current-context: dev
kind: Config
preferences: {}
users:
- name: dev
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - eu-central-1
      - eks
      - get-token
      - --cluster-name
      - dev
      command: aws
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
- name: test
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - eu-central-1
      - eks
      - get-token
      - --cluster-name
      - test
      command: aws
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

Yes, but only a workaround.
I believe the issue is related to the Header value `authorization: Bearer undefined` provided in the request:

Image

Simply removing the `r.Header.Get("Authorization")` part here made everything work flawlessly.
https://github.com/kubernetes-sigs/headlamp/blob/bb0b5562ffc561fa460b47488ddc7e34002d4172/backend/pkg/portforward/handler.go#L135
`reqToken := "" // r.Header.Get("Authorization")`

Image

ashu8912:
fixes #3605 

- #3605

### DIFF
--- a/frontend/src/lib/k8s/api/v1/portForward.ts
+++ b/frontend/src/lib/k8s/api/v1/portForward.ts
@@ -74,14 +74,18 @@ export async function startPortForward(
   id: string = ''
 ): Promise<PortForward> {
   const kubeconfig = await findKubeconfigByClusterName(cluster);
-  const headers: HeadersInit = {
-    Authorization: `Bearer ${getToken(cluster)}`,
+  const headers: HeadersInit = new Headers({
     ...JSON_HEADERS,
-  };
+  });
+
+  const token = getToken(cluster);
+  if (!!token) {
+    headers.set('Authorization', `Bearer ${token}`);
+  }
 
   // This means cluster is dynamically configured.
   if (kubeconfig !== null) {
-    headers['X-HEADLAMP-USER-ID'] = getUserIdFromLocalStorage();
+    headers.set('X-HEADLAMP-USER-ID', getUserIdFromLocalStorage());
   }
 
   const request: PortForwardRequest = {
@@ -97,7 +101,7 @@ export async function startPortForward(
   };
   return fetch(`${getAppUrl()}portforward`, {
     method: 'POST',
-    headers: new Headers(headers),
+    headers: headers,
     body: JSON.stringify(request),
   }).then(async (response: Response) => {
     const contentType = response.headers.get('content-type');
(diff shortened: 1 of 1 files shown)


frontend/src/lib/k8s/api/v1/portForward.ts
```diff
@@ -74,14 +74,18 @@ export async function startPortForward(
   id: string = ''
 ): Promise<PortForward> {
   const kubeconfig = await findKubeconfigByClusterName(cluster);
-  const headers: HeadersInit = {
-    Authorization: `Bearer ${getToken(cluster)}`,
+  const headers: HeadersInit = new Headers({
```
sniok, line 77: new Headers() is not really needed, just plain object works fine
but it doesn't really matter

PR #3718, opened 2025-07-30
tools/releaser: Bump octokit/rest to 22.0.0

skoeva:
This change follows up on missing updates in #3521 after bumping octokit/rest to the new ESM-only v22.

To support this new version, the tools/releaser directory is migrated to ESM:
- Switched the `tsconfig.json` to `"module": "node16"` and `"moduleResolution": "node16"`
- Added `"type": "module"` to `package.json`
- Updated all imports to use `.js` extensions for local modules
- Replaced CJS globals (`__dirname`, `__filename`) with ESM shims
- Kept `Octokit` type import for type safety

### Testing
- [X] Run `cd tools/releaser && npm i`
- [X] Run `npm run build and npm run start` and ensure no errors appear
- [X] Run commands and ensure they work as expected

### DIFF
--- a/tools/releaser/package-lock.json
+++ b/tools/releaser/package-lock.json
@@ -8,7 +8,7 @@
       "name": "headlamp-releaser",
       "version": "0.1.0",
       "dependencies": {
-        "@octokit/rest": "^19.0.13",
+        "@octokit/rest": "^22.0.0",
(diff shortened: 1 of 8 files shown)


joaquimrocha:
Tested and it works.

/lgtm
/approve

PR #3722, opened 2025-07-30
backend: Dockerfile: Use golang version 1.24.4

skoeva:
This change sets the go version in the backend and Dockerfile to 1.24.4 to avoid complications with the latest version.


illume (changes requested):
...

Dockerfile
```diff
@@ -3,7 +3,7 @@
 ARG IMAGE_BASE=alpine:3.20.6@sha256:de4fe7064d8f98419ea6b49190df1abbf43450c1702eeb864fe9ced453c1cc5f
 FROM ${IMAGE_BASE} as image-base
 
-FROM --platform=${BUILDPLATFORM} golang:1.24.5@sha256:2b1cbf278ce05a2a310a3d695ebb176420117a8cfcfcc4e5e68a1bef5f6354da as backend-build
+FROM --platform=${BUILDPLATFORM} golang:1.24.4@sha256:2b1cbf278ce05a2a310a3d695ebb176420117a8cfcfcc4e5e68a1bef5f6354da as backend-build
```
sniok: I think hash should also be updated?

yolossn: +1
```suggestion
FROM --platform=${BUILDPLATFORM} golang:1.24.4@sha256:f981a85e7f0c5ec2b0ecef4a9746db2fe697f0d2fb90844824c7b7f8aa06c428 as backend-build
```

https://hub.docker.com/layers/library/golang/1.24.4/images/sha256-f981a85e7f0c5ec2b0ecef4a9746db2fe697f0d2fb90844824c7b7f8aa06c428

PR #3724, opened 2025-07-31
Fixing hover styling on close button in activity bar

Exar04:
The over circle over the close icon was not proper circle
Before - 
Screenshot 2025-07-31 at 6 06 18 PM
Screenshot 2025-07-31 at 6 06 28 PM

Now - 
Screenshot 2025-07-31 at 6 08 58 PM
Screenshot 2025-07-31 at 6 09 04 PM

### DIFF
--- a/frontend/src/components/activity/Activity.tsx
+++ b/frontend/src/components/activity/Activity.tsx
@@ -948,6 +948,8 @@ export const ActivityBar = React.memo(function ({
           key={it.id}
           sx={theme => ({
             display: 'flex',
+            alignItems: 'center',
+            padding: '3px',
             height: '100%',
             position: 'relative',
             border: '1px solid',
@@ -958,6 +960,7 @@ export const ActivityBar = React.memo(function ({
         >
           <Button
             sx={{
+              height: '100%',
               padding: '0px 5px 0 10px',
               lineHeight: 1,
               whiteSpace: 'nowrap',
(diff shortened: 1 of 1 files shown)


sniok:
This will make some empty spots around the button, maybe instead we can make the close button square and full height?

frontend/src/components/activity/Activity.tsx
```diff
@@ -1008,7 +1011,7 @@ export const ActivityBar = React.memo(function ({
               e.stopPropagation();
               Activity.close(it.id);
             }}
-            sx={{ width: '42px', flexShrink: 0 }}
+            sx={{ width: '42px', height: '42px', flexShrink: 0 }}
```
skoeva: following the previous suggestion
```suggestion
            sx={{ width: '42px', height: '100%', flexShrink: 0, borderRadius: '5px' }}
```

skoeva (changes requested):
could you rebase against main and try to resolve the frontend test failure?

Exar04:
> could you rebase against main and try to resolve the frontend test failure?

Rebase done

Ran the test for Activity and it seems to pass. Not sure if failing tests are related to this
Screenshot 2025-08-05 at 4 12 22 PM

illume (approved):
Thanks! 🎉

ps. There seems to be an unrelated issue with the Activity tests still. They pass on Ubuntu, so I ran them on a local Ubuntu and pushed the snapshot. Then the tests pass on the github checks.

PR #3727, opened 2025-07-31
backend: HeadlampServer: Move router initialization codes to createHeadlampHandler

### LINKED ISSUE #3708: Threre are some dead codes here
## Dead code position
The router object is initialized but never used

https://github.com/kubernetes-sigs/headlamp/blob/f4e864e0428c8e6d7385d03a9c5d88689236fc63/backend/cmd/headlamp.go#L1142

```go
	router := mux.NewRouter()

	if config.Telemetry != nil && config.Metrics != nil {
		router.Use(telemetry.TracingMiddleware("headlamp-server"))
		router.Use(config.Metrics.RequestCounterMiddleware)
	}

```

[Yes (I will propose a PR).]

LinPr:
## Summary
Remove the dead code, and move router  use middle logic into createHeadlampHandler function

## Related Issue
Fixes #3708

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -453,6 +453,11 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 		r = baseRoute.PathPrefix(config.BaseURL).Subrouter()
 	}
 
+	if config.Telemetry != nil && config.Metrics != nil {
+		r.Use(telemetry.TracingMiddleware("headlamp-server"))
+		r.Use(config.Metrics.RequestCounterMiddleware)
+	}
+
 	fmt.Println("*** Headlamp Server ***")
 	fmt.Println("  API Routers:")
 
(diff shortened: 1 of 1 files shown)


skoeva (changes requested):
There are two different changes in this PR, I would recommend separating them and maybe condensing the PR title

BTW, you don't need to create new PRs to change the titles. You can use the interactive rebase (`git rebase -i`) which should be very useful here

LinPr:
> There are two different changes in this PR, I would recommend separating them and maybe condensing the PR title
> 
> BTW, you don't need to create new PRs to change the titles. You can use the interactive rebase (`git rebase -i`) which should be very useful here

ok thanks for your suggestions, I will make it better

skoeva (approved):
thanks for this 👍

illume:
Had to revert this here: 
- https://github.com/kubernetes-sigs/headlamp/pull/3798

PR #3730, opened 2025-08-01
backend: Remove unnecessary assignment in range for loop

LinPr:
## Summary

Remove unnecessary assignment in range for loop, the return values of range for are already local values, dont have to do it again in the range for loop body


yolossn:
With the changes that were introduced in v1.22 the reinitialisation is redundant.
https://go.dev/doc/go1.22#language

@LinPr Thanks for the PR.

PR #3733, opened 2025-08-01
app: mac: Fix broken first load of clusters and plugins

### LINKED ISSUE #3596: Plugins not loading on first run after installation on Mac
## Describe the bug
On Mac, when Headlamp is run for the first time after installation, plugins do not seem to be running. This includes the plugin catalog and app catalog not being shown, which suggests that core plugins are not loading properly on the initial startup.

## To Reproduce
Steps to reproduce the bug:
1. Install Headlamp on Mac (via Homebrew or other method)
2. Launch Headlamp for the first time
3. Check for plugin catalog and app catalog functionality
4. Observe that plugins appear to not be loaded/running

 - Installation type: Mac-Homebrew (and possibly other Mac installation methods)
 - Headlamp Version: 0.32.0

## Additional Context
This appears to be a Mac-specific issue that only affects the first run of Headlamp after installation. Subsequent runs may work correctly.

illume:
This starts loading the backend right away, rather than waiting for the window to load.

With the mac losing signed binaries it meant they are additionally taking longer to load on first run.

Fixes #3596
-  #3596

### How to test

- compile app on Mac, and install it
- run app
- See that 1) **clusters are there** (without having to refresh app)
- See that 2) **plugins are loaded** (without refreshing app)
- There should not be that much (or any) delay waiting (plugins and clusters should be there quickly... on main there is a delay currently)

### Testing done

- Intel and Silicon Macs with `npm start` and with compiled app (could reproduce before)
- Windows and Ubuntu compiled apps (could not reproduce before)

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -1118,6 +1118,94 @@ function startElecron() {
   console.log('Check for updates: ', shouldCheckForUpdates);
 
   async function createWindow() {
+    if (!useExternalServer) {
+      serverProcess = await startServer();
+      attachServerEventHandlers(serverProcess);
+
+      serverProcess.addListener('exit', async e => {
+        const ERROR_ADDRESS_IN_USE = 98;
+        if (e === ERROR_ADDRESS_IN_USE) {
+          const runningHeadlamp = await getRunningHeadlampPIDs();
+          let shouldWaitForKill = true;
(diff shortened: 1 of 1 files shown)


illume:
Thanks @ashu8912

PR #3735, opened 2025-08-01
app: main: Fix scriptjs for long running scripts in compiled app

illume:
This is when using the compiled app to run scripts.

Otherwise for long running scripts (or on a slow computer) the
app runs some normal stuff and exits before the rest of the
script has had a chance to run.


skoeva (commented):
should the flag be set to false after the script is done running?

illume:
>should the flag be set to false after the script is done running?

No, because it is used later on to stop other parts from running.

skoeva (approved):
makes sense, thanks 👍

Issue #3731, opened 2025-08-01
Use Cobra CLI framework to initialize the backend


LinPr:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
Currently the backend use many command line arguments, and flags, but the style currently is not very fluent and not compatible with the mainstream project like kubectl, helm, and other Cloud Native project. Most of them use  [Cobra](https://github.com/spf13/cobra) to initialize the project  It provide pretty neat flag and command line argument style, and manry useful features. 

## Describe the solution you'd like
Use Cobra to refactor the backend initialization.

## What users will benefit from this feature?
Make backend initially cleaner, and compatible wtih the main stream of Clould Native projects. It will also improve the extensibility of this project 

## Are you able to implement this feature?
[Yes (I will try to propose a PR when I have spare time, but not guarantee， if anyone who wants to PR this,  feel free to do that)]
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

LinPr:
Currently the backed mainly contains two fucntions, so I want to impliment two sub commands for headlamp backend with Cobra 
1. server: run the headlamp server 
2. plugins: list all headmap plugins

And then the useage will be like 

```bash
./headlamp server 

./headlamp plugins 
```


And it is possible to add more sub commands in the future if needed , the style just like other Cobra based projects. if we want to see the command usage details , we can execute 

```bash
./headlamp --help

./headlamp server --help 
```

We can also import [viper](https://github.com/spf13/vipe)r to manage flags and environment variables, and we can bind the flags with config file together, which make it more convenient to a bunch of configs.

And what do you guys think of this, need some feedback, thanks!

skoeva:
Looks like this is [not planned](https://github.com/kubernetes-sigs/headlamp/pull/3738#issuecomment-3147182364)

PR #3738, opened 2025-08-02
backend: refactor backend initialization with cobra cli framework

LinPr:
## Summary

This PR  refactor backend initialization with cobra cli framework
Support  pass configurations through flag and environment， the flag precedence if higher than env

## Related Issue

refer to issue: #3731 

image

## Notes for the Reviewer


yolossn:
Thanks for the PR, @LinPr! We deliberately chose not to use Viper in order to keep the backend lightweight and avoid adding unnecessary complexity. Instead, we went with [koanf](https://github.com/knadh/koanf), which aligns better with our goals for simplicity and modularity. You can see more on why we preferred koanf here: [koanf – Alternatives to Viper](https://github.com/knadh/koanf?tab=readme-ov-file#alternative-to-viper).

Is there something missing in Headlamp’s current setup that you’re trying to solve? We’d be happy to understand your use case better.

PR #3742, opened 2025-08-04
frontend: ClusterContextMenu: Fix menu to not show two deletes

illume:
Do not show a delete if there are custom menu items.
Because they are responsible for adding delete.

Before (two deletes):

Screenshot 2025-08-04 at 21 38 02

After:

Screenshot 2025-08-04 at 21 36 17


skoeva:
is the two delete issue limited to clusters created with the minikube plugin? I'm not able to reproduce this but it may be that I'm having trouble getting the plugin working

illume:
@skoeva it's the only plugin using this so far. It has it's own way of deleting clusters (deleting the cluster rather than removing it from the kubeconfig file).

skoeva (approved):
lgtm

Issue #3746, opened 2025-08-05
a11y: Dark mode issues


illume:
I spotted a few contrast issues in dark mode...

1) This one on the Notifications action menu:

<img width="380" height="316" alt="Image" src="https://github.com/user-attachments/assets/2ff62204-eac0-4f96-81ee-fcb70bbf5130" />

2) I noticed also the "Cancel" button on actions is unreadable

<img width="747" height="244" alt="Image" src="https://github.com/user-attachments/assets/31749d81-bc45-4aa5-bf57-09ccdad4cac4" />

3) The issue pasted in slack headlamp channel today was similar to one I saw in my plugin where `<Alert severity="warning">` has text that's too dark.

<img width="1229" height="125" alt="Image" src="https://github.com/user-attachments/assets/709261ea-e848-48e7-bb48-e40c443eb62e" />

vyncent-t:
working on a fix

illume:
@vyncent-t thanks kindly

illume:
I think these are all fixed now 🎉

@vyncent-t please reopen if I'm wrong.

PR #3747, opened 2025-08-05
frontend: Notifications: Fix contrast for menu items on all themes

vyncent-t:
## Summary

This PR fixes the issue where the text for notification menu items were not readable on darker themes

## Related Issue

Related to issue https://github.com/kubernetes-sigs/headlamp/issues/3746

## Changes

- Added contrast text via theme.palette to all notification menu items
- Keeps buttons disabled when needed, just not readable

## Steps to Test

1. Navigate to notifications
2. Click on the menu button
3. Note that they are now contrast friendly

## Before
image
image
image

### disabled
image
image
image

### enabled
image
image
image

### DIFF
--- a/frontend/src/components/App/Notifications/List/List.tsx
+++ b/frontend/src/components/App/Notifications/List/List.tsx
@@ -104,11 +104,29 @@ export default function NotificationList() {
           <Icon icon="mdi:dots-vertical" />
         </IconButton>
         <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
-          <MenuItem onClick={markAllAsRead} disabled={!hasUnseenNotifications}>
-            <Typography color={'primary'}>{t('translation|Mark all as read')}</Typography>
+          <MenuItem
+            sx={{
+              '&.Mui-disabled': {
+                opacity: 0.7,
+              },
+            }}
+            onClick={markAllAsRead}
+            disabled={!hasUnseenNotifications}
+          >
+            <Typography color={theme.palette.text.primary}>
+              {t('translation|Mark all as read')}
+            </Typography>
           </MenuItem>
-          <MenuItem onClick={clearAllNotifications} disabled={allNotificationsAreDeleted}>
-            <Typography color="primary">{t('translation|Clear all')}</Typography>
+          <MenuItem
+            sx={{
+              '&.Mui-disabled': {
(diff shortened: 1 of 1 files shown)


sniok:
But now the buttons don't look disabled, they look like you can click on them which is misleading.
Disabled items don't have contrast requirement [[source]](https://www.w3.org/TR/WCAG22/#contrast-minimum)

vyncent-t:
> But now the buttons don't look disabled, they look like you can click on them which is misleading. Disabled items don't have contrast requirement [[source]](https://www.w3.org/TR/WCAG22/#contrast-minimum)

current contrast was too hard to read, tweaked the new contrast to still be faded and not solid, let me know if this is readable

frontend/src/components/App/Notifications/List/List.tsx
```diff
@@ -104,11 +104,31 @@ export default function NotificationList() {
           <Icon icon="mdi:dots-vertical" />
         </IconButton>
         <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
-          <MenuItem onClick={markAllAsRead} disabled={!hasUnseenNotifications}>
-            <Typography color={'primary'}>{t('translation|Mark all as read')}</Typography>
```
sniok, line 108: setting color as `<Typography color="text.primary">...` fixes this
there's no need for any additional styles 

```tsx
 <MenuItem onClick={markAllAsRead} disabled={!hasUnseenNotifications}>
    <Typography color="text.primary">{t('translation|Mark all as read')}</Typography>
  </MenuItem>
```

Image

Image

sniok (changes requested):

frontend/src/components/App/Notifications/List/List.tsx
vyncent-t: It looks like that adjustment does work but only for the lights out contrast (it was blue before).

however for the dark mode contrast it is still the same contrast as the original issue https://github.com/kubernetes-sigs/headlamp/issues/3746


before
image

after
image

vyncent-t:
Just to be clear I believe the main goal here is to make the disabled text pop a little bit more while still looking faded enough to not be mistaken as enabled.

If the current colors are still too solid and do not fit disabled styles enough I can adjust it to be lighter.

PR #3755, opened 2025-08-06
app: plugins: tools/releaser: Bump tmp to 0.2.5

skoeva:
This change updates the tmp dependency in the app and plugins directories to the desired 0.2.5 version.

This is temporary until the example plugins are updated with the newest version of headlamp-plugin, which no longer has the tmp dependency.


illume (approved):
🎉 great :)

PR #3765, opened 2025-08-08
Frontend: LogViewer: Fix layout for log viewer buttons and switches

vyncent-t:
## Summary

This PR aims to update the current layout structure for the switches and inputs of the log viewer button, and also fixes misaligned display issue for the inputs.

These changes come from a previous PR with the wrap texts feature (retired https://github.com/kubernetes-sigs/headlamp/pull/3027). 

## Related Issue

https://github.com/kubernetes-sigs/headlamp/issues/3052

## Changes

- Reformatted the display layout using boxed grids for the log viewer inputs
- Fixes current misaligned log viewer input

## Steps to Test

1. Navigate to deployments
2. Open a deployment item and click view resource
3. Notice the misalignment is now fixed

## Before

- Deployment view image

- Log viewer Responsiveness before image

## After

- Fixed Deployment view image

- Log viewer Responsiveness after 
image

### DIFF
--- a/frontend/src/components/pod/Details.tsx
+++ b/frontend/src/components/pod/Details.tsx
@@ -21,7 +21,7 @@ import InputLabel from '@mui/material/InputLabel';
 import MenuItem from '@mui/material/MenuItem';
 import Select from '@mui/material/Select';
 import Switch from '@mui/material/Switch';
-import { styled } from '@mui/system';
+import { Box, styled } from '@mui/system';
 import { Terminal as XTerminal } from '@xterm/xterm';
 import _ from 'lodash';
 import React from 'react';
@@ -31,7 +31,7 @@ import { KubeContainerStatus } from '../../lib/k8s/cluster';
 import Pod from '../../lib/k8s/pod';
 import { DefaultHeaderAction } from '../../redux/actionButtonsSlice';
 import { EventStatus, HeadlampEventType, useEventCallback } from '../../redux/headlampEventSlice';
-import { Activity } from '../activity/Activity';
+import { Activity, useActivity } from '../activity/Activity';
 import ActionButton from '../common/ActionButton';
 import Link from '../common/Link';
 import { LogViewer, LogViewerProps } from '../common/LogViewer';
@@ -57,6 +57,7 @@ const PaddedFormControlLabel = styled(FormControlLabel)(({ theme }) => ({
 
 interface PodLogViewerProps extends Omit<LogViewerProps, 'logs'> {
   item: Pod;
+  location: 'split-left' | 'split-right' | 'full' | 'window';
 }
 
 export function PodLogViewer(props: PodLogViewerProps) {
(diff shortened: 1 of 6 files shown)


vyncent-t:
Note that I have tried using grid earlier in the PR but had seen issues when come to spacing and styling. These issues were leading to more box solutions needed to be added, thus I switched to using boxes.

vyncent-t:
I feel the switches may be awkwardly high, been trying to tweak a top margin to have them lower to the log window, may look fine

vyncent-t:
> Tested and noticed that the new buttons disappear for a split sec when I click them. Also I couldn't see anything noticeable when I clicked the "format" button.

I tried the new `format` and `prettify` on the current main while viewing pod logs for the default `storage-provisioner` and I see the `format` button also not doing anything, as well as the new buttons disappearing for a split second when used.

Those are separate issues from this PR, this PR focuses on changing the layout of the buttons and switches to now have proper responsiveness at different screen sizes and also fixing a misalignment issue.

I was using words layout and format interchangeably here, which coincidentally enough, the actual Format button needs to be fixed (lol)

I will be updating the wording now

vyncent-t:
need to make a fix after rebasing, moving back to draft

frontend/src/components/common/Resource/LogsButton.tsx
```diff
@@ -371,104 +371,209 @@ function LogsButtonContent({ item }: LogsButtonProps) {
   const topActions = [
     <Box
       key="container-controls"
-      sx={{ display: 'flex', gap: 2, alignItems: 'center', width: '100%' }}
+      sx={{
+        display: 'flex',
+        flexDirection: { xs: 'row', lg: 'row' },
+        wrap: {
```
sniok, line 385: wrap is not a valid css property, did you mean flexWrap?
Also I don't think we can rely on screen breakpoints for this, because the log viewer can be narrow on wide screens when it's open in an activity (split left or right)

for example: 

Image

vyncent-t: Good point, if I cannot find a work around I will just close this PR. 

This PR was a remnant of some useful changes from the toggle word wrap feature, which we also already closed.

vyncent-t: I was able to adjust a few more things making use of the location from the activities logic.

This should make it so that the inputs are still usable when snapping to left or right side.

I also cleaned up some of the UI usage to make things cleaner and more uniform.

Let me know what you think

sniok: I think this is a bit overengineered, having flex-wrap on a container should be enough

sniok (changes requested):
The way it already wraps on in Pods logs on main is good, could we do the same thing for deployments too?

Image

frontend/src/components/common/Resource/LogsButton.tsx
vyncent-t: No problem, the initial goal here was to preserve a column look when resizing. Since that look is no longer desired and a simpler fix can be made here I will be closing this PR and opening up a smaller one.

vyncent-t:
> The way it already wraps on in Pods logs on main is good, could we do the same thing for deployments too?
> 
> Image

Will do

vyncent-t:
Closing PR as noted in the conversation with @sniok above

my main goal of this PR had two parts: to restructure the layout inputs into a column format, and to fix some display and responsiveness issues. 

I was using some logic from an old PR (the trashed wrap button toggle) to accomplish this, but the column look is not needed.

I will be closing this PR and opening up a new one linked here that will keep the layout as it is on main but fix the rendering issues.

vyncent-t:
New PR that discards this one - https://github.com/kubernetes-sigs/headlamp/pull/3977

illume:
It seems this was continued in:
- #3977

Which is merged now, so closing this one.

(@vyncent-t  Please reopen if I made a mistake)

PR #3766, opened 2025-08-10
backend: cmd/pagination: add pagination API for getting limited resource size 

upsaurav12:
this add new pagination API which would help to send only request that the user is veiwing only not just full resource list.


iamundefinedguy:
@yolossn @illume I have tested this with POSTMAN client tool where i have to add ``continueToken`` for the next batch of resource list which means it is working as expected. 

only thing we would need is to set ``continueToken`` in frontend .

iamundefinedguy:
@illume @yolossn the only problem in this approach would be that if there are > 1000's of pods/nodes or other heavy resources it would take 800ms for returning that paginating data.

iamundefinedguy:
@illume @yolossn I am trying to implement the pagination part but here the problem is the it is somehow still taking 800ms or even 1s taking from the backend server. I checked in postman it is giving response time like 10ms etc. 


Here in this code i am adding query that will go to backend server which results adding ``p`` unneccesarily into URL. 
for example if i am on secrets page it should add /api/v1/secrets?p=1,2,3,4 etc but it is also adding /api/v1/namespace=5,6,7 even mutating pageNo with many resources here is the screenshot from network tab, it seems backend is working fine but rather frontend is doing more work 
Screenshot 2025-10-19 165738


I think i need to manage the pageState in such a way that it should send pageNo to backend when the user is going to next or prev page

PR #3771, opened 2025-08-11
frontend: app: a11y: Add faster dev mode alternative with rsbuild, fix loader

illume:
## Add faster dev mode alternative with rsbuild

Before: app in dev mode takes over a minute to load app on windows cmd
After: app loads in seconds

This adds an alternative dev mode to app:
```shell
make frontend backend
cd app
npm run star
```

Also for the frontend:
```shell
cd frontend
npm run star
```

It leaves the previous "npm start" there, and leaves the build
and test systems as before.

This uses rsbuild which is a webpack compatible build tool that
does not send 850 files in dev mode.

Rather than minutes or 10s of seconds it is closer to the speed
of the compiled app or the previous webpack based dev mode.

### How to test

- `cd frontend/ && npm run star`
- `cd app/ && npm run star`
  - compare app speed of dev mode on a slow windows computer in cmd
  - even on a fast new mac laptop, it is seconds quicker
- This still works: `cd frontend/ && npm run start`
- win app still works `make frontend backend app-win`
- mac app `make frontend backend app-mac`
- container e2e tests still work

## Loader improvements

For users with fast machines the loader now doesn't flash for a second or under. It also now respects motion preferences to not animate if the user asked for that. For users with slower connections/machines the loader should still display.

- frontend/index.html: Fix a11y issue with loader. Gives loader a label.
- frontend/index.html: Fix a11y loader to respect motion preferences
- frontend/index.html: Fix animation loader animation to start later

### How to test

- Run the app with `npm run start` and compare it with `npm run star`. With star you should not see any loader. With start, it may be visible but later. Compare to main branch where you should see the loader.
- Change the motion preference to reduced
- For users with a slow connection, you should see the loader after about 1 second. Set the dev console network tab networking to 4g speed or so. This way it should take longer to load and you should see the loader appear after about 1 second longer than before.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -92,7 +92,7 @@
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root">
       <div class="splash">
-        <span class="loader"></span>
+        <span class="loader" role="status" aria-live="polite" aria-label="Loading"></span>
       </div>
     </div>
     <!--


vyncent-t:
looks like this works, builds with warnings about two things but other than that works! 
image

illume:
Thanks @vyncent-t 

- The first warning I tried a few things, but couldn't resolve it. The editor works though... so I think it's fine for now.
- The second warning is for a function that's not used anymore I think for 4 years. So I left the warning there. But I'll wait till Joaquim comes back before removing that function probably.

iamundefinedguy:
@illume I had tested the PR using codespaces which was running correctly or fine for me. but i will try to run ``saparately`` on windows systems also to give actual output as well
If it is running both ``frontend`` and ``backend`` using a single command then it would definetely  increase the development speed.

skoeva (approved):
Tested and works very well, good alternative for users with slower machines

PR #3776, opened 2025-08-12
frontend: MapView: Build MapView for Gateway API

userAdityaa:
### PR Description

This PR introduces a **MapView for the Gateway API**.

* Added a dedicated **Gateway API source option** in the source picker.
* Updated the **icon for Gateway API** in the source picker for better clarity.
* Built the **resource relationship mapping** for Gateway API objects (Gateways, Routes, etc.).
* Implemented **node-to-details linking**, so selecting a node opens its details section.

This enables users to visualize and explore Gateway API resources directly within the MapView.

---

### How to Test

1. Navigate to the **Map** section from the Headlamp sidebar.
2. Open the **Source Picker** and select **Gateway API** as the source.

   * Ensure that Gateway API resources are already applied in the cluster (Gateways, Routes, etc.), otherwise no nodes will show.
3. Verify that:

   * Nodes for Gateway API resources are displayed on the map.
   * Relationships between resources are correctly shown.
   * Clicking on a node opens the correct **details section**.
4. Confirm that the **icon for Gateway API** is correctly displayed in the source picker.

---

### Screenshots of the changes
Screenshot 2025-08-18 at 1 54 10 PM
Screenshot 2025-08-18 at 1 54 20 PM
Screenshot 2025-08-18 at 1 54 28 PM
Screenshot 2025-08-18 at 1 54 39 PM


frontend/src/components/resourceMap/sources/definitions/sources.tsx
```diff
@@ -174,6 +182,20 @@ export function useGetAllSources(): GraphSource[] {
         // leases
       ],
     },
+    {
+      id: 'gateway-api',
+      label: 'Gateway API',
+      icon: <img src={gatewayIcon} alt="gateway-api" style={{ width: '120%', height: '100%' }} />,
```
sniok: let's keep the icon consistent with the sidebar 

Image

`<Icon icon='mdi:lan-connect' />`

userAdityaa: Thank you, should I use that for all the nodes as well ?

frontend/src/components/resourceMap/sources/definitions/sources.tsx
```diff
@@ -174,6 +182,20 @@ export function useGetAllSources(): GraphSource[] {
         // leases
       ],
     },
+    {
+      id: 'gateway-api',
```
sniok: let's also make sure it's not selected by default

```
     isEnabledByDefault: false,
```

frontend/src/components/resourceMap/sources/definitions/sources.tsx
sniok: just for the 'Gateway API' item

sniok: oh also let's keep the name consistent and name it "Gateway (beta)" as well

PR #3778, opened 2025-08-12
fix: Handle CRDs with identical plural names

pablogarciamiranda:
## Summary

This PR fixes the routing conflict that occurs when multiple Custom Resource Definitions (CRDs) have the same plural name but belong to different API groups.

## Related Issue

Fixes #ISSUE_NUMBER  

## Changes

- Added `group` parameter to CRD routes
- Updated components to use `{plural}.{group}` format for CRD identification
- Modified CRD list to show API group in parentheses
- Fixed navigation links to include both group and plural name

## Steps to Test

1. Create two CRDs with the same plural name but different groups
2. Navigate to the CRDs list
3. Click on each CRD and verify correct resources are shown
4. Check that navigation between CRDs works properly

## Notes for the Reviewer

- Backwards compatible with existing CRDs
- Group now visible in CRD list for better identification


sniok:
Correct me if I'm wrong but the group is already included in the CRD name, so I'm not sure if this change is needed.

image

Could you share a more detailed way to reproduce a problem you're experiencing? The issue number you included in the description just points to this PR

PR #3780, opened 2025-08-12
frontend: Remove unused dependencies

skoeva:
These changes remove the unused dependencies in the frontend directory, sourced using `depcheck`.

### Testing
- [X] No issues running `npm run build`, `npm run test`, `npm run start`


illume:
Did you check the docs browser?

skoeva:
@illume They're all transitive dependencies that aren't used directly in those directories, so there isn't really a reason to keep them listed directly; they aren't actually deleted from the `package-lock.json` files, they're just removed from the list

illume:
@skoeva I'm wondering if the reason they were added is still valid. Did you look back to see why they were added?

I expect it was to fix some npm related dependency issue. Probably it doesn't apply anymore, but it would be good to know the reason.

skoeva:
@illume some were never actually used (i.e. `@dagrejs/dagre`, `base64-arraybuffer`, etc.) and some were previously used but no longer (i.e. `react-jwt`). ~~that last one specifically still exists as an artifact in the headlamp-plugin package though (don't think that's intentional) so opened an issue for it [here](https://github.com/kubernetes-sigs/headlamp/issues/3801)~~

PR #3784, opened 2025-08-13
Headlamp backstage support

yolossn:
This PR contains the changes that are required for the Backstage plugin.

The changes include

1. Configuring the default state of the sidebar during frontend build based on `REACT_APP_HEADLAMP_SIDEBAR_DEFAULT_OPEN` env variable
2. Add embedded Single page handler that bundles the frontend static files to the binary and serves it from the embedded files
3. Add make recipes for building embedded headlamp for multiple platforms and compress it
4. Add message receiver to frontend for receiving the messages sent from the parent (backstage).
5. Add Backstage auth_token to requests if it exists in localStorage.

### DIFF
--- a/frontend/src/helpers/backstageMessageReceiver.ts
+++ b/frontend/src/helpers/backstageMessageReceiver.ts
@@ -84,7 +84,7 @@ async function storeKubeconfigFromBackstage(kubeconfig: string) {
         await statelessFunctions.findAndReplaceKubeconfig(context.name, newKubeconfigBase64, true);
       }
     );
-
+    console.log('Promises', promises);
     // Wait for all kubeconfig operations to complete
     await Promise.all(promises);
   } catch (error) {
--- a/backstage-test/README.md
+++ b/backstage-test/README.md
@@ -1,0 +1,68 @@
+# Backstage Integration Testing Guide
+
+This guide provides a simplified way to test the core features of the Headlamp-Backstage integration without setting up a complete Backstage environment with OIDC authentication and managed Kubernetes from a cloud provider.
+
+## Prerequisites
+
+- Go development environment
+- Python 3 or Node.js
+- Valid Kubernetes configuration file (without OIDC/exec-based authentication)
+
+## Testing Steps
+
+### 1. Build the Headlamp Embedded Binary
+
+From the project root directory, run:
+
+```bash
+make backend-embed
+```
+
+### 2. Run the Headlamp Server
+
+Execute the binary with the required parameters:
+
+```bash
+./backend/headlamp-server --base-url="/api/headlamp" --enable-dynamic-clusters
+```
+
+### 3. Serve the Frontend
+
+Start a local web server to serve the `index.html` file. You can use either:
+
+**Option A: Python HTTP Server**
+```bash
+python3 -m http.server -p 8000
+```
+
+**Option B: Node.js HTTP Server**
+```bash
+npx http-server -p 8000
+```
+
+### 4. Access the Application
+
+Open your web browser and navigate to:
+```
+http://localhost:8000
+```
+
+## Testing the Integration
+
+### Test 1: Backstage Token Authentication
+
+1. Enter any random text in the Backstage authentication token field
+2. Click "Share Backstage Token"
+3. **Verification**: Check if the requests made include the `x-backstage-token` header
+
+### Test 2: Kubernetes Configuration
+
+1. Base64 encode a valid kubeconfig file (ensure it doesn't use OIDC/exec-based authentication)
+2. Click "Share Kubeconfig"
+3. **Verification**: A new cluster should appear in the interface
+
+## Troubleshooting
+
+- Ensure the Headlamp server is running before accessing the test frontend
+- Check that your kubeconfig is valid and doesn't contain OIDC or exec-based authentication
+- Verify that the `--enable-dynamic-clusters` flag is set when starting the server
--- a/backstage-test/index.html
+++ b/backstage-test/index.html
@@ -1,0 +1,222 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Headlamp Backstage Integration</title>
+    <style>
+        body {
+            font-family: Arial, sans-serif;
+            margin: 20px;
+            background-color: #f5f5f5;
+        }
+        .container {
+            max-width: 1200px;
+            margin: 0 auto;
+        }
+        .header {
+            background-color: #fff;
+            padding: 20px;
+            border-radius: 8px;
+            margin-bottom: 20px;
+            box-shadow: 0 2px 4px rgba(0,0,0,0.1);
+        }
+        .controls {
+            background-color: #fff;
+            padding: 20px;
+            border-radius: 8px;
+            margin-bottom: 20px;
+            box-shadow: 0 2px 4px rgba(0,0,0,0.1);
+        }
+        .input-group {
+            margin-bottom: 15px;
+        }
+        .input-group label {
+            display: block;
+            margin-bottom: 5px;
+            font-weight: bold;
+            color: #333;
+        }
+        .input-group input {
+            width: 100%;
+            padding: 10px;
+            border: 1px solid #ddd;
+            border-radius: 4px;
+            font-size: 14px;
+            box-sizing: border-box;
+        }
+        .button-group {
+            display: flex;
+            gap: 10px;
+            flex-wrap: wrap;
+        }
+        button {
+            background-color: #007bff;
+            color: white;
+            border: none;
+            padding: 12px 24px;
+            border-radius: 4px;
+            cursor: pointer;
+            font-size: 14px;
+            transition: background-color 0.3s;
+        }
+        button:hover {
+            background-color: #0056b3;
+        }
+        button:disabled {
+            background-color: #6c757d;
+            cursor: not-allowed;
+        }
+        .iframe-container {
+            background-color: #fff;
+            padding: 20px;
+            border-radius: 8px;
+            box-shadow: 0 2px 4px rgba(0,0,0,0.1);
+        }
+        iframe {
+            width: 100%;
+            height: 600px;
+            border: 1px solid #ddd;
+            border-radius: 4px;
+        }
+        .status {
+            margin-top: 10px;
+            padding: 10px;
+            border-radius: 4px;
+            font-size: 14px;
+        }
+        .status.success {
+            background-color: #d4edda;
+            color: #155724;
+            border: 1px solid #c3e6cb;
+        }
+        .status.error {
+            background-color: #f8d7da;
+            color: #721c24;
+            border: 1px solid #f5c6cb;
+        }
+    </style>
+</head>
+<body>
+    <div class="container">
+        <div class="header">
+            <h1>Headlamp Backstage Integration</h1>
+            <p>Share your Backstage authentication token and kubeconfig with the Headlamp application.</p>
+        </div>
+
+        <div class="controls">
+            <div class="input-group">
+                <label for="backstageToken">Backstage Authentication Token:</label>
+                <input type="text" id="backstageToken" placeholder="Enter your Backstage auth token">
+            </div>
+            
+            <div class="input-group">
+                <label for="kubeconfig">Kubeconfig:</label>
+                <input type="text" id="kubeconfig" placeholder="Enter your kubeconfig content">
+            </div>
+
+            <div class="button-group">
+                <button onclick="shareBackstageToken()">Share Backstage Token</button>
+                <button onclick="shareKubeconfig()">Share Kubeconfig</button>
+                <button onclick="clearInputs()">Clear Inputs</button>
+            </div>
+
+            <div id="status" class="status" style="display: none;"></div>
+        </div>
+
+        <div class="iframe-container">
+            <h3>Headlamp Application</h3>
+            <iframe 
+                id="headlampIframe" 
+                src="http://localhost:4466/api/headlamp/"
+                title="Headlamp Application">
+            </iframe>
+        </div>
+    </div>
+
+    <script>
+        const iframeRef = document.getElementById('headlampIframe');
+        const targetOrigin = 'http://localhost:4466/api/headlamp';
+
+        function showStatus(message, isSuccess = true) {
+            const statusDiv = document.getElementById('status');
+            statusDiv.textContent = message;
+            statusDiv.className = `status ${isSuccess ? 'success' : 'error'}`;
+            statusDiv.style.display = 'block';
+            
+            // Hide status after 3 seconds
+            setTimeout(() => {
+                statusDiv.style.display = 'none';
+            }, 3000);
+        }
+
+        function shareBackstageToken() {
+            const token = document.getElementById('backstageToken').value.trim();
+            
+            if (!token) {
+                showStatus('Please enter a Backstage authentication token', false);
+                return;
+            }
+
+            try {
+                iframeRef.contentWindow.postMessage(
+                    {
+                        type: 'BACKSTAGE_AUTH_TOKEN',
+                        payload: { token: token },
+                    },
+                    targetOrigin,
+                );
+                showStatus('Backstage token shared successfully!');
+            } catch (error) {
+                showStatus('Error sharing Backstage token: ' + error.message, false);
+            }
+        }
+
+        function shareKubeconfig() {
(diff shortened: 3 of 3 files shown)


sniok:
so does this mean that the entire frontend can now be included in the backend binary? should we do this in all cases (app/in-cluster) by default? this would probably simplify logic a lot

yolossn:
> so does this mean that the entire frontend can now be included in the backend binary? should we do this in all cases (app/in-cluster) by default? this would probably simplify logic a lot

Yes the entire frontend is included in the binary. We can eventually move this to default way of bundling.

frontend/src/lib/k8s/api/v1/clusterApi.ts
```diff
@@ -59,6 +60,7 @@ export async function testClusterHealth(cluster?: string) {
 
 export async function setCluster(clusterReq: ClusterRequest) {
   const kubeconfig = clusterReq.kubeconfig;
+  const headers = helpers.addBackstageAuthHeaders(JSON_HEADERS);
```
illume: Please import directly from your file and not the whole helpers, otherwise you import everything in helpers and it messes up imports.

Makefile
```diff
@@ -66,6 +69,122 @@ app-tsc:
 backend:
 	cd backend && go build -o ./headlamp-server${SERVER_EXE_EXT} ./cmd
 
+.PHONY: backend-embed
```
illume, line 72: Can you please add these builds to be run in CI? 
(bonus points if you can see if the builds are working)

frontend/src/components/App/Home/index.tsx
```diff
@@ -18,6 +18,7 @@ import { isEqual } from 'lodash';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router-dom';
+import helpers from '../../../helpers';
```
illume: Please never import helpers, but import only the thing you need. A full helpers import messes up imports that affect speed and even breaking compilation through import cycles.

frontend/src/stateless/index.ts
```diff
@@ -264,6 +266,46 @@ function findMatchingContexts(
...
 }
 
+/**
+ * Finds and replaces a kubeconfig by cluster name.
+ * @param clusterName - The name of the cluster to find and replace.
+ * @param kubeconfig - The base64 encoded kubeconfig to replace the existing one with.
+ * @param create - If true, create a new kubeconfig if it doesn't exist. If false, only replace existing kubeconfigs.
+ * @returns A promise that resolves when the kubeconfig is successfully replaced.
+ */
+export function findAndReplaceKubeconfig(
```
illume: A test for this would be nice.

backend/cmd/headlamp.go
```diff
@@ -166,6 +168,76 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+// embeddedSpaHandler serves the static files embedded in the binary.
+type embeddedSpaHandler struct {
+	staticFS  embed.FS
+	indexPath string
+	baseURL   string
+}
+
+// ServeHTTP serves the static files embedded in the binary.
+func (h embeddedSpaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
illume: We don't want to add anything new to headlamp.go

Instead it should go in the appropriate package, and if it needs to be called from headlamp.go then it can be imported there. (The less code in headlamp.go the better).

backend/cmd/headlamp.go
```diff
@@ -166,6 +168,76 @@ func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	http.ServeFile(w, r, path)
 }
 
+// embeddedSpaHandler serves the static files embedded in the binary.
+type embeddedSpaHandler struct {
+	staticFS  embed.FS
+	indexPath string
```
illume: I'm not sure what this field is. Can you please document all your fields?

frontend/src/helpers/addBackstageAuthHeaders.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function addBackstageAuthHeaders(headers: HeadersInit = {}): HeadersInit {
+  const token = getBackstageToken();
+
+  if (!token) {
+    return headers;
+  }
+
+  // If headers is already a Headers object, create a new one and copy all headers
+  if (headers instanceof Headers) {
```
illume, line 33: I'm wondering how this could be moved into the backstage plugin?

backend/cmd/headlamp.go
yolossn: moved it to separate package and added comments

backend/cmd/headlamp.go
yolossn: created a separate package for SPA handlers

frontend/src/stateless/index.ts
yolossn: Done

frontend/src/helpers/addBackstageAuthHeaders.ts
yolossn: Nope this is not possible because the backstage plugin only loads the headlamp frontend in iframe.

Makefile
yolossn: Added CI for embed binary builds

backend/pkg/exec/exec_test.go
```diff
@@ -609,7 +609,7 @@ func TestRefreshCreds(t *testing.T) {
 					"expirationTimestamp": "2006-01-02T15:04:05Z"
 				}
 			}`,
-			wantExpiry: time.Date(2006, 01, 02, 15, 04, 05, 0, time.UTC),
+			wantExpiry: time.Date(2006, 0o1, 0o2, 15, 0o4, 0o5, 0, time.UTC),
```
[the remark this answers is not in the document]
yolossn: Unrelated to the PR so removed

frontend/src/stateless/index.ts
```diff
@@ -264,6 +266,46 @@ function findMatchingContexts(
...
+        // If not found and create is false, reject
+        reject(
+          new Error(`No existing kubeconfig found for cluster ${clusterName} and create is false`)
+        );
+      }
+    } catch (error) {
+      console.error('Error in findAndReplaceKubeconfig:', error);
+      reject(error);
+    }
+  });
```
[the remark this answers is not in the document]
yolossn: fixed

frontend/src/lib/k8s/api/v1/portForward.ts
```diff
@@ -72,9 +73,7 @@ export async function startPortForward(
   id: string = ''
 ): Promise<PortForward> {
   const kubeconfig = await findKubeconfigByClusterName(cluster);
-  const headers: HeadersInit = new Headers({
-    ...JSON_HEADERS,
-  });
+  const headers = new Headers(addBackstageAuthHeaders(JSON_HEADERS));
```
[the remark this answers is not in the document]
yolossn: the header cannot be reused in this case

yolossn:
Added a simplified way to test the token and kubeconfig sharing from backstage to headlamp by creating a mock setup. The steps to test is added in `backstage-test/README.md`

image

frontend/src/components/App/AppContainer.tsx
```diff
@@ -31,6 +31,39 @@ window.desktopApi?.receive('backend-token', (token: string) => {
...
+  const location = useLocation();
+
+  useEffect(() => {
+    // Get the current URL search params
+    const searchParams = new URLSearchParams(location.search);
+
+    // Check if 'to' parameter exists
+    const redirectPath = searchParams.get('to');
+
+    if (redirectPath) {
```
illume, line 85: This would be an "open redirect". If a malicious link like ?to=https://evil.example.com is opened, the app could redirect users off-site.

suggestion:
`if (redirectPath && redirectPath.startsWith('/')) {`

frontend/src/components/App/AppContainer.tsx
```diff
@@ -31,6 +31,39 @@ window.desktopApi?.receive('backend-token', (token: string) => {
   setBackendToken(token);
 });
 
+/**
+ * QueryParamRedirect is a component that checks for a 'to' query parameter and redirects accordingly
+ * This should be placed near the top of your component hierarchy,
+ * typically in your main App component
+ * @returns
```
illume: Something is missing from "returns" here.

frontend/src/components/App/AppContainer.tsx
```diff
@@ -31,6 +31,39 @@ window.desktopApi?.receive('backend-token', (token: string) => {
   setBackendToken(token);
 });
 
+/**
+ * QueryParamRedirect is a component that checks for a 'to' query parameter and redirects accordingly
```
illume, line 69: Could you please add a section to here about when it is used? 
(I assume it's used by backstage somehow? Can you please let me know how?)

The comment in the git commit would be useful here too.
> Example: /notifications?to=/settings/general redirect to /settings/general

frontend/src/helpers/addBackstageAuthHeaders.ts
```diff
@@ -0,0 +1,51 @@
...
+import { getBackstageToken } from './backstageMessageReceiver';
+
+/**
+ * Adds backstage authentication token to headers if available
+ *
+ * @param headers - optional existing headers to add the token to
+ * @returns headers with backstage token if available
+ */
+export function addBackstageAuthHeaders(headers: HeadersInit = {}): HeadersInit {
+  const token = getBackstageToken();
```
illume, line 26: Either inside addBackstageAuthHeaders, or at the call site of addBackstageAuthHeaders: it should only do things if it's running isBackstage?

I guess getBackstageToken handles this?

illume (changes requested):
Thanks for all those changes.

I left a few notes.

(Btw. Can you please add to the commit messages the part they changed other than just "backend" or "frontend" ?  Also we don't require "Signed-off-by" in the commit messages in the headlamp repo anymore.)

frontend/src/helpers/addBackstageAuthHeaders.ts
illume: Ok, it does. So no backstage stuff is added without backstage being there. This is good enough, I'll mark this as resolved.

frontend/src/plugin/index.ts
```diff
@@ -286,12 +287,17 @@ function handlePluginRunError(error: unknown, packageName: string, packageVersio
  * @param delay The initial delay in milliseconds.
  * @returns A promise that resolves to the response of the fetch request.
  */
-async function fetchWithRetry(url: string, retries = 8, delay = 50): Promise<any> {
+async function fetchWithRetry(
+  url: string,
+  headers: HeadersInit,
+  retries = 8,
+  delay = 50
+): Promise<any> {
```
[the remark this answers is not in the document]
illume: This is an internal function, so this is not an issue.

frontend/src/stateless/index.ts
```diff
@@ -16,6 +16,8 @@
 
 import * as jsyaml from 'js-yaml';
 import _ from 'lodash';
+import { addBackstageAuthHeaders } from '../helpers/addBackstageAuthHeaders';
+import { JSON_HEADERS } from '../lib/k8s/api/v1/constants';
```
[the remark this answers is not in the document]
illume: We don't do mid program imports.

frontend/src/components/Sidebar/sidebarSlice.test.ts
```diff
@@ -108,6 +110,12 @@ describe('setInitialSidebarOpen', () => {
     Object.defineProperty(window, 'innerWidth', {
       value: originalInnerWidth,
     });
+    // Restore original environment variable
+    if (originalEnv !== undefined) {
+      (import.meta.env as any).REACT_APP_HEADLAMP_SIDEBAR_DEFAULT_OPEN = originalEnv;
+    } else {
+      delete (import.meta.env as any).REACT_APP_HEADLAMP_SIDEBAR_DEFAULT_OPEN;
+    }
```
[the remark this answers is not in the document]
illume: The code works ok... I think it's fine.

Makefile
```diff
@@ -66,6 +69,122 @@ app-tsc:
...
+	$(MAKE) backend-embed-linux VERSION=$(VERSION)
+	@echo "All builds completed successfully for version $(VERSION)!"
+
+.PHONY: backend-embed-all-compressed
+backend-embed-all-compressed: backend-embed-all
+	@echo "Compressing all binaries with version: $(VERSION)..."
+	cd backend/dist && for file in *; do \
+		if [ -f "$$file" ] && [[ ! "$$file" == *.tar.gz ]]; then \
+			tar -czf "$$file.tar.gz" "$$file" && \
+			rm "$$file"; \
```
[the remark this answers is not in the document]
illume: This works in bash and zsh. I think that's good enough.

frontend/src/components/App/AppContainer.tsx
yolossn: The Backstage plugin provides an "Open in Headlamp" button.
When clicked, the user is redirected to the Backstage page containing a Headlamp iframe. The iframe URL includes a redirect query parameter (to), which Headlamp uses to navigate the user to the correct page or resource (e.g., a Deployment or other Kubernetes resource). I didn't include this flow in the commit message because this redirect query param can be used by plugins and other places.

In short: clicking "Open in Headlamp" on a resource in Backstage opens that resource directly in Headlamp.

backend/pkg/spa/embeddedSPAHandler.go
```diff
@@ -0,0 +1,94 @@
...
+			http.Error(w, "Unable to read index file", http.StatusInternalServerError)
+			return
+		}
+	}
+
+	// if request is for / or index.html replace the headlampBaseUrl with the baseURL in the index.html file
+	if path == h.indexPath || path == "/"+h.indexPath || path == "/"+h.indexPath+"/" {
+		content = bytes.ReplaceAll(content, []byte("headlampBaseUrl = './"), []byte("headlampBaseUrl = '"+h.baseURL+""))
+		content = bytes.ReplaceAll(content, []byte("'./'"), []byte(h.baseURL+"/"))
+		content = bytes.ReplaceAll(content, []byte("url("), []byte("url("+h.baseURL+"/"))
```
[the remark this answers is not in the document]
yolossn: Added a check to replace only when the baseURL is configured.

frontend/src/lib/k8s/api/v1/portForward.ts
```diff
@@ -139,17 +138,16 @@ export async function stopOrDeletePortForward(
...
   // This means cluster is dynamically configured.
   if (kubeconfig !== null) {
-    headers['X-HEADLAMP-USER-ID'] = getUserIdFromLocalStorage();
+    headers.set('X-HEADLAMP-USER-ID', getUserIdFromLocalStorage());
   }
 
   return clusterFetch(`/portforward`, {
     method: 'DELETE',
+    headers: headers,
     body: JSON.stringify({
```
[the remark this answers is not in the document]
yolossn: The header is being used in the clusterFetch so closing this.

frontend/src/components/App/AppContainer.tsx
yolossn: Added a validation for the path

frontend/src/stateless/index.ts
```diff
@@ -264,6 +266,42 @@ function findMatchingContexts(
...
+  kubeconfig: string,
+  create: boolean = false
+): Promise<void> {
+  try {
+    // First try to find the existing kubeconfig
+    const existingKubeconfig = await findKubeconfigByClusterName(clusterName);
+
+    if (existingKubeconfig) {
+      // If found, delete the old one
+      await deleteClusterKubeconfig(clusterName);
```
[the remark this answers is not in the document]
yolossn: Handled this

frontend/src/helpers/backstageMessageReceiver.ts
```diff
@@ -0,0 +1,148 @@
...
+          users: [authInfo],
+        };
+
+        // Convert back to YAML and base64 encode
+        const newKubeconfigYaml = jsyaml.dump(newKubeconfig, { lineWidth: -1 });
+        const newKubeconfigBase64 = btoa(newKubeconfigYaml);
+        await statelessFunctions.findAndReplaceKubeconfig(context.name, newKubeconfigBase64, true);
+      }
+    );
+    console.log('Promises', promises);
```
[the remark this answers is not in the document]
yolossn: removed this

illume:
Weird, it says there’s a conversation open so I can’t merge. But I don’t see one open… do you see one @yolossn ?

illume:
Not to worry, found them. They were hidden under an outdated link.

Issue #3769, opened 2025-08-10
Button to copy value from Secret


jimmyjones2:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
Some charts generate a password and store it within a Secret, e.g. for logging onto the web interface deployed by the chart. Currently I need to click to reveal the value, select the whole string, then copy it.

## Describe the solution you'd like
A button to copy the value of the key within the Secret into the clipboard

## What users will benefit from this feature?
Helm chart users with password authenticated apps

## Are you able to implement this feature?
No

## Additional context

e.g. could look like:

<img width="565" height="286" alt="Image" src="https://github.com/user-attachments/assets/d2daaa45-e385-41e8-addd-6450a38fd472" />

skoeva:
looking into this

PR #3787, opened 2025-08-14
frontend: Secret: Add button to copy to clipboard

skoeva:
This change adds a button in the secret details page allowing users to copy the secret value to clipboard.

Fixes: [#3769](https://github.com/kubernetes-sigs/headlamp/issues/3769)

### Testing
- [X] Navigate to the details page for a secret
- [X] Click the copy button and ensure the tooltip confirms the value is copied
- [X] Paste anywhere and ensure the value is copied

image


frontend/src/components/common/Resource/Resource.tsx
```diff
@@ -501,15 +501,47 @@ export function DataField(props: DataFieldProps) {
...
+    } catch (err) {
+      console.error(`Error copying secret to clipboard: ${err}`);
+    }
+  }
+
+  const tooltipTitle = copied ? t('translation|Copied') : t('translation|Copy to clipboard');
+  const copyButton: JSX.Element = (
+    <IconButton
+      edge="start"
+      aria-label={t('copy to clipboard')}
```
illume: Can this be "translation|Copy to clipboard" so there's one less translation key?

skoeva: done

PR #3795, opened 2025-08-15
app/scripts/esrp.js: Replace execSync with execFileSync

skoeva:
This change prevents the shell from interpreting special characters in the arguments.

Follow-up from https://github.com/kubernetes-sigs/headlamp/pull/3454 by @odaysec


illume:
I'll leave this one for @joaquimrocha

PR #3796, opened 2025-08-17
fontend: Loader: Fix to not freeze when main thread freezes

illume:
This avoids SVG that mui CircleProgress uses, so it doesn't freeze.

Also avoids flashing if the loading is finished before 0.25 seconds.

Fixes a11y issue to not use progressbar, because we always do not
know the progress, but instead are announcing a status.


frontend/src/components/common/Loader.tsx
```diff
@@ -15,17 +15,169 @@
...
+  /** Size of the loader. Defaults to 40. */
+  size?: number;
+  /** Color of the loader. Defaults to primary color. */
+  color?: 'primary' | 'secondary' | 'error' | 'info' | 'success' | 'warning' | string;
+}
+
+/**
+ * Injects custom loader styles into the document, once for each theme.
+ */
+function injectLoaderStyle(theme: any) {
```
sniok, line 40: why do styles need to be injected like this?

illume: What’s your problem with it?

sniok: why not `<Box sx={...} />` ?

sniok: we're using react so we don't have to deal with manually creating dom nodes/styles. 
this kind of manual style injection is going to be harder to maintain, prone to bugs, less readable and unnecessary since we have nice abstractions like sx prop. or alternatively you could write a `.css` file and import it. 
I wanted to ask you first, maybe you had some reason to apply styles like this

PR #3809, opened 2025-08-20
frontend: Add Namespace based projects

sniok:
This PR introduces Projects feature

Project is a collection of kubernetes resources, used for organizing deployed applications or any other workload.

Testing done
Manual testing
A11y fast pass check

### DIFF
--- a/frontend/src/components/resourceQuota/Details.tsx
+++ b/frontend/src/components/resourceQuota/Details.tsx
@@ -21,6 +21,48 @@ import { compareUnits, normalizeUnit } from '../../lib/util';
 import { DetailsGrid } from '../common/Resource';
 import SimpleTable from '../common/SimpleTable';
 
+export function ResourceQuotaTable({
+  resourceStats,
+}: {
+  resourceStats: {
+    name: string;
+    hard: string;
+    used: string;
+  }[];
+}) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/App/Home/index.tsx
```diff
@@ -103,34 +110,61 @@ function HomeComponent(props: HomeComponentProps) {
...
+      </>
     ),
     [customNameClusters, errors, versions, warningLabels]
   );
 
-  return memoizedComponent;
+  return (
+    <PageGrid>
+      <SectionBox title="Home" headerProps={{ headerStyle: 'main' }}>
+        <Box sx={{ borderBottom: 1, borderColor: 'divider' }}>
```
joaquimrocha: We need a gap between the tabs and the clusters table **when** there are no clusters. Maybe a `marginBottom: 1` here would work but would affect also when there are clusters and I am not sure yet if it would look good with that.
image

joaquimrocha (commented):
Left a nit about the UX. However, I also don't see how we can just make a namespace as a project without deploying new YAML. I only see the option of creating from YAML.

joaquimrocha:
Thanks. Tested and it works as intended.
/lgtm
/approve

PR #3817, opened 2025-08-21
frontend: plugins: Bump sha.js to 2.4.12

skoeva:
This change updates the sha.js dependency in the frontend and plugins directories to the desired 2.4.12 version.


illume (approved):
🎉 thanks!

Issue #1103, opened 2023-04-25
Headlamp Icons does not load in isolated environment.


jgyy:
I am trying to host headlamp in cluster on an isolated environment without internet access. However the frontend dashboard is trying to load the following icons from the internet (links below) which will not work at all if I am accessing the dashboard from an isolated environment without internet access.

Can somebody look into it and make the icons load as assets locally instead of trying to fetch a public icon?  Not just icons, any other public CDN will not work in isolated environment. Thanks!

https://api.iconify.design/mdi.json
https://api.simplesvg.com/mdi.json
https://api.unisvg.com/mdi.json

illume:
Hello.There are a number of fixes for this merged into main. We are going to do another release this week (hopefully).best regards,On 25. Apr 2023, at 10:33, jgyy ***@***.***> wrote:﻿
I am trying to host headlamp in cluster on an isolated environment without internet access. However the frontend dashboard is trying to load the following icons from the internet (links below) which will not work at all if I am accessing the dashboard from an isolated environment without internet access.
Can somebody look into it and make the icons load as assets locally instead of trying to fetch a public icon?  Not just icons, any other public CDN will not work in isolated environment. Thanks!
https://api.iconify.design/mdi.json
https://api.simplesvg.com/mdi.json
https://api.unisvg.com/mdi.json

jgyy:
I hope the list of fixes also includes having local icon assets though. thanks anyway

joaquimrocha:
@jgyy , thanks. Closing this as it should be now available as mentioned.
If you spot any missing icon, please let us know. We will also work on a way to detect any missing icons from the cache in #1109 .

PR #3818, opened 2025-08-22
Make monaco-editor work in an offline/air-gapped environment and fix localisation

DanielLeone:
## Summary

This PR fixes the Monaco Editor not working in an air-gapped environment (no internet). By default the monaco-react package loads resources using the `jsdeliver.net` CDN, which I think is no good (we use Headlamp heavily in air-gapped environments at my workplace).

You can see the JSDeliver default config here: https://github.com/suren-atoyan/monaco-loader/blob/fc86117fb2f221070a509aa4539dd4644a607381/src/config/index.js#L3

I saw a previous issue related to air gapped environments (https://github.com/kubernetes-sigs/headlamp/issues/1103), so I assume this will be regarded as a bug here as well?

## Changes

- I've configured the Monaco Editor loader so it loads the editor files from the assets folder, instead of importing the ESM module - this was the only way I could find to get the localisation working, and also supporting air-gapped environments.

## Steps to Test

1. Turn off internet
2. Load headlamp
3. All the Monaco Editor instances should still work
4. Changing the locale setting should work in the monaco editor as well, but you'll have to reload the page for it to take effect if you've already loaded a single monaco-editor instance

image

## Notes for the Reviewer

If you set your headlamp language to one of the supported languages it should apply to the monaco editor now - see screenshot above. I'm sure that currently this does not work in master.

This is what we did have in master, but it doesn't work as you expect.

```
loader.config({ 'vs/nls': { availableLanguages: { '*': lang } }, monaco });
```

1. You can only call `loader.config()` before the Editor has loaded, calling it after the Editor as loaded has no effect, so we can't change language this way.
3. When you supply the `monaco` object to `loader.config` it ignores the `vs/nls` option, as it's saying you've given the `monaco`  module already, there's nothing more to load. This is the ESM module format we were using.

Here's a sandbox of the locale setting working: https://codesandbox.io/p/sandbox/hlcpwz
But you'll notice the radio button doesn't effect the language once it's loaded - only the initial state value does.
This is also loading in AMD format, hence the localisation setting does work.

Currently:
* Using ESM monaco-editor bundled via Vite, workers are loaded via CDN
* No language support because of that

Current PR:
* Changes to the AMD format bundled monaco-editor to work offline instead of via CDN

Previous commit in this branch
Keeps using the ESM module and loads the worker files locally, but this means language support is still not working.

Keen for anyones thoughts

I've tested macos electron, native macos vite dev server, and the default docker image, we should probably test on all other variants and with custom PUBLIC_URLs as well, if we're happy with this approach.

### DIFF
--- a/frontend/src/components/common/Resource/EditorDialog.tsx
+++ b/frontend/src/components/common/Resource/EditorDialog.tsx
@@ -15,7 +15,7 @@
  */
 
 import '../../../i18n/config';
-import Editor, { loader } from '@monaco-editor/react';
+import Editor from '@monaco-editor/react';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import DialogActions from '@mui/material/DialogActions';
@@ -27,7 +27,6 @@ import Switch from '@mui/material/Switch';
 import Typography from '@mui/material/Typography';
 import * as yaml from 'js-yaml';
 import _ from 'lodash';
-import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
@@ -97,9 +96,6 @@ export default function EditorDialog(props: EditorDialogProps) {
     readOnly: isReadOnly(),
     automaticLayout: true,
   };
-  const { i18n } = useTranslation();
-  const [lang, setLang] = React.useState(i18n.language);
-
   const initialCode = typeof item === 'string' ? item : yaml.dump(item || {});
   const originalCodeRef = React.useRef({ code: initialCode, format: item ? 'yaml' : '' });
   const [code, setCode] = React.useState(originalCodeRef.current);
@@ -182,17 +178,6 @@ export default function EditorDialog(props: EditorDialogProps) {
   React.useEffect(() => {
     codeRef.current = code;
   }, [code]);
-
-  React.useEffect(() => {
-    i18n.on('languageChanged', setLang);
-    return () => {
-      // Stop the timeout from trying to use the component after it's been unmounted.
-      clearTimeout(lastCodeCheckHandler.current);
-
-      i18n.off('languageChanged', setLang);
-    };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
 
   function isReadOnly() {
     return onSave === null;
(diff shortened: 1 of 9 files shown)


frontend/src/components/common/Resource/EditorDialog.tsx
```diff
@@ -372,31 +357,21 @@ export default function EditorDialog(props: EditorDialogProps) {
...
-    return useSimpleEditor ? (
-      <Box height="100%">
-        <SimpleEditor
-          language={originalCodeRef.current.format || 'yaml'}
-          value={code.code}
-          onChange={onChange}
-        />
-      </Box>
-    ) : (
+    const language = originalCodeRef.current.format || 'yaml';
```
DanielLeone, line 362: * remove the language checking, as it doesn't work (at least in the builds I'm running locally)
* small code duplication cleanup

frontend/src/helpers/initializeMonacoEditor.ts
```diff
@@ -0,0 +1,54 @@
...
+        case 'razor':
+          return new htmlWorker();
+        case 'typescript':
+        case 'javascript':
+          return new tsWorker();
+        default:
+          return new editorWorker();
+      }
+    },
+  };
```
DanielLeone: there are a few Vite plugins that do something similar, but I find this to be the simplest/most explicit way to make clear what's going on. Most other plugins will just inject a script into the `<head>` which does something very similar: https://github.com/vdesjs/vite-plugin-monaco-editor/blob/85ecb8421448c94bfd291f8422fd4ab9f6dab5c0/src/index.ts#L122

This is also mentioned in the docs: https://github.com/suren-atoyan/monaco-react#loader-config:~:text=If%20you%20use%20Vite%2C%20you%20need%20to%20do%20this%3A

frontend/vite.config.ts
```diff
@@ -51,6 +52,17 @@ export default defineConfig({
     nodePolyfills({
       include: ['process', 'buffer', 'stream'],
     }),
+    // Make sure we copy the minified monaco-editor source into the static folder
+    // since it's loaded dynamically and not bundled via ESM. We do it this way
+    // to support setting the localization language
+    viteStaticCopy({
+      targets: [
+        {
+          src: "node_modules/monaco-editor/min/vs",
```
DanielLeone, line 61: there's possibly some files we can exclude from this path so we don't have to bundle them all - if we're happy with this approach then we should figure out which files are required and only include those (workers files, language files, etc.)

frontend/src/App.tsx
```diff
@@ -47,7 +48,9 @@ function AppWithRedux(props: React.PropsWithChildren<{}>) {
 
   return (
     <I18nextProvider i18n={i18n}>
-      <ThemeProviderNexti18n theme={muiTheme}>{props.children}</ThemeProviderNexti18n>
+      <MonacoEditorProvider>
```
illume: Can this MonacoEditorProvider be moved to be closer where it's used?

At some point I'd like the monaco editor to be lazy loaded, but the code is all a bit tangled at the moment to do this easily. For this reason it would be good if it's possible to move this closer to where the editor is used.

illume (commented):
Thank you very much for this!

Working i18n in Monaco again, and working Monaco in air gapped environments again are great improvements.

(It's unfortunate we have to resort to copying files into the static folder. There's quite a lot of vite config specific to Monaco editor now.)

frontend/vite.config.ts
illume: If the approach works, that's great.

I would be happier of course if we didn't need to do with this vite and Monaco.

DanielLeone: Yeah absolutely - I guess the best solution might be using esm modules, with Vite chunk splitting, but to get i18n working with that approach is a lot more work, as it doesn't seem to be supported by default.

Here's a related issue for this: https://github.com/microsoft/monaco-editor/issues/1322

It's just more vite plugins though and more copying files; wrapping all the workers, etc.
Can ofcourse be done, but I felt this was a much lighter approach for now.

frontend/src/App.tsx
DanielLeone: In this PR, the editor is still lazy loaded, it only fetches the loader (which  in turn fetches the files and workers) once someone calls

```
import { loader } from '@monaco-editor/react';
loader.init()
```

and this happens inside the `<Editor>` component; so no files are downloaded until an `<Editor>` is rendered.

Did you still want to move it lower? It could go anywhere, as long as it's before any actual `<Editor>` instance is rendered; where would you like it?
Maybe on the outside of
```
<Plugins />
<Layout />
```
in `<AppContainer>`?

I'd be hesitant to put it any lower myself just incase someone adds an `<Editor>` to a plugin or at the root of the app layout, etc. But of-course, up to you.

I think I'll rename this component to `MonacoEditorLoaderInitializer` to make clear we're just configuring the loader here, not doing the loading.

---

These changes have been made in [`bc6e4d1` (#3818)](https://github.com/kubernetes-sigs/headlamp/pull/3818/commits/bc6e4d12f1ccae47d935710ec53222a6a5df8450)

illume: Thanks for the explanation. That sounds pretty good I think.

illume (changes requested):
With the app the editor doesn't seem to work.

eg.  `make app-mac` or `make app-linux`

illume:
Pushed a fix for the compiled headlamp apps.

illume (approved):
🎉 thanks!

DanielLeone:
Thank you for your efforts on this one as well @illume 🙏

PR #3820, opened 2025-08-22
frontend: Bump cipher-base to 1.0.6

skoeva:
This change updates the cipher-base dependency in the frontend to the desired 1.0.6 version.


illume (approved):
🎉 thanks

PR #3825, opened 2025-08-23
frontend: AddCluster: Add link to minikube in plugin-catalog

illume:
If in electron, and no plugin is already installed... then we add
a link to the plugin-catalog.

This will allow Headlamp users to more easily add local clusters.

### DIFF
--- a/frontend/src/components/App/CreateCluster/AddCluster.tsx
+++ b/frontend/src/components/App/CreateCluster/AddCluster.tsx
@@ -24,6 +24,7 @@ import Typography from '@mui/material/Typography';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router-dom';
+import { isElectron } from '../../../helpers/isElectron';
 import { createRouteURL } from '../../../lib/router';
 import { ClusterProviderInfo } from '../../../redux/clusterProviderSlice';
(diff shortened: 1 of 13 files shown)


frontend/src/components/App/CreateCluster/AddCluster.tsx
```diff
@@ -81,9 +82,17 @@ export default function AddCluster(props: DialogProps & { onChoice: () => void }
...
+          </Grid>
+          {isElectron() && addClusterProviders.length === 0 && (
             <Grid item xs={12}>
-              <Typography variant="h4">{t('translation|Providers')}</Typography>
+              <Button
+                onClick={() => history.push('/#/plugin-catalog/headlamp-plugins/headlamp_minikube')}
+                startIcon={<InlineIcon icon="mdi:plus-box-outline" />}
+              >
+                {t('translation|Add Local Cluster Provider')}
+              </Button>
```
joaquimrocha, line 100: For this to be compatible with builds that may not use the plugin catalog, we should probably check if the plugin catalog plugin is available and enabled.

illume: Is there one you know of?

illume: Added a check for plugin catalog.

Issue #3707, opened 2025-07-29
In-cluster installation certificate error


DamsPer:
## Describe the bug
We have configured OIDC to access clusters define in kubeconfig. We can access to the local cluster but not to a remote cluster. We have this error : 
`http: proxy error: tls: failed to verify certificate: x509: certificate signed by unknown authority`

Config OIDC : 
```
config:
  oidc:
    clientID: "<CLIENT_ID>"
    clientSecret: "CLIENT_SECRET"
    issuerURL: "<URL_OIDC>"
    useAccessToken: false
```

Config "Kubeconfig": 
```
extraManifests:
- |
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: cm-kubeconfig
  data:
    config: |
        apiVersion: v1
        clusters:
        - cluster:
            insecure-skip-tls-verify: true
            certificate-authority-data: <CERTIFICATE>
            server: https://<API_SERVER_URL>
          name: myCluster
        contexts:
        - context:
            cluster: myCluster
            user: oidc
          name: myCluster
        current-context: ""
        kind: Config
        # preferences: {}
        users:
        - name: oidc
          user:
            exec:
              apiVersion: "client.authentication.k8s.io/v1beta1"
              args:
              - oidc-login
              - get-token
              - --oidc-issuer-url=<URL_OIDC>
              - --oidc-client-id=<CLIENT_ID>
              - --oidc-client-secret=<CLIENT_SECRET>
              - --grant-type=authcode
              command: kubectl
              env: null
              interactiveMode: Never
              provideClusterInfo: false
```

## Environment (please provide info about your environment):
 - Installation type: in-cluster
 - Headlamp Version: [0.33.0]

## Are you able to fix this issue?
No

oliverbaehler:
Same here, but with the following values:

```yaml
        volumeMounts:
          - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
            name: token-ca
        env:
          - name: OIDC_USE_ACCESS_TOKEN
            value: "false"
          - name: KUBERNETES_SERVICE_HOST
            value: "capsule-proxy.capsule-system.svc"
          - name: KUBERNETES_SERVICE_PORT
            value: "9001"
        volumes:
          - name: token-ca
            projected:
              sources:
              - serviceAccountToken:
                  path: token
              - secret:
                  name: capsule-tls
                  items:
                  - key: ca.crt
                    path: ca.crt
              - downwardAPI:
[...]

PR #3826, opened 2025-08-25
backend(oidc): Add support for custom CA and skipping TLS verification

yolossn:
this patch adds support for custom CA and skipping TLS verification for OIDC provider verification. This improves flexibility for users working with self-signed OIDC providers.

Fixes: 
https://github.com/kubernetes-sigs/headlamp/issues/127 
https://github.com/kubernetes-sigs/headlamp/issues/3707
https://github.com/kubernetes-sigs/headlamp/issues/2207 (part 1)
> OIDC doesn't support self signed CA - Going through the instructions for OIDC ignores trusting a certificate. This makes OIDC mostly useless for identity prociders that aren't publicly hosted, like OpenUnison, in an enterprise environment which usually has its own internal CA.

Steps to test:
1. Deploy Dex or any other OIDC provider with custom CA
2. Configure the minikube cluster to use OIDC auth with your OIDC provider with custom CA
3. Deploy headlamp without skip-tls-verify or customCAs, you will the following error when you click "Sign In"
image
4. Deploy headlamp with skip-tls-verify alone configure. You will get a warning message that this isn't a safe way to use and suggest not do this in production.
image
On clicking Sign In you should be able to see the OIDC providers login page
image
5. Deploy headlamp with customCA configured, you will not see the warning and the sign in should work.
image
image

Note: 
re-opened based on https://github.com/kubernetes-sigs/headlamp/pull/2880 due to branch restrictions.


illume (changes requested):
This note from the last PR:

https://github.com/kubernetes-sigs/headlamp/pull/2880#pullrequestreview-2607593252
> I left some notes inline.
> 
> * Can you please add tests and a description of how to test manually in the PR?
> * Probably the related issue(s) should be in the PR description.
> * There will need to be support added to the helm chart.
> * docs in oidc.md ?

I didn't check if the notes after this made in the other PR were addressed here. Were they?

ashu8912:
synced with @yolossn  offline and tested this with both skip TLS verify and custom CA cert and it works for both

illume (approved):
Great work. 🎉 thanks!

(Ultimately we need to add support for at least one of these flags to the helm chart. But I think we can merge this in now, and that be done in a separate PR?)

yolossn:
@illume  Thanks for the review, I will create a follow up PR for the helm chart and update the `oidc.md` to include the helm charts too. Would like to get this merged for now.

PR #3829, opened 2025-08-26
headlamp-plugin: Remove unused jest dependency

skoeva:
Following #3780 and #3803, this change removes the unused dependency on `jest` in the headlamp-plugin package.


illume (approved):
🎉 thanks :)

Issue #3018, opened 2025-03-25
Searchbar overlaps vertical scroll


Andrew1988123123:
## Describe the bug
Searchbar overlaps vertical scroll

## To Reproduce
Steps to reproduce the bug:
1. Go to pod logs and open searchbar

## Environment (please provide info about your environment):
 - Installation type: Windows
 - Headlamp Version: 0.29.0

## Are you able to fix this issue?
No.

## Additional Context
![Image](https://github.com/user-attachments/assets/dced10dd-3957-4e7f-a29f-dbfbec9d0ae6)

skoeva:
Will take a look into this

illume:
This still affects Windows users, so is still a bug.

PR #3721, opened 2025-07-30
Fixed Searchbar overlaps vertical scroll 

greedy-wudpeckr:
## Summary

Fixed Search-bar that overlaps vertical scroll in pod logs. 

## Related Issue

Fixes #3018

## Changes

before : 
Screenshot 2025-07-31 001350

after : 
Screenshot 2025-07-31 001329

## Steps to reproduce the bug:

Go to pod logs and open search-bar

PR #3830, opened 2025-08-26
frontend: LogViewer: Prevent search overlapping scroll

skoeva:
This change fixes an issue in Windows where the search bar in the log viewer overlaps with the scroll on the side.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/3018, follow-up from [#3721 ](https://github.com/kubernetes-sigs/headlamp/pull/3721) by @greedy-wudpeckr

### Testing
- [X] Navigate to a pod in Headlamp and click on 'Show Logs'
- [X] Click on the search icon and note that the search bar does not overlap with scroll

image


illume (approved):
🎉

Thanks @greedy-wudpeckr and @skoeva

PR #3839, opened 2025-08-28
frontend: Sanitize Projects' names as we type

joaquimrocha:
## Summary

Projects' names get sanitized into an ID after the user types them which can be a surprising thing.
This PR instead transforms the typed name into the ID format as the user types.

In the future maybe we add support for more human-readable names.

## Steps to Test

1. Create a new Project
2. Type capital letters and spaces -> observe how the name turns into lowercase letters + dashes
3. Type a trailing or repeated dash and observe how the ID changes to remove these after focusing on another UI element


skoeva (commented):
could use updated translations (causing failure in CI)

vyncent-t (approved):
tried and works LGTM 👍

PR #3840, opened 2025-08-28
frontend: Add options when deleting a project

joaquimrocha:
The projects were being deleted by completely deleting the associated namespace and this may be not what the user wants. So these changes allow the user to choose exactly how the project should be deleted (just the project notion or the full namespaces associated with it).

## Steps to Test

1. Create a project
2. Try to delete that project
3. Test the different options there


frontend/src/components/project/ProjectDeleteButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+import AuthVisible from '../common/Resource/AuthVisible';
+import { ProjectDeleteDialog } from './ProjectDeleteDialog';
+
+interface ProjectDeleteButtonProps {
+  project: ProjectDefinition;
+  namespaces: Namespace[];
+  buttonStyle?: ButtonStyle;
+}
+
+export function ProjectDeleteButton({
```
skoeva, line 31: I wonder if we should further abstract this and the `DeleteButton` component for KubeObjects and make them into wrappers? just a thought that came to mind

frontend/src/components/project/ProjectDeleteDialog.tsx
```diff
@@ -0,0 +1,172 @@
...
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={onClose} color="secondary" variant="contained">
+          {t('Cancel')}
+        </Button>
+        <Button
+          onClick={handleDelete}
+          color="primary"
+          variant="contained"
+          sx={{ minWidth: '200px' }} // Fixed width to prevent button resize
```
skoeva, line 165: I think the fixed width looks awkward IMO, could be removed

image

frontend/src/components/project/ProjectDeleteDialog.tsx
```diff
@@ -0,0 +1,172 @@
...
+      aria-labelledby="project-delete-dialog-title"
+      aria-describedby="project-delete-dialog-description"
+      maxWidth="sm"
+      fullWidth
+    >
+      <DialogTitle id="project-delete-dialog-title">{t('Delete Project')}</DialogTitle>
+      <DialogContent>
+        <DialogContentText id="project-delete-dialog-description" component="div">
+          <Typography variant="body1" paragraph>
+            {t('What would you like to do with project "{{projectName}}"?', {
```
skoeva: I think this might be more appropriate / in line with our other delete dialogs:
```suggestion
            {t('Are you sure you want to delete project "{{projectName}}"?', {
```

skoeva (commented):
Tested and works well, thanks! Left a few comments

frontend/src/components/project/ProjectDeleteDialog.tsx
```diff
@@ -0,0 +1,172 @@
...
+        <Button onClick={onClose} color="secondary" variant="contained">
+          {t('Cancel')}
+        </Button>
+        <Button
+          onClick={handleDelete}
+          color="primary"
+          variant="contained"
+          sx={{ minWidth: '200px' }} // Fixed width to prevent button resize
+        >
+          {deleteNamespaces ? t('Delete Project & Namespaces') : t('Remove Project')}
```
skoeva: Maybe it makes more sense to stay consistent with the term "Delete", WDYT?
```suggestion
          {deleteNamespaces ? t('Delete Project & Namespaces') : t('Delete Project')}
```

frontend/src/components/project/ProjectDeleteDialog.tsx
joaquimrocha: This is to prevent the dialog from jumping in width when we select that box.

frontend/src/components/project/ProjectDeleteButton.tsx
joaquimrocha: Let's think about that later. For now, since this is inside the project folder, it's fine.

joaquimrocha:
Updated. Thanks for the review @skoeva !

frontend/src/components/project/ProjectDeleteDialog.tsx
skoeva: The dialog looks like it's the same width when I commented the line out:

image
image

The button still resizes when opting to delete the namespaces, it just grows less dramatically with the styling

image
image

PR #3843, opened 2025-08-28
frontend: improve filter table

farodin91:
## Changes

- Added job and volume sort for column filtering
- Fixed issue confusing sort values

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -416,7 +416,7 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
               id: 'namespace',
               header: t('glossary|Namespace'),
               gridTemplate: 'auto',
-              accessorFn: (item: RowItem) => item.getNamespace() ?? '',
+              accessorFn: (item: RowItem) => item.getNamespace() ?? '-',
               filterVariant: 'multi-select',
               Cell: ({ row }: { row: MRT_Row<RowItem> }) =>
                 row.original?.getNamespace() ? (
--- a/frontend/src/components/common/Table/Table.tsx
+++ b/frontend/src/components/common/Table/Table.tsx
@@ -46,6 +46,7 @@ import { MRT_Localization_PT } from 'material-react-table/locales/pt';
 import { MRT_Localization_ZH_HANS } from 'material-react-table/locales/zh-Hans';
 import { MRT_Localization_ZH_HANT } from 'material-react-table/locales/zh-Hant';
 import { memo, ReactNode, useEffect, useMemo, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
 import { useTranslation } from 'react-i18next';
 import { getTablesRowsPerPage } from '../../../helpers/tablesRowsPerPage';
 import { useURLState } from '../../../lib/util';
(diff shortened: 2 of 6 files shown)


farodin91:
@joaquimrocha I tried to fix it. What do you think?

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -416,7 +416,7 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
               id: 'namespace',
               header: t('glossary|Namespace'),
               gridTemplate: 'auto',
-              accessorFn: (item: RowItem) => item.getNamespace() ?? '',
+              accessorFn: (item: RowItem) => item.getNamespace() ?? '(cluster-wide)',
```
joaquimrocha: This would have to be translatable. I would prefer an empty string or a `-`.

joaquimrocha: I also think it belongs to a different commit.

frontend/src/components/job/List.tsx
```diff
@@ -29,6 +29,24 @@ import ResourceListView from '../common/Resource/ResourceListView';
 import { SimpleTableProps } from '../common/SimpleTable';
 import LightTooltip from '../common/Tooltip/TooltipLight';
 
+export function makeJobStatusValue(job: Job) {
+  if (!job?.status?.conditions) {
+    return 'Unknown';
```
joaquimrocha: Please make this string translatable since it's not coming from Kubernetes' statuses.

frontend/src/components/job/List.tsx
```diff
@@ -29,6 +29,24 @@ import ResourceListView from '../common/Resource/ResourceListView';
...
+  }
+  const conditionOptions = ['Failed', 'Complete', 'Suspended'];
+
+  const condition = job.status.conditions.find(
+    ({ status, type }: { status: string; type: string }) =>
+      conditionOptions.includes(type) && status === 'True'
+  );
+
+  if (!condition) {
+    return 'Unknown';
```
joaquimrocha: Same.

joaquimrocha (changes requested):
Found a couple of things to be changed. Thanks.

frontend/src/components/common/Resource/ResourceTable.tsx
farodin91: done

frontend/src/components/job/List.tsx
farodin91: i also used dash as it also undefined

frontend/src/components/job/List.tsx
farodin91: i also used dash as it also undefined

frontend/src/components/common/Table/Table.tsx
```diff
@@ -366,6 +367,17 @@ export default function Table<RowItem extends Record<string, any>>({
     },
   });
 
+  // toggle
+  useHotkeys(
+    'ctrl+shift+t',
```
joaquimrocha, line 372: This will collide with the browser's same shortcut for re-opening the last closed tab. Maybe `ctrl+alt+f`?
We are planning on having a way to edit the shortcuts but we don't have it yet.

farodin91: Also a shortcut overview?

PR #3849, opened 2025-08-29
frontend: Export functions for plugin Project overrides

joaquimrocha:
## Summary

This PR adds the needed plugin exports so the type checker doesn't complain when plugins use it.

## Steps to Test

1. Check that VScode complains about the unknown imports in the Project example plugin 
2. Check that in this branch it works


illume (approved):
🎉 thanks!

PR #3862, opened 2025-09-02
[helm] add optional PodDisruptionBudget

jcpunk:
## Summary

This PR adds an optional PodDisruptionBudget for the helm deplopyment

## Related Issue

Fixes #ISSUE_NUMBER  

## Changes

- Added optional PodDisruptionBudget for helm

## Steps to Test

1. render template as is
2. Set `.Values.podDisruptionBudget.enabled` to true
3. render again

## Notes for the Reviewer

- This should leave the default install without a PDB
- With `enabled` it should setup a PDB that lets headlamp be offline during a disruption

### DIFF
--- a/.github/workflows/helm-chart-template-test.yml
+++ b/.github/workflows/helm-chart-template-test.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
   lint-test:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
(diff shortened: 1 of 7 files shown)


illume:
Thank you very much for this.

There's a couple of additions that would be good to add if you want:

- Adding docs to the charts/headlamp/README.md ? This is shown on the artifact hub page and elsewhere people are viewing charts.
- To prevent regressions in the future... it would be good if an example could be added into the [test cases folder](https://github.com/kubernetes-sigs/headlamp/tree/main/charts/headlamp/tests/test_cases), with the [expected_templates](https://github.com/kubernetes-sigs/headlamp/tree/main/charts/headlamp/tests/expected_templates) folder having the expected render result.

We can also merge this as is and do these later in another PR if you prefer. What do you think?

jcpunk:
I hesitate to merge code in the hope that tests will come later, but the last time I tried writing helm tests went very poorly.

illume:
How about I do the tests tomorrow, and then push them to this PR for you to review?

illume:
I added a section to the README.md, and a test case. @jcpunk can you please review?

---

(The test case was failing on CI. I guess because the Kubernetes we are testing against is too old, so it doesn't render unhealthyPodEvictionPolicy (which came out in Kubernetes 1.26). So I updated the GitHub actions to use the newer helm, and now it works.) @skoeva Please review?

jcpunk:
looks fine to me

PR #3863, opened 2025-09-02
app: Update dependencies

skoeva:
These changes perform the following updates to the app dependencies:
- Remove the `electron/notarize` dependency, which is no longer needed after https://github.com/kubernetes-sigs/headlamp/pull/1741
- Adds the missing `glob` dependency which is required in `app/scripts/setup-plugins.js`


illume (approved):
👍 thanks!

PR #3872, opened 2025-09-03
frontend: k8s: Remove references to makeKubeObject

skoeva:
This change removes references to the deprecated `makeKubeObject()` function and replaces them with `KubeObject`.

### Testing
- [X] Open Headlamp and try to create a `ClusterRole` or `ClusterRoleBinding`
- [X] Ensure that they are successfully created with no errors


frontend/src/plugin/__snapshots__/pluginLib.snapshot
```diff
@@ -298,7 +298,6 @@
       "HEADLAMP_ALLOWED_NAMESPACES": "headlamp.allowed-namespaces",
       "KubeObject": [Function],
       "getAllowedNamespaces": [Function],
-      "makeKubeObject": [Function],
```
sniok, line 301: this function is used in some plugins, we have to keep it in the pluginLib @skoeva @illume

PR #3873, opened 2025-09-03
frontend: lib/k8s: Fixing some circular imports

illume:
Using import type in the lib/k8s folder. Only v1 and v2 folders done for now.

Why use import type?

-  No runtime code, removed from compiled JavaScript, so zero bundle impact
-  Clear intent, signals you’re importing types only, not values
-  Avoids circular dependencies, safer when working across interdependent modules
-  Faster builds, helps TypeScript and bundlers optimize
-  Cleaner output, keeps runtime imports minimal and focused


skoeva (approved):
🎉

PR #3874, opened 2025-09-03
frontend: stateless: Extract functions to new files

skoeva:
These changes extract the following functions to new files, which are imported by `frontend/src/lib/k8s/api/v2/fetch.ts`:
- `findKubeconfigByClusterName`
- `getUserIdFromLocalStorage`

This aims to remove circular imports caused by `fetch.ts` importing from `stateless/index.ts`.

### Testing
- [X] Run `make frontend-test` and ensure that all tests pass


illume (approved):
🎉

PR #3883, opened 2025-09-05
backend: Fix base URL feature

### LINKED ISSUE #3870: Headlamp v0.35 incluster is broken: `Getting the cluster version: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`
## Describe the bug
I've upgraded the Headlamp Helm chart to v0.35 today and now it's no longer usable.
It's failing on `/version` endpoint, which is returning the `index.html` instead of the expected json.

## To Reproduce
Steps to reproduce the bug:
1. Deploy Helm chart v0.35.0 with `baseURL=/_hl`
2. Open Headlamp in browser
3. Open browser console
4. [See error]

 - Installation type: Helm
 - Headlamp Version: 0.35.0

No

## Additional Context
Maybe a regression on `baseURL` handling when Headlamp is deployed to a sub path?

DanielLeone:
## Summary

This PR fixes #3870 - an explanation of the issue is here: https://github.com/kubernetes-sigs/headlamp/issues/3870#issuecomment-3254748948

## Related Issue

Fixes #3870

## Changes

- Updated `index.html` to have a consistent replacement target for the sever's `baseURL`
- Cleaned up `headlamp.go` a little bit while I was there, added a test for the new replacement code

## Steps to Test

See screenshots

These were the manual tests I ran, all seem to be working for me in this PR.
In v0.35.0 when using a `baseUrl` on the server, the webpage doesn't load properly at all.

#### Docker image with base url configured:

image

#### Docker image with no base url configured:

image

#### webpack dev server without base url configured:

image

#### webpack dev server with base url configured:

image

#### rspack dev server without base url configured:

image

#### rspack dev server with base url configured:

image

### DIFF
--- a/frontend/rsbuild.config.ts
+++ b/frontend/rsbuild.config.ts
@@ -25,7 +25,7 @@ export default defineConfig({
   html: {
     template: './index.html',
     templateParameters: {
-      BASE_URL: process.env.BASE_URL || './',
+      BASE_URL: process.env.BASE_URL || '/',
     },
   },
   server: {


backend/cmd/headlamp.go
```diff
@@ -176,35 +176,26 @@ func fileExists(filename string) bool {
 	return !info.IsDir()
 }
 
-// copy a file, whilst doing some search/replace on the data.
-func copyReplace(src string, dst string,
-	search []byte, replace []byte,
-	search2 []byte, replace2 []byte,
-) {
-	data, err := os.ReadFile(src)
+func mustReadFile(path string) []byte {
```
DanielLeone, line 129: just shuffled these functions around a little bit to make the test case a bit easier to write without worrying about files on disk, there should be no logic changes within - instead of having 1 function with does:
* reading file
* maybe replace 1
* maybe replace 2
* writing file

there's now a function for each 🙏 
this will mean less file reading and writing to and from disk as well

backend/cmd/headlamp.go
```diff
@@ -213,20 +204,38 @@ func baseURLReplace(staticDir string, baseURL string) {
...
 	}
 
-	if !fileExists(indexBaseURL) {
-		copyReplace(index, indexBaseURL, []byte(""), []byte(""), []byte(""), []byte(""))
-	}
+	// Replacement for headlampBaseUrl - matched from the known index.html content
+	data = bytes.ReplaceAll(data,
+		[]byte("headlampBaseUrl = __baseUrl__"),
+		[]byte(fmt.Sprintf("headlampBaseUrl = '%s'", replaceURL)),
+	)
```
DanielLeone, line 163: This replacement is the real change.

Instead of looking for `headlampBaseUrl = './'` we now look for `headlampBaseUrl = __baseUrl__`.

That was effectively the bug; since rspack support was added, the contents of `index.html` no longer contained `headlampBaseUrl = './'`, so this replacement never ran.

backend/cmd/headlamp_test.go
```diff
@@ -878,7 +849,7 @@ func TestBaseURLReplace(t *testing.T) {
 	indexContent := []byte(`<!DOCTYPE html>
 <html>
 <head>
-    <script>var headlampBaseUrl = './';</script>
+    <script>var headlampBaseUrl = __baseUrl__;</script>
```
DanielLeone, line 784: this test input had to change to meet the new interface of the replacement.

whilst `headlampBaseUrl = './'` probably looks more correct to have here, it's not that straight forward whilst supporting both rspack and webpack

backend/cmd/headlamp_test.go
```diff
@@ -942,6 +913,36 @@ func TestBaseURLReplace(t *testing.T) {
...
+	// after the front-end is built, one of the BASE_URL variables will have been replaced with string contents
+	// depending on which build system was used. In either case though, if the server knows the base URL, we want to ensure it is used;
+	// so headlampBaseUrl will be forced to the value we want
+	output := makeBaseURLReplacements([]byte(`
+<!DOCTYPE html>
+<html>
+<head>
+    <script>
+        __baseUrl__ = ['%BASE_URL%', '<%= BASE_URL %>'].filter(x => !x.includes("BASE_URL")).join('')
+        headlampBaseUrl = __baseUrl__;
```
DanielLeone, line 859: this is what appears in the real index.html now, more or less

backend/cmd/headlamp_test.go
```diff
@@ -942,6 +913,36 @@ func TestBaseURLReplace(t *testing.T) {
...
+</html>
+`), "/headlamp")
+
+	assert.Equal(t, string(output), `
+<!DOCTYPE html>
+<html>
+<head>
+    <script>
+        __baseUrl__ = ['%BASE_URL%', '<%= BASE_URL %>'].filter(x => !x.includes("BASE_URL")).join('')
+        headlampBaseUrl = '/headlamp';
```
DanielLeone, line 871: and it is replaced as expected now - hopefully that makes sense.

`__baseUrl__` is the URLs injected by either `rspack` or `webpack`
and `headlampBaseUrl` will default to that.

But if the server wants to inject the baseURL, like we're testing here, it will just override it and ignore the contents of `__baseUrl__`

frontend/rsbuild.config.ts
```diff
@@ -25,7 +25,7 @@ export default defineConfig({
   html: {
     template: './index.html',
     templateParameters: {
-      BASE_URL: process.env.BASE_URL || './',
+      BASE_URL: process.env.BASE_URL || '/',
```
DanielLeone, line 28: I don't think this has any implication, since we account for `.` vs `./` vs `/` in `getBaseUrl()` but I thought I'd make it at least consistent with what webpack is defaulting to (from what I can tell)

backend/cmd/headlamp.go
skoeva: might be a good idea to move this to a separate commit, since this isn't directly related to the issue

DanielLeone: thanks - it has been done - feel free to adjust as necessary

DanielLeone:
split and updated commits as requested 🙏 feel free to further adjust

joaquimrocha:
Thank you @DanielLeone ! I believe the tests failure is due to my rebase on main, so we have to fix that issue in the main branch.
/lgtm
/approve

Issue #3502, opened 2025-06-20
Dynamic clusters self deleting on renames


vyncent-t:
## Describe the bug
When using the add cluster button, adding resources from one kubeconfig file is flakey. (Users have reported it self deletes)

Adding an additional resources from another kubeconfig, then renaming a resource, removes the initial kubeconfig resources.

## To Reproduce
Steps to reproduce the bug:
1. Open headlamp in app mode
2. Click add cluster button at home view
3. Click and add clusters from a kubeconfig
4. Return to home page and click add cluster button again
5. Add additional resources from a different kubeconfig
6. Return to homepage and rename one of the new resources
7. Notice that the initial resources vanish

## Environment (please provide info about your environment):
 - Installation type: Linux
 - Headlamp Version: 0.31.1

## Are you able to fix this issue?
No current fix yet

## Additional Context
Experienced by other users https://github.com/kubernetes-sigs/headlamp/issues/2979#issuecomment-2983805902

vyncent-t:
could be related to 
- https://github.com/kubernetes-sigs/headlamp/issues/3498

PR #3896, opened 2025-09-10
frontend: updateStatelssClusterKubeconfig: Fix renaming dynamic cluster spawning stale names

vyncent-t:
## Summary

This PR fixes a major issue https://github.com/kubernetes-sigs/headlamp/issues/3502 for dynamic clusters where renaming a dynamic cluster would spawn links to the stale name and the new custom name. This also solves the issue where renaming clusters would remove all previous clusters from a different kubeconfig source.

Users can now rename dynamic clusters without having the old name still appearing on the list.

## Related Issue

Fixes [#ISSUE_NUMBER  ](https://github.com/kubernetes-sigs/headlamp/issues/3502)

## Changes

- added updateStatelessClusterKubeconfig into its own file
- updated updateStatelessClusterKubeconfig logic to now handle renaming of dynamic clusters
- added tests for updateStatelessClusterKubeconfig changes

## Steps to Test

1. Open headlamp in app mode
2. Click add cluster button at home view
3. Click and add clusters from a kubeconfig
4. Return to home page and click add cluster button again
5. Add additional resources from a different kubeconfig
6. Return to homepage and rename one of the new resources
7. Notice that the initial resources no longer vanish
8. Notice that the old name of the cluster you just named is no longer in the list


frontend/src/stateless/updateStatelessClusterKubeconfig.test.ts
```diff
@@ -0,0 +1,282 @@
...
+          )
+      );
+      return { matchingKubeconfig: ctx || null, matchingContext: ctx || null };
+    }),
+    handleDataBaseError: vi.fn(),
+    handleDatabaseUpgrade: vi.fn(),
+  };
+});
+
+import { updateStatelessClusterKubeconfig } from './updateStatelessClusterKubeconfig';
```
skoeva: this should be at the top with the other imports

frontend/src/stateless/updateStatelessClusterKubeconfig.test.ts
```diff
@@ -0,0 +1,282 @@
...
+if (typeof (global as any).atob === 'undefined') {
+  (global as any).atob = (s: string) => Buffer.from(s, 'base64').toString('utf8');
+}
+if (typeof (global as any).btoa === 'undefined') {
+  (global as any).btoa = (s: string) => Buffer.from(s, 'utf8').toString('base64');
+}
+
+type Row = { key: string; kubeconfig: string };
+
+function makeFakeIndexedDB(rows: Row[], opts?: { failPut?: boolean }) {
```
skoeva, line 51: using [fake-indexeddb](https://github.com/dumbmatter/fakeIndexedDB) might be worth it here

frontend/src/stateless/updateStatelessClusterKubeconfig.test.ts
vyncent-t: fixed

frontend/src/stateless/updateStatelessClusterKubeconfig.test.ts
vyncent-t: will consider in future for test improvements as the deleteClusterKubeconfig test also makes use of makeFakeIndexedDB

Issue #3907, opened 2025-09-13
Use Kind instead of Type by default in Table Column header


LinPr:
Currently the header name type, which is not very correspond to the  kubernetes term, I think use Kind is better

```typescript
          case 'type':
          case 'kind':
            return {
              id: 'kind',
              header: t('translation|Type'),
              accessorFn: (resource: RowItem) => String(resource?.kind),
              filterVariant: 'multi-select',
              gridTemplate: 'min-content',
            };
```

<img width="1993" height="312" alt="Image" src="https://github.com/user-attachments/assets/6bbe9b5c-ad8d-4066-ae6e-54efa57305dc" />

skoeva:
Fixed by #3908

PR #3908, opened 2025-09-13
frontend: ResourceTable: Update header translation from 'Type' to 'Kind'

LinPr:
## Summary
see this issue
#3907


skoeva (commented):
nice! all that's left is to squash the commits together and should be good

Issue #3215, opened 2025-05-09
Revamp the plugin development documentation


joaquimrocha:
## Describe the bug

Recently a user tried to run Headlamp for the first time and also build a plugin from scratch. They gave me this feedback:

### Problems when running headlamp-plugin

- [ ] Running `npx @kinvolk/headlamp-plugin create` takes a long time
  Maybe we cannot solve it but we could look into it at least
- [ ] Creating a plugin outputs lots of deprecated warnings and engine incompatibility
  Some deprecations are inevitable, and the engines incompatibility is something the user needs to take care of, but we should lower the number of packages that are deprecated

### Outdated/invalid docs

- [ ] [Building & Shipping](https://headlamp.dev/docs/latest/development/plugins/building#shipping-and-deploying-plugins) plugins seems outdated
  - [ ] Doesn't e.g. mention that plugins have a package.json and a main.js
  - [ ] Doesn't mention deploying plugins using gitops like [we have in the flux example](https://github.com/headlamp-k8s/plugins/tree/main/flux#installing-the-flux-headlamp-plugin-in-a-headlamp-deployment), which many users seem to like
  - [ ] Seems unrelated to the Creating a Plugin section

- [ ] The [How to Create a Plugin](https://headlamp.dev/docs/latest/development/plugins/how-to) section starts mentioning types...
  - [ ] There's no mention of the dependencies that are needed (node)

Seems like we need to rethink these sections, maybe make them into one.

### More context to take into account

- [ ] Cover more basic steps
  - Mention e.g. that the plugins can be created in any place, likely in a user's favorite projects folder or something
  - We need people with little experience to be able to understand the process of zero-to-a-headlamp-plugin

vyncent-t:
I see that the previous assignee was unassigned, I will pick this back up to add with other website changes I am planning next

vyncent-t:
> - [ ] Cover more basic steps

I have created a PR https://github.com/kubernetes-sigs/headlamp/pull/3926 to update some of the wording , mostly focusing on the `Quick Start` and `Create the plugin` steps. 

While doing a dry run of the instructions I ran into these points:
- an npm error when I tried to run the `npx command`
- confusion on the "where" to get started (the "how" was fine)

This PR should look to add clarity to these points.

vyncent-t:
Non urgent @joaquimrocha 

I also noticed that some of these to do items are out of date, feel free to update them when possible. I will add any unfinished or new items to my to do list.

[...]

PR #3926, opened 2025-09-18
docs: getting-started: Update quick start and create plugin

vyncent-t:
## Summary

This PR updates the wording for the getting started page of the plugins docs. I added some clarity around how and where to get started with creating a directory to develop a plugin. I also noticed that I was seeing errors for missing modules when following the current docs so I also added a troubleshoot tip to ensure the `npx` cmd works for creating the plugin foundation, as this worked for me.

## Related Issue

Fixes the "Cover more basic steps" in linked issue here
Fixes [#3215 ](https://github.com/kubernetes-sigs/headlamp/issues/3215)

## Changes

- Updated wording for "quick start" and "Create the Plugin"

### DIFF
--- a/docs/development/plugins/getting-started.md
+++ b/docs/development/plugins/getting-started.md
@@ -32,12 +32,23 @@ Before starting plugin development, ensure you have:
 
 ## Quick Start: Your First Plugin
 
-Let's create a simple plugin that displays "Hello Headlamp!" in the top navigation bar.
+Let's create a simple plugin that displays "Hello Headlamp!" in the top navigation bar. Before we start, it's important to note that plugins can be created and developed outside of the Headlamp repository. This allows for greater flexibility and ease of management.
+
+Please be sure to check that the required plugin development dependencies are correctly installed.
+
+```bash
+# Navigate to the headlamp-plugin directory within the Headlamp repository
+cd headlamp/plugins/headlamp-plugin
+
+# Install dependencies
+npm install
+```
 
 ### Step 1: Create the Plugin
 
-Run the following command to scaffold a new plugin using the headlamp-plugin
-tool. This command should be run where your plugin project (code) will live, typically in a dedicated projects directory (do not run this in Headlamp's
+With dependencies installed and your preferred instance of Headlamp running, navigate to where you would like to create your plugin. In your preferred location, run the following command to scaffold a new plugin using the headlamp-plugin tool.
+
+This command should be run where your plugin project (code) will live, typically in a dedicated projects directory (do not run this in Headlamp's
 plugin installation directory).
 
 ```bash
@@ -88,6 +99,7 @@ npm run start
 ```
 
 This command:
+
 - Makes the plugin available to Headlamp
 - Watches file changes to automatically rebuild the plugin
 
@@ -141,7 +153,7 @@ import {
   registerAppBarAction,
   registerRoute,
   registerSidebarEntry,
-  registerDetailsViewSection
+  registerDetailsViewSection,
 } from '@kinvolk/headlamp-plugin/lib';
 ```
 
@@ -179,7 +191,7 @@ function PodList() {
   return (
     <div>
       <h3>Pods ({pods.length})</h3>
-      {pods.map(pod => (
+      {pods.map((pod) => (
         <div key={pod.metadata.uid}>{pod.metadata.name}</div>
       ))}
     </div>
@@ -192,6 +204,7 @@ function PodList() {
 #### 1. Development Mode
 
 Always use `npm run start` during development for:
+
 - Automatic rebuilding
 - Hot reloading
 - Real-time error checking
@@ -247,23 +260,27 @@ Now that you've created your first plugin, explore these advanced topics:
 ## Troubleshooting
 
 ### Plugin Not Loading
+
 - Ensure Headlamp is running and accessible
 - Check the browser console for JavaScript errors
 - Verify your plugin's `package.json` has correct metadata
 - Make sure you're running `npm run start` in the plugin directory
 
 ### Plugin's Changes Not Reflecting
+
 - Ensure you saved your changes
 - Check if the development server is running (`npm run start`) without errors
 - Remove the installed plugin from Headlamp's plugins folder (see [plugin locations](./architecture.md#plugin-locations)) and re-run `npm run start`
 - Restart Headlamp if necessary
 
 ### Build Errors
+
 - Run `npm run lint` to check for code issues
 - Ensure all imports are correct
 - Check TypeScript errors with `npm run tsc`
 
 ### Hot Reloading Issues
+
 - Restart the development server (`npm run start`)
 - Make sure you do not have several Headlamp tabs in case you are running
   in development mode


docs/development/plugins/getting-started.md
```diff
@@ -32,12 +32,23 @@ Before starting plugin development, ensure you have:
 
 ## Quick Start: Your First Plugin
 
-Let's create a simple plugin that displays "Hello Headlamp!" in the top navigation bar.
+Let's create a simple plugin that displays "Hello Headlamp!" in the top navigation bar. Before we start, it's important to note that plugins can be created and developed outside of the Headlamp repository. This allows for greater flexibility and ease of management.
+
+Please be sure to check that the required plugin development dependencies are correctly installed.
```
illume, line 37: I don't think installing headlamp-plugin like this is needed.

Why is it needed?

docs/development/plugins/getting-started.md
```diff
@@ -32,12 +32,23 @@ Before starting plugin development, ensure you have:
...
+
+# Install dependencies
+npm install
+```
 
 ### Step 1: Create the Plugin
 
-Run the following command to scaffold a new plugin using the headlamp-plugin
-tool. This command should be run where your plugin project (code) will live, typically in a dedicated projects directory (do not run this in Headlamp's
+With dependencies installed and your preferred instance of Headlamp running, navigate to where you would like to create your plugin. In your preferred location, run the following command to scaffold a new plugin using the headlamp-plugin tool.
```
illume, line 49: I don't think this change is necessary if we are not telling users to run npm install inside headlamp-plugin.

illume (commented):
Can you please explain why this change is needed?

(I don't think it is)

docs/development/plugins/getting-started.md
vyncent-t: I feel like this is needed information to emphasize for people trying this out for the first time who may run into the same issue I did.

For context I was doing a dry run of the docs as they are now, and ran into an issue where the `npx` command was having issues finding a module (I already had an instance running before). 

After troubleshooting with running the `npx` command inside and outside of headlamp I found that an extra check of running `npm install` inside the headlamp-plugin folder is what fixed it

Should I reword this to be more of a troubleshooting tip instead?

docs/development/plugins/getting-started.md
vyncent-t: Similar to my other comment, I was having trouble getting the `npx` command to work and found the fix while troubleshooting to be an extra `npm install` inside headlamp-plugin while I developed the plugin elsewhere

This could maybe be changed from a prerequired step into a troubleshooting tip then?

docs/development/plugins/getting-started.md
illume: How can I reproduce the issue? Can you please show me the exact error given?

The documented npx command works for me.

vyncent-t: I think I found the issue, I had accidentally closed my running instance of headlamp when changing workspaces to test the docs

I was able to reproduce it by doing the following:

Replicate someone trying headlamp for the first time
- clone a fresh headlamp (with no external preexisting headlamp node modules accessible)
- run `make frontend` to install everything
- in a different dir, run `npx --yes @kinvolk/headlamp-plugin create my-first-plugin`

image

can confirm that the missing step of not running `make run-frontend` before running the npx cmd works as normal.

running another `npm install` inside `plugins/headlamp-plugin` can also fix it, but if they're running headlamp it isn't needed like you said so we can close this.

illume: Ah, weird.

I see fs-extra is in the dependencies…
https://github.com/kubernetes-sigs/headlamp/blob/main/plugins/headlamp-plugin/package.json#L72

illume: Also with this?

> npx --yes @kinvolk/headlamp-plugin@alpha create my-first-plugin

illume: It looks like it’s picking up ~/workspace/headlamp/ somehow?

What sort of workspace is that? A npm one?

I think this should show why it’s picking up that folder:

> NODE_OPTIONS=--trace-module-resolution npx @kinvolk/headlamp-plugin create my-plugin

vyncent-t: > Also with this?
> 
> > npx --yes @kinvolk/headlamp-plugin@alpha create my-first-plugin

That is odd, `npx --yes @kinvolk/headlamp-plugin create my-first-plugin` (the one in the docs) does not work and prints the error from the picture above, but if I run `npx --yes @kinvolk/headlamp-plugin@alpha create my-first-plugin` it does work. 

Should we default to `npx --yes @kinvolk/headlamp-plugin@alpha create my-first-plugin` or should it remain a separate cmd for a different scenario?

vyncent-t: > It looks like it’s picking up ~/workspace/headlamp/ somehow?

> What sort of workspace is that? A npm one?

my ~/workspace/headlamp/ is my current work environment where headlamp lives, it isn't something npm creates



> NODE_OPTIONS=--trace-module-resolution npx @kinvolk/headlamp-plugin create my-plugin

running this command (I am running it where I plan to make the plugin via npx) returns this error here 

`node: --trace-module-resolution is not allowed in NODE_OPTIONS`

am I meant to run this inside headlamp or where the plugin will be created?

sniok: using `@latest` version should resolve this issue. otherwise npx uses the local version (which is in ~/workspace/headlamp/ and may not have all node_modules installed)

image

illume: How is it finding the local one? Is it linked with npm link or something else?

sniok: I think so. Running `npm rm --global @kinvolk/headlamp-plugin` also fixes this issue
To check you can also run `npm ls --global @kinvolk/headlamp-plugin`

vyncent-t: so should we update the cmd to be @latest?

vyncent-t: > using `@latest` version should resolve this issue. otherwise npx uses the local version (which is in ~/workspace/headlamp/ and may not have all node_modules installed)
> 
> image

so should we update the cmd to be @latest for the docs if that one fixes it?

I would also like to know how it is linked to the local

illume: It seems like this only affects people who npm link, or npm install headlamp-plugin globally. Which would only be people developing headlamp-plugin who forget that.

How about adding something in the headlamp-plugin README.md for developers telling them not to install globally and not to use `npm link` but instead use npm pack?  Npm link seems to have too many issues and corner cases which work strangely.

Adding `@latest` is only a minor inconvenience to the majority of people who will use this doc, but maybe it will help someone.

Can we remove all of the text in this PR? (if this is only affecting headlamp-plugin developers)

vyncent-t: > How about adding something in the headlamp-plugin README.md for developers telling them not to install globally and not to use npm link but instead use npm pack? Npm link seems to have too many issues and corner cases which work strangely.

I am up for adding a new tip in the README for not making it global and directing people to use npm pack, I can also remove all this text and just draft up the README part, maybe add the @latest part too for people

I think it would be a good way to go, I may just close this PR and draft up the other one then.

any thoughts? @sniok

Issue #3913, opened 2025-09-15
How to disable nightly build on forks


Exar04:
<img width="1172" height="170" alt="Image" src="https://github.com/user-attachments/assets/ebab0e0c-f32d-4e74-bc2f-2a628b4f6454" />

I don't want to do nightly build on my fork. How do I disable it for my forked repo?

skoeva:
looking into this

PR #3930, opened 2025-09-22
github: Disable nightly build workflow on forks

skoeva:
This change disables the nightly build workflow on forked repos so that it only gets created on the main repo `(kubernetes-sigs/headlamp)`.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/3913


skoeva:
failing frontend test will be fixed by https://github.com/kubernetes-sigs/headlamp/pull/3932

PR #3936, opened 2025-09-22
frontend: make naming behavior equal on all resources

farodin91:


frontend/src/i18n/locales/de/translation.json
```diff
@@ -322,6 +322,7 @@
   "Init Containers": "Init-Container",
   "Conditions": "Bedingungen",
   "Volumes": "Speicher-Volumes",
+  "Kind": "",
```
skoeva: you can remove these reordering diffs, they're unnecessary

frontend/src/components/endpoints/Details.tsx
```diff
@@ -41,7 +41,7 @@ export default function EndpointDetails(props: {
       name={name}
       namespace={namespace}
       cluster={cluster}
-      title={t('Endpoint')}
+      title={`Endpoint: ${name}`}
```
illume: This should be using a translation:

```ts
t('Endpoint: {{ name }}', { name })
```

Then in the translation file:

```
"Endpoint: {{name}}": "Endpoint: {{name}}"
```

illume (changes requested):
Thanks @farodin91 

I left a note how do the Endpoint translation with the name.

frontend/src/i18n/locales/de/translation.json
illume: Left another note about the translation which will make these changes not necessary... I think. So resolving this.

frontend/src/components/endpoints/Details.tsx
farodin91: I know that this possible but all other resource aren't translate at this location. This fix is only due to backward compatibility.

skoeva: @farodin91 you're correct on that, translations can be taken care of elsewhere since none of the details pages translate the resource in the title

illume: There are some translations for Kind and Endpoint, so I expect they need to keep being translated.

farodin91:
@illume I added translation.

Issue #3918, opened 2025-09-16
Headlamp session expires in about 2 minutes when you log in via oidc


rift0nix:
## Describe the bug
Running Headlamp in-cluster, authenticating through a Keycloak OIDC server.
About 2 minutes after authentication a "Lost connection to the cluster" error is displayed, and the user is returned to the login dialog.

## To Reproduce
Steps to reproduce the bug:
1. Create values.yaml with folowing content
```
image:
  registry: ghcr.io
  repository: headlamp-k8s/headlamp
  pullPolicy: IfNotPresent
  tag: v0.35.0
config:
  watchPlugins: true
  extraArgs:
    - "-oidc-ca-file=/usr/local/share/ca-certificates/extra/ca-chain.pem"
  oidc:
    clientID: "${headlamp_oidc_client_id}"
    clientSecret: "${headlamp_oidc_client_secret}"
    issuerURL: "${headlamp_oidc_issuer_url}"
    callbackURL: "${headlamp_oidc_callback_url}"
    scopes: "openid,email,profile"
pluginsManager:
  enabled: true
  configContent: |
    plugins:
      - name: cert-manager
        source: "https://artifacthub.io/packages/headlamp/headlamp-plugins/headlamp_cert-manager"
        version: "0.1.0"
initContainers:
  - name: extra-certs
    image: alpine:3.18
    command:
      - /bin/sh
      - -c
      - |
        -----BEGIN CERTIFICATE-----
        MIIDXzCCAkegAwIBAgILBAAAAAABIVhTCKIwDQYJKoZIhvcNAQELBQAwTDEgMB4G
        A1UECxMXR2xvYmFsU2lnbiBSb290IENBIC0gUjMxEzARBgNVBAoTCkdsb2JhbFNp
        Z24xEzARBgNVBAMTCkdsb2JhbFNpZ24wHhcNMDkwMzE4MTAwMDAwWhcNMjkwMzE4
        MTAwMDAwWjBMMSAwHgYDVQQLExdHbG9iYWxTaWduIFJvb3QgQ0EgLSBSMzETMBEG
        A1UEChMKR2xvYmFsU2lnbjETMBEGA1UEAxMKR2xvYmFsU2lnbjCCASIwDQYJKoZI
        hvcNAQEBBQADggEPADCCAQoCggEBAMwldpB5BngiFvXAg7aEyiie/QV2EcWtiHL8
[...]

PR #3942, opened 2025-09-24
backend: auth: Improve ParseClusterAndToken function

skoeva:
These changes add a clarifying comment for better readability of the `ParseClusterAndToken` function and re-introduce the call to `GetTokenFromCookie` that was removed in #3494.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/3918


illume:
Thanks @skoeva 

- Would you mind adding a test for this as well?
- The function comment could use an update.

PR #3945, opened 2025-09-24
frontend: Projects: Add isEnabled method to ProjectDetailsTab to allow conditional tab rendering

sniok:
This PR adds a way for plugins to define a new tab in projects that is only rendered on certain projects.

It is implemented as an async `isEnabled` function 

```tsx
// Example of Tab that is only enabled for certain projects
registerProjectDetailsTab({
  id: 'special-tab',
  label: 'Special tab',
  icon: 'mdi:circle',
  component: () => <div>Special tab content</div>,
  isEnabled: async ({ project }) => {
    // In this example tab will only be displayed for projects
    // that have more than 1 cluster selected
    // Note: This function is async so you can make network requests here
    return project.clusters.length > 1;
  },
});
```

### DIFF
--- a/frontend/src/components/project/ProjectDetails.tsx
+++ b/frontend/src/components/project/ProjectDetails.tsx
@@ -392,14 +392,44 @@ function ProjectDetailsContent({ project }: { project: ProjectDefinition }) {
 
   const { items, isLoading } = useProjectItems(project);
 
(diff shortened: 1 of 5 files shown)


yolossn (approved):
Looks good. Just a small suggestion, it will be better to add an example to the example projects plugin that we already have.

PR #3965, opened 2025-09-29
github: e2e-tests: Save screenshot if there is an a11y failure

illume:
This also dumps the html which can be useful for debugging.

- See e2e test CI
- Look in e2e test logs for dumped html (only if there is an a11y failure)
- Look for link in the logs to the saved screenshot artifact


illume:
I don't think this can work ok with external fork PRs.

PR #3967, opened 2025-09-29
Expose cluster-scoped user info via endpoint

skoeva:
These changes do the following:
- Add a configurable `/clusters/{clusterName}/me` endpoint that reads the per-cluster cookie, decodes the JWT via JMESPath, and returns user info
- Update the frontend to fetch `/clusters/{clusterName}/me` in the top bar, show the user info in the menu, and proxy `/api` and `/clusters`

### Testing
- [X] Open a cluster in Headlamp and authenticate with your choice OIDC client
- [X] Refresh the page (ensuring you are still logged in)
- [X] Open the user menu in the top bar and ensure the user credentials appear

image

Related to:
- https://github.com/kubernetes-sigs/headlamp/issues/3562


backend/cmd/headlamp.go
```diff
@@ -594,6 +598,8 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 
 	// Auth token management
 	r.HandleFunc("/auth/set-token", config.handleSetToken).Methods("POST")
+	// Expose user info so the frontend can show the current user in the top bar using the per-cluster auth cookie.
+	r.HandleFunc("/api/me", config.handleMe).Methods("GET")
```
sniok: instead of `/api/me?cluster=clustername` it should be `/clusters/{clusterName}/me`
this way we can keep the original cookie path `"/clusters/" + cluster` without reducing security

skoeva: I suppose it's functionally cluster scoped anyway, will make this change

skoeva (commented):
updated the endpoint and commits/PR descriptions

frontend/src/components/App/TopBar.tsx
```diff
@@ -87,6 +87,59 @@ export default function TopBar({}: TopBarProps) {
...
+      }
+
+      const headlampBaseUrl = (window as any).headlampBaseUrl || '';
+      const isFile = typeof window !== 'undefined' && window.location?.protocol === 'file:';
+      const base = isFile ? 'http://localhost:4466/' : headlampBaseUrl || '';
+      const apiBase = base === '' ? '/' : base.replace(/\/?$/, '/');
+
+      try {
+        // Use per-cluster endpoint so cookies stay scoped with cluster name in the path
+        const url = `${apiBase}cluster/${cluster}/me`;
```
illume: URL is wrong here? clusters not cluster?

backend/cmd/headlamp.go
```diff
@@ -1465,6 +1472,146 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
+	groupsPaths := c.meGroupsPaths
+	if strings.TrimSpace(groupsPaths) == "" {
+		groupsPaths = cfg.DefaultMeGroupsPath
+	}
+
+	username := stringValueFromJMESPaths(claims, usernamePaths)
+	email := stringValueFromJMESPaths(claims, emailPaths)
+	groups := stringSliceFromJMESPaths(claims, groupsPaths)
+
+	// Prevent caching to avoid 304 Not Modified responses from intermediaries/dev server
```
illume: Just in case, ask proxy servers not to save this information:
```
w.Header().Set("Vary", "Cookie")
```

backend/cmd/headlamp.go
```diff
@@ -1465,6 +1472,146 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
+	w.Header().Del("ETag")
+	_ = json.NewEncoder(w).Encode(map[string]interface{}{
+		"username": username,
+		"email":    email,
+		"groups":   groups,
+	})
+}
+
+// stringValueFromJMESPaths tries multiple comma-separated JMESPath expressions and returns the first string result.
+func stringValueFromJMESPaths(payload map[string]interface{}, pathCSV string) string {
```
illume: I don't see stringValueFromJMESPaths specific unit tests. Do the other tests in this PR cover it?

backend/cmd/headlamp.go
```diff
@@ -1465,6 +1472,146 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
 	handleClusterAPI(c, router)
 }
 
+// handleMe returns user info extracted from the per-cluster OIDC cookie so the UI can display the current user.
+// Claim discovery is controlled via the Headlamp configuration (HEADLAMP_CONFIG_ME_* env / --me-*-path flags),
+// which accept comma-separated JMESPath expressions to support different identity providers and nested claim layouts.
+func (c *HeadlampConfig) handleMe(w http.ResponseWriter, r *http.Request) {
```
illume: Instead of adding things to headlamp.go, can these be added to another package?

frontend/src/components/App/TopBar.tsx
```diff
@@ -87,6 +87,59 @@ export default function TopBar({}: TopBarProps) {
   const history = useHistory();
   const { appBarActions, appBarActionsProcessors } = useAppBarActionsProcessed();
 
+  // Fetch current user info from backend
```
illume: Can you please add some story book states for these changes?

backend/cmd/headlamp.go
```diff
@@ -1465,6 +1472,146 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
+	groups := stringSliceFromJMESPaths(claims, groupsPaths)
+
+	// Prevent caching to avoid 304 Not Modified responses from intermediaries/dev server
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate, private")
+	w.Header().Set("Pragma", "no-cache")
+	w.Header().Set("Expires", "0")
+	w.Header().Del("ETag")
+	_ = json.NewEncoder(w).Encode(map[string]interface{}{
+		"username": username,
```
illume: What if the token is expired?

It could show stale information about who is logged in. Can the expiry of the data be checked here? Or is expiry already handled somehow?

frontend/src/components/App/TopBar.tsx
skoeva: I think this is pending additional testing, will update on this

backend/cmd/server.go
```diff
@@ -54,32 +54,50 @@ func main() {
 	StartHeadlampServer(headlampConfig)
 }
 
+func buildHeadlampCFG(conf *config.Config, kubeConfigStore kubeconfig.ContextStore) *headlampconfig.HeadlampCFG {
```
illume, line 65: These build* functions need docs.

backend/cmd/headlamp.go
```diff
@@ -1465,6 +1472,146 @@ func (c *HeadlampConfig) handleClusterRequests(router *mux.Router) {
...
+			}
+		case float64:
+			return fmt.Sprintf("%v", v)
+		case int64:
+			return fmt.Sprintf("%v", v)
+		case map[string]interface{}:
+			b, _ := json.Marshal(v)
+			if len(b) > 0 {
+				return string(b)
+			}
```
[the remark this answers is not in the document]
skoeva: irrelevant. bad copilot.

backend/cmd/server.go
skoeva: added a short sentence for each, it's just extracted logic from the larger function so nothing new

backend/cmd/headlamp.go
skoeva: added a call to `GetExpiryUnixTimeUTC()` to be safe, but theoretically should be covered by verification from `/oidc-callback`

frontend/src/components/App/TopBar.tsx
skoeva: done

backend/cmd/headlamp.go
skoeva: moved to auth and adjusted the logic a bit

backend/cmd/headlamp.go
skoeva: added some tests for the endpoint which covers the helpers

backend/cmd/headlamp.go
```diff
@@ -83,6 +83,9 @@ type HeadlampConfig struct {
 	telemetryConfig           cfg.Config
 	oidcScopes                []string
 	telemetryHandler          *telemetry.RequestHandler
+	meUsernamePaths           string
```
illume: Are these really once per backend? I'm wondering about multiple clusters,

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
 
 	return newToken, nil
 }
+
+// MeHandlerOptions configures how user identity information is extracted from a JWT.
+type MeHandlerOptions struct {
+	UsernamePaths string
```
illume, line 280: Please document these fields?

frontend/src/components/App/TopBar.stories.tsx
```diff
@@ -119,3 +119,26 @@ TwoCluster.args = {
   cluster: 'ak8s-desktop',
   clusters: { 'ak8s-desktop': '', 'ak8s-desktop2': '' },
 };
+
```
illume, line 127: Could you please add a loading state and an error state?

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+	EmailPaths    string
+	GroupsPaths   string
+}
+
+// HandleMe returns a handler that reads the per-cluster auth cookie and responds with user info.
+func HandleMe(opts MeHandlerOptions) http.HandlerFunc {
+	usernamePaths, emailPaths, groupsPaths := resolveMePaths(opts)
+
+	return func(w http.ResponseWriter, r *http.Request) {
```
illume, line 298: Can you please check what happens in the frontend when an error happens? 

I see the request() function used in the frontend assumes that the errors returned will also be in json format with a message field.

frontend/src/components/App/TopBar.tsx
```diff
@@ -87,6 +88,34 @@ export default function TopBar({}: TopBarProps) {
...
+  const [me, setMe] = React.useState<{ username?: string; email?: string } | null>(null);
+  React.useEffect(() => {
```
illume: Naming the function used inside the useEffect would make it a bit more readable.

illume (commented):
I'm interested in the error+loading states, and what happens with multi clusters?

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+	w.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate, private")
+	w.Header().Set("Pragma", "no-cache")
+	w.Header().Set("Expires", "0")
+	w.Header().Set("Vary", "Cookie")
+	w.Header().Del("ETag")
+	_ = json.NewEncoder(w).Encode(map[string]interface{}{
+		"username": username,
+		"email":    email,
+		"groups":   groups,
+	})
```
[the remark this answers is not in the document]
illume: This looks valid.

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+		if p == "" {
+			continue
+		}
+
+		res, err := jmespath.Search(p, payload)
+		if err != nil || res == nil {
+			continue
+		}
+
+		switch v := res.(type) {
```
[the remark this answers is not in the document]
illume: It's probably not much, but a valid point.

backend/pkg/config/config.go
```diff
@@ -218,6 +227,21 @@ func setKubeConfigPath(config *Config) {
...
+
+	if strings.TrimSpace(config.MeEmailPath) == "" {
+		config.MeEmailPath = DefaultMeEmailPath
+	}
+
+	if strings.TrimSpace(config.MeGroupsPath) == "" {
+		config.MeGroupsPath = DefaultMeGroupsPath
+	}
+}
+
```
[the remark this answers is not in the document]
illume: Valid point.

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+			vs := v.String()
+			if vs != "" {
+				return vs
+			}
+		case float64:
+			return fmt.Sprintf("%v", v)
+		case int64:
+			return fmt.Sprintf("%v", v)
+		case map[string]interface{}:
+			b, _ := json.Marshal(v)
```
illume: This was mentioned in another convo that's resolved now, but this error is silently ignored, and I don't think it should be.

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+			for _, it := range v {
+				switch s := it.(type) {
+				case string:
+					out = append(out, s)
+				case float64:
+					out = append(out, fmt.Sprintf("%v", s))
+				case int64:
+					out = append(out, fmt.Sprintf("%v", s))
+				default:
+					b, _ := json.Marshal(s)
```
illume: The error is ignored here, and I don't think it should be.

frontend/src/components/App/TopBar.tsx
```diff
@@ -87,6 +88,34 @@ export default function TopBar({}: TopBarProps) {
...
+    let aborted = false;
+    async function loadMe() {
+      if (!cluster) {
+        setMe(null);
+        return;
+      }
+
+      try {
+        const res = await request('/me');
+        if (!res) return;
```
illume: What about this case:

- A) there is a user logged in
- B) there is not a user logged in

In this case it would should stale data from A?

backend/pkg/auth/auth.go
skoeva: the request call throws, catch block runs, fall back to `setMe(null)` and info is not displayed

if the backend gets plain text, then the `response.json()` call throws, we log "Unable to parse error json", and reject with the HTTP status text

updating the error logging to return a json error body

backend/pkg/auth/auth.go
skoeva: adding logging for these

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,175 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+		return nil, http.StatusUnauthorized, "invalid token"
+	}
+
+	claims, err := DecodeBase64JSON(parts[1])
+	if err != nil {
+		return nil, http.StatusUnauthorized, "invalid token claims"
+	}
+
+	return claims, 0, ""
+}
```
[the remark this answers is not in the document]
illume: This was opened before. @skoeva checked and the token is validated in a separate part of the code. Resolving.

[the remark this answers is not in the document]
skoeva: I see the token is only validated once though, I suppose the cookie can be overwritten later? could run the oidc verifier on every request and add a server side signature to the cookie

edit: the first of these should suffice unless we want to add more

backend/pkg/auth/auth.go
illume: More than a lack of an error message to the user, I'm worried about stale data. 

If it shows "UserA", but then there is an error... it should not show "UserA" still. Would be best if it could show an error there where the user name is IMHO.

skoeva: the request is made every time the topbar is re-rendered, so there would be stale data if there was an error but everything else on the page would also be stale until reload.

to test this, i blocked the request url and reloaded -> no user info displayed
unblocked the request url and reloaded -> user info displayed

illume: Ok, great. Thanks for testing that. I'll mark this as resolved.

backend/pkg/auth/auth.go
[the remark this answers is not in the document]
illume: In normal secure cookies the cookie can only be set by the server side... but there was a method added to make it possible for javascript to set these cookies too.

It looks like no verification of the cookie signature is done there though: `auth/cookies.go@SetTokenCookie`.

It's still not clear to me that allowing JavaScript to bypass the protections Secure Cookies provide is a good idea at all.

backend/pkg/auth/auth_test.go
```diff
@@ -852,3 +854,106 @@ func TestConfigureTLSContext_CACert(t *testing.T) {
...
+
+	payloadJSON, err := json.Marshal(claims)
+	require.NoError(t, err)
+
+	return fmt.Sprintf("%s.%s.%s",
+		base64.RawURLEncoding.EncodeToString(headerJSON),
+		base64.RawURLEncoding.EncodeToString(payloadJSON),
+		"signature",
+	)
+}
```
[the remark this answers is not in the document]
skoeva: added a signed test token and signature enforcement in tests

frontend/src/components/App/TopBar.tsx
skoeva: the useEffect is dependent on `cluster` so if the context switches then we get `setMe(null)` before making a new request

on log out or any error, the data should be cleared

backend/pkg/auth/auth.go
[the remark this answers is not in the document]
illume: @skoeva marking this as unresolved. Can you please write why it should be resolved?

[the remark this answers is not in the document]
skoeva: the server is still setting the httponly cookie, same as before

I guess the only change is rerunning the cluster's oidc verifier so that part is addressed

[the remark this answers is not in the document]
illume: Closing this because the token is verified for the code in this PR.

backend/cmd/headlamp.go
```diff
@@ -84,6 +85,16 @@ type HeadlampConfig struct {
...
+	meEmailPaths              string
+	meGroupsPaths             string
+	meVerifierCache           sync.Map
+}
+
+type oidcVerifierCacheEntry struct {
```
illume: This type and its fields need docs.

backend/cmd/headlamp.go
```diff
@@ -2382,3 +2403,87 @@ func (c *HeadlampConfig) handleSetToken(w http.ResponseWriter, r *http.Request)
...
+			"token verification failed")
+		return http.StatusUnauthorized, fmt.Errorf("invalid token")
+	}
+
+	return 0, nil
+}
+
+// getClusterOIDCVerifier returns a cached OIDC ID token verifier for the given cluster,
+// initialising it when necessary with the cluster-specific TLS and client settings.
+func (c *HeadlampConfig) getClusterOIDCVerifier(ctx context.Context, cluster string,
```
illume: getClusterOIDCVerifier needs tests

backend/cmd/headlamp.go
```diff
@@ -2382,3 +2403,87 @@ func (c *HeadlampConfig) handleSetToken(w http.ResponseWriter, r *http.Request)
 
 	w.WriteHeader(http.StatusOK)
 }
+
+// verifyMeToken re-validates the per-cluster auth cookie using the cluster's OIDC verifier.
+func (c *HeadlampConfig) verifyMeToken(ctx context.Context, r *http.Request,
```
illume: verifyMeToken needs tests

illume:
For the case where there are multiple clusters and the user selects multiple clusters... what does the Topbar say? To me it looks like Topbar could be modified to show the different users (and handle logout). But lets try and get this PR merged before addressing this issue.
- https://github.com/kubernetes-sigs/headlamp/issues/4082

For the setToken not verifying started an issue
- https://github.com/kubernetes-sigs/headlamp/issues/4083

backend/cmd/headlamp.go
illume: There's an issue here for Topbar needing support here:
- https://github.com/kubernetes-sigs/headlamp/issues/4082

I wonder if these should be per context? A future extension might allow supplying multiple things for different context given a separator?

backend/cmd/headlamp.go
```diff
@@ -2382,3 +2403,87 @@ func (c *HeadlampConfig) handleSetToken(w http.ResponseWriter, r *http.Request)
...
+			return entry.verifier, nil
+		}
+	}
+
+	providerCtx := auth.ConfigureTLSContext(ctx, oidcAuthConfig.SkipTLSVerify, oidcAuthConfig.CACert)
+	if c.oidcValidatorIdpIssuerURL != "" {
+		providerCtx = oidc.InsecureIssuerURLContext(providerCtx, c.oidcValidatorIdpIssuerURL)
+	}
+
+	provider, err := oidc.NewProvider(providerCtx, oidcAuthConfig.IdpIssuerURL)
```
illume: I see you set "issuer" above? Why isn't that used here instead of oidcAuthConfig.IdpIssuerURL?

backend/pkg/auth/auth.go
```diff
@@ -271,3 +274,217 @@ func RefreshAndCacheNewToken(ctx context.Context, oidcAuthConfig *kubeconfig.Oid
...
+			for _, it := range v {
+				switch s := it.(type) {
+				case string:
+					out = append(out, s)
+				case float64:
+					out = append(out, fmt.Sprintf("%v", s))
+				case int64:
+					out = append(out, fmt.Sprintf("%v", s))
+				default:
+					if encoded, ok := marshalToString(v); ok && encoded != "" {
```
illume: This function needs tests too.

I'm not sure about this part...

In stringSliceFromJMESPaths, the default case inside the inner loop appends marshalToString(v) (the whole slice) instead of the current element.

s instead of v
>if encoded, ok := marshalToString(s)

backend/cmd/headlamp.go
```diff
@@ -2382,3 +2403,87 @@ func (c *HeadlampConfig) handleSetToken(w http.ResponseWriter, r *http.Request)
...
+	}
+
+	return 0, nil
+}
+
+// getClusterOIDCVerifier returns a cached OIDC ID token verifier for the given cluster,
+// initialising it when necessary with the cluster-specific TLS and client settings.
+func (c *HeadlampConfig) getClusterOIDCVerifier(ctx context.Context, cluster string,
+	oidcAuthConfig *kubeconfig.OidcConfig,
+) (*oidc.IDTokenVerifier, error) {
```
illume: For this one... what are the different things that can change for the cache to be invalidated?

Variables affecting cache validity:

- cluster – The cluster name (primary key).
- issuer – Resolved issuer URL (from oidcAuthConfig.IdpIssuerURL or c.oidcValidatorIdpIssuerURL).
- clientID – Resolved client ID (from oidcAuthConfig.ClientID or c.oidcValidatorClientID).
- oidcAuthConfig.SkipTLSVerify – Boolean controlling TLS verification.
- oidcAuthConfig.CACert – CA certificate string (PEM content).
- (Optional but relevant) c.oidcValidatorIdpIssuerURL – If override changes, issuer changes.
- (Optional but relevant) c.oidcValidatorClientID – If override changes, client ID changes.

backend/cmd/headlamp.go
```diff
@@ -2382,3 +2403,87 @@ func (c *HeadlampConfig) handleSetToken(w http.ResponseWriter, r *http.Request)
...
+		return http.StatusNotFound, fmt.Errorf("cluster not found")
+	}
+
+	oidcAuthConfig, err := kContext.OidcConfig()
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{"cluster": cluster}, err,
+			"failed to get oidc config for token verification")
+		return http.StatusUnauthorized, fmt.Errorf("unauthorized")
+	}
+
```
illume: We would need to consider access tokens here. `if c.oidcUseAccessToken`. Access tokens are used by Microsoft Entra Id and others.

backend/pkg/config/config.go
```diff
@@ -20,6 +20,12 @@ import (
 
 const defaultPort = 4466
 
+const (
+	DefaultMeUsernamePath = "preferred_username,username,name"
```
illume: I think we can add support for "upn" here.

- https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims?tabs=appui
- https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference

> In Azure Active Directory (AAD), the user’s primary identifier is often the User Principal Name (UPN), which looks like an email address (e.g., alice@example.com).

> DefaultMeUsernamePath = "preferred_username,upn,username,name"

illume:
Verification is incredibly complex, and we need to test with every idp to make sure it works (and handle all the different combinations of how it can be done and how things can be configured). I think we SHOULD do this, but as part of separate work to verify the token over all (not just for the /me endpoint).

As discussed, let's finish off tests for these functions as is, make notes in the code or in our issue... and have a feature flag turn the verification off.

backend/cmd/headlamp.go
skoeva: added

backend/cmd/headlamp.go
skoeva: added

backend/cmd/headlamp.go
skoeva: added

backend/cmd/headlamp.go
skoeva: we get the metadata and keys directly from the issuer url (`oidcAuthConfig.IdpIssuerURL`), and we might need a different value to verify (`issuer` can be set to a different value)

backend/pkg/auth/auth.go
skoeva: done

sniok:
Correct me if I'm wrong but since this endpoint only shows a field from a token and doesn't perform authorization then we don't really need to verify the signature? If token was still available on the frontend we would be able to do that completely client side, since it's just base64

illume:
The "Build Container and test" out of disk space failure is unrelated to this PR.

illume (approved):
🎉 thanks!

PR #3974, opened 2025-09-30
frontend: Fix project creation when multiple clusters are involved

### LINKED ISSUE #3887: headlamp.dev/project-id is only added to 1 namespace from multiple clusters
## Describe the bug
With `v0.35.0` version i've seen that there is a new feature called projects. After testing it i can see 1 issue. When creating a project you have an option to choose multiple clusters and namespaces inside them. The list of namespaces is logically updated in a way that if the certain namespace already exists in previous clusters it is not added to the list multiple times. The issue is that if you choose that certain namespace in the project it only applies to the cluster from which it was initially added. So in other words the `headlamp.dev/project-id` label is only added to 1 instance of namespace from multiple clusters instead of all instances with the targeted name.
## To Reproduce
Steps to reproduce the bug:
1. Go to headlamp home view
2. Click on Projects
3. Click on Create Project
4. Choose New Project
5. Choose 2 clusters with the same namespace on them
6. Choose the targeted namespace
7. Check newly created project to see resources from 1 cluster lacking

No

## Additional Context
* `in-cluster` headlamp `v0.35.0` version

joaquimrocha:
## Summary

It was picking only one namespace for setting the project metadata instead of using several.

## Related Issue

fixes #3887


frontend/src/components/project/NewProjectPopup.tsx
```diff
@@ -130,15 +130,20 @@ function ProjectFromExistingNamespace({ onBack }: { onBack: () => void }) {
...
-      const maybeExistingNamespace = namespaces?.find(it => it.metadata.name === selectedNamespace);
-      if (maybeExistingNamespace) {
-        await maybeExistingNamespace.patch({
-          metadata: {
-            labels: {
-              [PROJECT_ID_LABEL]: projectName,
-            },
-          },
-        });
+      const existingNamespaces = namespaces?.filter(it => it.metadata.name === selectedNamespace);
```
sniok, line 133: this assumes that either all clusters have the given namespace or all clusters don't have it.
the case when `cluster 1` has the namespace and `cluster 2` doesn't have it won't work

instead I think we should iterate over each cluster and check for namespace individually 

```tsx
 /**
   * Creates or updates namespaces for the proejct
   */
  const handleCreate = async () => {
    if (!isReadyToCreate || isCreating) return;

    setIsCreating(true);
    try {
      for (const cluster of selectedClusters) {
        const existingNamespace = namespaces?.find(
          it => it.metadata.name === selectedNamespace && it.cluster === cluster
        );
        if (existingNamespace) {
          await existingNamespace.patch({
            metadata: {
              labels: {
                [PROJECT_ID_LABEL]: projectName,
              },
            },
          });
        } else {
          const namespace = {
            kind: 'Namespace',
            apiVersion: 'v1',
            metadata: {
              name: toKubernetesName(typedNamespace),
              labels: {
                [PROJECT_ID_LABEL]: projectName,
              },
            } as any,
          } as KubeObjectInterface;
          await apply(namespace, cluster);
        }
      }

      history.push(createRouteURL('projectDetails', { name: projectName }));
    } catch (e: any) {
      setError(e);
    } finally {
      setIsCreating(false);
    }
  };
```

illume:
The Mac app build being cancelled is due to GitHub runner issues for mac at the moment.

frontend/src/components/project/NewProjectPopup.tsx
illume: This was done

PR #3982, opened 2025-10-01
frontend: Layout: Disable backend loader for now

illume:
On Mac it causes the app on Mac to show error page when switching
branches or sometimes on first load of compiled mac app.
Additionally it flashes white on a dark theme during this error.


illume:
The Mac app check being cancelled is because GitHub Mac runners are having issues at the moment.

joaquimrocha:
Code looks fine to me.

/lgtm
/approve

PR #3993, opened 2025-10-02
frontend: ResourceTable: Correctly use age as a default sorting column

sniok:
This PR fixes default sorting column

## Steps to Test

1. Go to any resource table
2. Resources should be sorted by age


skoeva (approved):
Tested and seems to work as expected 🔥

PR #4005, opened 2025-10-03
frontend: Expose Activity API to plugins

sniok:


sniok:
this PR is cursed, github action keeps failing because it's running out of space. I'll break it up into two PRs

PR #4010, opened 2025-10-05
frontend: ClusterTable: align Actions column and disable sorting/filtering

### LINKED ISSUE #4009: All Clusters Table: Actions column shows sort icon (no label) and misaligned row menu
## Describe the bug
In Home → All Clusters, the rightmost “Actions” column shows a sort icon with no header label, and the three-dots menu in rows looks visually isolated/misaligned compared to other lists.

## To Reproduce
Steps to reproduce the bug:
1. Go to Home.
2. Click "All Clusters"
3. Observe the rightmost column header and row actions.

 - Installation type: Mac-Homebrew
 - Headlamp Version: 0.36.0

Yes (I will propose a PR)

## Additional Context

Image

jaehanbyun:
## Summary

Fix the Actions column in Home → All Clusters: disable sorting/filtering, ensure it stays at right, center-align the three dots menu, and unify the icon to a horizontal ellipsis for consistency with other cells.

## Related Issue

Fixes #4009

## Changes

- Added/Updated [component/file/logic]
- Fixed [bug/issue/typo]
- Refactored [code/module] for clarity/performance

## Steps to Test

1. Go to Home.
2. Click "All Clusters"
3. Observe the rightmost column header and row actions.

**Before**
Screenshot 2025-10-05 at 5 11 16 PM

**After**
Screenshot 2025-10-05 at 5 22 02 PM


jaehanbyun:
@illume ptal :)

frontend/src/components/App/Home/ClusterTable.tsx
```diff
@@ -174,15 +174,18 @@ export default function ClusterTable({
         },
         {
           id: 'actions',
-          header: '',
+          header: t('Actions'),
```
skoeva, line 183: don't think this is necessary

frontend/src/components/App/Home/ClusterTable.tsx
```diff
@@ -174,15 +174,18 @@ export default function ClusterTable({
...
+          gridTemplate: 'min-content',
           muiTableBodyCellProps: {
-            align: 'right',
+            align: 'center',
```
skoeva: tested this change locally and doesn't look like it does anything?

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -126,7 +126,7 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
           aria-controls={menuId}
           aria-label={t('Actions')}
         >
-          <Icon icon="mdi:more-vert" />
+          <Icon icon="mdi:more-horiz" />
```
skoeva: the icon change isn't necessary IMO, it seems fine

skoeva (commented):
thanks for this! left a few comments

frontend/src/components/App/Home/ClusterTable.tsx
jaehanbyun: Looking at other resource lists (e.g., Pods, Services, PVCs), they all show "Actions" as the column header. 
It felt inconsistent to hide it only in the ClusterTable. 

**Cluster Table Icon Tooltip**
image

**Storage Classes List**
Screenshot 2025-10-08 at 1 05 34 PM

frontend/src/components/App/Home/ClusterTable.tsx
jaehanbyun: Given that we have the "Actions" header, I aligned it with how other resource lists display their Actions column - both the header and the three-dot icon are center-aligned. Here's a comparison:
**Before: [right-aligned]**
image
**After: [center-aligned with header, matching other resource lists]**
Note: ClusterTable uses the base Table component, while other resource lists use ResourceListView, but the Actions column styling should be consistent across both.

image

frontend/src/components/App/Home/ClusterContextMenu.tsx
jaehanbyun: I checked the Actions columns in other resource tables (Pods, Services, etc.), and they all use horizontal ellipsis rather than vertical. 
This change unifies the icon choice across all resource lists.

jaehanbyun:
@skoeva Thanks for the review! 
I've left my thoughts on each of your comments above. 
Would appreciate your feedback on my approach. Thanks!

frontend/src/components/App/Home/ClusterTable.tsx
skoeva: I stand corrected. good catch

frontend/src/components/App/Home/ClusterContextMenu.tsx
skoeva: I think the resourcetable (as well as this clustertable) could probably use the vertical icon, you're right that it makes no sense why they would be different

frontend/src/components/App/Home/ClusterTable.tsx
skoeva: thanks for this as well, I realize this change is only needed with the horizontal three dots

frontend/src/components/App/Home/ClusterContextMenu.tsx
illume: We have horizontal "..." used inside the data columns, so for that reason I think vertical is better.

@jaehanbyun sound ok?

I see @skoeva made an issue for making them consistent vert
- https://github.com/kubernetes-sigs/headlamp/issues/4020

jaehanbyun: @illume yep, I understood.

jaehanbyun:
@skoeva @illume 
As discussed, I've updated the Actions column to use the vertical menu icon. PTAL.

joaquimrocha:
Thanks @jaehanbyun !

PR #4015, opened 2025-10-07
backend: auth: Include baseURL in Cookie Path

### LINKED ISSUE #4014: Set-Cookie path does not respect --baseURL option, causing authentication failure with reverse proxy
## Describe the bug
I am trying to deploy headlamp under reverse proxy with prefix path like `example.com/headlamp` using the baseURL option, e.g., --baseURL=/headlamp, but SA token authentication fails with this configuration.

I checked the request and suspect this is a bug due to the behavior of the /set-cookie endpoint. The response from the /set-cookie endpoint includes the `Set-Cookie` header with the path `/clusters/main` as follows:

```
Set-Cookie: headlamp-auth-main.0=XXXXXX; HttpOnly; SameSite=Strict; Path=/clusters/main; Max-Age=86400
```

I expect the endpoint to respect the baseURL option, so the path should include it as a prefix, e.g., `/headlamp/clusters/main`.	

https://github.com/kubernetes-sigs/headlamp/blob/7f6635292f7a135e2fede45ee8e12966f6b7ff0e/backend/pkg/auth/cookies.go#L87-L95

Authentication works when no prefix configuration is used.	

## To Reproduce
Steps to reproduce the bug:

1. Deploy headlamp Helm chart with the following values.yaml.

```yaml
ingress:
  enabled: true
  ingressClassName: nginx
  hosts:
    - host: example.com
      paths:
        - path: /headlamp
          type: Prefix
config:
  baseURL: /headlamp
```

2. Generate a token using `kubectl create token headlamp -n headlamp`

3. Input the token to `https://example.com/headlamp/` and get the `Error authenticating`

 - Installation type: Helm
 - Headlamp Version: 0.36.0

Yes (I will propose a PR)

appare45:
## Summary

This PR fixes a bug where the Cookie's `Path` attribute does not respect the `base url` option.

## Related Issue

Fixes #4014

## Changes

- Added `GetCookiePath` function for getting the path using the `baseURL`.
- Fixed the bug where the Cookie `Path` did not respect the `baseURL` option.

## Steps to Test

1. Start the backend with the `baseURL` option, e.g., run `/backend/headlamp-server -base-url=/headlamp -port=8081`.
2. Request the token setting endpoint (e.g., `/set-token`) using `curl`:

```
curl --request POST
--url http://localhost:8081/headlamp/clusters/main/set-token
--data '{"token": "XXXXX"}'
```

3.  Verify the `Set-Cookie` header in the response. The `Path` attribute should start with the configured `baseURL` (`/headlamp/clusters/main` in this example).

Expected result:

```
Set-Cookie: headlamp-auth-main.0=XXXXX; Path=/headlamp/clusters/main; Max-Age=86400; HttpOnly; SameSite=Strict
```

## Notes for the Reviewer

- I could not test this change with the OIDC authentication flow locally. As this change affects how the Cookie Path is set, please also verify that the OIDC login/logout flow works correctly when the base-url option is used.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -755,7 +755,7 @@ func createHeadlampHandler(config *HeadlampConfig) http.Handler {
 			}
 
 			// Set auth cookie
-			auth.SetTokenCookie(w, r, string(decodedState), rawUserToken)
+			auth.SetTokenCookie(w, r, string(decodedState), rawUserToken, config.BaseURL)
 
 			redirectURL += fmt.Sprintf("auth?cluster=%1s", decodedState)
 			http.Redirect(w, r, redirectURL, http.StatusSeeOther)
@@ -841,7 +841,7 @@ func (c *HeadlampConfig) refreshAndSetToken(oidcAuthConfig *kubeconfig.OidcConfi
 		}
 
 		// Set refreshed token in cookie
-		auth.SetTokenCookie(w, r, cluster, newTokenString)
+		auth.SetTokenCookie(w, r, cluster, newTokenString, c.BaseURL)
 
 		c.telemetryHandler.RecordEvent(span, "Token refreshed successfully")
 	}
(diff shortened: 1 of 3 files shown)


illume (approved):
🎉 thanks!

illume:
@sniok can you please also have a look at this one?

sniok (approved):
Looks good. Thanks for the fix!

Issue #4008, opened 2025-10-04
Set up current namespace from the search


joaquimrocha:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

I would like the global search to have a shortcut for setting up the current namespace. This is so we avoid having to select the namespace from the dropdown.

## Describe the solution you'd like
1. When writing in the global search, if it matches a known namespace, it should offer to set up that namespace completion as the current namespace, e.g.:
```
[ my- ]
| Set namespace: my-namespace1
| Set namespace: my-own-namespace
```
2. In addition to the above, always offer to set up the namespace to the value currently in the global search. This could work as a way to set up allowed namespaces without actually having to remember to configure them

PR #4016, opened 2025-10-07
frontend: GlobalSearch: Set namespace from search

skoeva:
This change creates a shortcut for setting the current namespace from the global search. The options list will now allow users to set the current namespace either to the closest match in the namespace list or the value currently in the search bar.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/4008

### Testing
- [X] Open Headlamp in-cluster and navigate to the search bar at the top
- [X] Search for a namespace and ensure the option to set the namespace is visible
- [X] Search for a random string and ensure the option to set the namespace is visible

image


frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -175,7 +178,68 @@ export function GlobalSearchContent({
 
   // Resource search items
   const resources = useSearchResources();
+  const resourcesByKind = useMemo(
```
sniok: Since only namespaces are used from this map, maybe we can skip building the map? I think it'll be a bit simpler
```tsx
const namespaces = useMemo(() => resources.filter(r => r.kind === 'Namespace'), [resources]);
```

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -175,7 +178,68 @@ export function GlobalSearchContent({
...
+
+    const trimmedQuery = query.trim();
+    if (trimmedQuery.length > 0) {
+      addOption(trimmedQuery);
+    }
+
+    Array.from(knownNamespaces)
+      .sort((a, b) => a.localeCompare(b))
+      .forEach(namespaceValue => {
+        if (namespaceValue !== trimmedQuery) {
```
sniok, line 236: I'm not sure why this condition is needed. If I fully type a namespace, for example "default" then it won't show the option

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -175,7 +178,68 @@ export function GlobalSearchContent({
...
+    const options: SearchResult[] = [];
+
+    const addOption = (namespaceValue: string) => {
+      if (!namespaceValue) {
+        return;
+      }
+
+      options.push({
+        id: `set-namespace-${namespaceValue}`,
+        subLabel: t('translation|Namespace'),
```
sniok: I think sublabel should be different from the regular search result for the namespace so it stands out a bit

frontend/src/components/globalSearch/GlobalSearchContent.tsx
```diff
@@ -175,7 +178,68 @@ export function GlobalSearchContent({
...
   // Resource search items
   const resources = useSearchResources();
+  const resourcesByKind = useMemo(
+    () => new Map(resources.map(resource => [resource.kind, resource])),
+    [resources]
+  );
   const loading = resources.filter(it => it.isLoading).map(it => it.kind);
+  const namespaceItems = (resourcesByKind.get(Namespace.kind)?.items ?? []) as Namespace[];
+  const namespaceOptions = useMemo(() => {
+    const knownNamespaces = new Set<string>();
```
sniok: This is more of a styling comment, so feel free to ignore, but the same Set can be initialized more concisely 

```tsx
const knownNamespaces = new Set<string>([
  // Loaded namespaces
  ...namespaceItems.map(n => n.metadata.name),
  // Allowed namespaces from settings
  ...selectedClusters.flatMap(c => loadClusterSettings(c)?.allowedNamespaces)
].filter(Boolean));
```

frontend/src/components/globalSearch/GlobalSearchContent.tsx
skoeva: done

joaquimrocha:
Nice! Works great!
/approve
/lgtm

PR #3953, opened 2025-09-25
Allow running the whole project using npm

joaquimrocha:
## Summary

This PR allows building and running Headlamp by using npm alone, instead of having to install and use make for some things and using npm for others.

## Steps to Test

1. Check the scrips with `npm run`
2. Check different npm commands to build/run Headlamp to see if it works

## Still Missing
- [ ] Replacing make's inner commands by npm
- [ ] Replacing Github workflows to use npm

PR #4034, opened 2025-10-13
frontend: Add webSocket polling and initial timeouts

joaquimrocha:
## Summary

This change will avoid tests failing unexpectedly.

Separated from #3953 .


joaquimrocha:
I've closed the PR where this change was needed. So I don't think this is needed in the end.

PR #4056, opened 2025-10-17
Add root npm no workspaces

joaquimrocha:
## Summary

This PR is based on #3953 but doesn't convert the project to workspaces, so it should be easier to test and accept.

This PR allows building and running Headlamp by using npm alone, instead of having to install and use make for some things and using npm for others.

## Steps to Test

- [ ] `npm install` in the root of Headlamp; check if it succeeds
  - [ ] Pass (@sniok: Works)
- [ ] Make sure Headlamp is run in development mode: `npm run start` 
  - [ ] Pass (@sniok: Works although I wish this command didn't run `npm install` multiple times and didn't build backend)
- [ ] Make sure Headlamp app is built when running `npm run app:build`; test the build application
  - [ ] Pass (@sniok: Works. There are duplicated npm installs in this script. I'd expect that all scripts assume that the package are already installed so the usage would be `npm ci && npm run app:build`)
- [ ] Make sure Headlamp image is built using `npm run image:build` + test the image
  - [ ] Pass
- [ ] Run the Build app Actions on Github, make sure they work and test their apps
  - [ ] Linux
    - [ ] Pass
  - [ ] Windows
    - [ ] Pass
  - [ ] Mac
    - [ ] Pass
- [ ] Build headlamp-plugin like previously + generate a new plugin with this new version
  - [ ] Pass

### DIFF
--- a/app/e2e-tests/README.md
+++ b/app/e2e-tests/README.md
@@ -41,17 +41,17 @@ To run the tests for the web mode, you will need to have the backend running. Fo
   `cd headlamp`
 
 - run the following command
-  `make backend` followed by `make run-backend`
+  `npm run backend:build` followed by `npm run backend:start`
 
 ### Frontend
 
 To run the tests for the web mode, you will need to have the frontend running. Follow the steps below to run the frontend:
 
 - cd into the headlamp directory in a separate terminal
-  `cd headlamp/frontend`
+  `cd headlamp`
 
 - run the following command
-  `make frontend` followed by `make run-frontend`
+  `npm run frontend:build` followed by `npm run frontend:start`
 
 ### Running the tests
 
--- a/backend/README.md
+++ b/backend/README.md
@@ -1,8 +1,8 @@
 # Quickstart
 
 ```bash
-make backend
-make run-backend
+npm run backend:build
+npm run backend:start
 ```
 
 See more detailed [Headlamp backend documentation on the web](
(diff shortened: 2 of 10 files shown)


package-lock.json
```diff
@@ -0,0 +1,662 @@
...
+        "eslint-config-prettier": "^9.1.0",
+        "eslint-plugin-import": "^2.24.2",
+        "eslint-plugin-jsx-a11y": "^6.9.0",
+        "eslint-plugin-react": "7.35.0",
+        "eslint-plugin-react-hooks": "^4.6.2",
+        "eslint-plugin-simple-import-sort": "^12.1.1",
+        "eslint-plugin-unused-imports": "^4.1.3"
+      }
+    },
+    "frontend": {
```
illume: Looks like the package-lock.json along with node_modules needs to be removed and regenerated?

package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "headlamp-root",
```
illume, line 2: I'm not sure it matters... but just in case: `@headlamp-k8s/headlamp-root` ?

illume (approved):
🎉 thanks!

package.json
joaquimrocha: Doesn't matter since the package is never meant to be published. I thought an appropriate name would just be headlamp, but if we then add workspaces, it could clash with the frontend. Let's keep this one.

package-lock.json
illume: @joaquimrocha I don't think you pushed this change yet.

joaquimrocha: Ah yeah. I had a failed origin... Now should be good.

PR #4063, opened 2025-10-19
.github/workflows/backend-test.yml: Fix backend coverage comment

illume:
It used to be that a coverage report would be sent as a comment to
the PR if the PR has changes in backend/ folder.

### Testing

- Try adding a backend/ folder commit, and see the PR comment come

### DIFF
--- a/.github/workflows/backend-test.yml
+++ b/.github/workflows/backend-test.yml
@@ -22,6 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      issues: write # needed for commenting on issues for coverage changes
       pull-requests: write # needed for commenting on PRs for coverage changes
     steps:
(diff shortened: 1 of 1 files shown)


illume:
The main issue is that forked PRs don't get access to a secret which gives them permission to comment on the PR.
Because none of our PRs use our own branches any more, it's not commenting anymore.

A possible fix?

Two step workflow:
- backend-test.yml, runs tests, uploads artefacts
- post-coverage-comment.yml, is triggered on workflow_run.completed of backend test finishes, downloads artefacts from the github.event.workflow_run.id and posts the comment (because it has more privs this is allowed)

Still not running code with secrets, the stuff requiring privs is done in a separate workflow. It can download from the forked artefacts because triggered jobs are given a run id.

PR #4071, opened 2025-10-21
frontend: plugins: Bump vite to 6.4.1

skoeva:
These changes bump vite in frontend and the plugins to v6.4.1 as desired.

### Testing
- [x] Run `npm run start` and make sure the app is working as expected


illume (approved):
🎉 thanks!

PR #4076, opened 2025-10-22
frontend: docs: headlamp-plugin: Fix storybook theme, use correct new logo

illume:
- **docs: frontend: headlamp-plugin: Fix storybook fonts and use new correct logo**
- **headlamp-plugin: Fix storybook to include base mocks and css**

### Screenshot

Note the Headlamp font and logo match the frontend storybook now.

Plugin storybook screenshot

Screenshot 2025-10-23 at 15 58 28

#### Headlamp storybook

Screenshot 2025-10-23 at 15 58 28

### Testing

```shell
# build the headlamp-plugin tar.gz from that 4076  in headlamp repo 
gh pr checkout 4076
cd plugins/headlamp-plugin
rm kinvolk*
npm run build && npm pack
```
 
Install the new storybook with theme changes, and note the fonts are slightly different.
```shell
cd plugins/examples/podcounter
npm i ../../headlamp-plugin/kinvolk*
npm run storybook
```


plugins/headlamp-plugin/config/.storybook/headlamp_light.svg
joaquimrocha, line 1: We should use the right logo.

frontend/.storybook/baseMocks.ts
```diff
@@ -17,6 +17,8 @@
 import { http, HttpResponse } from 'msw';
 import { NODE_DUMMY_DATA } from '../src/components/node/storyHelper';
 
+// NOTE: Please keep this synced with headlamp/plugins/headlamp-plugin/config/.storybook/baseMocks.ts
```
joaquimrocha, line 20: Can we have a test or a way to copy this? If we do already maybe the comment is not needed?

joaquimrocha (changes requested):

frontend/.storybook/baseMocks.ts
illume: Feel free to propose something in a separate or.

I don’t see how considering the number of slightly different things that are needed. At least this gets it working.

plugins/headlamp-plugin/config/.storybook/headlamp_light.svg
illume: Ed: Ah, right… I get it now. I can update the storybooks and docs to use the right logo.

frontend/.storybook/baseMocks.ts
illume: Yeah, I don’t think it would be practical to do in a short time and the changes would have to be manually verified anyway. So I don’t think it’s useful for the large amount of work required.

plugins/headlamp-plugin/config/.storybook/headlamp_light.svg
illume: Added the new logo to the docs, and frontend storybook as well.

vyncent-t:
tested LGTM 👍 

seen w/ @illume @skoeva

PR #4079, opened 2025-10-23
frontend/KubeMetadata: plugins/examples/projects: Small fixes

illume:
Some typescript related fixes which show up with new headlamp-plugin version.

- **headlamp-plugin: Add missing registerProjectDeleteButton export**
- **frontend: KubeMetadata: Add version of metadata for creation**
- **plugins/examples/projects: Fix for using id instead of name field**

@sniok I'm not sure if the name field is supposed to be there? I wasn't sure, so used the id field in the plugin example.


illume:
I had to change metadata so there's a separate type for creation and one for otherwise. So creation with apply can avoid supplying uid and creationTimestamp. So now there's a KubeMetadataCreate, and apply has an extra signature for that.
(just making it optional caused like 50 errors inside frontend/)

illume:
@sniok thanks for the clarification

frontend/src/lib/k8s/api/v1/apply.ts
```diff
@@ -26,11 +26,19 @@ import { resourceDefToApiFactory } from './factories';
...
+ * Overloads:
+ * - When called with a KubeObjectInterfaceCreate body, the parameter type is the create type.
+ * - Otherwise it accepts a full KubeObjectInterface.
+ *
  * @param body - The kubernetes object body to apply.
  * @param clusterName - The cluster to apply the body to. By default uses the current cluster (URL defined).
  *
  * @returns The response from the kubernetes API server.
  */
+export async function apply<T extends KubeObjectInterfaceCreate>(
```
sniok, line 38: I think you could use `RecursivePartial` type helper here to mark all fields in T as optional

illume: Sorry, I don't understand. This change is about making the metadata uid and creationTimestamp optional. At least metadata name is required.

Can you please expand on what you mean?

PR #4088, opened 2025-10-24
.github/build-container.yml: Cleanup disk space to avoid running out

illume:
Sometimes this job is running out of disk space.

- removed node_modules, apt packages, and deleted clusters before saving container images to cache
- compressed container images with gzip

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -51,20 +51,21 @@ jobs:
     - name: Try the cluster!
       run: kubectl get pods -A
     - name: Restore image-cache Folder
-      id: cache-image-restore
+      id: cache-image-restore2
       uses: actions/cache/restore@0c907a75c2c80ebcb7f088228285e798b750cf8f # v4.2.1
       with:
         path: ~/image-cache
         # cache the container image. All the paths this PR depends on except the e2e-tests folder for the key.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-container.yml
```diff
@@ -61,8 +61,10 @@ jobs:
       if: steps.cache-image-restore.outputs.cache-hit == 'true'
       run: |
         export SHELL=/bin/bash
-        docker load -i ~/image-cache/headlamp-plugins-test.tar
-        docker load -i ~/image-cache/headlamp.tar
+        mkdir -p ~/image-cache
+        gzip -dc ~/image-cache/headlamp-plugins-test.tar.gz | docker load
+        gzip -dc ~/image-cache/headlamp.tar.gz | docker load
+
```
skoeva: these extra lines could probably be removed to match the rest of the file

illume:
Seems to work... https://github.com/kubernetes-sigs/headlamp/actions/runs/18788084557/job/53611366877?pr=4088#step:19:199

.github/workflows/build-container.yml
illume: Fixed, thanks.

illume:
Changed the cache key, because there will be a few old PRs saving non .gz files into the cache. With the new cache key the two of them can coexist.

PR #4092, opened 2025-10-27
backend: auth: Fetch user info from header

skoeva:
This change swaps the `GetTokenFromCookie` call in the `HandleMe` function to `ParseClusterAndToken`, which allows us to fetch user info from a header with a fall back to the cookie.


ashu8912 (approved):
Tested with entra id and aks cluster. With oauth2proxy config, and it works

PR #4098, opened 2025-10-28
frontend: EditButton: Use JSON patch to prevent resource version conflicts

### LINKED ISSUE #4097: YAML editor produces 409 Conflict errors on frequently updated resources (e.g., Nodes)
## Describe the bug
When editing Kubernetes resources through the YAML editor, a 409 Conflict error occurs if the resource is modified by the system between opening the editor and saving changes. 
This is particularly problematic for resources that are frequently updated by the system, such as Nodes.

## To Reproduce
Steps to reproduce the bug:
1. Open the YAML editor for a Node resource
2. Make a small change (e.g., modify a taint or label)
3. Wait a few seconds (during which the kubelet updates the Node's status)
4. Click "Save & Apply"
5. Observe 409 Conflict error: `"Operation cannot be fulfilled on nodes \"node-name\": the object has been modified; please apply your changes to the latest version and try again"`

**What happens internally:**

```yaml
# Step 1: User opens the editor (resourceVersion: 1000)
apiVersion: v1
kind: Node
metadata:
  name: jhb-master-1
  resourceVersion: "1000"  # Captured at this moment
spec:
  taints:
    - key: node-role
      effect: NoSchedule

# Step 2: While user is editing, system updates the node status (resourceVersion: 1001)
status:
  conditions:
    - type: Ready
      status: True

# Step 3: User clicks Save&Apply button→ 409 Conflict!
# Editor sends resourceVersion 1000, but actual version is 1001
```

 - Installation type: Mac-Homebrew
 - Headlamp Version: latest

Yes (I will propose a PR)

## Additional Context

jaehanbyun:
## Summary

This PR fixes 409 Conflict errors in the YAML editor. 
The editor now sends only the changed fields instead of the entire resource, significantly reducing conflicts on frequently updated resources like Nodes.

## Related Issue

Fixes #4097 

### Before
The editor performed a full PUT request with the entire resource object:

```typescript
async function updateFunc(newItem: KubeObjectInterface) {
  // Simply sent the entire edited resource
  await item.update(newItem);  // Uses PUT request
}
```

**Problem:** If the resource's `resourceVersion` changed between opening the editor and saving (e.g., kubelet updating Node status), Kubernetes rejected the request with 409 Conflict.

### After
The editor now captures the original snapshot and sends only the diff:

```typescript
async function updateFunc(newItem: KubeObjectInterface) {
  // Calculate JSON Patch: diff between original and edited
  const patches = compare(originalResourceRef.current, newItem);
  
  // Send PATCH with json-patch format
  await patch(url, patches, true, {
    cluster: item._clusterName,
    headers: { 'Content-Type': 'application/json-patch+json' }
  });
}
```

**Benefit:** Only changed fields are sent, dramatically reducing conflict probability even when the resource is modified by other processes.

## Steps to Test

1. Navigate to a Node page
2. Click the Edit button to open the YAML editor
3. Make a small change (e.g., add a label or modify a taint)
4. Wait a few seconds (kubelet will update the Node status in the background)
5. Click "Save & Apply"
6. Verify the change is applied successfully without a 409 Conflict error

### Before
image

### After
image

## Notes for the Reviewer

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -56,6 +56,7 @@
         "eslint-plugin-simple-import-sort": "^12.1.1",
         "eslint-plugin-unused-imports": "^4.1.3",
         "fake-indexeddb": "^6.0.0",
+        "fast-json-patch": "^3.1.1",
         "fuse.js": "^7.0.0",
         "humanize-duration": "^3.27.2",
         "i18next": "^23.15.1",
@@ -7942,6 +7943,11 @@
       "engines": {
         "node": ">= 6"
       }
+    },
+    "node_modules/fast-json-patch": {
+      "version": "3.1.1",
+      "resolved": "https://registry.npmjs.org/fast-json-patch/-/fast-json-patch-3.1.1.tgz",
+      "integrity": "sha512-vf6IHUX2SBcA+5/+4883dsIjpBTqmfBjmYiWK1savxQmFk4JfBMLa7ynTYOs1Rolp/T1betJxHiGD3g1Mn8lUQ=="
     },
     "node_modules/fast-json-stable-stringify": {
       "version": "2.1.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -57,6 +57,7 @@
     "eslint-plugin-simple-import-sort": "^12.1.1",
     "eslint-plugin-unused-imports": "^4.1.3",
     "fake-indexeddb": "^6.0.0",
+    "fast-json-patch": "^3.1.1",
     "fuse.js": "^7.0.0",
     "humanize-duration": "^3.27.2",
     "i18next": "^23.15.1",
--- a/frontend/src/components/common/Resource/EditButton.tsx
+++ b/frontend/src/components/common/Resource/EditButton.tsx
@@ -15,10 +15,13 @@
  */
 
 import { Icon } from '@iconify/react';
+import { compare } from 'fast-json-patch';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { useLocation } from 'react-router-dom';
+import { patch } from '../../../lib/k8s/api/v1/clusterRequests';
+import { KubeObjectEndpoint } from '../../../lib/k8s/api/v2/KubeObjectEndpoint';
 import { KubeObject } from '../../../lib/k8s/KubeObject';
 import { KubeObjectInterface } from '../../../lib/k8s/KubeObject';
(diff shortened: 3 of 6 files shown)


illume:
This is a really nice fix.

I need some time to test it though. So I marked it for the 0.38.0 release.

My initial question is what happens if someone else updates it in the mean time? Do you handle this case?

jaehanbyun:
@illume 
Thanks for the review!

I tested concurrent edits to answer your question:

- Different fields (e.g., different label keys): both updates apply successfully.

- Same field (same label key, different values): the later request wins. The second replace on the same path overwrites the first one (effectively last-write-wins).

- Delete vs modify on the same key: also last-write-wins.
  - If A removes labels.key1 and B changes labels.key1 from value1 → value2, the one that reaches the API server later determines the outcome:
    - A later → key1 is deleted.
    - B later → key1: value2 is present.

illume:
I wonder if last-write-wins is the way in these cases? I think it could be a bit dangerous.

Currently it doesn't let them apply, so they have to try again. It's not a good experience, but at least they need to try make the change again and take into consideration the changes.

The alternative could be:
- 1) let Different fields (e.g., different label keys): show the user there are different fields. 
- 2) Same field (same label key, different values): show the user the conflicted field and values, with last one shown to apply.
- 3) Delete vs modify on the same key: show the user the conflicted field and values, with last one shown to apply.

Basically show them the conflict with a default of the last write... but make them confirm the change... and let them have an opportunity to change it (maybe keeping the other change).

What do you think?

jaehanbyun:
@illume Yeah, handling the three cases as you described would definitely provide a better UX.
So, before implementing it, I think we should follow an approach(3 ver diff) like this:

1. Prepare the base (when the editor opens), theirs (latest from server at "Save & Apply"), and ours (user edits).

2. Compare base→theirs and base→ours to detect only “path-level” conflicts.

3. If no conflicts, but there are differences between base and theirs, show a modal highlighting those changes.

4. If conflicts exist, show a modal listing the conflicted fields with ours/theirs previews.
Default to ours (last edit) but require confirmation.
Then merge based on selections and apply a JSON Patch against the latest version.

What do you think about proceeding with this approach before we start implementing it?

illume:
@jaehanbyun I spent some time on Friday thinking about this....
I think your proposal would be very very good.

(Sorry it took quite some time to get to this)

jaehanbyun:
@illume 
Totally understand it takes time, so no worries.
I’ll go ahead and implement this based on my proposal and push a commit soon.

illume:
@jaehanbyun There's a related PR to this at:
- https://github.com/kubernetes-sigs/headlamp/pull/4269

What do you think?

PR #4099, opened 2025-10-28
Remove usage of execSync in scripts

skoeva:
These changes replace `execSync` with `execFileSync` in the two places where
dynamic values are interpolated into a shell command string.

- `app/electron/main.ts`: `list-plugins` built its command from `backendPath`,
  which comes from `process.resourcesPath`. An install path with a space or a
  shell metacharacter would break the command or run arbitrary shell.
- `tools/releaser/src/utils/git.ts`: `version` and `branchName` come from
  releaser CLI arguments, and `git add` interpolated filesystem paths.

Note: the `git` calls in the releaser now use `stdio: 'inherit'` where they
previously defaulted to `'pipe'`, so their output is visible.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { ChildProcessWithoutNullStreams, exec, execSync, spawn } from 'child_process';
+import { ChildProcessWithoutNullStreams, exec, execFileSync, spawn } from 'child_process';
 import { randomBytes } from 'crypto';
 import dotenv from 'dotenv';
 import {
@@ -132,7 +132,7 @@ const args = yargs(hideBin(process.argv))
     () => {
       try {
         const backendPath = path.join(process.resourcesPath, 'headlamp-server');
-        const stdout = execSync(`${backendPath} list-plugins`);
+        const stdout = execFileSync(backendPath, ['list-plugins']);
         process.stdout.write(stdout);
         process.exit(0);
       } catch (error) {
@@ -1383,7 +1383,7 @@ async function getHeadlampPIDsOnPort(port: number): Promise<number[] | null> {
 function killProcess(pid: number) {
   if (process.platform === 'win32') {
     // Otherwise on Windows the process will stick around.
-    execSync('taskkill /pid ' + pid + ' /T /F');
+    execFileSync('taskkill', ['/pid', String(pid), '/T', '/F']);
   } else {
     process.kill(pid, 'SIGHUP');
   }
--- a/tools/releaser/src/utils/git.ts
+++ b/tools/releaser/src/utils/git.ts
@@ -1,11 +1,13 @@
-import { execSync } from 'child_process';
+import { execFileSync } from 'child_process';
 import path from 'path';
 import fs from 'fs';
 import { isValidVersion } from './version.js';
 
 export function getRepoRoot(): string {
   try {
-    const gitRoot = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
+    const gitRoot = execFileSync('git', ['rev-parse', '--show-toplevel'], {
+      encoding: 'utf-8',
+    }).trim();
     return gitRoot;
   } catch (error) {
     console.error('Error: Not in a git repository');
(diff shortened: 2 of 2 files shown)


illume:
Thanks.

I want to hold off on this until after 0.37.0

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1024,15 +1023,23 @@ function upgrade(packageFolder, skipPackageUpdates, headlampPluginVersion) {
    */
   function runCmd(cmd, folder) {
     console.log(`Running cmd:"${cmd}" inside of ${folder} abs: "${resolve(folder)}"`);
-    try {
-      child_process.execSync(cmd, {
-        stdio: 'inherit',
-        encoding: 'utf8',
-      });
-    } catch (e) {
+    const result = child_process.spawnSync(cmd, {
```
joaquimrocha: Curious, why is this one not execFileSync? (assuming you can split cmd?)
Not saying it's better, just curious.

plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -1260,16 +1267,16 @@ function tsc(packageFolder) {
...
-      {
-        stdio: 'inherit',
-        cwd: packageFolder,
-        encoding: 'utf8',
-      }
-    );
-  } catch (e) {
+  const command =
+    './node_modules/.bin/storybook dev -p 6007 -c node_modules/@kinvolk/headlamp-plugin/config/.storybook';
+  const result = child_process.spawnSync(command, {
```
joaquimrocha: Same question about execFileSync here.

app/electron/main.ts
```diff
@@ -1090,7 +1098,13 @@ async function getRunningHeadlampPIDs() {
 function killProcess(pid: number) {
   if (process.platform === 'win32') {
     // Otherwise on Windows the process will stick around.
-    execSync('taskkill /pid ' + pid + ' /T /F');
+    const taskKill = spawnSync('taskkill', ['/pid', String(pid), '/T', '/F'], { stdio: 'ignore' });
```
joaquimrocha: Same question.

plugins/headlamp-plugin/bin/headlamp-plugin.js
skoeva: just copilot doing things. they can both work (for the other cases), I guess execFileSync would be simpler. will look at these individually

in this case, looks like splitting the cmd is a headache when it's a shell string anyway, so reverted

app/electron/main.ts
skoeva: both work but execfilesync seems simpler here, updating

app/scripts/build-backend.js
```diff
@@ -15,7 +15,7 @@ exports.default = async context => {
     osName = 'Windows_NT';
   }
 
-  execSync('make backend', {
+  execFileSync('make', ['backend'], {
```
illume: Unrelated to this PR... but we shouldn't use `make` now, but use the npm backend build script instead.

plugins/headlamp-plugin/bin/headlamp-plugin.js
illume: looks like this was done

tools/releaser/src/commands/start.ts
```diff
@@ -22,7 +22,10 @@ export function startRelease(releaseVersion: string): void {
 
     // Run npm install in app directory
     console.log(chalk.blue('Running npm install in app directory...'));
-    execSync('npm install', { stdio: 'inherit', cwd: path.join(repoRoot, 'app') });
+    execFileSync('npm', ['install'], {
+      stdio: 'inherit',
+      cwd: path.join(repoRoot, 'app'),
```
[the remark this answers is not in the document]
illume: This is a real issue, I tested it in node on windows with cmd.

illume (changes requested):
I think these require testing. The copilot found one issue which turned out to be a real one, so I'm worried that these changes will break other things.

It might be worth considering which if any of these is a security problem really to reduce the amount of testing that needs to be done.

tools/releaser/src/commands/start.ts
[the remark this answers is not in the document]
skoeva: this could probably be reverted, doing this now

illume (approved):
🎉 thanks!

PR #4100, opened 2025-10-29
frontend: Improve empty state for the cluster list view

joaquimrocha:
## Summary

When the cluster list is empty, Headlamp looks very empty and there is no clue on what to do.
This PR fixes that by adding a call to action (in app mode). When in-cluster, it just states there are no clusters (since one cannot add it).

## Steps to Test

1. Make a backup of your ~/.kube/config, remove it, and then open headlamp -> Realize how now there's a much better looking empty state for the cluster list view

Screenshot showing the changes


illume (approved):
🎉 thanks!

PR #4103, opened 2025-10-30
frontend: TopBar: backend: auth: Add me-user-info-url config

illume:
This PR adds support for optionally fetching additional user information from an external endpoint (like oauth2proxy's userinfo endpoint) to populate the user info displayed in the TopBar.

- Frontend fetches additional user info from the meUserInfoURL if provided
- Adds a new meUserInfoURL configuration option across frontend, backend(`me-user-info-url`), and Helm charts
- Backend passes the userInfoURL to the frontend in the /me endpoint response
- Adds a `config.oidc.meUserInfoURL` string with a default of `""`  to be returned by the `/me` endpoint (which is then fetch()'d by the frontend/. For oauth2proxy `"/oauth2/userinfo"` can be used.

### DIFF
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -24,6 +24,7 @@ const (
 	DefaultMeUsernamePath = "preferred_username,upn,username,name"
 	DefaultMeEmailPath    = "email"
 	DefaultMeGroupsPath   = "groups,realm_access.roles"
+	DefaultMeUserInfoURL  = ""
 )
 
 type Config struct {
@@ -56,6 +57,7 @@ type Config struct {
 	MeUsernamePath            string `koanf:"me-username-path"`
 	MeEmailPath               string `koanf:"me-email-path"`
 	MeGroupsPath              string `koanf:"me-groups-path"`
+	MeUserInfoURL             string `koanf:"me-user-info-url"`
 	OidcUsePKCE               bool   `koanf:"oidc-use-pkce"`
 	// telemetry configs
 	ServiceName        string   `koanf:"service-name"`
@@ -230,7 +232,7 @@ func setKubeConfigPath(config *Config) {
 }
 
 // ApplyMeDefaults trims and applies defaults to the JMESPath expressions used for the /me endpoint.
-func ApplyMeDefaults(usernamePath, emailPath, groupsPath string) (string, string, string) {
+func ApplyMeDefaults(usernamePath, emailPath, groupsPath, userInfoURL string) (string, string, string, string) {
 	username := strings.TrimSpace(usernamePath)
 	if username == "" {
 		username = DefaultMeUsernamePath
@@ -246,15 +248,21 @@ func ApplyMeDefaults(usernamePath, emailPath, groupsPath string) (string, string
 		groups = DefaultMeGroupsPath
 	}
 
-	return username, email, groups
+	userInfo := strings.TrimSpace(userInfoURL)
+	if userInfo == "" {
+		userInfo = DefaultMeUserInfoURL
+	}
+
+	return username, email, groups, userInfo
 }
 
 // setMeDefaults ensures the /clusters/{clusterName}/me claim paths fall back to defaults when unset.
 func setMeDefaults(config *Config) {
-	config.MeUsernamePath, config.MeEmailPath, config.MeGroupsPath = ApplyMeDefaults(
+	config.MeUsernamePath, config.MeEmailPath, config.MeGroupsPath, config.MeUserInfoURL = ApplyMeDefaults(
 		config.MeUsernamePath,
 		config.MeEmailPath,
 		config.MeGroupsPath,
+		config.MeUserInfoURL,
 	)
 }
 
@@ -407,6 +415,8 @@ func addOIDCFlags(f *flag.FlagSet) {
 		"Comma separated JMESPath expressions used to read email from the JWT payload")
 	f.String("me-groups-path", DefaultMeGroupsPath,
 		"Comma separated JMESPath expressions used to read groups from the JWT payload")
+	f.String("me-user-info-url", DefaultMeUserInfoURL,
+		"URL to fetch additional user info for the /me endpoint. For oauth2proxy /oauth2/userinfo can be used.")
 }
 
 func addTelemetryFlags(f *flag.FlagSet) {
--- a/charts/headlamp/README.md
+++ b/charts/headlamp/README.md
@@ -90,6 +90,7 @@ $ helm install my-headlamp headlamp/headlamp \
 | config.oidc.secret.name | string | `"oidc"` | Name of the OIDC secret |
 | config.oidc.externalSecret.enabled | bool | `false` | Enable using external secret for OIDC |
 | config.oidc.externalSecret.name | string | `""` | Name of external OIDC secret |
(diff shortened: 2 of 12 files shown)


backend/pkg/config/config.go
```diff
@@ -24,6 +24,8 @@ const (
 	DefaultMeUsernamePath = "preferred_username,upn,username,name"
 	DefaultMeEmailPath    = "email"
 	DefaultMeGroupsPath   = "groups,realm_access.roles"
+	// DefaultMeUserInfoURL  = "/oauth/userinfo"
```
illume: I set this default this to on to the oauth2 proxy path. I think we wanted to move towards recommending most folks use oauth2proxy?

Existing installations not using oauth2proxy will get one 404 for each user (until they configure it off). So it's not such a bad outcome for existing installations. The benefit is that oauth2proxy installations (which we are moving towards recommending) is that they will work without configuration.

illume:
@skoeva updated the path to /oauth2/userinfo

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,43 @@ export default function TopBar({}: TopBarProps) {
...
+        }
+
+        const shouldFetchUserInfo =
+          typeof res.userInfoURL === 'string' && res.userInfoURL.length > 0;
+        if (shouldFetchUserInfo) {
+          try {
+            const ui: any = await fetch(res.userInfoURL, {
+              credentials: 'include',
+              headers: {
+                'Content-Type': 'application/json',
```
skoeva: is this header necessary to include? the fetch worked in the console without it

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,43 @@ export default function TopBar({}: TopBarProps) {
...
+              if (!r.ok) {
+                throw new Error('Could not fetch user info');
+              }
+              return r.json();
+            });
+
+            if (!ui) {
+              return null;
+            }
+            if (!ui.preferredUsername && !ui.email) {
```
skoeva: I feel like these two null checks can be combined

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,43 @@ export default function TopBar({}: TopBarProps) {
...
 
-        return res ? { username: res.username, email: res.email } : null;
+        if (!res) {
+          return null;
+        }
+
+        const shouldFetchUserInfo =
+          typeof res.userInfoURL === 'string' && res.userInfoURL.length > 0;
+        if (shouldFetchUserInfo) {
+          try {
```
skoeva: maybe this can be optimized? I don't think we should have a nested try

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,43 @@ export default function TopBar({}: TopBarProps) {
...
         });
 
-        return res ? { username: res.username, email: res.email } : null;
+        if (!res) {
+          return null;
+        }
+
+        const shouldFetchUserInfo =
+          typeof res.userInfoURL === 'string' && res.userInfoURL.length > 0;
+        if (shouldFetchUserInfo) {
```
skoeva: we could probably also avoid using this flag

maybe something like this would suffice?

```tsx
const userInfoURL = res.userInfoURL?.trim();
if (!userInfoURL) return null;
```

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,43 @@ export default function TopBar({}: TopBarProps) {
...
+          try {
+            const ui: any = await fetch(res.userInfoURL, {
+              credentials: 'include',
+              headers: {
+                'Content-Type': 'application/json',
+              },
+            }).then(r => {
+              if (!r.ok) {
+                throw new Error('Could not fetch user info');
```
skoeva: is it necessary to throw this? we already return null silently in the catch block, wouldn't the check on line 126 suffice to handle this case?

skoeva (commented):
thanks for taking care of this! left some comments

frontend/src/components/App/TopBar.tsx
illume: It's strictly correct to set it when requesting json, even if the endpoint currently works without it.

frontend/src/components/App/TopBar.tsx
illume: fixed

frontend/src/components/App/TopBar.tsx
illume: fixed

frontend/src/components/App/TopBar.tsx
illume: I was thinking this should throw because the URL is explicitly configured, and if there is an error then the issue could be noticed more easily with the throw.

(I defer to you though... what should the code be here?)

illume:
@skoeva I addressed your notes, and added helm chart configuration. Please have a look at the two convos still open.

Note, I defaulted to leaving this off. Because I think it's a bit better to not have 404s and such for people who don't use oauth2proxy. Sound ok? 

So when testing you'll need to set the `-me-user-info-url=/oauth2/userinfo` for the backend.

frontend/src/components/App/TopBar.tsx
illume: fixed

frontend/src/components/App/TopBar.tsx
```diff
@@ -103,7 +103,34 @@ export default function TopBar({}: TopBarProps) {
...
-        return res ? { username: res.username, email: res.email } : null;
+        if (!res) {
+          return null;
+        }
+
+        if (!(typeof res.userInfoURL === 'string' && res.userInfoURL.length > 0)) {
+          return { username: res.username, email: res.email };
+        }
+
+        const ui: any = await fetch(res.userInfoURL, {
```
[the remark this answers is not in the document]
illume: fixed

charts/headlamp/README.md
```diff
@@ -90,6 +90,7 @@ $ helm install my-headlamp headlamp/headlamp \
 | config.oidc.secret.name | string | `"oidc"` | Name of the OIDC secret |
 | config.oidc.externalSecret.enabled | bool | `false` | Enable using external secret for OIDC |
 | config.oidc.externalSecret.name | string | `""` | Name of external OIDC secret |
+| config.oidc.meUserInfoURL | string | `""` | Name of url to oauth2proxy userinfo endpoint. Set to /oauth2/userinfo if using oauth2proxy |
```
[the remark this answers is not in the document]
illume: fixed

illume:
I fixed a couple of things the copilot mentioned.

frontend/src/components/App/TopBar.tsx
skoeva: fair, maybe the error message should be more descriptive to reflect this

we could add the url in the message to easily catch a typo or wrongly submitted one

``throw new Error(`Could not fetch user info from ${res.userInfoURL}`);``

illume: fixed

joaquimrocha (approved):

illume:
@skoeva Sorry, didn't notice you pushed. It was the same change I pushed though.

skoeva:
@illume just tried to rebase but I realized it was against my fork's main branch and not yours so I undid it

illume:
Ok, cool. I added that log change you mentioned and rebased against main.

PR #4106, opened 2025-10-30
frontend: TopBar: Hide empty user info fields

skoeva:
This change hides the user info menu item when the fields are null.


illume:
@skoeva 

Let's add some story states for this?

```ts
export const EmptyUserInfo = PureTemplate.bind({});
EmptyUserInfo.args = {
  appBarActions: [],
  logout: () => {},
  cluster: 'ak8s-desktop',
  clusters: { 'ak8s-desktop': '' },
  userInfo: {
    email: '',
    username: '',
  },
};

export const UndefinedUserInfo = PureTemplate.bind({});
UndefinedUserInfo.args = {
  appBarActions: [],
  logout: () => {},
  cluster: 'ak8s-desktop',
  clusters: { 'ak8s-desktop': '' },
  userInfo: {
    email: undefined,
    username: undefined,
  },
};
```

frontend/src/components/App/TopBar.tsx
```diff
@@ -309,7 +309,7 @@ export const PureTopBar = memo(
         ? userInfo.email
         : undefined;
 
-    const renderUserMenu = !!isClusterContext && (
+    const renderUserMenu = !!isClusterContext && !!userDisplayName && (
```
illume: We could still want to render the user menu if there is no display name?

Because what if there's no name, and they want to logout?

frontend/src/components/App/TopBar.tsx
```diff
@@ -327,7 +327,7 @@ export const PureTopBar = memo(
           },
         }}
       >
-        {userInfo && (
+        {!!userDisplayName && (
```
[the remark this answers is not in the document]
skoeva: this isn't correct reasoning but looks like this line could still be removed. doing this now

frontend/src/components/App/TopBar.tsx
skoeva: you're entirely right, updating this now

PR #4137, opened 2025-11-07
Fix missing PKCE arg in Helm

### LINKED ISSUE #4136: Two issues: (1) Helm chart doesn’t pass -oidc-use-pkce in some paths, (2) empty in-cluster OIDC CA cert triggers failed to append ca cert to pool
Overview

I ran into two related problems when enabling OIDC with PKCE in Headlamp:

Helm chart arg omission for PKCE
In some render paths, the Helm chart did not add -oidc-use-pkce=$(OIDC_USE_PKCE) to the container args even though the environment variable was set and PKCE was enabled. After adding the missing snippet in the else branch (where ExternalSecret is enabled), the flag finally appeared in args and PKCE worked.

Empty in-cluster OIDC CA cert causes TLS error
When running in-cluster with PKCE enabled (so the backend hits the token endpoint), the backend logs:

{"level":"error","source":"/headlamp/backend/pkg/auth/auth.go","line":229,
 "error":"failed to append ca cert to pool","time":"2025-11-07T16:34:17Z",
 "message":"couldn't add custom cert to context"}

Root cause: in kubeconfig.GetInClusterContext(), oidcCACert is always assigned to OidcConfig.CACert even when it is an empty string. Later, ConfigureTLSContext() checks caCert != nil (pointer non-nil), tries to parse PEM, and fails on the empty string. This incorrectly forces the “custom CA” code path and breaks TLS for public CAs (e.g., Let’s Encrypt), where no custom CA is needed.

Environment

Headlamp v0.37.0 (container ghcr.io/headlamp-k8s/headlamp:v0.37.0)

Kubernetes: in-cluster deployment

OIDC provider: Keycloak (TLS served by cert-manager / Let’s Encrypt)

PKCE enabled

How to reproduce

Deploy Headlamp via Helm with PKCE enabled.

Before the Helm change, -oidc-use-pkce might not appear under args depending on the ExternalSecret branch.

Run in-cluster with an empty oidcCACert (e.g., value present but empty, or not intended to be used).

Attempt login: backend performs token exchange over TLS → error log:

failed to append ca cert to pool / couldn't add custom cert to context

Actual:

PKCE flag sometimes not present in container args.

TLS error when oidcCACert is empty string (pointer non-nil).

Expected:

PKCE flag should always be passed when enabled.

If oidcCACert is empty, backend should not attempt to parse/append it; it should rely on the system trust store (works fine for Let’s Encrypt).

Proposed fixes

Helm chart: ensure -oidc-use-pkce=$(OIDC_USE_PKCE) is added in both branches (ExternalSecret on/off).
Example (added in the else branch):

{{- if or (eq ($oidc.usePKCE | toString) "true") (eq $usePKCE "true") }}
- "-oidc-use-pkce=$(OIDC_USE_PKCE)"
{{- end }}

Backend (in-cluster OIDC): set OidcConfig.CACert to nil when the provided string is empty/whitespace.
Suggested diff:

diff --git a/backend/pkg/kubeconfig/kubeconfig.go b/backend/pkg/kubeconfig/kubeconfig.go
@@ func GetInClusterContext(..., oidcSkipTLSVerify bool, oidcCACert string) (*Context, error) {
-        oidcConf = &OidcConfig{
-            ...
-            SkipTLSVerify: &oidcSkipTLSVerify,
-            CACert:        &oidcCACert,
-        }
+        var caPtr *string
+        if strings.TrimSpace(oidcCACert) != "" {
+            caPtr = &oidcCACert
+        }
+        oidcConf = &OidcConfig{
+            ...
+            SkipTLSVerify: &oidcSkipTLSVerify,
+            CACert:        caPtr,
+        }

This prevents ConfigureTLSContext() from entering the custom-CA branch when there’s no CA to add.

Workarounds

For the TLS error: temporarily set -oidc-skip-tls-verify=true (test only, not for prod) or make sure no empty CACert is passed.

For Helm: patch the chart to always render the PKCE arg in both branches.

Logs / Snippets

Error:

failed to append ca cert to pool / couldn't add custom cert to context

Helm before/after: adding the PKCE arg to the else branch fixed the missing arg.

Request

If this approach looks good, I can open a PR with:

Helm template fix for PKCE arg.

Backend guard for empty oidcCACert.

Thanks!

rayshoo:
This PR addresses issues reported in #4136 :

1. Helm chart PKCE arg omission
Ensure `-oidc-use-pkce=$(OIDC_USE_PKCE)` is rendered in both branches (when ExternalSecret is enabled or not). Previously, some paths did not include the arg although PKCE was enabled.

```
# values.yaml
config:
  oidc:
    secret:
      create: false
    externalSecret:
      enabled: true
      name: headlamp

env:
- name: OIDC_CALLBACK_URL
  value: https://headlamp-dev.example.com/oidc-callback
- name: OIDC_USE_PKCE
  value: 'true'
```

# Changes

- Helm: add `-oidc-use-pkce=$(OIDC_USE_PKCE)` in the `else` branch as well.

# Rationale
- PKCE should be consistently enabled when configured.

# Testing
- Verified Helm renders -oidc-use-pkce in both ExternalSecret paths.

# Related
- Fixes #4136 .

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -290,6 +290,9 @@ spec:
             # Check if callbackURL is non empty either from env or oidc.config
             - "-oidc-callback-url=$(OIDC_CALLBACK_URL)"
             {{- end }}
+            {{- if or (eq ($oidc.usePKCE | toString) "true") (eq $usePKCE "true") }}
+            - "-oidc-use-pkce=$(OIDC_USE_PKCE)"
+            {{- end }}
             {{- if or (ne $oidc.validatorClientID "") (ne $validatorClientID "") }}
             # Check if validatorClientID is non empty either from env or oidc.config
             - "-oidc-validator-client-id=$(OIDC_VALIDATOR_CLIENT_ID)"


illume (approved):
🎉 thanks!

PR #4144, opened 2025-11-10
app: Fix url decoding for external links handling

joaquimrocha:
## Summary

When clicking Reload Now after installing a plugin on Windows, the system's browser is opened on the index.html file.

## Steps to Test

1. Open Headlamp desktop on Windows
2. Go to the plugin catalog and install a plugin, then click Reload Now in the notification -> An index.html is mistakenly opened in the system's browser


illume (approved):
🎉 thanks!

PR #4158, opened 2025-11-13
.github: build-container.yml: Add arm64 container tests

illume:
This is so that arm64 builds are also tested by e2e tests.


illume:
Looks like the ubuntu arm64 runners are backed up and take a long time to start. (Over 26 minutes and still queued).

illume:
I don't think this is usable at the moment. The jobs are generally queued for hours.

Maybe when github gets more arm64 runners we can look at this again.

PR #4166, opened 2025-11-14
app: frontend: plugins: Bump js-yaml to 4.1.1

skoeva:
These changes update the js-yaml dependency to 4.1.1 as desired.


illume (approved):
👍 thanks

PR #4180, opened 2025-11-19
Update image digests in Dockerfiles

skoeva:
This change updates the image digests for golang & node in the Dockerfiles, fixing the failure in the Build Container test in CI.


illume:
I guess these images are updated quite often, and then the digest is also updated quite often.

Probably I need to change the test,
so that it only runs when one of these digests are changed. Rather than on every container job run.

illume:
@skoeva I added a commit to change this test so it only runs when someone modifies the digest. I wanted to check it out with your change of the digests here in this PR. Can you please have a look? 
- https://github.com/kubernetes-sigs/headlamp/pull/4181

illume:
@skoeva I temporarily added a commit to change this test so it only runs when someone modifies the digest. I wanted to check it out with your change of the digests here in this PR.

Here's the other PR that only verifies the digests when the PR changes:
- https://github.com/kubernetes-sigs/headlamp/pull/4181

PR #4200, opened 2025-11-24
app: Fix type for isDev constant

sniok:
This PR coerces isDev into proper boolean

Change from aks-dekstop


skoeva (approved):

illume:
nit: Probably we should have a test if an external project requires this. A boolean type on there would make it more explicit.

Reminder to please no pushes to branches on kubernetes-sigs/headlamp, because they become locked and it's a security issue.

PR #4204, opened 2025-11-24
frontend: Remove multiplexer from useKubeObject function

sniok:
I've noticed that the details view still uses websocket multiplexer, which we disabled a while ago due to issues.

This PR reverts back to normal direct websocket connection,
Also moves multiplexer to a separate file so it's clearer (it was mixed with default websockets and probably that's why we forgot to disable it for the useKubeObject hook)

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.ts
+++ b/frontend/src/lib/k8s/api/v2/hooks.ts
@@ -25,6 +25,8 @@ import type { KubeListUpdateEvent } from './KubeList';
 import { KubeObjectEndpoint } from './KubeObjectEndpoint';
 import { makeUrl } from './makeUrl';
 import { useWebSocket } from './multiplexer';
+import { getWebsocketMultiplexerEnabled } from './useKubeObjectList';
+import { useWebSockets } from './webSocket';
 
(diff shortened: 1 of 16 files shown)


illume:
Can you make it not hard to enable again?

frontend/src/lib/k8s/api/v2/hooks.ts
```diff
@@ -149,21 +151,51 @@ export function useKubeObject<K extends KubeObject>({
...
+      enabled: !!endpoint && !!data,
+      cluster,
+      onMessage(update: KubeListUpdateEvent<K>) {
+        if (update.type !== 'ADDED' && update.object) {
+          client.setQueryData(queryKey, new kubeObjectClass(update.object));
+        }
+      },
+    });
+  } else {
+    useWebSockets({
```
joaquimrocha, line 194: Doesn't this violate the no-conditional hooks rule?

sniok: kind of but not really, I explained it in the comment. Think of it as a compile time 'if', it will never change during the runtime

PR #4209, opened 2025-11-25
frontend: Add support for project header actions

sniok:
## Summary
This PR adds ability for plugins to add custom header actions in project details

Change from aks-desktop


illume:
For changes that external projects rely on, can we please have tests?

Plugin functionality should be included in an example.

illume (approved):
🎉 thanks!

PR #4211, opened 2025-11-26
backend: Bump golang.com/x/crypto to 0.45.0

skoeva:
This change updates the golang.com/x/crypto dependency in the backend from 0.40.0 to 0.45.0.


illume (approved):
🎉 thanks

PR #4219, opened 2025-11-28
I18n improvements

joaquimrocha:
## Summary

This PR adds a couple of tools to improve the experience for translators.
It adds a couple of top level commands:
npm run i18n:status -> prints progress bars for each locale
npm run i18n:list -> lists locales and the percentage of completion

Also moves the former copy/missing translation commands to this top level variant (check docs).

Also adds docs.

### How to test

Check the docs and run the commands to test them out.


frontend/src/i18n/tools/extract-empty-translations.js
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+// ⚠️ DEPRECATED: This script is deprecated. Use the new i18n tool instead:
```
illume: I wonder why we can't remove it?

joaquimrocha:
Closed by mistake. Reopened a new one.

PR #4228, opened 2025-12-01
backend: embed: open Headlamp in default browser

yolossn:
## Summary

Open Headlamp (localhost:4466) in the default
browser when the binary is built with the 'embed'
tag. This default behaviour can be disabled by
using the -no-browser flag.

## Related Issue

Fixes #ISSUE_NUMBER  

## Changes

- Added/Updated [component/file/logic]
- Fixed [bug/issue/typo]
- Refactored [code/module] for clarity/performance

## Steps to Test

1. Build the backend embedded binary by running 
> make backend-embed
2. Run the binary 
> ./backend/headlamp-server
The binary should start the backend server and open locahost:4466 in the default browser
3. Run the binary with the `--no-browser` flag
> ./backend/headlamp-server --no-browser
The binary should just start the backend server

### DIFF
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -24,7 +24,9 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
+	"github.com/cli/browser"
 	"github.com/gorilla/mux"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/cache"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/config"
@@ -33,6 +35,7 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/plugins"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/spa"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/telemetry"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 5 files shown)


backend/cmd/server.go
```diff
@@ -57,6 +60,17 @@ func main() {
 		return
 	}
 
+	// Open browser if running locally with embedded frontend and not in cluster and no-browser is not set
+	if spa.UseEmbeddedFiles && !conf.NoBrowser && !conf.InCluster {
+		go func() {
+			time.Sleep(500 * time.Millisecond) // give server time to start
+
+			if err := browser.OpenURL("http://localhost:4466"); err != nil {
```
illume: A recent change is it can be that it listens on another port than 4466.

backend/cmd/server.go
```diff
@@ -33,7 +34,9 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/plugins"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/spa"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/telemetry"
+	"github.com/pkg/browser"
```
illume: How about https://github.com/cli/browser instead? The pkg/browser one has a few open vulns. This alternative is a bit better maintained and used by GitHub cli... so they will probably take care of updates going forward.

backend/cmd/server.go
yolossn: AFAIU the default port isn't changed, the change that @joaquimrocha introduced in this PR https://github.com/kubernetes-sigs/headlamp/pull/4121 involves changing the port arg and not the default port itself. I have fixed the part where the default port is hardcoded to use the value from port arg.

PR #4230, opened 2025-12-01
backend: auth: Extract RefreshAndSetToken from headlamp.go

skoeva:
This change extracts the RefreshAndSetToken function from headlamp.go into the auth package.

Part of:
- #3482

### Updates
- Add params struct to help move future functions
- Add comprehensive tests in `auth_test.go`

### Testing
```shell
cd backend
go test -v ./pkg/auth -run RefreshAndSetToken
```

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -933,49 +933,6 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	return r
 }
 
-func (c *HeadlampConfig) refreshAndSetToken(oidcAuthConfig *kubeconfig.OidcConfig,
-	cache cache.Cache[interface{}], token string,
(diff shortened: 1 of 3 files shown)


joaquimrocha (changes requested):
Is there a chance (honest question) we could have the tests being done for the same logic,  before the extraction commit. This way we could verify that everything keeps working whereas right now we are testing for the new changes already.

skoeva:
There aren't any changes in the function logic so coupling the tests with the extracted code should be fine

illume (changes requested):
Unfortunately this has a merge conflict now.

I read through this the other week, and looks good to me. Thanks for the improvement.

illume (changes requested):
Thanks for these changes.


illume (approved):
🎉 thanks!

Issue #3425, opened 2025-06-03
plugins: node_modules in the newly installed plugin is very big


sniok:
## Describe the bug

When creating new plugin, node_modules in that plugin is extremely big. This results in a quite long install time and big impact on the disk usage.

```
$ du -sh node_modules
852M    node_modules
```

By default it installs a lot of modules that plugin might never need. And considering that even if the module like mui or monaco-editor is used, the code from those modules is not actually used for building and running the plugin.

## To Reproduce
Steps to reproduce the bug:
1. npx @kinvolk/headlamp-plugin@latest create plugin-test 
2. cd plugin-test && du -sh node_modules

## Are you able to fix this issue?
We need to think and discuss about how to fix this first

BABAR-TAHSEEN55:
I just researched this issue and found this insights . Let me know your thoughts on this ? @sniok 
1) Total size is very large 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/43061d6a-af84-480a-9cae-4c8082129404" />

When I sort them with sizes , we get this : 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/02724b64-01cd-406f-8eab-6ef3865328a5" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e3edeb0a-21ce-4fba-83cc-28790e4205b1" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f3d20bfe-cfef-4db1-919c-d963bd461f0b" />


We'll have to find and cut down the redundant modules 
Do you have anything in mind ?
We could create a clean up script after that

PR #4235, opened 2025-12-01
headlamp-plugin: Remove unused dependencies

skoeva:
This continues the work of removing unused dependencies in Headlamp.

Related issue: https://github.com/kubernetes-sigs/headlamp/issues/3425

### Testing
- [x] Run `npm i && npm run build` in `headlamp-plugin` and ensure no failures

### DIFF
--- a/plugins/headlamp-plugin/dependencies-sync.js
+++ b/plugins/headlamp-plugin/dependencies-sync.js
@@ -71,6 +71,7 @@ const dependenciesToNotCopy = [
   'openapi-types',
   'resize-observer-polyfill',
   'vitest-canvas-mock',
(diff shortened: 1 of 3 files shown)


load-tests/package.json
```diff
@@ -13,7 +13,6 @@
   "author": "Microsoft",
   "license": "Apache-2.0",
   "dependencies": {
-    "@types/node": "^20.14.2",
```
illume, line 16: This was for type checking the code, is this not necessary anymore?

For type checking inside vscode for example.

skoeva: there isn't any TS code here so it isn't really necessary, but I guess the typings could be helpful. I'll add it back

PR #4242, opened 2025-12-02
frontend: Enable verbatimModuleSyntax and autofix imports

sniok:
Enabled `verbatimModuleSyntax` ts compiler option that is generally recommended. 
Nice side effect of it is that it now requires you to use `import type ...` everywhere and will throw error if it's not used.


sniok:
Somehow removing circular dependencies by switching to type imports resulted in a more broken build and untangling it doesn't look straightforward, so I'm closing this for now. Maybe we can return later when we have more time for this.

PR #4248, opened 2025-12-05
backend: Dockerfile: Update golang to 1.24.11

skoeva:
This change bumps golang in the backend and Dockerfile to address a security vulnerability in golang v1.24.10.


illume (approved):
🎉 thanks!

PR #4256, opened 2025-12-09
frontend: Add cluster column to custom resources list

kahirokunn:
## Summary

This PR adds a **Cluster** column to the Custom Resources objects list so that it behaves consistently with the Workloads lists.  
When multiple clusters are configured, users can now quickly see which cluster each custom resource instance belongs to.

## Related Issue

Fixes #ISSUE_NUMBER *(replace with the actual issue number, if any)*

## Changes

- Updated `CustomResourceListTable` in `CustomResourceList` to include the `cluster` column.

## Steps to Test

1. Start Headlamp with at least two cluster configured (ideally multiple clusters for full verification).
2. Navigate to **Custom Resources** in the sidebar.
3. Select a CRD that has instances and open its objects list.

CleanShot 2025-12-09 at 16 51 05@2x

### DIFF
--- a/frontend/src/components/crd/CustomResourceList.tsx
+++ b/frontend/src/components/crd/CustomResourceList.tsx
@@ -18,11 +18,16 @@ import { JSONPath } from 'jsonpath-plus';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
+import { useSelectedClusters } from '../../lib/k8s';
 import CRD, { KubeCRD } from '../../lib/k8s/crd';
 import { KubeObject } from '../../lib/k8s/KubeObject';
 import { localeDate } from '../../lib/util';
 import { EmptyContent as Empty, Link, Loader, ResourceListView } from '../common';
-import { ResourceTableColumn, ResourceTableProps } from '../common/Resource/ResourceTable';
+import {
+  ColumnType,
+  ResourceTableColumn,
+  ResourceTableProps,
+} from '../common/Resource/ResourceTable';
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/crd/CustomResourceList.tsx
```diff
@@ -142,20 +142,20 @@ export function CustomResourceListTable(props: CustomResourceTableProps) {
...
-      ...additionalPrinterCols,
-      'age',
-    ] as ResourceTableProps<KubeObject<KubeCRD>>['columns'];
+    ];
 
     if (crd.isNamespacedScope) {
       colsToDisplay.splice(1, 0, 'namespace');
     }
 
+    colsToDisplay.push('cluster', ...additionalPrinterCols, 'age');
```
joaquimrocha: Wouldn't it be more readable if we just added the `cluster` column to the original assignment above before `...additionalPrinterCols`?
Also, the cluster column is supposed to be shown only when in a multi-cluster view and I think your patch is making it show it even when there's only one cluster.

joaquimrocha (changes requested):
Thanks @kahirokunn ! This is a nice catch. The code looks fine but I had a couple of things I'd like to understand. Please see my comments.

frontend/src/components/crd/CustomResourceList.tsx
kahirokunn: Thanks for the feedback! Fixed both points - moved cluster to the initial array for better readability, and made it conditional on multi-cluster view using useSelectedClusters().
Tested with both single and multi-cluster selections, working as expected.

PR #4271, opened 2025-12-14
frontend: Guard header actions when resource is null

kahirokunn:
## Summary

This PR fixes an intermittent console error: `TypeError: Cannot read properties of null (reading 'metadata')`.  
The root cause was that `MainInfoSectionHeader` could render header actions (including `EditButton`) while the `resource` was still `null`, passing a null `item` down the tree.  
We now **guard header actions so they only render once `resource` is available**, preventing null from being passed to actions.

## Related Issue

- Fixes #ISSUE_NUMBER

## Changes

- Guarded header action rendering in `frontend/src/components/common/Resource/MainInfoSection/MainInfoSectionHeader.tsx` to avoid passing a null `resource` to action components (e.g. `EditButton`).

## Steps to Test

1. Open a resource details page (ideally under slow network / first load conditions).
2. Check the browser console.
3. Confirm `TypeError: Cannot read properties of null (reading 'metadata')` no longer appears.
4. After the resource finishes loading, confirm header actions (Edit, Delete, etc.) appear and work as expected.

Before:
CleanShot 2025-12-14 at 11 58 59@2x

After:
No Error 🙌

### DIFF
--- a/frontend/src/components/common/Resource/MainInfoSection/MainInfoSectionHeader.tsx
+++ b/frontend/src/components/common/Resource/MainInfoSection/MainInfoSectionHeader.tsx
@@ -83,7 +83,7 @@ export function MainInfoHeader<T extends KubeObject>(props: MainInfoHeaderProps<
       return <ErrorBoundary>{Action}</ErrorBoundary>;
     } else if (Action === null) {
       return null;
-    } else if (typeof Action === 'function') {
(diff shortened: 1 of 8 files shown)


joaquimrocha:
Nice catch! Thanks @kahirokunn !
/approve
/lgtm

PR #4276, opened 2025-12-15
app: backend: Fix for plugin watch doing a loop on mac app

illume:
- **backend: plugins: Fix for skipping CHMOD events in plug watcher**
- **app: plugin-management: Fix to only chmod a file if it is needed**

### Testing done

- use a plugin that has a bin/ file with an executable permission (minikube prerelease plugin)
- open app on mac, note that it doesn't reload over and over again


illume:
Thanks @sniok

PR #4279, opened 2025-12-16
frontend,backend: support exec/teleport/cert based auth to clusters

rushrs:
### Problem
Headlamp didn't properly handle Teleport's exec credential plugin authentication. When users loaded clusters configured with tsh kube credentials, they would see:
- Token prompt dialogs (which don't apply to Teleport)
- Redirect loops
- Failed auth tests (because SelfSubjectRulesReview API is blocked by Teleport's RBAC proxy)

### Solution

Added first-class support for Teleport's tsh authentication:

- Backend: Detect tsh auth type (backend/pkg/kubeconfig/kubeconfig.go) 
  - Added detection for tsh in exec command
  - Returns auth_type: "tsh" instead of generic "exec" for Teleport clusters
- Frontend: Use /version endpoint for auth test (frontend/src/lib/k8s/api/v1/clusterApi.ts)
  - For tsh/exec/client-cert auth, use /version endpoint instead of SelfSubjectRulesReview
  - The /version endpoint is accessible to any authenticated user without RBAC restrictions
  - Keeps original SelfSubjectRulesReview flow for token/OIDC auth
- Frontend: Teleport-specific error messages (frontend/src/components/App/RouteSwitcher.tsx)
  - When tsh auth fails (expired session), shows: "Teleport authentication failed. Please run 'tsh login' to refresh your credentials"
  - No redirect to login page (since Teleport auth is handled externally)
## Steps to Test

I built the frontend backend and app locally and tried it with these changes and the tsh based auth worked.

## Notes for the Reviewer

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -143,6 +143,7 @@ interface AuthRouteProps {
   requiresAuth: boolean;
   requiresCluster: boolean;
   requiresToken: () => boolean;
+  clusters: ReturnType<typeof useClustersConf>;
   [otherProps: string]: any;
 }
 
@@ -153,15 +154,24 @@ function AuthRoute(props: AuthRouteProps) {
     requiresAuth = true,
     requiresCluster = true,
     computedMatch = {},
+    clusters,
     ...other
   } = props;
   const redirectRoute = getCluster() ? 'login' : 'chooser';
   useSidebarItem(sidebar, computedMatch);
   const cluster = useCluster();
+
+  // Get the cluster's auth type - for tsh/exec/client-cert, the backend handles auth
+  const clusterConfig = cluster && clusters ? clusters[cluster] : null;
+  const authType = clusterConfig?.auth_type || '';
+  const isExecOrCertAuth = authType === 'tsh' || authType === 'exec' || authType === 'client-cert';
+
   const query = useQuery({
-    queryKey: ['auth', cluster],
-    queryFn: () => testAuth(cluster!),
-    enabled: !!cluster && requiresAuth,
+    queryKey: ['auth', cluster, authType],
+    queryFn: () => testAuth(cluster!, { authType }),
+    // Run auth test for all auth types including exec/client-cert
+    // Wait for clusters to be loaded to ensure we have the correct authType
+    enabled: !!cluster && requiresAuth && !!clusters,
     retry: 0,
   });
 
@@ -182,6 +192,30 @@ function AuthRoute(props: AuthRouteProps) {
     }
 
     if (query.isError) {
+      // For tsh/exec/client-cert auth, show a helpful error message
+      if (isExecOrCertAuth) {
+        let errorMessage: string;
+        if (authType === 'tsh') {
+          errorMessage =
+            'Teleport authentication failed. Please run `tsh login` to refresh your credentials, then reload this page in your browser.';
+        } else if (authType === 'exec') {
+          errorMessage =
+            'Exec credential plugin authentication failed. Please refresh your credentials and try again.';
+        } else {
+          errorMessage =
+            'Client certificate authentication failed. Please check your certificates.';
+        }
+        return (
+          <ErrorComponent
+            error={{
+              name: 'AuthenticationError',
+              message: errorMessage,
+              stack: query.error instanceof Error ? query.error.message : String(query.error),
+            }}
+          />
+        );
+      }
+
       return (
         <Redirect
           to={{
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


vyncent-t (commented):
Submitting this for a copilot suggestions, feel free to adjust as you see fit.

rushrs:
> Submitting this for a copilot suggestions, feel free to adjust as you see fit.

Just had a scan through and added the tsh check. happy with it now

rushrs:
addressed the failing github actions steps and rebased, thanks all

illume:
Thanks @rushrs 

I will review and test over the next days. I ran the CI jobs again (they have to be manually run the first time someone contributes a PR).

Would it be possible to squash some of the commits together and to get rid of the merge commits? We follow a Linux kernel style commit message formatting… please see https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines

Are there instructions somewhere for teleport with the exec credential so we can test this manually?

illume:
It looks like teleport isn't open source exactly, so I can not manually test it myself.

rushrs:
> It looks like teleport isn't open source exactly, so I can not manually test it myself.

exec based auth
- tsh
- aws-iam-auth
- gke gcloud auth

cert based auth
- minikube
- kind
- k3s

sniok:
Hi, we already support client cert and exec auth. Is there a reason for changes you're introducing for those? 
Also I'm not sure if we should add support for commercial product like teleport

rushrs:
there is a self hosted opensource version which has an apache 2.0 license

rushrs:
> Hi, we already support client cert and exec auth. Is there a reason for changes you're introducing for those?
Also I'm not sure if we should add support for commercial product like teleport

Hey! the exec based support doesnt work in the frontend, it results to `auth_type = ""` this is likely to be teleport specific - when connecting via the `gke-auth-plugin `for the clusters, I do not run into the same issues. 

When connecting via teleport headlamp asks for a client id which isn't present as `tsh` creates short lived certs for access. In the kube config it runs an exec for  `tsh kube credentials` for the given cluster which is proxied behind teleport

`selfsubjectrulesreviews` causes problems for the connection check some exec based solutions like teleport have minimal rbac - users can access resources but can't call authorization APIs.

illume:
> there is a self hosted opensource version which has an apache 2.0 license

Ah right. Thanks for the clarification.

illume:
@rushrs do you have a link that explains the licensing? There is this file that seems to be a non open source license? https://github.com/gravitational/teleport/blob/master/build.assets/LICENSE-community

I especially want to know how to use the "self hosted opensource version which has an apache 2.0 license" version in order to be able to manually test this.

rushrs:
> @rushrs do you have a link that explains the licensing? There is this file that seems to be a non open source license? https://github.com/gravitational/teleport/blob/master/build.assets/LICENSE-community
> 
> I especially want to know how to use the "self hosted opensource version which has an apache 2.0 license" version in order to be able to manually test this.

ah I hadn't realised they moved away from that licence a while back. i've just used the enterprise version mainly. looks like the new one builds on apache 2.0 but restricts building products from their source code. we are not doing that in this case just enabling support for it

illume:
> ah I hadn't realised they moved away from that licence a while back. i've just used the enterprise version mainly. looks like the new one builds on apache 2.0 but restricts building products from their source code. we are not doing that in this case just enabling support for it


Cool. It's just none of the headlamp maintainers has Teleport experience, and I guess we can't manually test it ourselves. Do you think you would be around if any Teleport related issues pop up? We don't have regression tests for this code either so far... so we'd have to rely on you and other Teleport users to report/fix issues. 

Are you able to write testing steps in the PR description? This way another teleport user in the future would find the testing steps useful for testing this if you're not around.

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -182,6 +192,30 @@ function AuthRoute(props: AuthRouteProps) {
...
+        let errorMessage: string;
+        if (authType === 'tsh') {
+          errorMessage =
+            'Teleport authentication failed. Please run `tsh login` to refresh your credentials, then reload this page in your browser.';
+        } else if (authType === 'exec') {
+          errorMessage =
+            'Exec credential plugin authentication failed. Please refresh your credentials and try again.';
+        } else {
+          errorMessage =
+            'Client certificate authentication failed. Please check your certificates.';
```
illume, line 206: Strings need to be translated with t() from useTranslate. See other strings in different components.

Afterwards “npm run i18n” needs to be run.

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -182,6 +192,30 @@ function AuthRoute(props: AuthRouteProps) {
...
+      // For tsh/exec/client-cert auth, show a helpful error message
+      if (isExecOrCertAuth) {
+        let errorMessage: string;
+        if (authType === 'tsh') {
+          errorMessage =
+            'Teleport authentication failed. Please run `tsh login` to refresh your credentials, then reload this page in your browser.';
+        } else if (authType === 'exec') {
+          errorMessage =
+            'Exec credential plugin authentication failed. Please refresh your credentials and try again.';
+        } else {
```
illume, line 204: Maybe match on "client-cert" to be more explicit?

frontend/src/components/App/RouteSwitcher.tsx
```diff
@@ -182,6 +192,30 @@ function AuthRoute(props: AuthRouteProps) {
     }
 
     if (query.isError) {
+      // For tsh/exec/client-cert auth, show a helpful error message
+      if (isExecOrCertAuth) {
```
illume, line 196: I think this could code be in it's own component. Called something like AuthenticationError?

rushrs:
I will action these changes and don't mind being a maintainer for this part of the integration.

ptr1120:
Hi 👋
Just wanted to say that I’m very interested in this PR and would love to see it merged when possible. Happy to help test if useful. Thanks!

illume (changes requested):
Can you please have a look at the open comments?

If it's merged by Tuesday we can put it into 0.40.0 release.

illume (changes requested):
Thanks for these changes.


<details>

Good examples:

</details>

PR #4284, opened 2025-12-17
frontend: Label: Fix warning label color contrast

vyncent-t:
## Summary

image

Ran FastPass and it seems there is an issue when using white text and orange background. I have tried adjusting the background color to be darker to match allowed contrast, but this results in the "allowed" orange basically being just red. (Please refer to the image above for details)

This clashes with the red color used for failures so I am opting for a use of black text when displaying orange labels in light mode.  

## Changes

- Adjusts label text color to black when on light mode

## Steps to Test

1. Navigate to a resource that is displaying warning label
2. Note that the white text is now black
3. Run FastPass if available and observe no issues

### Before

image

### After

image

### DIFF
--- a/frontend/src/components/common/Label.stories/__snapshots__/StatusLabel.Warning.stories.storyshot
+++ b/frontend/src/components/common/Label.stories/__snapshots__/StatusLabel.Warning.stories.storyshot
@@ -1,7 +1,7 @@
 <body>
   <div>
     <span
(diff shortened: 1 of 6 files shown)


skoeva (commented):
would be good to test this with a color contrast checker like [this one](https://accessibleweb.com/color-contrast-checker/)

skoeva (commented):
CI failure should also go away with a rebase

illume (approved):
🎉 thanks!

PR #4300, opened 2025-12-20
helm: Add extraContainers support for sidecar containers

### LINKED ISSUE #4239: Helm chart allows extra containers
Headlamp right now only supports otlp protocol for tracing. Since i don't use that, the easiest solution that i could think of is to use sidecard container to convert a protocol. Headlamp right now does not allow to create extra containers via values.yaml

values.yaml will have `extraContainers` field which will be loaded to the deployment template

Users can customize their deployments with additional containers for whatever are their needs.

Yes

hxrshxz:
## Summary

This PR adds extraContainers support to the Helm chart for sidecar containers.

## Related Issue

Fixes #4239

## Changes

- Added `extraContainers` field to `values.yaml`
- Updated `deployment.yaml` template to include extra containers

## Steps to Test

1. Add an extra container in values:
```yaml
extraContainers:
- name: my-sidecar
  image: busybox
```
2. Run helm template test ./charts/headlamp --set 'extraContainers[0].name=sidecar'
3. Verify sidecar container appears in deployment

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -382,6 +382,9 @@ spec:
               {{- toYaml $defaultSC | nindent 12 }}
             {{- end }}
         {{- end }}
+        {{- with .Values.extraContainers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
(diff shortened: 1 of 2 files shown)


illume:
Oh. There was already a PR for this here:
- #4241

illume:
I'll merge the other one. Seems we still need this to be documented in the README.md

PR #4313, opened 2025-12-29
charts: add ability to specify topologySpreadConstraints

irasnyd:
## Summary

This PR adds the ability to specify a `topologySpreadConstraints` for the headlamp deployment.

An example was included which prefers to schedule multiple Pods in different Availability Zones, and requires that multiple Pods are on different Nodes. When uncommented, the example has no effect on the default configuration with `replicaCount: 1`.

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -394,6 +394,17 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- range $constraint := . }}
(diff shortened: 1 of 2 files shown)


charts/headlamp/templates/deployment.yaml
```diff
@@ -394,6 +394,17 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.topologySpreadConstraints }}
```
illume, line 397: This just needs documenting in the charts/README.md

illume: @copilot please document this in charts/README.md

ashu8912:
Hii @irasnyd thanks for the PR, i have continued the docs addition here in #4559

Issue #2085, opened 2024-06-26
fuzz testing / property testing


illume:
Currently we're not using fuzzing or property testing which is checked by the [headlamp OpenSSF scorecard](https://scorecard.dev/viewer/?uri=github.com/headlamp-k8s/headlamp).

![image](https://github.com/headlamp-k8s/headlamp/assets/9541/3f5b0651-4630-4afb-8898-2b34fcfa8281)

> Fuzzing, or fuzz testing, is the practice of feeding unexpected or random data into a program to expose bugs. Regular fuzzing is important to detect vulnerabilities that may be exploited by others, especially since attackers can also use fuzzing to find the same flaws.
-- https://github.com/ossf/scorecard/blob/7ce8609469289d5f3b1bf5ee3122f42b4e3054fb/docs/checks.md#fuzzing

### Additional information

It's especially useful on any code that parses inputs like network data or user provided content or arguments. During testing we might uncover areas which don't validate or parse input data at all.

#### Some tools

- https://go.dev/doc/security/fuzz/
- https://fast-check.dev/
- https://github.com/google/oss-fuzz

illume:
scorecard considers fastcheck for tsx files now https://github.com/ossf/scorecard/commit/59ed1b6b04d3ba5a721d494d4a277cdca33659e1#diff-9aa0457d2d787488ba6d1d65d0f7c9d17ba3cd68bd43d4652380a427267897e3R124

illume:
@tejhan please /assign yourself for issue refinement

PR #4320, opened 2026-01-01
frontend: units.test: Add fast-check property tests for parsers

illume:
- **frontend: headlamp-plugin: Add fast-check and audit fix**
- **frontend: units.test: Add some property tests for parsers**

Adds fast-check for property testing in typescript. See https://fast-check.dev/ 
This is a requirement for openssf scorecard.

Related:
- #2085

These tests are run when you do "npm run test" in the frontend/.


illume:
@skoeva rebased against main

skoeva (approved):
LGTM

PR #4337, opened 2026-01-06
AGENTS.md: Add basic AGENTS.md

### LINKED ISSUE #4324: Add AGENTS.md

illume:
This does speed up AGENTS, and it fixes a few problems different ones have. Including the style of changes it makes and detecting commit message issues, and agents not formatting code or running tests properly.

Note, this is used by both review and code generating agents. Made with some manual tweaks from the combination of a few different AGENTS.md creators and researcher agents. It is one file rather than several, because some agents do not support nested AGENTS.md well. Even copilot in vscode at time of writing.

Fixes #4324
- https://github.com/kubernetes-sigs/headlamp/issues/4324

### Testing

Was tested with copilot agent via vscode, and github. Also Claude and a custom agent. Also tried it with the copilot review bot.

I think we can use this as a beginning and add more documentation or things in here as we spot other issues.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,367 @@
+# AGENTS.md
+version: 1
+default_agent: "@dev-agent"
+
+> **Consulted files (agent must populate before committing):**  
+> 
+> **README.md files:**
+> - `/README.md` - Main project README
+> - `/frontend/README.md` - Frontend-specific README
+> - `/backend/README.md` - Backend quickstart guide
+> - `/app/README.md` - Desktop app README
+> - `/e2e-tests/README.md` - End-to-end testing README
+> - `/load-tests/README.md` - Load testing guide
+> - `/plugins/README.md` - Plugins overview
+> - `/plugins/headlamp-plugin/README.md` - Plugin development tools
+> - `/plugins/pluginctl/README.md` - Plugin control utility
+> - `/backend/pkg/telemetry/README.md` - Telemetry module
+> - `/docker-extension/README.md` - Docker extension
+> - `/charts/headlamp/README.md` - Helm chart
+> - `/backstage-test/README.md` - Backstage integration test
+> - `/tools/i18n/README.md` - Internationalization tools
+> - `/eslint-config/README.md` - ESLint configuration
+> - `/frontend/src/i18n/README.md` - Frontend i18n
+> - `/app/e2e-tests/README.md` - Desktop app e2e tests
+> - Plugin examples: `/plugins/examples/*/README.md` (multiple examples)
+>
+> **Documentation files:**
+> - `/docs/development/index.md` - Main development guide with build/run commands
+> - `/docs/development/frontend.md` - Frontend development guide
+> - `/docs/development/backend.md` - Backend development guide with testing commands
+> - `/docs/development/testing.md` - Load testing guide
+> - `/docs/development/architecture.md` - Architecture documentation
+> - `/docs/development/plugins/index.md` - Plugin system overview
+> - `/docs/development/plugins/building.md` - Plugin building guide
+> - `/docs/development/plugins/getting-started.md` - Plugin getting started
+> - `/docs/development/plugins/publishing.md` - Plugin publishing guide
+> - `/docs/development/i18n/index.md` - Internationalization guide
+> - `/docs/development/release-guide.md` - Release process
+> - `/docs/installation/index.mdx` - Installation instructions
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,356 @@
...
+- **`frontend/`** — UI code (TypeScript/React); see `/frontend/README.md` and `/docs/development/frontend.md` for build/run/test commands.
+- **`backend/`** — Go server and API; see `/backend/README.md` and `/docs/development/backend.md` for server commands.
+- **`docs/`** — all developer and user docs; reference specific files under `docs/` for workflows.
+- **`plugins/`** — plugin system and examples; see `/plugins/README.md` and `/docs/development/plugins/`.
+- **`app/`** — desktop application (Electron); see `/app/README.md`.
+- **CI & infra:** `.github/workflows/`, Dockerfiles (`/Dockerfile`, `/Dockerfile.plugins`), Kubernetes manifests (`kubernetes-headlamp*.yaml`), Helm charts (`/charts/`) — treat as manual-review-only.
+
+---
+
+### Primary entrypoints (exact commands from repository)
```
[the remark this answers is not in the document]
illume: fixed

AGENTS.md
```diff
@@ -0,0 +1,356 @@
...
+- **Steps to Test:** Numbered steps to verify the changes
+- **Screenshots:** Include for UI changes
+- **Notes for the Reviewer:** Any relevant context or areas to focus on
+
+#### PR guidelines (from `/docs/contributing.md`):
+- Run tests: `npm run frontend:test`, `npm run backend:test`
+- Run linters: `npm run frontend:lint`, `npm run backend:lint`
+- Summarize changes and explain _why_ they are needed
+- Provide steps to test the changes
+- Link to related issue via `fixes #ISSUE_NUMBER`
```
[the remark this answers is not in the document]
illume: fixed

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+3. ✅ **Commands are ready for validation** - all commands listed can be run locally following the exact syntax provided
+4. ✅ **File is ready for commit** - consulted-files list is complete and all placeholders have been replaced with exact commands and paths from the repository
+
+**Version Information:**
+- Node.js: >=20.11.1 (from `/package.json`)
+- npm: >=10.0.0 (from `/package.json`)
+- Go: 1.24.11 (from `/backend/go.mod`)
+
+**Key Command Sources:**
+- Build/test commands: `/package.json`
```
yolossn, line 364: Lets also add make file here?

AGENTS.md
```diff
@@ -0,0 +1,357 @@
...
+> - `/CONTRIBUTING.md` - Contributing guidelines
+> - `/OWNERS` - Code reviewers and approvers
+
+---
+
+### Agent persona and scope
+- **@dev-agent** — pragmatic, conservative, test-first, risk-averse.
+- **Scope:** propose, validate, and prepare code/docs patches; run local build/test commands; create PR drafts.
+- **Not allowed:** push images/releases, modify CI or infra, or merge without human approval.
+
```
yolossn, line 58: shall we add this to agent scope to avoid unrelated changes?
```
### Explicit non-goals
The agent should NOT:
- Propose refactors without a clear bug, performance, or maintenance justification
- Change public APIs without explicit request
- Reformat unrelated code
- Rename files or symbols for stylistic reasons
- Introduce new dependencies unless required to fix a bug or implement a requested feature
```

AGENTS.md
joaquimrocha: We are trying to avoid using the Makefile and use the npm run scripts instead, for simplicity. So I would say we don't suggest it to agents.

AGENTS.md
joaquimrocha: I wonder if this will restrict agents in a way we don't intend. I get your point, but I think the best practices below already cover some of the intent here. Maybe you can add some of what you propose as a best practice instead of a "prohibition".

AGENTS.md
illume: Okie, I’ll avoid make

AGENTS.md
illume: Ok, let me try some of these out and I’ll see how it goes.

illume: - Changed them a bit to be more "unless explicitly requested or strictly necessary for the change" rather than a flat out "prohibition".
- Seems to work ok...

But yeah, we can tweak these rules as we go over the next weeks/months/decades ;)

Issue #3979, opened 2025-10-01
Distributed Headlamp settings


joaquimrocha:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

A common request from our users, like @jda258 , has been to be able to configure certain parts of Headlamp or its plugins in a way that will apply to all users (when running Headlamp as a web service).

For example, if someone is using the AI Assistant, instead of having each client/user set up the AI provider, with the needed keys, endpoint, etc. they should just connect to Headlamp and it should be pre-configured to them.

## Describe the solution you'd like

Even if each plugin can implement solutions for distributed configuration however they want, I think that Headlamp should offer plugins a standardized way to read configuration from a cluster (or clusters).
For example, this could be having a specific configmap, in a specific namespace, like "ConfigMap: headlamp-general-config", in a namespace with a specific label like _headlamp.dev.settings-general: configmap:headlamp-general-config_ .

Then a plugin could have its own settings like:
_headlamp.dev.settings-plugin/headlamp-ai-assistant: configmap:headlamp-ai-assistant-settings .

Every time we connect to a cluster, we could check if a namespace has the special label and check for the settings if it does.

When getting the settings via the Headlamp-provided lib, we could have a flag "check-remote = true|false" for allowing the settings to be distributed.

## What users will benefit from this feature?

Vendors who deploy Headlamp and need to have centralized settings for Headlamp or its plugins.

illume:
@guidemetothemoon wrote in #4280 


# Persistent storage for configuration of Headlamp and Headlamp plugins

## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
From what I understand in the Headlamp's source code, configuration settings like for example "Allowed namespaces" are stored locally in the browser so if the user uses a different browser or clears data in the current browser he/she will need to add this information again.

Especially when there is no option currently to at least define distributed settings like Prometheus configuration that would be the same for all users. Ref. existing issue https://github.com/kubernetes-sigs/headlamp/issues/4051

For in-cluster Headlamp this creates confusion among users and unnecessary, repeatable work which negatively affects user experience.

At least I couldn't find any information about possible workarounds for this but please let me know if there's some work around this already.

## Describe the solution you'd like
It would be great to have a possibility to store this configuration permanently somewhere, without being bound to a single browser session. For example in a persistent volume or some kind of config file that can be modified on-demand but that would persist when f.ex. a user switches to a different browser.

[...]

PR #4357, opened 2026-01-15
Add appearance configuration per cluster

### LINKED ISSUE #15: Use accent colors for clusters
Since we support multiple clusters, a nice way to quickly identify them (besides the name) is to assign them a different color. We shall call this the cluster's accent color as it will be reflected in a few parts of the dashboard.

We understand this is not an accessible option for color-blind users, but that should be covered by #14 .

This feature also requires having a settings view (#16).

**Acceptance Criteria:**
- [ ] Allow the user to select a color for each configured cluster
- [ ] Use this color in the cluster's icon (if not custom) and other elements in the UI (TBD)

alexmathieu22:
## Summary

This PR is a rebase of this [PR](https://github.com/kubernetes-sigs/headlamp/pull/4329).

This PR adds accent colors / logo / warning message for each cluster by adding an "Appearance" section in the settings of the clusters. The color change was kept to a minimum as to not disrupt the design, but let us modify it further when a more elaborate design is approved by board. The warning message is optional and it's color is the accent color decided by user. Finally, the icon is currently an icon from the iconify library. This choice was made to kept the change as small as possible, and to use an already used library.

## Related Issue

Fixes #15

## Changes

- Added accent color by cluster
- Added warning message by cluster
- Added icon by cluster

## Steps to Test

1. Navigate to cluster settings
2. Modify appearance as wanted
3. Click on apply
4. Confirm
5. Verify change is persisted for cluster only and looks as required

- See thread of rebased PR

## Notes for the Reviewer

- This touches the i18n layer, so please check language consistency.

### DIFF
--- a/frontend/src/components/App/Home/__snapshots__/index.Base.stories.storyshot
+++ b/frontend/src/components/App/Home/__snapshots__/index.Base.stories.storyshot
@@ -618,11 +618,8 @@
                         href="/c/cluster0/"
                       >
                         <div
-                          class="MuiBox-root css-1or8la4"
+                          class="MuiBox-root css-172hjuw"
                         >
-                          <div
-                            class="MuiBox-root css-iyhyxy"
-                          />
                           <span
                             class="MuiTypography-root MuiTypography-caption css-chqac-MuiTypography-root"
                           >
@@ -729,11 +726,8 @@
                         href="/c/cluster1/"
                       >
                         <div
-                          class="MuiBox-root css-1or8la4"
+                          class="MuiBox-root css-172hjuw"
                         >
-                          <div
-                            class="MuiBox-root css-iyhyxy"
-                          />
                           <span
                             class="MuiTypography-root MuiTypography-caption css-chqac-MuiTypography-root"
                           >
@@ -840,11 +834,8 @@
                         href="/c/cluster2/"
                       >
                         <div
-                          class="MuiBox-root css-1or8la4"
+                          class="MuiBox-root css-172hjuw"
                         >
-                          <div
-                            class="MuiBox-root css-iyhyxy"
-                          />
                           <span
                             class="MuiTypography-root MuiTypography-caption css-chqac-MuiTypography-root"
                           >
--- a/frontend/src/components/Sidebar/ClusterBadge.test.tsx
+++ b/frontend/src/components/Sidebar/ClusterBadge.test.tsx
@@ -50,14 +50,17 @@ describe('ClusterBadge', () => {
     expect(container).toBeInTheDocument();
   });
 
-  it('renders with default icon when no icon provided', () => {
-    const { container } = render(
+  it('renders no icon when no icon provided', () => {
+    render(
       <TestContext>
         <ClusterBadge name="test-cluster" />
       </TestContext>
     );
 
-    expect(container).toBeInTheDocument();
+    // Should not render the icon element
+    expect(screen.queryByTestId('cluster-badge-icon')).toBeNull();
+    // Still renders the cluster name
+    expect(screen.getByText('test-cluster')).toBeInTheDocument();
   });
 
   it('renders with all props', () => {
--- a/frontend/src/components/Sidebar/ClusterBadge.tsx
+++ b/frontend/src/components/Sidebar/ClusterBadge.tsx
@@ -35,8 +35,6 @@ export interface ClusterBadgeProps {
  */
 export default function ClusterBadge({ name, accentColor, icon }: ClusterBadgeProps) {
   const theme = useTheme();
-  const defaultIcon = 'mdi:hexagon-multiple-outline';
-
   const iconColor = theme.palette.primary.main;
   const backgroundColor = theme.palette.secondary.main;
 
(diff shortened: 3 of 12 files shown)


docs/platforms.md
skoeva, line 30: was this meant to be added here?

alexmathieu22: I'm unsure why this appears as a diff... It is the content of a commit I made in another PR which was already merged on master 🤔

joaquimrocha: Yeah. This part should not be here.

joaquimrocha:
I only noticed now that you have backend code related to this appearance PR. I do not think that's where the appearance should be stored.
We have many UI related settings (like the number of rows to show by default in tables, and even the allowed namespaces) set in the local storage. We are planning on adding some ways to then distribute such settings more or less automatically via a ConfigMap or similar.
So please keep this PR frontend only (use a similar way to store the cluster appearance settings like we are doing for the ones I mentioned above), in order to make it more in line with the other settings and the code safer as well.

alexmathieu22:
Does that mean I should also remove the warning banner? The whole point of this MR was to have a unified setup for all devs. Putting this in localstorage would remove the main idea of adding a warning. 

I can do it via localstorage, but do you have a link to the issue / thread regarding:
>  We are planning on adding some ways to then distribute such settings more or less automatically via a ConfigMap or similar.

This is something that I wouldn't mind working on and is quite important for my team to continue using headlamp going forward.

alexmathieu22:
Hi @joaquimrocha @skoeva ! Following your suggestions, I've modified the PR to only include frontend changes, the setting is now applied in the localstorage.

I've also removed the warning banner to keep changes to the minimum.

Here is a demo of the final result:

https://github.com/user-attachments/assets/92930cf6-30ba-4c92-8bd8-c19c2eb22c09

Please let me know how that looks.

Regarding being able to share those settings across users, is there something I could look into to follow progress / be able to help in any way (edit: saw the following issue: https://github.com/kubernetes-sigs/headlamp/issues/3979)? This is something that would be very useful for me and my team to use this - it is a big UX enabler for devs having access to a shared headlamp instance to be able to easily see they are working on a prod cluster without having to set those settings themselves :)

alexmathieu22:
@illume @skoeva @joaquimrocha @sniok any updates regarding this PR?

joaquimrocha:
Hi @alexmathieu22 . Thanks for the PR clean up and updates. I think it looks pretty good in your video. I am a bit concerned how a larger group of clusters will look in the side bar, but I think that the rest should be mergeable. I want to take a closer look and unfortunately didn't have time for it at all this week. Maybe next week I will able to, as I really want to have this.
@sniok , WDYT?

sniok:
image

This button in the sidebar serves so many purposes: link to the cluster overview page, indicator for selected clusters, button that expands submenu. And now with icons it just looks way too "busy". Maybe as a interim solution (before we rework how cluster selector works and moved to the sidebar) we can make it so clusters don't have icons by default, so without manually assigning icons it will look more or less like it used to, what do you think?

frontend/src/helpers/clusterAppearance.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * Extracts and sanitizes appearance fields from localStorage for a given cluster.
+ *
+ * @param clusterNameOrMetaData - The cluster name, or for backward compatibility, cluster metadata.
+ * @returns {ClusterAppearance} Sanitized per-cluster appearance settings from localStorage.
+ */
+export function getClusterAppearanceFromMeta(
+  clusterNameOrMetaData: string | ClusterMetadata | undefined
```
sniok: this function covers cases where the argument is string or undefined but I don't see the `ClusterMetadata` ?

frontend/src/helpers/clusterAppearance.ts
```diff
@@ -0,0 +1,98 @@
...
+): ClusterAppearance {
+  // Handle both old signature (metaData) and new signature (clusterName)
+  let clusterName: string | undefined;
+
+  if (typeof clusterNameOrMetaData === 'string') {
+    clusterName = clusterNameOrMetaData;
+  } else {
+    // Old signature: first param is metadata, try to extract from window location
+    clusterName = window.location.pathname.split('/c/')[1]?.split('/')[0];
```
sniok: we have a `getCluster` function for this that better handles edge cases

frontend/src/components/Sidebar/sidebarSlice.ts
```diff
@@ -33,7 +34,7 @@ export interface SidebarEntry {
   /**
    * Text to display under the name.
    */
-  subtitle?: string;
+  subtitle?: string | React.ReactNode;
```
sniok: nit: ReactNode type already includes string so this can be just `subtitle?: React.ReactNode;`

joaquimrocha:
> Maybe as a interim solution (before we rework how cluster selector works and moved to the sidebar) we can make it so clusters don't have icons by default, so without manually assigning icons it will look more or less like it used to, what do you think?

I assume you are suggesting we keep the icon for the cluster if it's just 1 cluster being shown), ideally replacing the sidebar icon we now have; and if more than one, we don't show the cluster icon at all (keep it as it was).
If that is the case, I agree.

sniok:
> I assume you are suggesting we keep the icon for the cluster if it's just 1 cluster being shown), ideally replacing the sidebar icon we now have; and if more than one, we don't show the cluster icon at all (keep it as it was). If that is the case, I agree.

I'm suggesting that clusters by default don't have an icon. Currently it adds default icons to every cluster.

image

illume:
> I'm suggesting that clusters by default don't have an icon.

Yeah, I think this will look better by default (to have no icons).

It could be a default minikube, kind, k3d (and other testing clusters) icons could be good. But even then some people have like 10 test clusters... so it would have the same problem of lots of un-unique icons.

alexmathieu22:
Good idea to have no default icon, just modified it:
Here's the final look:

with no icon set:
image
image
image
image




with an icon set for a cluster:
image
image
image
image

frontend/src/helpers/clusterAppearance.ts
illume: This was fixed. Thanks!

frontend/src/helpers/clusterAppearance.ts
illume: This was fixed. Thanks!

frontend/src/components/Sidebar/sidebarSlice.ts
illume: This was fixed. Thanks!

joaquimrocha (approved):

joaquimrocha:
@illume I have approved this changes. I think there are a couple of things we could improve but it's also in a good state already and I love having icons for clusters.
@alexmathieu22 , if you can solve the current conflict, we'll then merge your PR. Thanks.

joaquimrocha:
@alexmathieu22 I removed your merge branch hoping I am helping you land this PR.
I am having trouble with headlamp in my own machine because it keeps generating different snapshots and thus failing the tests, so I had to play it a bit by ear here, but I hope this is working fine and is cleaned up now. I did test and things worked fine for me.

alexmathieu22:
Thanks for this! Had trouble with rebasing for one of the files on my local machine, was going to have another look later, but thanks for doing it for me, greatly appreciated :)

PR #4360, opened 2026-01-15
frontend: Remove unused code

skoeva:
This change removes the unused `overrideKubeObject` function from the frontend.

### DIFF
--- a/frontend/src/test/index.tsx
+++ b/frontend/src/test/index.tsx
@@ -58,12 +58,3 @@ export function TestContext(props: TestContextProps) {
     </Provider>
   );
 }
-
-export function overrideKubeObject<U>(kubeObject: U, propsToOverride: Partial<U>) {
-  for (const [key, value] of Object.entries(propsToOverride)) {
-    if (value !== undefined) {
-      // @ts-ignore
-      kubeObject[key] = value;
-    }
-  }
-}


joaquimrocha:
I would double check that this code was not importable by plugins. But I think @sniok can do a better job at reviewing this one.

skoeva:
I did catch that `makeJobStatusLabel` is not imported but copied [in the keda plugin](https://github.com/headlamp-k8s/plugins/blob/c1eb87709df5650cc8fc8c3d038730c64247dea1/keda/src/components/common/CommonComponents.tsx#L501), will take a look through the community plugins. The ones in the plugin repo are clean

joaquimrocha:
> I did catch that `makeJobStatusLabel` is not imported but copied [in the keda plugin](https://github.com/headlamp-k8s/plugins/blob/c1eb87709df5650cc8fc8c3d038730c64247dea1/keda/src/components/common/CommonComponents.tsx#L501), will take a look through the community plugins. The ones in the plugin repo are clean

I think you just need to check if importing any of that code is at all possible. If it is, you should assume anyone can be using it even in a non-public way and we'd be breaking things for them.
If it's not possible to import then we're not breaking things and we can have this one.

skoeva:
@joaquimrocha Got it, that pretty much only leaves `overrideKubeObject` since the rest should be published to the npm package; will update now

illume:
I can't see overrideKubeObject used anywhere on public github https://github.com/search?q=overrideKubeObject&type=code

illume:
btw. I feel if we can't find any users on github/gitlab/google at all, I think it's fine to mark them deprecated, mention them in the release notes that we will remove them in after several releases(6 months or so). We can even ask a few times in our slack channel. If no one gets back to us... I feel it's probably safe at that point to remove them. Something to consider.

PR #4371, opened 2026-01-16
charts: Permit setting hostUsers for deployment

jcpunk:
## Summary

Adds a flag for folks to opt into user namespaces.

## Changes

- Added flag to values.yaml and deployment.yaml

## Steps to Test

1. render chart, get current kubernetes default
2. change hostUsers
3. render chart, get new value

## Notes for the Reviewer

https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/

### DIFF
--- a/charts/headlamp/README.md
+++ b/charts/headlamp/README.md
@@ -151,6 +151,7 @@ config:
 | clusterRoleBinding.create | bool | `true` | Create cluster role binding |
 | clusterRoleBinding.clusterRoleName | string | `"cluster-admin"` | Kubernetes ClusterRole name |
 | clusterRoleBinding.annotations | object | `{}` | Cluster role binding annotations |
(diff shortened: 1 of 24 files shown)


jcpunk:
In theory, copilot's review has been addressed.

jcpunk:
I've updated the documentation. Since it defaults to the current kubernetes default setting, I don't think existing users will notice any change. It will start letting folks opt-in to user namespaces.

illume:
@jcpunk thanks for that. Are you able to see about the github check failure?

```shell
make helm-template-test
```

jcpunk:
perhaps that is better

ashu8912:
Hii @jcpunk the test seems to be failing can you take a look

jcpunk:
Looks like my editor strips trailing white space by default.

Updated the snapshots with a different tool so the spaces remain.

PR #4400, opened 2026-01-20
frontend: a11y: Fix a11y issues and add storybook based a11y testing to CI

illume:
These are a bunch of a11y fixes found by running axe tests on all the storybook states.

### DIFF
--- a/frontend/src/components/verticalPodAutoscaler/List.tsx
+++ b/frontend/src/components/verticalPodAutoscaler/List.tsx
@@ -85,6 +85,7 @@ export default function VpaList() {
                     href="https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler#installation"
                     target="_blank"
                     rel="noopener"
+                    sx={{ textDecoration: 'underline' }}
                   >
                     Learn More
                   </Link>


frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -319,7 +319,7 @@ export const PureSidebar = memo(
           direction="column"
           justifyContent="space-between"
           wrap="nowrap"
-          aria-hidden={!largeSideBarOpen}
+          {...(!largeSideBarOpen && { inert: 'true' as any })}
```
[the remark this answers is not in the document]
illume: Without this the inert is not actually written to the storybook html, and the test fails. So I think it's needed.

If I do the suggested change it doesn't work.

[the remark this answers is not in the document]
illume: Without this the inert is not actually written to the storybook html, and the test fails. So I think it's needed.

If I do the suggested change it doesn't work.

frontend/package.json
```diff
@@ -92,7 +92,6 @@
     "yaml": "^2.7.0"
   },
   "overrides": {
-    "axe-core": "npm:dry-uninstall",
```
sniok, line 95: we've had issues with license scanning for `axe-core` package since it's using MPL which is not allowed for CNCF projects, are you sure we can add it back?

illume: This adds it to dev dependencies rather than a dependency.

It's already included through playwright dev dependency I think.
grep axe package-lock.json 
    "node_modules/axe-core": {
        "axe-core": "^4.9.1",

illume:
@skoeva rebased against main

frontend/package.json
sniok: I think dependency policy also applies to dev dependencies 

"overrides" with uninstall is there to remove it from any existing dependency
in package-lock it's resolved to "dry-uninstall"

```
"node_modules/axe-core": {
      "name": "dry-uninstall",
```

illume: Ok. Let me remove this commit.

We can apply for an exception later. https://contribute.cncf.io/projects/submit-license-exception-request/ I checked their list of already approved exceptions, however no one has yet applied for axe.

illume: removed

illume:
Removed the testing stuff (because CNF not liking MPL), only the a11y fixes remain to review.

vyncent-t:
> running on new changes

tests run fine LGTM

PR #4409, opened 2026-01-21
frontend: src: Use kind and apiVersion for Link component routing

### LINKED ISSUE #4267: Link component routes Knative Service to core Service page when using `kubeObject`
## Describe the bug

When using `<Link kubeObject={r}>{r.metadata.name}</Link>`, if the provided `kubeObject` is a **Knative Service**, Headlamp incorrectly treats it as a **core Kubernetes Service** and navigates to the core Service details page instead of the Knative Service page.

**Expected behavior:**
The Link should route to the correct resource page based on both `kind` and `apiVersion`.

## To Reproduce

Steps to reproduce the bug:

1. Create or obtain a Knative Service resource.
2. Render a link using `<Link kubeObject={r}>{r.metadata.name}</Link>`.
3. Click the link.
4. Observe that it navigates to the core Service page instead of the Knative Service page.

## Environment

* Installation type: Mac-Homebrew
* Headlamp Version: 0.38.0
* Other: Occurs with Knative Service (`apiVersion: serving.knative.dev/v1`)

No

## Additional Context

The root cause appears to be that `Link` determines the target page **only by `kind`**, and does not take `apiVersion` into account.
Since both core Service and Knative Service share the same `kind: Service`, this results in incorrect routing.

This is not specific to Knative and may affect any CRDs that reuse core resource kinds.

**Example Knative Service class definition:**

```typescript
export class KService extends KubeObject {
  static kind = 'Service';
  static apiName = 'services';
  static apiVersion = 'serving.knative.dev/v1';
  static isNamespaced = true;

    return '/knative/services/:namespace/:name';
  }

  static get listRoute(): string {
    return '/knative/services';
  }
}
```

Despite having a custom `detailsRoute`, the `Link` component ignores this and routes based solely on `kind: Service`, leading to incorrect navigation.

dibyanshu-pal-kushwaha:
## Summary

This PR fixes the incorrect routing of resources that share the same Kind name as core Kubernetes resources (but belong to different apiGroups), such as Knative Services.Currently in the codebase, any Non core service with kind Service navigates to the core Service details page instead of the dedicated Service page.So , This PR helps in correct routing.

## Related Issue

Fixes #4267

## Changes

1. Link component: Updated the routing logic to check both kind and apiGroupName.
2.Collision Detection: Added a check to verify if a resource's API group matches the "standard" core group before attempting to open the right-side details drawer.
3.Routing: Ensures that non-standard resources (like Knative Services) bypass the standard drawer and correctly navigate to their specific full-page views.

## Steps to Test

1. Verify Core Resources: Navigate to Network > Services. Click on the kubernetes service.
Expected: The right-side drawer should open normally.
2. Verify colliding CRDs (Knative): Navigate to Custom Resources > services.serving.knative.dev. Click on a Knative      Service link. 
Expected: You should be redirected to the Knative Service details page. The core Service drawer should not open.

## Screenshots (After Fixing)
For Core Services : 
Core

For Custom Services : 
As you can see clearly , here I have created a fake Knative service to check and it is working positively , routing to the dedicated page rather than Core K8s services.
Custom

## Notes for the Reviewer

- The implementation is generic and does not hardcode "Knative". It will work for any CRD that happens to reuse a core Kind name .
- This change specifically touches the Link component which is used throughout the dashboard, so it improves routing across all resource tables.


kahirokunn (approved):

kahirokunn:
The matchesStandard check appears to be functioning correctly.

kahirokunn:
Did it function properly when you actually tested it manually? 👀

skoeva (changes requested):
Hi! Would you mind addressing the copilot review comments and the linting failure in the CI?

kahirokunn:
After reviewing the source code, it appears that the comparison is likely being made solely based on the kind without checking the apiGroup. Could you explain why this resolved the issue? 🙏

dibyanshu-pal-kushwaha:
@kahirokunn 
>After reviewing the source code, it appears that the comparison is likely being made solely based on the kind without checking the apiGroup. Could you explain why this resolved the issue? 

Yes sir Sorry for the late reply , I currently applied the kind and apiGroup check for correct routing. I verified this locally on Minikube by creating a collision scenario . The Link component now correctly identifies it as a generic resource and routes to the generic page instead of the core Service drawer.Please review sir

frontend/src/components/common/Link.tsx
```diff
@@ -141,68 +144,102 @@ function PureLink(
...
+    let groupMatches = false;
+
+    if (obj.jsonData && obj.jsonData.apiVersion && objClass.apiVersion) {
+      const instanceGroup = getApiGroup(obj.jsonData.apiVersion);
+      const classVersions = Array.isArray(objClass.apiVersion)
+        ? objClass.apiVersion
+        : [objClass.apiVersion];
+      const classGroups = classVersions.map(v => getApiGroup(v));
+
+      if (classGroups.includes(instanceGroup)) {
```
kahirokunn: Using `find` completes the task in a single operation, rather than the two-step process of `map` followed by `includes`.

frontend/src/components/common/Link.tsx
```diff
@@ -141,68 +144,102 @@ function PureLink(
...
+  return apiVersion.split('/')[0];
+}
+
 export default function Link(props: React.PropsWithChildren<LinkProps | LinkObjectProps>) {
   const drawerEnabled = useTypedSelector(state => state?.drawerMode?.isDetailDrawerEnabled);
 
   const { tooltip, ...propsRest } = props as LinkObjectProps;
 
-  const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props?.routeName;
+  const kind = 'kubeObject' in props ? props.kubeObject?.kind : props?.routeName;
```
kahirokunn, line 162: Why are we changing this? Was there a problem?

frontend/src/components/common/Link.tsx
```diff
@@ -141,68 +144,102 @@ function PureLink(
...
+    if (!kindMatches || !groupMatches) {
+      matchesStandard = false;
+    }
+  }
+
   const openDrawer =
-    drawerEnabled && canRenderDetails(kind)
+    drawerEnabled && canRenderDetails(kind) && matchesStandard
       ? () => {
-          // Object information can be provided throught kubeObject or route parameters
```
kahirokunn, line 169: Why was this comment deleted?

frontend/src/components/common/Link.tsx
dibyanshu-pal-kushwaha: Here if we were using :
>const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props?.routeName;

then for the Generic resources whose kind is generally undefined then using this becomes undefined and can cause other issue where this kind is used but if we replace it to the 

> const kind = 'kubeObject' in props ? props.kubeObject?.kind : props?.routeName;

then  this current issue is also managed along with not creating the inconsistency.

illume (changes requested):
Thanks @dibyanshu-pal-kushwaha 

It looks like the lint CI check is failing. Please check the commands locally? format, lint, tsc, and test

Can you please check the open review comments?

frontend/src/components/common/Link.tsx
illume: Please add this back, unless there is a good reason to remove it?

frontend/src/components/common/Link.tsx
illume: Good point.

frontend/src/components/common/Link.tsx
illume: @kahirokunn does this change make sense now?

@dibyanshu-pal-kushwaha If there is a change like this, usually it's a good idea to explain why it was made in the git commit message.

kahirokunn: Thank you for asking. Looking at the `KubeObject` implementation:

```ts
export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
  get kind() {
    return this.jsonData.kind;
  }

  _class() {
  }
}
```

I couldn't find a rational reason why we would need to go through the constructor just to retrieve the `kind` value. Since `kind` is already accessible directly from `jsonData.kind` via the getter, I believe this change makes sense and we can adopt it.

frontend/src/components/common/Link.tsx
dibyanshu-pal-kushwaha: I restored this in the latest commit.

frontend/src/components/common/Link.tsx
dibyanshu-pal-kushwaha: Fixed

dibyanshu-pal-kushwaha:
@kahirokunn @illume , I made the changes by you and copilot , please review it.

frontend/src/components/common/Link.tsx
illume: Thanks!

frontend/src/components/common/Link.tsx
illume: thanks!

frontend/src/components/common/Link.tsx
illume: > I believe this change makes sense and we can adopt it.

Thanks for the explanation. I'll mark this as resolved.

illume (changes requested):
Thank you. Looking pretty good now.

Just the commit message needs tidying up. Can you please squash it?

Also the first commit message body should be rewritten. It currently contains previously squashed subject lines in the commit body.

```
frontend: src: Use kind and apiGroup for Link component routing
frontend: src: Change apiVersion to apiGroup and fix linting error

frontend: src: Fix check using kind and apiGroup
```

alokdangre:
https://github.com/user-attachments/assets/275f0592-d8da-42d4-8c6e-bd56651b2ab6

@illume @dibyanshu-pal-kushwaha it did'nt worked as expected

alokdangre:
@dibyanshu-pal-kushwaha plzz test it locally and provide ss/sr

frontend/src/components/common/Link.tsx
```diff
@@ -141,19 +144,52 @@ function PureLink(
...
   const { tooltip, ...propsRest } = props as LinkObjectProps;
 
-  const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props?.routeName;
+  const kind = 'kubeObject' in props ? props.kubeObject?.kind : props?.routeName;
   const cluster =
     'kubeObject' in props && props.kubeObject?.cluster
       ? props.kubeObject?.cluster
       : props.activeCluster ?? getCluster() ?? '';
 
+  let matchesStandard = true;
```
alokdangre, line 168: the matchesStandard check currently returns true for colliding crds because it compares the object's api group against its own constructor's apigroup,,, headlamp instantiates Custom Resources using dynamic classes that correctly match the crd api group, the check passes, and the standard drawer is still triggered.

alokdangre (changes requested):
you can check if the object's class has the customResourceDefinition property. This property is typically present on dynamic classes created for erdbut absent on standard core classes

illume:
Thanks for the review @alokdangre

dibyanshu-pal-kushwaha:
> @illume @dibyanshu-pal-kushwaha it did'nt worked as expected 

@alokdangre , please go through the issue #4267  (You misunderstood this issue) as this states that currently in the codebase , when any non-core Kubernetes services with same kind Service is there and we click it over there it navigates to the core Service details page instead of the dedicated Service page. Your Screenshot shows correctly as it does not navigate to any core services.

I have already tested it locally ......

This is core service navigation->
Core

This is for Custom Service:

Custom

dibyanshu-pal-kushwaha:
@illume Sir it worked correctly, I think.

alokdangre:
@dibyanshu-pal-kushwaha i think maybe i have misunderstood it, but the same ss is working in current version as well

kahirokunn:
You need to create a Custom Resource KubeObject like this:

```ts
class KService extends KubeObject<KServiceResource> {
  static kind = 'Service';
  static apiName = 'services';
  static apiVersion = 'serving.knative.dev/v1';
  static isNamespaced = true;
}
```

And to properly test this, You need to pass the retrieved object to the Link component as follows:

```tsx
<Link kubeObject={knativeServiceObject} />
```

Currently, this navigates to the core Service resource (networking).

My issue would be resolved if the link navigates to the route defined for KService instead of the networking Service resource. If there is no route defined for KService, I believe it would be appropriate to navigate to the target resource page in the custom resources tab.

illume:
@dibyanshu-pal-kushwaha can you please reply to the last review comment?

frontend/src/components/common/Link.tsx
dibyanshu-pal-kushwaha: Thanks for pointing out this!.
Fixed in the latest commit.

frontend/src/components/common/Link.tsx
```diff
@@ -141,19 +144,52 @@ function PureLink(
...
+
+      if (classVersions.find(v => getApiGroup(v) === instanceGroup)) {
+        groupMatches = true;
+      }
+    }
+
+    if (!kindMatches || !groupMatches) {
+      matchesStandard = false;
+    }
+  }
```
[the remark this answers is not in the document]
dibyanshu-pal-kushwaha: Fixed!

frontend/src/components/common/Link.tsx
```diff
@@ -141,19 +144,52 @@ function PureLink(
...
+
   const openDrawer =
-    drawerEnabled && canRenderDetails(kind)
+    drawerEnabled && canRenderDetails(kind) && matchesStandard
       ? () => {
           // Object information can be provided throught kubeObject or route parameters
```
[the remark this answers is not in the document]
dibyanshu-pal-kushwaha: Done!

illume (changes requested):
Thanks for the contribution.


<details>

</details>


There are some open Copilot review comments — could you take a look at them? (edit... I marked all the ones done resolved, so there's none left open now).


Can you please look at the filing CI jobs?

illume (changes requested):
Thanks for this PR.

PR #4446, opened 2026-01-25
charts: headlamp: Use HEADLAMP_CONFIG_ env var 

### LINKED ISSUE #4080: Helm chart uses wrong environment variables
## Describe the bug
When deploying Headlamp with the Helm chart, it always includes these args:
```
args:
- "-oidc-client-id=$(OIDC_CLIENT_ID)"
- "-oidc-client-secret=$(OIDC_CLIENT_SECRET)"
- "-oidc-idp-issuer-url=$(OIDC_ISSUER_URL)"
- "-oidc-scopes=$(OIDC_SCOPES)"
```
https://github.com/kubernetes-sigs/headlamp/blob/74ddd85dce9f485aaf2f3c8ceb767b4d3f0a310e/charts/headlamp/templates/deployment.yaml#L249

These environment variables do not exist in Headlamp, the correct keys are `HEADLAMP_CONFIG_OIDC_CLIENT_ID`, etc. It is confusing to add an inconsistent set of environment variables and conflicts with Headlamp documentation.

If following security best practices and using an external secret to set OIDC config parameters, this is exceptionally confusing. Users need to use a confusing mixture of envvars, with the above 4 needing to be used without the `HEADLAMP_CONFIG_` prefix in the secret. If users need to add any additional OIDC config, like `HEADLAMP_CONFIG_OIDC_VALIDATOR_CLIENT_ID` for AKS environments, those must have the HEADLAMP_CONFIG_ prefix. If a user follows Headlamp Documentation and tries to use consistent `HEADLAMP_CONFIG_` prefixed envvars, then the user will get crashes because the Helm chart has overridden the standard envvar-sourced OIDC config with empty strings. If a user tries to remove the prefixes to prevent these crashes, then variables like `HEADLAMP_CONFIG_OIDC_VALIDATOR_CLIENT_ID` will be silently ignored, causing confusing, hard-to-debug OIDC issues. The way the Helm chart requires users to use a mixture of prefixes is very confusing and not documented.

It does not seem useful for the Helm chart to set these args in the first place, my understanding is that it is much simpler and consistent to just use envvars directly instead of translating envvars to args at runtime.

Yes (I will propose a PR)

dibyanshu-pal-kushwaha:
Summary
This PR fixes an issue where the Helm chart could override valid OIDC settings with empty values.It removes hardcoded command-line flags and relies only on the standard HEADLAMP_CONFIG_ environment variables. This ensures that user-provided configuration is read correctly and not ignored.

Related Issue
Fixes #4080

Changes
-Updated charts/headlamp/templates/deployment.yaml to strictly use HEADLAMP_CONFIG_ prefixed environment variables for all OIDC settings, aligning with Headlamp's native configuration system.
-Removed the logic that passed OIDC settings as command-line arguments. This prevents empty values from overriding valid environment variable configuration.
-Removed redundant variable extraction logic from the top of the deployment template.

Steps to Test
1.Navigate to chart directory : cd charts/headlamp
2.Use any dummy OIDC values :
 >helm template . --set config.oidc.clientID=test-client-id --set config.oidc.issuerURL=https://test-issuer.com

3.After inspecting we will get that environment variables like HEADLAMP_CONFIG_OIDC_CLIENT_ID are present and correctly set with no no command-line arguments.

Screenshots (if applicable)

Notes for the Reviewer

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -1,51 +1,7 @@
 {{- $oidc := .Values.config.oidc }}
 {{- $env := .Values.env }}
 
-{{- $clientID := "" }}
-{{- $clientSecret := "" }}
-{{- $issuerURL := "" }}
-{{- $scopes := "" }}
-{{- $callbackURL := "" }}
-{{- $validatorClientID := "" }}
-{{- $validatorIssuerURL := "" }}
-{{- $usePKCE := "" }}
-{{- $useAccessToken := "" }}
-{{- $meUserInfoURL := "" }}
 
-# This block of code is used to extract the values from the env.
-# This is done to check if the values are non-empty and if they are, they are used in the deployment.yaml.
-{{- range $env }}
-  {{- if eq .name "OIDC_CLIENT_ID" }}
-    {{- $clientID = .value }}
-  {{- end }}
-  {{- if eq .name "OIDC_CLIENT_SECRET" }}
-    {{- $clientSecret = .value }}
-  {{- end }}
-  {{- if eq .name "OIDC_ISSUER_URL" }}
-    {{- $issuerURL = .value }}
-  {{- end }}
-  {{- if eq .name "OIDC_SCOPES" }}
-    {{- $scopes = .value }}
-  {{- end }}
-  {{- if eq .name "OIDC_CALLBACK_URL" }}
-    {{- $callbackURL = .value }}
-  {{- end }}
-  {{- if eq .name "OIDC_VALIDATOR_CLIENT_ID" }}
-    {{- $validatorClientID = .value }}
-  {{- end }}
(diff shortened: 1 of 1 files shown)


illume (changes requested):
@dibyanshu-pal-kushwaha I think this breaks backwards compatibility 

I see the test cases are failing because of it. To change the names, I think the old ones should still work, otherwise when people update their installs won't work.

Also, it's a complicated issue... it's a bit confusing to me anyway. Probably I'd need to sit and look at this for a while.

Please check the open copilot comments?

illume (changes requested):
Thanks for these changes.


<details>

Good examples:

</details>


skoeva (commented):
this is superseded by #5446 and hasn't had activity in months, will close this now

PR #4450, opened 2026-01-25
frontend: projects: Check for existing project names

### LINKED ISSUE #4087: Projects: Can add clusters to existing projects via same name
## Describe the bug

**When creating a project, if you use the same name as an already existing project, the new cluster will be added to the already existing project.**

If you are creating a project and unintentionally use the name of a project that already exists, the cluster and namespace that you are trying to create a project with will be added to that existing project.

There is no way to know if a name is already in use as of this moment, the UI just lets you create, and in this scenario it just adds whatever you were trying to make into that existing project.

This could be dangerous as there does not seem to be a way to undo this within the Headlamp UI.

If this is not a bug, there should maybe be a warning about using a project name that is already in use.

We could also maybe make use of this by reworking this flawed logic into a "Add cluster" functionality but within the Overview page of a project.

## To Reproduce
Steps to reproduce the bug:
1. Navigate to Projects tab at the home page
2. Click create project and choose "New Project" and not YAML (have not tried this for yaml yet)
3. Create a project named `example-project` etc, 
4. Choose one default minikube cluster as your `clusters` and set `namespace` to default
5. After project is made return to project list
6. Create another project using the same name `example-project` but add a different cluster with same or different namespace.
7. Notice that the `example-project` now has the newly added clusters / namespaces

 - Ubuntu
 - Minikube
 - Docker
 - Headlamp Version: 0.36.0

maybe

pallava-joshi:
## Summary

Fixes #4087

## Changes

Added validation to prevent creating projects with duplicate names in both project creation flows:
- NewProjectPopup.tsx - "New Project" dialog
- ProjectCreateFromYaml.tsx - "Create from YAML" page
- When a user enters a project name that already exists, an error message is displayed and the Create button is disabled
- Added the translation string "A project with this name already exists" to frontend/src/i18n/locales/en/translation.json

## Steps to Test

1. Navigate to the Projects tab on the home page
2. Click "Create Project" and select "New Project"
3. Create a project named test-project with any cluster and namespace
4. After the project is created, return to the project list
5. Click "Create Project" again and try to create another project with the same name test-project
6. Observe that an error message "A project with this name already exists" appears and the Create button is disabled
7. repeat fro yaml

https://github.com/user-attachments/assets/2fcb0659-1bc4-448f-a70c-dc53cc7e45e3

### DIFF
--- a/frontend/src/components/project/NewProjectPopup.tsx
+++ b/frontend/src/components/project/NewProjectPopup.tsx
@@ -29,7 +29,7 @@ import {
   useTheme,
 } from '@mui/material';
 import { uniq } from 'lodash';
-import { ReactNode, useCallback, useState } from 'react';
+import { ReactNode, useCallback, useMemo, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router';
 import { useClustersConf } from '../../lib/k8s';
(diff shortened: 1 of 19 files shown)


frontend/src/components/project/NewProjectPopup.tsx
```diff
@@ -119,8 +119,30 @@ function ProjectFromExistingNamespace({ onBack }: { onBack: () => void }) {
     clusters: selectedClusters,
   });
 
+  // Fetch all namespaces with project labels to check for existing projects
+  const { items: allProjectNamespaces } = Namespace.useList({
```
sniok: we're already fetching namespaces directly above, please reuse the `namespaces` list

illume:
Thanks @vyncent-t

PR #4453, opened 2026-01-25
frontend: fix plugin settings props not passed

### LINKED ISSUE #4062: registerPluginSettings components are not passed props?
## Describe the bug

Components given to registerPluginSettings  are told they will get a data, and onDataChange props.

But this doesn't seem to be the case.

A bunch of plugins (but not all) also use "ConfigStore", which seems to be a separate way of saving settings.

## To Reproduce

Steps to reproduce the bug:
1. Go to a plugin that uses registerPluginSettings 
2. In the component given to registerPluginSettings log the props.
3. See in web dev console that data and onDataChange are undefined.

 - Installation type: MacOS headlamp app in dev mode

## Additional

My guess is this is tech debt, and some plugins are broken that aren't using ConfigStore. But what to do from here? 

Should data and onDataChange be passed in? Will removing them ConfigStore using plugins?

Which plugin settings are working or broken? The typechecker is now catching components passed into registerPluginSettings  that do not use onDataChange and data props.

Are onDataChange and data props deprecated but not formally? There's not deprecated note in any docs.

alokdangre:
## Summary
Fix plugin settings behavior so `registerPluginSettings` components always receive `data` and `onDataChange`, while preserving existing save-button semantics.

## Related Issue
Fixes #4062

## Changes
- Always pass `data` and `onDataChange` to function components used in plugin settings.
- Preserve `displaySettingsComponentWithSaveButton` behavior:
  - `true`: explicit save flow remains unchanged.
  - `false`: changes are persisted immediately via `onSave` from `onDataChange`.
- Add regression tests for both modes:
  - auto-save mode (no save button)
  - explicit save mode (save button shown)

## Steps to Test
   1. Register a plugin settings component with `displaySaveButton=true` and confirm edits do not persist until Save.
   2. Register one with `displaySaveButton=false` and confirm edits persist immediately through `onDataChange`.

## Notes for the Reviewer
This addresses the original props mismatch while keeping the save-button contract intact, which avoids making `displaySettingsComponentWithSaveButton` a no-op.

### DIFF
--- a/frontend/src/components/App/PluginSettings/PluginSettingsDetails.test.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettingsDetails.test.tsx
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from '@mui/material/styles';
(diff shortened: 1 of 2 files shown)


alokdangre:
@illume ptal

sniok:
I'm pretty sure that this fix is not needed, the props should only be passed if the value is true `displaySaveButton`

Also left a comment here https://github.com/kubernetes-sigs/headlamp/issues/4062#issuecomment-4215291114

PR #4454, opened 2026-01-25
backend, frontend: Enable plugin reload on non-cluster pages

### LINKED ISSUE #3040: plugin reloading is broken on non cluster pages (like settings)
## Describe the bug

For me reloading with "npm start" is broken except on the home page.

Note, only seems to happen only on non-cluster pages that don't do backend requests.

## To Reproduce

Steps to reproduce the bug:

- Open headlamp app (in dev mode or with compiled app)
- Go to a non home page (eg settings)
- cd plugins/examples/podcounter/ && npm i && npm start
- Note how the app doesn't reload
- Make a change to plugin, note how it doesn't reload

WSL ubuntu, running the app

alokdangre:
## Summary

The backend detects the change, and the frontend (polling every 5s) sees the signal and triggers the refresh automatically.

## Related Issue

Fixes #3040 

## Changes

- Added usePluginRefreshCheck hook: Implemented a new hook that polls the /plugin-refresh endpoint to detect plugin changes. Use setInterval to check periodically (default 5s).
- Updated Plugins.tsx: Integrated the new hook to run specifically in Electron or Development environments when not viewing a cluster page.
- Fixed plugin hot-reloading: Resolved an issue where plugin updates were ignored (no reload triggered) when users were on non-cluster pages (e.g., Home, Settings).
- Added Unit Tests: Added comprehensive tests in usePluginRefreshCheck.test.ts to verify polling logic, enabled/disabled states, error handling, and cleanup.

## Steps to Test

1. Navigate to Non-Cluster Page: Go to the main Home page or Settings page (ensure you are not inside a specific cluster view).
2. Modify a Plugin: In your local plugins directory, make a small change to a plugin (e.g., update its package.json version).
3. Observe Behavior: usage of the app should automatically trigger a page reload within 5 seconds to apply the plugin change.
4. Verify Polling: Open the browser's Network tab and confirm that requests to /plugin-refresh are occurring every 5 seconds while on the Home page, and that they stop when you navigate to a specific cluster page.

https://github.com/user-attachments/assets/a928db5e-f53f-4a6d-8ea3-3f38f1e9cee5

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -265,6 +265,7 @@ func addPluginRoutes(config *HeadlampConfig, r *mux.Router) {
 	}
 
 	addPluginListRoute(config, r)
+	addPluginRefreshRoute(config, r)
 
 	// Serve development plugins
 	pluginHandler := http.StripPrefix(config.BaseURL+"/plugins/", http.FileServer(http.Dir(config.PluginDir)))
@@ -387,6 +388,31 @@ func addPluginListRoute(config *HeadlampConfig, r *mux.Router) {
 			} else if config.Telemetry != nil {
 				span.SetStatus(codes.Ok, "Plugin list retrieved successfully")
 			}
+		}
+	}).Methods("GET")
+}
+
+// addPluginRefreshRoute registers a GET endpoint handler at "/plugin-refresh" that checks
+// if plugins need to be reloaded. This allows non-cluster pages (like Settings)
+// to poll this endpoint for plugin changes or reload signals.
+func addPluginRefreshRoute(config *HeadlampConfig, r *mux.Router) {
+	r.HandleFunc("/plugin-refresh", func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		var span trace.Span
+
+		if config.Telemetry != nil {
+			_, span = telemetry.CreateSpan(ctx, r, "plugins", "checkPluginRefresh")
+
+			defer span.End()
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		plugins.HandlePluginReload(config.cache, w)
+		response := map[string]bool{"ok": true}
+		if err := json.NewEncoder(w).Encode(response); err != nil {
+			logger.Log(logger.LevelError, nil, err, "encoding plugin refresh response")
+		} else if config.Telemetry != nil {
+			span.SetStatus(codes.Ok, "Plugin refresh check completed")
 		}
 	}).Methods("GET")
 }
--- a/frontend/src/plugin/Plugins.tsx
+++ b/frontend/src/plugin/Plugins.tsx
@@ -19,11 +19,13 @@ import { SnackbarKey, useSnackbar } from 'notistack';
 import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
-import { useHistory } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import { isElectron } from '../helpers/isElectron';
+import { getClusterPathParam } from '../lib/cluster';
 import { useTypedSelector } from '../redux/hooks';
 import { fetchAndExecutePlugins } from './index';
 import { pluginsLoaded, setPluginSettings } from './pluginsSlice';
+import { usePluginRefreshCheck } from './usePluginRefreshCheck';
 
 /**
  * For discovering and executing plugins.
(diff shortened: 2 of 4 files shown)


illume (changes requested):
Thanks :)

There is a github CI check failing. Please see locally `npm run backend:test` and `npm run backend:lint` `npm run backend:format`?

backend/cmd/headlamp.go
```diff
@@ -391,6 +392,30 @@ func addPluginListRoute(config *HeadlampConfig, r *mux.Router) {
 	}).Methods("GET")
 }
 
+// addPluginRefreshRoute registers a GET endpoint handler at "/plugin-refresh" that checks
+// if plugins need to be reloaded. This allows non-cluster pages (like Settings) to poll
```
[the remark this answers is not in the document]
illume: looks like a good suggestion

frontend/src/plugin/Plugins.tsx
```diff
@@ -43,6 +44,10 @@ export default function Plugins() {
 
   const settingsPlugins = useTypedSelector(state => state.plugins.pluginSettings);
 
+  // Poll for plugin refresh status to enable hot-reload on non-cluster pages
+  // This is needed in Electron mode or development mode where local plugin development is done
+  usePluginRefreshCheck(isElectron() || import.meta.env.DEV);
```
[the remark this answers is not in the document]
illume: Ah right. Yeah, maybe add a check to see if it's currently in a cluster view URL, and if so avoid the check?

backend/cmd/headlamp.go
```diff
@@ -391,6 +392,30 @@ func addPluginListRoute(config *HeadlampConfig, r *mux.Router) {
...
+func addPluginRefreshRoute(config *HeadlampConfig, r *mux.Router) {
+	r.HandleFunc("/plugin-refresh", func(w http.ResponseWriter, r *http.Request) {
```
sniok, line 399: this should be a WebSocket IMHO
unless developer is working on a plugin, 'plugin refresh' is a very rare event and having 5 second poll seems like an overkill.
and also if we're going to have a separate mechanism to initiate reload for plugins we should remove the old one

sniok (changes requested):

backend/cmd/headlamp.go
illume: There are some cons for websocket. I'll write more later where this issue pops up.

illume: > this should be a WebSocket IMHO

An optional WebSocket implementation would be good that falls back to polling. WebSockets don't work for some people.

>unless developer is working on a plugin, 'plugin refresh' is a very rare event and having 5 second poll seems like an overkill.

When developing plugins it's a very common experience. Currently none of the non-cluster pages reload. Note developing plugins happens in dev mode and with just the app.

This means that installing and changing plugin settings give a pretty broken experience at the moment. 

> and also if we're going to have a separate mechanism to initiate reload for plugins we should remove the old one
The old one reuses the cluster check, so this can't be used.

frontend/src/plugin/usePluginRefreshCheck.ts
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    intervalRef.current = setInterval(checkPluginRefresh, pollInterval);
+    return () => {
+      if (intervalRef.current) {
+        clearInterval(intervalRef.current);
+        intervalRef.current = null;
+      }
+    };
+  }, [enabled, pollInterval]);
```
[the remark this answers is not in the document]
alokdangre: added tests

backend/cmd/headlamp.go
alokdangre: @illume @sniok  should i strictly scope it for dev mode

illume: The problem is that people develop plugins with the app, as well as in dev mode.

sniok: > WebSockets don't work for some people.

we use websockets for hot reloads on the frontend in dev mode so it's safe to assume that in dev environment websockets will work. most websocket issues were with `in-cluster` setup

alokdangre: my current implementation we call plugin-refresh in both always but in non-cluster page only, so not much calling

illume: @alokdangre @sniok 

Does this sound good?
- A **websocket implementation only** is best for this. (No polling implementation is necessary. Because the websocket issues are mostly for people behind an nginx-ingress with default settings, or for people behind a corp proxy that doesn't do websockets.). Because polling on these views uses power and resources... and it's much better to not have this (unless we really need to).
- However, it should be **kept separate from our other plugin reloading code** (as it is now). Because it keeps this simpler, without risking introducing bugs into the other code. Also the other code is good because for this case lots of people will be using it where websocket issues remain.
- For app, and dev mode (as it is now). Because people develop plugins with both the app, and in dev mode.

alokdangre: this idea is very nice and optimised , when @sniok  agrees, i will start making changes according to the this plan,

illume (changes requested):
Thanks for these changes.


<details>

Good examples:

</details>

PR #4459, opened 2026-01-26
frontend: CopyableCell: Integrate copy in tables

### LINKED ISSUE #3983: Add a button to quickly copy values from table cells where it matters

Some tables shows values that the user may wish to copy, like the ID/name of the resource, IPs, etc.
Recently while copying an IP I felt like we should have a quick way to copy values like this. There is some work to be done in order to identify which values make sense to be copied.

I believe that the values that are likely to be copied by the user should have a copy button overlaying them on hover.

Everyone who uses Headlamp to visualize resources and will use the data visualized in other tools (like connecting to an IP they have seen in a table).

pallava-joshi:
## Summary

This PR adds copy-to-clipboard functionality to table cells by introducing a new CopyableCell component that displays a copy button on hover. This allows users to easily copy IP addresses and other values from resource tables.

Look at IP address columns (Pod IP, Cluster IP, External IP, Internal IP, Addresses)

## Related Issue

Fixes #3983

## Steps to Test

https://github.com/user-attachments/assets/be452653-5cb0-49e1-983c-41a6dc4ea8ed


illume (changes requested):
@pallava-joshi thanks for this.

Could you please update the commit message to be something like this?
> frontend: CopyableCell: Integrate copy in tables

Also, please remove the bun.lock file. (We only support using node and npm, but if it helps you please raise a new PR to add bun.lock files to the .gitignore?)

illume (approved):
Thanks! 🎉 

This looks good to me. I'll wait for another reviewer though. Can you please ask for someone to test or review your PR in the headlamp channel on Kubernetes slack?

joaquimrocha:
Hey @pallava-joshi . Did you take into account the built-in copy feature for the table component we use?
https://v2.material-react-table.com/docs/guides/click-to-copy

pallava-joshi:
The main difference I noticed is that the built-in feature makes the entire cell content a clickable button (always visible), whereas this custom implementation shows a copy icon only on hover, which I thought was a bit more subtle UX-wise. Would you prefer I switch to the built-in approach?

illume:
How it's not showing all the time is nice.

pallava-joshi:
@joaquimrocha final thoughts?

joaquimrocha:
Hi @pallava-joshi . I'm at an event this weekend. So I haven't been able to investigate what's the best solution. Ideally we'd try to find a way to use the built in copy if it does allow to show it only on hover. I'm surprised if there's no such option.

I'll try to look it up closer next week.

illume (changes requested):
With the latest changes the snapshot tests are failing.

`npm run test -- -u`

joaquimrocha:
@pallava-joshi I was checking and it seems like it's possible to have your own renderer for the enableClickToCopy. Something like:
```ts
const CopyOnHoverCell = ({ value }) => {
  const [hover, setHover] = useState(false);

  const copy = () => navigator.clipboard.writeText(value);

  return (
    <div
      style={{ display: 'flex', alignItems: 'center', position: 'relative' }}
      onMouseEnter={() => setHover(true)}
      onMouseLeave={() => setHover(false)}
    >
      <span style={{ flex: 1 }}>{value}</span>

      {hover && (
        <Tooltip title="Copy">
          <IconButton
            size="small"
            onClick={copy}
          >
            <Icon ... />
          </IconButton>
        </Tooltip>
      )}
    </div>
  );
};

```

But it should also work if you use a CSS approach (probably cleaner too).

I think the advantage of using the built-in feature is that likely we get less code to maintain and also we can easily choose to turn it on per table, etc.
Do you think it's worth adapting your code to use this?

illume:
Thanks @joaquimrocha 

@pallava-joshi

Note the code written there should not be copied as is... it's demo/proof of concept code.

- This should be keyboard navigable, not just usable with mouse.
- Use mui components not html ones

Maybe the built in functionality also did many iterations to handle cases like keyboard navigation gracefully. Apart from being easier to maintain going forward, I expect they have considered more and other design problems. We can also expect improvements from them going forward.

pallava-joshi:
@joaquimrocha sure, I'll dig into it.

illume (changes requested):
Thanks for these changes.


skoeva (changes requested):
hi, could you rebase this pr against main?

I also noticed we have an existing `CopyButton` in the code - we could probably render that next to the cell content to avoid duplicating code

skoeva (commented):
this PR has gone stale, closing in favor of #7097

PR #3339, opened 2025-05-23
frontend: ClusterTable: Add delete cluster functionality

userAdityaa:
#### Overview
This PR introduces new features to enhance cluster management in the Kubernetes dashboard. It modifies `ClusterContextMenu.tsx` to allow single cluster deletion for all cluster types (previously limited to plugin-based clusters) and adds bulk delete functionality in `ClusterTable.tsx`. Additionally, it improves the toolbar button layout for a more polished UI.

#### Why This Change?
Previously, cluster deletion was restricted to plugin-based (`dynamic_cluster`) clusters, which limited the ability to manage other cluster types, such as those sourced from `kubeconfig` or `in_cluster`. Expanding single delete to all cluster types provides greater flexibility for users to manage their clusters. Additionally, introducing bulk delete functionality enables users to remove multiple clusters at once, streamlining workflows and improving efficiency in cluster management.

PR #4461, opened 2026-01-27
frontend: Enable cluster deletion in browser

### LINKED ISSUE #3053: Unable to delete cluster
## Describe the bug
Old kubeconfig for docker-desktop got regenerated and it doesn't allow you do add new cluster with same name and no option to remove old cluster
## To Reproduce
Steps to reproduce the bug:
1. [Go to '...']
2. [Click on '....']
3. [Scroll down to '....']
4. [See error]

 - Installation type: [e.g. Linux-Flatpak/Linux-Tarball/Linux-AppImage/Windows//Windows-Chocolatey/Windows-Winget/WSL/Mac/Mac-Homebrew/... or Container-Image/In-Cluster/Helm/...]
 - Headlamp Version: [X.Y.Z]

[Yes (I will propose a PR) / No.]

## Additional Context

alokdangre:
## Summary

This change enables the deletion of clusters when running Headlamp in a browser, provided that dynamic cluster management is enabled in the backend configuration.

Previously, cluster deletion was restricted to the Electron app. 

## Related Issue

Fixes #3053 

## Changes

- Adds isDynamicClusterEnabled to the Redux config state.
- Updates ClusterContextMenu to check this flag in addition to isElectron.
- Ensures the X-HEADLAMP_BACKEND-TOKEN is set correctly in development mode for API requests.

## Steps to Test

1. Go to view cluster
2. click on three dots
3. click delete
4. confirm

https://github.com/user-attachments/assets/a58135aa-ff79-4f99-96f7-bcf4b607b835


alokdangre:
@illume ptal

alokdangre:
@illume  fixed the changes ptal

frontend/src/helpers/getHeadlampAPIHeaders.ts
```diff
@@ -21,8 +21,12 @@
  *
  * The app also sets HEADLAMP_BACKEND_TOKEN in the headlamp-server environment,
  * which the server checks to validate requests containing this same token.
+ *
+ * For development, when running the frontend separately from the backend,
+ * the token can be initialized from the REACT_APP_HEADLAMP_BACKEND_TOKEN
+ * environment variable to authenticate API requests.
  */
-let backendToken: string | null = null;
+let backendToken: string | null = import.meta.env.REACT_APP_HEADLAMP_BACKEND_TOKEN || null;
```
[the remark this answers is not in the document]
illume: I don't think this suggested change is necessary?

frontend/src/components/App/Home/ClusterContextMenu.tsx
```diff
@@ -52,6 +52,7 @@ export default function ClusterContextMenu({ cluster }: ClusterContextMenuProps)
   const [openConfirmDialog, setOpenConfirmDialog] = React.useState<string | null>(null);
   const dialogs = useTypedSelector(state => state.clusterProvider.dialogs);
   const menuItems = useTypedSelector(state => state.clusterProvider.menuItems);
+  const isDynamicClusterEnabled = useTypedSelector(state => state.config.isDynamicClusterEnabled);
```
illume, line 55: I don't see where isDynamicClusterEnabled is set? How does the app get into this mode?

alokdangre: HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS=true this enables it from backend @illume

alokdangre:
> I am not sure we should make this dependent on the dynamic clusters being available because the dynamic clusters don't have much to do with being able to delete a cluster from the kubeconfig. Still we should make sure that we do not offer this option (and especially that we do not effectively remove clusters) when Headlamp has been deployed with kubeconfig as an option. We need to be extremely cautious with this, as we don't want users removing clusters from their company-deployed Headlamp.
> 
> So maybe this means we should have a flag --allow-kubeconfig-removal or something, which would be false by default, but we'd set it to true when developing and when running the app (which would work in headless mode).
> 
> @sniok / @illume , WDYT?
> 
> BTW, there was already this PR in draft from last year: #3339

that's  very nice point, it seems good to me, im implementing the suggested review
also the draft pr was stale from long time, and also the issue was unassigned from that user so i took the issue

joaquimrocha:
Changes look good. I was gonna suggest we use `ALLOW_KUBECONFIG_CHANGES` as the removal alone is too broad, but then I double checked how we are allowing adding new clusters to the kubeconfig just on the condition we are running in Electron, and now I am having doubts if we need a new flag or we just use isElectron.
I think in the past we refrained from removing clusters from the kubeconfig as this could surprise the user, but as Headlamp has become more established, I wonder if we have enough support that it's a major Kubernetes UI and thus it's expected that it changes the kubeconfig.

So to be clear: either we add a new "AllowKubeConfigChanges" flag, or we do not any new flags and allow deleting clusters from the kubeconfig files only when isElectron() is true.

Any thoughts @sniok / @illume ?

sniok:
> or we do not any new flags and allow deleting clusters from the kubeconfig files only when isElectron() is true.

doesn't it already work like this? we can delete clusters in the app

alokdangre:
> > or we do not any new flags and allow deleting clusters from the kubeconfig files only when isElectron() is true.
> 
> doesn't it already work like this? we can delete clusters in the app

yes its the logic currently used in app, but i added a flag in this pr, so i will remove the flag from the electron app

alokdangre:
> Thanks for making those changes. It got me thinking about something else...
> 
> Can we enable this flag by default in "headless" mode? See the app/electron/main.ts for headless mode.
> 
> Since this is a local thing, if we can enable this functionality for headless mode it will more closely match the abilities of the app. Since the security context of these two modes are pretty much the same... they are both app mode run by an individual.

done

illume (approved):
Thanks! 🎉 

(I’ll leave this one open to give the others a chance to have a look)

app/electron/main.ts
```diff
@@ -755,6 +755,14 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
   process.env.HEADLAMP_CONFIG_ENABLE_HELM = 'true';
   process.env.HEADLAMP_CONFIG_ENABLE_DYNAMIC_CLUSTERS = 'true';
 
+  // In headless mode the app runs in the browser (not Electron), so the
+  // frontend's isElectron() check won't apply. We explicitly enable
+  // kubeconfig removal here because headless shares the same single-user
+  // security context as the desktop app.
+  if (isHeadlessMode) {
```
joaquimrocha: I think you can always set this as true since we'll always want to allow deletion as this logic only runs if it's the app.

backend/cmd/stateless.go
```diff
@@ -178,7 +178,11 @@ func (c *HeadlampConfig) parseKubeConfig(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	clientConfig := clientConfig{contexts, c.EnableDynamicClusters}
+	clientConfig := clientConfig{
+		Clusters:                contexts,
+		IsDynamicClusterEnabled: c.EnableDynamicClusters,
+		AllowKubeconfigRemoval:  c.AllowKubeconfigRemoval,
```
joaquimrocha: I thought we had agreed to make this AllowKubeconfigChanges or similar.

backend/pkg/config/config.go
```diff
@@ -433,6 +434,8 @@ func addGeneralFlags(f *flag.FlagSet) {
 	f.Bool("insecure-ssl", false, "Accept/Ignore all server SSL certificates")
 	f.String("log-level", "info", "Set backend log verbosity. Options: debug, info (default), warn, error")
 	f.Bool("enable-dynamic-clusters", false, "Enable dynamic clusters, which stores stateless clusters in the frontend.")
+	f.Bool("allow-kubeconfig-changes", false,
+		"Allow users to modify and remove clusters in their kubeconfig via the UI")
```
joaquimrocha: "Allow Headlamp to make changes to the known kubeconfigs when needed. E.g. to remove a cluster via the UI. May not be recommendable when Headlamp is deployed as a service."

PR #4467, opened 2026-01-27
frontend: Fix broken cluster settings navigation

### LINKED ISSUE #4041: Cluster settings 404
## Describe the bug
When clicking on "Cluster settings" on the login screen, a 404 page is returned.
When reloading the page without changing anything else, the settings are shown.

## To Reproduce
Steps to reproduce the bug:
1. Open the Headlamp page without logging in
2. Click on "Cluster settings"
3. 404 Page is shown
4. reload page
5. settings are shown

## Environment
 - Installation type: In-Cluster via Helm/ArgoCD
 - Headlamp Version: 0.36.0
 - Other: exposed via ingress-nginx

emmanuel-ferdman:
## Summary

The "Cluster settings" button on the token and authentication pages was broken - clicking it would either navigate to the home page or fail to load the settings page properly. The issue was that the link URL included query parameters in the wrong place, which React Router couldn't handle during client-side navigation. This PR fixes the button to properly navigate to the cluster settings page with the correct cluster selected.

## Related Issue

Fixes #4041

## Changes

- Fixed AuthChooser non-error case to use `settingsClusterHomeContext` with proper search prop.
- Fixed AuthChooser error case to pass the cluster parameter that was missing from original fix.
- Fixed Auth token page to use same pattern.
- Updated snapshots to reflect correct URLs.
- Omit empty `?c=` query param when no cluster is set, producing cleaner URLs.

## Steps to Test

1 Navigate to `/c/<cluster-name>/token`
2. Click "Cluster settings" link
3. Verify it navigates to `/settings/cluster?c=<cluster-name>` and loads correctly

## Notes for the Reviewer

This PR is a follow-up to issue #2404 and PR #2405 - the original fix was incomplete (removed cluster parameter entirely instead of passing it correctly).

### DIFF
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -303,7 +303,7 @@ export function PureAuthChooser({
                 <ColorButton onClick={handleTokenAuth}>{t('Use A Token')}</ColorButton>
               </Box>
               <Box m={2} textAlign="center">
-                <Link routeName="settingsCluster" params={{ clusterID: clusterName }}>
+                <Link routeName="settingsClusterHomeContext" search={{ c: clusterName }}>
                   {t('translation|Cluster settings')}
                 </Link>
               </Box>
@@ -321,7 +321,7 @@ export function PureAuthChooser({
                         errorMessage: error!.message,
                       })}
                 </Empty>
-                <Link routeName="settingsClusterHomeContext">
+                <Link routeName="settingsClusterHomeContext" search={{ c: clusterName }}>
                   {t('translation|Cluster settings')}
                 </Link>
               </Box>
--- a/frontend/src/components/account/Auth.tsx
+++ b/frontend/src/components/account/Auth.tsx
@@ -168,7 +168,10 @@ export function PureAuthToken({
           <div style={{ flex: '1 0 0' }}></div>
         </DialogActions>
         <Box overflow="hidden" textAlign="center">
-          <HeadlampLink routeName="settingsCluster" params={{ clusterID: cluster || '' }}>
+          <HeadlampLink
+            routeName="settingsClusterHomeContext"
+            search={cluster ? { c: cluster } : ''}
+          >
             {t('translation|Cluster settings')}
           </HeadlampLink>
         </Box>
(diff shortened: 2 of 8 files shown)


frontend/src/components/authchooser/index.tsx
```diff
@@ -303,7 +303,7 @@ export function PureAuthChooser({
                 <ColorButton onClick={handleTokenAuth}>{t('Use A Token')}</ColorButton>
               </Box>
               <Box m={2} textAlign="center">
-                <Link routeName="settingsCluster" params={{ clusterID: clusterName }}>
+                <Link routeName="settingsClusterHomeContext" search={`c=${clusterName}`}>
```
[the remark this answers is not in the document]
emmanuel-ferdman: This pattern is consistent with the existing codebase - see `createRouteURL.tsx:98`, `SettingsCluster.tsx:138`, `ClusterSelector.tsx:38` which all use `?c=${cluster}` without encoding. If URL encoding is needed, it should be addressed codebase-wide in a separate PR.

illume:
@emmanuel-ferdman thanks. Can you please squash the commits into one?

You can add the two parts into the context so it's easier for people to see where the change is:
> frontend: auth: authchooser: Fix broken cluster settings navigation

For this repo we do not use signed-off, so please remove that from the commit message?

frontend/src/components/authchooser/index.tsx
illume: I'm pretty sure this works:
```
search={{ c: clusterName }
```

This way it avoids enoding/injection issues. Can you please try that?

emmanuel-ferdman: @illume Thanks for the review! Tried it - doesn't work. The `Link` component passes search directly to React Router, which expects a [string](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/components/common/Link.tsx#L128). Passing an object causes `search.charAt is not a function` error. We'd need to update [`search`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/components/common/Link.tsx#L43) to accept objects and convert via `URLSearchParams`. Want me to add that change to this PR? It will probably require to do something like:
```typescript
+ const searchString = typeof search === 'object' ? new URLSearchParams(search).toString() : search;
...
   return (
     <MuiLink
       component={RouterLink}
       to={{
         pathname: createRouteURL(routeName, params),
-        search,
+        search: searchString,
         state,
       }}
       {...otherProps}
 ...
 ```

illume: Ok, cool! Yes, please add it to this PR. In a separate commit first if possible?

btw. please rebase to remove the merge commit?

emmanuel-ferdman: @illume sure, fixed 😄

PR #4473, opened 2026-01-28
Frontend : Fixed Activity-Component responsiveness

### LINKED ISSUE #4472: Frontend : Activity Component Text Overflow on Small Screens
## Describe the bug
The Activity component had responsiveness issues on small screens where text content was overflowing instead of being properly truncated with ellipsis.

## To Reproduce
Steps to reproduce the bug:
- Open multiple pods , minimize them , then click on overview button in the bottom panel

[Yes (I will propose a PR)]

-Screenshots 

Image

greedy-wudpeckr:
## Summary

This PR fixes Activity Component Text Overflow on Small Screens

## Related Issue

Fixes #4472 

Before : 
Screenshot 2026-01-28 at 11 23 58 PM

After : 
Screenshot 2026-01-28 at 11 26 31 PM

## Notes for the Reviewer

### DIFF
--- a/frontend/src/components/activity/Activity.tsx
+++ b/frontend/src/components/activity/Activity.tsx
@@ -244,10 +244,19 @@ export function SingleActivityRenderer({
       const rows = 5;
       const gapPx = 20;
       const box = container.getBoundingClientRect();
-      const x = (box.width / cols) * (index % 3) + gapPx;
-      const y = (box.height / rows) * Math.floor(index / 3) + gapPx;
-      const width = box.width / cols - gapPx * (cols - 2);
-      const height = box.height / rows - gapPx * (rows - 2);
+      const row = Math.floor(index / 3);
+      const col = index % 3;
+
+      // Calculate cell dimensions accounting for gaps
+      const cellWidth = (box.width - gapPx * (cols + 1)) / cols;
+      const cellHeight = Math.max(150, (box.height - gapPx * (rows + 1)) / rows);
+
+      // Position with gaps
+      const x = gapPx + col * (cellWidth + gapPx);
+      const y = gapPx + row * (cellHeight + gapPx);
+
+      const width = cellWidth;
+      const height = cellHeight;
 
       oldTranslation = activity.style.transform ?? '';
       oldHeight = activity.style.height;
@@ -390,16 +399,40 @@ export function SingleActivityRenderer({
           {isOverview && (
             <Box
               sx={{
-                fontSize: '18px',
+                fontSize: { xs: '1rem', md: '0.875rem' },
                 display: 'flex',
+                flexDirection: 'column',
                 alignItems: 'center',
                 justifyContent: 'center',
-                padding: 1,
-                gap: 1,
+                padding: 2,
+                gap: 1.5,
                 height: '100%',
+                minHeight: '150px',
               }}
             >
-              <Box sx={{ width: '48px', height: '48px', flexShrink: 0 }}>{icon}</Box> {title}
+              <Box
+                sx={{
+                  width: { xs: '32px', sm: '40px', md: '48px' },
+                  height: { xs: '32px', sm: '40px', md: '48px' },
+                  flexShrink: 0,
+                }}
+              >
+                {icon}
+              </Box>
+              <Box
+                sx={{
+                  minWidth: 0,
+                  flex: 1,
+                  textAlign: 'center',
+                  wordBreak: 'break-word',
+                  display: 'flex',
+                  alignItems: 'center',
+                  justifyContent: 'center',
+                  width: '100%',
+                }}
+              >
+                {title}
+              </Box>
             </Box>
           )}
           <>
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Interesting, thanks!

Can you please check the commit message guideline documentation? Have a look at the `git log --oneline` commit messages for examples.

The text looks a bit small. With the text overflow at least people can read all the text. It might be increasing the size of the boxes is better? If the text is cut off, there should be some way for users to read and copy the full string IMHO.

greedy-wudpeckr:
@illume wt are your opinions on this ? 

Screenshot 2026-01-29 at 2 26 50 AM

added hover tooltips ?

illume:
hrmm...

- the text in the bubble should be at a similar size to the other text, so it's readable. Generally you want the body text to be the same size as other body text. Most of the text in headlamp defaults to a size that most people can read. It's also an a11y standard to have minimum sizes.
- cutting off the text makes it less readable to the user than the original. Having a hover makes it readable to people who hover, but just looking at it without having to go to the extra effort of hovering is harder


Could you try to increase the size of the box the text lives in instead? I think vertical height increase would perhaps be better, because looking at the original bumping up the vertical size a bit would allow them both to fit.

greedy-wudpeckr:
> hrmm...
> 
> * the text in the bubble should be at a similar size to the other text, so it's readable. Generally you want the body text to be the same size as other body text. Most of the text in headlamp defaults to a size that most people can read. It's also an a11y standard to have minimum sizes.
> * cutting off the text makes it less readable to the user than the original. Having a hover makes it readable to people who hover, but just looking at it without having to go to the extra effort of hovering is harder
> 
> Could you try to increase the size of the box the text lives in instead? I think vertical height increase would perhaps be better, because looking at the original bumping up the vertical size a bit would allow them both to fit.

@illume  any suggestion regarding the taskbar ? does it need to show the whole content or just oky with the one in the image ?

greedy-wudpeckr:
Screenshot 2026-02-01 at 2 01 36 AM

@illume Does this works ?

illume:
@greedy-wudpeckr I think the text should be the same size as in the "Before" screenshot.

~~Something to try: Maybe the Icon could be top left? Then the text could continue on the same line making good use of the space. It might cut off the text though and be more readable with the icon by itself.~~ Edit: I think how you have the icon on the top row by itself means the text doesn't wrap as much making it more readable.

For the task bar, I think it's important to be able to read it without having to hover or such.
Maybe add one more row in height? Then perhaps it won't be so wide, and the full text could wrap.

greedy-wudpeckr:
> @greedy-wudpeckr I think the text should be the same size as in the "Before" screenshot.
> 
> ~Something to try: Maybe the Icon could be top left? Then the text could continue on the same line making good use of the space. It might cut off the text though and be more readable with the icon by itself.~ Edit: I think how you have the icon on the top row by itself means the text doesn't wrap as much making it more readable.
> 
> For the task bar, I think it's important to be able to read it without having to hover or such. Maybe add one more row in height? Then perhaps it won't be so wide, and the full text could wrap.

Screenshot 2026-02-01 at 6 11 38 PM

i think we are sorted with taskbar , and for items do we need change in font size or this is okay ?

illume:
Try another variation of the task bar? Where the icons are on the same row as the cluster name? "docker-desktop". I expect this will allow the name of it have more space and to flow more.

The text is still smaller than other text I see in the background, and smaller than before. I believe the smallest text shown is the minimum size we should be showing text, according to accessibility and usability guidelines.

greedy-wudpeckr:
> Try another variation of the task bar? Where the icons are on the same row as the cluster name? "docker-desktop". I expect this will allow the name of it have more space and to flow more.
> 
> The text is still smaller than other text I see in the background, and smaller than before. I believe the smallest text shown is the minimum size we should be showing text, according to accessibility and usability guidelines.

Screenshot 2026-02-01 at 11 00 17 PM

increased the font size little bit , and that task bar icon things , is this okay now ?

illume:
Nice one. Looking better. We can await feedback from others :)

In order to test this with different content, it can be good to create different storybook states. 

For example:
- very long cluster name
- very long pod name
- very long resource name (Pod is a very short resource name SomeAreVeryLongAndMightLookDifferent)
- very short cluster name
- very short pod name
- some real world example values found on the internet

Then it can be quicker to see if your change works with all the states by quickly flicking through the states in storybook.

greedy-wudpeckr:
https://github.com/user-attachments/assets/e3cf9516-07ec-4863-803f-56c8004736b6

PTAL @illume

illume:
@greedy-wudpeckr good work adding all those states.

With all the states you can look through them to see how the implementation works, and fix any design inconsistencies.

For example, you can see that on some of them the left icon is in the middle, and on others it's at the top. (same with the cluster name).

I think next is to try to fix these inconsistencies, and try to make them render well in all the different states. When you're ready, I think it's a good time to ask for some design feedback on this in the headlamp slack channel.

illume (changes requested):
Thanks for these changes.


<details>

Good examples:

</details>

skoeva (commented):
this PR is stale, closing this in favor of #7146. thanks for contributing!

PR #4484, opened 2026-01-29
backend: cmd: pkg: Add flag for non-in cluster OIDC authentication

### LINKED ISSUE #4481: Allow OIDC authentication for kubeconfig clusters when inCluster mode is false
When running Headlamp with inCluster: false and kubeconfig clusters that use OIDC auth-provider with oidc config. The OIDC login flow works correctly and refresh tokens are cached. However when proxy does requests to kubernetes API the cached token is not used and the headlamp forces to reauthenticate. Headlamp with inCluster mode set to false expects tokens to be stored in the kubeconfig file which is not viable for unique user tokens. 

The solution is already enabled for [inCluster mode set to true](https://github.com/kubernetes-sigs/headlamp/blob/main/backend/cmd/headlamp.go#L1306), i would like to be able to use the same thing. If removing blocking condition from if statement is not viable it would be nice to have flag (for example --OIDC-USE-COOKIE) to be able to enable this behavior.

* Users with kubeconfigs that contains OIDC auth-provider config
* Users that deployed headlamp in k8s, but do not use inCluster mode
* Users who authenticate with OIDC
* Users that use multiple clusters with the same kubeconfig

No

mudit06mah:
## Summary

This PR adds a flag to allow OIDC authentication for kubeconfig clusters when inCluster mode is false.

## Related Issue

Fixes #4481

## Changes

- Added `oidc-allow-context` flag
- Added flag logic to use token from cookie if set
- Modified test name and message

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -72,6 +72,7 @@ import (
 
 type HeadlampConfig struct {
 	*headlampcfg.HeadlampCFG
+	OidcAllowContext          bool
 	oidcClientID              string
 	oidcValidatorClientID     string
 	oidcClientSecret          string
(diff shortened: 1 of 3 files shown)


mudit06mah:
@illume I have opened a draft PR for now since I am not sure about the approach. Can you please review this and share your opinion?

mudit06mah:
@sniok @yolossn Please Review this :)

PR #4495, opened 2026-01-30
headlamp-plugin: Fix package failing on dist subdirectories

### LINKED ISSUE #4471: headlamp-plugin package fails with EISDIR when dist folder contains subdirectories
### Description
The `headlamp-plugin package` command fails when the `dist/` folder contains subdirectories (e.g., `locales/`). This commonly occurs when using i18n support, as the `copyExtraDistFiles` function auto-copies the `locales` directory into `dist/`.

### Error Message
`Error: EISDIR: illegal operation on a directory, copyfile '/path/to/plugin/dist/locales' -> '/tmp/headlamp-plugin-xxx/plugin-name/locales'`

### Root Cause
The `extractPackage` and `extractFolderOfPackages` functions in `plugins/headlamp-plugin/bin/headlamp-plugin.js` use `fs.copyFileSync()` to copy all items from the `dist/` directory. However, `copyFileSync` only works on files, not directories.

#### Affected code (extractPackage, lines 230-236):
```js
const files = fs.readdirSync(distPath);
files.forEach(file => {
  const srcFile = path.join(distPath, file);
  const destFile = path.join(plugName, file);
  console.log(`Copying "${srcFile}" to "${destFile}".`);
  fs.copyFileSync(srcFile, destFile);  // Fails on directories
});
```
The same issue exists in `extractFolderOfPackages` (lines 262-268).

skoeva:
This change fixes `extractPackage` and `extractFolderOfPackages` to handle directories in the dist folder.

Fixes #4471 

### Testing
- Create a new plugin with `npx @kinvolk/headlamp-plugin create test-plugin`
- Add a `locales/en/translation.json` file with `{}`
- Add `"headlamp": { "i18n": ["en"] }` to `package.json`
- Run `npm run build && npm run package`
- Without fix: fails with `EISDIR` error
- With fix: creates tarball successfully

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -232,7 +232,11 @@ function extract(pluginPackagesPath, outputPlugins, logSteps = true) {
         const srcFile = path.join(distPath, file);
         const destFile = path.join(plugName, file);
         console.log(`Copying "${srcFile}" to "${destFile}".`);
-        fs.copyFileSync(srcFile, destFile);
+        if (fs.statSync(srcFile).isDirectory()) {
+          fs.copySync(srcFile, destFile);
+        } else {
+          fs.copyFileSync(srcFile, destFile);
+        }
       });
 
       const inputPackageJson = path.join(pluginPackagesPath, 'package.json');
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

PR #4497, opened 2026-01-31
frontend: NameValueTable: Fix spacing issue

alokdangre:
## Summary

Adjust the label column width on medium/large screens to reduce excessive whitespace between label and value.

## Related Issue

Fixes #4494 

## Changes

- Updated NameValueTable.tsx to utilize md and lg breakpoints for better responsiveness.
- Fixed excessive whitespace between "Name" (label) and "Value" columns on desktop screens, resolving the "hard to scan" issue.
- Refactored the Grid column widths to shift from a fixed 33% label width to a tighter ~16% on larger screens.

## Steps to Test

1. Navigate to any resource details page (e.g., Network > Services > select a service) OR open Storybook to the NameValueTable component.
2. Check the metadata section at the top (where it lists Name, Namespace, Labels, etc.).
3. Resize your browser window to a desktop width (>900px).
4. Observe that the space between the label (left) and the value (right) is now reduced, making it easier to visually scan the information.

## Screenshots (if applicable)
Before --> 
Screenshot 2026-01-31 132408
After -->
Screenshot 2026-01-31 132212

PR #4532, opened 2026-02-03
frontend: NameValueTable: Fix alignment in narrow layouts

mahmoudmagdy1-1:
## Summary

This PR fixes #4522 by fixing text collision and layout misalignment in `NameValueTable` when used in narrow containers like in side drawers (eg. pods) also fixes it in the settings page and other places.

## Related Issue

Fixes #4522

## Changes

- NameValueTable.tsx:
  - Changed `minWidth` from `'10rem'` to `0` to allow shrinking without issues on smaller layouts
  - Added `wordBreak: 'break-word'` to labels to prevent text collision and overflow

## Steps to Test

1. Open Headlamp and go to the "Pods" list.
2. Click on any Pod to open the details drawer (split-right view).
3. Observe the "Labels", "Annotations", and other fields in the metadata section
4. Resize the browser window to a narrower width so the drawer becomes smaller.
5. Verify that :
   - The "Name" and "Value" columns stay aligned (borders line up).
   - Long label names wrap correctly instead of colliding with the value.
   - The drawer content does not overflow horizontally

## Screenshots (if applicable)
Pods BEFORE
Pods Before

Pods AFTER
Pods After

PR #4534, opened 2026-02-03
frontend: namevaluetable: fix the key value pair structure

### LINKED ISSUE #4522: UI Bug: Side drawer layout breaks
## Describe the bug
The right-side drawer layout is broken when viewing Pod details. Several information boxes/cards are misaligned or overflow the drawer width, making the content hard to read. The expected behavior is a properly aligned, readable side drawer layout.

Image

## To Reproduce
Steps to reproduce the bug:
1. Open Headlamp
2. Navigate to **Pods**
3. Click on any Pod 
4. Observe the broken layout in the side drawer

- Installation type: Linux (built from source)
- Headlamp Version: latest `main`
- Other: Kubernetes via minikube, Dark mode enabled

not yet tried

## Additional Context
This issue is consistently reproducible for Pod detail views. A screenshot is attached showing the broken layout in the side drawer. Not sure if this is a regression.

ChayanDass:
## Summary

This PR rolls back the changes introduced in  **PR #4497**  that caused the **NameValueTable key–value rows to break when rendered in Headlamp’s side window**.  
The layout continues to work correctly in full-screen details view, but breaks specifically in the **side details panel**, which is a primary interaction pattern in Headlamp.

---

## Related Issue

Regression introduced by  PR  #4497 

Fixes  #4522 

---

## Changes

- Rolled back NameValueTable layout changes from PR #4497
- Restored correct key–value row alignment in the **side window**
- Ensured consistent rendering between **side window** and **full-screen** views
- Avoided further layout regressions until a proper fix is applied

---

## Steps to Test

1. Open Headlamp
2. Select any resource (Pod, Deployment, etc.)
3. View the resource details in the **side window**
4. Verify that the key and value appear on the **same row**
5. Open the same resource in **full-screen details view**
6. Confirm the layout is correct in both views

---

Before->
Screenshot from 2026-02-02 14-44-07
After->
image

### DIFF
--- a/frontend/src/components/common/NameValueTable/NameValueTable.tsx
+++ b/frontend/src/components/common/NameValueTable/NameValueTable.tsx
@@ -123,8 +123,7 @@ export default function NameValueTable(props: NameValueTableProps) {
               item
               key={i}
               xs={12}
-              sm={hideValueGridItem ? 12 : 3}
-              md={hideValueGridItem ? 12 : 2}
+              sm={hideValueGridItem ? 12 : 4}
               component="dt"
               className={className}
               id={nameID}
@@ -140,7 +139,7 @@ export default function NameValueTable(props: NameValueTableProps) {
                   fontSize: '1rem',
                   textAlign: 'left',
                   maxWidth: '100%',
-                  minWidth: '8rem',
+                  minWidth: '10rem',
                   verticalAlign: 'top',
                   color: theme.palette.text.secondary,
                   borderBottom:
@@ -169,8 +168,7 @@ export default function NameValueTable(props: NameValueTableProps) {
                 item
                 key={i + 10000}
                 xs={12}
-                sm={valueFullRow ? 12 : 9}
-                md={valueFullRow ? 12 : 10}
+                sm={valueFullRow ? 12 : 8}
                 component="dd"
                 sx={theme => {
                   const extra = withHighlightStyle
--- a/frontend/src/components/App/Settings/__snapshots__/ClusterNameEditor.Default.stories.storyshot
+++ b/frontend/src/components/App/Settings/__snapshots__/ClusterNameEditor.Default.stories.storyshot
@@ -4,7 +4,7 @@
       class="MuiGrid-root MuiGrid-container css-kxuems-MuiGrid-root"
     >
       <dt
-        class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-3 MuiGrid-grid-md-2 css-1779c3p-MuiGrid-root"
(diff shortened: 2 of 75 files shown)


ChayanDass:
The previous PR used `md` breakpoints, but the thing is, screen size doesn’t change. Only the drawer/window width does.
Because MUI breakpoints are viewport-based, the layout still uses `md/lg ` in the side drawer, which causes the key–value rows to break.

suggestion: Using a dynamic drawer/container width calculation would fix this, and I’ve also seen a related issue with a resizable drawer where this approach would help as well.

alokdangre:
@ChayanDass  thanks for addressing it,
It was my mistake, but in your pr, which i reviewed it again increase the space back
Screenshot 2026-01-31 132408

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -123,8 +123,7 @@ export default function NameValueTable(props: NameValueTableProps) {
               item
               key={i}
               xs={12}
-              sm={hideValueGridItem ? 12 : 3}
-              md={hideValueGridItem ? 12 : 2}
```
alokdangre, line 127: just remove this line 127

frontend/src/components/common/NameValueTable/NameValueTable.tsx
```diff
@@ -169,8 +168,7 @@ export default function NameValueTable(props: NameValueTableProps) {
                 item
                 key={i + 10000}
                 xs={12}
-                sm={valueFullRow ? 12 : 9}
-                md={valueFullRow ? 12 : 10}
```
alokdangre, line 173: and remove this line 173

alokdangre:
https://github.com/user-attachments/assets/2d8ced76-12c1-4be9-a1f2-5d538b26eabd

my local testing

sniok:
Thanks for opening a PR for this.

Something that I've wanted to change about this component for a while is to use CSS grid instead of the mui Grid component, that would allow us to use `min-content` for the labels in the first column and greatly simplify logic here. Possibly also using container query to properly handle resizing would be cool too. Let me know if you'd be interested in implementing something like that

ChayanDass:
HI @sniok, and of course, I like to implement this!

illume:
@ChayanDass it looks like someone else worked on a PR for this at the same time. Can you please see this comment and coordinate on which approach to take?
- https://github.com/kubernetes-sigs/headlamp/issues/4522#issuecomment-3860360882

sniok:
#4532 seems to have fixed the same issue. We can iterate on this in the future, for now I'm removing it from 0.40.1 milestone

PR #4571, opened 2026-02-05
charts: Bump Headlamp version to v0.40.0

illume:


illume:
Let's wait for the container image to be published before we merge this.

Issue #3854, opened 2025-09-01
backend: k8cache: future work


illume:
There were a few items after the project was done that need further looking into.

- https://github.com/kubernetes-sigs/headlamp/pull/3740

### backend/pkg/k8cache/cacheInvalidation_test.go

https://github.com/kubernetes-sigs/headlamp/pull/3740/files#r2307849205

>Only Add is used. TestRunInformerToWatch tests Add events but not Update/Delete handlers.
> Need to add some tests for Update and Delete as well.

### backend/pkg/k8cache/cacheInvalidation.go

https://github.com/kubernetes-sigs/headlamp/pull/3740/files#r2307821464

> discoveryClient.ServerPreferredResources has a strange API where it can return an error, but one error case is "there are some results returned, some with errors".

https://github.com/kubernetes-sigs/headlamp/pull/3740/files#r2307828847

> If a kube context is deleted or reloaded, we may need a public StopWatcher(contextKey) to call cancel and delete from watcherRegistry.

### backend/cmd/server.go

https://github.com/kubernetes-sigs/headlamp/pull/3740/files#r2294146727

> This returns an error, but it's ignored?

### backend/cmd/headlamp_test.go

https://github.com/kubernetes-sigs/headlamp/pull/3740/files#r2308041318

> We probably need integration tests with a real K8s server. These create a fake server.

iamundefinedguy:
@illume also that one for moving the items of HeadlampConfig into headlampConfig pkg, so that we can move CacheMiddleware to k8cache pkg

iamundefinedguy:
can I still raise pr ?

illume:
@upsaurav12 hi, sure if you want to.

I added sub issues for the remaining work. Someone already started on some of the work in the PR
- #4503

iamundefinedguy:
@illume I wanted to add pagination/search but I wasn't able to do it. can i add it here. 
for pagination things to work i need to understand the frontend part of the headlamp. It is a little bit tricky.

PR #4617, opened 2026-02-06
feat: pagination

iamundefinedguy:
## Summary
This PR add pagination implementation in server side to make sure that the headlamp's backend don't send full k8s response to client. rather it will send the page data which the user is currently viewing. 

for example let's say user is watching secrets list having 200 secrets so by default the list size would be 151.8KB. 
now let's assume per page there are 15 secrets so now the server will send only 15 pages (i.e 7.15KB). 

## Related Issue
Ref #3854  

## Steps to Test

1. To test you would need to switch the branch to ``pagination-with-cache``.
2. in config.go there is a flag named as ``cache-enabled``, by default it is false switch to true. 
3. after that run ``make backend``. 
4. make sure you have tools like POSTMAN, if not u can use VSCODE extension ``THUNDERCLIENT``. 
5. make sure that in your cluster you have number of secrets > 100 or more would be good fit. 
6. in POSTMAN, add this URL http://localhost:4466/clusters/<your-cluster-name>/api/v1/<your-resource-name>?limit=15(or 20 25), according to you.
7. Hence you won't see full list. 

### default
image

### paginated response: 
image
image

## Notes for the Reviewer


iamundefinedguy:
@illume raising new PR for this because previous PR became messy. 
now working on implementing pagination on server side, and making pagination full functional

vyncent-t (commented):
Greetings, is this feature still being worked on? If so, when this is review, ready please be sure to clean up the commit message history and rebase with the current main. 👍

iamundefinedguy:
@vyncent-t thanks for the reply I will push the commit now. was waiting for a bug fix in frontend side so that I can test the implementation

PR #4628, opened 2026-02-07
frontend: Activity: Ensure complementary landmark has a valid label

### LINKED ISSUE #4606: **Activity**: `frontend/src/components/Activity/Activity.stories.tsx` - Add unique aria-label to complementary landmark to distinguish from other complementary landmarks

mastermaxx03:
## Summary
Adds unique `aria-label` attributes to the `Activity` component landmarks to fix accessibility issue #4606. This ensures that when multiple activities are open (e.g., split-view), screen readers can distinguish between them.

## Related Issue
Fixes #4606

## Changes
- **Activity.tsx**: Added `aria-label` to the `role="complementary"` container, using the activity title or a translated fallback.
- **Activity.stories.tsx**: Updated the `Basic` story and helper to provide unique titles for multiple activities.
- **Snapshots**: Updated the `Activity` storyshots to reflect the new accessibility attributes.

## Steps to Test
1. Run Storybook: `npm run storybook`.
2. Navigate to `Activity` -> `Basic`.
3. Inspect the two sidebar containers and verify they have unique `aria-label` values ("Activity 1" and "Activity 2").

## Notes for the Reviewer
N/A

### DIFF
--- a/frontend/src/components/activity/Activity.tsx
+++ b/frontend/src/components/activity/Activity.tsx
@@ -328,6 +328,7 @@ export function SingleActivityRenderer({
     <ActivityContext.Provider value={activity}>
       <Box
         role="complementary"
+        aria-label={typeof title === 'string' ? title : undefined}
         sx={{
           display: minimized && !isOverview ? 'none' : undefined,
           gridColumn: '2 / 3',
--- a/frontend/src/components/activity/Activity.stories.tsx
+++ b/frontend/src/components/activity/Activity.stories.tsx
@@ -64,6 +64,7 @@ const makeActivity = (activity: Partial<Activity>): Activity => ({
   id: 'id',
   location: 'window',
   content: 'Activity Content',
+  title: activity.title,
   ...activity,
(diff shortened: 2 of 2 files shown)


frontend/src/components/activity/Activity.tsx
```diff
@@ -328,6 +328,7 @@ export function SingleActivityRenderer({
     <ActivityContext.Provider value={activity}>
       <Box
         role="complementary"
+        aria-label={typeof title === 'string' && title.trim() ? title.trim() : t('Activity')}
```
illume: I'm not sure if we should be having an Activity as fallback here. Or if instead we can require the title.

Can you please change it to just use title with no fallback for now? I will ask for some opinions of others.

mastermaxx03: okay!

illume: This is done, marking as resolved. Thanks

PR #4641, opened 2026-02-08
docs: add warning about API server OIDC configuration

### LINKED ISSUE #4618: docs: oidc needs some explanation of what needs to be configured on the API server
Have had a couple of folks come to slack trying to debug oidc only to not have setup their clusters to support OIDC.  The oidc docs page doesn't mention this - https://headlamp.dev/docs/latest/installation/in-cluster/oidc/#token-validation-overrides

## To Reproduce
Steps to reproduce the bug:
1. Configure headlamp incluster for oidc
2. don't configure the api server for oidc

 - Installation type: incluster
 - Headlamp Version: [X.Y.Z]

Yes, can submit a PR

## Additional Context

zyzzmohit:
## Summary
This PR updates the OIDC installation documentation to explicitly state that the Kubernetes API server **must** be configured to accept OIDC tokens. This addresses confusion where users configure Headlamp for OIDC but forget the underlying cluster configuration.

## Related Issue
Fixes #4618

## Changes
- Added an "Important" warning note to `docs/installation/in-cluster/oidc.md` linking to the official Kubernetes OIDC documentation.

### DIFF
--- a/docs/installation/in-cluster/oidc.md
+++ b/docs/installation/in-cluster/oidc.md
@@ -5,6 +5,8 @@ sidebar_label: OIDC
 
 Headlamp supports OIDC for cluster users to effortlessly log in using a "Sign in" button.
 
+> **⚠️ Important:** Headlamp performs the OIDC login flow and obtains a token, but it then calls Kubernetes APIs on the user's behalf — so the API server **must** be configured to accept and validate that token. The OIDC **issuer, client ID, and claims** configured in your cluster must match your Headlamp OIDC configuration. If the API server is not configured, Headlamp login may succeed but subsequent API calls will be rejected with **401/403 errors**. See the [Kubernetes documentation on configuring the API server for OIDC](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server) for details. For managed Kubernetes services (EKS, AKS, GKE), refer to your cloud provider's documentation for OIDC configuration.
+
 ![screenshot the login dialog for a cluster](./oidc_button.png)
 
 To use OIDC, Headlamp needs to know how to configure it, so you have to provide the following OIDC-related arguments to Headlamp from your OIDC provider:
@@ -100,7 +102,6 @@ For quick reference if you are already familiar with setting up Entra ID,
 - Set `--oidc-validator-idp-issuer-url` to `https://sts.windows.net/<Your Directory (tenant) ID>/`
 - Set `-oidc-validator-client-id` to `6dae42f8-4368-4678-94ff-3960e28e3630`
 - Set `-oidc-use-access-token=true`
-
 
 ### Example: OIDC with Dex
 


zyzzmohit:
@illume @sniok 
Friendly bump! 👋 Just checking in — this is a small docs-only change (2 lines) adding an important warning about API server OIDC configuration, which addresses the confusion reported in #4618. Would appreciate a review when you get a chance. Thanks :)

illume:
Maybe what you have is ok and fine enough. But I’m a bit unclear about the conditions that bring people here. Maybe you can help me understand?

How will people find this? What happens if it’s not? Maybe we can the error behavior here for folks to search for it.

I’m also wondering if those docs are adequate for folks configuring a cloud k8s instance? Or is this not something they will encounter?


Not like below(the details are probably wrong), but I think add the error condition they encounter to what you have and a note how to do this on a cloud provider is cloud provider specific (if it’s relevant at all to them?)

> **⚠️ Important:** For Headlamp’s OIDC authentication to work, your Kubernetes API server **must** be configured to accept and validate OIDC tokens (for example by setting `--oidc-issuer-url`, `--oidc-client-id`, and the username/groups claims). Headlamp performs the login flow and returns a token, but the API server validates that token. If the API server is not configured, Headlamp’s login may succeed but subsequent API calls will be rejected with `401`/`403` errors. See the Kubernetes OIDC docs (link) and your cloud provider’s guidance for details.

Or maybe what you have already is better, and these two things are not needed.

illume:
@mlbiam this PR enough for the issue you reported?

illume:
I wonder if we can add into an error message into headlamp when people encounter this something like: “Have  you configured your api server? If not see: xxx”

mlbiam:
@illume 

1. +1 for an error message that tells the user that the cluster isn't integrated 
2. I'd recommend starting with that headlamp calls kubernetes APIs on the user's behalf, so the token must be accepted by kubernetes. Also include that the issuer and audience must be the same.

zyzzmohit:
Thanks for the feedback @illume and @mlbiam! Great suggestions.

I agree the warning can be more actionable. I'll update it to include:

The specific API server flags (--oidc-issuer-url, --oidc-client-id, etc.)
The error behavior users will see (login succeeds but API calls fail with 401/403)
A note that the issuer and audience must match between Headlamp's config and the API server
A mention that cloud provider OIDC configuration is provider-specific
Something along the lines of what @illume suggested:

⚠️ Important: Headlamp performs the OIDC login flow and obtains a token, but it then calls Kubernetes APIs on the user's behalf — so the API server must be configured to accept and validate that token. The --oidc-issuer-url and --oidc-client-id flags (along with username/groups claim flags) must be set on the API server, and the issuer and audience must match your Headlamp OIDC configuration. If the API server is not configured, Headlamp login may succeed but subsequent API calls will be rejected with 401/403 errors. See the [Kubernetes OIDC docs](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens) for details. For managed Kubernetes services (EKS, AKS, GKE), refer to your cloud provider's documentation for OIDC configuration.

Regarding the in-app error message idea — I think that's a great enhancement and worth tracking separately. I can open a follow-up issue for detecting when OIDC token validation fails at the API server level and surfacing a helpful message like "Your cluster's API server may not be configured to accept OIDC tokens. See: [link]". Would that work, or would you prefer it as part of this PR?

I'll push the docs update shortly!

zyzzmohit:
Thanks for the great feedback @illume and @mlbiam! 🙏

I've updated the warning to be much more actionable. It now includes:

- That Headlamp calls Kubernetes APIs on the user's behalf, so the API server must validate the token
- Specific API server flags (`--oidc-issuer-url`, `--oidc-client-id`, username/groups claims)
- The requirement that **issuer and audience must match** between Headlamp's config and the API server
- The error behavior users will encounter (login succeeds but API calls fail with **401/403 errors**)
- A note about cloud provider-specific OIDC configuration (EKS, AKS, GKE)
- Updated link pointing to the [API server OIDC configuration docs](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)

Regarding the in-app error message idea — I think that's a great enhancement! I'll open a separate issue for detecting when OIDC token validation fails at the API server level and surfacing a helpful message to the user.

Please take a look at the updated changes and let me know if anything else needs tweaking!

zyzzmohit:
I've updated the warning to be much more actionable. It now includes:

That Headlamp calls Kubernetes APIs on the user's behalf, so the API server must validate the token
Specific API server flags (--oidc-issuer-url, --oidc-client-id, username/groups claims)
The requirement that issuer and audience must match between Headlamp's config and the API server
The error behavior users will encounter (login succeeds but API calls fail with 401/403 errors)
A note about cloud provider-specific OIDC configuration (EKS, AKS, GKE)
Updated link pointing to the [API server OIDC configuration docs](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server)

Please take a look at the updated changes and let me know if anything else needs tweaking :)

mlbiam:
@zyzzmohit looks great.  nitpick: I'd remove

> The `--oidc-issuer-url` and `--oidc-client-id` flags (along with username/groups claim flags) must be set on the API server, 

and just say that the issuer, client id, and claims must match.  Nearly every commercial distro that supports OIDC has its own custom mechanism for configuration (even so far as to not support direct API server integration but instead rely on their own token validation webhook).  Even with vanilla kube like kubeadm or even talos, with structured authentication now being GA there's a push to move off of the API server flags and onto the structured authentication spec.

zyzzmohit:
> @zyzzmohit looks great. nitpick: I'd remove
> 
> > The `--oidc-issuer-url` and `--oidc-client-id` flags (along with username/groups claim flags) must be set on the API server,
> 
> and just say that the issuer, client id, and claims must match. Nearly every commercial distro that supports OIDC has its own custom mechanism for configuration (even so far as to not support direct API server integration but instead rely on their own token validation webhook). Even with vanilla kube like kubeadm or even talos, with structured authentication now being GA there's a push to move off of the API server flags and onto the structured authentication spec.

@mlbiam Good point! Updated — removed the reference to specific API server flags and now it just says the OIDC issuer, client ID, and claims must match. This keeps it distro-agnostic. Please take another look when you get a chance :)

gambtho (commented):
Thanks for the commit - Well-targeted documentation improvement that addresses a real user pain point. Accurately explains the OIDC flow and provides authoritative references. Commit format is correct.

zyzzmohit:
Thanks for the review, @gambtho :)

@illume / @mlbiam — since the text and commit format look good now, could one of you please drop a `/lgtm` and `/approve` when you have a moment so the CI bot can merge this? Thank you

docs/installation/in-cluster/oidc.md
```diff
@@ -101,7 +103,6 @@ For quick reference if you are already familiar with setting up Entra ID,
 - Set `-oidc-validator-client-id` to `6dae42f8-4368-4678-94ff-3960e28e3630`
 - Set `-oidc-use-access-token=true`
 
-
```
skoeva, line 104: this line doesn't need to be removed

skoeva (changes requested):
can you rebase this against main? setting this to draft until it's ready to review

PR #4649, opened 2026-02-08
docs: clarify plugin installation for in-cluster deployments

iashutoshyadav:
This PR clarifies the recommended plugin installation approaches for
In-cluster Headlamp deployments.

It documents InitContainer + shared volume as a recommended option for
declarative and GitOps-based setups, and clarifies the role of the
sidecar-based plugin manager as an alternative supported mechanism.


illume:
>The plugin manager is intended mainly for development/testing.

This is not true.

PR #4651, opened 2026-02-08
make: build: Explicitly set DOCKER_PLATFORM to Linux architectures on…

### LINKED ISSUE #4578: make image doesn't work by default on silicon mac
`make image` does not work by default on a silicon mac.

`DOCKER_PLATFORM=linux/arm64 make image` does work.

Inside the Makefile I see that local is the default. Maybe we need to detect that we are on a arm mac, and if so make linux/arm the default.

Vaishnav-Dhaval:
… macOS to prevent buildx targeting darwin

## Summary
Detected macOS builds now default to a Linux platform (arm64/amd64) instead of `local`. This prevents `docker buildx` from incorrectly targeting `darwin`

## Related Issue

Fixes #4578

## Changes

- Updated `Makefile` to set `DOCKER_PLATFORM` to `linux/arm64` or `linux/amd64` when on macOS.

## Steps to Test
1. Run `make image` on a Mac.
2. Building should proceed using a `linux` platform instead of failing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,17 @@ DOCKER_PLUGINS_IMAGE_NAME ?= plugins
 DOCKER_IMAGE_VERSION ?= $(shell git describe --tags --always --dirty)
 DOCKER_PLATFORM ?= local
 DOCKER_PUSH ?= false
+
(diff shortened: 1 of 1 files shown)


illume:
hello. Thanks for this contribution!!

However, this issue was already assigned to someone else. Please check if someone else has been assigned before doing an issue? It’s also ok to ask them if they are going to finish it if some weeks have past.

PR #4657, opened 2026-02-09
Fix: Auto-detect platform for Apple Silicon Macs

### LINKED ISSUE #4578: make image doesn't work by default on silicon mac

Mandalorian7773:
- Auto-detect Darwin arm64 and set DOCKER_PLATFORM=linux/arm64
- Preserves manual override capability

## Summary

Auto-detect the system architecture and set the appropriate Docker platform:

Apple Silicon (Darwin + arm64) → linux/arm64
All other systems → local

## Related Issue

Fixes #4578

## Changes

Added system detection using uname -s and uname -m
Auto-sets DOCKER_PLATFORM=linux/arm64 for Apple Silicon Macs
Preserves ability to override via environment variable
Updated docker build to docker buildx build --platform to support platform flag

## Steps to Test

- run `make image` on Apple Silicon Mac

Screenshot 2026-02-09 at 2 11 51 PM

## Notes for the Reviewer
Tested on:

- Apple Silicon Mac (M3) - make image works 
- Built image runs successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,22 @@ DOCKER_EXT_REPO ?= docker.io/headlamp
 DOCKER_IMAGE_NAME ?= headlamp
 DOCKER_PLUGINS_IMAGE_NAME ?= plugins
 DOCKER_IMAGE_VERSION ?= $(shell git describe --tags --always --dirty)
-DOCKER_PLATFORM ?= local
+# Detect platform (Windows, macOS, Linux)
+ifeq ($(OS),Windows_NT)
+    DOCKER_PLATFORM ?= local
+else
+    UNAME_S := $(shell uname -s)
+    ifeq ($(UNAME_S),Darwin)
+        UNAME_M := $(shell uname -m)
+        ifeq ($(UNAME_M),arm64)
+            DOCKER_PLATFORM ?= linux/arm64
+        else
+            DOCKER_PLATFORM ?= linux/amd64
+        endif
+    else
+        DOCKER_PLATFORM ?= local
+    endif
+endif
 DOCKER_PUSH ?= false
 EMBED_BINARY_NAME := headlamp_app
 # Get version and app name from app/package.json


Mandalorian7773:
> @Mandalorian7773 please see the failing easycla check?

done @illume

Makefile
```diff
@@ -9,7 +9,16 @@ DOCKER_EXT_REPO ?= docker.io/headlamp
 DOCKER_IMAGE_NAME ?= headlamp
 DOCKER_PLUGINS_IMAGE_NAME ?= plugins
 DOCKER_IMAGE_VERSION ?= $(shell git describe --tags --always --dirty)
-DOCKER_PLATFORM ?= local
+# Auto-detect Apple Silicon and set platform accordingly
+UNAME_M := $(shell uname -m)
```
sniok: does this work on windows?

illume: Good point. You could wrap it like we do further below in a check for windows and only call uname on non windows.

Also we still support Intel Macs… so we’ll need to add that detection too.

Something like this… (untested)


# Default
DOCKER_PLATFORM ?= local

# Detect Windows first (safe in CMD, PowerShell, Git Bash, MSYS)
ifeq ($(OS),Windows_NT)
    # leave DOCKER_PLATFORM=local
else
    # POSIX systems: macOS, Linux, BSD, WSL, etc.
    UNAME_S := $(shell uname -s)

    ifeq ($(UNAME_S),Darwin)
        # macOS detected
        UNAME_M := $(shell uname -m)

        ifeq ($(UNAME_M),arm64)
            DOCKER_PLATFORM := linux/arm64
        else ifeq ($(UNAME_M),x86_64)
            DOCKER_PLATFORM := linux/amd64
        endif
    endif
endif

Mandalorian7773: Hii @illume, I have applied the new changes pls review them, i consolidated nested conditionals into a cleaner structure

illume (changes requested):
Thanks.

Can you please see the contributing guide git commit guidelines? We follow a Linux kernel style message format. See git log Makefile for other examples.

Something like:
> Makefile: Fix make image on macOS
>
> On Mac local resolves to Darwin and does not work when trying to build the container image for Linux. So on macOS we detect the architecture and use linux/arm64 or linux/amd64.

Please squash the commits so there is one atomic commit in the PR?

Mandalorian7773:
> Thanks.
> 
> Can you please see the contributing guide git commit guidelines? We follow a Linux kernel style message format. See git log Makefile for other examples.
> 
> Something like:
> 
> > Makefile: Fix make image on macOS
> > On Mac local resolves to Darwin and does not work when trying to build the container image for Linux. So on macOS we detect the architecture and use linux/arm64 or linux/amd64.
> 
> Please squash the commits so there is one atomic commit in the PR?

Sry about that, i read the guides and, now i have applied the changes @illume

Dockerfile
```diff
@@ -11,10 +11,10 @@ ARG TARGETOS
 ARG TARGETARCH
 ENV GOPATH=/go \
     GOPROXY=https://proxy.golang.org \
-	GO111MODULE=on\
-	CGO_ENABLED=0\ 
-	GOOS=${TARGETOS}\
-	GOARCH=${TARGETARCH}
+    GO111MODULE=on\
```
illume: Sorry, I just noticed these changes. They seem not needed. Can you please remove?

Mandalorian7773: Fixed it @illume . It was just a formatting difference, there was no changes in codes

illume: Thanks for that.

PR #4668, opened 2026-02-10
frontend: Table: Fix the rows-per-page settings persist among views

### LINKED ISSUE #3227: Table: Remember rows per page option
When I am navigating Headlamp and adjusting the number of rows displayed per page,
I want my selected "Rows per page" setting to persist across page reloads and menu navigation,
so I don’t have to repeatedly change it, and when I share my view it shows them the same number of rows.

---

### Discussion

It seems we had this in SimpleTable before, but that functionality got lost.

We do have the settings option... but is it discoverable? Is not the rows per page selector also a setting? 

In some other apps this setting is remembered. You don't need to go to a separate settings page to make sure it remembers your rows per page setting.

Additionally, browsers when reloading remember the items in the form. Our implementation of Table forgets rows per page on page reload.

Should rows per page be in the URL? I think so. Because it will change what is shown, and we want people be able to share what they are seeing by sharing URLs. We didn't have this before, and it would complicate the code.

### Additional context

- Material react table docs on pagination: https://www.material-react-table.com/docs/guides/pagination
- https://github.com/kubernetes-sigs/headlamp/issues/553
- requested in https://github.com/kubernetes-sigs/headlamp/issues/522
- implemented in SimpleTable https://github.com/kubernetes-sigs/headlamp/pull/525
- This was also a popular dashboard feature request: https://github.com/kubernetes/dashboard/issues/1387

ChayanDass:
# Summary

This PR adds persistent rows per page functionality to tables by storing the selected rows per page value in localStorage . Users can now change the number of rows displayed per page, and their preference is automatically saved and restored on page refresh.

## Related Issue

Fixes #3227 

## Changes

- **Added localStorage persistence for rows per page** - When users change the rows per page in any table, the value is automatically saved to localStorage with key `tables_rows_per_page`.
-  remove `useMemo` for fetching `getTablesRowsPerPage`
- I changed the default `perPage` value in the URL to match the `storeRowsPerPageOptions[0] `value, ensuring shared links reflect the actual table state.

## Steps to Test
- Open a cluster resource table (e.g. Workloads → Namespaces).
- Change the rows-per-page value (e.g. 25 or 50).
- Reload the page or switch to another view and come back.
- Confirm the same rows-per-page value is still selected (persisted via localStorage).

Note: We should avoid using an id for `reflectInURL`, as it makes the URL unnecessarily long, for example:
`http://localhost:3000/c/minikube/?headlamp-cluster.overview.events.perPage=25&headlamp-cluster.overview.events.p=3`

### DIFF
--- a/frontend/src/components/common/Table/Table.tsx
+++ b/frontend/src/components/common/Table/Table.tsx
@@ -216,7 +216,7 @@ export default function Table<RowItem extends Record<string, any>>({
   const storeRowsPerPageOptions = useSettings('tableRowsPerPageOptions');
   const rowsPerPageOptions = rowsPerPage || storeRowsPerPageOptions;
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const defaultRowsPerPage = useMemo(() => getTablesRowsPerPage(rowsPerPageOptions[0]), []);
+  const defaultRowsPerPage = getTablesRowsPerPage(rowsPerPageOptions[0]);
   const [pageSize, setPageSize] = useURLState(shouldReflectInURL ? 'perPage' : '', {
     defaultValue: defaultRowsPerPage,
     prefix,
--- a/frontend/src/helpers/tablesRowsPerPage.test.ts
+++ b/frontend/src/helpers/tablesRowsPerPage.test.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { getTablesRowsPerPage, setTablesRowsPerPage } from './tablesRowsPerPage';
+
+describe('getTablesRowsPerPage', () => {
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
+  it('returns the default when localStorage is empty', () => {
+    expect(getTablesRowsPerPage(15)).toBe(15);
+  });
+
+  it('returns the stored value when localStorage holds a valid number', () => {
+    setTablesRowsPerPage(25);
+    expect(getTablesRowsPerPage(15)).toBe(25);
+  });
+
+  it('returns the default when localStorage contains a non-numeric string', () => {
+    localStorage.setItem('tables_rows_per_page', 'not-a-number');
+    expect(getTablesRowsPerPage(15)).toBe(15);
+  });
+
+  it('returns the default when localStorage contains an empty string', () => {
+    localStorage.setItem('tables_rows_per_page', '');
+    expect(getTablesRowsPerPage(15)).toBe(15);
+  });
+
+  it('returns the parsed integer when localStorage contains a partial numeric string like "12abc"', () => {
+    localStorage.setItem('tables_rows_per_page', '12abc');
+    // parseInt('12abc') === 12, which is a valid integer
+    const result = getTablesRowsPerPage(15);
+    expect(result).toBe(12);
+
+    const options = [15, 25, 50];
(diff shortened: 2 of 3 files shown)


ChayanDass:
HI @illume ..

I think we can remove `useSettings` here. `getTablesRowsPerPage` already reads the value from localStorage, and `useSettings` also persists the same value there.
```bash
  const storeRowsPerPageOptions = useSettings('tableRowsPerPageOptions');
  const rowsPerPageOptions = rowsPerPage || storeRowsPerPageOptions;
  const defaultRowsPerPage = getTablesRowsPerPage(rowsPerPageOptions[0]);
```
Also, we can set `reflectInURL` to default true, to set the table state values at the URL,  which makes it shareable, with preserved table format
.

ChayanDass:
I removed unnecessary useMemo usage and now pass reflectInURL from parent components. I also noticed that perPage is omitted from the URL when it matches the local default, but defaults can differ between users(if the perpage value already updated and stored in localstorage). This can cause shared links to render different table views, so it may be better to always include perPage in the URL when URL reflection is enabled.

illume (commented):
If you want, you could improve the git message subject body. By including some of the explanation from the PR description and your comment.

illume (changes requested):
Can you please add some steps to test in the PR description?

ChayanDass:
updated, PTAL!

illume:
nit, you could fix the first letter of the subject after the context to be capitalized. Also add some of the Changes part of the PR description into the git message body, and any Why you can think of that may help people.

> frontend: table: Add row-per-page persist and update url

illume (approved):
Thanks! 🎉

ChayanDass:
updated, Thanks!

joaquimrocha:
> Note: We should avoid using an id for reflectInURL, as it makes the URL unnecessarily long, for example:
http://localhost:3000/c/minikube/?headlamp-cluster.overview.events.perPage=25&headlamp-cluster.overview.events.p=3

This is needed if we have to distinguish between different places (tables) that use the same key (perPage).

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -328,7 +328,7 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
     hideColumns = [],
     filterFunction,
     errorMessage,
-    reflectInURL,
+    reflectInURL = true,
```
joaquimrocha: This is changing the default behavior of a component that is used by plugins. So we shouldn't do this, and AFAIU you don't need this setting to be true by default for the issue you are fixing, right?

joaquimrocha (changes requested):
Please check my comment about changing the default behavior of a plugin-exposed component.

frontend/src/components/common/Resource/ResourceTable.tsx
ChayanDass: yes ! true ,  i have removed that for now! PTAL

gambtho:
Hi @ChayanDass, thanks for this PR! Could you rebase to resolve the merge conflicts with `main`? Let us know if you need any help.

illume (changes requested):
Thanks for this PR.


<details>

</details>


ChayanDass:
https://github.com/kubernetes-sigs/headlamp/pull/4918
Someone already merged his pr for this fix. One thing I can do now is to remove the use of useMemo.  else can close this pr!
Thanks

illume (changes requested):
Thanks for these changes.


<details>

</details>

illume (changes requested):
Thanks for working on this.


frontend/src/components/common/Table/Table.tsx
```diff
@@ -216,7 +216,7 @@ export default function Table<RowItem extends Record<string, any>>({
   const storeRowsPerPageOptions = useSettings('tableRowsPerPageOptions');
   const rowsPerPageOptions = rowsPerPage || storeRowsPerPageOptions;
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  const defaultRowsPerPage = useMemo(() => getTablesRowsPerPage(rowsPerPageOptions[0]), []);
+  const defaultRowsPerPage = getTablesRowsPerPage(rowsPerPageOptions[0]);
   const [pageSize, setPageSize] = useURLState(shouldReflectInURL ? 'perPage' : '', {
     defaultValue: defaultRowsPerPage,
```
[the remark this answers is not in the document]
illume: pushed a fix for this

illume (approved):
🎉 thanks!

Issue #3222, opened 2025-05-11
Support rollout undo, like kubectl rollout undo


illume:
The UI should have a way so that the user can rollback deployments(plus daemonsets and statefulsets)  like kubectl can do.

```
  # Roll back to the previous deployment
  kubectl rollout undo deployment/abc
  
  # Roll back to daemonset revision 3
  kubectl rollout undo daemonset/abc --to-revision=3
  
  # Roll back to the previous deployment with dry-run
  kubectl rollout undo --dry-run=server deployment/abc
```

See [kubectl docs for rollout undo](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_rollout/kubectl_rollout_undo/)

Valid resource types include:

- deployments
- daemonsets
- statefulsets

## Describe the solution you'd like

I should be able to select the rollout I want to undo. Maybe also see the history of rollouts to select 

Since it's a potentially dangerous operation, it should do the same as other dangerous ops... like show cluster actions with a grace time to cancel and such.

Note: this needs more design thought, and discussion before implementation.

## What users will benefit from this feature?

People who want to quickly rollback a deployment (or daemonset/statefulset).

## Additional context

See open kubectl issues for rollout undo: https://github.com/search?q=repo%3Akubernetes%2Fkubectl+undo&type=issues

Also in kubernetes/kubernetes where the feature was originally made: https://github.com/search?q=repo%3Akubernetes%2Fkubernetes+rollout+undo&type=pullrequests

See PR for [Add kubectl rollout undo](https://github.com/kubernetes/kubernetes/pull/19835)

This is a top requested feature on Kubernetes Dashboard: https://github.com/kubernetes/dashboard/issues/2889

> Rolling back to a previous release is a fairly important operation as it allows to remedy critical situations quickly. While the functionality is accessible in kubectl as well, I find an integration into the Dashboard very desirable given that it's easier to use, especially for users that are not that familiar with kubectl.

niladrix719:
@illume Hey can I take this issue

illume:
Hi @niladrix719 sure! 

(I see you found the assign command with the k8s bot.)
Also if you decide not to do it, please let us know?

niladrix719:
Hey, give me two days. I’ll try to do it—if not, I’ll let you know

kallal79:
I'd like to take on this issue. Could you please assign it to me?

kallal79:
Thanks for the feedback!

[...]

PR #4701, opened 2026-02-13
Implement rollback to specific revision and Revision history view

alokdangre:
## Summary

This PR adds a feature to rollback to specific revision and also adds a revision history view in workloads details

## Related Issue

Related #3222  

## Changes

- Refactored rollback logic to support optional toRevision parameter
- Added RevisionHistorySection to workload details pages
- Replaced RollbackButton confirmation with RollbackDialog for revision selection
- Added RevisionInfo type to manage revision history data

## Steps to Test

1. Go to Workloads
2. go to any of these -->Deployments, statefulset, daemonset 
3. click on the selected resource to see details in drawer
4. scroll down to see the revision history
5. click on rollback button
6. see rollback dailog and select any specific revision you want to go back to
7. see the changes implemented

Screenshot 2026-02-13 153527
image

https://github.com/user-attachments/assets/830bab35-37b0-421b-be1b-641fa675089b

## Notes for the Reviewer

- Plzz check the design for any improvement and you r views on it

### DIFF
--- a/frontend/src/lib/k8s/daemonSet.ts
+++ b/frontend/src/lib/k8s/daemonSet.ts
@@ -48,6 +48,7 @@ class DaemonSet extends KubeObject<KubeDaemonSet> {
   static apiName = 'daemonsets';
   static apiVersion = 'apps/v1';
   static isNamespaced = true;
+  private revisionHistoryCache?: { resourceVersion?: string; history: RevisionInfo[] };
 
   get spec() {
     return this.jsonData.spec;
@@ -156,13 +157,6 @@ class DaemonSet extends KubeObject<KubeDaemonSet> {
         .filter(rev => rev.revision > 0)
         .sort((a, b) => b.revision - a.revision);
 
-      if (sortedRevisions.length < 2) {
-        return {
-          success: false,
-          message: 'No previous revision available to rollback to',
-        };
-      }
-
       // Find target revision: specific or previous (second in sorted list)
       let targetRev;
       if (toRevision !== undefined && toRevision > 0) {
@@ -180,6 +174,12 @@ class DaemonSet extends KubeObject<KubeDaemonSet> {
           };
         }
       } else {
+        if (sortedRevisions.length < 2) {
+          return {
+            success: false,
+            message: 'No previous revision available to rollback to',
+          };
+        }
         targetRev = sortedRevisions[1];
       }
 
@@ -211,6 +211,7 @@ class DaemonSet extends KubeObject<KubeDaemonSet> {
       const url = `${apiRoot}/namespaces/${this.getNamespace()}/daemonsets/${this.getName()}`;
 
       await jsonPatch(url, patchOperations, true, { cluster: this.cluster });
+      this.revisionHistoryCache = undefined;
 
       return {
         success: true,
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/Resource/RollbackDialog.tsx
```diff
@@ -0,0 +1,214 @@
...
+    return () => {
+      isActive = false;
+    };
+  }, [open, getRevisionHistory]);
+
+  function handleConfirm() {
+    onConfirm(selectedRevision);
+  }
+
+  const hasMultipleRevisions = revisions.filter(r => !r.isCurrent).length > 0;
```
Kaniikura: suggestion: It looks like this variable specifically checks for the existence of "rollback-able" revisions. The current name might be slightly misleading—would a more specific name make sense here?

```suggestion
  const hasPreviousRevisions = revisions.filter(r => !r.isCurrent).length > 0;
```

frontend/src/components/common/Resource/RollbackButton.tsx
```diff
@@ -137,19 +138,13 @@ export function RollbackButton(props: RollbackButtonProps) {
...
-        description={t(
-          'translation|Are you sure you want to rollback "{{ name }}" to the previous version? This will replace the current pod template with the one from the previous revision.',
-          {
-            name: resource.metadata.name,
-          }
-        )}
-        handleClose={() => setOpenDialog(false)}
+        resourceKind={resourceKind}
+        resourceName={resource.metadata.name}
+        getRevisionHistory={() => resource.getRevisionHistory()}
```
Kaniikura: issue: I noticed the selected revision can reset while the dialog is open, which might lead to an accidental rollback.

The getRevisionHistory prop is an inline arrow function that creates a new reference on every render. Since RollbackDialog's useEffect has getRevisionHistory in its dependency array (line 95 in RollbackDialog.tsx), this could trigger unnecessary re-fetches while the dialog is open if the parent re-renders.

This means a user can pick revision X, then a parent re-render can silently reset selection to the default before confirm.

frontend/src/i18n/locales/zh/translation.json
```diff
@@ -393,13 +393,24 @@
...
   "Restart items": "重启资源",
+  "Revision History": "",
+  "Loading revision history…": "",
+  "Failed to load revision history: {{ error }}": "",
+  "No revision history available.": "",
+  "Revision": "",
+  "Created": "",
+  "Images": "",
+  "Current": "当前",
+  "Previous": "上一页",
```
Kaniikura: suggestion: I am not an expert in Chinese, but based on my research in the Kubernetes Chinese Documentation, the term "Revision" is officially translated as "修订版本".

In the context of revision history, "上一页" (Previous Page) is misleading as it refers to pagination. Replacing it with "上一版" (Previous Version/Revision) or "上一修订版本" is much more accurate.

```suggestion
  "Previous": "上一版",
```

Kaniikura (commented):
Noticed this on Slack and wanted to leave feedback.
Local behavior looked very good, and the feature feels solid in practice.
I've left a few comments.

frontend/src/i18n/locales/zh/translation.json
alokdangre: I did not use "上一版" because this key is also used in Pod logs, where “previous version” is not ideal. 
I am using "上一个"

illume:
Thanks for the review @Kaniikura

illume:
Excellent work.

A little nit pick, the last two commits needs a capital letter at the start of the sentence after the context part.

> frontend: workloads: Validate explicit rollback targets and cache history

It's not necessary, but would you mind making a short silent screen capture demoing it? (To add after the screenshots in the PR description) I think it will make a nice feature to show a video of to people when we promote the next release.

PR #4723, opened 2026-02-17
Backend: Frontend: Add ability to transfer files to/from containers in pods 

mudit06mah:
## Summary

This PR adds feature to upload/download files to/from pod containers

## Related Issue

Fixes #2348

## Changes

- Added `pkg/transfer` to support upload and download
- Added tests for `copy.go` and `handler.go`
- Added route in `headlamp.go`
- Added upload & download button to terminal component along with dialog boxes

### Download:
1. Open Pod Terminal (pod should not be distro-less).
2. Click on `Download Files From Container` button and add absolute path to file/dir
3. Click on `Download`, file should be downloaded to your system.

### Upload:
1. Open Pod Terminal (pod should not be distro-less).
2. Click on `Upload Files to Container` button and select file from your system.
3. Add absolute path of dir where the file is to be transfered.
4. Click on `Upload`, file should be uploaded to container (check using ls)

### Download

https://github.com/user-attachments/assets/e3bcd7b6-5c7d-46ed-96a0-14a6a187c596

### Upload

https://github.com/user-attachments/assets/bbb96a3c-7279-45af-92f6-68d96fb77c8f

## Summary by CodeRabbit

* **New Features**
  * Added terminal controls to download files from and upload files to container environments.
  * Added path and destination dialogs with validation, progress feedback, success messages, and error handling.
  * Downloads preserve suggested filenames, while successful uploads refresh the terminal view.
  * Added backend support for secure file transfers and authentication.

* **Localization**
  * Added file-transfer terminology and messages across supported language locales.

* **Tests**
  * Added coverage for transfer workflows, validation, error responses, and terminal states.

PR #4729, opened 2026-02-18
frontend:kubectlcp: Add kubectl cp command at pod details view

### LINKED ISSUE #3952: File upload / download from containers (kubectl cp)
A way of uploading and downloading files/directories to/from containers - i.e. `kubectl cp`

The ability to specify the path of the file/directory within the container, and to browse local files/directories.

So they can better debug their containers, and perform manual testing/admin

No

ChayanDass:
# PR Description

## Summary

This PR adds pod copy (kubectl cp–style upload/download). 

## Related Issue

Fixes #3952

## Changes

- **Pod copy (upload/download):**
  - `frontend/src/lib/k8s/podCopy.ts`: copyFromPod (download via tar), copyToPod (upload: single file via cat, multiple via tar), path normalization, stderr handling with path-error hints.
  - `frontend/src/components/pod/PodCopyDialog.tsx`: Copy files dialog (container select, Download/Upload tabs, path inputs, file picker).
  - Pod details: "Copy files" action (icon) opens PodCopyDialog; gated by AuthVisible subresource="exec".

## Steps to Test

1. **Upload single file:** Workloads → Pods → open a pod → Copy files → Upload to container → destination `/tmp`, choose one file → Upload. Expect "Upload completed". Verify with `kubectl exec <pod> -- ls -la /tmp` and `cat` the file.
2. **Upload to nonexistent dir:** Use destination `/temp/file.txt` (where `/temp` does not exist). Expect error like "cannot create ... Directory nonexistent" with hint "The destination directory must exist (e.g. /tmp)" and no raw JSON.
3. **Download:** Copy files → Download from container → path e.g. `/tmp` or `/usr/share/nginx/html/index.html` → Download. Expect a .tar download. For nonexistent path, expect error and "The path may not exist in the container."
4. **Multi-file upload:** Copy files → Upload → choose two or more files → destination `/tmp`. Expect success; requires tar in container.

image


frontend/src/components/pod/PodCopyDialog.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const url = URL.createObjectURL(blob);
+  const a = document.createElement('a');
+  a.href = url;
+  a.download = filename;
+  a.click();
+  URL.revokeObjectURL(url);
+}
+
+export function PodCopyDialog({ open, onClose, pod }: PodCopyDialogProps) {
+  const { t } = useTranslation(['glossary', 'translation']);
```
gambtho: as we're using the glossary namespace here, i think you needed to add the copy key to glossary.json, rather than translation.json, otherwise it will always return the english version

mudit06mah:
Hi! @ChayanDass 
It appears that I have already opened a PR that introduces the same feature :)
Check: #4723 
My PR has a different approach to it, I have added this in the Terminal component instead of pod details

illume:
Oh there's two different issues as well.

illume:
@ChayanDass @mudit06mah What are the pros and cons of each approach?

mudit06mah:
> What are the pros and cons of each approach?

I don't think there are necessarily "pros" and "cons", it's just a preference in where you want this feature in the UI.

I personally think that it will be better suited in the terminal view since you can then access those files from the terminal (see the attached video in my PR) and you already have the container selection drop down menu there.
This solution feels more "friction-less" to me, for example: if someone wants to check which file they want to download, the can just use the terminal and then directly click the download button instead of going to the Pod Details view, similarly for uploading, they can verify their upload to the path.

I think maybe we should ask the author of both issues as to which approach they prefer :)

ChayanDass:
Yup, you’re right, I liked your approach. It’s more accessible for users,  they can verify things easily.

One suggestion: after uploading, it would be helpful if you automatically ran the ls command on the target folder so users can immediately confirm that the files were uploaded successfully.

edit: I think you already have this

mudit06mah:
> One suggestion: after uploading, it would be helpful if you automatically ran the ls command on the target folder so users can immediately confirm that the files were uploaded successfully.

Good suggestion! I have added the functionality to run `ls` command after successful uploads 👾

illume:
Ok, let's close this one.

Thanks both.

PR #4736, opened 2026-02-19
fix(storybook): return 404 for unsupported API versions in baseMocks

### LINKED ISSUE #4735: storybook: Use 404 instead of network error for unsupported API versions in baseMocks
In `.storybook/baseMocks.ts`, the v1beta1 CRD endpoint uses `HttpResponse.error()` which simulates a network failure (fetch rejection) rather than an HTTP error. Real clusters that don't support a deprecated API version return a 404 or 410. This could cause fallback logic to behave differently in Storybook vs real clusters.

Should use something like:

`() => new HttpResponse(null, { status: 404 })`

kunaldevsahu:
## Summary

This PR updates Storybook baseMocks to return a 404 HTTP response
for unsupported API versions instead of throwing a generic network error.
This aligns mock behavior with actual backend responses and improves
error handling consistency.

## Related Issue

Fixes #4735  

## Changes

- Updated baseMocks to return HTTP 404 for unsupported API versions
- Replaced generic network error handling logic

## Steps to Test

1. Run Storybook locally
2. Trigger a request using an unsupported API version
3. Verify that the response returns a 404 status instead of a network error

N/A

## Notes for the Reviewer

- This change only affects Storybook mock behavior.
- No backend logic was modified.


frontend/.storybook/baseMocks.ts
```diff
@@ -25,7 +25,7 @@ import { NODE_DUMMY_DATA } from '../src/components/node/storyHelper';
  *
  */
 export const baseMocks = [
-  http.get('http://localhost:4466/wsMultiplexer', () => HttpResponse.error()),
+  http.get('http://localhost:4466/wsMultiplexer', () => new HttpResponse(null, { status: 404 })),
```
gambtho, line 28: I think the linked issue #4735 describes a problem with `apiextensions.k8s.io/v1beta1/customresourcedefinitions` endpoints using `HttpResponse.error()` to simulate network failures instead of returning 404. However, this PR changes the `wsMultiplexer` endpoint instead. The `wsMultiplexer` HTTP mock exists intentionally to suppress MSW "unhandled request" warnings for WebSocket URLs — `HttpResponse.error()` is the correct behavior there since MSW cannot intercept WebSocket upgrade connections via `http.get()` handlers.

PR #4759, opened 2026-02-22
frontend: Terminal: useTerminalStream: Focus terminal when showing it

### LINKED ISSUE #4754: Keyboard focus should be in terminal when launched
When I launch a terminal in a pod, I need to click on the terminal window before I can type commands. If I type commands before clicking in the terminal, nothing happens. Every time a user launches a terminal, they intend on typing commands into it.

As soon as the terminal is launched, the keyboard focus moves to it.

This will save a click every time someone launches a terminal. It'll also be a less confusing experience for new users who will need to figure out they need to click into the terminal before they can type commands.

No

illume:
## Summary

Users had to click inside the terminal window before typing — this PR auto-focuses the xterm terminal immediately after it mounts.

Fixes #4754

## Related Issue

- #4754

## Changes

- Added `xterm.focus()` after `xterm.open()` in `Terminal.tsx` (pod exec terminal) and `useTerminalStream.ts` (used by `PodDebugTerminal` and `NodeShellTerminal`)

## Steps to Test

1. Open a pod's terminal (Exec tab)
2. Without clicking, type a command
3. Characters should appear immediately — no click required

Also verify with:
- Node shell terminal
- Pod debug/ephemeral container terminal

### DIFF
--- a/frontend/src/components/common/Terminal.tsx
+++ b/frontend/src/components/common/Terminal.tsx
@@ -78,6 +78,7 @@ export default function Terminal(props: TerminalProps) {
     }
 
     xterm.open(containerRef);
+    xterm.focus();
 
     let lastKeyPressEvent: KeyboardEvent | null = null;
     xterm.onData(data => {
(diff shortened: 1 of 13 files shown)


illume:
Thanks @vyncent-t

Issue #4710, opened 2026-02-15
Projects by label selector/helm deployment


segevfiner:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
In some Kubernetes setups it is common to deploy multiple applications in one namespace with shared access control and quotas assigned to that namespace, currently projects are assigned to a whole namespace (Even across clusters). It would be cool to be able to create a project based on a label selector instead, which assuming labels are properly assigned to all related resources, makes it possible to view all resources for a certain application even when there are multiple application deployed in the namespace.

## Describe the solution you'd like
Allow creating projects which are assigned to a label selector in a namespace

## What users will benefit from this feature?
Users that use headlamp (External or in-cluster) in Kubernetes clusters that have multiple applications in one namespace.

## Are you able to implement this feature?
[Yes (I will propose a PR) / No.]
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

illume:
@segevfiner 

This is for multiple applications for a single namespace?
(Not so projects could work across multiple namespaces?)

Can you please give a concrete example or some examples of the selectors you’re thinking of? Maybe a minimal scenario with a bit more detail.

segevfiner:
Yeah, multiple applications in a single namespace.

For example, one applications resources all have:
```
app.kubernetes.io/name: Foo
app.kubernetes.io/instance: main-app
```
And another has:
```
app.kubernetes.io/name: Bar
app.kubernetes.io/instance: test
```

Or any other labeling convention there off. And they both live in one namespace, let's say `default` and I want to focus on only one or the other in projects view, or even in the listing views of Headlamp (which is a separate different issue https://github.com/kubernetes-sigs/headlamp/issues/2547), rather than only filtering by namespace.

illume:
Thanks for that.

I should edit that issue to make it clear that filtering by labels is left to do there.

illume:
I've been working on the label filter in listing view this week:
- https://github.com/kubernetes-sigs/headlamp/pull/4760

PR #3160, opened 2025-04-23
app: backend: Add flag for watching plugins changes

skoeva:
This change creates a CLI flag in the backend + desktop for watching changes to the plugins or their directory.

Fixes: #2894, follow-up from #2939

#### Backend
- [X] Run `make backend`
- [X] Run `cd backend && ./headlamp-server -watch-plugins-changes false`

#### Desktop
- [X] Run `cd app && npm run build`
- [X] Run `cd dist/linux-unpacked && ./headlamp -watch-plugins-changes  false` and ensure that the flag is passed in

PR #4760, opened 2026-02-22
frontend: Add first-class label and selector support to resource views

### LINKED ISSUE #2547: Filtering/searching by pod labels
People put labels on things, and it would be good to use the labels.

Searching or filtering with labels.

Ideally filtering AND global search would work with labels.

- [x] filtering by labels in search
- [ ] filtering by labels in list view

All the people who use labels. I think lots of people use labels (most?).

## Additional context

- Requested in a [slack thread](https://kubernetes.slack.com/archives/C01FXB5E8ER/p1730995623614169).
- This is one of the top requested features for Kubernetes dashboard: https://github.com/kubernetes/dashboard/issues/3160
- This was also requested by a user in person.

Similar to kubectl (see [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#api) for reference).

For example, here's a manifest for a Pod that has two labels environment: production and app: nginx:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: label-demo
  labels:
    environment: production
    app: nginx
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```

`kubectl get pods -l environment=production`

illume:
## Summary

Imagine not being able to click on labels or tags. Now in Headlamp, you can.

This PR adds Kubernetes label filtering to resource list views. It supports the full Kubernetes
label selector syntax, exactly the same syntax accepted by `kubectl -l`.

It also supports the workflow described in #6932: labels on resource detail pages are links. Clicking one opens the list for the same resource type with that label applied as a URL-backed filter. The shared metadata component makes this available to Pods and other Kubernetes resource types.

Namespaces and Label Selector are positioned like subtitles beneath the resource title. Previously, on
larger browser windows, the namespace filter could appear far to the right and out of the user's
immediate view, making it unclear that the current page was filtered by namespace. Placing both
filters beneath the resource title lets people quickly understand what the page contains and how
it is scoped. The active values are subtitle links: they adopt link styling on hover or focus, and
clicking one opens the editor. A filter icon beside Create always toggles both fields, and applying
a value with Enter returns to subtitle display mode. The subtitle location is the right place for
Namespaces and Label Selector because together they define the current view.

Related requests:
- [Headlamp #2547: Filtering/searching by pod labels](https://github.com/kubernetes-sigs/headlamp/issues/2547)
- [Headlamp #4710: Projects by label selector/Helm deployment](https://github.com/kubernetes-sigs/headlamp/issues/4710)
- [Headlamp #6820: Label selectors in Resource Map filtering](https://github.com/kubernetes-sigs/headlamp/issues/6820)
- [Headlamp #6929: Show pods targeted by a Service](https://github.com/kubernetes-sigs/headlamp/issues/6929)
- [Headlamp #6930: Show pods targeted by a NetworkPolicy](https://github.com/kubernetes-sigs/headlamp/issues/6930)
- [Headlamp #6932: Clicking a label adds a search filter](https://github.com/kubernetes-sigs/headlamp/issues/6932)
- [Headlamp #6982: Copy resource labels and annotations](https://github.com/kubernetes-sigs/headlamp/issues/6982)
- [Headlamp #7028: Clicking a pod selector shows matched pods](https://github.com/kubernetes-sigs/headlamp/issues/7028)
- [Kubernetes Dashboard #3160: Allow the user to search for labels](https://github.com/kubernetes-retired/dashboard/issues/3160) - one of most highly requested Kubernetes Dashboard issues (35 thumbs-up reactions)
- [Kubernetes Dashboard #7547: Filter by pod status and labels](https://github.com/kubernetes-retired/dashboard/issues/7547)

## Why Labels Matter

People use labels in kubernetes. A lot.

Namespaces are essential for scope, access control, and isolation, but they provide only one coarse
dimension: a namespaced resource belongs to one namespace. Labels complement that boundary with
many simultaneous dimensions. The same object can be grouped by application, component,
environment, team, owner, release, tier, and topology, without changing where it lives.

For discovery and day-to-day operations, labels can therefore be as important as namespaces, and
often more useful. Labels can identify related resources within a large shared namespace, across
multiple namespaces, and among cluster-scoped resources that have no namespace. Kubernetes also
understands label selectors directly: controllers, Services, and policies use them to associate
resources in ways that namespace membership alone cannot express.

Helm charts, GitOps workflows, monitoring, cost management, security policy, and automation also
rely heavily on labels. As clusters grow, namespaces remain the isolation boundary, while labels
become the flexible index used to find and operate on related resources. Making label selectors a
first-class list filter gives both organizational mechanisms appropriate weight in Headlamp.

## Performance and Scalability

Label selectors are passed to the Kubernetes API, so Kubernetes filters resources before returning
them to Headlamp. Headlamp does not need to download every object and then filter the full set in
the browser, reducing network transfer, memory use, and client-side processing for large clusters.

Server-side filtering is also required for correct results when list requests are paginated. With
30,000 or more resources, client-side filtering can inspect only the pages fetched so far and miss
matching objects on pages that have not been downloaded. Passing the selector to Kubernetes makes
each returned page part of the complete filtered result set without fetching all unfiltered items.

## Changes

- Add a **Label Selector** filter to resource list headers.
- Support the full Kubernetes label selector syntax, including equality (`app=nginx`), inequality
  (`tier!=backend`), set-based (`env in (production,staging)` and `env notin (dev)`), existence
  (`partition`), and non-existence (`!partition`) queries.
- Show common query examples from an accessible information icon on hover, keyboard focus, or tap.
- Show active Namespaces and Label Selector values as MUI-style subtitle links.
- Keep a filter icon beside Create to toggle the editable fields in every state.
- Return to subtitle display mode after applying a filter with Enter.
- Use stable, matching editor widths: full-row on mobile and wide enough for realistic values on
  medium and large screens.
- Apply label selectors to generic resource tables and the paginated Pod list.
- Persist selectors per cluster and reflect them in the URL for bookmarks and sharing.
- Add reusable URL query parameter helpers with tests.
- Make detail-page labels keyboard-accessible links to filtered lists.
- Link Pod Node Selectors to filtered Nodes and resource Selectors to filtered Pods.
- Validate selector syntax before saving it or sending a Kubernetes request.
- Recognize explicit selectors in global search and show matching resource kinds.
- Probe Kubernetes server-side with the complete selector and `limit=1` instead of downloading
  unfiltered resource lists for selector searches.
- Keep annotations and other metadata dictionaries non-interactive.
- Document filtering and the detail-to-list workflow.

## Steps to Test

1. Navigate to a resource list such as Pods, Deployments, or Services.
2. Click the filter icon beside Create, enter `app=nginx` in **Label Selector**, and press Enter or
  leave the field.
3. Verify only matching resources remain and `labelSelector=app%3Dnginx` appears in the URL.
4. Clear the field and open a resource detail page with labels.
5. Click a label such as `app: nginx`.
6. Verify the same resource type's list opens with `app=nginx` applied.
7. Repeat with a cluster-scoped resource or another namespaced resource.
8. Enter an invalid selector such as `app in (` and verify the field shows an error without
  changing the URL or results.
9. On a Pod detail page, click a Node Selector and verify the Nodes list opens with that selector.
10. On a workload detail page, click its Selector and verify the matching Pods list opens.
11. Open global search, enter `environment in (production),tier in (frontend)`, and verify a
  **Pods** result appears when matching Pods exist.
12. Select that result and verify the Pods list opens with the complete selector in the URL.

## Validation

- `npm --prefix frontend test -- --run --reporter=dot --silent=passed-only`
  - 241 test files passed
  - 3,215 tests passed
- `npm --prefix frontend run build`
- `npx playwright test tests/labelSelectors.spec.ts`
  - 3 tests passed against Kind: validation/list filtering, detail-label navigation, and global
    selector search.
- Live Kind verification:
  - Clicking `app: nginx` opened `#/c/kind-test/pods?labelSelector=app%3Dnginx`.
  - The **Label Selector** input contained `app=nginx`.
  - Two nginx Pods were shown and the redis Pod was excluded.
  - Three realistic namespaces and a long selector appeared as compact subtitle links.
  - The filter action appeared immediately after Create and toggled both editor fields.
  - Pressing Enter after editing returned to subtitle display mode.
  - Mobile subtitles stayed within the viewport with three matching Pods visible.
  - The examples tooltip used an opaque surface (`opacity: 1`).
  - Invalid selectors stayed in the editor and did not update state, storage, or the URL.
  - Pod labels, Pod Node Selectors, workload Selectors, and workload Node Selectors all opened the
    corresponding filtered lists.
  - Global search sent the full set-based selector in 17 resource-kind probes with `limit=1`.
  - The matching **Pods** result opened the server-filtered list, and nonmatching Pods were absent.

### Cleaner when no filters are set

If no selectors are set, then they are not shown. This is a bit cleaner than before, because now
there is no empty Namespaces input when Namespaces is empty. Instead, people can press the filter
icon to filter by namespace or label selector.

Resource list with no active filters and a filter action beside Create

### Detail pages: labels and selectors are clickable now

Before this change, labels and selectors on resource detail pages were static text. They are now
clickable links that open the corresponding resource list with the selected label or selector
applied.

Pod detail page with clickable labels

Click on a label and it goes to the List view, so you can see all the pods with that label.

### Pod labels and Node Selectors

Pod labels open a filtered Pods list. Node Selector chips open the filtered Nodes list, using the
same selector in the Kubernetes request.

Pod labels and Node Selector linked to filtered lists

### Workload Selectors

Workload Selector chips open matching Pods, while Node Selector chips open matching Nodes.

Workload Pod Selector and Node Selector linked to filtered lists

### Filtered Pod list

The URL, all three Namespaces, and the long **Label Selector** remain visible as subtitle links.

Populated namespace and label selector subtitles

### Global search works with kubernetes selectors now

Type a Kubernetes label selector into global search, for example
`environment in (production),tier in (frontend)`. Headlamp shows a result for each resource type
with at least one match, such as **Pods**. Click the result to open that resource list with the
selector applied.

Headlamp asks Kubernetes for only one matching object per resource type while building these
results. This keeps the search lightweight and avoids downloading entire unfiltered resource lists.

Global search result for a set-based label selector

### Query examples tooltip

Seasoned Kubernetes users may already know label selector syntax from `kubectl -l`. New users can
discover the same syntax by hovering over, focusing, or tapping the information icon. The tooltip
uses an opaque surface so the page behind it cannot interfere with readability.

Opaque Kubernetes label query examples tooltip

### Subtitles, before they could be missed

In this 'before' screenshot, look at 'Pods' title, now tell me which namespaces this view is for? All? Only some? Which ones? You can't tell.

On the main branch, a large screen pushes the selected namespace context far to the right of the
resource title. Even with three namespaces selected, the empty gap makes the active scope easy to
miss. Labels are not shown because the main branch does not yet have the Labels filter.

Namespace filter pushed to the right on the main branch

#### After

Now you know what namespaces this view is for. Before it was very easy to miss, and even if you
noticed, you needed to click the namespace filter box to see.

Namespaces and label selector shown as editable subtitles

#### Hover

The values keep subtitle styling at rest. On hover or keyboard focus, link styling and a descriptive
tooltip make the editing action discoverable.

Namespace subtitle showing its link hover state

#### Clicking filter selector icon or clicking on a selector brings up selector edit input fields

When neither Namespaces nor Label Selector is set, the filter icon directly after Create shows both
editable fields. While editing, the same action hides the fields and returns to display mode.

Filter editor opened from the title action

### Mobile

Namespaces and Label Selector are shown as compact subtitle links. Three matching Pods keep the
filtered result visible below them.

Namespace and label selector subtitles on mobile

#### Input selectors variation

For comparison, the editor keeps both labeled inputs full-width while leaving matching Pods
visible below.

Namespace and label selector inputs on mobile

### Medium

All three namespaces and the realistic long selector remain visible as editable subtitles.

Namespace and label selector subtitles on a medium screen

### Large

The subtitle treatment keeps page scope easy to scan without stretching controls across the row.

Namespace and label selector subtitles on a large screen

Assisted by copilot


mudit06mah:
PR works as intended, I tested it from the UI for both single and multi-clusters for all cases present in steps to test.
UI also renders appropriate syntax error on incorrect syntax.
All stories pass as well!

mudit06mah (approved):
LGTM 👾

joaquimrocha (changes requested):
I don't think we should have a dedicated filter selector next to the namespace one. This will fill the space too much, leaving little space for the resources' title, and I do not think it's of interest for the majority of users.

IMO, one way we could implement this without making it somewhat intrusive would be to have it in the search that is part of the table. We could also have a labels column that's hidden by default but which allows then to have the search, if implementing something like `label:MY-TERM` is too complicated in the table's general search.

skoeva (commented):
agree two different label filters is too crowded, a table search could be a good idea to reducing crowding

sniok:
Since there wasn't activity for a while and I had an idea how to implement this without too many changes, I've made a different PR for this. I hope you don't mind 

I've opened a PR here https://github.com/kubernetes-sigs/headlamp/pull/5035 @joaquimrocha @skoeva @illume please take a look if that is good enough. I've basically just added a new column that table can search by with existing search bar

illume:
Some cleanups...

- Rebased onto current `main` and reorganized the work into seven atomic commits 
- Added TSDoc to `isValidLabelName`, `isValidLabelKey`, `isValidComparisonValue`, and `splitRequirements` in `labelSelectorValidation`.
- Removed the new `urlUtils` module and reused the existing `useQueryParamsState` hook, moving it to `frontend/src/lib` so common filters, tables, and Resource Map share one neutral implementation.
- Added coverage that query updates preserve unrelated URL parameters.
- Kept selector foundations, list filtering, detail navigation, global search, E2E coverage, generated locale messages, and docs in separate reviewable commits.

frontend/src/redux/filterSlice.ts
```diff
@@ -21,15 +21,23 @@ import { useMemo } from 'react';
...
 export interface FilterState {
   /** The namespaces to filter on. */
   namespaces: Set<string>;
+  /** The label selector to filter on. */
+  labelSelector: string;
 }
 
 export const initialState: FilterState = {
   namespaces: new Set(getSavedNamespaces()),
+  labelSelector: getSavedLabelSelector(),
```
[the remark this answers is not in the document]
illume: Addressed. Layout now dispatches restoreFiltersForCluster on every active-cluster change, and that reducer reloads both saved namespaces and the saved label selector for the target cluster. Added a reducer regression test plus a two-cluster Playwright scenario that verifies each cluster shows its own saved selector.

frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -248,6 +267,19 @@ export function MetadataDictGrid(props: MetadataDictGridProps) {
 
     let labelComponent = <MetadataEntry>{shortText}</MetadataEntry>;
 
+    if (labelFilterRoute) {
+      labelComponent = (
+        <Link
+          routeName={labelFilterRoute}
+          activeCluster={activeCluster}
+          search={{ labelSelector: `${key}=${dict[key]}` }}
```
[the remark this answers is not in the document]
illume: Addressed. MetadataDictGrid now has an explicit all-entry mode for Selector and Node Selector rows, while resource metadata labels keep their existing per-label links. All workload, Service, Job, PDB, Pod node-selector, DaemonSet, and StatefulSet callers use the complete dictionary. Added a two-key unit regression and a Service Playwright scenario that verifies the complete selector reaches the Pod list request.

Suyog241005 (commented):
This is such a great addition! Being able to click on labels and filter by them is something I've always wanted in Headlamp. The subtitle placement for namespaces and label selectors is a really nice touch too.

A couple of small ideas that I think could make it even better:

1. **`kubectl` copy helper** :- would be super handy to have a button that copies the equivalent `kubectl` command (e.g. `kubectl get pods -l app=nginx`). Lots of devs jump between the UI and terminal, so that would save some time!

2. **Label key autocomplete** :- as you type in the Label Selector field, showing hints from the actual labels in the cluster would make filtering even faster, especially on big clusters with tons of custom labels.

LGTM

frontend/src/components/podDisruptionBudget/Details.tsx
```diff
@@ -59,7 +44,14 @@ export default function PDBDetails(props: { name?: string; namespace?: string; c
           },
           {
             name: t('glossary|Selector'),
-            value: <>{selectorsToJSX(item.selectors)}</>,
+            value: (
+              <MetadataDictGrid
+                activeCluster={item.cluster}
+                dict={item.spec.selector.matchLabels}
+                labelFilterMode="all"
```
unlikelyzero: This is the dropped optional chaining mentioned above - `item.spec?.selector?.matchLabels ?? {}` would restore the old guard. `MetadataDictGrid` itself is null-safe, so the crash is entirely on this call site.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -90,15 +100,85 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
   if (!!propsActions) {
     actions = actions.concat(propsActions);
   }
+  const titleSideActions = [...(headerProps.titleSideActions || [])];
 
+  const filters: React.ReactNode[] = [];
   if (!noNamespaceFilter) {
-    actions.push(<NamespacesAutocomplete />);
+    filters.push(<NamespacesAutocomplete key="namespace-filter" onApply={closeFilterEditor} />);
+  }
```
unlikelyzero, line 125: This is the resourceClass gate mentioned above - worth checking against role/List.tsx, crd/List.tsx, and the other list pages that pass `data=` instead of `resourceClass`, since none of them thread the Redux label selector into their own `useList` calls. `pod/List.tsx` does this correctly via `useLabelSelector()`, might be worth applying the same pattern to the others, or setting `noLabelFilter` on them if filtering genuinely isn't supported there.

frontend/src/components/common/SectionFilterHeader.tsx
```diff
@@ -82,6 +86,12 @@ export default function SectionFilterHeader(props: SectionFilterHeaderProps) {
     []
   );
 
+  React.useEffect(() => {
+    if (labelSelectorFromURL && labelSelectorFromURL !== filter.labelSelector) {
+      dispatch(setLabelSelectorFilter(labelSelectorFromURL));
+    }
```
unlikelyzero, line 109: This only dispatches when `labelSelectorFromURL` is truthy - there's no `else` branch to clear the Redux value when the param disappears from the URL, which is what causes the Back-button issue mentioned above. Also worth validating this before dispatching - right now a malformed selector from a bookmarked/shared URL goes straight to the API with none of the checks `LabelSelectorInput` applies to manual edits.

frontend/src/lib/labelSelectorValidation.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    const setMatch = requirement.match(/^(.+?)\s+(in|notin)\s*\((.*)\)$/);
+    if (setMatch) {
+      const [, key, , rawValues] = setMatch;
+      if (!isValidLabelKey(key.trim())) {
+        return `invalid label key "${key.trim()}"`;
+      }
+
+      const values = rawValues.split(',').map(value => value.trim());
```
unlikelyzero, line 143: This accepts an empty value list for `in`/`notin` (e.g. `env in ()`), which the real Kubernetes API server rejects - the PR's own test file asserts this case as valid. In the other direction, `isValidComparisonValue` requires `/^\d+$/` and rejects negative operands like `replicas>-1`, which apimachinery's `strconv.ParseInt`-based parsing actually accepts. Both are worth fixing since the backend already vendors the real parser (`backend/pkg/config/config.go` uses `labels.Parse`) - just not reachable from here.

frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -185,18 +191,52 @@ export function MetadataDisplay<T extends KubeObject>(props: MetadataDisplayProp
...
+  activeCluster?: string;
+}
+
+/**
+ * Converts a match-label dictionary to an AND-combined Kubernetes selector.
+ *
+ * @param dict - Match-label key-value pairs.
+ * @returns A comma-separated equality selector.
+ */
+function labelDictionaryToSelector(dict: MetadataDictGridProps['dict']): string {
```
unlikelyzero: This is the matchLabels-only join mentioned above - swapping in `labelSelectorToQuery({ matchLabels: dict })` from `lib/k8s` would pick up matchExpressions for free and drop this custom joiner entirely.

frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -248,6 +288,21 @@ export function MetadataDictGrid(props: MetadataDictGridProps) {
 
     let labelComponent = <MetadataEntry>{shortText}</MetadataEntry>;
 
+    if (labelFilterRoute) {
+      const labelSelector =
+        labelFilterMode === 'all' ? labelDictionaryToSelector(dict) : `${key}=${dict[key]}`;
+      labelComponent = (
+        <Link
+          routeName={labelFilterRoute}
```
unlikelyzero, line 301: This link only carries `labelSelector`, not `namespace` - the global namespace filter stays whatever it was, so clicking a label from a resource outside the currently-filtered namespace can land on a guaranteed-empty list with no indication why. Might be worth setting the namespace param (and dispatching the namespace filter) to the source resource's own namespace here.

e2e-tests/tests/multiCluster.spec.ts
```diff
@@ -112,6 +112,30 @@ test.describe('multi-cluster setup', () => {
     await headlampPage.hasTitleContaining(/Choose a cluster/);
   });
 
+  test('switching clusters restores each saved label selector', async ({ page }) => {
```
unlikelyzero, line 135: This seeds localStorage directly and does a full page reload per cluster, which only exercises `filterSlice`'s `initialState` reading from storage - not the actual in-app cluster-switch path (`Layout`'s `restoreFiltersForCluster` effect firing with no reload) that the original Copilot comment was about. This would still pass if that reducer path were completely broken.

unlikelyzero (commented):
`podDisruptionBudget/Details.tsx` dropped the optional chaining the old `selectors` getter had - it's now `item.spec.selector.matchLabels` instead of `item.spec?.selector?.matchLabels`. `spec.selector` is legitimately optional on a PDB, so a PDB with none (or one that arrives partially populated mid-watch) throws and takes out the whole details page.

Bigger picture: the label filter UI shows up on every list page but only actually applies inside `TableFromResourceClass`'s `resourceClass` path - roles, CRDs, CR instances and a few others fetch their own data and silently ignore the selector while still showing it as active. The `noLabelFilter` prop that looks built for exactly this has no call sites anywhere, and wouldn't stop the filter from applying even if used.

Selector links also only read `matchLabels`, dropping `matchExpressions` - there's already a `labelSelectorToQuery` in `lib/k8s` that handles both, used elsewhere in this same PR, just not here. And the URL-to-Redux sync for the selector is one-way, so Back doesn't clear a filter and it can silently follow you across pages.

A few more things inline.

frontend/src/components/podDisruptionBudget/Details.tsx
illume: Restored the optional guard before rendering the selector metadata. I also corrected the PDB model to use the shared Kubernetes LabelSelector shape, so matchExpressions remains an array and can be serialized safely. The focused PDB tests and TypeScript check pass.

frontend/src/components/common/SectionFilterHeader.tsx
illume: Data-driven ResourceListView pages now set noLabelFilter because they cannot propagate a selector server-side. Resource-class-driven pages keep the filter. Added focused coverage for both cases and updated the affected storyshots.

frontend/src/components/common/SectionFilterHeader.tsx
illume: SectionFilterHeader now observes URL selector changes, clears Redux when browser history removes the parameter, and validates URL values before dispatch. Malformed selectors resolve to an empty selector and are not sent to the API. Added focused Back/history-removal and malformed-URL tests.

frontend/src/lib/labelSelectorValidation.ts
illume: I checked this against the exact k8s.io/apimachinery labels.Parse behavior used by this branch. `env in ()` and `env notin ()` both parse successfully. `replicas>-1` is rejected because `-1` is not a valid label value, while `replicas>1` parses successfully. The frontend tests match those parser results, so I left these semantics unchanged.

frontend/src/components/common/Resource/MetadataDisplay.tsx
illume: Selector rows now receive a serialized complete LabelSelector instead of rebuilding only matchLabels in MetadataDisplay. The serializer was extracted into a pure labelSelector module and callers pass the full workload selector, including matchExpressions. Focused tests and the complex StatefulSet storyshot cover this.

frontend/src/components/common/Resource/MetadataDisplay.tsx
illume: Namespaced metadata and selector links now include the source resource namespace in the query string, while cluster and parameterized route data remain preserved. MetadataDisplay tests and detail storyshots cover the namespace propagation.

e2e-tests/tests/multiCluster.spec.ts
illume: Updated the E2E to perform a real in-app switch: it navigates once to the test cluster, opens the cluster chooser, selects test2 without a reload, and verifies the restored test2 selector. This now exercises Layout restoreFiltersForCluster rather than only initial state hydration.

illume:
Thanks @unlikelyzero and @Suyog241005 for the reviews!!

Work in progress, marking it draft.

illume:
> Label key autocomplete :- as you type in the Label Selector field, showing hints from the actual labels in the cluster would make filtering even faster, especially on big clusters with tons of custom labels.

@Suyog241005
Implemented label-key autocomplete in the Label Selector field. Suggestions come from label keys on resources already loaded by the active server-backed list, so this adds no duplicate cluster request and remains bounded by the list data already fetched. Completion is prefix-based, applies only to the active selector requirement, preserves earlier requirements and negation, and supports mouse or keyboard selection. Added focused extraction, propagation, syntax, mouse, and keyboard tests.

illume:
> `podDisruptionBudget/Details.tsx` dropped the optional chaining the old `selectors` getter had - it's now `item.spec.selector.matchLabels` instead of `item.spec?.selector?.matchLabels`. `spec.selector` is legitimately optional on a PDB, so a PDB with none (or one that arrives partially populated mid-watch) throws and takes out the whole details page.
> 
> Bigger picture: the label filter UI shows up on every list page but only actually applies inside `TableFromResourceClass`'s `resourceClass` path - roles, CRDs, CR instances and a few others fetch their own data and silently ignore the selector while still showing it as active. The `noLabelFilter` prop that looks built for exactly this has no call sites anywhere, and wouldn't stop the filter from applying even if used.
> 
> Selector links also only read `matchLabels`, dropping `matchExpressions` - there's already a `labelSelectorToQuery` in `lib/k8s` that handles both, used elsewhere in this same PR, just not here. And the URL-to-Redux sync for the selector is one-way, so Back doesn't clear a filter and it can silently follow you across pages.
> 
> A few more things inline.

@unlikelyzero
Thanks for the detailed review. These issues have now been addressed:

- Restored the optional PDB guard with item.spec?.selector?.matchLabels, so an absent or partially populated selector no longer crashes the details page. The PDB selector type now also uses the shared Kubernetes LabelSelector shape.
- Label filtering is capability-gated. Resource-class lists keep the filter because [TableFromResourceClass](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) sends it server-side; unsupported data-driven lists hide it. The paginated Pod list explicitly enables the filter because it applies the selector in its own useList call.
- Selector links now use labelSelectorToQuery with the complete selector object. This preserves matchExpressions alongside matchLabels, and each rendered selector chip links using the complete selector.
- URL synchronization now handles parameter removal and malformed URL selectors. Browser Back/history removal clears Redux, and invalid selectors are not sent to the API.
- Namespaced selector links now carry the source namespace, avoiding empty results caused by a stale namespace filter.
- Added focused coverage for the PDB guard, data-driven/resource-class filter gating, complete selectors including expressions, namespace propagation, Back/history removal, and malformed URL selectors.

Selectors remain cluster-global during ordinary navigation between supported resource lists, matching the existing namespace-filter behavior, but they are now always shown in the active-filter summary and are hidden on pages that cannot apply them.

frontend/src/components/globalSearch/labelSelectorSearch.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param labelSelector - Optional validated selector to pass directly to Kubernetes.
+ * @returns Options for `KubeObject.useList`.
+ */
+export function makeGlobalSearchListOptions(
+  inACluster: boolean,
+  labelSelector: string | null
+): { clusters: string[] | undefined; labelSelector?: string; limit?: number } {
+  return {
+    clusters: inACluster ? undefined : [],
+    ...(labelSelector ? { labelSelector, limit: 1 } : {}),
```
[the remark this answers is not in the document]
illume: Fixed by separating the Kubernetes API page limit from list-request scheduling. Global selector search now passes `fetchAllRequests: true` with `limit: 1`, so every selected-cluster/allowed-namespace request starts immediately while each probe still fetches at most one item. Ordinary paginated lists keep the existing limit-based scheduling behavior. Added a hook-level regression covering four probes across two clusters and four restricted namespaces, asserting `limit=1` on every request and no `loadMore`; focused tests, the full 2,843-test frontend suite, TypeScript, lint/format, and production build all pass.

frontend/src/components/pod/List.tsx
```diff
@@ -572,9 +575,11 @@ export function PodListRenderer(props: PodListProps) {
 }
 
 export default function PodList() {
```
rootp1, line 577: The Pod request is filtered, but the parallel PodMetrics request is still unfiltered. Since metrics pagination follows the unfiltered ordering and is only advanced when the filtered Pod list has loadMore, matching Pods beyond the first metrics page can permanently show no metrics. Pass the same labelSelector to PodMetrics.useList so both result sets stay aligned.

@illume maybe this can be considered

rootp1 (commented):
Once can you consider this copliot comment before proceeding  @illume

frontend/src/components/workload/Details.tsx
rootp1, line 1: Why is this a being shown as a binary file? @illume this must be a real issue, digging into this file for more findings because this has never been found by copilot, because it can't see diff for github detected binary files

rootp1, line 1: Image

I think this is making an issue @illume (line:52)

rootp1 (changes requested):

frontend/src/components/pod/List.tsx
illume: Fixed. `PodMetrics.useList` now receives the same `labelSelector` as `Pod.useList`, keeping the two paginated result sets aligned. `List.test.tsx` covers both calls and the focused suite passes.

frontend/src/components/workload/Details.tsx
illume: Fixed. The binary classification came from a literal NUL byte in this source file. The sentinel is now written as the escaped TypeScript source `\0loading`, and `*.tsx` is explicitly treated as LF text with a text diff driver. HEAD contains zero literal NUL bytes and Git now reports a normal text diff (`11` insertions, `1` deletion).

illume: Yes, the literal source NUL on this line was the cause of the binary-file issue. It has been replaced with the escaped source `\0loading`: this preserves the collision-resistant runtime sentinel without embedding a NUL byte in the file. The focused workload tests pass. Thanks!!!!

rootp1: suree

rootp1: true dat.

illume:
Fixed some CI failures and rebased against main to fix conflicts.

PR #4763, opened 2026-02-22
Add websocket-mode config with UI toggle (websockets/multiplexer/off)

### LINKED ISSUE #4505: Make websocket multiplexer a setting rather than just an environment variable
Add an ability to turn multiplexer on and off in general settings. 

It would be easier to debug if any issues are related to websocket in general or specific to multiplexer.

Currently people need to use an environment variable, but this means only people who build headlamp can try this feature.

In the settings area it should also mention "Experimental feature", or maybe even a whole section of experimental features, and this setting could go under that "Experimental feature" heading.

Everyone who needs a "web socket multiplexer". That is people with multiple clusters, or who are looking at many resources at once.

It should still respect the environment variable: REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER

If that is true then it should be enabled, if false disabled.

Should the user setting over ride this? Yes. If the user has not changed the setting at all, then just use the environment variable.

## Additional context

See previous multiplexer PRs

https://github.com/kubernetes-sigs/headlamp/pulls?q=is%3Apr+multiplexer+is%3Aclosed

See frontend/src/lib/k8s/api/v2/useKubeObjectList.ts

Currently you need to set this when running or building frontend.

```
REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER=true npm run frontend:build
REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER=true npm run frontend:start
```

illume:
## Summary

Some users requested to turn web sockets off entirely, and some want to use the web socket multiplexer.
This allows configuration of this variable via an environment variable and via settings in the app.

Adds a `websocket-mode` configuration option to control WebSocket real-time update behaviour. Three modes are supported:

- `websockets` (default): standard WebSocket connections
- `multiplexer`: experimental multiplexed connection for improved performance
- `off`: disable real-time updates entirely (useful for very busy clusters where real-time updates make the UI unusable)

The mode can be set at multiple levels (highest priority wins):
1. **User preference in General Settings UI** (persisted in localStorage) — always overrides env vars and server config
2. Build-time env var `REACT_APP_WEBSOCKET_MODE` (or legacy `REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER`) — used as fallback when user is on "Default"
3. Server-configured value via `HEADLAMP_CONFIG_WEBSOCKET_MODE` env var or `websocketMode` Helm value (no CLI flag added)
4. Default: `websockets`

## Related issues

Fixes #4505
- https://github.com/kubernetes-sigs/headlamp/issues/4505

## Changes

**Backend:**
- Added `WebsocketMode` string field (koanf: `websocket-mode`, default: `websockets`) to `Config` and `HeadlampCFG`
- Configured via `HEADLAMP_CONFIG_WEBSOCKET_MODE` env var (no CLI flag added, to avoid long-term maintenance commitment)
- Validates the value on startup (errors on unknown modes; unset defaults to `websockets`)
- Updated `/config` API endpoint to expose `websocketMode` to the frontend

**Frontend:**
- Added `websocketMode` string field to Redux `ConfigState`; validated before storing
- Added `websocketModeUserOverride` to persisted app settings (localStorage) so the user's choice survives page reloads
- Added `WebsocketMode` type, `useWebsocketMode()` hook and `getWebsocketMode()` non-hook; both use `import.meta.env` only (browser-safe), honour legacy `REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER` for backwards compatibility, and validate the runtime config value
- `useWatchKubeObjectLists` and `useKubeObject` route to multiplexed or legacy watcher based on mode; both hooks called unconditionally (Rules of Hooks); `off` disables both
- **New `WebsocketModeSettings` component** in General Settings: a Select with four options — Default, Websockets, Multiplexer, Off
- Added `start-websockets-off`, `start-websockets-multiplexer` and `start-websockets-websockets` npm dev scripts
- Added `'off'` mode regression test and legacy `false`-override tests

**Charts:**
- Added `websocketMode` string value to Helm chart (`values.yaml`, schema, deployment template, README)
- Regenerated all expected templates with Helm v3.18.4

**Docs:**
- Updated in-cluster and development docs to describe `HEADLAMP_CONFIG_WEBSOCKET_MODE` env var configuration (CLI flag section removed)

## Steps to Test

- There are frontend, backend, and charts tests added.
- There are also e2e tests which cover multiple scenarios.

**Manual testing:**

1. Start with multiplexer mode: `HEADLAMP_CONFIG_WEBSOCKET_MODE=multiplexer make run-backend`
2. Open Headlamp → Settings → General Settings → **WebSocket mode** → verify the dropdown shows the current mode and can be changed
3. Change to `Off` — confirm real-time updates stop (resource list no longer auto-refreshes)
4. Change to `Multiplexer` — check the browser/app console to see the multiplexed WebSocket in the `/config` response
5. Helm: `helm install headlamp . --set config.websocketMode=multiplexer`

Screenshot 2026-02-22 at 19 07 39
Screenshot 2026-02-22 at 18 59 17

## Notes for the Reviewer

- The `off` mode is particularly useful for very busy clusters where real-time updates make the UI hard to use
- Both `useWebSocket` (multiplexer) and `useWebSockets` (legacy) hooks are always called unconditionally to comply with Rules of Hooks; the mode controls which one receives active connections
- The user's UI preference overrides both build-time env vars and server config; env vars are a fallback when the user is on "Default"


frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
```diff
@@ -31,11 +34,43 @@ import { WebSocketManager } from './multiplexer';
...
+  // Check build-time environment variable first (supports testing and build-time config)
+  if (import.meta.env.REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER === 'true') {
+    return true;
+  }
+
+  // Fall back to runtime configuration from backend
+  const isWebsocketMultiplexerEnabled = useSelector(
+    (state: RootState) => state.config.isWebsocketMultiplexerEnabled
+  );
+  return isWebsocketMultiplexerEnabled ?? false;
```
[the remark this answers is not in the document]
illume: fixed

frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
```diff
@@ -140,21 +175,25 @@ export function useWatchKubeObjectLists<K extends KubeObject>({
...
+  });
+
+  const legacyResult = useWatchKubeObjectListsLegacy({
+    kubeObjectClass,
+    endpoint,
+    lists: !isMultiplexerEnabled ? lists : [],
+    queryParams,
+  });
+
+  return isMultiplexerEnabled ? multiplexedResult : legacyResult;
```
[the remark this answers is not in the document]
illume: fixed

backend/pkg/config/config.go
```diff
@@ -437,6 +438,7 @@ func addGeneralFlags(f *flag.FlagSet) {
 	f.Bool("allow-kubeconfig-changes", false,
 		"Allow Headlamp to make changes to the known kubeconfigs when needed. "+
 			"E.g. to remove a cluster via the UI. May not be recommendable when Headlamp is deployed as a service.")
+	f.Bool("enable-websocket-multiplexer", false, "Enable experimental WebSocket multiplexer for improved performance")
```
joaquimrocha: If we add this, then we are "documenting" that this CLI exists and we have to carry it forward for a few versions at least. Why not just turn this feature on with an env var and document it in documentation only? It would have less implications for maintenance IMO.

joaquimrocha (commented):
I don't really agree with adding this CLI option as I think it has implications for future maintenance. I also think that the people willing to test this option would be able to just run the program/container with a special env var.

skoeva (changes requested):
this pr needs to be rebased against main

illume:
Rebased onto current `main`.

illume:
Missed a new commit on main... rebased again and updated test.

illume:
I added some new e2e tests to give more confidence this is working properly.

rootp1:
@illume  
I pulled this PR locally and ran the changed backend and frontend paths. I found two blocking issues before I could sign off. The new `--websocket-mode` CLI flag is documented and validated but not actually registered, and the frontend only updates the runtime `websocketMode` when the cluster list changes, which defeats the new runtime config path when clusters stay stable.\n\nWhat I ran locally\n\n```bash\ncd backend && go test -v ./cmd ./pkg/config\ncd backend && go test -v ./pkg/k8cache ./pkg/kubeconfig\ncd frontend && npm test -- src/lib/k8s/api/v2/multiplexer.test.ts src/lib/k8s/api/v2/useKubeObjectList.test.tsx src/lib/k8s/api/v2/hooks.test.tsx\n```\n\nAll of those passed, so these look like coverage gaps rather than failures already caught by the suite.

backend/pkg/config/config.go
```diff
@@ -235,6 +236,31 @@ func (c *Config) validateClusterInventory() error {
...
+// validateWebsocketMode validates the WebsocketMode field and applies a default.
+// An empty value (i.e. the field was not set via any configuration source) is
+// normalised to "websockets". Any other value that is not one of the three
+// supported modes ("websockets", "multiplexer", "off") is rejected with an error.
+//
+// Configuration sources in priority order (highest wins):
+//  1. HEADLAMP_CONFIG_WEBSOCKET_MODE environment variable
+//  2. websocketMode Helm value (injected as the env var above by the chart)
+//  3. Default: "websockets"
+func (c *Config) validateWebsocketMode() error {
```
rootp1, line 248: Blocking. This adds the new `websocket-mode` config surface and validation, but the CLI flag itself is still not registered in `addGeneralFlags`. I reproduced `config.Parse([]string{"headlamp", "--websocket-mode=off"})` locally and it fails with `flag provided but not defined: -websocket-mode`. So env based config works, but the documented CLI path is broken. This needs the flag registration plus a parse test for the CLI path.

frontend/src/components/App/Layout.tsx
```diff
@@ -147,7 +147,11 @@ const fetchConfig = (dispatch: Dispatch<UnknownAction>) => {
       clustersToConfig[cluster.name] = cluster;
     });
 
-    const configToStore = { ...config, clusters: clustersToConfig };
+    const configToStore = {
+      ...config,
+      clusters: clustersToConfig,
+      websocketMode: config?.websocketMode ?? 'websockets',
```
rootp1, line 155: Blocking. `websocketMode` is added to `configToStore` here, but later in `fetchConfig` it is only dispatched when `clusters === null` or `isEqualClusterConfigs(...)` reports a cluster diff. If the backend changes `/config.websocketMode` while the cluster list stays the same, Redux never gets the new mode and `useWebsocketMode()` stays stale indefinitely. This runtime config needs to be compared and dispatched independently of cluster membership.

backend/pkg/config/config.go
illume: Addressed in 297540dee: addGeneralFlags now registers --websocket-mode with the documented values, and TestParseWithEnv covers --websocket-mode=off through config.Parse.

frontend/src/components/App/Layout.tsx
illume: Addressed in 2b16e5e31: config refresh now compares websocketMode independently of cluster membership and dispatches when only the mode changes. Focused tests cover changed mode, unchanged config, and changed clusters.

illume:
This needs some pretty major changes to how the multiplexer works. Basically, it was broken in-cluster when we moved how authentication worked.

This can't be fixed in time for 0.45.0, so moving it to the next one.

PR #4771, opened 2026-02-24
github: Fix picking the hash for chocolatey version bump

joaquimrocha:
## Summary

The logic to update the chocolatey version was checking for the hash
of the Headlamp file by checking for the right architecture only.
If, for some reason, we have more than one Headlamp version in a file,
it could pick up the wrong file for the bump. So this change makes
sure we choose the hash corresponding to the right architecture and
version.

## Steps to Test

This can only be tested next time we release. But it's an easy, logical change.


joaquimrocha:
Since this was already tested and is low risk. I am merging it after addressing Copilot's review.

Issue #4779, opened 2026-02-25
Overlay/Drawer opens default details view instead of plugin's custom detailsRoute


yolossn:
## Describe the bug
When a plugin creates a custom resource class (e.g., extending Pod) and defines a custom `detailsRoute`, clicking the resource name in a `ResourceListView` table opens the **default** details page for the original/built-in resource class in the overlay/drawer panel instead of navigating to the route defined by the plugin's `detailsRoute`.

Hovering over the link shows the **correct** URL (from `getDetailsLink()`, which respects the custom `detailsRoute`), but clicking opens the wrong component in the drawer.

## To Reproduce
Steps to reproduce the bug:
1. Create a plugin with a custom resource class that overrides `detailsRoute`:

```typescript
export class MyPod extends KubeObject<MyPodInterface> {
  static apiVersion = 'v1';
  static apiName = 'pods';
  static kind = 'Pod';
  static isNamespaced = true;

  static get detailsRoute(): string {
    return `/my-plugin/:namespace/pods/:name`;
  }
}
```

2. Register a custom detail page route and use `ResourceListView` with the custom class:

```typescript
registerRoute({
  path: '/my-plugin/:namespace/pods/:name',
  sidebar: 'my-plugin-pods',
  component: MyPodDetailPage,
});

function MyPodsPage() {
  return (
    <ResourceListView
      title="My Pods"
      resourceClass={MyPod}
      columns={['name', 'namespace', 'age']}
    />
  );
}
```

3. Enable the detail drawer/overlay feature (the split-right panel)
4. Click any pod name in the "My Pods" list view

### Expected behavior

The link navigates to the plugin's custom detail page at `/my-plugin/:namespace/pods/:name`, or at minimum does not open the wrong component in the overlay.

### Actual behavior

The overlay drawer opens and renders the **default** `PodDetails` component (the built-in Headlamp pod details view), completely ignoring the plugin's custom `detailsRoute` and registered detail page.

### Root Cause

In `src/components/common/Link.tsx`, the `Link` component determines whether to open the drawer using the Kubernetes `kind` string:

```typescript
// Link.tsx line 149
const kind = 'kubeObject' in props ? props.kubeObject?._class().kind : props?.routeName;

// Link.tsx line 155-156
const openDrawer =
    drawerEnabled && canRenderDetails(kind)
      ? () => { /* opens drawer */ }
      : undefined;
```

[...]

PR #4781, opened 2026-02-25
frontend: Link: Respect plugin detailsRoute

yolossn:
The overlay drawer used the Kubernetes kind to pick a built-in details component and intercepted navigation, so plugin-defined detailsRoute components were never reached. When a KubeObject class has a custom detailsRoute, skip opening the drawer and let the router navigate to the plugin’s details page instead.

Fixes: kubernetes-sigs/headlamp#4779


illume (approved):
🎉 thanks!

PR #4782, opened 2026-02-25
frontend: useKubeObjectList: ensure correct class instantiation

### LINKED ISSUE #4780: Query cache sharing causes wrong class instantiation for resources with the same API endpoint
### Description

When a plugin creates a custom resource class that targets the same Kubernetes API as a built-in class (e.g., a custom Pod class wrapping `/api/v1/pods`), the `useList` query results are non-deterministic. The `_class()` method on returned items may return either the built-in class or the custom class, depending on **which page was visited first**. This causes `getDetailsLink()`, `detailsRoute`, and other class-dependent instance methods to resolve incorrectly.

### Steps to Reproduce

1. Create a plugin with a custom Pod class:

```typescript
export class MyPod extends KubeObject<MyPodInterface> {
  static apiVersion = 'v1';
  static apiName = 'pods';
  static kind = 'Pod';
  static isNamespaced = true;

    return `/my-plugin/:namespace/pods/:name`;
  }
}
```

2. Register a list page using the custom class:

```typescript
function MyPodsPage() {
  return (
    <ResourceListView
      title="My Pods"
      resourceClass={MyPod}
      columns={['name', 'namespace', 'age']}
    />
  );
}
```

3. Navigate to the **built-in** Pods list page (`/c/:cluster/pods`)
4. Then navigate to the **plugin's** custom Pods list page (`/my-plugin/pods`)
5. Observe the name column links in the plugin's table

### Expected behavior

Links in the custom Pods list should always point to the plugin's detail route (e.g., `/my-plugin/default/pods/my-pod-xyz`), as defined by `MyPod.detailsRoute`.

### Actual behavior

Links point to the **default** Pod detail route (e.g., `/c/:cluster/pods/default/my-pod-xyz`), because the items were instantiated with the built-in `Pod` class from the query cache.

The behavior is **order-dependent**:

| Navigation order | Custom Pods list links | Built-in Pods list links |
|---|---|---|
| Built-in Pods → Custom Pods | **WRONG** (default route) | Correct |
| Custom Pods → Built-in Pods | Correct | **WRONG** (custom route) |
| Refresh on Custom Pods | Correct | — |
| Refresh on Built-in Pods | — | Correct |

### Root Cause

In `src/lib/k8s/api/v2/useKubeObjectList.ts`, `kubeObjectListQuery` generates a React Query cache key that does **not** include any class discriminator:

```typescript
// useKubeObjectList.ts lines 75-83
queryKey: [
  'kubeObject',
  'list',
  kubeObjectClass.apiVersion,  // 'v1' for both Pod and MyPod
  kubeObjectClass.apiName,      // 'pods' for both Pod and MyPod
  cluster,
  namespace ?? '',
  queryParams,
],
```

Both the built-in `Pod` class and `MyPod` have `apiVersion = 'v1'` and `apiName = 'pods'`, producing **identical** query keys.

The `queryFn` creates instances using its own `kubeObjectClass`:

```typescript
// useKubeObjectList.ts lines 95-103
list.items = list.items.map(item => {
  const itm = new kubeObjectClass({
    ...item,
    kind: list.kind.replace('List', ''),
    apiVersion: list.apiVersion,
  });
  itm.cluster = cluster;
  return itm;
});
```

But since the keys are identical, **whichever class triggers the query first "wins"** — its `queryFn` runs and the result is cached. The second class gets the cached result (items instantiated with the wrong constructor) without its own `queryFn` ever executing.

Since `_class()` returns `this.constructor`:

```typescript
// KubeObject.ts lines 481-483
_class() {
}
```

All instance methods that delegate to the class (`detailsRoute`, `listRoute`, `getDetailsLink()`, etc.) resolve based on whichever class happened to instantiate the items first.

### Flow diagram

```
Scenario: User visits built-in Pods page first, then plugin's Pods page

1. Built-in Pods page renders
   → Pod.useList() called
   → kubeObjectListQuery creates key: ['kubeObject', 'list', 'v1', 'pods', cluster, '', {}]
   → queryFn runs: items = data.map(d => new Pod(d))
   → Cache stores Pod instances under this key

2. User navigates to plugin's Pods page
   → MyPod.useList() called
   → kubeObjectListQuery creates key: ['kubeObject', 'list', 'v1', 'pods', cluster, '', {}]
   → SAME KEY as step 1 → cache hit!
   → queryFn does NOT run (data is fresh/not stale)
   → Returns cached Pod instances (not MyPod instances)

3. ResourceTable renders name column
   → <Link kubeObject={item} />
   → item._class() returns Pod (wrong — should be MyPod)
   → item.getDetailsLink() uses Pod.detailsRoute ("Pod") instead of MyPod.detailsRoute
   → Link shows default Pod detail URL ← WRONG
```

### Proposed Fix

In `useKubeObjectList`, after retrieving items from the query cache, verify they are instances of the expected class. If not, re-instantiate them. This preserves cache sharing (avoiding duplicate API calls) while ensuring each consumer gets items with the correct class constructor:

```typescript
const items = useMemo(() => {
  if (!query.items || query.items.length === 0) return query.items;
  if (query.items[0]._class() === kubeObjectClass) return query.items;

  return query.items.map(item => {
    if (item._class() === kubeObjectClass) return item;
    const corrected = new kubeObjectClass(item.jsonData);
    corrected.cluster = item.cluster;
    return corrected as K;
  });
}, [query.items, kubeObjectClass]);
```

### Alternative approaches considered

1. **Include class discriminator in query key** — Would correctly separate caches but causes duplicate API calls for the same data. Not ideal for performance.
2. **Use React Query `select` option** — Called on every cache read, but creates new references each time, potentially causing infinite re-renders without careful memoization.
3. **Post-query re-instantiation with `useMemo`** (proposed) — Non-invasive, preserves shared cache, only re-instantiates when needed, memoized to avoid unnecessary work.

### Affected files

- `src/lib/k8s/api/v2/useKubeObjectList.ts`

### Impact

Any plugin that creates a custom class targeting a built-in Kubernetes resource API will experience non-deterministic behavior in list views. Links, route resolution, and any logic depending on `_class()` will be incorrect depending on navigation order. This fundamentally breaks the plugin pattern of creating custom views for standard Kubernetes resources.

https://github.com/user-attachments/assets/e5b5d684-d063-48e9-b20d-0e9045a021b2

yolossn:
Ensure items returned from useKubeObjectList are instantiated with the expected KubeObject class even when React Query cache entries are shared between resources targeting the same API endpoint. This prevents non-deterministic routing and detail link generation for plugins that wrap built-in Kubernetes resources.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/4780

### DIFF
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
@@ -536,9 +536,26 @@ export function useKubeObjectList<K extends KubeObject>({
 
   const errors = query.errors.filter(it => it !== null);
 
+  // When multiple classes share the same API endpoint (e.g. a plugin's custom Pod class
+  // and the built-in Pod class), they produce identical query keys and share the cache.
+  // Items may have been instantiated with a different class than what the caller expects.
+  // Re-instantiate with the correct class so that _class() returns the right constructor
+  // and instance getters like detailsRoute resolve correctly.
+  const items = useMemo(() => {
+    if (!query.items || query.items.length === 0) return query.items;
+    if (query.items[0]._class() === kubeObjectClass) return query.items;
+
+    return query.items.map(item => {
+      if (item._class() === kubeObjectClass) return item;
+      const corrected = new kubeObjectClass(item.jsonData);
+      corrected.cluster = item.cluster;
+      return corrected as K;
+    });
+  }, [query.items, kubeObjectClass]);
+
   // @ts-ignore - TS compiler gets confused with iterators
   return {
-    items: endpointError ? [] : query.items,
+    items: endpointError ? [] : items,
     errors: endpointError ? [endpointError] : errors.length > 0 ? errors : null,
     error: endpointError ?? query.errors.find(it => it !== null) ?? null,
     clusterResults: query.clusterResults,
@@ -547,7 +564,7 @@ export function useKubeObjectList<K extends KubeObject>({
     isFetching: query.isFetching,
     isSuccess: query.isSuccess,
     *[Symbol.iterator](): ArrayIterator<ApiError | K[] | null> {
-      yield query.items;
+      yield items;
       yield endpointError ?? query.errors.find(it => it !== null) ?? null;
     },
   };


sniok:
I don't think we should be recreating all the items in the cache just because a plugin for some reason recreated built-in classes.

In that case, if for example plugin recreates Pod classes and doesn't implement `getLogs` or some other method that we expect then the whole application will break.

Allowing plugins to override default KubeObject classes is not supported and shouldn't be since the rest of the app is expecting our own instances.

yolossn:
IMHO we are not recreating all the items in the cache. The data in the cache stays untouched and only the class it is resolved to changes per consumer.

I agree that there can be very limited to no usecase for a plugin to recreate a resource class that already exists, but it is very common for devs to extend existing classes to add extra functionalities (e.g. `class MyPod extends ResourceClasses.Pod`). The developer will not be aware that just by extending the class they are unknowingly sharing a cache key with the built-in class and that their instances will resolve incorrectly depending on navigation order.

The fix also doesn't re-instantiate unconditionally, it uses the first item as a fast-path check to skip re-instantiation entirely when the class already matches, and inside the map each item is checked individually so already correctly typed items are returned as-is.

illume (changes requested):
Thanks for these changes.

PR #4812, opened 2026-03-02
frontend: NavigationTabs: Refactor component so it doesn't require cloning

sniok:
We recently had a regression with tabs. Sidebar items can now have jsx and cluster has jsx element defined, tabs are incredibly slow in development because the NavigationTabs implementation is deepcloning the items (which in development have `window`).

This PR refactors the implementation to not require cloning at all

You'll also notice that the storybook snapshots changed quite drastically, because they now use correct mocked store. For example checkout the "WorkloadsParentSelected" story, in the previous snapshot it has items like "stateful sets", "daemon sets" which are not present [in the mocked store ](https://github.com/sniok/headlamp-o/blob/62831987b6f82e69c46846ddb5935131c8dda3ee/frontend/src/components/Sidebar/NavigationTabs.stories.tsx#L55). So there was a bug somewhere in the previous implementation which is now fixed. 

To reproduce:

start headlamp in development mode
click on different sidebar items
it will be slow, you can record performance trace in devtools to see that it spends a lot of time cloning items

### DIFF
--- a/frontend/src/components/Sidebar/NavigationTabs.tsx
+++ b/frontend/src/components/Sidebar/NavigationTabs.tsx
@@ -18,8 +18,7 @@ import Box from '@mui/material/Box';
 import Divider from '@mui/material/Divider';
 import { useTheme } from '@mui/material/styles';
 import useMediaQuery from '@mui/material/useMediaQuery';
-import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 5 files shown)


illume:
@sniok can you please put the steps to see the regression in the PR description? (I guess there's no issue?)

sniok:
added repro steps

PR #4873, opened 2026-03-12
Update French translation for 'Sign In'

adam-carruthers:
## Summary

This PR fixes a French translation. "Sign in" was translated as "S'inscrire" which actually means "Sign up". This confused some of our users who were connecting via SSO and already have an account.

## Changes

- Updated the french i8n file


adam-carruthers:
@yolossn would you mind reviewing please?

PR #4924, opened 2026-03-20
frontend: Table: SimpleTable: Fix to announce no data

illume:
image

### DIFF
--- a/frontend/src/components/common/SimpleTable.tsx
+++ b/frontend/src/components/common/SimpleTable.tsx
@@ -28,6 +28,7 @@ import TableHead from '@mui/material/TableHead';
 import TablePagination from '@mui/material/TablePagination';
 import TableRow from '@mui/material/TableRow';
 import { SxProps } from '@mui/system';
+import { visuallyHidden } from '@mui/utils';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { getTablesRowsPerPage, setTablesRowsPerPage } from '../../helpers/tablesRowsPerPage';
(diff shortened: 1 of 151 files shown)


frontend/src/components/common/Table/Table.tsx
```diff
@@ -475,69 +473,86 @@ export default function Table<RowItem extends Record<string, any>>({
...
-              cells={row.getVisibleCells() as MRT_Cell<Record<string, any>, unknown>[]}
-              table={table as MRT_TableInstance<Record<string, any>>}
-              isSelected={row.getIsSelected()}
-              onRowClick={handleRowClick}
-            />
-          ))}
-        </StyledBody>
-      </MuiTable>
-      <MRT_BottomToolbar table={table} />
+      <Box role="status" aria-live="polite" aria-atomic="true" sx={visuallyHidden}>
```
illume, line 558: There is one return point now, and the visuallyHidden announce area is always in the dom.

frontend/src/components/common/SimpleTable.tsx
```diff
@@ -300,192 +301,221 @@ export default function SimpleTable(props: SimpleTableProps) {
...
-    </TableContainer>
+        </TableContainer>
+      );
+    }
+  }
+
+  return (
+    <>
+      <Box role="status" aria-live="polite" aria-atomic="true" sx={visuallyHidden}>
+        {announcedStatus}
```
illume, line 515: There is one return point now, and the visuallyHidden announce area is always in the dom.

sniok (approved):

Issue #4927, opened 2026-03-20
The metrics for pods with the same name are reported incorrectly.


Fausto-R91:
## Describe the bug
In Workloads > Pods, Headlamp shows incorrect CPU/memory metrics when multiple pods share the same name across different namespaces. All pods with the same name display identical metrics. Filtering by namespace shows the correct values.

## To Reproduce
Steps to reproduce the bug:
1. Deploy multiple pods with the same name in different namespaces
2. Open Headlamp and navigate to Workloads > Pods.
3. Observe that pods with the same name show identical metrics (see screenshot).

## Environment (please provide info about your environment):
 - Installation type: In-cluster with Helm
 - Headlamp Version: 0.40.1

## Are you able to fix this issue?
No

## Additional Context
The API response from `/apis/metrics.k8s.io/v1beta1/pods` contains correct per‑namespace metrics, so the issue appears to be in the frontend.

The problematic code seems to be here:

https://github.com/kubernetes-sigs/headlamp/blob/5ebe4ee3e80e48d25fb16b239208b2535a284316/frontend/src/components/pod/List.tsx#L203
https://github.com/kubernetes-sigs/headlamp/blob/5ebe4ee3e80e48d25fb16b239208b2535a284316/frontend/src/components/pod/List.tsx#L212

The matching logic uses only the pod name and does not consider the namespace.

<img width="705" height="763" alt="Image" src="https://github.com/user-attachments/assets/b600f062-62f9-4967-9ca0-4cb094e1cc3f" />

WasThatRudy:
This was fixed on main by [b88d01be0](https://github.com/kubernetes-sigs/headlamp/commit/b88d01be0e9e6da0b7b6d547295d60c355072f55) (PR #4933) on 2026-03-21, which adds a namespace check to the metric `find()` callbacks in `PodListRenderer`. The issue can be closed.

PR #4933, opened 2026-03-21
fix: Pod metrics matching missing namespace (#4927)

Utkarshpandey0001:
Fixes issue #4927 

This PR fixes a bug where Headlamp displays incorrect CPU and memory metrics for pods that share the same name across different namespaces (e.g., in the Workloads > Pods view).

Previously, the 
getCpuUsage and getMemoryUsage functions in List.tsx matched metrics to pods using only the pod name (it.getName() === pod.getName()). This caused all pods with the same name to display the metrics of the first matched pod.

This fix updates the lookup logic to also compare the namespace (it.getNamespace() === pod.getNamespace()), ensuring each pod correctly displays its own individual metrics.

### DIFF
--- a/frontend/src/components/pod/List.tsx
+++ b/frontend/src/components/pod/List.tsx
@@ -200,7 +200,9 @@ export function PodListRenderer(props: PodListProps) {
   const { t } = useTranslation(['glossary', 'translation']);
 
   const getCpuUsage = (pod: Pod) => {
-    const metric = metrics?.find(it => it.getName() === pod.getName());
+    const metric = metrics?.find(
+      it => it.getName() === pod.getName() && it.getNamespace() === pod.getNamespace()
+    );
     if (!metric) return;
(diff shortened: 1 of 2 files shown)


sniok (changes requested):
thanks for contributing! could you please make sure the PR follows our contributing guidelines https://headlamp.dev/docs/latest/contributing#submitting-a-pull-request-pr

Utkarshpandey0001:
@sniok Thanks for the review and for pointing me to the contributing guidelines. I've updated the PR description to remove the boilerplate text and amended the commit message to follow the proper format . I also verified that all local tests and snapshots are passing cleanly. Let me know if there's anything else you need me to adjust!

sniok:
there's still a merge commit and old PR description

sniok (approved):
thanks!

/lgtm

PR #4935, opened 2026-03-21
fix(frontend): handle cluster-scoped CRD links without failing on empty namespace

### LINKED ISSUE #4921: Crash Report: Expected "namespace" to match "[^\/]+?", but received ""
I get an error when I click the CVE link. In Headlamp, the namespace field is blank, which causes it to retrieve all namespaces.

## Crash Summary
Expected "namespace" to match "[^\/]+?", but received ""

## Error Stack
```
TypeError: Expected "namespace" to match "[^\/]+?", but received ""
    at https://headlamp.lakearch.co.uk/assets/index-Cn9qClYY.js:18:25924
    at gl (https://headlamp.lakearch.co.uk/assets/index-Cn9qClYY.js:25:6811)
    at ls (https://headlamp.lakearch.co.uk/assets/index-Cn9qClYY.js:1198:35893)
    at Cin (https://headlamp.lakearch.co.uk/assets/index-Cn9qClYY.js:1198:237776)
    at xd (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:38:17652)
    at Nh (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:40:44746)
    at Eh (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:40:40337)
    at Ek (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:40:40265)
    at nc (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:40:40118)
    at Gd (https://headlamp.lakearch.co.uk/assets/vendor-mui-B6u-D06-.js:40:36403)
```

## Headlamp Version
0.40.1

## Git Commit
434a12263a6bd9d25f86f8bfa00d9992ac2672e3

## System Information
- User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36
- Platform: MacIntel
- Language: en

Utkarshpandey0001:
Fixes #4921
This PR address issue #4921 where clicking the details link of a cluster-scoped custom resource (such as a ClusterVulnerabilityReport or other CVE-related plugins) crashed the UI with TypeError: Expected "namespace" to match "[^\/]+?", but received "".

The crashing behavior occurred because react-router requires all path parameters (like :namespace) to be non-empty, but cluster-scoped CRs naturally do not have a namespace. Attempting to use the customresource route fallback resulted in an empty parameter validation crash.

Changes made:

Added clusterCustomResource route in frontend/src/lib/router/index.tsx that explicitly omits the :namespace parameter (/customresources/:crd/:crName).
Updated Link Generation in CustomResourceList.tsx and CustomResourceInstancesList.tsx. It now dynamically checks if a namespace is present (cr.metadata.namespace). If it is, it uses the standard customresource route. If not, it uses the new clusterCustomResource route and omits the namespace param entirely.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ARG IMAGE_BASE=alpine:3.23.3@sha256:25109184c71bdad752c8312a8623239686a9a2071e8825f20acb8f2198c3f659
 FROM ${IMAGE_BASE} AS image-base
 
-FROM --platform=${BUILDPLATFORM} golang:1.25.8@sha256:dfae680962532eeea67ab297f1166c2c4e686edb9a8f05f9d02d96fc9191833e AS backend-build
+FROM --platform=${BUILDPLATFORM} golang:1.25.8@sha256:3ac2864710f25e84381bf5d4272261c7ba73ada0339d62034df4de20dabb33ca AS backend-build
 
 WORKDIR /headlamp
 
@@ -27,7 +27,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg/mod \
     cd ./backend && go build -o ./headlamp-server ./cmd/
 
-FROM --platform=${BUILDPLATFORM} node:22@sha256:7e791fc54bd02fc89fd4fb39eb37e5bea753c75679c8022478d81679367d995a AS frontend-build
+FROM --platform=${BUILDPLATFORM} node:22@sha256:ecabd1cb6956d7acfffe8af6bbfbe2df42362269fd28c227f36367213d0bb777 AS frontend-build
 
 # We need .git and app/ in order to get the version and git version for the frontend/.env file
 # that's generated when building the frontend.
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ARG IMAGE_BASE=alpine:3.23.3@sha256:25109184c71bdad752c8312a8623239686a9a2071e8825f20acb8f2198c3f659
 FROM ${IMAGE_BASE} AS image-base
 
-FROM --platform=${BUILDPLATFORM} golang:1.25.8@sha256:f55a6ec7f24aedc1ed66e2641fdc52de01f2d24d6e49d1fa38582c07dd5f601d AS backend-build
+FROM --platform=${BUILDPLATFORM} golang:1.25.8@sha256:dfae680962532eeea67ab297f1166c2c4e686edb9a8f05f9d02d96fc9191833e AS backend-build
 
 WORKDIR /headlamp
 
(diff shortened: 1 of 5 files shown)


sniok (changes requested):
could you please provide how to reproduce this error and how to test it?

Utkarshpandey0001:
@sniok Thanks for the review. Here is a screenshot of the fix in action, tested locally ,showing that the details page for a cluster scoped custom resource now renders correctly without throwing the expected namespace to match crash.
 
Screenshot 2026-04-04 at 1 42 28 AM

1. Open Headlamp, navigate via the left-sidemenu to Custom Resources, and find globalconfigs.test.com.
2. Click on the my-global-config instance. 

Before this PR: The UI immediately crashes with a React Router error boundary: TypeError: Expected "namespace" to   match "[^\/]+?", but received " ".

After this PR: The link directs successfully to the details page (bypassing the crashing namespace requirement entirely) and renders normally as seen in the screenshot above.

sniok (changes requested):
hi @Utkarshpandey0001 , I think there's a little misunderstanding. I'm wondering if you could provide the steps that cause a crash without a fix. We need this to verify how headlamp behaves both before and after the fix.

> Before this PR: The UI immediately crashes with a React Router error boundary: TypeError: Expected "namespace" to match "[^\/]+?", but received " ".

I don't think that's true. Here's a screenshot of a cluster scoped custom resource rendering just fine 

image

Utkarshpandey0001:
@sniok strange, let me check it once over my local system and then i will give you detail about it. For now i feel i did something wrong, anyway i will back to you shortly

Utkarshpandey0001:
@sniok Ah, I understand what happened on PR
why you couldn't reproduce the crash: In Headlamp, there is a Detail Drawer feature. When that feature is turned on, clicking a customresource link intercepts the click and opens the data in a sliding side drawer instead of explicitly navigating to the raw React Router URL. Because no explicit URL generation or validation happens in drawer mode, it never crashed for you.
The crash I solved was as I experienced the react-router crash because i middle-clicked to open in a new tab and also i had the Details Drawer disabled. React Router has strict parameter regex checking, and sending an empty namespace like this ' ' directly crashes the router.
My original fix: I created an entirely new route called clusterCustomResource to avoid the namespace parameter constraint.
The reason why it was rejected by copilot As Copilot pointed out the Detail Drawer component only recognizes links labeled as customresource. By creating a new route name I inadvertently disabled the Drawer feature entirely for cluster-scoped resources.

The Solution for it as Headlamp already has a mechanism to handle this and as Copilot suggested, the correct way to handle cluster-scoped custom resources is to keep the link mapped to customresource but pass a hyphen ("-") as a placeholder namespace. The CustomResourceDetails.tsx file natively understands that ns === '-' means the resource is cluster-scoped.

What I've done now - 
I have reverted structural route changes and properly applied Copilot's placeholder logic

sniok:
On the main branch middle clicking on a custom resource instance without namespace also works, so I'm not sure what this PR is supposed to fix

https://github.com/user-attachments/assets/e99a1787-53bb-4623-a98f-114b59ce9d33

Before updating PR further could you please provide steps to reproduce original issue? When does UI crash currently?

I'm marking this as a draft for now until we can figure out the original issue

PR #5006, opened 2026-03-29
frontend: Fix cross-namespace resource linking in Map view

### LINKED ISSUE #5003: Map view: Resource relations ignore namespaces, causing incorrect cross-namespace grouping
## Description

The Map view's resource relation predicates match resources by **name only** without checking namespace boundaries. This causes resources from different namespaces to be incorrectly linked, which in turn produces non-obvious grouping in the graph.

## Steps to Reproduce

1. Have resources with identical names in different namespaces (e.g., a ConfigMap named `config` in both namespace A and namespace B, or Services/Secrets with the same name)
2. Open the Map view with multiple namespaces visible
3. Observe that resources from different namespaces are grouped together

## Expected Behavior

Resources should only be connected within the same namespace (for namespace-scoped resources). A ConfigMap `config` in namespace A should not create an edge to a Pod in namespace B, even if the Pod references a ConfigMap named `config`.

## Actual Behavior

- Resources from different namespaces are merged into the same connected component
- Groups are named after the highest-weight node, which may be from a completely different namespace
- The resulting grouping appears arbitrary and non-obvious to the user

## Root Cause

In `frontend/src/components/resourceMap/sources/definitions/relations.tsx`, the `makeRelation` helper (line 61-74) only checks cluster equality:

```typescript
predicate(fromNode, toNode) {
    const fromObject = fromNode.kubeObject as InstanceType<From>;
    const toObject = toNode.kubeObject as InstanceType<To>;
    return fromObject.cluster === toObject.cluster && Boolean(selector(fromObject, toObject));
}
```

Almost all relation predicates match by **name only** without namespace checks:

- `configMapUsedInPods`: `volume.configMap?.name === configMap.metadata.name`
- `secretsUsedInPods`: `secret.metadata.name === env.valueFrom?.secretKeyRef?.name`
- `serviceToPods`: label selector match only
- `endpointsToServices`: `getName() === getName()`
- `ingressToService`: `service.metadata.name === path?.backend?.service?.name`
- `roleBindingsToRole`: `role.metadata.name === binding.roleRef.name`
- `hpaToDeployment`: `scaleTargetRef?.name === deployment.metadata.name`
- `hpaToStatefulSet`: `scaleTargetRef?.name === statefulSet.metadata.name`
- ...and most others

Only **two** relations correctly check namespace: `serviceAccountToDeployments` (line 201) and `serviceAccountToDaemonSets` (line 209).

## Suggested Fix

Add a namespace check to `makeRelation` for namespace-scoped resources:

```typescript
predicate(fromNode, toNode) {
    const fromObject = fromNode.kubeObject as InstanceType<From>;
    const toObject = toNode.kubeObject as InstanceType<To>;
    return fromObject.cluster === toObject.cluster 
        && fromObject.metadata?.namespace === toObject.metadata?.namespace
        && Boolean(selector(fromObject, toObject));
}
```

Cluster-scoped resources (where namespace is `undefined`) would still match correctly since `undefined === undefined`.

Resources that intentionally cross namespace boundaries (e.g., ClusterRole referenced by a RoleBinding) would need a separate `makeClusterRelation` helper without the namespace check.

## Related Issues

- #4310 (RWX PVC breaks groups) — related but different root cause (shared PVC within a namespace)

## Environment

- Headlamp version: main branch (commit e8c803fbf)
- Affected component: Map view (`frontend/src/components/resourceMap/`)

itvi-1234:
**Summary**

This PR solves the bug where resources with the same name in different namespaces were being incorrectly linked in the Map view.

Fixes #5003 

**Changes**

1.Modified makeRelation in relations.tsx to include a namespace check for resource connection.
2.Added a new makeClusterRelation helper for cluster-wide resources (like Webhook and GatewayClass) to still allow cross-namespace communication.
3.Updated vwcToService, mwcToService, and gatewayToGatewayClass to use this new helper.

Before:

image

After:

image

**Steps to Test**

1.Create a ConfigMap named 'config' in Namespace-A and Namespace-B.
2.View them in the Map.
3.Earlier, they would have been connected incorrectly, but now they are isolated in their own namespaces.

### DIFF
--- a/frontend/src/components/resourceMap/sources/definitions/relations.tsx
+++ b/frontend/src/components/resourceMap/sources/definitions/relations.tsx
@@ -69,7 +69,16 @@ const makeRelation = <From extends KubeObjectClass, To extends KubeObjectClass>(
     const fromObject = fromNode.kubeObject as InstanceType<From>;
     const toObject = toNode.kubeObject as InstanceType<To>;
 
-    return fromObject.cluster === toObject.cluster && Boolean(selector(fromObject, toObject));
+    const hasSameNamespace =
+      !fromObject.metadata?.namespace || !toObject.metadata?.namespace
+        ? true
+        : fromObject.metadata.namespace === toObject.metadata.namespace;
+
+    return (
+      fromObject.cluster === toObject.cluster &&
+      hasSameNamespace &&
+      Boolean(selector(fromObject, toObject))
+    );
   },
 });
 


frontend/src/components/resourceMap/sources/definitions/relations.tsx
```diff
@@ -69,6 +69,25 @@ const makeRelation = <From extends KubeObjectClass, To extends KubeObjectClass>(
     const fromObject = fromNode.kubeObject as InstanceType<From>;
     const toObject = toNode.kubeObject as InstanceType<To>;
 
+    return (
+      fromObject.cluster === toObject.cluster &&
+      fromObject.metadata?.namespace === toObject.metadata?.namespace &&
```
sniok: what if instead of new `makeClusterRelation` this function would only check for namespace equality when both object have namespace defined? that would make it a bit simpler

sniok (approved):
good catch, thanks! I've left one comment that could make this a bit simpler

sniok (changes requested):
sorry, meant to request changes

itvi-1234:
Great catch @sniok  That does make it much simpler and avoids the need for a separate makeClusterRelation helper altogether. I've updated makeRelation to conditionally check for namespace equality only when both objects define a namespace, and removed makeClusterRelation. Thanks for the feedback! Just pushed the changes

sniok (changes requested):
thanks, could you also squash all the changes into a single commit? so it follows our [guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines)

itvi-1234:
Done @sniok :rocket:

sniok (approved):

itvi-1234:
"Hey @sniok, while pushing the latest changes, I observed that the `pr-merge-commit-check` failed again despite my branch being a single, clean squash commit.

I did some research and found that the issue is likely in our `.github/workflows/pr-merge-commit-check.yml` script. The use of --depth=1 to fetch the base branch (main) causes Git to lose its ancestry relationship with the PR branch if they have diverged. Because of this 'shallow graft', git rev-list --merges ends up scanning the entire repository history and falsely flagging hundreds of old merge commits as being 'new' in the PR.

The PR is clean, but I'd be happy to open a separate PR to fix this CI check (e.g., by increasing the fetch depth or using merge-base logic) so it doesn't block other contributors in the future! Let me know if you'd like me to go ahead with that fix too."

PR #5047, opened 2026-04-04
frontend: redux: Surface original error messages in cluster actions

### LINKED ISSUE #5013: Admission controller errors not shown for Namespace creation/edit failures
## Describe the bug

Admission policies that control namespace resources do not surface errors in Headlamp - only the generic error `Failed to apply changes to XXXX.` is shown. For other resources, e.g. pods, a full error message from the admission controller is shown.

I have an admission policy that blocks namespaces without certain PSS labels (see below). If I create a namespace, or edit an existing namespace that lacks these labels, it will fail, but the details of the failure are not shown - only the generic error is shown.

## To Reproduce
Steps to reproduce the bug:
1. Apply admission control (see below example)
2. Create namespace in Headlamp
3. See error

Alternative to 2 - edit an existing namespace created before the rule, same error in 3.

`kubectl edit` says:
```
# * : ValidatingAdmissionPolicy 'enforce-restricted-pss' with binding 'enforce-restricted-pss-default' denied request: expression 'has(object.metadata.labe>
# ' resulted in error: no such key: pod-security.kubernetes.io/enforce
```

 - Installation type: Linux-Flatpak
 - Headlamp Version: 0.40.1

No

## Additional Context

Sample policy that causes the issue:
```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: enforce-restricted-pss
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
      - apiGroups: [""]
        apiVersions: ["v1"]
        operations: ["CREATE", "UPDATE"]
        resources: ["namespaces"]
  validations:
    - expression: >
        has(object.metadata.labels) &&
        object.metadata.labels["pod-security.kubernetes.io/enforce"] == "restricted" &&
        object.metadata.labels["pod-security.kubernetes.io/enforce-version"] == "latest"
      message: >
        Namespaces must enforce Pod Security Standards at the 'restricted' level.
---
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: enforce-restricted-pss-default
spec:
  policyName: enforce-restricted-pss
  validationActions:
    - Deny
```

itvi-1234:
## Summary
Refactor `dispatchError` in `clusterActionSlice` to surface detailed API error messages in notifications. 

Currently, many cluster actions (like creating or editing a Namespace) show a generic error message (e.g., "Failed to apply changes to XXX.") even when the Kubernetes API provides a specific reason for rejection (e.g., a `ValidatingAdmissionPolicy` failure). While some resources like Pods may show more detail, others only show the generic fallback.

This change ensures that if an error object is caught, its original message is appended to the generic error string, providing users with the necessary context (like admission policy details) to understand why an action failed.

## Changes
- Modified `dispatchError` in `frontend/src/redux/clusterActionSlice.ts` to accept an optional `err` parameter and append its message to the generic notification text.
- Added punctuation handling to ensure messages are concatenated cleanly.
- Updated `executeClusterAction` to pass caught errors to `dispatchError`.
- Updated `clusterActionSlice.test.ts` to verify the new concatenated error message format.

## Steps to Test
1. Set up a `ValidatingAdmissionPolicy` that blocks Namespace creation or updates (as described in the reported issue).
2. Attempt to perform an action in Headlamp that violates this policy.
3. **Before**: Observe a generic "Failed to apply changes to..." message.
4. **After**: Observe the generic message followed by the specific detail from the admission controller.

## Related Issue
Fixes #5013

### DIFF
--- a/frontend/src/redux/clusterActionSlice.ts
+++ b/frontend/src/redux/clusterActionSlice.ts
@@ -236,13 +236,22 @@ export const executeClusterAction = createAsyncThunk(
         })
       );
     }
-    function dispatchError() {
+    function dispatchError(err?: Error | string | null) {
+      let message = errorMessage || '';
+      if (err) {
+        const originalMessage = typeof err === 'string' ? err : err.message;
+        if (originalMessage) {
+          const separator = message ? (message.endsWith('.') ? ' ' : '. ') : '';
+          message = `${message}${separator}${originalMessage}`;
+        }
+      }
+
       dispatch(
         updateClusterAction({
           buttons: undefined,
           dismissSnackbar: actionKey,
           id: actionKey,
-          message: errorMessage,
+          message,
           state: 'error',
           snackbarProps: errorOptions,
           url: errorUrl,
@@ -278,7 +287,8 @@ export const executeClusterAction = createAsyncThunk(
         }
         dispatchSuccess();
       } catch (err) {
-        if ((err as Error).message === 'Action cancelled' || controller.signal.aborted) {
+        const error = err instanceof Error ? err : new Error(String(err));
+        if (error.message === 'Action cancelled' || controller.signal.aborted) {
           dispatchCancelled();
 
           if (cancelCallback) {
@@ -289,7 +299,7 @@ export const executeClusterAction = createAsyncThunk(
             }
           }
         } else {
-          dispatchError();
+          dispatchError(error);
         }
       } finally {
         controllers.delete(actionKey);
--- a/frontend/src/redux/clusterActionSlice.test.ts
+++ b/frontend/src/redux/clusterActionSlice.test.ts
@@ -167,7 +167,7 @@ describe('clusterActionSlice', () => {
         expect.objectContaining({
           type: updateClusterAction.type,
           payload: expect.objectContaining({
-            message: 'Error',
+            message: 'Error. Something went wrong',
           }),
         })
       );


itvi-1234:
Hi @illume , I think the PR looks fine , Lemme know if any changes needed

frontend/src/redux/clusterActionSlice.ts
```diff
@@ -289,7 +298,7 @@ export const executeClusterAction = createAsyncThunk(
             }
           }
         } else {
-          dispatchError();
+          dispatchError(err as Error);
```
illume: Can we remove the "err as Error" here?

It's turning the type checker off, so we usually don't use as... unless there's a good reason?

illume:
ps. I'm not sure about how we do translation here of these incoming messages. But maybe we can try and solve this later.

frontend/src/redux/clusterActionSlice.ts
itvi-1234: Thanks for catching that! I've removed the unsafe type assertion (as Error) and replaced it with a proper instanceof Error check. This ensures type safety while correctly surfacing the admission error messages. Updated in the latest push.

itvi-1234:
Agreed. Translating dynamic server-side admission messages is indeed challenging, as we do not have direct control over the keys returned by the API yet. Maybe ,  we can revisit this in a follow-up task once a more structured pattern for localizing server-side errors is established.

itvi-1234:
Hi @illume I have dis the suggested changes , could you please review

sniok (approved):
Tried it and works nicely! Thanks

Image

It's a bit hard to read such a big error message in our toasts but that's another issue and it's definitely better than having just a generic error message.

Issue #5036, opened 2026-04-02
Show per Node total resource requests and limits


peter-facko:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
I want to see the resource allocation on my Nodes.

## Describe the solution you'd like
In the Node detail view, I would like to have the total number of resource requests and limits. Similar to resource usage we currently have.

## What users will benefit from this feature?
Anyone wanting to monitor resource restrictions on their cluster.

## Are you able to implement this feature?
No.
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

itvi-1234:
Nice suggestion @peter-facko , Lemme work on it 🚀

itvi-1234:
Hi @peter-facko , I have added the resouce allocation section as per ur requirement , Lemme know if any changes needed

<img width="1803" height="964" alt="Image" src="https://github.com/user-attachments/assets/da0f0c1e-9c9a-401d-b079-dffd0b637d09" />

peter-facko:
Hello @itvi-1234, that looks amazing!

Maybe a pie or bar chart would be a nicer visualization and could also display the request and limit at once, but this already covers my use case, so thank you very much.

itvi-1234:
Welcome @peter-facko , THanks for the feedback , Let this feature got merged into the main , then I will also try to look into the ways of implementing pie or bar chart there :rocket:

peter-facko:
I use this feature in 0.42 so I am closing this.

PR #5048, opened 2026-04-04
frontend: nodes: Add resource allocation summary in Node Details

itvi-1234:
**Summary**
The Node details view currently only shows actual resource usage (metrics). It lacks a summary of the total resource requests and limits allocated to the pods running on that node.

This PR adds a new "Resource Allocation" section to the Node Details view, providing a clear summary of total CPU/Memory requests and limits, along with their percentage relative to the node's capacity.

Fixes Issue #5036 

**Changes**
1.Added a new AllocatedResourcesSection component to NodeDetails.tsx.
2.Implemented logic to calculate the total CPU/Memory requests and limits from all pods assigned to the node.
3.Integrated the new section as a full-width block in extraSections for consistent UI alignment with other details.
4.Included color-coded status labels for quick visual monitoring of allocation levels.

**Steps to Test**
1.Navigate to any Node's details page.
2.Locate the new "Resource Allocation" block below the usage charts.
3.Verify that the totals and percentages correctly reflect the assigned pod resources

**Screenshot** (Added the new resource allocation section)
image


itvi-1234:
Hi @ashu8912 , @vyncent-t , @sniok Lemme know if any changes needed

itvi-1234:
Hey @illume , I have implemented the suggested changes , lemme know if further required

itvi-1234:
Hey , I have updated the implementation to address all the feedback provided in the previous comments. I have force-pushed the changes, though some review comments may not have been automatically marked as 'outdated' by copilot yet.

Summary of improvements:
1.Multi-cluster scoping: Updated Pod.useList to use the node's cluster context.
2.Init containers: Included init container resources in the allocation totals using the Kubernetes standard formula.
3.Terminal pods: Added a filter to exclude pods in Succeeded or Failed phases.
4.Calculation safety: Implemented a guard against division by zero for node capacity.
5.Formatting: Standardized the display to use a space between values and units (${value} ${unit}) to match existing UI patterns.

itvi-1234:
Hi @sniok , @illume , Could you please review this PR

sniok (changes requested):
hi, could you please resolve merge conflicts and address the comments

itvi-1234:
Hi @sniok , I've rebased this PR onto the latest `main` and also addressed 
all the Copilot review suggestions:

- **Multi-cluster scoping**: `Pod.useList()` now passes `cluster: node?.cluster` to ensure pods are fetched from the 
  correct cluster in multi-cluster setups.

- **Terminal pod filtering**: Pods in `Succeeded` or `Failed` phase are now skipped before summing CPU/memory requests and limits, preventing inflated allocation totals.

- **Division-by-zero guard**: `getPercentage()` now returns `'0'` when capacity is `0`, avoiding `Infinity` / `NaN` in the UI.

- **Value + unit formatting**: Resource values are rendered with a space between the numeric value and unit (e.g. `500 m`) for consistency with the rest of the UI.

Please let me know if anything else needs to be addressed. Thanks!

frontend/src/components/node/Details.tsx
```diff
@@ -367,6 +373,140 @@ function ChartsSection(props: ChartsSectionProps) {
...
+  let cpuRequests = 0;
+  let cpuLimits = 0;
+  let memoryRequests = 0;
+  let memoryLimits = 0;
+
+  pods?.forEach((pod: KubePod) => {
+    const phase = pod.status?.phase;
+    if (phase === 'Succeeded' || phase === 'Failed') {
+      return;
+    }
```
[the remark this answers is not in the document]
sniok: yeah we already fetch pods with this fieldSelector `spec.nodeName=${name},status.phase!=Succeeded,status.phase!=Failed`  in NodeDetails so we can just reuse that list here and as a nice side effect avoid the `if (phase === 'Succeeded' || phase === 'Failed') {` check

itvi-1234:
Hi @sniok You are totally spot on! 

I have refactored `AllocatedResourcesSection` to accept the `nodePods` array as a prop directly from the parent `NodeDetails` component. This eliminates the duplicate API call entirely.

And yes, as you cleverly pointed out, since the parent's `Pod.useList()` already provides a list pre-filtered with `status.phase!=Succeeded,status.phase!=Failed` inside the field selector, I was able to completely drop the manual `if (phase === 'Succeeded' || phase === 'Failed')` client-side filtering block inside the capacity loop.

Thanks for the excellent catch! I've updated the PR. Let me know if everything looks good!

frontend/src/components/node/Details.tsx
```diff
@@ -367,6 +373,130 @@ function ChartsSection(props: ChartsSectionProps) {
...
+
+  const cpuCapacity = units.parseCpu(node?.status.capacity?.cpu || '0');
+  const memoryCapacity = units.parseRam(node?.status.capacity?.memory || '0');
+
+  let cpuRequests = 0;
+  let cpuLimits = 0;
+  let memoryRequests = 0;
+  let memoryLimits = 0;
+
+  pods?.forEach((pod: KubePod) => {
```
sniok: oh one more thing, since this can be an expensive calculation if node has a lot of resources, lets put this in a useMemo to avoid recalculating this on every render

frontend/src/components/node/Details.tsx
```diff
@@ -367,6 +373,130 @@ function ChartsSection(props: ChartsSectionProps) {
...
+      podMemoryLimits = Math.max(podMemoryLimits, initMemLimit);
+    });
+
+    cpuRequests += podCpuRequests;
+    cpuLimits += podCpuLimits;
+    memoryRequests += podMemoryRequests;
+    memoryLimits += podMemoryLimits;
+  });
+
+  const getPercentage = (value: number, capacity: number) => {
```
sniok: this function doesn't use anything from the closure so it's safe to move it outside of the component

itvi-1234: Done! Moved it to the outer module scope.

frontend/src/components/node/Details.tsx
itvi-1234: Good catch! I have wrapped the resource allocation summation logic inside a useMemo block with pods as the dependency array. It will smoothly avoid re-calculating parsed capacity boundaries on every render now.

itvi-1234:
hey @sniok , lemme know if any further changes required :smiley:

sniok:
hmm something doesn't look right, values seem to align with kubectl output, but percentages are different

image

image

I also think adding a `Total limits may be over 100 percent, i.e., overcommitted.` message (like kubectl does) would be a good idea

itvi-1234:
You are perfectly right. @sniok  `kubectl describe node` calculates the percentages against the node's `Allocatable` resources (ignoring the kube reserved amounts), whereas the component was previously calculating against the total absolute `Capacity`.
I have updated the calculation to use `node.status.allocatable` resources fallbacking to `capacity` if absolutely needed. The percentages should completely map to CLI output now.
I've also added the `Total limits may be over 100 percent, i.e., overcommitted.` hint on top of the section to keep the UX transparent, echoing `kubectl`.
Thanks a lot for the review! I've pushed the fix. Let me know what you think. :smiley:

itvi-1234:
Hi @sniok , any views regarding the change

sniok (approved):
tested and looks good, thank you!

/lgtm

itvi-1234:
Thank you too @sniok for guiding me through it :rocket:

PR #5059, opened 2026-04-05
frontend: pods: Fix status icon stuck on warning after pod becomes ready

### LINKED ISSUE #4905: Pod status stuck with warning icon despite being Running/Completed and healthy — WebSocket events received but UI not reflecting state changes
## Describe the bug

When viewing the Pods list, pod status visual indicators do not update correctly in real time despite WebSocket events being received. This manifests in two scenarios:

1. **Running pods:** A pod that has already passed its health checks and is fully `Running` remains displayed with a yellow warning icon (as if it were still initialising). The status text shows "Running" but the visual indicator does not update to reflect a healthy state.

2. **Completed pods (Jobs/CronJobs):** Pods created by Jobs that have finished successfully and are in `Completed` state also display a warning icon, even though they terminated with exit code 0 and the Job itself reports success. This specifically occurs when the pod was created and completed while the user was already viewing the Pods list for that namespace. Pods that completed before navigating to the view display correctly on load.

The visual state can be corrected by either **manually refreshing the page** or **navigating away to another section and back** to the Pods view. Other actions such as deleting or evicting a different pod do not trigger a refresh of the affected pods' visual state. Furthermore, if the deleted pod is part of a Deployment and a new pod is created to replace it, the new pod also ends up stuck with the warning icon after passing its health checks — just like the rest.

**Important:** when opening an individual pod's detail view (by clicking on a pod), the status and conditions **do update correctly** in real time. The problem is isolated to the **pod list view**.

**What we verified:**

1. WebSocket connections are established correctly (`pods?watch=1`, status 101 Pending).
2. Events (`MODIFIED`, `ADDED`, `DELETED`) are flowing through the WebSocket — confirmed via browser DevTools → Network → Socket → Messages.
3. Real-time metrics (CPU and memory) **do update** correctly in the pod list without manual refresh.
4. The pod status visual indicator (icon/color) does **not** update despite the watch events arriving.
5. Deleting/evicting an unrelated pod does **not** cause other pods' status indicators to refresh. New pods spawned by Deployments after a delete also get stuck with the warning icon.

This suggests the frontend is receiving the watch data but the pod list view is not re-rendering the pod status/conditions correctly, while it does process the metrics data properly and the individual pod detail view handles state updates as expected.

**Expected behavior:**

- Pod status icon and visual state should update in real time as watch events arrive, just like CPU and memory metrics do.
- `Running` pods that have passed health checks should show a healthy (green) indicator without requiring a manual refresh.
- `Completed` pods from successful Jobs should show a neutral or success indicator, not a warning.

### Scenario 1 — Running pods
1. Go to the Pods view in Headlamp for a given namespace.
2. Trigger a pod restart or new deployment.
3. Wait for the pod to reach `Running` with all containers ready (confirm via `kubectl get pod`).
4. See that the pod still shows a yellow warning icon in the list despite being healthy.
5. Refresh the page or navigate away and back — the correct status is now shown.

### Scenario 2 — Completed pods
1. Go to the Pods view for a namespace and stay on it.
2. Run a Job (or wait for a CronJob to trigger) while remaining on the Pods view.
3. Wait for the Job to complete successfully (confirm via `kubectl get pods`).
4. See that the resulting pod shows a warning icon despite being in `Completed` state with exit code 0.
5. Refresh the page or navigate away and back — the correct status is now shown.

### Scenario 3 — Delete does not fix other pods
1. With several pods already showing the incorrect warning icon, delete or evict an unrelated pod.
2. See that the deleted pod is removed from the list correctly.
3. If the deleted pod belongs to a Deployment, see that the replacement pod also gets stuck with the warning icon after becoming healthy.
4. See that the previously affected pods remain unchanged — still showing the warning icon.

- Installation type: Helm (in-cluster)
- Headlamp Version: 0.40.1 (helm chart 0.40.1)
- Browsers tested: Brave (Chromium-based), Google Chrome, Safari
- Access: via ingress-nginx (correctly configured to keep the webSockets which works correctly for many other data updates)

No.

## Additional Context
The problem appears to be a rendering issue in the pod list view specifically. The individual pod detail view updates status and conditions correctly in real time, and metrics (CPU/memory) in the list view also update without issues. Only the status icon/indicator in the list view fails to re-render when watch events arrive.

It is unknown whether this is a regression, as this is the first deployment of Headlamp in our environment.

nidhi-9900:
## Summary

The pod status icon in the list view was getting stuck on warning even after the pod was Running and healthy. It only fixed itself after a page refresh. The issue was that getValue was only returning the reason string ("Running"), which stayed the same even when the Ready condition changed : so the table never re-rendered. Added the Ready condition status to the return value so it picks up the change.

## Related Issue

Fixes #4905

## Changes

- Updated `getValue` in the status column in `frontend/src/components/pod/List.tsx` to return both the reason and the Ready condition status
- Earlier it only returned the reason string which was "Running" in both warning and success states, so nothing triggered a re-render

## Steps to Test

1. Start Headlamp locally
2. Go to Workloads → Pods
3. Run `kubectl create deployment nginx-test --image=nginx`
4. Watch the pod without refreshing the page
5. Icon should turn green when pod becomes ready

## Notes for the Reviewer

- Only `List.tsx` is changed
- Snapshot files changed too but that's just MUI CSS class names, nothing to do with this fix

### DIFF
--- a/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartNoMetrics.stories.storyshot
+++ b/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartNoMetrics.stories.storyshot
@@ -1,7 +1,7 @@
 <body>
   <div>
     <div
-      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-1oobngp-MuiPaper-root"
+      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-u96y7l-MuiPaper-root"
     >
       <div
         class="MuiBox-root css-qrw4x1"
--- a/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartPod.stories.storyshot
+++ b/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartPod.stories.storyshot
@@ -1,7 +1,7 @@
 <body>
   <div>
     <div
-      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-1oobngp-MuiPaper-root"
+      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-u96y7l-MuiPaper-root"
     >
       <div
         class="MuiBox-root css-qrw4x1"
--- a/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartWithMetrics.stories.storyshot
+++ b/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.CpuChartWithMetrics.stories.storyshot
@@ -1,7 +1,7 @@
 <body>
   <div>
     <div
-      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-1oobngp-MuiPaper-root"
+      class="MuiPaper-root MuiPaper-outlined MuiPaper-rounded css-u96y7l-MuiPaper-root"
     >
       <div
         class="MuiBox-root css-qrw4x1"
--- a/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.MemoryChartNoMetrics.stories.storyshot
+++ b/frontend/src/components/cluster/Charts/__snapshots__/ResourceCharts.MemoryChartNoMetrics.stories.storyshot
@@ -1,7 +1,7 @@
 <body>
   <div>
     <div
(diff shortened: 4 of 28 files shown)


sniok (changes requested):
thanks for the PR, the code looks good but could you make sure not to include a merge commit

nidhi-9900:
Hi, I've removed the merge commit. Please let me know if anything else needs to be changed.

sniok (approved):
looks good, thanks!

J4NS-R:
Thank you @nidhi-9900 !

PR #5061, opened 2026-04-06
frontend: Replace ts-ignore with proper type fixes

skoeva:
These changes replace `@ts-ignore` suppressions with proper fixes across three files:
- `KubeObject.ts` - targeted function casts on `delete` and `patch` spread calls
- `hooks.ts` - `Object.assign` tuple pattern to satisfy the intersection return type
- `graphLayout.tsx` - added missing `label?` field to `ElkEdgeWithoutData`

### DIFF
--- a/frontend/src/lib/k8s/KubeObject.ts
+++ b/frontend/src/lib/k8s/KubeObject.ts
@@ -497,8 +497,11 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
       params.gracePeriodSeconds = 0;
     }
 
-    // @ts-ignore
-    return this._class().apiEndpoint.delete(...args, params, this._clusterName);
+    return (this._class().apiEndpoint.delete as (...args: any[]) => Promise<any>)(
+      ...args,
+      params,
+      this._clusterName
+    );
   }
 
   update(data: KubeObjectInterface) {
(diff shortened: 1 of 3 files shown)


ashu8912 (approved):
Nice one, LGTM!!

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -497,8 +497,11 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
       params.gracePeriodSeconds = 0;
     }
 
-    // @ts-ignore
-    return this._class().apiEndpoint.delete(...args, params, this._clusterName);
+    return (this._class().apiEndpoint.delete as (...args: any[]) => Promise<any>)(
```
illume, line 500: I don't think these should be "any"?

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -547,8 +550,11 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
 
     args.push(this.getName());
 
-    // @ts-ignore
-    return this._class().apiEndpoint.patch(...args, {}, this._clusterName);
+    return (this._class().apiEndpoint.patch as (...args: any[]) => Promise<any>)(
```
illume, line 553: Also, the any here I think is wrong. 

(It is a bit better than using ts-ignore)

PR #5063, opened 2026-04-06
frontend: ClusterSettings page refactor

### LINKED ISSUE #5062: 'Cluster Settings' page is broken when switching between clusters
When using `Cluster` dropdown in the Cluster Settings page, settings values are incredibly broken, sometimes appearing, sometimes disappearing and overriding cluster settings with values from previously selected cluster

## To Reproduce
Steps to reproduce the bug:
1. Have at least two clusters in headlamp
2. Go to Cluster Settings page
3. Go to cluster1 and set a setting like `allowed namespace`
4. Switch to cluster2
5. Switch back to cluster1
6. Repeat steps 4,5 
7. Observe that the allowed namespace value sometimes is missing from cluster1, sometimes appears in cluster2
8. Observe the localStorage values in the DevTools when you switch between clusters using the dropdown. Cluster Settings values in local storage are updated when just switching between clusters. This shouldn't cause any writes or updates!

## Demo

Video where I change allowed namespace in one cluster and then just toggle between two clusters

https://github.com/user-attachments/assets/21707fa0-69bd-457c-b652-36db9f316597

yeah

## notes

This is caused by duplicated state, lack of single source of truth, broken rules of hooks (we should REALLY enable those eslint rules) and ad-hoc syncing with localStorage. The page really needs a refactor.

sniok:
Fixes #5062 

Simplify cluster settings by having SettingsCluster own the settings state and pass it down to child components, removing duplicated load/store logic

 - Added useClusterSettings hook for loading/storing cluster settings from localStorage
 - SettingsCluster now passes settings to NodeShellSettings, PodDebugSettings, and ClusterNameEditor as props
 - Appearance changes (accent color, icon) apply immediately, consistent with all other settings
 - Removed debouncing (localStorage writes are fast) and per-component localStorage calls

This also reduces amount of eslint warnings in the src/components/App/Settings folder from 23 to 10

## Steps to Test

1. Have at least two clusters in headlamp
2. Go to Cluster Settings page
3. Go to cluster1 and set a setting like allowed namespace
4. Switch to cluster2
5. Switch back to cluster1
6. Repeat steps 4,5


Athang69:
@sniok  ig it the issue number #5062
You mentioned #5060 in PR description

sniok:
whoops, thanks for pointing that out

joaquimrocha (approved):
Seems like the second commit could be split in a couple, but it also works like this. I am approving so feel free to rebase + merge.

illume (changes requested):
Thanks for these changes.


illume (approved):
🎉 thanks!

vyncent-t (approved):
works on my end lgtm

PR #5075, opened 2026-04-07
frontend: auth.test: Replace 'as any' with strictly typed 'as Response'

itvi-1234:
##Summary##

This PR resolves TypeScript technical debt in the authentication tests by replacing loose as any type bypasses with strict Partial<Response> interfaces. This ensures our `mockBackendFetch` expectations are strongly bound to the native Response interface, preventing future silent false-positive test passes if the backend's expected return shape changes.


itvi-1234:
hi @illume , @sniok , @vyncent-t  Could you please review this PR , Lemme know if any changes needed

sniok (approved):
thanks!

/lgtm

PR #5080, opened 2026-04-07
 docs: tutorials: Add tutorial for building a WhoAmI plugin

yolossn:
Add a WhoAmI plugin tutorial using the SelfSubjectReview API to
 display user identity in the cluster sidebar.


docs/tutorials/plugin-development/whoami-plugin/index.md
```diff
@@ -0,0 +1,554 @@
...
+// Type for the SelfSubjectReview response
+interface UserInfo {
+  username: string;
+  uid?: string;
+  groups?: string[];
+  extra?: Record<string, string[]>;
+}
+
+function WhoAmIPage() {
+  const cluster = getCluster();
```
sniok, line 180: here a `useSelectedClusters()` hook should be used instead 
we can either demonstrate how to do whoami lookup for multiple clusters or explicitly say in a comment that we're only going to use one

I feel like multicluster feature sometimes gets forgotten even when we develop core features so I think it's important that we provide an example that takes it into consideration for plugin development

docs/tutorials/plugin-development/whoami-plugin/index.md
```diff
@@ -0,0 +1,554 @@
...
+  const [userInfo, setUserInfo] = useState<UserInfo | null>(null);
```
sniok: minor nit but I find using `undefined` for useState a bit cleaner

```tsx
const [userInfo, setUserInfo] = useState<UserInfo>();
```

resulting type for `userInfo` will be `UserInfo | undefined`

sniok (changes requested):
overall looks good, a really nice usecase for a simple plugin showcase!

the only concern I have is one comment I've left about multicluster support

docs/tutorials/plugin-development/whoami-plugin/index.md
yolossn: Agree, I will add one more section to the tutorial discussing this will fail for multicluster and to support that we should use the useSelectedClusters hook

sniok (approved):
thanks!

/lgtm

PR #5081, opened 2026-04-07
SECURITY.md: Update to use kubernetes-sigs SECURITY.md

illume:
Note, keeping the file in the repo like minikube does so it
is easier to find for people browsing via github files or via
a checked out copy of the repo.

https://github.com/kubernetes-sigs/.github/blob/master/SECURITY.md

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,7 +1,22 @@
 # Security Policy
 
+## Security Announcements
+
+Join the [kubernetes-security-announce] group for security and vulnerability announcements.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,7 +1,22 @@
 # Security Policy
 
+## Security Announcements
+
+Join the [kubernetes-security-announce] group for security and vulnerability announcements.
```
sniok, line 5: TIL you could reference links like this in markdown 😮

sniok (approved):
thanks!

PR #5085, opened 2026-04-07
frontend: cronjob: Handle undefined spec.suspend in List and Details

Athang69:
## Summary
This PR fixes a crash in the CronJob list and details pages when `spec.suspend` is not explicitly set in the CronJob manifest. Kubernetes defaults `spec.suspend` to `false` but does not always include it in the API response, causing `.toString()` to be called on `undefined` and throwing a `TypeError`

## Related Issue

Fixes #5083

## Changes
- Fixed `frontend/src/components/cronjob/List.tsx` added `?? false` fallback before `.toString()` on `spec.suspend `in the Suspend column `getValue`
- Fixed `frontend/src/components/cronjob/Details.tsx` added `?. `optional chaining on `spec` and `?? false` fallback for `isCronSuspended` derivation
- Fixed `frontend/src/components/cronjob/Details.tsx` added `?? false` fallback before `.toString()` on `spec.suspend` in the `extraInfo` Suspend field

## Steps to Test

1. Apply a CronJob manifest that omits `spec.suspend` e.g. a Helm generated CronJob
2. Navigate to Workloads -> CronJobs
3. Verify the list renders correctly and shows `false` in the Suspend column without crashing
4. Click on the CronJob to open its details page
5. Verify the details page renders correctly and shows `false` for the Suspend field
6. Repeat with a CronJob that has `spec.suspend:true` explicitly set and verify it still shows `true` correctly

## Notes for the Reviewer
`spec.suspend` is defined as optional in the Kubernetes `batch/v1 CronJobSpec` its absence from the API response is valid behaviour, not a bug in the cluster
The `?? false` fallback exactly matches the Kubernetes-defined default for this field, so there is no behaviour change for CronJobs that have the field explicitly set
The additional `?.` on `spec` in `isCronSuspended` guards against the loading phase before the API response arrives, which the original `cronJob?.spec.suspend` did not cover

### DIFF
--- a/frontend/src/components/cronjob/Details.tsx
+++ b/frontend/src/components/cronjob/Details.tsx
@@ -149,7 +149,7 @@ export default function CronJobDetails(props: {
   const { items: jobs, errors } = Job.useList({ namespace, cluster: cronJob?.cluster });
   const [isSpawnDialogOpen, setIsSpawnDialogOpen] = useState(false);
   const [isPendingSuspend, setIsPendingSuspend] = useState(false);
-  const isCronSuspended = cronJob?.spec.suspend;
+  const isCronSuspended = cronJob?.spec?.suspend ?? false;
 
   const ownedJobs = useMemo(
     () =>
@@ -238,7 +238,7 @@ export default function CronJobDetails(props: {
           },
           {
             name: t('translation|Suspend'),
-            value: item.spec.suspend.toString(),
+            value: (item.spec?.suspend ?? false).toString(),
           },
           {
             name: t('Starting deadline'),
--- a/frontend/src/components/cronjob/List.tsx
+++ b/frontend/src/components/cronjob/List.tsx
@@ -66,7 +66,7 @@ export default function CronJobList() {
         {
           id: 'suspend',
           label: t('translation|Suspend'),
-          getValue: cronJob => cronJob.spec.suspend.toString(),
+          getValue: cronJob => (cronJob.spec?.suspend ?? false).toString(),
           gridTemplate: 0.6,
         },
         {


kannon92:
Do you know why there is such a big difference in the code for Job and CronJob?

I wanted to check if there was a similar bug for Job but I don't see us using suspend at all in that code path.

Athang69:
Hey @kannon92 good catch on the Job side. I checked the Job components and Job does not expose a `suspend` field in the list or details views at all, which is why there is no similar bug there.
The difference comes down to how the two resources use `suspend`. For CronJob, `spec.suspend` is a first class control that stops the scheduler from firing new jobs, so it makes sense to surface it in the UI as an actionable field (we even have the suspend/resume button wired to it). For Job, `spec.suspend` exists in the Kubernetes API but it is more of an advanced/low level knob, mainly used by higher level controllers like JobSet, and Headlamp simply does not render it in the UI currently.

With that said, if someone were to add `spec.suspend` to the Job list or details view in the future, they would hit the exact same bug since the field is optional in the API response for Jobs as well. Might be worth a note somewhere or a shared utility like `getSuspendValue(resource)` that handles the undefined case in one place rather than repeating the `?? false` pattern across different resource types.

sniok (changes requested):
thanks for the fix. the code looks good but could you please update commit message so it follows our guidelines? https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines

Athang69:
@sniok Yeah edited the commit message according to guidelines. Thanks

Athang69:
Hey @illume @sniok, sorry for the confusion when I amended the commit message earlier, some unrelated files accidentally got included in the push. I've now reset back to the original clean commit and only updated the message to follow the contribution guidelines. The PR now contains only the two CronJob files as intended. Sorry for the noise!

frontend/src/components/cronjob/Details.tsx
```diff
@@ -238,7 +238,7 @@ export default function CronJobDetails(props: {
           },
           {
             name: t('translation|Suspend'),
-            value: item.spec.suspend.toString(),
+            value: (item.spec.suspend ?? false).toString(),
```
sniok: this part is still reading `.suspend` from the `.spec` that might be undefined, it should use `?.` like you've done above in line 152

the original reported issue is `TypeError: can't access property "suspend", c.spec is undefined`

same applies to the List.tsx change

sniok (changes requested):

frontend/src/components/cronjob/Details.tsx
Athang69: @sniok Good catch! Fixed both `Details.tsx` and `List.tsx` to use `item.spec?.suspend ?? false` with optional chaining on `spec` as well. Thanks!
Can you re-review?

sniok (approved):
thanks!

/lgtm

PR #5095, opened 2026-04-08
backend: Replace context.TODO() with explicit background context in long-running goroutine

itvi-1234:
## Summary
This PR cleans up technical debt in the Golang backend inside the `drainNode` asynchronous API handler. It replaces repeated `context.TODO()` usages with the correctly initialized, explicit `context.Background()` wrapper variable `ctx` that the goroutine actually instantiated but failed to pass downstream. 

## Changes
- `backend/cmd/headlamp.go`: Inside the `drainNode` goroutine, replaced `context.TODO()` in Kubernetes `nodeClient` and `clientset.CoreV1().Pods()` calls with the available `ctx` variable to explicitly signify that this work outlives the HTTP request scope.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -2553,7 +2553,7 @@ func (c *HeadlampConfig) drainNode(clientset *kubernetes.Clientset, nodeName str
 		cacheKey := uuid.NewSHA1(uuid.Nil, []byte(nodeName+cluster)).String()
 		cacheItemTTL := DrainNodeCacheTTL * time.Minute
 
-		node, err := nodeClient.Get(context.TODO(), nodeName, v1.GetOptions{})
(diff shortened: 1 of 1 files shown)


itvi-1234:
Hi @sniok  @vyncent-t @skoeva This is a small code cleanup. It replaces the temporary context.TODO() placeholders with the proper ctx (context.Background()) variable that was already defined at the start of the function.

Because Node Draining takes a long time, it runs in the background. Using the correct background context ensures the draining process safely finishes its job, even if the user accidentally closes their browser before it completes.  Could you please review this PR , lemme know if any changes needed :rocket:

itvi-1234:
Hi @illume , @sniok The PR looks fine , lemme know if any changes needed

itvi-1234:
Hi @sniok , any directions regarding this PR

skoeva (approved):
makes sense, thanks for this!

PR #5097, opened 2026-04-08
Update dependencies using npm audit

sniok:
the only audit warnings that are left are about the `elliptic` issue that doesn't have a fix yet

the only breaking change is upgrade of `electron` version. this needs to be tested
I've looked through the [breaking changes](https://www.electronjs.org/docs/latest/breaking-changes) and the only thing that we should update is navigation methods like goBack. But since they're not removed yet and just marked as deprecated we can do it in a separate PR.
Other breaking changes do no affect us.

Tested electron app on linux, tested on windows as well


illume (approved):
🎉 thanks!

PR #5100, opened 2026-04-09
frontend: Storybook: fix ResourceTable Name Search filter

### LINKED ISSUE #5082: frontend: ResourceTable Name Search story is broken
NameSearch story is supposed to demonstrate that the table will search using the `search` value but it doesn't actually do that. 

Image

Screenshot of this story with an empty search input

## To Reproduce

1. npm run storybook in frontend/
2. Open 'Name Search' story
3. Observe empty input

Maybe later but if anyone else wants to take this feel free to do so

## Additional Context
This is a logic error in the story, it sets the search to the redux store, which ResourceTable doesn't use
ResourceTable also accepts defaultGlobalFilter prop that is forwards to the Table component but that also doesn't work after changes made in #3354, global filter is read from the query param

Vikaspal8923:
## Summary

This PR fixes the `ResourceTable` **Name Search** Storybook example so the search input is actually prefilled and the table filters as intended.

## Related Issue

Fixes #5082

## Changes

- Updated `frontend/src/components/common/Resource/ResourceTable.stories.tsx`
- Switched the story from Redux-only search state to the URL query param that `ResourceTable` actually reads
- Enabled `reflectInURL` on the `ResourceTable` story so the global filter is applied correctly
- Kept the existing example data and story structure intact

## Steps to Test

1. Run `npm run frontend:storybook`
2. Open `ResourceTable` in Storybook
3. Select the `Name Search` story
4. Confirm the search box is prefilled with `mypod3`
5. Confirm the table is filtered to the matching row only

Screenshot from 2026-04-09 15-57-09

- Storybook screenshot showing the `Name Search` story with the search field populated and filtered results

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.stories.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.stories.tsx
@@ -47,14 +47,13 @@ export default {
 const TemplateWithFilter: StoryFn<{
   resourceTableArgs: ResourceTableFromResourceClassProps<typeof MyPod>;
   namespaces: string[];
-  search: string;
 }> = args => {
-  const { resourceTableArgs, search, namespaces = [] } = args;
+  const { resourceTableArgs, namespaces = [] } = args;
 
   const storeWithFilterAndSettings = configureStore({
     reducer: (
       state = {
-        filter: { namespaces: new Set<string>(), search: '' },
+        filter: { namespaces: new Set<string>() },
         config: { settings: { tableRowsPerPageOptions: [10, 20, 50, 100] } },
         ui: { ...uiSlice.getInitialState() },
         drawerMode: { isDetailDrawerEnabled: false },
@@ -65,7 +64,6 @@ const TemplateWithFilter: StoryFn<{
       ui: { ...uiSlice.getInitialState() },
       filter: {
         namespaces: new Set(namespaces),
-        search,
       },
       config: {
         settings: {
@@ -174,17 +172,17 @@ const withHiddenCols: ResourceTableFromResourceClassProps<typeof MyPod> = {
 export const NoFilter = TemplateWithFilter.bind({});
 NoFilter.args = {
   resourceTableArgs: podData,
-  search: '',
 };
 
 export const NameSearch = TemplateWithFilter.bind({});
 NameSearch.args = {
-  resourceTableArgs: podData,
-  search: 'mypod3',
+  resourceTableArgs: {
+    ...podData,
+    defaultGlobalFilter: 'mypod3',
+  },
 };
 
 export const WithHiddenCols = TemplateWithFilter.bind({});
 WithHiddenCols.args = {
   resourceTableArgs: withHiddenCols,
-  search: '',
 };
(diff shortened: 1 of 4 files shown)


sniok:
while this technically fixes the story via url search parameters, this doesn't address the root cause of `defaultGlobalFilter` prop being ignored in the ResourceTable component

I'm not against of getting this fix in but I think it would be better to address the underlying issue, what do you think @Vikaspal8923 ?

Vikaspal8923:
Sure, I will try to find out the actual root cause.

Vikaspal8923:
@sniok  can you take a look now. 
Screenshot from 2026-04-10 12-46-44

frontend/src/components/common/Table/Table.tsx
```diff
@@ -196,10 +196,16 @@ export default function Table<RowItem extends Record<string, any>>({
   const prefix = reflectInURL === true ? '' : reflectInURL || '';
   const [page, setPage] = usePageURLState(shouldReflectInURL ? 'p' : '', prefix, initialPage);
   const filterKey = prefix ? `${prefix}filter` : 'filter';
-  const [globalFilter, setGlobalFilter] = useQueryParamsState<string | undefined>(
+  const [localGlobalFilter, setLocalGlobalFilter] = useState<string | undefined>(
```
sniok: `localGlobalFilter` is quite a confusing name for this state. 'local' is a complete opposite to 'global' and they're used together. maybe something like `stateGlobalFilter` or `filterState`/`filterQueryParam`? it'd be nice to also have a comment explaining why we need two separate states

frontend/src/components/common/Resource/ResourceTable.stories.tsx
```diff
@@ -179,8 +179,11 @@ NoFilter.args = {
 
 export const NameSearch = TemplateWithFilter.bind({});
 NameSearch.args = {
-  resourceTableArgs: podData,
-  search: 'mypod3',
+  resourceTableArgs: {
+    ...podData,
+    defaultGlobalFilter: 'mypod3',
+  },
+  search: '',
```
sniok: I don't think we need search anymore? If it's not used

sniok (changes requested):

frontend/src/components/common/Table/Table.tsx
Vikaspal8923: I think we need two state because of shouldReflectInURL , if it is true then we need to sync the url query with filter as well. and when it false then we don't need to sync it with url query , so for that we need a plain react state

sniok: I understand, I meant that the comment should be in the code somewhere

frontend/src/components/common/Resource/ResourceTable.stories.tsx
sniok: the search field is still present

Vikaspal8923: done :+1:

frontend/src/components/common/Resource/ResourceTable.stories.tsx
```diff
@@ -65,7 +64,7 @@ const TemplateWithFilter: StoryFn<{
       ui: { ...uiSlice.getInitialState() },
       filter: {
         namespaces: new Set(namespaces),
-        search,
+        search: '',
```
sniok: the search field doesn't exist in the filter slice

Vikaspal8923: done :+1:

sniok (approved):
thanks!

/lgtm

PR #5135, opened 2026-04-12
frontend: Add missing Pi and P memory unit support to normalizeUnit

### LINKED ISSUE #5134: frontend: normalizeUnit missing Pi (Pebibyte) and P (Petabyte) memory unit support
### Description

The `normalizeUnit` function in `frontend/src/lib/util.ts` is missing support for `Pi` (pebibyte) and `P` (petabyte) memory units, despite them being documented as valid Kubernetes resource quantity suffixes in the function's own comments.

At [line 499-502](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/util.ts#L499-L502), the comment documents all valid units:

```
Decimal: m | n | "" | k | M | G | T | P | E
Binary:  Ki | Mi | Gi | Ti | Pi | Ei
```

However, the `if-else` chain that follows (lines 505-531) only implements:
- **Binary:** `Ki`, `Mi`, `Gi`, `Ti`, `Ei` — **`Pi` is missing**
- **Decimal:** `m`, `u`, `n`, `k`, `M`, `G`, `T`, `E` — **`P` is missing**

Additionally, this causes the `Ei` and `E` multipliers to be incorrect:
- `Ei` currently uses `1024^5` (which is actually the `Pi` multiplier). It should be `1024^6`.
- `E` currently uses `1000^5` (which is actually the `P` multiplier). It should be `1000^6`.

**Impact:**
- Any Kubernetes resource reporting memory in `Pi` or `P` units will display an incorrect value — the raw integer with no unit conversion applied.
- `Ei` and `E` values are also off by a factor of 1024x and 1000x respectively.
- This affects the ResourceQuota display in the Headlamp frontend where `normalizeUnit` is used.

### Steps to Reproduce
1. Create a ResourceQuota or LimitRange with a memory limit using `Pi` or `P` suffix (e.g., `1Pi`).
2. View the resource in Headlamp.
3. The displayed value will be `1 Bytes` instead of `~1.13 PB`.

### Expected Behavior
All units documented in the function's comment block should be correctly handled:
- `1Pi` → `~1.13 PB`
- `1P` → `1 PB`
- `1Ei` → `~1.15 EB` (not `~1.13 PB` as currently displayed)
- `1E` → `1 EB` (not `1 PB` as currently displayed)

KumarADITHYA123:
Fixes #5134 

### Summary

The `normalizeUnit` function was missing support for `Pi` (pebibyte) and `P` (petabyte) memory units, despite them being documented in the function's own comments as valid Kubernetes quantity suffixes. This also caused `Ei` and `E` to use incorrect multipliers (off by one order of magnitude).

### Changes

- **`frontend/src/lib/util.ts`:** Added `Pi` and `P` branches to the memory unit `if-else` chain. Corrected `Ei` from `1024^5` to `1024^6` and `E` from `1000^5` to `1000^6`.
- **`frontend/src/lib/util.test.ts`:** Added comprehensive test suite for `normalizeUnit` covering all binary units (Ki, Mi, Gi, Ti, Pi, Ei), all decimal units (k, M, G, T, P, E), edge cases (zero bytes, plain bytes), and CPU normalization. The `Pi` and `P` test cases serve as regression tests.

### Steps to Test
1. Run `npm run frontend:test -- src/lib/util.test.ts`
2. All 92 tests should pass, including the new `normalizeUnit` suite.

### Screenshots
N/A — This is a data formatting utility function, not a visual component.

### Notes for the Reviewer
The function's comment block at line 499-502 already documents `Pi` and `P` as valid units, confirming this was an oversight rather than an intentional omission. The `Ei`/`E` multiplier correction is a direct consequence of inserting the missing `Pi`/`P` entries in the correct order.

### DIFF
--- a/frontend/src/lib/util.test.ts
+++ b/frontend/src/lib/util.test.ts
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
-import { combineClusterListErrors, flattenClusterListItems, formatDuration, timeAgo } from './util';
+import {
+  combineClusterListErrors,
+  flattenClusterListItems,
+  formatDuration,
+  normalizeUnit,
+  timeAgo,
+} from './util';
 
 const SECOND = 1000;
 const MINUTE = 60 * SECOND;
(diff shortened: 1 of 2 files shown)


frontend/src/lib/util.test.ts
```diff
@@ -204,3 +210,83 @@ describe('timeAgo', () => {
     expect(timeAgo(start, { format: 'mini' })).toBe('90d');
   });
 });
+
+describe('normalizeUnit', () => {
+  describe('memory — binary units', () => {
+    it('converts Ki (kibibytes)', () => {
+      expect(normalizeUnit('memory', '1Ki')).toBe('1.02 KB');
```
sniok, line 217: could you add test cases where the value is not a round number? like `1.5Ki` or `2.34Pi`

frontend/src/lib/util.test.ts
```diff
@@ -204,3 +210,83 @@ describe('timeAgo', () => {
...
+    it('converts Gi (gibibytes)', () => {
+      expect(normalizeUnit('memory', '1Gi')).toBe('1.07 GB');
+    });
+
+    it('converts Ti (tebibytes)', () => {
+      expect(normalizeUnit('memory', '1Ti')).toBe('1.1 TB');
+    });
+
+    it('converts Pi (pebibytes)', () => {
+      // Regression test: Pi was missing before the fix, causing incorrect values.
```
sniok: we shouldn't put implementation history in comments like this

frontend/src/lib/util.ts
```diff
@@ -510,8 +510,10 @@ export function normalizeUnit(resourceType: string, quantity: string) {
         bytes *= 1024 * 1024 * 1024;
       } else if (quantity.endsWith('Ti')) {
         bytes *= 1024 * 1024 * 1024 * 1024;
-      } else if (quantity.endsWith('Ei')) {
+      } else if (quantity.endsWith('Pi')) {
         bytes *= 1024 * 1024 * 1024 * 1024 * 1024;
+      } else if (quantity.endsWith('Ei')) {
+        bytes *= 1024 * 1024 * 1024 * 1024 * 1024 * 1024;
```
sniok: minor nit, I think using `1024**6` is bit more readable

sniok (changes requested):
thanks for the fix, I've left a couple of comments

KumarADITHYA123:
Thanks for the review @sniok, good catches all around.

Updated in b9da289:

Switched all the multiplier chains to use ** — agreed, way easier to scan at a glance
Dropped the implementation history from the test comments, makes sense to keep those clean
Added a handful of non-round value test cases (512Mi, 256Ki, 1500Mi, 1500k, 2500M) so we're not just testing the trivial 1X paths
Let me know if there's anything else you'd like adjusted.

frontend/src/lib/util.test.ts
```diff
@@ -266,6 +264,28 @@ describe('normalizeUnit', () => {
     });
   });
 
+  describe('memory — non-round values', () => {
+    it('converts 512Mi', () => {
+      expect(normalizeUnit('memory', '512Mi')).toBe('536.87 MB');
```
sniok: I think there's a misunderstanding here, I meant non round values as inputs like `normalizeUnit('memory', '1.5Gi')`

KumarADITHYA123:
hii @sniok  lgtm !

KumarADITHYA123:
@sniok Could you please review this PR at your convenience? Thanks!

sniok (approved):
thanks!

/lgtm

PR #5136, opened 2026-04-12
frontend: Hide create button for pods in owned resource sections

### LINKED ISSUE #5133: Nested pods within a Job should not a create button
Image

As I was working on JobSet (#4810) I found that nested objects like pods within a Job have a create button.

This is confusing because Jobs do not allow you to add pods to existing jobs. So I think this is actually invalid.

kannon92:
## Summary

Pods listed within a parent resource's detail view (e.g. Job, Deployment) should not show a "Create" button since pods cannot be added to existing workloads. Add hideCreateButton prop to PodListRenderer and set it in OwnedPodsSection.

## Related Issue

Fixes #5133 

## Changes

- Fixed #5133

## Steps to Test

1. Create a Job
2. See that pods no longer have the Create action.

old:

image

new:
image

## Notes for the Reviewer

None.


sniok (approved):
thanks!

/lgtm

PR #5145, opened 2026-04-15
frontend: portforward: Add user notification toast on failure

itvi-1234:
### Summary

This PR improves user feedback for the Port-Forwarding feature. Previously, if `startPortForward` failed (e.g., due to the port already being in use or connection issues), the failure was only logged silently to the console (`console.error`). This change adds an explicit snackbar toast notification so the user is properly informed when their port-forward action fails.

### Changes
- Updated the `.catch` block in `frontend/src/components/portforward/index.tsx`.
- Dispatched `enqueueSnackbar` to immediately render the error message string directly to the UI, falling back to a sanitized translation string if the error object varies.

### Steps to Test
1. Navigate to Headlamp -> Port Forwarding.
2. Attempt to Start a Port Forward on a port that is explicitly unavailable (already in use on your host machine) or forcefully break the connection.
3. Observe that instead of a silent error, a red error toast notification now appears at the bottom of the screen with a readable error message.

### DIFF
--- a/frontend/src/components/portforward/index.tsx
+++ b/frontend/src/components/portforward/index.tsx
@@ -347,6 +347,17 @@ export default function PortForwardingList() {
             })
             .catch(error => {
               setPortForwardInAction(null);
+              const errorMessage =
+                error instanceof Error ? error.message : typeof error === 'string' ? error : '';
+              const displayMessage = errorMessage
+                ? `${t('translation|Error starting port forward')}: ${errorMessage}`
+                : t('translation|Error starting port forward');
+
+              enqueueSnackbar(displayMessage, {
+                key: 'portforward-error',
+                preventDuplicate: true,
+                variant: 'error',
+              });
               console.error('Error starting port forward:', error);
             });
         }}
--- a/frontend/src/i18n/locales/de/translation.json
+++ b/frontend/src/i18n/locales/de/translation.json
@@ -603,6 +603,7 @@
   "Loading port forwarding": "Laden der Portweiterleitung",
   "Pod Port": "Pod-Port",
   "Local Port": "Lokaler Port",
+  "Error starting port forward": "",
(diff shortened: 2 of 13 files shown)


itvi-1234:
Hi @sniok @illume @joaquimrocha , Could you please review this PR whenever you get some time :rocket:

frontend/src/components/portforward/index.tsx
```diff
@@ -347,6 +347,13 @@ export default function PortForwardingList() {
             })
             .catch(error => {
               setPortForwardInAction(null);
+              const errorMessage = error instanceof Error ? error.message : error;
+              enqueueSnackbar(errorMessage || t('translation|Error starting port forward'), {
```
sniok: I think it should always says "Error starting port forward" along with the message from the error (if any)

itvi-1234:
Thanks for the suggestion @sniok , I have implemented the changes :rocket:

frontend/src/components/portforward/index.tsx
```diff
@@ -347,6 +347,18 @@ export default function PortForwardingList() {
...
+              const errorMessage =
+                error instanceof Error ? error.message : typeof error === 'string' ? error : '';
+              const displayMessage = errorMessage
+                ? `${t('translation|Error starting port forward')}: ${errorMessage}`
+                : t('translation|Error starting port forward');
+
+              enqueueSnackbar(displayMessage, {
+                key: 'portforward-error',
+                preventDuplicate: true,
+                autoHideDuration: 3000,
```
sniok: let's keep the default value instead of 3000 here for consistency with the rest of the app

itvi-1234:
Thanks for pointing it out @sniok , I have removed it

sniok (approved):
thanks!

/lgtm

PR #5159, opened 2026-04-17
backend: add per-request cluster filtering support

### LINKED ISSUE #5153: Per-user cluster visibility for in-cluster deployments embedded in multi-tenant apps
Hi, looking for advice / a feature direction before we commit to Headlamp
as the k8s UI layer in our platform, or work around it.

We run a multi-tenant web app that embeds Headlamp as an iframe. Each user
has access to a different subset of clusters (can be 100+ across the
platform), and the host app already knows which clusters a given session is
allowed to see. What we need is for Headlamp's cluster list to reflect that
per session - not be a pod-wide constant.

From reading the code and the docs, the cluster list comes from the backend
kubeconfig and is the same for everyone hitting the pod. Which means:

- If we preload all clusters into the backend kubeconfig, every user sees
  every cluster in the picker.
- If we want to switch a user from cluster A to cluster B mid-session
  (either by navigating in the host app, or by them opening a second tab
  on a different cluster), there's no per-session way to do that - routing
  via `/c/{cluster}` only works when the cluster is already in the shared
  list, and reconfiguring the backend affects everyone.

RBAC obviously filters what a user can *do* inside a cluster, but it
doesn't hide the cluster from the picker, so the UX is still noisy and
confusing.

We looked at stateless cluster management (https://headlamp.dev/blog/2024/04/09/user-added-cluster-support-in-shared-headlamp-deployments/) -
that's the only per-user mechanism today, but it relies on the user
pasting a kubeconfig manually. No way to drive it from the host app, and
IndexDB isn't shareable across iframe origins anyway.

We also went through related threads to make sure we're not duplicating:
#2710 (watcher is off in-cluster), #4283 / #4198 / #4077 (all about
identity propagation and RBAC inside clusters, not the cluster list),
#4919 (proxy auth - solves auth, doesn't touch the list).

A few shapes that would unblock us, from least to most invasive:

- Have the backend honour a request header (something like
  `X-Headlamp-Allowed-Clusters`) and filter `/clusters` and
  cluster-scoped paths based on it. Composes nicely with the proxy-auth
  work in #4919.
- A postMessage API for iframe embedders that scopes the picker on the
  frontend.
- Scoping driven by an OIDC claim listing allowed clusters.

Is this something the project would be open to? Happy to prototype one
of these if there's a preferred direction - or if there's a supported
path we missed, I'd love to hear it.

kunaldevsahu:
## Description

This PR introduces per-request cluster filtering in Headlamp to support multi-tenant deployments where cluster visibility and access are determined externally (e.g., by a host application or proxy).

---

## Motivation

In multi-tenant environments, different users should only be able to view and interact with a subset of clusters. Currently, Headlamp loads clusters from a shared kubeconfig, making all clusters visible to all users.

This change enables dynamic, request-scoped cluster filtering without requiring modifications to the underlying kubeconfig.

---

### Backend Filtering

* The `/config` endpoint now respects the `X-Headlamp-Allowed-Clusters` request header.
* When the header is present, only the specified clusters are returned to the frontend.
* When the header is not provided, the existing behavior remains unchanged.

### Access Enforcement

* All cluster-scoped API requests are validated against the same header.
* Requests targeting clusters not included in the allowed list return a `403 Forbidden` response.
* Enforcement is implemented centrally in the cluster API routing layer to ensure consistency.

### Tests

* Added test coverage in `backend/cmd/headlamp_test.go` to verify:

  * Cluster visibility filtering
  * Access restriction enforcement
  * Backward compatibility when the header is absent

---

## Usage

To enable per-user cluster scoping, inject the following header from a proxy or host application:

```http
X-Headlamp-Allowed-Clusters: cluster-a,cluster-b
```

This ensures that only the specified clusters are visible in the UI and accessible via the API.

---

## Design Considerations

* The implementation is request-driven and stateless.
* Core cluster loading logic remains unchanged.
* Backward compatibility is preserved for deployments that do not use the header.

---

## Additional Information

Further details on implementation and testing can be found in `walkthrough.md`.


skoeva (changes requested):
looks like the linter is raising issues...could you take a look at the backend test output?

sniok:
we still haven't landed on a solution to the issue, it's way too early for an implementation imho

skoeva:
> we still haven't landed on a solution to the issue, it's way too early for an implementation imho

I suppose we can set this PR to draft for now. will keep track of this

illume (changes requested):
Thanks for these changes.


<details>

Good examples:

</details>

skoeva (changes requested):
echoing the previous comments about commits, would also suggest rebasing against main while addressing the open review comments

illume (changes requested):
Thanks for these changes.


illume (changes requested):
Thanks for these changes.


I noticed the GitHub CI backend test job is failing. Can you please fix the failing tests? You can run `cd backend && go test ./...` locally to see the errors.

PR #5170, opened 2026-04-20
Update node version to 22.x LTS

sniok:
Node 20 will be EOL by the  end of this month so let's update all our ci workflows to use 22.x LTS
24.x was causing too many issues, so for now upgrading just to 22.x

## Changes

- Updated all workflows to use Node 22.x
- Updated "engines" fields in all package.json 

## Steps to Test

1. Check if CI workflows pass


skoeva (approved):
thanks!

PR #5173, opened 2026-04-21
docs: Update platforms.md to include NKP

jlnhnng:
I've thoroughly tested headlamp with NKP clusters, and I'm pleased to report it is functioning well without any workarounds, since NKP is based on kubeadm.


skoeva (approved):
It would be insightful, if you have any additional info that may be relevant from your testing, to include that somewhere, but this is awesome! thanks 🎉

skoeva (dismissed):
could you squash the merge commit as noted [here](https://headlamp.dev/docs/latest/contributing/#2-follow-commit-guidelines)? thanks!

illume (approved):
🎉 thanks!

PR #5177, opened 2026-04-22
frontend: stateless: Use console.debug for operational logs

NAME-ASHWANIYADAV:
## What this PR does

Replaces `console.log` with `console.debug` for IndexedDB operational
messages in `deleteClusterKubeconfig.ts` and `stateless/index.ts`.

## Why

These messages (`Kubeconfig deleted/updated/added to IndexedDB`) are
informational success confirmations, not debug artifacts. Using
`console.debug` keeps them accessible during development but hides
them from the production console by default, reducing noise for users.


sniok (approved):
thanks!

/lgtm

NAME-ASHWANIYADAV:
Thank You !! @sniok

PR #5221, opened 2026-04-24
backend: telemetry: use structured logger instead of log.Printf

NAME-ASHWANIYADAV:
## What this PR does / why we need it:
Replaces `log.Printf` with the project's structured `logger.Log` in the telemetry package for logging consistency.

### Changes:
- Replaced `log.Printf("Warning: ...")` with `logger.Log(logger.LevelWarn, ...)` including structured fields (`configured`, `selected`) for the multiple trace exporters warning.
- Removed unused `"log"` stdlib import.
- Added `"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"` import.

### Why:
The rest of the backend consistently uses the structured `logger` package (zerolog-based), but this single call used the stdlib `log.Printf`, breaking structured logging and making it harder to filter/parse logs in production.

### DIFF
--- a/backend/pkg/telemetry/telemetry.go
+++ b/backend/pkg/telemetry/telemetry.go
@@ -3,9 +3,10 @@ package telemetry
 import (
 	"context"
 	"fmt"
-	"log"
 	"strings"
 	"time"
+
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

NAME-ASHWANIYADAV:
Thank You !! 
@illume for review

skoeva (approved):
lgtm, thanks

PR #5229, opened 2026-04-25
cloudbuild: Pin gcb-docker-gcloud image by digest

### LINKED ISSUE #5228: cloudbuild: Fix broken `post-headlamp-push-images` job due to missing image tag
### Problem

The `post-headlamp-push-images` job fails on every merge to main. The job cannot pul the Cloud Build step image:

Image

The tag `v20250513-9264efb079` has been removed from the GCR registry, so the build exhausts its 10 retry attempts and fails.

### Impact

No container images are being pushed to the staging registry on merge. This blocks all image releases.

skoeva:
This change pins the `gcb-docker-gcloud` Cloud Build image by digest to fix the `post-headlamp-push-images` Prow job, which fails on every merge because the previous tag was garbage-collected from the registry.

Fixes: #5228

### Changes

- Pin image by digest instead of tag to prevent future tag GC breakage
- Align with the pattern used by other kubernetes-sigs projects (e.g. cluster-api)

### Testing
- [x] Verified current tag `v20250513-9264efb079` returns "no such manifest" via `docker manifest inspect`
- [x] Verified replacement digest `sha256:8d6a3a5b...` exists and is pullable
- [x] Verified `cloudbuild.yaml` is the only file referencing this image
- [x] Prow `post-headlamp-push-images` job should pass after merge

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,7 +4,7 @@ timeout: 3000s
 # For each build step, cloudbuild executes a docker container.
 steps:
   # see https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing
-  - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20250513-9264efb079'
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

PR #5267, opened 2026-04-27
frontend: node: Extract useMetrics into standalone hook

### LINKED ISSUE #5246: frontend: node: Fix `react-hooks/rules-of-hooks` — `useState`, `useErrorState`, and `useConnectApi` are called inside a static method of the `Node` class component. Hooks cannot be used in class components — refactor to a standalone hook function. File: `frontend/src/lib/k8s/node.ts`

aabhinavvvvvvv:
## Summary

Fixes a `react-hooks/rules-of-hooks` violation by extracting the `Node.useMetrics()` static class method into a standalone exported hook `useNodeMetrics()`, and updating all call sites to use it directly.

## Related Issue

Fixes #5246

## Changes

- Extracted `static useMetrics()` and `static useNodeSummaryStats()` from the `Node` class in `frontend/src/lib/k8s/node.ts` into standalone exported hooks `useNodeMetrics()` and `useNodeSummaryStats()`
- Updated `frontend/src/components/node/Details.tsx` to import and call both hooks directly
- Updated `frontend/src/components/node/List.tsx` to use `useNodeMetrics()`
- Updated `frontend/src/components/cluster/Overview.tsx` to use `useNodeMetrics()`
- Updated the plugin library snapshot to reflect the two new named exports

## Steps to Test

1. Run `npm run lint` in `frontend/` — should pass with no `react-hooks/rules-of-hooks` errors

## Notes for the Reviewer
- The extracted `useNodeMetrics` function is behaviour-identical to the original static method; only the call site syntax changes.

### DIFF
--- a/frontend/src/components/cluster/Overview.tsx
+++ b/frontend/src/components/cluster/Overview.tsx
@@ -22,7 +22,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router';
 import Event from '../../lib/k8s/event';
-import Node from '../../lib/k8s/node';
+import Node, { useNodeMetrics } from '../../lib/k8s/node';
 import Pod from '../../lib/k8s/pod';
 import { useFilterFunc } from '../../lib/util';
 import { useNamespaces } from '../../redux/filterSlice';
@@ -48,7 +48,7 @@ export default function Overview() {
   const { t } = useTranslation(['translation']);
   const [pods] = Pod.useList();
   const [nodes] = Node.useList();
-  const [nodeMetrics, metricsError] = Node.useMetrics();
+  const [nodeMetrics, metricsError] = useNodeMetrics();
   const chartProcessors = useTypedSelector(state => state.overviewCharts.processors);
 
   const noMetrics = metricsError?.status === 404;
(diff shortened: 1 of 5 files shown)


sniok (changes requested):
I don't think we need to extract static methods that are hooks. Eslint is giving false positive errors here

This also breaks compatibility for plugins

illume:
Hmm. It's technically correct that the rule of hooks rules out static methods, for valid reasons (not just stylistic reasons). I don't know if we're encountering any actual issues because of it(didn't check). But maybe it's worth fixing just for the in-case and stylistic/consistency reasons?

How about implementing the static method with the hook function, and marking the static method `@deprecated` with a code comment?

I searched GitHub and [our plugins repo](https://github.com/search?q=repo%3Aheadlamp-k8s%2Fplugins%20Node.useMetrics&type=code)... couldn't find users of Node.useMetrics, but that's not all plugins.

aabhinavvvvvvv:
@illume will the deprecated wrapper approach work as i wanted to make sure before i implement it

sniok:
Here's the error that eslint is reporting

```
React Hook "React.useState" cannot be called in a class component. React Hooks must be called in a React function component or a custom React Hook function.
```

Node is not a class component. This is a 100% a false positive, not valid and doesn't need to be fixed. There's nothing to fix here just in case

aabhinavvvvvvv:
@illume just dropping by to ask if the deprecated wrapper technique works for you?

PR #5291, opened 2026-04-28
frontend: e2e: Add e2e tests for CreateResourceForm

vyncent-t:
## Summary

Adds a Playwright e2e test for the new data-driven `CreateResourceForm`, exercising the Pod resource type end-to-end via the form (no YAML typing). The test covers two flows: validation of the Apply button state as required fields are filled, and a happy-path pod creation that verifies the new pod appears in the list afterward.

## Changes

- **New page object** `e2e-tests/tests/createResourceFormPage.ts`
  - Opens the Create Pod activity from the global `Create Pod` button and switches to the Form tab
  - Exposes stable role/`aria-label` locators for the fields exposed by `CreateResourceForm` / `ContainerTextField`: `Name`, `Namespace`, `Container name`, `Container image`, `Apply`, `Close`
  - Provides small helpers: `fillMetadata`, `fillFirstContainer`, `addContainer`, `clickApply`, `closeCreateDialog`, `waitForAppliedToast`
  - Includes an `a11y()` helper using `@axe-core/playwright`, matching other page objects
- **New spec** `e2e-tests/tests/createResourceForm.spec.ts` (`test.describe('CreateResourceForm – Pod')`)
  - Authenticates against the `test` cluster and short-circuits when pod permissions are missing (same pattern as `namespaces.spec.ts` and `podsPage.spec.ts`)
  - `Apply stays disabled until required fields are filled` — asserts the Apply button stays disabled at each intermediate state (nothing filled, metadata only, container name only) and only becomes enabled once the container image is filled too
  - `creates a pod via the form` — opens the form, fills metadata and the seeded first container, clicks Apply, closes the activity via the `X`, waits for the `Applied …` toast, navigates back to the pods list, and confirms the new pod link is visible; uses a per-retry unique name so retries don't collide with leftovers

## Steps to Test

1. `cd e2e-tests && npm install`
2. Start Headlamp against a cluster with pod permissions
3. Run: `HEADLAMP_TEST_URL=<url> HEADLAMP_TEST_TOKEN=<token> npx playwright test tests/createResourceForm.spec.ts`
4. Both tests should pass

## Notes for the Reviewer

- The page object targets the Pod form specifically. The base Pod seeds one empty container row, so `fillFirstContainer` fills the seeded row rather than adding a new one.
- Pod names are suffixed with `testInfo.retry` so a leftover pod from a failed prior attempt doesn't turn the Apply click into a duplicate-create no-op.
- The `Namespace` combobox lookup falls back to the first combobox in the form because the underlying MUI Autocomplete label association isn't always picked up by role/name.


illume (changes requested):
Please see copilot review comments?

PR #5292, opened 2026-04-29
backend: k8cache: Fix corrupted cache keys due to fragile ExtractNamespace logic

Joshna907:
### **Changes:**

1. `backend/pkg/k8cache/cacheStore.go (ExtractNamespace)`: Replaced the fragile len(urls) > 2 check with robust URL parsing logic. The function now ensures it only extracts a resource "Kind" if the URL is a valid Kubernetes API path (verifying it contains an "api" or "apis" segment). Non-API URLs now safely return empty values, preventing them from being improperly cached.
2. `backend/pkg/k8cache/cacheStore_test.go`: Added new unit tests to explicitly verify that ExtractNamespace correctly ignores internal cluster URLs and plugin URLs, confirming they return empty namespace and kind strings.

### **Steps to Test:**

1. Run the targeted backend unit tests to verify the new URL parsing logic: cd backend && go test -v ./pkg/k8cache/... -run TestExtractNamespace
2. Run the full backend test suite to ensure no regressions across the caching layer: cd backend && go test -v ./pkg/k8cache/...
3. (Optional) Start the application locally (npm start) and navigate between clusters and plugins in the UI. Ensure that the backend logs no longer show unexpected "Kinds" (like cluster names) being stored in the cache.

### DIFF
--- a/backend/pkg/k8cache/cacheStore.go
+++ b/backend/pkg/k8cache/cacheStore.go
@@ -35,6 +35,11 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 )
 
+const (
+	apiPathSegment  = "api"
+	apisPathSegment = "apis"
+)
+
 // CachedResponseData stores information such as StatusCode, Headers, and Body.
 // It helps cache responses efficiently and serve them from the cache.
 type CachedResponseData struct {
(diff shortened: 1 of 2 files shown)


Joshna907:
@illume I have fixed the comments, PTAL !!

illume (approved):
This looks good to me. Thanks a lot!

PR #5300, opened 2026-04-29
frontend: App: Fix react-hooks lint warnings

skoeva:
These changes fix two `react-hooks` lint warnings in `AddCluster.stories.tsx` decorator by moving the `window.process` mutation into a `useEffect` and capturing the original value in a local variable rather than a ref. The related storyshot snapshot is updated accordingly.

### Testing

- Run `cd frontend && npm run ci-lint` and ensure it passes


skoeva (commented):
let's just revert the original branch

PR #5171, opened 2026-04-20
Add Storybook stories for AddCluster component

Dharmiksarvaiya24:
## Summary
Adds Storybook stories for the `AddCluster` component to visualize
its real UI states during development.

The component has three meaningful states driven by Redux:
- No providers registered
- No providers but plugin catalog registered in sidebar
- With cluster providers registered

## Changes
- Added `frontend/src/components/App/CreateCluster/AddCluster.stories.tsx`

## Related Issue
Fixes #4687 

## Steps to Test
1. Navigate to frontend: `cd frontend`
2. Run Storybook: `npm run storybook`
3. In sidebar find `App > AddCluster`
4. Verify each story renders a visually distinct UI state
5. Check browser console for errors

Screenshot 2026-04-27 at 6 58 04 PM
Screenshot 2026-04-27 at 6 58 11 PM

PR #5301, opened 2026-04-29
Revert "Add Storybook stories for AddCluster component"

skoeva:
There's a story that's causing the CI to fail... we need to investigate this further.


vyncent-t:
Related - #5171

vyncent-t (approved):
Builds are passing LGTM

PR #5306, opened 2026-04-29
fix(backend): preserve kubeconfig server path prefix in WebSocket multiplexer and port-forward, prefer WS port-forward

### LINKED ISSUE #5304: Streaming and port-forward fail when kubeconfig server URL contains a path prefix
## Describe the bug
When the cluster `server` URL in a `kubeconfig` carries a **path prefix**
-  typical when the apiserver is fronted by a path-routing reverse proxy or bastion (e.g. [Warpgate](https://github.com/warp-tech/warpgate)'s Kubernetes proxy, custom `kubectl proxy --append-server-path …` setups) 
- every Headlamp interaction that requires an HTTP **upgrade** (WebSocket or SPDY/3.1) fails:

* Pod **logs** (`?follow=true`) and resource **watches** (`?watch=1`)
  surface as `WebSocket connection to 'ws://localhost:…/events?…'
  failed` in the browser console; the corresponding tabs stay empty or
  hang.
* Pod **exec / terminal**: same WebSocket error, terminal never opens.
* **Port-forward**: returns `500 Internal Server Error`. Server log
  reads `error upgrading connection: Upgrade request required`.

Plain HTTP requests (resource list, get, create, delete) **keep working**, which made the issue confusing at first since the cluster clearly wasn't unreachable.

Expected outcome: the same operations work as they do with the standard 
`kubectl` CLI on the same `kubeconfig`, since `kubectl` handles path-prefixed `server` URLs natively.

## To Reproduce
The simplest local reproduction uses `kubectl proxy` with a path prefix
to mimic any path-routing front-proxy.

1. Start a path-prefixed proxy in front of any reachable cluster:
   ```bash
   kubectl proxy \
     --address 127.0.0.1 --port 9000 \
     --append-server-path /my-cluster \
     --accept-hosts='.*'
Save a minimal kubeconfig pointing at that proxy:

```
apiVersion: v1
kind: Config
clusters:
  - cluster:
      server: http://127.0.0.1:9000/my-cluster
    name: prefixed
contexts:
  - context: { cluster: prefixed, user: anon }
    name: prefixed
current-context: prefixed
users:
  - name: anon
    user: {}

```

Open Headlamp on this kubeconfig.
Resource lists load fine. Now click on any pod → Logs tab. Observe the failure in DevTools: WebSocket connection to 'ws://localhost:4466/clusters/prefixed/api/v1/namespaces/.../pods/.../log?follow=true&…' failed.
Same pod → click Port Forward. Observe a 500 returned by /clusters/prefixed/portforward, with body containing error upgrading connection: Upgrade request required in Headlamp's server logs.
As a sanity check, identical operations through kubectl against the same kubeconfig succeed:

A real-world repro is any cluster behind Warpgate's k8s proxy (it routes by URL path) — but the bug is purely on Headlamp's side and reproduces with the lightweight kubectl proxy setup above.

## Environment :
Installation type: Linux-AppImage (also reproduced from sources, build from main)
Headlamp Version: 0.41.0
Other:
Backend Go version: 1.25
client-go / apimachinery: v0.35.1
Browser: Chromium-based (Headlamp's embedded Electron 41.x), also reproduces in Firefox/Chrome on the served frontend.
Reverse-proxy frontends confirmed affected: Warpgate (Kubernetes proxy module, v0.23.x), kubectl proxy --append-server-path.

 - Installation type: [e.g. Linux-Flatpak/Linux-Tarball/Linux-AppImage/Windows//Windows-Chocolatey/Windows-Winget/WSL/Mac/Mac-Homebrew/... or Container-Image/In-Cluster/Helm/...]
 - Headlamp Version: [X.Y.Z]

Yes — PR proposed. The root causes are three independent path-handling

## Additional Context
Likely a long-standing regression — the affected functions (createWebSocketURL, initPortForwarder) are old, but path-prefixed kubeconfigs are uncommon outside SaaS bastion setups, which is probably why it slipped through.
An existing unit test in backend/cmd/multiplexer_test.go (TestCreateWebSocketURL/HTTPS_with_path_in_host) actually documented the buggy behavior — it asserts that the host path is dropped. The PR rewrites it.
No frontend change is needed: the browser already forwards the correct URL to headlamp-server. The bug is entirely in how headlamp-server rebuilds the upstream URL before dialing.

plopoyop:
## Summary

This PR fixes the cluster proxy in the backend so that interactions requiring an HTTP upgrade (WebSocket / SPDY) keep working when the `kubeconfig` cluster `server` URL contains a path prefix — e.g. clusters fronted by reverse proxies such as Warpgate or any path-routing bastion.
It also enables WebSocket-first port-forward negotiation (SPDY fallback) so port-forward survives proxies that don't transparently propagate the `Upgrade: SPDY/3.1` header.

## Related Issue

Fixes #5304  

## Changes

- Fixed `backend/cmd/multiplexer.go::createWebSocketURL`: the upstream  WebSocket URL was rebuilt with `u.Path = path`, which dropped any path  prefix carried by the kubeconfig server URL. Replaced by  `singleJoiningSlash(u.Path, path)` (mirroring `net/http/httputil`).
- Fixed `backend/pkg/portforward/handler.go::initPortForwarder`: the  upstream port-forward URL was built with
  `hostURL.ResolveReference(&url.URL{Path: path})`, which silently drops  the host's path because the reference path is absolute. Extracted  `buildPortForwardURL`, which preserves any prefix on `rConf.Host`.
- Added `buildPortForwardDialer`: tries WebSocket first via `portforward.NewSPDYOverWebsocketDialer` and falls back to the existing  SPDY dialer with `portforward.NewFallbackDialer`. Same negotiation pattern that `kubectl port-forward` adopted in v1.31.
- Added/updated unit tests:
  - `TestCreateWebSocketURL`: new cases for trailing-slash prefix and multi-segment prefix; fixed an existing case that was asserting the buggy "drop the prefix" behavior.
  - `TestBuildPortForwardURL`: covers no prefix, single-segment prefix, trailing-slash prefix, multi-segment prefix and invalid host.

## Steps to Test

1. Use a `kubeconfig` whose cluster `server` field contains a path
   prefix, for example:
   ```yaml
   clusters:
     - cluster:
         server: https://k8s.example.com:443/dev-cluster
         insecure-skip-tls-verify: true
       name: warpgate-dev

Any setup where a reverse proxy in front of the apiserver routes by URL path will reproduce the bug. A kubectl proxy --address 127.0.0.1    --port 9000 --append-server-path /dev-cluster instance is the
simplest local repro.
2. Open Headlamp on that cluster.
3. Before the patch: navigating to a pod's "Logs" or "Events" tab
yields WebSocket connection to 'ws://localhost:.../events?...&watch=1'    failed in the browser console; clicking "Port Forward" returns
500 Internal Server Error with error upgrading connection: Upgrade request required.
4. After the patch: log streaming works, exec terminal opens, and port-forward succeeds. The upstream proxy logs the request as GET /<prefix>/api/v1/.../portforward → 101 Switching Protocols, confirming the WebSocket path is taken.
5. Regression check on a vanilla cluster (no path prefix): all existing flows continue to work — singleJoiningSlash("", "/api/...") ==    "/api/..." and the WebSocket dialer falls back to SPDY transparently on apiservers that don't speak the WS variant.


## Notes for the Reviewer

buildPortForwardDialer takes rConf *rest.Config because NewSPDYOverWebsocketDialer needs the auth/TLS config to build the WebSocket transport — it can't reuse the SPDY round-tripper.
The fallback condition (IsUpgradeFailure || IsHTTPSProxyError) is intentionally identical to kubectl's (pkg/cmd/portforward/portforward.go::createDialer) so we benefit from the same upstream behavior and bug fixes.
For setups where the cluster server URL contains characters that need URL-encoding (uncommon in practice), url.Parse already returns the decoded Path; the constructed URL is then re-encoded by Go's net/url and request writers, so no double-encoding occurs.
No public API or CLI surface changes; the multiplexer and port-forward routes accept the exact same payloads as before.

### DIFF
--- a/backend/pkg/portforward/internal_test.go
+++ b/backend/pkg/portforward/internal_test.go
@@ -234,6 +234,63 @@ func TestPortForwardRequestValidate(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+// TestBuildPortForwardURL ensures the upstream port-forward URL preserves the
+// kubeconfig server's path prefix (relevant when the cluster is fronted by a
+// path-routing reverse proxy such as Warpgate).
+func TestBuildPortForwardURL(t *testing.T) {
+	tests := []struct {
+		name      string
+		host      string
+		namespace string
+		podName   string
+		want      string
+	}{
+		{
+			name:      "no path prefix",
+			host:      "https://kubernetes.default.svc:443",
+			namespace: "default",
+			podName:   "my-pod",
+			want:      "https://kubernetes.default.svc:443/api/v1/namespaces/default/pods/my-pod/portforward",
+		},
+		{
+			name:      "single segment path prefix",
+			host:      "https://example.com/k8s",
+			namespace: "default",
+			podName:   "my-pod",
+			want:      "https://example.com/k8s/api/v1/namespaces/default/pods/my-pod/portforward",
+		},
+		{
+			name:      "trailing slash path prefix",
+			host:      "https://example.com/k8s/",
+			namespace: "default",
+			podName:   "my-pod",
+			want:      "https://example.com/k8s/api/v1/namespaces/default/pods/my-pod/portforward",
+		},
+		{
+			name:      "Warpgate-style multi-segment prefix",
+			host:      "https://k8s.example.com:443/proxy-routed-cluster",
+			namespace: "kube-system",
+			podName:   "traefik-69fpr",
+			want: "https://k8s.example.com:443/proxy-routed-cluster" +
+				"/api/v1/namespaces/kube-system/pods/traefik-69fpr/portforward",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := buildPortForwardURL(tt.host, tt.namespace, tt.podName)
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got.String())
+		})
+	}
+}
+
+// TestBuildPortForwardURLInvalidHost ensures an invalid host yields an error.
+func TestBuildPortForwardURLInvalidHost(t *testing.T) {
+	_, err := buildPortForwardURL("://not a url", "ns", "pod")
+	assert.Error(t, err)
+}
+
 // TestStopOrDeletePortForwardRequest.Validate() function.
 func TestStopOrDeletePortForwardRequestValidate(t *testing.T) {
 	req := stopOrDeletePortForwardRequest{}
--- a/backend/pkg/portforward/internal_test.go
+++ b/backend/pkg/portforward/internal_test.go
@@ -31,6 +31,8 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/cache"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
 )
 
 // TestHandlePortForwardReadiness tests handlePortForwardReadiness function.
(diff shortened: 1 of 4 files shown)


gambtho:
Thanks for the PR!  could you consider adjusting your commit message slightly to match the project format  something like   `backend: multiplexer: Preserve kubeconfig path prefix in WS proxy and port-forward`. maybe?

backend/pkg/portforward/internal_test.go
```diff
@@ -165,6 +165,63 @@ func TestPortForwardRequestValidate(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+// TestBuildPortForwardURL ensures the upstream port-forward URL preserves the
```
gambtho, line 239: might be worth adding a test for the spdy fallback too

illume (changes requested):
Thanks for working on this.


Would you mind addressing the open Copilot review comments?

illume (changes requested):
Thanks for the contribution.


<details>

</details>

illume (changes requested):
Thanks for these changes.


It looks like the [Contributor License Agreement (CLA) has not been signed](https://github.com/kubernetes-sigs/headlamp/pull/5306#issuecomment-4345872346). Kubernetes requires all contributors to sign the CLA before a PR can be merged. Could you please take a look at the EasyCLA bot comment and sign it?

<details>
<summary>About the CLA</summary>

The Kubernetes project uses the Linux Foundation EasyCLA system. Signing takes only a minute through the link in the EasyCLA bot comment. Individual contributors sign electronically; corporate contributors may need their employer to sign a Corporate CLA.

</details>

illume (approved):
🎉 thanks!

PR #5307, opened 2026-04-29
backend: multiplexer: Fix lock-order inversion deadlock in sendDataMe…

### LINKED ISSUE #5305: backend: multiplexer: Lock-order inversion deadlock between updateStatus and sendDataMessage
##  Description

There is a lock-order inversion deadlock in `backend/cmd/multiplexer.go` between two concurrently-called functions on the `Connection` struct. Under load, this causes the Headlamp server to freeze permanently, requiring a process restart.

## Root Cause

The `Connection` struct has two mutexes:
- `mu` (`sync.RWMutex`) — protects connection state (`Status`, `closed`, etc.)
- `writeMu` (`sync.Mutex`) — protects WebSocket write operations

Two functions acquire these locks in **opposite order**, creating a classic circular wait:

**`updateStatus`** (called on state changes, errors, reconnects):
```go
c.mu.Lock()          // Lock 1
  c.writeMu.Lock()   // Lock 2 (nested inside mu)
```

**`sendDataMessage`** (called on every incoming Kubernetes message):
```go
conn.writeMu.Lock()  // Lock 1
  conn.mu.Lock()     // Lock 2 (nested inside writeMu)
```

## Deadlock Scenario

When data is flowing over a WebSocket connection while the connection's status is being updated concurrently:

1. **Goroutine A** (in `sendDataMessage`): acquires `writeMu`, blocks waiting for `mu`
2. **Goroutine B** (in `updateStatus`): acquires `mu`, blocks waiting for `writeMu`

Both goroutines wait forever — the server freezes for that client connection.

## Trigger Conditions

This deadlock is triggered when:
- A Kubernetes watch stream is actively sending data (`sendDataMessage` runs continuously)
- AND a connection error, heartbeat failure, or reconnect occurs (`updateStatus` is called)

This is a realistic production scenario under any non-trivial cluster load.

## Impact

- **Severity:** P1 — server deadlock (frozen WebSocket connections)
- **Affected users:** Any Headlamp deployment with active watch streams and intermittent cluster connectivity issues
- **Scope:** Individual client connections freeze permanently; server process stays alive but affected connections become unresponsive

## Expected Behavior

`sendDataMessage` should not hold `writeMu` while acquiring `mu`. The two critical sections should be sequential (non-nested), maintaining a consistent global lock order.

KumarADITHYA123:
## Summary

Fix a lock-order inversion deadlock between `updateStatus` and `sendDataMessage` in `backend/cmd/multiplexer.go` that can freeze WebSocket connections under concurrent load.

## Related Issue

Fixes #5305

## Root Cause

The `Connection` struct uses two mutexes: `mu` (state) and `writeMu` (WebSocket writes). Two functions acquired them in opposite order, creating a circular wait:

| Function | Lock Order | Nested? |
|---|---|---|
| `updateStatus` | `mu.Lock()` → `writeMu.Lock()` | Yes |
| `sendDataMessage` | `writeMu.Lock()` → `mu.Lock()` | Yes (BUG) |

When a Kubernetes watch stream sends data (`sendDataMessage`) while a connection error triggers a status update (`updateStatus`), both goroutines block forever waiting for each other's held lock.

## Fix

Make `sendDataMessage`'s lock acquisitions sequential (non-nested) instead of nested. The `writeMu` is released before `mu` is acquired, eliminating the circular wait while preserving correct synchronization:

```diff
 func (m *Multiplexer) sendDataMessage(...) error {
     dataMsg := m.createWrapperMessage(conn, messageType, message)

     conn.writeMu.Lock()
-    defer conn.writeMu.Unlock()
-
-    if err := clientConn.WriteJSON(dataMsg); err != nil {
+    err := clientConn.WriteJSON(dataMsg)
+    conn.writeMu.Unlock()
+
+    if err != nil {
         return err
     }

     conn.mu.Lock()
     conn.Status.LastMsg = time.Now()
     conn.mu.Unlock()

     return nil
 }
```

This is safe because:
- `writeMu` only protects the WebSocket write operation
- `mu` only protects the `Status.LastMsg` update
- These two operations are logically independent

## Post-fix Lock Ordering (Verified)

| Function | Lock Order | Nested? | Deadlock Risk? |
|---|---|---|---|
| `updateStatus` | `mu` → `writeMu` | Yes | No |
| `sendDataMessage` (fixed) | `writeMu` → release → `mu` | **No** | **No** |
| `sendCompleteMessage` | `mu.RLock` → release → `writeMu` | No | No |
| `cleanupConnection` | `mu` only | N/A | No |

## Notes for the Reviewer

- This is a 3-line diff (excluding comment additions) — minimal and surgical.
- The fix follows the same pattern already used by `sendCompleteMessage` (lines 710-739), which correctly releases `mu.RLock()` before acquiring `writeMu.Lock()`.
- Existing tests (`TestSendDataMessage`) validate the function's behavior (text, binary, and closed connection cases). All `cmd` package tests and linters pass cleanly.

### DIFF
--- a/backend/cmd/multiplexer.go
+++ b/backend/cmd/multiplexer.go
@@ -739,6 +739,10 @@ func (m *Multiplexer) sendCompleteMessage(conn *Connection, clientConn *WSConnLo
 }
 
 // sendDataMessage sends the actual data message to the client.
+//
+// Lock ordering note: writeMu is released before acquiring mu to
+// maintain a consistent lock order (mu → writeMu) with updateStatus
+// and prevent a lock-order inversion deadlock.
 func (m *Multiplexer) sendDataMessage(
 	conn *Connection,
 	clientConn *WSConnLock,
@@ -748,9 +752,10 @@ func (m *Multiplexer) sendDataMessage(
 	dataMsg := m.createWrapperMessage(conn, messageType, message)
 
 	conn.writeMu.Lock()
-	defer conn.writeMu.Unlock()
+	err := clientConn.WriteJSON(dataMsg)
+	conn.writeMu.Unlock()
 
-	if err := clientConn.WriteJSON(dataMsg); err != nil {
+	if err != nil {
 		return err
 	}
 
--- a/backend/cmd/multiplexer_test.go
+++ b/backend/cmd/multiplexer_test.go
@@ -1570,3 +1570,101 @@ func TestSendDataMessage(t *testing.T) {
 	err = m.sendDataMessage(conn, clientConn, websocket.TextMessage, textMsg)
 	assert.NoError(t, err) // Should return nil even for closed connection
 }
+
+// runConcurrentLockStress fires updateStatus and sendDataMessage
+// simultaneously for n iterations to surface lock-order deadlocks.
+func runConcurrentLockStress(
+	m *Multiplexer,
+	conn *Connection,
+	clientConn *WSConnLock,
+	iterations int,
+) <-chan struct{} {
+	finished := make(chan struct{})
+
+	go func() {
+		defer close(finished)
+
+		var start sync.WaitGroup
+
+		for i := 0; i < iterations; i++ {
+			start.Add(1)
+
+			var wg sync.WaitGroup
+
+			wg.Add(2)
+
+			go func() {
+				defer wg.Done()
+
+				start.Wait()
+
+				conn.updateStatus(StateConnected, nil)
+			}()
+
+			go func() {
+				defer wg.Done()
+
+				start.Wait()
+
+				_ = m.sendDataMessage(conn, clientConn, websocket.TextMessage, []byte("test"))
+			}()
+
+			// Release both goroutines simultaneously.
+			start.Done()
+
+			wg.Wait()
+		}
+	}()
+
+	return finished
+}
+
+// TestConcurrentUpdateStatusAndSendDataMessage is a regression test that verifies
+// updateStatus and sendDataMessage can run concurrently without deadlocking.
+//
+// Before the fix, sendDataMessage acquired writeMu then mu (nested), while
+// updateStatus acquired mu then writeMu, creating a lock-order inversion
+// (ABBA deadlock). The fix changed sendDataMessage to acquire each lock
+// sequentially (writeMu → unlock → mu → unlock), eliminating the nested
+// acquisition entirely.
+func TestConcurrentUpdateStatusAndSendDataMessage(t *testing.T) {
+	m := NewMultiplexer(kubeconfig.NewContextStore())
+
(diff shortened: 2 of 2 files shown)


backend/cmd/multiplexer.go
```diff
@@ -748,9 +752,10 @@ func (m *Multiplexer) sendDataMessage(
 	dataMsg := m.createWrapperMessage(conn, messageType, message)
 
 	conn.writeMu.Lock()
-	defer conn.writeMu.Unlock()
+	err := clientConn.WriteJSON(dataMsg)
```
gambtho, line 755: clean fix, possibly worth adding a negative test to prevent regression?

KumarADITHYA123: Apologies for the delay! Thanks for the review. 

I've added a regression test (`TestConcurrentUpdateStatusAndSendDataMessage`) that exercises the concurrent lock-acquisition paths. It runs 50 iterations of simultaneous `updateStatus` and `sendDataMessage` calls with a 5s deadlock timeout to ensure the lock-order inversion is caught if it ever regressions. 

Also rebased onto latest `main`.

illume (changes requested):
Thanks for this PR.


<details>

</details>

KumarADITHYA123:
hey @illume @gambtho 
Fixed the golangci-lint wsl_v5 failures (missing blank lines after defer in goroutine bodies) and force-pushed.

Regarding the Copilot suggestion on test determinism: the current approach uses the standard Go concurrency testing pattern (many iterations + timeout). With 50 iterations spawning 100 goroutines, the scheduling interleaving is effectively guaranteed. Pre-locking internal mutexes from test code would couple the test to internal lock implementation details, making it fragile if the locking strategy evolves.

All checks passing locally:
- golangci-lint: 0 issues
- gofumpt: clean
- go test ./cmd/: PASS

KumarADITHYA123:
Thanks @illume for the Copilot review! Addressed the bot's feedback:

Goroutine leak: Added t.Cleanup to close clientConn and properly shut down the reader goroutine.
Nondeterministic test: Added a sync.WaitGroup start barrier to ensure simultaneous goroutine execution.
(Note on the bot's TryLock suggestion: It assumes nested locking, but our fix changed sendDataMessage to sequential locking. Since both locks are never held simultaneously, TryLock wouldn't detect anything useful here. The concurrent stress test is the correct approach.)

I also fixed a funlen linter issue in the test and rebased on the latest main.

illume (approved):
Awesome. Thanks for finding this tricky bug.

🎉

KumarADITHYA123:
"Thank you for the review and approval, @illume! Tracking down that deadlock was a fun challenge. Glad to help make the backend more robust!"

Issue #5302, opened 2026-04-29
Add deep-link support for workload logs (Deployment/StatefulSet/DaemonSet) via ?view=logs


bhat-ganesh:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

Headlamp already supports deep-linking directly to a pod's log viewer via `?view=logs` on the Pod details page (commit [`dc7e7e4`](https://github.com/kubernetes-sigs/headlamp/commit/dc7e7e408334b359ddadefb5646fed3c76ccd2ab), "frontend: pod: Add deeplink params to view pod logs"), and the same pattern was extended to exec in #4941. However, pod names are ephemeral — they change on every deploy, rollout, or restart — so a bookmarked URL to a specific pod's logs stops working almost immediately.

In practice, developers and on-call engineers usually want "the logs for my workload," not "the logs for pod `foo-abc123-xyz`." Today there is no stable URL that answers that need. The workaround is either (a) navigating manually each time (Deployments → pick one → Pods → pick newest → Logs), or (b) scripting `kubectl get pods -l …` + URL interpolation — neither is easily shareable in chat, dashboards, or runbooks.

## Describe the solution you'd like

Extend the existing `?view=logs` deep-link pattern from Pod details to the three workload kinds that own pods:

- `/deployments/:namespace/:name?view=logs`
- `/statefulsets/:namespace/:name?view=logs`
- `/daemonsets/:namespace/:name?view=logs`

When the param is present, the workload details page would resolve a pod via the workload's label selector, pick one sensibly (newest-created among `Ready` pods; fall back to newest pod overall if none are `Ready`), and open the same `PodLogViewer` activity that Pod details already opens.

Behavior notes / edge cases to consider:
- No pods matching the selector yet → render the details page normally (no error dialog).
- Pod-selection strategy should be documented. "Newest Ready pod" is a reasonable default; we could also support `?view=logs&pod=<name>` as a future extension to pin a specific replica.
- Scope of this issue is intentionally just `logs`. `?view=exec` on workloads is a reasonable follow-up but is out of scope here — happy to file separately if there's interest.

## What users will benefit from this feature?

- Anyone maintaining runbooks, on-call playbooks, or dashboards that link to Headlamp — they get stable, shareable URLs that survive pod restarts.
- Plugin developers and platform teams embedding Headlamp, who today have to build their own pod-lookup layer to produce a usable logs link.
[...]

PR #5312, opened 2026-04-29
frontend: workload/Details: Support ?view=logs deep-link

bhat-ganesh:
## Summary

Adds `?view=logs` URL deep-link support on workload details pages — opening a URL like `…/deployment/<ns>/<name>?view=logs` auto-launches the existing aggregated Logs activity.

- Applies to **Deployment, ReplicaSet, DaemonSet, StatefulSet** (all route through the shared `WorkloadDetails` component).
- Also adds `'StatefulSet'` to the workload-details `isLoggable` action list so the Logs action button appears for StatefulSets (`LogsButton` itself already supports StatefulSet — this was a pre-existing gap in `workload/Details.tsx`).

Refs #5302 (design discussed and acked by maintainers in the issue thread).

## Motivation

Pod names are ephemeral (regenerated on every deploy / restart) so bookmarked pod-level log URLs break quickly. Workload names are stable, so a `?view=logs` URL on a Deployment/StatefulSet is shareable in runbooks and on-call docs.

This mirrors the existing pattern on `pod/Details.tsx` (`?view=logs` and `?view=terminal` — precedent: commit `dc7e7e4`, PR #4941).

## Implementation

One behavior change, one scaffolding refactor, one pre-existing bug fix:

1. **`LogsButton.tsx`** — extract `launchWorkloadLogs(item)` as a standalone export so the Activity can be launched programmatically (not just from a button click). No behavior change to the button itself; `onClick` now just delegates.
2. **`workload/Details.tsx`** — read `?view` from the URL, hold the fetched workload in state via `onResourceUpdate`, and `useEffect` auto-launches logs once per unique workload `uid`. A `useRef` makes it idempotent across polling rerenders and resets on navigation away.
3. **`workload/Details.tsx`** — add `'StatefulSet'` to `isLoggable` (one line).

## Testing

**Unit/integration tests** (12 total, all passing, lint clean):
- Extended `LogsButton.test.tsx` with a case asserting `launchWorkloadLogs` can be called programmatically.
- New `workload/Details.test.tsx` covers:
  - auto-launches when `?view=logs`
  - does NOT launch without the param
  - does NOT launch for other `view` values
  - does NOT re-launch on rerender for the same workload (idempotency)

**Manual testing against a real Kubernetes cluster** (not just locally):
- Deployment: `…/deployments/dev/ganesh-bhat-simple-app?view=logs` ✅
- StatefulSet: `…/statefulsets/dev/…-mysql?view=logs` ✅
- DaemonSet: `…/daemonsets/kube-system/ebs-csi-node?view=logs` ✅
- Idempotency verified (no double-launch on DetailsGrid polling rerenders)
- Re-trigger verified (close activity, navigate away, navigate back re-fires)

## Test plan

- [x] Unit + integration tests added and passing
- [x] Manual verification against live cluster (Deployment, StatefulSet, DaemonSet)
- [x] Idempotency: no double-launch on rerender
- [x] Re-trigger: closing activity + navigating back re-fires
- [x] ESLint clean (`--max-warnings 0`)

Co-authored with Claude (Anthropic).


bhat-ganesh:
I tested this against a real cluster and the deep links worked 🎉 , attaching screenshots

image

image

image

illume (changes requested):
Thanks for these changes.


<details>

</details>


<details>

Good examples:

</details>


<details>

</details>


I noticed the GitHub CI frontend lint job is failing. Can you please fix the lint errors? You can run `cd frontend && npm run lint` locally to see the errors.

<details>

</details>

frontend/src/components/common/Resource/LogsButton.tsx
```diff
@@ -529,19 +529,23 @@ function LogsButtonContent({ item }: LogsButtonProps) {
   );
 }
 
+export function launchWorkloadLogs(item: KubeObject) {
```
gambtho: Could we have launchWorkloadLogs accept a dispatchHeadlampEvent callback? As written, this helper opens the activity but doesn't fire HeadlampEventType.LOGS (similar to what pod/Details.tsx::launchLog does)

bhat-ganesh:
Thanks for the review @illume, @gambtho, and @copilot-pull-request-reviewer! I've pushed a revised version that addresses each point. Force-push notes below.

### @illume
- **Rebased against main** — force-pushed a single linear commit on top of latest `main`, no more merge commits.
- **Commit message style** — rewrote the title from conventional-commits (`feat(workload): ...`) to the project's `<area>: <subarea>: <description>` style (`frontend: workload/Details: Support ?view=logs deep-link`). Expanded the body to explain intent and the three pieces of the change.
- **Lint** — confirmed `cd frontend && npm run lint` is clean locally (exit 0, no warnings).
- **Snapshot mismatches** — I looked into this carefully: I ran `src/storybook.test.tsx` against plain `main` (no PR changes) and got the same 115 snapshot failures. They're not caused by this PR. I didn't want to update them here because that would bundle unrelated visual drift into my commit and attribute it to my feature. Happy to do a separate snapshot-refresh PR if that's useful, but I'd rather not sneak it into this one. Let me know if you'd prefer I include it anyway.

### Copilot reviewer
- **Good catch on `Job`** — yes, `WorkloadDetails` is also routed for `Job` and `LogsButtonContent` doesn't support Jobs. Fixed: lifted `LOGGABLE_KINDS = ['Deployment', 'ReplicaSet', 'DaemonSet', 'StatefulSet']` to module scope and gated the auto-launch effect on it. Added a regression test (`does not launch logs for non-loggable workload kinds (e.g. Job)`) to cover the case. `?view=logs` on a Job URL is now a no-op rather than opening a broken activity.

### @gambtho
- **`dispatchHeadlampEvent` parity with `pod/Details.tsx::launchLogs`** — good spot. Updated `launchWorkloadLogs` to accept an optional `dispatchHeadlampEvent` callback and fire `HeadlampEventType.LOGS` / `EventStatus.OPENED` on launch, matching the pod precedent. Both call sites (click path in `LogsButton` and deep-link path in `WorkloadDetails`) now wire in `useEventCallback()` and pass the dispatcher through. Added a unit test asserting the event fires when a dispatcher is passed. Previously the click path was silently missing this event; this fixes that gap too.

---

All changes verified:
- Unit/integration tests: **14 passing** (LogsButton: 9, WorkloadDetails: 5, incl. new Job regression and LOGS-event dispatch tests)
- ESLint clean (`--max-warnings 0`)
- Re-verified manually against a live cluster after the rebase (Deployment, StatefulSet, DaemonSet all launch; Job is a no-op)

bhat-ganesh:
Thanks again for the second pass @illume @gambtho and @copilot-pull-request-reviewer! Pushed commit [`e0fafdc`](https://github.com/kubernetes-sigs/headlamp/commit/e0fafdc83) addressing both Copilot comments from the last review round.

### Copilot comment 1 — [type-only imports](https://github.com/kubernetes-sigs/headlamp/pull/5312#discussion_r3169533773)
Switched `Workload` / `WorkloadClass` in `workload/Details.tsx` to `import type { ... }` and combined them into one line. Verified they are type-only exports in `lib/k8s/Workload.ts`.

### Copilot comment 2 — [`launchWorkloadLogs` accepts any `KubeObject`](https://github.com/kubernetes-sigs/headlamp/pull/5312#discussion_r3169533806)
Good catch — the call-site already guarded, but the exported helper's contract was still loose, so any plugin author could call it with a Pod / Job / etc. and get a blank activity.

Fixed with a runtime guard **inside the helper itself**, so the contract is enforced at the API boundary rather than only at the call site:

```ts
export function launchWorkloadLogs(
  item: KubeObject,
  dispatchHeadlampEvent?: (event: HeadlampEvent) => void
) {
  if (!LOGGABLE_WORKLOAD_KINDS.has(item.kind)) {
    return;
  }
  // ... existing body
}
```

Also exported `LOGGABLE_WORKLOAD_KINDS` as a `ReadonlySet<string>` so:
- `workload/Details.tsx` no longer maintains its own duplicate `LOGGABLE_KINDS` array — both files now reference one source of truth (dedup win; the two were a drift bug waiting to happen).
- Plugin authors who want to pre-check before calling the helper can do so without hard-coding the kind list.

Went with the runtime-guard approach (over tightening the `KubeObject` parameter to a union type) to keep the helper callable from any `KubeObject`-typed context — matches the pattern of the call-site gate already in place. Happy to switch to a stricter signature if maintainers would prefer that.

### Testing
- New regression test: `launchWorkloadLogs no-ops for unsupported workload kinds (e.g. Job)` — asserts both `Activity.launch` and the LOGS event dispatcher are **not** called for unsupported kinds.
- Updated `MockKubeObject` to expose the `kind` getter the guard reads.
- Regenerated `pluginLib.snapshot` to include the new exported `LOGGABLE_WORKLOAD_KINDS` symbol on the plugin API surface (2 occurrences, both at the same export path).

All checks green locally:
- 16/16 tests passing (LogsButton: 11, WorkloadDetails: 5)
- `npm run lint` clean (exit 0, no warnings)
- Snapshot diff is scoped to the new symbol only

No rebase needed — this is a regular push on top of [`a6be68f`](https://github.com/kubernetes-sigs/headlamp/commit/a6be68fd8) so the previous review diff is still visible separately.

bhat-ganesh:
Thanks for the follow-up catch @copilot-pull-request-reviewer! Pushed [`31faadb`](https://github.com/kubernetes-sigs/headlamp/commit/31faadbdf) addressing the [cross-bundle `instanceof` concern](https://github.com/kubernetes-sigs/headlamp/pull/5312#discussion_r3170xxxxx).

### Why this was a real bug for the programmatic path
Exactly as flagged — my previous commit guarded `launchWorkloadLogs` on `item.kind` (cross-bundle safe), but `LogsButtonContent` internally used `instanceof Deployment/ReplicaSet/DaemonSet/StatefulSet` in three places. A plugin passing a `Deployment` from its own bundle would:
1. Pass `LOGGABLE_WORKLOAD_KINDS.has(item.kind)` ✅
2. Open the Activity ✅
3. Fail `instanceof` inside `getRelatedPods` and `onMount` ❌
4. Result: Activity opens, stays empty silently.

### Fix
Introduced a local `isLoggableWorkload` type guard that uses `KubeObject.isClassOf` — which the [KubeObject docstring explicitly documents](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/KubeObject.ts#L145-L147) as the canonical cross-bundle alternative to `instanceof`:

```ts
function isLoggableWorkload(
  item: KubeObject | null
): item is Deployment | ReplicaSet | DaemonSet | StatefulSet {
  return (
    !!item &&
    (Deployment.isClassOf(item) ||
      ReplicaSet.isClassOf(item) ||
      DaemonSet.isClassOf(item) ||
      StatefulSet.isClassOf(item))
  );
}
```

Replaces all three `instanceof` chains in `LogsButtonContent` (`getRelatedPods` x2, `onMount` x1). As a drive-by, also simplified the resource-type lookup inside `getRelatedPods` — the old code chained `instanceof` again just to map to a lowercase string; now it reads `item.kind.toLowerCase()` directly.

### Scope note
Left the button-render gate at the bottom of `LogsButton` using `instanceof`. That path only runs on host-rendered workload detail pages where class identity is stable (host code import host's class), so it's not part of the programmatic-plugin failure mode. Happy to convert that one too in a follow-up if reviewers prefer.

### Testing
- New regression test **`LogsButtonContent fetches pods for a cross-bundle Deployment (isClassOf)`** — constructs a `ForeignDeployment` class with identical `kind: 'Deployment'` but a distinct class identity (the exact cross-bundle scenario), asserts `foreign instanceof Deployment === false` as a precondition, then calls `launchWorkloadLogs(foreign)` and verifies `LogsButtonContent` mounts and kicks off its pod fetch via `clusterFetch`. Without `isClassOf`, this test fails.
- Existing tests updated: `MockKubeObject` now has a `static kind` + inherited `isClassOf`; the four class mocks each declare their own `static kind`.

### Verification
- 16/16 tests passing (11 LogsButton + 5 WorkloadDetails), including the new cross-bundle regression
- `npm run build-typedoc` clean (0 errors)
- `npm run lint` clean (exit 0)
- `pluginLib.snapshot` diff vs main unchanged (only the intended additions remain)

bhat-ganesh:
Thanks @copilot-pull-request-reviewer — both catches are fair. Pushed [`cb3558e`](https://github.com/kubernetes-sigs/headlamp/commit/cb3558e3a).

### Comment 1 — inconsistent guards in `launchWorkloadLogs`
Agreed: after the `isClassOf` refactor in [`31faadb`](https://github.com/kubernetes-sigs/headlamp/commit/31faadbdf), `LogsButtonContent` guards with `isLoggableWorkload` (kind + apiGroup), but the launcher itself still only checked `item.kind` via `LOGGABLE_WORKLOAD_KINDS`. A CRD with `kind: "Deployment"` and a non-core apiGroup would pass the launcher and then silently no-op on the pod fetch. Switched the launcher to `isLoggableWorkload` so every gate in this file uses the same cross-bundle-safe check.

`LOGGABLE_WORKLOAD_KINDS` remains exported — plugin authors can still use it as a lightweight pre-check without pulling in `isClassOf`/class imports.

### Comment 2 — dead `vi.mock` for a type-only import
Agreed: `Workload` / `WorkloadClass` in `workload/Details.tsx` are now `import type { ... }`, so the module is erased at build time and the stub never runs. Removed.

### Verification
- 16/16 tests passing
- `tsc --noEmit` clean
- `npm run lint` clean
- Snapshot diff vs main unchanged

illume (approved):
This is great. Thanks a lot! 🎉

PR #5321, opened 2026-04-30
Add missing details views info

joaquimrocha:
## Summary

This PR adds missing details to different resources' details views. The additional info was based on what kubectl gives.
E.g. the Service details view was not showing the port when a service is a NodePort, except in the list view.

## Steps to Test

Verify that the details views are working fine and that they show now similar info to what kubectl describe does, without being too overwhelming.


joaquimrocha:
I think the snapshots are showing changes that shouldn't be here. Probably the node version differences still.
I will fix it.

sniok (approved):
thanks!

/lgtm

PR #5328, opened 2026-05-01
frontend: Fix useEvict setting not persisting as false

### LINKED ISSUE #5325: bug: "Use evict for pod deletion" setting resets to enabled on every page refresh
## Describe the bug

The "Use evict for pod deletion" (`useEvict`) setting in Headlamp's General Settings cannot be persisted as `false`. When a user disables this toggle and refreshes the page, the setting silently resets back to `true` (enabled). This makes it impossible to opt out of the eviction behavior for pod deletion.

The root cause is a JavaScript logical OR (`||`) bug in `frontend/src/redux/configSlice.ts` (line 92):

```js
useEvict: storedSettings.useEvict || true,
```

In JavaScript, `false || true` evaluates to `true` because `false` is falsy. So when the stored value is explicitly `false`, the `||` operator ignores it and falls through to `true`. The correct fix is to use the nullish coalescing operator (`??`) instead:

```js
useEvict: storedSettings.useEvict ?? true,
```

### Expected behavior
When a user disables the "Use evict for pod deletion" toggle in Settings and refreshes the page, the setting should remain `false` (disabled). Deleting a Pod should use the standard `DELETE` API, not the `Evict` API.

### Actual behavior
The `useEvict` setting always resets to `true` on every page load, regardless of the user's stored preference. This means:
- The Delete button for Pods always shows **"Evict"** instead of **"Delete"**
- Pod deletion always uses the Evict API, which respects PodDisruptionBudgets and may prevent deletion of stuck/unresponsive pods
- Users who specifically need to force-delete stuck pods cannot do so through the UI without toggling the setting *every session*

## To Reproduce
Steps to reproduce the bug:
1. Open Headlamp and navigate to **Settings** (gear icon in the top right → General Settings)
2. Find the **"Use evict for pod deletion"** toggle and turn it **OFF** (disabled)
3. Refresh the page (F5 or Cmd+R)
4. Go back to Settings and observe the **"Use evict for pod deletion"** toggle has silently turned itself **ON** again
5. Navigate to any Pod detail page and click the Delete icon — it will show **"Evict"** instead of **"Delete"**

 - Other: Reproducible on all platforms. The bug exists in `frontend/src/redux/configSlice.ts` line 92.

Yes (I will propose a PR)

### Root Cause (one-line fix)

**File:** `frontend/src/redux/configSlice.ts`, line 92

```diff
- useEvict: storedSettings.useEvict || true,
+ useEvict: storedSettings.useEvict ?? true,
```

The `||` (logical OR) operator treats `false`, `0`, `""`, `null`, and `undefined` as falsy and falls through to the default. The `??` (nullish coalescing) operator only falls through on `null` or `undefined`, correctly preserving an explicit `false` value from `localStorage`.

### Impact
This bug affects the **Pod deletion flow** across the entire UI:
- `DeleteButton.tsx` — Uses `settingsObj.useEvict` to decide between `pod.evict()` and `item.delete()`
- `DeleteMultipleButton.tsx` — Same evict-vs-delete logic for bulk pod deletion
- The button label, dialog title, and confirmation text all change based on this setting

### Is this a regression?
This bug has existed since the `useEvict` setting was introduced. It was never possible to persist `useEvict = false` across page reloads.

Utkarshpandey0001:
## Summary

This PR fixes a bug where the "Use evict for pod deletion" setting could not be persisted as disabled across page reloads.

## Related Issue

Fixes #5325 

## Changes

- Fixed a logical OR (`||`) bug in `frontend/src/redux/configSlice.ts` by replacing it with the nullish coalescing operator (`??`). This ensures that an explicit `false` value from `localStorage` is preserved instead of falling through to the default `true`.

## Steps to Test

1. Navigate to Settings (gear icon in the top right).
2. Under "General Settings", toggle **OFF** "Use evict for pod deletion".
3. Refresh the browser page (F5 or Cmd+R).
4. Navigate back to Settings and verify that the toggle is still **OFF**.
5. Navigate to a Pod details page and click the Delete icon. Verify the confirmation dialog says "Delete" instead of "Evict".

## Notes for the Reviewer

- The `||` operator treats `false` as falsy, which caused the setting to always revert to `true` upon initialization from localStorage. The `??` operator correctly only falls back to `true` when the stored value is `null` or `undefined`.

### DIFF
--- a/frontend/src/redux/configSlice.ts
+++ b/frontend/src/redux/configSlice.ts
@@ -89,7 +89,7 @@ export const initialState: ConfigState = {
       storedSettings.tableRowsPerPageOptions || defaultTableRowsPerPageOptions,
     timezone: storedSettings.timezone || defaultTimezone(),
     sidebarSortAlphabetically: storedSettings.sidebarSortAlphabetically || false,
-    useEvict: storedSettings.useEvict || true,
+    useEvict: storedSettings.useEvict ?? true,
   },
 };
 


illume (approved):
Thanks! 🎉

PR #5329, opened 2026-05-01
frontend: apiDiscovery: Log core API discovery failures

Joshna907:
### **Changes:**

1. frontend/src/lib/k8s/api/v2/apiDiscovery.tsx: Replaced the empty catch block in the core API discovery loop with a console.debug statement. Failures are now properly logged with the cluster name and API version details, improving diagnosability without spamming the main UI with error toasts.

### **Steps to Test:**

1. Run npm run frontend:test to ensure no regressions.
2. Verify logging: Temporarily block or simulate a network failure for a core API fetch request (e.g., block the /api/v1 request in the browser's network tab).
3. Open the browser's developer tools and verify that a helpful message (Failed to fetch core API resources for cluster...) appears in the console instead of the failure happening silently.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/apiDiscovery.tsx
+++ b/frontend/src/lib/k8s/api/v2/apiDiscovery.tsx
@@ -204,7 +204,12 @@ export async function apiDiscovery(clusters: string[]): Promise<ApiResource[]> {
                 }
               }
             } catch (e) {
-              // This catch block intentionally left blank.
+              // Log failure to fetch core API resources for better observability.
+              console.debug(
+                `Failed to fetch core API resources for cluster ${cluster}:`,
+                { version: v },
+                e
+              );
             }
           });
           await Promise.allSettled(coreResourceFetchPromises);


Joshna907:
ptal @illume

illume:
> This catch block intentionally left blank

I'm assuming this is here because with version retries it's quite common on different clusters for the. What does git blame on that say? (in the Pr the introduced the comment too?)

Joshna907:
> > This catch block intentionally left blank
> 
> I'm assuming this is here because with version retries it's quite common on different clusters for the. What does git blame on that say? (in the Pr the introduced the comment too?)

I checked git blame. That comment was added in PR #3228 when the apiDiscovery.tsx file was originally implemented.
Your assumption is completely correct! It was left blank because 404s and errors are very common when trying to discover legacy or unsupported API groups across different clusters.
However, because it completely swallows the error, it also hides genuine issues (like RBAC permission denials or network timeouts for Core APIs like /api/v1). This makes it extremely difficult to debug when core resources like Pods or Nodes are silently missing from the UI.
I changed it to console.debug() specifically so that we don't spam console.error (or the UI) with expected 404s, but developers can still see what happened in the debug logs if they are investigating missing resources. Does that approach sound okay to you, or would you prefer I add an if check to specifically ignore 404 Not Found errors? @illume

illume:
@Joshna907 if you find it useful for debugging, then it's fine by me. Thanks for the explanation.

illume (approved):
🎉 thanks!

PR #5335, opened 2026-05-01
frontend: docs: Makefile: ci: Enable react compiler in vscode, and precommit hook

illume:
## Summary

This PR makes `npm run lint` fast and quiet for the frontend by keeping the slow React Compiler `react-hooks/*` rules off the lint fast path, while still enforcing them everywhere they fit cheaply: the husky pre-commit hook (changed files only), the VSCode editor (open file only), and CI (full tree, parallel job). It also renames `ci-lint` → `lint:ci` for consistency, adds an autofix counterpart, fixes a real lint bug where local copies of `frontend/src/` under `plugins/headlamp-plugin/src/` were being scanned, and documents the rationale.

`npm run lint` is the **fast path** the "quick check" developers run constantly while editing, switching branches, or rewriting commits. Several common workflows defeat the ESLint cache (e.g. `git rebase -I` to clean up PR commits, switching to a fresh worktree, CI without a cache hit, bisecting), so the uncached time is what actually gets
experienced. A 1-second lint stays in the flow of work; a 15-second lint does not, and developers start skipping it. It is even more true on common Windows lower-specced machines where it can instead take 2.5 minutes to run lint with the React Compiler enabled! Consider that people might run that for bisecting 20 commits… and it adds up.

The react compiler is wired into:
- CI (.github/workflows/frontend.yml → frontend:lint:slow)
- husky lint-staged pre-commit hook
- root package.json frontend:lint:slow / frontend:lint:slow:fix scripts
- vscode uses react compiler so all warnings shown in IDE

---

Github Action time and memory savings.

  | Before | After
-- | -- | --
npm run lint elapsed | 9.40 s | 8.93 s
npm run lint peak RSS | 666 MB | 658 MB
npm run lint:slow elapsed | 22.70 s | 21.23 s
npm run lint:slow peak RSS | 1380 MB | 715 MB

Peak RSS dropped ~665 MB because the legacy hooks rules no longer get re-declared on top of the compiler analysis.

## Related

Previous react compiler PRs:

- react compiler enabled, many warnings in terminal, 5x slow "npm run lint", CI warning quality gate turned off  https://github.com/kubernetes-sigs/headlamp/pull/5155
- disable warnings when running "npm run lint", make issues for each disabled lint issue, run react compiler in CI and separate local ci-lint command locally, no react compiler warnings in vscode, no react compiler checks in precommit hook: https://github.com/kubernetes-sigs/headlamp/pull/5196

## Changes

- **`frontend/package.json`**: rename `ci-lint` → `lint:ci`; add `lint:ci:fix` (eslint `--fix` + prettier `--write`); keep `lint` fast (only `react-hooks/rules-of-hooks` enabled, no `--max-warnings 0`).
- **`package.json` (root)**: add `frontend:lint:ci` and `frontend:lint:ci:fix` passthroughs; root `frontend:lint` runs `npm run lint && npm run format-check` — no `lint:ci`, no `--max-warnings 0`, no slow `react-hooks/*` rules.
- **`frontend/.eslintignore`**: ignore the gitignored copies of `frontend/src/` populated under `plugins/headlamp-plugin/src/{assets,components,helpers,i18n,lib,plugin,redux,resources,stateless,react-app-env.d.ts}` by the plugin's setup scripts. ESLint does not honor `.gitignore`, so without these entries users with the copies present saw `react-hooks/*` warnings on `plugins/headlamp-plugin/src/lib/k8s/node.ts`, `…/lib/router/index.tsx`, etc.
- **`frontend/src/`**: 177 inline `// eslint-disable-next-line react-hooks/…` comments across 94 files, grandfathering the 195 pre-existing `react-hooks/*` violations so `lint:ci` is green from day one. New violations are still flagged.
- **`docs/development/frontend.md`**: documents `lint` vs `lint:ci` vs `lint:ci:fix`, lists the React Hooks rules enabled in CI, and adds a rationale section for keeping the React Compiler rules out of the fast path. Notes that those rules are still checked in pre-commit, VSCode and CI — but not in `npm run lint`, because pre-commit and VSCode operate on a small set of files at a time while `lint` runs over the whole tree and must stay fast.

Assisted by copilot.

## Steps to Test

Manual verification of the lint fast path / slow path split:

2. **Fast path (`npm run lint`) — should be quiet and fast:**
   ```bash
   cd frontend && rm -rf node_modules/.cache && time npm run lint
   ```
   Expected: exits 0, no warnings printed, ~1-5s on a modern laptop, ~30s+ on slower hardware. From the repo root, `npm run frontend:lint` should also be clean.
3. **Slow path (`npm run lint:ci`) — strict, all `react-hooks/*` rules on, `--max-warnings 0`:**
   ```bash
   cd frontend && rm -rf node_modules/.cache && time npm run lint:ci
   ```
   Expected: exits 0, no warnings printed. Same for `npm run frontend:lint:ci` from the root.
4. **Autofix variant:**
   ```bash
   cd frontend && npm run lint:ci:fix
   ```
   Expected: exits 0, runs ESLint `--fix` then Prettier `--write`, no diff produced on a clean tree.
5. **Pre-commit / VSCode**: stage a frontend `.tsx` file with a deliberate `react-hooks/exhaustive-deps` violation and confirm the husky pre-commit hook (via `lint-staged`) blocks the commit, and that VSCode shows the warning inline.
6. **CI**: `.github/workflows/frontend.yml` now uses `npm run lint:ci`; the lint job should be green on this PR.
7. **Existing test suite still passes:**
   ```bash
   npm run frontend:test
   ```

N/A — no UI changes.

## Notes for the Reviewer

- The slow `react-hooks/*` rules are still enforced in four places — pre-commit, VSCode, locally with npm run lint:slow and in CI — just not in the bulk `npm run lint`. That mirrors how `tsc`, `format-check`, and integration tests are also kept off the fast paths.


docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
 This shows an alert when an a11y issue is detected.
 
+## Linting
+
+The frontend has two lint modes that share the same ESLint base config but
+differ in which rules are enforced and how strictly warnings are treated.
+
+The slow React Compiler `react-hooks/*` rules are checked in three places:
+the husky **pre-commit hook**, the **VSCode** editor, and **CI** — but
+**not** in `npm run lint`. The pre-commit hook and VSCode each operate on
```
sniok, line 108: if the rules are not part of default eslint config then contributors might not see issues while coding. this might be problematic for external contributors, people who are not using vscode

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
 
 This shows an alert when an a11y issue is detected.
 
+## Linting
+
+The frontend has two lint modes that share the same ESLint base config but
```
sniok, line 103: I'm not sure if this is a good approach since it's quite unusual and can be confusing when npm run lint passes locally but fails in ci

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
+differ in which rules are enforced and how strictly warnings are treated.
+
+The slow React Compiler `react-hooks/*` rules are checked in three places:
+the husky **pre-commit hook**, the **VSCode** editor, and **CI** — but
+**not** in `npm run lint`. The pre-commit hook and VSCode each operate on
+a small set of files at a time so the cost is negligible there, whereas
```
sniok, line 109: if linting in IDEs is fast per file that supports keeping the rules enabled in the default config

the full codebase linting can be handled with cache or a separate fast command if really needed
I don’t think we should solve that by making the default rules weaker

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
+`npm run lint` runs over the whole tree and needs to stay fast.
+
+### `npm run lint` (local development)
+
+The fast, permissive mode you run while editing:
```
sniok, line 114: calling `npm run lint` permissive mode is confusing since most developers expect npm run lint to match what CI enforces

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
+It uses the ESLint config in `frontend/package.json` (`eslintConfig`).
+Only `react-hooks/rules-of-hooks` is enabled (as `warn`); the rest of the
+slow `react-hooks/*` rules are off here.
+
+### `npm run lint:slow` (CI / pre-commit / strict)
+
+Run this before opening a PR, or rely on it via `husky` pre-commit and CI:
+
+```bash
+cd frontend && npm run lint:slow
```
sniok, line 132: if this is the real lint rules that will be enforced, I would prefer not to hide it behind lint:slow

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
+`lint:slow` is the strict mode used by:
+
+- the GitHub Actions workflow `.github/workflows/frontend.yml`
+- the `make frontend-lint` target
+- the `husky` pre-commit hook through `lint-staged`
+- developers verifying their work locally before pushing
+
+It differs from `npm run lint` in two ways:
+
+1. It points ESLint at a dedicated config, `frontend/.eslintrc.slow.cjs`, which
```
sniok, line 154: second config adds more maintenance, configs can drift over time. a single config is simpler to understand and less likely to surprise contributors

docs/development/frontend.md
```diff
@@ -98,6 +80,141 @@ REACT_APP_SKIP_A11Y=false npm run frontend:start
...
+`use-memo`. See the
+[plugin docs](https://react.dev/reference/eslint-plugin-react-hooks)
+for what each one catches.
+
+### Why the React Compiler rules are kept out of `npm run lint`
+
+The `react-hooks/*` rules from `eslint-plugin-react-hooks` v7+ are powered
+by the React Compiler. They are excellent at catching real bugs, but they
+are also **roughly 5× slower** than the rest of the lint pass. Turning
+them on for `npm run lint` would push it from ~1 second to ~15 seconds on
```
sniok, line 174: I think we should verify this with different conditions cache on/off, machine, OS and whether this is representative. In my local testing with cache, npm run lint with all rules enabled is around 1s

sniok (changes requested):
I understand that the performance concern is a real problem

but with this change the default local lint command won't match the CI, this is going to be surprising especially for external contributors. `lint:slow` is what is enforced by CI then it's not 'slow' but 'real lint'

I think we should have a single set of rules for all environments, with `npm run lint` meaning run all the eslint rules for this project

Can I suggest that instead we can optimize and introduce a separate path for a quicker linting by linting only staged files with `npm run lint:staged` command?

docs/development/frontend.md
illume: Please read the rest of this.

A number of cases are listed where the cache is not used.

If you want to test more, please test more. I'll leave that to you because you just enabled this without performance testing being posted anywhere.

docs/development/frontend.md
illume: What's worse? A file that changes once every... what two years? Or a penalty on each lint run for every single commit?

docs/development/frontend.md
illume: This is running the react compiler through the linter.

We don't run the TypeScript checker through the linter - because that is a bad idea for performance. We don't run the long running integration tests when running tests, because that is slow and an incredibly bad idea.

Running the react compiler checks in the lint fast path is also a bad idea for the same reason.

docs/development/frontend.md
illume: Please read the PR description and docs for why the cache is not suitable.

docs/development/frontend.md
illume: Please read the docs in this PR.

Running separate slow compiler checks out of the faster lint path is not at all unusual and already done.

docs/development/frontend.md
illume: You're saying they will miss the IDE warnings, and the prerecommit hook for these rarely occurring rules?

For those that miss those, they will be shown in the GitHub PR.

Patches for supporting other IDEs gladly accepted.

illume:
> I understand that the performance concern is a real problem
> 
> but with this change the default local lint command won't match the CI, this is going to be surprising especially for external contributors. `lint:slow` is what is enforced by CI then it's not 'slow' but 'real lint'
> 
> I think we should have a single set of rules for all environments, with `npm run lint` meaning run all the eslint rules for this project
> 
> Can I suggest that instead we can optimize and introduce a separate path for a quicker linting by linting only staged files with `npm run lint:staged` command?

You seem to have missed reading some of the PR description and docs. Please see my responses.

We can also disable the react compiler.

docs/development/frontend.md
illume: >calling npm run lint permissive mode is confusing since most developers expect npm run lint to match what CI enforces

Can you back this claim up with anything?

The error found by the react compiler are relatively rare. There's IDE support, a precommit hook, as well as the error being shown in GitHub.

PR #5398, opened 2026-05-05
backend: externalproxy: Forward status code and content type

ayushmaan-16:
### Description
I noticed that the `/externalproxy` handler was not sending back the correct status codes and `Content-Type` from the external server. Because of this, the frontend was always getting a `200 OK` response, even if the external request had actually failed (like a 404 or 500 error). This was making it a bit confusing for the UI to show the right error message to the user.

### Changes made
- I have updated the code in `headlamp.go` so that it now forwards the actual status code using `w.WriteHeader`. 
- I also added a small logic to copy the `Content-Type` header from the external response so the browser knows exactly what kind of data it is receiving.

### Tests verified
- I have built the backend properly using `npm run backend:build` and it is working fine.
- I also ran the full tests with `npm run backend:test`. All the relevant tests have passed, so we are good to go.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -647,10 +647,15 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 			return
 		}
 
+		if contentType := resp.Header.Get("Content-Type"); contentType != "" {
+			w.Header().Set("Content-Type", contentType)
+		}
+
+		w.WriteHeader(resp.StatusCode)
+
 		_, err = w.Write(respBody)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Oh! Thanks for this.

Would you mind adding a test case?

ayushmaan-16:
@illume I have added a test case in headlamp_test.go, this new test TestExternalProxyForwarding verifies that both the HTTP status code and the Content-Type header are correctly forwarded by the proxy. I have also checked it locally and it is passing properly.
image

ayushmaan-16:
@illume done with the changes. you can review it when you get the time :)

illume (approved):
🎉 thanks a lot!

PR #5400, opened 2026-05-05
docs: frontend: resourceMap: Deduplicate merged graph elements

### LINKED ISSUE #5364: resourceMap: Add source ownership for overlapping nodes

Resource map sources can currently overlap. If a plugin source and a built-in source both provide the same Kubernetes object, the map can show duplicate nodes or duplicate relationships.

This is a problem for plugins that want to provide a focused view of built-in resources. For example, a plugin may define custom workload resources and want its map source to show only the Pods owned by those workloads. That makes the plugin map useful on its own, because users can enable the plugin source and see the relevant custom resources and Pods together.

However, if the user also enables Headlamp’s built-in `Workloads > Pods` source, those same Pods may be added again by the built-in source. The map then has overlapping nodes for the same Kubernetes objects, which can make the graph harder to understand.

If the plugin avoids adding those Pods, there is no duplication, but users must enable the broad built-in Pods source to see them. That can add many unrelated Pods and make the plugin-specific map less focused.
If the plugin removes its Pods source, there is no duplication, but users must enable `Workloads > Pods` to see Job Pods, which shows all cluster Pods rather than only Volcano Pods.

Add a generic way for the resource map to handle overlapping nodes across sources.

This could be implemented as one of the following, or another design maintainers prefer:
- Deduplicate nodes by Kubernetes object UID when multiple enabled sources provide the same object.
- Allow a source to declare ownership of a subset of nodes from another source.
- Allow plugin sources to provide filtered subsets of built-in resources without duplicating nodes when the built-in source is also enabled.

The goal is to let plugin authors build focused map views that include relevant built-in resources, while keeping the graph clean when users enable multiple overlapping sources.

Plugin developers and users of plugin-provided map sources.

Plugin users would be able to enable a plugin map source and see the resources relevant to that plugin without needing to enable broad built-in sources that include unrelated resources.
Plugin developers would have a cleaner way to model relationships involving built-in resources such as Pods, Nodes, Services, Jobs, or HPAs.

Yes. I can help investigate the design and propose a PR after discussing the preferred approach with maintainers.

## Additional context
A common case is a plugin that defines custom workload resources but also wants to show the native Pods created by those workloads. The plugin can provide a filtered Pods source, but that source may overlap with Headlamp’s built-in Pods source if both are enabled.

Related Images from WIP Volcano plugin map:
pods enabled from the plugin source only: 

Image

pods enabled from the plugin source + Workloads -> Pods:

Image

mahmoudmagdy1-1:
## Summary
Deduplicate resource map graph nodes and edges after selected sources and relations are merged.
## Related Issue
Fixes #5364
## Changes
- Added graph element deduplication by `GraphNode.id` and `GraphEdge.id`.
- Preserves the first graph element when duplicate IDs are returned by selected sources.
- Applies deduplication before filtering, grouping, lookup, layout, and rendering.
- Added tests for duplicate nodes, duplicate edges, and distinct edges with the same source/target.
- Documented duplicate ID behavior.
## Steps to Test
1. Open the resource map with overlapping sources enabled.
2. Verify duplicate node or edge IDs render as a single graph element.
Before (pods enabled from the plugin source + Workloads -> Pods):
image
After (pods enabled from the plugin source + Workloads -> Pods):
image

### DIFF
--- a/frontend/src/components/resourceMap/graph/graphModel.test.ts
+++ b/frontend/src/components/resourceMap/graph/graphModel.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { getNodeWeight, GraphNode } from './graphModel';
+import { deduplicateGraphElements, getNodeWeight, GraphNode } from './graphModel';
 
 describe('getNodeWeight', () => {
   it('correctly assigns default weights to different Kubernetes resource types', () => {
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

- `docs: frontend: resourceMap: Deduplicate merged graph elements` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>


mahmoudmagdy1-1:
> Thanks for the contribution.
> 
> Could you take a look at the commit messages in this PR? We follow a Linux kernel style for git commits
> There are some open Copilot review comments — could you take a look at them? Please mark each one as resolved once you've addressed it.

Thanks for reviewing this, I've addressed the copilot comment and fixed the commit message

ashu8912 (approved):
LGTM!!

frontend/src/components/resourceMap/graph/graphModel.test.ts
```diff
@@ -58,3 +58,91 @@ describe('getNodeWeight', () => {
...
+          icon: 'second icon',
+          kubeObject: { kind: 'Pod', metadata: { name: 'second' } } as any,
+          data: { queue: 'other' },
+          weight: 1000,
+        },
+      ],
+      []
+    );
+
+    expect(nodes[0]).toMatchObject({
```
illume, line 109: Should nodes.length be 1?

skoeva: fixed this

skoeva (approved):
added an additional test and reorganized the line in the docs. looks good now

illume (approved):
Thanks!

PR #5406, opened 2026-05-06
frontend: AlertNotification: Fix interval leak on mount

### LINKED ISSUE #5375: Bug: Memory leak and duplicate setInterval polling in AlertNotification
## Describe the bug

The `AlertNotification` component in Headlamp leaks a `setInterval` timer on mount, causing the application to repeatedly poll the cluster health endpoint (or other checker functions) in the background indefinitely. This results in unnecessary network traffic, degraded performance over time, and a silent memory/resource leak.

The root cause is the presence of two conflicting `useEffect` hooks that both attempt to manage the same `setInterval` polling loop.

**File:** `frontend/src/components/common/AlertNotification.tsx`, lines 74–102

```tsx
  React.useEffect(
    () => {
      const id = registerSetInterval();
      setIntervalID(id);
      return () => clearInterval(id);
    },
    // eslint-disable-next-line
    []
  );

  React.useEffect(
    () => {
      if (intervalID) {
        clearInterval(intervalID);
      }
      const id = registerSetInterval();
      setIntervalID(id);
      return () => clearInterval(id);
    },
    // eslint-disable-next-line
    [networkStatusCheckTimeFactor]
  );
```

On initial mount, both `useEffect` hooks execute. The first hook creates an interval and queues a state update (`setIntervalID(id)`). The second hook also executes, observes that `intervalID` is currently `null` (because state updates are asynchronous), creates a **second** interval, and queues another state update. 

The state is ultimately overwritten by the second interval's ID. Because the first interval's ID is lost and its cleanup function only runs on component unmount, the first interval is never cleared while the component remains mounted. It continues to poll every 5000ms indefinitely.

### Expected behavior
The component should only create a single interval at any given time. The polling interval should be properly managed and cleared when the polling frequency (`networkStatusCheckTimeFactor`) changes or when the component unmounts.

### Actual behavior
Two intervals are created immediately upon mount. One of them becomes completely orphaned (its ID is lost from state) and continues running in the background indefinitely, hitting the cluster health check endpoint every 5 seconds, causing silent background resource consumption.

## To Reproduce
Steps to reproduce the bug:
1. Open Headlamp and navigate to any cluster view (e.g., the Cluster Overview page).
2. Open the browser's Developer Tools and switch to the **Network** tab.
3. Observe the network requests.
4. Notice that the cluster health endpoint (or the network check) is being polled **twice** as frequently as expected. Alternatively, place a `console.log` inside `registerSetInterval`'s callback and observe that it is firing twice per cycle.

 - Other: Reproducible on all platforms. The bug exists in `frontend/src/components/common/AlertNotification.tsx`.

Yes (I will propose a PR)

### Proposed Fix
The first `useEffect` hook (with the empty dependency array) is completely redundant and causes the race condition. It should be removed entirely. The second `useEffect` hook correctly handles both the initial mount (when `networkStatusCheckTimeFactor` is `0`) and subsequent updates.

```diff
-  React.useEffect(
-    () => {
-      const id = registerSetInterval();
-      setIntervalID(id);
-      return () => clearInterval(id);
-    },
-    // eslint-disable-next-line
-    []
-  );

  React.useEffect(
    () => {
-      if (intervalID) {
-        clearInterval(intervalID);
-      }
      const id = registerSetInterval();
      setIntervalID(id);
      return () => clearInterval(id);
    },
    // eslint-disable-next-line
    [networkStatusCheckTimeFactor]
  );
```
Removing the first hook and removing the unsafe `if (intervalID) clearInterval(intervalID)` (since `useEffect` already returns a cleanup function) will completely resolve the leak.

Utkarshpandey0001:
## Summary

This PR fixes a bug in the `AlertNotification` component by preventing an orphaned `setInterval` from leaking on component mount, which was causing unnecessary double-polling of the cluster health check in the background.

## Related Issue

Fixes #5375

## Changes

- Fixed a silent memory and network resource leak by removing a redundant `useEffect` hook in `frontend/src/components/common/AlertNotification.tsx` that created a race condition with the interval loop.
- Refactored `AlertNotification` logic to remove the manual `clearInterval` call, instead relying entirely on React's native `useEffect` cleanup function.
- Removed the now-unused `intervalID` state entirely to clean up the code, which successfully satisfies linting constraints.

## Steps to Test

1. Launch Headlamp and navigate to any cluster view (e.g., the Cluster Overview page).
2. Open your browser's Developer Tools and switch to the **Network** tab.
3. Observe the network requests hitting the cluster health check endpoint.
4. Verify that the endpoint is only being polled exactly once every 5 seconds (5000ms), rather than twice concurrently. 

## Notes for the Reviewer

By removing the `intervalID` state alongside the redundant `useEffect`, the component state is much leaner and more idiomatic to React's lifecycle.

### DIFF
--- a/frontend/src/components/common/AlertNotification.tsx
+++ b/frontend/src/components/common/AlertNotification.tsx
@@ -40,7 +40,7 @@ const ROUTES_WITHOUT_ALERT = ['login', 'token', 'settingsCluster'];
 export function PureAlertNotification({ checkerFunction }: PureAlertNotificationProps) {
   const [networkStatusCheckTimeFactor, setNetworkStatusCheckTimeFactor] = React.useState(0);
   const [error, setError] = React.useState<null | string | boolean>(null);
-  const [intervalID, setIntervalID] = React.useState<NodeJS.Timeout | null>(null);
+
   const { t } = useTranslation();
   const { pathname } = useLocation();
 
@@ -71,16 +71,6 @@ export function PureAlertNotification({ checkerFunction }: PureAlertNotification
     }, (networkStatusCheckTimeFactor + 1) * NETWORK_STATUS_CHECK_TIME);
   }
 
-  React.useEffect(
-    () => {
-      const id = registerSetInterval();
-      setIntervalID(id);
-      return () => clearInterval(id);
-    },
-    // eslint-disable-next-line
-    []
-  );
-
   // Make sure we do not show the alert notification if we are not on a cluster route.
   React.useEffect(() => {
     if (!getCluster()) {
@@ -90,11 +80,7 @@ export function PureAlertNotification({ checkerFunction }: PureAlertNotification
 
   React.useEffect(
     () => {
-      if (intervalID) {
-        clearInterval(intervalID);
-      }
       const id = registerSetInterval();
-      setIntervalID(id);
       return () => clearInterval(id);
     },
     // eslint-disable-next-line


skoeva (changes requested):
hi, thanks for submitting! would you mind editing the title to match our [commit guidelines](https://headlamp.dev/docs/latest/contributing/#2-follow-commit-guidelines)?

Utkarshpandey0001:
@skoeva Hi i changed pr title to match guidelines. Pls have a look ,thanks

skoeva (approved):
lgtm, thanks for this fix!

illume (approved):
🎉 thanks!

PR #5407, opened 2026-05-06
frontend: TimeAgo: Fix stale closure on date change

### LINKED ISSUE #5376: Bug: TimeAgo component stale closure ignores prop updates and displays wrong time
## Describe the bug

The `TimeAgo` component (used within `DateLabel` across the application) suffers from a severe React stale closure bug. If the component receives a new `date` prop after its initial mount, it will completely ignore the new date and continually overwrite the UI with the time calculated from the *initial* date prop.

This causes incorrect timestamps to be displayed in the UI when data updates (e.g., in tables or lists that update their data without completely remounting the row components), leading to confusing and misleading timestamps for users.

**File:** `frontend/src/components/common/Label.tsx`, lines 258–272

```tsx
function TimeAgo({ date, format }: { date: number | string | Date; format?: DateFormatOptions }) {
  const [formattedDate, setFormattedDate] = useState<string>(() => timeAgo(date, { format }));

  useEffect(() => {
    const id = setInterval(() => {
      const newFormattedDate = timeAgo(date, { format });
      setFormattedDate(newFormattedDate);
    }, 1_000);

    return () => clearInterval(id);
  }, []);

  return formattedDate;
}
```

Because the `useEffect` hook has an empty dependency array `[]`, the `setInterval` closure captures the `date` and `format` props exactly as they were during the component's first render. 
If the parent component passes a new `date` prop, the interval callback continues to evaluate `timeAgo(initialDate)` and `setFormattedDate` overwrites the component's state with the old calculation every single second.

### Expected behavior
When the `date` or `format` prop passed to `TimeAgo` changes, the component should immediately calculate the time against the new date, and the background interval should seamlessly continue polling based on the new date.

### Actual behavior
When the `date` prop changes, the new prop is entirely ignored by the interval loop. Every 1 second, the interval recalculates the time difference based on the *original* date the component was mounted with, locking the UI into displaying the wrong time for the item.

## To Reproduce
Steps to reproduce the bug:
1. Render a `DateLabel` or `TimeAgo` component with an initial `date` value.
2. Trigger an update in the parent component that passes a new, different `date` value to the same component instance.
3. Observe that for a split second the UI might attempt to update, but the 1-second interval immediately overwrites it with the relative time calculated from the first `date`.
4. (To test via code: temporarily add a button in any component that updates a state variable containing a Date passed into `<DateLabel date={stateDate} />`. Clicking the button will fail to update the displayed TimeAgo effectively).

 - Other: Reproducible on all platforms. The bug exists in `frontend/src/components/common/Label.tsx`.

Yes (I will propose a PR)

### Proposed Fix
The fix is two-fold:
1. Update the dependency array of the `useEffect` so that the interval restarts with the fresh closure whenever `date` or `format` changes.
2. Remove the `useState` entirely, as it's an anti-pattern to mirror props into state when the value can be derived on the fly or just force-updated. However, since we need to tick every second, we can store a dummy "tick" state or simply keep the existing pattern but fix the dependencies.

A simple and robust fix keeping the current pattern:

```diff
  function TimeAgo({ date, format }: { date: number | string | Date; format?: DateFormatOptions }) {
-   const [formattedDate, setFormattedDate] = useState<string>(() => timeAgo(date, { format }));
+   const [formattedDate, setFormattedDate] = useState<string>('');

    useEffect(() => {
+     // Immediately set it so we don't wait 1 second for the first update
+     setFormattedDate(timeAgo(date, { format }));
      const id = setInterval(() => {
        const newFormattedDate = timeAgo(date, { format });
        setFormattedDate(newFormattedDate);
      }, 1_000);

      return () => clearInterval(id);
-     // eslint-disable-next-line react-hooks/exhaustive-deps
-   }, []);
+   }, [date, format]);

    return formattedDate;
  }
```
This correctly handles the `date` prop changing without requiring a full remount of the component, and it satisfies React's exhaustiveness checks.

Utkarshpandey0001:
## Summary

This PR fixes a stale closure bug in the `TimeAgo` component by removing the interval callback's dependency on captured `date` and `format` values. The interval now only triggers a re-render tick, while the displayed relative time is calculated during render from the latest props.

## Related Issue

Fixes #5376

## Changes

- Fixed a stale closure bug in `frontend/src/components/common/Label.tsx` within the `TimeAgo` component where the interval loop could ignore new `date` or `format` props.
- Changed the interval callback to update a tick counter instead of storing a formatted date derived from stale closure values.
- Kept the effect dependency array empty because the effect only manages the timer lifecycle and does not read `date` or `format`.
- Removed the unsafe `as unknown as JSX.Element` assertion and explicitly typed `TimeAgo` as returning `React.ReactNode`, allowing the `timeAgo(...)` string to be returned directly.

## Steps to Test

1. Navigate to a page in the Headlamp UI that utilizes the `DateLabel` or `TimeAgo` component (e.g., a table with creation timestamps).
2. Trigger an update that changes the date of an existing item in the table without completely remounting the row (or test via an explicit state update in a parent component).
3. Observe that the displayed relative time correctly jumps to the time relative to the *new* date immediately, and the 1-second interval ticking continues properly.

## Notes for the Reviewer

- The ESLint rule bypass (`// eslint-disable-next-line react-hooks/exhaustive-deps`) was removed as it is no longer necessary.
- `TimeAgo` intentionally keeps the timer effect independent from `date` and `format`; prop changes naturally re-render the component and the next displayed value is computed from the latest props.

### DIFF
--- a/frontend/src/components/common/Label.test.tsx
+++ b/frontend/src/components/common/Label.test.tsx
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { timeAgo } from '../../lib/util';
+import { TestContext } from '../../test';
+import { DateLabel } from './Label';
+
+vi.mock('../../lib/util', async importOriginal => {
+  const actual = await importOriginal<typeof import('../../lib/util')>();
+
+  return {
+    ...actual,
+    timeAgo: vi.fn((date: string) => `time ago for ${date}`),
+  };
+});
+
+describe('DateLabel', () => {
+  it('updates the relative time when the date prop changes without remounting', () => {
+    const firstDate = '2026-05-08T10:00:00.000Z';
+    const secondDate = '2026-05-08T10:01:00.000Z';
+
+    const { rerender } = render(
+      <TestContext>
+        <DateLabel date={firstDate} />
+      </TestContext>
+    );
+
+    expect(screen.getByText(`time ago for ${firstDate}`)).toBeInTheDocument();
+
+    rerender(
+      <TestContext>
+        <DateLabel date={secondDate} />
+      </TestContext>
+    );
+
+    expect(screen.getByText(`time ago for ${secondDate}`)).toBeInTheDocument();
+    expect(screen.queryByText(`time ago for ${firstDate}`)).not.toBeInTheDocument();
+    expect(timeAgo).toHaveBeenLastCalledWith(secondDate, { format: 'brief' });
+  });
+});
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -256,17 +256,15 @@ export function DateLabel(props: DateLabelProps) {
  * Automatically refreshes
  */
 function TimeAgo({ date, format }: { date: number | string | Date; format?: DateFormatOptions }) {
-  const [formattedDate, setFormattedDate] = useState<string>(() => timeAgo(date, { format }));
+  const [, setTick] = useState(0);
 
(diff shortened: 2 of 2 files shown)


skoeva (changes requested):
hi, thanks for submitting! would you mind editing the title to match our [commit guidelines](https://headlamp.dev/docs/latest/contributing/#2-follow-commit-guidelines)?

Utkarshpandey0001:
@illume Hi, open reviews are solved and description of pr has changed as suggested by copilot. pls go through it once. Thanks

illume (changes requested):
Thanks for these changes.


Utkarshpandey0001:
@illume , copilots reviews are resolved.

Utkarshpandey0001:
@illume Hi, open comments are resolved and also CI tests passes succcessfully. Could you have look pls

illume (changes requested):
Thanks for working on this.


illume (approved):
🎉 thanks!

PR #5408, opened 2026-05-06
docs: e2e-tests: ci: Add new OIDC e2e tests with Dex + Oauth tutorial

illume:
## Summary

Replaces the original "Headlamp talks OIDC to Dex directly" tutorial with the
recommended pattern: Headlamp sits behind
[OAuth2-Proxy](https://oauth2-proxy.github.io/oauth2-proxy/configuration/integrations/headlamp),
which handles the OIDC flow with Dex and forwards the user's `id_token` to
Headlamp as an `Authorization: Bearer` header.

The tutorial is made reproducible and testable:

- **Runnable scripts** in `test-scripts/` bring up the full stack locally with
  one command.
- **A smoke test** (`test.sh`) verifies the redirect chain and Dex discovery
  document without a browser.
- **An opt-in Playwright smoke test** drives the complete OIDC flow in a real
  browser and asserts 14 security-relevant behaviors against the pinned chart.
- **A CI workflow** tests the tutorial stack when its docs, scripts, or spec
  change. It does not build the current checkout's Headlamp source.

## Changes

**`docs/installation/in-cluster/dex/index.md`**
New tutorial: architecture overview, Dex config, Minikube setup, RBAC,
Headlamp install, OAuth2-Proxy install, login walkthrough, production checklist,
troubleshooting. Old "direct OIDC" content and screenshots removed.

**`docs/installation/in-cluster/dex/test-scripts/`**

| File | Purpose |
|---|---|
| `dex-config.yaml` | Dex with two static users (`admin@`, `alice@`, password `password`) |
| `run.sh` | Starts or reuses its owned Dex/Minikube stack, installs pinned charts, and port-forwards on `:8080` |
| `test.sh` | Smoke test: unauthenticated redirect chain + Dex discovery + pod readiness |
| `cleanup.sh` | Tears down the port-forward, Dex, Helm releases, and Minikube profile owned by these scripts |
| `headlamp-values.yaml`, `oauth2-proxy-values.yaml.tpl` | Helm values used by the local stack |
| `README.md` | Prerequisites with per-OS install steps (Ubuntu/Debian and macOS/Homebrew), usage |

Note: Minikube starts without `apiserver.oidc-*` flags (Dex runs over plain HTTP
and kube-apiserver rejects `http://` issuer URLs). K8s API access in this demo
runs under Headlamp's in-cluster ServiceAccount. For per-user RBAC, follow the
manual HTTPS setup in `index.md`.

**`e2e-tests/tests/dexOauth2Proxy.spec.ts`**

Opt-in tutorial-stack smoke test (skipped unless
`HEADLAMP_TEST_DEX_OAUTH2_PROXY=1`). Covers:

- Unauthenticated visit and deep-link both hit the OAuth2-Proxy sign-in splash
- `/ping` liveness endpoint is reachable without auth
- Invalid Dex credentials stay on Dex
- Full sign-in provides successful Kubernetes API access through Headlamp
- `/oauth2/userinfo` returns the signed-in user's email
- Session cookie survives a page reload (no second Dex round-trip)
- Sign-out invalidates the session
- Deep-link redirect preserved after sign-in
- Session cookie has `HttpOnly` and `SameSite=Lax`
- Second browser context is gated (no session leakage)
- Forged `Authorization` header without a session cookie is rejected
- Second Dex user (`alice@`) gets their own identity through the round-trip
- Forged OAuth2-Proxy callback state is rejected (CSRF guard)

When `HEADLAMP_TEST_DEX_OAUTH2_PROXY_MANAGE=1` is also set, the spec invokes
`run.sh` in `beforeAll` and `cleanup.sh` in `afterAll`.

**`.github/workflows/dex-oauth2-proxy-e2e.yml`**

Runs the spec on PRs/pushes touching the tutorial or spec. It smoke-tests the
pinned published Headlamp chart rather than building the current checkout.
The workflow builds Dex from source (pinned to `v2.45.1`), starts and cleans up
the stack around Playwright, collects stack logs before cleanup on failure,
adds `host.minikube.internal` to `/etc/hosts` so the browser can reach host-side
Dex, and always uploads the Playwright report.

## Steps to Test

1. Install prerequisites (see `test-scripts/README.md` for per-OS steps).
2. `cd docs/installation/in-cluster/dex/test-scripts && ./run.sh`
3. `./test.sh` -- all checks should pass.
4. Open http://localhost:8080 and sign in as `admin@example.com` / `password`.
5. `./run.sh` again to verify idempotency.
6. `./cleanup.sh`

**Playwright e2e (with stack already running):**

```bash
cd e2e-tests
HEADLAMP_TEST_DEX_OAUTH2_PROXY=1 npx playwright test tests/dexOauth2Proxy.spec.ts
```

**Self-managed (spec brings up and tears down the stack):**

```bash
cd e2e-tests
HEADLAMP_TEST_DEX_OAUTH2_PROXY=1 HEADLAMP_TEST_DEX_OAUTH2_PROXY_MANAGE=1 \
  npx playwright test tests/dexOauth2Proxy.spec.ts
```

On Linux also add `127.0.0.1 host.minikube.internal` to `/etc/hosts` so the
browser can follow redirects to host-side Dex.

## Screenshots

- `docs/installation/in-cluster/dex/images/01-oauth2-proxy-signin.png`
- `docs/installation/in-cluster/dex/images/02-dex-login.png`
- `docs/installation/in-cluster/dex/images/03-headlamp-after-login.png`

Assisted by Copilot.


illume:
Did a little edit of the tutorial and fixed some pilot review comments.

The script and the tutorial drifted a little bit... so I'm trying to make them be a little bit closer to each other.

gambtho:
Ran this locally end to end. The flow works, and `cleanup.sh` was especially solid — it tore everything down cleanly on the first try.

I hit one setup issue and have a couple of questions about what we want the CI job to protect. I left those inline.

e2e-tests/README.md
```diff
@@ -188,6 +188,51 @@ npx playwright test -g "404 page is present"
 npx playwright test -g "404 page is present" --headed
 ```
 
+## OAuth2-Proxy + Dex e2e test (opt-in)
+
+The spec `tests/dexOauth2Proxy.spec.ts` exercises the
+[Headlamp + OAuth2-Proxy + Dex tutorial](../docs/installation/in-cluster/dex/index.md)
```
gambtho, line 194: One setup detail tripped me up here. `host.minikube.internal` worked inside Minikube, but it didn't resolve from my Windows browser when I ran the scripts in WSL. The login redirect then went to a hostname the browser couldn't reach, which looked like the login had simply failed.

CI already handles this by adding a hosts entry. Could we add the same prerequisite to the README and print a reminder in `run.sh`'s final output? When the browser runs outside the environment hosting Dex, the entry needs to be made on the browser side and may need to point to the WSL/VM address rather than `127.0.0.1`.

e2e-tests/tests/dexOauth2Proxy.spec.ts
```diff
@@ -0,0 +1,443 @@
...
+        failOnStatusCode: false,
+      }
+    );
+
+    // OAuth2-Proxy returns 403 ("Permission Denied") for missing /
+    // invalid CSRF state. Accept any 4xx (or 500-level handler error)
+    // so the test isn't pinned to a specific status code, but a 200
+    // or 302 (redirect to the upstream) would indicate a real bypass.
+    expect(response.status(), 'callback with forged state must not succeed').not.toBe(200);
+    expect(response.status()).toBeGreaterThanOrEqual(400);
```
gambtho: Could we require a 4xx rather than accepting any error response? Right now a 500 from a panic or broken callback handler would also pass, which would hide the more useful failure.

```ts
expect(response.status()).toBeGreaterThanOrEqual(400);
expect(response.status()).toBeLessThan(500);
```

That keeps the assertion flexible across OAuth2-Proxy versions without treating a server error as a successful CSRF rejection.

.github/workflows/dex-oauth2-proxy-e2e.yml
```diff
@@ -0,0 +1,139 @@
...
+      - 'e2e-tests/tests/dexOauth2Proxy.spec.ts'
+      - 'e2e-tests/package.json'
+      - 'e2e-tests/package-lock.json'
+      - 'e2e-tests/playwright.config.ts'
+      - '.github/workflows/dex-oauth2-proxy-e2e.yml'
+
+permissions:
+  contents: read
+
+jobs:
```
gambtho, line 37: One question on scope: since this runs `run.sh`, it tests the published Headlamp 0.44.0 image and only triggers for tutorial/e2e changes. So this feels more like a smoke test for the tutorial stack than regression coverage for the current Headlamp code.

That’s still useful — could we describe it that way? Covering Headlamp regressions too would mean building the PR image and triggering on the relevant source changes.

gambtho (approved):

.github/workflows/dex-oauth2-proxy-e2e.yml
illume: Yeah, they are end to end tests against the published images.

I updated the wording. Thanks!

e2e-tests/tests/dexOauth2Proxy.spec.ts
illume: Makes sense, thanks.

e2e-tests/README.md
illume: Good ideas. cheers

illume:
Addressed those review comments.

PR #5410, opened 2026-05-06
backend: Bump go version to 1.25.9

yolossn:
The stdlib contains a couple of CVEs and those are fixed in go version 1.25.9 so updated the go version. Other dep update suggestions from Artifacthub is already fixed in main

image


illume (approved):
🎉 thanks!

PR #5421, opened 2026-05-06
build: bump local golangci-lint installer to v2.11.3 to match CI

gambtho:
## Summary

`npm run backend:lint` is currently broken on local machines that respect `backend/go.mod`'s toolchain pin. This one-line change in `package.json` aligns the local installer with what CI uses.

## The bug

`package.json:30` installs golangci-lint **v1.64** via:

```
GOBIN=\`pwd\`/backend/tools go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64
```

But:

- **CI uses v2.11.3** (`.github/workflows/backend-test.yml`'s `golangci-lint-action` step pins `version: v2.11.3`).
- `backend/.golangci.yml` is in **v2 schema** — line 1 is `version: "2"`. v1.64 cannot parse it.
- v1.64 was built against **Go 1.24** and refuses to lint modules declaring `go 1.25.x` in `go.mod`. `backend/go.mod` declares `go 1.25.9`.

So locally:

```
$ npm run backend:lint
Error: can't load config: the Go language version (go1.24) used to build golangci-lint
is lower than the targeted Go version (1.25.9)
```

CI works because the workflow ignores the npm script and pins its own version.

## The fix

Update `package.json` to install v2.11.3 (matching CI) with the v2 module path:

```diff
-"backend:install:linter": "GOBIN=\`pwd\`/backend/tools go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64",
+"backend:install:linter": "GOBIN=\`pwd\`/backend/tools go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3",
```

v2.11.3 requires Go ≥ 1.25, which matches `backend/go.mod`'s declaration. Go's toolchain auto-switch handles older host installations transparently — verified locally on a host running 1.24.11 where `go install` auto-fetched 1.25.9 to satisfy the requirement.

## Verification

```
$ rm backend/tools/golangci-lint
$ npm run backend:lint
> GOBIN=\`pwd\`/backend/tools go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.3
go: github.com/golangci/golangci-lint/v2@v2.11.3 requires go >= 1.25.0; switching to go1.25.9
0 issues.
```

## Why this is its own PR

Found while preparing #5420 (OIDC characterization tests). That PR depends on local lint working, but the lint fix is genuinely orthogonal to the test additions, so it's its own one-line change here. Reviewers can ack independently.

Issue #5402, opened 2026-05-05
EKS/GKE auth: consolidate cloud-provider exec creds and OIDC-without-API-server path


gambtho:
## Summary

A second cluster of OIDC/SSO issues all stem from the same root: Headlamp doesn't have a clean, well-tested path for managed Kubernetes (EKS, GKE) where authentication is mediated by the cloud provider rather than by Headlamp doing OIDC directly against an IdP that the API server trusts.

There are two sub-paths. They meet at kubeconfig load and proxy request handling, but the fixes should be separate because A is desktop client-credential behavior and B is a security-sensitive server-side impersonation mode. Two stacked PRs (A first, B on top) — bundling would inflate the security review surface for A reviewers without benefit.

**Scope boundary with #5401:** This issue covers *identity → API server* — how an established identity gets accepted by the cluster API, including any session/token-storage changes needed to make impersonation safe. #5401 covers *identity establishment* — the browser-side OIDC round trip. Post-login 401/403 issues (#2848, #4198) belong here, on sub-path B.

## Sub-path A — Desktop exec credentials

User's kubeconfig uses an `exec` credential plugin (`aws`, `aws-iam-authenticator`, `gke-gcloud-auth-plugin`, `kubelogin`) and auth happens out of band. Headlamp either fails to invoke it, doesn't pick up refreshed creds, or surfaces a generic TLS/proxy error. **Likely root cause is `SetupProxy`'s silent fallback to Go's default transport when `RESTConfig()` or `makeTransportFor()` fails (`backend/pkg/kubeconfig/kubeconfig.go:423-435`)** — needs reproduction to confirm per issue.

- #1716 — Cannot connect to AWS EKS with SSO (`x509: certificate signed by unknown authority`, "failed to get context")
- #358 — [Desktop] Can't login when using `aws sso` authentication scheme
- #1338 — GKE + OIDC integration (exec-plugin half — `gke-gcloud-auth-plugin`)
- #1829 — GKE + Keycloak base64 decode error (exec-plugin half)

**A explicitly does not promote Headlamp's Windows-only custom exec authenticator (`backend/pkg/exec/exec.go`) to all platforms by default.** First fail-closed and surface diagnostics while preserving stock client-go on non-Windows; only expand if parity tests prove stock client-go is insufficient. Lifting the fork would change behavior for every Linux/macOS desktop user.

## Sub-path B — In-cluster OIDC decoupled from API server

[...]

Issue #5401, opened 2026-05-05
OIDC login flow: consolidate redirect, callback, and autologin UX bugs


gambtho:
- [ ] split up into tasks/sub issues

---

## Summary

Several open issues describe overlapping bugs and UX gaps in the OIDC browser login flow. They share the same code path (frontend OIDC flow + backend `/oidc` and `/oidc-callback`) and could reasonably be addressed together. Filing this as a tracking/consolidation issue so the work can be scoped as one effort.

**Scope boundary with #5402:** This issue covers *identity establishment* — the browser-side round trip and the ephemeral state that makes it work. #5402 covers *identity → API server* — how the established identity gets accepted by the cluster API. Post-login 401/403 issues (e.g. #2848) belong on #5402 sub-path B, not here.

## In scope

- #4877 — OIDC auth redirect does not preserve original deep-linked URL *(verify + regression coverage; current `authchooser/index.tsx:154,160,164` already does `history.replace(from)` on success — residual likely the page-reload / full-page autologin variant where `location.state.from` is lost)*
- #2134 — Successful sign-in always redirects to home rather than original URL *(same path as #4877; verify the residual)*
- #2126 — 404 not found on callback URI **(desktop callback handoff only — comments on #2126 also reference kubeconfig OIDC token handling that may belong on #5402 or a separate kubeconfig-auth bug)**
- #4019 — OIDC login fails "Invalid Request" in fullscreen until refresh (multi-replica state loss)
- #4721 — UI does not redirect to the Login page (`testAuth` false positive)
- #4343 — Enable autologin on OIDC *(via PR #4475 if it lands; otherwise fold scope in)*
- #3414 — Default authentication method / OIDC autologin (autologin half; "disable token auth" half is separate)

Common theme: the post-auth redirect target is lost or wrong, the callback is fragile under multi-replica or desktop entry conditions, and there's no first-class "start at login / autologin" behavior.

## Proposed approach

Two PRs, sequenced after #4475 lands — **or fold #4475's scope in** if it remains stale (currently merge-conflicted, failing checks).

[...]

Issue #5025, opened 2026-04-01
OIDC login with Microsoft Entra fails with `failed to append ca cert to pool` and `failed to verify id token signature`


jnabizada:
## Description

When running Headlamp in-cluster behind Envoy Gateway and configuring OIDC with Microsoft Entra ID, login reaches the provider correctly, the callback returns to Headlamp, and the token exchange succeeds once the client secret is corrected, but Headlamp then fails to verify the ID token signature.

The repeating backend errors are:

```text
failed to append ca cert to pool
failed to verify id token signature
```

This appears to happen even though:
- OIDC discovery is reachable from the pod
- JWKS is reachable from the pod
- no custom CA env vars or CA volume mounts are present in the deployment
- OIDC config is otherwise minimal and valid

## Environment

- Headlamp image: `ghcr.io/headlamp-k8s/headlamp:v0.39.0`
- In-cluster deployment behind Envoy Gateway
- Base URL: `/dashboard`
- OIDC provider: Microsoft Entra ID
- Kubernetes route exposed at:
  `https://....com/dashboard`

## Headlamp configuration

Deployment args:

```yaml
args:
- -in-cluster
- -base-url=/dashboard
```

Runtime env inside the pod:

```text
HEADLAMP_CONFIG_OIDC_CLIENT_ID=<app-client-id>
HEADLAMP_CONFIG_OIDC_CLIENT_SECRET=<actual-secret-value>
HEADLAMP_CONFIG_OIDC_IDP_ISSUER_URL=https://login.microsoftonline.com/<tenant-id>/v2.0
HEADLAMP_CONFIG_OIDC_SCOPES=openid,email,profile
HEADLAMP_CONFIG_OIDC_USE_ACCESS_TOKEN=false
```

There are no extra validator env vars anymore, and no CA/TLS-related env vars are present in the container.

## Entra app setup

- Redirect URI:
  `https://....com/dashboard/oidc-callback`
- Response type sent by Headlamp:
  `code`
- Scopes:
  `openid email profile`
- API permissions granted

Observed authorize request:

```text
https://login.microsoftonline.com/<tenant-id>/oauth2/v2.0/authorize?client_id=<app-client-id>&redirect_uri=https%3A%2F%2F....com%2Fdashboard%2Foidc-callback&response_type=code&scope=openid+email+profile&state=...
```

## What works

From inside the Headlamp pod, OIDC discovery works:

```bash
[...]

PR #5427, opened 2026-05-06
backend, frontend: oidc: portable state, return-URL preservation, testAuth fix

gambtho:
## Summary

This PR is the implementation half of issue #5401 (web side; desktop is PR 2). It makes OIDC login portable across replicas, preserves the deep-link the user came from, and fixes the SSRR-based auth probe that returns false-positives when `system:basic-user` is granted to `system:unauthenticated`.

**Stacked on #5420** (characterization tests). Until #5420 merges, the diff against `main` here contains both stage 0 and stages 1-7. After #5420 merges, this branch will be rebased and the diff will narrow to just stages 1-7.

Eight commits, one per stage plus a harness-move commit, each independently building/linting/passing tests:

1. `backend: auth: add ValidateReturnTo helper` — pure helper, no callers.
2. `backend: auth: add signed state encode/decode helpers` — pure helper, no callers.
3. `backend: oidc: replace per-process state map with signed state` — wires StateSigner into /oidc and /oidc-callback; adds `--oidc-state-signing-key-file`. Multi-replica characterization test flips from asserting the bug to asserting the fix.
4. `backend, frontend: oidc: returnTo end-to-end with mode=popup|fullPage` — `returnTo` validated at issue time, embedded in signed state, re-validated on callback (defense in depth). Frontend AuthChooser builds the URL via a pure helper; OIDCAuth schedules a defense-in-depth fallback navigation.
5. `backend: oidc: HTML error pages on /oidc-callback` — content-negotiated; non-HTML callers keep the existing plain-text contract.
6. `frontend: clusterApi: testAuth cookie verification for OIDC clusters` — calls `/clusters/{cluster}/me` for OIDC clusters; SSRR unchanged elsewhere.
7. `tools: oidc-repro: add developer / E2E harness` — kind + Dex + two-replica nginx round-robin harness used to characterize the four #5401 in-scope failure modes. Notes/findings intentionally not shipped; users create them via the per-issue scripts.
8. `e2e-tests: oidc: scenarios on the tools/oidc-repro harness` — Playwright specs covering deep-link → IdP, deep-link with reload, fullPage entry, stale state HTML error page, and an optional multi-replica scenario gated behind an env var. Whole describe block is gated behind `HEADLAMP_OIDC_E2E=1` so it's safe to ship.

## Multi-replica caveat

Without `--oidc-state-signing-key-file`, Headlamp generates a per-process HMAC key at startup. This works for single-replica deployments. For multi-replica rollouts, operators must point every replica at the same key file (the default key is logged as a single INFO line so operators have a pointer at the flag). Helm chart wiring for the flag is intentionally **out of scope** here — separate follow-up.

## Out of scope (deferred)

- Desktop OIDC code-handoff (PR 2 of #5401)
- In-cluster OIDC impersonation (#5402 sub-path B)
- Desktop exec credentials (#5402 sub-path A)
- Helm chart values for `--oidc-state-signing-key-file` (separate follow-up)
- OIDC nonce hardening (separate task)
- Provider-specific bug fixes (#4876, #5025, #3884, etc.)
- CI integration of the `tools/oidc-repro/` harness — `HEADLAMP_OIDC_E2E` keeps the Playwright spec opt-in for now; wiring `up.sh` into a workflow is a follow-up

## References

- Issue: #5401
- Per-issue evidence: #4019 (multi-replica state loss), #4877 / #2134 (returnTo regression), #4721 (SSRR false-positive)
- Stage 0: #5420 (characterization tests, awaiting review)
- Lint version fix: #5421 (golangci-lint v1.64 → v2.11.3, prerequisite for local lint)
- Related in-flight work: #4230 (RefreshAndSetToken extraction), #3482 (auth package extraction tracking)
- Maintainer feedback that drove the staged restructure: illume's comment on #5401 (#issuecomment-4385554361)

## Test plan

- [x] Backend: `go test ./...` green (auth helpers, OIDC handlers, characterization tests including new HTML error / mode / multi-replica subtests)
- [x] Frontend: `tsc --noEmit` clean; vitest specs for `buildOauthUrl`, `OIDCAuth`, and `testAuth` all pass
- [x] Each commit builds independently (verified by running tests after each)
- [ ] CI green on this draft
- [ ] Manual run of `tools/oidc-repro/scripts/up.sh` + Playwright `oidc.spec.ts` once a reviewer wants to exercise the E2E (locally)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -866,6 +866,45 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	r.HandleFunc("/oidc", func(w http.ResponseWriter, r *http.Request) {
 		cluster := r.URL.Query().Get("cluster")
 
+		// mode controls how /oidc-callback redirects after a successful
+		// token exchange. "popup" preserves the legacy /auth?cluster=...
+		// contract; "fullPage" redirects directly to returnTo (or the
+		// cluster root if returnTo is absent). "desktop" is reserved for
+		// PR 2 of #5401 and is rejected here so its eventual semantics
+		// can't be silently established by abuse.
+		mode := r.URL.Query().Get("mode")
+		switch mode {
+		case "":
+			mode = "popup"
+		case "popup", "fullPage":
+			// ok
+		case "desktop":
+			http.Error(w,
+				"mode=desktop is reserved for the desktop OIDC code-handoff flow (PR 2 of #5401) "+
+					"and is not yet supported",
+				http.StatusBadRequest)
+
+			return
+		default:
+			http.Error(w, "invalid mode", http.StatusBadRequest)
+			return
+		}
+
+		var validatedReturnTo string
+
+		if rawReturnTo := r.URL.Query().Get("returnTo"); rawReturnTo != "" {
+			cleaned, vErr := auth.ValidateReturnTo(rawReturnTo, "")
+			if vErr != nil {
+				logger.Log(logger.LevelError, map[string]string{"cluster": cluster},
+					vErr, "rejected returnTo on /oidc")
+				http.Error(w, "invalid returnTo", http.StatusBadRequest)
+
+				return
+			}
+
+			validatedReturnTo = cleaned
+		}
+
 		flow, err := buildOIDCFlow(r.Context(), config, r, cluster)
 		if err != nil {
 			handleOIDCFlowError(w, r, cluster, err)
@@ -883,7 +922,8 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		payload := auth.StatePayload{
 			V:         auth.StateSchemaVersion,
 			Cluster:   cluster,
-			Mode:      "popup",
+			Mode:      mode,
+			ReturnTo:  validatedReturnTo,
 			ExpUnixMs: stateSigner.Now().Add(stateSigner.TTL()).UnixMilli(),
 			CSRF:      hex.EncodeToString(csrfBytes),
 		}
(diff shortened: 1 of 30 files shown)


gambtho:
Closing this in favour of a set of smaller, independently reviewable PRs.

This grew to 29 files / +3,636, and the only open blocker it addressed (#4721) was buried in the middle of it. Splitting along the existing stage boundaries lets the blocker land on its own and gives the state-map redesign the review attention it actually needs.

Open now:

- #6985 — `testAuth` cookie verification for OIDC clusters. This is the #4721 fix. 2 files, no backend changes; it relies only on `/clusters/{cluster}/me`, which already exists on `main`.
- #6987 — the `tools/oidc-repro/` kind+Dex harness. New directory, no production code, not touched by CI.

Still to come, as separate PRs:

- signed OIDC state (stages 1–3 here). Deliberately held back: `backend/cmd/headlamp.go` has seen ~72 commits since this branch was cut, at least five of them reworking the exact per-process state map these stages delete (`evictExpiredOidcStates`, `OidcStateTTL`, the injectable `oidcStateReader`). Whether signed state replaces that or layers on top of it is a design conversation with the folks who landed that work, not a rebase. It is also still stacked on #5420.
- `returnTo` end-to-end and the HTML `/oidc-callback` error pages (stages 4–5). These stack on the signed-state PR — stage 4 calls `auth.ValidateReturnTo` and `auth.StatePayload` from stages 1–2, and stages 3–5 all edit overlapping regions of `headlamp.go`.
- the Playwright specs, which will go with the PR that implements the behaviour they assert.

No work is lost — the branch stays available and every stage carries over as its own commit. Refs #5401.

PR #5428, opened 2026-05-06
Fix/cluster redirect OIDC

### LINKED ISSUE #5409: Bug: cluster switching redirects to incorrect cluster
## Describe the bug
I have two clusters A and B. Access to both of them is behind authentication. After logging in to cluster A when i try to switch to a cluster B, Headlamp attemps to authorize automatically. Since i am not logged into cluster B yet, it fails. After that i am redirected to authentication page to log in. However after successfully logging in (via OIDC), the application redirects me back to the cluster and fails to complete the redirection, so i end up on cluster A again. Switching again to cluster B at this point works without issues.

## To Reproduce
Steps to reproduce the bug:
1. Click on cluster A in home view
2. Log in to cluster A
3. Switch to cluster B via clusterName button next to the settings
4. Log in to cluster B

 - Installation type: Helm
 - Headlamp Version: 0.41.0

No

## Additional Context
The culprit is likely the cached state. From what i understand looking at [RouteSwitcher](https://github.com/kubernetes-sigs/headlamp/blob/55a50436655649a85cf1ed68eddd23db2d140aac/frontend/src/components/App/RouteSwitcher.tsx) and [index](https://github.com/kubernetes-sigs/headlamp/blob/55a50436655649a85cf1ed68eddd23db2d140aac/frontend/src/components/authchooser/index.tsx) it seems like invalid cache is never actually cleared after the successful login so it cannot redirect me to cluster B.

kunaldevsahu:
## Summary

This PR fixes an issue where, after OIDC authentication, Headlamp redirects users back to the previously active cluster instead of the intended one.

The fix preserves the intended cluster using sessionStorage and restores it in RouteSwitcher after login.

## Related Issue

Fixes #5409

## Changes

- Added logic in RouteSwitcher to restore the intended cluster after authentication
- Used sessionStorage to temporarily store the selected cluster across redirects
- Ensured correct URL update after login to avoid fallback to previous cluster

## Steps to Test

1. Start Headlamp locally
2. Navigate to a cluster (e.g., cluster A)
3. Switch to another cluster (e.g., cluster B) that requires authentication
4. Complete the login process
5. Verify that the application redirects to cluster B instead of cluster A

N/A

## Notes for the Reviewer

- The fix is isolated to RouteSwitcher and does not affect other routing logic
- No changes were made to authentication flow itself, only post-login redirection handling
- Verified behavior manually using simulated pendingCluster in sessionStorage

### DIFF
--- a/frontend/src/components/App/RouteSwitcher.tsx
+++ b/frontend/src/components/App/RouteSwitcher.tsx
@@ -37,6 +37,25 @@ import { useSidebarItem } from '../Sidebar';
 export default function RouteSwitcher(props: { requiresToken: () => boolean }) {
   // The NotFoundRoute always has to be evaluated in the last place.
   const routes = useTypedSelector(state => state.routes.routes);
+  const cluster = useCluster();
+  const history = useHistory();
+
+  React.useEffect(() => {
+    const pendingCluster = sessionStorage.getItem('pendingCluster');
+
+    if (!history?.location?.pathname) return;
+
+    if (!history.location.pathname.startsWith('/c/')) return;
+
+    if (pendingCluster && pendingCluster !== cluster) {
+      sessionStorage.removeItem('pendingCluster');
+
+      const newPath = history.location.pathname.replace(/^\/c\/[^/]+/, `/c/${pendingCluster}`);
+
+      history.replace(newPath);
+    }
+  }, [cluster, history]);
+
   const routeFilters = useTypedSelector(state => state.routes.routeFilters);
   const defaultRoutes = Object.values(getDefaultRoutes()).concat(NotFoundRoute);
   const clusters = useClustersConf();


beep-boopp:
Hey @kunaldevsahu , thanks for looking into this — OIDC redirect bugs across clusters are a pain to debug.
One thing I noticed going through the diff: the effect reads pendingCluster from sessionStorage, but I couldn't find anything in this PR or the existing codebase that writes that key. So in an actual cluster-switching flow, sessionStorage.getItem('pendingCluster') would always return null and the effect would never fire.

I think the root cause is actually further upstream in authchooser/index.tsx:

The from state (line 64, derived from location.state) seems to go stale or default to the wrong cluster during the OIDC roundtrip.

Then, history.replace(from) on lines 154/160/212 sends you back to the old cluster.

That's probably where the fix needs to live rather than a post-redirect correction here.

Also, since RouteSwitcher.test.tsx already exists, it'd be great to add test coverage for whatever new routing logic ends up landing!

illume (changes requested):
Thanks for these changes.


<details>

- `Fix cluster redirect after OIDC login by preserving intended cluster` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for this PR.


skoeva (commented):
closing this - nothing in the repo ever calls `sessionStorage.setItem('pendingCluster',...)`. this was flagged a month ago by copilot and another user and no follow-up. the real cause is the stale `from`/`location.state` used in the OIDC roundtrip (the `history.replace(from)` calls in the oidcauth flow), so the fix should be there

PR #5434, opened 2026-05-07
downgrade electron

joaquimrocha:
## Summary

There's some font processing related issue with Electron 41 which means electron doesn't load and the app window is left blank.
I have tried several electron versions and settled on the one in this PR (as the highest one that works) + the needed change for gtk.

## Steps to Test

1. `npm run app:start` and verify Headlamp actually loads


illume (approved):
🎉 thanks!

(tried it on Mac and Ubuntu 22.04 via WSL)

PR #5471, opened 2026-05-08
Use `sh` as the default `Node` debugger `command`

dabico:
## Summary

By default, the `debugger` container of `Node` debuggers created by Headlamp don't have the `command` set in the `Pod` spec. As a result, the following issues can manifest under specific circumstances[^1]:

1. Default `CMD`/`ENTRYPOINT` of certain images exit immediately. As a result, Headlamp will fail to attach a terminal session, as the container exits almost instantly.

2. While the container `command` can be omitted for images that define a `CMD`/`ENTRYPOINT`, using custom images that define neither would result in the `Pod` failing to start:

	```
	Error: failed to generate container "..." spec: failed to generate spec: no command specified
	```

In both cases, the Headlamp UI terminal only displays:

```
Trying to open a shell
```

This same behaviour is not exhibited by Headlamp's pod debug feature, which spawns ephemeral containers and explicitly defaults their `command` to `sh`:

https://github.com/kubernetes-sigs/headlamp/blob/0b4eeacc97ff0293e8c920ecda28df44fdd39d9d/frontend/src/components/pod/PodDebugTerminal.tsx#L104

This PR applies the same default to the `Node` debugger Pod, aligning the two features behavior-wise.

## Changes

- Updated `frontend/src/components/node/NodeShellTerminal.tsx`

## Steps to Test

1. Launch locally with Minikube:

    ```sh
    $ minikube start \
      --container-runtime=containerd \
      --docker-opt containerd=/var/run/containerd/containerd.sock \
      --kubernetes-version=v1.35.0
    $ tmux new-session -d -s frontend 'make frontend-install run-frontend'
    $ tmux new-session -d -s backend 'make backend run-backend'
    ```

2. Open Headlamp UI in the browser.
3. Set the <kbd>Linux Image</kbd> to something like `cgr.dev/chainguard/busybox:latest-glibc` under <kbd>Settings</kbd> › <kbd>Cluster Settings</kbd> › <kbd>Node Shell Settings</kbd>.
4. Switch to <kbd>Cluster</kbd> › <kbd>Nodes</kbd> and select `minikube`.
5. Trigger the <kbd>Debug Node</kbd> action in the node details, the terminal should open and accept prompts.

[^1]:
    As to why these went unnoticed, it probably has to do with the default image (`busybox:latest`) having `CMD`:

    ```shell
    $ docker inspect busybox:latest | jq '.[].Config.Cmd.[]'
    "sh"
    ```

### DIFF
--- a/frontend/src/components/node/NodeShellTerminal.tsx
+++ b/frontend/src/components/node/NodeShellTerminal.tsx
@@ -35,7 +35,13 @@ interface NodeShellTerminalProps {
   onClose?: () => void;
 }
 
-const shellPod = (name: string, namespace: string, nodeName: string, nodeShellImage: string) => {
+const shellPod = (
+  name: string,
+  namespace: string,
+  nodeName: string,
+  nodeShellImage: string,
+  command: string[] = ['sh']
+) => {
   return {
     kind: 'Pod',
     apiVersion: 'v1',
@@ -59,6 +65,7 @@ const shellPod = (name: string, namespace: string, nodeName: string, nodeShellIm
         {
           name: 'debugger',
           image: nodeShellImage,
+          command,
           terminationMessagePolicy: 'File',
           tty: true,
           stdin: true,


illume:
Thanks for this.

I wonder if we have to first get the Cmd out of the pod spec? If they don't have it, then set a default?

dabico:
> Thanks for this.

Thank you for the review! Regarding your question:
 
> I wonder if we have to first get the Cmd out of the pod spec? If they don't have it, then set a default?

I think you mean the image's `CMD`/`ENTRYPOINT` rather than the `Pod` spec (constructed by Headlamp). Kubernetes itself doesn't expose the `CMD`/`ENTRYPOINT` from images until the container is created. Inspecting it would mean fetching the OCI image details directly from the source registry. Implementing this would involve solving non-trivial problems like authenticating with arbitrary registries.

For the time being, I'd suggest keeping this PR scoped to bringing the `Node` debugger in line with Headlamp's existing `Pod` debugger, which already hardcodes `['sh']`. A smart, image-aware fallback would be a worthwhile follow-up, but the effort and scope are considerably larger than this small fix warrants.

illume (changes requested):
Thanks for these changes.


<details>

</details>


<details>

- `Use `sh` as default node shell command` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

illume:
@dabico thanks for the explanation.

illume (approved):
🎉

dabico:
> @dabico thanks for the explanation.

Thank you @illume! It was my pleasure to contribute. Looking forward to doing more of that! 😊

PR #5498, opened 2026-05-09
frontend: resourceMap: Prefer custom node details page

mahmoudmagdy1-1:
## Summary
This PR fixes resource map node details rendering so a node-provided `detailsComponent` replaces the default kube object details instead of rendering both.
This affects plugin-provided map nodes that include both `detailsComponent` and `kubeObject`. Without this change, Headlamp can render duplicate fallback details for built-in resources, unnecessary fallback content for plugin resources, and incorrect fallback details when a plugin resource shares a kind with a built-in Kubernetes resource.
## Changes
- Updated `GraphNodeDetails` to render `detailsComponent` as the full node details view when provided.
- Kept the existing `KubeObjectDetails` fallback for nodes without a custom details component.
- Prevented plugin map nodes from triggering unnecessary requests or incorrect default kube object details after their custom details render.
## Steps to Test
1. Open the resource map with a plugin source that provides custom node details.
2. Click a plugin map node that has both `detailsComponent` and `kubeObject`.
3. Verify only the custom detail view renders in the side panel.
4. Verify no extra fallback kube object detail request or warning is triggered for that node.
Before (Console Logs after clicking on Job to get job detail view):
image
After (Console Logs after clicking on Job to get job detail view):
image
Also another thing is that it loads the other view in the end of the actual working job details view which doesn't work
Before:
image
After:
image

### DIFF
--- a/frontend/src/components/resourceMap/details/GraphNodeDetails.tsx
+++ b/frontend/src/components/resourceMap/details/GraphNodeDetails.tsx
@@ -40,9 +40,12 @@ function NodeDetailsRenderer({ node }: { node?: GraphNode }) {
   const hasContent = node && (node.detailsComponent || node.kubeObject);
   if (!hasContent) return null;
 
+  if (node.detailsComponent) {
+    return <node.detailsComponent node={node} />;
+  }
+
   return (
     <>
-      {node.detailsComponent && <node.detailsComponent node={node} />}
       {node.kubeObject && (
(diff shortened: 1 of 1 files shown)


mahmoudmagdy1-1:
The CI Failure is unrelated to this PR I think

illume (approved):
🎉 thanks!

PR #5509, opened 2026-05-09
fix: Return error from GetDefaultKubeConfigPath instead of calling os.Exit

### LINKED ISSUE #5493: GetDefaultKubeConfigPath calls os.Exit(1) instead of returning an error
## Describe the bug

While exploring the backend codebase to understand how Headlamp resolves the default kubeconfig path, I noticed that `GetDefaultKubeConfigPath()` in `backend/pkg/config/config.go` calls `os.Exit(1)` directly when it fails to get the current OS user.

This caught my attention because the function lives in `pkg/` which in Go convention means it is library/utility code meant to be reused by callers. A utility function should never make the decision to kill the process. That decision belongs to the caller.

This creates three real problems:

**1. The process crashes with no recovery**
If `user.Current()` fails which can happen in containers or restricted environments the entire Headlamp process dies instantly. The caller has no chance to show a friendly error, fall back to a default path, or handle it in any way.

**2. It makes the function untestable**
`GetDefaultKubeConfigPath` is called in test files. If `user.Current()` ever fails in a CI environment, `os.Exit(1)`
would kill the entire Go test runner not just the failing test. You cannot write a unit test for the failure case at all.

**3. Deferred cleanups are skipped**
`os.Exit` is immediate it does not run any `defer` statements. Any open resources or connections that rely on defers to clean up would be left in a broken state.

## Steps to Reproduce

This can be triggered in containers or restricted CI environments where the current OS user cannot be resolved.

## Expected Behavior

The function should return `(string, error)` and let the caller decide how to handle the failure.

## Actual Behavior

The process exits immediately with code 1.

## Proposed Fix

```go
func GetDefaultKubeConfigPath() (string, error) {
    user, err := user.Current()
    if err != nil {
        return "", fmt.Errorf("getting current user: %w", err)
    }
    return filepath.Join(user.HomeDir, ".kube", "config"), nil
}
```

The callers in `config.go` and the test files would also be updated to handle the returned error appropriately.

## Environment

## Additional context

I'd like to work on a fix for this if the maintainers agree.

YadavAkhileshh:
## Summary
`GetDefaultKubeConfigPath()` in `backend/pkg/config/config.go` was calling `os.Exit(1)` directly when it failed to resolve the current OS user. This is a Go anti-pattern for library code utility functions should return errors and let the caller decide how to handle failure.

This PR refactors the function to return `(string, error)` and updates all callers accordingly.

## Related Issue
Fixes #5493

## Changes
- Changed `GetDefaultKubeConfigPath()` signature from `string` to `(string, error)`.
- Replaced `os.Exit(1)` with a proper error return.
- Updated all callers in `config.go` and test files to handle the returned error.

## Testing
1. Run backend tests: `cd backend && go test ./pkg/config/... ./pkg/kubeconfig/... ./cmd/...`
2. Start Headlamp normally and verify kubeconfig loads correctly.
3. Confirm no regression in cluster connection behavior.

### DIFF
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -108,19 +108,6 @@ func getResponseFromRestrictedEndpoint(handler http.Handler, method, url string,
 	return rr, nil
 }
 
-// getDefaultKubeConfigPathForTest is a test helper that wraps config.GetDefaultKubeConfigPath
-// and fails the test immediately if retrieving the default path returns an error.
-func getDefaultKubeConfigPathForTest(t *testing.T) string {
-	t.Helper()
-
-	path, err := config.GetDefaultKubeConfigPath()
-	if err != nil {
-		t.Fatalf("failed to get default kubeconfig path: %v", err)
-	}
-
-	return path
-}
-
 func TestGetConfigIncludesDefaultPodDebugImage(t *testing.T) {
 	c := &HeadlampConfig{
 		HeadlampConfig: &headlampconfig.HeadlampConfig{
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for working on this.


<details>

</details>

illume (changes requested):
Thanks for the contribution.


illume (changes requested):
Thanks for the contribution.


illume (changes requested):
Thanks for the contribution.


<details>

- `fix: Return error from GetDefaultKubeConfigPath instead of calling os.Exit` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for working on this.


illume (changes requested):
Thanks for these changes.


illume (approved):
🎉 thanks!

PR #5528, opened 2026-05-10
backend: externalproxy: Add timeout to HTTP client

ayushmaan-16:
### Description:
The `/externalproxy` handler was making requests using an `http.Client` that had no timeout set. This is a bit risky because if an external website (like ArtifactHub) hangs or responds very slowly, Headlamp would wait forever. This could eventually lead to "goroutine leaks" and might even crash the server if too many requests get stuck.

This PR adds a 30-second timeout to that client to ensure that we always release system resources if a request takes too long. This follows the same safe pattern that is already used in other parts of the codebase (like in `serviceproxy/http.go`).

### Changes Made:
Updated the `http.Client` initialization in `backend/cmd/headlamp.go` to include a `Timeout` of 30 seconds.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -610,7 +610,12 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		w.Header().Set("Pragma", "no-cache")
 		w.Header().Set("X-Accel-Expires", "0")
 
-		client := http.Client{}
+		proxyCtx, cancel := context.WithTimeout(r.Context(), 30 * time.Second)
+		defer cancel()
+
+		proxyReq = proxyReq.WithContext(proxyCtx)
+
+		client := http.Client{Timeout: 30 * time.Second}
 
 		resp, err := client.Do(proxyReq) //nolint:gosec
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ayushmaan-16:
@illume I've fixed the changes addressed by copliot. You can review this further when you have time :)

illume (changes requested):
Thanks for these changes.


ayushmaan-16:
@illume The Copilot comment about adding spaces around the `*` operator was incorrect. I ran `gofmt -w` on the file, which reverted the change back to `30*time.Second` (without spaces). This is the correct format according to the project's gofmt/gofumpt configuration in [.golangci.yml](cci:7://file:///c:/Users/ayush/Downloads/headlamp-1/backend/.golangci.yml:0:0-0:0). Also you can even see that the CI check fails now when I pushed the new code with spaces. How do you suggest to move ahead ? Should I go back to my previous commit ?

illume (changes requested):
Thanks for working on this.


<details>

</details>

illume (changes requested):
Thanks for these changes.


illume (approved):
🎉 thanks!

PR #5538, opened 2026-05-10
frontend: ContainerInfo: Fix key prop placement in ports list

### LINKED ISSUE #5537: bug: React key prop placed on wrong element in ContainerInfo ports list causes reconciliation warning
## Describe the bug

In `frontend/src/components/common/Resource/Resource.tsx`, the `Ports` row inside `ContainerInfo` maps over `container.ports` and wraps each item in a shorthand React Fragment (`<>`). The `key` prop is placed on the inner `<Grid>` child instead of the outermost element, which React silently ignores. This triggers a console warning on any Pod detail page that has ports defined.

The shorthand `<>` cannot accept props — `key` must be on the outermost element returned from `.map()`. Using `<React.Fragment key={...}>` instead fixes it.

Expected outcome: no React key warning in the console when viewing Pod details with ports.

## To Reproduce

1. Open Headlamp and connect to a cluster.
2. Navigate to **Workloads → Pods**.
3. Click on any Pod that has ports defined in its container spec.
4. Open browser DevTools console — observe the warning:
Warning: Each child in a list should have a unique "key" prop.
Check the render method of ContainerInfo.

## Environment

- Installation type: Any (reproducible in local dev with `npm start`)
- Headlamp Version: latest (`main` branch)
- Other: Visible in browser DevTools console on any Pod with ports

Yes — I plan to submit a PR.

## Additional Context

Affected file:

```txt
frontend/src/components/common/Resource/Resource.tsx
```

Approximate location:

```txt
ContainerInfo → containerRows() → Ports
```

This does not appear to be a regression and is likely a long-standing issue.

iashutoshyadav:
## What?

Fix incorrect `key` prop placement in the `ContainerInfo` ports list inside `Resource.tsx`.

---

## Why?

The shorthand React fragment syntax (`<>...</>`) cannot accept props. The `key` prop was applied to the inner `<Grid>` element instead of the outermost element returned from `.map()`, which React ignores during reconciliation.

This triggers the following console warning on Pod detail pages that contain ports:

```txt
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `ContainerInfo`.
```

---

## How?

Replace the shorthand fragment with `<React.Fragment key={...}>` so React can properly track list items during reconciliation.

```tsx
// Before
container.ports?.map(({ containerPort, protocol, name }, index) => (
  <>
    <Grid item xs={12} key={`port_line_${index}`}>
      ...
    </Grid>
  </>
))

// After
container.ports?.map(({ containerPort, protocol, name }, index) => (
  <React.Fragment key={`port_line_${index}`}>
    <Grid item xs={12}>
      ...
    </Grid>
  </React.Fragment>
))
```

---

## Testing

1. Open Headlamp and navigate to **Workloads → Pods**
2. Open any Pod that has ports defined (for example: `argocd-server`)
3. Open the browser DevTools console
4. Verify that the warning below no longer appears:

```txt
Warning: Each child in a list should have a unique "key" prop.
```

---

Fixes #5537

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -1650,8 +1650,8 @@ export function ContainerInfo(props: ContainerInfoProps) {
         value: (
           <Grid container>
             {container.ports?.map(({ containerPort, protocol, name }, index) => (
-              <>
-                <Grid item xs={12} key={`port_line_${index}`}>
+              <React.Fragment key={`port_line_${index}`}>
+                <Grid item xs={12}>
                   <Box display="flex" alignItems={'center'}>
                     <Box px={0.5} minWidth={120}>
                       {name && <ValueLabel>{`${name} `}</ValueLabel>}
@@ -1670,7 +1670,7 @@ export function ContainerInfo(props: ContainerInfoProps) {
                     </Box>
                   </Grid>
                 )}
-              </>
+              </React.Fragment>
             ))}
           </Grid>
         ),


skoeva (commented):
looks like this has already been fixed on main - thanks for the contribution!

PR #5541, opened 2026-05-10
Frontend: fix unstable React reconciliation in ContainerInfo ports list

### LINKED ISSUE #5540: Unstable reconciliation causing stale port rendering or incorrect UI updates
## Describe the issue

While reviewing the `ContainerInfo` rendering flow, I noticed that the React `key` prop for the container ports list is attached to a nested child element instead of the top-level mapped element returned by the iterator.

This causes React to emit reconciliation warnings, but more importantly it can also lead to unstable list reconciliation behavior during container updates.

## Why this matters

Incorrect key placement in mapped React lists can cause:

* stale UI rows being preserved incorrectly
* inconsistent rerender behavior
* incorrect DOM reuse during updates
* flickering or stale port information after container changes
* harder-to-debug rendering inconsistencies in dynamic Pod/container updates

This is especially important in Kubernetes resource UIs where container port lists may change dynamically during rolling updates or live resource refreshes.

## Current behavior

The `key` prop is applied to an inner child component instead of the outermost JSX element returned from the `.map()` callback.

As a result:

* React emits reconciliation warnings
* reconciliation stability is not guaranteed

## Expected behavior

The `key` prop should be attached to the top-level mapped element returned by the iterator so React can correctly track list identity during rerenders.

## Suggested fix

Move the `key` prop to the outermost JSX element returned from the ports list mapping callback in `ContainerInfo`.

## Potential impact

This should:

* remove React reconciliation warnings
* improve rendering stability
* ensure correct list identity tracking during Pod/container updates
* reduce risk of stale UI state preservation

gouravi19:
## Summary

Fixes unstable React reconciliation behavior in the `ContainerInfo` ports list by moving the `key` prop to the outermost mapped element returned from the ports iterator.

## What changed

* Replaced the shorthand fragment (`<>...</>`) in the ports `.map()` callback with an explicit keyed `React.Fragment`
* Moved the `key` prop from the inner `<Grid>` element to the top-level mapped element
* Used a stable composite key based on container port/protocol values
* Removed the now-redundant inner `key` prop

## Why this matters

Previously, the mapped ports list returned an un-keyed top-level fragment, which caused React reconciliation warnings and could potentially lead to unstable list reconciliation behavior during rerenders or container updates.

This change ensures React can correctly track list identity during updates and removes reconciliation warnings without altering rendering behavior or layout.

## Notes

* Kept the diff intentionally minimal and isolated to the affected rendering block
* No UI/layout behavior changes
* Verified linting locally

Fixes #5540
Fixes #5540

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -1649,29 +1649,47 @@ export function ContainerInfo(props: ContainerInfoProps) {
         name: t('Ports'),
         value: (
           <Grid container>
-            {container.ports?.map(({ containerPort, protocol, name }, index) => (
-              <>
-                <Grid item xs={12} key={`port_line_${index}`}>
-                  <Box display="flex" alignItems={'center'}>
-                    <Box px={0.5} minWidth={120}>
-                      {name && <ValueLabel>{`${name} `}</ValueLabel>}
-                      <ValueLabel>{`${protocol}:`}</ValueLabel>
-                      <ValueLabel>{containerPort}</ValueLabel>
-                    </Box>
-                    {!!resource && ['Service', 'Pod'].includes(resource.kind) && (
-                      <PortForward containerPort={containerPort} resource={resource} />
-                    )}
-                  </Box>
-                </Grid>
-                {index < container.ports!.length - 1 && (
+            {container.ports?.map(({ containerPort, protocol, name, hostPort, hostIP }, index) => {
+              const baseKey = `${name ?? 'unnamed'}-${protocol ?? 'TCP'}-${containerPort}-${
+                hostPort ?? 'no-host-port'
+              }-${hostIP ?? 'no-host-ip'}`;
+
+              const duplicateIndex =
+                container.ports
+                  ?.slice(0, index)
+                  .filter(
+                    port =>
+                      `${port.name ?? 'unnamed'}-${port.protocol ?? 'TCP'}-${port.containerPort}-${
+                        port.hostPort ?? 'no-host-port'
+                      }-${port.hostIP ?? 'no-host-ip'}` === baseKey
+                  ).length ?? 0;
+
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

- `frontend: fix unstable React reconciliation in ContainerInfo ports list` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

gouravi19:
Thanks for the review! I updated the key generation to avoid relying on the array index and amended the commit message to better match the repository commit conventions.
@illume

gouravi19:
Thanks for the review! I updated the commit message to better match the repository commit conventions and also adjusted the key generation to avoid relying on the array index.

illume (changes requested):
Thanks for the contribution.


gouravi19:
Thanks for the review! I updated the key generation to avoid relying on unstable reconciliation behavior and resolved the remaining review comments.

gouravi19:
Thanks for the follow-up reviews! I addressed the reconciliation and key stability concerns by moving the key to the top-level mapped element and using a deterministic composite key including additional stable port identity fields. I also updated the commit message to match the repository conventions and resolved the related review comments.

illume (changes requested):
Thanks for the contribution.


gouravi19:
Thanks for the follow-up review! I updated the key generation to keep stable identity fields as the primary key source while also adding a deterministic disambiguator to guarantee uniqueness for duplicate unnamed port entries. The key remains attached to the top-level mapped element, preserving React reconciliation stability without changing rendering behavior or layout.

illume (changes requested):
Thanks for working on this.


gouravi19:
Thanks for the follow-up review! its done

illume (approved):
🎉 thanks!

PR #5560, opened 2026-05-11
frontend: Multiplexer: Fix setInterval leak when WebSocket connection fails

### LINKED ISSUE #5559: frontend: multiplexer: setInterval leaks forever when WebSocket connection fails
## Describe the bug

In `WebSocketManager.connect()` (`frontend/src/lib/k8s/api/v2/multiplexer.ts`), when a connection attempt is already in progress, concurrent callers wait via a `setInterval` polling loop. The interval only clears itself when the socket reaches `OPEN` state. If the connection fails (`socket.onerror`), `this.connecting` is reset to `false` but the interval is never cleared it keeps firing every 100ms forever, and the returned `Promise` never resolves or rejects, causing all waiting callers to hang silently with no error shown to the user.

The code's own comment acknowledges this: *"No timeout - could theoretically run forever if connection never opens"* (line 68).

Expected outcome: when the connection fails, all pending polling intervals should be cleared and their Promises rejected so callers can surface the error or retry.

## To Reproduce

Steps to reproduce the bug:
1. Open Headlamp with a Kubernetes cluster configured.
2. Open browser DevTools → Network tab and block WebSocket connections (or point Headlamp at a cluster where the backend is unreachable).
3. Navigate to any resource list page (e.g. Workloads → Pods) so that multiple components call `WebSocketManager.subscribe()` concurrently.
4. Open DevTools → Performance profiler and record for a few seconds — observe `setInterval` callbacks firing at ~10/s indefinitely with no data loading and no error message displayed.

- Installation type: [e.g. Linux-Tarball / In-Cluster / Helm / ...]
- Headlamp Version: [X.Y.Z]
- Other: Reproducible whenever the WebSocket multiplexer endpoint is unreachable and more than one subscription is initiated simultaneously.

Yes (I will propose a PR).

## Additional Context

Adding a `!this.connecting` check inside the interval is sufficient to detect a failed connection and reject:

```typescript
if (this.connecting) {
  return new Promise((resolve, reject) => {
    const checkConnection = setInterval(() => {
      if (this.socketMultiplexer?.readyState === WebSocket.OPEN) {
        clearInterval(checkConnection);
        resolve(this.socketMultiplexer!);
      } else if (!this.connecting) {
        clearInterval(checkConnection);
        reject(new Error('WebSocket connection failed'));
      }
    }, 100);
  });
}

RajPrakash681:
## Summary

This PR fixes a `setInterval` leak in `WebSocketManager.connect()` where concurrent callers waiting for an in-progress WebSocket connection would spin forever if the connection failed.

## Related Issue

Fixes #5559

## Changes

- Fixed `WebSocketManager.connect()` in `frontend/src/lib/k8s/api/v2/multiplexer.ts` to reject waiting Promises and clear the polling interval when the WebSocket connection fails

## Steps to Test

1. Open Headlamp with a Kubernetes cluster configured
2. Open browser DevTools → Network tab and block WebSocket connections (or point Headlamp at an unreachable cluster)
3. Navigate to any resource list page (e.g. Workloads → Pods) to trigger concurrent `WebSocketManager.subscribe()` calls
4. Open DevTools → Performance profiler and verify no `setInterval` callbacks keep firing after the connection fails ,an error should be propagated instead of a silent hang

N/A

## Notes for the Reviewer

- The fix piggybacks on the existing `this.connecting = false` reset that `socket.onerror` already performs , the interval now checks `!this.connecting` on each tick to detect failure and reject, keeping the change minimal and consistent with the existing pattern

### DIFF
--- a/frontend/src/lib/k8s/api/v2/multiplexer.test.ts
+++ b/frontend/src/lib/k8s/api/v2/multiplexer.test.ts
@@ -412,6 +412,26 @@ describe('WebSocket Multiplexer', () => {
   });
 
   describe('WebSocket error handling', () => {
+    it('should reject concurrent callers when in-progress connection fails', async () => {
+      vi.useFakeTimers();
+
+      // Manually set connecting = true to simulate an in-progress attempt
+      WebSocketManager.connecting = true;
+
+      // Start a concurrent caller — it enters the polling branch
+      const concurrentPromise = WebSocketManager.connect();
+
+      // Simulate the primary connection failing by resetting connecting to false
+      WebSocketManager.connecting = false;
+
+      // Advance timers so the setInterval tick fires and detects the failure
+      await vi.advanceTimersByTimeAsync(200);
+
+      await expect(concurrentPromise).rejects.toThrow('WebSocket connection failed');
+
+      vi.useRealTimers();
+    });
+
     it('should handle polling timeout', async () => {
       const OriginalWebSocket = window.WebSocket;
 
--- a/frontend/src/lib/k8s/api/v2/multiplexer.test.ts
+++ b/frontend/src/lib/k8s/api/v2/multiplexer.test.ts
@@ -415,21 +415,23 @@ describe('WebSocket Multiplexer', () => {
     it('should reject concurrent callers when in-progress connection fails', async () => {
       vi.useFakeTimers();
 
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

PR #5572, opened 2026-05-11
headlamp-plugin: Fix broken plugin tests

skoeva:
This change fixes `make plugins-test` which is broken due to deleted ArtifactHub test packages. The `test-123` ArtifactHub repo (used by plugin e2e/unit tests) had several packages removed (`appcatalog_headlamp_plugin`, `ai_plugin`, `prometheus_headlamp_plugin`), causing 404s and test failures. The shebang `#!/bin/env` also doesn't work on macOS.

### Changes

- Fix shebang `#!/bin/env` -> `#!/usr/bin/env` in `test-headlamp-plugin.js`, `plugin-management.e2e.js`, and `test-pluginctl.js`
- Replace deleted ArtifactHub test packages with live ones from the same `test-123` repo in `plugin-management.e2e.js`, `plugin-management.test.js`, `multi-plugin-management.test.js`, and `test-pluginctl.js`

### Testing

- [x] `make plugins-test` passes locally on macOS
- [x] All 14 tests in `plugin-management.test.js` pass
- [x] All 14 tests in `multi-plugin-management.test.js` pass
- [x] Plugin install/update/uninstall e2w flow completes successfully

PR #5573, opened 2026-05-11
Fix/OIDC external secret scopes arg

### LINKED ISSUE #4220: Using external secret adds empty oidc-scopes arg parameter even if scopes are not provided
## Describe the bug
We switched our oidc secret configuration to the `externalSecret`. We specified only `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET` and the `OIDC_ISSUER_URL`, omitting the `OIDC_SCOPES` variable. However, the `Deployment` template in the helm chart still adds the `-oidc-scopes` arg.

## To Reproduce
Steps to reproduce the bug:
1. Install the helm chart with these values:
```
config:
  oidc:
    secret:
      create: false
    externalSecret:
      enabled: true
      name: oidc
```
2. Make sure to have a k8s secret with the above three mentioned variables set.
3. Logout of your IDP, in our case Entra
4. Try to login to headlamp via `SignIn`
5. Receive this error from the IDP
```
Failed to exchange token: oauth2: "invalid_request" "AADSTS900144: The request body must contain the following parameter: 'code'. Trace ID: xxx Correlation ID: xxx Timestamp: 2025-11-28 15:10:09Z" "https://login.microsoftonline.com/error?code=900144"
```

 - Installation type: In-Cluster via Helm
 - Headlamp Version: 0.38.0
 - Other: Entra applications are configured according to the documentation, without the AKS part:
https://headlamp.dev/docs/latest/installation/in-cluster/azure-entra-id/

Yes. We can add another `if` block to add the parameter only if it is actually set at this location:
https://github.com/kubernetes-sigs/headlamp/blob/47c10d484867069dc53734038059df5eb2a9c090/charts/headlamp/templates/deployment.yaml#L288

iashutoshyadav:
## Summary

This PR fixes the empty `-oidc-scopes` argument being unconditionally added to the Headlamp deployment when using `externalSecret` for OIDC configuration without an `OIDC_SCOPES` key in the secret.

## Related Issue

Fixes #4220

## Changes

- Added `externalSecret.hasScopes` boolean field (default `false`) to `values.yaml` and `values.schema.json`
- Updated `deployment.yaml` to conditionally include:
  ```yaml
  -oidc-scopes=$(OIDC_SCOPES)
  ```
  only when `hasScopes: true`
- Updated expected test template for `oidc-external-secret` to reflect the new default behavior
- Added new test case `oidc-external-secret-with-scopes` and its expected template for the `hasScopes: true` scenario

### 1. Install the Helm chart with an external OIDC secret that does not include `OIDC_SCOPES`

```yaml
config:
  oidc:
    secret:
      create: false
    externalSecret:
      enabled: true
      name: oidc
```

### 2. Verify

- Deployment args do **not** contain:
  ```yaml
  -oidc-scopes=$(OIDC_SCOPES)
  ```
- Login succeeds without an `invalid_request` error from the identity provider (IDP)

### 3. Test opt-in scopes behavior

Set:

```yaml
config:
  oidc:
    externalSecret:
      enabled: true
      name: oidc
      hasScopes: true
```

Verify the deployment args now contain:

```yaml
-oidc-scopes=$(OIDC_SCOPES)
```

## Screenshots

image

## Notes for the Reviewer

- `hasScopes` defaults to `false`, which changes behavior for existing users who already have `OIDC_SCOPES` in their external secret
- Those users will now need to explicitly set:
  ```yaml
  externalSecret.hasScopes: true
  ```
  to restore the previous behavior
- The non-`externalSecret` code path is unaffected

## Validation

Run:

```bash
helm lint charts/headlamp
bash charts/headlamp/tests/test.sh
```

### DIFF
--- a/charts/headlamp/templates/deployment.yaml
+++ b/charts/headlamp/templates/deployment.yaml
@@ -305,7 +305,9 @@ spec:
             - "-oidc-client-id=$(OIDC_CLIENT_ID)"
             - "-oidc-client-secret=$(OIDC_CLIENT_SECRET)"
             - "-oidc-idp-issuer-url=$(OIDC_ISSUER_URL)"
+            {{- if $oidc.externalSecret.hasScopes }}
             - "-oidc-scopes=$(OIDC_SCOPES)"
+            {{- end }}
             {{- if or (ne $oidc.callbackURL "") (ne $callbackURL "") }}
             # Check if callbackURL is non empty either from env or oidc.config
             - "-oidc-callback-url=$(OIDC_CALLBACK_URL)"
--- a/charts/headlamp/tests/expected_templates/oidc-external-secret-with-scopes.yaml
+++ b/charts/headlamp/tests/expected_templates/oidc-external-secret-with-scopes.yaml
@@ -110,6 +110,7 @@ spec:
             - "-oidc-client-id=$(OIDC_CLIENT_ID)"
             - "-oidc-client-secret=$(OIDC_CLIENT_SECRET)"
             - "-oidc-idp-issuer-url=$(OIDC_ISSUER_URL)"
+            - "-oidc-scopes=$(OIDC_SCOPES)"
           ports:
             - name: http
               containerPort: 4466
--- a/charts/headlamp/tests/test_cases/oidc-external-secret-with-scopes.yaml
+++ b/charts/headlamp/tests/test_cases/oidc-external-secret-with-scopes.yaml
@@ -1,0 +1,11 @@
+# This is a test case for OIDC external secret with hasScopes enabled.
+# When hasScopes is true, the -oidc-scopes argument is included so that
+# the OIDC_SCOPES key from the external secret is passed to Headlamp.
+config:
+  oidc:
+    secret:
+      create: false
+    externalSecret:
+      enabled: true
+      name: oidc
+      hasScopes: true
--- a/charts/headlamp/values.schema.json
+++ b/charts/headlamp/values.schema.json
@@ -229,6 +229,10 @@
                 "enabled": {
                   "type": "boolean",
                   "description": "Enable the external secret"
+                },
+                "hasScopes": {
+                  "type": "boolean",
+                  "description": "Set to true if the external secret contains an OIDC_SCOPES key"
                 }
               }
             }
--- a/charts/headlamp/values.yaml
+++ b/charts/headlamp/values.yaml
@@ -104,6 +104,10 @@ config:
     externalSecret:
       enabled: false
       name: ""
+      # -- Set to true if your external secret contains an OIDC_SCOPES key.
+      # When false (default), the -oidc-scopes argument is omitted so that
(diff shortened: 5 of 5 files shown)


iashutoshyadav:
Hi @illume , I checked the failing CI job and it appears unrelated to this PR.

The failure comes from `plugins/pluginctl/src/plugin-management.e2e.js:60`, which hardcodes the following ArtifactHub URL:

```text id="10r7kj"
https://artifacthub.io/packages/headlamp/test-123/prometheus_headlamp_plugin
```

The `test-123` repository on ArtifactHub no longer exists, so this test currently fails independently of the changes in this PR.

I also verified that similar `test plugins` failures are appearing on other recent upstream PRs as well, which suggests this is a pre-existing repository issue rather than a regression introduced here.

The Helm-related changes in this PR were validated locally with:

```bash id="c4bgj2"
helm lint charts/headlamp
bash charts/headlamp/tests/test.sh
```

Thanks!

skoeva (commented):
the CI failure should be fixed by #5572

iashutoshyadav:
Hi @illume,
Could you please review this PR when you get a chance?

charts/headlamp/templates/deployment.yaml
```diff
@@ -305,7 +305,9 @@ spec:
             - "-oidc-client-id=$(OIDC_CLIENT_ID)"
             - "-oidc-client-secret=$(OIDC_CLIENT_SECRET)"
             - "-oidc-idp-issuer-url=$(OIDC_ISSUER_URL)"
+            {{- if $oidc.externalSecret.hasScopes }}
```
illume, line 308: Could the if check OIDC_SCOPES is empty instead of a new "hasScopes"?

illume (changes requested):
Thanks for this.

Left a question for you.

charts/headlamp/templates/deployment.yaml
iashutoshyadav: Hi @illume, thanks for the suggestion!

One option would be to reuse `config.oidc.scopes` as the opt-in signal — if it's non-empty, include `-oidc-scopes=$(OIDC_SCOPES)`.

However, there's a trade-off: in the `externalSecret` path, the runtime value of `$(OIDC_SCOPES)` always comes from the external secret via `envFrom`, not from the Helm value. So `config.oidc.scopes` would only act as a trigger flag — its value would be silently ignored at runtime. That felt a bit confusing, which is why `hasScopes` was introduced as an explicit, purpose-built signal.

That said, if you'd prefer removing `hasScopes` in favor of reusing `config.oidc.scopes` as the trigger, I'm happy to implement that — just let me know which approach you'd like to go with.

illume: Ok, I see now why hasScapes is better. Thanks for the explanation.

illume (approved):
🎉 thanks!

Issue #5183, opened 2026-04-23
Rule of hooks fixes needed (ubmrella issue)


illume:
### Suppressed `react-hooks` lint violations — TODO to fix properly

Each item below is a self-contained description of a suppressed violation. Each could become a standalone sub-issue. They follow the `<area>: <description>` commit-message convention.

#### `react-hooks/rules-of-hooks` violations (58 suppressions across 21 files)
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5184 
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5185
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5186
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5187
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5188
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5189
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5190
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5191
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5192
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5193
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5194
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5195
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5240
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5241
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5242
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5243
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5244
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5245
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5246
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5247
- [x] https://github.com/kubernetes-sigs/headlamp/issues/5248

#### `react-hooks/exhaustive-deps` violations (145 suppressions across 73 files)

- [ ] frontend: App: Fix `react-hooks/exhaustive-deps` — missing dependency in `useEffect`. File: `frontend/src/App.tsx`
[...]

PR #5576, opened 2026-05-12
frontend: CustomResourceList: Document why t is excluded from cols deps

### LINKED ISSUE #5575: frontend: CustomResourceList: Fix `react-hooks/exhaustive-deps` — `cols` useMemo omits the `t` translation function from its deps
Sub-issue of #5183. Fix the `react-hooks/exhaustive-deps` suppression in `frontend/src/components/crd/CustomResourceList.tsx` — the `cols` `useMemo` at line 167 reads `t` from `useTranslation` inside its callback but does not list it in the deps array.

WasThatRudy:
## Summary

This PR documents the existing `react-hooks/exhaustive-deps` suppression in `frontend/src/components/crd/CustomResourceList.tsx` with a comment explaining why `t` is intentionally excluded from the `cols` `useMemo` deps array. Per maintainer feedback, including `t` in this deps array breaks the hook in practice, so the suppression remains in place but is now self-documenting for future contributors.

## Related Issue

Fixes #5575 (sub-issue of umbrella #5183).

## Changes

- Replaced the bare `// eslint-disable-next-line react-hooks/exhaustive-deps` directive with a two-line comment noting that `t` is intentionally omitted and pointing readers to umbrella #5183 for context.
- Reformatted the `useMemo` call to a multi-line shape so the new comments sit directly above the `eslint-disable-next-line` and the deps array (per Copilot review feedback) rather than after `return colsToDisplay;` where they read like dead code.
- No functional change.

### Iteration history

The first revision added `t` to the deps array (the conventional react-i18next fix). Maintainer @illume noted in review that including `t` here breaks the hook in practice, so the change was reverted and replaced with this documentation-only approach. Layout subsequently adjusted per Copilot's inline suggestion so the rationale sits clearly above the suppression rather than after the `return`.

## Steps to Test

2. `npm run lint`: passes (the existing suppression continues to silence the rule for this hook).
3. `npm run tsc`: passes.
4. `npx vitest run src/components/crd`: existing crd tests still pass.

N/A. No functional change.

## Notes for the Reviewer

- Single-file diff. Functional change is the new explanatory comment; the rest of the diff is a prettier-driven re-indent caused by moving the deps array onto its own line.
- Force-pushed twice on this PR: once to swap the deps approach for documentation per @illume's review, and once to address Copilot's comment-placement feedback.

### DIFF
--- a/frontend/src/components/crd/CustomResourceList.tsx
+++ b/frontend/src/components/crd/CustomResourceList.tsx
@@ -145,26 +145,31 @@ export function CustomResourceListTable(props: CustomResourceTableProps) {
     return cols;
   }, [crd, apiGroup]);
 
-  const cols = React.useMemo(() => {
-    const colsToDisplay: ResourceTableProps<KubeObject<KubeCRD>>['columns'] = [
-      {
-        label: t('translation|Name'),
-        getValue: resource => resource.metadata.name,
-        render: resource => <CustomResourceLink resource={resource} crd={crd} />,
-      },
-      ...(isMultiCluster ? (['cluster'] as ColumnType[]) : ([] as ColumnType[])),
-      ...additionalPrinterCols,
-      'labels',
-      'age',
-    ];
+  const cols = React.useMemo(
+    () => {
+      const colsToDisplay: ResourceTableProps<KubeObject<KubeCRD>>['columns'] = [
+        {
+          label: t('translation|Name'),
+          getValue: resource => resource.metadata.name,
+          render: resource => <CustomResourceLink resource={resource} crd={crd} />,
+        },
+        ...(isMultiCluster ? (['cluster'] as ColumnType[]) : ([] as ColumnType[])),
+        ...additionalPrinterCols,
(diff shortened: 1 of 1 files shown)


illume (changes requested):
I don't think t should be in there... I think it always breaks it.

Instead maybe add a comment there that t is left out on purpose.

WasThatRudy:
Thanks @illume. Reverting the deps change and replacing the bare suppression with a comment noting that `t` is intentionally excluded. Will push shortly.

WasThatRudy:
Pushed `20d62c1fd` addressing @Copilot's placement note: reformatted the `useMemo` to a multi-line shape so the explanatory comment now sits directly above the `eslint-disable-next-line` and the deps array, rather than after `return colsToDisplay;` where it could read as dead code. Prettier drove the re-indent of the callback body. Functional change is unchanged; updated the PR description to match.

illume (approved):
Thanks! 🎉

PR #5595, opened 2026-05-12
fix: Replace String.fromCharCode.apply with TextDecoder in KubeConfigLoader

### LINKED ISSUE #5594: KubeConfigLoader crashes with RangeError on large KubeConfig files
## Describe the bug

The `KubeConfigLoader` component crashes when a user uploads a large KubeConfig file. This is caused by `String.fromCharCode.apply` which has a maximum argument limit based on the JavaScript call stack size.

When the file exceeds this limit the browser throws a `RangeError: Maximum call stack size exceeded` and the upload fails. This affects enterprise users who manage KubeConfigs with many clusters or large embedded certificates.

## Steps to Reproduce

1. Open Headlamp and go to the Add Cluster page
2. Select KubeConfig as the source
3. Upload a KubeConfig file larger than 128KB or with many entries
4. The application crashes and the console shows the `RangeError`

## Expected Behavior

KubeConfig files should be decoded and parsed safely regardless of their size.

## Actual Behavior

The application crashes due to a stack overflow in the file decoding logic.

## Proposed Fix

Replace `String.fromCharCode.apply` with the modern `TextDecoder` API which handles large buffers without stack size limitations:

```ts
const data = new TextDecoder().decode(reader.result as ArrayBuffer);
```

## Environment

- Affected file: `frontend/src/components/cluster/KubeConfigLoader.tsx`

YadavAkhileshh:
## Summary

Fixes a crash that occurs when uploading a large KubeConfig file. `String.fromCharCode.apply` spreads the entire file as function
arguments, which exceeds the JavaScript call stack limit on large files and throws a `RangeError`.

Replaced with the `TextDecoder` API which is the modern standard for decoding binary data in the browser and has no stack size limit.

## Related Issue

Fixes #5594

## Changes

- Replaced `String.fromCharCode.apply(null, [...new Uint8Array(...)])` with `new TextDecoder().decode(reader.result as ArrayBuffer)`  in `KubeConfigLoader.tsx`

## Testing

1. Upload a small KubeConfig and confirm it works as before
2. Upload a large KubeConfig with many clusters or large CA certs
3. Confirm no `RangeError` and file loads correctly
4. Confirm existing snapshots and tests still pass

### DIFF
--- a/frontend/src/components/cluster/KubeConfigLoader.tsx
+++ b/frontend/src/components/cluster/KubeConfigLoader.tsx
@@ -398,9 +398,7 @@ function KubeConfigLoader() {
     reader.onerror = () => setError(t("translation|Couldn't read kubeconfig file"));
     reader.onload = () => {
       try {
-        const data = String.fromCharCode.apply(null, [
-          ...new Uint8Array(reader.result as ArrayBuffer),
-        ]);
+        const data = new TextDecoder().decode(reader.result as ArrayBuffer);
         const doc = yaml.load(data) as kubeconfig;
         if (!doc.clusters) {
           throw new Error(t('translation|No clusters found!'));


illume (approved):
🎉 thanks!

PR #5602, opened 2026-05-13
frontend: DateLabel: Show full creation timestamp as tooltip on age hover

### LINKED ISSUE #5577: frontend: Age: Show exact creation timestamp in tooltip on hover
## Is your feature request related to a problem? Please describe.

The Age column across all resource list views (Pods, Deployments, Services,
ConfigMaps etc.) only shows relative time like "3d" or "5h". There is no
way to see the exact creation timestamp without clicking into the resource
detail page. When you are debugging an incident or doing a quick audit,
having to open each resource just to check a timestamp adds unnecessary
friction.

When hovering over the Age value in any resource list table, a tooltip
should appear showing the full creation timestamp. For example hovering
over "3d" would show "2026-05-09 14:32:07 UTC". A dashed underline on
the age value would act as a visual hint that hover is available.

Since the Age cell is rendered from a shared component, the change would
apply across all resource list views automatically without needing
individual updates to each one.

## Describe alternatives you've considered

Clicking into the resource detail page works but adds unnecessary navigation
overhead. Adding a separate column for the full timestamp would clutter
the table unnecessarily.

## Are you able to work on this?

Yes, I am planning to open a PR for this.

## Additional Context

Tools like kubectl, Lens and k9s all surface exact timestamps easily.
This would bring Headlamp in line with what users expect coming from
those tools. The fix is self-contained and minimal — only the shared
Age cell component needs to be updated.

**Current behavior vs proposed behavior:**

Image

vikash7485:
## Summary

This PR adds a tooltip to the Age column across all resource list views 
so users can see the exact creation timestamp on hover without having to 
navigate into the resource detail page.

The issue was that the DateLabel component passed hoverInfo to 
HoverInfoLabel but the tooltip was only triggered by hovering over the 
calendar icon, not the age text itself. Adding the labelFirst prop fixes 
this so the tooltip is accessible from the text as well.

Since DateLabel is a shared component the fix applies automatically across 
all resource list views including Pods, Deployments, Services and ConfigMaps.

fixes #5577 

## Steps to test

1. Open Workloads > Pods in Headlamp
2. Hover over any value in the Age column
3. Verify tooltip appears showing the full creation timestamp
4. Repeat for Deployments, Services and ConfigMaps
5. Run npm run frontend:lint
6. Run npm run frontend:test

### DIFF
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -247,6 +247,7 @@ export function DateLabel(props: DateLabelProps) {
       hoverInfo={localeDate(date)}
       icon="mdi:calendar"
       iconProps={iconProps}
+      labelFirst
     />
   );
 }
--- a/frontend/src/components/common/Label.tsx
+++ b/frontend/src/components/common/Label.tsx
@@ -247,7 +247,7 @@ export function DateLabel(props: DateLabelProps) {
       hoverInfo={localeDate(date)}
       icon="mdi:calendar"
       iconProps={iconProps}
-      labelFirst
+      iconPosition="end"
     />
   );
 }


illume (changes requested):
Thanks for the contribution.


<details>

</details>

vikash7485:
@illume , thanks for the feedback. I have rebased the branch onto main. 
Please let me know if anything else needs to be fixed.

illume (changes requested):
Thanks for these changes.


illume (changes requested):
Can you please explain a bit more how I can reproduce the issue?

In main branch, if I hover I can see the full date:

Image

vikash7485:
@illume , thanks for checking this. You are right, I can see the 
tooltip was already working on hover over the calendar icon 
in main. I misunderstood the original issue — I thought the 
tooltip was not showing at all but it was only visible on 
the icon and not on the text itself.

Since the tooltip is already working correctly, I think this 
PR does not add value and should be closed. Sorry for the 
confusion and thank you for your patience in reviewing it.

illume:
Thanks @vikash7485 for the info.

I'll close this now. I hope this experience was still useful to learn some of the contributing guidelines.

PR #5605, opened 2026-05-13
frontend: ClusterChooserPopup: Make cluster filter case-insensitive

### LINKED ISSUE #5600: Cluster filter in ClusterChooserPopup is case-sensitive
## Describe the bug

The cluster search filter in the cluster chooser popup uses a case-sensitive `includes()` check. Typing "PROD" will not find "production" and typing "prod" will not find "PROD-cluster".

## Steps to Reproduce

1. Open Headlamp with multiple clusters configured
2. Open the cluster chooser popup
3. Type the cluster name in uppercase
4. Observe matching clusters in lowercase are not shown

## Expected Behavior

Filter should match clusters regardless of case.

## Actual Behavior

Only exact case matches are returned.

## Proposed Fix

```ts
allClusters = allClusters.filter(cluster =>
  cluster.name.toLowerCase().includes(filter.toLowerCase())
);
```

YadavAkhileshh:
## Summary
Fixes #5600  makes the cluster chooser filter case-insensitive.

## Changes
- Converted both cluster name and filter to lowercase before comparison

## Testing
1. Open the cluster chooser popup
2. Type a cluster name in any case (upper, lower, mixed)
3. Verify matching works regardless of case

### DIFF
--- a/frontend/src/components/cluster/ClusterChooserPopup.tsx
+++ b/frontend/src/components/cluster/ClusterChooserPopup.tsx
@@ -109,7 +109,9 @@ function ClusterChooserPopup(props: ChooserPopupPros) {
   const [recentClusters, clustersToShow] = React.useMemo(() => {
     let allClusters = Object.values(clusters || {});
     if (filter !== '') {
-      allClusters = allClusters.filter(cluster => cluster.name.includes(filter));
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

PR #5519, opened 2026-05-10
electron: quit app on window close outside macOS

Mohammed-Thaha:
## Summary

Fixes window close behavior on Linux and Windows.
Clicking the window close button (X) now fully quits the app instead of keeping Headlamp running in the background.

## Related Issue

Closes #5445

## Changes

- updated electron window close behavior
- fixed app remaining active in the background after clicking the window close button
- preserved existing macOS tray behavior

## Steps to Test

1. Build and launch the Headlamp desktop application
1. Run `ps aux | grep -i headlamp` and confirm Headlamp processes are running
1. Click the window close button (X)
1. Run `ps aux | grep -i headlamp` again
1. Verify that Electron and headlamp-server processes are no longer running

## Screenshots 

Build and run 
Screenshot from 2026-05-10 13-36-20

Verify processes before and after closing the application using the window close button (X)

Screenshot from 2026-05-10 13-35-31

## Notes for the Reviewer

Tested on Ubuntu 22.04.5 LTS using both development mode and packaged production build.

PR #5618, opened 2026-05-13
app: electron: Fix window reopen failure and server process cleanup on Windows

### LINKED ISSUE #5445: Closing window via window close button hides UI but doesn't shut down the process (Linux)
## Description

On the Headlamp desktop app, clicking the window close button (the "X" in the title bar) only closes the window — the Headlamp process keeps running in the background. Using the menu **File → Quit** correctly shuts the process down.

The two actions should be consistent: clicking the window close button should fully exit the application (or at least free the underlying server process), not leave it lingering.

## Impact

- Users assume the app has been closed when the window disappears, but the process and its child server keep running, holding ports and resources.
- Re-launching Headlamp can lead to confusing behavior because a previous instance is still alive.
- Only users who know to use **File → Quit** actually shut it down.

## Steps to reproduce

1. Launch Headlamp desktop app on Linux.
2. Click the window close button ("X") in the title bar.
3. Check running processes (e.g. `ps aux | grep -i headlamp`) — the Headlamp process is still running.
4. Compare with: launch Headlamp → use **File → Quit** → process is gone as expected.

## Expected behavior

Clicking the window close button should fully terminate the Headlamp process, the same way **File → Quit** does (or, if backgrounding is intentional, this should be made explicit, e.g. via a tray icon and/or a setting).

## Environment

- OS: Linux
- Headlamp version: 0.42.0
- Git commit: a69a04cc28ede29c263b40e00654efadd4fa5b24

ChandanKT-git:
## Summary

This PR fixes a bug where Headlamp fails to reopen after closing on Windows, and ensures the backend server process tree is fully terminated on quit.

## Related Issue

Fixes #5445

## Changes

- Fixed `second-instance` handler in `app/electron/main.ts` to call `mainWindow.show()` before `focus()`, so hidden windows (from tray hide-on-close) become visible when re-launching
- Fixed `quitServerProcess()` in `app/electron/main.ts` to use `taskkill /T /F` on Windows for reliable process tree cleanup, matching the existing `killProcess()` behavior

## Steps to Test

1. Build and launch the Headlamp desktop application
2. Close the window via the X button
3. Run `tasklist | findstr headlamp` (Windows) or `ps aux | grep headlamp` (Linux/macOS) and confirm no Headlamp processes remain
4. Re-launch Headlamp and verify it starts normally without needing to force-kill background processes
5. (macOS only) Close the window — verify it hides to tray. Re-launch from Dock — verify the window reappears

N/A — no UI changes, behavioral fix only.

## Notes for the Reviewer

- The `second-instance` fix is needed for macOS where the tray hide-on-close behavior is still active (via PR #5519). Without `show()`, `focus()` alone does not make a hidden `BrowserWindow` visible.
- The `taskkill /T /F` approach in `quitServerProcess()` is consistent with the existing `killProcess()` function (line ~1318) which already uses `taskkill` for Windows. The server is spawned with `detached: true`, so `serverProcess.kill()` only terminates the root process — not child processes that hold ports.
- Tested locally: `app:tsc` ✅, `app:lint` ✅, `app:test:unit` ✅ (103/103 passed).

## Note on Dependency Updates: 

- You will notice a second commit in this PR that updates package-lock.json and package.json in the plugins/headlamp-plugin directory. This was necessary to fix a failing CI check-dependencies job. A recent merge to main updated the frontend dependencies (adding `remark-gfm` and bumping Storybook) but did not sync them to the plugin infrastructure. I ran npm run update-dependencies to fix the CI drift so this PR could pass.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -857,9 +857,20 @@ function quitServerProcess() {
   }
 
   serverProcess.stdin.destroy();
-  // @todo: should we try and end the process a bit more gracefully?
-  //       What happens if the kill signal doesn't kill it?
-  serverProcess.kill();
+
+  if (process.platform === 'win32') {
+    // On Windows, use taskkill with /T to kill the entire process tree.
+    // serverProcess.kill() only terminates the root process, leaving
+    // detached children (spawned with detached: true) running.
+    try {
+      execSync('taskkill /pid ' + serverProcess.pid + ' /T /F');
+    } catch (e) {
+      console.error('Failed to kill server process tree with taskkill:', e);
+      serverProcess.kill();
+    }
+  } else {
+    serverProcess.kill();
+  }
 
   serverProcess = null;
 }
@@ -1606,6 +1617,7 @@ function startElectron() {
       app.on('second-instance', () => {
         // Someone tried to run a second instance, we should focus our window.
         if (mainWindow) {
+          mainWindow.show();
           if (mainWindow.isMinimized()) mainWindow.restore();
           mainWindow.focus();
         }
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -844,7 +844,12 @@ let intentionalQuit: boolean;
 let serverProcessQuit: boolean;
 
 function quitServerProcess() {
-  if (!serverProcess || serverProcessQuit) {
+  if (!serverProcess) {
+    console.info('server process already not running');
+    return;
+  }
+
+  if (serverProcessQuit && process.platform !== 'win32') {
     console.info('server process already not running');
     return;
   }
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -857,10 +857,6 @@ function quitServerProcess() {
   intentionalQuit = true;
   console.info('stopping server process...');
 
-  if (!serverProcess) {
-    return;
-  }
-
   serverProcess.stdin.destroy();
 
   if (process.platform === 'win32') {
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -845,7 +845,7 @@ let serverProcessQuit: boolean;
 
 function quitServerProcess() {
   if (!serverProcess || serverProcessQuit) {
-    console.error('server process already not running');
+    console.info('server process already not running');
     return;
   }
 
@@ -1855,6 +1855,8 @@ if (!isRunningScript) {
  * @param  {ChildProcess} serverProcess to attach the error handlers to.
  */
 function attachServerEventHandlers(serverProcess: ChildProcessWithoutNullStreams) {
+  serverProcessQuit = false;
+
   serverProcess.on('error', err => {
     console.error(`server process failed to start: ${err}`);
   });
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -859,13 +859,16 @@ function quitServerProcess() {
   serverProcess.stdin.destroy();
 
   if (process.platform === 'win32') {
-    // On Windows, use taskkill with /T to kill the entire process tree.
+    // On Windows, reuse killProcess() so the process-tree termination logic
+    // stays centralized and does not diverge from the shared implementation.
     // serverProcess.kill() only terminates the root process, leaving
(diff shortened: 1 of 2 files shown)


illume:
I thought this was already fixed?

ChandanKT-git:
> I thought this was already fixed?

PR #5519 partially addressed the issue it restricted the hide-on-close behavior to macOS only (via `process.platform === 'darwin`'), which prevents the window from being hidden on Windows/Linux. That was a great fix for the primary reported symptom.

However, this PR addresses two remaining gaps that #5519 didn't cover:

1. `second-instance` handler is missing `mainWindow.show()` , On macOS, where the tray hide-on-close is still active, if a user closes the window (it hides) and then re-launches the app from Dock, the second-instance handler only calls focus(), which doesn't make a hidden BrowserWindow visible. Adding show() before focus() fixes this.

2. `quitServerProcess()` doesn't kill the full process tree on Windows , The server is spawned with detached: true, so `serverProcess.kill()` only terminates the root process, not its children. This can leave orphaned server processes holding ports. The existing `killProcess()` function (line 1317) already handles this correctly with `taskkill /T /F`, but `quitServerProcess()` wasn't using it. This PR brings them in line.

Happy to split this into separate PRs if you'd prefer, or make any other adjustments!

illume (changes requested):
Thanks for the explanation.

Can you please remove the headlamp-plugin related changes?
(also please rebase against main)

illume (changes requested):
Thanks for these changes.


illume (changes requested):
Thanks for the contribution.


The frontend lint job in CI is failing. Run `cd frontend && npm run lint` to see the errors.

<details>

</details>

app/electron/main.ts
```diff
@@ -1606,6 +1617,7 @@ function startElectron() {
       app.on('second-instance', () => {
         // Someone tried to run a second instance, we should focus our window.
         if (mainWindow) {
+          mainWindow.show();
```
joaquimrocha, line 1620: This should be in a separate commit. We shouldn't have a commit that fixes window showing + cleaning up the app quitting process in one.

app/package-lock.json
joaquimrocha, line 1: Why is the package-lock.json changed in this commit?

app/electron/main.ts
```diff
@@ -859,13 +859,16 @@ function quitServerProcess() {
   serverProcess.stdin.destroy();
 
   if (process.platform === 'win32') {
-    // On Windows, use taskkill with /T to kill the entire process tree.
+    // On Windows, reuse killProcess() so the process-tree termination logic
+    // stays centralized and does not diverge from the shared implementation.
```
joaquimrocha, line 863: Do not make this a separate commit, since the changed code was introduced in a commit in this PR. Look into the Headlamp contribution docs that mention atomic commits please.

app/electron/main.ts
```diff
@@ -844,7 +844,7 @@ let intentionalQuit: boolean;
 let serverProcessQuit: boolean;
 
 function quitServerProcess() {
-  if ((!serverProcess || serverProcessQuit) && process.platform !== 'win32') {
+  if (!serverProcess || serverProcessQuit) {
```
joaquimrocha, line 847: Not sure if the commit description is accurate to these changes.

app/electron/main.ts
```diff
@@ -844,7 +844,12 @@ let intentionalQuit: boolean;
 let serverProcessQuit: boolean;
 
 function quitServerProcess() {
-  if (!serverProcess || serverProcessQuit) {
+  if (!serverProcess) {
+    console.info('server process already not running');
+    return;
+  }
+
+  if (serverProcessQuit && process.platform !== 'win32') {
```
joaquimrocha, line 852: This seems another commit that should be squashed.

joaquimrocha (changes requested):

sniok:
this issue seems to be already fixed in https://github.com/kubernetes-sigs/headlamp/pull/5519
could you please check if this issue still appears for you on main? if not I'd like to close this PR

PR #5639, opened 2026-05-14
docs: Update traefik middleware apiVersion

skoeva:
This change updates the Traefik Middleware apiVersion in the basic auth documentation from the deprecated `traefik.containo.us/v1alpha1` to the current `traefik.io/v1alpha1`.


sniok (approved):
thanks!

/lgtm

PR #5641, opened 2026-05-14
fix(frontend): resolve owner references using apiVersion+kind to prevent wrong resource links

### LINKED ISSUE #5569: Resource: Owner references can link to the wrong resource when kinds overlap
## Describe the bug
Resource metadata owner references are currently resolved by `kind` when rendering `Controlled by`.
This can route users to the wrong details page when a custom resource shares a kind with a built-in Kubernetes resource. For example, a Volcano PodGroup can have an owner reference like `apiVersion: batch.volcano.sh/v1alpha1`, `kind: Job`, but Headlamp treats it as a built-in Kubernetes Job and links to the built-in Job details page.
The expected behavior is that Headlamp should not link to the wrong built-in resource. Ideally, plugins should also have a clean way to route these owner references to their own detail pages, or hide the default owner reference row and render their own.
## To Reproduce
Steps to reproduce the bug:
1. Install a plugin or custom resource that uses an owner reference with the same `kind` as a built-in Kubernetes resource (for example: Volcano).
2. Open the details page for the owned resource.
3. Look at the `Controlled by` row in the metadata section.
4. Click the owner reference link.
5. See that Headlamp navigates to the built-in resource page instead of the plugin/custom resource detail page.
 - Installation type: Development build / Plugin development
 - Headlamp Version: main
 - Other: Seen while working on Volcano resources where `batch.volcano.sh/v1alpha1 Job` conflicts with built-in Kubernetes `batch/v1 Job`.
Yes. I can propose a PR after discussing the preferred approach with maintainers.
## Additional Context
A minimal fix would be to only link built-in owner references when both API group and kind match. If the API group does not match, Headlamp could render the owner reference as plain text instead of linking to the wrong page (Simple but i don't think it's ideal).
For plugin UX, there may be better options worth discussing such as:
- Add an owner-reference resolver API so plugins can route owner references to plugin detail pages.
- Add a way for plugins/details pages to hide the default `Controlled by` metadata row and render their own.
Screenshot example for more understanding:

Image

Here the PodGroup is controlled by the Job, but when clicking on that link, it treats that Job as a Kubernetes Job not a Volcano Job, so the details page doesn't load correctly

DashratRajpurohit:
# Description

### Latest Changes
- Updated OwnerReference resolution to match both kind and apiVersion.
- - Added support for array-based apiVersion definitions (e.g. Ingress).
- - Added unit tests in MetadataDisplay.test.tsx.
- - Fixed a template literal bug in the console error log.
- 

Fixes #5569

## Problem

When Headlamp shows **"Controlled by"** in a resource's metadata panel, it resolved
the owner reference link using **only the `kind` field**. But Kubernetes allows
different API groups to use the **same kind name**:

| Kind | apiVersion | Source |
|------|-----------|--------|
| `Job` | `batch/v1` | Built-in Kubernetes |
| `Job` | `batch.volcano.sh/v1alpha1` | Volcano CRD |
| `ReplicaSet` | `apps/v1` | Built-in Kubernetes |
| `ReplicaSet` | `custom.io/v1` | Custom CRD |

So if a Pod was owned by a Volcano `Job`, Headlamp would silently link to a
`batch/v1 Job` — **the wrong resource entirely**.

This is a silent correctness bug — the UI shows a clickable link that navigates
to the wrong place with no error shown to the user.

---

## Root Cause

In `frontend/src/components/common/Resource/MetadataDisplay.tsx`, the
`makeOwnerReferences` function matched resources using only `ownerRef.kind`:

```tsx
// ❌ BEFORE — only kind is checked, apiVersion is ignored
if (ownerRef.kind in ResourceClasses) {
  routeName = ResourceClasses[ownerRef.kind as keyof typeof ResourceClasses].detailsRoute;
  // ^ picks batch/v1 Job even when the actual owner is a volcano.sh Job
}
```

The `KubeOwnerReference` type **already has** an `apiVersion` field that
uniquely identifies the API group — it was simply never used in the lookup.

Additionally, a minor pre-existing bug was found in the same function:
the `console.error` call used `{ownerRef.kind}` (a plain object literal syntax)
instead of `${ownerRef.kind}` (a template literal), so error messages were
never correctly interpolated.

---

## Fix

```tsx
// ✅ AFTER — both kind AND apiVersion must match
const matchingClass =
  ownerRef.kind in ResourceClasses
    ? (() => {
        const cls = ResourceClasses[ownerRef.kind as keyof typeof ResourceClasses];
        // Only return the class if the apiVersion also matches the owner reference
        return cls.apiVersion === ownerRef.apiVersion ? cls : null;
      })()
    : null;

if (matchingClass) {
  routeName = matchingClass.detailsRoute; // now guaranteed to be the correct class
}
```

If the `apiVersion` does **not** match (e.g. a CRD with the same kind name as a
built-in resource), the owner reference is rendered as **plain text** instead of
an incorrect clickable link.

---

## Changes

| File | Description |
|------|-------------|
| `frontend/src/components/common/Resource/MetadataDisplay.tsx` | Narrow owner reference lookup to require both `kind` and `apiVersion` to match; fix template literal in `console.error` |
| `frontend/src/components/common/Resource/MetadataDisplay.test.tsx` | Added unit tests to assert correct owner reference matching with `apiVersion` and fallback to plain text |

**Diff size:** 2 files · +229 additions · −4 deletions

---

### ✅ Case 1 — Standard built-in resource (link still works correctly)

```yaml
# Pod owned by a standard built-in Kubernetes Job
ownerReferences:
  - apiVersion: batch/v1        # matches Job.apiVersion in ResourceClasses ✅
    kind: Job
    name: my-batch-job
    uid: abc-123
```

**Before fix:** Clickable link → correct `batch/v1 Job` ✅  
**After fix:** Clickable link → correct `batch/v1 Job` ✅ *(no regression)*

---

### ✅ Case 2 — CRD with the same kind name (wrong link is now prevented)

```yaml
# Pod owned by a Volcano CRD Job (NOT a built-in batch/v1 Job)
ownerReferences:
  - apiVersion: batch.volcano.sh/v1alpha1   # does NOT match Job.apiVersion ❌
    kind: Job
    name: my-volcano-job
    uid: def-456
```

**Before fix:** Broken link → navigates to `batch/v1 Job` named `my-volcano-job`
which likely does not exist ❌  
**After fix:** Plain text `Job: my-volcano-job` — no incorrect link ✅

---

### ✅ Case 3 — Completely unknown CRD (unchanged behaviour)

```yaml
# Resource owned by an unrecognised CRD
ownerReferences:
  - apiVersion: custom.example.com/v1
    kind: FooController
    name: foo-ctrl
    uid: ghi-789
```

**Before fix:** Plain text `FooController: foo-ctrl` ✅  
**After fix:** Plain text `FooController: foo-ctrl` ✅ *(no change)*

---

### ✅ Case 4 — console.error template literal fix

```ts
// Before (bug): prints the literal string "{ownerRef.kind}" — not the value
console.error(`Error getting routeName for {ownerRef.kind}`, e);

// After (fixed): correctly prints the actual kind value
console.error(`Error getting routeName for ${ownerRef.kind}`, e);
```

---

## How to Test Manually

1. Install a CRD on your cluster that uses a `kind` name identical to a built-in
   resource (e.g. `kind: Job` with `apiVersion: batch.volcano.sh/v1alpha1`)
2. Create a workload whose `ownerReferences` points to that CRD resource
3. Open Headlamp and navigate to the owned resource's detail page
4. Observe the **"Controlled by"** row in the metadata panel:
   - **Before fix:** Shows a clickable link to the wrong `batch/v1 Job`
   - **After fix:** Shows plain text since the `apiVersion` does not match

---

## Checklist

- [x] Linked to issue #5569
- [x] Files modified (`MetadataDisplay.tsx` and `MetadataDisplay.test.tsx` for unit tests)
- [x] `npm run lint` passes — exit code `0`
- [x] No breaking changes — built-in resources continue to produce correct links
- [x] Backward compatible with all existing `ownerReferences` rendering behaviour
- [x] Inline comment added explaining the fix with a link to the issue
- [x] Bonus bug fixed: `console.error` template literal interpolation

### DIFF
--- a/frontend/src/components/common/Resource/MetadataDisplay.tsx
+++ b/frontend/src/components/common/Resource/MetadataDisplay.tsx
@@ -69,13 +69,32 @@ export function MetadataDisplay<T extends KubeObject>(props: MetadataDisplayProp
 
     return ownerReferences
       .map((ownerRef, i) => {
-        if (ownerRef.kind in ResourceClasses) {
+        // Match both kind AND apiVersion to avoid linking to the wrong resource when
+        // different API groups share the same kind name (e.g. batch/v1 Job vs a CRD Job).
+        const matchingClass =
+          ownerRef.kind in ResourceClasses
+            ? (() => {
+                const cls = ResourceClasses[ownerRef.kind as keyof typeof ResourceClasses];
+                const apiVersions = Array.isArray(cls.apiVersion)
+                  ? cls.apiVersion
+                  : [cls.apiVersion];
+                const apiVersionMatches = apiVersions.includes(ownerRef.apiVersion);
+                return apiVersionMatches ? cls : null;
+              })()
+            : null;
+
+        if (matchingClass) {
           let routeName;
           try {
-            routeName = ResourceClasses[ownerRef.kind as keyof typeof ResourceClasses].detailsRoute;
+            routeName = matchingClass.detailsRoute;
           } catch (e) {
-            console.error(`Error getting routeName for {ownerRef.kind}`, e);
-            return null;
+            console.error(`Error getting routeName for ${ownerRef.kind}`, e);
+            return (
+              <>
+                {`${ownerRef.kind}: ${ownerRef.name}`}
+                {i < numItems - 1 && <br />}
+              </>
+            );
           }
           return (
             <>
--- a/frontend/src/components/common/Resource/MetadataDisplay.test.tsx
+++ b/frontend/src/components/common/Resource/MetadataDisplay.test.tsx
@@ -1,0 +1,206 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from '@mui/material/styles';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { KubeObject } from '../../../lib/k8s/KubeObject';
+import { createMuiTheme } from '../../../lib/themes';
+import { TestContext } from '../../../test';
+import { MetadataDisplay } from './MetadataDisplay';
+
+// Use vi.hoisted to ensure the mock data is available before the vi.mock call
+const { mockResourceClasses } = vi.hoisted(() => {
+  return {
+    mockResourceClasses: {
+      Job: {
+        apiVersion: 'batch/v1',
+        detailsRoute: 'job',
+      },
+      Ingress: {
+        apiVersion: ['networking.k8s.io/v1', 'extensions/v1beta1'],
+        detailsRoute: 'ingress',
+      },
+    },
+  };
+});
+
+vi.mock('../../../lib/k8s', () => ({
+  ResourceClasses: mockResourceClasses,
+  useSelectedClusters: vi.fn(() => ['test-cluster']),
+}));
+
+const theme = createMuiTheme({ base: 'light', name: 'light' });
+
+describe('MetadataDisplay', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    // Reset any mutations to mockResourceClasses
+    mockResourceClasses.Job.detailsRoute = 'job';
+  });
+
+  const mockResource = {
+    metadata: {
+      name: 'test-resource',
+      namespace: 'test-namespace',
+      creationTimestamp: '2023-01-01T00:00:00Z',
+    },
+    cluster: 'test-cluster',
+    jsonData: {
+      metadata: {
+        name: 'test-resource',
+        namespace: 'test-namespace',
+        creationTimestamp: '2023-01-01T00:00:00Z',
+      },
+    },
+  } as unknown as KubeObject;
+
+  const renderWithTheme = (ui: React.ReactElement) => {
+    return render(
+      <TestContext>
+        <ThemeProvider theme={theme}>{ui}</ThemeProvider>
+      </TestContext>
+    );
+  };
+
+  it('renders a link when kind and apiVersion (string) match', () => {
+    const ownerReferences = [
+      {
+        apiVersion: 'batch/v1',
+        kind: 'Job',
+        name: 'test-job',
+        uid: 'uid-1',
+      },
+    ];
+    const resourceWithOwners = {
+      ...mockResource,
+      metadata: { ...mockResource.metadata, ownerReferences },
+    } as unknown as KubeObject;
+
+    renderWithTheme(<MetadataDisplay resource={resourceWithOwners} />);
+
+    const link = screen.getByRole('link', { name: 'Job: test-job' });
+    expect(link).toBeInTheDocument();
+  });
+
+  it('renders a link when kind and apiVersion (array) match', () => {
+    const ownerReferences = [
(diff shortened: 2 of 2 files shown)


illume (changes requested):
Thanks for these changes.


<details>

- `fix(frontend): resolve owner references using apiVersion+kind` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for the contribution.


DashratRajpurohit:
@illume all feedback has been addressed with the latest changes and unit tests. I have updated the PR description with the summary of changes.

illume (changes requested):
Thanks for these changes.


<details>

</details>


<details>

- `frontend: fix owner ref matching with apiVersion` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for working on this.


illume (changes requested):
Thanks for this PR.


<details>

</details>

DashratRajpurohit:
Hi @illume! I have successfully rebased this branch on top of the latest main as well to remove the merge commit. The history is now completely clean and linear with a single, DCO-signed commit. Tests are verified and passing!

illume (changes requested):
Thanks for this PR.


illume (changes requested):
Thanks for the contribution.


frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -69,13 +69,32 @@ export function MetadataDisplay<T extends KubeObject>(props: MetadataDisplayProp
 
     return ownerReferences
       .map((ownerRef, i) => {
-        if (ownerRef.kind in ResourceClasses) {
+        // Match both kind AND apiVersion to avoid linking to the wrong resource when
+        // different API groups share the same kind name (e.g. batch/v1 Job vs a CRD Job).
+        // See: https://github.com/kubernetes-sigs/headlamp/issues/5569
```
mahmoudmagdy1-1: I think it might be better to leave out the issue link as the lines above already explains it well

frontend/src/components/common/Resource/MetadataDisplay.tsx
```diff
@@ -69,13 +69,32 @@ export function MetadataDisplay<T extends KubeObject>(props: MetadataDisplayProp
...
+        const matchingClass =
+          ownerRef.kind in ResourceClasses
+            ? (() => {
+                const cls = ResourceClasses[ownerRef.kind as keyof typeof ResourceClasses];
+                const apiVersionMatches = ([] as string[])
+                  .concat(cls.apiVersion)
+                  .includes(ownerRef.apiVersion);
```
mahmoudmagdy1-1: This is a bit hard to read, how about this instead?

```suggestion
                const apiVersions = Array.isArray(cls.apiVersion) ? cls.apiVersion : [cls.apiVersion];
                const apiVersionMatches = apiVersions.includes(ownerRef.apiVersion);
```

mahmoudmagdy1-1 (commented):
Thank you for working on this, I've added some small comments, other than that it looks good

DashratRajpurohit:
Hello! I have updated the branch and resolved the review feedback:

- **Refactored `apiVersion` array check**: Replaced the `.concat()` logic with a cleaner `Array.isArray(cls.apiVersion)` check to determine if the resource matches the owner's `apiVersion` definition, improving readability and type safety.
- **Removed Issue Comment**: Deleted the extra `// See: https://github.com/kubernetes-sigs/headlamp/issues/5569` comment, as the surrounding logic and inline comments already explain the context clearly.
- **Verification**: Verified that all frontend tests pass cleanly (`npm run frontend:test`) and the TypeScript compiler checks run with zero errors (`npm run frontend:tsc`).

Please let me know if there's anything else that needs adjustment. Thank you!

mahmoudmagdy1-1:
LGTM

mahmoudmagdy1-1 (approved):

illume (approved):
🎉 thanks!

Issue #5627, opened 2026-05-14
backend: parseKubeConfig should stop after invalid JSON body


harrshita123:
## Description

The backend `/parseKubeConfig` handler keeps processing even after JSON decoding fails. In `backend/cmd/stateless.go`, the handler calls `http.Error(w, "Invalid JSON request body", http.StatusBadRequest)` when `decoder.Decode(&kubeconfigReq)` fails, but it does not return immediately. The request then continues with the zero-value payload, which can lead to unnecessary work and an invalid follow-up response.

## Steps to Reproduce

1. Run Headlamp with dynamic clusters enabled.
2. Send malformed JSON to the `/parseKubeConfig` endpoint.
3. Inspect the response and server behavior.

## Expected Behavior

The handler should stop as soon as JSON decoding fails and return only a clear `400 Bad Request` response.

## Actual Behavior

The handler writes the `400 Bad Request` error, then continues processing and may attempt to build and encode a config response afterward.

## Code Reference

`backend/cmd/stateless.go`

The decode error path currently writes an error but does not return:

```go
if err := decoder.Decode(&kubeconfigReq); err != nil {
    logger.Log(logger.LevelError, nil, err, "decoding config")
    http.Error(w, "Invalid JSON request body", http.StatusBadRequest)
}
```

## Possible Fix

Return immediately after writing the bad request response.

```go
if err := decoder.Decode(&kubeconfigReq); err != nil {
    logger.Log(logger.LevelError, nil, err, "decoding config")
    http.Error(w, "Invalid JSON request body", http.StatusBadRequest)
    return
}
```

PR #5644, opened 2026-05-14
backend: Stop parsing invalid kubeconfig JSON

### LINKED ISSUE #5598: backend: parseKubeConfig continues after invalid JSON body
## Description

The backend `/parseKubeConfig` handler continues processing after failing to decode an invalid JSON request body.

In `backend/cmd/stateless.go`, `parseKubeConfig` calls `http.Error(w, "Invalid JSON request body", http.StatusBadRequest)` when `decoder.Decode(&kubeconfigReq)` fails, but it does not return. The handler then continues with the zero-value request, calls `parseClusterFromKubeConfig`, builds a `clientConfig`, and attempts to encode a JSON response.

## Steps to Reproduce

1. Run Headlamp with dynamic clusters enabled.
2. Send malformed JSON to the parse kubeconfig endpoint:

```bash
curl -i -X POST http://localhost:4466/parseKubeConfig \
  -H 'Content-Type: application/json' \
  --data '{'
```

3. Inspect the response body and headers.

## Expected Behavior

The handler should stop immediately after the decode error and return only a clear `400 Bad Request` response.

## Actual Behavior

The handler writes the `400 Bad Request` error, then continues processing and attempts to write a config JSON response afterward.

## Code Reference

`backend/cmd/stateless.go`

The decode error path writes an error but does not return:

```go
if err := decoder.Decode(&kubeconfigReq); err != nil {
    logger.Log(logger.LevelError, nil, err, "decoding config")
    http.Error(w, "Invalid JSON request body", http.StatusBadRequest)
}
```

## Possible Fix

Return immediately after writing the bad request response.

```go
if err := decoder.Decode(&kubeconfigReq); err != nil {
    logger.Log(logger.LevelError, nil, err, "decoding config")
    http.Error(w, "Invalid JSON request body", http.StatusBadRequest)
    return
}
```

Sambhram1:
## Summary

This PR fixes the /parseKubeConfig handler so malformed JSON requests stop immediately after the decode error. Without the early return, the handler writes a 400 response and then continues with the zero-value request payload.

## Related Issue

Fixes #5598
Related #5627

## Changes

- Return immediately after /parseKubeConfig fails to decode the JSON request body.
- Add a regression test that posts malformed JSON and verifies only the bad request response is returned.

## Steps to Test

1. Run go test ./cmd -run TestParseKubeConfigInvalidJSONReturnsBadRequest -count=1 from the backend directory.
2. Run go test ./cmd -run TestStatelessClustersKubeConfig -count=1 from the backend directory.
3. Run git diff --check from the repository root.

Not applicable; backend handler change only.

## Notes for the Reviewer

Issue #5627 reports the same behavior as #5598, so this PR targets the older existing issue and references the newer duplicate.

### DIFF
--- a/backend/cmd/stateless.go
+++ b/backend/cmd/stateless.go
@@ -165,6 +165,8 @@ func (c *HeadlampConfig) parseKubeConfig(w http.ResponseWriter, r *http.Request)
 		logger.Log(logger.LevelError, nil, err, "decoding config")
 
 		http.Error(w, "Invalid JSON request body", http.StatusBadRequest)
+
+		return
 	}
 
 	kubeconfigs := kubeconfigReq.Kubeconfigs
--- a/backend/cmd/stateless_test.go
+++ b/backend/cmd/stateless_test.go
@@ -131,6 +131,7 @@ func TestParseKubeConfigInvalidJSONReturnsBadRequest(t *testing.T) {
 
 	token := uuid.New().String()
 	require.NoError(t, os.Setenv("HEADLAMP_BACKEND_TOKEN", token))
+
 	defer func() { _ = os.Unsetenv("HEADLAMP_BACKEND_TOKEN") }()
 
 	req := httptest.NewRequestWithContext(
--- a/backend/cmd/stateless_test.go
+++ b/backend/cmd/stateless_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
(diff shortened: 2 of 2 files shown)


skoeva (changes requested):
could you run `make backend-lint-fix` locally to address the backend linting issue?

illume (approved):
🎉 thanks!

illume (approved):
🎉 thanks!

PR #5645, opened 2026-05-14
helm: Simplify oidc-scopes conditional with externalSecret

### LINKED ISSUE #5744: charts test failing in CI
the charts templates need to be updated

skoeva:
This replaces the `hasScopes` parameter introduced in #5573 with a simpler conditional that reuses the existing `$oidc.scopes`/`$scopes` values check, consistent with how all other optional OIDC args are handled in the same template.

The `hasScopes` field is removed from `values.yaml` and the schema since it's no longer needed -- scopes are included when explicitly configured via `oidc.scopes` or the `OIDC_SCOPES` env var.

Fixes: #5744, supercedes from #5573


illume:
@skoeva I'm sorry. There was already a PR opened for this and someone assigned to the issue.

If you still want to make changes, feel free to reopen.

skoeva:
not necessary - does not handle edge case

PR #5653, opened 2026-05-14
backend: Return error when repository lock is not acquired

### LINKED ISSUE #5634: backend: Helm repo handlers should fail when lock acquisition does not succeed
## Description

The Helm repository helpers in `backend/pkg/helm/repository.go` return both `locked` and `err` from `lockRepositoryFile()`, but the callers only branch on `err`.

That means `addRepository()`, `RemoveRepository()`, and `UpdateRepository()` can continue into repo file reads and writes even if the lock was not actually acquired, as long as `err` is nil.

This is a separate bug from the lock timing issue: even with a correct timeout or retry delay, the code should still stop if the lock was not obtained.

## Steps to Reproduce

1. Start Headlamp with Helm support enabled.
2. Trigger two Helm repository operations at nearly the same time.
3. Force one operation to hold the repo lock while the other enters the helper.
4. Observe that the losing call can continue past the lock helper path instead of failing closed.

## Expected Behavior

If the Helm repository lock cannot be acquired, the operation should stop and return an error before reading or writing the repository file.

## Actual Behavior

The helpers only check `err` after `lockRepositoryFile()` returns. If `locked` is false and `err` is nil, the code still proceeds to load and modify the repo file.

## Code Reference

`backend/pkg/helm/repository.go`

```go
locked, fileLock, err := lockRepositoryFile(lockCtx, settings.RepositoryConfig)
if err == nil && locked {
    defer func() {
        err := fileLock.Unlock()
        if err != nil {
            logger.Log(logger.LevelError, nil, err, "unlocking repository config file")
        }
    }()
}

if err != nil {
    logger.Log(logger.LevelError, nil, err, "locking repository config file")
    return err
}
```

## Possible Fix

Treat `locked == false` as an error and return immediately before continuing with repository reads or writes.

infinityanant:
## Summary
This PR fixes a concurrency bug in the Helm repository helpers where a failed 
lock acquisition could go undetected, allowing unsafe reads and writes to the 
repository config file.

## Related Issue
Fixes #5634

## Changes
- Fixed lock check logic in `addRepository`, `RemoveRepository`, and 
  `UpdateRepository` in `backend/pkg/helm/repository.go`
- All three functions now explicitly check `!locked` after `lockRepositoryFile` 
  returns, and return an error immediately if the lock was not acquired
- Moved the `defer fileLock.Unlock()` to only run after both guards pass, 
  so we never try to unlock a lock we don't own

## Steps to Test
1. Run `go build ./pkg/helm/...` from the `backend/` directory — should 
   produce no output (clean build)
2. Trigger two concurrent Helm repository operations (e.g., two simultaneous 
   AddRepo calls)
3. The operation that loses the lock race should now return 
   `"failed to acquire repository lock"` instead of silently proceeding
4. Check logs — no unlocking errors should appear from a lock that was 
   never held

N/A — this is a backend-only logic fix with no UI impact

## Notes for the Reviewer
- The bug was subtle: `lockRepositoryFile` returns three values (`locked`, 
  `fileLock`, `err`), but the original code only branched on `err`. If 
  `locked == false` and `err == nil`, the code would silently fall through 
  and proceed without the lock. This PR closes that gap in all three callers.
- The fix pattern is now consistent across all three functions — check `err`, 
  then check `!locked`, then set up the defer. Same order everywhere.

### DIFF
--- a/backend/pkg/helm/repository.go
+++ b/backend/pkg/helm/repository.go
@@ -108,19 +108,21 @@ func addRepository(name string, url string, settings *cli.EnvSettings) error {
 	defer cancel()
 
 	locked, fileLock, err := lockRepositoryFile(lockCtx, settings.RepositoryConfig)
-	if err == nil && locked {
-		defer func() {
-			err := fileLock.Unlock()
-			if err != nil {
-				logger.Log(logger.LevelError, nil, err, "unlocking repository config file")
-			}
-		}()
-	}
-
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "locking repository config file")
 		return err
 	}
+
+	if !locked {
+		return errors.New("failed to acquire repository lock")
+	}
+
+	defer func() {
+		err := fileLock.Unlock()
+		if err != nil {
+			logger.Log(logger.LevelError, nil, err, "unlocking repository config file")
+		}
+	}()
 
 	// read repo file
 	repoFile, err := repo.LoadFile(settings.RepositoryConfig)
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for these changes.


<details>

- `backend: Return error when repository lock is not acquired` — Only one file changed inside `backend/`; add a sub-area so it's clear what was touched (e.g. `backend: ComponentName: description`).

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for this PR.


<details>

</details>

illume (changes requested):
Thanks for these changes.


<details>

</details>

infinityanant:
This PR references an issue that has already been resolved and closed, so I’m closing this PR.

PR #5654, opened 2026-05-14
backend: helm: Fail when repo lock is not acquired

### LINKED ISSUE #5634: backend: Helm repo handlers should fail when lock acquisition does not succeed

harrshita123:
## Summary

  This PR fixes a backend Helm repository locking bug by making repository add, remove, and update operations fail closed when the repository file lock is not acquired.

  Previously, these paths only checked err from lockRepositoryFile(). If locked was false and err was nil, the code could continue into repository file reads and writes without holding the lock.

  ## Related Issue

  Fixes #5634

  ## Changes

  - Added an explicit errRepositoryLockNotAcquired error for unsuccessful lock acquisition.
  - Added a small helper to validate both locked and err from lockRepositoryFile().
  - Updated Helm repository add, remove, and update paths to stop before reading or writing when the lock is not confirmed.
  - Added regression coverage for the locked == false && err == nil case.

  ## Steps to Test

  1. Run cd backend && go test -v ./pkg/helm.
  2. Run npm run backend:lint.
  3. Confirm Helm repository tests pass and lint reports 0 issues.

  Not applicable. This is a backend-only change.

  ## Notes for the Reviewer

  This keeps the existing lock/unlock flow intact and only adds the missing fail-closed check after lock acquisition.

### DIFF
--- a/backend/pkg/helm/repository.go
+++ b/backend/pkg/helm/repository.go
@@ -40,7 +40,10 @@ const (
 	defaultNewConfigFolderMode os.FileMode = os.FileMode(0o770)
 )
 
-var errRepositoryNotFound = errors.New("repository not found")
+var (
+	errRepositoryNotFound        = errors.New("repository not found")
+	errRepositoryLockNotAcquired = errors.New("repository lock not acquired")
+)
 
 // add repository.
 type AddUpdateRepoRequest struct {
@@ -94,6 +97,18 @@ func lockRepositoryFile(lockCtx context.Context, repositoryConfig string) (bool,
 	return locked, fileLock, err
 }
 
+func ensureRepositoryFileLocked(locked bool, err error) error {
+	if err != nil {
+		return err
+	}
+
+	if !locked {
+		return errRepositoryLockNotAcquired
+	}
+
+	return nil
+}
+
 const timeoutForLock = 30 * time.Second
 
 // Adds a repository with name, url to the helm config. Returns error if there is one.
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

PR #5657, opened 2026-05-14
backend: headlamp_test: Add negative test for unauthed callers

vyncent-t:
## Summary

The backend's mutating/proxy endpoints (`/cluster`, `/plugins/{name}`, `/clusters/{name}/helm/...`) are gated by the `X-HEADLAMP_BACKEND-TOKEN` header, which must match the `HEADLAMP_BACKEND_TOKEN` env var set by the desktop app. Until now there was no negative test proving the gate actually rejects unauthenticated callers a regression that removed `checkHeadlampBackendToken` from any of those handlers would have shipped green. The gate also has a deliberate carve-out (`UseInCluster == true` skips the check entirely), and that was equally untested.

This PR adds three focused tests in `backend/cmd/headlamp_test.go` that pin the gate's behavior on both sides.

## Changes

Added in `backend/cmd/headlamp_test.go`:

- **`TestRestrictedEndpointsRequireToken`** — table-driven over the four currently-gated routes:
  - `POST /cluster`
  - `DELETE /cluster/{name}`
  - `DELETE /plugins/{name}`
  - `GET /clusters/{name}/helm/...` (Helm proxy)

  For each route it runs three subtests:
  - missing `X-HEADLAMP_BACKEND-TOKEN` header → expects `403 Forbidden`
  - wrong token value → expects `403 Forbidden`
  - correct token → expects status to NOT be `401`/`403` (we only pin the auth gate, not the response body)

Each subtest builds a fresh handler with its own `t.TempDir()` so the "valid token" subtest, which actually mutates state, can't bleed into the others. Env var is set with `t.Setenv` so it's restored automatically. The test handler sets `EnableHelm: true` so the Helm route is actually registered (otherwise the generic `/clusters/{name}/...` proxy catches the path and the test would observe proxy-layer status codes instead of the auth gate).

- **`TestRestrictedEndpointsRejectEmptyEnvToken`** — pins a second property of `checkHeadlampBackendToken`: when `HEADLAMP_BACKEND_TOKEN` is empty, every request must be rejected, including one with an empty header. Without this, clearing the env var would silently disable the gate.

- **`TestRestrictedEndpointsBypassedInCluster`**  — pins the other half of the gate. With `UseInCluster: true` and a non-empty `HEADLAMP_BACKEND_TOKEN` env var, requests with **no** `X-HEADLAMP_BACKEND-TOKEN` header must not be rejected on any of the four gated routes. A regression that made the gate apply in-cluster would break every in-cluster install; this test makes that a CI failure instead of a deployment incident.

## Steps to Test

1. `cd backend && go test -count=1 -run 'TestRestrictedEndpoints' ./cmd/...` all three tests should pass.
3. Full backend suite: `npm run backend:test`.

## Notes for the Reviewer

- **Scope is intentionally tight.** No production code is changed; this is purely test coverage of behavior that already exists.
- **`PUT /cluster/{name}` (`renameCluster`) is deliberately excluded from the table.** While writing these tests I noticed `renameCluster` does not call `checkHeadlampBackendToken`, unlike its `addCluster`/`deleteCluster` siblings so it's currently reachable without the backend token. Adding it to this table would assert behavior that does not exist and fail CI. The omission is documented in a one-line comment on `TestRestrictedEndpointsRequireToken`.
- The expected rejection status is `403 Forbidden` (not `401`) because that's what `checkHeadlampBackendToken` returns via `http.Error`. The "valid token" and in-cluster branches assert `!= 403 && != 401`, so they stay robust to handlers returning `200`/`201`/`400`/`404`/`5xx` depending on body and cluster state.

### DIFF
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -1252,6 +1252,248 @@ func TestDeletePlugin(t *testing.T) {
 	assert.True(t, os.IsNotExist(err))
 }
 
+// TestRestrictedEndpointsRequireToken is the canary for the backend-token gate:
+// dropping checkHeadlampBackendToken from any listed route would fail here.
+// PUT /cluster/{name} (renameCluster) is omitted because it isn't gated yet.
+func TestRestrictedEndpointsRequireToken(t *testing.T) {
+	const validToken = "valid-token-for-test"
+
+	t.Setenv("HEADLAMP_BACKEND_TOKEN", validToken)
+
+	kubeConfigBytes, err := os.ReadFile("./headlamp_testdata/kubeconfig")
+	require.NoError(t, err)
+
+	kubeConfigB64 := base64.StdEncoding.EncodeToString(kubeConfigBytes)
+
+	cases := []struct {
+		name   string
+		method string
+		path   string
+		body   interface{}
+	}{
+		{
+			name:   "POST /cluster (addCluster)",
+			method: http.MethodPost,
+			path:   "/cluster",
+			body:   ClusterReq{KubeConfig: &kubeConfigB64},
+		},
+		{
+			name:   "DELETE /cluster/{name} (deleteCluster)",
+			method: http.MethodDelete,
+			path:   "/cluster/" + minikubeName,
+			body:   nil,
+		},
+		{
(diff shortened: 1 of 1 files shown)


sniok (approved):
thanks!

/lgtm

PR #5391, opened 2026-05-05
frontend: router: Remove hardcoded redirect hack in createRouteURL 

Joshna907:
### **Changes:**
 `frontend/src/lib/router/Route.tsx:`

Updated the Route interface to include an optional redirect field, which supports both string templates and dynamic functions.

`frontend/src/lib/router/createRouteURL.tsx:`

1. Refactored the URL generation logic to generically handle route redirections.
2. Removed the hardcoded if (routeName === 'settingsCluster') hack that previously bypassed the configuration system.

`frontend/src/components/App/RouteSwitcher.tsx:`

1. Updated RouteComponent to perform automatic redirection when a route with a redirect property is matched.
2. Integrated useParams and generatePath to support dynamic, parameter-aware redirection.

`frontend/src/lib/router/index.tsx:`

1. Updated the settingsCluster route to utilize the new redirect field.
2. Removed the redundant SettingsClusterRedirect component, as its logic is now handled generically by the routing engine.

### **Steps to Test:**

1. Navigate to the frontend directory.
2. Run the frontend tests to ensure no regressions: npm run frontend:test
3. Start the application: npm start
4. Click on Cluster Settings in the sidebar. Verify it correctly redirects to /settings/cluster?c=CLUSTER_NAME.
5. Manually enter /c/CLUSTER_NAME/settings in the address bar and verify it successfully redirects to the correct settings page.

### **Validation Results:**

1. Unit Tests: Passed.
2. Build/TSC: npm run frontend:tsc successful.
3. Lint: npm run frontend:lint passed with no new errors.
4. Manual Verification: Verified correct redirection behavior in the browser.

PR #5679, opened 2026-05-15
frontend: router: Handle missing route params gracefully in createRouteURL

### LINKED ISSUE #4863: Plugin routes with :namespace param cause dashboard crash on initialization
## Bug

When a plugin registers detail routes containing `:namespace/:name` params,
the Headlamp dashboard crashes on load with:

  TypeError: Expected "namespace" to be defined

All plugin-specific sidebar views work fine — only the main dashboard is affected.

## Root Cause

Headlamp's internal URL builder (path-to-regexp) tries to construct URLs for 
all registered routes during dashboard initialization. When a plugin route 
contains `:namespace` but the dashboard renders without concrete param values, 
path-to-regexp throws the TypeError and crashes the dashboard view.

## Environment

Headlamp v0.40.1, discovered while using the headlamp-longhorn plugin 
(giantswarm/headlamp-longhorn) on k3s with Longhorn v1.5.3.

## Workaround

For plugins where all resources live in a fixed namespace, removing 
`:namespace` from detail route paths and hardcoding the namespace in 
`useGet()`/`useList()` calls resolves the crash.

The fix for the headlamp-longhorn plugin is available here:
https://github.com/arx666x/headlamp-longhorn/tree/fix/dashboard-crash-namespace-route-params

## Question

Is this expected behavior? Should plugin developers avoid `:namespace` in 
route params, or should Headlamp handle this more gracefully during 
dashboard initialization?

WasThatRudy:
## Summary

`createRouteURL` calls `generatePath(url, fullParams)` from `react-router` without guarding against missing required params. When a plugin registers a detail route with `:namespace` / `:name` placeholders and the dashboard iterates registered routes during initialization without concrete param values, `generatePath` throws `TypeError: Expected \"namespace\" to be defined` and the entire dashboard view crashes.

This PR wraps the `generatePath` call in a try/catch and returns an empty string on failure, matching the existing `!route` and missing-cluster early-return behaviour higher up in the same function. A `console.warn` keeps the diagnostic visible to plugin developers.

## Related Issue

Fixes #4863.

## Changes

- `frontend/src/lib/router/createRouteURL.tsx`: wrapped `generatePath(url, fullParams)` in a try/catch. On failure, logs a warning with the route name + path + error and returns `''`.
- `frontend/src/lib/router/createRouteURL.test.ts`: new suite covering the missing-name and unknown-route cases, the happy path for parameterized and non-parameterized routes, and the new missing-required-params behaviour. Five tests, all passing.

### Why this approach

The minimum-disruption fix. The function already has two existing fallbacks that return early without crashing (`!route` returns `''` at line 76; missing cluster returns `'/'` at line 83). Adding a third for the missing-param case is symmetric with that behaviour and avoids any change to the success path. Plugin developers still see the warning so they can fix their route registration if they want a proper URL, but their plugin won't take down the dashboard.

A more ambitious fix would expose a plugin-side API for routing owner references / unknown-namespace routes (the original reporter raised this as a follow-up question). That's worth a separate design discussion and is out of scope for this defensive fix.

## Steps to Test

2. `npm run lint`: passes.
3. `npm run tsc`: passes.
4. `npx vitest run src/lib/router/createRouteURL.test.ts`: all 5 tests pass.
5. Manual repro: in a plugin, call `registerRoute({ path: '/myplugin/:namespace/:name', ... })`, build the plugin, install it, and load the dashboard. Before this PR the dashboard crashes on initialization with the `TypeError`; after, the dashboard renders normally and a `console.warn` appears in DevTools.

N/A. The bug is a crash; the fix prevents it.

## Notes for the Reviewer

- Two files: a six-line production change and a new ~70-line test file.
- A prior PR (#2383) attempted a similar graceful fix but was auto-closed by the K8s stale-bot in August 2025 after 120 days of inactivity, not on merit. This PR revives the direction with a smaller surface and tests.
- A concurrent PR (#5391) refactors `createRouteURL` to remove the `settingsCluster` hardcode. The two PRs touch the same file but different lines; rebasing either onto the other is straightforward.

### DIFF
--- a/frontend/src/lib/router/createRouteURL.test.ts
+++ b/frontend/src/lib/router/createRouteURL.test.ts
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { createRouteURL } from './createRouteURL';
+
+vi.mock('./getRoute', () => ({
+  getRoute: vi.fn((name?: string) => {
+    if (name === 'pluginParameterizedRoute') {
+      return { path: '/plugin/:namespace/:name', useClusterURL: false };
+    }
+    if (name === 'pluginSimpleRoute') {
+      return { path: '/plugin', useClusterURL: false };
+    }
+    return undefined;
+  }),
+}));
+
+vi.mock('./getRoutePath', () => ({
+  getRoutePath: vi.fn((route: { path: string }) => route.path),
+}));
+
+vi.mock('./getRouteUseClusterURL', () => ({
+  getRouteUseClusterURL: vi.fn((route: { useClusterURL?: boolean }) => !!route.useClusterURL),
+}));
+
+describe('createRouteURL', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('returns an empty string when no route name is provided', () => {
+    expect(createRouteURL()).toBe('');
+  });
+
+  it('returns an empty string when the route name is unknown', () => {
+    expect(createRouteURL('does-not-exist')).toBe('');
+  });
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

- `frontend: router: Handle missing route params gracefully in createRouteURL` — Contains `Fixes #NN`; please remove the issue reference from the commit message.

</details>

<details>

Good examples:

</details>

WasThatRudy:
Updated the commit message per the contributing guidelines @illume — subject is now 63 chars (under the 72 soft limit) and `Fixes #4863.` is removed from the body. The PR description still has `Fixes #4863` in the Related Issue section so GitHub auto-closes the issue on merge. Also added a one-line comment above `generatePath` explaining the try/catch trigger, since stripping all context out felt too terse for the next reader. Force-pushed as `284a8584f`.

illume (approved):
🎉 thanks!

PR #5683, opened 2026-05-16
backend: auth: Bound FuzzSanitizeClusterName input to prevent CI timeout

### LINKED ISSUE #5682: backend: auth: FuzzSanitizeClusterName times out in CI due to unbounded input size
## Describe the bug
`FuzzSanitizeClusterName` in `backend/pkg/auth/cookies_fuzz_test.go` intermittently fails in CI with `context deadline exceeded` after 30s because the fuzzer generates unbounded string inputs with no size cap, causing some iterations to run long enough to exhaust the fuzz budget.

Expected: fuzz test completes within the 30s `-fuzztime` budget.
Actual: `--- FAIL: FuzzSanitizeClusterName (31.00s) context deadline exceeded`

## To Reproduce
1. Run `cd backend && go test -fuzz=FuzzSanitizeClusterName -fuzztime=30s ./pkg/auth/`
2. Observe `context deadline exceeded` after 31s.

## Environment
- Installation type: N/A (backend CI only)

Yes (I will propose a PR).

## Additional Context
The fix is to add a 256 character input size guard inside the fuzz function so pathological inputs are skipped and each iteration stays bounded. Cluster names are not expected to exceed 256 characters in practice. This is not a regression; the fuzz test has always had unbounded input.

Athang69:
## Summary
This PR fixes a CI timeout in the fuzz test `FuzzSanitizeClusterName` by adding a 256 character input size guard to skip pathological inputs that exhaust the 30s fuzz budget.

## Related Issue
Fixes #5682

## Changes
- Added a 256 character input size guard in `FuzzSanitizeClusterName` inside `backend/pkg/auth/cookies_fuzz_test.go` to skip unbounded inputs that cause CI timeouts.

## Steps to Test
1. Run `cd backend && go test -fuzz=FuzzSanitizeClusterName -fuzztime=30s ./pkg/auth/`
2. Confirm the test completes with PASS and no `context deadline exceeded`.
3. Verified locally: 395,171 executions in 30s, PASS.

Screenshot from 2026-05-16 10-47-57

## Notes for the Reviewer
- Cluster names are not expected to exceed 256 characters in practice so this guard does not reduce meaningful fuzz coverage.
- All existing invariants in the fuzz function are preserved.

### DIFF
--- a/backend/pkg/auth/cookies_fuzz_test.go
+++ b/backend/pkg/auth/cookies_fuzz_test.go
@@ -46,6 +46,10 @@ func FuzzSanitizeClusterName(f *testing.F) {
 	validCharsRegex := regexp.MustCompile(`^[a-zA-Z0-9\-_]*$`)
 
 	f.Fuzz(func(t *testing.T, input string) {
+		if len(input) > 256 {
+			t.Skip()
+		}
+
 		result := auth.SanitizeClusterName(input)
 
 		// Invariant 1: Result should never be longer than 50 characters


illume (approved):
🎉 thanks!

PR #5685, opened 2026-05-16
backend: Require kubeconfigs in parseKubeConfig

### LINKED ISSUE #5658: backend: /parseKubeConfig returns 200 when kubeconfigs is missing
### What happened?

The backend `POST /parseKubeConfig` endpoint accepts a JSON body that omits the required `kubeconfigs` field and returns a successful response with no parsed clusters. This makes malformed client requests look valid, so callers can silently drop stateless/dynamic cluster configuration instead of receiving a validation error.

The backend request type is:

```go
type KubeconfigRequest struct {
    Kubeconfigs []string `json:"kubeconfigs"`
}
```

`parseKubeConfig` decodes into that struct, then immediately calls `parseClusterFromKubeConfig(kubeconfigReq.Kubeconfigs)`. When the field is absent, `Kubeconfigs` is nil, the loop in `parseClusterFromKubeConfig` does not run, `setupErrors` remains empty, and the handler writes HTTP 200 with an empty/null clusters payload.

This is also easy to trigger accidentally because some frontend code paths pass a singular `kubeconfig` property while the backend only validates/reads plural `kubeconfigs`.

### What did you expect to happen?

`POST /parseKubeConfig` should reject requests that do not include at least one kubeconfig, for example with HTTP 400 and a clear validation error such as `kubeconfigs is required`.

### How can we reproduce it?

1. Start Headlamp backend with dynamic clusters enabled.
2. Send a request that uses `kubeconfig` instead of `kubeconfigs`, or omits `kubeconfigs` entirely:

```bash
curl -i -X POST http://localhost:4466/parseKubeConfig \
  -H "Content-Type: application/json" \
  -d "{\"kubeconfig\":\"bad-or-empty-value\"}"
```

3. Observe that the backend returns a successful response instead of a 400 validation error.

### Code pointers

- `backend/cmd/cluster.go`: `KubeconfigRequest` only defines `kubeconfigs`.
- `backend/cmd/stateless.go`: `parseKubeConfig` does not validate that `kubeconfigReq.Kubeconfigs` is non-empty before parsing and encoding a success response.
- `backend/cmd/headlamp.go`: `parseClusterFromKubeConfig` returns no error for an empty input slice.

### Duplicate check

I searched existing issues for:

- `parseKubeConfig kubeconfig`
- `kubeconfig singular plural parseKubeConfig`
- `parseKubeConfig empty clusters`
- `"/parseKubeConfig" "empty"`
- `"kubeconfigs" "kubeconfig" stateless`

The related existing issue I found is #5598, but that covers malformed JSON continuing after an error. This report is about syntactically valid JSON with the required `kubeconfigs` field missing or empty being treated as success.

### Environment

Current main branch/local checkout. Backend code path only; no cluster-specific behavior required to reproduce the validation gap.

harrshita123:
## Summary

  This PR fixes backend validation for POST /parseKubeConfig. The endpoint now returns 400 Bad Request when the request body does not include at least one kubeconfigs entry.

  ## Related Issue

  Fixes #5658

  ## Changes

  - Added validation for missing, empty, or null kubeconfigs in parseKubeConfig.
  - Added regression tests for malformed payloads that previously returned success.
  - Kept existing behavior unchanged for valid kubeconfig requests and malformed JSON.

  ## Steps to Test

  1. Run cd backend && go test ./cmd.
  2. Run npm run backend:lint.
  3. Send a request without kubeconfigs:

     curl -i -X POST http://localhost:4466/parseKubeConfig \
       -H "Content-Type: application/json" \
       -d '{"kubeconfig":"bad-or-empty-value"}'

  4. Verify the endpoint returns 400 Bad Request with kubeconfigs is required.

  Not applicable. This is a backend validation change.

  ## Notes for the Reviewer

  This is intentionally limited to validating the request body before kubeconfig parsing. It does not change the parsing behavior for valid kubeconfigs payloads.

### DIFF
--- a/backend/cmd/stateless.go
+++ b/backend/cmd/stateless.go
@@ -170,6 +170,11 @@ func (c *HeadlampConfig) parseKubeConfig(w http.ResponseWriter, r *http.Request)
 	}
 
 	kubeconfigs := kubeconfigReq.Kubeconfigs
+	if len(kubeconfigs) == 0 {
+		http.Error(w, "kubeconfigs is required", http.StatusBadRequest)
+
+		return
+	}
 
 	contexts, setupErrors := parseClusterFromKubeConfig(kubeconfigs)
 	if len(setupErrors) > 0 {
--- a/backend/cmd/stateless_test.go
+++ b/backend/cmd/stateless_test.go
@@ -150,6 +150,62 @@ func TestParseKubeConfigInvalidJSONReturnsBadRequest(t *testing.T) {
 	assert.NotContains(t, resp.Body.String(), "clusters")
 }
 
+func TestParseKubeConfigRequiresKubeconfigs(t *testing.T) {
+	tests := []struct {
+		name string
+		body map[string]interface{}
+	}{
+		{
+			name: "missing kubeconfigs",
+			body: map[string]interface{}{},
+		},
+		{
+			name: "singular kubeconfig",
+			body: map[string]interface{}{
+				"kubeconfig": "bad-or-empty-value",
+			},
+		},
+		{
+			name: "empty kubeconfigs",
+			body: map[string]interface{}{
+				"kubeconfigs": []string{},
+			},
+		},
+		{
+			name: "null kubeconfigs",
+			body: map[string]interface{}{
+				"kubeconfigs": nil,
+			},
+		},
+	}
+
+	for _, tc := range tests {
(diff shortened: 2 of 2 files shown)


harrshita123:
@illume 
I resolved all the copilot comments .

illume (approved):
🎉 thanks!

Issue #5691, opened 2026-05-16
refactor: rename confusing stopOrDelete parameter in portForward API


shubhtrek:
Description

The stopOrDelete parameter in frontend/src/lib/k8s/api/v1/portForward.ts is confusing and reduces code readability.

A TODO comment already suggests renaming it to something clearer like justStop.

Proposed Change
Rename stopOrDelete to justStop (or similar)
Update all related call sites
Ensure functionality remains unchanged
Benefits
Improves readability
Makes port forwarding logic easier to understand
Better developer experience

PR #5692, opened 2026-05-16
refactor: rename stopOrDelete parameter for better readability

shubhtrek:
This PR improves readability in the port forwarding API by renaming the confusing stopOrDelete parameter to justStop.

Changes Made
Renamed stopOrDelete → justStop
Updated all related call sites and references
Kept existing functionality unchanged

Testing
Verified the project builds successfully
Confirmed port forwarding behavior works as expected

#5691

### DIFF
--- a/frontend/src/components/common/Resource/PortForward.tsx
+++ b/frontend/src/components/common/Resource/PortForward.tsx
@@ -315,7 +315,7 @@ function PortForwardContent(props: PortForwardProps) {
       return;
     }
     setLoading(true);
-    stopOrDeletePortForward(cluster, portForward.id, true)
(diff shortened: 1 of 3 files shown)


illume (changes requested):
Thanks for this.

How about adding a "stop" and a "remove" parameter? Combining the two things is quite confusing.

Probably the name change should be passed all the way through to the backend service.

skoeva (changes requested):
per @illume's suggestion, the better fix is to split into `stopPortForward()` and `deletePortForward()` end-to-end - every call site is already hardcoded to one or the other. want to rework along those lines? otherwise I'll close and leave the `@todo` for someone else

PR #5712, opened 2026-05-16
fix: Replace clearInterval with clearTimeout in KubeObjectNode hover-expand effect

### LINKED ISSUE #5711: fix: Replace `clearInterval` with `clearTimeout` in `KubeObjectNode` hover-expand effect
## Bug Summary

In `frontend/src/components/resourceMap/nodes/KubeObjectNode.tsx`, the `useEffect` hook that manages the hover-expand delay for resource map nodes uses `setTimeout` to delay expansion but incorrectly calls `clearInterval(id)` in the cleanup function instead of `clearTimeout(id)`.

## Current Code

```tsx
const id = setTimeout(() => setIsExpanded(true), EXPAND_DELAY);
return () => clearInterval(id); // wrong cleanup
```

## Fixed Code

```tsx
const id = setTimeout(() => setIsExpanded(true), EXPAND_DELAY);
return () => clearTimeout(id); // correct cleanup
```

## Impact

Because `clearInterval` is called on a `setTimeout` ID, the timeout is never actually cancelled when the component unmounts or the dependency changes. This causes:

- Timer is never cleared when user un-hovers the node quickly
- Node may expand even after the mouse has moved away
- Memory/timer leak over repeated hover interactions

## Steps to Reproduce

1. Open the Resource Map view in Headlamp
2. Hover over a node briefly, then move the mouse away before the expand delay completes
3. The node expands anyway, even though the mouse is no longer over it

## Expected Behavior

When the user moves the mouse away before `EXPAND_DELAY` completes, the timeout should be cancelled and the node should not expand.

YadavAkhileshh:
Fixes the cleanup function in the hover-expand useEffect in KubeObjectNode.tsx. The timeout was created with setTimeout but cleaned up with clearInterval, which does nothing on a timeout ID. This meant the expand would still fire even after the user moved their mouse away.

Changed `clearInterval(id)` to `clearTimeout(id)`

Closes #5711

### DIFF
--- a/frontend/src/components/resourceMap/nodes/KubeObjectNode.tsx
+++ b/frontend/src/components/resourceMap/nodes/KubeObjectNode.tsx
@@ -182,7 +182,7 @@ export const KubeObjectNodeComponent = memo(({ id }: NodeProps) => {
     }
 
     const id = setTimeout(() => setIsExpanded(true), EXPAND_DELAY);
-    return () => clearInterval(id);
+    return () => clearTimeout(id);
   }, [isHovered]);
 
   const icon = kubeObject ? (


illume (approved):
🎉 thanks!

PR #5713, opened 2026-05-17
frontend: units: Handle decimal values in parseRam and parseDiskSpace

### LINKED ISSUE #4875: Resources for pods with defined memory are reporting incorrectly (Mi/Gi)
## Describe the bug
Dashboard for workload pods is getting mixed up between Gi and Mi when reporting memory.

## To Reproduce
Steps to reproduce the bug:
1. Go to Workloads → Pods
2. Casually scroll thru pods and notice that they are reporting the digits correctly, but the units incorrectly. (289.9 Gi when it should be Mi) --- Please see screenshots in Additional Context.

 - Headlamp Version: 0.40.0
 - Other: N/A

No.

## Additional Context

Image

Image

Not sure if this is a regression, this is my first install of Headlamp.

WasThatRudy:
## Summary

`parseUnitsOfBytes` (used by `parseRam` and `parseDiskSpace` in `frontend/src/lib/units.ts`) silently mangles any quantity expressed with a decimal portion. The regex matches only `\d+` and runs the captured group through `parseInt`, so:

- `parseRam('289.9Mi')` returns `289` (the integer, interpreted as raw bytes — the `.9Mi` suffix is dropped because the regex never reaches the unit class after the dot)
- `parseRam('1.5Gi')` returns `1`
- `parseRam('0.5Gi')` returns `0`

Pod memory requests and limits expressed with decimal binary units (common in Helm charts, HPA outputs, kubectl-top metrics) consequently render with the wrong magnitude and unit in workload tables and tooltips.

## Related Issue

Fixes #4875.

## Changes

- `frontend/src/lib/units.ts`: widened the digit group from `(\d+)` to `(\d+(?:\.\d+)?)` and switched `parseInt(..., 10)` to `parseFloat(...)`. Integer behaviour is unchanged; decimals are now preserved and the unit suffix is matched correctly.
- `frontend/src/lib/units.test.ts`: added two cases covering decimal binary units (`1.5Ki`, `1.5Mi`, `0.5Gi`, `289.9Mi`) and decimal decimal units (`1.5K`, `2.5M`, `0.5G`). All 27 tests pass.

## Steps to Test

2. `npm run lint`: passes.
3. `npm run tsc`: passes.
4. `npx vitest run src/lib/units.test.ts`: all 27 tests pass (5 new assertions for decimal inputs).
5. Manual repro: deploy a Pod with `resources.requests.memory: 0.5Gi` (or any decimal binary value). Before this PR, the Memory column on the Pods list renders that container's request as `0 Bi`. After, it renders as `512 Mi`.

N/A. The fix is in the parsing layer; visible effect is correct numbers/units in existing tables.

## Notes for the Reviewer

- Two-file diff, +15 / -2 lines.
- The original reporter (#4875) had screenshots showing "289.9 Gi when it should be Mi". I couldn't access the images, but tracing the parser end-to-end surfaced this decimal-handling bug which manifests at every site that calls `parseRam`. If the reporter's specific repro turns out to be a different code path, happy to file that separately as well.

### DIFF
--- a/frontend/src/lib/units.test.ts
+++ b/frontend/src/lib/units.test.ts
@@ -40,6 +40,19 @@ describe('parseRam', () => {
     expect(parseRam('1e6')).toBe(1000000);
   });
 
+  it('should parse decimal values with binary units', () => {
+    expect(parseRam('1.5Ki')).toBe(1.5 * 1024);
+    expect(parseRam('1.5Mi')).toBe(1.5 * 1024 * 1024);
+    expect(parseRam('0.5Gi')).toBe(0.5 * 1024 * 1024 * 1024);
+    expect(parseRam('289.9Mi')).toBeCloseTo(289.9 * 1024 * 1024, 5);
+  });
+
+  it('should parse decimal values with decimal units', () => {
+    expect(parseRam('1.5K')).toBe(1500);
+    expect(parseRam('2.5M')).toBe(2500000);
+    expect(parseRam('0.5G')).toBe(500000000);
+  });
+
   it('should scale binary units by powers of 1024', () => {
     fc.assert(
       fc.property(fc.integer({ min: 1, max: 100 }), num => {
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

PR #5724, opened 2026-05-17
backend: oidc-callback: Stop logging unrelated outer-scope error on missing state

### LINKED ISSUE #4839: Insecure Error Logging
The OIDC callback handler in backend/cmd/headlamp.go  logs an incorrect and unrelated error variable when the OIDC state query parameter is missing.
 **Location**: `backend/cmd/headlamp.go`
Specifically, at line 773, the code calls logger.Log(..., err, ...) where err refers to a variable from the outer function scope (specifically, the result of the kubeconfig loading that happens when the server starts). This leads to misleading logs: if a user hits the callback without a state, the log may show a successful "nil" error or, worse, a completely unrelated error from the server's initialization phase.

govindup63:
## Summary

The `/oidc-callback` handler in `backend/cmd/headlamp.go` logged a stray `err` variable when the OIDC `state` query parameter was missing. The variable was not declared in the handler closure: it resolved to the outer-scope `err` returned by the kubeconfig load that runs once at server startup inside `createHeadlampHandler`.

The practical effect is that every malformed callback request surfaced whatever happened at boot, attached to an unrelated 400 response. For example, a typical in-cluster pod without a local kubeconfig logs this on every empty-state callback:

```
error loading kubeconfig files: error reading kubeconfig file:
open /home/headlamp/.config/Headlamp/kubeconfigs/config: no such file or directory
```

That is the exact misleading line reported in the issue.

## Related Issue

Fixes https://github.com/kubernetes-sigs/headlamp/issues/4839

## Changes

- `backend/cmd/headlamp.go`: pass `nil` to `logger.Log` on the missing-state branch (instead of the captured outer-scope `err`). Added a short comment explaining the scoping trap so the capture is not reintroduced.
- `backend/cmd/headlamp_test.go`: new regression test `TestOidcCallbackEmptyStateDoesNotLogStaleError`. Hooks the logger via `logger.SetLogFunc`, runs `createHeadlampHandler` with a deliberately broken `KubeConfigPath` so the kubeconfig load returns a known error into the captured outer-scope `err`, fires `GET /oidc-callback` with no `state`, then asserts:
  - The response is `400 Bad Request` (unchanged behavior).
  - The captured log entry for `invalid request state is empty` carries a `nil` error value (not the stale kubeconfig load error).

## Steps to Test

1. `cd backend && go test ./cmd/ -run TestOidcCallbackEmptyStateDoesNotLogStaleError -v`. Passes.
2. Reverting only `headlamp.go` (keeping the test) reproduces the bug: the test fails with `Expected nil, but got: &errors.errorString{s:"error loading kubeconfig files: error reading kubeconfig file: open .../kubeconfigs/config: no such file or directory"}`. That is exactly the misleading log the reporter described.
3. `cd backend && go test ./cmd/... -race`. All existing `cmd/` tests still pass.
4. `cd backend && go build ./... && go vet ./cmd/...`. Clean.

## Notes for the Reviewer

- This is a minimal correctness fix. The behavior of the endpoint (`400 Bad Request` on missing `state`) is unchanged. Only the log line argument changes.
- There is a sibling logger misuse at the type-assertion branch a few lines below (`logger.Log(... , nil, err, ...)` when the oauth2 token has no expected token field). At that point the closure's local `err` is always `nil`, so the call is misleading-but-harmless rather than a real outer-scope leak. I left it untouched to keep this PR scoped to the issue. Happy to clean it up in a follow-up.
- The regression test uses `logger.SetLogFunc`, which the logger package already exposes for tests (see `backend/pkg/logger/logger_test.go`). No new test helpers were added.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -846,7 +846,9 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		state := r.URL.Query().Get("state")
 
 		if state == "" {
-			logger.Log(logger.LevelError, nil, err, "invalid request state is empty")
+			// nil err: the closure-local err is declared below, so any err
+			// here would leak from createHeadlampHandler's outer scope.
+			logger.Log(logger.LevelError, nil, nil, "invalid request state is empty")
 			http.Error(w, "invalid request state is empty", http.StatusBadRequest)
 
 			return
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -843,12 +843,15 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		config.handleNodeDrainStatus).Methods("GET").Queries("cluster", "{cluster}", "nodeName", "{node}")
 
 	r.HandleFunc("/oidc-callback", func(w http.ResponseWriter, r *http.Request) {
+		// Shadow createHeadlampHandler's outer-scope err so any log call in
+		// this handler is guaranteed to reference this closure's err and
+		// never the startup kubeconfig-load error. See issue #4839.
+		var err error
+
 		state := r.URL.Query().Get("state")
 
 		if state == "" {
-			// nil err: the closure-local err is declared below, so any err
-			// here would leak from createHeadlampHandler's outer scope.
-			logger.Log(logger.LevelError, nil, nil, "invalid request state is empty")
+			logger.Log(logger.LevelError, nil, err, "invalid request state is empty")
 			http.Error(w, "invalid request state is empty", http.StatusBadRequest)
 
 			return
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for working on this.


govindup63:
Thanks @illume, marked them as resolved.

govindup63:
golangci-lint was failing on the new test (funlen + wsl_v5). Pushed a fix in `b7d6729bb`.

govindup63:
Thanks @copilot, addressed in `b829fbefe` (use `t.Setenv` to pin the static-plugin dir, dropped the no-op field assignment). Marked resolved.

govindup63:
Thanks @copilot, applied in `e0e34ef78` — hoisted `var err error` to the top of the closure so the shadow is enforced by the language. Marked resolved.

backend/cmd/headlamp.go
illume: I guess it makes sense to not log err here?

govindup63:
Right, fixed in `47bc03b6e` — passing `nil` directly. Marked resolved.

illume (approved):
🎉 thanks!

PR #5757, opened 2026-05-19
resourceMap: Hide Gateway API filter when empty

### LINKED ISSUE #5770: hide from map when no resources are installed
On the Map view, the "Gateway (beta)" filter group is always present in the filter list. On clusters that don't have the Gateway API CRDs installed, clicking it triggers list requests to `gateway.networking.k8s.io/v1` and `v1beta1`, which fail (404), and the Map gets stuck in an infinite loading state.

related: #2738

skoeva:
This change hides the "Gateway (beta)" group from the map's filter list when the cluster doesn't serve the Gateway API, preventing the infinite-loading state caused by clicking a filter whose CRDs aren't installed.

Fixes: #5770 

### Changes
- In `useGetAllSources()`, check via `apiDiscovery` whether the `gateway.networking.k8s.io` API group is served by any selected cluster, and gate the `gateway-beta` group on that
- Drop the now-unused MSW handlers for the `gateways` list endpoints from `GraphView.stories.tsx`

### Testing
- [x] Verified the "Gateway (beta)" filter is hidden on a cluster without Gateway CRDs
- [x] Storybook tests pass and no TS errors added

image

### DIFF
--- a/frontend/src/components/resourceMap/sources/definitions/sources.tsx
+++ b/frontend/src/components/resourceMap/sources/definitions/sources.tsx
@@ -15,8 +15,10 @@
  */
 
 import { Icon } from '@iconify/react';
+import { useQuery } from '@tanstack/react-query';
 import React, { useMemo } from 'react';
-import { useCluster } from '../../../../lib/k8s';
+import { useCluster, useSelectedClusters } from '../../../../lib/k8s';
+import { apiDiscovery } from '../../../../lib/k8s/api/v2/apiDiscovery';
 import BackendTLSPolicy from '../../../../lib/k8s/backendTLSPolicy';
 import BackendTrafficPolicy from '../../../../lib/k8s/backendTrafficPolicy';
 import ConfigMap from '../../../../lib/k8s/configMap';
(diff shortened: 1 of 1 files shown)


kahirokunn (approved):
LGTM

frontend/src/components/resourceMap/sources/definitions/sources.tsx
```diff
@@ -127,10 +127,16 @@ const generateCRSources = (crds: CRD[], vpaEnabled: boolean): GraphSource[] => {
...
 export function useGetAllSources(): GraphSource[] {
-  const { items: CustomResourceDefinition } = CRD.useList({ namespace: useNamespaces() });
+  const namespaces = useNamespaces();
+  const { items: CustomResourceDefinition } = CRD.useList({ namespace: namespaces });
   const cluster = useCluster();
   const [vpaEnabled, setVpaEnabled] = React.useState(false);
 
+  // Hide the Gateway (beta) group when there are no Gateway instances
+  // (also covers the CRDs-not-installed case, where the list stays null).
+  const [gatewayItems] = Gateway.useList({ namespace: namespaces });
```
sniok: when there are 0 gateways but some httproutes or gatewayclasses it will still hide all of the gateway resources
I think a better way to check if gateway resources are available is to use `apiDiscovery` 

example usage:
https://github.com/kubernetes-sigs/headlamp/blob/432da0152a0573ec1f87c5b86a09f57c3c05d6d8/frontend/src/components/advancedSearch/AdvancedSearch.tsx#L70

skoeva: done

sniok (approved):

sniok (approved):
/lgtm

PR #5761, opened 2026-05-19
frontend: Update stale snapshots

skoeva:
These changes refresh outdated Storybook snapshots so the frontend tests pass on main.


illume (approved):
Thanks!

PR #5803, opened 2026-05-21
frontend: SimpleTable: Fix stale defaultRowsPerPage after settings change

ayushmaan-16:
## Description
In `SimpleTable.tsx`, the `defaultRowsPerPage` value was memoized using `useMemo` with an empty dependency array `[]`. Due to this, it was only capturing the value when the table loaded for the first time. 

If the user went to the Settings page and changed the default rows per page options, the table still showed the old default value instead of updating to the new one. The issue only got fixed after doing a full page refresh/remount.

To fix this, we have removed `useMemo` completely and we are now calling `getTablesRowsPerPage` directly. This also allows us to remove the `eslint-disable-next-line` comment which was hiding the lint warning.

## Changes Made
* **`frontend/src/components/common/SimpleTable.tsx`**:
  * Removed `useMemo` and the empty dependency array from `defaultRowsPerPage`.
  * Removed the `eslint-disable-next-line react-hooks/exhaustive-deps` comment.
  * Aligned the logic with `Table.tsx` by calling `getTablesRowsPerPage` directly.

### DIFF
--- a/frontend/src/components/common/SimpleTable.tsx
+++ b/frontend/src/components/common/SimpleTable.tsx
@@ -175,8 +175,7 @@ export default function SimpleTable(props: SimpleTableProps) {
   const [displayData, setDisplayData] = React.useState(data);
   const storeRowsPerPageOptions = useSettings('tableRowsPerPageOptions');
   const rowsPerPageOptions = props.rowsPerPage || storeRowsPerPageOptions;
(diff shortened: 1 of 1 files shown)


sniok:
could you provide reproduction steps for the original issue?

illume (changes requested):
Thanks for this PR.

PR #5805, opened 2026-05-21
frontend: SimpleTable: Fix synchronous setPage(0) during render

ayushmaan-16:
## Description
In `SimpleTable.tsx`, we were previously calling `setPage(0)` directly during the render cycle if the `filteredData` became empty or too small for the current page size. This is a React anti-pattern because calling a state setter during the render phase forces React to synchronously drop the current render and start over.

The original code likely did this because there is an early return (`if (displayData === null) { ... }`) higher up, which prevented using a `useEffect` hook below it (Hooks cannot be placed after early returns). 

To fix this properly and align with React best practices, we have hoisted the `filteredData` calculation above the early return using `useMemo`. This allows us to cleanly handle the page boundary correction inside a `useEffect` hook, completely removing the synchronous state update from the render body.

## Changes Made
* **`frontend/src/components/common/SimpleTable.tsx`**:
  * Moved the `filteredData` calculation above the early return and wrapped it in `useMemo`.
  * Moved the `setPage(0)` boundary correction logic inside a proper `useEffect` hook.
  * Cleaned up the `else` block to remove the inline calculation.

### DIFF
--- a/frontend/src/components/common/SimpleTable.tsx
+++ b/frontend/src/components/common/SimpleTable.tsx
@@ -328,11 +328,9 @@ export default function SimpleTable(props: SimpleTableProps) {
     return filterFunction ? displayData.filter(filterFunction) : displayData;
   }, [displayData, filterFunction]);
 
-  React.useEffect(() => {
-    if (filteredData && filteredData.length <= page * rowsPerPage && page !== 0) {
-      setPage(0);
-    }
-  }, [filteredData, page, rowsPerPage, setPage]);
+  if (filteredData && filteredData.length <= page * rowsPerPage && page !== 0) {
+    setPage(0);
+  }
 
   let content;
   if (displayData === null) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/SimpleTable.tsx
```diff
@@ -331,23 +342,11 @@ export default function SimpleTable(props: SimpleTableProps) {
...
-      filteredData = displayData?.filter(filterFunction);
-    }
-
     function getPagedRows() {
       const startIndex = page * rowsPerPage;
       return filteredData!.slice(startIndex, startIndex + rowsPerPage);
     }
 
-    if (
-      (filteredData?.length === 0 || (filteredData?.length ?? 0) < page * rowsPerPage) &&
```
sniok, line 345: setting state in render is not an antipattern and in this case it's better than useEffect

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

sniok (changes requested):

sniok:
can you explain a bit what is the actual issue you're trying to fix and how to reproduce it? does this issue affect users?

skoeva (commented):
calling a state setter during render of the same component is a [supported react pattern](https://react.dev/reference/react/useState#storing-information-from-previous-renders) when it's conditional and derives state from props. I don't think there's any demonstrated user impact here, so going to close this one. thanks for the contribution

PR #5824, opened 2026-05-22
frontend: ObjectEventList: Guard optional event source in From column

### LINKED ISSUE #5823: Events table crashes when event has no source field
On resource detail pages, the Events table reads `item.source.component` without checking if source exists. Kubernetes often omits source (especially with newer event APIs), which causes a TypeError and can break the Events section.

anurag-p6:
## Summary

Fix a runtime error in the resource **Events** table when Kubernetes returns events without a `source` field. The **From** column now safely reads `source.component` using optional chaining and falls back to an empty string.

## Related Issue

Fixes #5823   

## Changes

- In the Kubernetes API, `Event.source` is optional (`EventSource` with optional `component` and `host`).
- Headlamp’s `Event.source` getter returns `undefined` when the field is absent (`getValue('source')`).
- Accessing `.component` on `undefined` throws and can break rendering of the events table for affected resources.
- Events from newer pipelines (e.g. `events.k8s.io`) may omit `source` entirely; this is expected cluster behavior, not bad data.
- Aligns with existing defensive access elsewhere (e.g. `event.source?.host` in cluster Overview).

## Steps to Test

1. Run the frontend: `npm run frontend:start`
2. Open any resource details view that shows the **Events** section (e.g. a Pod).
3. Confirm the Events table renders when events have no `source` (or only `source.host`).
4. Confirm events with `source.component` (e.g. `kubelet`, `default-scheduler`) still show the correct **From** value.
5. (Optional) `npm run frontend:storybook` → **common/ObjectEventList** stories still render.

## Notes for the Reviewer

- crash fix only, no mapping of `reportingController` into **From** (can be a follow-up).
- Same empty-cell behavior as other optional event fields in this table (`message || ''`).

### DIFF
--- a/frontend/src/components/common/ObjectEventList.tsx
+++ b/frontend/src/components/common/ObjectEventList.tsx
@@ -119,7 +119,7 @@ export default function ObjectEventList(props: ObjectEventListProps) {
           {
             label: t('From'),
             getter: item => {
-              return item.source.component;
+              return item.source?.component ?? '';
             },
           },
           {


skoeva:
closing this in favor of #5864

anurag-p6:
works.

PR #5826, opened 2026-05-22
frontend: Pod: Show resolved environment variables for running containers

### LINKED ISSUE #4798: Feature Request: Display Resolved Environment Variables for Running Pods
---

In the previous Kubernetes Dashboard, it was possible to view the fully resolved environment variables of a running Pod directly from the UI.

In Headlamp, only the environment variable definitions from the Pod spec are shown (e.g., `secretKeyRef`, `configMapKeyRef`). The actual resolved environment variable values inside the running container are not visible.

This makes it difficult to quickly verify whether a Pod is using the latest Secret value, especially in setups like AWS Secrets Manager → ExternalSecret → Kubernetes Secret → Pod.

Currently, users must open the Pod terminal and manually run `printenv` to confirm the actual environment variables, which slows down troubleshooting and incident resolution.

---

Add support in the Pod details view to display the fully resolved Environment Variables of a running container.

For example:

* Add a dedicated tab named “Environment Variables”
* Or provide a button to fetch and display the container’s runtime environment output

This would provide similar functionality to what was available in Kubernetes Dashboard.

---

* DevOps engineers
* SRE teams
* Platform engineers
* Teams using Headlamp as a replacement for Kubernetes Dashboard
* Users running Headlamp in production clusters (e.g., EKS)

---

Not Sure, I Need to check the code 

---

## Additional context

This enhancement would improve observability and reduce the need to exec into containers for simple environment variable validation.

VijayabaskarR-06:
## Summary

closes:  kubernetes-sigs/headlamp/issues/4798.
closes #4798 

Pod details now has an Environment variables action that opens a panel listing the fully resolved env vars of a running container. Today you can only see the spec level definitions like secretKeyRef and configMapKeyRef, and the only way to confirm the actual values a container ended up with is to open a shell and run printenv. This change brings that workflow into the UI, similar to what the older Kubernetes Dashboard offered, so verifying a Secret rotation or a ConfigMap reference takes one click instead of an exec session.

## What is in the change

A new PodEnvironmentVariables component lives next to the rest of the Pod components. It uses the existing pod exec endpoint to run env inside the chosen container, captures stdout over the apiserver channel multiplexing protocol, and waits for the metav1.Status object on channel three to know when the command has finished. When env produces no output, which is the case for distroless and other minimal images that ship without it on PATH, the component falls back to reading the null separated stream at /proc/1/environ. Each entry is parsed by splitting on the first equals sign so values that legally contain equals signs are preserved verbatim.

The panel shows the entries in a sortable, searchable table. Values are masked by default to keep secrets off the screen during demos and screen shares, with a per row eye toggle and a Reveal all switch for when you actually want to see them. There is a copy to clipboard button on each row and a download as .env action that produces a file named pod_container.env. A container selector at the top lists main, init, and ephemeral containers. The refresh button re reads the environment from the live container.

The action button is wired into Details.tsx behind an AuthVisible gate for create exec, matching the existing Terminal action. A new POD_ENVIRONMENT entry has been added to DefaultHeaderAction so plugins can target the button. The panel also auto launches when the URL carries view=env and honors the container query parameter, mirroring how view=logs and view=exec already work. The plugin compatibility snapshot and the offline icon cache have been updated to cover the new enum entry and the variable icon used by the action button.

## Why exec and not something cheaper

Resolved environment variables are runtime state. They are not part of the Pod spec returned by the apiserver and there is no read only endpoint that exposes them. The only authoritative source is the running container itself, which is exactly what the exec subresource gives us. Using exec here is the same trust model as the existing Terminal and Attach actions, so the panel reuses the same RBAC gate.

## Tradeoffs and edge cases

Some images do not ship env or printenv. The /proc/1/environ fallback handles most of these, including the common distroless variants. For images that ship without cat either, the panel surfaces the apiserver Failure message and offers Try again, so the user gets a real reason rather than a silent empty table. There is a fifteen second safety timeout so a slow or hung container cannot leave the panel spinning forever.

Values are masked by default. This is purely UX and does not change the trust boundary, since anyone who can open this panel already has create exec on the pod and can read the same values from a shell.

## Test plan

* npx tsc noEmit passes for the frontend
* npx eslint passes for the changed files
* vitest run for the pod and redux folders passes
* New PodEnvironmentVariables.test.tsx covers the parser for env style output, the parser for /proc/1/environ, the not running guard, the happy path with env, the fallback to /proc/1/environ when env returns nothing, value masking and reveal toggling, and the error path when both exec attempts fail
* Details.test.tsx covers auto launch for view=env, the no relaunch behavior on subsequent renders for the same pod, and relaunch when the container query parameter changes
* Plugin compatibility snapshot updated to include POD_ENVIRONMENT and the offline icon cache updated to include the variable icon
* Manual run against a real cluster: open a Pod, click the Environment variables button, verify the action button is gated by create exec, change containers, search by name and by value, toggle Reveal all, copy a single value, download the .env file, and confirm the fallback path on a distroless image


skoeva (changes requested):
hi, could you run npm run i18n locally to update the translations? that should fix the failing CI test

VijayabaskarR-06:
Hi @skoeva  thanks for the quick review. I ran npm run i18n locally and committed the regenerated translation and glossary JSON files for every locale, then pushed.them because the test job on the latest commit just completeted so the failing CI should be resolved  after that , Waiting for your feedback if there is anything else you would like me to change and Thanks once again!!

Issue #5828, opened 2026-05-22
frontend: Add unit test coverage for untested helpers in src/helpers/


RajPrakash681:
## Problem

Several modules under `frontend/src/helpers/` contain non-trivial logic but
have **no unit tests**. They are imported across many parts of the app, so
silent regressions in them (e.g., a corrupted `localStorage` payload, a
container-fallback bug, an off-by-one in the dedup logic) are hard to spot.

The umbrella under #931 (Stories for components) covers visual / Storybook
gaps; this issue tracks the same gap for the **pure-logic helper modules**.

## Modules affected

| File | LOC | Imported from | Logic worth pinning down |
|------|-----|---------------|--------------------------|
| `frontend/src/helpers/clusterSettings.ts` | 73 | 14 files | localStorage round-trip; empty-name no-op; key-prefix isolation; corrupt-JSON behaviour |
| `frontend/src/helpers/podContainer.ts` | 77 | 2 files | `getAllContainers` (main + init + ephemeral concat); `getDefaultContainer` fallback chain (running main → running init → first main); `resolveContainerName` preferred-name match |
| `frontend/src/helpers/recentClusters.ts` | 45 | 4 files | dedup on re-add, 3-item cap, string/Cluster-object polymorphism, non-array stored payload, corrupt JSON behaviour |

Sibling helper [`tableSettings.ts`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/helpers/tableSettings.ts)
already has tests at [`tableSettings.test.ts`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/helpers/tableSettings.test.ts);
the same pattern (Vitest, `beforeEach(localStorage.clear())`, nested
`describe`-per-function) applies cleanly to the three modules above.

## Proposed approach

One PR per helper (the three modules are unrelated logically — keeping them
separate makes review and bisect cleaner):

1. `frontend: clusterSettings: Add unit tests`
2. `frontend: podContainer: Add unit tests`
3. `frontend: recentClusters: Add unit tests`

Each PR adds **one new `<helper>.test.ts` file**, no changes to production
code, no new translation strings.

## Out of scope (for this issue)

- Fixing the latent JSON-corruption behaviour in `loadClusterSettings` and
  `getRecentClusters` (both throw on malformed payloads rather than
  recovering). The tests will document the current behaviour with
  `expect(...).toThrow(SyntaxError)`; a follow-up issue can decide whether
[...]

PR #5829, opened 2026-05-22
frontend: clusterSettings: Add unit tests

RajPrakash681:
## Summary

This PR adds **unit test coverage** for `frontend/src/helpers/clusterSettings.ts`, which manages cluster-specific configuration in `localStorage` (default namespace, node-shell terminal, pod-debug terminal, appearance). The module is imported by **14 files** across the app but previously had zero test coverage, so silent regressions in `storeClusterSettings` / `loadClusterSettings` would not be caught by CI.

## Related Issue
#5828 

## Changes

- Added `frontend/src/helpers/clusterSettings.test.ts` with 11 tests covering:
  - Round-trip of a full settings object (every nested field — `nodeShellTerminal`, `podDebugTerminal`, `appearance`)
  - Overwriting an existing entry
  - Isolation between cluster names (no aliasing across keys)
  - No-op behavior when `clusterName` is empty (both `store` and `load`)
  - Persistence of an empty settings object
  - Empty-object return when no entry exists or the cluster name is empty
  - Documented current behavior on corrupted localStorage payloads (`SyntaxError` thrown)
- No changes to the source module.
- No new translation strings.

## Steps to Test

1. From the repo root: `cd frontend && npx vitest run src/helpers/clusterSettings.test.ts`
2. Confirm **11 tests pass**.
3. Run the full frontend lint to confirm no regressions: `npm run frontend:lint`

N/A — test-only PR, no UI.

## Notes for the Reviewer

- The follows the existing pattern in [`tableSettings.test.ts`](frontend/src/helpers/tableSettings.test.ts): nested `describe`-per-function, `beforeEach(localStorage.clear())`, copyright header.
- One test (`throws when the stored payload is not valid JSON`) intentionally **documents the current behavior** — `loadClusterSettings` throws on corrupt JSON rather than recovering. A follow-up could harden the helper; this test should be updated alongside such a change.

### DIFF
--- a/frontend/src/helpers/clusterSettings.test.ts
+++ b/frontend/src/helpers/clusterSettings.test.ts
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


RajPrakash681:
@illume can this be merged?

illume (approved):
🎉 thanks!

PR #5831, opened 2026-05-22
frontend: recentClusters: Add unit tests

RajPrakash681:
## Summary

This PR adds **unit test coverage** for `frontend/src/helpers/recentClusters.ts`, which maintains the user's recently-accessed clusters list in `localStorage`. The module is imported by 4 files and contains non-trivial logic (string-or-`Cluster`-object polymorphism, dedup on re-add, hard cap at 3 entries, non-array storage handling) that previously had zero test coverage.

## Related Issue

#5828 
## Changes

- Added `frontend/src/helpers/recentClusters.test.ts` with 11 tests covering:
  - `setRecentCluster`: writing into empty storage, accepting a string vs. a `Cluster` object, ordering (most-recent first), dedup on re-add, 3-item cap dropping the oldest, mixed string/object dedup
  - `getRecentClusters`: empty storage, well-formed stored array, non-array shapes (object → `[]`, JSON `null` → `[]`)
  - Documented current behavior on corrupted payloads (`SyntaxError` thrown)
- No changes to the source module.
- No new translation strings.

## Steps to Test

1. From the repo root: `cd frontend && npx vitest run src/helpers/recentClusters.test.ts`
2. Confirm **11 tests pass**.
3. Run the full frontend lint to confirm no regressions: `npm run frontend:lint`

N/A  test-only PR, no UI.

## Notes for the Reviewer

- A small private `read()` helper asserts on raw localStorage state independently of `getRecentClusters`, so the dedup / cap / ordering tests don't depend on the read path being correct.
- One test (`throws when the stored payload is not valid JSON`) intentionally **documents the current behavior** — `getRecentClusters` throws on corrupt JSON rather than recovering. A follow-up could harden the helper; this test should be updated alongside such a change.

### DIFF
--- a/frontend/src/helpers/recentClusters.test.ts
+++ b/frontend/src/helpers/recentClusters.test.ts
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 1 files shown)


RajPrakash681:
@illume please have a look at this also! i think this is good to go.

illume (approved):
🎉 thanks!

Issue #4304, opened 2025-12-23
Crash Report: Expected "name" to be defined


mpalu:
## Crash Summary
Error loading Cluster
Expected "name" to be defined

## Error Stack
```
TypeError: Expected "name" to be defined
    at file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-D68s4_Sf.js:18:25395
    at pl (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-D68s4_Sf.js:25:6811)
    at as (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-D68s4_Sf.js:1195:35861)
    at kg.getDetailsLink (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-D68s4_Sf.js:888:11966)
    at knn (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-D68s4_Sf.js:1195:237436)
    at xd (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-1viqCYs3.js:38:17652)
    at Nh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-1viqCYs3.js:40:44746)
    at Eh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-1viqCYs3.js:40:40337)
    at Ek (file:///Applications/Headlamp.app/Contents/Resources/
... (truncated)
```

## Headlamp Version
0.39.0

## Git Commit
ec014c531adf1bdbdf4f97343035c5f85df8bca0

## System Information
- User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Headlamp/0.39.0 Chrome/126.0.6478.127 Electron/31.2.0 Safari/537.36
- Platform: MacIntel
- Language: en-US

## Additional Context
This happens when I access the "Clusters" page, only for one out of the 23 clusters I have, named "arn:aws:eks:eu-west-1:XXXXXXXXXXXX:cluster--xxx-xxxxxx-xxxx"

Sagar-6203620715:
Hi @illume,
[...]

Issue #4303, opened 2025-12-22
Add "lint:commit" to check commits follow our guidelines


illume:
There should be a command to check commits follow our [commit guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines).

- [ ] `npm run lint:commit` at top level
  - [ ] `npm run lint` should also check commits
  - [ ] Probably should also check commits when working in the frontend/ folder, not just top level folder.
- [ ] Gather more good, and bad examples to test with. The [commit guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines) has some, but more are probably needed to test this tool properly.
- [ ] documentation for lint:commit is updated in contribution/dev guide
- [ ] the lint:commit runs in CI, in a GitHub workflow
- [ ] precommit/husky check
- [ ] people notice their git commit message needs work and modify it themselves

This will help us all check the commits are ok, but also save review time.

Harshul23:
### Request for assignment on Issue #4303

Hi @illume! I’m a 1st-year B.Tech CSE AI student and have been closely following the Headlamp roadmap for the Spring 2026 LFX Mentorship cycle. I am very interested in improving the Developer Experience (DX) through this lint:commit task.

I have already set up the Headlamp development environment locally and have experience with the Node.js/TypeScript ecosystem. My proposed plan is to:

1. Integrate commitlint with husky to automate these checks at the git-hook level.

2. Ensure the linter covers both the top-level and the frontend/ directory as requested.

3. Add documentation to the CONTRIBUTING.md so other contributors can easily follow the new rules.

I am ready to submit a PR immediately once the approach is approved. May I be assigned to this issue?

dibyanshu-pal-kushwaha:
Problems : 
1) There is no command to check if commits follow the commit guidelines.
2)Linting of Commits only runs at top level (npm run lint:commit)
3)npm run lint does not include commit message validation
4)There is also no pre-commit hook for commit validation.
5)Reliance on the manual correction - No Automation 
6)Updating the documentation for lint:commit in contribution/dev guide

For the very first problem  : 
In the root configuration , we will modify the package.json ,
1. add  lint:commit script: commitlint --from=origin/main
=> It checks every commit on your current branch that has not yet been     merged into origin/main.
2.Update lint script to   "lint": "npm run backend:lint && npm run frontend:lint &&npm run lint:commit".  
It ensures that commits are also checked.
[...]

PR #4403, opened 2026-01-20
fix: prevent crash when cluster name is undefined in getDetailsLink

Sagar-6203620715:
## Summary
This PR fixes a crash that occurs when accessing the Clusters page for clusters with undefined `metadata.name` (e.g., AWS EKS clusters with ARN-style names containing special characters like `arn:aws:eks:eu-west-1:XXXXXXXXXXXX:cluster--xxx-xxxxxx-xxxx`).

## Related Issue
Fixes #4304

## Changes
- Added null check in `KubeObject.getDetailsLink()` before generating route URLs
- Returns empty string with console warning when object name is undefined
- Prevents `TypeError: Expected "name" to be defined` from `react-router`'s `generatePath()`

## Root Cause
The crash occurred because `getDetailsLink()` called `getName()` which returned `undefined` for certain cluster configurations. This undefined value was then passed to `createRouteURL()` → `generatePath()`, which expects all route parameters to be defined, causing the application to crash.

## Steps to Test
1. Add a cluster with an ARN-style name (or any cluster where `metadata.name` is undefined)
2. Navigate to the Clusters page (`/c/:cluster/cluster`)
3. Verify the page loads without crashing
4. Check browser console for warning message about undefined cluster name
5. Confirm other clusters with valid names still generate proper detail links

## Notes for the Reviewer
- The fix is defensive at the `getDetailsLink()` layer rather than modifying `getName()` to preserve its current behavior for other use cases
- Console warning provides debugging context (kind, cluster, namespace) to help identify problematic resources
- Returning empty string prevents malformed URLs while allowing calling code to handle gracefully

PR #5855, opened 2026-05-26
frontend: KubeObject: guard against undefined name in getDetailsLink

### LINKED ISSUE #5791: Crash Report: Expected "name" to be defined
## Crash Summary
Expected "name" to be defined

## Error Stack
```
TypeError: Expected "name" to be defined
    at file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-CE3nuifS.js:18:25395
    at mc (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-CE3nuifS.js:25:6811)
    at ls (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-CE3nuifS.js:1198:36218)
    at Ng.getDetailsLink (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-CE3nuifS.js:888:12002)
    at Ain (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-CE3nuifS.js:1198:238411)
    at xd (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-Dv5SuZR_.js:38:17652)
    at _h (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-Dv5SuZR_.js:40:44746)
    at Lh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-Dv5SuZR_.js:40:40337)
    at Lk (file:///Applications/Headlamp.app/Contents/Resources/
... (truncated)
```

## Headlamp Version
0.41.0

## Git Commit
066c5be9deb14b12bcef3485b531c64a84a2a219

## System Information
- User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Headlamp/0.41.0 Chrome/126.0.6478.127 Electron/31.2.0 Safari/537.36
- Platform: MacIntel
- Language: en-GB

rajeshsingh241:
When a Kubernetes resource has no metadata.name (which can happen
with newer event APIs), getDetailsLink() would pass undefined to
createRouteURL causing a TypeError crash.

Guard against this by returning null from getDetailsLink() when name
is missing, and handle the null case in KubeObjectLink by rendering
a plain span instead of a clickable link.

Fixes #5791

### DIFF
--- a/frontend/src/components/common/Link.tsx
+++ b/frontend/src/components/common/Link.tsx
@@ -15,6 +15,7 @@
  */
 
 import MuiLink from '@mui/material/Link';
+import Typography from '@mui/material/Typography';
 import { useQueryClient } from '@tanstack/react-query';
 import React from 'react';
 import { Link as RouterLink } from 'react-router-dom';
@@ -66,6 +67,16 @@ function KubeObjectLink(props: {
   const { namespace, name } = kubeObject.metadata;
   const { endpoint } = useEndpoints(kubeObject._class().apiEndpoint.apiInfo, kubeObject.cluster);
 
+  const detailsLink = kubeObject.getDetailsLink();
+
+  if (!detailsLink) {
+    return (
+      <Typography component="span" {...otherProps}>
+        {props.children || kubeObject!.getName()}
+      </Typography>
+    );
+  }
+
   return (
     <MuiLink
       onClick={e => {
@@ -75,10 +86,7 @@ function KubeObjectLink(props: {
           namespace,
           name,
         });
-        // prepopulate the query cache with existing object
         client.setQueryData(key, kubeObject);
-        // and invalidate it (mark as stale)
-        // so that the latest version will be downloaded in the background
         client.invalidateQueries({ queryKey: key });
 
         if (onClick) {
@@ -87,7 +95,7 @@ function KubeObjectLink(props: {
         }
       }}
       component={RouterLink}
-      to={kubeObject.getDetailsLink()}
+      to={detailsLink}
       {...otherProps}
     >
       {props.children || kubeObject!.getName()}
--- a/frontend/src/components/common/Resource/RollbackButton.tsx
+++ b/frontend/src/components/common/Resource/RollbackButton.tsx
@@ -102,7 +102,7 @@ export function RollbackButton(props: RollbackButtonProps) {
 
   function handleConfirm(toRevision?: number) {
     const itemName = rollbackResource.metadata.name;
-
+    const detailsLink = rollbackResource.getDetailsLink();
     dispatch(
       clusterAction(() => performRollback(toRevision), {
         startMessage: t('Rolling back {{ itemName }} to previous version…', { itemName }),
@@ -110,8 +110,8 @@ export function RollbackButton(props: RollbackButtonProps) {
         successMessage: t('Rolled back {{ itemName }} to previous version.', { itemName }),
         errorMessage: t('Failed to rollback {{ itemName }}.', { itemName }),
         cancelUrl: location.pathname,
-        startUrl: rollbackResource.getDetailsLink(),
-        errorUrl: rollbackResource.getDetailsLink(),
+        startUrl: detailsLink ?? undefined,
+        errorUrl: detailsLink ?? undefined,
       })
     );
 
--- a/frontend/src/lib/k8s/KubeObject.ts
+++ b/frontend/src/lib/k8s/KubeObject.ts
@@ -181,14 +181,16 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
     return this.jsonData.kind;
   }
 
-  getDetailsLink() {
+  getDetailsLink(): string | null {
+    const name = this.getName();
+    if (!name) {
+      return null;
+    }
     const selectedClusters = getSelectedClusters();
-
     const cluster = formatClusterPathParam(selectedClusters, this.cluster);
-
     const params = {
       namespace: this.getNamespace(),
-      name: this.getName(),
+      name,
       cluster,
     };
     const link = createRouteURL(this.detailsRoute, params);
--- a/frontend/src/components/endpointSlices/__snapshots__/EndpointSliceList.Items.stories.storyshot
+++ b/frontend/src/components/endpointSlices/__snapshots__/EndpointSliceList.Items.stories.storyshot
@@ -711,9 +711,8 @@
               <td
                 class="MuiTableCell-root MuiTableCell-alignLeft MuiTableCell-sizeMedium css-1adoll5-MuiTableCell-root"
               >
-                <a
-                  class="MuiTypography-root MuiTypography-inherit MuiLink-root MuiLink-underlineHover css-2ugbm1-MuiTypography-root-MuiLink-root"
-                  href="/"
+                <span
+                  class="MuiTypography-root MuiTypography-body1 css-1ezega9-MuiTypography-root"
                 />
               </td>
               <td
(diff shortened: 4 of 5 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

- `frontend: KubeObject: guard against undefined name in getDetailsLink` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>


The GitHub CI test job has snapshot failures. Run `cd frontend && npm run test -- -u` to regenerate the snapshots.

<details>

</details>

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -181,13 +181,15 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
   }
 
   getDetailsLink() {
+    const name = this.getName();
+    if (!name) {
+      return null;
+    }
```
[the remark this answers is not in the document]
rajeshsingh241: Good point — added explicit string | null return type annotation to getDetailsLink() so downstream consumers and plugins know to handle the null case safely.

frontend/src/components/common/Link.tsx
```diff
@@ -66,6 +66,12 @@ function KubeObjectLink(props: {
   const { namespace, name } = kubeObject.metadata;
   const { endpoint } = useEndpoints(kubeObject._class().apiEndpoint.apiInfo, kubeObject.cluster);
 
+  const detailsLink = kubeObject.getDetailsLink();
+
+  if (!detailsLink) {
+    return <span {...otherProps}>{props.children || kubeObject!.getName()}</span>;
+  }
```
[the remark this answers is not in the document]
rajeshsingh241: Fixed — replaced the plain <span> with MUI Typography with component='span' so MUI props are consumed correctly without leaking to the DOM.

frontend/src/components/App/__snapshots__/VersionDialog.VersionDialog.stories.storyshot
```diff
@@ -28,46 +28,6 @@
...
-                    class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"
-                  />
-                </button>
-              </div>
-            </div>
-          </div>
-        </h2>
         <div
           class="MuiDialogContent-root css-ypiqx9-MuiDialogContent-root"
         >
```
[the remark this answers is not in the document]
rajeshsingh241: This snapshot was already outdated before this PR. The title element is present in the updated snapshot — regenerated with npm run test -- -u and all 1590 tests pass.

frontend/src/components/common/Resource/RollbackButton.tsx
```diff
@@ -110,8 +110,8 @@ export function RollbackButton(props: RollbackButtonProps) {
         successMessage: t('Rolled back {{ itemName }} to previous version.', { itemName }),
         errorMessage: t('Failed to rollback {{ itemName }}.', { itemName }),
         cancelUrl: location.pathname,
-        startUrl: rollbackResource.getDetailsLink(),
-        errorUrl: rollbackResource.getDetailsLink(),
+        startUrl: rollbackResource.getDetailsLink() ?? undefined,
+        errorUrl: rollbackResource.getDetailsLink() ?? undefined,
       })
```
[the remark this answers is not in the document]
rajeshsingh241: Fixed — computed once as detailsLink and reused for both startUrl and errorUrl.

rajeshsingh241:
@illume All review comments addressed, snapshots regenerated, and all CI checks passing. Please take another look when you get a chance.

frontend/src/lib/k8s/KubeObject.ts
```diff
@@ -181,14 +181,16 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
     return this.jsonData.kind;
   }
 
-  getDetailsLink() {
+  getDetailsLink(): string | null {
+    const name = this.getName();
```
sniok, line 185: every kubernetes resource has to have a name, the root cause of the crash must be somewhere else

rajeshsingh241: Thanks @sniok. I found #4304 which has the exact same crash in getDetailsLink(). I also noticed PR #4403 attempted the same null guard fix but was never merged — you fixed #4304 differently. Could you share what that actual fix was? For #5791 the trigger seems different (newer Event APIs), so I want to understand if this is a new occurrence of the same underlying problem before deciding where to put the fix.

sniok: I haven't fixed #4303, the original reporter closed it after saying that it's not a problem for them anymore.

Your proposed fix is addressing the symptom of the problem, in kubernetes every resource is required to have a name. The root cause must be somewhere else, we can't just add a check for a missing name because it means something went wrong in someplace else and we won't be able to trace it down because this check will make it a silent failure.

To properly fix this issue we first need to be able to reproduce it. Have you been able to reproduce this crash?

rajeshsingh241: @sniok You're right - every Kubernetes resource should have a name. I dug into this more and found where the actual problem is.

The crash isn't from Event objects themselves. It's coming from the Event.involvedObjectInstance getter (around line 180-192 in event.ts).

This getter creates a KubeObject from event.involvedObject, which is an ObjectReference. Turns out all fields in ObjectReference are optional according to the K8s spec, including name: https://github.com/kubernetes-client/go/blob/master/kubernetes/docs/V1ObjectReference.md

So when an Event has an involvedObject without a name, we end up creating a KubeObject with `metadata.name = undefined`. Then when the Cluster Overview page calls `getDetailsLink()` on it (Overview.tsx around line 162-167), it crashes.

I think the fix should be in `Event.involvedObjectInstance` - just don't create the object if there's no name:

```typescript
get involvedObjectInstance(): KubeObject | null {
  if (!this.involvedObject || !this.involvedObject.name) {
    return null;
  }
  // ... rest of existing code
}
```

The nice thing is makeObjectLink() already handles null - it just shows the plain text name instead. So this keeps KubeObject instances valid without needing changes elsewhere.

Does this seem like the right fix to you?

sniok: when involved object in the event doesn't have a name we don't render a link, so I don't think that's the root cause 

image

rajeshsingh241: Thanks @sniok for testing that! I see the plain text rendering in your screenshot. 

Looking at the code more carefully - if `involvedObject.name` is missing but the `kind` is in `ResourceClasses` (like Pod), then `involvedObjectInstance` would create an object with `metadata.name = undefined`, and `makeObjectLink()` would try to render `<Link kubeObject={obj} />` which calls `getDetailsLink()`.

But you're showing it renders as plain text instead. Could you help me understand:

1. Is the "involvedObject" in your test missing a `name` field, or is it an empty string?
2. When you click where the name would be, does anything happen or are there any console errors?
3. Could there be error boundaries or try-catch somewhere that's catching this before it becomes visible?

Also, do you have more context on the original crash reported in #5791? The issue mentions a crash with stack trace pointing to `getDetailsLink()` - I want to make sure I'm fixing the right code path.

sniok: it doesn't render plain text, please don't use ai to respond

I don't have more context on the original crash since the question about determining root cause was my initial comment

PR #5861, opened 2026-05-27
headlamp-plugin: Spawn Windows .cmd shims via cmd.exe

skoeva:
These changes fix `headlamp-plugin` CLI commands still failing on Windows after #4850 by routing the `.cmd` shims that npm installs in `node_modules/.bin/` through `cmd.exe /d /s /c` when invoking them via `child_process.execFileSync` / `spawnSync`. 

Without this, Node (>= 20.12.2, via the CVE-2024-27980 mitigation) refuses to spawn `.cmd`/`.bat` files and the commands exit with `spawnSync <name>.cmd EINVAL`. Routing through `cmd.exe` is the approach [recommended by the Node docs](https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows); we avoid `shell: true` because it is [deprecated when combined with an `args` array (DEP0190)](https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild_process-execfilespawn-with-shell-option). 

A small `viaCmd(file, args)` helper centralizes this and is a no-op on non-Windows platforms. The `.cmd` shim suffix is also applied to the remaining `npm`, `npx`, and `i18next` call sites that were missed previously.

### Testing

- [x] Run commands like `npm run tsc`, `npm run lint`, and `npm run i18n` locally on Windows
- [x] Ensure that they do not fail with a vague error message

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -33,13 +33,36 @@ const { PluginManager, MultiPluginManager } = require('@headlamp-k8s/pluginctl')
 const { table } = require('table');
 const tar = require('tar');
 
-// On Windows, use the .cmd shim so execFileSync can run it without a shell.
-const binExt = process.platform === 'win32' ? '.cmd' : '';
-
 // ES imports
 const viteCopyPluginPromise = import('vite-plugin-static-copy');
 const viteConfigPromise = import('../config/vite.config.mjs');
 const vitePromise = import('vite');
+
+// On Windows, npm-installed CLIs are `.cmd` shims that Node can't spawn directly.
+// We append this suffix when looking them up on disk.
+const binExt = process.platform === 'win32' ? '.cmd' : '';
+
+/**
+ * Adapt a `(file, args)` pair for spawning on the current platform.
+ *
+ * On Windows, `.cmd`/`.bat` shims can't be spawned directly by Node, so we route them
+ * through `cmd.exe /d /s /c`, as recommended by
+ * https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows.
+ * This avoids `shell: true`, which is deprecated with an `args` array (DEP0190).
+ *
+ * Spread the result into `spawnSync` / `execFileSync`:
+ *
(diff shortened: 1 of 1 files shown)


plugins/headlamp-plugin/bin/headlamp-plugin.js
```diff
@@ -101,8 +103,9 @@ function create(name, link, noInstall) {
   // This can be used to make testing locally easier.
   if (link) {
     console.log('Linking @kinvolk/headlamp-plugin');
-    child_process.spawnSync('npm', ['link', '@kinvolk/headlamp-plugin'], {
+    child_process.spawnSync('npm' + binExt, ['link', '@kinvolk/headlamp-plugin'], {
       cwd: dstFolder,
+      shell: winShell,
```
sniok: node docs don't recommend using shell argument since it's potentially insecure and [deprecated](https://nodejs.org/api/deprecations.html#dep0190-passing-args-to-nodechild_process-execfilespawn-with-shell-option)

https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows

could you try one of the alternative options mentioned there? I think doing `spawn('cmd.exe', ...` should work

sniok (changes requested):

plugins/headlamp-plugin/bin/headlamp-plugin.js
skoeva: done

sniok:
running npm run lint,tsc,i18n in frontend on windows seems to work currently on main, need to verify the issue in the first place here

skoeva (commented):
I think this can be safely closed for now

skoeva:
reopening - verified this is still broken on main and found why

things work currently on the **frontend** i18n, which runs npm as a script (npm launches the `.cmd` via a shell)

the headlamp-plugin CLI spawns the `.cmd` programatically (`execFileSync('npx'+binExt, ...)` and `i18next.cmd`, both with no shell) which node >= 20.12.2 rejects with `spawnSync npx.cmd EINVAL` (CVE-2024-27980). different code paths, so both observations were right

repro on node v20.20.2: `npm run i18n` in any plugin fails; the same call with shell/`cmd.exe` succeeds. #4850 is necessary but not sufficient - so this PR's `cmd.exe` routing is needed

illume (approved):
🎉 thanks!

PR #5865, opened 2026-05-27
frontend: useKubeObjectList: Use Set for O(1) watched-list lookup

ayushmaan-16:
## Description

In `useKubeObjectList`, the reconciliation logic that determines which lists to start watching uses a nested loop pattern. For every item returned by the query, it scans through the entire `listsToWatch` array using `.find()` to check if that `(cluster, namespace)` combination is already being watched. This results in `O(N×M)` time complexity, which scales poorly when there are many clusters and namespaces.

This PR optimizes the `listsNotYetWatched` check by pre-computing a `Set` of currently watched `(cluster, namespace)` keys via `useMemo`. This changes the nested `.find()` scan into a direct `O(1)` `Set.has()` lookup, reducing per-render overhead without altering any existing behaviour.

A `watchKey` helper is also extracted to centralize the `cluster:namespace` key format used across the watch-list logic, avoiding future drift if the separator or normalization ever changes.

### `frontend/src/lib/k8s/api/v2/useKubeObjectList.ts`

- Added `watchKey(cluster, namespace)` helper function that returns a stable string key for a `(cluster, namespace)` pair, used across all `Set`/`Map` lookups in the reconciliation logic.

- Added `currentlyWatchedKeys` — a `useMemo`-wrapped `Set` built from `listsToWatch`, replacing the previous nested `.find()` scan.

- Updated `listsNotYetWatched` filter to use `currentlyWatchedKeys.has()` for `O(1)` lookups instead of a linear `.find()` over `listsToWatch`.

- Replaced `.filter(Boolean)` with a typed predicate:
  ```ts
  (data): data is ListResponse<K>
  ```
  so the pipeline is properly type-narrowed and downstream non-null assertions are eliminated.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts
@@ -124,6 +124,15 @@ export function kubeObjectListQuery<K extends KubeObject>(
 }
 
 /**
+ * Returns a stable string key for a (cluster, namespace) pair.
+ * Used to build Sets and Maps for O(1) membership checks across the
+ * watch-list reconciliation logic.
+ */
+function watchKey(cluster: string, namespace: string | undefined): string {
+  return `${cluster}:${namespace || ''}`;
+}
+
+/**
  * Accepts a list of lists to watch.
  * Upon receiving update it will modify query data for list query
  */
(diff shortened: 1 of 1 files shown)


sniok (changes requested):
amount of requests is determined by amount of selected clusters multiplied by amount of selected namespaces, so that number will be small. there's no performance issue here and constructing a set is not needed and might actually be slower for most cases (1 cluster, 1 namespace)

skoeva (commented):
thanks for the contribution! as noted in the previous comment, `listToWatch` is bounded by selected clusters x namespaces (typically very small), so swapping `.find()` for a `Set` is a net loss for the common case and isn't solving a real hotspot. going to close this one.

PR #5866, opened 2026-05-27
backend: oidc: Evict abandoned state tokens to prevent memory leak

ayushmaan-16:
## Description

This PR fixes an unbounded memory growth issue in the OIDC login flow.

When a user starts OIDC login, the backend creates a random state token and stores the related login config in `oauthRequestMap`. That entry is normally deleted when the user completes login and reaches `/oidc-callback`.

However, if the user closes the tab, abandons the IdP login, loses network, or repeatedly starts login without completing it, the callback never happens and the entry currently stays in memory until the Headlamp process restarts.

This PR adds a TTL-based cleanup path for those abandoned OIDC state entries. Each entry now records when it was created, and a background cleanup goroutine periodically removes entries older than `OidcStateTTL`.

The TTL is set to 60 minutes to avoid interrupting slow/manual login flows while still bounding memory growth. Since cleanup runs every `OidcStateTTL / 2`, the eviction check runs every 30 minutes.

## Changes Made

- Added `OidcStateTTL = 60 * time.Minute`.
- Added `createdAt` to `OauthConfig`.
- Set `createdAt` when storing a new OIDC state entry.
- Added cleanup for abandoned OIDC state entries older than the TTL.
- Ensured the cleanup goroutine exits when the server context is cancelled.
- Added `TestOidcStateMapEviction` to verify stale entries are removed and fresh entries remain.

## Testing

```sh
go test ./cmd -run TestOidcStateMapEviction -count=1 -v

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -184,6 +184,23 @@ type OauthConfig struct {
 	createdAt    time.Time
 }
 
+// evictExpiredOidcStates removes entries from m whose createdAt timestamp is
+// older than ttl. It is called by the background cleanup goroutine in
+// createHeadlampHandler and is also used directly in tests so that both paths
+// exercise the same production logic.
+func evictExpiredOidcStates(m map[string]*OauthConfig, mu *sync.Mutex, ttl time.Duration) {
+	cutoff := time.Now().Add(-ttl)
+
+	mu.Lock()
+	defer mu.Unlock()
+
+	for state, entry := range m {
+		if entry.createdAt.Before(cutoff) {
+			delete(m, state)
+		}
+	}
+}
+
 // returns True if a file exists.
 func fileExists(filename string) bool {
 	info, err := os.Stat(filename)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ayushmaan-16** [comment]
@illume hey I guess there are no conflicts anymore and i've resolved all the copilot comments.

File: backend/cmd/headlamp.go [hunk 1]
```diff
@@ -134,6 +134,11 @@ const ContextUpdateCacheTTL = 20 * time.Second // seconds
 
 const JWTExpirationTTL = 10 * time.Second // seconds
 
+// OidcStateTTL is the maximum time an OIDC state token is kept in memory
+// waiting for the callback. Entries that are never completed (e.g. the user
+// closed the tab) are evicted after this duration to prevent unbounded growth.
+const OidcStateTTL = 10 * time.Minute
```
illume: Can we bump this to 60 minutes?

For some people it can take a long time to do tasks, so I think 60 minutes should allow people plenty of time. But still provide the benefits of this PR.

What do you think?

illume (changes requested):
I left a comment about the time required? Can you please check it to see if it's ok?

ayushmaan-16:
@illume Yes, 60 minutes sounds reasonable.

This PR prevents abandoned OIDC login attempts from staying in `oauthRequestMap` until the headlamp process restarts. Completed logins are already removed on `/oidc-callback`, but flows where the user closes the tab, abandons IdP login, or repeatedly starts login without completing it currently remain in memory indefinitely.

A 60-minute TTL keeps that memory bounded while giving users plenty of time to complete slower/manual login flows. The tradeoff is only that callbacks after 60 minutes must restart login.

ayushmaan-16:
@illume Well I just had to update the PR description. I've done that

illume (approved):
🎉 thanks!

PR #5868, opened 2026-05-28
frontend: authchooser: Make back button keyboard-accessible

### LINKED ISSUE #5867: frontend: a11y: Back button in AuthChooser is not keyboard-accessible
## Describe the bug
I was checking the keyboard navigation on the cluster authentication screen (where you choose between OIDC/token login) and noticed a minor accessibility issue with the "Back" action. 
While tab-navigating the page, the focus completely skips the "Back" button. I checked the DOM and noticed it's rendered as a `div`/`Box` with a click handler and `role="button"`, but it doesn't have a `tabIndex={0}` or a keyboard event listener (`onKeyDown`) to handle the `Enter`/`Space` keys.
Because of this, keyboard-only or screen reader users are locked out from going back to the main cluster chooser screen.

## To Reproduce
Steps to reproduce the bug:
1. Run Headlamp and configure a cluster requiring authentication.
2. Select the cluster to go to the authentication chooser screen.
3. Try to navigate the page elements using only the keyboard (`Tab` key).
4. Observe that the focus completely skips the "Back" button at the bottom.
5. Even if focused manually via dev tools, pressing `Enter` or `Space` does not trigger the back navigation.

## Environment
 - Installation type: Frontend dev
 - Headlamp Version: main branch (latest)

Yes, I am happy to propose a PR to fix this!

YadavAkhileshh:
### Summary
Makes the "Back" action on the cluster authentication chooser screen keyboard-accessible. 

### Changes
* Added `tabIndex={0}` to the custom Back button (`Box` element) to allow keyboard focus/tabbing.
* Added an `onKeyDown` listener to handle `Enter` or `Space` key presses.
* Updated corresponding storybook snapshots to reflect the new `tabindex` attribute.

Fixes #5867

### DIFF
--- a/frontend/src/components/authchooser/__snapshots__/AuthChooser.AnError.stories.storyshot
+++ b/frontend/src/components/authchooser/__snapshots__/AuthChooser.AnError.stories.storyshot
@@ -137,6 +137,7 @@
               class="MuiBox-root css-oo0cqs"
               role="button"
               style="cursor: pointer;"
+              tabindex="0"
             >
               <div
                 class="MuiBox-root css-37urdo"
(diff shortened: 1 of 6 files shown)


illume (changes requested):
Thanks for working on this.

PR #5876, opened 2026-05-28
frontend: ObjectEventList: Handle events without a source field

### LINKED ISSUE #5823: Events table crashes when event has no source field

ayush-that:
Fixes #5823.

The `From` column getter in `ObjectEventList` reads `item.source.component` directly, which throws a `TypeError` when the Kubernetes API omits `source`. Events emitted under the `events.k8s.io/v1` API routinely lack this field, breaking the events table on resource detail pages.

Use optional chaining so the column renders empty for events without `source`, matching the existing null-safe pattern used elsewhere in the same file.

Also add a Storybook mock event without `source` to the `WithEvents` story so it exercises the no-source path and prevents regression.

### Testing

- `npx tsc --noEmit` clean
- `eslint` clean
- `npx vitest run src/storybook.test.tsx` — 542 tests pass with the updated snapshot
- Reverting the `?.` change makes the `WithEvents` story throw, confirming the snapshot covers the regression

### DIFF
--- a/frontend/src/components/common/ObjectEventList.stories.tsx
+++ b/frontend/src/components/common/ObjectEventList.stories.tsx
@@ -100,6 +100,31 @@ const mockEvents: KubeEvent[] = [
     count: 1,
     type: 'Normal',
   },
+  {
+    kind: 'Event',
+    apiVersion: 'v1',
+    metadata: {
+      name: 'event3.789',
+      namespace: 'default',
+      uid: 'event-uid-3',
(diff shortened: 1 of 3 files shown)


ayush-that:
sorry for the duplicate.

PR #5877, opened 2026-05-28
backend: auth: Drop IncrementRequestCounter from headlamp.go

skoeva:
This change removes the `IncrementRequestCounter` helper from `headlamp.go`. The same `RequestCounter` is already incremented for every HTTP request by `RequestCounterMiddleware` in `pkg/telemetry/metrics.go`, and again by `RecordRequestCount` in downstream handlers, so this was producing duplicate counts on the same metric.

Part of:
- #3482 

### Updates

- Drop `IncrementRequestCounter` and its call site in `OIDCTokenRefreshMiddleware`

### DIFF
--- a/backend/pkg/auth/middleware.go
+++ b/backend/pkg/auth/middleware.go
@@ -28,7 +28,6 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/telemetry"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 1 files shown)


backend/pkg/auth/auth.go
```diff
@@ -574,6 +575,20 @@ func marshalToString(val interface{}) (string, bool) {
 	return string(b), true
 }
 
+// IncrementRequestCounter records the start of an OIDC token refresh
+// middleware request in the shared request counter metric.
+func IncrementRequestCounter(ctx context.Context, metrics *telemetry.Metrics) {
```
sniok: hm there's a very similar function `RecordRequestCount` in the `requesthandler.go` which also increases requestcounter, I wonder if we're double counting?

skoeva: looked deeper and honestly yeah - I think this could be removed

illume (approved):
🎉 thanks!

PR #5880, opened 2026-05-28
frontend: useKubeObjectList: Fix kind name replacement for individual items within a list

sniok:
Originally reported on [slack](https://kubernetes.slack.com/archives/C01FXB5E8ER/p1779984731068429) 

```
So if you have resources with "List" in their names and try to map them, loco things happen :smiley:

In my case for tekton pipelines EventListener became evenTenerList or something like that
```

Fixes replacement to make sure we only replace List suffix, and not in the middle of the word

## Steps to Test

1. Go to a resource list page that has a `List` in the middle of the kind name, for example Event*List*ener
2. Inspect individual item and validate that the `kind` is EventListener and not `evenTenerList `


joaquimrocha (approved):
Code LGTM. Haven't tested.

PR #5883, opened 2026-05-29
github: Use GH_TOKEN for website trigger

skoeva:
This change updates the website trigger workflow to provide the GitHub CLI token through `GH_TOKEN`, which fixes authentication for the `gh api` dispatch call.

### Error output

image

### Testing

- Confirmed the workflow now uses `GH_TOKEN` as required by GitHub CLI in Actions
- Not run locally; workflow-only change.


sniok (approved):
thanks!

/lgtm

PR #5891, opened 2026-05-29
redraw lost edges when nodes move

gattytto:
frontend: resourceMap.graphFiltering: add edge redrawing for nodes that move

Nodes can move when switching from cloud to namespace-specific view. This requires a second pass to redraw lost edges for the nodes that moved.

image


gattytto:
@sniok thank you so much for all the help with this PR

sniok (approved):
nice fix, thank you!

/lgtm

PR #5895, opened 2026-06-01
CI: Parallelize and speed up plugins-test

skoeva:
This change reduces plugin-test CI wall-clock time from ~16 min to ~8 min by splitting jobs and parallelizing the example plugin tests.

Follow-up from #5885

### Changes
- Split `testplugins` into `test-headlamp-plugin`, `test-plugin-examples`, and `test-pluginctl` jobs that run in parallel
- Each job caches only the lockfiles it needs and installs its own dependencies
- In `test-plugins-examples.sh`, run example plugin tests in parallel (default `PARALLEL=4`, override with env var) using `xargs -P`
- In `Makefile`, remove the redundant `npm install` before `test-headlamp-plugin.js` (it installs internally), use `npm ci` for pluginctl, combine two Jest commands into one `--runInBand` invocation, and add the missing `.PHONY` declaration

### Testing

- [x] `git diff --check` passes (no whitespace issues)
- [x] `make -n plugins-test` shows expected command flow
- [x] CI run confirms all three jobs pass and total wall-clock time is reduced

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -113,8 +113,8 @@ jobs:
         run: |
           make frontend-build-rsbuild
 
-  testplugins:
-    name: test plugins
+  test-headlamp-plugin:
+    name: test headlamp-plugin
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/frontend.yml
```diff
@@ -129,11 +129,62 @@ jobs:
...
+          cache-dependency-path: |
+            plugins/headlamp-plugin/package-lock.json
+            plugins/examples/*/package-lock.json
+
+      - name: Test plugin examples
+        run: |
+          cd plugins/headlamp-plugin && npm ci && ./test-plugins-examples.sh
+
+  test-pluginctl:
+    name: test pluginctl
```
illume, line 163: After name: block you could put this to stop it running unless there were changes in that folder...

```yaml
    paths:
      - 'plugins/pluginctl/**'
```

illume (approved):
Nice one. I left an optional suggestion to only run the pluginctl tests when there's a change to that folder.

plugins/headlamp-plugin/test-plugins-examples.sh
```diff
@@ -9,22 +9,42 @@ npm run build
...
-    npm run format
-    npm run build
-    npm run tsc
-    cd ..
-  fi
-done
 
+# Run the per-example test pipeline in parallel via xargs -P. GNU xargs
+# returns exit 123 if any worker fails, which `set -e` then propagates.
+ls -d */ | tr -d / | xargs -P "$PARALLEL" -I {} sh -c '
```
illume: Hmm. Reading the log files I can't really see how things are with this, because all the lines get combined from different ones.

Maybe we can use that npm package from the repo root to do this, because it prepends something to the start of each log line saying where it's from.

illume (changes requested):
oh, I think those cp ones are worth a look.

But maybe replace those xargs things with the repo root npm package?

plugins/headlamp-plugin/test-plugins-examples.sh
skoeva: switched to concurrently so each example's lines are prefixed and colored by dir name

skoeva:
> oh, I think those cp ones are worth a look.
> 
> But maybe replace those xargs things with the repo root npm package?

@illume Took both. xargs is gone (replaced by concurrently), and removed the redundant `npm install` from `test-pluginctl.js`

Makefile
```diff
@@ -349,12 +349,13 @@ lint: backend-lint frontend-lint
...
-	cd plugins/headlamp-plugin && npm install && ./test-headlamp-plugin.js
+	cd plugins/headlamp-plugin && ./test-headlamp-plugin.js
 	cd plugins/headlamp-plugin && ./test-plugins-examples.sh
-	cd plugins/pluginctl/src && npm install && node ./plugin-management.e2e.js
-	cd plugins/pluginctl && npx jest src/multi-plugin-management.test.js
-	cd plugins/pluginctl && npx jest src/plugin-management.test.js
+	cd plugins/pluginctl && npm ci
+	cd plugins/pluginctl/src && node ./plugin-management.e2e.js
+	cd plugins/pluginctl && npx jest --runInBand src/multi-plugin-management.test.js src/plugin-management.test.js
 	cd plugins/pluginctl && npm run test
```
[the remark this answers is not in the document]
skoeva: removed the `npm install` call

plugins/headlamp-plugin/test-plugins-examples.sh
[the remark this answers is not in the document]
skoeva: replaced with a plain `for dir in */` glob, and the loop builds the command list passed to concurrently

.github/workflows/frontend.yml
skoeva: this could maybe be added in a follow-up pr if it's worth adding

illume (approved):
🎉 thanks!

PR #5767, opened 2026-05-20
frontend: NumRowsInput: Fix react-hooks/exhaustive-deps 

kishore08-07:
## Summary
This PR removes the `react-hooks/exhaustive-deps` suppression in `frontend/src/components/App/Settings/NumRowsInput.tsx` by adding missing options dependency to `defaultRowsPerPageValue` and `storedCustomValue` in `useMemo` and adding 'dispatch' to the `useEffect` dependency array.

## Related Issue
Fixes #5766 (Sub-issue of #5183)

## Changes

- Added missing options dependency to defaultRowsPerPageValue and storedCustomValue in useMemos
- Added 'dispatch' to the useEffect dependency array.
- This allowed all eslint suppressions to be safely removed without causing re-renders

## Steps to Test

1.`cd frontend && npm install`
2.`npm run lint` passes and no react-hooks/exhaustive-deps violation surfaces for this hook.
3.`npm run tsc` passes.

## Screenshots (if applicable)
N/A. No functional change.

## Notes for the Reviewer
- DCO sign-off applied (Signed-off-by: in the commit).
- Listed as an unchecked exhaustive-deps bullet in umbrella issue https://github.com/kubernetes-sigs/headlamp/issues/5183

PR #5898, opened 2026-06-01
frontend: NumRowsInput: Refactor state initialization using lazy state initializers

YadavAkhileshh:
### Summary
This PR refactors the state initialization logic inside the settings module (`NumRowsInput.tsx`) to use React's standard lazy state initialization pattern instead of redundant `useMemo` hooks.

### Rationale
In the current implementation, `useMemo` hooks with empty dependency arrays were being used solely to compute the initial values of `selectedValue` and `customValue` state variables. Since these hook blocks access outer variables but restrict re-runs on updates, they violated standard React linter rules, requiring inline `eslint-disable` overrides to pass checks.

Using `useMemo` to initialize state is a known React anti-pattern. Migrating this logic to lazy state initializers passed directly into the `useState` hook ensures that the initial values are computed exactly once when the component mounts, aligns perfectly with official React guidelines, and cleanly eliminates the need for linter silencing flags.

### Changes
*   Removed the redundant `useMemo` calculations and their corresponding React imports.
*   Refactored `selectedValue` and `customValue` hooks to use clean, standard lazy state initialization functions.
*   Removed deprecated `eslint-disable-next-line react-hooks/exhaustive-deps` directives from the settings component.

### DIFF
--- a/frontend/src/components/App/Settings/NumRowsInput.tsx
+++ b/frontend/src/components/App/Settings/NumRowsInput.tsx
@@ -25,7 +25,7 @@ import MenuItem from '@mui/material/MenuItem';
 import { SelectChangeEvent } from '@mui/material/Select';
 import Select from '@mui/material/Select';
 import TextField from '@mui/material/TextField';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


skoeva (commented):
closing - this conflicts with #5767

PR #5902, opened 2026-06-02
frontend: Improve French translations

codeurluce:
## Summary

This PR updates a French translation string to better reflect the meaning of "Issue" in the context of GitHub-style issue tracking.

### Changes

* `Open an Issue`

  * From: `Créer un ticket d'incident`
  * To: `Signaler un problème`

### Rationale

An issue is not necessarily an incident. It can represent a bug report, feature request, task, or question. "Signaler un problème" is broader and better matches the meaning of "Issue" in this context.


sniok (approved):
thanks for contributing!

/lgtm

PR #5907, opened 2026-06-03
app: frontend: plugins: Bump vitest and vitest/coverage-istanbul

skoeva:


illume:
@skoeva there's a bunch of non dependency update changes in here? Are they needed for the dependency updates?

app/electron/runCmd.ts
```diff
@@ -316,7 +317,8 @@ export function runScript() {
     process.exit(1);
   }
 
-  import(scriptPath);
+  // Use a file:// URL so absolute paths (especially on Windows) are valid ESM specifiers.
+  return import(pathToFileURL(scriptPath).href);
```
illume, line 343: I'm not sure about his. Why is this in this PR? This would need to be tested with the minikube plugin.

frontend/src/components/App/Settings/IconPicker.test.tsx
```diff
@@ -26,12 +26,12 @@ vi.mock('react-i18next', () => ({
...
-  let onClose: ReturnType<typeof vi.fn>;
-  let onSelectIcon: ReturnType<typeof vi.fn>;
+  let onClose: ReturnType<typeof vi.fn<() => void>>;
+  let onSelectIcon: ReturnType<typeof vi.fn<(icon: string) => void>>;
 
   beforeEach(() => {
-    onClose = vi.fn();
-    onSelectIcon = vi.fn();
+    onClose = vi.fn<() => void>();
+    onSelectIcon = vi.fn<(icon: string) => void>();
```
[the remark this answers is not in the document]
skoeva: this is actually incorrect but using `Mock<T>` is probably still a good idea

app/electron/runCmd.ts
skoeva: ~~this is an outstanding windows issue - will test/open the issue for it~~

the vitest 4 change surfaces a windows-only bug that got swallowed in the previous version. now it shows up in the CI: https://github.com/kubernetes-sigs/headlamp/actions/runs/27559033539/job/81465712107?pr=5907

illume: This code change should be tested with the Minikube plugin first in a compiled build. I’m unclear if there is adequate test coverage for this on all the platforms so it would need to be tested manually.

mahesh-09-12:
@illume while investigating #5980, I independently narrowed the failures down to the Vitest/jsdom path and ended up on the same Vitest upgrade approach. After @skoeva pointed me to this PR, I confirmed it resolves the Node.js 24 frontend test failures I was seeing locally

PR #5920, opened 2026-06-05
frontend: fix: Harden isValidRedirectPath against whitespace and URL-encoded protocol bypass

Athang69:
## Summary
This PR fixes open-redirect bypass vulnerabilities in `isValidRedirectPath` by trimming leading whitespace before all protocol checks and decoding URL-encoded sequences before matching dangerous protocols.

## Changes
- Added `trimmedPath` variable so all protocol checks (`http://`, `https://`, `//`) operate on the whitespace-stripped path instead of the raw input
- Added `decodeURIComponent` with `try/catch` fallback before the dangerous-protocol check to block URL-encoded bypass attempts like `java%73cript:` or `%6aavascript:`
- Expanded `AppContainer.test.tsx` with six new test blocks covering leading-whitespace bypass, URL-encoded protocol bypass (including the malformed-percent-encoding fallback), scheme-based URLs without `//`, protocol-relative URLs with whitespace, valid paths with query strings and fragments, and edge cases like bare words `javascript` and `data` without a colon

## Steps to Test
1. Run `npm test` from `frontend/` and verify all 11 tests in `AppContainer.test.tsx` pass
2. In a running Headlamp instance navigate to `/?to=%6aavascript:alert(1)` and confirm no redirect occurs
3. Navigate to `/?to= //malicious.com` and confirm no redirect occurs
4. Navigate to `/?to=/settings/cluster` and confirm the redirect works correctly

N/A

## Notes for the Reviewer
- The `decodeURIComponent` call is wrapped in `try/catch` because malformed percent-sequences like `%zz` throw a `URIError`; in that case the raw trimmed path is used as a safe fallback
- The two whitespace cases inside the existing `should block dangerous protocols` block are intentional duplicates of cases in the new dedicated whitespace block; the first documents the fix in context, the second groups all whitespace coverage together

### DIFF
--- a/frontend/src/components/App/AppContainer.test.tsx
+++ b/frontend/src/components/App/AppContainer.test.tsx
@@ -53,4 +53,41 @@ describe('isValidRedirectPath', () => {
     expect(isValidRedirectPath(null as any)).toBe(false);
     expect(isValidRedirectPath(undefined as any)).toBe(false);
   });
+
+  it('should block URL-encoded protocol bypass attempts', () => {
+    expect(isValidRedirectPath('java%73cript:alert("XSS")')).toBe(false);
+    expect(isValidRedirectPath('%64ata:text/html,<script>alert("XSS")</script>')).toBe(false);
+    expect(isValidRedirectPath('javascript%3Aalert("XSS")')).toBe(false);
+    expect(() => isValidRedirectPath('%zz')).not.toThrow();
+    expect(isValidRedirectPath('%zz')).toBe(true);
+  });
+
+  it('should block scheme-based URLs without //', () => {
+    expect(isValidRedirectPath('http:evil.com')).toBe(false);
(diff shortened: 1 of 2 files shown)


Athang69:
@illume PTAL whenever convenient.
Thanks!

illume (changes requested):
Thanks for this PR.


Athang69:
@illume PTAL whenever convenient.
Thankss!

illume (changes requested):
Thanks for the contribution.


Athang69:
> Thanks for the contribution.
> 
> There are some open Copilot review comments — could you take a look at them? Please mark each one as resolved once you've addressed it.

@illume Done. Fixed the copilot's reviews PTAL.
Thanks!

illume (approved):
🎉 thanks!

Great work!

PR #5929, opened 2026-06-05
backend: Broadcast OIDC token across kubeconfig contexts with same IdP

Naga15:
## Summary

When a Headlamp instance manages multiple clusters that all trust the
same OIDC identity provider, a user must today log in once per cluster
even though one IdP issued the token. This PR adds an opt-in
`--oidc-use-token-broadcast` flag (default false). When enabled, after a
successful OIDC login on cluster X the auth cookie is also set for every
other kubeconfig context whose OIDC auth-provider has the same
`idp-issuer-url` AND `client-id`.

This eliminates per-cluster re-login in deployments where one OIDC
application is trusted by every cluster's apiserver — a common pattern
with Okta / Keycloak / Dex fronting EKS + AKS + GKE.

## Related issue

Refs #4283

That discussion surfaced two approaches: (A) token broadcast across
contexts sharing the same IdP, and (B) impersonation headers injected
by the backend. This PR implements (A) as an opt-in flag and does not
preclude (B) — they address different deployment shapes (A: same IdP
trusted everywhere; B: ServiceAccount-on-remote-cluster impersonating
the OIDC identity). Happy to follow up with (B) in a separate PR if
maintainers want both.

## Why a flag rather than default-on?

@illume asked whether `--oidc-use-token-broadcast` is needed at all, or
whether broadcast could be the default with no flag. Answering here for
the record.

**Downsides of making it the default:**

- **Wider credential distribution by default** — every multi-cluster OIDC
  user would start getting an auth cookie set for *all* sibling clusters
  on each login: a silent behavior change for deployments that never
  opted in.
- **Blast radius** — any context matching issuer+client-id (including a
  misconfigured or less-trusted one in the kubeconfig) would
  automatically receive the token. Opt-in preserves a least-privilege
  default for a feature whose whole job is distributing a credential more
  widely.
- **The `--oidc-use-access-token` caveat would affect everyone** — with
  access tokens the `aud` is provider-specific and often isn't the
  client-id, so a sibling apiserver can reject the broadcast cookie.
  Opt-in surfaces that only for users who chose the behavior.
- **Cookie volume** — large multi-cluster kubeconfigs emit many
  `Set-Cookie` headers per login; default-on approaches browser/proxy
  cookie limits sooner.
- **Consistency** — the sibling OIDC behaviors (`oidc-use-cookie`,
  `oidc-use-access-token`) are all opt-in flags today.

**The case for default-on is real too:** one fewer knob, and the token is
already valid at the siblings — there is no privilege escalation, since
the user could log into those clusters directly.

**The deciding argument:** IdP-sharing isn't the same as "should share a
session." Several clusters can be registered with the same OIDC app yet
still warrant separate sessions — e.g. prod and dev on the same IdP, or
(per @harse's 10+ cluster setup in #4283) a large fleet where one login
shouldn't seed cookies across everything. Even matching clusters
shouldn't automatically share a session unless the operator says so.

**Natural follow-up:** beyond a global on/off, broadcast could be scoped
to a named group of contexts — e.g.
`--oidc-token-broadcast-groups=prod:ctxA,ctxB;dev:ctxC`, or a per-context
tag carried in the kubeconfig as a `headlamp` extension so group
membership travels with the context. Broadcast would then only reach
siblings in the source's group. Suggest keeping this PR as the simple
opt-in MVP that unblocks the multi-cluster case, and adding group-scoping
as a follow-up for fleets wanting finer control.

**If you'd prefer default-on (dropping the flag), I'm happy to make that
change — just say the word.**

## Scope

In scope:
- New flag `--oidc-use-token-broadcast` (default `false`)
- `broadcastOIDCToken` helper: walks the kubeconfig, sets the cookie
  on every target context whose OIDC config matches the source's
  `idp-issuer-url` + `client-id`
- 9-case unit test covering: matching broadcast, mismatched issuer
  skipped, mismatched client_id skipped, non-OIDC target skipped,
  source-without-OIDC no-op, source never receives a self-broadcast,
  mixed contexts broadcast only to matching, missing source cluster
  handled cleanly

Out of scope (follow-ups):
- Refresh-path broadcasting — refresh logic now lives in `pkg/auth`
  and extending it requires an API change there. Tracked as a
  follow-up rather than rolled into this PR.
- Impersonation-header path (Approach B above).
- Group-scoped broadcast (see "Why a flag rather than default-on?" above).

## Preconditions and caveats

- The token's `aud` claim defaults to the `client-id`, so matching
  issuer + client_id keeps the token valid against the target
  apiserver. **Audience overrides via `--oidc-extra-audience` on a
  target apiserver are not detected here** and must be aligned by
  deployment configuration. Documented in code; called out here for
  reviewers.

## How tested

- Unit: `go test ./backend/cmd -run TestBroadcastOIDCToken` — 9 cases pass
- E2E (Tier 1, local): 2 × kind clusters trusting one Keycloak realm,
  merged kubeconfig with OIDC auth-provider for both contexts. Login on
  cluster-a → `headlamp-auth-*` cookies set for both clusters → cluster-b
  reachable without re-login. Negative test (different `client-id` on
  cluster-b): only cluster-a cookie set, cluster-b prompts re-login.
- DevTools screenshots and full runbook available on request; I can post
  them as a follow-up comment if useful.

## Notes for reviewers

- The flag is intentionally **opt-in and default false** — zero behavior
  change for existing deployments. Rationale in "Why a flag rather than
  default-on?" above.
- The source cluster never receives a duplicate broadcast cookie (it
  already has the original).
- Failure of one target broadcast (e.g., bad context) does not abort
  the others.

/kind feature

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1118,6 +1118,18 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		// Set auth cookie
 		auth.SetTokenCookie(w, r, oauthConfig.Cluster, rawUserToken, config.BaseURL, config.SessionTTL)
 
+		if config.OidcUseTokenBroadcast {
+			auth.BroadcastOIDCToken(auth.BroadcastOIDCTokenParams{
+				Writer:          w,
+				Request:         r,
+				KubeConfigStore: config.KubeConfigStore,
+				SourceCluster:   oauthConfig.Cluster,
+				Token:           rawUserToken,
+				BaseURL:         config.BaseURL,
+				SessionTTL:      config.SessionTTL,
+			})
+		}
+
 		redirectURL += fmt.Sprintf("auth?cluster=%1s", oauthConfig.Cluster)
 
 		http.Redirect(w, r, redirectURL, http.StatusSeeOther)
--- a/backend/pkg/auth/broadcast.go
+++ b/backend/pkg/auth/broadcast.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth
+
+import (
+	"net/http"
+
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
+)
+
+const (
+	logFieldSourceCluster = "sourceCluster"
+	logFieldTargetCluster = "targetCluster"
+)
+
+// BroadcastOIDCTokenParams groups the inputs required to broadcast an OIDC
+// token's auth cookie to sibling kubeconfig contexts after a successful login.
+type BroadcastOIDCTokenParams struct {
+	Writer          http.ResponseWriter
+	Request         *http.Request
+	KubeConfigStore kubeconfig.ContextStore
+	SourceCluster   string
+	Token           string
+	BaseURL         string
+	SessionTTL      int
+}
+
+// isOIDCAuthContext reports whether the kubeconfig context's auth-provider is
+// OIDC. kubeconfig.Context.AuthType() is too coarse: it returns "oidc" for any
+// AuthProvider (e.g. gcp, azure), so we check AuthProvider.Name explicitly to
+// avoid treating non-OIDC contexts as broadcast candidates. Note that exec
+// credential plugins live on AuthInfo.Exec, not AuthInfo.AuthProvider, and are
+// not relevant here.
+func isOIDCAuthContext(kCtx *kubeconfig.Context) bool {
+	if kCtx == nil {
+		return false
+	}
+
+	if kCtx.OidcConf != nil {
+		return true
+	}
+
+	if kCtx.AuthInfo == nil || kCtx.AuthInfo.AuthProvider == nil {
+		return false
+	}
+
+	return kCtx.AuthInfo.AuthProvider.Name == "oidc"
+}
+
+// BroadcastOIDCToken sets the auth cookie for every kubeconfig context whose
+// OIDC auth-provider has the same idp-issuer-url AND client-id as the source
+// cluster.
+//
+// Precondition: the kube-apiserver of each target cluster must trust the same
+// OIDC application (issuer + client-id) as the source cluster. For the default
+// id_token the audience (aud) claim is the client-id, so broadcasting only when
+// both issuer and client-id match keeps the token valid against the target
+// cluster's apiserver without re-validating per cluster. Two audience caveats
+// are NOT detected here and must be ensured by deployment configuration:
+//   - A target apiserver configured to accept a different or additional
+//     audience than the client-id (via the "audiences" list in a structured
+//     AuthenticationConfiguration): the broadcast cookie may be set but the
+//     target apiserver could reject the token.
+//   - --oidc-use-access-token: the broadcast then carries the access_token,
+//     whose audience is provider-specific and frequently NOT the client-id
+//     (e.g. a resource/API identifier on Okta, Entra ID, Auth0). A matching
+//     issuer+client-id therefore does not by itself guarantee the access token
+//     is accepted by the target apiserver; align token audiences fleet-wide.
+//
+// Refresh-path caveat: this runs only at initial OIDC login. The token-refresh
+// middleware (OIDCTokenRefreshMiddleware, backed by RefreshAndSetToken)
+// refreshes each cluster's cookie independently and is not yet broadcast-aware.
+// Because refresh happens per-cluster, sibling cookies diverge as soon as any
+// one cluster refreshes its token; with short token lifetimes (e.g., EKS'
+// default ~1h expiry) siblings then fall back to per-cluster re-login.
+// Whichever token the flag broadcasts (id_token, or access_token when
+// --oidc-use-access-token is set) is the same one SetTokenCookie stores.
+// Broadcast-on-refresh is tracked as a follow-up PR.
+//
+// Cookie-path caveat (pre-existing, general): SetTokenCookie's chunk-clear
+// pre-step reads existing chunk cookies via r.Cookie(), which only sees cookies
+// the browser sent on the current request path. Cluster cookies are scoped to
+// /clusters/<cluster> while OIDC login completes on /oidc-callback, so stale
+// chunks from a prior longer token are not actively cleared during login. This
+// applies equally to the source cluster; broadcasting just makes the path
+// mismatch hit more clusters per login. In the rare case a re-issued token uses
+// fewer chunks than the previous one, the affected cluster(s) may need a
+// one-time re-login. Also surfaced in the flag's help text.
+//
+// Gated by the --oidc-use-token-broadcast flag (disabled by default).
+func BroadcastOIDCToken(params BroadcastOIDCTokenParams) {
+	sourceOIDCConfig, ok := loadBroadcastSourceOIDCConfig(params.KubeConfigStore, params.SourceCluster)
+	if !ok {
+		return
+	}
+
+	kContexts, err := params.KubeConfigStore.GetContexts()
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{logFieldSourceCluster: params.SourceCluster}, err,
+			"failed to get contexts for broadcasting OIDC token")
+
+		return
+	}
+
+	for _, kCtx := range kContexts {
+		if kCtx.Name == params.SourceCluster || !isOIDCAuthContext(kCtx) {
+			continue
+		}
+
+		broadcastToTarget(params, sourceOIDCConfig, kCtx)
+	}
+}
+
+// loadBroadcastSourceOIDCConfig validates that the source cluster has a usable
+// OIDC config (auth-provider type, non-nil config, non-empty issuer+client-id).
+// Logs and returns ok=false on any check failure so the caller can return early.
+func loadBroadcastSourceOIDCConfig(
+	store kubeconfig.ContextStore, sourceCluster string,
+) (*kubeconfig.OidcConfig, bool) {
+	sourceContext, err := store.GetContext(sourceCluster)
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{logFieldSourceCluster: sourceCluster}, err,
+			"failed to get source context for broadcasting OIDC token")
+
+		return nil, false
+	}
+
+	if !isOIDCAuthContext(sourceContext) {
+		logger.Log(logger.LevelInfo,
+			map[string]string{logFieldSourceCluster: sourceCluster},
+			nil, "skipping OIDC token broadcast: source cluster does not use the OIDC auth-provider")
+
+		return nil, false
+	}
+
+	oidcCfg, err := sourceContext.OidcConfig()
+	if err != nil || oidcCfg == nil {
+		logger.Log(logger.LevelInfo,
+			map[string]string{logFieldSourceCluster: sourceCluster},
+			err, "skipping OIDC token broadcast: source cluster has no usable OIDC config")
+
+		return nil, false
+	}
+
+	if oidcCfg.IdpIssuerURL == "" || oidcCfg.ClientID == "" {
+		logger.Log(logger.LevelInfo,
+			map[string]string{logFieldSourceCluster: sourceCluster},
+			nil, "skipping OIDC token broadcast: source cluster OIDC config is missing issuer or client-id")
+
+		return nil, false
+	}
+
+	return oidcCfg, true
+}
+
+// oidcIssuerAndClientID reads the issuer URL and client ID from a context's
+// OIDC configuration without triggering the filesystem I/O (idp-certificate-
+// authority read) that Context.OidcConfig() performs. Used for a cheap pre-match
+// so the CA-file cost is only paid for targets that actually share the source's
+// issuer + client-id.
+func oidcIssuerAndClientID(kCtx *kubeconfig.Context) (issuer, clientID string) {
+	if kCtx.OidcConf != nil {
+		return kCtx.OidcConf.IdpIssuerURL, kCtx.OidcConf.ClientID
+	}
+
+	if kCtx.AuthInfo != nil && kCtx.AuthInfo.AuthProvider != nil {
+		cfg := kCtx.AuthInfo.AuthProvider.Config
+		return cfg["idp-issuer-url"], cfg["client-id"]
+	}
+
+	return "", ""
+}
+
+// broadcastToTarget evaluates one candidate target context and, if it shares
+// the same issuer + client-id as the source, sets the auth cookie for it.
+//
+// The issuer/client-id comparison is done first via a cheap read that avoids
+// filesystem I/O, so large kubeconfigs with many non-matching contexts do not
+// pay the CA-file read cost during /oidc-callback. Only a matching target is
+// then run through OidcConfig() to validate the config is usable before the
+// cookie is set.
+//
+// Missing or empty issuer/client-id, or an unusable OIDC config on a matching
+// target, is logged at Info level so operators can spot misconfiguration.
+// Legitimate issuer/client-id mismatches are intentionally silent because
+// logging them per-target on every login would be noisy in fleets with many
+// non-matching contexts. Either way one bad target does not affect the others.
+func broadcastToTarget(
+	params BroadcastOIDCTokenParams,
+	sourceOIDCConfig *kubeconfig.OidcConfig,
+	kCtx *kubeconfig.Context,
+) {
+	issuer, clientID := oidcIssuerAndClientID(kCtx)
+	if issuer == "" || clientID == "" {
+		logger.Log(logger.LevelInfo,
+			map[string]string{logFieldSourceCluster: params.SourceCluster, logFieldTargetCluster: kCtx.Name},
+			nil, "skipping OIDC token broadcast: target cluster OIDC config is missing issuer or client-id")
+
+		return
+	}
+
+	if issuer != sourceOIDCConfig.IdpIssuerURL || clientID != sourceOIDCConfig.ClientID {
+		return
+	}
+
+	// Matching target: now pay OidcConfig()'s cost (may read the CA file), which
+	// also validates the config is usable before we set the cookie.
+	if _, err := kCtx.OidcConfig(); err != nil {
+		logger.Log(logger.LevelInfo,
+			map[string]string{logFieldSourceCluster: params.SourceCluster, logFieldTargetCluster: kCtx.Name},
+			err, "skipping OIDC token broadcast: target cluster has no usable OIDC config")
+
+		return
+	}
+
+	SetTokenCookie(params.Writer, params.Request, kCtx.Name, params.Token, params.BaseURL, params.SessionTTL)
+	logger.Log(logger.LevelInfo,
+		map[string]string{logFieldSourceCluster: params.SourceCluster, logFieldTargetCluster: kCtx.Name},
+		nil, "broadcasted OIDC token to cluster")
+}
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -80,6 +80,7 @@ type Config struct {
 	OidcScopes                   string `koanf:"oidc-scopes"`
 	OidcUseAccessToken           bool   `koanf:"oidc-use-access-token"`
 	OidcUseCookie                bool   `koanf:"oidc-use-cookie"`
+	OidcUseTokenBroadcast        bool   `koanf:"oidc-use-token-broadcast"`
 	OidcSkipTLSVerify            bool   `koanf:"oidc-skip-tls-verify"`
 	OidcCAFile                   string `koanf:"oidc-ca-file"`
 	MeUsernamePath               string `koanf:"me-username-path"`
@@ -600,6 +601,26 @@ func addOIDCFlags(f *flag.FlagSet) {
 	f.String("oidc-ca-file", "", "CA file for OIDC")
 	f.Bool("oidc-use-access-token", false, "Setup oidc to pass through the access_token instead of the default id_token")
 	f.Bool("oidc-use-cookie", false, "Enable OIDC cookie usage even when not running in-cluster")
+	f.Bool("oidc-use-token-broadcast", false,
+		"After a successful OIDC login on one cluster, also set the auth cookie for every other "+
+			"kubeconfig context whose OIDC auth-provider has the same idp-issuer-url AND client-id. "+
+			"This precondition is required because the issued token's audience (aud) claim defaults "+
+			"to the client-id; broadcasting only when both match keeps the token valid against the "+
+			"target cluster's apiserver. Disabled by default. "+
+			"Scope: broadcasting fires only at initial OIDC login, not on token refresh. Whichever "+
+			"token is in use is broadcast (the id_token, or the access_token when "+
+			"--oidc-use-access-token is set). Because refresh happens per-cluster, sibling cookies "+
+			"diverge as soon as any cluster refreshes its token; with short token lifetimes (e.g., "+
+			"~1h on EKS by default) siblings then fall back to per-cluster re-login. "+
+			"Broadcast-on-refresh is tracked as a follow-up. "+
+			"Operational caveats: each target cluster receives one or more Set-Cookie headers per "+
+			"login, so enabling this with very large multi-cluster kubeconfigs may approach browser "+
+			"and proxy cookie count/size limits. Pre-existing chunk-cookie limitation: chunk "+
+			"cleanup at login runs only against cookies the browser sent for the current request "+
+			"path (cluster cookies live under /clusters/<cluster>, while login completes on "+
+			"/oidc-callback). This applies to all OIDC logins; broadcast simply increases how many "+
+			"clusters experience it at once. In the rare case a re-issued token is shorter than the "+
+			"previous one, the affected cluster(s) may need a one-time re-login.")
 	f.Bool("oidc-use-pkce", false, "Use PKCE (Proof Key for Code Exchange) for enhanced security in OIDC flow")
 	f.String("me-username-path", DefaultMeUsernamePath,
 		"Comma separated JMESPath expressions used to read username from the JWT payload")
--- a/docs/installation/in-cluster/oidc.md
+++ b/docs/installation/in-cluster/oidc.md
@@ -78,6 +78,24 @@ By default, headlamp leverages the `id_token` provided back from the OIDC Provid
 
 - `-oidc-use-access-token=true` or env var `HEADLAMP_CONFIG_OIDC_USE_ACCESS_TOKEN`
 
+### Multi-cluster: broadcast the OIDC token across sibling clusters
+
+When a single Headlamp instance serves several Kubernetes clusters that all trust the **same** OIDC application (same issuer URL and client ID), an operator can opt in to broadcasting the auth cookie to every matching sibling cluster after a successful login. This eliminates per-cluster re-authentication for the common deployment shape where one OIDC app (Okta, Keycloak, Dex, Entra ID, etc.) is registered with every `kube-apiserver` in the fleet.
+
+- `-oidc-use-token-broadcast=true` or env var `HEADLAMP_CONFIG_OIDC_USE_TOKEN_BROADCAST`
+
+**Precondition.** A sibling cluster receives the broadcast only when its kubeconfig context's OIDC auth-provider has BOTH a non-empty `idp-issuer-url` AND a non-empty `client-id` that match the source cluster's. Contexts using a different auth-provider (e.g. `gcp`, `azure`) or a static token are skipped silently.
+
+**Scope.** Broadcasting fires at initial OIDC login and broadcasts whichever token is in use (the `id_token`, or the `access_token` when `-oidc-use-access-token=true`). Token-refresh broadcasting is tracked as a follow-up. Because token refresh happens independently per cluster, sibling cookies diverge as soon as any one cluster refreshes its token; until refresh broadcasting lands, the affected clusters fall back to per-cluster re-login (commonly ~1h on EKS / Okta with default settings).
+
+**Caveats to be aware of before enabling.**
+
+- The flag is **disabled by default**; existing deployments see zero behavior change.
+- Audience mismatches are not detected here. A target apiserver's accepted audiences come from its OIDC client ID (`--oidc-client-id`) or the `audiences` list in a structured [AuthenticationConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration); if a target is configured to require a different or additional audience than the source, the broadcast cookie may be set but the target apiserver could reject the token. Align deployment configuration in that case.
+- When `-oidc-use-access-token=true`, the broadcast carries the `access_token` rather than the `id_token`. Unlike the `id_token`, an access token's audience is provider-specific and is frequently **not** the client ID (many IdPs — Okta, Entra ID, Auth0 — set it to a resource/API identifier). Matching issuer + client ID therefore does not by itself guarantee the access token is accepted by a sibling apiserver; ensure the access-token audience is honored fleet-wide before relying on broadcast with this flag.
+- Each target cluster receives one or more `Set-Cookie` headers per login, so enabling this with very large multi-cluster kubeconfigs may approach browser and proxy cookie count / size limits.
+- Pre-existing chunk-cookie limitation: stale chunk cookies on cluster paths are not actively cleared during login because cookies live under `/clusters/<cluster>` while OIDC login completes on `/oidc-callback`. In the rare case a re-issued token uses fewer chunks than the previous one, the affected cluster(s) may need a one-time re-login.
+
 ### Example: OIDC with Keycloak in Minikube
 
 If you are interested in a comprehensive example of using OIDC and Headlamp,
--- a/backend/cmd/server.go
+++ b/backend/cmd/server.go
@@ -118,6 +118,7 @@ func buildHeadlampCFG(conf *config.Config, kubeConfigStore kubeconfig.ContextSto
 		NodeShellImage:                        conf.NodeShellImage,
 		NodeShellNamespace:                    conf.NodeShellNamespace,
 		OidcUseCookie:                         conf.OidcUseCookie,
+		OidcUseTokenBroadcast:                 conf.OidcUseTokenBroadcast,
 		DefaultLightTheme:                     conf.DefaultLightTheme,
 		DefaultDarkTheme:                      conf.DefaultDarkTheme,
 		ForceTheme:                            conf.ForceTheme,
--- a/backend/pkg/auth/broadcast_test.go
+++ b/backend/pkg/auth/broadcast_test.go
@@ -1,0 +1,404 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package auth_test
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"sort"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/auth"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+const (
+	testIssuerA        = "https://idp-a.example.com"
+	testIssuerB        = "https://idp-b.example.com"
+	testClientFoo      = "client-foo"
+	testClientBar      = "client-bar"
+	testBroadcastToken = "test-token"
+)
+
+// broadcastCookieInfo is the reassembled auth cookie for one cluster: the full
+// token value (all chunks concatenated in order) and the cookie Path.
+type broadcastCookieInfo struct {
+	value string
+	path  string
+}
+
+// broadcastCookies reassembles the chunked headlamp-auth-<cluster>.<i> cookies
+// set on w into a per-cluster {value, path}. Clearing cookies (empty value,
+// non-positive MaxAge emitted by SetTokenCookie's ClearTokenCookie pre-step) are
+// ignored. Reassembling the value and path lets tests assert that the broadcast
+// carried the exact source token, scoped to the target cluster — not merely that
+// some cookie was set for the cluster.
+func broadcastCookies(t *testing.T, w *httptest.ResponseRecorder) map[string]broadcastCookieInfo {
+	t.Helper()
+
+	resp := w.Result()
+	defer func() { _ = resp.Body.Close() }()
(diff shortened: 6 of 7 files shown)


Naga15:
Attaching the Tier-1 E2E screenshots referenced in the PR description. Captured locally from a 2 × kind + Keycloak setup with `--oidc-use-token-broadcast=true`.

### Positive path: one login on cluster-a → both cluster cookies set

After signing in to cluster-a via Keycloak (only), DevTools shows `headlamp-auth-cluster-a` **and** `headlamp-auth-cluster-b` cookies present. The broadcast set the cluster-b cookie automatically because both contexts share the same `idp-issuer-url` + `client-id`.

Cluster-a after broadcast login

### Cluster-b is accessible with no re-login

Same browser session, navigating to cluster-b — overview renders immediately. The broadcast cookie is honored by cluster-b's apiserver (same OIDC app trusted on both kind clusters).

Cluster-b with no re-login

### Negative test: precondition enforced

Cluster-b reconfigured with a **different** `oidc-client-id`. After logging into cluster-a, only `headlamp-auth-cluster-a.0` is set — no broadcast cookie for cluster-b. Cluster-b correctly prompts for re-login. Keycloak rejects the mismatched client as "Client not found", confirming the precondition guard is doing its job and not setting cookies that the apiserver would reject.

Negative test: mismatched client_id

Happy to share the full runbook (kind configs, Keycloak realm setup, merged kubeconfig) if useful for reviewers wanting to reproduce locally.

Naga15:
Marking this PR ready for review.

Quick recap of where things stand:

- Implementation: ~50 lines in `backend/cmd/headlamp.go` behind an opt-in `--oidc-use-token-broadcast` flag (default `false`) — zero behavior change for existing deployments
- Unit tests: 9 cases pass (`go test ./backend/cmd -run TestBroadcastOIDCToken`) covering matching/mismatched issuer, mismatched client_id, non-OIDC targets, source-without-OIDC, mixed contexts, and self-exclusion
- E2E (local): 2 × kind + Keycloak verified end-to-end — DevTools screenshots posted in the comment above showing positive path (broadcast cookie set, no re-login on second cluster) and negative path (mismatched client_id correctly skipped)
- Design context: discussed in #4283 (community thread with multiple users hitting this pain)

Out of scope for this PR (called out in the description):
- Refresh-token broadcast — refresh logic now lives in `pkg/auth` and extending it needs an API change there; planned as a separate follow-up
- Impersonation-header path (Approach B from #4283) — different deployment shape, not mutually exclusive with this

Happy to address feedback quickly. cc @joaquimrocha @illume — would appreciate a steer when bandwidth allows on whether the Approach A scope here is the direction you'd take, or if you'd prefer to see the impersonation path in parallel.

illume (changes requested):
Thanks for this PR.


<details>

- `backend: broadcast OIDC token across kubeconfig contexts sharing same IdP` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.
- `backend: add unit tests for broadcastOIDCToken` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

Naga15:
Thanks for the review @illume! Pushed an update rewriting both commit titles to follow the project's commit guidelines:

- `backend: broadcast OIDC token...` → `backend: Broadcast OIDC token across kubeconfig contexts with same IdP` (capitalized; kept under 72 chars)
- `backend: add unit tests...` → `backend: Add unit tests for broadcastOIDCToken`

Also removed the `Refs #4283` line from commit bodies since the contributing guide asks not to include issue references in commit messages — the design link still lives in the PR description.

**Update:** rebased the branch on latest `main`, which picks up the updated `node:22` image digest. The `build discover and deploy` check should now pass on its own — no separate Dockerfile PR needed.

Also folded in @copilot's two observability suggestions (logs when source/target context has no usable OIDC config) into the broadcast commit. Unit tests still pass.

Naga15:
Thanks @copilot — both points are fair. Operators losing visibility into why broadcasting was skipped is exactly the kind of silent failure that's painful to debug.

Pushed an amendment to the broadcast commit adding `LevelInfo` logs at both spots:

- When the **source** cluster's `OidcConfig()` returns nil/error: logs `"skipping OIDC token broadcast: source cluster has no usable OIDC config"` with `sourceCluster` and the underlying error
- When a **target** cluster's `OidcConfig()` returns nil/error: logs `"skipping OIDC token broadcast: target cluster has no usable OIDC config"` with both `sourceCluster` and `targetCluster` so it's clear which context misbehaved

Folded into the existing broadcast commit (rather than a separate follow-up) since the contributing guide asks for atomic commits and these logs belong to the same logical unit.

Unit tests still all pass — the `source_without_oidc_config_is_a_no-op` case now also exercises the new log line.

Also rebased on latest `main` to pick up the recent `node:22` image digest update, which should fix the `build discover and deploy` check too.

illume (changes requested):
Thanks for working on this.


The commit messages could use some tidying up to match our contribution guidelines. We use Linux kernel style — the [contributing guide](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines) has the details, and `git log` shows good examples.

<details>

- `backend: Add unit tests for broadcastOIDCToken` — Only one file changed inside `backend/`; add a sub-area so it's clear what was touched (e.g. `backend: ComponentName: description`).

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for the contribution.


The backend test job in CI is failing. Run `cd backend && go test ./...` to reproduce the errors locally.

<details>

</details>

Naga15:
Thanks for the follow-up @illume — pushed an update addressing all three threads.

**Commit-style fix:**
- `backend: Add unit tests for broadcastOIDCToken` → `backend: cmd: Add unit tests for broadcastOIDCToken`

**Backend `golangci-lint` failures (all 5 resolved locally):**
- `errcheck` on `defer resp.Body.Close()` → wrapped in `defer func() { _ = resp.Body.Close() }()`
- `funlen` on `TestBroadcastOIDCToken` (132 lines) → refactored into a thin loop over `broadcastTestCases()`, which composes from `matchingTargetCases`/`mismatchedTargetCases`/`mixedAndMisconfiguredCases`. Assertion logic moved into `assertBroadcastResult`. Top-level test fn is now ~6 lines
- `noctx` (2 sites) → `httptest.NewRequest(...)` → `httptest.NewRequestWithContext(context.Background(), ...)`
- `wsl_v5` at `headlamp.go:1202` → added the missing blank line before the bare `return`

**Copilot review comments (all 5 resolved):**
- Source/target `OidcConfig()` silent skips → already addressed in the previous push with `LevelInfo` logs (incl. cluster name)
- Cookie-path stale-chunk caveat → documented as a function-level godoc on `broadcastOIDCToken` AND surfaced in the `--oidc-use-token-broadcast` flag's help text so operators see it
- Cookie count/size operational limit → added to the flag help text (warn about large multi-cluster kubeconfigs hitting browser/proxy header limits)
- Missing test case for `target with OIDC AuthProvider but OidcConfig() returns error` → added `newOIDCContextWithUnreadableCA` helper + a new case `"target with unreadable OIDC CA is skipped but others still broadcast"` proving one bad context doesn't block valid targets

Verified locally:
- `cd backend && ./tools/golangci-lint run --path-mode=abs --timeout 5m` → `0 issues.`
- `cd backend && go test ./cmd/ -run TestBroadcastOIDCToken -v` → all 9 cases pass (8 in `TestBroadcastOIDCToken` + the no-contexts test)

All Copilot threads marked resolved.

Naga15:
Good catch @copilot — `kubeconfig.Context.AuthType()` does indeed coarsely return `"oidc"` for any `AuthInfo.AuthProvider`, so a context using a `gcp`/`azure`/`exec` auth-provider would have been let into the broadcast loop and only later filtered out by `OidcConfig()` returning a partial result.

Pushed a fix:

- Added an `isOIDCAuthContext(kCtx)` helper that checks `OidcConf != nil` OR `AuthInfo.AuthProvider.Name == "oidc"` explicitly
- The broadcast loop uses it instead of `kCtx.AuthType() != "oidc"`
- Also added the same guard on the **source** context, so a non-OIDC source (e.g. gcp auth-provider) returns early rather than running `OidcConfig()` and potentially matching an empty-issuer/empty-client against another non-OIDC target
- Documented the rationale in the helper's godoc so it's clear why we don't use `AuthType()` here

Test coverage:
- New case: `"non-oidc auth-provider (gcp) target is skipped"` (target side)
- New case: `"source with non-oidc auth-provider (gcp) is a no-op"` (source side)
- Renamed the existing `"non-oidc target is skipped"` → `"static-token target is skipped"` since it now specifically covers the static-token path, not the auth-provider path
- New helper `newGCPAuthProviderContext` builds the kubeconfig shape

Verified locally:
- `golangci-lint run --path-mode=abs --timeout 5m` → `0 issues.`
- `go test ./cmd/ -run TestBroadcastOIDCToken -v` → 10/10 cases PASS

illume (changes requested):
Thanks for working on this.


Naga15:
Pushed an update addressing all three of @copilot's new comments:

- **`isOIDCAuthContext` godoc** — removed `exec` from the example list. `exec` credential plugins live on `AuthInfo.Exec`, not `AuthInfo.AuthProvider`, so it was a misleading example. Now reads `e.g. gcp, azure` with an explicit note that exec is not relevant here
- **`broadcastOIDCToken` godoc — cookie-path caveat** — rewritten to clarify that the stale-chunk issue is **pre-existing and general** (applies to source-cluster login too because that also happens on `/oidc-callback`); broadcast just increases how many clusters hit it at once per login
- **Flag help text** — same reframing: removed the "target clusters" scoping, now explains it's a pre-existing OIDC-login limitation that broadcast surfaces more often

All three Copilot threads marked resolved. Verified locally:
- `go test ./cmd/ -run TestBroadcastOIDCToken` → 10/10 PASS
- `golangci-lint run --path-mode=abs --timeout 5m` → 0 issues

Naga15:
Self-initiated polish — added an explicit **refresh-scope caveat** to both the operator-facing docs:

- `--oidc-use-token-broadcast` flag help text now includes: *"Scope: broadcasting fires only at initial OIDC login, not on token refresh. With short id-token lifetimes (e.g., ~1h on EKS by default), sibling clusters fall back to per-cluster re-login once the source token refreshes; broadcast-on-refresh is tracked as a follow-up."*
- `broadcastOIDCToken` godoc gained the same refresh-scope paragraph, naming `OIDCTokenRefreshMiddleware` / `auth.RefreshAndSetToken` so future contributors know exactly where the follow-up needs to plug in

This was previously documented only in the PR description and commit body — neither of which an operator sees from `--help`. Without this caveat in the flag help, someone could enable broadcast on EKS and be surprised when users start needing per-cluster re-logins after the ~1h token refresh cycle.

Verified locally: `go test ./cmd/ -run TestBroadcastOIDCToken` → 10/10 PASS, `golangci-lint` → 0 issues.

illume (changes requested):
Thanks for this PR.


Naga15:
Pushed an update addressing all three of @copilot's new comments:

**1. Literal `"cluster"` key replaced with `logFieldCluster` constant** — the source-context fetch error log now uses the existing constant for consistency with other log sites in the cmd package. Added `logFieldSourceCluster` and `logFieldTargetCluster` constants too while I was there, since `"sourceCluster"`/`"targetCluster"` was about to hit goconst from repetition.

**2 + 3. Empty issuer/client-id guards on both source and target** — real correctness concern Copilot caught. If a context passes `isOIDCAuthContext` but has empty `IdpIssuerURL`/`ClientID`, the string-equality match would incorrectly pair empty-vs-empty across misconfigured contexts. Added explicit guards (with distinct log messages) on both sides:
- Source: empty fields → return early, log at LevelInfo
- Target: empty fields → skip that target, continue with others, log at LevelInfo

**Refactor to keep funlen happy:** the new guards pushed `broadcastOIDCToken` over the 60-line limit, so I split it into three small pieces with clear responsibilities:
- `broadcastOIDCToken` — the top-level entry point (just resolves source + loops)
- `loadBroadcastSourceOIDCConfig` — all source-side validation + logging
- `broadcastToTarget` — per-target validation, match check, and cookie set

Test coverage for the new code paths:
- `source with empty issuer or client-id is a no-op`
- `target with empty issuer or client-id is skipped (would otherwise match empty source incorrectly)`

Both Copilot threads marked resolved.

Verified locally:
- `go test ./cmd/ -run TestBroadcastOIDCToken -v` → 12/12 PASS
- `golangci-lint run --path-mode=abs --timeout 5m` → 0 issues

CI is now fully green on the previous HEAD as well, so this is doc-and-correctness polish on an otherwise passing PR.

Naga15:
One more pass of self-review while waiting on reviewer attention — pushed two additional changes:

**User-facing docs added** at [`docs/installation/in-cluster/oidc.md`](https://github.com/kubernetes-sigs/headlamp/blob/main/docs/installation/in-cluster/oidc.md). The file already documents every other OIDC flag (`-oidc-use-access-token`, `-oidc-validator-*`, etc.); a new "Multi-cluster: broadcast the OIDC token across sibling clusters" section now covers `-oidc-use-token-broadcast` with the precondition, scope (login-only for now), refresh-time follow-up caveat, and the cookie-limit / chunk-cookie operational notes. Mirrors the language in the flag's `--help` text so operators see the same caveats regardless of where they're reading.

**PR title updated** to match the commit titles (`Broadcast` capitalized + "with same IdP"): `backend: Broadcast OIDC token across kubeconfig contexts with same IdP`.

Both changes folded into the existing broadcast commit (no new commit). Tests + lint still green locally.

Naga15:
Fair point @copilot — the `if !c.broadcastToTarget(...) { continue }` was dead-equivalent to just calling it, since nothing followed in the loop. The bool return wasn't used anywhere.

Cleaner fix: drop the bool return from `broadcastToTarget` entirely. The function now has a `void`-style signature and the loop body is one line. Less surface, same behavior.

Thread marked resolved. `go test ./cmd/ -run TestBroadcastOIDCToken` → 12/12 PASS, `golangci-lint` → 0 issues.

Naga15:
Both @copilot points addressed:

**License header** — `headlamp_broadcast_test.go` now carries the full Apache 2.0 header including the "AS IS" warranty disclaimer, matching `headlamp_test.go` and the rest of the repo. (Was truncated after the license URL before.)

**Godoc / behavior mismatch** — the `broadcastToTarget` doc comment previously claimed all skip conditions log at Info, but issuer/client-id mismatches return silently by design. Updated the comment to say so explicitly and explain why: logging every non-matching target on each login would be noisy in fleets with many unrelated contexts. Missing/empty OIDC config still logs at Info (that's a real misconfiguration worth surfacing); a legitimate issuer/client-id mismatch stays silent.

Both threads resolved. `go test ./cmd/ -run TestBroadcastOIDCToken` → 12/12 PASS, `golangci-lint` → 0 issues.

illume (changes requested):
Thanks for the contribution.


illume (changes requested):
Thanks for this PR.


<details>

</details>

Naga15:
Rebased on latest `main` to clear the merge conflict (the conflict was with #6211's OIDC middleware extraction, which removed `setTokenFromCookie` from this file — took `main`'s deletion and kept the broadcast additions). Also addressed all three @copilot comments in the same push:

**Performance — avoid CA-file I/O for non-matching targets.** `broadcastToTarget` previously called `kCtx.OidcConfig()` (which can read `idp-certificate-authority` from disk) for every OIDC-ish context. It now does a cheap issuer/client-id pre-check via a new `oidcIssuerAndClientID` helper that reads `OidcConf` / `AuthProvider.Config` directly with no I/O, and only calls `OidcConfig()` for a target that actually matches the source — so a large multi-context kubeconfig no longer pays the CA-file cost per non-matching context during `/oidc-callback`. The unusable-CA skip behavior is preserved (the `target with unreadable OIDC CA is skipped` test still passes, because that target's issuer/client-id match and it's the `OidcConfig()` validation that fails).

**Docs + flag help wording — refresh trigger.** Reworded both the flag help text and `docs/installation/in-cluster/oidc.md` (and the `broadcastOIDCToken` godoc for consistency): the divergence trigger is per-cluster refresh, not "the source token refreshing." Sibling cookies diverge as soon as *any* cluster refreshes its token.

**Flag help — token type.** Called out that broadcast carries whichever token is in use — the `id_token`, or the `access_token` when `--oidc-use-access-token` is set — rather than implying id_token specifically.

All three threads resolved. Verified locally on the rebased branch:
- `go build ./...` clean
- `go test ./cmd/ -run TestBroadcastOIDCToken -v` → 12/12 PASS
- `golangci-lint run --path-mode=abs --timeout 5m` → 0 issues

Naga15:
Both @copilot logging-consistency points addressed:

- The source-context fetch error now uses the `sourceCluster` field key (was the generic `cluster` key), matching the rest of the broadcast helpers so log queries/alerts can filter on one consistent field.
- The `GetContexts()` failure log now includes `sourceCluster` in its structured fields (was `nil`), so it's clear which login triggered a failed broadcast attempt when multiple users/clusters are active.

Both threads resolved. `go test ./cmd/ -run TestBroadcastOIDCToken` → 12/12 PASS, `golangci-lint` → 0 issues.

Naga15:
Pushed two follow-up commits (`b989d24..6fdedfc`) from another pass over the broadcast logic:

1. **Document the access-token audience limitation.** The precondition justifies broadcasting on the basis that the token's `aud` defaults to the client-id. That holds for the `id_token`, but with `--oidc-use-access-token` the broadcast carries the `access_token`, whose audience is provider-specific and frequently *not* the client-id (e.g. a resource/API identifier on Okta, Entra ID, Auth0). Matching issuer+client-id therefore doesn't by itself guarantee the target apiserver accepts the token. Added this to the `broadcastOIDCToken` godoc and the OIDC install docs, alongside the existing `--oidc-extra-audience` note.

2. **Harden the broadcast tests.** They previously asserted only *which* clusters received a cookie, not the cookie's value or path — so a regression that broadcast an empty/truncated/wrong token, or mis-scoped the path, would still have passed. They now reassemble the chunked `headlamp-auth-<cluster>` cookie and assert `value == source token` and `path == /clusters/<target>`. (Confirmed this catches a regression: tampering the broadcast token makes the suite fail.)

Two smaller things I noticed but did **not** change — happy to address either if you'd like:

- **Issuer matching is byte-exact.** `OidcConfig()` does no normalization, so `https://idp.example.com` vs `https://idp.example.com/` (trailing slash), or a case difference across hand-edited kubeconfigs, will silently not broadcast. It's fail-safe (never over-broadcasts), but operators may hit "why didn't it broadcast?" with no signal. Worth normalizing the issuer before comparing, or at least documenting the exact-match requirement?
- **Security note in the docs.** Broadcast changes token delivery from user-initiated to automatic for any context sharing issuer+client-id. Kubeconfig is trusted operator config so this is low severity, but since the flag is fleet-wide it may be worth one sentence noting that any matching context receives the token on every login.

Also, as noted in the PR, broadcast-on-refresh remains a follow-up (the refresh path in `pkg/auth` refreshes each cluster independently); I have a design for it and can open it as a separate PR once this lands.

docs/installation/in-cluster/oidc.md
```diff
@@ -78,6 +78,24 @@ By default, headlamp leverages the `id_token` provided back from the OIDC Provid
...
+
+**Scope.** Broadcasting fires at initial OIDC login and broadcasts whichever token is in use (the `id_token`, or the `access_token` when `-oidc-use-access-token=true`). Token-refresh broadcasting is tracked as a follow-up. Because token refresh happens independently per cluster, sibling cookies diverge as soon as any one cluster refreshes its token; until refresh broadcasting lands, the affected clusters fall back to per-cluster re-login (commonly ~1h on EKS / Okta with default settings).
+
+**Caveats to be aware of before enabling.**
+
+- The flag is **disabled by default**; existing deployments see zero behavior change.
+- Audience overrides via `--oidc-extra-audience` on a target apiserver are not detected here. If you've configured `--oidc-extra-audience` on one cluster but not another, the broadcast cookie may be set but the target apiserver could reject the token. Align deployment configuration in that case.
```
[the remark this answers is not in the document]
Naga15: Thanks for catching that this was imprecise — though it turns out neither `--oidc-extra-audience` (singular) nor `--oidc-extra-audiences` (plural) is a kube-apiserver flag. The only audience-related flag is `--api-audiences` (for service-account token binding). A target apiserver's accepted OIDC audiences come from its `--oidc-client-id` or the `audiences` list in a structured [AuthenticationConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration). Reworded the godoc and docs to reference that mechanism instead in f2848c8.

backend/cmd/headlamp.go
```diff
@@ -1162,6 +1168,196 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
...
+//
+// Precondition: the kube-apiserver of each target cluster must trust the same
+// OIDC application (issuer + client-id) as sourceCluster. For the default
+// id_token the audience (aud) claim is the client-id, so broadcasting only when
+// both issuer and client-id match keeps the token valid against the target
+// cluster's apiserver without re-validating per cluster. Two audience caveats
+// are NOT detected here and must be ensured by deployment configuration:
+//   - --oidc-extra-audience on a target apiserver: the broadcast cookie may be
+//     set but the target apiserver could reject the token.
+//   - --oidc-use-access-token: the broadcast then carries the access_token,
```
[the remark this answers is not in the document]
Naga15: Thanks for catching that this was imprecise — though it turns out neither `--oidc-extra-audience` (singular) nor `--oidc-extra-audiences` (plural) is a kube-apiserver flag. The only audience-related flag is `--api-audiences` (for service-account token binding). A target apiserver's accepted OIDC audiences come from its `--oidc-client-id` or the `audiences` list in a structured [AuthenticationConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration). Reworded the godoc and docs to reference that mechanism instead in f2848c8.

illume (changes requested):
Thanks for the contribution.


Naga15:
Thanks @illume — done. All open Copilot comments are now addressed and marked resolved (0 open threads).

The last two were about the audience caveat referencing `--oidc-extra-audience`. On checking, neither that nor the plural `--oidc-extra-audiences` is actually a kube-apiserver flag — a target apiserver's accepted OIDC audiences come from its `--oidc-client-id` or the `audiences` list in a structured [AuthenticationConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration). So rather than the singular→plural swap, I reworded the godoc and docs to reference that mechanism instead (f2848c8), and replied on both threads.

The earlier merge conflict is also resolved (PR is mergeable). Locally, `go test ./cmd/ -run TestBroadcastOIDCToken` passes and `go vet` / `gofmt` are clean. Ready for another look when you have a chance.

illume:
Could you please answer this in the PR description: 

I wonder if we need "oidc-use-token-broadcast"? Could we instead make this the default behaviour and have no flag.
What are the downsides of having it?

backend/cmd/headlamp.go
```diff
@@ -1162,6 +1168,198 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	return r
 }
 
+// isOIDCAuthContext reports whether the kubeconfig context's auth-provider is
+// OIDC. kubeconfig.Context.AuthType() is too coarse: it returns "oidc" for any
+// AuthProvider (e.g. gcp, azure), so we check AuthProvider.Name explicitly to
+// avoid treating non-OIDC contexts as broadcast candidates. Note that exec
+// credential plugins live on AuthInfo.Exec, not AuthInfo.AuthProvider, and are
+// not relevant here.
+func isOIDCAuthContext(kCtx *kubeconfig.Context) bool {
```
illume: Is it possible to move these changes to the auth package?

We're slowly moving as much stuff out of headlamp.go as possible.

illume (changes requested):
Thanks for this.

I left some notes...

Naga15:
@illume good question. The downsides of making broadcast the default (dropping the flag):

- **Wider credential distribution by default.** Every multi-cluster OIDC user would start getting an auth cookie set for *all* sibling clusters on each login — a silent behavior change for existing deployments that never opted in.
- **Blast radius.** Any context matching issuer+client-id (including a misconfigured or less-trusted one in the kubeconfig) would automatically receive the token. Keeping it opt-in preserves a least-privilege default for a feature whose whole job is distributing a credential more widely.
- **The `--oidc-use-access-token` caveat would affect everyone.** With access tokens the `aud` is provider-specific and often isn't the client-id, so a sibling apiserver can reject the broadcast cookie — surfacing that for all users rather than only those who chose the behavior.
- **Cookie volume.** Large multi-cluster kubeconfigs emit many `Set-Cookie` headers per login; default-on approaches browser/proxy cookie limits sooner.
- **Consistency.** The related sibling OIDC behaviors (`oidc-use-cookie`, `oidc-use-access-token`) are all opt-in flags today.

The upside of default-on is real too: one fewer knob, and the token is already valid at the siblings (no privilege escalation — the user could log into those clusters directly).

On balance I leaned opt-in so the default stays conservative and existing deployments are unaffected — but I'm happy to make it the default (dropping the flag) if you'd prefer. What's your call? I'll fold the outcome and this rationale into the PR description.

backend/cmd/headlamp.go
Naga15: Yes, that makes sense. I'll move `broadcastOIDCToken` and its helpers (`isOIDCAuthContext`, `oidcIssuerAndClientID`, `broadcastToTarget`, `loadBroadcastSourceOIDCConfig`) plus their tests out of `headlamp.go` into `pkg/auth`, and have the `/oidc-callback` handler call the exported function.

That also sets up the planned broadcast-on-refresh follow-up nicely, since the refresh middleware (also in `pkg/auth`) can then reuse the same helper.

I'll do it in the next push — and if you decide broadcast should be default rather than flagged, I'll fold that into the same refactor so it's a single change.

Naga15:
One more design angle on "flag vs default", since it shapes the answer:

IdP-sharing isn't the same as "should share a session." Several clusters can be registered with the same OIDC app yet still be things you'd want kept as separate sessions — e.g. prod and dev on the same IdP, or (per @harse's 10+ cluster setup in #4283) a large fleet where you wouldn't want one login seeding cookies across everything.

So beyond a global on/off, a natural refinement is to scope broadcast to a **named group of contexts** — opt-in/opt-out per group, rather than every context that happens to match issuer+client-id. Roughly one of:

- a group map, e.g. `--oidc-token-broadcast-groups=prod:ctxA,ctxB;dev:ctxC`, or
- a per-context tag carried in the kubeconfig (a `headlamp` extension), so group membership travels with the context.

Broadcast would then only reach siblings in the source's group.

My suggestion: keep **this** PR as the simple opt-in flag (broadcast to all issuer+client-id matches) — it's the MVP and unblocks the multi-cluster case — and add **group-scoping as a follow-up** for fleets that want finer control. That also keeps the default conservative, which is really the core reason I'd keep it opt-in rather than default-on: even matching clusters shouldn't automatically share a session unless the operator says so.

Does group-scoping match how you'd want operators to control this? Happy to spec it out in #4283 if that's a useful direction.

Naga15:
@illume — folded the flag-vs-default answer into the PR description as you asked, under a new **"Why a flag rather than default-on?"** section (sorry, I'd originally answered in-thread rather than in the description).

Short version: I'd keep the flag, because IdP-sharing isn't the same as "should share a session" — prod and dev can sit behind the same OIDC app, and in a large fleet (e.g. @harse's 10+ cluster setup in #4283) one login shouldn't seed cookies across everything. Opt-in also keeps `--oidc-use-access-token`'s audience caveat and the cookie-volume limits contained to operators who chose the behavior. The description lists the full trade-off, including the honest case *for* default-on.

**That said — this is your call, and I'll implement whichever you prefer.** If you'd rather drop the flag and make broadcast the default, say the word and I'll push that change.

Current state: all Copilot threads resolved (0 open), CI green across the board, no conflicts. Ready for another look whenever you have a moment.

backend/cmd/headlamp.go
Naga15: Done — moved the whole broadcast implementation into `pkg/auth` (new `broadcast.go`), exported as `BroadcastOIDCToken` following the package's existing params-struct convention (like `RefreshAndSetTokenParams`). The tests moved with it to `pkg/auth/broadcast_test.go` (`auth_test` package, exercising only the exported API).

`cmd/headlamp.go` now contains just the call site in the OIDC callback handler (+12 lines total); the `HeadlampConfig` methods, helpers, and the broadcast-specific log-field constants are gone from it.

Also squashed the branch back to two atomic commits (implementation, tests) per the commit guidelines, since the later fix-up commits were rewriting earlier ones.

Verified locally: `go build ./...` clean, `go test ./pkg/auth ./cmd` all pass, `golangci-lint` 0 issues.

Naga15:
@illume — both of your July 6 notes are now addressed:

- **Moved to the auth package** — the whole implementation (including `isOIDCAuthContext`) now lives in `pkg/auth/broadcast.go`, exported as `BroadcastOIDCToken` following the package's existing `…Params` convention. `cmd/headlamp.go` is just the call site now. Tests moved with it to `pkg/auth/broadcast_test.go`.
- **Flag vs. default** — folded the rationale into the PR description under "Why a flag rather than default-on?" (short version: opt-in keeps the default conservative because IdP-sharing isn't the same as "should share a session").

CI is green across the board and Copilot's latest pass found nothing blocking. Whenever you have a moment, this should be ready for another look — and if you'd still prefer broadcast be the default rather than a flag, I'm happy to make that change.

KhaledELG:
@illume, it would be great if this feature were implemented 👍
Thanks @Naga15 for your work on this. I’m in the same situation as you.

backend/cmd/headlamp.go
Naga15: Marking this one resolved since the move is complete — the implementation now lives in `pkg/auth/broadcast.go` (exported as `BroadcastOIDCToken`, with `isOIDCAuthContext` and the helpers alongside it), the tests moved to `pkg/auth/broadcast_test.go`, and `cmd/headlamp.go` is down to just the call site in the OIDC callback handler.

Happy to reopen or reshape it if you'd like the split done differently.

Naga15:
@KhaledELG thanks for chiming in — useful to know the multi-cluster/one-IdP setup is hitting others the same way. That's now a few of us on this thread and in #4283 (@gadzillos, @harse, @minimalmattt, @mw-0) running the same shape.

@illume — short status so this is easy to pick back up, since both of your July 6 notes are done:

- **Auth package move** — the feature now lives in `pkg/auth/broadcast.go` as `BroadcastOIDCToken`; `cmd/headlamp.go` is just the call site. Tests in `pkg/auth/broadcast_test.go`. (Resolved that thread just now.)
- **Flag vs. default-on** — rationale is in the PR description under *"Why a flag rather than default-on?"*.

Current state: CI green across all 9 checks, no conflicts, all review threads resolved.

On flag-vs-default, my recommendation is to keep it opt-in — sharing an IdP isn't the same as "should share a session" (prod and dev often sit behind the same OIDC app, and in a 10+ cluster fleet one login shouldn't seed cookies everywhere). **Unless you'd prefer default-on, I'll treat that as settled** — and if you do want it default, say so and I'll push that change.

Would you be able to give it another look when you have a window? And if you're stretched, would you be open to assigning or suggesting a second reviewer? Either works — I just don't want it sitting on you alone.

backend/pkg/auth/broadcast.go
```diff
@@ -0,0 +1,235 @@
...
+	kContexts, err := params.KubeConfigStore.GetContexts()
+	if err != nil {
+		logger.Log(logger.LevelError, map[string]string{logFieldSourceCluster: params.SourceCluster}, err,
+			"failed to get contexts for broadcasting OIDC token")
+
+		return
+	}
+
+	for _, kCtx := range kContexts {
+		if kCtx.Name == params.SourceCluster || !isOIDCAuthContext(kCtx) {
```
gambtho, line 121: Could we skip `kCtx.Internal` here, like `getClusters()` does? `GetContexts()` also includes other users’ private stateless contexts, not just the configured clusters.

There’s a cookie collision here too: an internal key like `prod\x00-other` becomes `prod-other` after cookie sanitization. That can overwrite the real `prod-other` cluster’s cookie even if its issuer/client-ID didn’t match.

Could you add a test covering internal-context exclusion and that collision?

backend/cmd/headlamp.go
```diff
@@ -1118,6 +1118,18 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		// Set auth cookie
 		auth.SetTokenCookie(w, r, oauthConfig.Cluster, rawUserToken, config.BaseURL, config.SessionTTL)
 
+		if config.OidcUseTokenBroadcast {
+			auth.BroadcastOIDCToken(auth.BroadcastOIDCTokenParams{
+				Writer:          w,
```
gambtho, line 1249: Could we also invalidate the sibling clusters’ `clusterMe` queries when login completes? The frontend currently only invalidates the cluster we logged into.

If I visit B as Alice, then log into A as Bob, broadcast replaces B’s cookie but the UI can still show Alice from its five-minute cache. Requests to B now run as Bob.

Invalidating the whole `['clusterMe']` prefix would cover this. A test for switching accounts across two clusters would help too.

backend/pkg/auth/broadcast.go
Naga15: Good catch — both parts of this were real. Fixed in the latest push:

**Internal contexts skipped.** The broadcast loop now skips `kCtx.Internal`, matching `getClusters()`. Per-user stateless contexts are private to one user and should never receive another user's token, and this also removes the collision path entirely since we never call `SetTokenCookie` for one.

**The collision.** Confirmed exactly as you described: `AddContextWithKeyAndTTL` sets `Name` to the `"<cluster>\x00<user>"` key, and `SanitizeClusterName` strips the `\x00`, so `prod\x00-other` and a real `prod-other` share `headlamp-auth-prod-other`. Before the fix a matching internal context could overwrite the real cluster's cookie even when the real cluster's issuer/client-id didn't match.

Tests added in `pkg/auth/broadcast_test.go`:
- Table case: `internal (per-user stateless) target is skipped even when issuer and client-id match`
- `TestBroadcastOIDCToken_InternalContextCannotClobberRealClusterCookie` — real `prod-other` on a *different* issuer, internal `prod\x00-other` on the *matching* issuer; asserts no `prod-other` cookie is set. I verified it fails against the pre-fix code (it reported `got: [prod-other]`) and passes with the skip in place.

Comment on the skip explains the `\x00` sanitization reason so it doesn't get "simplified" away later.

backend/cmd/headlamp.go
Naga15: Agreed — with broadcast in play, invalidating only the logged-into cluster leaves siblings serving a stale identity from cache (TopBar's `clusterMe` has a 5-minute `staleTime`) while requests already run as the new user. Fixed in a separate `frontend:` commit:

- Added `invalidateClusterUserInfo()` in `lib/auth.ts`, which invalidates the whole `['clusterMe']` prefix (no `exact`), with a doc comment explaining the broadcast reason.
- `handleOidcAuth` in the auth chooser now calls it instead of the single-cluster `exact: true` invalidation.

Test in `lib/auth.test.ts` models your scenario: identity cached for cluster B (and A) as `alice`, then the login-complete path runs; asserts both `['clusterMe','cluster-a']` and `['clusterMe','cluster-b']` are `isInvalidated`, and that an unrelated `['auth', …]` entry is untouched.

I kept it unconditional rather than gating on whether broadcast is enabled — the frontend doesn't know the backend flag, a refetch of the visible clusters' `/me` is cheap, and refreshing every cluster's identity after any OIDC login is the correct behavior regardless.

Naga15:
Rebased onto current `main` to clear the merge conflict — no functional changes from my side.

The only conflict was in `authchooser/index.tsx`, where `main` added the `oidc-login-attempted` sessionStorage tracking (#2848) inside the same `handleOidcAuth` handler this PR touches. Resolved by keeping that logic as-is and replacing only the single-cluster `queryClient.invalidateQueries({ queryKey: ['clusterMe', clusterName], exact: true })` with the prefix-wide `invalidateClusterUserInfo()` from @gambtho's review:

```ts
handleOidcAuth={() => {
  if (clusterName) {
    try {
      sessionStorage.setItem(`oidc-login-attempted.${clusterName}`, 'true');
    } catch {
      // sessionStorage unavailable (e.g. private browsing with strict settings).
    }
  }
  // Invalidate every cluster's cached identity, not just this one: the
  // backend may have broadcast the new token to sibling clusters.
  invalidateClusterUserInfo();
  history.replace(from);
}}
```

Re-verified on the new base: `go build ./...` clean, `go test ./pkg/auth ./cmd` pass, `golangci-lint` 0 issues, frontend `vitest src/lib/auth.test.ts` 8/8, eslint + prettier + `tsc` clean.

Naga15:
CI caught a real problem with my previous push — fixed, and worth explaining rather than just re-pushing.

`src/plugin/pluginLib.test.ts` failed with:

```
Snapshot `pluginLib variable > should stay the same for plugin compatibility` mismatched
      "auth": {
+       "invalidateClusterUserInfo": [Function],
```

`lib/util.ts` does `export * as auth from './auth'`, and `util.ts` is exposed to plugins as `Utils`. So putting `invalidateClusterUserInfo` in `lib/auth.ts` silently added it to the **public plugin API**.

Updating the snapshot would have been the wrong fix. The sibling test in that same file is explicit about the policy:

> `should not expose queryClient to plugins` — *"queryClient is an internal singleton that plugins must not access directly. Exposing it lets plugins invalidate or mutate queries belonging to other plugins or to Headlamp core, breaking isolation…"*

A function that invalidates the whole `['clusterMe']` prefix across every cluster is precisely that capability, so it shouldn't be reachable from plugins either.

**Fix:** moved `invalidateClusterUserInfo` from `lib/auth.ts` into `lib/queryClient.ts`, which already owns the singleton and is not re-exported to plugins. Its test moved to `lib/queryClient.test.ts` alongside it. The doc comment now records why it lives there, so it doesn't get "tidied" back into `lib/auth` later.

Net effect:
- `lib/auth.ts` and `lib/auth.test.ts` are back to **zero diff vs `main`**
- `__snapshots__/pluginLib.snapshot` is **unmodified** — the plugin API surface is unchanged by this PR
- The frontend commit is now just 3 files: `queryClient.ts`, `queryClient.test.ts`, `authchooser/index.tsx`

Verified locally: `vitest src/plugin/pluginLib.test.ts` passes (the exact CI failure), `vitest src/lib/queryClient.test.ts src/lib/auth.test.ts` 23/23, eslint + prettier + `tsc` clean, and the backend is untouched by this change.

gambtho (approved):

backend/pkg/config/config.go
```diff
@@ -665,6 +666,26 @@ func addOIDCFlags(f *flag.FlagSet) {
 	f.String("oidc-ca-file", "", "CA file for OIDC")
 	f.Bool("oidc-use-access-token", false, "Setup oidc to pass through the access_token instead of the default id_token")
 	f.Bool("oidc-use-cookie", false, "Enable OIDC cookie usage even when not running in-cluster")
+	f.Bool("oidc-use-token-broadcast", false,
```
yolossn: IMHO the arg description is very big and should be made crisp. The details can be documented somewhere and referenced in the help text.

Naga15:
Good call — it had grown to ~1500 characters across review rounds, about 9x the next-longest flag in `config.go`. Trimmed to the essentials with a pointer to the docs:

```
After a successful OIDC login, also set the auth cookie for other kubeconfig
contexts whose OIDC auth-provider has the same idp-issuer-url and client-id.
Disabled by default. Preconditions and caveats:
https://headlamp.dev/docs/latest/installation/in-cluster/oidc/
```

Now 264 chars, in line with the other flags here (next-longest is `allow-kubeconfig-changes` at 170).

Nothing was lost — every detail that was inline (the issuer/client-id precondition, login-vs-refresh scope, the audience caveats including the `access_token` case, cookie count/size limits, and the pre-existing chunk-cookie limitation) is already written up in the "Multi-cluster: broadcast the OIDC token across sibling clusters" section this PR adds to `docs/installation/in-cluster/oidc.md`, which the help text now links to.

Only `config.go` changed since @gambtho's approval — the diff against the reviewed commit is this one flag's help string.

backend/pkg/config/config.go
Naga15: Trimmed — it had grown to ~1500 characters across review rounds, about 9x the next-longest flag in this file. Now 264 chars, in line with the others (next-longest is `allow-kubeconfig-changes` at 170), with the detail moved behind a link:

```
After a successful OIDC login, also set the auth cookie for other kubeconfig
contexts whose OIDC auth-provider has the same idp-issuer-url and client-id.
Disabled by default. Preconditions and caveats:
https://headlamp.dev/docs/latest/installation/in-cluster/oidc/
```

Nothing was lost: the precondition, the login-vs-refresh scope, the audience caveats (including the `access_token` case), the cookie count/size limits, and the pre-existing chunk-cookie limitation are all already in the "Multi-cluster: broadcast the OIDC token across sibling clusters" section this PR adds to `docs/installation/in-cluster/oidc.md`, which the help text now points at.

PR #5946, opened 2026-06-07
app: runCmd: runPlugin: Fix for missing commands

illume:
- **app: runCmd: runPlugin: Allow running az and gh commands**
- **app: runCmd: Handle child process error events**

### DIFF
--- a/app/electron/runCmd.test.ts
+++ b/app/electron/runCmd.test.ts
@@ -14,13 +14,26 @@
  * limitations under the License.
  */
 
+import { EventEmitter } from 'events';
 import path from 'path';
 import { afterEach, beforeEach, describe, expect, it, Mock, vi } from 'vitest';
-import { checkPermissionSecret, validateCommandData } from './runCmd';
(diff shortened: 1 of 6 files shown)


ashu8912 (commented):
Hii @illume  thanks for the PR, I was thinking if we should make this config based, so anybody who is running headlamp can pass a config(How i am not sure) and that run of headlamp app would allow those commands, otherwise we would keep on modifying headlamp core to allow certain commands.

ashu8912 (approved):
LGTM!!

ashu8912:
> Hii @illume thanks for the PR, I was thinking if we should make this config based, so anybody who is running headlamp can pass a config(How i am not sure) and that run of headlamp app would allow those commands, otherwise we would keep on modifying headlamp core to allow certain commands.

Discussed this with rené offline and this is part of broader discussion to make things like this scale, right now this one is safer to get in, as it allows these commands for minikube and ai-assistant plugin

PR #5950, opened 2026-06-07
Frontend: use lazy useState in NumRowsInput

### LINKED ISSUE #5897: frontend: Refactor state initialization to lazy state initializers in NumRowsInput
### Description
In `frontend/src/components/App/Settings/NumRowsInput.tsx`, there are redundant `useMemo` hooks used solely to calculate the initial values of `selectedValue` and `customValue` state variables. 

These `useMemo` hooks use variables like `options` but pass an empty dependency array `[]` (forcing calculation only on mount). This violates the ESLint rule for hooks, requiring the use of `// eslint-disable-next-line react-hooks/exhaustive-deps` hacks.

### Proposed Improvement
Using `useMemo` to initialize state is a React anti-pattern. Instead, we can use React's standard **Lazy State Initializer** pattern. By passing a initialization function directly inside `useState(() => { ... })`, we can:
1. Ensure the computation runs exactly once on mount.
2. Completely remove the redundant `useMemo` hooks.
3. Clean up the codebase by removing the `eslint-disable` workarounds.

### Expected Behavior
The settings UI works exactly as before, but with cleaner code, better performance, and standard-compliant React hooks usage.

piyusshkumar:
## Summary

Refractor NumRowsInput to use Reacts lazy state initialization pattern instead of using useMemo for initial state values.

## Related Issue
Fixes #5897 

## Changes
-Replaced useMemo based initial state calculations with lazy useState(() => ... ) initializers .
-Removed unused useMemo hook
-Removed unnecessary eslint-disabled-next-line-react-hooks/exhaustive-deps commits.

## Validation

1. ran npm  run lint
2. ran npm run test

## Notes for the Reviewer
This is a refracting changes only and does not modify the component behavior.


frontend/src/components/App/Settings/NumRowsInput.tsx
```diff
@@ -48,24 +48,22 @@ export default function NumRowsInput(props: { defaultValue: number[]; nameLabelI
...
-  }, []);
-  const [selectedValue, setSelectedValue] = useState(defaultRowsPerPageValue);
-  const storedCustomValue = useMemo(() => {
+  });
+
+  const [customValue, setCustomValue] = useState(() => {
     const val = options.find(val => !defaultTableRowsPerPageOptions.includes(val));
     if (!val) {
-      return defaultTableRowsPerPageOptions[0];
+      return defaultTableRowsPerPageOptions[0].toString();
```
ashu8912, line 63: Why this change??

ashu8912 (commented):
Overall LGTM, added a comment for a change that this brings to how we use defaulRowsPerPageValue

frontend/src/components/App/Settings/NumRowsInput.tsx
piyusshkumar: Earlier,  customvalue was initialized with `storedCustomvalue.toString(),` so the state value was always a string. 
After removing `storedCustomvalue` , i added `.toString()` so that customvalue continues to be always a string. to keep the behavior same after removing `storedCustomvalue`.
Happy to change , if i am missing something.

skoeva (changes requested):
change looks fine, just retitle the PR/commit to something like `frontend: Use lazy useState in NumRowsInput`

piyusshkumar:
@illume I have addressed the feedback.  Please review .

Issue #5702, opened 2026-05-16
frontend: Resource search input loses focus after table refresh


prisha-sh:
## Current Behavior

When typing into a resource search/filter input, the field occasionally loses focus after automatic table refreshes.

This interrupts typing and forces users to click back into the input repeatedly.

## Expected Behavior

The search field should maintain focus during background refreshes unless explicitly blurred by the user.

## Steps to Reproduce

1. Open a resource list page
2. Start typing into the search/filter input
3. Wait for automatic refresh
4. Observe focus loss

## Environment

- Headlamp version: v0.42.0
- Browser: Chrome
- OS: Windows 11

prisha-sh:
Hi @illume sir, I’m interested in working on this issue and would like to contribute a fix for it.  
I’ve started looking into the related components already. Please assign this issue to me if it’s available. Thanks!

Issue #5701, opened 2026-05-16
frontend: Resource tables lose scroll position after auto refresh


prisha-sh:
## Current Behavior

While viewing long resource lists such as Pods or Deployments, the table periodically refreshes and resets the scroll position back to the top.

This becomes disruptive when monitoring workloads or inspecting resources in large clusters because users lose their place after every refresh cycle.

## Expected Behavior

The current scroll position should be preserved after automatic data refreshes.

## Steps to Reproduce

1. Open a cluster with a large number of Pods
2. Scroll down in the Pods table
3. Wait for the automatic refresh interval
4. Observe that the table jumps back to the top

## Environment

- Headlamp version: v0.42.0
- Browser: Chrome
- OS: Windows 11

## Possible Direction

The table component could preserve and restore the scroll offset during polling based updates.

prisha-sh:
Hi @illume sir, I’m interested in working on this issue and would like to contribute a fix for it.  
I’ve started looking into the related components already. Please assign this issue to me if it’s available. Thanks!

PR #5952, opened 2026-06-08
frontend: ResourceTable: Stable row id when metadata.uid is missing

### LINKED ISSUE #5707: frontend: Table row selection clears after periodic data refresh
## Current Behavior

Selected rows in resource tables become deselected after automatic refreshes.

This interrupts workflows involving batch operations.

## Expected Behavior

Selections should remain preserved as long as the resources still exist in the refreshed dataset.

## Steps to Reproduce

1. Select multiple resources in a table
2. Wait for automatic refresh
3. Observe selections disappearing

WasThatRudy:
## Summary

Fixes #5707. Row selections in resource tables vanish after every polling refresh whenever any visible row lacks a `metadata.uid`.

`getRowId` was `item => item?.metadata?.uid`. When that returns `undefined`, Material React Table falls back to the row index for identity. After the throttled refetch hands a new `data` array down (different order or length), index-based ids no longer point at the same rows and the selection map effectively resets. Rows without a uid (events, virtual rows, synthetic items) also collapse onto a single `undefined` id, which causes duplicate-key warnings on top of broken selection.

## Approach

Extract the id logic into a leaf module `getResourceRowId(item)` with no `lib/k8s` dependency:

- Use `metadata.uid` when present (the common path, unchanged behaviour).
- Fall back to `cluster/namespace/name` so resources without a uid still get a stable, distinct identifier across polling refreshes.
- Include `cluster` so identically-named resources from different clusters stay distinct in multi-cluster views.

`ResourceTable` now points `getRowId` at the helper directly, which also gives the prop a stable function identity across renders.

## Test plan

- [x] New unit test `getResourceRowId.test.ts` covers: uid present, uid missing (composite fallback), cross-cluster distinctness, idempotent calls (selection stability), `null`/`undefined`/empty inputs, missing-metadata path, and cluster-scoped resources. 7/7 pass.
- [x] Full `src/components/common/Resource` suite: 95/95 pass (including the export-tracking `index.test.ts`, where the new private helper is added to `avoidCheck`).
- [x] `npm run tsc` clean.
- [x] `npx eslint -c .eslintrc.ci.cjs --max-warnings 0` clean over the touched files and the whole `Resource` directory.
- [ ] Manual: select rows in any resource table, wait for the auto-refresh interval, confirm the same rows stay selected.

## Notes for the reviewer

- Scope is intentionally narrow. Two sibling reports (#5701 scroll position, #5702 search input focus) describe the same "resets on auto-refresh" pattern but their likely root causes (pagination toggle remounting the body, focus loss on rerender) are independent and would need their own investigation in a running cluster.

### DIFF
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -59,6 +59,7 @@ import { getA8RMetadata } from './A8RInfo';
 import DeleteButton from './DeleteButton';
 import DownloadButton from './DownloadButton';
 import EditButton from './EditButton';
+import { getResourceRowId } from './getResourceRowId';
 import ResourceTableMultiActions from './ResourceTableMultiActions';
 import { RestartButton } from './RestartButton';
 import ScaleButton from './ScaleButton';
@@ -718,7 +719,7 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
         }}
         globalFilterFn="kubeObjectSearch"
         filterFunction={filterFunc}
-        getRowId={item => item?.metadata?.uid}
+        getRowId={getResourceRowId}
       />
     </>
   );
--- a/frontend/src/components/common/Resource/getResourceRowId.test.ts
+++ b/frontend/src/components/common/Resource/getResourceRowId.test.ts
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


WasThatRudy:
@illume addressed both @Copilot comments in `b9f0a5c4b`: `getResourceRowId` now takes MRT's row index as a second argument and uses `row-${index}` as the last-resort fallback, so null/empty inputs and items with no identifying metadata get distinct ids instead of collapsing onto `''` or `'//'`. Tests updated to match the new signature (9/9 pass). Ready for another look.

illume (changes requested):
Thanks for these changes.


illume (approved):
🎉 thanks!

PR #5973, opened 2026-06-09
frontend: pluginLib: Expose stateless and queryClient to plugins

sniok:
Adding some missing things that are present in the pluginLib types but weren't actually exported


vyncent-t (approved):
LGTM

PR #5977, opened 2026-06-10
frontend: Replace native title tooltips with styled tooltips

### LINKED ISSUE #5960: UI: Redundant and unstyled tooltip appears when hovering over health status badges
### Describe the bug
When navigating the dashboard, hovering over various text elements, status badges (e.g., "Rebooted"), and table cells triggers an unnecessary, unstyled native tooltip. The tooltip simply duplicates the exact text already clearly visible on the screen. This appears to be a global frontend issue rather than an isolated component bug, leading to a cluttered and visually jarring user experience across multiple views.

### To Reproduce

- Start Headlamp locally using npm run start.

- Connect to any local cluster.

- Navigate to almost any view containing tables, lists, or status badges (e.g., Nodes, Pods, Events).

- Hover the mouse cursor over different text blocks, object names, or health status badges.

- Notice the large, redundant native popup reproducing the element's text over the cursor.

### Expected behavior
Tooltips should only appear if text is explicitly truncated (ellipses) or if they provide meaningful supplementary context (e.g., full timestamps, raw error codes). Fully visible standard text elements and badges should not trigger a redundant, unstyled tooltip on hover.

Environment:

Headlamp Version: Local development (Main branch)

Deployment: Local (npm run start)

OS: Fedora

Browser: (Add your browser here, e.g., Chrome / Firefox)

### Additional context
Because this behavior is observable across multiple distinct sections of the frontend, this strongly points to a widespread issue in a shared base component. Potential culprits include:

A high-level layout/wrapper component or a utility function automatically injecting native HTML title attributes into text nodes.

A global configuration issue with Material-UI Tooltip components wrapping broad element groups without checking if the text is truncated or requires a tooltip.

### Screenshots

Image

Image

xonas1101:
## What this does

Across the UI, several text elements set the native HTML `title` attribute, which the browser/OS renders as an unstyled tooltip that looks out of place next to Headlamp's own styled tooltips. This PR replaces those native tooltips with the styled `LightTooltip`, fixes the root prop-forwarding that leaked them, and tidies up tooltip sizing.

Fixes #5960 

### Changes

- **Root fix in `StatusLabel`** — it spread `{...other}` (including any `title`) onto its underlying `<span>`, forwarding a native tooltip. Now a `title` is routed through `LightTooltip` instead. This is what surfaced unstyled tooltips on status badges (including plugin-sourced ones).
- **Replaced native `title` tooltips with `LightTooltip`** on text/DOM elements in: Home `SquareButton`, advanced-search `SavedSearches`, cluster `Chooser`, `GlobalSearchContent`, resource-map `KubeObjectGlance`, `SelectionBreadcrumbs`, and `GroupNode`; deployments `List` conditions now pass the message through `StatusLabel`.
- **Reduced the oversized global tooltip font** — the global `MuiTooltip` override set the font to `1.3em` (~double MUI's default), making every tooltip look bloated. Lowered to `0.875rem`.
- **Removed a redundant always-on tooltip** on the Home cluster-table Name column that merely repeated the cluster name already shown beside the icon.

### Testing

- `tsc --noEmit`, eslint, and prettier all clean
- All 548 storyshot snapshot tests pass (snapshots updated where markup legitimately changed)
- Verified visually against a running cluster

### Screenshots
image
Screenshot From 2026-06-10 20-45-55

### DIFF
--- a/frontend/src/components/App/Home/ClusterTable.tsx
+++ b/frontend/src/components/App/Home/ClusterTable.tsx
@@ -18,7 +18,6 @@ import { Icon } from '@iconify/react';
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import { useTheme } from '@mui/material/styles';
-import Tooltip from '@mui/material/Tooltip';
 import Typography from '@mui/material/Typography';
 import {
   MRT_ColumnFiltersState,
@@ -260,17 +259,13 @@ export default function ClusterTable({
           Cell: ({ row: { original } }) => {
             const appearance = getClusterAppearanceFromMeta(original.name);
             return (
-              <Tooltip title={original.name} arrow>
-                <span>
-                  <Link routeName="cluster" params={{ cluster: original.name }}>
-                    <ClusterBadge
-                      name={original.name}
-                      icon={appearance.icon}
-                      accentColor={appearance.accentColor}
-                    />
-                  </Link>
-                </span>
-              </Tooltip>
+              <Link routeName="cluster" params={{ cluster: original.name }}>
+                <ClusterBadge
+                  name={original.name}
+                  icon={appearance.icon}
+                  accentColor={appearance.accentColor}
+                />
+              </Link>
             );
           },
         },
--- a/frontend/src/components/App/Home/ClusterTable.tsx
+++ b/frontend/src/components/App/Home/ClusterTable.tsx
@@ -259,13 +259,17 @@ export default function ClusterTable({
           Cell: ({ row: { original } }) => {
             const appearance = getClusterAppearanceFromMeta(original.name);
             return (
-              <Link routeName="cluster" params={{ cluster: original.name }}>
-                <ClusterBadge
-                  name={original.name}
-                  icon={appearance.icon}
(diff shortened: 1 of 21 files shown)


skoeva (changes requested):
could you address the copilot comments? it would also be nice if you could post screenshots to show the difference

xonas1101:
@skoeva I addressed the copilot comments and have added a couple of screenshots for the changes. As for the old state, you can check the issue linked in the PR body.

xonas1101:
@skoeva work's done. Would appreciate a review. PTAL!

xonas1101:
@illume Copilot comments fixed. PTAL!

illume (changes requested):
Thanks for this PR.


xonas1101:
@illume The comment was actually addressed. Just forgot to mark it resolved. PTAL!

xonas1101:
@illume Copilot seems to be satisfied with this. PTAL

illume (approved):
🎉 thanks!

(I love changes like this where it makes the UI better through consistency)

PR #5982, opened 2026-06-10
app: Add opt-in Chromium remote debugging port for devtools/MCP

yolossn:
# Summary 

Register a --remote-debugging-port flag on the Electron app and expose the Chromium remote-debugging endpoint only when it is explicitly passed. This lets developers attach Chrome DevTools tooling such as chrome-devtools-mcp to the running app. It is opt-in on purpose: enabling it by default would expose an unauthenticated debugging endpoint in production builds. Invalid or absent port values fall back to 9222.

# Example usage:
Start the app (full dev stack) with remote debugging enabled:

```bash
npm run start:with-app:debug
```

This launches the app with `--remote-debugging-port=9222`, exposing a Chrome
DevTools Protocol endpoint at `http://localhost:9222`.

Then register the endpoint with an MCP-capable client (e.g. Claude, Cursor) so
it can drive and inspect the running app via `chrome-devtools-mcp`:

```json
{
  "mcpServers": {
    "headlamp-electron-devtools": {
      "command": "npx",
      "args": ["-y", "chrome-devtools-mcp@latest", "--browser-url", "http://localhost:9222"]
    }
  }
}
```


sniok (approved):
nice!

can this be used with default devtools in chrome? it'd be nice to use devtools that have plugins like react/redux that bultin electron devtools don't have

yolossn:
>can this be used with default devtools in chrome? it'd be nice to use devtools that have plugins like react/redux that bultin electron devtools don't have

Yes, this works with default Chrome DevTools too. I'm not sure Redux state is exposed via the MCP though since those panels come from extensions

PR #5985, opened 2026-06-10
frontend: Clean up dev-server build warnings

skoeva:
This change cleans up two noisy warnings in `frontend/`:

- **monaco-editor** `Critical dependency: require function is used...` — unreachable `require()` in the ESM build. Suppressed with a scoped `ignoreWarnings` entry.
- **Node `MaxListenersExceededWarning`** — `http-proxy-middleware` adds a `'close'` listener per proxy entry; 14 routes exceeded Node's default cap of 10. Collapsed into one HTTP + one WS proxy via `pathFilter`.

Dev-build only; no runtime impact on the packaged app.

### Testing

- `npm run star` from `app/` — both warnings gone.
- All proxied routes still work; Monaco editor still loads.


illume (approved):
Thanks! 🎉

PR #5989, opened 2026-06-11
frontend: plugin: Fix main branch tests, add test that implementation detail is not exported

illume:
## Summary

Fixes a pre-existing snapshot instability: `Lodash` serializes differently across environments (local macOS includes `module.exports`, CI Ubuntu does not), causing the `pluginLib.snapshot` test to fail on CI. Fixed by adding `Lodash` to the `externalLibs` list so it gets the same `"Present"/"Missing"` normalization as other bundled libraries.

Also stops exporting the internal `queryClient` singleton to plugins via `window.pluginLib`. Plugins that access the shared QueryClient can invalidate or mutate queries belonging to other plugins or Headlamp core, breaking isolation. A test is added to prevent re-exporting it in the future.

## Changes

- Added `Lodash` to the `externalLibs` normalization list to fix cross-platform snapshot instability
- Removed `queryClient` import and export from `frontend/src/plugin/index.ts`
- Added test in `pluginLib.test.ts`: `should not expose queryClient to plugins`
- Updated `pluginLib.snapshot` to reflect both changes

## Steps to Test

1. Run `cd frontend && npx vitest run src/plugin/pluginLib.test.ts`
2. Both tests should pass: the snapshot test and the new queryClient exclusion test
3. Verify no existing plugin in the repo imports `queryClient` from `pluginLib`

## Notes for the Reviewer

- No plugin in the repo uses `pluginLib.queryClient` — they all create their own `QueryClient` instances
- This is a breaking change only for third-party plugins that directly access `window.pluginLib.queryClient`, which is an undocumented internal
- The Lodash snapshot fix is independent of the queryClient change but included here since it affects the same snapshot file

### DIFF
--- a/frontend/src/plugin/__snapshots__/pluginLib.snapshot
+++ b/frontend/src/plugin/__snapshots__/pluginLib.snapshot
@@ -537,9 +537,7 @@
       "default": [Function],
     },
   },
-  "Lodash": Function {
-    "default": [Function],
-  },
+  "Lodash": "Present",
   "MonacoEditor": "Present",
   "MuiCore": "Present",
   "MuiLab": "Present",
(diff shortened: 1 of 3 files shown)


ashu8912 (approved):
LGTM!!!

PR #5994, opened 2026-06-11
frontend: node: Separate roles list with spaces

skoeva:
This change updates the roles column in the node list to join entries with `", "` instead of `","`, matching the formatting used on the node details page for a consistent look across views.

#### Before

image

#### After

image


sniok (approved):
thanks!

/lgtm

PR #5802, opened 2026-05-21
frontend: Headlamp web app crash on firefox browsers

DhairyaMajmudar:
## Summary

This PR fixes Headlamp web app crashing on firefox browsers

## Related Issue

Fixes #5801

## Changes

- Removed i18 normalisation due to which the headlamp app is crashing on firefox browsers

## Steps to Test

1. Run `npm run start`
2. Open `localhost:3000` on firefox browser window

## Screenshots

Screenshot-494

## Notes for the Reviewer

This is quite a major regression because currently, Headlamp web app will be crashing for all Firefox users. It would be helpful if this could be taken in P0 priority. Thank you!

cc: @illume @joaquimrocha

PR #6026, opened 2026-06-15
frontend: i18n: Add regression test for undefined locale handling

### LINKED ISSUE #6025: frontend: Add regression coverage for undefined locale handling in ThemeProviderNexti18n
# Describing the issue

`frontend/src/i18n/ThemeProviderNexti18n.tsx` currently has no dedicated regression test coverage for handling an undefined i18n language.

A previous bug caused the Headlamp web application to crash on Firefox when `locale` was undefined and `toLowerCase()` was called directly. This issue was fixed in PR #5802 by introducing a safe fallback:

```ts
const normalizedLocale = locale?.toLowerCase() ?? 'en';
```

While the bug has been fixed, there is currently no automated test that verifies this behavior. Without regression coverage, future refactors could accidentally reintroduce the crash.

# Expected behavior

There should be a regression test that verifies:

* `ThemeProviderNexti18n` handles an undefined i18n language without crashing.
* Child components continue to render successfully when the language is undefined.
* The fallback locale path remains protected against future regressions.

# Steps to reproduce

Revert the locale normalization logic in `ThemeProviderNexti18n.tsx` to call `locale.toLowerCase()` directly and run the i18n test suite.

Without the fix from PR #5802, rendering the provider with an undefined language causes a runtime error. A regression test should detect this behavior automatically.

mahesh-09-12:
## Summary

This PR adds regression test coverage for `ThemeProviderNexti18n` to prevent reintroducing a previously fixed crash caused by an undefined i18n language.

The added test verifies that the provider successfully renders its children when `i18n.language` is undefined, covering the fallback locale handling path introduced in PR #5802 to resolve the Firefox crash.

## Related Issue

Fixes #6025

Related to PR #5802 (`frontend: i18n: Fix crash on Firefox when locale is undefined`)

## Changes

* Added `frontend/src/i18n/ThemeProviderNexti18n.test.tsx`
* Added regression coverage for undefined locale handling
* Verified that `ThemeProviderNexti18n` renders successfully when `i18n.language` is undefined
* Added coverage for the fix introduced in PR #5802
* Prevented accidental reintroduction of the previously fixed Firefox crash scenario

## Steps to Test

1. Run:

   ```bash
   npm test -- src/i18n/ThemeProviderNexti18n.test.tsx
   ```

2. Verify the test passes.

3. Temporarily replace:

   ```ts
   const normalizedLocale = locale?.toLowerCase() ?? 'en';
   ```

   with:

   ```ts
   const normalizedLocale = locale.toLowerCase();
   ```

4. Re-run the test and verify it fails.

## Notes for the Reviewer

* This is a focused regression test covering the Firefox crash fixed in PR #5802.
* The test reproduces the undefined locale code path and confirms that the provider continues rendering successfully.
* Reverting the locale fallback logic causes the test to fail, demonstrating regression coverage for the original bug.
* The added test file passes locally, and lint checks for the file also pass.

### DIFF
--- a/frontend/src/i18n/ThemeProviderNexti18n.test.tsx
+++ b/frontend/src/i18n/ThemeProviderNexti18n.test.tsx
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createTheme } from '@mui/material/styles';
+import { render, screen } from '@testing-library/react';
+import { vi } from 'vitest';
+import ThemeProviderNexti18n from './ThemeProviderNexti18n';
+
+vi.mock('react-i18next', async () => {
+  const actual = await vi.importActual<typeof import('react-i18next')>('react-i18next');
+
+  return {
+    ...actual,
+    useTranslation: () => ({
+      i18n: {
+        language: undefined,
+        on: vi.fn(),
+        off: vi.fn(),
+      },
+      ready: true,
+    }),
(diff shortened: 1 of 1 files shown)


mahesh-09-12:
@illume thanks for taking a look. I've addressed the Copilot review suggestion regarding import ordering and updated the branch. I also rebased onto the latest main so the checks run against the current base branch.

mahesh-09-12:
@illume just following up on this PR - the latest Copilot review completed without any additional comments after the update.

illume (approved):
🎉 thanks!

It's great to have regression tests for bugs fixed.

PR #6057, opened 2026-06-16
frontend: useKubeObjectList: Refactor how request watching is being stopped

sniok:
Tested on flux plugin and various namespace scenarios

# how to test

- install flux, open overview page, it shouldn't crash
- go to any resource list (like deployments), select two namespaces, then clear all namespaces, there should be just one websocket connection for that resource


vyncent-t (approved):
Works on my end, doesn't crash

PR #6087, opened 2026-06-18
frontend: CustomResourceInstancesList: Wrap title and column labels with i18n

### LINKED ISSUE #6085: frontend: CustomResourceInstancesList: title and column labels not wrapped with i18n
## Describe the bug
On the Custom Resource Instances page (`frontend/src/components/crd/CustomResourceInstancesList.tsx`), a few strings are hardcoded in English instead of going through the `t()` i18n function: the page title "Custom Resource Instances" and the column headers "Instance name", "CRD", and "Categories". The rest of the component (loading/empty/error text) already uses `t('translation|...')`, so these were just missed. Because they aren't wrapped, they don't get extracted into the locale files and always render in English. Expected: they should translate along with the rest of the UI when the language is changed.

## To Reproduce
Steps to reproduce the bug:
1. Switch the language to a non-English locale in Settings (e.g. Español)
2. Open a cluster and go to Custom Resources
3. Open any CRD and view its instances list
4. The page title and the "Instance name" / "CRD" / "Categories" column headers are still in English while the surrounding UI is translated

 - Installation type: Mac (built from source / main)
 - Headlamp Version: latest (main)
 - Other: reproducible regardless of cluster, any CRD will do

Yes (I will propose a PR)

## Additional Context
Not a regression, the strings have been hardcoded since the component was added. Fix is to wrap them with the existing `t()` and run `npm run i18n` to pull the new keys into the locale files. "Categories" already exists as a translation key, so that one mainly needs the wrapping.

krsatyamthakur-droid:
## Summary

The Custom Resource Instances page had a few strings hardcoded in English instead of going through the `t()` i18n function, so they were never extracted into the locale files and stayed in English even when the rest of the UI was translated. This wraps them so they get translated like everything else.

## Related Issue

Fixes #6085

## Changes

- Wrapped the page title "Custom Resource Instances" and the column labels "Instance name", "CRD", and "Categories" with `t('translation|...')` in `CustomResourceInstancesList.tsx`, consistent with the other strings already translated in the same component
- Ran `npm run i18n` to extract the new keys into the locale files

## Steps to Test

1. Switch the language to a non-English locale in Settings (e.g. Español)
2. Open a cluster and go to Custom Resources, then open any CRD's instances list
3. Before this change the title and column headers stayed in English; now they pick up the selected language like the rest of the page

## Notes for the Reviewer

- This only touches the i18n layer (wrapping existing strings); no logic changed.
- "Categories" already existed as a translation key, so it just needed wrapping. The other three are new keys added with empty values for translators.


illume (approved):
🎉 thanks!

PR #6103, opened 2026-06-19
frontend: Fix /auth proxy path for OIDC callback and enforce LF line endings

jooskim:
## Summary

Fix the local dev proxy so the OIDC login flow completes without 404s, and enforce LF line endings for TypeScript/JavaScript/JSON files.

## Related Issue

N/A

## Changes

- **`frontend/rsbuild.config.ts` / `frontend/vite.config.ts`:** Change the proxy key for the auth route from `/auth` to `/auth/`. `/auth` is a React route that the Go backend's `/oidc-callback` handler redirects users to after a successful OIDC login. Without the trailing slash, the dev proxy was matching `/auth` as a prefix path and forwarding it to the backend instead of serving the frontend bundle, resulting in 404s after the OIDC callback.
- **`.gitattributes`:** Add `text eol=lf` rules for `*.ts`, `*.tsx`, `*.js`, `*.jsx`, and `*.json`, matching the existing rule for `*.go`. This is to enforce consistent EOL rules across the entire stack.
- **`frontend/vite.config.ts`:** Convert from CRLF to LF to align with the new `.gitattributes` policy.

## Steps to Test

1. Configure an OIDC provider in Headlamp and start the frontend dev server (`npm run frontend`).
2. Initiate the OIDC login flow.
3. Confirm the `/oidc-callback` redirect lands on the `/auth` React route without a 404.

### OIDC callback ends up hitting 404s without `/auth` proxy prefix
image

## Notes for the Reviewer

The root cause was that Vite and rsbuild both treat proxy keys without a trailing slash as exact-match routes, not prefix matches. Adding `/` makes them proxy only requests whose path starts with `/auth/`, leaving the bare `/auth` React route to be served by the dev server as intended.


skoeva (changes requested):
hi - could you sign the CLA so the PR is ready to review?

jooskim:
Hello @skoeva , I am currently in the process of going through the CLA process with my employer. I will get back with an update as soon as possible, thanks!

jooskim:
Closed in lieu of https://github.com/kubernetes-sigs/headlamp/pull/6184

PR #6110, opened 2026-06-20
frontend: PluginSettings: Fix User-installed chip contrast

### LINKED ISSUE #4592: **Settings/PluginSettings**: `frontend/src/components/App/PluginSettings/PluginSettings.tsx` - Fix color contrast for "User-installed" chip - Current: 3.85:1 (#ffffff on #0288d1), Required: 4.5:1 (affects MultipleLocations and MigrationScenario stories)

vtushar06:
The "User-installed" plugin chip rendered white text on the info blue (#0288d1), which is 3.85:1 and fails the WCAG AA minimum of 4.5:1 for normal text.

Fix: keep the info background and set the chip text to a dark color, scoped to this chip in PluginSettings so the theme is not changed globally. I checked it across themes: dark text on info.main is 5.44:1 on the light theme and 9.1:1 on the dark theme (info.main #29b6f6), both above 4.5:1. getContrastText is not used here because at our default contrastThreshold it returns white for info.main and would not reach AA.

The typeLabels map is now typed with ChipProps and SxProps instead of any.

## Testing
- tsc, eslint (ci config) and prettier pass.
- The PluginSettings story tests pass. Snapshots for the MultipleLocations and MigrationScenario stories are updated for the new chip style.

## Screenshots
The change only affects the text color of the "User-installed" chip. Before: white text on #0288d1 (3.85:1). After: dark text on the same blue (5.44:1 light, 9.1:1 dark). Happy to attach a before/after image if useful.

Fixes #4592

### DIFF
--- a/frontend/src/components/App/PluginSettings/PluginSettings.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettings.tsx
@@ -16,9 +16,9 @@
 
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
-import Chip from '@mui/material/Chip';
+import Chip, { ChipProps } from '@mui/material/Chip';
 import Link from '@mui/material/Link';
-import { useTheme } from '@mui/material/styles';
+import { SxProps, Theme, useTheme } from '@mui/material/styles';
 import { SwitchProps } from '@mui/material/Switch';
 import Switch from '@mui/material/Switch';
 import Tooltip from '@mui/material/Tooltip';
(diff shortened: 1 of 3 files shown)


vtushar06:
cc @illume @sniok would love to hear your thoughts on this.

vtushar06:
Small a11y fix for the User-installed chip contrast. @illume @sniok would appreciate a review whenever you get a chance, happy to adjust if anything looks off.

illume:
@vtushar06 I wonder if you checked this with the different themes?

vtushar06:
Thanks @illume. I checked the User-installed chip in both the light and dark themes (screenshots below). The dark text gives 5.44:1 on light and 9.1:1 on dark, both above the 4.5:1 AA minimum, while the default white text was only 3.85:1. The change is scoped to just this one chip, so the rest of PluginSettings is unaffected.
Screenshot 2026-06-25 at 2 19 35 AM
Screenshot 2026-06-25 at 2 22 32 AM



Separately, while testing in the dark theme I noticed a couple of the other status chips look low-contrast on main too either text color doesn't changes as per theme, unrelated to this change.
here is ss from main branch:

Screenshot 2026-06-25 at 2 16 23 AM

illume (approved):
🎉 thanks!

PR #6153, opened 2026-06-23
backend: frontend: Split stopOrDelete port forward into two functions

### LINKED ISSUE #5691: refactor: rename confusing stopOrDelete parameter in portForward API

r0hansaxena:
## Summary

Replaces the confusing `stopOrDelete` boolean in the port forward API with two
dedicated functions, `stopPortForward` and `deletePortForward`, across the
frontend and backend. The wire payload changes from a `stopOrDelete` boolean to
an explicit `action` of either `stop` or `delete`.

## Related Issue

Fixes #5691

## Changes

- Backend splits the store logic into `stopPortForward` and `deletePortForward`,
  sharing a `closePortForward` helper. The two only differ in what they do after
  the channel is closed.
- The backend handler and its request type are renamed to
  `HandlePortForwardAction` and `portForwardActionRequest`, so the names match
  the `action` payload rather than the removed boolean.
- `Validate` now rejects an unknown or missing `action` with 400.
- Frontend replaces `stopOrDeletePortForward(cluster, id, bool)` with the two
  functions and updates all call sites.
- `stopOrDeletePortForward` is kept as a **deprecated wrapper** over the two new
  functions and is still exported from `ApiProxy`, so existing plugins keep
  working. The plugin API snapshot keeps the entry.
- Tests assert which `action` each helper puts on the wire, so the two cannot
  silently send the same one.

## Steps to Test

1. Start a port forward, then click Stop. It stops and can be restarted.
2. Start it again, then Delete. It is removed from the list.

## Notes for the Reviewer

No breaking change for plugins. `stopOrDeletePortForward` keeps its original
signature, including the `stopOrDelete = true` default, and now maps `true` to
`stop` and `false` to `delete`.

The `action` payload is a backend wire change. Headlamp ships the frontend and
backend together, so they always move in step, but it is worth knowing that an
older cached frontend against a newer backend would get a 400 on stop/delete.

All five Copilot review comments are addressed.


illume (changes requested):
Thanks for this PR.


illume (changes requested):
Thanks for working on this.


Looks like there are git conflicts in this PR. Can you fix them up?

<details>

</details>

r0hansaxena:
@illume hey, could you please review this?

PR #6163, opened 2026-06-23
frontend: LogsButton: Fix translation key case mismatch

srajang1805:
## Summary

Fixes a translation key case mismatch in `LogsButton` that prevented the localized string from being resolved correctly.

## Changes

- Updated the translation key passed to `t(...)` to match the defined i18n key casing.

## Note for Reviewers

This is a small i18n fix only. No behavior or UI logic was changed; the update ensures the correct translation key is used and existing localized strings are displayed as intended.


vyncent-t (changes requested):
Please address the copilot comments and also update w/ a rebase, other than that looks good

srajang1805:
@vyncent-t I resolved the comments. The snapshot update was a temporary workaround as tests were failing yesterday which are now resolved. Thank you for your valuable time for reviewing my PR.

illume (approved):
🎉 thanks!

PR #6185, opened 2026-06-25
frontend: workload: Fix Jobs, CronJobs, JobSets overview charts stuck at 100%

### LINKED ISSUE #6173: Workloads overview: Jobs, CronJobs, and JobSets tiles always show 100%
### Description

The Workloads overview page reuses `WorkloadCircleChart` for every workload tile. Its default health logic counts an item as failed only when `getReadyReplicas() != getTotalReplicas()`. Jobs, CronJobs, and JobSets have none of those replica fields, so both helpers return 0, the check is always false, and these three tiles can only ever render 100% Running regardless of the actual state of the underlying objects.

This is the same root cause as the Pods tile bug fixed in #6114 (issue #6090). The Pods tile got an optional per-item `categorize()` classifier so it renders a real multi-segment health ring; Jobs, CronJobs, and JobSets still fall back to the binary replica check and remain stuck at 100%.

### Steps to reproduce

1. Run some Jobs that have failed (or are still active), or CronJobs/JobSets.
2. Open the Workloads overview page.
3. The Jobs / Cron Jobs / Job Sets tiles show 100% Running even when jobs have failed or are not complete.

### Expected

Each tile reflects the real status of its objects, using kind-specific status fields:

- Jobs: `status.succeeded` / `status.failed` / `status.active`.
- CronJobs: healthy when scheduled and the last run succeeded, degraded when suspended (`spec.suspend`), transitional when it has `status.active` jobs running, failed when `status.lastScheduleTime` is newer than `status.lastSuccessfulTime` (last run did not succeed).
- JobSets: conditions plus per-replicatedJob counts.

All four tiles should keep rendering a ring so the overview stays visually consistent; each kind just supplies its own categories.

### Notes

Follow-up to #6114 / #6090. The fix should follow the same approach: pass a `categorize()` classifier per kind and add a kind-specific health helper in the k8s model layer, keeping it small and scoped to these tiles. Deployments/StatefulSets/DaemonSets/ReplicaSets are out of scope here.

xonas1101:
## What this does

The Workloads overview reuses `WorkloadCircleChart` for every tile. Its default
health logic counts an item as failed only when `getReadyReplicas() !=
getTotalReplicas()`. Jobs, CronJobs, and JobSets have none of those replica
fields, so the check is always false and those three tiles are permanently stuck
at 100% regardless of the real state of the objects.

This is the same root cause as the Pods tile fixed in #6114. This PR extends that
PR's optional `categorize()` classifier to the three remaining replica-less
kinds, so every non-replica tile renders a real multi-segment health ring and the
overview row stays visually consistent.

Fixes #6173.

## How

Each kind gets a `getHealth()` in the k8s model layer, mirroring the status logic
its own list page already shows:

- **Job** — `Failed` → failed, `Complete` → healthy, `Suspended` → degraded,
  otherwise (still running) → transitional.
- **CronJob** — suspended (`spec.suspend`) → degraded, has `status.active` jobs →
  transitional, scheduled but last run never succeeded
  (`lastScheduleTime` newer than `lastSuccessfulTime`) → failed, otherwise healthy.
- **JobSet** — `Failed` → failed, `Completed` → healthy, `Suspended` → degraded,
  otherwise → transitional.

`Overview.tsx` wires these in via a per-kind `categorizers` map. Unit tests cover
each category and the priority ordering; the `Overview.Workloads` storyshot is
updated.

## Dependency

> #6114 (the Pods tile fix) — which introduced the `categorize()` prop and
> `WorkloadHealthCategory` type this builds on — has now merged. This branch is
> rebased onto `main`, so the diff is just the Jobs/CronJobs/JobSets delta:
> the model `getHealth()` helpers, their tests, and the `Overview.tsx` wiring.

## Testing

- New unit tests for `Job`/`CronJob`/`JobSet` `getHealth()` (18 cases) pass.
- Storyshots pass; `Overview.Workloads` regenerated to reflect real health.
- `tsc`, ESLint, Prettier clean.

### DIFF
--- a/frontend/src/components/workload/Overview.tsx
+++ b/frontend/src/components/workload/Overview.tsx
@@ -124,6 +124,14 @@ export default function Overview() {
     return <Link routeName={workload.pluralName}>{workloadLabel[workload.className]}</Link>;
   }
 
+  // Jobs/CronJobs/JobSets have no replica fields either (like Pods), so they
+  // classify health per item instead of by replica match.
+  const jobHealth: { [className: string]: (item: Workload) => ReturnType<Job['getHealth']> } = {
+    [Job.className]: item => (item as Job).getHealth(),
+    [CronJob.className]: item => (item as CronJob).getHealth(),
+    [JobSet.className]: item => (item as JobSet).getHealth(),
+  };
+
   return (
     <PageGrid>
       <SectionBox py={2} mt={1}>
@@ -134,8 +142,16 @@ export default function Overview() {
                 workloadData={workloadsData[workload.className] || null}
                 title={<ChartLink workload={workload} />}
                 partialLabel={t('translation|Failed')}
-                totalLabel={workload === Pod ? t('translation|Healthy') : t('translation|Running')}
-                categorize={workload === Pod ? item => (item as Pod).getHealth() : undefined}
+                totalLabel={
+                  workload === Pod || jobHealth[workload.className]
+                    ? t('translation|Healthy')
+                    : t('translation|Running')
+                }
+                categorize={
+                  workload === Pod
+                    ? item => (item as Pod).getHealth()
+                    : jobHealth[workload.className]
+                }
               />
             </Grid>
           ))}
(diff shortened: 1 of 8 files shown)


illume (changes requested):
Thanks for this PR.


xonas1101:
@illume copilot seems fine with the changes. PTAL!

illume (approved):
🎉 thanks!

PR #6210, opened 2026-06-26
github: app.yml: Cache deps and cancel stale app builds

skoeva:
The `Build App` workflow had no dependency caching (unlike `app-artifacts-*` and `frontend`) and no concurrency control, making it one of the slowest PR checks. This caches npm and Go deps and cancels superseded runs on new pushes.

### Changes

- Add `cache: 'npm'` to `setup-node` (app + frontend lockfiles) and a Go module cache path to `setup-go`, matching the existing artifact workflows
- Add a `concurrency` group that cancels in-progress runs when a PR is updated

### Testing

- [x] `make app-test`, `make app-tsc`, and the OS builds pass
- [x] Cached runs skip the cold dependency overloads

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -23,6 +23,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-linux:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.github/workflows/app.yml
```diff
@@ -23,6 +23,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
illume, line 28: Not sure about this?

>  cancel-in-progress: true

If there is a failed download this would stop all the other builds.

illume: Let's try it.

illume (approved):
🎉 thanks!

Issue #5776, opened 2026-05-20
backend: informer cache invalidation leaves all-namespace list cache stale


harrshita123:
## Describe the bug
The backend Kubernetes response cache can keep serving stale all-namespace list responses after a namespaced resource changes.

`GenerateKey` stores list responses with the namespace extracted from the request path. For an all-namespace list like `/api/v1/pods`, the namespace portion is empty, so the cache key is equivalent to `+pods++<context>`.

Informer-driven invalidation builds the key from the changed object's namespace, so a pod update in `default` deletes only `+pods+default+<context>`. The all-namespace cache key remains untouched and can continue serving stale data.

Relevant code:
- `backend/pkg/k8cache/cacheStore.go`: `ExtractNamespace` returns an empty namespace for all-namespace list paths, and `GenerateKey` includes that namespace in the cache key.
- `backend/pkg/k8cache/cacheInvalidation.go`: `handleKeyGenerationAndDeletion` deletes only the key containing `unstructuredObj.GetNamespace()`.
- `DeleteKeys` already shows the intended behavior for non-GET invalidation by deleting both the namespaced key and the namespace-stripped key, but informer invalidation does not use the same logic.

## To Reproduce
Steps to reproduce the bug:
1. Enable the backend cache.
2. Open an all-namespace resource list, for example Pods across all namespaces, so `/api/v1/pods` is cached.
3. Change a pod in a namespace, for example create/update/delete a pod in `default`.
4. Return to or refresh the all-namespace pod list.
5. The backend can serve the cached all-namespace response even though the informer invalidated only the namespaced key.

## Expected behavior
When an informer observes a namespaced resource change, the cache invalidation should delete both:
- the namespace-specific key, for example `+pods+default+<context>`
- the all-namespace key, for example `+pods++<context>`

This should match the broader invalidation behavior already used by `DeleteKeys` for non-GET requests.

## Environment (please provide info about your environment):
 - Installation type: Backend / any deployment with cache enabled
 - Headlamp Version: current main
 - Other: Found by code inspection in `backend/pkg/k8cache`

## Are you able to fix this issue?
Yes, I can propose a PR.

## Additional Context
Duplicate check performed against existing issues for `cache invalidation namespace all namespaces stale`, `stale cache backend k8cache`, and `all namespaces cache`. I found related cache work, but not this specific all-namespace invalidation gap.

PR #6230, opened 2026-06-27
backend: k8cache: Invalidate named GET cache keys on informer events

Pragati5-DEBUG:
### Summary
Informer-driven cache invalidation in k8cache only deleted list-shaped keys built from gvr.Resource (e.g. +pods+default+<context>). That left two stale-cache gaps:

Named GET / detail views — single-object GETs are cached with the object name in the key segment (e.g. +my-pod+default+<context>), which informer invalidation never cleared.
All-namespace lists — same gap as [#5776](https://github.com/kubernetes-sigs/headlamp/issues/5776): list invalidation did not use DeleteKeys, so +pods++<context> could stay stale.
This PR routes informer invalidation through a new helper that evicts list keys (via DeleteKeys) and named GET keys.

### Motivation
When backend cache is enabled, a user can open a pod detail page, then the pod changes elsewhere (controller, kubectl, another tab). The informer fires, but the cached named GET response can still be served — detail views show stale data.

HandleNonGETCacheInvalidation already uses DeleteKeys for mutations; informer invalidation did not match that behavior.

### Changes
Add invalidateCacheKeysForResourceEvent() in cacheInvalidation.go
On informer add/update/delete, evict:
namespaced list key — e.g. +pods+default+<context>
all-namespace list key — e.g. +pods++<context> (via DeleteKeys)
named GET key — e.g. +<objectName>+default+<context>
Add TestInvalidateCacheKeysForResourceEvent and informer integration coverage
Export helper in export_test.go for unit testing

###Test plan

 cd backend && go test ./pkg/k8cache/ -run TestInvalidateCacheKeysForResourceEvent -v

 Manual: enable --cache-enabled, open pod detail, change pod via kubectl label, refresh detail — data should update (no stale X-HEADLAMP-CACHE body)

### DIFF
--- a/backend/pkg/k8cache/cacheInvalidation.go
+++ b/backend/pkg/k8cache/cacheInvalidation.go
@@ -370,5 +370,7 @@ func invalidateCacheKeysForResourceEvent(
 	}
 
 	namedKey := buildCacheKey(gvr.Group, name, namespace, contextKey)
-	_ = k8scache.Delete(context.Background(), namedKey)
+	if err := k8scache.Delete(context.Background(), namedKey); err != nil {
+		logger.Log(logger.LevelError, nil, err, "error while deleting key")
+	}
 }
--- a/backend/pkg/k8cache/cacheInvalidation.go
+++ b/backend/pkg/k8cache/cacheInvalidation.go
@@ -340,13 +340,35 @@ func handleKeyGenerationAndDeletion(obj interface{}, gvr schema.GroupVersionReso
 		return
 	}
 
-	namespace := unstructuredObj.GetNamespace()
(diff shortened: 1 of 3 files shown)


illume (changes requested):
Thanks for these changes.


Pragati5-DEBUG:
@illume Addressed the Copilot feedback — named GET cache eviction now logs Delete() errors instead of ignoring them. Thread resolved, ready for another look.

Pragati5-DEBUG:
@illume All CI green on #6230, Copilot thread resolved (named GET Delete errors are now logged). Ready for review — thanks!

illume (approved):
🎉 thanks!

PR #6237, opened 2026-06-27
frontend: workload: Hide Job duration while the job is still running

### LINKED ISSUE #6229: Job detail page shows "Duration: 0s" for running (incomplete) Jobs
## Description

On a **Job detail page**, a Job that is still running (or otherwise hasn't completed) shows **`Duration: 0s`** instead of either its real elapsed time or no Duration row at all.

## Root cause

`Job.getDuration()` returns `-1` when the job has no `completionTime` yet (i.e. it's still running):

```ts
// frontend/src/lib/k8s/job.ts
getDuration(): number {
  const startTime = this.status?.startTime;
  const completionTime = this.status?.completionTime;
  if (startTime && completionTime) {
    return new Date(completionTime).getTime() - new Date(startTime).getTime();
  }
  return -1;
}
```

The detail page (`jobExtraInfo`) passes that `-1` straight into `formatDuration` and tries to hide the row when empty:

```ts
// frontend/src/components/workload/extraInfo.tsx
const duration = formatDuration(item.getDuration());
...
{ name: t('glossary|Duration'), value: duration, hide: !duration }
```

But `formatDuration(-1)` → `shortHumanDuration(-1)` returns the **truthy string `"0s"`** (because `Math.trunc(-1 / 1000) === -0`, which is not `< 0`), so `hide: !duration` is `false` and the row renders `Duration: 0s`.

The Jobs **list** view handles this correctly by guarding on the numeric value first:

```ts
// frontend/src/components/job/List.tsx
const duration = job.getDuration();
return duration > 0 ? formatDuration(duration, { format: 'mini' }) : '-';
```

So list and detail disagree for the same running Job.

## Steps to reproduce

1. Create a Job that runs for a while (e.g. `kubectl create job sleeper --image=busybox -- sleep 600`).
2. While it's still running, open its **detail page** in the Workloads section.
3. The info table shows **`Duration: 0s`**.

## Expected

While a Job is running, the detail page should not show a misleading `0s` — either hide the Duration row (matching the empty-value intent of `hide`) or show the live elapsed time (`now - startTime`), consistent with how the list view shows `-`.

## Environment

- Affects current `main`.
- Component: frontend (`Job` details / `workload/extraInfo.tsx`).

xonas1101:
Fixes #6229.

On a Job detail page, a Job that is still running showed `Duration: 0s` instead of no Duration row.

`Job.getDuration()` returns `-1` while a job has no `completionTime`. `jobExtraInfo` passed that into `formatDuration`, which returns the truthy string `"0s"` (because `Math.trunc(-1 / 1000) === -0`, not `< 0`), so the `hide: !duration` guard never fired.

This guards on the numeric value first — only formatting a duration once `getDuration() > 0` — matching how the Jobs list view already handles running jobs. Added a small unit test for the running and completed cases.

### DIFF
--- a/frontend/src/components/workload/__snapshots__/extraInfo.Completed.stories.storyshot
+++ b/frontend/src/components/workload/__snapshots__/extraInfo.Completed.stories.storyshot
@@ -1,0 +1,112 @@
+<body>
+  <div>
+    <dl
+      class="MuiGrid-root MuiGrid-container css-kxuems-MuiGrid-root"
+    >
+      <dt
+        class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-3 MuiGrid-grid-md-2 css-9i4s51-MuiGrid-root"
+      >
+        Completions
+      </dt>
+      <dd
+        class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-9 MuiGrid-grid-md-10 css-1dbzfsd-MuiGrid-root"
+      >
+        <span
+          class="MuiTypography-root MuiTypography-body1 css-e06lsu-MuiTypography-root"
+        >
+          1/1
+        </span>
+      </dd>
+      <dt
+        class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-3 MuiGrid-grid-md-2 css-9i4s51-MuiGrid-root"
+      >
+        Parallelism
+      </dt>
+      <dd
+        class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-12 MuiGrid-grid-sm-9 MuiGrid-grid-md-10 css-1dbzfsd-MuiGrid-root"
+      >
+        1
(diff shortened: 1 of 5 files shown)


skoeva (commented):
could you also add an additional storybook test for this? would be helpful

xonas1101:
@skoeva made the changes.

xonas1101:
@illume PTAL.

xonas1101:
@illume copilot generated no new comments. PTAL!

illume (approved):
🎉 thanks!

sniok:
Here's a job that's running before the changes in this PR 
image

here's job details after the change
image

Both have `Duration -` unlike PR and issue claim

I don't see a difference, what did this PR change? Was this even tested?

xonas1101:
@sniok The thing that is presented to the users was aleady fine, this was some other version of this code that was actually buggy ( return 0s ). I could open a PR to just remove that old code - if thats fine? 
edit: This "bug" wasn't found from the UI - that was fine. It came up while grepping through the Job details code, which surfaced the stale jobExtraInfo path.

PR #6251, opened 2026-06-28
frontend: NavigationTabs: Fix active tab index when sidebar items are hidden

### LINKED ISSUE #6250: frontend: NavigationTabs: Active tab indicator missing for Network Policies in browser mode
## Describe the bug

When running Headlamp in browser mode (non-Electron), navigating to **Network → Network Policies** does not highlight the active tab in the horizontal navigation.

The page loads correctly, but the selected tab has no active indicator.

## Root cause

`NavigationTabs` filters hidden sidebar items before rendering tabs, but `defaultIndex` is still calculated from the unfiltered list.

Since the `portforwards` item is hidden outside Electron, the rendered tab indices shift while the selected index does not, causing MUI Tabs to receive an out-of-range active index.

## To Reproduce

1. Run Headlamp in browser mode.
2. Navigate to **Network → Network Policies**.
3. Observe that no navigation tab is highlighted.

## Expected behavior

The **Network Policies** tab should be highlighted like every other navigation section.

## Additional context

The issue appears to originate from computing `defaultIndex` using the unfiltered tab list while rendering a filtered list.

This seems limited to browser mode because `portforwards` is hidden outside Electron.

Adii-45:
## Summary

Fixes the active tab calculation in `NavigationTabs` when rendering a filtered list of sidebar items.

In browser mode, hidden sidebar entries (such as `portforwards`) caused the selected tab index to be calculated from the unfiltered list while the rendered tabs came from the filtered list. As a result, the **Network Policies** tab was not highlighted and tab navigation could become inconsistent.

Fixes #6250

## Changes

- Compute active tab indices from the filtered tab arrays used for rendering.
- Keep the tab change handlers consistent with the filtered tab lists.
- Add a Storybook scenario covering the Network Policies regression.
- Update the corresponding snapshot.

## Verification

- ✅ TypeScript checks pass (`tsc --noEmit`)
- ✅ ESLint passes
- ✅ Storybook tests pass
- ✅ Added regression coverage for the affected navigation scenario

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,6 @@ default_agent: "@dev-agent"
 > **Consulted files (agent must populate before committing):**
 > - `/backend/pkg/k8cache/cacheInvalidation.go`
 > - `/backend/pkg/kubeconfig/contextStore.go`
-> - `/frontend/src/components/Sidebar/NavigationTabs.tsx`
 >
 > **README.md files:**
 > - `/README.md` - Main project README
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,6 +5,7 @@ default_agent: "@dev-agent"
 > **Consulted files (agent must populate before committing):**
 > - `/backend/pkg/k8cache/cacheInvalidation.go`
 > - `/backend/pkg/kubeconfig/contextStore.go`
+> - `/frontend/src/components/Sidebar/NavigationTabs.tsx`
 >
 > **README.md files:**
 > - `/README.md` - Main project README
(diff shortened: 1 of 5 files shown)


AGENTS.md
```diff
@@ -5,6 +5,7 @@ default_agent: "@dev-agent"
 > **Consulted files (agent must populate before committing):**
 > - `/backend/pkg/k8cache/cacheInvalidation.go`
 > - `/backend/pkg/kubeconfig/contextStore.go`
+> - `/frontend/src/components/Sidebar/NavigationTabs.tsx`
```
illume: Can you please remove this AGENTS.md change?

illume (changes requested):
Thanks for this!!

Could you please add a test case for this bug fix?

Adii-45:
@illume Thanks for the review! I've addressed the requested changes:

- Added a regression test covering this bug.
- Removed the unintended `AGENTS.md` change.

I'd appreciate another look whenever you have a chance. Thanks!

illume (approved):
Thanks! 🎉

PR #6253, opened 2026-06-28
Synchronize placeholders for `Node` debugger settings with back-end configuration

dabico:
## Summary

Cluster settings for `Node` will always display the default Docker image and namespace as placeholders, even when deployed with custom image / namespace overrides.

## Changes

- Updated `NodeShellSettings` component to source placeholder using the same pattern used in `PodDebugSettings`.
- Updated translation files to parametrize the namespace instead of hardcoding it.
- Added translations for German and Italian.

## Steps to Test

1. Launch locally, specifying a custom `node-shell-image` and `node-shell-namespace` on the back-end.
2. Open Headlamp UI in the browser.
3. Navigate to `/settings/cluster`, and ensure the placeholder text in <kbd>Node Shell Settings</kbd> fields matches what was passed to the back-end.


illume (approved):
🎉 thanks!

PR #6255, opened 2026-06-28
feat: network-resilient log streaming with auto-reconnect

ayesigaisaac:
- Automatically reconnects log streams after network drops without user intervention
- Tracks the last seen RFC3339 timestamp and passes \`sinceTime=\` on reconnect to resume from exact position
- Deduplicates log lines at the reconnect boundary to prevent repeated output
- Exponential backoff: 1s → 2s → 4s … capped at 30s
- Shows a \"Reconnecting…\" spinner banner in the LogViewer during auto-retry
- Falls back to the manual Reconnect button only after all retries are exhausted
- Resets reconnect state cleanly when the user switches pod or container

## Test plan
- [ ] Start log stream, disconnect network, verify banner appears and stream resumes automatically
- [ ] Verify no duplicate log lines appear after reconnect
- [ ] Verify manual Reconnect button still appears after prolonged failure
- [ ] Verify switching containers resets reconnect state


ashu8912:
Hii @ayesigaisaac can you first create an issue to discuss this and why this would be needed, and suggest your approach there?

PR #6271, opened 2026-06-29
frontend: i18n: Restore translations removed by Diagnostics revert

sniok:
The revert in be93ef37a0129f8667857e83f7ae467453d3692b accidentally removed existing translation values to empty strings across many locale files


skoeva (approved):
thanks!

PR #6293, opened 2026-06-30
frontend: Sidebar: Minimize activities and close details drawer on click

### LINKED ISSUE #4792: Ability to go back to main view, when a navigation menu item is clicked

Scenario:
User clicks Workloads/Pods, then a pod, and views logs. Now wants to go back to pods view, clicks Workloads/Pods again, and nothing happens. 

Side issue:
If the user clicks another tree item, e.g. Workloads/Deployments, pod log view minimizes, but the previously clicked pod details pane (one that stick to the right by default) pops-up.

Proposed solution: 
When an item from the main tree view is clicked, everything else gets minimized, and the main view is displayed. 

I think even a better solution would be to close the detail view, unless the user snaps, minimizes, made it full screen etc.
If the user is interested in a detail view (e.g. pod log) they would interact beyond navigating, or else, it would mean the user has moved on. 

All

No

Rohith-Saran:
When navigating through the main sidebar, open details panels (split-right activities) and active log/terminal activities should be closed or minimized so the user is presented with a clean main view.

## Summary
This PR fixes issue #4792 by clearing or minimizing active panels and drawers upon sidebar link interaction.

## Related Issue
Fixes #4792

## Changes
- **`activitySlice`**: Added a `minimizeAll` action that removes temporary activities (such as split-right resource detail panels) and minimizes other active ones (such as logs or terminals).
- **`Activity` Helper**: Added the `minimizeAll()` method to dispatch this action.
- **`ListItemLink`**: Added an `onClick` handler to sidebar buttons that calls `Activity.minimizeAll()` and dispatches `setSelectedResource(undefined)` to reset the details drawer state.
- **`pluginLib.snapshot`**: Updated the snapshot file to include `minimizeAll` within the registered `Activity` library namespace.

## Steps to Test
1. Select a cluster and navigate to **Workloads** -> **Pods**.
2. Click on a pod name to show its details.
3. Click on the logs or terminal button to open log/terminal activities.
4. Click on **Workloads/Pods** (or any other item like **Workloads/Deployments**) in the sidebar.
5. Verify that the split-right details panel is closed and the logs activity gets minimized, returning you cleanly to the list view.

## Notes for the Reviewer
- The changes have been fully type-checked with TypeScript (`npm run frontend:tsc`) and verified with the test suite (`npm run frontend:test`).

### DIFF
--- a/frontend/src/components/Sidebar/ListItemLink.tsx
+++ b/frontend/src/components/Sidebar/ListItemLink.tsx
@@ -22,7 +22,10 @@ import { styled } from '@mui/material/styles';
 import Tooltip from '@mui/material/Tooltip';
 import { alpha } from '@mui/system/colorManipulator';
 import React from 'react';
+import { useDispatch } from 'react-redux';
 import { Link as RouterLink, LinkProps as RouterLinkProps } from 'react-router-dom';
+import { setSelectedResource } from '../../redux/drawerModeSlice';
+import { Activity } from '../activity/Activity';
 
 const ExpandedIconSize = 20;
 const CollapsedIconSize = 24;
@@ -40,6 +43,7 @@ interface ListItemLinkProps {
   level?: number;
   fullWidth?: boolean;
   divider?: boolean;
+  onClick?: (event: React.MouseEvent<any, any>) => void;
   containerProps?: {
     [prop: string]: any;
   };
@@ -52,6 +56,7 @@ const StyledLi = styled('li')<{ level: number }>(({ level }) => ({
 }));
 
 export default function ListItemLink(props: ListItemLinkProps) {
+  const dispatch = useDispatch();
   const {
     primary,
     pathname,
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for these changes.


sniok:
this reimplements logic that minimizes activities that we already have in `frontend/src/components/activity/Activity.tsx:990`
I'd rather we update that logic to handle the case when user clicks on the same page they're on instead of introducing new, conflicting mechanism for minimizing activities

Rohith-Saran:
@illume and @sniok  Before I rework this: should the same-page-click case just reuse the minimize path that Activity.tsx:990 already calls, or does it need its own branch there since the trigger (route click vs explicit close) is different? Want to make sure I extend it the right way rather than just relocating the duplication.

illume:
@Rohith-Saran by the way, there's a PR here which I think adds a test case which covers the feature mentioned above.
(If I've understood it properly)

illume:
Maybe related to this one:
- https://github.com/kubernetes-sigs/headlamp/issues/7410

cc @tejhan

PR #6302, opened 2026-07-01
backend: renameCluster: Return errors on collision and cache failure

vtushar06:
## What

`handleClusterRename` (backend) had two failure paths that were reported as success:

- On a **duplicate cluster name** it writes a 400, but then returns the outer `err`, which is `nil` at that point. So `renameCluster` falls through and records the failed rename as a success (it logs "Completed renameCluster request" and records success metrics/duration).
- On an **`updateCustomContextToCache` failure** it passes that same `nil` `err` to `handleError`, which does `http.Error(w, err.Error(), ...)` and **panics** on the nil dereference. The real errors in `errs` were also discarded.

This returns a real error in both cases: a non-nil error for the name collision, and `errors.Join(errs...)` for the cache failure (which also surfaces the actual cause instead of a generic message).

## Why

A name collision was silently logged and measured as a successful rename, and a cache-update failure crashed the request handler via a nil-pointer panic.

## Testing

- Added `TestHandleClusterRename_NameCollision`, which asserts the collision path returns a non-nil error and a 400. Confirmed it **fails without the fix** ("An error is expected but got nil") and passes with it.
- Existing `TestRenameCluster` still passes; `gofmt` and `go vet` are clean.

```
$ go test ./cmd/ -run 'TestRenameCluster|TestHandleClusterRename_NameCollision'
ok  	github.com/kubernetes-sigs/headlamp/backend/cmd
```

cc @illume

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -2741,10 +2741,12 @@ func (c *HeadlampConfig) handleClusterRename(w http.ResponseWriter, r *http.Requ
 	isUnique := CheckUniqueName(config.Contexts, clusterName, reqBody.NewClusterName)
 	if !isUnique {
 		http.Error(w, "custom name already in use", http.StatusBadRequest)
-		logger.Log(logger.LevelError, map[string]string{logFieldCluster: clusterName},
-			err, "cluster name already exists in the kubeconfig")
 
-		return err
+		renameErr := errors.New("cluster name already in use")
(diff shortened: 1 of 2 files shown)


skoeva (changes requested):
could you run `make backend-lint-fix` to address the failing linter?

vtushar06:
Thanks @skoeva, good catch. Ran `make backend-lint-fix`, it was just two missing blank lines (wsl). Lint is clean now and the tests still pass.

illume (approved):
🎉 thanks!

PR #6310, opened 2026-07-01
frontend: recentClusters: Add try/catch around JSON.parse for localSt…

### LINKED ISSUE #6309: frontend: Unhandled JSON.parse and type leakage in getRecentClusters causes startup crash
## Bug description

The `getRecentClusters()` function in [`frontend/src/helpers/recentClusters.ts`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/helpers/recentClusters.ts) retrieves the `recent_clusters` key from `localStorage` and attempts to parse it using `JSON.parse` without a `try/catch` block.

Because `getRecentClusters` is invoked on app startup to render the cluster-selection UI, if the `localStorage` key contains malformed JSON (due to a browser crash, interrupted write, or storage quota failure), the unhandled `SyntaxError` prevents the application's cluster UI from loading entirely.

Furthermore, even if the JSON is valid, the parsed result is forcibly cast `as string[]` without verifying that the elements are actually strings. If the storage is corrupted with an array of objects, this bypasses the `Array.isArray` check and leaks incorrect types into the rest of the application, causing runtime crashes downstream.

Lastly, `setRecentCluster()` calls `localStorage.setItem()` without catching potential `QuotaExceededError` or `SecurityError` exceptions, which can disrupt application flow.

## Blast Radius
This bug crashes the cluster selection process if the storage is corrupted, meaning the user cannot connect to any clusters until they manually clear their browser's Local Storage.

## Expected behavior
The helper should catch `JSON.parse` errors, log a warning, and return an empty array `[]`. It should also validate that the parsed array strictly contains only strings. Similarly, write failures should be caught and logged gracefully.

## Context
This is part of a broader effort to harden `localStorage` usage and type safety across Headlamp (similar to fixes in `tableSettings.ts` and `useLocalStorageState.ts`).

Kakumanu-Harshitha:
## Summary
This PR hardens the `recentClusters` helper functions against corrupted `localStorage` data, browser storage exceptions, and type leakage, preventing application startup crashes.

## Related Issue
Fixes #6309 

## Root Cause
`getRecentClusters()` retrieves the `recent_clusters` key from `localStorage` and calls `JSON.parse` without a `try/catch`. If the data is corrupted, a `SyntaxError` is thrown on application startup, crashing the cluster-selection UI before the user can connect. Furthermore, the parsed output was forcefully cast `as string[]` without runtime validation; if the storage contained an array of objects, it would leak incorrect types and crash downstream. Finally, `setRecentCluster()` called `localStorage.setItem` without catching `QuotaExceededError`.

## Changes
- **`frontend/src/helpers/recentClusters.ts`:**
  - Wrapped `JSON.parse` in `getRecentClusters()` with a `try/catch`. On failure, logs a warning and returns `[]`.
  - Added strong runtime type validation (`recentClusters.every(item => typeof item === 'string')`) to prevent non-string elements from leaking past the `as string[]` cast.
  - Wrapped `localStorage.setItem` in `setRecentCluster()` with a `try/catch` to gracefully log errors if the quota is exceeded.
- **`frontend/src/helpers/recentClusters.test.ts`:**
  - Replaced the `toThrow(SyntaxError)` test with one that verifies it correctly catches the error, logs a warning, and returns `[]`.
  - Added a test to ensure it returns `[]` if the stored array contains non-string elements.
  - Added a test to verify `localStorage.setItem` throws are caught and logged.

## Steps to Test
1. Run `cd frontend && npx vitest run src/helpers/recentClusters.test.ts` to verify all 13 tests pass.
2. Open Headlamp, navigate to Local Storage in DevTools, and manually corrupt the `recent_clusters` key (e.g., set it to `{ broken JSON ]` or `[{"unexpected": "object"}]`).
3. Refresh the page. The app should load normally (ignoring the corrupted recent clusters) rather than crashing to a blank screen. 

## Verification
- [x] Linting and type checking (`npm run frontend:tsc`) pass.
- [x] 13/13 tests in `recentClusters.test.ts` pass.
- [x] Diff scope is minimal and targeted.

### DIFF
--- a/frontend/src/helpers/recentClusters.test.ts
+++ b/frontend/src/helpers/recentClusters.test.ts
@@ -27,6 +27,7 @@ function read(): unknown {
 describe('recentClusters', () => {
   beforeEach(() => {
     localStorage.clear();
+    vi.restoreAllMocks();
   });
 
   describe('setRecentCluster', () => {
@@ -75,6 +76,19 @@ describe('recentClusters', () => {
 
       expect(read()).toEqual(['alpha']);
     });
+
+    it('catches and logs errors when localStorage.setItem throws', () => {
+      const spyError = vi.spyOn(console, 'error').mockImplementation(() => {});
+      vi.spyOn(localStorage, 'setItem').mockImplementation(() => {
+        throw new Error('Quota exceeded');
+      });
+
+      setRecentCluster('alpha');
+      expect(spyError).toHaveBeenCalledWith(
+        `Failed to set ${STORAGE_KEY} in localStorage:`,
+        expect.any(Error)
+      );
+    });
   });
 
   describe('getRecentClusters', () => {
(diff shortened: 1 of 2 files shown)


Kakumanu-Harshitha:
@illume ptal ! when your are free?

Kakumanu-Harshitha:
@illume all copilot conversation issues are resolved
ready for full review and merge?

illume (approved):
🎉 thanks!

Issue #5198, opened 2026-04-23
[Plugin Proposal] Velero: contextual backup integration for workload, namespace, and cluster-wide views


MattiZZZ241:
### Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

I am a platform engineer running Velero in production and using Headlamp as management UI.

Velero's CRDs (`backups.velero.io`, `schedules.velero.io`, `restores.velero.io`) are already browsable in Headlamp through the generic CRD viewer, the raw data is accessible.

The real issue is that there is no contextual integration with workload or namespace views. Someone browsing a Deployment or StatefulSet has to leave that view entirely, navigate to the CRD section, and read raw YAML to determine whether that workload is covered or not by a backup schedule. The coverage signal exists in the tool but there is no path from a workload or namespace view to the relevant Velero data.

The problem compounds with Velero's targeting model. A schedule can scope by namespace, but also by label selector (`spec.labelSelector`) and resource type (`spec.includedResources`). In practice, it is common to run schedules that cover only a single application (e.g. a StatefulSet and its PVCs) within a namespace that contains other workloads with no coverage at all. A namespace-level indicator is not enough from my point of view: some of us need more granularity.

**This is not a proposal for a standalone Velero dashboard.** 

Two of those already exist: [otwld/velero-ui](https://github.com/otwld/velero-ui) and [seriohub/vui-ui](https://github.com/seriohub/vui-ui).

The value here is contextual integration within the cluster management workflow, this is the same pattern established by the Karpenter (#3231) and Kubeflow (#3710) plugins, where the tool has its own UI and the Headlamp plugin adds integration into how operators actually navigate the cluster.

This plugin idea was also raised in `#headlamp` on Kubernetes Slack before filing this issue.

---

#### Shared data layer

A `useVeleroData()` hook that:

- Lists `backups.velero.io` and `schedules.velero.io` CRDs via the Kubernetes API
- Joins them via `metadata.labels["velero.io/schedule-name"]` present on Backup CRs
- Computes per-schedule health: last run timestamp, last run status, next scheduled run from the cron expression
- Is consumed by both plugin surfaces without duplication

No backend required, this is just pure CRD watching.

---

#### Phase 1 — Workload and namespace backup coverage panel *(first target)*

A detail panel injected into Deployment, StatefulSet, and PVC views showing:

- Which Velero schedule(s) cover this specific resource, evaluated against:
  - `spec.includedNamespaces` — namespace match
[...]

PR #6320, opened 2026-07-01
frontend: globalSearch: Lazy-load searchable resource classes

Pragati5-DEBUG:
## Summary

Lazy-load the 16 Kubernetes resource class modules used by global search in a separate chunk, instead of statically importing them all into `GlobalSearchContent`.

When the user opens search, the input appears immediately while resource models load in the background. List fetching only starts after those modules are available.

## Test plan

- [x] `npx tsc --noEmit` in `frontend/`

### Fix:

searchResourceClasses.ts — loads each resource class via dynamic import() in a separate chunk (cached after first load).

GlobalSearchContent.tsx split into:

Placeholder — search field + spinner while classes load (user can still type)
Loaded — runs useList hooks only after classes are available
Removed eager KubeIcon import — uses LazyKubeIcon everywhere.

### DIFF
--- a/frontend/src/components/globalSearch/searchResourceClasses.ts
+++ b/frontend/src/components/globalSearch/searchResourceClasses.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { KubeObject, type KubeObjectClass } from '../../lib/k8s/KubeObject';
+
+// Ensure the base class is initialized before resource models are loaded dynamically.
+void KubeObject;
+
+let searchResourceClassesPromise: Promise<KubeObjectClass[]> | null = null;
+
+/**
+ * Loads Kubernetes resource classes used by global search in a separate chunk so
+ * opening the search UI does not eagerly bundle every list model up front.
+ */
+export function loadSearchResourceClasses(): Promise<KubeObjectClass[]> {
+  if (!searchResourceClassesPromise) {
+    searchResourceClassesPromise = Promise.all([
+      import('../../lib/k8s/pod'),
+      import('../../lib/k8s/deployment'),
+      import('../../lib/k8s/service'),
+      import('../../lib/k8s/job'),
+      import('../../lib/k8s/cronJob'),
+      import('../../lib/k8s/configMap'),
+      import('../../lib/k8s/namespace'),
+      import('../../lib/k8s/statefulSet'),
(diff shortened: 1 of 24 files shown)


Pragati5-DEBUG:
@illume Copilot feedback on #6320 is now addressed in 65c36bfea on frontend/global-search-lazy-resource-classes.

Pragati5-DEBUG:
Hi @illume — quick check-in on Velero #5198.

You already OK’d Phase 1 scope (useVeleroData + read-only coverage panels). I’ve started a thin slice locally on headlamp-k8s/plugins:

- useVeleroData() hook (schedules + backups CRDs)
- Deployment detail panel (“Velero backup coverage”)
- coverage matching logic + unit tests
- verified on minikube with Velero + a sample schedule/backup

I also commented on the issue and pinged Matthis (issue author) so we don’t duplicate work.

Before I open a PR: is a thin first PR (Deployment panel + hook + tests) the right landing approach, or do you prefer waiting for full Phase 1 (StatefulSet/PVC/namespace panels too)?

Happy to align with whatever you/Matti prefer. Thanks!

Pragati5-DEBUG:
Anything to add @illume or is it fine?

illume (commented):
Can you show me the built chunks for each case?

I want to see if this actually creates separate chunks, because I expect these imports are already inside a larger chunk.

illume (changes requested):
Can you show me the built chunks for each case?

I want to see if this actually creates separate chunks, because I expect these imports are already inside a larger chunk.

Pragati5-DEBUG:
Built main vs this branch: __no per-resource chunk file_s_; the 16 models stay in index-*.js (~6 MB). GlobalSearchContent is its own ~20 KB chunk on both sides.
This PR defers loadSearchResourceClasses() and the 16 useList hooks until Global Search mounts, 

_rather than splitting those models into new files._

Happy to adjust approach if you’d prefer a different splitting strategy.

Pragati5-DEBUG:
Addressed the remaining Copilot feedback: skip logging after unmount, and load resource classes via a single aggregator chunk.

Pragati5-DEBUG:
Hi @illume — when you have a moment, could you review the Velero plugin PR first?
https://github.com/headlamp-k8s/plugins/pull/877
Happy to keep iterating on #6320 after that.

frontend/src/components/globalSearch/searchResourceClasses.ts
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { KubeObjectClass } from '../../lib/k8s/KubeObject';
+
+let searchResourceClassesPromise: Promise<KubeObjectClass[]> | null = null;
+
+/**
+ * Loads Kubernetes resource classes used by global search in a separate chunk so
```
sniok, line 22: all the resource classes are already imported on page load in pluginLib so this doesn't result in the claimed separate chunk

sniok (changes requested):

Pragati5-DEBUG:
Thanks @sniok @illume — agreeing with the chunk point. I also replied earlier (Jul 6) after comparing builds: no per-resource chunks; the 16 models stay in index-*.js, and this PR mainly defers loadSearchResourceClasses() / the 16 useList hooks until Global Search mounts, not a real bundle split.

So the remaining value is runtime deferral of those list queries (plus placeholder/error UI), not smaller initial JS. Prefer I reframe the PR around that, or close it?

illume (changes requested):
Thanks for this PR.


<details>

</details>

PR #6335, opened 2026-07-02
backend: telemetry: Fix metrics gauge leak and double-count in panic recovery

srajang1805:
## Summary

Fix a bug in `RequestCounterMiddleware` where panicking requests caused request metrics to be recorded twice and the active requests gauge to be decremented twice.

## Changes

* Restructure the deferred metrics handling to record metrics exactly once.
* Update the response status to `500` when a panic is recovered.
* Re-panic only after all metrics have been recorded.
* Preserve existing behavior for successful and non-panic error requests.

## Result

* `ActiveRequestsGauge` is decremented exactly once per request.
* `RequestCounter` is recorded exactly once, including on panic paths.
* Panic propagation is preserved so `net/http.Server` can perform its normal recovery.

## Testing 

* Relevant Tests passed.

## Note for reviewers

The deferred metrics logic has been restructured so metrics are recorded exactly once before re-panicking, eliminating the double-counting issue while preserving the existing panic propagation behavior.

### DIFF
--- a/backend/pkg/telemetry/metrics.go
+++ b/backend/pkg/telemetry/metrics.go
@@ -132,26 +132,28 @@ func (m *Metrics) RequestCounterMiddleware(next http.Handler) http.Handler {
 		wrapper := newResponseWriter(w)
 
 		defer func() {
+			statusCode := wrapper.statusCode
+
+			// Capture panic before recording metrics so both paths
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this PR.

PR #6345, opened 2026-07-02
frontend: hooks: Fix stale WebSocket watch on same-kind resource navigation

### LINKED ISSUE #6344: frontend: useKubeObject keeps watching the previous resource's WebSocket stream after in-place navigation between same-kind resources
## Describe the bug

When navigating directly from one resource's **Details** page to another resource of the same kind (for example, Pod A → Pod B) without performing a full page reload, the Details page for the newly selected resource stops receiving live updates via the WebSocket watch.

The initial data is loaded correctly (via the REST request), but subsequent changes to the resource (status transitions, restarts, condition changes, etc.) are no longer reflected until the page is manually refreshed.

## To Reproduce

1. Open the Details page for a resource (e.g. `Pod A`).
2. From the Details page, navigate to another resource of the same kind (e.g. `Pod B`) using an in-app link.
3. Modify the second resource (restart it, delete/recreate it, or trigger any status change).
4. Observe that the Details page no longer receives live updates until the page is refreshed.

## Expected behavior

The currently displayed resource should continue receiving live updates after navigating between resources of the same kind.

## Actual behavior

The initial resource data loads correctly, but the WebSocket subscription continues tracking the previously viewed resource, so live updates for the newly selected resource are never received until a full page reload.

## Environment

- Headlamp Version: current `main`
- Installation type: Any
- Browser: Any
- OS: Any

Yes. I can submit a pull request.

## Additional Context

While tracing the live-update flow, I found that `useKubeObject` memoizes the WebSocket connection descriptor using a dependency array that only includes `endpoint`, even though the memoized value also depends on `namespace`, `name`, `cluster`, `cleanedUpQueryParams`, and `queryKey`.

When navigating between two resources of the same endpoint (for example, two Pods), the endpoint remains unchanged, so the memoized connection descriptor is reused. Since the WebSocket effect depends on this memoized descriptor, the existing subscription is never recreated and continues watching the previously selected resource.

This appears to be localized to:

- `frontend/src/lib/k8s/api/v2/hooks.ts`

A possible fix would be to update the memoization dependencies so the WebSocket connection descriptor is recomputed whenever the resource identity changes, allowing the subscription to reconnect to the currently displayed resource.

Adii-45:
## Summary

Fixes a stale WebSocket subscription in `useKubeObject` when navigating between resources of the same kind (for example, Pod A → Pod B) without a full page reload.

`connectionsRequests` was memoized using only `endpoint`, even though the memoized value also depends on the resource identity. Since the endpoint remains the same for resources of the same kind, the existing WebSocket subscription continued watching the previous resource and live updates stopped for the newly displayed Details page.

This change updates the dependency array so the watch connection is recreated whenever the resource identity changes.

Fixes #6344

## Changes

- Add `namespace` and `name` to the `connectionsRequests` `useMemo` dependency array
- Add a regression test covering navigation between same-kind resources
- Preserve existing behavior for unrelated code paths

## Testing

- [x] Regression test added
- [x] `npx vitest run src/lib/k8s/api/v2/hooks.test.ts`
- [x] `npm run tsc`
- [x] `npm run eslint`
- [x] Full `src/lib/k8s/api/v2` test suite passes

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.test.tsx
+++ b/frontend/src/lib/k8s/api/v2/hooks.test.tsx
@@ -396,4 +396,43 @@ describe('useKubeObject watch wiring', () => {
       expect(lastCall.url()).toContain('namespaces/my-ns');
     });
   });
+
+  it('re-subscribes to the new resource when navigating between resources of the same kind', async () => {
+    vi.stubEnv('REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER', 'false');
+    mockClusterFetch.mockResolvedValue(
+      mockJsonResponse({
+        apiVersion: 'v1',
+        kind: 'Pod',
+        metadata: { name: 'pod-a', namespace: 'my-ns' },
+      })
+    );
+
+    const { rerender } = renderHook(
+      ({ name }: { name: string }) =>
+        useKubeObject({
+          kubeObjectClass: MockPod,
+          name,
+          namespace: 'my-ns',
+          cluster: 'test',
+        }),
+      { wrapper: createWrapper(), initialProps: { name: 'pod-a' } }
+    );
+
+    // Same endpoint ("pods") is used for both resources, so only the watch
+    // connection's fieldSelector should change between renders.
+    await waitFor(() => {
+      const calls = mockUseWebSockets.mock.calls;
+      const lastCall = calls[calls.length - 1][0];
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for these changes.


Adii-45:
@illume All review comments have been addressed and marked as resolved. Thanks for the review!

illume (approved):
🎉 thanks!

PR #6351, opened 2026-07-03
frontend: plugins: Show error notification when plugin loading fails

### LINKED ISSUE #6349: Plugins: plugin load failure is swallowed silently with no user-facing error
When plugin loading fails (e.g. backend is unreachable), the error gets swallowed silently by `.catch(console.error)` in `Plugins.tsx`. The user gets no notification at all — the app just loads with an empty plugin list and no explanation. You'd only know something went wrong if you had DevTools open.

The component already uses `enqueueSnackbar` for incompatible plugin warnings just a few lines above, so there's a clear pattern to follow — it's just not being used when the whole `fetchAndExecutePlugins` call fails.
1. Start Headlamp with the backend not running (or kill the backend after startup with `pkill headlamp-server`)
2. The app loads normally — plugin list is empty, no error shown anywhere in the UI
3. Open DevTools → Console tab and you'll see:

GET http://localhost:4466/plugins net::ERR_CONNECTION_REFUSED
fetchWithRetry @ index.ts:402
fetchAndExecutePlugins @ index.ts:450
(anonymous) @ Plugins.tsx:48

The error is there in the console but the user has no idea.
- Installation type: Mac (running from source)
- Headlamp Version: 0.43.0
Yes (I will propose a PR)
The relevant code is in `frontend/src/plugin/Plugins.tsx:85-90`:

.finally(() => {
  dispatch(pluginsLoaded());
  window.desktopApi?.send('pluginsLoaded');
})
.catch(console.error);

Replacing `.catch(console.error)` with something that also calls `enqueueSnackbar` would fix this.

krsatyamthakur-droid:
Previously, if fetchAndExecutePlugins() rejected (e.g. the backend was
unreachable), the error was silently swallowed by a bare .catch(console.error).
Users had no indication that plugins failed to load.

Replace it with a handler that logs the error and displays an error
snackbar via enqueueSnackbar(), matching the pattern used elsewhere in
the codebase for user-visible fetch failures.

## Summary

When plugin loading fails, users now see an error snackbar instead of
nothing. The fix replaces the bare `.catch(console.error)` with a handler
that both logs and surfaces the error to the user.

## Related Issue

Fixes #6349

## Changes

- `frontend/src/plugin/Plugins.tsx`: replaced `.catch(console.error)` with
  a handler that logs the error and calls `enqueueSnackbar()` with `variant: 'error'`
- Updated all 17 locale files with the new translation string via `npm run i18n`

## Steps to Test

1. In `frontend/src/plugin/index.ts`, temporarily make `fetchAndExecutePlugins`
   throw (e.g. add `throw new Error('test')` at the top of the function)
2. Run Headlamp and wait for it to load
3. An error snackbar should appear saying "Failed to load plugins. Please try reloading the app."
4. Remove the test throw and confirm normal plugin loading still works

N/A — the change is a snackbar that only appears on failure

## Notes for the Reviewer

- The new string follows the `translation|` namespace prefix convention
  already used in Plugins.tsx for the incompatible plugins warning
- All locale files were regenerated with `npm run i18n`; non-English
  entries are placeholder strings that translation contributors can fill in


krsatyamthakur-droid:
Good catch — I missed that index.ts does Promise.reject(resp) in a few places, so err isn't always an Error. Changed the type to unknown in the latest commit.

illume (changes requested):
Thanks for these changes.


krsatyamthakur-droid:
Fixed. `dispatch()` was outside the try/catch so a throw there would go unhandled. Moved it inside and cleaned up the wrong comment too.

frontend/src/plugin/Plugins.tsx
```diff
@@ -82,12 +82,20 @@ export default function Plugins() {
         }
       }
     )
-      .finally(() => {
-        dispatch(pluginsLoaded());
-        // Warn the app (if we're in app mode).
```
illume, line 87: Please add the comment back.

illume (changes requested):
Thanks for this. 

I wonder if you saw this error, or how you found it?

Could you please add a test case?

frontend/src/plugin/Plugins.tsx
krsatyamthakur-droid: Done, added it back. Missed it when wrapping the block in try/catch.

krsatyamthakur-droid:
`While going through the plugin loading flow in index.ts, I noticed that when fetchAndExecutePlugins rejects — for example if the backend is down — the error just gets swallowed by .catch(console.error). Nothing shows up for the user. That felt like something worth fixing.`

`For the test, I added Plugins.test.tsx with two cases — one where the fetch fails and we verify the error snackbar actually shows up, and one where it succeeds and we check no error snackbar is triggered.
`

illume (changes requested):
Thanks for these changes.


<details>

- `frontend: plugins: add test for error snackbar and restore inline comment` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

krsatyamthakur-droid:
Done — squashed all the fixup commits into two clean commits and removed the issue reference from the commit body.

illume (approved):
🎉 thanks!

PR #6355, opened 2026-07-03
frontend: oidcauth: Fix Safari OIDC login stuck due to ITP storage event suppression

### LINKED ISSUE #6353: In-cluster OIDC login stuck using Safari browser
## Describe the bug

After upgrading from Headlamp 0.42.0 to 0.43.0, OIDC auth no longer works using Safari browser in in-cluster mode.

## To Reproduce

1. Deploy headlamp in in-cluster mode, configured with OIDC auth.
2. Click *Sign in*. It opens redirect popup, and get stuck on it:

Image

The main window gets stuck waiting for the popup to close, meaning auth can never complete.

Switching to Chrome or Firefix and it works, so this is probably a Safari specific behavior.

 - Installation type: In-cluster helm chart
 - Headlamp Version: 0.43.0, same chart version
 - Other: Safari version 26.5.

No, but I'm happy to contribute more info if I can.

anushkagupta200615-jpg:
Safari's ITP suppresses cross-window storage events, causing the OIDC popup auth flow to get stuck. The beforeUnloadListener and the setInterval closed-window branch both called onClose() without checking localStorage first.

Fix both paths to read AUTH_STATUS_KEY from localStorage before treating the popup close as a cancellation. Use setTimeout(..., 0) in beforeUnloadListener to ensure the popup's useEffect has committed the write.

## Summary

This PR fixes a bug where in-cluster OIDC login gets stuck when using the Safari browser. Safari's Intelligent Tracking Prevention (ITP) suppresses the cross-window storage events that the main window relies on. The fallback mechanisms (beforeunload and window polling) were incorrectly treating the popup closing as a login cancellation without first checking if the auth token was successfully written to localStorage.

## Related Issue

Fixes #6353

## Changes

- Updated `OauthPopup.tsx` `beforeUnloadListener` to defer checking `localStorage` (via `setTimeout(..., 0)`) before triggering cancellation.
- Updated `OauthPopup.tsx` `setInterval` closed-window fallback to read `localStorage` for completed auth before triggering `onClose()`.
- Added two new Safari-specific test cases in `OauthPopup.test.tsx` to cover the ITP fallback scenarios.
- Updated an existing test in `OauthPopup.test.tsx` to correctly handle the now-asynchronous `beforeUnloadListener`.

## Steps to Test

1. Deploy Headlamp in in-cluster mode with OIDC authentication configured.
2. Open Headlamp using the Safari browser.
3. Click on "Sign in" to open the OIDC auth popup.
4. Complete the authentication flow in the popup.
5. Observe that the popup closes and the main window successfully redirects and logs you in, rather than getting stuck waiting.

*(N/A - underlying logic fix)*

## Notes for the Reviewer

- The `setTimeout(..., 0)` used in `beforeUnloadListener` is a workaround for a timing issue specific to Safari where `beforeunload` can fire before the popup window's React `useEffect` commits the `localStorage` write. Deferring the check to the next event loop tick ensures the data is readable.

### DIFF
--- a/frontend/src/components/oidcauth/OauthPopup.test.tsx
+++ b/frontend/src/components/oidcauth/OauthPopup.test.tsx
@@ -64,7 +64,7 @@ describe('OauthPopup', () => {
     expect(popupWindow.close).toHaveBeenCalled();
   });
 
-  it('removes the storage listener when the popup closes without completing auth', () => {
+  it('removes the storage listener when the popup closes without completing auth', async () => {
     const popupListeners: Record<string, () => void> = {};
     const popupWindow = {
       addEventListener: vi.fn((eventName: string, listener: () => void) => {
@@ -74,6 +74,7 @@ describe('OauthPopup', () => {
       close: vi.fn(),
     } as unknown as Window;
 
+    vi.useFakeTimers();
     vi.spyOn(window, 'open').mockReturnValue(popupWindow);
     const removeEventListenerSpy = vi.spyOn(window, 'removeEventListener');
     const onClose = vi.fn();
@@ -94,8 +95,13 @@ describe('OauthPopup', () => {
 
     popupListeners.beforeunload?.();
 
+    // beforeUnloadListener defers via setTimeout(..., 0) — flush it
+    await vi.runAllTimersAsync();
+
     expect(removeEventListenerSpy).toHaveBeenCalledWith('storage', expect.any(Function));
     expect(onClose).toHaveBeenCalled();
+
+    vi.useRealTimers();
   });
 
   it('closes the popup and removes listeners when auth completes', () => {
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for these changes.


illume (changes requested):
Thanks for the contribution.


<details>

- `﻿frontend: oidcauth: Fix Safari OIDC login stuck` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -106,14 +112,65 @@ const OauthPopup: React.FC<OauthPopupProps> = props => {
     storageListenerRef.current = storageListener;
     window.addEventListener('storage', storageListener);
 
+    intervalRef.current = window.setInterval(() => {
```
gambtho, line 115: **Blocker: a stale `auth_status` is consumed as a fresh success.**

This interval reads `localStorage.getItem(AUTH_STATUS_KEY)` unconditionally, starting the instant the popup opens. `oidcauth/index.tsx` writes `'success'` on *any* load of the redirect route carrying a `?cluster=` param, and the only code that ever removes the key is this component on consumption.

So the key can outlive a flow: the parent tab is closed or navigated away mid-login, or a user opens/reloads the redirect URL directly. On the next "Sign In" click, the first interval tick — roughly 500 ms in, while the IdP page is still loading — fires `onCode('success')` and `cleanupPopup(true)`, closing the popup and reporting a successful login that never happened.

Under the old event-only model this was harmless, since a pre-existing key fires no `storage` event.

Suggested fix: in `createPopup`, before `window.open`, either `localStorage.removeItem(AUTH_STATUS_KEY)`, or snapshot the current value and require the polled value to differ from it before treating it as a result.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -106,14 +112,65 @@ const OauthPopup: React.FC<OauthPopupProps> = props => {
...
+      if (externalWindowRef.current?.closed) {
```
gambtho, line 116: **Blocker: the interval leaks indefinitely when the popup is blocked.**

`window.open` returns `null` when a popup blocker intervenes, so `externalWindowRef.current` is `null` and `externalWindowRef.current?.closed` evaluates to `undefined` — falsy. This branch never runs, `cleanupPopup()`/`props.onClose()` are never called, and the poll keeps firing every 500 ms for as long as `AuthChooser` stays mounted.

Combined with the stale-key issue above, a blocked popup can produce a "successful" login with no popup ever having opened.

Suggested fix: treat a null window as closed — `if (!externalWindowRef.current || externalWindowRef.current.closed)` — or skip installing the interval altogether when `window.open` returned `null`.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -106,14 +112,65 @@ const OauthPopup: React.FC<OauthPopupProps> = props => {
...
         const beforeUnloadListener = () => {
-          cleanupPopup();
-          externalWindowRef.current = null;
-          if (!!props.onClose) {
-            props.onClose();
-          }
+          // Safari fires beforeunload before the popup's useEffect has committed
+          // the localStorage write, and also suppresses the cross-window storage
+          // event due to ITP. Defer by one task so localStorage is readable.
+          setTimeout(() => {
```
gambtho, line 156: **Blocker: this timer is never cancelled and can tear down a later popup session.**

`cleanupPopup` clears the interval and both listeners, but holds no handle on this `setTimeout`, so the callback survives cleanup.

Concrete failure: the popup unloads and the timer is scheduled; before it fires, either the component unmounts, or the user clicks "Sign In" again — which runs `cleanupPopup(true)`, opens a *new* window, and installs a *new* interval and storage listener. The pending callback then reads an (already-consumed) empty `localStorage`, falls through to `cleanupPopup()`, and clears the **new** interval and removes the **new** storage listener, then sets `externalWindowRef.current = null`. The second login flow is left permanently dead with no detection path, and `props.onClose()` fires against a session that is still open.

Suggested fix: store the id in a ref (e.g. `unloadTimeoutRef`) and clear it in `cleanupPopup` and on unmount, alongside `intervalRef`.

frontend/src/components/oidcauth/OauthPopup.tsx
```diff
@@ -106,14 +112,65 @@ const OauthPopup: React.FC<OauthPopupProps> = props => {
...
+
+      try {
+        const authStatus = localStorage.getItem(AUTH_STATUS_KEY);
+        if (authStatus) {
+          onCode(authStatus);
+          localStorage.removeItem(AUTH_STATUS_KEY);
+          cleanupPopup(true);
+        }
+      } catch (e) {
+        // ignore
```
gambtho, line 146: Non-blocking: this swallows exactly the failure mode the PR exists to fix. Safari in private-browsing mode can throw `SecurityError` from `localStorage.getItem`, and here that would fail silently 120 times a minute with no diagnostic — while the structurally identical branch ~25 lines above does `console.error`. Consider logging once (guarded by a ref so it isn't spammed) and clearing the interval, since a throwing `localStorage` will never succeed on a later tick.

frontend/src/components/oidcauth/OauthPopup.test.tsx
```diff
@@ -74,6 +74,7 @@ describe('OauthPopup', () => {
       close: vi.fn(),
     } as unknown as Window;
 
+    vi.useFakeTimers();
```
gambtho, line 77: Non-blocking: `vi.useFakeTimers()` is restored by a `vi.useRealTimers()` at the end of the test *body* rather than in `afterEach`. If an assertion in one of these tests fails, the restore is skipped and fake timers leak into every subsequent test in the file — which now all start a real 500 ms interval. The result is cascading unrelated failures that obscure the original one. Moving the restore into the existing `afterEach` makes it failure-safe.

gambtho (changes requested):
Thanks for the contribution - approach looks good

The change in detection model does introduce three issues that I think need addressing before merge. All three come from the same root cause: the previous code only reacted to storage *events*, so a leftover `auth_status` key was inert and the listener was inherently tied to a live popup. An unconditional `getItem` poll removes both of those properties.

1. A stale `auth_status` value is now consumed as a fresh success, which can complete a login the user never performed.
2. The interval never terminates when `window.open` is blocked.
3. The deferred `setTimeout` is not cancellable and can tear down a subsequent popup session.

Details inline. Extracting a single `consumeAuthStatus(): boolean` helper would also collapse the four copies of the read/`onCode`/`removeItem`/`cleanupPopup(true)` block into one place, making #1 a single-site fix.

Two non-blocking notes are inline as well.

PR #6357, opened 2026-07-04
frontend: activity: Wrap Close button aria-label with t()

### LINKED ISSUE #6350: frontend: activity: ActivityBar Close button aria-label is not translated
Describe the bug

The × close button in the ActivityBar has aria-label="Close" hardcoded as a raw string. Every other button in the same component (Fullscreen, Snap Left, Snap Right, etc.) uses t() — this one was just missed. Screen readers on non-English locales will read "Close" in English no matter what language is set.

To Reproduce

Switch Headlamp to a non-English language (Settings → General → Language)
Open pod logs or a terminal so the activity bar appears at the bottom
Inspect the × button in the activity bar — aria-label is still "Close" in English
Environment

Not environment-specific — it's a code issue that affects any Headlamp installation.

Are you able to fix this issue?

Yes, I'll open a PR.

Additional context

The string "Close" already exists in all locale translation files, so the fix is a single-line change — just wrapping the hardcoded string in t() like the surrounding buttons already do.

krsatyamthakur-droid:
The 'Close' IconButton in ActivityBar had its aria-label hardcoded as the
English string "Close" while every other aria-label in the same component
(Window, Fullscreen, Snap Left/Right/Top/Bottom, Overview) correctly uses
the t() translation helper.

Wrap it in t('Close') for consistency and to ensure screen readers announce
the label in the user's chosen language.

## Summary

Fixes a missed `t()` wrap on the ActivityBar close button's `aria-label`.
Every other button in the same component was already translated — this one
got skipped.

## Related Issue

Fixes #6350

## Changes

- `Activity.tsx`: `aria-label="Close"` → `aria-label={t('Close')}`
- Added missing `"Close"` translations for 11 locales (de, es, fr, hi, it,
  ja, ko, pt, ta, zh, zh-tw) that previously had empty strings and were
  silently falling back to English

## Steps to Test

1. Switch Headlamp to any non-English language (Settings → General → Language)
2. Open pod logs or a terminal so the ActivityBar appears at the bottom
3. Use a screen reader or inspect the × button in the activity bar —
   the `aria-label` should now reflect the translated string

N/A

## Notes for the Reviewer

- `"Close"` already existed as a key in all locale files but 11 of them
  had empty string values — with `returnEmptyString:false` in the i18next
  config, those fell back to English silently
- ar, bn, ru, ur already had correct translations; he and en were fine as-is

### DIFF
--- a/frontend/src/components/activity/Activity.tsx
+++ b/frontend/src/components/activity/Activity.tsx
@@ -1212,7 +1212,7 @@ export const ActivityBar = React.memo(function ({
               Activity.close(it.id);
             }}
             sx={{ width: '42px', height: '100%', borderRadius: 1, flexShrink: 0 }}
-            aria-label="Close"
+            aria-label={t('Close')}
           >
             <Icon icon="mdi:close" />
           </IconButton>
--- a/frontend/src/i18n/locales/de/translation.json
+++ b/frontend/src/i18n/locales/de/translation.json
@@ -16,7 +16,7 @@
   "Snap Top": "",
   "Snap Bottom": "",
   "Minimize": "",
-  "Close": "",
+  "Close": "Schließen",
   "Close All": "",
   "Overview": "",
   "Loading": "",
--- a/frontend/src/i18n/locales/es/translation.json
+++ b/frontend/src/i18n/locales/es/translation.json
@@ -16,7 +16,7 @@
   "Snap Top": "",
   "Snap Bottom": "",
   "Minimize": "",
-  "Close": "",
+  "Close": "Cerrar",
   "Close All": "",
   "Overview": "",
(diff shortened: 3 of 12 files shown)


krsatyamthakur-droid:
Hey, good catch. I checked all the locale files and found 11 of them had "Close" as an empty string — de, es, fr, hi, it, ja, ko, pt, ta, zh, zh-tw. Since returnEmptyString: false is set in the i18next config, those were silently falling back to English the whole time.

Added the translations for all 11 in a follow-up commit and updated the PR description to reflect the actual scope. All 17 locales now have a proper value for "Close".

illume (approved):
🎉 thanks!

PR #6378, opened 2026-07-05
backend: kubeconfig: Pre-allocate slices in contextStore methods

### LINKED ISSUE #6377: backend: Slice Pre-allocation Optimization in ContextStore
### Issue Description:

 In backend/pkg/kubeconfig/contextStore.go, the GetContexts() and GetContextKeys() methods append items to zero-length slices while iterating over contextMap. Since the map is fetched first, we already know the exact final size of the slices via len(contextMap). Pre-allocating these slices using make([]*Context, 0, len(contextMap)) and make([]string, 0, len(contextMap)) will prevent the Go runtime from dynamically reallocating memory and copying elements as the slices grow. This provides a clear code readability improvement and memory micro-optimization that is perfectly backward compatible and easily mergeable.

CygnusMaximillian:
**Summary:** Pre-allocate slices in contextStore methods for memory efficiency
**Related Issue:** Fixes #6377 
**Changes:**
- Modified `GetContexts()` in `backend/pkg/kubeconfig/contextStore.go` to pre-allocate the `contexts` slice using `make([]*Context, 0, len(contextMap))`.
- Modified `GetContextKeys()` in `backend/pkg/kubeconfig/contextStore.go` to pre-allocate the `keys` slice using `make([]string, 0, len(contextMap))`.

**Steps to Test:**
1. Run backend tests: `npm run backend:test`
2. Verify all `contextStore` tests continue to pass.

**Notes for the Reviewer:**
A small memory optimization to prevent unnecessary slice growth reallocations. Since the target sizes are exactly known from the map lengths beforehand, this enhances code efficiency and readability with zero side effects.

### DIFF
--- a/backend/pkg/kubeconfig/contextStore.go
+++ b/backend/pkg/kubeconfig/contextStore.go
@@ -120,11 +120,9 @@ func (c *contextStore) GetContexts() ([]*Context, error) {
 		return nil, err
 	}
 
-	contexts := make([]*Context, len(contextMap))
-	i := 0
+	contexts := make([]*Context, 0, len(contextMap))
 	for _, ctx := range contextMap {
-		contexts[i] = ctx
-		i++
+		contexts = append(contexts, ctx)
 	}
 
 	return contexts, nil
@@ -137,11 +135,9 @@ func (c *contextStore) GetContextKeys() ([]string, error) {
 		return nil, err
 	}
 
-	keys := make([]string, len(contextMap))
-	i := 0
+	keys := make([]string, 0, len(contextMap))
 	for key := range contextMap {
-		keys[i] = key
-		i++
+		keys = append(keys, key)
 	}
 
 	return keys, nil
--- a/backend/pkg/kubeconfig/contextStore.go
+++ b/backend/pkg/kubeconfig/contextStore.go
@@ -115,15 +115,16 @@ func (c *contextStore) AddContext(headlampContext *Context) error {
 
 // GetContexts returns all contexts in the store.
 func (c *contextStore) GetContexts() ([]*Context, error) {
-	contexts := []*Context{}
-
 	contextMap, err := c.cache.GetAll(context.Background(), nil)
 	if err != nil {
 		return nil, err
 	}
 
+	contexts := make([]*Context, len(contextMap))
+	i := 0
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this PR.


<details>

</details>

illume (changes requested):
Thanks for this PR.


vtushar06:
the slice pre-allocation reads good now, `make([]T, 0, len)` + append is the idiomatic pattern and covers the copilot note. tests pass locally for me too (`go test ./pkg/kubeconfig/` is green).

one small thing worth a look @illume: the TestContextAuthType addition seems unrelated to the slice change, it covers AuthType() in kubeconfig.go which this PR doesn't touch. probably cleaner as its own commit given the atomic-commit guideline, unless it was there to cover a specific failing test.

yolossn (changes requested):
Thanks for the contribution, @CygnusMaximillian. A few things before this can move forward:
  
On the optimization. `GetContexts()` and `GetContextKeys()` are called on cold paths (startup, watcher sync, cluster inventory) with a handful of contexts — not thousands. Pre-allocating saves a couple of tiny append grows per call, which isn't measurable in practice.

If you think the impact isn't negligible, please share a `go test -bench` result (before/after, with realistic context counts) so we have numbers to justify the change.

Copilot's comment is still open. `GetContextKeys() `used to return nil for an empty store and now returns a non-nil empty slice — that's an observable change (JSON null vs [], == nil). Please either preserve the old behavior or reply confirming it's safe, and mark the comment resolved.

Unrelated test. As @vtushar06 noted, TestContextAuthType doesn't belong here, it tests code this PR doesn't touch. Please split it into its own PR.

**More broadly:** if you'd like to contribute to Headlamp, there are plenty of good first issue and help wanted tickets that would have more impact than a micro-optimization.

CygnusMaximillian:
Closing this PR. After review feedback and local benchmarks, the slice pre-allocation doesn’t show a meaningful win for realistic context counts (a handful to ~50). These methods run on cold paths, so the savings aren’t worth the review cost. Thanks everyone for reviewing this @vtushar06 @yolossn  and @illume :)

PR #6405, opened 2026-07-06
frontend: pod: Remove unreachable duplicate condition

### LINKED ISSUE #6404: frontend: pod: Remove duplicate unreachable cache-hit check in getDetailedStatus
## Summary

The `getDetailedStatus()` method in `frontend/src/lib/k8s/pod.ts` contains an **identical cache-hit guard duplicated back-to-back**. The second copy is unreachable dead code — if the condition is true, execution already returns at the first check; if false, the condition will still be false at the second check.

## Location

**File:** `frontend/src/lib/k8s/pod.ts`
**Method:** `getDetailedStatus()`

skools-here:
## Summary

Remove the duplicate unreachable cache-hit check in getDetailedStatus(). The second guard was identical to the first and could never be executed because the function would already have returned. This change removes dead code without altering functionality, improving readability and maintainability

## Related Issue

Fixes #6404

### DIFF
--- a/frontend/src/lib/k8s/pod.ts
+++ b/frontend/src/lib/k8s/pod.ts
@@ -375,14 +375,6 @@ class Pod extends KubeObject<KubePod> {
       return this.detailedStatusCache.details;
     }
 
-    // We cache this data to avoid going through all this logic when nothing has changed
-    if (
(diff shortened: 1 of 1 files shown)


skoeva (approved):
lgtm

PR #6412, opened 2026-07-06
frontend: hpa: Fix incorrect check and break statement

### LINKED ISSUE #6411: hpa.ts : averageUtilization renders as empty string instead of  '0%' and dead break after return
## Describe the bug
 In hpa.ts,  metricTargetValue()  has two issues:
- Incorrect falsy check ( if (target.averageUtilization) ) causes  averageUtilization = 0  to display as  ''  instead of '0%'.
- Unreachable break on line 73 after return '' dead code that should be removed.

## Fix 
- Use (target.averageUtilization != null) instead of (target.averageUtilization).
- Remove Unnecessary break statement.

skools-here:
## Summary

This PR adds/fixes [feature/bug] by [brief description of what the change does].

## Related Issue

Fixes #6411 

## Changes

Two bugs fixed in  metricTargetValue()  in  frontend/src/lib/k8s/hpa.ts :
- Bug:  if (target.averageUtilization)  used a truthiness check, so a metric with  averageUtilization == 0  would silently render as  ''  instead of  '0%' .
Fix: Changed to  if (target.averageUtilization != null)  to correctly handle the zero case.
- Bug:  break  after  return ''  was unreachable dead code inside a  switch  case.
Fix: Removed the dead  break .

Files changed:  frontend/src/lib/k8s/hpa.ts

### DIFF
--- a/frontend/src/lib/k8s/hpa.test.ts
+++ b/frontend/src/lib/k8s/hpa.test.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


skools-here:
ptal @illume

skools-here:
ptal @illume

illume (approved):
🎉 thanks!

PR #6421, opened 2026-07-07
backend: serviceproxy: Use context.Background() instead of context.TODO() in tests

ssz2605:
## Summary

Replace `context.TODO()` with `context.Background()` in the service proxy tests.

`context.TODO()` is intended as a temporary placeholder when the appropriate context is not yet known. Since test code always has a valid root context, `context.Background()` is the appropriate choice. This aligns with Go best practices.

## Changes

- `backend/pkg/serviceproxy/service_test.go`
  - Replaced 2 occurrences of `context.TODO()` with `context.Background()`
- No functional changes; this is a test code cleanup for consistency with Go conventions.

### DIFF
--- a/backend/pkg/serviceproxy/service_test.go
+++ b/backend/pkg/serviceproxy/service_test.go
@@ -28,7 +28,7 @@ func TestGetServiceInternal(t *testing.T) {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


skoeva (approved):
this change is purely stylistic, not fixing any linter warnings. updated the description to reflect this, thanks

PR #6446, opened 2026-07-09
frontend: AlertNotification: reset health-check backoff on recovery

### LINKED ISSUE #6445: AlertNotification: cluster health-check polling backs off on failure and never resets after recovery
### Summary

The cluster health-check banner (`AlertNotification`) increases its polling interval by 5s on **every** failed check and never resets it after the cluster recovers on its own. As a result, the error/offline banner can linger for a long time after connectivity is restored, and health polling stays slow for the rest of the session.

### Component

`frontend/src/components/common/AlertNotification.tsx`

### Details

The interval delay is computed as `(networkStatusCheckTimeFactor + 1) * NETWORK_STATUS_CHECK_TIME` (`NETWORK_STATUS_CHECK_TIME = 5000`).

- On each failed check, the catch handler runs `setNetworkStatusCheckTimeFactor(prev => prev + 1)`, and the effect (dep `[networkStatusCheckTimeFactor]`) re-registers the interval with the new, longer delay. So the interval grows linearly and unbounded: 5s → 10s → 15s → …
- The factor is reset **only** by the manual **"Try Again"** button (`onClick={() => setNetworkStatusCheckTimeFactor(0)}`).
- The automatic success path only does `setError(false)` — it never resets the factor.

### Steps to reproduce

1. Open Headlamp connected to a cluster.
2. Make the cluster/API unreachable for a while (e.g. ~2 minutes) so several consecutive health checks fail. The polling interval climbs (e.g. ~15 failures → ~80s).
3. Restore connectivity (no user interaction with the banner).

### Expected

Once the cluster is reachable again, the error/offline banner clears promptly and health polling returns to the normal 5s cadence.

### Actual

- The banner lingers for up to the grown interval (e.g. ~80s) after recovery, because the next automatic check that would call `setError(false)` is that far away.
- For the rest of the session, polling stays at the elevated interval, so the next real outage is detected slowly. A longer or repeated outage pushes the interval to several minutes.

### Suggested fix

Reset the factor in the success path so normal cadence resumes automatically after recovery:

```js
.then(() => {
  setError(false);
  setNetworkStatusCheckTimeFactor(0);
})
```

Resetting to `0` when it's already `0` is a no-op (no interval churn); it only re-arms the interval when the factor was elevated.

xonas1101:
Fixes #6445.

## Problem

`AlertNotification` polls cluster health on an interval of `(networkStatusCheckTimeFactor + 1) * 5000ms`. Every failed check increments `networkStatusCheckTimeFactor`, and the effect re-registers the interval with the longer delay, so the interval backs off 5s → 10s → 15s → … linearly and unbounded.

The factor is reset **only** by the manual "Try Again" button. The automatic success path just calls `setError(false)` and never resets it. So after a transient outage:

- the error/offline banner lingers for up to the grown interval after the cluster is reachable again (the next automatic check that would clear it is that far away), and
- polling stays slow for the rest of the session, delaying detection of the next outage. A longer/repeated outage pushes the interval to several minutes.

## Fix

Reset `networkStatusCheckTimeFactor` to `0` in the success path so polling returns to the normal 5s cadence once a check succeeds. Resetting to `0` when it is already `0` is a no-op (no interval churn); it only re-arms the interval when the factor was elevated.

## Test

Added `AlertNotification.test.tsx`, which drives the poll interval with fake timers and asserts the cadence returns to 5s after a failure → recovery. It fails without the fix (the third check does not fire at t=20s) and passes with it. Note the suite-wide `fakeTimers.toFake` only covers `setTimeout`/`clearTimeout`, so the test opts `setInterval`/`clearInterval` in explicitly.

### DIFF
--- a/frontend/src/components/common/AlertNotification.test.tsx
+++ b/frontend/src/components/common/AlertNotification.test.tsx
@@ -25,9 +25,13 @@ vi.mock('../../lib/cluster', async importOriginal => ({
 }));
 
 describe('PureAlertNotification', () => {
-  // The suite-wide config only fakes setTimeout/clearTimeout, so opt setInterval
+  // The suite-wide config fakes Date + setTimeout/clearTimeout; opt setInterval/clearInterval
   // (the health-check poller) in as well for these timing assertions.
-  beforeEach(() => vi.useFakeTimers({ toFake: ['setInterval', 'clearInterval'] }));
+  beforeEach(() =>
+    vi.useFakeTimers({
+      toFake: ['Date', 'setTimeout', 'clearTimeout', 'setInterval', 'clearInterval'],
+    })
+  );
   afterEach(() => vi.useRealTimers());
 
   // A failed check backs the poll interval off from 5s to 10s. Once a check
--- a/frontend/src/components/common/AlertNotification.test.tsx
+++ b/frontend/src/components/common/AlertNotification.test.tsx
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { act, render } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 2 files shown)


xonas1101:
@illume fixed

ashu8912:
Hii @xonas1101 can you share a small recording of how the polling is fixed and how it is reflected in UI?

xonas1101:
https://github.com/user-attachments/assets/2ac4d413-2974-4e12-80d9-c2f865f57a2c

@ashu8912 here it is!

We can see the cadence went onto around 15-20s after some failures, but when I fixed the problem, it went back to 5s instead of still being at the last given interval.

xonas1101:
@illume copilot generated no new comments

xonas1101:
@illume @ashu8912 one interesting thing - if the server is lazy in responding to bad requests, the cadence doesn't follow - requests on healthz are still made every 5s, maybe we could do this so that the consequent requests are made only when the precedent is resolved by the server. I could work on the fix in this PR itself.

illume (approved):
Thanks! 🎉

PR #6452, opened 2026-07-10
backend: auth: Add RFC 8693 Token Exchange (STS) client support

Suyog241005:
## Summary

Adds an RFC 8693 OAuth 2.0 Token Exchange (STS) client to the Go backend. This allows Headlamp to dynamically exchange a user's session token for a target cluster-scoped token right before forwarding requests to the cluster, avoiding native credential sharing/impersonation.
Additionally, supports RFC 7523 Workload Identity client authentication, enabling the Headlamp pod to authenticate to the STS provider using its own Kubernetes Service Account token.

## Changes

- `backend/pkg/auth/sts.go`: Core OIDC discovery and token exchange logic.
- `backend/pkg/auth/sts_test.go`: Complete test coverage for client secrets and client assertions.
- `backend/cmd/headlamp.go` & `backend/cmd/server.go`: Proxy routing integration.
- `backend/pkg/config/config.go`: Mapping for `--oidc-sts-enabled`, `--oidc-sts-audience-map`, etc.

## Steps to Test

1. Run backend tests: `go test -v ./pkg/auth/...`
2. Start Headlamp with `--oidc-sts-enabled=true` and configuration flags pointing to an OIDC issuer that supports token exchange.

## Notes for the Reviewer

## Summary by CodeRabbit

- **New Features**
  - Added optional OIDC Security Token Service (STS) support for Kubernetes API requests.
  - Added configuration options for STS enablement, issuer, client credentials, and cluster audience mappings.
  - Supports secure token exchange using configured TLS settings and workload identity credentials.

- **Bug Fixes**
  - Invalid token exchanges now return a clear `401 Unauthorized` response instead of forwarding the request.


ashu8912:
Hii @Suyog241005 can you add steps for which oidc issuer you are testing this on, and the entire flow for it?

Suyog241005:
Hi @ashu8912! 

The design was aligned in this Slack thread: https://kubernetes.slack.com/archives/C01FXB5E8ER/p1783504788655199?thread_ts=1782401998.746849&cid=C01FXB5E8ER.

For local testing and verification, I simulated an OIDC issuer that supports token exchange (STS). Here is the flow to reproduce it:

### 1. Run a local Mock STS Server
```bash
python3 -c '
import http.server, json
class MockSTS(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path == "/.well-known/openid-configuration":
            self.send_response(200)
            self.send_header("Content-Type", "application/json")
            self.end_headers()
            self.wfile.write(json.dumps({"issuer": "http://localhost:8080", "token_endpoint": "http://localhost:8080/token"}).encode())
    def do_POST(self):
        if self.path == "/token":
            self.send_response(200)
            self.send_header("Content-Type", "application/json")
            self.end_headers()
            print("[Mock STS] Received token exchange request!")
            self.wfile.write(json.dumps({"access_token": "exchanged-cluster-scoped-token-xyz"}).encode())
http.server.HTTPServer(("localhost", 8080), MockSTS).serve_forever()
```

### 2. Start Headlamp Go backend
```bash
cd backend
go run ./cmd \
  -dev \
  --oidc-sts-enabled=true \
  --oidc-sts-issuer-url="http://localhost:8080" \
  --oidc-sts-audience-map="docker-desktop=https://rollercoaster.audience.com"
```

### 3. Verify via proxy request
```bash
curl -H "Cookie: headlamp-auth-docker-desktop.0=MY-MAIN-PARK-TICKET" \
  http://localhost:4466/clusters/docker-desktop/api/v1/namespaces
```
The mock server console will print: `[Mock STS] Received token exchange request!` and the proxy request will forward with the exchanged token.

Suyog241005:
Hi @kahirokunn and @mlbiam, I hope you're both doing well! Just wanted to gently follow up on this draft. The implementation is complete and the tests are passing. Whenever you have some time, I’d love to hear your thoughts on the approach or any changes you'd suggest to align this with project patterns. No rush at all, and thank you for your time!

kahirokunn:
I understand from our previous discussion that using STS is likely the ideal approach. However, I must apologize, as I am not at all familiar with the STS protocol, its implementation, or its ecosystem, and I do not have enough time to study or research it. Therefore, I believe someone else would be better suited to review this PR. 🙏

Suyog241005:
Hi @illume , I have addressed the copilot's comments.

Suyog241005:
Hi @illume , I have addressed the copilot's comments.

Suyog241005:
Hi @illume , I have addressed the copilot's comments.

Suyog241005:
Hi @vyncent-t, could you please review this PR when you have some time? Thanks!

Suyog241005:
Hi @illume copilot generated no new comments.

Suyog241005:
hi @illume , copilot has generated no new comments.

kahirokunn:
Based on current context, I believe this content is currently associated with the following Issue:
I recommend not merging this PR until the Issue progresses and a clear direction is established.
https://github.com/kubernetes-sigs/headlamp/issues/6857

/hold

kahirokunn:
> hi @illume , copilot has generated no new comments.

I also believe that we need to review not only Copilot but also all of CodeRabbit's content.

southball:
@Suyog241005 hello! I took a look at #6857 and this PR. It looks like a nice implementation to ensure that we use separate tokens between clusters.

I am slightly curious regarding the IdP support, since RFC 8693 support and implementation seems to differ a little bit between each provider. I would like to test this PR in my local environment as well; could you let me know what IdP you tested this with?

(For instance, Okta seems to require `subject_token_type` to be `subject_token_type = urn:ietf:params:oauth:token-type:access_token` which seems incompatible with this PR.)

Also, it seems that without a token cache, this PR performs a token exchange for each request made. Is that intentional?

Suyog241005:
Hi @southball, thanks for the review!

1. **IdP Support:** This was tested against **Keycloak** (with token-exchange enabled) and mock STS servers where `urn:ietf:params:oauth:token-type:jwt` is used. I can add an `--oidc-sts-subject-token-type` flag so providers like Okta (`urn:ietf:params:oauth:token-type:access_token`) are fully supported.
2. **Token Caching:** Exchanging per-request was the initial minimal approach. I can add an in-memory TTL token cache (using `expires_in`) to avoid latency and IdP rate-limiting.

I'll update the PR with both the token cache and configurable `subject_token_type`.

illume (changes requested):
Thanks for these changes.


illume (changes requested):
Thanks for these changes.


Suyog241005:
Hi @illume , I have addressed the copilot's comments.

Suyog241005:
Hi @illume just a gentle follow up on this PR.

PR #6466, opened 2026-07-10
frontend: ConfigMap: Base64 decode/encode binaryData in the details editor

### LINKED ISSUE #6462: frontend: ConfigMap Binary Data editor displays raw Base64 and does not preserve binary data when edited
## Describe the bug

The ConfigMap Details page's **Binary Data** section displays `ConfigMap.binaryData` values as raw Base64 strings instead of decoding them for editing.

Additionally, when a value is modified and saved, the updated text is sent back without being Base64-encoded again. Since the Kubernetes API expects `binaryData` values to be Base64-encoded (`map[string][]byte`), editing binary entries either fails validation or results in incorrect data being stored.

The sibling **Secret** details page already performs the required Base64 decode/encode roundtrip, but the ConfigMap Binary Data editor currently does not.

---

## To Reproduce

1. Create a ConfigMap containing a `binaryData` entry.
   ```bash
   kubectl create configmap demo-cm --from-file=hello.txt
   ```

   or

   ```bash
   kubectl patch configmap demo-cm --type merge \
     -p '{"binaryData":{"hello.txt":"SGVsbG8sIFdvcmxkIQo="}}'
   ```

2. Open the ConfigMap in Headlamp.
3. Navigate to **Details → Binary Data**.
4. Observe that the value is displayed as the raw Base64 string (`SGVsbG8sIFdvcmxkIQo=`) instead of the decoded content.
5. Edit the value and click **Save**.

---

## Expected behavior

- Binary Data values should be Base64-decoded before being displayed for editing.
- After editing, the updated content should be Base64-encoded before sending the update to the Kubernetes API.
- The editing experience should match the existing behavior of the Secret editor.

---

## Actual behavior

- Binary Data values are displayed as raw Base64.
- Edited values are sent back without re-encoding.
- Saving either fails with an API validation error (invalid Base64) or, if the edited value happens to be valid Base64, stores unintended binary content.

---

## Environment

- Headlamp Version: current `main`
- Installation Type: Any
- Browser: Any
- OS: Any

---

Yes. I can submit a pull request for this issue.

---

## Additional Context

While investigating, I noticed that the ConfigMap Binary Data editor uses a plain text editing flow for `binaryData`, whereas the existing Secret editor already performs the necessary Base64 decode/encode roundtrip before displaying and saving values.

The issue appears to be localized to the ConfigMap Binary Data editing path and should be possible to resolve by mirroring the encode/decode handling already used for Secret data.

Adii-45:
## Summary

Fixes #6462

The ConfigMap Details page's Binary Data section displayed Base64-encoded values directly and sent edited plaintext back to the API, which could result in incorrect binary data being stored.

This change mirrors the existing Secret implementation by:

- decoding `binaryData` values before displaying them in the editor
- encoding edited values before saving
- leaving `ConfigMap.data` handling unchanged

## Testing

- Added a regression test covering the `binaryData` encode/decode flow
- Ran ConfigMap component tests successfully
- Verified TypeScript and ESLint pass

### DIFF
--- a/frontend/src/components/configmap/Details.test.tsx
+++ b/frontend/src/components/configmap/Details.test.tsx
@@ -21,12 +21,10 @@ import { act, fireEvent, render, screen, within } from '@testing-library/react';
 import { Base64 } from 'js-base64';
 import _ from 'lodash';
 import type ConfigMap from '../../lib/k8s/configMap';
+import { CLUSTER_ACTION_GRACE_PERIOD } from '../../redux/clusterActionSlice';
 import { TestContext } from '../../test';
 import { ConfigMapDataSection } from './Details';
 import { BASE_BINARY_DATA_AND_DATA_CONFIG_MAP } from './storyHelper';
-
-// clusterAction defers the wrapped update by this grace period before running it.
-const CLUSTER_ACTION_GRACE_PERIOD = 5000;
 
 // Stub the ConfigMap class module. Evaluating it pulls the k8s barrel, which has an
 // init-order cycle when loaded cold in an isolated test (class-extends-undefined).
--- a/frontend/src/components/configmap/Details.test.tsx
+++ b/frontend/src/components/configmap/Details.test.tsx
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Adii-45:
@illume All Copilot review comments have been addressed and resolved. Another review would be greatly appreciated! Thank you!

Adii-45:
@illume All Copilot review comments have been addressed and resolved & there are no new comments. Another look would be greatly appreciated! Thank you!

PR #6468, opened 2026-07-10
frontend: Table: Hide columns to fit width

### LINKED ISSUE #1232: Mobile, service accounts cut off column
Seems on this view the data is cut off from display, whereas before it wasn't.

after:
image

before:
image

skoeva:
Headlamp rendered every table column regardless of viewport width, so on narrow windows and high zoom levels flexible columns were crushed toward zero or pushed behind a horizontal scrollbar, cutting off data. This hides columns that no longer fit, dropping the least useful ones first.

Fixes: #1232

### Changes

- Measure the table's available width with a `ResizeObserver` and hide columns that don't fit, instead of crushing them or relying on a horizontal scrollbar
- Add a `responsivePriority` column prop - columns are dropped by ascending priority, then right-to-left among equal priority, and the first column is always kept
- Assign priorities to the shared resource columns so Namespace/Labels drop before Cluster/Kind, and Age/Name are kept longest
- Columns without a priority default to `0`, so other tables keep previous right-to-left behavior unchanged
- Hidden columns remain reachable via the existing column-visibility menu

### Testing

- [x] `make frontend-test` passes
- [x] `make frontend-lint` passes
- [x] Resize/zoom a resource list and confirm columns drop by priority and come back when widened

### Screenshots

image

### DIFF
--- a/frontend/src/components/App/Home/__snapshots__/index.Base.stories.storyshot
+++ b/frontend/src/components/App/Home/__snapshots__/index.Base.stories.storyshot
@@ -90,846 +90,850 @@
                 </div>
               </div>
               <div
-                aria-atomic="true"
-                aria-live="polite"
-                class="MuiBox-root css-ty8jgw"
-                role="status"
-              />
-              <div
-                class="MuiBox-root css-1ipmh7v"
+                class="MuiBox-root css-8atqhb"
               >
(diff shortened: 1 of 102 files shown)


illume (approved):
🎉 thanks!

PR #6471, opened 2026-07-11
app: plugin-manager: guard against malformed JSON in IPC handler

### LINKED ISSUE #6456: app: plugin-manager IPC handler crashes on malformed JSON
## Describe the bug

The plugin-manager IPC handler in app/electron/main.ts parses its incoming data with JSON.parse and no try/catch:

    ipcMain.on('plugin-manager', async (event, data) => {
      const eventData = JSON.parse(data) as Action;
      ...
    });

If data isn't valid JSON, JSON.parse throws. Since the handler is async, that throw turns into an unhandled promise rejection in the Electron main process — nothing catches it, the renderer that sent the message gets no error back, and depending on the Node/Electron config it can take down the main process. Either way the failure is invisible and there's no graceful handling.

In normal use the frontend always sends well-formed JSON, so this won't fire day to day. But an IPC handler shouldn't crash the whole app on bad input — and other handlers in the same file already wrap their work in try/catch, so this one just looks like it's missing the same guard.

Expected: bad input should be caught and logged (or returned as an error to the renderer), not left to bubble up as an unhandled rejection.

## To Reproduce
Steps to reproduce the bug:
1. Send a 'plugin-manager' IPC message whose payload is not valid JSON (e.g. from a buggy caller or a plugin with IPC access)
2. JSON.parse throws inside the async handler
3. The throw becomes an unhandled promise rejection in the main process — no try/catch stops it, and nothing is surfaced to the caller

I reproduced the exact handler shape in a small script — a valid message is handled fine, but a malformed one throws in JSON.parse and the rejection reaches the process. Screenshot below.

 - Installation type: running from source (main branch), Electron desktop app
 - Other: the desktop app only — this is in app/electron/main.ts

Yes (I will propose a PR)

## Additional Context

The spot is the plugin-manager handler in app/electron/main.ts around line 267. Doesn't look like a regression — the missing try/catch has been there a while.

Attached a screenshot of a small repro that mirrors the handler: a valid message ({"action":"LIST"}) is handled fine, but a malformed one (not-json) throws in JSON.parse and the unhandled rejection reaches the process.

Image

krsatyamthakur-droid:
## Summary

The plugin-manager IPC handler parses the incoming event data with JSON.parse and no try/catch. Since the handler is async, a parse failure turns into an unhandled promise rejection in the Electron main process instead of a caught error — nothing tells the renderer that sent the message what went wrong.

In normal operation the frontend always sends well-formed JSON, so this doesn't fire during regular use. But an IPC handler shouldn't be able to take down the main process on bad input, and every other handler in this file already wraps its work in try/catch — this one was just missing the same guard.

This wraps the JSON.parse call in try/catch, logs the failure, and returns early instead of letting the exception escape uncaught.

## Related Issue

Fixes #6456

## Changes

- app/electron/main.ts: wrapped JSON.parse in the plugin-manager IPC handler in try/catch, logging and returning early on failure instead of letting it throw unhandled

## Steps to Test

1. Trigger the plugin-manager IPC channel with a non-JSON payload (e.g. from devtools: ipcRenderer.send('plugin-manager', 'not-json'))
2. Before this change: JSON.parse throws inside the async handler, which becomes an unhandled promise rejection in the main process
3. After this change: the error is caught, logged to the console, and the handler returns cleanly — no crash

N/A — error handling change, nothing visual.

## Notes for the Reviewer

I didn't add an automated test for this one. main.ts pulls in electron, yargs, MCPClient and a handful of other modules with real side effects at import time, and nothing else in this codebase imports main.ts directly in a test — so mocking all of that just for this one handler felt like more risk than it was worth. Happy to add one if you'd rather have it, just let me know what approach you'd prefer.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -265,7 +265,15 @@ class PluginManagerEventListeners {
    */
   setupEventHandlers() {
     ipcMain.on('plugin-manager', async (event, data) => {
-      const eventData = JSON.parse(data) as Action;
+      let eventData: Action;
+
+      try {
+        eventData = JSON.parse(data) as Action;
+      } catch (error) {
+        console.error('plugin-manager: failed to parse event data as JSON:', error);
+        return;
+      }
+
       const { identifier, action } = eventData;
       const updateCache = (progress: ProgressResp) => {
         const percentage = this.convertProgressToPercentage(progress);


illume (approved):
🎉 thanks!

PR #6477, opened 2026-07-11
frontend: SimpleTable: Fix pagination bug

### LINKED ISSUE #6476: Off-by-One Pagination Boundary Bug in SimpleTable
When using the `SimpleTable` component, if the total number of items in the table is an exact multiple of the `rowsPerPage` count (e.g. exactly 10 items, with `rowsPerPage` set to 10), navigating to the second page (page index `1` which is 0-indexed) leaves the user on a completely blank page.

This happens because the page boundary check expects `page * rowsPerPage > displayData.length` before correcting the page index. For exactly 10 items and page index `1`:
- `1 * 10 > 10` evaluates to `false`.
- The page index remains `1`.
- The component attempts to slice items from index `10` to `20`, returning an empty list, resulting in a blank table.

skools-here:
## Summary

Fixes a bug in  SimpleTable  where navigating to an empty page occurs when the total number of items is an exact multiple of the page size (e.g. exactly 10 items, viewing the second page).

## Related Issue

Fixes #6476

## Changes

- Updated page validation check in SimpleTable.tsx from  >  to  >=  to capture exact boundary alignments.
- Corrected the fallback page index calculation in SimpleTable.tsx to compute the maximum valid page index using Math.max(0, Math.floor((displayData.length - 1) / rowsPerPage)) .
- Updated the filtered data reset check in SimpleTable.tsx from  <  to  <=  to return the page index to  0  when filters matching exactly  page * rowsPerPage  items are applied.
- Added a new regression unit test in SimpleTable.test.tsx to verify correct page correction at page boundaries.

### DIFF
--- a/frontend/src/components/common/SimpleTable.test.tsx
+++ b/frontend/src/components/common/SimpleTable.test.tsx
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from '@mui/material/styles';
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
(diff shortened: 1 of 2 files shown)


skools-here:
ptal @illume

skools-here:
ptal @illume

itvi-1234:
I think we should take a look at the commit messages , not including fixes for previous commits in the same PR. The last couple of commits fall within that. Instead they should be added to the previous commit. This is so that when people review commit by commit they can understand it better. Otherwise they might be like "oh I spot a bug here, and here are some paragraphs explaining it"... then several commits later... "oh, they fixed it."

Btw also take a look at - https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#describe-your-changes

skools-here:
Thanks @itvi-1234  for the feedback! I've cleaned up the commit history by squashing commits.

skools-here:
ptal @illume

illume (changes requested):
Thanks for these changes.


<details>

</details>

PR #6509, opened 2026-07-14
frontend: ClusterSettings: Handle invalid json and add unit test

### LINKED ISSUE #6508: UI can crash when localStorage has invalid/corrupted JSON in loadClusterSettings
## Description

If the cluster settings JSON stored in `localStorage` is corrupted or invalid, calling the helper function `loadClusterSettings` crashes the rendering path because of an uncaught `JSON.parse` exception. 

## Proposed Solution

Wrap the `JSON.parse` call inside `loadClusterSettings` in a `try-catch` block:

skools-here:
## Summary

Wrap JSON parsing in `loadClusterSettings` with a try-catch block to prevent the Headlamp React UI rendering path from crashing if `localStorage` contains invalid or corrupted JSON.

## Related Issue

Fixes #6508

## Changes

- Modified clusterSettings.ts in clusterSettings.ts to safely handle JSON parsing exceptions.
- Added unit tests in clusterSettings.test.ts to verify correct loading, storing, and fallback parsing behavior.

### DIFF
--- a/frontend/src/helpers/clusterSettings.test.ts
+++ b/frontend/src/helpers/clusterSettings.test.ts
@@ -118,16 +118,37 @@ describe('clusterSettings', () => {
       expect(loadClusterSettings('prod.extra')).toEqual({});
     });
 
-    it('returns empty object and logs console error when the stored payload is not valid JSON', () => {
+    it('returns empty object and logs console warning when the stored payload is not valid JSON', () => {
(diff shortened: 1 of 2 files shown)


skools-here:
ptal @illume

skools-here:
@sniok @illume ptal

illume (approved):
🎉 thanks!

PR #6537, opened 2026-07-15
backend: portforward: Accept context in permission check and pod status helpers

srajang1805:
## Summary

Replace inline `context.Background()` with a `context.Context` parameter
in `checkPortForwardPermission` and `checkIfPodIsRunning`, enabling
callers to supply proper contexts for Kubernetes API calls.

## Changes

- Add `ctx context.Context` parameter to `checkPortForwardPermission`
  and `checkIfPodIsRunning`.
- Callers pass `context.Background()` matching previous behavior.
- Removes two hardcoded `context.Background()` calls.

## Testing

```bash
go test ./pkg/portforward/...   # PASS
golangci-lint run               # 0 issues


skoeva (commented):
idt this change is necessary - there's no real functional benefit to this change. closing now

PR #6573, opened 2026-07-19
backend: plugins: Fix delete for migrated catalog plugins

### LINKED ISSUE #4489: Allow all user-installed/development plugins to be deleted
As per #3882, I'm unable to uninstall ai-assistant in Headlamp. When running in dev mode, ai-assistant isn't shipped, so it should be able to get deleted.

Pritz395:
## Summary

- Fixes plugin deletion for catalog-installed plugins that still live under
the development `plugins/` directory but are listed as `type=user`
(migration from before `user-plugins/` existed).

- Previously, `DELETE /plugins/{name}?type=user` only looked in
`user-plugins/`, so deletes failed with:

`plugin 'headlamp_kubescape' not found in user-plugins directory`

`Delete()` now falls back to the development `plugins/` directory when a
`type=user` delete misses `user-plugins/`. Shipped plugins remain
undeletable.

Fixes #4489

## Changes

- `backend/pkg/plugins/plugins.go`: fall back to development dir for `type=user`
- `backend/pkg/plugins/plugins_test.go`: cover the migrated catalog plugin case

### DIFF
--- a/backend/pkg/plugins/plugins.go
+++ b/backend/pkg/plugins/plugins.go
@@ -295,8 +295,24 @@ func GeneratePluginPaths(
 
 // isCatalogInstalledPlugin checks if a plugin was installed via the catalog.
 // Catalog-installed plugins have isManagedByHeadlampPlugin: true in their package.json.
+// pluginName may come from a request path, so paths escaping pluginDir are rejected
+// before any file is read.
 func isCatalogInstalledPlugin(pluginDir, pluginName string) bool {
-	packageJSONPath := filepath.Join(pluginDir, pluginName, "package.json")
+	if pluginDir == "" {
+		return false
+	}
+
+	absDir, err := filepath.Abs(pluginDir)
+	if err != nil {
+		return false
+	}
+
+	pluginPath := filepath.Join(absDir, pluginName)
+	if !isSubdirectory(absDir, pluginPath) {
+		return false
+	}
+
+	packageJSONPath := filepath.Join(pluginPath, "package.json")
 
 	content, err := os.ReadFile(packageJSONPath) //nolint:gosec
 	if err != nil {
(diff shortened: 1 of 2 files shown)


vtushar06:
hey @illume reviewed this end to end. since it's delete-by-name i checked the traversal side first: the type=user fallback routes through tryDeletePlugin, which keeps the isSubdirectory guard, so a name with ../ can't escape the dev dir, and shipped plugins stay unreachable (separate dir, never passed to Delete). the branch logic traces cleanly and the added test covers the migrated type=user case. ran `go test ./pkg/plugins/` locally and it's green.

one small thing: the `not found in user-plugins directory` error now also fires after the dev-dir fallback runs, so it reads slightly off. something like `not found in user-plugins or development` would match what was actually checked.

Pritz395:
Thanks for the review - I updated the type=user not-found message to `not found in user-plugins or development directory` so it matches the directories that were actually checked.

vtushar06:
@Pritz395 nice, now I think is ready to go.

Pritz395:
@vtushar06 yupp! I think I'm just awaiting maintainer feedback/merge now

Pritz395:
@illume Please feel free to review it once again. Thank you!

itvi-1234:
Hi @Pritz395 , Thanks for the PR , we generally follow Linux kernel style commits - https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#describe-your-changes

Also ,Could you please rebase your PR

illume (changes requested):
Thanks for the contribution.


<details>

</details>

illume (changes requested):
Thanks for working on this.


Pritz395:
@illume Rebased onto latest main (dropped the merge commit) and addressed the Copilot notes: type=user now only falls back to development for catalog-installed plugins (isManagedByHeadlampPlugin), and the tests cover that. Please do take a look at it, thank you!

Pritz395:
@illume isCatalogInstalledPlugin now rejects names that escape the plugin directory before reading package.json, with a regression test. Also rebased onto latest main.

skoeva (approved):
tested and works for me, lgtm

PR #6602, opened 2026-07-22
backend: drainNode: Skip DaemonSet pods via ownerReferences

### LINKED ISSUE #5736: backend: Node drain deletes `DaemonSet` pods on Kubernetes 1.16+ due to removed label check
## Describe the bug

The node drain operation uses the deprecated label `kubernetes.io/created-by:
daemonset-controller` to identify and skip DaemonSet pods. This label was
deprecated in Kubernetes 1.8 and removed in 1.16. On any cluster running
Kubernetes 1.16+, the filter silently does nothing, causing the drain to delete
DaemonSet pods it should skip -- such as kube-proxy, CNI agents, and log
collectors. This can break cluster networking and observability during a drain.

The correct approach is to check `pod.OwnerReferences` for an entry with
`Kind: "DaemonSet"`.

Affected code: `drainNode` in `backend/cmd/headlamp.go` (line 2753).

## To Reproduce

1. Connect Headlamp to a Kubernetes cluster running version 1.16 or later.
2. Ensure at least one node has DaemonSet-managed pods (e.g. kube-proxy,
   a CNI plugin pod, or a log collector).
3. Navigate to the node's detail page and trigger a drain operation.
4. Observe that DaemonSet pods are deleted instead of being skipped.

- Installation type: [e.g. Container-Image / In-Cluster / Helm / Linux-Tarball]
- Headlamp Version: [X.Y.Z]
- Other: Reproducible on any Kubernetes cluster >= 1.16, which covers all
  actively supported versions.

Yes (I will propose a PR).

> Do you want to help fix this issue? Please see our
> [contribution docs](https://headlamp.dev/docs/latest/contributing). For any
> non-trivial contributions, we advise you to join
> [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss
> the implementation with Headlamp's core developers.

## Additional Context

The `kubernetes.io/created-by` annotation was deprecated in Kubernetes 1.8
(https://github.com/kubernetes/kubernetes/pull/54445) and removed in 1.16. All
modern clusters use `ownerReferences` on the pod object to express ownership.
The fix is to replace the label check with:

```go
func isDaemonSetPod(pod corev1.Pod) bool {
    for _, ref := range pod.OwnerReferences {
        if ref.Kind == "DaemonSet" {
            return true
        }
    }
    return false
}

### PR DESCRIPTION
**Pull request by Pritz395**
## Summary

Node drain used the deprecated `kubernetes.io/created-by: daemonset-controller` label to skip DaemonSet pods. That label was removed in Kubernetes 1.16, so on modern clusters the filter silently did nothing and drain deleted kube-proxy, CNI agents, and other DaemonSet pods that should be skipped.

This change detects `apps/v1` DaemonSet owner references instead. It prefers the controlling owner and safely falls back to a DaemonSet owner only when no controller reference exists.

Fixes #5736

## Changes

- Replace the removed label check with `apps/v1` owner-reference detection
- Preserve pods controlled by DaemonSets during drain
- Cover controller, fallback, multiple-owner, and invalid API cases
- Verify drain deletes regular pods while retaining DaemonSet pods

## Steps to Test

1. Connect Headlamp to a cluster with DaemonSet pods such as kube-proxy.
2. Drain a node that runs those pods.
3. Confirm DaemonSet pods remain and regular pods are deleted.
4. Run:

```bash
cd backend && go test ./cmd/ -run 'TestIsDaemonSetPod|TestDrainNode'
```

## Notes for the Reviewer

- The deprecated label alone no longer classifies a pod as DaemonSet-owned.
- Owner references from APIs other than `apps/v1` are rejected.
- Drain tests use realistic owner references and assert pod retention/deletion.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -64,6 +64,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -2918,6 +2919,32 @@ func (c *HeadlampConfig) drainNode(
 	}()
 }
 
+// isDaemonSetPod reports whether pod is managed by a DaemonSet. It replaces the
+// kubernetes.io/created-by label, which was removed in Kubernetes 1.16.
+//
+// A DaemonSet controller reference is the normal case. A pod owned by some other
+// controller is not a DaemonSet pod even if it also lists a DaemonSet owner. When
+// no controller reference is set at all, a plain DaemonSet owner still counts,
+// because drain must not delete a pod nothing will recreate.
+func isDaemonSetPod(pod corev1.Pod) bool {
+	if controller := v1.GetControllerOf(&pod); controller != nil {
+		return isDaemonSetOwner(*controller)
+	}
+
+	for _, ref := range pod.OwnerReferences {
+		if isDaemonSetOwner(ref) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func isDaemonSetOwner(ref v1.OwnerReference) bool {
+	return ref.APIVersion == appsv1.SchemeGroupVersion.String() &&
+		ref.Kind == "DaemonSet"
+}
+
 func (c *HeadlampConfig) drainNodePods(
 	ctx context.Context,
 	clientset kubernetes.Interface,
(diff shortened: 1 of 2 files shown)


Pritz395:
@illume Resolved the comment -please feel free to re-review. Thank you!

illume (changes requested):
Thanks for these changes.


<details>

</details>

illume (changes requested):
Thanks for this PR.


Pritz395:
@illume Rebased onto latest main (dropped the merge commit) and addressed the Copilot notes: DaemonSet detection now uses the controller OwnerReference via GetControllerOf, and the test fixture sets Controller=true. Please do take a look, thank you!

Pritz395:
@illume Rebased onto the latest main. For the DaemonSet detection, this and the earlier review comment were kind of pulling in opposite directions, so `isDaemonSetPod` now handles both cases.

This is slightly more conservative than `kubectl drain`, which only looks at the controller reference.

Pritz395:
@illume @yolossn @vyncent-t Please do take a look at this PR. Thank you!

illume:
Sorry, I noticed there was already an open PR started linked from the issue.

Pritz395:
@illume It's cool. Thank you!

PR #6611, opened 2026-07-22
backend: helm: Skip invalid repository cache

### LINKED ISSUE #6608: Helm chart listing fails when a repository cache cannot be loaded
## Problem

The Helm charts listing endpoint currently fails if any configured
repository cache cannot be loaded.

In `listCharts()`, the cached repository index is loaded using
`repo.LoadIndexFile()`. If a single repository cache is missing or
corrupted, the function returns an error immediately, causing the API to
respond with `500 Internal Server Error`, even when other configured
repositories are valid.

## Expected behavior

The endpoint should continue listing charts from repositories whose
cached indexes can be loaded, rather than failing the entire request
because of a single invalid repository cache.

## Steps to reproduce

1. Configure multiple Helm repositories.
2. Remove or corrupt the cached `index.yaml` for one repository.
3. Request the Helm charts endpoint.
4. Observe that the request returns HTTP 500.

## Expected result

Charts from valid repositories are returned, while the repository with
the invalid cache is skipped or otherwise handled gracefully.

mukeshdhadhariya:
## Summary

This PR makes the Helm chart listing endpoint resilient to partial
repository failures.

Previously, if `repo.LoadIndexFile()` failed for any configured
repository, the request returned an error and no charts were listed.

With this change, the backend logs a warning, skips the affected
repository, and continues listing charts from the remaining valid
repositories.

Fixes #6608 

## Changes

```go
indexFile, err := repo.LoadIndexFile(repoIndexFile)
if err != nil {
    logger.Log(logger.LevelWarn, nil, err, "loading repo index")
    continue
}
```

### DIFF
--- a/backend/pkg/helm/charts.go
+++ b/backend/pkg/helm/charts.go
@@ -60,7 +60,8 @@ func listCharts(filter string, settings *cli.EnvSettings) ([]chartInfo, error) {
 
 		indexFile, err := repo.LoadIndexFile(repoIndexFile)
 		if err != nil {
-			return nil, err
+			logger.Log(logger.LevelWarn, nil, err, "loading repo index")
+			continue
 		}
 
 		index.AddRepo(name, indexFile, true)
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for this PR.


mukeshdhadhariya:
PR #6625 has resolve this with proper ways @illume plz review

PR #6638, opened 2026-07-25
make: build-plugins-container: Follow host platform by default

### LINKED ISSUE #6632: make build-plugins-container ignores host architecture on Apple Silicon
## Describe the bug

`make build-plugins-container` hardcodes `--platform=linux/amd64`, so on an Apple Silicon Mac it produces an amd64 image regardless of the host. The platform is written inline rather than through a variable, so it cannot be overridden from the environment either.

This is the same defect as #4578 ("make image doesn't work by default on silicon mac"), which was fixed in #4657 by adding host architecture detection to the Makefile. That fix only covered the `image` target; `build-plugins-container` was missed.

Makefile:14-27 detects the host and sets `DOCKER_PLATFORM` accordingly:

```make
    ifeq ($(UNAME_S),Darwin)
        UNAME_M := $(shell uname -m)
        ifeq ($(UNAME_M),arm64)
            DOCKER_PLATFORM ?= linux/arm64
```

but Makefile:399-406 ignores it:

```make
build-plugins-container:
	$(DOCKER_CMD) $(DOCKER_BUILDX_CMD) build \
	--pull \
	--platform=linux/amd64 \
	--push=$(DOCKER_PUSH) \
	-t $(DOCKER_REPO)/$(DOCKER_PLUGINS_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
	Dockerfile.plugins \
	.
```

The practical impact is that you cannot reproduce the CI e2e run locally on an Apple Silicon Mac using the documented commands. What makes this awkward to diagnose is that nothing reports an error until the very end: `kind load docker-image` exits 0 and prints its usual "loading..." message, but the mismatched-architecture image is silently not added to the node. The failure only surfaces later as `ErrImageNeverPull` on the pod, which reads like a tagging or pull-policy mistake rather than an architecture mismatch.

Expected: `make build-plugins-container` builds for the host architecture by default, the way `make image` does, and honours an override variable.

## To Reproduce

Steps to reproduce the bug, on an Apple Silicon Mac:

1. Build both images the way `.github/workflows/build-container.yml` does:
   ```
   DOCKER_IMAGE_VERSION=latest make image
   DOCKER_IMAGE_VERSION=latest DOCKER_PLUGINS_IMAGE_NAME=headlamp-plugins-test make build-plugins-container
   ```
2. Compare the architectures of the two resulting images:
   ```
   docker image inspect ghcr.io/headlamp-k8s/headlamp:latest --format '{{.Os}}/{{.Architecture}}'
   docker image inspect ghcr.io/headlamp-k8s/headlamp-plugins-test:latest --format '{{.Os}}/{{.Architecture}}'
   ```
   The first reports `linux/arm64`, correctly following the host detection added in #4657. The second reports `linux/amd64`.

3. Confirm the platform cannot be overridden from the environment:
   ```
   DOCKER_PLATFORM=linux/arm64 DOCKER_IMAGE_VERSION=ovrtest \
     DOCKER_PLUGINS_IMAGE_NAME=headlamp-plugins-ovrtest make build-plugins-container
   docker image inspect ghcr.io/headlamp-k8s/headlamp-plugins-ovrtest:ovrtest --format '{{.Os}}/{{.Architecture}}'
   ```
   Still reports `linux/amd64`.

4. Load the amd64 image into an arm64 kind cluster and observe that it appears to succeed:
   ```
   kind create cluster --name test
   kind load docker-image ghcr.io/headlamp-k8s/headlamp-plugins-test:latest --name test
   echo "exit: $?"
   ```
   Prints `Image ... not yet present on node "test-control-plane", loading...` and exits 0.

5. Check whether it actually landed on the node:
   ```
   docker exec test-control-plane crictl images | grep plugins
   ```
   The amd64 image is absent. An arm64 build of the same image, loaded the same way, is present.

6. Deploy and watch it fail:
   ```
   envsubst < e2e-tests/kubernetes-headlamp-ci.yaml | kubectl --context=test apply -f -
   kubectl --context=test get pods -n kube-system -l app.kubernetes.io/name=headlamp
   ```
   The `headlamp-plugins` init container reports:
   ```
   Warning  ErrImageNeverPull  kubelet  Container image "..." is not present with pull policy of Never
   Warning  Failed             kubelet  Error: ErrImageNeverPull
   ```
   and the deployment never reaches Available. Note the manifest sets `imagePullPolicy: Never`, which is correct for a kind-loaded image; the image genuinely is not there.

Running the equivalent buildx command by hand with the correct platform is enough to fix it, which confirms the platform flag is the only problem:

```
docker buildx build --pull --platform=linux/arm64 --push=false --load \
  -t ghcr.io/headlamp-k8s/headlamp-plugins-test:latest -f Dockerfile.plugins .
```

With that image in place the deployment comes up and the full e2e suite passes locally: 35 passed, 11 skipped.

 - Installation type: Development environment, building container images locally
 - Other: macOS 26.5.2 arm64 (M-series), Docker 27.5.1, kind 0.32.0

Yes (I will propose a PR).

The fix mirrors what #4657 did for `image`: add a `DOCKER_PLUGINS_PLATFORM ?= $(DOCKER_PLATFORM)` variable and use it in the target, so host detection applies by default and an explicit platform can still be pinned.

## Additional Context

For reviewers: CI runs on linux/amd64 runners, where the detection in Makefile:14-27 resolves to `local` rather than an explicit platform, so this should be a no-op there. Any fix should leave the release pipeline's published image behaviour unchanged and only affect what a developer gets by default on their own machine. Worth confirming whether `local` is the desired value for the plugins image in CI, or whether that build should stay explicitly pinned to amd64.

unlikelyzero:
## Summary

This PR fixes the plugins container build ignoring the host architecture. `build-plugins-container` hardcoded `--platform=linux/amd64`, so on an Apple Silicon Mac it produced an amd64 image while `make image` correctly produced arm64. Because the flag was written inline rather than through a variable, it could not be overridden from the environment either.

This is the same defect as #4578, which was fixed for the `image` target in #4657 by adding host detection. That change did not cover `build-plugins-container`.

The failure mode is quiet and easy to misread. `kind load docker-image` exits 0 and prints its usual loading message, but the mismatched-architecture image is never added to the node. The problem only surfaces later as `ErrImageNeverPull` on the pod, which looks like a tagging or pull-policy mistake rather than an architecture mismatch. The practical effect is that the CI e2e run cannot be reproduced locally on an Apple Silicon Mac, because the plugins image is an init container in `e2e-tests/kubernetes-headlamp-ci.yaml` and the Headlamp pod never becomes ready.

## Related Issue

Fixes #6632

## Changes

- Added `DOCKER_PLUGINS_PLATFORM ?= $(DOCKER_PLATFORM)`, defaulting to the platform already derived from the host in Makefile:13-28
- Updated `build-plugins-container` to use `--platform=$(DOCKER_PLUGINS_PLATFORM)` instead of the hardcoded `linux/amd64`

Four lines changed in one file.

## Steps to Test

Resolution of the platform value, which can be checked on any host without building:

```
# 1. Default now follows the host (linux/arm64 on Apple Silicon, previously linux/amd64)
make -n build-plugins-container | tr -s ' \\\n' ' ' | grep -o '\-\-platform=[^ ]*'

# 2. Explicitly pinnable, which was not possible before
DOCKER_PLUGINS_PLATFORM=linux/amd64 make -n build-plugins-container | tr -s ' \\\n' ' ' | grep -o '\-\-platform=[^ ]*'

# 3. Also inherits an existing DOCKER_PLATFORM override
DOCKER_PLATFORM=linux/amd64 make -n build-plugins-container | tr -s ' \\\n' ' ' | grep -o '\-\-platform=[^ ]*'

# 4. Linux/CI path is unchanged: resolves to `local`, which is what the image target already passes there
make -n build-plugins-container DOCKER_PLATFORM=local | tr -s ' \\\n' ' ' | grep -o '\-\-platform=[^ ]*'
```

Expected: `linux/arm64`, `linux/amd64`, `linux/amd64`, `local`.

End to end on an Apple Silicon Mac:

```
# 5. Build and confirm the image architecture
DOCKER_IMAGE_VERSION=fixtest DOCKER_PLUGINS_IMAGE_NAME=headlamp-plugins-fixtest make build-plugins-container
docker image inspect ghcr.io/headlamp-k8s/headlamp-plugins-fixtest:fixtest --format '{{.Os}}/{{.Architecture}}'
# -> linux/arm64   (before this change: linux/amd64)

# 6. Confirm it actually lands on an arm64 kind node, which the amd64 build did not
kind create cluster --name test
kind load docker-image ghcr.io/headlamp-k8s/headlamp-plugins-fixtest:fixtest --name test
docker exec test-control-plane crictl images | grep fixtest
# -> present. With the previous amd64 build this grep returned nothing, despite kind load exiting 0.
```

I ran all six of the above on macOS 26.5.2 arm64 with Docker 27.5.1 and kind 0.32.0. With the corrected image in place the full e2e suite passes locally: 35 passed, 11 skipped.

Not applicable, build tooling change.

## Notes for the Reviewer

- CI behaviour should be unchanged. On Linux the detection in Makefile:13-28 resolves `DOCKER_PLATFORM` to `local`, which is exactly what the `image` target already passes on those runners, so the plugins build gets the same value it effectively had before. Worth a second opinion on whether `local` is what you want there, or whether that build should stay explicitly pinned to `linux/amd64`; if the latter is preferred I am happy to set `DOCKER_PLUGINS_PLATFORM=linux/amd64` in `build-container.yml` instead of relying on the default.
- `build-plugins-container` is invoked from exactly one place, the e2e job at `.github/workflows/build-container.yml:137`. It is not used by any release or publish workflow, so there is no impact on published images.
- I deliberately used a separate `DOCKER_PLUGINS_PLATFORM` variable rather than reusing `DOCKER_PLATFORM` directly in the target, so the plugins image can still be cross-built independently of the main image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ else
         DOCKER_PLATFORM ?= local
     endif
 endif
+# The plugins container follows the host platform by default, like the main
+# image. Override to cross-build, e.g. DOCKER_PLUGINS_PLATFORM=linux/amd64.
+DOCKER_PLUGINS_PLATFORM ?= $(DOCKER_PLATFORM)
 DOCKER_PUSH ?= false
 EMBED_BINARY_NAME := headlamp_app
 # Get version and app name from app/package.json
@@ -399,7 +402,7 @@ image-verify-digests:
 build-plugins-container:
 	$(DOCKER_CMD) $(DOCKER_BUILDX_CMD) build \
 	--pull \
-	--platform=linux/amd64 \
+	--platform=$(DOCKER_PLUGINS_PLATFORM) \
 	--push=$(DOCKER_PUSH) \
 	-t $(DOCKER_REPO)/$(DOCKER_PLUGINS_IMAGE_NAME):$(DOCKER_IMAGE_VERSION) -f \
 	Dockerfile.plugins \


illume (approved):
🎉 thanks!

Issue #6640, opened 2026-07-25
app/e2e-tests README is missing the compile-electron step


unlikelyzero:
## Describe the bug

`app/e2e-tests/README.md` is missing a required setup step, so following it does not run the tests.

`app/package.json` declares `"main": "build/main.js"`, and the specs launch the app with `electron .`, which resolves that field. `app/build/main.js` does not exist in a fresh checkout and nothing in the README or the specs creates it. It is produced by `npm run compile-electron -- --dev`, which today is only reachable via `npm run dev`.

The README's setup is just:

```
cd headlamp/app/e2e-tests
npm install
npm run test-app
```

Expected: following the README runs the tests. Actual: the app has no entry point to load, and the run fails.

## To Reproduce

1. From a fresh checkout, follow the README exactly:
   ```
   cd app/e2e-tests
   npm install
   npx playwright install chromium
   npm run test-app
   ```
2. Confirm the entry point is missing:
   ```
   node -p "require('./app/package.json').main"   # build/main.js
   ls app/build/main.js                           # No such file or directory
   ```
3. Add the missing step and the app starts:
   ```
   cd app && npm run compile-electron -- --dev
   ls -la build/main.js                           # now exists
   ```

## Environment (please provide info about your environment):
 - Installation type: Development environment
 - Headlamp Version: main at 2f091cf34 (0.43.0)
 - Other: macOS 26.5.2 arm64, node v24.14.1, npm 11.11.0

## Are you able to fix this issue?
Yes (I can propose a PR).

Either document the step in `app/e2e-tests/README.md`, or make `test-app` depend on it so the suite cannot run against a missing entry point:

```json
"test-app": "npm --prefix .. run compile-electron -- --dev && PLAYWRIGHT_TEST_MODE=app playwright test"
```

I would rather not guess which fits the project's preferences.

## Additional Context

`compile-electron --dev` is necessary but not sufficient for a green run. With `ELECTRON_DEV=true`, `app/electron/main.ts` also resolves `../frontend/build/index.html` (main.ts:91) and `../backend/headlamp-server` (main.ts:772), so the frontend and backend need building too, and the tests need a minikube cluster named `minikube`. The full prerequisite chain is written up in #6639.

[...]

Issue #6639, opened 2026-07-25
Desktop e2e suite (app/e2e-tests) does not run in CI


unlikelyzero:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.

The desktop e2e suite in `app/e2e-tests` does not run in any workflow, so the four tests it contains provide no automated protection.

```
clusterAutoConnect.spec.ts  2 tests   app mode only (skips unless PLAYWRIGHT_TEST_MODE=app)
clusterRename.spec.ts       1 test    app mode only (skips unless PLAYWRIGHT_TEST_MODE=app)
namespaces.spec.ts          1 test
```

Searching `.github/workflows/` for `app/e2e-tests`, `app:test:e2e`, `test-app`, `test-web` or `app:test` returns nothing. The only workflow that runs Playwright is `build-container.yml`, which runs the web suite in `e2e-tests/`.

Worth being precise about one thing, because it looks like coverage but is not: `app.yml` does run `make app-test` on Linux, but that Makefile target is

```make
app-test:
	cd app && npm install
	cd app && npm run test
```

and `app/package.json` defines `"test": "vitest run"`. So `make app-test` runs the Electron unit tests only. It is not the same as the root `npm run app:test` script, and it never reaches the e2e suite.

What is lost is specifically the desktop behaviour the web suite cannot cover: cluster rename, and cluster auto-connect via link click. Three of the four tests are gated behind `PLAYWRIGHT_TEST_MODE=app`, so even running the suite in web mode would not exercise them.

This also has a concrete downstream effect. #6633 (`npm run app:test:e2e` fails with `Missing script: "test"`) went unnoticed precisely because nothing runs this suite.

I want to be upfront that this may well be a deliberate decision. Running these four tests in CI is not cheap, for the reasons below, and a maintainer may have already weighed that. I am filing this to make the gap visible and to collect the requirements in one place, not to assert that it must be fixed.

## Describe the solution you'd like

Run the desktop e2e suite in CI, most naturally as a sibling job to `build-linux` in `app.yml`, which already provides ubuntu-22.04, Node 22 and Go.

Below is what I worked out by reading the specs and `app/electron/main.ts`. I have not yet run the suite end to end, so treat the requirements as verified from source and the timings as estimates.

### Prerequisites the suite needs

The specs do not launch the packaged app. They launch the dev Electron binary:

```js
_electron.launch({
  cwd: app/,
[...]

PR #6641, opened 2026-07-26
app: e2e-tests: Fix the desktop suite and stop it mutating the kubeconfig

### LINKED ISSUE #6648: Desktop e2e suite fails and modifies the developer's kubeconfig
## Describe the bug

The app-mode e2e suite in `app/e2e-tests` does not pass, and running it modifies the developer's real `~/.kube/config`. Because that change is never reverted, the suite can only pass once on a given machine.

Running it on macOS against a minikube cluster: 2 of 4 tests fail.

The kubeconfig mutation is the most serious part. `clusterRename.spec.ts` renames the cluster through the UI, which persists a `headlamp_info` extension into the user's kubeconfig:

```yaml
- context:
    cluster: minikube
    extensions:
      - extension:
          customName: test-cluster
        name: headlamp_info
  name: minikube
```

Nothing removes it. Every later run then sees a cluster named `test-cluster`, so the specs that expect `minikube` fail. Separately, `minikube delete` in `clusterAutoConnect`'s teardown clears `current-context`, which leaves `kubectl` unusable afterwards (`The connection to the server localhost:8080 was refused`).

Expected: the suite passes, is repeatable, and does not modify the developer's kubeconfig.

## To Reproduce

Requires a minikube cluster named `minikube` and the prerequisites in #6639 and #6640.

1. Run the suite:
   ```
   cd app/e2e-tests
   PLAYWRIGHT_TEST_MODE=app npx playwright test
   ```
   Two tests fail: `clusterRename` and `namespaces`.

2. Observe the kubeconfig was modified and never restored:
   ```
   grep -c customName ~/.kube/config      # 1
   kubectl config current-context         # error: current-context is not set
   ```

3. Run it again. It now fails differently, because the cluster is named `test-cluster`:
   ```
   PLAYWRIGHT_TEST_MODE=app npx playwright test
   ```

## Root causes

Seven, each of which masked the next:

1. `clusterRename` persists `customName` into the user's kubeconfig and never reverts it.
2. `minikube delete` in `clusterAutoConnect` teardown clears `current-context`.
3. `clusterRename` and `namespaces` never close their Electron app. The app takes a single-instance lock (`app/electron/main.ts:1634`), so the next spec file's launch is denied it and quits, reported as `electron.launch: Process failed to launch!`.
4. `namespaces.spec.ts:58` uses `test.setTimeout(0)`, disabling the test timeout, so a failure there hangs instead of failing.
5. `reporter: 'html'` blocks the run on `open --wait-apps` after a failure.
6. `getByRole('button', { name: 'Yes' })` matches nothing: `ConfirmDialog` sets `aria-label="confirm-button"`, which becomes the accessible name and overrides the visible text. The namespace delete dialog also reads "Cancel"/"Delete" rather than "No"/"Yes".
7. Two selectors have drifted from the UI: `namespacesPage` clicks a "Cluster" sidebar group that no longer exists (Namespaces is top-level now), and `getByRole('button', { name: 'Apply' })` matches both "Apply" and "Create / Apply", a strict mode violation.

 - Other: macOS 26.5.2 arm64, minikube v1.38.1, node v24.14.1

Yes, proposed in #6641.

That PR makes `clusterRename` and `namespaces` run against a throwaway copy of the kubeconfig in `os.tmpdir()`, which is what `clusterAutoConnect` already did, and fixes the other six. Result is 4 passed in about a minute, repeatable, leaving `current-context` and the kubeconfig unchanged and no `testing-e2e` namespace behind.

## Additional Context

This went unnoticed because nothing runs this suite; see #6639. Everything above was verified on macOS only, since the suite has never run on Linux or in CI.

unlikelyzero:
## Summary

This PR fixes the app-mode e2e suite in `app/e2e-tests`, which did not pass and which modified the developer's real kubeconfig as a side effect. Because that mutation was never reverted, the suite could only ever pass once on a given machine; every run after the first found a cluster named `test-cluster` instead of `minikube`.

Before: 2 of 4 tests failing, roughly 1.8 minutes, with a failure mode that looked like an indefinite hang.
After: 4 of 4 passing in about a minute, repeatable, and leaving no state behind.

There were seven distinct problems. They had to be fixed together, because each one masked the next.

## Related Issue

Fixes #6648

Also related to #6639, which asks whether this suite should run in CI and lists what that would take. Two of the fixes here (the disabled timeout and the cross-spec state mutation) are the robustness problems that issue identifies as worth fixing first, so this is a prerequisite rather than the CI change itself.

This PR does not add a workflow. That decision is still open in #6639.

## Changes

**Test isolation and reporting**

- `clusterRename` persisted `customName` as a `headlamp_info` extension in `~/.kube/config` and never reverted it. Both `clusterRename` and `namespaces` now run against a throwaway copy of the kubeconfig in `os.tmpdir()`, which is the approach `clusterAutoConnect` already used. The user's kubeconfig is no longer written to at all.
- `minikube delete` clears `current-context`, which left `kubectl` unusable after a run (`connection refused to localhost:8080`). `clusterAutoConnect` now records and restores it.
- `clusterRename` and `namespaces` never closed their Electron app. The app takes a single-instance lock (`app/electron/main.ts:1634`), so the next spec file's launch was denied the lock and quit immediately, surfacing as `electron.launch: Process failed to launch!`. Both now close the app in `afterAll`.
- `namespaces.spec.ts` used `test.setTimeout(0)`, disabling the test timeout entirely, so any failure in that file hung forever instead of failing. In CI that would burn a job to the workflow limit rather than reporting a failure. It now has a real timeout.
- The html reporter blocked the run on `open --wait-apps` after a failure. I lost 85 minutes to a run that had actually finished in 40 seconds. Reporting is now `list` plus `html` with `open: 'never'`, so runs stream progress and always terminate. This does not change CI behaviour, where Playwright already suppresses auto-open.

**Selectors that had drifted from the UI**

- `ConfirmDialog` sets `aria-label="confirm-button"` / `"cancel-button"`, which becomes the accessible name and overrides the visible text, so `getByRole('button', { name: 'Yes' })` matched nothing. These now query the same accessible name that the component's own unit tests use (`frontend/src/components/common/ConfirmButton.test.tsx`).
- The namespace delete confirmation reads "Cancel"/"Delete" rather than "No"/"Yes", so keying off the aria-label covers both dialogs.
- Namespaces is a top-level sidebar entry now. The page object still clicked through a "Cluster" group, which no longer exists outside Storybook stories.
- `getByRole('button', { name: 'Apply' })` matched both "Apply" and "Create / Apply", a strict mode violation. Made exact.

## Steps to Test

Requires a running minikube cluster named `minikube`, plus the prerequisites the suite needs (these are not currently documented; see #6640):

```
# 1. Build what the app loads in dev mode
make frontend                       # -> frontend/build/index.html
make backend                        # -> backend/headlamp-server
cd app && npm install
npm run compile-electron -- --dev   # -> app/build/main.js
cd e2e-tests && npm install && npx playwright install chromium

# 2. Run it
PLAYWRIGHT_TEST_MODE=app npx playwright test
```

Expected: `4 passed`, exit 0, in roughly a minute.

To confirm the isolation actually holds, which is the main point of this PR:

```
# Before and after a run, all three should be unchanged:
grep -c customName ~/.kube/config     # 0
kubectl config current-context        # minikube
kubectl get ns testing-e2e            # not found

# And it should be repeatable, which it was not before:
PLAYWRIGHT_TEST_MODE=app npx playwright test   # 4 passed
PLAYWRIGHT_TEST_MODE=app npx playwright test   # 4 passed again
```

I also ran it with `CI=true` to exercise `retries: 2` and `forbidOnly`; it passes and terminates cleanly.

Not applicable, test-only change.

## Notes for the Reviewer

- **The aria-label queries look odd on purpose.** `getByRole('button', { name: 'confirm-button' })` reads strangely, but it is the actual accessible name, and it is what the component's unit tests already assert against. Worth flagging separately: `aria-label="confirm-button"` overriding the visible "Yes" means a screen reader announces "confirm button" rather than "Yes". That looks like a real accessibility problem in `ConfirmDialog`, not just a test inconvenience. I have not filed that, and did not want to change component behaviour inside a test-only PR. Happy to open an issue if you agree it is one.
- **The kubeconfig isolation changes what the specs talk to.** They now see a minified, flattened copy of the `minikube` context rather than the developer's full kubeconfig. That is deliberate: it stops the suite mutating a real file, and it means the specs no longer depend on whatever else happens to be in the developer's config. The context is still named `minikube`, so the specs' expectations are unchanged.
- **Not validated on Linux.** Everything here was verified on macOS/arm64. The suite has never run in CI, so there may be Linux-specific issues (Electron sandbox flags, missing GTK/NSS libraries, xvfb behaviour) that only appear the first time it runs on a runner. That is one reason this PR deliberately stops short of adding a workflow.
- `app/e2e-tests` is outside the scope of `npm run app:lint` and its `format-check`, though the pre-commit hook does cover `app/**/*.{ts,tsx}`. These files were also checked directly against the same Prettier config.

### DIFF
--- a/app/e2e-tests/playwright.config.ts
+++ b/app/e2e-tests/playwright.config.ts
@@ -46,7 +46,9 @@ export default defineConfig({
   /* Opt out of parallel tests on CI. */
   workers: 1,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: 'html',
+  // 'list' streams progress; the html report is still written but never
+  // auto-opened, which otherwise blocks the run on `open --wait-apps`.
+  reporter: [['list'], ['html', { open: 'never' }]],
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
--- a/app/e2e-tests/tests/clusterAutoConnect.spec.ts
+++ b/app/e2e-tests/tests/clusterAutoConnect.spec.ts
@@ -34,7 +34,7 @@
  */
 
 import { expect, test } from '@playwright/test';
-import { execSync } from 'child_process';
+import { execFileSync, execSync } from 'child_process';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
@@ -56,7 +56,17 @@ function shell(cmd: string): string {
   return execSync(cmd, { encoding: 'utf8', stdio: ['pipe', 'pipe', 'inherit'] }).trim();
 }
 
+// `minikube delete` clears current-context in the user's kubeconfig, which
+// leaves kubectl unusable afterwards. Remember it so teardown can put it back.
+let previousContext = '';
+
 function setupExecCluster(): void {
+  try {
+    previousContext = shell('kubectl config current-context');
+  } catch {
+    /* no current context set; nothing to restore */
+  }
+
   // Start (or reconnect to) the dedicated minikube profile.
   // minikube automatically adds the context to the kubeconfig in ORIGINAL_KUBECONFIG.
   shell(`minikube start --profile ${CLUSTER_NAME}`);
@@ -72,7 +82,7 @@ function setupExecCluster(): void {
   // Use the kubeconfig that minikube already wrote (cert-based auth, proven to work).
   // Export it to a standalone file so the test doesn't inherit unrelated contexts.
   const exported = shell(`kubectl --context ${CLUSTER_NAME} config view --minify --raw --flatten`);
-  fs.writeFileSync(MERGED_KUBECONFIG, exported);
+  fs.writeFileSync(MERGED_KUBECONFIG, exported, { mode: 0o600 });
 }
 function teardownExecCluster(): void {
   // Stop and delete the minikube profile (also removes its kubeconfig entries).
@@ -84,6 +94,17 @@ function teardownExecCluster(): void {
   for (const f of [EXEC_KUBECONFIG, MERGED_KUBECONFIG]) {
     try {
       fs.unlinkSync(f);
+    } catch {
+      /* ignore */
+    }
+  }
+
+  // Restore the context `minikube delete` cleared.
+  if (previousContext) {
+    try {
+      execFileSync('kubectl', ['config', 'use-context', previousContext], {
+        stdio: ['pipe', 'pipe', 'inherit'],
+      });
     } catch {
       /* ignore */
     }
--- a/app/e2e-tests/tests/clusterRename.spec.ts
+++ b/app/e2e-tests/tests/clusterRename.spec.ts
@@ -15,6 +15,9 @@
  */
 
 import { expect, test } from '@playwright/test';
+import { execSync } from 'child_process';
+import fs from 'fs';
+import os from 'os';
 import path from 'path';
 import { _electron, Page } from 'playwright';
 import { HeadlampPage } from './headlampPage';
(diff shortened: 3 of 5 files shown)


unlikelyzero:
Heads up on an interaction with #6653.

This PR changes the e2e queries to `getByRole('button', { name: 'confirm-button' })` and `button[aria-label="confirm-button"]`, which matches `main` as it stands today. #6653 fixes the underlying accessibility problem by moving those `aria-label` values to `data-testid`, so the accessible name becomes the visible text.

The two do not conflict textually, they touch different files, but whichever merges second needs a small follow-up:

- If this PR merges first, #6653 should update `app/e2e-tests/tests/namespacesPage.ts` and `clusterRename.spec.ts` to the new hook.
- If #6653 merges first, this PR should use `data-testid` instead of the aria-label.

Either way I will do the adjustment; just flagging it so it is not a surprise in review. No action needed unless you want them landed in a particular order.

illume (approved):
🎉 thanks!

PR #6644, opened 2026-07-26
frontend: Activity: Add accessible aria-labels to icon buttons

Suyog241005:
## Summary

This PR improves accessibility (a11y) for keyboard-only and screen-reader users by adding explicit `aria-label` attributes to icon-only action buttons in `Activity.tsx` and `ShortcutsSettings.tsx`.

## Related Issue

N/A

## Changes

- Added `aria-label={t('Minimize')}` and `aria-label={t('Close')}` to header `<IconButton>` components in `Activity.tsx`.
- Added `aria-label={t('Reset to default')}` to shortcut key reset `<IconButton>` in `ShortcutsSettings.tsx`.

## Steps to Test

1. Open Headlamp and trigger an Activity panel (e.g., Create/Apply or Log Viewer).
2. Inspect the minimize and close icon buttons in the top-right header using Chrome DevTools Accessibility inspector or a screen reader.
3. Verify that the accessible name announces as "Minimize" and "Close".
4. Navigate to Settings > Shortcuts and modify a shortcut key to display the reset button.
5. Verify that the reset button announces as "Reset to default".

*N/A - Accessibility attribute additions.*

## Notes for the Reviewer

- Clean 6-line diff touching frontend accessibility compliance.
- Passes `npm run frontend:lint:fix` and `npm run frontend:test`.

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Added accessible labels to shortcut reset, activity minimize, and activity close buttons.
  * Improved screen reader clarity for window controls in activity panels.

### DIFF
--- a/frontend/src/components/App/Settings/ShortcutsSettings.tsx
+++ b/frontend/src/components/App/Settings/ShortcutsSettings.tsx
@@ -203,7 +203,7 @@ function ShortcutEditor({
       </Button>
       {isModified && (
         <Tooltip title={t('Reset to default')}>
-          <IconButton size="small" onClick={onReset}>
+          <IconButton size="small" onClick={onReset} aria-label={t('Reset to default')}>
             <Icon icon="mdi:restore" width={16} />
           </IconButton>
         </Tooltip>
--- a/frontend/src/components/activity/Activity.tsx
+++ b/frontend/src/components/activity/Activity.tsx
@@ -669,15 +669,19 @@ export function SingleActivityRenderer({
                     </ClickAwayListener>
                   </Popper>
                   <IconButton
-                    onClick={() => {
-                      Activity.update(id, { minimized: true });
-                    }}
+                    onClick={() => Activity.update(id, { minimized: true })}
                     size="small"
                     title={t('Minimize')}
+                    aria-label={t('Minimize')}
                   >
                     <Icon icon="mdi:minimize" />
                   </IconButton>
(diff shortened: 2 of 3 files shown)


Suyog241005:
Hi @illume , can you review this PR when you get the time.

Suyog241005:
hi @illume , copilot has generated no new comments.

Suyog241005:
hi @illume , copilot has generated no new comments.

unlikelyzero:
The `aria-label` additions here are correct — these are icon-only buttons (their only child is an `Icon` glyph, which iconify renders `aria-hidden` by default), so there's no visible text label for `aria-label` to conflict with, and `title` alone isn't a reliable accessible-name source. This matches the existing pattern elsewhere in the codebase (e.g. `ActionButton.tsx`).

One button in the same file was missed: the "Window" snap-menu `IconButton` at `Activity.tsx:466` (`title={t('Window')}`, no `aria-label`) has the identical gap this PR is fixing everywhere else. Worth adding `aria-label={t('Window')}` there too while this is fresh.

Smaller, non-blocking note: the `ShortcutsSettings.tsx` change (adding `aria-label` to the button already wrapped in MUI `Tooltip`) is a no-op — `Tooltip` already sets `aria-label` from its `title` prop on the child when `describeChild` is false, so that button already had the correct accessible name. Harmless to keep for clarity, just flagging that it isn't fixing a bug like the other two are.

Issue #6634, opened 2026-07-25
e2e-tests/README.md setup does not match what the tests require


unlikelyzero:
## Describe the bug

`e2e-tests/README.md` describes a minikube-based setup, but the specs in that directory cannot pass against it. They hardcode two cluster contexts named `test` and `test2`, which is what CI creates using kind. Following the README as written leads to failures that look like product bugs rather than a setup mismatch.

The README instructs:

```
minikube addons enable headlamp
minikube service headlamp -n headlamp
export HEADLAMP_TEST_URL=   # from the command above
export HEADLAMP_TOKEN=$(kubectl create token headlamp --duration 24h -n headlamp)
```

Three things do not line up with the tests:

1. Cluster names. The specs reference `test` and `test2` throughout, for example `navigateTopage('/c/test/pods')` in `podsPage.spec.ts` and the whole of `multiCluster.spec.ts`, which asserts a two-cluster home page listing `test` and `test2`. A single minikube cluster named `minikube` cannot satisfy this.
2. Token variable name. The README exports `HEADLAMP_TOKEN`, but the specs read `HEADLAMP_TEST_TOKEN` (and `HEADLAMP_TEST2_TOKEN` for the second cluster). Nothing reads `HEADLAMP_TOKEN`.
3. Deployment namespace. The README uses the minikube addon, which installs into the `headlamp` namespace. The tests expect the CI layout from `e2e-tests/kubernetes-headlamp-ci.yaml`, which deploys into `kube-system` with a `headlamp-admin` service account. `headlamp.spec.ts` asserts on that service account by name.

What CI actually does, in `.github/workflows/build-container.yml`, is create two kind clusters, rename their contexts, and deploy in-cluster:

```
kind create cluster --name test    # via helm/kind-action
kubectl config rename-context kind-test test
kind create cluster --name test2
kubectl config rename-context kind-test2 test2
kubectl create serviceaccount headlamp-admin --namespace kube-system
kubectl create clusterrolebinding headlamp-admin --serviceaccount=kube-system:headlamp-admin --clusterrole=cluster-admin
envsubst < e2e-tests/kubernetes-headlamp-ci.yaml | kubectl --context=test apply -f -
```

[...]

PR #6654, opened 2026-07-26
app: e2e-tests: Make the desktop suite runnable as documented

### LINKED ISSUE #6633: npm run app:test:e2e fails with Missing script: "test"
## Describe the bug

`npm run app:test:e2e` fails immediately with `Missing script: "test"`. The root `package.json` delegates to a script that does not exist in `app/e2e-tests/package.json`.

Root `package.json`:

```json
"app:test:e2e": "cd app/e2e-tests && npm run test",
"app:test": "npm run app:test:unit && npm run app:test:e2e",
```

but `app/e2e-tests/package.json` only defines:

```json
"scripts": {
  "test-app": "PLAYWRIGHT_TEST_MODE=app playwright test",
  "test-web": "PLAYWRIGHT_TEST_MODE=web playwright test"
}
```

There is no `test` script, so the delegation cannot resolve. This also breaks `npm run app:test`, since that runs `app:test:e2e` as its second step.

The command is referenced as a validation step in `AGENTS.md` (in the app validation list and in the frontend guidance about running e2e tests for significant UI changes), so anyone following that guidance hits the error.

Expected: `npm run app:test:e2e` runs the desktop e2e tests, or the documented entry point points at a script that exists.

## To Reproduce

Steps to reproduce the bug:

1. From a clean checkout of main:
   ```
   npm run app:test:e2e
   ```
2. Observe:
   ```
   > headlamp-root@0.1.0 app:test:e2e
   > cd app/e2e-tests && npm run test

   npm error Missing script: "test"
   npm error
   npm error To see a list of scripts, run:
   npm error   npm run
   ```
3. Confirm the available scripts do not include `test`:
   ```
   node -p "Object.keys(require('./app/e2e-tests/package.json').scripts)"
   [ 'test-app', 'test-web' ]
   ```

This fails before any cluster or browser setup, so it does not depend on having minikube or Playwright installed.

 - Other: macOS 26.5.2 arm64, node v24.14.1, npm 11.11.0

Yes, though I would like input on the intended behaviour first.

The two modes are not equivalent, so it is not obvious which one `app:test:e2e` should mean:

- `test-app` runs against the built Electron app and needs no token.
- `test-web` needs the backend and frontend running separately, per `app/e2e-tests/README.md`.

Both also require a running minikube instance named `minikube`. So the options are roughly:

1. Add `"test": "npm run test-app"` to `app/e2e-tests/package.json` and treat app mode as the default.
2. Point the root script directly at `test-app`, leaving the sub-package scripts explicit.
3. Split into `app:test:e2e:app` and `app:test:e2e:web` and drop the ambiguous single entry point.

Happy to send a PR once someone confirms which is preferred.

## Additional Context

Related observation while looking into this: the `app/e2e-tests` suite does not appear to run in any workflow. Searching `.github/workflows/` for `app/e2e-tests`, `app:test:e2e`, `test-app`, or `test-web` returns nothing, whereas the `e2e-tests/` web suite runs from `build-container.yml`. That may be intentional, since the desktop suite needs a real Electron build, but it does mean this broken script would not have been caught by CI. I am happy to file that separately if it is worth tracking on its own.

### LINKED ISSUE #6640: app/e2e-tests README is missing the compile-electron step

unlikelyzero:
## Summary

Two problems stopped the desktop e2e suite from being runnable by following the documentation.

`npm run app:test:e2e` failed immediately with `Missing script: "test"`. The root script delegated to a `test` script in `app/e2e-tests` that does not exist; that package only defines `test-app` and `test-web`.

Separately, the README omitted a required setup step. `app/package.json` declares `"main": "build/main.js"` and the specs launch the app with `electron .`, which resolves that field, but nothing in the README or the specs created it.

## Related Issue

Fixes #6633
Fixes #6640

## Changes

- **Root `app:test:e2e` now points at `test-app`**, the mode these tests are written for. Adding a delegating `test` script to `app/e2e-tests` would also have fixed the error, but it puts another `npm run` in the chain and arguments are not forwarded across those, so flags would be silently dropped.
- **Added a `pretest-app` hook** that runs `compile-electron -- --dev`, so the Electron entry point is built automatically and the suite cannot be run against a missing one. That seemed better than documenting one more manual step people can skip.
- **Rewrote `app/e2e-tests/README.md`** around what the suite actually requires.

The README changes are based on getting the suite running rather than on reading the source. The previously undocumented requirements:

- A minikube cluster named **`minikube`**. The name matters: `clusterRename.spec.ts` expects it, and `clusterAutoConnect.spec.ts` shells out to `minikube` directly to create and delete its own throwaway profile, so both `minikube` and `kubectl` must be on `PATH`.
- The **frontend and backend built**, because app mode runs the app from source. With `ELECTRON_DEV=true`, `app/electron/main.ts` resolves `../frontend/build/index.html` (main.ts:91) and `../backend/headlamp-server` (main.ts:772).
- **Chromium installed**, even though these are Electron tests. Every spec requests Playwright's `page` fixture before switching to the Electron window, and requesting it launches a browser. This one is easy to miss.

It also documents that flags cannot be passed through `npm run app:test:e2e` from the repository root, because that delegates to a second `npm run`; run the script from `app/e2e-tests` when you need flags.

## Steps to Test

The script fix, which needs no cluster:

```
# Before: Missing script: "test"
# After: resolves and starts Playwright
node -p "require('./package.json').scripts['app:test:e2e']"
# -> cd app/e2e-tests && npm run test-app
```

The pre-hook, which is the point of the second fix:

```
rm -f app/build/main.js
cd app/e2e-tests && npm run pretest-app
ls -la ../build/main.js     # rebuilt
```

That the suite is discoverable, without running it:

```
cd app/e2e-tests
PLAYWRIGHT_TEST_MODE=app npx playwright test --list
# -> Total: 4 tests in 3 files
```

Then follow the README from a clean checkout and confirm you reach a running suite.

Not applicable, docs and npm scripts only.

## Notes for the Reviewer

- **This does not make the suite pass.** It currently fails for unrelated reasons, tracked in #6648 and fixed in #6641. This PR only makes it possible to invoke it by following the docs. I have kept the two separate because they are different problems with different reviewers in mind, and #6641 is test-code while this is scripts and documentation.
- **On `pretest-app` versus documenting the step.** npm runs `pre<script>` automatically for `npm run test-app`, so this costs nothing at the call site. The alternative was a README line saying "remember to run compile-electron first", which is the kind of step people skip and which produced a confusing failure. Happy to switch to documentation only if you would rather the scripts stay minimal.
- **Web mode is unchanged.** Only `test-app` gets the pre-hook, because web mode does not launch Electron and so does not need `build/main.js`.
- I did not touch the top-level `e2e-tests` README, which has a similar but separate mismatch tracked in #6634.

### DIFF
--- a/app/e2e-tests/README.md
+++ b/app/e2e-tests/README.md
@@ -1,69 +1,105 @@
-# e2e test for local playwright app mode
+# e2e tests for the desktop (Electron) app
 
-Currently we have the original e2e tests for the web mode in the `e2e-tests` directory. We are adding new tests for the app mode in the `app-e2e-tests` directory for local testing. Unlike the other tests, these tests do not require a token to run so the setup is as followed:
+These are the Playwright tests for the desktop app. The web-mode tests for
+in-cluster Headlamp live in the top-level `e2e-tests` directory.
 
-## Running app mode tests
+There are two modes:
 
-## Setup
+- **app mode** (`npm run test-app`) drives the Electron app directly. This is
+  the mode most of these tests are written for; three of the four skip unless
+  `PLAYWRIGHT_TEST_MODE=app`.
+- **web mode** (`npm run test-web`) drives a browser against a separately
+  running backend and frontend.
 
-- Before running the tests, be sure to have an instance of Minikube running with the name `minikube`
+## Prerequisites
 
-### Running the tests
+Both modes need a running minikube cluster named `minikube`:
 
-To run the tests for the app mode, follow the steps below:
+```bash
+minikube start
+kubectl config current-context   # should print: minikube
+```
 
-- cd into the e2e-tests directory within the headlamp repository
-  `cd headlamp/app/e2e-tests`
+The name matters. `clusterRename.spec.ts` expects a cluster called `minikube`,
+and `clusterAutoConnect.spec.ts` shells out to `minikube` directly to create and
+delete its own throwaway profile, so `minikube` and `kubectl` must both be on
+`PATH`.
 
-- npm install the needed packages
-  `npm install`
+App mode runs the app from source rather than from a packaged build, so the
+things the app loads in dev mode have to exist first:
 
-- run the following command
-  `npm run test-app`
-  (optional: include `-- --headed` to run the tests in headed mode)
-  (optional: include `-- --ui` to run the tests in ui mode)
+```bash
+# from the repository root
+npm run frontend:install
+npm run frontend:build   # -> frontend/build/index.html
+npm run backend:build    # -> backend/headlamp-server
+npm run app:install
+```
 
-## Running web mode tests
+`app/build/main.js`, the Electron entry point, is built automatically by the
+`pretest-app` script, so there is no separate step for it.
 
-Running the tests for the web mode requires the backend and frontend to be running. Follow the steps below to run the tests for the web mode:
+Then install the test dependencies:
 
-Note: You may encouter issues switching from the app mode tests to the web mode tests. If you do, search for any running headlamp server processes and end them before running the web mode tests or app mode tests.
+```bash
+cd app/e2e-tests
+npm install
+npx playwright install chromium
+```
 
-## Setup
+Chromium is required even though these are Electron tests. The specs request
+Playwright's `page` fixture before switching to the Electron window, and
+requesting it launches a browser.
 
-- Before running the tests, be sure to have an instance of Minikube running with the name `minikube`
+## Running the tests
 
-### Backend
+### App mode
 
-To run the tests for the web mode, you will need to have the backend running. Follow the steps below to run the backend:
+```bash
+cd app/e2e-tests
+npm run test-app
+```
 
-- cd into the headlamp directory in a singular terminal
-  `cd headlamp`
+Or from the repository root:
 
-- run the following command
-  `npm run backend:build` followed by `npm run backend:start`
+```bash
+npm run app:test:e2e
+```
 
-### Frontend
+Optional flags can be passed from either directory:
 
-To run the tests for the web mode, you will need to have the frontend running. Follow the steps below to run the frontend:
+```bash
+# from app/e2e-tests
+npm run test-app -- --headed   # watch it run
+npm run test-app -- --ui       # Playwright UI mode
 
-- cd into the headlamp directory in a separate terminal
-  `cd headlamp`
+# from the repository root
+npm run app:test:e2e -- --headed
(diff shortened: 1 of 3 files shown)


illume (approved):
Thanks! 🎉

PR #6656, opened 2026-07-26
ci: push-release-assets: Add cosign keyless signing for checksums

Athang69:
## Summary
This PR adds cosign keyless signing of the release checksums.txt file to the push-release-assets workflow. It attaches a checksums.txt.sigstore.json bundle to each release, giving OpenSSF Scorecard a recognized signature file to detect, which should raise the Signed-Releases check score from the current 0/10 on future releases where this workflow runs.

## Related Issue
Relates to the OpenSSF Scorecard report for this repo showing a 0/10 Signed-Releases score. No existing tracked issue.

## Changes
- Added an id-token: write permission to the publish job, needed for cosign keyless signing via GitHub OIDC
- Added a step to install cosign using sigstore/cosign-installer pinned by SHA
- Added a step that signs checksums.txt with cosign in keyless mode, producing a checksums.txt.sigstore.json bundle, and uploads it to the release as an additional asset

## Steps to Test
1. Checked out this branch in a personal fork and confirmed cosign keyless signing works end to end in an isolated test workflow, since the real push-release-assets workflow needs release secrets and a real release to run against
2. The test workflow installed cosign, signed a sample file with cosign sign-blob --bundle, and ran cosign verify-blob --new-bundle-format against it
3. The run succeeded, ending in Verified OK, confirming the OIDC permission and cosign invocation are correct before applying the same pattern to checksums.txt in this workflow
4. Once merged, the real verification step is to confirm checksums.txt.sigstore.json appears as a release asset the next time push-release-assets is dispatched for an actual release, then re-run the OpenSSF Scorecard analysis against the repo to confirm the Signed-Releases score improves

## Screenshots 
Verified cosign keyless signing in an isolated test run on my fork: https://github.com/Athang69/headlamp/actions/runs/30211522217/job/89818442024

The run shows "Wrote bundle to file test.txt.sigstore.json" followed by "Verified OK" for the cosign verify-blob step, confirming the OIDC signing and verification flow works correctly.

image

## Notes for the Reviewer
This only affects future releases where push-release-assets is dispatched, it cannot retroactively sign already published releases, so the Scorecard score will not update until the next real release runs through this workflow. Also worth checking separately why this workflow was not dispatched for the last couple of releases, since the existing GPG signing step for checksums.txt.asc has the same gap today.

### DIFF
--- a/.github/workflows/push-release-assets.yml
+++ b/.github/workflows/push-release-assets.yml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write # need to write to releases and download artifacts
+      id-token: write # needed for cosign keyless signing via GitHub OIDC
     steps:
       - name: Checkout code
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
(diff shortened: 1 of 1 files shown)


Athang69:
@illume This PR adds cosign keyless signing to push-release-assets so checksums.txt gets a signed bundle attached to future releases, targeting the Signed-Releases check on our OpenSSF Scorecard which is currently 0/10. I tested the cosign signing and verification flow in an isolated run on my fork first (linked in the PR description) since the real workflow needs release secrets I don't have access to. One thing worth flagging separately, the existing GPG signing step for checksums.txt.asc doesn't seem to have run on the last couple of releases either, so that might be worth a look too.
PTAL whenever convenient. Thanks!

illume:
Thanks @Athang69 

Yeah, I think we can remove the gpg stuff. It hasn't worked since we moved to the Kubernetes-sigs org over a year ago. I don't think anyone ever checked it either. With cosign stuff checking is automated.

illume (approved):
🎉 thanks!

illume:
(The frontend test failure is unrelated and fixed in main)

PR #6657, opened 2026-07-26
backend: kubeconfig: Resolve relative PKI paths against file dir

### LINKED ISSUE #2413: Support pki file?
CleanShot 2024-10-11 at 17 11 46@2x

This kind of kubeconfig is not recognized by headlamp.

Pritz395:
## Summary

Kubeconfigs that use relative `certificate-authority`, `client-certificate`, and `client-key` paths work with kubectl but failed in Headlamp Desktop. Headlamp loaded kubeconfig bytes via `clientcmd.Load` without recording `LocationOfOrigin`, so relative PKI paths were never resolved against the kubeconfig file directory.

This matches client-go's `LoadFromFile` + `ResolveLocalPaths` behavior: when a kubeconfig comes from disk, relative paths are absolutized before `SetupProxy`. Base64 / in-memory configs have no filesystem origin and are left unchanged.

## Related Issue

Fixes #2413

## Changes

- Thread the source kubeconfig path through the file load path
- Resolve relative PKI paths with `clientcmd.ResolveLocalPaths` before proxy setup
- Keep the exported `ProcessContext` API unchanged (no origin → no resolution)
- Add regression tests for relative, absolute, parent-relative, multi-file, and base64 cases

## Steps to Test

1. Create a kubeconfig with relative PKI paths, e.g.:
   ```yaml
   apiVersion: v1
   clusters:
   - cluster:
       certificate-authority: pki/ca.pem
       server: https://<api>:6443
     name: kubernetes80
   contexts:
   - context:
       cluster: kubernetes80
       user: kubernetes-admin1
     name: kubernetes-80
   current-context: kubernetes-80
   kind: Config
   users:
   - name: kubernetes-admin1
     user:
       client-certificate: pki/admin.pem
       client-key: pki/admin.key
   ```
   with the PEM files under `pki/` next to the kubeconfig.
2. Confirm `kubectl` can use that config.
3. Point Headlamp Desktop at the same kubeconfig and select the context.
4. Confirm the cluster connects (no Bad Gateway from unresolved relative cert paths).
5. Confirm absolute-path kubeconfigs and base64/dynamic cluster configs still behave as before.

## Notes for the Reviewer

- Regression test fails on the old code with paths left as `pki/...` and passes after the fix.
- Validated locally with `npm run backend:lint` and `npm run backend:test`.
- Opaque Bad Gateway when cert files are missing after resolution is a separate concern (`SetupProxy` still swallows transport setup errors); this PR only fixes the relative-path resolution gap reported in #2413.

### DIFF
--- a/backend/pkg/kubeconfig/export_test.go
+++ b/backend/pkg/kubeconfig/export_test.go
@@ -10,6 +10,9 @@ var BuildUserAgent = buildUserAgent
 
 var DeriveInClusterName = deriveInClusterName
 
+// ResolveKubeconfigPaths is exported for testing.
+var ResolveKubeconfigPaths = resolveKubeconfigPaths
+
 // UserAgentRoundTripper is exported for testing.
 type UserAgentRoundTripper struct {
 	Base      roundTripperInterface
--- a/backend/pkg/kubeconfig/kubeconfig.go
+++ b/backend/pkg/kubeconfig/kubeconfig.go
@@ -9,6 +9,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


Pritz395:
@illume Could you please take a look at this PR

Pritz395:
@illume Is the PR good to go?

Pritz395:
@joaquimrocha Could you please take a look at this PR? Thank you!

illume (approved):
🎉 thanks!

Issue #6631, opened 2026-07-25
e2e-tests: themedXterm.spec.ts skips all 6 of its tests unconditionally


unlikelyzero:
## Describe the bug

All 6 tests in `e2e-tests/tests/themedXterm.spec.ts` skip unconditionally, on every run, including in CI. The xterm theme-awareness assertions and the axe accessibility checks in that file have never actually executed since the file was added in #5671.

Because Playwright reports skipped tests as non-failures, CI stays green and the skips are silent.

The cause is the permission guard in `openFirstPodDetails` (themedXterm.spec.ts:69-71):

```js
const content = await page.content();
if (!content.includes('Pods') || !content.includes('href="/c/test/pods')) {
  test.skip(true, 'No pods permission on this cluster.');
}
```

This infers "no pods permission" from the presence of a Pods link in the rendered sidebar. But Headlamp's sidebar has no top-level Pods entry: Pods lives under the Workloads section, which is collapsed by default, so `href="/c/test/pods` is never in the DOM at that point. The guard therefore fires even for a cluster-admin user on a cluster full of pods.

Expected: with sufficient permissions and pods present, the logs and exec cases should run. Actual: they always skip.

## To Reproduce

Steps to reproduce the bug:

1. Set up the e2e environment the way `.github/workflows/build-container.yml` does: a kind cluster whose context is named `test`, a `headlamp-admin` service account in `kube-system` bound to `cluster-admin`, and Headlamp deployed in-cluster from `e2e-tests/kubernetes-headlamp-ci.yaml`.
2. Confirm the cluster genuinely has pods and the token is cluster-admin:
   ```
   kubectl --context=test get pods -A | wc -l
   kubectl --context=test auth can-i list pods --as=system:serviceaccount:kube-system:headlamp-admin
   kubectl --context=test auth can-i create pods/exec --as=system:serviceaccount:kube-system:headlamp-admin
   ```
3. Run just this spec:
   ```
   cd e2e-tests
   export HEADLAMP_TEST_URL=<your headlamp url>
   export HEADLAMP_TEST_TOKEN=$(kubectl --context=test create token headlamp-admin --duration 24h -n kube-system)
   npx playwright test tests/themedXterm.spec.ts --workers=1 --reporter=list
   ```
4. All 6 cases report as skipped rather than passed:
   ```
   -  1 [chromium] > logs viewer in light theme: background matches theme + axe clean
[...]

PR #6658, opened 2026-07-26
e2e-tests: Document the setup the tests actually require

### LINKED ISSUE #6634: e2e-tests/README.md setup does not match what the tests require

unlikelyzero:
## Summary

`e2e-tests/README.md` described a single minikube cluster with the headlamp addon, but the specs in that directory cannot pass against it. They hardcode two kubectl contexts named `test` and `test2`, which is what CI creates with kind, and they assert on an in-cluster deployment rather than on the addon.

Following the README as written produces failures that look like product bugs rather than a setup mismatch.

## Related Issue

Fixes #6634

## Changes

Rewrote the setup section around what CI does in `.github/workflows/build-container.yml`, which was the only place the working configuration was recorded.

Three concrete mismatches are fixed:

- **Cluster names.** `podsPage.spec.ts` navigates to `/c/test/pods`, and `multiCluster.spec.ts` asserts a home page listing both `test` and `test2`. A single cluster named `minikube` cannot satisfy those.
- **Token variable.** The README exported `HEADLAMP_TOKEN`. Nothing reads that. The specs read `HEADLAMP_TEST_TOKEN` and `HEADLAMP_TEST2_TOKEN`.
- **Deployment shape.** The addon installs into the `headlamp` namespace; the tests expect the layout in `kubernetes-headlamp-ci.yaml`, which deploys into `kube-system` with a `headlamp-admin` service account. `headlamp.spec.ts` asserts on that service account by name.

It also documents three things that were not written down anywhere, each of which cost me time to rediscover:

- `HEADLAMP_TEST_URL` defaults to `http://localhost:3000` (`playwright.config.ts`), so leaving it unset surfaces as connection errors rather than as an obvious missing setting.
- `dynamicCluster.spec.ts` reads certificates from disk via `kubectl config view`, which redacts embedded certificate data. An embedded-certificate kubeconfig fails six tests in that file with `ENOENT`, and the paths must be absolute because the test resolves them relative to its own working directory.
- `incluster-api.spec.ts` and `clusterInventory.spec.ts` skip unless their variables are set. Neither is set in CI, so those suites do not run there either. That is worth knowing before assuming they provide coverage.

Every environment variable the specs read is now documented, including the optional `HEADLAMP_TEST_BACKEND_TOKEN`.

## Steps to Test

Follow the README from a clean checkout and confirm you reach a passing run. I did this and got `35 passed, 11 skipped` in about 1.2 minutes. The 11 skips are the two gated suites plus `themedXterm.spec.ts`, which skips for a separate reason tracked in #6631.

The two mismatches can also be confirmed without running anything:

```bash
# Nothing reads HEADLAMP_TOKEN
grep -ro "process\.env\.HEADLAMP[A-Z_0-9]*" e2e-tests/tests/ | sed 's/.*env\.//' | sort -u

# The specs hardcode test / test2
grep -ro "'/c/test2\?[^']*'" e2e-tests/tests/ | sort -u
```

Not applicable, documentation only.

## Notes for the Reviewer

- **I documented the kind setup rather than making minikube work**, because that is what the specs require today and what CI exercises. Whether minikube should also be supported is a separate question: it would mean the specs no longer hardcoding `test` and `test2`, which is a code change rather than a docs one. I have left that open in #6634 rather than deciding it here.
- **On the Service URL.** CI uses `http://<node InternalIP>:<nodePort>`, which is reachable from the host on Linux runners but not on every platform. Rather than document a platform-specific workaround, the README says any means of exposing the Service is fine as long as `HEADLAMP_TEST_URL` points at it. Happy to add platform-specific guidance if you would rather it be concrete.
- **The Azure VM section is unchanged**, with a note that it predates this setup. It refers to a single minikube cluster and to `kubernetes-headlamp-ci.yml`, which is now `.yaml`. I could not verify those steps, so I did not want to rewrite them speculatively. Removing or updating that section seems worth a separate decision.
- The certificate workaround is described as a workaround, not a recommended step, since it modifies the reader's kubeconfig. It is only needed if `dynamicCluster.spec.ts` fails.

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -101,33 +101,60 @@ absolute, because the test resolves them relative to its own working directory:
 
 ```bash
 mkdir -p ~/headlamp-e2e-certs
-kubectl config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority-data}' | base64 --decode > ~/headlamp-e2e-certs/ca.crt
+kubectl config view --raw --minify --context=test -o jsonpath='{.clusters[0].cluster.certificate-authority-data}' | base64 --decode > ~/headlamp-e2e-certs/ca.crt
 kubectl config set-cluster kind-test --certificate-authority="$HOME/headlamp-e2e-certs/ca.crt" --embed-certs=false
 kubectl config unset clusters.kind-test.certificate-authority-data
 ```
 
 Do the same for the `client-certificate-data` and `client-key-data` fields on
-the user entry.
+the user entry. To update the second cluster, use `--context=test2` and the
+`kind-test2` cluster entry.
 
-### Optional suites
+### Additional suites
 
-Two spec files skip unless their variable is set, so a normal run does not
-exercise them:
+`tests/incluster-api.spec.ts` tests a separate Headlamp deployment running in
+in-cluster mode. CI runs it in a dedicated step after deploying
+`kubernetes-headlamp-incluster-ci.yaml` to the `test2` cluster. To reproduce
+that setup locally:
 
 ```bash
-# tests/incluster-api.spec.ts
-export HEADLAMP_SA_TOKEN=$(kubectl create token headlamp --duration=1h -n kube-system)
+kubectl config use-context test2
+kubectl create serviceaccount headlamp --namespace kube-system
+kubectl create clusterrolebinding headlamp \
+  --serviceaccount=kube-system:headlamp --clusterrole=cluster-admin
+kubectl apply -f e2e-tests/kubernetes-headlamp-incluster-ci.yaml
+kubectl wait deployment -n kube-system headlamp \
+  --for condition=Available=True --timeout=120s
 
-# tests/clusterInventory.spec.ts
-export HEADLAMP_CLUSTER_INVENTORY_E2E=true
-
-# Optional, only read by clusterInventory.spec.ts. Defaults to "headlamp".
-export HEADLAMP_TEST_BACKEND_TOKEN=...
+kubectl port-forward -n kube-system service/headlamp 8080:80
 ```
 
-Neither is set in CI, so both suites are skipped there as well.
+Leave the port-forward running, then use a second terminal to set the URL and
+service account token before running the spec:
 
-IMPORTANT: Make sure that the following npx commands are ran in the same terminal session as the environment variables were set.
+```bash
+export HEADLAMP_TEST_URL=http://localhost:8080
+export HEADLAMP_SA_TOKEN=$(kubectl create token headlamp --duration=1h -n kube-system)
+
+cd e2e-tests
+npx playwright test tests/incluster-api.spec.ts
+```
+
+`tests/clusterInventory.spec.ts` requires Headlamp to be configured with a
+working cluster inventory source. It is skipped unless explicitly enabled and
+is not run in CI:
+
+```bash
+export HEADLAMP_CLUSTER_INVENTORY_E2E=true
+
+# Optional. This defaults to "headlamp".
+export HEADLAMP_TEST_BACKEND_TOKEN=...
+
+cd e2e-tests
+npx playwright test tests/clusterInventory.spec.ts
+```
+
+IMPORTANT: Make sure that the following npx commands are run in the same terminal session as the environment variables were set.
 
 ## Run all tests
 
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -1,35 +1,131 @@
-# end to end tests with playwright and minikube
+# end to end tests with playwright
 
-```
+These tests run against Headlamp deployed **in-cluster**, across **two**
+clusters whose kubectl contexts are named `test` and `test2`.
+
+Those names are not arbitrary. The specs hardcode them: `podsPage.spec.ts`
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

PR #6669, opened 2026-07-27
backend: Scope ClusterProfile discovery by namespace

kahirokunn:
## Summary

Scope ClusterProfile discovery to the namespace used by each Headlamp consumer.
This makes the recommended one-inventory-per-consumer model the default while
still allowing shared inventories through an explicit namespace list or `*`.

## Related Issue

https://github.com/kubernetes/enhancements/pull/6237

## Changes

- Default ClusterProfile discovery to the Headlamp pod namespace in-cluster and
  each kubeconfig context namespace out-of-cluster.
- Add server and Helm configuration for named namespaces or all namespaces.
- Validate namespace names and require `*` to be used on its own.
- Update the Cluster Inventory API dependency and provider examples to v0.1.3.

## Steps to Test

Prerequisites: Linux with Docker running, plus Node/npm, `curl`, `envsubst`, `jq`, `kind`, `kubectl`, and `make`.

From the repository root, run in order:

```bash
cd e2e-tests
npm ci
npx playwright install --with-deps
cd ..

./e2e-tests/scripts/setup-multicluster.sh

set -a
source e2e-tests/.env
set +a
cd e2e-tests
npx playwright test tests/multiCluster.spec.ts tests/clusterInventory.spec.ts
```

The setup creates `test` and `test2` when both are absent, builds and loads the Headlamp images, deploys the shared Headlamp and Inventory fixture, and writes the Playwright URL and temporary credentials to `e2e-tests/.env`. A later run reuses the same environment. The clusters remain available until explicitly deleted with `kind delete cluster --name test` and `kind delete cluster --name test2`.

this change has no UI changes.

CleanShot 2026-08-02 at 12 39 16

## Notes for the Reviewer

- Existing deployments that rely on cluster-wide discovery must set `*` explicitly.
- An explicit namespace list starts one informer per namespace and waits for all caches to sync before pruning contexts.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -502,24 +502,36 @@ func startClusterInventory(ctx context.Context, config *HeadlampConfig) error {
 		return nil
 	}
 
-	var hubConfig *rest.Config
+	var (
+		hubConfig    *rest.Config
+		hubNamespace string
+	)
 
 	if config.UseInCluster {
-		inClusterConfig, err := rest.InClusterConfig()
+		var err error
+
+		hubConfig, err = rest.InClusterConfig()
 		if err != nil {
 			return fmt.Errorf("get in-cluster config for cluster inventory: %w", err)
 		}
 
-		hubConfig = inClusterConfig
+		hubNamespace, _, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{}, &clientcmd.ConfigOverrides{},
+		).Namespace()
+		if err != nil {
+			return fmt.Errorf("get pod namespace for cluster inventory: %w", err)
+		}
 	}
 
 	runner, err := clusterinventory.NewRunner(clusterinventory.Options{
 		Store:                 config.KubeConfigStore,
 		ProviderFile:          config.ClusterInventoryProviderFile,
 		LabelSelector:         config.ClusterInventoryLabelSelector,
+		Namespaces:            config.ClusterInventoryNamespaces,
 		RootReconcileInterval: config.ClusterInventoryRootReconcileInterval,
 		NoCRDCacheTTL:         config.ClusterInventoryNoCRDCacheTTL,
 		HubConfig:             hubConfig,
+		HubNamespace:          hubNamespace,
 		DiscoverFromStore:     !config.UseInCluster,
 	})
 	if err != nil {
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -68,6 +68,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -145,6 +146,8 @@ const (
 	serverReadHeaderTimeout = 10 * time.Second
 	// serverIdleTimeout is the maximum time to wait for the next request on a keep-alive connection.
 	serverIdleTimeout = 120 * time.Second
+	// serviceAccountNamespacePath contains the namespace of an in-cluster pod.
+	serviceAccountNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 )
 
 // maxProxyResponseSize is the maximum size (in bytes) for proxied responses.
@@ -501,29 +504,57 @@ func addPluginListRoute(config *HeadlampConfig, r *mux.Router) {
 	}).Methods("GET")
 }
 
+func readServiceAccountNamespace() (string, error) {
+	data, err := os.ReadFile(serviceAccountNamespacePath)
+	if err != nil {
+		return "", fmt.Errorf("read service account namespace: %w", err)
+	}
+
+	return validateServiceAccountNamespace(data)
+}
+
+func validateServiceAccountNamespace(data []byte) (string, error) {
+	namespace := strings.TrimSpace(string(data))
+	if errs := validation.IsDNS1123Label(namespace); len(errs) > 0 {
+		return "", fmt.Errorf("invalid service account namespace %q: %s", namespace, strings.Join(errs, "; "))
+	}
+
+	return namespace, nil
+}
+
 func startClusterInventory(ctx context.Context, config *HeadlampConfig) error {
 	if !config.EnableClusterInventory {
 		return nil
 	}
 
-	var hubConfig *rest.Config
+	var (
+		hubConfig    *rest.Config
+		hubNamespace string
+	)
 
 	if config.UseInCluster {
-		inClusterConfig, err := rest.InClusterConfig()
+		var err error
+
+		hubConfig, err = rest.InClusterConfig()
 		if err != nil {
 			return fmt.Errorf("get in-cluster config for cluster inventory: %w", err)
 		}
 
-		hubConfig = inClusterConfig
+		hubNamespace, err = readServiceAccountNamespace()
+		if err != nil {
+			return fmt.Errorf("get pod namespace for cluster inventory: %w", err)
+		}
 	}
 
 	runner, err := clusterinventory.NewRunner(clusterinventory.Options{
 		Store:                 config.KubeConfigStore,
 		ProviderFile:          config.ClusterInventoryProviderFile,
 		LabelSelector:         config.ClusterInventoryLabelSelector,
+		Namespaces:            config.ClusterInventoryNamespaces,
 		RootReconcileInterval: config.ClusterInventoryRootReconcileInterval,
 		NoCRDCacheTTL:         config.ClusterInventoryNoCRDCacheTTL,
 		HubConfig:             hubConfig,
+		HubNamespace:          hubNamespace,
 		DiscoverFromStore:     !config.UseInCluster,
 	})
 	if err != nil {
--- a/backend/pkg/clusterinventory/clusterinventory.go
+++ b/backend/pkg/clusterinventory/clusterinventory.go
@@ -28,6 +28,7 @@ import (
 	"hash"
 	"net/http"
 	"net/url"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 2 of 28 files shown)


kahirokunn:
@illume Hi ✋ , could you take a look at this when you get a chance?

Looking at the surrounding ecosystem, I noticed there are quite a few cases where people want to use the Cluster Inventory API in scopes other than cluster scope. I've already verified this works locally.
Thank you 🙏

kahirokunn:
The conflict has been resolved.

unlikelyzero (commented):
Overall this looks good. I checked the namespace scoping pretty closely and it looks like the filtering is happening server-side through the per-namespace informer factories, rather than just filtering results in the client. The * validation also looks consistent between the Helm schema and the Go code.

I did notice two things that are outside the lines changed in this PR, so leaving them here instead of inline:

1.  `charts/headlamp/values.yaml` still defaults `clusterRoleBinding.clusterRoleName` to `cluster-admin`.

That means the new namespace setting limits what Headlamp asks for, but the default RBAC still gives it access to everything. That seems a little inconsistent with the tenant-isolation wording in the docs.

Probably worth a follow-up to either add namespaced Role/RoleBinding support or adjust the docs so they don't imply the namespace setting is an RBAC boundary.

2. `validateClusterInventory()` in backend/pkg/config/config.go validates `--cluster-inventory-label-selector` during config parsing, but `--cluster-inventory-namespaces` isn't validated until `clusterinventory.NewRunner` is created during startup.

It's not a functional issue since bad input still prevents the server from starting, but it would be more consistent to validate both flags in the same place. It would also make sense to cover this in TestClusterInventoryValidation.

I left the smaller findings inline.

backend/cmd/headlamp.go
```diff
@@ -502,24 +502,36 @@ func startClusterInventory(ctx context.Context, config *HeadlampConfig) error {
...
-		inClusterConfig, err := rest.InClusterConfig()
+		var err error
+
+		hubConfig, err = rest.InClusterConfig()
 		if err != nil {
 			return fmt.Errorf("get in-cluster config for cluster inventory: %w", err)
 		}
 
-		hubConfig = inClusterConfig
+		hubNamespace, _, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
```
unlikelyzero: One concern with the default namespace handling: `clientcmd` prefers `POD_NAMESPACE` over the service account namespace file, and that value also skips the validation used for `--cluster-inventory-namespaces`.

That means changing the Deployment env can silently change the discovery scope.

I'd prefer reading the service account namespace file directly, validating it the same way as the flag, and logging the resolved namespace at startup.

backend/pkg/clusterinventory/clusterinventory.go
```diff
@@ -422,13 +467,23 @@ func (r *Runner) activateRoot(state *rootState) (*rootState, bool) {
...
+
+	for _, watch := range state.watches {
+		watch.factory.Start(state.ctx.Done())
 
-	factory.Start(state.ctx.Done())
+		synced = append(synced, watch.informer.HasSynced)
+	}
 
-	if !cache.WaitForCacheSync(state.ctx.Done(), state.informer.HasSynced) {
+	if !cache.WaitForCacheSync(state.ctx.Done(), synced...) {
```
unlikelyzero: `WaitForCacheSync` currently waits on every namespace for a root. If one namespace can't sync, stale-context pruning never runs for any of the healthy namespaces either.

I'd track sync per namespace and run `completeRootSyncFromCache` only for the namespaces that successfully synced.

docs/installation/in-cluster/index.md
```diff
@@ -76,7 +76,25 @@ Cluster Inventory access provider. The `plugins` entries are Kubernetes `image`
 volumes that mount provider binaries into the Headlamp container, not Headlamp
 UI plugins. By default, Headlamp watches `ClusterProfile` resources that do not
 have the `headlamp.dev/ignore` label because the chart sets
-`config.clusterInventory.labelSelector` to `!headlamp.dev/ignore`.
+`config.clusterInventory.labelSelector` to `!headlamp.dev/ignore`. Discovery is
+scoped to the Headlamp pod's namespace. This follows the namespaced consumer
```
unlikelyzero, line 80: This changes the default discovery behavior from cluster-wide to namespace-scoped, so existing installs may stop seeing previously discovered clusters after upgrading.

I think this needs an explicit upgrade note mentioning that `namespaces: ["*"]` restores the old behavior.

Also, kubeconfig roots without a namespace fall back to `default`, which could make clusters disappear unexpectedly. Logging the resolved namespaces at startup would make this much easier to diagnose.

backend/pkg/clusterinventory/clusterinventory.go
```diff
@@ -270,42 +296,45 @@ func (r *Runner) collectStoreSeedRoots(
 			continue
 		}
 
-		desiredRoots[rootID] = seedConfig
+		desiredRoots[rootID] = rootConfig{
+			restConfig: seedConfig,
+			namespaces: r.namespacesForRoot(headlampContext.KubeContext.Namespace),
```
unlikelyzero, line 302: Namespace values from kubeconfig contexts (and `POD_NAMESPACE` in `headlamp.go`) skip the `IsDNS1123Label` validation used for the CLI flag.

Since `client-go` uses `path.Join`, malformed values can change the request path instead of being rejected. Low impact, but I'd run `namespacesForRoot` inputs through the same validation for consistency.

backend/pkg/clusterinventory/clusterinventory.go
```diff
@@ -880,6 +940,61 @@ func normalizeLabelSelector(selector string) (labels.Selector, error) {
...
+			namespaces = append(namespaces, namespace)
+		}
+	}
+
+	if allNamespaces {
+		if len(namespaces) > 0 {
+			return nil, errors.New(`invalid cluster-inventory-namespaces: "*" must be used on its own`)
+		}
+
+		return []string{metav1.NamespaceAll}, nil
```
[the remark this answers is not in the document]
kahirokunn: Good catch. normalizeNamespaces now rejects a second *, and I added cases for *,* in both the normalization and config-validation tests.

e2e-tests/scripts/setup-multicluster.sh
```diff
@@ -0,0 +1,345 @@
...
+log "Create a workload on ${SPOKE_CLUSTER}"
+apply_generated "${SPOKE_CONTEXT}" create namespace inventory-demo
+apply_generated "${SPOKE_CONTEXT}" -n inventory-demo create deployment inventory-demo \
+  --image=registry.k8s.io/pause:3.10.1
+kubectl --context="${SPOKE_CONTEXT}" -n inventory-demo \
+  rollout status deployment/inventory-demo --timeout=300s
+
+log "Install the Cluster Inventory API on ${HUB_CLUSTER}"
+kubectl --context="${HUB_CONTEXT}" apply -f \
+  "https://raw.githubusercontent.com/kubernetes-sigs/cluster-inventory-api/${CLUSTER_INVENTORY_API_VERSION}/config/crd/bases/multicluster.x-k8s.io_clusterprofiles.yaml"
```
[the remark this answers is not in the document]
kahirokunn: I kept the v0.1.3 tag here. This is an E2E setup script, and the Go dependency and docs already refer to the same released version. Using the Kubernetes SIG release tag keeps those references aligned; pinning the raw commit feels unnecessary for this fixture.

kahirokunn:
@unlikelyzero I reworded the installation docs so they only describe discovery scope and no longer imply that the namespace setting is an RBAC boundary. Changes to the chart’s default RBAC are outside this PR.

Namespace values are now also validated in `validateClusterInventory()`, with config parsing tests for invalid labels, empty entries, and repeated `*`.

backend/cmd/headlamp.go
kahirokunn: Updated. In-cluster startup now reads `/var/run/secrets/kubernetes.io/serviceaccount/namespace` directly and validates it as a DNS-1123 label. Each watch logs the resolved namespace when it starts, along with the root and server.

backend/pkg/clusterinventory/clusterinventory.go
kahirokunn: Changed this to sync and prune per namespace. A healthy namespace can now remove its own stale contexts even if another namespace never syncs, while contexts from the failed namespace are left alone. I added a regression test for that case.

docs/installation/in-cluster/index.md
kahirokunn: The upgrade impact is already called out in the PR description under “Notes for the Reviewer,” including that existing cluster-wide setups need to set `*` explicitly. I kept the installation page focused on the current behavior rather than adding a version-specific migration note there.

Resolved namespaces are now logged when each watch starts, so the `default` fallback is visible.

backend/pkg/clusterinventory/clusterinventory.go
kahirokunn: I fixed the in-cluster path by reading and validating the service-account namespace directly, but I’m leaving kubeconfig context namespaces unchanged in this PR. Rejecting values that currently follow the existing kubeconfig/client-go behavior would be a separate behavior change, so I’d prefer to handle that separately.

e2e-tests/scripts/setup-multicluster.sh
```diff
@@ -0,0 +1,345 @@
...
+  kubectl --context="${SPOKE_CONTEXT}" -n kube-system \
+    create token headlamp-admin --duration=24h
+)"
+export TEST_CA_DATA TEST_SERVER TEST2_CA_DATA TEST2_SERVER INVENTORY_SPOKE_TOKEN
+
+log "Deploy Headlamp and the inventory-test2 ClusterProfile on ${HUB_CLUSTER}"
+envsubst \
+  '${TEST_CA_DATA} ${TEST_SERVER} ${TEST2_CA_DATA} ${TEST2_SERVER} ${INVENTORY_SPOKE_TOKEN}' \
+  <"${REPO_ROOT}/e2e-tests/kubernetes-headlamp-ci.yaml" |
+  kubectl --context="${HUB_CONTEXT}" apply -f -
```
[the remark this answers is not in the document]
kahirokunn: Good catch. On reused clusters, the setup now restarts `deployment/headlamp` after applying the manifest and waits with `kubectl rollout status`, so Playwright does not continue against the previous pod. Fixed in c3ebe7917.

e2e-tests/scripts/setup-multicluster.sh
```diff
@@ -0,0 +1,345 @@
...
+    printf '%s\n' "${env_values[@]}" >>"${GITHUB_ENV}"
+    log "Playwright environment added to GITHUB_ENV"
+    return
+  fi
+
+  env_file="${REPO_ROOT}/e2e-tests/.env"
+  (
+    umask 077
+    printf '%s\n' "${env_values[@]}" >"${env_file}"
+  )
```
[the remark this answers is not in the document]
kahirokunn: Good catch. I updated the PR Steps to Test to wrap `source e2e-tests/.env` with `set -a`/`set +a`, matching the README, so the values are exported to `npx`.

kahirokunn:
Completed copilot review

kahirokunn:
just fixed conflict

illume (approved):
🎉 thanks!

PR #6675, opened 2026-07-27
frontend: streamingApi: Keep the newest resources when limiting

### LINKED ISSUE #6674: frontend: Streaming list limit discards the newest resources instead of the oldest
## Describe the bug

`streamResultsForCluster` in `frontend/src/lib/k8s/api/v1/streamingApi.ts` truncates the result set when a `limit` query parameter is set. It sorts the values so the most recent are first, then removes items from the front of that array, so it discards exactly the resources it just sorted to the top and keeps the oldest ones instead.

`frontend/src/lib/k8s/api/v1/streamingApi.ts:261-267`:

```js
values.sort((a, b) => {
  const aTime = new Date(a.lastTimestamp || a.metadata.creationTimestamp!).getTime();
  const bTime = new Date(b.lastTimestamp || b.metadata.creationTimestamp!).getTime();
  // Reverse sort, so we have the most recent resources at the beginning of the array.
  return 0 - (aTime - bTime);
});
values.splice(0, values.length - maxResources);
```

`splice(0, n)` removes `n` items starting at index 0. After a descending sort those are the newest. For the sort order the comment describes, the truncation should drop everything after the first `maxResources` entries.

## To Reproduce

With `limit: 2` and three resources created at 2023-04-25, 2023-04-26 and 2023-04-27:

1. After the sort, the array is `[newest, middle, oldest]`.
2. `values.splice(0, 3 - 2)` removes index 0, the newest.
3. The callback receives `[middle, oldest]`.

Expected `[newest, middle]`, actual `[middle, oldest]`.

I confirmed this against the real code rather than by reading it, by driving `streamResultsForCluster` through the existing mock WebSocket setup in `apiProxy.test.ts` and asserting on what the callback received.

## Impact

This affects the v1 list path, which is still reachable through `KubeObject.useApiList`/`useConnectApi` and through the `apiFactory`, `apiFactoryWithNamespace` and `streamResultsForCluster` functions exported from `lib/k8s/apiProxy` for plugin authors. `KubeObject.useList` now goes through the v2 `useKubeObjectList`, so the resource list views in the app are not affected.

The sort keys off `lastTimestamp`, which suggests events were the intended consumer, and showing the oldest events rather than the most recent is the least useful possible result for that case.

## Environment

- Installation type: development environment, frontend unit tests
- Headlamp Version: main at 50f0c7f14

Yes, I have a fix and a regression test ready to submit.

xevrion:
## Summary

`streamResultsForCluster` limits the result set by sorting the values so the most recent are first, then calling `values.splice(0, values.length - maxResources)`. `splice(0, n)` removes from the front of the array, which after that sort is the newest items, so the function discarded exactly the resources it had just sorted to the top and kept the oldest ones.

With `limit: 2` and three resources, the callback received `[middle, oldest]` where `[newest, middle]` was intended.

The fix drops everything after the first `maxResources` entries instead, which is what the existing sort and its comment already describe. The sort itself is unchanged.

## Related Issue

Fixes #6674

## Changes

- `streamingApi.ts`: `values.splice(0, values.length - maxResources)` becomes `values.splice(maxResources)`.
- `apiProxy.test.ts`: a regression test in the existing `streamResultsForCluster` block, plus three fixtures differing only in `creationTimestamp`, `name` and `uid`.

## Steps to Test

```
cd frontend

npx vitest run src/lib/k8s/api/v1/apiProxy.test.ts   # 68 passed
npx vitest run src/lib/k8s                           # 462 passed

npm run frontend:lint
npm run frontend:tsc
```

To confirm the new test is load-bearing, restore the old `splice` call and re-run `apiProxy.test.ts`. It fails with `expected [ 'middle', 'oldest' ] to deeply equal [ 'newest', 'middle' ]`, which is the bug itself. I verified this rather than assuming it.

## Notes for the Reviewer

The affected path is the v1 list code, still reachable through `KubeObject.useApiList`/`useConnectApi` and through `apiFactory`, `apiFactoryWithNamespace` and `streamResultsForCluster` as exported from `lib/k8s/apiProxy` for plugin authors. `KubeObject.useList` now goes through the v2 `useKubeObjectList`, so the in-app resource list views do not hit this. I mention it so the impact is not overstated: this is the v1 and plugin-facing path, not the main list views.

The sort keys off `lastTimestamp`, which suggests events were the intended consumer, and returning the oldest events rather than the most recent is the least useful outcome for that case.

Unrelated to this change, `src/storybook.test.tsx` has one failing snapshot on my machine, `Version Dialog > VersionDialog`. It fails identically on a clean `main` with my changes stashed, so I have left it alone.

### DIFF
--- a/frontend/src/lib/k8s/api/v1/apiProxy.test.ts
+++ b/frontend/src/lib/k8s/api/v1/apiProxy.test.ts
@@ -85,6 +85,33 @@ const modifiedConfigMap = {
   metadata: { ...mockConfigMap.metadata, resourceVersion: '5678' },
   data: { ...mockConfigMap.data, volumeName: 'pvc-xyz-5678' },
 };
+const oldestConfigMap = {
+  ...mockConfigMap,
+  metadata: {
+    ...mockConfigMap.metadata,
+    creationTimestamp: '2023-04-25T20:31:27Z',
+    name: 'oldest',
+    uid: 'abc-oldest',
+  },
+};
+const middleConfigMap = {
+  ...mockConfigMap,
+  metadata: {
+    ...mockConfigMap.metadata,
+    creationTimestamp: '2023-04-26T20:31:27Z',
+    name: 'middle',
+    uid: 'abc-middle',
+  },
+};
+const newestConfigMap = {
+  ...mockConfigMap,
+  metadata: {
+    ...mockConfigMap.metadata,
+    creationTimestamp: '2023-04-27T20:31:27Z',
+    name: 'newest',
+    uid: 'abc-newest',
+  },
+};
 
 describe('apiProxy', () => {
   describe('clusterRequest', () => {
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

PR #6680, opened 2026-07-27
frontend: filterSlice: guard uid and name in search filter

### LINKED ISSUE #6624: frontend: Search filter crashes on a resource missing metadata.uid or name
## Describe the bug

filterResource in frontend/src/redux/filterSlice.ts calls .toLowerCase() straight on metadata.uid and metadata.name when there's a search active, with nothing guarding them:

    const usedMatchCriteria = [
      item.metadata.uid.toLowerCase(),                                     // no guard
      item.metadata.namespace ? item.metadata.namespace.toLowerCase() : '', // guarded
      item.metadata.name.toLowerCase(),                                    // no guard
      ...Object.keys(item.metadata.labels || {}).map(...),                 // guarded
      ...
    ];

namespace and labels are handled defensively, but uid and name aren't. So if a list has even one row whose metadata is missing uid or name, the moment you type in the search box filterResource runs over every row, hits that one, and undefined.toLowerCase() throws. And since this happens during filtering/render, the whole list view drops to the error boundary instead of just skipping that row.

I checked it locally by calling filterResource directly. a normal item (uid + name present) filters fine, but an item with no uid throws "TypeError: Cannot read properties of undefined (reading 'toLowerCase')", and same thing for a missing name.

Expected: a row missing uid or name should just not match (or match on whatever fields it does have), not crash the list.

## To Reproduce
Steps to reproduce the bug:
1. Open a resource list that has an item whose metadata is missing uid or name (some custom/aggregated objects, or anything run through useFilterFunc that isn't a standard API object)
2. Type anything into the list's search box
3. The list crashes instead of filtering

 - Installation type: running from source (main branch)
 - Other: frontend only, it's in the shared filter helper (filterSlice.ts)

Yes (I will propose a PR)

## Additional Context

Fix is basically to guard uid and name the same way namespace and labels already are, like item.metadata.uid?.toLowerCase() ?? ''. Standard API objects always have a server-assigned uid/name so most people won't run into this, but the fact that namespace and labels right next to it are already guarded shows some fields were known to be missable. Doesn't look like a regression.

krsatyamthakur-droid:
## Summary

filterResource dereferences metadata.uid and metadata.name with .toLowerCase() when a search is active, but neither is guarded even though the namespace and labels right next to them are. A resource list that contains a row without a uid or name therefore throws "Cannot read properties of undefined" as soon as the user types in the search box, and because it happens during filtering the whole list drops to the error boundary instead of just skipping that row.

Guard uid and name the same way namespace and labels already are, so a row missing either field simply doesn't match on it rather than crashing the list.

## Related Issue

Fixes #6624

## Changes

- Updated `filterResource` in `frontend/src/redux/filterSlice.ts` to guard `metadata.uid` and `metadata.name` with `?.toLowerCase() ?? ''`
- Added `filterResource` search tests in `frontend/src/redux/filterSlice.test.ts` for a normal item and items missing uid or name

## Steps to Test

1. Open a resource list that has a row whose metadata is missing uid or name (some custom/aggregated objects, or anything run through useFilterFunc that isn't a standard API object).
2. Type anything into the list's search box.
3. Before the fix the list crashes to the error boundary; after it, search filters normally and the incomplete row is just skipped.

## Notes for the Reviewer

- Standard API objects always get a server-assigned uid/name so most people won't hit this, but namespace and labels in the same array were already guarded, so this just brings uid and name in line with them. Doesn't look like a regression.

### DIFF
--- a/frontend/src/redux/filterSlice.test.ts
+++ b/frontend/src/redux/filterSlice.test.ts
@@ -16,6 +16,7 @@
 
 import { getSavedNamespaces } from '../lib/storage';
 import filterReducer, {
+  filterResource,
   FilterState,
   initialState,
   resetFilter,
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

PR #6707, opened 2026-07-28
frontend: fix useClustersVersion polling when tab is in background

shreyas-acharya:
## Problem

The `/version` endpoint was being polled every 10s via a raw `setInterval` in `useClustersVersion`, which kept firing even when the browser tab was hidden (e.g. user switched to another tab). This caused unnecessary network requests to every connected cluster while the user wasn't actively using Headlamp.

## Solution

Replace the manual `setInterval` with React Query's `useQueries` and `refetchInterval`. React Query automatically pauses polling when the tab is hidden, since `refetchIntervalInBackground` defaults to `false`.

A ref with `_.isEqual` is used to stabilise the `clusterNames` reference so that interval timers are not reset on every render when the `clusters` prop gets a new array reference (e.g. while a new cluster is connecting).

## Testing

- Opened Headlamp with multiple clusters connected and observed `/version` requests in browser network tab
- Switched to a different browser tab — requests stopped
- Switched back — requests resumed after 10s
- Connected a new cluster — requests continued at the correct 10s interval without continuous firing

### DIFF
--- a/e2e-tests/tests/multiCluster.spec.ts
+++ b/e2e-tests/tests/multiCluster.spec.ts
@@ -68,6 +68,41 @@ test.describe('multi-cluster setup', () => {
     }
   });
 
+  test('cluster version polling pauses in the background and resumes immediately', async ({
+    page,
+  }) => {
+    await page.evaluate(() => localStorage.setItem('recent_clusters', JSON.stringify(['test'])));
+
+    let versionRequests = 0;
+    page.on('request', request => {
+      if (new URL(request.url()).pathname.endsWith('/clusters/test/version')) {
+        versionRequests++;
+      }
+    });
+
+    await page.reload();
+    await expect.poll(() => versionRequests).toBe(1);
+
+    await page.evaluate(() => {
+      Object.defineProperty(document, 'visibilityState', {
+        configurable: true,
+        get: () => 'hidden',
+      });
+      window.dispatchEvent(new Event('visibilitychange'));
+    });
+    await page.waitForTimeout(11_000);
+    expect(versionRequests).toBe(1);
+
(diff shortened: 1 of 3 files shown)


skoeva (changes requested):
hi, could you address the linting error in the CI?

shreyas-acharya:
@skoeva Have addressed the linting errors in the CI. Could you please re-review the PR.

illume (changes requested):
Thanks for this PR.


<details>

</details>


<details>

- `frontend: fix useClustersVersion polling when tab is in background` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.
- `frontend: address review feedback in useClustersVersion` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

shreyas-acharya:
@illume Thanks for the response.

Have addressed the merge-main commit and commit messages by rebasing the commits. Also have addressed the copilot comments.

Could you please re-review the PR.

shreyas-acharya:
Copilot suggests here that `isPending` attribute of the result might be undefined resulting the cluster being incorrectly marked. 

But the interface of the `tanstack/react-query (v5.51.24)` defines `isPending` as a strict boolean value. Reference to the result interface (https://github.com/TanStack/query/blob/eee3948b455521e3eff3d3c77e22ca09bda90531/packages/query-core/src/types.ts#L605). So the attribute will never be undefined.

@illume Can you please check this out.

illume (approved):
Thanks! 🎉

PR #6715, opened 2026-07-28
app-manifest: Update the prometheus and app-catalog links

yolossn:
this patch bumps the prometheus and app-catalog links


illume (approved):
🎉 thanks!

PR #6717, opened 2026-07-28
frontend: CircularChart: Show loading state while metrics load

### LINKED ISSUE #6716: frontend: CPU and Memory usage tiles briefly show "0.0 %" while metrics are loading
**Describe the bug**

On the Cluster Overview and Node details pages, the **CPU** and **Memory** usage tiles briefly display `0.0 %` (and the memory legend shows `0.00 / N GB`) right after the page loads, then jump to the real usage values a moment later.

It's a quick flash, but on a slower connection or a larger cluster where metrics take longer to return, the false `0.0 %` stays visible long enough to mislead. At a glance it looks like the cluster is using no CPU or memory at all.

**To Reproduce**
1. Use a cluster with `metrics-server` installed and serving metrics.
2. Open the **Cluster Overview** page (or a **Node → details** page).
3. Watch the **CPU** and **Memory** tiles as the page loads. Throttling the network in browser dev tools makes it easier to see.
4. The tiles show `0.0 %` first, then update to the actual usage once the metrics request completes.

**Expected behavior**

While metrics are still loading, the tiles should show a loading indicator (like the Pods tile already does with `…`) instead of `0.0 %`. A real value should appear only once metrics have arrived. `0.0 %` should mean "zero usage", not "still loading".

**Additional context**

The Pods usage tile on the same page doesn't have this problem, it shows `…` until its data is ready. The CPU and Memory tiles seem to be missing the same loading guard.

xonas1101:
**What this PR does**

Fixes the CPU and Memory usage tiles briefly showing `0.0 %` while metrics are still loading.

`CircularChart` computed `used` as `0` whenever metrics had not arrived yet, which is indistinguishable from real zero usage. On first load (or on a slow connection) the CPU and Memory tiles flashed a misleading `0.0 %` before the real values came in.

Now, when metrics have not loaded yet (no data and no error, i.e. not the `noMetrics` state), the chart reports a loading state instead of `0`, so the tile shows a loading indicator until the real usage arrives. This matches how the Pods usage tile already behaves. The `noMetrics` path (metrics-server not installed) and a loaded-but-empty result are unaffected.

**Testing**

Added `MemoryChartLoading` and `CpuChartLoading` stories covering the "metrics not loaded yet" state, with snapshots confirming the tile renders a loading indicator rather than `0.0 %`.

Closes #6716

### DIFF
--- a/frontend/src/components/cluster/Charts/ResourceCharts.stories.tsx
+++ b/frontend/src/components/cluster/Charts/ResourceCharts.stories.tsx
@@ -196,6 +196,18 @@ MemoryChartPod.args = {
   resource: 'memory',
 };
 
+// Metrics not loaded yet (items present, metrics null, no error): should show a
+// loading state rather than a misleading 0%.
+export const MemoryChartLoading = Template1.bind({});
+MemoryChartLoading.args = {
+  chart_props: {
+    items: [mockNode],
+    itemsMetrics: null,
+    noMetrics: false,
+  },
+  resource: 'memory',
+};
+
 // CPU Chart Stories
 export const CpuChartWithMetrics = Template1.bind({});
 CpuChartWithMetrics.args = {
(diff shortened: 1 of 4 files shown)


xonas1101:
@illume this PR is ready for review. PTAL!

xonas1101:
@illume PTAL

PR #6718, opened 2026-07-28
frontend: resourceMap: Fix fit to screen, and resize issues

illume:
## Summary

Restore Resource Map viewport fitting after selecting a grouped node. The map now calculates bounds from the newly generated layout instead of consulting React Flow's stale lookup for the previously rendered topology.

The change also hides the map performance diagnostic controls by default. They remain available by setting `REACT_APP_HEADLAMP_ENABLE_MAP_PERFORMANCE_FEATURES=true`.

## Related Issue

Regression after #4991.

## Changes

- Build a parent-aware node lookup for newly laid-out nodes so nested resource bounds use absolute graph coordinates.
- Keep toolbar viewport actions on React Flow's live node lookup.
- Forward viewport movement through `onMoveStart`, matching the callback contract used to detect manual movement.
- Add direct coverage for the `GraphRenderer` move-start wiring and nested layout bounds.
- Hide **Incremental Updates** and **Performance Stats** unless the performance feature flag is enabled.
- Update the affected Resource Map story snapshots.
- Expand selected namespaces from filtered resources, then apply namespace-local simplification.
- Include collapsed state in layout cache keys so resizing cannot restore the wrong parent/child layout.

## Behavior Comparison

Compared against pre-performance commit `ea308a9bb` using the same BasicExample story and viewport sizes:

- Both revisions show the same three resources after selecting **Node with children**.
- In default `100%` mode, the fixed map recenters the selected layout and keeps every node in view.
- In mobile Fit mode, selection updates the transform with every selected node inside the viewport.
- Nested/sub-flow bounds no longer depend on the previous React Flow topology.

#### 0.43.0 (before before)

https://github.com/user-attachments/assets/84cac7d8-317c-483c-9855-7804aa3349db

#### Main branch (before, broken)

https://github.com/user-attachments/assets/79d5a6b7-af27-4230-b4e5-02b28b20902d

#### this PR (after, fixed)

https://github.com/user-attachments/assets/da0aea47-f275-4478-9f0a-63449c165660

#### Nodes inside namespace not there issue

> map simplification seems glitchy, namespace node doesn't show any child elements when simplification is enabled (edited)

https://github.com/user-attachments/assets/e6801320-bfb7-4e3a-9996-876280dcdd5f

#### Simplification resizing issue

> there's also a visual bug when resizing window, parent node (namespace) becomes broken and children nodes appear or disappear

https://github.com/user-attachments/assets/837d5fe4-ef66-467f-9e97-0d73bd47a50e

## Steps to Test

0. See videos.
1. Open the Resource Map and choose **Fit to screen**.
2. Select a grouped node such as **Node with children**.
3. Verify the selected resources are fitted and centered in the viewport.
4. Verify **Incremental Updates** and **Performance Stats** are hidden by default.
5. Start the frontend with `REACT_APP_HEADLAMP_ENABLE_MAP_PERFORMANCE_FEATURES=true` and verify both controls are visible.
6. With simplification enabled, select a small namespace and verify all resources are shown.
7. Select a very large namespace and verify it remains simplified.
8. Resize repeatedly between desktop and mobile sizes; verify the namespace parent and children remain stable.
9. Also check mobile view. (As well as improving over main branch... things fit a bit better also compared to previous Headlamp release 0.43.0)

## Validation

- `npm test -- --run src/components/resourceMap` (296 tests)
- `npm run frontend:tsc`
- Focused ESLint for changed Resource Map files
- `cd e2e-tests && npx playwright test tests/resourceMap.spec.ts` (desktop/mobile resize loop)

Assisted by copilot.

### DIFF
--- a/frontend/src/components/resourceMap/GraphRenderer.test.tsx
+++ b/frontend/src/components/resourceMap/GraphRenderer.test.tsx
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { act, render } from '@testing-library/react';
+import { GraphRenderer } from './GraphRenderer';
+
+const mocks = vi.hoisted(() => ({
+  reactFlowProps: undefined as Record<string, any> | undefined,
+}));
+
+vi.mock('@xyflow/react', () => ({
+  Background: () => null,
+  BackgroundVariant: { Dots: 'dots' },
+  ConnectionMode: { Loose: 'loose' },
+  Controls: ({ children }: { children: React.ReactNode }) => children,
+  ReactFlow: (props: Record<string, any>) => {
+    mocks.reactFlowProps = props;
+    return <div>{props.children}</div>;
+  },
+}));
+
+vi.mock('./GraphControls', () => ({
+  GraphControls: ({ children }: { children: React.ReactNode }) => children,
+}));
+
+vi.mock('./edges/GraphEdgeComponent', () => ({
+  GraphEdgeComponent: () => null,
+}));
+
+vi.mock('./nodes/KubeObjectNode', () => ({
+  KubeObjectNodeComponent: () => null,
+}));
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+describe('GraphRenderer', () => {
+  beforeEach(() => {
+    mocks.reactFlowProps = undefined;
+  });
+
+  it('forwards continuous viewport movement through onMove', () => {
+    const onMove = vi.fn();
+    render(<GraphRenderer nodes={[]} edges={[]} onMove={onMove} />);
+
+    expect(mocks.reactFlowProps?.onMove).toBe(onMove);
+    expect(mocks.reactFlowProps?.onMoveStart).toBeUndefined();
+
+    const viewport = { x: 10, y: 20, zoom: 1 };
+    act(() => mocks.reactFlowProps?.onMove(null, viewport));
+    expect(onMove).toHaveBeenCalledWith(null, viewport);
+  });
+});
--- a/frontend/src/components/resourceMap/GraphRenderer.tsx
+++ b/frontend/src/components/resourceMap/GraphRenderer.tsx
@@ -26,7 +26,7 @@ import {
   EdgeMouseHandler,
   Node,
   NodeMouseHandler,
-  OnMoveStart,
+  OnMove,
   ReactFlow,
 } from '@xyflow/react';
 import React from 'react';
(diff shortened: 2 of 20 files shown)


frontend/src/components/resourceMap/GraphRenderer.tsx
```diff
@@ -125,7 +125,7 @@ export function GraphRenderer({
       elementsSelectable
       onNodeClick={onNodeClick}
       onEdgeClick={onEdgeClick}
-      onMove={onMoveStart}
+      onMoveStart={onMoveStart}
```
rootp1: onMove changed to onMoveStart here. onMove fires continuously during a drag or zoom while onMoveStart fires once at the start. Please confirm no other logic depended on continuous onMove updates. The new test only checks the prop is forwarded, it does not exercise viewportMovedRef in GraphView through an actual drag.

frontend/src/components/resourceMap/useGraphViewport.ts
```diff
@@ -31,6 +31,56 @@ import { maxZoom, minZoom, viewportPaddingPx } from './graphConstants';
...
+      node.id,
+      {
+        ...node,
+        measured: {
+          width: node.measured?.width ?? node.width,
+          height: node.measured?.height ?? node.height,
+        },
+        internals: {
+          positionAbsolute: getAbsolutePosition(node),
+          z: 0,
```
rootp1, line 75: z is hardcoded to 0 for every node in this manually built nodeLookup. If getNodesBounds from xyflow ever uses z ordering internally this could produce wrong results in edge cases. Worth a short comment explaining why 0 is always safe here.

frontend/src/components/resourceMap/graph/graphForNamespaceSelection.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import type Namespace from '../../lib/k8s/namespace';
+import type { GraphEdge, GraphNode } from './graph/graphModel';
+import { getGraphForSelectedNamespace } from './graph/graphSimplification';
+
+type Graph = { nodes: GraphNode[]; edges: GraphEdge[] };
+type SimplifiedGraph = Graph & { simplified: boolean };
+
+export function getGraphForNamespaceSelection(
```
rootp1, line 24: This is a small pure function extracted into its own top level file mainly for testability. Reasonable tradeoff, but GraphView.tsx now depends on three small graph transforming modules for closely related behavior, graphSimplification, graphForNamespaceSelection, and graphLayout. Worth considering consolidating these under graph as the module count grows.

frontend/src/components/resourceMap/GraphRenderer.tsx
illume: Restored the original continuous `onMove` contract and updated the renderer test to assert `onMove` is forwarded while `onMoveStart` is absent. This avoids changing the existing `viewportMovedRef` semantics; the resize behavior is now covered separately by the routed Playwright e2e test.

frontend/src/components/resourceMap/useGraphViewport.ts
illume: Added a comment at the synthetic InternalNode construction explaining that `getNodesBounds` only consumes geometry, while `z` is required by the type and does not affect the computed box. The value is therefore intentionally `0` for every node.

frontend/src/components/resourceMap/graph/graphForNamespaceSelection.ts
illume: Consolidated `graphForNamespaceSelection` and its test under `resourceMap/graph/`, next to `graphSimplification` and `graphLayout`, and updated GraphView imports accordingly.

frontend/src/components/resourceMap/GraphRenderer.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { act, render } from '@testing-library/react';
+import { GraphRenderer } from './GraphRenderer';
```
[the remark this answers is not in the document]
illume: doesn't fail

PR #5719, opened 2026-05-17
backend: k8cache: Fix named resource authorization

harrshita123:
## Summary

  Fixes cache authorization for named Kubernetes resource requests. The cache middleware now builds SelfSubjectAccessReview requests with the Kubernetes resource type, such as pods, instead of incorrectly using the object name, such as nginx.

  ## Related Issue

  Fixes #5718

  ## Changes

  - Added structured parsing for proxied Kubernetes API paths in backend/pkg/k8cache/authorization.go.
  - Updated cache authorization to include resource, name, namespace, API group, version, and subresource in ResourceAttributes.
  - Added regression tests for named resources, subresources, and namespace paths.

  ## Steps to Test

  1. Run GOCACHE=/private/tmp/headlamp-go-build-cache npm run backend:lint.
  2. Run GOCACHE=/private/tmp/headlamp-go-build-cache npm run backend:test.
  3. Confirm named resource paths like /api/v1/namespaces/default/pods/nginx authorize against resource pods with name nginx.

  ## Screenshots

  Not applicable. Backend-only change.

  ## Notes for the Reviewer

  This only changes cache authorization path parsing. The intent is to avoid false forbidden responses when response caching is enabled for named Kubernetes resource requests.

PR #6720, opened 2026-07-28
backend: k8cache: key cache entries by resource, not path tail

### LINKED ISSUE #6713: backend: k8cache: Cache keys do not uniquely identify responses, causing cross-resource data disclosure
## Describe the bug

k8cache generates cache keys that do not uniquely identify the response they store. `GenerateKey` builds a key from only `apiGroup + kind + namespace + contextID`, where `kind` is simply **the last segment of the URL path** (`ExtractNamespace` returns `urls[n-1]`), and the query string is dropped entirely.

This produces three classes of key collision on **valid** Kubernetes API paths:

```
/api/v1/namespaces/default/secrets/nginx           => "+nginx+default+mycluster"
/api/v1/namespaces/default/configmaps/nginx        => "+nginx+default+mycluster"
/api/v1/namespaces/default/services/nginx          => "+nginx+default+mycluster"
/api/v1/namespaces/default/pods/nginx              => "+nginx+default+mycluster"
/api/v1/namespaces/default/serviceaccounts/nginx   => "+nginx+default+mycluster"

/api/v1/namespaces/default/pods/podA/log           => "+log+default+mycluster"
/api/v1/namespaces/default/pods/podB/log           => "+log+default+mycluster"

/api/v1/namespaces/default/pods?labelSelector=app=foo => "+pods+default+mycluster"
/api/v1/namespaces/default/pods?labelSelector=app=bar => "+pods+default+mycluster"
/api/v1/namespaces/default/pods?limit=1               => "+pods+default+mycluster"
/api/v1/namespaces/default/pods                       => "+pods+default+mycluster"
```

1. **Resource type is absent.** For a named GET, the object *name* occupies the `kind` segment, so every core-group resource sharing a name in a namespace shares one cache entry.
2. **Subresources are not distinguished.** `pods/A/log` and `pods/B/log` both key on `log`.
3. **Query parameters are ignored.** A `labelSelector`-filtered, `fieldSelector`-filtered, or `limit`-paginated list shares a key with the unfiltered list.

### Why this is a data-disclosure issue, not only a correctness issue

`contextID` is the *cluster* context key, not a per-user key, so entries are shared by every user of that cluster. The middleware authorizes the real request and then serves whatever sits under the coarse key:

1. An admin opens Secret `nginx` in `default`. Its body is cached under `+nginx+default+prod`.
2. A low-privilege user requests ConfigMap `nginx` in `default`.
3. `IsAllowed` runs a `SelfSubjectAccessReview` for `get configmaps/nginx` — this legitimately **passes**.
4. `LoadFromCache` returns the entry under that key: **the Secret's body**, with `X-HEADLAMP-CACHE: true`.

The authorization check is never asked about the resource that is actually served. Sharing a name between a Secret and a ConfigMap (`app-config`, `nginx`, `redis`) is routine, so this needs no unusual cluster layout.

The same mechanism serves pod A's logs in response to a pod B log request, and serves an unfiltered or paginated list in response to a `labelSelector` query.

### Expected behavior

A cache key should uniquely identify the response it stores. Two requests may share a key only if they are guaranteed to produce the same body. Distinct resource types, distinct named objects, distinct subresources, and distinct query parameters must never collide.

### Affected code

- `backend/pkg/k8cache/cacheStore.go` — `GenerateKey`, `buildCacheKey`, `ExtractNamespace`
- `backend/pkg/k8cache/cacheInvalidation.go` — `invalidateCacheKeysForResourceEvent` mirrors the same assumption; its doc comment states *"named GET paths use the object name as the kind segment"*

Scope: only reachable with `--cache-enabled`, which defaults to `false`.

### A. Unit-level, no cluster required

Drop this in `backend/pkg/k8cache/` and run `go test ./pkg/k8cache/ -run TestKeyCollision -v`:

```go
package k8cache_test

import (
	"net/url"
	"testing"

	"github.com/kubernetes-sigs/headlamp/backend/pkg/k8cache"
)

func TestKeyCollision(t *testing.T) {
	gen := func(raw string) string {
		u, _ := url.Parse(raw)
		k, err := k8cache.GenerateKey(u, "mycluster")
		if err != nil {
			t.Fatal(err)
		}
		return k
	}

	secret := gen("/clusters/c/api/v1/namespaces/default/secrets/nginx")
	cm := gen("/clusters/c/api/v1/namespaces/default/configmaps/nginx")
	if secret == cm {
		t.Errorf("Secret and ConfigMap share a cache key: %q", secret)
	}

	logA := gen("/clusters/c/api/v1/namespaces/default/pods/podA/log")
	logB := gen("/clusters/c/api/v1/namespaces/default/pods/podB/log")
	if logA == logB {
		t.Errorf("logs of two pods share a cache key: %q", logA)
	}

	all := gen("/clusters/c/api/v1/namespaces/default/pods")
	filtered := gen("/clusters/c/api/v1/namespaces/default/pods?labelSelector=app%3Dfoo")
	if all == filtered {
		t.Errorf("filtered and unfiltered lists share a cache key: %q", all)
	}
}
```

All three assertions fail.

### B. End-to-end

1. Start the backend with response caching on: `./backend/headlamp-server --cache-enabled --kubeconfig ~/.kube/config`
2. Create two resources sharing a name in one namespace:
   ```
   kubectl -n default create secret generic nginx --from-literal=password=hunter2
   kubectl -n default create configmap nginx --from-literal=greeting=hello
   ```
3. As a user with full access, request the Secret so it enters the cache:
   ```
   curl -H "Authorization: Bearer $ADMIN_TOKEN" \
     http://localhost:4466/clusters/my-cluster/api/v1/namespaces/default/secrets/nginx
   ```
4. As a user whose RBAC grants `get` on `configmaps` but **not** on `secrets`, request the ConfigMap:
   ```
   curl -i -H "Authorization: Bearer $LIMITED_TOKEN" \
     http://localhost:4466/clusters/my-cluster/api/v1/namespaces/default/configmaps/nginx
   ```
5. The response carries `X-HEADLAMP-CACHE: true` and contains the **Secret's** body, including the base64 `password` value — a resource that user cannot read.

 - Installation type: In-Cluster / Container-Image (any deployment started with `--cache-enabled`)
 - Headlamp Version: reproduced on `main` @ `bc37b0f` (2026-07-28); `GenerateKey`, `buildCacheKey`, and `ExtractNamespace` on `main` are unchanged from the code quoted above
 - Other: requires `--cache-enabled` (default `false`). Backend-only; no frontend involvement.

Yes (I will propose a PR).

## Additional Context

Not a regression in the strict sense — the key format has behaved this way since `k8cache` was introduced. However, **#5719** ("backend: k8cache: Fix named resource authorization", merged) fixed exactly this parsing flaw on the *authorization* side, so `SelfSubjectAccessReview` now correctly uses the resource type (`pods`) rather than the object name (`nginx`). The cache-key side was not updated to match. The result is that the middleware now reliably authorizes the correct resource and then serves a different one, which makes the wrong-body path *more* reachable than before that fix.

Related but distinct prior work:

- **backend: k8cache: Fix named resource authorization #5719** (merged) — fixed the authorization half of this same parsing flaw; left the key untouched.
- **backend: k8cache: percent-encode '+' in cache key fields to avoid delimiter collision #6127** (merged) — percent-encoded `+` in key fields. Addresses delimiter ambiguity, not semantic collision.
- **Corrupted cache keys and "garbage" resource kinds due to fragile URL parsing logic #5287** (closed) — "garbage" kinds extracted from non-API paths. Fixed by the `IsKubernetesAPIPath` guards; concerns *invalid* paths, whereas this report concerns *valid* paths colliding.
- **backend: k8cache: Invalidate named GET cache keys on informer events #6230** (merged) — added named-GET invalidation; it encodes the same assumption and would need updating alongside any fix.
- **backend: informer cache invalidation leaves all-namespace list cache stale #5776**, **backend: k8cache misses invalidation for non-allowlisted resources #6457** (open) — invalidation coverage gaps; adjacent, but about *when* entries are dropped rather than *what* a key identifies.

Anjali-Chauhan1:
## Summary

This PR fixes a data-disclosure bug in k8cache: cache keys did not identify the response they stored. `GenerateKey` used `apiGroup + kind + namespace + contextID`, where `kind` was just the last path segment and the query string was dropped — that segment is the resource type only for lists, the object name for a named GET, and the subresource for a subresource GET. Different objects therefore shared one entry.

Entries are keyed by cluster context, not by user, so they are shared across every user of a cluster. A user with `get` on configmaps but not secrets could request ConfigMap `nginx`, pass the SelfSubjectAccessReview legitimately, and be served the Secret of the same name cached earlier by an admin. Authorization is never asked about the resource actually served.

#5719 fixed this parsing flaw on the authorization side; only key generation was left behind. This PR routes it through the same parser.

## Related Issue

Fixes #6713

## Changes

- **Key is now `group + resource + namespace + context + name + variant`**, resource type and object name in separate segments. `variant` hashes the API version, subresource, and canonicalised query, so multi-version CRDs and filtered lists get their own entries. The version stays out of the prefix, keeping invalidation version-agnostic.
- **Requests resolve through `parseAPIResourceRequest`**, the parser SelfSubjectAccessReview already uses, so authorization and caching cannot disagree about what a request addresses. Discovery paths fall back to the trailing segment of the *API-relative* path, so proxied and direct routing agree and `/api` and `/apis` stay distinct.
- **Invalidation moved to prefix deletes**, since an object now has one entry per variant. One sweep per event, duplicate prefixes dropped, eviction still exact — a Pod `nginx` changing no longer drops the Service or ConfigMap of that name.
- **Eviction brackets modifying requests** (before and after forwarding), so a read racing the write cannot leave pre-write state behind. The old refetch-and-recache is gone: it stored under the write's query, a variant no read can address. A collection `DELETE` also evicts the objects it removed; a create does not touch its siblings.
- **Cacheability and invalidation are separate questions.** `NewAPIRequest` resolves any API request; `Cacheable()` covers GETs excluding watch streams, subresources, and self-subject reviews; `InvalidatesCache()` covers the mutating methods. A subresource write is not cacheable but still evicts its parent object.
- **Smaller items:** one predicate for self-subject reviews (bypass and store can't drift), 1-minute TTL for single-use `continue`-token pages, no cache sweep for informer events with an unchanged `resourceVersion`, key parsers in `context_cleanup.go`/`redactCacheKey` bumped to the new segment count, and `CacheKey` (`key.go`), `ExtractNamespace`, and `ErrHandled` removed as unused.
- **Tests:** every collision class from the issue, an end-to-end store/serve test proving a cached Secret is not returned for a ConfigMap request, a round-trip test pinning `GenerateKey` output to the invalidation prefixes, per-method cacheability coverage, and a subresource-write-evicts-parent test. `MockCache.GetAll` was a `nil, nil` stub and is now implemented.

## Steps to Test

**1. Confirm the bug on `main`** — drop this in `backend/pkg/k8cache/`, run `go test ./pkg/k8cache/ -run TestKeyCollision -v`:

```go
package k8cache_test

import (
	"net/url"
	"testing"

	"github.com/kubernetes-sigs/headlamp/backend/pkg/k8cache"
)

func TestKeyCollision(t *testing.T) {
	gen := func(raw string) string {
		u, _ := url.Parse(raw)
		k, err := k8cache.GenerateKey(u, "mycluster")
		if err != nil {
			t.Fatal(err)
		}
		return k
	}

	pairs := [][2]string{
		{"/clusters/c/api/v1/namespaces/default/secrets/nginx", "/clusters/c/api/v1/namespaces/default/configmaps/nginx"},
		{"/clusters/c/api/v1/namespaces/default/pods/podA/log", "/clusters/c/api/v1/namespaces/default/pods/podB/log"},
		{"/clusters/c/api/v1/namespaces/default/pods", "/clusters/c/api/v1/namespaces/default/pods?labelSelector=app%3Dfoo"},
		{"/clusters/c/apis/example.io/v1alpha1/namespaces/default/widgets", "/clusters/c/apis/example.io/v1/namespaces/default/widgets"},
		{"/api", "/apis"},
	}

	for _, p := range pairs {
		if gen(p[0]) == gen(p[1]) {
			t.Errorf("%s and %s share a cache key: %q", p[0], p[1], gen(p[0]))
		}
	}
}
```

All five fail on `main` and pass on this branch.

**2. Package suite:** `go test ./pkg/k8cache/ -v` — 87 tests / 253 including subtests, among them `TestCachedResponseIsNotServedToADifferentRequest`, which drives the real store→serve path.

**3. Live cluster:** run the server with `--cache-enabled`, create a Secret and a ConfigMap both named `nginx` in `default`, warm the cache by fetching the Secret as an admin, then fetch the ConfigMap as a user with `get` on configmaps only:

```bash
./backend/headlamp-server --cache-enabled --kubeconfig ~/.kube/config

kubectl -n default create secret generic nginx --from-literal=password=hunter2
kubectl -n default create configmap nginx --from-literal=greeting=hello

curl -H "Authorization: Bearer $ADMIN_TOKEN" \
  http://localhost:4466/clusters/my-cluster/api/v1/namespaces/default/secrets/nginx

curl -i -H "Authorization: Bearer $LIMITED_TOKEN" \
  http://localhost:4466/clusters/my-cluster/api/v1/namespaces/default/configmaps/nginx
```

On `main` the second response carries `X-HEADLAMP-CACHE: true` and the Secret's base64 password; on this branch it returns the ConfigMap.

**4. Full backend:** `go test ./...` and `golangci-lint run ./...` — 15/15 packages green, 0 lint issues.

## Notes for the Reviewer

- **Behaviour change:** subresource and watch *responses* are no longer cached. Subresource *writes* still invalidate, so this costs cache hits, not correctness. Happy to split the bypass into its own PR if you'd rather keep the key fix narrow.
- **Memory:** a simulated 10-namespace × 50-pod session goes from ~520 to 550 entries (+6%). The key fix alone would have doubled it; the subresource bypass holds it down, and the 1-minute TTL bounds continue-token pages. The cache is TTL-only with no size bound, so this is worth being deliberate about.
- **Benchmark:** `BenchmarkGenerateKey` goes from 1139 ns/op, 336 B, 3 allocs to 1994 ns/op, 664 B, 10 allocs — the cost of hashing the variant. Partly undoes #6420's hot-path work; happy to discuss.
- **Invalidation cost:** `deleteByPrefixes` walks every entry where the old exact-key delete was O(1), and a mutation sweeps twice. Bracketing closes the common race but not the one where a read fetches before the write and stores after the second eviction. A prefix index or debouncing would fix it; I'd rather not fold that into a bugfix.
- **Removed exported symbols:** `CacheKey`, `ExtractNamespace`, `ErrHandled` — all callerless after this change. Say the word and I'll restore them rather than touch public surface in a bugfix.
- **Integration tests:** `TestCacheMiddleware_CacheInvalidation` asserted a hit on the first GET after a write, which only held while invalidation refilled what it evicted; it now requires a miss then a hit, so it fails if eviction stops. The two `_RealK8s` tests need a cluster I can't run locally — please watch them in CI.
- **Not run locally:** `go test -race` (my cgo is 32-bit only). CI doesn't run it either.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache invalidation after create, update, and delete operations, including collection, namespaced, subresource, and named-resource changes.
  * Prevented stale or conflicting cached results across API versions, query variations, and resource names.
  * Excluded watch requests and sensitive authorization review requests from caching.
  * Improved handling of direct and proxied API requests.

* **Performance**
  * Added shorter cache lifetimes for paginated requests and ignored unchanged resource updates to reduce unnecessary work.


Anjali-Chauhan1:
cc : @illume

illume (changes requested):
Thanks for these changes.


<details>

- `backend: k8cache: key cache entries by resource, not path tail` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for these changes.


illume (changes requested):
Thanks for these changes.


I noticed the GitHub CI backend test job is failing. Can you please fix the failing tests? You can run `cd backend && go test ./...` locally to see the errors.

<details>

</details>

illume (changes requested):
Thanks for the contribution.


Anjali-Chauhan1:
@illume , I have addressed the copilot comments . Please have a look whenever you get the chance :)

Issue #6607, opened 2026-07-22
WIP: 0.44.0 release notes draft and testing coordination


illume:

illume:
# Release testing

• Windows: https://github.com/kubernetes-sigs/headlamp/actions/runs/30350911146
• Mac: https://github.com/kubernetes-sigs/headlamp/actions/runs/30350914335
• Linux: https://github.com/kubernetes-sigs/headlamp/actions/runs/30350917355
Nightly container image is 0.44.0: https://github.com/orgs/headlamp-k8s/packages/container/headlamp/1074097520?tag=nightly


---

Please write a comment if you're going to test an item and if you've completed testing an item.

- [x] **Test web sockets** (@yolossn @illume) (Tested and Ran into issues, Deleting a pod isn't reflecting in the UI automatically, bug fixed and e2e test added using kubectl to delete)
  - Run the app in static server mode (`./backend/headlamp-server -static-html-dir ./frontend/build...`)
  - Open the pods list view
  - Delete a pod using kubectl
  - Verify that the list reflects the changes

- [x] **App: Run with plugins** (@yolossn)
  - Build app with plugins: `make && make app`
  - Run the app (UI should load in less than 1 minute)
  - Verify plugins are present: App Catalog, Prometheus, App Catalog

- [x] **App: Test Prometheus plugin** (@yolossn)
  - Run the app
  - Go to the Pods page and choose a pod
  - Verify the Prometheus plugin shows at the top and displays correctly

- [ ] **Image: Test Prometheus plugin** ( )
  - Run headlamp from image
  - Go to the Pods page and choose a pod
  - Verify the Prometheus plugin shows at the top and displays correctly

- [x] **App: Test Plugin Catalog**   ( @vyncent-t  )
  - Run the app
  - Verify the Plugin Catalog sidebar item is available and open it
  - Verify plugins appear from ArtifactHub
  - Install a plugin
  - Verify the plugin shows as installed
  - Delete the plugin
  - Verify the plugin shows as available to install

- [x] **Test creating a resource through edit button** ( @vyncent-t )
  - Open Headlamp via app or image
  - Choose a cluster where you have admin rights
  - Click Create (bottom left), add a resource (e.g., ConfigMap), and apply
  - Verify the resource is created

- [x] **Check Workloads page**   ( @vyncent-t )
  - Open Headlamp via app or image
  - Go to the Workloads page of a cluster with diverse resource types
  - Verify resources and top charts display correctly

[...]

PR #6738, opened 2026-07-29
Docs/fix html static dir flag

Pragati5-DEBUG:
## Summary
- Fixes the release testing checklist generator to use `-html-static-dir` instead of the invalid `-static-html-dir`.

## Related
Noticed while testing websockets for #6607.


illume (approved):
🎉 thanks!

PR #4810, opened 2026-03-01
frontend: Add JobSet as a first-class workload resource

kannon92:
## Summary

This PR adds JobSet (from `jobset.x-k8s.io/v1alpha2`) as a first-class workload resource in Headlamp. JobSet is a Kubernetes-native API for managing a group of Jobs as a unit, commonly used for ML/AI training workloads, HPC, and batch processing.

## Related Issue

N/A

## Changes

- Added `JobSet` KubeObject class (`frontend/src/lib/k8s/jobSet.ts`) with proper API group, version, and type definitions
- Added JobSet list view with conditions column (`frontend/src/components/jobset/List.tsx`)
- Added JobSet detail view reusing the workload details pattern with owned Jobs and Pods sections
- Added sidebar entry under Workloads for "Job Sets"
- Added routes for `/jobsets` (list) and `/jobsets/:namespace/:name` (detail)
- Integrated JobSet into global search, resource map (graph model, relations, sources, icons), and workload overview
- Added Storybook stories for the JobSet list component
- Fixed pod listing for JobSet to filter by `jobset.sigs.k8s.io/jobset-name` label
- Added `jobset.x-k8s.io` to the Workloads resource category API groups
- Added i18n glossary entry for "Job Sets"

## Steps to Test

1. Install the JobSet CRD on a cluster (`kubectl apply --server-side -f https://github.com/kubernetes-sigs/jobset/releases/latest/download/manifests.yaml`)
2. Create a JobSet resource in a namespace
3. Navigate to the Workloads section in the sidebar — "Job Sets" should appear
4. Click "Job Sets" to see the list view with name, namespace, conditions, and age columns
5. Click a JobSet to see the detail view with owned Jobs and Pods sections
6. Verify JobSet appears in global search results
7. Verify JobSet appears in the resource map with correct relations to Jobs

## Screenshots (if applicable)

Screenshot From 2026-03-02 20-16-48
Screenshot From 2026-03-02 20-17-26

## Notes for the Reviewer

- JobSet uses the `jobset.x-k8s.io/v1alpha2` API version which is the current stable alpha release
- The owned pods section uses the `jobset.sigs.k8s.io/jobset-name` label selector since JobSets don't have a standard `spec.selector`
- An `OwnedJobsSection` was added to show Jobs belonging to a JobSet in the detail view
- This follows the same patterns used by other workload resources (Jobs, CronJobs, Deployments, etc.)

PR #6744, opened 2026-07-30
frontend: Add LeaderWorkerSet as a first-class workload resource

### LINKED ISSUE #6613: Add LeaderWorkerSet integration to workload
Headlamp should display LeaderWorkerSets.

LWS are a workload and LWS should be a workload option in the workloads dropdown (similar to Job, Deployment, StatefulSet, JobSet).

Inference or AI users.

No.

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

adityachawla005:
## Summary

This adds LeaderWorkerSet (`leaderworkerset.x-k8s.io/v1`) as a first-class workload, following the JobSet implementation from #4810 as suggested in the issue. A LeaderWorkerSet runs a group of pods as one unit — a leader plus its workers — and is mostly used for distributed inference and AI serving, so it fits alongside the other workloads in the sidebar.

## Related Issue

Fixes #6613

## Changes

- A `LeaderWorkerSet` model class with types and unit tests
- A list view with ready groups, group size and conditions columns
- A detail view reusing the workload details pattern, showing the owned StatefulSets and their Pods
- Sidebar entry under Workloads, plus list and detail routes
- Global search, resource map and workloads overview integration
- The resource category entry and i18n strings

To list the owned StatefulSets I pulled a `StatefulSetsListRenderer` out of the StatefulSet list, the same way the JobSet detail view reuses the Job list renderer.

## Steps to Test

1. Install the CRD and controller:

   ```
   kubectl apply --server-side -f https://github.com/kubernetes-sigs/lws/releases/latest/download/manifests.yaml
   ```

2. Create a LeaderWorkerSet:

   ```yaml
   apiVersion: leaderworkerset.x-k8s.io/v1
   kind: LeaderWorkerSet
   metadata:
     name: inference-server
   spec:
     replicas: 2
     leaderWorkerTemplate:
       size: 3
       leaderTemplate:
         spec:
           containers: [{ name: leader, image: nginx:1.27 }]
       workerTemplate:
         spec:
           containers: [{ name: worker, image: nginx:1.27 }]
   ```

3. Go to Workloads → Leader Worker Sets — it should show up in the sidebar and list
4. Open one — the detail view shows its StatefulSets, then its Pods
5. Check that it also shows up in the workloads overview and the resource map

I tested all of this on a local kind cluster with the LWS controller installed.

## Screenshots

1-sidebar-and-list
2-details-conditions
3-details-statefulsets-and-pods
4-workloads-overview
5-resource-map

## Notes for the Reviewer

**The resource map matches on a label rather than owner references.** 
Only the leader StatefulSet is actually owned by the LeaderWorkerSet, and each worker group's StatefulSet is owned by that group's leader pod. I had it on owner references first, and the worker groups showed up disconnected in the map. Switching to the controller's `leaderworkerset.sigs.k8s.io/name` label fixed it and made the map agree with the detail view.

**Health is worked out per item instead of by comparing replica counts.** A plain ready-vs-desired check can't tell a partially ready group apart from a failed one, and it misses rollouts entirely. So a rollout (`UpdateInProgress`) reads as transitional, a workload that's still starting up (`Progressing` with nothing ready yet) also reads as transitional rather than failed, and a deliberate scale to zero reads as healthy.

**Two side effects worth flagging.** Extracting `StatefulSetsListRenderer` gave the StatefulSets page a table `id` and `reflectInURL` it didn't have before, which brings it in line with the pod, job, secret and role lists. And I skipped the containers section for LeaderWorkerSet — it keeps its containers under `spec.leaderWorkerTemplate` rather than `spec.containers`, so the section rendered as an empty placeholder. The leader and worker containers are visible in the owned StatefulSets instead.

Happy to hear your suggestions.


kannon92:
@coderabbitai review

kannon92:
@illume can you request a copilot review?

frontend/src/lib/k8s/leaderWorkerSet.ts
```diff
@@ -0,0 +1,143 @@
...
+   * Classifies the leader worker set into a coarse health category for the
+   * Workloads overview chart. Unlike job sets, a leader worker set does have
+   * replica fields, so readiness is judged by comparing ready against desired
+   * replicas. A group that has been explicitly scaled to zero is treated as
+   * healthy rather than failed, since that is the requested state, and one that
+   * is still coming up is transitional rather than failed.
+   */
+  getHealth(): WorkloadHealthCategory {
+    const conditions = this.status?.conditions || [];
+    const isTrue = (type: string) => conditions.some(c => c.type === type && c.status === 'True');
```
kannon92: nit: this utility is duplicated across LWS, JobSet and Job.

Can we consider declaring this elsewhere and reusing? Ideally add some tests around it also.

frontend/src/components/leaderworkerset/List.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useTranslation } from 'react-i18next';
+import LeaderWorkerSet from '../../lib/k8s/leaderWorkerSet';
+import ResourceListView from '../common/Resource/ResourceListView';
+
+// Explicit priority to make the rendered condition stable and meaningful: an
+// in-flight rollout or scale-up says more about the current state than the
+// steady-state Available condition it sits alongside.
+const conditionPriority = ['UpdateInProgress', 'Progressing', 'Available'];
+
+function getLeaderWorkerSetCondition(leaderWorkerSet: LeaderWorkerSet): string {
```
kannon92: nit: this is really similar to JobSet's function on conditions.

```
function getJobSetCondition(jobSet: JobSet): string {
  const conditions = jobSet.status?.conditions;
  if (!conditions) return '-';

  const trueConditions = conditions.filter(c => c.status === 'True');
  if (trueConditions.length === 0) {
    return '-';
  }

  let selected = trueConditions[0];
  let bestPriorityIndex = conditionPriority.length;

  for (const cond of trueConditions) {
    const idx = conditionPriority.indexOf(cond.type);
    if (idx !== -1 && idx < bestPriorityIndex) {
      bestPriorityIndex = idx;
      selected = cond;
    }
  }

  return selected.type ?? '-';
}
```

Could we consider just passing in `conditions` and reusing this function in the two areas it is used?

frontend/src/lib/k8s/leaderWorkerSet.ts
```diff
@@ -0,0 +1,143 @@
...
+    // which briefly match while pods are being replaced.
+    if (isTrue('UpdateInProgress')) {
+      return 'transitional';
+    }
+
+    // The CRD defaults replicas to 1, so an object read from a cluster always
+    // has it. Mirror that default rather than falling back to 0, which would
+    // otherwise take the scale-to-zero branch below and call a partially
+    // constructed leader worker set healthy.
+    const desired = this.spec?.replicas ?? 1;
```
kannon92: AI review:

2. Cleanup — frontend/src/components/leaderworkerset/List.tsx:49: the list's Ready column defaults spec.replicas to 0, but getHealth() in leaderWorkerSet.ts defaults it to 1. A resource with missing spec.replicas shows "0/0" (looks fine) in the list while being flagged failed in the Overview health chart.

frontend/src/components/statefulset/List.tsx
```diff
@@ -15,12 +15,43 @@
...
 import { KubeContainer } from '../../lib/k8s/cluster';
 import StatefulSet from '../../lib/k8s/statefulSet';
+import { useNamespaces } from '../../redux/filterSlice';
+import { CreateResourceButton } from '../common';
 import ResourceListView from '../common/Resource/ResourceListView';
+import { SimpleTableProps } from '../common/SimpleTable';
 import LightTooltip from '../common/Tooltip/TooltipLight';
 
 export default function StatefulSetList() {
+  const { items: statefulSets, errors } = StatefulSet.useList({ namespace: useNamespaces() });
```
kannon92: My LLM is flagging this as a correctness issue.

1. Correctness (most severe) — frontend/src/components/statefulset/List.tsx:28: StatefulSetList now fetches its own data and passes it into ResourceListView instead of using resourceClass, which silently bypasses ResourceTable's plugin LIST_VIEW event dispatch for the Stateful Sets page. Any plugin listening for that event stops firing.

kannon92: More details:

The two code paths in ResourceTable (frontend/src/components/common/Resource/ResourceTable.tsx:174-186):

```
export default function ResourceTable(props) {
  if (!!props.resourceClass) {
    return <TableFromResourceClass resourceClass={resourceClass!} {...otherProps} />;
  }
  return <ResourceTableContent {...props} />;
}
```

- Pass resourceClass → goes through TableFromResourceClass, which calls resourceClass.useList() itself and dispatches a HeadlampEventType.LIST_VIEW event every time the items/errors change (lines 193-212).
- Pass data directly → goes straight to ResourceTableContent, skipping TableFromResourceClass entirely — no event dispatch, ever.

What changed: before this PR, frontend/src/components/statefulset/List.tsx passed resourceClass={StatefulSet} into ResourceListView, so it took the event-dispatching path. The PR (commit 2c94cee4c) refactored it to support reuse from the LeaderWorkerSet details page — it now does:

```
const { items: statefulSets, errors } = StatefulSet.useList({ namespace: useNamespaces() });
return <StatefulSetsListRenderer statefulSets={statefulSets} errors={errors} />;
// ...
<ResourceListView data={statefulSets} errors={errors} ... />
```

That's the data path — the plain top-level /statefulsets page lost the LIST_VIEW event dispatch as a side effect of the refactor.

Why it matters: HeadlampEventType.LIST_VIEW is a documented plugin hook (ResourceListViewLoadedEvent) — plugins register listeners for it to react to a list page loading/erroring (e.g. show a badge, log analytics, surface a custom error banner). Any plugin relying on that event for Stateful Sets specifically would silently stop receiving it once a user visits /statefulsets, with no error or warning — it just never fires.

Fix shape: StatefulSetsListRenderer needs to still pass resourceClass={StatefulSet} for the plain-list case (default export), only switching to the data prop when it's being reused from the LWS details page with pre-fetched items. Something like accepting resourceClass as an optional override, or giving TableFromResourceClass's event-dispatch logic to callers that supply their own data. Want me to implement that fix?

illume (changes requested):
Thanks for the contribution.


<details>

</details>


frontend/src/components/leaderworkerset/List.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return selected.type ?? '-';
+}
+
+function getReadyGroups(leaderWorkerSet: LeaderWorkerSet): string {
+  const ready = leaderWorkerSet.status?.readyReplicas ?? 0;
+  const desired = leaderWorkerSet.spec?.replicas ?? 0;
+
+  return `${ready}/${desired}`;
+}
```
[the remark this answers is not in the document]
kannon92: Same as https://github.com/kubernetes-sigs/headlamp/pull/6744#discussion_r3716003674.

adityachawla005:
Thanks for the review, all of these should be covered now.

On the duplication: pulled the condition handling out into `frontend/src/lib/k8s/conditions.ts`, two functions there.

- `isConditionTrue(conditions, type)` replaces the little `isTrue` closure that Job, JobSet and LeaderWorkerSet were each declaring inside their own `getHealth()`.
- `getTopCondition(conditions, priority)` replaces the copy-pasted picker in the JobSet and LeaderWorkerSet list columns. It returns `undefined` rather than `'-'` so the display fallback stays in the component instead of being baked into the helper.

Added `conditions.test.ts` alongside it. Included the ordering case specifically, since that's the whole reason the priority list exists in the first place: `status.conditions` has no guaranteed order, so reversing the input shouldn't change what the column ends up showing.

On the `0/0` mismatch, good catch, that one was real. LeaderWorkerSet now has `getDesiredReplicas()` and `getReadyReplicas()`, and both the Ready column and `getHealth()` go through them, so there's a single place defaulting `spec.replicas` to the CRD's 1. That also covers the Copilot comment on `getReadyGroups`.

On the LIST_VIEW one, your read was right, and thanks for actually tracing it through `ResourceTable`. `StatefulSetsListRenderer` now only takes the `data` path when a caller hands it pre-fetched items, which is what the LWS details page does. With the prop unset the Stateful Sets page passes `resourceClass={StatefulSet}` like it did before, so it stays on `TableFromResourceClass` and the event keeps firing. One nice side effect: the page no longer needs its own `useNamespaces()` call, since `TableFromResourceClass` already reads the namespace filter itself.

Also fixed a rebase casualty while I was in there. The statefulset -> leaderworkerset map relation was missing the id that `makeRelation` started requiring after the relation provider API landed.

frontend/src/lib/k8s/leaderWorkerSet.ts
adityachawla005: Moved to `lib/k8s/conditions.ts` as `isConditionTrue`, now shared by Job, JobSet and LeaderWorkerSet, with tests in `conditions.test.ts`.

frontend/src/components/leaderworkerset/List.tsx
adityachawla005: Both are gone, replaced by `getTopCondition` in `lib/k8s/conditions.ts`. It returns `undefined` rather than `'-' so each list keeps its own display fallback.

frontend/src/lib/k8s/leaderWorkerSet.ts
adityachawla005: Fixed, `getDesiredReplicas()` is the one place the CRD default of 1 lives now, and both the Ready column and `getHealth()` read through it.

frontend/src/components/statefulset/List.tsx
adityachawla005: Good catch. The page passes `resourceClass={StatefulSet}` again so the event still fires, and only the LWS details page takes the `data` path with pre-fetched items.

frontend/src/components/leaderworkerset/List.tsx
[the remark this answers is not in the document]
adityachawla005: Fixed via `getDesiredReplicas()`, which defaults to the CRD's 1 and is shared with `getHealth()` so the list and the overview agree.

kannon92:
LGTM

/assign @illume

illume (approved):
🎉 thanks!

Issue #4051, opened 2025-10-16
Improve UX for non-cluster-admins (


siegenthalerroger:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
Many views only function if either cluster-scoped view permissions exist or a subset of namespaces are selected. This means that developers often tend to believe that they do not have access to certain information/resources, limiting the visibility and causing unnecessary noise for cluster operators.

## Describe the solution you'd like
In cases where the permissions of a user are constrained, cluster-scoped requests should no longer be made (as these will fail) and they should be replaced by requests with a list of allowed namespaces. Currently this is possible by manually selecting the namespaces that the the user has access to with the namespace dropdown.

This list of namespaces should not have to be manually defined by the user, but rather the namespace selector should automatically select the namespaces that are visible to the user. Alternatively a "select all" button would already do a lot of the work.

A minimal solution would be a config option where one could statically set the accessible namespaces per cluster. This is suboptimal as any newly added namespaces would require manual steps from each user (and on every device as well).

## What users will benefit from this feature?
Any user that doesn't have cluster-wide read permissions.

## Are you able to implement this feature?
Sadly no

## Additional context
Currently we allow all users to see all available namespaces, regardless of their actual capability to view or do anything within the namespaces. If the above suggestion was implemented we would change this to be limited to the relevant namespaces with the `resourceNames` property.

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: view-namespaces
  
rules:
  - apiGroups: [""]
    resources: ["namespaces"]
    verbs: ["get", "watch", "list"]
    # resourceNames: ["namespace-a", "namespace-c"]
```

sniok:
Hi, thanks for opening an issue
Headlamp has a setting called "allowed namespaces" which should help with this problem

<img width="1259" height="539" alt="Image" src="https://github.com/user-attachments/assets/41fd1830-05f6-40fe-ab73-01b804302453" />

but as you've mentioned this is not optimal and requires action from each cluster user.

The problem is essentially that there's no way for user to know which namespaces they have access to.
We're thinking of having a ConfigMap in clusters for headlamp settings (allowed namespaces and others), there's an issue for that [here](https://github.com/kubernetes-sigs/headlamp/issues/3979). 
What do you think?

siegenthalerroger:
[...]

PR #6759, opened 2026-07-30
frontend: Add allowed namespaces label selector to cluster settings

guillaumebernard84:
## Summary

At the company I work for, users from several teams share clusters and only have access to namespaces carrying a certain label. If they don't select a namespace explicitly in Headlamp, they have an error when browsing pods, deployments, etc... 
The workaround in  #4051 is to manually maintain an allowed namespaces list to avoid unauthorized errors when browsing without a selected namespace. That is not convenient.

This change adds a cluster setting to define a namespace label selector. Namespaces matching the selector are fetched from the API, cached, and combined with the manually configured allowed namespaces list. The list is refreshed when the cluster loads, so newly created namespaces matching the selector are picked up automatically.

## Changes

- Added: `allowedNamespacesSelector` cluster setting with UI in cluster settings
  (text field + chips showing the resolved namespaces)
- Added: `lib/k8s/allowedNamespaces.ts` to resolve the selector via
  `/api/v1/namespaces?labelSelector=...` and cache the result
- Updated: all readers of `allowedNamespaces` now use the combined
  (manual + resolved) list
- Updated: resolved namespaces are refreshed on cluster load (`Layout.tsx`)

## Steps to Test

1. Use a kubeconfig with a user that only has access to namespaces carrying a given label (RBAC must allow listing namespaces).
2. Go to Settings → Cluster and set the label selector (e.g. `team=frontend`), then press Enter.
3. The matching namespaces appear as chips.
4. Browse resource lists (Pods, Deployments, …) without selecting a namespace: resources from the allowed namespaces are listed without unauthorized errors.
5. Create a new namespace with the label, reload the page: it is picked up automatically.

## Screenshots

image

## Notes for the Reviewer

Resolving the selector requires the user to be able to list namespaces with that selector. The resolved list is cached in localStorage next to the existing cluster settings.


illume (changes requested):
Thanks for this PR.


frontend/src/lib/k8s/allowedNamespaces.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * This is useful for users who don't have the authority to list all namespaces
+ * but do have access to the namespaces carrying a certain label; the API server
+ * still needs to allow listing namespaces with that selector.
+ *
+ * @param cluster - The cluster to fetch the namespaces from.
+ * @param selector - A Kubernetes label selector (e.g. "team=frontend,env=prod").
+ * @returns The sorted list of namespace names matching the selector.
+ */
+export async function fetchNamespacesBySelector(
```
unlikelyzero: This reimplements what `Namespace.useList({ clusters, labelSelector })` already does elsewhere in the codebase (see `ProjectList.tsx`) - `KubeObject.useList` forwards `labelSelector` through `QueryParameters` natively. Going through that instead would get you react-query caching, retry, and websocket watch for free instead of a hand-rolled fetch with none of that.

frontend/src/lib/k8s/allowedNamespaces.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @returns The resolved namespace names, or null if no selector is configured.
+ */
+export async function syncAllowedNamespacesFromSelector(cluster: string): Promise<string[] | null> {
+  if (!cluster) {
+    return null;
+  }
+
+  const selector = loadClusterSettings(cluster).allowedNamespacesSelector?.trim();
+  if (!selector) {
+    // Clear any stale cache left from a previously configured selector.
```
unlikelyzero: This clears the cache when the selector is cleared, which is good, but there's no equivalent clear on a *failed* resolution - if the fetch throws (network error, 403, malformed selector), the previous selector's namespace list just stays cached with nothing but a console.debug in Layout.tsx marking the failure. Might be worth failing closed here the same way the empty-selector path does.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -130,6 +136,50 @@ export default function SettingsCluster() {
...
+  const allowedNamespacesSelector = clusterSettings.allowedNamespacesSelector || '';
+  React.useEffect(() => {
+    setNamespacesSelectorInput(allowedNamespacesSelector);
+    setNamespacesSelectorError('');
+  }, [allowedNamespacesSelector, cluster]);
+
+  function applyNamespacesSelector(selector: string) {
+    const trimmedSelector = selector.trim();
+    setNamespacesSelectorError('');
+    setClusterSettings(settings => {
```
unlikelyzero, line 175: The selector is persisted before the fetch resolves, and on failure only the UI error gets set - `allowedNamespacesFromSelector` is left untouched. Combined with the resolution logic not being tagged to a specific selector value, two edits in quick succession (or this racing the on-load sync in Layout.tsx) can leave the persisted selector and the cached namespace list disagreeing indefinitely.

frontend/src/components/App/Settings/SettingsCluster.tsx
```diff
@@ -435,6 +486,78 @@ export default function SettingsCluster() {
...
+                    error={!!namespacesSelectorError}
+                    value={namespacesSelectorInput}
+                    helperText={
+                      namespacesSelectorError ||
+                      t(
+                        'translation|A label selector used to automatically find the namespaces you are allowed to access in this cluster. Press Enter or the apply button to resolve it. The resolved namespaces are combined with the allowed namespaces list above.'
+                      )
+                    }
+                    autoComplete="off"
+                    inputProps={{
```
unlikelyzero, line 514: `inputProps` sets `form` to an object (`{ autocomplete: 'off' }`) instead of a form-id string, which renders as the invalid `form="[object Object]"` attribute and doesn't actually do anything - `autoComplete="off"` on the TextField already covers the intent. This is one of the two Copilot comments from earlier in the review that's still open.

frontend/src/helpers/clusterSettings.ts
```diff
@@ -22,6 +22,16 @@ export interface ClusterSettings {
   defaultNamespace?: string;
   /** Only allow namespaces in this list to be selected */
   allowedNamespaces?: string[];
+  /**
```
unlikelyzero, line 25: This persists a live-API-derived value indefinitely in the same localStorage blob as genuine user config, with no timestamp or TTL to signal staleness. Elsewhere in this codebase that kind of derived data goes through react-query with a staleTime (see the SelfSubjectReview cache in TopBar.tsx) rather than localStorage - might be worth following that pattern here too, which would also sidestep the race with useClusterSettings mentioned above.

unlikelyzero (commented):
The two-source approach makes sense, but there are a few issues here.

The biggest one is fail-open behavior: `getCombinedAllowedNamespaces` returns `[]` both for "no restriction" and "selector matched nothing." `useKubeObjectList.ts` treats `[]` as an unscoped cluster-wide request, so a selector that matches nothing can fall back to unrestricted access instead of no results.

There are also a couple of cache races. Older selector requests can overwrite newer results, failed resolution leaves stale namespaces cached, and the Settings page can overwrite background-sync changes because it holds its own localStorage snapshot.

One other thing worth checking: Kubernetes RBAC doesn't scope `list namespaces` by label selector. If a user can't list namespaces cluster-wide, they generally won't be able to resolve this selector either. That seems different from the use case described in the PR.

illume (changes requested):
Thanks for these changes.


<details>

</details>

frontend/src/lib/k8s/allowedNamespaces.ts
guillaumebernard84: You are right, I use useList now

frontend/src/components/App/Settings/SettingsCluster.tsx
guillaumebernard84: Fixed

frontend/src/lib/k8s/allowedNamespaces.ts
guillaumebernard84: You are right, I fixed that too

frontend/src/helpers/clusterSettings.ts
guillaumebernard84: Implemented TTL

frontend/src/components/App/Settings/SettingsCluster.tsx
guillaumebernard84: fixed

guillaumebernard84:
I rebased, implemented the reviewers feedback, and added the i18n strings for the new language (tests were failing)

illume (changes requested):
Thanks for the contribution.


frontend/src/helpers/clusterSettings.ts
```diff
@@ -81,3 +86,141 @@ export function loadClusterSettings(clusterName: string): ClusterSettings {
...
+ *
+ * @param cluster - The name of the cluster.
+ * @returns The combined list of allowed namespaces.
+ */
+export function getCombinedAllowedNamespaces(cluster: string): string[] {
+  const settings = loadClusterSettings(cluster);
+  const selector = settings.allowedNamespacesSelector?.trim() || '';
+  const cache = selector ? loadResolvedAllowedNamespaces(cluster) : null;
+  const resolved = cache && cache.selector === selector ? cache.namespaces : [];
+  return [...new Set([...(settings.allowedNamespaces || []), ...resolved])].sort();
```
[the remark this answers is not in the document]
guillaumebernard84: No namespace selected resolves to "all namespaces" in the "allowed namespace" box. I kept this behavior. If no namespaces are retrieved with the label selector, it stays empty and all namespaces are selected (then it's merged with the   "allowed namespace" list, so the final list may be empty or not)

To prevent any ambiguity, I added an warning message if the selector returns no label so the user knows it "failed"

frontend/src/components/App/Layout.tsx
```diff
@@ -235,6 +237,13 @@ export default function Layout({}: LayoutProps) {
     }
   }, [cluster, dispatch]);
 
+  // Keep the namespaces resolved from the configured label selector (if any) in
+  // sync for the active cluster so getCombinedAllowedNamespaces stays up to date.
+  useAllowedNamespacesFromSelector(
+    cluster || '',
+    loadClusterSettings(cluster || '').allowedNamespacesSelector
+  );
```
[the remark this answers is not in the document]
guillaumebernard84: fixed

frontend/src/components/App/Layout.tsx
```diff
@@ -301,6 +318,9 @@ export default function Layout({}: LayoutProps) {
       >
         {t('Skip to main content')}
       </Link>
+      {selectedClusters.map(clusterName => (
+        <AllowedNamespacesSelectorSync key={clusterName} cluster={clusterName} />
+      ))}
```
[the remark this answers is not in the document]
guillaumebernard84: Fixed with delayed rendering

illume (approved):
🎉 thanks!

PR #6761, opened 2026-07-30
backend: health: Add /healthz and /readyz endpoints

### LINKED ISSUE #6760: Add /healthz and /readyz health check endpoints to the backend server
## Problem Description

The Headlamp backend server has no dedicated health or readiness endpoints. The Helm chart currently uses `GET /` (the SPA index page) as both the liveness and readiness probe target. This only verifies that the HTTP server is responding, not that it is actually ready to serve traffic. During graceful shutdown, the server continues to pass both probes until the process exits, which means the load balancer keeps routing new traffic to a dying instance.

## Solution

Add two Kubernetes-standard health check endpoints:

- **`GET /healthz`** (liveness): Returns `200 OK` with `{"status":"ok"}` as long as the HTTP server can process requests.

- **`GET /readyz`** (readiness): Returns `200 OK` with `{"status":"ok"}` when the server is ready to accept traffic. Returns `503 Service Unavailable` with error details when the server has  received a SIGINT/SIGTERM and is draining connections.

Both endpoints should bypass authentication and telemetry middleware so they remain fast and always accessible.

Production operators deploying Headlamp in-cluster via the Helm chart or Kubernetes manifests. Once these endpoints exist, the Helm chart probe paths can be updated from `/` to `/healthz` and `/readyz` for proper health checking. This also benefits any reverse proxy or service mesh that relies on standard health endpoints.

## Additional context

The Helm chart (`charts/headlamp/templates/deployment.yaml`) already has configurable `livenessProbe` and `readinessProbe` sections pointing at `/`. Updating those paths to `/healthz` and `/readyz` would be a one-line follow-up change in the chart. The backend already has a `setupGracefulShutdown` function that handles SIGINT/SIGTERM, making it a natural place to wire the readiness state change.

Slambot01:
## Summary

This PR adds Kubernetes-standard `/healthz` (liveness) and `/readyz` (readiness) health check endpoints to the Headlamp backend server. The Helm chart currently probes `GET /` (the SPA page) for both liveness and readiness, which only verifies the HTTP server is alive. These new endpoints provide proper health semantics, with `/readyz` returning 503 during graceful shutdown so the load balancer stops routing traffic before the process exits.

## Related Issue

Fixes #6760 

## Changes

- Added new `backend/pkg/health` package with `Checker` struct using `sync/atomic.Bool` for lock-free shutdown state tracking
- Added `HandlerHealthz()` returning `200 OK` with `{"status":"ok"}`
- Added `HandlerReadyz()` returning `200 OK` when ready, `503` with error details when shutting down
- Registered both endpoints on the top-level router in `StartHeadlampServer` (bypasses auth and telemetry middleware)
- Wired `SetShuttingDown()` into `setupGracefulShutdown` so `/readyz` returns 503 on SIGINT/SIGTERM
- Added comprehensive tests (5 test cases covering both endpoints and shutdown behavior)

## Steps to Test

1. Build and start the backend: `cd backend && go build ./cmd/ && ./cmd --in-cluster=false`
2. `curl http://localhost:4466/healthz` -- should return `{"status":"ok"}` with HTTP 200
3. `curl http://localhost:4466/readyz` -- should return `{"status":"ok"}` with HTTP 200
4. Send SIGTERM to the server process
5. Immediately `curl http://localhost:4466/readyz` -- should return HTTP 503

## Notes for the Reviewer

- Both endpoints bypass auth/telemetry middleware by being registered on the top-level mux router before the handler chain. This is intentional so health checks remain fast and unauthenticated.
- The `Checker` struct uses `sync/atomic.Bool` (no mutex needed) for the shutdown flag.
- The Helm chart update (changing probe paths from `/` to `/healthz` and `/readyz`) is left as a follow-up to keep this PR focused on the backend implementation.
- `runServer` and `setupGracefulShutdown` signatures now accept a `*health.Checker` parameter.This is a private API change with no external consumers.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -52,6 +52,7 @@ import (
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/clusterinventory"
 	cfg "github.com/kubernetes-sigs/headlamp/backend/pkg/config"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/headlampconfig"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/health"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/helm"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
@@ -1398,6 +1399,14 @@ func StartHeadlampServer(config *HeadlampConfig) {
 
 	router := mux.NewRouter()
 
+	// Health and readiness endpoints are registered on the top-level
+	// router so they bypass auth and telemetry middleware.
+	healthChecker := health.NewChecker()
+	router.HandleFunc("/healthz", healthChecker.HandlerHealthz()).Methods(http.MethodGet)
+	router.HandleFunc("/readyz", healthChecker.HandlerReadyz()).Methods(http.MethodGet)
+
+	logger.Log(logger.LevelInfo, nil, nil, "health endpoints registered: /healthz, /readyz")
+
 	if config.Telemetry != nil && config.Metrics != nil {
 		router.Use(telemetry.TracingMiddleware("headlamp-server"))
 		router.Use(config.Metrics.RequestCounterMiddleware)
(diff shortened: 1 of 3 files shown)


Slambot01:
cc @illume

PR #6767, opened 2026-07-31
frontend: Tabs: Respect defaultIndex on initial render

### LINKED ISSUE #6409: Tabs component ignores defaultIndex prop always initialises at tab 0
## Describe the bug
The `Tabs` component in `frontend/src/components/common/Tabs.tsx` always renders with the **first tab selected**, regardless of what value is passed to the `defaultIndex` prop.

In `Tabs.tsx` at the `useState` initialiser (line 68), the code reads:

```ts
defaultIndex && Math.min(defaultIndex as number, 0)
```

This contains **two compounding bugs:**

1. **`Math.min(defaultIndex, 0)` always returns `0`** (or a negative value) — it clamps to the *smaller* of `defaultIndex` and `0`, so any positive index is discarded.
2. **The `&&` short-circuit is incorrect** — `0 && ...` is falsy in JavaScript, so `defaultIndex=0` returned `0` only by accident, masking the bug for the default case.

The intended expression was `Math.max(defaultIndex, 0)` — clamp to non-negative.

Harsh63870:
## Summary
`Tabs` was ignoring `defaultIndex` on first render — it always opened on tab 0 no matter what you passed in. Turned out the initial state was doing `Math.min(defaultIndex, 0)` instead of `Math.max(...)`, so anything >= 0 just got clamped down to 0. Storybook's "Starting Tab" story was a good example of this being broken.

## Related Issue
Fixes #6409

## Changes
- `frontend/src/components/common/Tabs.tsx`: swapped `Math.min(defaultIndex, 0)` → `Math.max(defaultIndex, 0)` for numeric values. Left `false`/`null` handling alone so disabling initial selection still works.
- Added tests in `Tabs.test.tsx` for:
  - `defaultIndex={2}` actually opening on the 3rd tab
  - default behavior (no `defaultIndex`) still starting on tab 0
  - making sure `onTabChanged` didn't break from this change

## Steps to Test
```bash
cd frontend
npm install
npm test -- --run src/components/common/Tabs.test.tsx
npm run lint
```
Also checked the "Starting Tab" story in Storybook manually to confirm it opens on the right tab now.

### DIFF
--- a/frontend/src/components/common/Tabs.test.tsx
+++ b/frontend/src/components/common/Tabs.test.tsx
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import Tabs from './Tabs';
+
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this PR.


Harsh63870:
@illume, resolved all the comments..

illume:
Hi, sorry. In the issue I noticed there was already a PR open for this one.

PR #6779, opened 2026-07-31
frontend: useLocalStorageState: Fix react-hooks/exhaustive-deps

### LINKED ISSUE #6777: frontend: useLocalStorageState: Fix react-hooks/exhaustive-deps — missing dependency. File: frontend/src/components/globalSearch/useLocalStorageState.tsx
Sub-issue of #5183 

> frontend: useLocalStorageState: Fix react-hooks/exhaustive-deps — missing dependency. File: frontend/src/components/globalSearch/useLocalStorageState.tsx

southball:
## Summary

This PR rewrites the `useLocalStorageState` to simplify the behavior and resolve lint errors.

## Related Issue

Fixes #6777 (part of #5183)  

## Changes

* Simplify the behavior: the original hook has non-inter-hook-synced `useLocalStorageState(...)[1]` and synced `useLocalStroageState.update`. In then new version, all 
* Enable batched updates to the same value
* Use `useSyncExternalStore`
* Fix the suppressed lint error below (#6777 part of #5183)
  > frontend: useLocalStorageState: Fix react-hooks/exhaustive-deps — missing dependency. File: frontend/src/components/globalSearch/useLocalStorageState.tsx

## Steps to Test

Visiting all pages that use the hook and ensure there are no regressions.

No UI changes

## Notes for the Reviewer

N/A

### DIFF
--- a/frontend/src/components/globalSearch/useLocalStorageState.tsx
+++ b/frontend/src/components/globalSearch/useLocalStorageState.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 /** Store listeners to allow updates outside of the hook */
 const updateListeners: Record<string, Array<(newValue: any) => void>> = {};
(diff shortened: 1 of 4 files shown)


frontend/src/components/globalSearch/useLocalStorageState.tsx
```diff
@@ -14,90 +14,122 @@
...
+    listener();
+  }
+
+  writeStoredValue(key, newValue);
 }
 
 /**
- * Update the value in local storage and notify all `useLocalStorageState` hooks
+ * Custom hook to manage state synchronized with localStorage and other hook instances
+ * in the same window. Values must be serializable to JSON.
```
southball, line 101: > `in the same window`

I wonder if there is appetite for making this also handle cross-window sync using `window.addEventListener("storage", ...)`. I did not make this change since that would be a bit too much behavior change, and also it could be easily added in another PR.

(In addition, we might want to think about what states should be synced cross-window and what should not, e.g. do we want to actually sync the table sort column, or do we want to have a global setting for default...?)

illume: I'm not sure about this. But feel free to open an issue if you want. To see if anyone is interested in this.

illume (approved):
🎉 thanks!

PR #6787, opened 2026-07-31
frontend: Table: Preserve scroll position across auto refresh

### LINKED ISSUE #5701: frontend: Resource tables lose scroll position after auto refresh

WasThatRudy:
Fixes #5701

## Summary
- Resource tables using React Query polling receive a fresh array on every refresh, which caused the MRT scroll container to jump back to the top even when the user had scrolled down.
- Adds a small hook `useScrollPreservationOnDataChange` that captures `scrollTop` via `onScroll` into a ref and, in a layout effect keyed on the data reference, restores that value when the container has just been reset to 0.
- Wires the hook into `muiTableContainerProps` on the `useMaterialReactTable` call in `Table.tsx`.
- Clears the saved position when the container becomes unscrollable (data went empty) so a later refill does not restore a position that no longer maps to the new dataset.
- Seeds the saved position from a non-zero mounted `scrollTop` before any scroll event fires, so hydration or deep links that land the user mid-list are preserved on the first refresh too.

## Test plan
- [x] `npx vitest run src/components/common/Table src/components/common/Resource` (94 tests pass, includes 5 new tests for the hook: restore after reset, does not fight scroll-to-top, empty then refill drop, mount seed then restore, ref/onScroll identity stability)
- [x] `npm run tsc` (clean)
- [ ] Manual: open any resource list with more rows than fit on screen, scroll partway down, wait for the auto refresh interval, confirm the scroll position is preserved
- [ ] Manual: on the same list, scroll partway down then change the search filter, confirm the container still shows the filtered results near where the user was (not fought back to a stale offset)
- [ ] Manual: on a list that goes empty and then repopulates (e.g. namespace switch), confirm the new list starts at the top

### DIFF
--- a/frontend/src/components/common/Table/Table.tsx
+++ b/frontend/src/components/common/Table/Table.tsx
@@ -59,6 +59,7 @@ import { useSettings } from '../../App/Settings/hook';
 import { useQueryParamsState } from '../../resourceMap/useQueryParamsState';
 import Empty from '../EmptyContent';
 import Loader from '../Loader';
+import { useScrollPreservationOnDataChange } from './useScrollPreservation';
 
 /**
  * Column definition
@@ -298,6 +299,11 @@ export default function Table<RowItem extends Record<string, any>>({
     return (tableProps.data ?? []).filter(it => filterFunction(it));
   }, [tableProps.data, filterFunction]);
 
+  // Preserve scroll position across auto-refresh re-renders where React Query
+  // hands us a fresh array reference (kubernetes-sigs/headlamp#5701).
+  const { ref: scrollContainerRef, onScroll: onScrollContainer } =
+    useScrollPreservationOnDataChange(tableData);
+
   const paginationSelectProps = import.meta.env.UNDER_TEST
     ? {
         inputProps: {
@@ -458,6 +464,10 @@ export default function Table<RowItem extends Record<string, any>>({
       showFirstButton: false,
       showLastButton: false,
       SelectProps: paginationSelectProps,
+    },
+    muiTableContainerProps: {
+      ref: scrollContainerRef,
+      onScroll: onScrollContainer,
     },
     muiTableBodyCellProps: {
       sx: {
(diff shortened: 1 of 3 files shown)


illume (approved):
🎉 thanks!

frontend/src/components/common/Table/Table.tsx
```diff
@@ -459,6 +465,10 @@ export default function Table<RowItem extends Record<string, any>>({
       showLastButton: false,
       SelectProps: paginationSelectProps,
     },
+    muiTableContainerProps: {
```
sniok, line 468: this PR doesn't make much sense to me. the table container is never the container that gets scrolled, it's the page container that has an overflow and a scrollbar. the useScrollPreservationOnDataChange hook doesn't do anything since it returns early. I've debugged this on a local env and the scrollContainerRef never even gets populated, so the hook never executes at all. given all that I'm suspecting that this PR is partly or completely generated by LLM

@illume have you tested this at all PR before merging?

WasThatRudy: You're right, sorry. I just checked and the ref never actually attaches, MRT in our setup doesn't render a TableContainer element, so it stays null and the hook just goes on the if (!el) return. The scrolling element is a page level ancestor, not the one I was targeting.
I'll send a revert now and take another look at whether there's a real fix that hooks into the actual scroll container.
Revert: #7076

PR #6792, opened 2026-08-01
frontend: DeleteButton: Offer both delete and evict for pods

### LINKED ISSUE #6764: Instead of having a toggle in the settings between delete/evict, just have both options available at the same time
I sometimes need to evict a pod, and sometimes to straight out delete it. e.g. It's stuck and not working, but it has a PDB that prevents evict, so I want to just delete it so it gets recreated.

Why not just have both options available at the same time? It's cumbersome to have to go to the settings to toggle between them when they both have their uses.

Any

[Yes (I will propose a PR) / No.]
> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context

Image

krsatyamthakur-droid:
## Summary

Deleting and evicting a pod both have their uses (for example evict respects a PodDisruptionBudget while delete does not), but the button only exposed one at a time, chosen by the global useEvict setting, so switching meant a trip to Settings.

For pods this now shows both as a split button: the primary action follows the useEvict setting and the other option is in the dropdown. Inside a table row's overflow menu, where a split button can't nest, both are offered as separate menu items. Non-pod resources are unchanged, and useEvict just selects the default action instead of hiding one.

## Related Issue

Fixes #6764

## Changes

- Updated DeleteButton.tsx: for pods, render a split button (primary action + dropdown with Delete/Evict) in the action style, and two menu items in the overflow-menu style. The confirm dialog and callback now follow a per-action state instead of the global useEvict setting
- Non-pod resources keep the single Delete button; useEvict now selects the default pod action rather than hiding one
- Added tests for the pod split button and updated the affected storyshots
- Updated the DeleteButton PodEvict story description

## Steps to Test

1. Open a Pod's details page and confirm the delete action shows a split button, with the other option (Delete/Evict) in the dropdown
2. Choose each and confirm the right dialog appears (Delete shows the Force Delete option, Evict doesn't) and performs the right action
3. Open the Pods list, use a row's overflow (...) menu, and confirm both Delete and Evict appear as separate items
4. Confirm non-pod resources still show a single Delete button, and that the useEvict setting flips which action is the default

Pod delete action now offers both Delete and Evict as a split button:

Screenshot 2026-08-01 at 2 38 42 PM

Shown in a resource details header:
Screenshot 2026-08-01 at 2 39 01 PM

## Notes for the Reviewer

- Discussed the split-button approach with @illume on Slack before building
- The overflow-menu context can't nest a split button, so it shows both as separate menu items instead. Non-pod behaviour is untouched


illume (changes requested):
Thanks for these changes.


frontend/src/components/common/Resource/DeleteButton.test.tsx
```diff
@@ -122,6 +131,25 @@ describe('DeleteButton', () => {
...
+  it('opens the evict confirmation when Evict is chosen for a pod', async () => {
+    renderButton(makePod({ name: 'my-pod' }));
+
+    fireEvent.click(await screen.findByLabelText('translation|More delete options'));
+    fireEvent.click(await screen.findByRole('menuitem', { name: 'translation|Evict' }));
+
+    const dialog = await screen.findByRole('dialog');
+    expect(within(dialog).getByText('translation|Evict Pod')).toBeInTheDocument();
+  });
+
```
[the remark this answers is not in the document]
krsatyamthakur-droid: Good catch, added a test that renders with buttonStyle="menu" and checks both Delete and Evict show up as separate menu items.

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -59,14 +61,22 @@ export default function DeleteButton(props: DeleteButtonProps) {
...
       if (!item) {
         return;
       }
 
       let callback = item!.delete;
-      if (settingsObj.useEvict && item.kind === 'Pod') {
+      if (action === 'evict' && item.kind === 'Pod') {
         const pod = item as Pod;
         callback = pod.evict;
       }
```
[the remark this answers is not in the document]
krsatyamthakur-droid: Fixed. The messages now switch to eviction wording (Evicting/Evicted/Cancelled eviction/Error evicting) when evict is the selected action, and stay as delete wording otherwise.

krsatyamthakur-droid:
hii @illume ,Addressed both comments — evict now uses eviction-worded snackbar messages instead of delete wording, and added a test for the overflow-menu branch. Resolved both threads. ptal

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -109,39 +138,74 @@ export default function DeleteButton(props: DeleteButtonProps) {
...
+            onClick={() => openDialogFor(defaultAction)}
+            icon="mdi:delete"
+          />
+          <ActionButton
+            description={t('translation|More delete options')}
+            buttonStyle={buttonStyle}
+            onClick={(event: React.MouseEvent<HTMLElement>) => setMenuAnchorEl(event.currentTarget)}
+            icon="mdi:menu-down"
+          />
+          <Menu
```
unlikelyzero: This still shows both Delete and Evict in the dropdown rather than just the alternate action - that's Copilot's still-open comment from the last round. Worth taking seriously given useEvict defaults to true: with the default config, the primary button is Evict and the dropdown right next to it lists Delete first, with no styling difference to indicate it's the destructive one. Filtering the menu to the non-default action (matching what the PR description and the PodEvict story text both say) would fix both the spec mismatch and the accidental-hard-delete risk.

frontend/src/components/common/Resource/DeleteButton.tsx
```diff
@@ -109,39 +138,74 @@ export default function DeleteButton(props: DeleteButtonProps) {
...
+        // Split button: the default action from the useEvict setting, with the
+        // other option available in the dropdown.
+        <>
+          <ActionButton
+            description={
+              defaultAction === 'evict' ? t('translation|Evict') : t('translation|Delete')
+            }
+            buttonStyle={buttonStyle}
+            onClick={() => openDialogFor(defaultAction)}
+            icon="mdi:delete"
```
unlikelyzero: The dropdown trigger only has an `aria-label` - no `aria-haspopup`, `aria-expanded`, or `aria-controls` linking it to the menu. This is new surface area from this PR, so worth adding those rather than leaving a screen-reader user with no indication the button opens a menu.

frontend/src/components/common/Resource/DeleteButton.test.tsx
```diff
@@ -122,6 +131,36 @@ describe('DeleteButton', () => {
     expect(container).toBeEmptyDOMElement();
   });
 
+  it('offers both delete and evict for a pod', async () => {
```
unlikelyzero: This test asserts both Delete and Evict are present in the dropdown, which locks in the exact behavior Copilot's open comment is asking to change. Once the dropdown filtering gets fixed this test will need to change with it - just flagging so it doesn't read as evidence the current both-options behavior is correct.

unlikelyzero (commented):
This is a good feature and worth having, but there's an RBAC bug that needs fixing before this merges, plus a couple of things I'd want settled on the design.

The one I'd flag as most serious: the whole control - both Delete and Evict - is gated by a single `AuthVisible authVerb="delete"`. Eviction is actually a `create` against the `pods/eviction` subresource, not a pod delete. `AuthVisible` already supports a `subresource` prop and this exact pattern is already used elsewhere in the codebase for other pod subresources (exec, logs, scale), so it's a straightforward fix. As written, a user with delete-only RBAC sees an Evict option that 403s after they confirm it, and a user with eviction-only RBAC (a pretty common least-privilege setup - create on pods/eviction but not delete on pods) never sees the control at all.

That RBAC gap is really a symptom of the component doing two structurally different things internally rather than being two separately-gated components, which is the pattern every other action button in that directory follows (Restart, Scale, Edit, Rollback all render one ActionButton with its own AuthVisible). Splitting Delete and Evict into two components each with the right verb would fix the RBAC issue and make the two actions' different confirm copy and API calls less awkward to maintain together.

Separately, `Pod.evict()` calls `post(url, body)` with no cluster option, while `KubeObject.delete()` explicitly threads `this._clusterName` through. In a multi-cluster view - `ProjectResourcesTab` renders this exact button per row alongside a cluster column - evicting a pod could end up posting to the wrong cluster if there's a same-named pod there (common with GitOps-managed identical workloads). Before this PR that path only mattered if useEvict was on; now it's always reachable.

A few more inline.

illume (changes requested):
Thanks for these changes.


<details>

</details>

frontend/src/components/common/Resource/DeleteButton.tsx
krsatyamthakur-droid: good catch, the dropdown now only shows the alternate action (the primary button already covers the default one).

frontend/src/components/common/Resource/DeleteButton.tsx
krsatyamthakur-droid: added aria-haspopup, aria-expanded, and aria-controls to the trigger.

frontend/src/components/common/Resource/DeleteButton.test.tsx
krsatyamthakur-droid: updated, this test now checks the dropdown only has the alternate action.

krsatyamthakur-droid:
fixed the conflicts and addressed everything from the review — delete/evict now check their real RBAC verbs independently instead of one authVerb gating both, and evict passes the pod's cluster through now too. dropdown, aria attrs, and the test are also updated. ptal

krsatyamthakur-droid:
hii @unlikelyzero thanks for the thorough review, the RBAC and cluster catches were both real bugs — appreciate you digging in

krsatyamthakur-droid:
hii @illume , addressed unlikelyzero's review and the conflicts are resolved too — should be good for another look whenever you get a chance

illume (approved):
🎉 thanks!

PR #6795, opened 2026-08-01
backend: deduplicate concurrent OIDC token refreshes to fix invalid_grant/forced-relogin race

### LINKED ISSUE #6793: OIDC token-refresh middleware has no concurrency guard, causing invalid_grant errors and forced re-logins under refresh-token rotation
## Describe the bug

`NewOIDCTokenRefreshMiddleware` (`backend/pkg/auth/middleware.go`) refreshes an OIDC token whenever it is within `JWTExpirationTTL` (10s) of expiring. This check-and-refresh runs independently, once per incoming HTTP request, with no locking or de-duplication between concurrent requests for the same token/cluster:

```go
// backend/pkg/auth/middleware.go
if !IsTokenAboutToExpire(token) {
    ...
    next.ServeHTTP(w, r)
    return
}

RefreshAndSetToken(RefreshAndSetTokenParams{ ...Token: token... })
next.ServeHTTP(w, r)
```

`RefreshAndSetToken` → `RefreshAndCacheNewToken` → `GetNewToken` (`backend/pkg/auth/auth.go`) then does an unsynchronized read-refresh-write against the shared token cache:

```go
// backend/pkg/auth/auth.go
func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
    tokenType string, token string, tokenURL string, ctx context.Context,
) (*oauth2.Token, error) {
    refreshToken, err := cache.Get(ctx, oidcKeyPrefix+token)   // read
    ...
    newToken, err := conf.TokenSource(ctx, &oauth2.Token{RefreshToken: rToken}).Token() // exchange with IdP
    ...
    if err := CacheRefreshedToken(newToken, tokenType, token, rToken, cache); err != nil { // write
        ...
    }
    return newToken, nil
}
```

The Headlamp frontend routinely fires many parallel `/clusters/<name>/...` requests (list/watch calls for a cluster overview, multiple resource kinds, RBAC checks, etc.). It's normal for several of these to land within the same 10‑second expiry window, so multiple goroutines read the *same* cached refresh token and each independently calls the IdP's token endpoint with it before any of them has written the new one back.

Most OIDC providers Headlamp documents support (Keycloak, Okta, Azure AD, Auth0, Dex) implement **refresh-token rotation with reuse detection**: a refresh token is single-use, and presenting an already-consumed one either returns `invalid_grant` or revokes the whole token family (forcing full re-authentication). Under this middleware:

1. Only the first of the N concurrent exchanges the IdP processes succeeds.
2. The rest fail with `invalid_grant`. That error is logged and swallowed — `RefreshAndSetToken` just leaves the (soon-to-expire) old token in place and moves on (`auth.go`, `RefreshAndSetToken`'s error branch).
3. On providers with reuse-detection, this can revoke the entire refresh chain, forcing the user to fully re-authenticate mid-session, with no clear error message tying it back to a refresh race.
4. Even without reuse-detection, `CacheRefreshedToken` itself has no synchronization: if two goroutines each get a distinct rotated token back (IdP grace period), whichever `cache.Set` call runs last "wins" and the other newly-issued refresh token is only reachable via the old-token mapping, which is deliberately expired after `oldTokenTTL` (10s). That token is effectively orphaned.
5. `SetTokenCookie`, called at the end of `RefreshAndSetToken`, writes `Set-Cookie` on the HTTP response for every request that reached the refresh branch. Concurrent responses race to set the cookie in the browser, so the browser can end up persisting a cookie tied to a refresh chain a sibling request already invalidated.

The net effect is intermittent, hard-to-reproduce forced logouts and "broken session" reports for any multi-cluster or actively-polling session authenticated via an OIDC provider with rotating refresh tokens — a plausible explanation for previously reported vague OIDC session-expiry symptoms (e.g. #3918) that were never traced to a concurrency root cause.

## To Reproduce

1. Configure Headlamp with an OIDC provider that rotates refresh tokens on use (e.g. Keycloak with "Revoke Refresh Token" / rotation enabled, or Okta).
2. Log in and open a cluster view that fires several concurrent `/clusters/<name>/...` API calls (e.g. the default overview page, or any view listing multiple resource kinds).
3. Wait until the cached token is within 10 seconds of expiry, then trigger several simultaneous requests (e.g. switch between resource list views quickly, or have several browser tabs/panels open against the same cluster).
4. Observe backend logs for `"failed to refresh token"` / `invalid_grant` errors occurring for some of the concurrent refresh attempts, and/or the user being unexpectedly logged out despite the token nominally being refreshable.

This can also be reproduced deterministically with a unit test that calls `NewOIDCTokenRefreshMiddleware` (or `GetNewToken`) concurrently from multiple goroutines against a fake token endpoint that returns `invalid_grant` when a refresh token is reused — `go test -race` will additionally flag the unsynchronized cache read/write pair in `GetNewToken`/`CacheRefreshedToken`.

## Root cause

There is no mutual exclusion or de-duplication (e.g. a per-token/per-cluster mutex, or a `singleflight`-style guard) around the "check expiry → read refresh token → exchange with IdP → write new token" critical section. The refresh decision and execution happen independently inside every request-handling goroutine instead of being centralized, so N concurrent requests for the same near-expiry token each perform a full, independent refresh cycle against a token cache designed for single-writer use.

## Impact

- Intermittent forced logouts / broken sessions for OIDC users on providers with refresh-token rotation and reuse detection — the more clusters or concurrent views a user has open, the more likely this triggers.
- Silent failures: refresh errors from the losing goroutines are logged and discarded, giving users and operators no actionable signal that this is a concurrency issue rather than an IdP misconfiguration.
- Orphaned refresh tokens accumulate in the cache under `oidc-token-<newToken>` keys that are never looked up again, since only the last write "wins" the update.
- Affects any deployment using OIDC auth, which is a primary supported authentication path documented for Headlamp.

## Affected files/components

- `backend/pkg/auth/auth.go` — `GetNewToken`, `CacheRefreshedToken`, `RefreshAndCacheNewToken`, `RefreshAndSetToken`
- `backend/pkg/auth/middleware.go` — `NewOIDCTokenRefreshMiddleware`
- `backend/pkg/auth/auth_test.go`, `backend/pkg/auth/middleware_test.go` — need new concurrency regression tests
- Related, but not overlapping: #6214 (structural relocation of this middleware into `backend/pkg/auth`, already done) and #3918 (closed symptom report, likely the same root cause, never diagnosed as a race)

## Acceptance criteria

- Concurrent requests that observe the same near-expiry token result in exactly one outbound refresh call to the IdP per token (verified with a test that fires N goroutines through `NewOIDCTokenRefreshMiddleware` against a call-counting fake token endpoint).
- Requests arriving while a refresh for their token is already in flight either wait for its result or proceed using the still-valid old token — they must never independently re-exchange the same refresh token.
- `CacheRefreshedToken`'s cache reads/writes are race-free under `go test -race`.
- No `Set-Cookie` clobbering: only the request that completed the winning refresh sets the new cookie value; other concurrent requests do not overwrite it with stale data.
- Existing single-request refresh behavior, telemetry events, and status attributes are unchanged for the non-concurrent case.
- New regression tests reproduce the current race (fail on `main`, pass with the fix).

## High-level implementation plan

1. **`backend/pkg/auth/auth.go`**: introduce per-token (or per-cluster+user) de-duplication around the refresh critical section — e.g. `golang.org/x/sync/singleflight` keyed on `oidcKeyPrefix+token`, or an internal map of mutexes — so concurrent callers for the same token collapse into a single upstream refresh and all receive the same resulting token.
2. **`backend/pkg/auth/middleware.go`**: update `NewOIDCTokenRefreshMiddleware` to use the de-duplicated refresh result before calling `next.ServeHTTP`, and avoid issuing a redundant `Set-Cookie` when a concurrent request already refreshed the token moments earlier.
3. **Tests**: add concurrency tests in `auth_test.go` and `middleware_test.go` that assert a single upstream call per token under concurrent load (via a call-counting fake token server) and pass under `-race`.
4. **Docs**: document the refresh de-duplication guarantee (package doc or `docs/`), including the interaction with IdPs that rotate/invalidate refresh tokens on reuse, so operators debugging intermittent OIDC session drops (Keycloak/Okta/Azure AD) have something to point to.

Yes (I will propose a PR).

## Additional Context

Not a regression tied to a single release — the middleware has lacked synchronization since the OIDC refresh logic was first introduced/extracted into `backend/pkg/auth`. This is likely the underlying cause of previously reported, never-root-caused OIDC session-expiry complaints such as #3918.

bhuvan-somisetty:
Fixes #6793

## Problem

`NewOIDCTokenRefreshMiddleware` checks whether a token is about to expire and refreshes it independently on every incoming request, with no synchronization. When several concurrent requests for the same cluster observe the same near-expiry token (Headlamp routinely fires many parallel `/clusters/<name>/...` calls), each one raced to read the same cached refresh token and exchange it with the IdP.

On providers that rotate refresh tokens on use (Keycloak, Okta, Azure AD, Dex), only the first exchange succeeds; the rest fail with `invalid_grant`, which was logged and silently swallowed. On providers with reuse detection, presenting an already-consumed refresh token can revoke the entire token family, forcing the user to fully re-authenticate mid-session.

## Fix

`GetNewToken` in `backend/pkg/auth/auth.go` now deduplicates concurrent refreshes for the same token using a `singleflight.Group`. Only one goroutine performs the actual cache read, IdP exchange, and cache write per token; all other concurrent callers for that same token wait and receive the identical resulting token instead of independently re-exchanging it.

The shared refresh runs on a context detached from any single caller's cancellation (`context.WithoutCancel`), so one request finishing or its client disconnecting can't abort the refresh that other concurrent requests are waiting on.

`golang.org/x/sync` was already an indirect dependency (via other transitive deps); it's now a direct one after `go mod tidy`.

## Testing

- `TestGetNewToken_ConcurrentRefresh_Deduplicates`: fires 20 concurrent `GetNewToken` calls for the same token against a call-counting fake token endpoint and asserts exactly one upstream call happened and every caller received the same refreshed token.
- `TestGetNewToken_SequentialRefreshesAfterDeduplication`: confirms the dedup key is forgotten once its refresh completes, so a later, independent refresh cycle still hits the IdP.
- Full existing `backend/pkg/auth` suite passes unchanged (`go test ./pkg/auth/...`).
- `go build ./...` and `go vet ./pkg/auth/...` pass clean.

I wasn't able to run `-race` locally (no cgo toolchain in this environment), but the fix relies on `singleflight.Group`, a synchronization primitive designed specifically to make this pattern race-free, and CI's race job should confirm it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication token refresh handling to prevent duplicate refresh requests during concurrent activity.
  - Reuses recent successful refresh results, reducing failures caused by refresh-token rotation.
  - Ensures requests continue reliably while token status and refresh activity are tracked correctly.
  - Added handling for tokens that do not require expiration-based refreshes.

### DIFF
--- a/backend/pkg/auth/auth.go
+++ b/backend/pkg/auth/auth.go
@@ -229,14 +229,17 @@ func CacheRefreshedToken(token *oauth2.Token, tokenType string, oldToken string,
 func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
 	tokenType string, token string, tokenURL string, ctx context.Context,
 ) (*oauth2.Token, error) {
-	// Detach from the caller's context so that one request's cancellation (e.g.
-	// the client disconnecting) doesn't abort the shared refresh that other
-	// concurrent requests for the same token are waiting on. Bound it with its
-	// own timeout so a hung IdP can't block waiting callers indefinitely.
-	refreshCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), refreshTimeout)
-	defer cancel()
-
 	result, err, _ := refreshGroup.Do(token, func() (interface{}, error) {
+		// Detach from the caller's context so that one request's cancellation (e.g.
+		// the client disconnecting) doesn't abort the shared refresh that other
+		// concurrent requests for the same token are waiting on. Bound it with its
+		// own timeout so a hung IdP can't block waiting callers indefinitely.
+		//
+		// Created here, inside the singleflight closure, so it's allocated exactly
+		// once per in-flight refresh rather than once per waiting caller.
+		refreshCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), refreshTimeout)
+		defer cancel()
+
 		return refreshTokenFromIDP(refreshCtx, clientID, clientSecret, cache, tokenType, token, tokenURL)
 	})
 	if err != nil {
--- a/backend/pkg/auth/auth.go
+++ b/backend/pkg/auth/auth.go
@@ -40,12 +40,37 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/oauth2"
+	"golang.org/x/sync/singleflight"
 )
 
 const (
-	oldTokenTTL   = time.Second * 10 // seconds
-	oidcKeyPrefix = "oidc-token-"
+	oldTokenTTL             = time.Second * 10 // seconds
+	oidcKeyPrefix           = "oidc-token-"
+	refreshedTokenKeyPrefix = "oidc-refreshed-token-" //nolint:gosec // cache key prefix, not a credential
+	// refreshTimeout bounds the detached context used for the shared IdP
+	// exchange in GetNewToken, so a hung IdP can't block waiting callers
+	// indefinitely.
+	refreshTimeout = 30 * time.Second
 )
+
+// refreshGroup deduplicates concurrent OIDC refreshes for the same token.
+//
+// The OIDC refresh middleware runs independently for every incoming request, and
+// Headlamp's frontend commonly fires many concurrent requests against a cluster.
+// Without this, every request that observes the same soon-to-expire token would
+// race to exchange the same cached refresh token with the IdP: only the first
+// exchange succeeds, and the rest fail (or, on providers with refresh-token
+// rotation and reuse detection, can revoke the whole token family and force the
+// user to fully re-authenticate). Keying on the old token string is sufficient
+// since it uniquely identifies the in-flight refresh for a given session.
+//
+// This only dedupes calls that overlap in time. A request can still observe the
+// old, about-to-expire token via RefreshTokenIfNeeded's IsTokenAboutToExpire
+// check, then reach GetNewToken after some other request already completed a
+// refresh for that same token (the singleflight.Group forgets a key as soon as
+// its Do call returns). refreshedTokenKeyPrefix below covers that non-overlapping
+// case.
+var refreshGroup singleflight.Group
 
 const JWTExpirationTTL = 10 * time.Second // seconds
 
@@ -186,9 +211,70 @@ func CacheRefreshedToken(token *oauth2.Token, tokenType string, oldToken string,
 // GetNewToken uses the provided credentials and fetches the old refresh
 // token from the cache to obtain a new OAuth2 token
 // from the specified token URL endpoint.
+//
+// Concurrent calls for the same token are deduplicated via refreshGroup: only
+// one of them performs the actual cache read, IdP exchange, and cache write,
+// and all of them receive the same resulting token. This avoids racing
+// multiple exchanges of the same refresh token, which fails outright on IdPs
+// that rotate refresh tokens on use.
+//
+// A caller can also reach this function for a token that was already refreshed
+// by an earlier, now-completed call (e.g. the middleware read the old token
+// before a concurrent request refreshed it, but only got scheduled here
+// afterwards). refreshTokenFromIDP checks for that case first, inside the
+// singleflight execution, so it reuses the already-refreshed token instead of
+// attempting a second exchange with a refresh token the IdP already rotated
+// out. Checking outside the singleflight call would leave the same gap: the
+// check and the exchange itself must happen as one atomic unit per key.
 func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
 	tokenType string, token string, tokenURL string, ctx context.Context,
 ) (*oauth2.Token, error) {
+	result, err, _ := refreshGroup.Do(token, func() (interface{}, error) {
+		// Detach from the caller's context so that one request's cancellation (e.g.
+		// the client disconnecting) doesn't abort the shared refresh that other
+		// concurrent requests for the same token are waiting on. Bound it with its
+		// own timeout so a hung IdP can't block waiting callers indefinitely.
+		//
+		// Created here, inside the singleflight closure, so it's allocated exactly
+		// once per in-flight refresh rather than once per waiting caller.
+		refreshCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), refreshTimeout)
+		defer cancel()
+
+		return refreshTokenFromIDP(refreshCtx, clientID, clientSecret, cache, tokenType, token, tokenURL)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	newToken, ok := result.(*oauth2.Token)
+	if !ok {
+		return nil, fmt.Errorf("unexpected refresh result type %T", result)
+	}
+
+	return newToken, nil
+}
+
+// refreshTokenFromIDP performs the actual refresh-token exchange with the IdP
+// and updates the token cache. It is only ever invoked once per in-flight
+// token, via refreshGroup in GetNewToken, even when multiple goroutines call
+// GetNewToken concurrently for the same token.
+//
+// The very first thing it does is check whether this old token was already
+// refreshed by an earlier, now-completed singleflight call for the same key.
+// That check has to live here, inside the singleflight execution, rather than
+// in GetNewToken before calling refreshGroup.Do: checking there would leave a
+// window between the check and the Do call where a concurrent refresh could
+// complete, so a late caller could still retry the exchange with a refresh
+// token the IdP already rotated out.
+func refreshTokenFromIDP(ctx context.Context, clientID, clientSecret string, cache cache.Cache[interface{}],
+	tokenType, token, tokenURL string,
+) (*oauth2.Token, error) {
+	if cached, err := cache.Get(ctx, refreshedTokenKeyPrefix+token); err == nil {
+		if cachedToken, ok := cached.(*oauth2.Token); ok {
+			return cachedToken, nil
+		}
+	}
+
 	// get refresh token
 	refreshToken, err := cache.Get(ctx, oidcKeyPrefix+token)
 	if err != nil {
@@ -218,6 +304,14 @@ func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
 	// update the refresh token in the cache
 	if err := CacheRefreshedToken(newToken, tokenType, token, rToken, cache); err != nil {
 		return nil, fmt.Errorf("caching refreshed token: %w", err)
+	}
+
+	// Remember the outcome of this refresh under the old token for a short
+	// grace period, so a request that checked the old token before this
+	// refresh happened but only reaches GetNewToken afterwards reuses this
+	// result instead of retrying the exchange with a now-stale refresh token.
+	if err := cache.SetWithTTL(ctx, refreshedTokenKeyPrefix+token, newToken, oldTokenTTL); err != nil {
+		logger.Log(logger.LevelError, nil, err, "failed to cache refresh outcome")
 	}
 
 	return newToken, nil
@@ -649,4 +743,23 @@ func RefreshAndSetToken(params RefreshAndSetTokenParams) {
 
 		params.TelemetryHandler.RecordEvent(params.Span, "Token refreshed successfully")
 	}
+}
+
+// RefreshTokenIfNeeded checks whether params.Token is close to expiring and, if
+// so, refreshes it via RefreshAndSetToken. It reports whether a refresh was
+// attempted, so callers can tell "token still valid, nothing to do" apart from
+// "token was expiring, refresh was attempted" without checking expiry themselves.
+//
+// This keeps the expiry decision next to the refresh machinery it gates,
+// instead of split out into the caller (e.g. the middleware): the caller no
+// longer needs its own IsTokenAboutToExpire check before deciding whether to
+// invoke a refresh.
+func RefreshTokenIfNeeded(params RefreshAndSetTokenParams) bool {
+	if !IsTokenAboutToExpire(params.Token) {
+		return false
+	}
+
+	RefreshAndSetToken(params)
+
+	return true
 }
--- a/backend/pkg/auth/auth.go
+++ b/backend/pkg/auth/auth.go
@@ -44,8 +44,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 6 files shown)


bhuvan-somisetty:
Hi maintainers, this PR fixes a concurrency bug in the OIDC token-refresh path (issue 6793) — concurrent requests near token expiry were racing to refresh the same cached token, which fails against IdPs that rotate refresh tokens on use (Keycloak, Okta, Azure AD) and can force users to fully re-authenticate. Fixed by deduplicating concurrent refreshes with a singleflight group, plus regression tests reproducing the race.

@illume @sniok @skoeva @vyncent-t would appreciate a review when you get a chance. Happy to make any changes needed.

backend/pkg/auth/auth.go
```diff
@@ -186,8 +199,41 @@ func CacheRefreshedToken(token *oauth2.Token, tokenType string, oldToken string,
...
+// that rotate refresh tokens on use.
 func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
 	tokenType string, token string, tokenURL string, ctx context.Context,
+) (*oauth2.Token, error) {
+	// Detach from the caller's context so that one request's cancellation (e.g.
+	// the client disconnecting) doesn't abort the shared refresh that other
+	// concurrent requests for the same token are waiting on.
+	refreshCtx := context.WithoutCancel(ctx)
+
+	result, err, _ := refreshGroup.Do(token, func() (interface{}, error) {
```
uhthomas, line 232: Although this helps prevent using the refresh token concurrently, it may not alleviate the race condition in the actual middleware which checks if a token needs refreshing. It's possible that another go routine may check IsTokenAboutToExpire before the token is refreshed, but then actually get here after it has been refreshed.

https://github.com/kubernetes-sigs/headlamp/blob/99a230be9c9c679a70d59c219cc246c00ae2be45/backend/pkg/auth/middleware.go#L124

bhuvan-somisetty: @uhthomas Good catch, thanks — you're right that the singleflight dedup only covers overlapping refreshes, not this sequential case.

Fixed in 41aa300: a request that reads the old token before a concurrent refresh completes, but only reaches `GetNewToken` after that refresh's singleflight call has already resolved (and forgotten the key), now reuses the just-completed refresh instead of retrying the exchange with a refresh token the IdP already rotated out. Added `TestGetNewToken_ReusesRecentRefreshForSameOldToken` to cover it.

Let me know if this looks right to you.

bhuvan-somisetty: @uhthomas all green now, CI's passing on the latest commit. Thanks again for catching that 🙏

backend/pkg/auth/auth.go
```diff
@@ -186,8 +206,55 @@ func CacheRefreshedToken(token *oauth2.Token, tokenType string, oldToken string,
...
+// by an earlier, now-completed call (e.g. the middleware read the old token
+// before a concurrent request refreshed it, but only got scheduled here
+// afterwards). refreshedTokenKeyPrefix caches the outcome of each refresh
+// briefly so that case reuses the already-refreshed token instead of
+// attempting a second exchange with a refresh token the IdP already rotated
+// out.
 func GetNewToken(clientID, clientSecret string, cache cache.Cache[interface{}],
 	tokenType string, token string, tokenURL string, ctx context.Context,
+) (*oauth2.Token, error) {
+	if cached, err := cache.Get(ctx, refreshedTokenKeyPrefix+token); err == nil {
```
uhthomas: I think this is still racy. As there is a chance that the token could be refreshed in between this condition and the single flight fetch. It should be resolved by moving this into the single flight execution?

backend/pkg/auth/auth.go
bhuvan-somisetty: @uhthomas good point, you're right — checking before the singleflight call still left a window. Moved the check into refreshTokenFromIDP so it runs as part of the singleflight execution itself. CI's green on the latest commit.

illume (changes requested):
Thanks for these changes.


<details>

- `backend: deduplicate concurrent OIDC token refreshes` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.
- `backend: close remaining race in OIDC refresh middleware` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.
- `backend: silence gosec false positive on refresh cache key prefix` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.
- `backend: move recent-refresh check inside the singleflight execution` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

bhuvan-somisetty:
Thanks for the reviews! Capitalized the commit descriptions per the contribution guide, and addressed both points from the Copilot review: the detached refresh context now has a 30s timeout so a hanging IdP can't block waiting callers forever, and fakeCache.Set now updates the store like SetWithTTL does, so the test double matches the real cache. Force-pushed the squashed/reworded history.

uhthomas:
I was thinking about this last night and I wonder if this is the right place to do it. I know that the current proposal works, but it seems a bit odd to me that we check if a token has expired in one place, and then rely on a timed cache to prevent refreshing that token multiple times should be the single flight group move into the actual handler instead? It might be a smaller and more focussed change.

bhuvan-somisetty:
Fair point, and I went back and forth on it. The catch is singleflight.Group forgets its key the moment Do() returns, so keying by cluster and moving the check inside wouldn't remove the race by itself a late request still needs to know "this was just refreshed," which is what the timed cache does today.

Actually dropping that cache would mean keying by cluster/session instead of token value, so a late request can check against a live "current token" pointer rather than the specific old token it's carrying. That's a bigger structural change than it looks. Happy to take it on as a follow-up if you think it's worth it didn't want to reshape this PR around it mid-review.

illume (changes requested):
Thanks for these changes.


bhuvan-somisetty:
Thanks for the nudge — went through and resolved all four open threads. Also circled back on the design question about moving the singleflight check into the actual handler: added RefreshTokenIfNeeded so the expiry check now lives right next to the refresh logic in auth.go instead of being split out in the middleware, and the middleware just calls it unconditionally now. Tests + go vet are green.

bhuvan-somisetty:
CI's golangci-lint caught something I hadn't hit locally before: findAuthCookie's cluster param was flagged as unparam since every call site in the test file passes the same "test" literal. Dropped the param and hardcoded it in the helper. All checks are green now.

illume (changes requested):
Thanks for these changes.


<details>

- `backend: Silence gosec false positive on refresh cache key prefix` — Only one file changed inside `backend/`; add a sub-area so it's clear what was touched (e.g. `backend: ComponentName: description`).
- `backend: Move recent-refresh check inside the singleflight execution` — Only one file changed inside `backend/`; add a sub-area so it's clear what was touched (e.g. `backend: ComponentName: description`).

</details>

<details>

Good examples:

</details>

bhuvan-somisetty:
Thanks both — fixed.

Added the `auth:` sub-area to the two commits that were missing it (they only touched auth.go), and moved the detached refresh context creation inside the singleflight `Do` closure so it's allocated once per in-flight refresh instead of once per waiting caller, per Copilot's note. Build, vet, and the auth test suite are all green.

illume (changes requested):
Thanks for the contribution.


<details>

</details>

bhuvan-somisetty:
Hi @illume, I have rebased the branch onto latest main and resolved the conflicts in backend/go.mod. All backend tests are passing.

bhuvan-somisetty:
@illume addressed the latest round of Copilot feedback: renamed the middleware's `refreshed` bool to `refreshAttempted` so it actually matches what `RefreshTokenIfNeeded` reports, pulled the duplicated store-init logic in `fakeCache` into a small `storeSet` helper, swapped the `require.*` calls inside the new tests' httptest handler goroutines for `assert.*` (avoids a possible fatal-off-test-goroutine panic), and restored `golang.org/x/net` to v0.56.0 since the downgrade to v0.55.0 was unintentional from `go mod tidy`. Build, vet, and the full auth suite are green.

illume (changes requested):
Thanks for these changes.


<details>

- `auth: address latest Copilot review feedback on PR 6795` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for the contribution.


bhuvan-somisetty:
@illume squashed the review-correction commits into two focused ones — the singleflight dedup fix itself, and a new test for the concurrency gap Copilot flagged (the middleware tests only covered bypass paths before; now there's one that fires concurrent requests through the actual middleware and checks a single token-endpoint call plus a correctly-refreshed cookie on every response). Build, vet, and the full auth suite are green.

bhuvan-somisetty:
Hi @illume, I've addressed both Copilot review comments:

1. Added TestOIDCTokenRefreshMiddleware_ConcurrentRequests in ackend/pkg/auth/middleware_test.go to exercise the full expiry-check -> discovery -> singleflight deduplicated refresh -> Set-Cookie flow under concurrent requests, asserting a single token-endpoint call and current cookie values on all responses.
2. Interactively squashed the commits into a single clean, focused logical commit.

All tests and checks are passing cleanly. Could you please take another look?

illume (changes requested):
Thanks for this PR.


<details>

- `backend: deduplicate concurrent OIDC token refreshes` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

bhuvan-somisetty:
Good catch, thanks for pointing that out. I'd already squashed things down to one commit, but missed that the title itself still didn't follow the capitalization rule. Fixed it to `backend: Deduplicate concurrent OIDC token refreshes` and force-pushed.

PR #6801, opened 2026-08-01
backend: auth: Fix cluster name cookie collision for long cluster names

### LINKED ISSUE #6800: [backend/auth] Cookie key collision for long cluster names in SanitizeClusterName
## Describe the bug

`SanitizeClusterName()` in `backend/pkg/auth/cookies.go` truncates sanitized cluster names to 50 characters.

If two Kubernetes cluster names are longer than 50 characters and share the same first 50 sanitized characters, they produce identical sanitized names. Since the sanitized name is used as part of the authentication cookie key, different clusters can end up using the same cookie.

As a result, authenticating to one cluster can overwrite the authentication cookie of another cluster, causing authentication failures when switching back to the first cluster.

## To Reproduce

1. Create two Kubernetes cluster names longer than 50 characters that share the same first 50 sanitized characters.
2. For example:
   - `production-k8s-us-east-1-apps-microservices-cluster-alpha`
   - `production-k8s-us-east-1-apps-microservices-cluster-bravo`
3. Call `SanitizeClusterName()` for both names.
4. Observe that both return the same sanitized value because the names are truncated to 50 characters.
5. Use the resulting values with `SetTokenCookie()`.
6. Observe that both clusters generate the same cookie name, allowing the later cookie to overwrite the earlier one.

## Expected behavior

Distinct cluster names should always generate unique authentication cookie names, even when the names exceed the truncation limit, so that authentication sessions remain isolated across clusters.

## Environment (please provide info about your environment)

- Installation type: Local development build (`main` branch)
- Headlamp version: `main`
- Component: `backend/pkg/auth`

Yes. I have a fix prepared and can open a pull request.

### Technical Root Cause

`SanitizeClusterName()` removes unsupported characters and truncates the sanitized string to 50 characters:

```go
func SanitizeClusterName(cluster string) string {
	reg := regexp.MustCompile(`[^a-zA-Z0-9\-_]`)
	sanitized := reg.ReplaceAllString(cluster, "")
	if len(sanitized) > 50 {
		sanitized = sanitized[:50]
	}
	return sanitized
}
```

The sanitized value is later incorporated into the authentication cookie name by `SetTokenCookie()`. When two distinct cluster names produce the same truncated sanitized value, they also produce identical cookie names, allowing one cluster's authentication cookie to overwrite another's.

AYUSH-P-SINGH:
## Summary

This PR fixes an authentication cookie collision in `backend/pkg/auth` caused by `SanitizeClusterName()` truncating long cluster names to a fixed 50-character prefix. When two cluster names shared the same first 50 sanitized characters, they generated identical authentication cookie names, allowing one cluster's session cookie to overwrite another's.

The fix preserves the 50-character limit while ensuring distinct long cluster names generate unique cookie names.

## Related issue

Fixes #6800 

## Changes

- **Updated `backend/pkg/auth/cookies.go`**
  - Modified `SanitizeClusterName()` to preserve uniqueness for cluster names longer than 50 characters by truncating the sanitized prefix and appending a deterministic 8-character SHA-256 hash suffix.

- **Added `backend/pkg/auth/cookies_test.go`**
  - Added unit tests verifying that:
    - Long cluster names with identical prefixes generate unique sanitized names.
    - Sanitized output remains within the 50-character limit.
    - Existing behavior for shorter cluster names is preserved.

- **Updated `backend/pkg/k8cache/cacheStore_bench_test.go`**
  - Modernized benchmark loops to use Go 1.24's `b.Loop()` API.

## Steps to Test

1. Run the authentication package tests:

   ```bash
   go test -v ./pkg/auth/...
   ```

2. Configure two Kubernetes clusters with names longer than 50 characters that share the same prefix, for example:

   - `production-k8s-us-east-1-apps-microservices-cluster-alpha`
   - `production-k8s-us-east-1-apps-microservices-cluster-bravo`

3. Authenticate to both clusters.

4. Verify that each cluster receives a distinct authentication cookie and that authenticating to one cluster does not overwrite the other's session cookie.

N/A (backend authentication logic change).

## Notes for the Reviewer

- The maximum sanitized cluster name length remains 50 characters.
- The hash suffix is deterministic, so the same cluster name always produces the same sanitized value.
- Existing behavior for cluster names of 50 characters or fewer is unchanged.
- All authentication package tests, including the new collision tests, pass successfully.

### DIFF
--- a/backend/pkg/auth/cookies.go
+++ b/backend/pkg/auth/cookies.go
@@ -17,6 +17,7 @@ limitations under the License.
 package auth
 
 import (
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"net/http"
@@ -45,9 +46,11 @@ func SanitizeClusterName(cluster string) string {
 	reg := regexp.MustCompile(`[^a-zA-Z0-9\-_]`)
 	sanitized := reg.ReplaceAllString(cluster, "")
 
-	// Limit length to prevent issues
+	// Limit length to prevent issues, appending a short hash slice for uniqueness when truncated.
 	if len(sanitized) > 50 {
-		sanitized = sanitized[:50]
+		hashBytes := sha256.Sum256([]byte(cluster))
+		hash := fmt.Sprintf("%x", hashBytes[:16])
+		sanitized = sanitized[:17] + "-" + hash
 	}
 
 	return sanitized
--- a/backend/pkg/auth/cookies.go
+++ b/backend/pkg/auth/cookies.go
@@ -17,6 +17,7 @@ limitations under the License.
 package auth
 
 import (
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"net/http"
@@ -45,9 +46,10 @@ func SanitizeClusterName(cluster string) string {
 	reg := regexp.MustCompile(`[^a-zA-Z0-9\-_]`)
 	sanitized := reg.ReplaceAllString(cluster, "")
 
-	// Limit length to prevent issues
+	// Limit length to prevent issues, appending a short hash slice for uniqueness when truncated.
 	if len(sanitized) > 50 {
-		sanitized = sanitized[:50]
+		hash := fmt.Sprintf("%x", sha256.Sum256([]byte(cluster)))[:8]
+		sanitized = sanitized[:41] + "-" + hash
 	}
 
 	return sanitized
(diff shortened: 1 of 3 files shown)


illume (changes requested):
Thanks for working on this.


AYUSH-P-SINGH:
Addressed the review comments and updated the commit. Thanks!

illume (changes requested):
Thanks for the contribution.

PR #6806, opened 2026-08-01
frontend: Logviewer: Fix memory leak

anubhavsingh2106:
## Summary

This PR fixes a memory leak in the `LogViewer` component by ensuring that terminal resources and event listeners are properly disposed of when the component unmounts.

## Changes

- **Updated** [frontend/src/components/common/LogViewer.tsx] to include proper cleanup logic within the component cleanup functions.
- **Fixed** a memory leak issue where log rendering resources (e.g., terminal instances or observers) were not being fully freed, leading to degraded performance over time.
- **Refactored** resource management in the log viewer for better long-term application stability and performance.


illume (changes requested):
Thanks for the contribution.

Issue #6692, opened 2026-07-27
frontend: Table: actions column is never hidden due to column id mismatch


Aamod007:
## Describe the bug
The Table component has an effect meant to hide the row-actions column when all data columns are hidden, but it toggles visibility for a column id that doesn't exist. In `frontend/src/components/common/Table/Table.tsx`:

```ts
// Hide actions column when others are hidden
useEffect(() => {
  const visibility = table.getState().columnVisibility || {};
  const shouldHideActions = tableColumns
    .filter(col => (col.id ?? '') !== 'actions')
    .every(col => visibility[col.id ?? ''] === false);
  if (shouldHideActions && visibility['actions'] !== false) {
    table.setColumnVisibility(prev => ({ ...prev, actions: false }));
  } ...
```

Material React Table's built-in row actions column id is `mrt-row-actions` — the same file pushes `'mrt-row-actions'` into `columnOrder` a few hundred lines earlier. Setting `visibility['actions']` therefore toggles a key no rendered column uses, and the intended behavior never fires: when all data columns are responsively hidden, the empty actions column and its grid track are still rendered instead of collapsing.

Expected: hiding all data columns also hides the actions column.

## To Reproduce
Steps to reproduce the bug:
1. Open any resource list with row actions enabled.
2. Hide all data columns via the column visibility menu (or use gridTemplateColumns/hideColumns so every data column is hidden).
3. The actions column remains rendered (empty) instead of being hidden; `columnVisibility` state gains an `actions: false` entry that matches no column.

## Environment (please provide info about your environment):
 - Installation type: any (frontend bug)
 - Headlamp Version: main (d6c619a88)
 - Other: Found by code inspection.

## Are you able to fix this issue?
Yes (I will propose a PR).

## Additional Context
Fix: use the actual MRT id `mrt-row-actions` in the visibility check/update.

PR #6811, opened 2026-08-02
frontend: Table: Let column toggles override responsive hiding

### LINKED ISSUE #6733: bug: hidden columns cannot be made visible again
## Describe the bug
It seems that after the changes in https://github.com/kubernetes-sigs/headlamp/pull/6468, when a column is hidden due to the screen size it cannot be made visible again.

## To Reproduce
Steps to reproduce the bug:
1. Open any page with a table
2. Resize the headlamp window until columns start being hidden
3. Try to make a hidden column visible again
4. The toggle for that column will not move.
5. Other columns, like "Labels" can be made visible and the behavior reverts to 0.43 with an horizontal scroll bar 

 - Installation type: Mac - Helm in cluster
 - Headlamp Version: 0.44.0

No

## Additional Context
Testable on mobile or on desktop by manually resizing the headlamp window

Anjali-Chauhan1:
## Summary

This PR fixes the column chooser being unable to restore a column that responsive hiding had removed. The merged visibility map applied the automatic decision last, so a column hidden to fit the width was re-hidden on every render regardless of what the user asked for — the toggle simply snapped back.

Investigating the report surfaced two further defects in the same code path, both included here:

- The responsive pass measured only the columns the caller had left visible. `Labels` is hidden by default, so it was never counted: turning it on made it appear, but no width was reserved for it and the table fell back to a horizontal scrollbar. This is the second behaviour the reporter distinguishes.
- The internal visibility handler was installed *after* `...tableProps` was spread into the table options, so it overwrote the handler `ResourceTable` passes in. That handler is where `storeTableSettings` runs, which left column preferences loading on mount but never saving again.

## Related Issue

Fixes #6733

## Changes

- **Gave the automatic and explicit decisions a defined order** in `frontend/src/components/common/Table/Table.tsx`. Precedence is now `initialState` → `state` → responsive hiding → user choice, with a new `baseColumnVisibility` layer for the caller's two inputs. Only columns the user actually toggled are tracked as user intent, so responsive hiding keeps applying to everything they haven't touched.
- **Counted user-enabled columns towards the width budget.** A column the user turns on is excluded from the hide candidates but still measured, so a lower-priority column is dropped rather than the table overflowing.
- **Composed the visibility handler instead of replacing it.** It records the user's deltas and forwards the same deltas to the caller as an updater, so responsively hidden columns are never written into stored settings.
- **Derived the actions column from the merged map** and removed the `useEffect` that wrote it back through `table.setColumnVisibility`. With controlled `columnVisibility` that round trip would be recorded as user intent. Both the caller-defined `actions` id and MRT's built-in `mrt-row-actions` id are handled, and neither is offered to the responsive pass.
- **Added `frontend/src/components/common/Table/Table.test.tsx`** — the directory had stories but no test file, which is how this shipped.

## Steps to Test

1. Open any resource list page (for example Pods) and narrow the window until columns start disappearing.
2. Open **Show/Hide columns** and switch on a column that responsive hiding removed. Before this change the switch snaps back; now the column appears and stays.
3. Switch on **Labels**, which is hidden by default. Before this change it appeared but pushed the table into a horizontal scrollbar; now a lower-priority column is dropped to make room.
4. Toggle any column, reload the page, and confirm the choice persists. Before this change it was lost, because `ResourceTable`'s handler never ran.
5. Widen the window again and confirm the columns you explicitly enabled stay enabled.

Automated: `cd frontend && npx vitest run src/components/common/Table/Table.test.tsx` — 6 tests.

To confirm they are real regression tests:

```
cp src/components/common/Table/Table.tsx /tmp/f.tsx
git checkout -- src/components/common/Table/Table.tsx
npx vitest run src/components/common/Table/Table.test.tsx   # failures
cp /tmp/f.tsx src/components/common/Table/Table.tsx
```

## Notes for the Reviewer

- **This touches the effect that #6697 and #6702 modify** (both for #6692). Those PRs change `'actions'` to `'mrt-row-actions'` inside the `useEffect` this PR removes. The removal is not incidental: with `columnVisibility` controlled, that effect can no longer write through `table.setColumnVisibility`, so it had to become a derived value. I handle **both** ids so their fix is not lost either way — happy to rebase on top if theirs lands first, or to drop the actions handling entirely if you would rather sequence it the other way. Worth deciding before either merges.

- **One behaviour change worth your judgement.** Now that user-enabled columns are counted, turning one on hides a lower-priority column instead of showing a horizontal scrollbar. That matches #6468's intent, but the reporter described the scrollbar neutrally as "0.43 behavior" and may have considered it acceptable. If you prefer that an explicit user choice is allowed to overflow, that is a one-line change.

- **The persistence fix widens the scope beyond what #6733 reports.** It is the same PR and the same code path, but nobody has filed it — it is invisible until you reload. Flagging so the diff touching `onColumnVisibilityChange` is not a surprise.

- **Verification status, stated plainly:** `tsc`, `eslint`, and `prettier` are clean, and the full frontend suite passes (183 files). Four of the six new tests are confirmed to fail without the fix; the two covering the actions column are **not yet confirmed as regression tests** — repeated attempts to run the without-fix comparison timed out on a loaded machine. Worth a second look in review or CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsive table behavior so columns hide and reappear correctly based on available space and user preferences.
  * Preserved user-enabled columns when responsive adjustments occur.
  * Ensured action columns and row-action layout remain consistent as column visibility changes.
  * Improved synchronization of column visibility updates with application settings.

### DIFF
--- a/frontend/src/components/common/Table/Table.responsiveColumns.test.tsx
+++ b/frontend/src/components/common/Table/Table.responsiveColumns.test.tsx
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from '@mui/material/styles';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { createMuiTheme } from '../../../lib/themes';
+import { TestContext } from '../../../test';
+import Table, { TableColumn } from './Table';
+
+const theme = createMuiTheme({ base: 'light', name: 'light' });
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key, i18n: { language: 'en' } }),
+  initReactI18next: { type: '3rdParty', init: () => {} },
+}));
+
+type Row = { name: string; namespace: string; labels: string; age: string };
+
+const rows: Row[] = [{ name: 'pod-a', namespace: 'default', labels: 'app=a', age: '1d' }];
+
+const columns: TableColumn<Row>[] = [
+  { id: 'name', header: 'Name', accessorFn: (r: Row) => r.name },
+  {
+    id: 'namespace',
+    header: 'Namespace',
+    accessorFn: (r: Row) => r.namespace,
+    responsivePriority: -2,
+  },
+  { id: 'labels', header: 'Labels', accessorFn: (r: Row) => r.labels, responsivePriority: -2 },
+  { id: 'age', header: 'Age', accessorFn: (r: Row) => r.age, responsivePriority: 1 },
+];
+
+/**
+ * Drives the ResizeObserver that Table uses to measure available width, so the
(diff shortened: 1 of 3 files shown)


illume (changes requested):
Thanks for this PR.

PR #6814, opened 2026-08-02
frontend: shortcutsSlice: Fix i18next uninitialized warning

KeerthiKumarR:
This fixes the `i18next was not initialized` and `missingKey` warnings that spam the console on app startup. 

### What changed
The issue was that `DEFAULT_SHORTCUTS` in `shortcutsSlice.ts` was calling `i18next.t()` directly when the Redux store first initialized. Since i18next isn't ready at that point, it threw warnings and cached the English strings before the actual locale could load.

To fix it:
- I deferred the actual translation to render-time in `ShortcutsSettings.tsx` using the `useTranslation` hook.
- Renamed `.name` and `.description` in the config to `.labelKey` and `.descriptionKey` so it's obvious they just hold translation keys, not the actual display strings.
- Added `// t('...')` comment markers above `DEFAULT_SHORTCUTS` so `i18next-parser` can still extract the strings properly during `npm run i18n` without us having to write a fake identity function.

### Steps to test
1. Start the frontend (`npm run frontend:start`) and check the browser console on startup — you shouldn't see those `i18next was not initialized` warnings anymore.
2. Go to **Settings > Keyboard Shortcuts**, switch the app language, and confirm the shortcut names/descriptions actually translate now (they used to just stay in English).
3. Run `npm run frontend:test -- src/redux/shortcutsSlice.test.ts` and `cd frontend && npm run i18n` to make sure the tests pass and the parser still extracts everything correctly.

### DIFF
--- a/frontend/src/redux/shortcutsSlice.ts
+++ b/frontend/src/redux/shortcutsSlice.ts
@@ -23,10 +23,10 @@ import { createSlice } from '@reduxjs/toolkit';
 export interface ShortcutConfig {
   /** The unique identifier for the shortcut */
   id: string;
-  /** The human-readable name of the shortcut */
-  name: string;
-  /** A description of what the shortcut does */
-  description: string;
+  /** The translation key for the human-readable name of the shortcut */
+  labelKey: string;
+  /** The translation key for the description of what the shortcut does */
+  descriptionKey: string;
   /** The current key combination assigned to the shortcut */
   key: string;
   /** The default key combination for the shortcut */
@@ -35,40 +35,46 @@ export interface ShortcutConfig {
   category: 'navigation' | 'search' | 'general';
 }
 
-import i18next from 'i18next';
-
 /**
  * All available shortcuts with their default configurations
  */
+// t('Global Search')
+// t('Open the global search dialog')
+// t('Cluster Chooser')
+// t('Open the cluster chooser popup')
+// t('Toggle Table Filters')
+// t('Toggle column filters in tables')
+// t('Log Viewer Search')
+// t('Toggle search in log viewer')
 export const DEFAULT_SHORTCUTS: Record<string, ShortcutConfig> = {
   GLOBAL_SEARCH: {
     id: 'GLOBAL_SEARCH',
-    name: i18next.t('Global Search'),
-    description: i18next.t('Open the global search dialog'),
+    labelKey: 'Global Search',
+    descriptionKey: 'Open the global search dialog',
     key: '/',
     defaultKey: '/',
     category: 'search',
   },
   CLUSTER_CHOOSER: {
     id: 'CLUSTER_CHOOSER',
-    name: i18next.t('Cluster Chooser'),
-    description: i18next.t('Open the cluster chooser popup'),
+    labelKey: 'Cluster Chooser',
+    descriptionKey: 'Open the cluster chooser popup',
     key: 'ctrl+shift+l',
     defaultKey: 'ctrl+shift+l',
     category: 'navigation',
   },
   TABLE_COLUMN_FILTERS: {
     id: 'TABLE_COLUMN_FILTERS',
-    name: i18next.t('Toggle Table Filters'),
-    description: i18next.t('Toggle column filters in tables'),
+    labelKey: 'Toggle Table Filters',
+    descriptionKey: 'Toggle column filters in tables',
     key: 'alt+shift+t',
     defaultKey: 'alt+shift+t',
     category: 'general',
   },
   LOG_VIEWER_SEARCH: {
     id: 'LOG_VIEWER_SEARCH',
-    name: i18next.t('Log Viewer Search'),
-    description: i18next.t('Toggle search in log viewer'),
+    labelKey: 'Log Viewer Search',
+    descriptionKey: 'Toggle search in log viewer',
     key: 'ctrl+shift+f',
     defaultKey: 'ctrl+shift+f',
     category: 'search',
--- a/frontend/src/redux/shortcutsSlice.ts
+++ b/frontend/src/redux/shortcutsSlice.ts
@@ -35,7 +35,7 @@ export interface ShortcutConfig {
   category: 'navigation' | 'search' | 'general';
 }
 
-import i18next from 'i18next';
+const t = (s: string) => s;
 
 /**
  * All available shortcuts with their default configurations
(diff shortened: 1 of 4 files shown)


KeerthiKumarR:
@illume @gambtho @joaquimrocha 
PTAL when you got time.
Thank You!

illume (changes requested):
Thanks for the contribution.


<details>

</details>

KeerthiKumarR:
@illume  Thanks for the feedback!
I've rebased the branch against the latest main to remove the merge commit and keep the history clean as requested. Let me know if anything else is needed!

illume (changes requested):
Thanks for working on this.


<details>

</details>

KeerthiKumarR:
@illume I have made the requested changes, let me know if there's anything else that needs to be changed.
Thank you!

illume (changes requested):
Thanks for these changes.


I noticed the GitHub CI frontend i18n check is failing. It looks like the translation files need to be updated. You can run `cd frontend && npm run i18n` locally to update them and then commit the result.

<details>
<summary>How to update translation files</summary>

Run `cd frontend && npm run i18n` to regenerate the translation files. Commit the updated files alongside your other changes.

</details>

KeerthiKumarR:
@illume Thanks for the feedback
I've updated shortcutsSlice to use a const t = (s: string) => s to notice marker function. This allows i18next-parser to recognize shortcut translation keys during extraction without modifying any translation.json files while completely avoiding early i18next initialization on app startup.

KeerthiKumarR:
> ### 🟡 Changes recommended
> `shortcutsSlice.ts` still wraps i18n keys in a misleading local identity `t()` helper (instead of storing raw strings), and several CI checks are still in progress.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> _This review doesn't count toward merge requirements. [Sign up for the private preview](https://forms.cloud.microsoft/r/zLCqnkB1FJ) to control whether Copilot approvals count._
> 
> Review details
> ### Suppressed comments (1)
> **frontend/src/redux/shortcutsSlice.ts:38**
> 
> * `DEFAULT_SHORTCUTS` is meant to store plain i18n keys, but this file now introduces a local `t` identity helper and wraps every key with `t(...)`. This is misleading (looks like a real translation call) and makes it easier for future edits to accidentally reintroduce early i18next evaluation; consider storing raw string keys directly and removing the identity helper.
> 
> ```
> const t = (s: string) => s;
> ```
> 
> * **Files reviewed:** 2/2 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite
> 
> We're testing this review assessment. Please use 👍 or 👎 to tell us if it's correct.




The const t = (s: string) => s identity function is intentional as it serves as a marker so that i18next-parser recognizes these strings as translation keys during extraction npm run without it the scanner removes these keys from all local files and CI fails with failonupdate.

KeerthiKumarR:
> > ### 🟡 Changes recommended
> > `shortcutsSlice.ts` still wraps i18n keys in a misleading local identity `t()` helper (instead of storing raw strings), and several CI checks are still in progress.
> > _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> > _This review doesn't count toward merge requirements. [Sign up for the private preview](https://forms.cloud.microsoft/r/zLCqnkB1FJ) to control whether Copilot approvals count._
> > Review details
> > ### Suppressed comments (1)
> > **frontend/src/redux/shortcutsSlice.ts:38**
> > 
> > * `DEFAULT_SHORTCUTS` is meant to store plain i18n keys, but this file now introduces a local `t` identity helper and wraps every key with `t(...)`. This is misleading (looks like a real translation call) and makes it easier for future edits to accidentally reintroduce early i18next evaluation; consider storing raw string keys directly and removing the identity helper.
> > 
> > ```
> > const t = (s: string) => s;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * **Files reviewed:** 2/2 changed files
> > * **Comments generated:** 0 new
> > * **Review effort level:** Lite
> > 
> > We're testing this review assessment. Please use 👍 or 👎 to tell us if it's correct.
> 
> The const t = (s: string) => s identity function is intentional as it serves as a marker so that i18next-parser recognizes these strings as translation keys during extraction npm run without it the scanner removes these keys from all local files and CI fails with failonupdate.

@illume is this a right approach!

vyncent-t (commented):
Could you provide some testing instructions in the PR body and expand on the context of how this approach fixes this?

KeerthiKumarR:
@vyncent-t 
updated the PR description with the background context and steps to test and also put screenshots for the fix, Let me know if anything else is needed.

KeerthiKumarR:
@vyncent-t 

Before Fix

Before_Fix

After Fix

After_Fix

frontend/src/redux/shortcutsSlice.ts
```diff
@@ -35,40 +35,40 @@ export interface ShortcutConfig {
   category: 'navigation' | 'search' | 'general';
 }
 
-import i18next from 'i18next';
+const t = (s: string) => s;
```
unlikelyzero: This is the marker mentioned above. I ran the repo's own i18next-parser against a comment-based alternative (`// t('Global Search')` above the object, plain string literals below) and it extracts fine under the existing config with no changes needed - might be worth switching to that instead of shadowing `t` here.

unlikelyzero (commented):
Shadowing `t` with a no-op identity works, but I don't think it's the right call - I tested the repo's actual i18next-parser against it. A `// t('Global Search')` comment marker above `DEFAULT_SHORTCUTS` is the parser's documented workaround for this exact case, needs no config change, and doesn't leave a fake `t` in the file for someone to "fix" later by wiring in the real i18next and reintroducing this bug.

Also worth noting since the PR description undersells it: this fixes more than the console warning. Shortcut names were rendering in English in every locale before, since the old module-scope call never re-ran on language switch.

`ShortcutConfig.name`/`description` mean "translation key" now, not "display string," but the JSDoc still says otherwise - the codebase's existing `labelKey` convention (A8RInfo.tsx, ResourceTable.tsx) would make that clearer.

A few smaller things inline.

KeerthiKumarR:
> Shadowing `t` with a no-op identity works, but I don't think it's the right call - I tested the repo's actual i18next-parser against it. A `// t('Global Search')` comment marker above `DEFAULT_SHORTCUTS` is the parser's documented workaround for this exact case, needs no config change, and doesn't leave a fake `t` in the file for someone to "fix" later by wiring in the real i18next and reintroducing this bug.
> 
> Also worth noting since the PR description undersells it: this fixes more than the console warning. Shortcut names were rendering in English in every locale before, since the old module-scope call never re-ran on language switch.
> 
> `ShortcutConfig.name`/`description` mean "translation key" now, not "display string," but the JSDoc still says otherwise - the codebase's existing `labelKey` convention (A8RInfo.tsx, ResourceTable.tsx) would make that clearer.
> 
> A few smaller things inline.

@unlikelyzero Thanks for the review and for pointing out the documented workaround for i18next-parser
I have made all the changes and updated the PR PTAL!

illume (changes requested):
Thanks for the contribution.


<details>

</details>

illume (changes requested):
Thanks for working on this.


KeerthiKumarR:
@illume all the requested changes are made PTAL!

vyncent-t (approved):
Looks alright

illume (approved):
🎉 thanks!

PR #6815, opened 2026-08-02
frontend: clusterRequests: Reject with the real request failure

### LINKED ISSUE #6804: Intermittent "Unexpected end of input" / "Request timed-out" on testAuth (selfsubjectrulesreviews) despite server responding correctly, in proxy-auth + service-account-token mode
### Environment
- Headlamp version: v0.44.0 (`ghcr.io/headlamp-k8s/headlamp:v0.44.0`)
- Deployment mode: in-cluster, behind nginx ingress + oauth2-proxy (`auth_request`/`auth-signin`)
- Backend flags:
  ```
  -in-cluster
  -in-cluster-context-name=make-dev
  -session-ttl=86400
  -unsafe-use-service-account-token
  -base-url=/headlamp
  -proxy-auth
  -proxy-auth-username-header=X-Auth-Request-User
  -proxy-auth-email-header=X-Auth-Request-Email
  ```
- Ingress: ingress-nginx v1.12.1 (chart 4.12.1), `auth_request` pointed at an internal oauth2-proxy ClusterIP service
- Browser: Chrome/Edge 150 on macOS

### Description
After Headlamp has loaded and been used successfully for a while, the browser console intermittently logs:

```
Unable to parse error json at url: https://<host>/headlamp/clusters/<ctx>/apis/authorization.k8s.io/v1/selfsubjectrulesreviews
{err: SyntaxError: Unexpected end of input
    at ac (index-C6h6Guz1.js:868:3730)}
with request data: {signal: AbortSignal, credentials: 'include', headers: {…}, method: 'POST', body: '{"spec":{"namespace":"default"}}'}

Testing auth at authchooser
Requiring token for <ctx> as testing auth failed: Error: Request timed-out
```

This appears to originate from the frontend's `testAuth()` call (`frontend/src/lib/k8s/api/v1/clusterApi.ts`), which POSTs to `selfsubjectrulesreviews` with a hardcoded 5-second `AbortSignal` timeout. When it fails, the UI gets stuck / prompts for a token, even though the deployment uses `-unsafe-use-service-account-token` and doesn't require one.

### Evidence that this is not a server/network/proxy issue
We captured real ingress-nginx access log entries for this exact browser session and endpoint around the time these errors occur. Every single real request (not synthetic) returns cleanly:

```
"POST /headlamp/clusters/<ctx>/apis/authorization.k8s.io/v1/selfsubjectrulesreviews HTTP/2.0" 201 3992 ... 0.011 [headlamp-headlamp-80] [] 10.2.14.44:4466 3992 0.010 201
```

- Status: always `201`
- Body size: always `3992` bytes (a full, valid `SelfSubjectRulesReview` JSON body)
- Upstream response time: consistently `10–27ms`

We also independently verified oauth2-proxy's `/oauth2/auth` (used by nginx's `auth_request`) returns a well-formed `401 Unauthorized` with a 13-byte body (never empty) when unauthenticated, and confirmed cookie/session settings (`--cookie-refresh`, stable `cookie-secret`) rule out session-expiry as the trigger for this specific recurrence.

Given the server consistently returns a fast, complete, valid response, but the browser's `fetch()` still throws `SyntaxError: Unexpected end of input` (i.e. attempts to `JSON.parse()` an empty/truncated buffer) on an `AbortSignal`-carrying request, this looks like a client-side race — e.g. the frontend cancelling/superseding its own in-flight `testAuth()` call (possibly via re-render, route change, or repeated "authchooser" checks) while the response body is still being read, rather than a genuine network timeout.

### Impact
When this happens, the UI becomes stuck / prompts to re-authenticate ("Requiring token for `<ctx>`") even though the cluster connection and RBAC check both succeeded seconds earlier and the backend doesn't use per-user tokens at all (`-unsafe-use-service-account-token`). A full page reload is required to recover.

### Expected behavior
`testAuth()` should not misreport a successful, fast, valid API response as a failure, and/or should not enter a "requiring token" stuck state when using `-unsafe-use-service-account-token` + `-proxy-auth`, since no user-provided token is applicable in this mode.

### Reproduction difficulty
Intermittent / not reliably reproducible on demand — appears sporadically during normal use after the app has been open and used for a while. Happy to provide HAR captures or additional diagnostics if useful.

Anjali-Chauhan1:
## Summary

Fixes intermittent auth failures where a cluster that responded correctly was reported as timed out, alongside a misleading `SyntaxError: Unexpected end of input`.

`clusterRequest()` seeded its result with a placeholder response and only replaced it on abort:

```js
let response = new Response(undefined, { status: 502, statusText: 'Unreachable' });
try {
  response = await fetch(url, requestData);
} catch (err) {
  if (err.name === 'AbortError') {
    response = new Response(undefined, { status: 408, statusText: 'Request timed-out' });
  }
}
```

Both placeholders carry an **empty body**, and the failure path that follows reads that body to recover an error message — so every timed-out request ended in `Unexpected end of input`, logged against the request url as though the server had truncated its response. That reconciles the reporter's evidence: the ingress logs (`201`, 3992 bytes, 10–27ms) were correct, the response simply never reached the JS layer, and the code then parsed its own placeholder.

Two further defects in the same path:

- **Non-abort failures were discarded.** A `TypeError: Failed to fetch`, connection reset or TLS error left the placeholder 502 in place, so the real error was never logged or surfaced — which is what made this class undiagnosable.
- **A transient failure left the auth chooser demanding a token**, and the debug line claimed a token was required even on the branch that reports a connection error instead.

## Related Issue

Fixes #6804

## Changes

- **`clusterRequests.ts`** — track the failure instead of standing in a response for it, and reject with what actually happened. Same 408/502 status codes, so downstream consumers are unchanged. The url goes to the log, not the message, which is rendered in the auth chooser.
- **`authchooser/testAuthWithRetry.ts`** (new) — retry once on `408`/`502`/`504`. A check that never reached the cluster says nothing about whether a token is needed, so failing on the first attempt interrupts a working session. The retry uses a shorter timeout (3s vs 5s) so a hanging cluster still reports promptly.
- **`authchooser/index.tsx`** — the debug line now only claims a token is required on the branch that requires one.
- **`usesServiceAccountToken` per cluster** (`backend/cmd/cluster.go`, `headlamp.go`) — under `--unsafe-use-service-account-token` no user token is ever applied, so the chooser must not fall back to asking for one. Taken from the same helper the proxy uses to drop user credentials, so the two cannot disagree. Per cluster, since it is a property of the context.
- **Extracted `clusterMetadata()`** from `getClusters` to stay within the `funlen` limit.
- **Tests:** `clusterRequests.test.ts` (6), `testAuthWithRetry.test.ts` (9), `TestGetClustersUsesServiceAccountToken` (both directions).

## Steps to Test

1. `cd frontend && npx vitest run src/lib/k8s/api/v1/clusterRequests.test.ts src/components/authchooser/`
2. Confirm they are genuine regression tests — on `main` these fail:
   ```
   × keeps the original message when the network fails
     → expected 'Unreachable' to contain 'Failed to fetch'
   × does not log a body parse failure when there was no response
     → expected true to be false
   ```
3. `cd backend && go test ./cmd/ -run TestGetClustersUsesServiceAccountToken -v`
4. Against a live cluster, throttle or suspend the tab so the 5s abort fires during an auth check. Before: `Unable to parse error json … SyntaxError`. After: `Request timed-out after 5000ms at url: …`, and the check retries and recovers.
5. With `--unsafe-use-service-account-token`, confirm a failed auth check no longer routes to the token page.

## Notes for the Reviewer

- **The `frontend:` prefix undersells the commit** — it also touches `backend/cmd/` for the `usesServiceAccountToken` field. Happy to split that into its own `backend: cluster: …` commit.

- **The root trigger is not proven.** I believe the hardcoded 5s abort fires during background-tab timer throttling or an event-loop stall, which fits "intermittent, after the app has been open a while" — but it is not reliably reproducible and I could not demonstrate it. The retry recovers regardless of cause, which is why I chose it over simply raising the timeout.

- **Worth your judgement:** a genuinely unreachable cluster now takes up to 8s to report instead of 5s. I kept the retry budget shorter than the first attempt to bound it; retrying only on `408` is a one-line change if you disagree.

- **For the reporter:** a **Try Again** button already exists in that error state (wired to `runTestAuthAgain`), so "a full page reload is required" may not be accurate. If it genuinely was not visible, that is a separate bug.

- **Verification:** backend builds clean, `golangci-lint` 0 issues; frontend `tsc`, `ci-lint` (`--max-warnings 0`) and prettier clean; 2408 tests pass. Two pre-existing failures confirmed on a clean tree and unrelated: `pkg/spa`'s `TestURLToRelAcceptsSafeRelativePath` (Windows path separators) and some `ColorPicker`/`icons` tests that only time out under machine load.

## Summary by CodeRabbit

* **New Features**
  * Improved cluster authentication with service-account token awareness.
  * Added automatic retries for temporary connection failures.
  * Added clearer timeout and unreachable-cluster error reporting.

* **Bug Fixes**
  * Authentication failures are now distinguished more accurately from transient connection issues.
  * Improved cancellation and timeout handling for cluster requests.

* **Tests**
  * Added coverage for authentication retries, request failures, timeout handling, and token behavior.

### DIFF
--- a/backend/cmd/cluster.go
+++ b/backend/cmd/cluster.go
@@ -26,6 +26,9 @@ type Cluster struct {
 	AuthType string                 `json:"auth_type"`
 	Metadata map[string]interface{} `json:"meta_data"`
 	Error    string                 `json:"error,omitempty"`
+	// UsesServiceAccountToken reports that requests to this cluster are authenticated with
+	// the backend's own service account token, so a user-supplied token is never applied.
+	UsesServiceAccountToken bool `json:"usesServiceAccountToken,omitempty"`
 }
 
 type ClusterReq struct {
--- a/backend/cmd/cluster.go
+++ b/backend/cmd/cluster.go
@@ -26,6 +26,9 @@ type Cluster struct {
 	AuthType string                 `json:"auth_type"`
 	Metadata map[string]interface{} `json:"meta_data"`
 	Error    string                 `json:"error,omitempty"`
+	// UsesServiceAccountToken reports that requests to this cluster are authenticated with
+	// the backend's own service account token, so a user-supplied token is never applied.
+	UsesServiceAccountToken bool `json:"uses_service_account_token,omitempty"`
 }
 
 type ClusterReq struct {
--- a/backend/cmd/cluster.go
+++ b/backend/cmd/cluster.go
@@ -28,7 +28,7 @@ type Cluster struct {
 	Error    string                 `json:"error,omitempty"`
 	// UsesServiceAccountToken reports that requests to this cluster are authenticated with
 	// the backend's own service account token, so a user-supplied token is never applied.
-	UsesServiceAccountToken bool `json:"usesServiceAccountToken,omitempty"`
+	UsesServiceAccountToken bool `json:"uses_service_account_token,omitempty"`
 }
 
 type ClusterReq struct {
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -25,7 +25,6 @@ import { generatePath, useHistory, useLocation } from 'react-router-dom';
 import { getAppUrl } from '../../helpers/getAppUrl';
 import { getCluster, getClusterPrefixedPath } from '../../lib/cluster';
 import { useClustersConf } from '../../lib/k8s';
-import { testAuth } from '../../lib/k8s/api/v1/clusterApi';
 import { queryClient } from '../../lib/queryClient';
 import { createRouteURL } from '../../lib/router/createRouteURL';
 import { getRoute } from '../../lib/router/getRoute';
@@ -37,6 +36,7 @@ import Empty from '../common/EmptyContent';
 import Link from '../common/Link';
 import Loader from '../common/Loader';
 import OauthPopup from '../oidcauth/OauthPopup';
+import { testAuthWithRetry, TRANSIENT_STATUSES } from './testAuthWithRetry';
 
 function ColorButton({ children, ...rest }: ComponentProps<typeof Button>) {
   return (
@@ -106,7 +106,9 @@ function AuthChooser({ children }: AuthChooserProps) {
       // With clusterAuthType == oidc,
       //   they are presented with a choice of login or enter token.
       if (clusterAuthType !== 'oidc' && cluster.useToken === undefined) {
-        let useToken = true;
+        // When the backend authenticates with its own service account, a user token is
+        // never applied, so a failed check must not fall back to asking for one.
+        let useToken = !cluster.usesServiceAccountToken;
 
         setTestingAuth(true);
 
@@ -114,20 +116,24 @@ function AuthChooser({ children }: AuthChooserProps) {
 
         console.debug('Testing auth at authchooser');
 
-        testAuth(clusterName)
+        testAuthWithRetry(clusterName)
           .then(() => {
             console.debug('Not requiring token as testing auth succeeded');
             useToken = false;
           })
           .catch(err => {
             if (!cancelledRef.current) {
-              console.debug(`Requiring token for ${clusterName} as testing auth failed:`, err);
-
               // Ideally we'd only not assign the error if it was 401 or 403 (so we let the logic
               // proceed to request a token), but let's first check whether this is all we get
               // from clusters that require a token.
-              if ([408, 504, 502].includes(err.status)) {
+              if (TRANSIENT_STATUSES.includes(err.status)) {
+                // The request did not reach the cluster, so whether a token is needed is still
+                // unknown. Surfacing this as a connection error keeps it apart from the auth
+                // failures below, which do require one.
+                console.debug(`Could not reach ${clusterName} while testing auth:`, err);
                 errorObj = err;
+              } else {
+                console.debug(`Requiring token for ${clusterName} as testing auth failed:`, err);
               }
 
               setTestingAuth(false);
--- a/frontend/src/components/authchooser/index.tsx
+++ b/frontend/src/components/authchooser/index.tsx
@@ -132,6 +132,12 @@ function AuthChooser({ children }: AuthChooserProps) {
                 // failures below, which do require one.
                 console.debug(`Could not reach ${clusterName} while testing auth:`, err);
                 errorObj = err;
+              } else if (cluster.uses_service_account_token) {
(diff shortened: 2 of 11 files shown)


illume (changes requested):
Thanks for this PR.


backend/cmd/cluster.go
```diff
@@ -26,6 +26,9 @@ type Cluster struct {
 	AuthType string                 `json:"auth_type"`
 	Metadata map[string]interface{} `json:"meta_data"`
 	Error    string                 `json:"error,omitempty"`
+	// UsesServiceAccountToken reports that requests to this cluster are authenticated with
+	// the backend's own service account token, so a user-supplied token is never applied.
+	UsesServiceAccountToken bool `json:"usesServiceAccountToken,omitempty"`
```
illume: Please change this field in the previous commit?

In general each commit should be atomic, and new commits in the same PR should not modify the same code as a previous commit in the same PR.

frontend/src/components/authchooser/index.tsx
```diff
@@ -106,28 +106,34 @@ function AuthChooser({ children }: AuthChooserProps) {
       // With clusterAuthType == oidc,
       //   they are presented with a choice of login or enter token.
       if (clusterAuthType !== 'oidc' && cluster.useToken === undefined) {
-        let useToken = true;
+        // When the backend authenticates with its own service account, a user token is
```
illume: Can you please explain why in this comment?

illume (changes requested):
Thanks for this.

> frontend: clusterRequests: Reject with the real request failure
Can you please add backend to this commit subject?
> backend: frontend: k8s: authChooser: Reject with the real request failure

The headlamp repo doesn't use Signed-off-by, can you please remove?
> Signed-off-by:

Please add something to the git message body? Especially about the why.

backend/cmd/cluster.go
Anjali-Chauhan1: Thanks @illume for review. I have removed sign-offs , added a body to each commit, and the field name is fixed in the commit that adds it.
 I split the backend part into its own commit instead, so nothing spans both. Can squash it into one if you prefer. Rebased on main also done.

frontend/src/components/authchooser/index.tsx
Anjali-Chauhan1: I have expanded it now. When the backend runs with its own service account it throws away any token the user gives, so asking for one would not help.

PR #6817, opened 2026-08-02
fix: parse milli-quantities in parseUnitsOfBytes

### LINKED ISSUE #6816: Bug: usage percentages wrong (~1000x) when metrics API emits milli quantities (e.g. prometheus-adapter)
## Summary

Cluster Overview and Pod List CPU/Memory columns display incorrect memory usage when the Metrics API is provided by **prometheus-adapter** (the standard `metrics.k8s.io` implementation backed by Prometheus, commonly used in OpenShift-style environments).

Memory usage is interpreted as approximately **1000× larger** than the actual value, resulting in incorrect percentages.

### Examples
Image
Image

The values occasionally correct themselves during the next metrics refresh, then become incorrect again.

---

## Why it happens

`prometheus-adapter` alternates between two valid Kubernetes quantity formats for the same memory value.

Example responses from the Metrics API:

```yaml
usage:
  memory: "2655456Ki"
```

and

```yaml
usage:
  memory: "2728711509333m"
```

Both represent approximately **2.5 GiB**.

`kubectl top` correctly parses both formats because it uses the standard Kubernetes quantity parser.

Headlamp, however, only parses the first format correctly.

---

## Steps to Reproduce

1. Install **prometheus-adapter** (Helm chart) configured to serve `metrics.k8s.io` using an existing Prometheus instance (`resourceRules` enabled).
2. Use a cluster without `metrics-server`.
3. Open **Cluster Overview** or **Workloads → Pods**.
4. Observe the CPU/Memory usage.
5. During periodic metrics refreshes, memory values alternate between:
   - correct values
   - approximately **1000× larger** values.

---

## Expected Behavior

Headlamp should correctly parse **all valid Kubernetes quantity formats**, including milli-byte (`m`) values, so memory usage and percentages remain consistent regardless of which notation the Metrics API returns.

---

## Environment

- **Headlamp:** 0.29.x / current `main`
- **Kubernetes:** v1.36
- **Metrics provider:** `prometheus-community/prometheus-adapter`
- **Cluster:** Docker Desktop Kubernetes
- **Metrics API:** `metrics.k8s.io/v1beta1`

Pratikdeb07:
## Summary

This PR fixes an issue where memory usage percentages are displayed approximately **1000× larger than expected** when the Metrics API returns memory quantities in **milli-units** (for example, `2687146666666m`).

This occurs with **prometheus-adapter**, which commonly serves the `metrics.k8s.io` API in Prometheus-backed Kubernetes clusters that do not use `metrics-server`.

## Related Issue

Fixes #6816

## Changes

- Added support for the Kubernetes **`m` (milli)** suffix in `parseUnitsOfBytes()` (`frontend/src/lib/units.ts`).
- Parsed milli-byte values as `parseFloat(value) / 1000`, matching the existing handling in `parseCpu()`.
- Added unit tests covering milli-byte quantities, including:
  - `1000m` → `1`
  - `2687146666666m` → `2687146666.666`
  - `1.5m` → `0.0015`

## Steps to Test

1. Set up a cluster where the Metrics API is provided by **prometheus-adapter** (with `resourceRules` enabled) instead of `metrics-server`.
2. Open **Cluster Overview**.
   - Verify memory usage displays approximately **32.8%** (about **2.5 GiB / 7.63 GiB**) instead of **32800%**.
3. Open **Workloads → Pods**.
   - Verify the CPU/Memory columns and the **"Request: X%"** tooltips display correct values.
4. Observe the UI for a few metrics refresh cycles.
   - Memory usage should remain stable even when the metrics provider alternates between `Ki` and `m` quantity formats.

## Screenshots

**Before**
Images are present in issue 
**After**
image

## Notes for Reviewers

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory quantity parsing for Kubernetes milli-unit values.
  * Correctly handles whole, large, and fractional milli-quantities by converting them to base units.

### DIFF
--- a/frontend/src/lib/units.test.ts
+++ b/frontend/src/lib/units.test.ts
@@ -53,6 +53,12 @@ describe('parseRam', () => {
     expect(parseRam('0.5G')).toBe(500000000);
   });
 
+  it('should parse milli quantities (e.g. from prometheus-adapter)', () => {
+    expect(parseRam('1000m')).toBe(1);
+    expect(parseRam('2687146666666m')).toBeCloseTo(2687146666.666, 1);
+    expect(parseRam('1.5m')).toBe(0.0015);
+  });
+
   it('should scale binary units by powers of 1024', () => {
     fc.assert(
       fc.property(fc.integer({ min: 1, max: 100 }), num => {
--- a/frontend/src/lib/units.ts
+++ b/frontend/src/lib/units.ts
@@ -38,6 +38,13 @@ export function parseRam(value: string) {
 function parseUnitsOfBytes(value: string): number {
   if (!value) return 0;
 
+  // Milli-quantity ex. "2687146666666m". Valid Kubernetes quantity syntax
+  // (e.g. emitted by prometheus-adapter for memory usage), but no unit is
+  // matched by the regex below, which would misread it as raw bytes.
+  if (value.endsWith('m')) {
+    return parseFloat(value) / 1000;
+  }
+
   const groups = value.match(/(\d+(?:\.\d+)?)([BKMGTPEe])?(i)?(\d+)?/) || [];
   const number = parseFloat(groups[1]);
 


illume (changes requested):
Thanks for the contribution.


<details>

- `fix: parse milli-quantities in parseUnitsOfBytes` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

Pratikdeb07:
Thanks for your review @illume ! I have updated the commit message.

unlikelyzero (commented):
This only handles the 'm' suffix, but Kubernetes decimalSI also has 'u' (micro) and 'n' (nano) on the same family - same bug class, just less common. parseCpu right below in this same file already handles the full n/u/m ladder, and util.ts's normalizeUnit already handles all three for memory too, so the pattern to copy is already sitting in the codebase.

That duplication isn't theoretical either - resourceQuota/Details.tsx renders the same table row using both parsers: the Used/Hard columns go through normalizeUnit, and the Usage bar in that same row goes through parseRam/parseUnitsOfBytes. Before this PR they disagreed 1000x on milli values in that row; after it they agree on 'm' but still disagree on 'u'/'n' and on lowercase 'k' (which is the canonical k8s suffix - the regex here only matches uppercase 'K', so '1k' returns 1 instead of 1000, same 1000x bug in the same function).

A few smaller things inline.

frontend/src/lib/units.ts
```diff
@@ -38,6 +38,13 @@ export function parseRam(value: string) {
 function parseUnitsOfBytes(value: string): number {
   if (!value) return 0;
 
+  // Milli-quantity ex. "2687146666666m". Valid Kubernetes quantity syntax
+  // (e.g. emitted by prometheus-adapter for memory usage), but no unit is
+  // matched by the regex below, which would misread it as raw bytes.
+  if (value.endsWith('m')) {
```
unlikelyzero, line 44: Would extend this to cover 'u' and 'n' too, mirroring parseCpu a few lines down: `/1000` for m, `/1e6` for u, `/1e9` for n. Same bug, just needs two more branches.

frontend/src/lib/units.ts
```diff
@@ -38,6 +38,13 @@ export function parseRam(value: string) {
...
+    return parseFloat(value) / 1000;
+  }
+
   const groups = value.match(/(\d+(?:\.\d+)?)([BKMGTPEe])?(i)?(\d+)?/) || [];
```
unlikelyzero, line 48: This character class only matches uppercase 'K', but the canonical Kubernetes suffix is lowercase 'k' (unlike M/G/T/P/E which are uppercase) - util.ts's normalizeUnit checks endsWith('k') for exactly this reason. Worth fixing while you're in here, since it's the same 1000x-off bug for the one suffix that happens to be lowercase.

frontend/src/lib/units.test.ts
```diff
@@ -53,6 +53,12 @@ describe('parseRam', () => {
     expect(parseRam('0.5G')).toBe(500000000);
   });
 
+  it('should parse milli quantities (e.g. from prometheus-adapter)', () => {
```
unlikelyzero, line 56: These test parseRam directly, but the actual reported bug is a percentage calculation through divideK8sResources - there's already a describe block for that with a comparable CPU nano/micro case a bit further down that this could extend with a memory-milli one.

illume (changes requested):
Thanks for working on this.


Looks like there are git conflicts in this PR. Can you fix them up?

<details>

</details>

PR #6755, opened 2026-07-30
Fix/validate backstage postmessage origin

vanshika2720:
## Summary

This PR fixes a security issue in the Backstage integration by validating the origin of incoming `window.postMessage` events before processing authentication messages. It also replaces wildcard (`"*"`) target origins with explicitly trusted origins when sending messages back to the parent window.

Additionally, the `HEADLAMP_READY` message sent from the Home page is now posted only to a configured trusted origin.

## Related Issue

Fixes #6754, #6757

## Changes

* Added trusted origin validation for incoming Backstage `postMessage` events.
* Introduced helper functions to determine trusted origins from `REACT_APP_BACKSTAGE_ORIGIN`.
* Ignored messages received from untrusted origins before processing authentication tokens or kubeconfig data.
* Replaced wildcard (`"*"` ) target origins with the validated `event.origin` when sending `BACKSTAGE_AUTH_TOKEN_ACK`.
* Updated the `HEADLAMP_READY` message to target only the configured trusted origin.
* Added comprehensive unit tests covering:

  * trusted origin acceptance,
  * untrusted origin rejection,
  * prevention of token storage from untrusted origins,
  * ACK messages sent only to validated origins,
  * preservation of existing behavior for valid messages, including kubeconfig storage.

## Steps to Test

1. Configure `REACT_APP_BACKSTAGE_ORIGIN` with a trusted Backstage origin.
2. Run the frontend and send a valid `BACKSTAGE_AUTH_TOKEN` message from the configured origin.
3. Verify that the authentication token and kubeconfig (when provided) are processed correctly.
4. Send the same message from an untrusted origin.
5. Verify that the message is ignored and no authentication token or kubeconfig is stored.
6. Confirm that `BACKSTAGE_AUTH_TOKEN_ACK` is sent only to the validated sender origin.
7. Verify that `HEADLAMP_READY` is sent only when a trusted origin is configured.

## Notes for the Reviewer

* This change is security-focused and preserves existing behavior for trusted Backstage integrations.
* The implementation affects only the Backstage `postMessage` communication flow and the `HEADLAMP_READY` message.
* Added unit tests to cover both trusted and untrusted origin scenarios and ensure existing functionality remains unchanged.

PR #6818, opened 2026-08-03
frontend: backstageMessageReceiver: Validate message payloads

### LINKED ISSUE #6757: Security: Validate Backstage postMessage payloads before processing
## Describe the bug

The Backstage message receiver processes incoming `window.postMessage` events by casting `event.data` to `BackstageMessage` using a compile-time type assertion (`as BackstageMessage`) without verifying the runtime structure of the received message.

As a result, malformed messages with incorrect field types (for example, a numeric `token`) may still be processed and update application state through implicit JavaScript type coercion.

Incoming messages should be validated at runtime before they are processed.

Incoming messages should be validated at runtime before any application state is updated.

### Expected behavior

Before processing any Backstage message:

- Verify that `event.data` is a non-null object.
- Verify that the message `type` is one of the supported values.
- Verify that `payload`, when present, is an object.
- Validate the expected fields for each supported message type (for example, `token` and `kubeconfig` should be strings when provided).
- Ignore malformed or unexpected messages without throwing errors or updating application state.

## To Reproduce

1. Enable the Backstage integration.
2. Send a `window.postMessage` event from a trusted origin with:
   - a non-object `event.data`,
   - a non-object `payload`,
   - an unsupported message type,
   - a `BACKSTAGE_AUTH_TOKEN` message whose `token` is not a string, or
   - a kubeconfig payload with invalid field types.
3. Observe that the message is accepted based on the unchecked runtime type assertion instead of being rejected before processing.

## Environment

- Installation type: Any
- Headlamp Version: Current `main`
- Browser: Any modern browser

Yes. I would like to work on this issue and submit a PR.

## Additional Context

The recent origin validation improvements ensure that only trusted origins can communicate with Headlamp. However, the structure of trusted messages is still assumed rather than verified at runtime.

This issue proposes adding lightweight runtime validation to ensure that only well-formed Backstage messages are processed. This provides defense in depth, prevents runtime type confusion, and enforces the expected message contract instead of relying solely on TypeScript type assertions.

Possible implementation approaches include:

- Runtime type guards.
- A shared message validation helper.
- Validation of supported message types and payload shapes without introducing new dependencies.

Suggested tests:

- Reject `null` or non-object `event.data`.
- Reject `null` or non-object payloads.
- Reject unknown message types.
- Reject messages with missing required fields.
- Reject messages with invalid field types (for example, a non-string `token` or `kubeconfig`).
- Verify that valid messages continue to behave as expected.

vanshika2720:
## Summary

The Backstage message receiver currently relies on a compile-time type
assertion (`as BackstageMessage`) when processing incoming `window.postMessage`
events. Because the message shape is not verified at runtime, malformed payloads
with incorrect field types (for example, a numeric `token`) can still be
processed and implicitly coerced before being stored.

This PR introduces lightweight runtime type guards to validate incoming
Backstage messages before processing them, ensuring that only well-formed
messages update application state while preserving existing behavior for valid
messages.

## Related Issue

Fixes #6757

## Changes

- Added runtime validation for incoming Backstage messages.
- Validated supported message types.
- Validated payload structure and expected field types.
- Ignored malformed messages before any state updates.
- Added unit tests covering valid and invalid payload scenarios.

## Steps to Test

1. Run `npm run frontend:test`.
2. Send a valid `BACKSTAGE_AUTH_TOKEN` message and verify existing behavior.
3. Send malformed messages, including:
   - non-object `event.data`
   - non-object `payload`
   - unknown message types
   - non-string `token`
   - non-string `kubeconfig`
4. Verify malformed messages are ignored and no application state is updated.

## Notes for the Reviewer

- Introduces lightweight runtime type guards only.
- Adds no new dependencies.
- Preserves existing behavior for valid Backstage messages.
- Enforces the runtime contract instead of relying solely on the compile-time
  `BackstageMessage` type assertion.

### DIFF
--- a/frontend/src/helpers/backstageMessageReceiver.test.ts
+++ b/frontend/src/helpers/backstageMessageReceiver.test.ts
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import jsyaml from 'js-yaml';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import * as statelessFunctions from '../stateless';
+import { getBackstageToken, setupBackstageMessageReceiver } from './backstageMessageReceiver';
+import { encodeBase64 } from './base64';
+
+vi.mock('./isBackstage', () => ({
+  isBackstage: () => true,
+}));
+
+function dispatchMessage(data: unknown) {
+  window.dispatchEvent(
+    new MessageEvent('message', { data, origin: 'https://backstage.example.com' })
+  );
+}
+
+describe('setupBackstageMessageReceiver', () => {
+  let cleanup: () => void;
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.spyOn(window.parent, 'postMessage').mockImplementation(() => {});
+    cleanup = setupBackstageMessageReceiver();
+  });
+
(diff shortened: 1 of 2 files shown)


skoeva (commented):
thanks for this work. we verified that #6755 now contains the same payload-validation implementation and test scenarios, in addition to origin validation. to avoid duplicate changes to the same handler and test file, we're consolidating both fixes in #6755 and closing this PR as superseded. we'll link #6755 to #6757 as well

PR #6828, opened 2026-08-03
upstream 0002: app: plugins: Support default-disabled bundled plugins

illume:
## Base

- https://github.com/kubernetes-sigs/headlamp/pull/6827 is merged.
- Rebased directly onto upstream `main` at `f69eae324993535cc0bcd85127357ec8e347e199`.
- The PR contains only the 6-commit, 8-file `0002` change.

## Summary

- preserve an explicit boolean `enabledByDefault` value from the build manifest in bundled plugin package metadata
- apply explicit defaults atomically and fail packaging when declared metadata cannot be applied
- use only strict `false` defaults for newly discovered shipped plugins; user and development plugins remain enabled
- preserve all existing settings, including entries automatically persisted by earlier releases
- share the Headlamp package metadata type between plugin discovery and i18n
- add focused app/frontend unit coverage and a shipped-plugin Electron end-to-end test

- Add unit and end-to-end characterization before implementation so default behavior and saved choices are reviewable independently.
- Validate and atomically replace package metadata so interrupted builds cannot ship a disabled plugin as enabled or leave truncated JSON.
- Apply explicit defaults only to newly discovered shipped plugins so user-installed plugins remain enabled and existing choices keep precedence.
- Treat missing type metadata as shipped and accept only strict boolean `false`, which fails closed without interpreting malformed values.
- Share package metadata types between discovery and i18n parsing to prevent the on-disk schema declarations from drifting.
- Use stable row, checkbox, and shipped-resource assertions so browser and Electron coverage does not depend on locale text or private fixtures.

## Source

This upstreams Azure/aks-desktop commit [`4a4e949d`](https://github.com/Azure/aks-desktop/commit/4a4e949ddb4be6c0df1b84f4a5b779541b74209d), originally authored by Joaquim Rocha on 2025-11-07. The same authored change is retained as [`85eb94ac`](https://github.com/Azure/aks-desktop/commit/85eb94ac8502999159d8b5250737213a8b28339c), followed by the default-disabled Kaito consumer in [`1eda1461`](https://github.com/Azure/aks-desktop/commit/1eda1461009a406ae4ce8d06a2d146c505307e51). GitHub reports no directly associated Azure PR for those commits. The current retained patch is part of [Azure/aks-desktop#823](https://github.com/Azure/aks-desktop/pull/823).

## Testing

- `npm --prefix app test` (19 files, 313 tests passed)
- `npm --prefix app test -- --run electron/build-manifest.test.ts` (73 passed)
- `npm --prefix frontend test -- --run src/plugin/updateSettingsPackages.test.ts` (10 passed)
- `npm run frontend:lint`
- `npm --prefix app run tsc`
- `npm --prefix frontend run build`
- Electron Playwright `pluginDefaultEnablement.spec.ts` with a shipped plugin fixture (1 passed)
- `git diff --check`

Assisted by copilot.

### DIFF
--- a/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
+++ b/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect, test } from '@playwright/test';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { _electron, ElectronApplication, Page } from 'playwright';
+
+const appPath = path.resolve(__dirname, '../../');
+const electronExecutable = process.platform === 'win32' ? 'electron.cmd' : 'electron';
+const electronPath = path.resolve(__dirname, `../../node_modules/.bin/${electronExecutable}`);
+const electronDistPath = path.join(appPath, 'node_modules', 'electron', 'dist');
+const electronResourcesPath =
+  process.platform === 'darwin'
+    ? path.join(electronDistPath, 'Electron.app', 'Contents', 'Resources')
+    : path.join(electronDistPath, 'resources');
+const pluginName = 'default-disabled-e2e';
+
+let electronApp: ElectronApplication;
+let electronPage: Page;
+let pluginPath: string;
+let userDataPath: string;
+
+test.describe('plugin default enablement', () => {
+  test.setTimeout(2 * 60 * 1000);
+
+  test.beforeEach(() => {
+    test.skip(process.env.PLAYWRIGHT_TEST_MODE !== 'app', 'This test only runs in app mode');
+  });
+
+  test.beforeAll(async ({}, testInfo) => {
+    testInfo.setTimeout(2 * 60 * 1000);
+
+    if (process.env.PLAYWRIGHT_TEST_MODE !== 'app') {
+      return;
+    }
+
+    pluginPath = path.join(electronResourcesPath, '.plugins', pluginName);
+    fs.mkdirSync(pluginPath, { recursive: true });
+    fs.writeFileSync(path.join(pluginPath, 'main.js'), '');
+    fs.writeFileSync(
+      path.join(pluginPath, 'package.json'),
+      JSON.stringify({
+        name: pluginName,
+        version: '1.0.0',
+        description: 'Default-disabled e2e fixture',
+        homepage: 'https://example.com/default-disabled-e2e',
+        isManagedByHeadlampPlugin: true,
+        artifacthub: {
+          title: 'Default-disabled e2e fixture',
+          url: 'https://example.com/default-disabled-e2e',
+          repoName: 'e2e',
+          author: 'Headlamp',
+          version: '1.0.0',
+        },
+        devDependencies: {
+          '@kinvolk/headlamp-plugin': '^0.8.0',
+        },
+        headlamp: {
+          enabledByDefault: false,
+        },
+      })
+    );
+
+    userDataPath = fs.mkdtempSync(path.join(os.tmpdir(), 'headlamp-plugin-default-e2e-'));
+    electronApp = await _electron.launch({
+      cwd: appPath,
+      executablePath: electronPath,
+      args: ['.', `--user-data-dir=${userDataPath}`],
+      env: {
+        ...process.env,
+        ELECTRON_DEV: 'true',
+        NODE_ENV: 'development',
+      },
+    });
+    electronPage = await electronApp.firstWindow();
+  });
+
+  test.afterAll(async () => {
+    await electronApp?.close();
+    if (pluginPath) {
+      fs.rmSync(pluginPath, { force: true, recursive: true });
+    }
+    if (userDataPath) {
+      fs.rmSync(userDataPath, { force: true, recursive: true });
+    }
+  });
+
+  test('starts a newly discovered default-disabled plugin disabled', async () => {
+    await electronPage.waitForLoadState('load');
+    const baseUrl = electronPage.url().split('#')[0];
+    await electronPage.goto(`${baseUrl}#/settings/plugins`);
+
+    const pluginRow = electronPage.getByRole('row').filter({ hasText: pluginName });
+    const pluginToggle = pluginRow.getByRole('checkbox');
+    await expect(pluginToggle).toBeVisible();
+    await expect(pluginToggle).not.toBeChecked();
+  });
+});
--- a/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
+++ b/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
@@ -104,7 +104,6 @@ test.describe('plugin default enablement', () => {
 
     const pluginRow = electronPage.getByRole('row').filter({ hasText: pluginName });
     const pluginToggle = pluginRow.getByRole('checkbox');
-    await expect(pluginRow).toContainText('Disabled');
     await expect(pluginToggle).toBeVisible();
     await expect(pluginToggle).not.toBeChecked();
   });
--- a/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
+++ b/app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2026 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect, test } from '@playwright/test';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { _electron, ElectronApplication, Page } from 'playwright';
+
+const appPath = path.resolve(__dirname, '../../');
+const electronExecutable = process.platform === 'win32' ? 'electron.cmd' : 'electron';
+const electronPath = path.resolve(__dirname, `../../node_modules/.bin/${electronExecutable}`);
+const pluginName = 'default-disabled-e2e';
+
+let electronApp: ElectronApplication;
+let electronPage: Page;
+let pluginPath: string;
+let pluginsPath: string;
+let userDataPath: string;
+
+test.describe('plugin default enablement', () => {
+  test.setTimeout(2 * 60 * 1000);
+
+  test.beforeEach(() => {
+    test.skip(process.env.PLAYWRIGHT_TEST_MODE !== 'app', 'This test only runs in app mode');
+  });
+
+  test.beforeAll(async ({}, testInfo) => {
+    testInfo.setTimeout(2 * 60 * 1000);
+
+    if (process.env.PLAYWRIGHT_TEST_MODE !== 'app') {
+      return;
+    }
+
+    pluginsPath = fs.mkdtempSync(path.join(os.tmpdir(), 'headlamp-plugins-e2e-'));
+    pluginPath = path.join(pluginsPath, pluginName);
+    fs.mkdirSync(pluginPath, { recursive: true });
+    fs.writeFileSync(path.join(pluginPath, 'main.js'), '');
+    fs.writeFileSync(
+      path.join(pluginPath, 'package.json'),
+      JSON.stringify({
+        name: pluginName,
+        version: '1.0.0',
+        description: 'Default-disabled e2e fixture',
+        homepage: 'https://example.com/default-disabled-e2e',
+        isManagedByHeadlampPlugin: true,
+        artifacthub: {
+          title: 'Default-disabled e2e fixture',
+          url: 'https://example.com/default-disabled-e2e',
+          repoName: 'e2e',
+          author: 'Headlamp',
+          version: '1.0.0',
+        },
+        devDependencies: {
+          '@kinvolk/headlamp-plugin': '^0.8.0',
+        },
+        headlamp: {
+          enabledByDefault: false,
+        },
+      })
+    );
+
+    userDataPath = fs.mkdtempSync(path.join(os.tmpdir(), 'headlamp-plugin-default-e2e-'));
+    electronApp = await _electron.launch({
+      cwd: appPath,
+      executablePath: electronPath,
+      args: ['.', `--user-data-dir=${userDataPath}`],
+      env: {
+        ...process.env,
+        ELECTRON_DEV: 'true',
+        HEADLAMP_CONFIG_PLUGINS_DIR: pluginsPath,
+        NODE_ENV: 'development',
+      },
+    });
+    electronPage = await electronApp.firstWindow();
+  });
+
+  test.afterAll(async () => {
+    await electronApp?.close();
+    if (pluginsPath) {
+      fs.rmSync(pluginsPath, { force: true, recursive: true });
+    }
+    if (userDataPath) {
+      fs.rmSync(userDataPath, { force: true, recursive: true });
+    }
+  });
+
+  test('starts a newly discovered default-disabled plugin disabled', async () => {
+    const baseUrl = electronPage.url().split('#')[0];
+    await electronPage.goto(`${baseUrl}#/settings/plugins`);
+
+    const pluginRow = electronPage.getByRole('row').filter({ hasText: pluginName });
+    const pluginToggle = pluginRow.getByRole('checkbox');
+    await expect(pluginRow).toContainText('Disabled');
+    await expect(pluginToggle).toBeVisible();
+    await expect(pluginToggle).not.toBeChecked();
+  });
+});
--- a/app/electron/build-manifest.test.ts
+++ b/app/electron/build-manifest.test.ts
@@ -30,6 +30,7 @@ import {
   resolveBuildManifestPath,
 } from '../scripts/build-manifest.ts';
 import {
+  applyEnabledByDefault,
   downloadFile,
   extractArchive,
   getArchiveFileName,
@@ -465,6 +466,19 @@ describe('plugin archive integrity', () => {
     ).toThrow('Invalid SHA-256');
   });
 
+  it('rejects non-boolean default enabled states', () => {
+    expect(() =>
+      validatePluginSource(
+        {
+          name: 'example',
+          file: './plugin.tar.gz',
+          enabledByDefault: 'false' as unknown as boolean,
+        },
+        false
+      )
+    ).toThrow('enabledByDefault must be a boolean');
+  });
+
   it('accepts matching digests and manifests without digests', () => {
     const archive = temporaryFile('plugin archive');
     const digest = crypto.createHash('sha256').update('plugin archive').digest('hex');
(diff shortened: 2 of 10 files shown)


app/scripts/setup-plugins.js
```diff
@@ -148,13 +147,27 @@ async function main() {
...
+        console.log(`Plugin ${name} enabledByDefault: ${enabledByDefault}`);
+      }
     }
   }
+}
 
-  process.exit(0);
+if (require.main === module) {
+  main().then(() => process.exit(0));
 }
```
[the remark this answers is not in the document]
illume: Addressed in e2fb47e2e: the CLI now handles both success and rejection explicitly, logs setup failures, and exits with status 1. Added a regression test for the rejected setup path.

app/scripts/setup-plugins.js
```diff
@@ -148,13 +147,27 @@ async function main() {
...
+      const packageJsonPath = path.join(pluginFolder, 'package.json');
+
+      if (enabledByDefault !== undefined && fs.existsSync(packageJsonPath)) {
+        const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+        packageJson.headlamp = packageJson.headlamp || {};
+        packageJson.headlamp.enabledByDefault = enabledByDefault;
+
+        fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
+        console.log(`Plugin ${name} enabledByDefault: ${enabledByDefault}`);
+      }
```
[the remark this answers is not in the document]
illume: Addressed in e2fb47e2e: package metadata read/write is isolated per plugin, errors are logged, and setup continues with later plugins. The new test uses malformed JSON followed by a valid plugin and verifies the latter is updated.

e2e-tests/tests/pluginSetting.spec.ts
```diff
@@ -38,10 +38,12 @@ test('plugin settings page should have a table', async () => {
   await headlampPage.a11y();
 });
 
-test('pod counter plugin should have setting option', async () => {
+test('pod counter plugin should have setting option', async ({ page }) => {
   const pluginName = 'headlamp-pod-counter';
 
   await headlampPage.navigateTopage('/settings/plugins', /Plugin/);
+  const pluginRow = page.getByRole('row').filter({ hasText: pluginName });
+  await expect(pluginRow).toContainText('Enabled');
```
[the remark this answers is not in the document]
illume: Addressed in 34b65b383: replaced the incorrect localized Enabled assertion with a stable row visibility check. This is also the cause of the failing container e2e job.

app/e2e-tests/tests/pluginDefaultEnablement.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  test('starts a newly discovered default-disabled plugin disabled', async () => {
+    const baseUrl = electronPage.url().split('#')[0];
+    await electronPage.goto(`${baseUrl}#/settings/plugins`);
+
+    const pluginRow = electronPage.getByRole('row').filter({ hasText: pluginName });
+    const pluginToggle = pluginRow.getByRole('checkbox');
+    await expect(pluginRow).toContainText('Disabled');
+    await expect(pluginToggle).toBeVisible();
+    await expect(pluginToggle).not.toBeChecked();
```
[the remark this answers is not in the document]
illume: Addressed in 34b65b383: removed the localized Disabled text assertion. The Electron test now relies on the visible unchecked checkbox and passes end to end locally.

frontend/src/plugin/index.ts
```diff
@@ -315,10 +315,9 @@ export function updateSettingsPackages(
     const index = settingsPlugins.findIndex(x => x.name === plugin.name && x.type === plugin.type);
 
     if (index === -1) {
-      // It's a new one settings doesn't know about, enable it by default
       return {
         ...plugin,
-        isEnabled: true,
+        isEnabled: plugin.headlamp?.enabledByDefault ?? true,
       };
```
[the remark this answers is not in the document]
illume: Addressed in 32f94b871: updated the JSDoc to describe headlamp.enabledByDefault with the true fallback.

app/scripts/setup-plugins.js
```diff
@@ -148,13 +147,41 @@ async function main() {
...
+        try {
+          const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+          packageJson.headlamp = packageJson.headlamp || {};
+          packageJson.headlamp.enabledByDefault = enabledByDefault;
+
+          fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
+          console.log(`Plugin ${name} enabledByDefault: ${enabledByDefault}`);
+        } catch (error) {
+          console.error(`Failed to update enabledByDefault for plugin ${name}:`, error);
+        }
```
unlikelyzero: This whole block is fail-open in a way that undercuts what the PR is for. If package.json wasn't extracted (existsSync is false) or the parse/write throws, we just log and move on - main() still resolves and runCli exits 0. So a plugin the manifest explicitly marks enabledByDefault:false can end up shipped enabled, with a green build. I think at least the false case needs to be fatal, since that's the direction with a real user-facing consequence, and the existing test that asserts "skips a manifest default when the plugin wasn't extracted" is effectively locking in the silent-success behavior rather than guarding against it.

app/scripts/setup-plugins.js
```diff
@@ -148,13 +147,41 @@ async function main() {
...
+      const pluginFolder = path.join(PLUGIN_FOLDER, name);
+      const packageJsonPath = path.join(pluginFolder, 'package.json');
+
+      if (enabledByDefault !== undefined && fs.existsSync(packageJsonPath)) {
+        try {
+          const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+          packageJson.headlamp = packageJson.headlamp || {};
+          packageJson.headlamp.enabledByDefault = enabledByDefault;
+
+          fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
```
unlikelyzero: This write isn't atomic - if the process dies mid-write we can leave a truncated package.json in the shipped `.plugins` tree. That matters more than it would look because the frontend fetches every plugin's package.json inside one `Promise.all` with `resp.json()` (frontend/src/plugin/index.ts around line 464) - one bad file rejects the whole batch and no plugins load at all, not just the one that got corrupted. I'd write to a temp path and rename into place.

app/scripts/setup-plugins.js
```diff
@@ -148,13 +147,41 @@ async function main() {
...
+      }
     }
   }
+}
+
+function runCli(setup = main, exit = process.exit) {
+  return setup().then(
+    () => exit(0),
+    error => {
+      console.error('Failed to set up plugins:', error);
```
unlikelyzero: Minor, but worth noting since this handler is new: the most likely rejection here is `extractArchive`'s "failed to find plugin content" path, which calls `reject()` with no argument. So in that case this prints `Failed to set up plugins: undefined` and drops the actually useful message that was logged two lines earlier in extractArchive. Wrapping that reject in a real Error would make this handler worth what it cost to add.

frontend/src/plugin/index.ts
```diff
@@ -315,10 +315,9 @@ export function updateSettingsPackages(
     const index = settingsPlugins.findIndex(x => x.name === plugin.name && x.type === plugin.type);
 
     if (index === -1) {
-      // It's a new one settings doesn't know about, enable it by default
       return {
         ...plugin,
-        isEnabled: true,
+        isEnabled: plugin.headlamp?.enabledByDefault ?? true,
```
unlikelyzero: Two things here worth thinking about:

1. "Not in settings" is the only signal for "newly discovered," but that settings entry gets auto-created and persisted the first time any plugin is seen (fetchAndExecutePlugins -> onSettingsChange -> localStorage). So if a maintainer changes a shipped plugin's default in a later release, every user who's ever launched the app before already has a persisted isEnabled entry and won't see the new default - only fresh installs will. That's a pretty different guarantee than "preserves the user's saved choice" implies, since the user never made a choice in that case.

2. This runs for every plugin type, not just bundled/shipped ones. A catalog-installed plugin whose own package.json happens to declare `headlamp.enabledByDefault: false` would load disabled right after a user explicitly installed it, with nothing in the UI explaining why.

Neither is a bug exactly, but I think both are worth deciding on purpose and writing into the docstring rather than leaving as emergent behavior.

unlikelyzero: Separately: `enabledByDefault` isn't validated as a boolean anywhere along the chain from manifest JSON to here. `??` only treats null/undefined as "use default," so something like the string `"false"` would be truthy and silently enable a plugin meant to ship disabled. Might be worth a stricter check, e.g. `=== false ? false : true`, instead of relying on `??`.

frontend/src/plugin/pluginsSlice.ts
```diff
@@ -104,6 +104,19 @@ export type PluginInfo = {
    */
   isCompatible?: boolean;
 
+  /**
+   * headlamp is an object containing Headlamp-specific plugin configuration
+   */
+  headlamp?: {
```
unlikelyzero: This `headlamp` shape duplicates `PluginPackageInfo.headlamp` in pluginI18n.ts, which already has `i18n?: string[]`. Now the same on-disk package.json object has two independent type declarations that can drift as fields get added to one and not the other. Probably worth pulling this into one shared exported type both files reference.

unlikelyzero (commented):
The five Copilot comments all look genuinely addressed, and the mechanism itself fits how Headlamp already stamps metadata onto extracted plugins (plugin-management.ts does the same read/mutate/write for isManagedByHeadlampPlugin), so this isn't importing a foreign pattern from the fork.

One thing worth flagging before this merges though: this PR edits `app/scripts/setup-plugins.js`, but #6827 ("0001: Verify external plugin manifests") deletes that exact file entirely and replaces it with `setup-plugins.ts`. This PR is titled "0002" like it stacks on 0001, but it's branched off plain main. Whichever of the two merges second is going to either silently lose this feature or have to re-port the manifest-validation work against a file that's about to disappear. Might be worth rebasing this onto #6827 so the stack is actually a stack.

A few more things inline.

illume:
Addressed the stack concern: this branch is now rebased directly onto #6827 head `bcf45d4df46652118949db7cf2c361fbee4928d4`, and the 0002 implementation is ported to `app/scripts/setup-plugins.ts`. Because #6827 lives in the contributor fork, GitHub cannot select that branch as the base of a `kubernetes-sigs/headlamp` PR; the dependency and expected post-merge 8-file diff are now explicit at the top of this PR description.

app/scripts/setup-plugins.js
illume: Addressed in f70a514f5 on the #6827 TypeScript stack. Any explicit enabledByDefault value now requires extracted package metadata and a successful atomic replacement; missing, malformed, or unwritable metadata throws before the staged plugin tree is published. Focused fail-closed tests cover missing and malformed package.json.

app/scripts/setup-plugins.js
illume: Addressed in f70a514f5. Package metadata is written to a same-directory exclusive temporary file and renamed over package.json; failures remove the temp file and abort the staged setup. The test verifies rename-based replacement, metadata preservation, and no orphan temp file.

app/scripts/setup-plugins.js
illume: Addressed by stacking on #6827: its TypeScript extractArchive already throws `new Error("Failed to find plugin content within archive: ...")`, and the CLI catch sets exitCode=1. The old reject-without-an-error JavaScript path no longer exists in this PR.

frontend/src/plugin/index.ts
illume: Addressed in 8f78e3985. The JSDoc now states that existing settings always win, including automatically persisted entries from earlier releases. Defaults are applied only to newly discovered `type === "shipped"` plugins; user-installed and development plugins remain enabled. Tests cover the user-plugin case.

illume: Addressed in f70a514f5 and 8f78e3985. Manifest setup rejects non-boolean enabledByDefault values. The frontend also fails safe at runtime: only strict `=== false` disables a newly discovered shipped plugin, so a string `"false"` remains enabled. Both boundaries have tests.

frontend/src/plugin/pluginsSlice.ts
illume: Addressed in 8f78e3985. `PluginPackageHeadlampConfig` is now exported once from pluginsSlice and referenced by both PluginInfo and pluginI18n via a type-only import, so i18n and enabledByDefault cannot drift into separate package.json shapes.

frontend/src/plugin/index.ts
```diff
@@ -317,10 +318,9 @@ export function updateSettingsPackages(
     const index = settingsPlugins.findIndex(x => x.name === plugin.name && x.type === plugin.type);
 
     if (index === -1) {
-      // It's a new one settings doesn't know about, enable it by default
       return {
         ...plugin,
-        isEnabled: true,
+        isEnabled: plugin.type !== 'shipped' || plugin.headlamp?.enabledByDefault !== false,
```
ashu8912: **Fail-open fallback when `type` is missing — points the wrong way for this feature.**

Elsewhere in this same file the convention is that an absent `type` *means* shipped:

```ts
const aPriority = priorityOrder.indexOf(a.type || 'shipped');
const bPriority = priorityOrder.indexOf(b.type || 'shipped');
```

Here the test is `plugin.type !== 'shipped'`, so an `undefined` type is treated as **not** shipped and short-circuits straight to `isEnabled: true`.

In practice this isn't a live bug — `type` is always populated from the `/plugins` metadata response in `fetchAndExecutePlugins`, so every real code path sets it. But the direction of the fallback is the unsafe one *for this specific feature*: if `type` ever goes missing, a plugin marked `enabledByDefault: false` silently ships **enabled**. That's precisely the outcome `3e60af2` ("preventing a disabled plugin from shipping enabled") is written to prevent, so it'd be nice for the frontend to fail the same way.

Suggested tweak, which also matches the surrounding convention:

```suggestion
      isEnabled: (plugin.type ?? 'shipped') !== 'shipped' || plugin.headlamp?.enabledByDefault !== false,
```

Nice touch that the test suite already pins the non-boolean case (`'false'` as a string stays enabled) — that's a genuinely easy thing to get wrong.

app/scripts/setup-plugins.ts
```diff
@@ -0,0 +1,591 @@
...
+ *
+ * @param plugin Plugin source declaration to validate.
+ * @param requireDigest Whether remote archives must declare a SHA-256 digest.
+ * @returns Nothing when the plugin source is valid.
+ * @throws When the plugin name or source is unsafe, ambiguous, or incomplete.
+ * @throws When a remote archive requires a digest but does not declare a valid one.
+ */
+export function validatePluginSource(
+  plugin: PluginSource,
+  requireDigest: boolean = externalManifest
```
ashu8912: **The first-party path is weaker than the third-party path — is that intended?**

`requireDigest` defaults to `externalManifest`, so:

- an **external** product manifest *must* declare a `sha256` for every remote archive
- Headlamp's **own** default manifest may download remote archives with **no integrity check at all**

That feels inverted. The default manifest is the one shipped to every desktop user, so it's the higher-value target, yet it's the one allowed to skip verification. Requiring a digest for *any* `archive:` source regardless of manifest origin would close that, and local `file:` sources would be unaffected.

If the asymmetry is deliberate (e.g. the default manifest currently has no remote archives at all, so the requirement would be vacuous today), a short comment here explaining that would save the next reader the same double-take.

app/scripts/setup-plugins.ts
```diff
@@ -0,0 +1,591 @@
...
+        extractionError = new Error(
+          `Plugin archive contains unsupported link or entry type: ${entryType}`
+        );
+        return false;
+      }
+      return true;
+    },
+    onReadEntry: entry => {
+      entryCount += 1;
+      extractedBytes += entry.size;
```
ashu8912, line 334: **`entry.size` is the *declared* header value, not bytes actually written.**

This accumulates the size field from the tar header rather than observed output, which would normally be a soft spot for a lying archive. It happens to be safe here because `node-tar` reads exactly `size` bytes per entry and won't emit more — so the header value *is* the bound.

That's load-bearing reasoning that isn't visible from the code though. Worth one line:

```ts
// node-tar reads exactly `size` bytes per entry, so the header value bounds real output.
```

Same thought applies to the `zlib.createGunzip()` in the pipeline below — it has no cap of its own, and these counters are the only backstop against a decompression bomb. Fine as built, just non-obvious.

app/scripts/setup-plugins.ts
```diff
@@ -0,0 +1,591 @@
...
+        )
+          .then(() => finish())
+          .catch(error => finish(error as Error));
+      } else {
+        response.resume();
+        finish(new Error(`Plugin archive download failed with status ${statusCode}: ${url}`));
+      }
+    });
+    request.on('error', error => finish(error));
+    const timeout = setTimeout(() => {
```
ashu8912, line 509: **The timeout resets on every redirect, so the real budget is ~6x the configured one.**

`finish()` clears this timer, then the recursive `downloadFile(...)` call for the redirect target starts a **fresh** `timeoutMs`. With the redirect cap of 5, a slow-but-not-dead server can hold the build for roughly `6 x 30s = 3 minutes` while still technically honouring a "30 second timeout".

Threading a single absolute deadline through would make the guarantee match the name:

```ts
downloadFile(url, dest, redirectCount + 1, limits, sourceHostname, deadline)
```

...and computing the per-request timeout as `deadline - Date.now()`.

Minor, and it affects build time rather than correctness — but the redirect chain is attacker-influenced when the archive URL is, so a bounded total is nicer.

Separately: `isAllowedRedirectHost` reads correctly to me. `sourceHostname.endsWith('.github.com')` with the leading dot means `evilgithub.com` does **not** match, which is the classic mistake avoided. :+1:

app/electron/build-manifest.test.ts
```diff
@@ -111,3 +131,455 @@ describe('platform metadata', () => {
...
+      fs.readFileSync(path.join(repositoryRoot, 'package.json'), 'utf8')
+    );
+    const appPackageJson = JSON.parse(
+      fs.readFileSync(path.join(repositoryRoot, 'app/package.json'), 'utf8')
+    );
+
+    for (const scriptName of ['app:build', 'app:build:dir', 'app:start']) {
+      expect(packageJson.scripts[scriptName]).toContain(expectedCommand);
+    }
+    expect(packageJson.engines.node).toBe('>=22.6.0');
```
ashu8912: This asserts a constant against itself — the test reads `engines.node` out of `package.json` and checks it equals the literal that was just written into `package.json`. It can only fail if someone edits the version, which is exactly when they'd also edit this line.

There's no behaviour being protected here, so I'd drop it.

app/electron/build-manifest.test.ts
```diff
@@ -111,3 +131,455 @@ describe('platform metadata', () => {
...
+    expect(appPackageJson.engines.node).toBe('>=22.6.0');
+  });
+
+  it('uses Node type stripping in Make and Windows build entrypoints', () => {
```
ashu8912: **Unit tests asserting on build-config *text* will break on legitimate refactors.**

This block reads the `Makefile` and the Windows workflow and asserts they contain the literal string `node --experimental-strip-types ./scripts/setup-plugins.ts`, including `toHaveLength(3)` on the Makefile match count.

That pins the repo's build wiring to exact prose. All of these would fail the test suite while being perfectly correct changes:

- adding a fourth call site to the Makefile
- wrapping the invocation in an npm script and calling that instead
- **dropping `--experimental-strip-types` once the Node floor reaches 22.18+**, where type stripping is on by default

That last one is plausibly within the year, and it'll surface as a red unit test in a file about build manifests — a confusing place to discover it.

The thing genuinely worth protecting is "the setup script still runs and produces a `.plugins` directory", which the e2e spec in `externalPluginManifest.spec.ts` already covers far better. I'd drop this `describe` block.

.github/workflows/app-artifacts-win.yml
```diff
@@ -97,7 +97,7 @@ jobs:
       run: |
         cd app
         npm ci
-        node scripts/setup-plugins.js
+        node --experimental-strip-types scripts/setup-plugins.ts
```
ashu8912, line 100: Flagging for process rather than correctness — the change itself is right and consistent with the other call sites.

`.github/workflows/**` is listed as a manual-review-only path in both `AGENTS.md` and `.github/copilot-instructions.md`, so this needs explicit human sign-off and can't ride along on an automated approval.

Practical note too: this is the one call site where a Node version mismatch surfaces only in Windows packaging CI rather than locally. The runner pins `22.x` (resolves to latest 22), so it's fine today — just the slowest place to find out if that ever changes.

package.json
```diff
@@ -10,7 +10,7 @@
   "license": "Apache-2.0",
   "engines": {
     "npm": ">=11.0.0",
-    "node": ">=22.0.0"
+    "node": ">=22.6.0"
```
ashu8912, line 13: This is a breaking requirement change for contributors (`>=22.0.0` -> `>=22.6.0`) and deserves a callout in the PR description.

Two follow-ups:

1. `docs/development/index.md` line 17 still advertises `>= 22.0.0`. Someone following the docs onto 22.4 gets an opaque `--experimental-strip-types` failure with nothing pointing at the cause.
2. Worth stating explicitly that Node type stripping is now a **hard build dependency**, not just a nicety — `app:build`, `app:start`, `make app-build` and Windows packaging all route through it.

CI itself is fine: every workflow pins `22.x`, which resolves to current 22 LTS.

.github/workflows/app-artifacts-win.yml
illume: Doesn't matter.

package.json
illume: does not matter

frontend/src/plugin/index.ts
illume: Fixed in e23a9b5ad. Missing type metadata now falls back to `shipped`, matching the surrounding priority convention, and a focused regression test verifies an explicit disabled default remains disabled when `type` is absent.

app/scripts/setup-plugins.ts
illume: The asymmetry is deliberate for compatibility: the reviewed in-repo default manifest currently contains three remote release archives without digest metadata, while externally selected product manifests are untrusted inputs. I added an explicit comment in 6e2af8a35 documenting that trust boundary. Migrating the built-in manifest to pinned digests can be handled separately without changing existing packaging inputs in this PR.

app/scripts/setup-plugins.ts
illume: Added the suggested node-tar invariant comment in 6e2af8a35, immediately before accumulating `entry.size`, so it is clear why the declared header size bounds actual extracted output.

app/scripts/setup-plugins.ts
illume: Fixed in 6e2af8a35. `downloadFile` now computes one absolute deadline and passes it through every redirect, with each request using only the remaining budget. A regression test advances the clock to the original deadline between redirect responses and verifies timeout plus partial-file cleanup.

app/electron/build-manifest.test.ts
illume: Removed this constant-against-itself assertion in 6e2af8a35.

app/electron/build-manifest.test.ts
illume: Removed the full text-based entrypoint `describe` block in 6e2af8a35. The setup behavior remains covered by the e2e path, and this focused suite now protects behavior rather than exact build command prose.

illume:
Rebased on main, since 0001 was merged.

illume:
Cleaned up the commits/messages, squashed a few together.

ashu8912 (approved):
LGTM!!! thanks

illume:
rebased against main, and fixed git conflicts

PR #6833, opened 2026-08-03
0027: Make the column selector keyboard accessible

illume:
## Summary

- replace MRT's nested column controls with semantic `menuitemcheckbox` items
- preserve table search, filter, and row-selection toolbar behavior
- align bulk actions with the columns they can actually change
- add focused unit coverage, current storyshots, and a deployed keyboard e2e test

## Compared with Azure/aks-desktop PR #316

This keeps the original custom column menu and toolbar integration, with these
changes for upstream Headlamp:

- individual columns expose `menuitemcheckbox` and `aria-checked` semantics
- the decorative switch is read-only, hidden from accessibility APIs, and
  removed from the input tab order
- Hide all and Show all derive their disabled state from hideable columns rather
  than table-wide visibility that can include locked or display columns
- the trigger only exposes `aria-expanded` and `aria-controls` while its menu is
  open
- caller-enabled density and fullscreen controls remain in MRT's default toolbar
  order alongside search, filters, and column visibility
- five unit tests cover menu state, filtering, disabled columns, bulk actions,
  and keyboard semantics with 100% branch coverage for the new component
- the Playwright scenario uses arrow-key navigation from MUI's initial menu focus
  and toggles from the item's observed state
- storyshots are refreshed against current upstream `main`

The development history is organized into two reviewable commits. The selector
implementation includes its unit tests and storyshots, so the first commit passes
frontend checks independently. The second commit contains only the e2e scenario.

## Provenance

This upstreams `0027-headlamp-upstream-accessible-column-selector.patch` from
Azure/aks-desktop PR [#823](https://github.com/Azure/aks-desktop/pull/823), where
it is retained as embedded patch commit
`26eee3137cde616dc089f035acb73a86208b9982`.

The original implementation came from Azure/aks-desktop PR
[#316](https://github.com/Azure/aks-desktop/pull/316), commit
[`08f52c0f5ef6344122bbd3f8670cfbdc3e470b4c`](https://github.com/Azure/aks-desktop/commit/08f52c0f5ef6344122bbd3f8670cfbdc3e470b4c),
by Oleksandr Dubenko. The frontend commit preserves Oleksandr's author date and
adds René Dudfield as co-author.

## Testing

Selector commit and final tree:

- `make frontend-test`
- `make frontend-i18n-check`
- `make frontend-lint`
- `make frontend-build`
- focused selector coverage: 100% statements, branches, functions, and lines

E2e commit:

- `playwright test tests/podsPage.spec.ts --list` (9 tests discovered)
- `make frontend-lint`
- `git diff --check`
- `git range-diff` confirms the selector and e2e patches are unchanged by the
  current-main rebase

The cluster-backed Playwright scenario was not executed locally because it
requires the documented two-kind-cluster Headlamp deployment.

Assisted by copilot

### DIFF
--- a/e2e-tests/tests/podsPage.spec.ts
+++ b/e2e-tests/tests/podsPage.spec.ts
@@ -114,6 +114,37 @@ test('loads the next page of pods', async ({ page }) => {
   expect(listRequests[1].searchParams.get('continue')).toBe('next-page');
 });
 
+test('changes column visibility with the keyboard', async ({ page }) => {
+  const headlampPage = new HeadlampPage(page);
+  await headlampPage.navigateToCluster('test', process.env.HEADLAMP_TEST_TOKEN);
+  await headlampPage.navigateTopage('/c/test/pods', /Pods/);
+
+  const columnSelector = page.getByRole('button', { name: 'Show/Hide columns' });
+  await columnSelector.focus();
+  await page.keyboard.press('Enter');
+
+  await expect(page.getByRole('menuitem', { name: 'Hide all' })).toBeFocused();
+
+  const columnItem = page.locator('[role="menuitemcheckbox"]:not([aria-disabled="true"])').first();
+  const initialChecked = await columnItem.getAttribute('aria-checked');
+  expect(initialChecked).toMatch(/^(true|false)$/);
+
+  const menuItemCount = await page.locator('[role="menuitem"], [role="menuitemcheckbox"]').count();
+  for (let index = 0; index < menuItemCount; index++) {
+    await page.keyboard.press('ArrowDown');
+    if (await columnItem.evaluate(element => element === document.activeElement)) {
+      break;
+    }
+  }
+  await expect(columnItem).toBeFocused();
+
+  await page.keyboard.press('Space');
+  await expect(columnItem).toHaveAttribute(
+    'aria-checked',
+    initialChecked === 'true' ? 'false' : 'true'
+  );
+});
+
 test('multi tab create delete pod', async ({ browser }) => {
   // This test may be slow to create and delete a pod
   test.setTimeout(60000);
--- a/e2e-tests/tests/podsPage.spec.ts
+++ b/e2e-tests/tests/podsPage.spec.ts
@@ -123,16 +123,18 @@ test('changes column visibility with the keyboard', async ({ page }) => {
   await columnSelector.focus();
   await page.keyboard.press('Enter');
 
-  await expect(page.getByRole('menuitem', { name: 'Hide all' })).toBeFocused();
-  await page.keyboard.press('ArrowDown');
-  await page.keyboard.press('ArrowDown');
-
   const columnItem = page.getByRole('menuitemcheckbox').first();
+  const initialChecked = await columnItem.getAttribute('aria-checked');
+  expect(initialChecked).toMatch(/^(true|false)$/);
+
+  await columnItem.focus();
   await expect(columnItem).toBeFocused();
-  await expect(columnItem).toHaveAttribute('aria-checked', 'true');
 
   await page.keyboard.press('Space');
-  await expect(columnItem).toHaveAttribute('aria-checked', 'false');
+  await expect(columnItem).toHaveAttribute(
+    'aria-checked',
+    initialChecked === 'true' ? 'false' : 'true'
+  );
 });
 
 test('multi tab create delete pod', async ({ browser }) => {
--- a/e2e-tests/tests/podsPage.spec.ts
+++ b/e2e-tests/tests/podsPage.spec.ts
@@ -114,6 +114,27 @@ test('loads the next page of pods', async ({ page }) => {
   expect(listRequests[1].searchParams.get('continue')).toBe('next-page');
 });
 
+test('changes column visibility with the keyboard', async ({ page }) => {
+  const headlampPage = new HeadlampPage(page);
+  await headlampPage.navigateToCluster('test', process.env.HEADLAMP_TEST_TOKEN);
+  await headlampPage.navigateTopage('/c/test/pods', /Pods/);
+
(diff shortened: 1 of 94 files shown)


frontend/src/components/common/Table/ColumnVisibilityButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+      >
+        <MenuItem
+          disabled={!table.getIsSomeColumnsVisible()}
+          onClick={() => handleToggleAll(false)}
+        >
+          {localization.hideAll}
+        </MenuItem>
+        <MenuItem disabled={table.getIsAllColumnsVisible()} onClick={() => handleToggleAll(true)}>
+          {localization.showAll}
+        </MenuItem>
```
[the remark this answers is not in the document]
illume: Addressed in 4d60cb676. Bulk action availability now derives from the same hideable column subset that the handlers update, with unit coverage for visible non-hideable and hidden hideable columns. The visual switch is also explicitly read-only and removed from the input tab order.

e2e-tests/tests/podsPage.spec.ts
```diff
@@ -114,6 +114,27 @@ test('loads the next page of pods', async ({ page }) => {
...
+  await page.keyboard.press('Enter');
+
+  await expect(page.getByRole('menuitem', { name: 'Hide all' })).toBeFocused();
+  await page.keyboard.press('ArrowDown');
+  await page.keyboard.press('ArrowDown');
+
+  const columnItem = page.getByRole('menuitemcheckbox').first();
+  await expect(columnItem).toBeFocused();
+  await expect(columnItem).toHaveAttribute('aria-checked', 'true');
+
```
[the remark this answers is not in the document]
illume: Addressed in d99d2dae5. The test now focuses the located menuitemcheckbox directly and toggles from its observed aria-checked state, so menu ordering, responsive visibility, and persisted settings do not affect the assertion.

e2e-tests/tests/podsPage.spec.ts
```diff
@@ -114,6 +114,29 @@ test('loads the next page of pods', async ({ page }) => {
...
+  const columnSelector = page.getByRole('button', { name: 'Show/Hide columns' });
+  await columnSelector.focus();
+  await page.keyboard.press('Enter');
+
+  const columnItem = page.getByRole('menuitemcheckbox').first();
+  const initialChecked = await columnItem.getAttribute('aria-checked');
+  expect(initialChecked).toMatch(/^(true|false)$/);
+
+  await columnItem.focus();
```
ashu8912: Opening the menu with Enter is genuinely covered, but this line then jumps focus
programmatically, which skips the part that actually regressed: arrow-key navigation
through the menu. `await expect(columnItem).toBeFocused()` on the next line only asserts
that `.focus()` worked.

Also note MUI autofocuses the first item on open, and the first item here is "Hide all",
not a menuitemcheckbox — so the test never demonstrates a keyboard user can *reach* a
column toggle. Suggested shape:

    await page.keyboard.press('Enter');
    await expect(page.getByRole('menuitem', { name: 'Hide all' })).toBeFocused();
    await page.keyboard.press('ArrowDown');
    await page.keyboard.press('ArrowDown');   // past "Show all" / divider
    await page.keyboard.press('Space');

That exercises the real path. Optionally also assert the column header disappears from
the table, so the test proves user-visible effect and not just the aria attribute.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -377,16 +380,37 @@ export default function Table<RowItem extends Record<string, any>>({
...
+
+      return (
+        <>
+          {enableFilters && enableGlobalFilter && !initState?.showGlobalFilter && (
+            <MRT_ToggleGlobalFilterButton table={tbl} />
+          )}
+          {enableFilters && enableColumnFilters && columnFilterDisplayMode !== 'popover' && (
+            <MRT_ToggleFiltersButton table={tbl} />
+          )}
+          {(enableHiding || enableColumnOrdering || enableColumnPinning) && (
```
ashu8912, line 413: This block reproduces MRT's default `MRT_ToolbarInternalButtons` so the custom column
menu can be swapped in, but it reproduces only 3 of the 5 default buttons. MRT's actual
default (index.esm.js:3996, v2.13.3) also renders:

    {enableDensityToggle && <MRT_ToggleDensePaddingButton table={table} />}
    {enableFullScreenToggle && <MRT_ToggleFullScreenButton table={table} />}

Headlamp defaults both to false (`tableProps.enableDensityToggle ?? false` at Table.tsx
:390-391), which is why no storyshot caught this — but both are caller-overridable, and
`Table` is part of the public plugin surface. So a caller or plugin that passes
`enableDensityToggle` / `enableFullScreenToggle` used to get those buttons and now
silently gets nothing.

Could you add the two missing branches? It's two lines and it makes this a true drop-in
replacement.

ashu8912 (commented):
Added a few comments

e2e-tests/tests/podsPage.spec.ts
illume: Addressed in a41f1038. The test now starts from MUI’s focused Hide all item, uses ArrowDown until the first enabled menuitemcheckbox receives focus, and activates it with Space. The loop is bounded by the actual menu item count, so disabled-item focus behavior does not make the test flaky.

frontend/src/components/common/Table/Table.tsx
illume: Addressed in e29d4da4. The custom toolbar now preserves MRT’s full default action set by rendering the density and fullscreen buttons when their public options are enabled. A focused Table integration test asserts search, filters, columns, density, and fullscreen controls together.

ashu8912 (approved):

illume:
rebased against main to fix conflict in test file

PR #6874, opened 2026-08-04
fix(backend): add X-Frame-Options and Content-Security-Policy headers

### LINKED ISSUE #6870: Security: Missing X-Frame-Options and Content-Security-Policy headers enable clickjacking
## Summary

The Headlamp backend does not set `X-Frame-Options` or `Content-Security-Policy: frame-ancestors` headers. As a result, the login page (and the full application) can be embedded in a cross-origin `<iframe>`, enabling clickjacking attacks against the authentication flow.

This was confirmed by a penetration test: the Headlamp login page loaded successfully inside a cross-origin iframe in a proof-of-concept clickjacking demonstration.

## Details

Inspecting `backend/cmd/headlamp.go`, there is no global security-headers middleware. Individual handlers only set `Content-Type` and caching headers. Neither `X-Frame-Options` nor `Content-Security-Policy` with a `frame-ancestors` directive is added to any response.

Modern browsers rely on one of these two mechanisms to prevent framing:
- `X-Frame-Options: DENY` (older, widely supported)
- `Content-Security-Policy: frame-ancestors 'none'` (newer, takes precedence over XFO in supporting browsers)

Without either, an attacker can embed the Headlamp login page in a malicious site and trick authenticated users into performing unintended actions (e.g. credential phishing via a transparent overlay).

## Proposed fix

Add a global security-headers middleware to the `mux.Router` in `createHeadlampHandler` that sets these headers on every response:

```go
func securityHeadersMiddleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        w.Header().Set("X-Frame-Options", "DENY")
        w.Header().Set("Content-Security-Policy", "frame-ancestors 'none'")
        next.ServeHTTP(w, r)
    })
}
```

Applied via `r.Use(securityHeadersMiddleware)` so all routes are covered without per-handler changes.

## Impact

- Clickjacking of the OIDC login flow
- UI redressing attacks against authenticated users
- Flagged as a Low severity finding in our penetration test

## References

- Related: #6868 (unauthenticated `/config`), #6869 (unauthenticated `/plugins`)
- OWASP Clickjacking Defence Cheat Sheet: https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html

harshitnub077:
Fixes #6870. This PR adds a new `securityHeadersMiddleware` to the main backend router to protect the application (including the OIDC login flow) from clickjacking attacks.

1. **`backend/cmd/headlamp.go`**: Created `securityHeadersMiddleware` and applied it to the root `mux.Router`.
2. **`backend/cmd/headlamp_test.go`**: Added `TestSecurityHeadersMiddleware` to assert that the `X-Frame-Options: DENY` and `Content-Security-Policy: frame-ancestors 'none'` headers are correctly returned on all endpoints.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -569,6 +569,14 @@ func setupInClusterContext(config *HeadlampConfig) {
 	}
 }
 
+func securityHeadersMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("X-Frame-Options", "DENY")
+		w.Header().Set("Content-Security-Policy", "frame-ancestors 'none'")
+		next.ServeHTTP(w, r)
+	})
+}
+
 //nolint:gocognit,funlen,gocyclo
 func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Handler {
 	kubeConfigPath := config.KubeConfigPath
@@ -645,6 +653,8 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		baseRoute := mux.NewRouter()
 		r = baseRoute.PathPrefix(config.BaseURL).Subrouter()
 	}
+
+	r.Use(securityHeadersMiddleware)
 
 	logger.Log(logger.LevelInfo, nil, nil, "*** Headlamp Server ***")
 	logger.Log(logger.LevelInfo, nil, nil, "  API Routers:")
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -4158,3 +4158,27 @@ func TestExternalProxyOversizeResponseGzip(t *testing.T) {
 	assert.Equal(t, http.StatusOK, rr.Code)
 	assert.Equal(t, int(maxProxyResponseSize), rr.Body.Len())
 }
+
+func TestSecurityHeadersMiddleware(t *testing.T) {
+	cache := cache.New[interface{}]()
+	kubeConfigStore := kubeconfig.NewContextStore()
+
+	handler := createHeadlampHandler(context.Background(), &HeadlampConfig{
+		HeadlampConfig: &headlampconfig.HeadlampConfig{
+			HeadlampCFG: &headlampconfig.HeadlampCFG{
+				UseInCluster:    false,
+				KubeConfigStore: kubeConfigStore,
+			},
+			Cache: cache,
+		},
+	})
+
+	req, err := http.NewRequestWithContext(context.Background(), "GET", "/config", nil)
+	require.NoError(t, err)
+
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+
+	assert.Equal(t, "DENY", rr.Header().Get("X-Frame-Options"))
+	assert.Equal(t, "frame-ancestors 'none'", rr.Header().Get("Content-Security-Policy"))
+}


illume (changes requested):
Thanks for working on this.


The commit messages could use some tidying up to match our contribution guidelines. We use Linux kernel style — the [contributing guide](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines) has the details, and `git log` shows good examples.

<details>

- `fix(backend): add global security headers to protect against clickjacking` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

harshitnub077:
@illume Thanks for the review! I've updated the commit message to follow the Linux kernel style as per the contribution guidelines. The prow label has been cleared. Could you take another look when you get a chance?

harshitnub077:
@yolossn Awesome catch—thanks for the thorough review! You are spot on locking this down unconditionally would have definitely broken folks embedding Headlamp via Backstage or other iframes. 

I've just force-pushed an update that incorporates your suggested escape hatch: 

- **Configurable `frame-ancestors`**: Added the `--allowed-frame-ancestors` flag to handle custom origins.
- **Secure Defaults Maintained**: If no flag is passed, we still default to the stricter `X-Frame-Options: DENY` and `Content-Security-Policy: frame-ancestors 'none'`.
- **Seamless Escape Hatch**: When operators configure specific origins, the middleware emits only the `Content-Security-Policy` with their allowlist and safely skips `X-Frame-Options` (since XFO doesn't support allowlists anyway). 
- **tests**: I also added test cases in `headlamp_test.go` to explicitly verify both the default and the allowlisted middleware behaviors.

really appreciate you pointing out that edge case. Let me know if this implementation aligns with what you had in mind.

backend/cmd/headlamp.go
```diff
@@ -569,6 +569,20 @@ func setupInClusterContext(config *HeadlampConfig) {
...
 }
 
+func (c *HeadlampConfig) securityHeadersMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if c.AllowedFrameAncestors != "" {
+			ancestors := strings.ReplaceAll(c.AllowedFrameAncestors, ",", " ")
+			w.Header().Set("Content-Security-Policy", fmt.Sprintf("frame-ancestors %s", ancestors))
+		} else {
+			w.Header().Set("X-Frame-Options", "DENY")
+			w.Header().Set("Content-Security-Policy", "frame-ancestors 'none'")
```
unlikelyzero, line 675: A malformed value here is a real problem: `AllowedFrameAncestors != ""` picks this branch for anything non-empty, including "," or a whitespace-only value, and `strings.ReplaceAll(ancestors, ",", " ")` doesn't filter empty tokens. That produces an empty `frame-ancestors` directive with X-Frame-Options dropped too, so the response ends up with no clickjacking protection at all. I'd split on `,`, trim, drop empties, and fall back to the default deny if nothing's left - and validate this at startup the way `proxy-urls` does rather than re-parsing it on every request.

backend/cmd/headlamp.go
```diff
@@ -646,6 +660,8 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 		r = baseRoute.PathPrefix(config.BaseURL).Subrouter()
 	}
 
+	r.Use(config.securityHeadersMiddleware)
```
unlikelyzero: `r.Use()` only applies to routes that match on this router. I checked and 404s, 405s, and (with `--base-url` set) anything outside the base prefix don't get these headers at all. `hostValidationMiddleware` makes it worse since it wraps the handler outside this router and returns before calling `next.ServeHTTP` on a rejected Host header, so that response skips the middleware too. I'd wrap the handler returned from `createHeadlampHandler` in `serverHandler` instead, next to `hostValidationMiddleware` - that covers all of this without depending on mux's route matching.

backend/pkg/config/config.go
```diff
@@ -591,6 +592,9 @@ func addGeneralFlags(f *flag.FlagSet) {
 	f.String("service-account-token-path", "",
 		"Path to the service account token. "+
 			"Only used when --unsafe-use-service-account-token is set and in-cluster")
+	f.String("allowed-frame-ancestors", "", "Comma-separated list of origins allowed to embed Headlamp "+
+		"in an iframe (e.g., 'https://example.com https://app.example.com'). If set, emits Content-Security-Policy "+
+		"with frame-ancestors and skips X-Frame-Options.")
```
unlikelyzero, line 749: This says comma-separated but the example uses spaces. Both work at runtime since the code just swaps commas for spaces, but the example should match what the docstring claims.

harshitnub077:
@unlikelyzero @yolossn Thank you for the thorough review. You raised excellent points regarding the middleware coverage for 404s/405s and the potential failure modes with malformed allowlists.

I've pushed an update that addresses all the feedback:

1. Robust Allowlist Parsing at Startup I refactored AllowedFrameAncestors to parse strictly at startup rather than on every request. It splits on commas, trims whitespace, and drops empty tokens. If a malformed string is provided (e.g., a trailing comma or spaces), it safely falls back to the default X-Frame-Options: DENY and frame-ancestors 'none', preventing any silent loss of clickjacking protection.

2. Global Middleware Coverage I removed r.Use(config.securityHeadersMiddleware) from the mux router. Instead, I wrapped the final root handler with config.securityHeadersMiddleware(handler) inside serverHandler(). This ensures that all edge-case responses—including 404s, 405s, paths outside the base URL, and rejected Host headers—receive the security headers unconditionally.

3. Config Documentation I updated the --allowed-frame-ancestors flag documentation so the example URL list correctly uses a comma separator, matching the description.

4. Backstage Integration Documentation To ensure developers embedding Headlamp aren't broken on upgrade, I updated the run command in backstage-test/README.md to explicitly include the --allowed-frame-ancestors="http://localhost:8000" flag as an escape hatch.

All tests and linters are passing locally thanks

sniok:
if I recall correctly, the backstage integration renders headlamp in an iframe. I'd recommend making sure that still works after this change

harshitnub077:
@sniok Thank you for pointing that out. This edge case is addressed by the new --allowed-frame-ancestors flag introduced in this PR.
While iframes are now restricted by default for security, operators embedding Headlamp via Backstage can use this flag to explicitly allow their origins (e.g., --allowed-frame-ancestors="https://backstage.example.com"). I have also updated the backstage-test/README.md to include this flag in the run commands so that local testing continues to work out-of-the-box.

sniok:
Having a parameter is good but might not be enough, we should try to avoid breaking existing setups. I'd test first to see if this is an issue in the first place, but then if it is maybe we can restrict those headers only to the login page instead of globally

harshitnub077:
@sniok Thanks for the feedback! You make a fair point about backwards compatibility.

While restricting the headers to the login page would prevent login-spoofing, Headlamp is an administrative dashboard. If an attacker frames the app while a user is already authenticated they could trick the user into inadvertently interacting with the cluster (e.g., deleting resources or changing configs). To be fully effective ,clickjacking protection needs to apply globally

This aligns with @yolossn and @unlikelyzero's earlier reviews, where we agreed to prioritize the secure default out-of-the-box while relying on the new --allowed-frame-ancestors flag as the explicit escape hatch for Backstage operators.
Let me know if you think we should revisit that tradeoff, but I believe the global secure default is the safest approach for the community

sniok:
Oh I completely missed the comment by [yolossn](https://github.com/yolossn) that already covers my concerns, disregard my comment then

harshitnub077:
No worries at all, @sniok! It's easy to miss with multiple threads going on

Also, I've just rebased the branch against the latest main to resolve the out-of-date conflicts. All CI tests are currently running. Let me know if there's anything else needed

harshitnub077:
Thanks for reviewing @illume I've fixed the merge conflicts and fully rebased against the latest main. I also squashed the commits and updated the message to perfectly follow the Linux kernel style guidelines let me know if anything else is needed

harshitnub077:
@illume @yolossn @unlikelyzero Thanks for the thorough reviews.

I've pushed an update that addresses the remaining feedback:

1. Commit Style: Rewrote the commit message to properly follow the Linux kernel style guidelines.
2. Build Fix: Removed the stray merge conflict marker that was breaking the backend compilation.
All CI tests and backend builds are green again. Please take another look when you have a moment.

backend/pkg/exec/exec.go
yolossn, line 1: changes in this file are not related to this PR

harshitnub077:
@yolossn Great catch! The exec.go changes slipped in from another branch by mistake, so I've reverted them entirely.
I've also wired up allowedFrameAncestors in the Helm chart (values.yaml, values.schema.json, and deployment.yaml) so operators can easily configure the escape hatch.
Everything is rebased and squashed. Thanks for the stellar review let me know if we're good to go

yolossn:
This PR is good to go, but the backstage plugin for headlamp needs to be updated to use this config. Including this feature in this release will break the integration. So moving this to the next milestone and integrate the changes in backstage plugin.

harshitnub077:
@illume squashed the commits down to a single atomic commit and updated the message to match the style guide. also fixed the helm chart fixture ordering that copilot pointed out. let me know if we are good to go

harshitnub077:
the backend test failed due to a github cache service outage returning a 400 error so could you please re run the ci when possible

harshitnub077:
I have updated the branch:
- Split the server config unit tests to satisfy the `funlen` linter rule in CI.
- Updated the Helm chart expected test template version to 0.45.0.
- All backend tests and Helm chart template checks pass locally.

illume (changes requested):
Thanks for working on this.


harshitnub077:
i've marked the previous copilot review threads as resolved they were addressed in commit `8f75da2` and updated the pr title to follow our style guidelines.

backend/cmd/headlamp.go
```diff
@@ -657,6 +657,20 @@ func loadDynamicClusters(config *HeadlampConfig, path string, skipFunc func(kube
 	}
 }
 
+func (c *HeadlampConfig) securityHeadersMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if len(c.AllowedFrameAncestors) > 0 {
+			ancestors := strings.Join(c.AllowedFrameAncestors, " ")
```
rootp1: AllowedFrameAncestors values get joined and inserted directly into the Content Security Policy header with no validation on what each entry actually contains. If one entry has a space in it, say from a copy paste mistake in a helm values file, whatever comes after that space effectively becomes an extra CSP directive rather than an origin, since it lands right after frame-ancestors in the header string. Worth at least trimming and validating each entry looks like an origin before joining.

backend/pkg/config/config.go
```diff
@@ -647,6 +648,9 @@ func addGeneralFlags(f *flag.FlagSet, appName string) {
 	f.String("service-account-token-path", "",
 		"Path to the service account token. "+
 			"Only used when --unsafe-use-service-account-token is set and in-cluster")
+	f.String("allowed-frame-ancestors", "", "Comma-separated list of origins allowed to embed Headlamp "+
```
rootp1, line 747: Same concern as the middleware comment, this is where the comma separated string first gets split, might be the better place to validate each origin before it ever reaches the header.

backend/cmd/headlamp_test.go
```diff
@@ -4581,3 +4581,56 @@ func TestExternalProxyOversizeResponseGzip(t *testing.T) {
 	assert.Equal(t, http.StatusOK, rr.Code)
 	assert.Equal(t, int(maxProxyResponseSize), rr.Body.Len())
 }
+
+func TestSecurityHeadersMiddleware(t *testing.T) {
```
rootp1, line 4588: Both cases here only hit /config. Could we add a case that requests the actual SPA static file route to confirm the middleware covers that path too, since that is the more commonly embedded page in practice.

rootp1 (commented):
Pulled this down, built clean, and confirmed the headers show up correctly against a running server with default and custom config. Good direction overall, left a few inline notes.

harshitnub077:
i pushed an update addressing the feedback:
- Added origin validation in `config.Validate()` to verify each configured entry and prevent accidental directive injection via spaces, semicolons, or invalid formats (covered with unit tests in `config_test.go`).
- Added defensive trimming in the middleware before header joining.
- Extended the middleware tests in `headlamp_test.go` to assert headers on the SPA static root (`/`) route as well as `/config`

harshitnub077:
gentle follow up on this @yolossn @illume when convenient latest comments are resolved and branch is ready for workflow approval and review thanks

PR #6880, opened 2026-08-04
backend: serviceproxy: Send no-cache headers before response writes

### LINKED ISSUE #6839: backend: serviceproxy: no-cache headers are never sent because disableResponseCaching is deferred
## Describe the bug

`RequestHandler` registers the cache-suppression helper with `defer`:

https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L26

```go
defer disableResponseCaching(w)
```

Because it is deferred, it runs *after* the status line and body have already been
written by `handleServiceProxy`. Per the `net/http` documentation for
`ResponseWriter.Header()`:

> Changing the header map after a call to WriteHeader (or Write) has no effect
> unless the HTTP status code was a 1xx or the modified headers are trailers.

So `disableResponseCaching` is a **no-op on every code path** in this handler:

- the successful proxy path (body written by `http.ServeContent`/`w.Write` before the defer runs), and
- all four early error returns, which call `w.WriteHeader(...)` directly at
  [L31](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L31),
  [L42](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L42),
  [L54](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L54)
  and [L62](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L62).

**Expected:** every response from `/clusters/{clusterName}/serviceproxy/{namespace}/{name}`
carries `Cache-Control: no-cache, private, max-age=0`, `Pragma: no-cache`,
`X-Accel-Expires: 0` and a past `Expires`, exactly as
[`disableResponseCaching`](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler.go#L121-L126)
intends.

**Actual:** none of those headers reach the client, so browsers and any
intermediate/reverse proxy are free to cache proxied in-cluster service
responses (dashboards, Prometheus output, etc.).

## To Reproduce

Steps to reproduce the bug:

1. Run `headlamp-server` against any cluster that has a service exposing an HTTP endpoint.
2. Request it through the service proxy, e.g.
   `curl -i "http://localhost:4466/clusters/<cluster>/serviceproxy/<namespace>/<service>?request=/"`
3. Inspect the response headers.
4. No `Cache-Control`, `Pragma`, `Expires` or `X-Accel-Expires` header is present.

### Self-contained proof

This reproduces the `net/http` behaviour in isolation, mirroring
`disableResponseCaching` and the deferred call site. Save as `main.go` and run
`go run main.go` — no Headlamp dependencies needed:

```go
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"
	"time"
)

// Copied verbatim from backend/pkg/serviceproxy/handler.go
func disableResponseCaching(w http.ResponseWriter) {
	w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
	w.Header().Set("Expires", time.Unix(0, 0).Format(http.TimeFormat))
	w.Header().Set("Pragma", "no-cache")
	w.Header().Set("X-Accel-Expires", "0")
}

// Mirrors RequestHandler's structure today.
func deferred(w http.ResponseWriter, r *http.Request) {
	defer disableResponseCaching(w)
	w.WriteHeader(http.StatusOK)
	_, _ = w.Write([]byte("proxied body"))
}

// The proposed fix.
func upfront(w http.ResponseWriter, r *http.Request) {
	disableResponseCaching(w)
	w.WriteHeader(http.StatusOK)
	_, _ = w.Write([]byte("proxied body"))
}

func probe(label string, h http.HandlerFunc) {
	srv := httptest.NewServer(h)
	defer srv.Close()
	resp, err := http.Get(srv.URL)
	if err != nil {
		fmt.Println(label, "ERR", err)
		return
	}
	defer resp.Body.Close()
	fmt.Printf("%-24s Cache-Control=%-30q Pragma=%q\n",
		label, resp.Header.Get("Cache-Control"), resp.Header.Get("Pragma"))
}

func main() {
	probe("current (deferred):", deferred)
	probe("fixed (before write):", upfront)
}
```

Output:

```
current (deferred):      Cache-Control=""                             Pragma=""
fixed (before write):    Cache-Control="no-cache, private, max-age=0"  Pragma="no-cache"
```

 - Installation type: any (backend defect — reproduced from source)
 - Headlamp Version: 0.44.0 (`506465d78ca162f65e46c57fab7b014fd771d047`)
 - Other: Go backend, `backend/pkg/serviceproxy`

Yes (I will propose a PR)

### Why this was not caught by the existing test

The only coverage calls the helper in isolation rather than through `RequestHandler`:

https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/serviceproxy/handler_test.go#L223

```go
disableResponseCaching(w)
assert.Equal(t, "no-cache, private, max-age=0", w.Header().Get("Cache-Control"))
```

Worth noting for whoever writes the regression test: a `httptest.ResponseRecorder`
based test would **also pass while the bug is live**, because
`ResponseRecorder.Header()` returns the live header map (`WriteHeader` only
snapshots into `snapHeader`, which is used by `Result()`). The regression test
therefore has to assert against headers observed from a real server —
`httptest.NewServer` + `http.Get`, or `rr.Result().Header` rather than
`rr.Header()`.

### Suggested fix

```diff
--- a/backend/pkg/serviceproxy/handler.go
+++ b/backend/pkg/serviceproxy/handler.go
@@
 	clusterName, namespace, name, requestURI := parseInfoFromRequest(r)
-
-	defer disableResponseCaching(w)
-	// Get the context
+	// Must run before anything writes to w: changing the header map after
+	// WriteHeader/Write has no effect.
+	disableResponseCaching(w)
+
 	ctx, err := kubeConfigStore.GetContext(clusterName)
```

### Prior art in this repo

The same bug class was already recognised and fixed in the Helm handlers —
#5055 ("Helm handlers: missing return after `http.Error` and Content-Type header
set after `WriteHeader`"). This is the same mistake in `serviceproxy`.

CygnusMaximillian:
## Summary

This PR fixes a bug in the service proxy handler where cache-suppression headers were never sent to clients. `disableResponseCaching` was registered with `defer`, so it ran after the status line and body had already been written. Per `net/http`, changing the header map after `WriteHeader` or `Write` has no effect, which made the helper a no-op on every code path and left proxied in-cluster service responses cacheable by browsers and intermediate proxies.

## Related Issue

Fixes #6839

## Changes

- Call `disableResponseCaching` at the start of `RequestHandler` instead of deferring it
- Add `TestRequestHandlerSendsNoCacheHeaders`, asserting headers observed from a real `httptest.NewServer` round-trip (a `ResponseRecorder`'s live header map would show them even when they were set too late to be sent)

## Steps to Test

1. Run the regression test:
   ```bash
   cd backend && go test ./pkg/serviceproxy -run TestRequestHandlerSendsNoCacheHeaders -count=1 -v

<img width="1029" height="106" alt="image" src="https://github.com/user-attachments/assets/1499f996-284b-4fa0-ae5f-af15576cec89" />

### DIFF
--- a/backend/pkg/serviceproxy/handler.go
+++ b/backend/pkg/serviceproxy/handler.go
@@ -23,7 +23,9 @@ func RequestHandler(
 ) {
 	clusterName, namespace, name, requestURI := parseInfoFromRequest(r)
 
-	defer disableResponseCaching(w)
+	// Must run before anything writes to w: changing the header map after
+	// WriteHeader or Write has no effect.
+	disableResponseCaching(w)
 	// Get the context
 	ctx, err := kubeConfigStore.GetContext(clusterName)
 	if err != nil {
--- a/backend/pkg/serviceproxy/handler_test.go
+++ b/backend/pkg/serviceproxy/handler_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/gorilla/mux"
+	"github.com/kubernetes-sigs/headlamp/backend/pkg/kubeconfig"
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -225,6 +226,41 @@ func TestDisableResponseCaching(t *testing.T) {
 	assert.Equal(t, "no-cache, private, max-age=0", w.Header().Get("Cache-Control"))
 	assert.Equal(t, "no-cache", w.Header().Get("Pragma"))
 	assert.Equal(t, "0", w.Header().Get("X-Accel-Expires"))
+}
+
+// TestRequestHandlerSendsNoCacheHeaders is a regression test for the
+// cache-suppression headers being registered with defer, which ran only after
+// the response had been written and therefore never reached the client. The
+// headers must be observed from a real server round-trip: a ResponseRecorder's
+// live header map would show them even when they were set too late to be sent.
+func TestRequestHandlerSendsNoCacheHeaders(t *testing.T) {
+	store := kubeconfig.NewContextStore()
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		r = mux.SetURLVars(r, map[string]string{"clusterName": "unknown", "namespace": "default", "name": "svc"})
+		RequestHandler(store, false, w, r)
+	}))
+	defer server.Close()
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, server.URL, nil)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	defer func() { _ = resp.Body.Close() }()
+
+	// The unknown cluster makes the handler take the early 404 error path;
+	// the no-cache headers must be present on every response.
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+	assert.Equal(t, "no-cache, private, max-age=0", resp.Header.Get("Cache-Control"))
+	assert.Equal(t, "no-cache", resp.Header.Get("Pragma"))
+	assert.Equal(t, "0", resp.Header.Get("X-Accel-Expires"))
+	assert.NotEmpty(t, resp.Header.Get("Expires"))
 }
 
 // createMockService creates a mock Kubernetes service for testing.


illume (changes requested):
Thanks for these changes.


CygnusMaximillian:
@illume I have resolved all the copilot issues and the PR is ready to be merged thanks :)

Suyog241005:
@CygnusMaximillian If you've resolved the copilot comments, please mark it as resolved.

yolossn (approved):
LGTM

PR #6881, opened 2026-08-04
Frontend/reduce pluginlib surface

### LINKED ISSUE #6879: frontend: Reduce the API surface exposed to plugins through window.pluginLib
## Summary

The current plugin runtime executes plugins using `new Function(...)` and exposes a large API surface through `window.pluginLib`, including internal utilities such as the Redux store, API helpers, Kubernetes client wrappers, React integrations, and other implementation details.

While plugins are intended to be trusted extensions, exposing internal implementation details through a global object tightly couples plugins to Headlamp internals and makes it difficult to evolve the plugin API safely.

Providing a smaller, documented plugin API would improve encapsulation while preserving existing plugin functionality.

### Expected behavior

The plugin runtime should expose only the public APIs required for plugin development.

In particular:

- Expose only documented plugin APIs.
- Avoid exposing internal implementation details directly.
- Avoid exposing mutable internal state (such as the Redux store) unless required.
- Clearly separate public plugin APIs from internal application APIs.

## Current behavior

Plugins execute with access to the entire `window.pluginLib` object, which includes many internal utilities beyond the documented plugin interfaces.

This makes it possible for plugins to depend on internal implementation details that are not part of the supported plugin API, increasing maintenance costs and making future refactoring more difficult.

## Environment

- Installation type: Any
- Headlamp Version: Current `main`
- Browser: Any modern browser

Yes. I would like to work on this issue and submit a PR.

## Additional Context

This proposal is intended as an architectural improvement rather than a functional bug fix.

Possible implementation approaches include:

- Defining a minimal public plugin API.
- Exposing only supported interfaces through `window.pluginLib`.
- Keeping internal application objects private.
- Documenting the supported plugin runtime APIs.

This would improve encapsulation, reduce accidental dependencies on internal implementation details, and make future maintenance easier without changing the plugin architecture.

vanshika2720:
## Summary

This PR documents the `stateless` module exposed through `window.pluginLib` as an internal, unsupported API rather than a supported part of the Headlamp plugin API.

Although `stateless` is currently exposed at runtime for historical compatibility, it is used internally for IndexedDB-backed kubeconfig storage, is not included in the `@kinvolk/headlamp-plugin` externals, and is not part of the documented plugin API. Removing it would introduce a breaking change for existing plugins, so this PR instead clarifies its status while preserving backward compatibility.

## Related Issue

Fixes #6879

## Changes

- Documented `window.pluginLib.stateless` as an internal-only compatibility API.
- Clarified that `stateless` is **not** part of the supported Headlamp plugin API.
- Added plugin API tests verifying that all documented public APIs remain exposed through `window.pluginLib`.
- Preserved backward compatibility for existing plugins that still reference `stateless`.
- No runtime behavior or supported plugin APIs were changed.

## Steps to Test

1. Run the frontend test suite:

   ```bash
   npm run frontend:test
   ```

2. Verify the plugin API tests pass.
3. Load an existing Headlamp plugin using the documented plugin API.
4. Confirm all documented plugin APIs continue to function normally.
5. Verify that `window.pluginLib.stateless` is still available for backward compatibility and is documented as an internal, unsupported API.

N/A

## Notes for the Reviewer

- This PR **does not remove** `stateless` from `window.pluginLib`.
- Removing it would be a breaking change because some existing plugins still rely on it.
- Instead, this change documents `stateless` as an internal compatibility API that is unsupported and should not be used by new plugins.
- The supported plugin API remains unchanged, and no runtime behavior is modified.
- This keeps the change narrowly scoped while avoiding breaking existing plugins.

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -103,6 +103,12 @@ window.pluginLib = {
   useTranslation,
   ...registryToExport,
   Activity,
+  /**
+   * @deprecated Internal implementation detail of Headlamp core (IndexedDB-backed
+   * kubeconfig storage for stateless clusters). Not part of the supported plugin API:
+   * it's absent from @kinvolk/headlamp-plugin's externals mapping and no plugin should
+   * rely on it. Kept for backwards compatibility; do not use in new plugin code.
+   */
   stateless,
 };
 
--- a/frontend/src/plugin/pluginLib.test.ts
+++ b/frontend/src/plugin/pluginLib.test.ts
@@ -52,4 +52,40 @@ describe('pluginLib variable', () => {
     // It's an implementation detail of our pluginLib that plugins must not rely on.
     expect(window.pluginLib).not.toHaveProperty('queryClient');
   });
+
+  it('should expose the documented public plugin API', () => {
+    // These are the properties plugins are documented to rely on, either via
+    // the prose docs (docs/development/plugins/functionality/index.md) or via
+    // the @kinvolk/headlamp-plugin externals mapping
+    // (plugins/headlamp-plugin/config/vite.config.mjs). Removing any of these
+    // would be a breaking change for existing plugins.
+    const documentedApi = [
+      // Named directly in the functionality docs as "the main ones".
+      'K8s',
+      'Headlamp',
+      'CommonComponents',
+      'Notification',
+      'Router',
+      'Activity',
+      // Mapped by @kinvolk/headlamp-plugin's externals config.
+      'ApiProxy',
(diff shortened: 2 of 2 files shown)


sniok:
stateless apis are used by some plugins, we can't just remove things from pluginLib, that would be a breaking change

vanshika2720:
@sniok Would you prefer restricting access through a different mechanism, or is this issue something that can't be addressed without redesigning the plugin architecture?

vanshika2720:
@illume @sniok @skoeva  pr description fixed- all resolved

PR #6885, opened 2026-08-05
backend: helm: Preserve server-side repo fields in AddRepo

### LINKED ISSUE #6840: backend: helm: AddRepo silently wipes existing repository credentials and TLS cert/key/CA paths
## Describe the bug

`addRepository` constructs a brand-new `repo.Entry` containing only `Name` and
`URL`, applies whichever request fields happen to be non-nil, and then calls
`repoFile.Update`:

https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L169-L188

```go
newRepo := &repo.Entry{
    Name: request.Name,
    URL:  request.URL,
}

applyRequestFields(newRepo, request)
...
repoFile.Update(newRepo)
```

`repo.File.Update` is a **wholesale replacement**, not a merge. From
helm.sh/helm/v3 v3.21.0 (the version pinned in `backend/go.mod`),
`pkg/repo/repo.go`:

```go
func (r *File) update(e *Entry) {
	for j, repo := range r.Repositories {
		if repo.Name == e.Name {
			r.Repositories[j] = e   // full replacement
			return
		}
	}
	r.Add(e)
}
```

So calling `AddRepo` for a repository that already exists silently discards every
field not present on the freshly built entry:

- `CertFile`, `KeyFile`, `CAFile` — always lost, since
  [`AddUpdateRepoRequest`](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L52-L59)
  deliberately does not carry them;
- `Username` / `Password` — lost whenever the request omits them, because
  [`applyRequestFields`](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L121-L137)
  only assigns non-nil pointers;
- `InsecureSkipTLSverify`, `PassCredentialsAll` — likewise.

**Expected:** re-adding an existing repository preserves the server-side state
that the HTTP API intentionally does not expose.

**Actual:** that state is destroyed, and subsequent chart operations against a
private or mTLS-protected repository fail with authentication or TLS errors until
the user reconfigures it outside Headlamp.

### Why this is clearly unintended, not a design choice

Three places in this same file say the opposite:

1. The request struct's own comment promises preservation —
   [L48-L51](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L48-L51):

   > `CertFile`, `KeyFile`, and `CAFile` are intentionally omitted from this
   > struct. Accepting filesystem paths from HTTP clients would allow arbitrary
   > local file reads on the server. **These values are preserved from the
   > existing repo entry only (server-side state).**

2. The sibling update path *does* implement that preservation, via
   `copyExistingRepoFields` —
   [L380-L388](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L380-L388)
   and its call site at
   [L446](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L446).
   `addRepository` is missing that one call.

3. `UpdateRepository`'s doc comment routes create-or-update traffic *to the
   broken path* —
   [L390-L393](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L390-L393):

   > It returns `errRepositoryNotFound` if the named repository does not exist.
   > **Callers that need create-or-update behaviour should use `AddRepo`
   > instead.**

## To Reproduce

Steps to reproduce the bug:

1. Configure a Helm repository with credentials outside Headlamp, so the
   server-side fields are populated:
   `helm repo add private https://charts.example.com --username u --password p --ca-file /etc/ssl/ca.pem`
2. Start Headlamp with Helm support enabled (`-enable-helm`).
3. `POST` to the add-repo endpoint with the same name and URL and no credentials:
   ```bash
   curl -X POST http://localhost:4466/helm/repositories \
     -H 'Content-Type: application/json' \
     -d '{"name":"private","url":"https://charts.example.com"}'
   ```
4. Inspect `repositories.yaml` (`helm env HELM_REPOSITORY_CONFIG`). The `private`
   entry has lost `username`, `password` and `caFile`.
5. Any subsequent chart listing or install against `private` now fails to
   authenticate.

 - Installation type: any install with Helm support enabled (`-enable-helm`)
 - Headlamp Version: 0.44.0 (`506465d78ca162f65e46c57fab7b014fd771d047`)
 - Other: `backend/pkg/helm`, helm.sh/helm/v3 v3.21.0

Yes (I will propose a PR)

### Suggested fix

Reuse the helper that already exists a few lines below, mirroring
`updateRepositoryWithRequest`:

```diff
--- a/backend/pkg/helm/repository.go
+++ b/backend/pkg/helm/repository.go
@@ func addRepository
 	newRepo := &repo.Entry{
 		Name: request.Name,
 		URL:  request.URL,
 	}
 
+	// repo.File.Update replaces the entry wholesale, so carry over the
+	// server-side fields (credentials, TLS file paths) that the HTTP API
+	// deliberately does not accept from clients.
+	if existing := repoFile.Get(request.Name); existing != nil {
+		copyExistingRepoFields(newRepo, existing)
+	}
+
 	applyRequestFields(newRepo, request)
```

Note the ordering: `copyExistingRepoFields` must run *before*
`applyRequestFields`, so that fields the caller *did* supply still win. This
matches the existing sequence in `updateRepositoryWithRequest`
([L446-L448](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L446-L448)).

Suggested test: seed a `repositories.yaml` containing an entry with
`Username`/`Password`/`CAFile`, call `addRepository` with only `Name`+`URL`, and
assert those fields survive; plus a case asserting an explicitly supplied
`Username` still overrides the stored one.

### Related but separate

While in this file I also noticed `defaultNewConfigFileMode` is `0o644`
([L39](https://github.com/kubernetes-sigs/headlamp/blob/506465d78ca162f65e46c57fab7b014fd771d047/backend/pkg/helm/repository.go#L39)),
whereas upstream Helm writes `repositories.yaml` with `0o600` because it can hold
plaintext basic-auth credentials (`pkg/cmd/repo_add.go`:
`f.WriteFile(o.repoFile, 0o600)`). I'll file that separately unless you'd prefer
it folded in here.

CygnusMaximillian:
## Summary 
addRepository built a fresh repo.Entry and called repo.File.Update, which replaces the existing entry wholesale. Re-adding an existing repository silently wiped Username/Password (when omitted from the request) and always wiped CertFile/KeyFile/CAFile, which the HTTP API intentionally never accepts from clients. Subsequent operations against private/mTLS repositories then failed until reconfigured outside Headlamp. This PR carries over the existing entry's fields before applying request fields, mirroring updateRepositoryWithRequest.

Related Issue Fixes #6840

## Changes

addRepository now calls the existing copyExistingRepoFields helper for pre-existing entries, before applyRequestFields, so request-supplied values still override.
Add TestAddRepoPreservesExistingServerSideFields: seeds repositories.yaml with username/password/CAFile, re-adds with only name+URL against a TLS + basic-auth index server (its cert is the CA file), asserts 200 and all fields preserved.
Add TestAddRepoRequestFieldsOverrideExisting: request credentials override stored ones.

## Steps to Test

cd backend && go test ./pkg/helm -run 'TestAddRepoPreserves|TestAddRepoRequestFields' -count=1 -v
Optionally the manual repro from the issue (helm repo with credentials, re-add via API, inspect repositories.yaml).

## Notes for the Reviewer 
The preservation also applies before DownloadIndexFile, so the index refresh itself uses the stored credentials/CA — that's the exact failure mode the TLS test locks in (it returns 500 without the fix)

##Screenshot of Tests

image

### DIFF
--- a/backend/pkg/helm/repository.go
+++ b/backend/pkg/helm/repository.go
@@ -171,6 +171,14 @@ func addRepository(request AddUpdateRepoRequest, settings *cli.EnvSettings) erro
 		URL:  request.URL,
 	}
 
+	// repo.File.Update replaces the entry wholesale, so carry over the
+	// server-side fields (credentials, TLS file paths) that the HTTP API
+	// deliberately does not accept from clients. Fields supplied in the
+	// request still win because applyRequestFields runs afterwards.
+	if existing := repoFile.Get(request.Name); existing != nil {
+		copyExistingRepoFields(newRepo, existing)
+	}
+
 	applyRequestFields(newRepo, request)
 
 	r, err := repo.NewChartRepository(newRepo, getter.All(settings))
--- a/backend/pkg/helm/repository_test.go
+++ b/backend/pkg/helm/repository_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"encoding/pem"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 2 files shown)


illume (approved):
🎉 thanks!

PR #6898, opened 2026-08-05
frontend: resourceMap: Persist source selection in localStorage

### LINKED ISSUE #6555: Map view intermittently never finishes loading; one default source can hold the whole view in the spinner state
# BUG: Map view intermittently never finishes loading; one default source can hold the whole view in the spinner state

## Summary

On a kind cluster running Kubernetes 1.36, the Map view
intermittently never finishes loading: it shows a spinner
indefinitely until the user manually deselects one default
source. When the hang occurs, deselecting the Workloads source
unblocks rendering immediately. The deselection does not persist
across navigation or reload, so on affected clusters the Map is
unreliable by default and the workaround must be repeated.

The bug is intermittent — the same cluster, same Headlamp build,
and same navigation path sometimes load fine. Detail on observed
frequency and conditions below, since that's the hard part of
reproducing this.

## Environment

- Headlamp: 0.43.0 (darwin/arm64); `headlamp-server` and the
  desktop app's frontend bundle from the same 0.43.0 install,
  frontend driven in Chrome against `headlamp-server`
- Cluster: kind, Kubernetes v1.36.1 node image
- Notable cluster state: APIs removed since ~1.22 are absent
  (Workloads-related legacy queries 404); JobSet CRD initially
  absent. Installing the JobSet CRDs did NOT prevent the hang,
  and — see below — deleting them again did not bring it back,
  so this is not simply "install the missing CRD".
- Plugins loaded: an in-development plugin that registers a map
  source via registerMapSource
  (https://github.com/danbruno101/plugins/tree/kro-plugin-pr).
  The hang implicates a core default source, not the plugin
  source — see isolation below, including one hung load with
  the plugin removed entirely.

## Observed behavior and frequency

Firsthand account from one working session (~2.5 h, one browser
tab, one `headlamp-server` process, one cluster):

- **~15 Map loads total: 7 hung, 8 rendered.** The outcomes were
  not interleaved — they came in two contiguous windows.
- **Hang window (~15 minutes, starting with the tab's first-ever
  visit to the Map):** 7 consecutive loads hung, including
  in-app navigations and full hard reloads (Cmd+Shift+R). During
  this window the hang was 100% reproducible; deselecting the
  Workloads source rendered the graph instantly every time, and
  the rest of the sources' data was visibly complete the moment
  it appeared.
- **Success window (starting ~75 minutes later, same tab, same
  server, same cluster):** 8 consecutive loads rendered in a few
  seconds each — in-app navigations, hard reloads, and
  `?node=<uid>` deep links alike. No hang has recurred since.
- Correlations checked, with negative results spelled out:
  - **Not the third-party plugin**: one hung load was performed
    with the plugin removed from the plugins directory entirely
    (hard reload; still hung).
  - **Not hard reload vs. in-app navigation**: both kinds of
    load hung inside the hang window and both succeeded inside
    the success window.
  - **Not browser cache**: cache-bypassing hard reloads hung
    during the hang window.
  - **Not the JobSet CRD / its 404** (our first theory): one
    load hung ~30 s *after* installing the JobSet CRDs (possibly
    an API-discovery race, but a hang nonetheless), and — the
    decisive negative — after the success window began, we
    **deleted the JobSet CRD again** and the Map still rendered
    fine on consecutive loads with `jobsets` 404ing.
  - **No monotonic correlation with server or cluster age**: the
    same long-running `headlamp-server` served both windows.
- Net observation: whatever wedges is **stable within a window
  and survives hard page reloads**, which suggests state outside
  the page's JS context (server-side, or the watch/multiplexer
  connection) rather than a per-load coin flip.

## Isolation

When the hang occurs, deselecting sources identifies the
**Workloads** source group as the wedge: with it deselected the
graph renders immediately with all other sources; re-selecting
it brings the spinner back. The aggregate isLoading appears to
be all-or-nothing across sources, so one sub-source whose query
never settles holds the entire view in the loading state.

(We did not manage to narrow it to a single child sub-source
inside Workloads before the hang stopped reproducing — and the
network evidence below suggests the wedged child is not
distinguishable at the HTTP layer at all.)

Devtools-level request behavior, captured on both hung and
successful loads:

- **On hung loads** (full request capture, ~245 requests):
  every Workloads child list GET returned **200 OK** — pods,
  deployments, statefulsets, daemonsets, replicasets, jobs,
  cronjobs. The only Workloads-adjacent failure was
  `apis/jobset.x-k8s.io/v1alpha2/jobsets` → **404, failing
  fast** (tens of ms, not pending), re-issued periodically
  (~once a minute). Other 404s (`extensions/v1beta1/ingresses`,
  `apiextensions.k8s.io/v1beta1/customresourcedefinitions`,
  `autoscaling.k8s.io/v1`) also failed fast. **Zero requests
  stayed pending** on hung loads.
- **On successful loads**: the request set and statuses are
  **identical** to the hung loads (modulo `jobsets` being 200 or
  404 depending on whether the CRD existed — both outcomes
  produced successful renders). The only long-pending request
  observed anywhere was a watch-style
  `apiextensions.k8s.io/v1/customresourcedefinitions` GET — on a
  load that rendered fine, so pending-ness there is not the
  wedge either.
- Conclusion from the capture: **the hung and successful loads
  are indistinguishable at the HTTP layer.** The blocking state
  is internal — a source's query result never being surfaced to
  the sources aggregation (or an initial watch snapshot never
  delivered over the WebSocket multiplexer, which the request
  log doesn't cover) — rather than any request outcome.

## Repro

1. kind cluster with a Kubernetes 1.36 node image
2. Open the Map view; if it renders, navigate away and return
   (or reload) — repeat. In our session the hang owned the tab's
   first ~15 minutes of Map visits (7/7 loads) and then never
   recurred (0/8); expect window-like behavior rather than a
   per-attempt probability.
3. When hung: deselect the Workloads source → renders
   immediately
4. Navigate away and back — deselection is reset, spinner risk
   returns

## Impact

Any plugin that deep-links users into the Map lands them on a
spinner when the hang strikes (hit while building a kro plugin
that adds "View in Map" buttons on detail pages). Because it's
intermittent, it's worse in demos and docs than a deterministic
failure — it can't be preflighted.

## Suggested fixes (any subset helps)

- Treat per-source loading independently: render what has
  settled, badge sources that are still loading or erroring,
  never let one source block the whole view. (Our capture shows
  the 404s already fail fast and nothing stays pending, so
  time-boxing requests likely isn't the fix — the wedge appears
  to be in how a completed source result reaches the loading
  aggregation, or in watch-snapshot delivery over the
  multiplexer. Independent per-source loading sidesteps
  whatever the internal race is.)
- Persist the user's source selection across navigation/reloads
  — this alone converts the workaround from "every visit" to
  "once". I'd be happy to submit a PR for this piece if the
  approach sounds right.

harshitnub077:
Fixes #6555. In `GraphSourceManager`, the selected resource map sources are currently reset on every navigation or page reload. On clusters where certain workload queries hang or fail, this forces the user to manually deselect the Workloads group on every single visit to avoid infinite spinners.

This PR loads the list of disabled sources from `localStorage` during initialization and saves any subsequent changes to `selectedSources` using a `useEffect`.

**Changes:**
- `frontend/src/components/resourceMap/sources/GraphSources.tsx`: Initialize `selectedSources` from `localStorage` and persist selections inside `useEffect`.
- `frontend/src/components/resourceMap/sources/GraphSources.test.tsx`: Add a unit test to verify that selections are persisted and correctly restored, along with `beforeEach` to reset `localStorage` state between test runs.

### DIFF
--- a/frontend/src/components/resourceMap/sources/GraphSources.test.tsx
+++ b/frontend/src/components/resourceMap/sources/GraphSources.test.tsx
@@ -109,6 +109,10 @@ describe('GraphSources helpers', () => {
 describe('GraphSourceManager', () => {
   let context: ReturnType<typeof useSources>;
 
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
   const Probe = () => {
     const value = useSources();
     useEffect(() => {
@@ -239,5 +243,38 @@ describe('GraphSourceManager', () => {
         label: undefined,
       },
     ]);
+  });
+
+  it('persists selection to localStorage and loads it on initialization', async () => {
+    const first = source('first', { nodes: [] });
+    const second = source('second', { nodes: [] });
+    const group: GraphSource = { id: 'group', label: 'Group', sources: [first, second] };
+
+    // Initially all are selected
+    const { unmount } = render(
+      <GraphSourceManager sources={[group]} relations={[]}>
+        <Probe />
+      </GraphSourceManager>
+    );
+    await waitFor(() => expect(context.isLoading).toBe(false));
+    expect(context.selectedSources).toEqual(new Set(['group', 'first', 'second']));
+
+    // Deselect 'first'
+    act(() => context.toggleSelection(first));
+    expect(context.selectedSources).toEqual(new Set(['group', 'second']));
+
+    // Check localStorage contains the disabled source ID
+    expect(localStorage.getItem('headlamp_resource_map_disabled_sources')).toContain('first');
+
+    unmount();
+
+    // Re-render and check that selection was restored
+    render(
+      <GraphSourceManager sources={[group]} relations={[]}>
+        <Probe />
+      </GraphSourceManager>
+    );
+    await waitFor(() => expect(context.isLoading).toBe(false));
+    expect(context.selectedSources).toEqual(new Set(['group', 'second']));
   });
 });
--- a/frontend/src/components/resourceMap/sources/GraphSources.tsx
+++ b/frontend/src/components/resourceMap/sources/GraphSources.tsx
@@ -185,17 +185,40 @@ export function GraphSourceManager({ sources, children, relations }: GraphSource
   const [selectedSources, setSelectedSources] = useState(() => {
     const _selectedSources = new Set<string>();
 
+    let disabledSources: string[] = [];
+    try {
+      const stored = localStorage.getItem('headlamp_resource_map_disabled_sources');
+      if (stored) {
+        disabledSources = JSON.parse(stored);
+      }
+    } catch (e) {
+      console.error('Error loading disabled map sources from localStorage:', e);
+    }
+    const disabledSet = new Set(disabledSources);
+
     const step = (source: GraphSource) => {
-      if (source.isEnabledByDefault ?? true) {
+      if (disabledSet.has(source.id)) {
+        // Skip selecting this source
+      } else if (source.isEnabledByDefault ?? true) {
         _selectedSources.add(source.id);
-        if ('sources' in source) {
-          source.sources.forEach(step);
-        }
+      }
+      if ('sources' in source) {
+        source.sources.forEach(step);
       }
     };
     sources.map(step);
     return _selectedSources;
   });
+
+  useEffect(() => {
+    const allFlatSources = getFlatSources(sources);
+    const disabled = allFlatSources.filter(s => !selectedSources.has(s.id)).map(s => s.id);
+    try {
+      localStorage.setItem('headlamp_resource_map_disabled_sources', JSON.stringify(disabled));
+    } catch (e) {
+      console.error('Error saving disabled map sources to localStorage:', e);
+    }
+  }, [selectedSources, sources]);
 
   const toggleSelection = useCallback(
     (source: GraphSource) => {


harshitnub077:
i have updated the PR to address all feedback:
- **Disabled-by-default groups**: child traversal now stays within the `isEnabledByDefault` branch during initial load ensuring groups like Security and Configuration don't unexpectedly load queries on a fresh visit.
- **Explicit selection persistence**: Switched from storing disabled leaf IDs to persisting the explicit list of selected IDs (`headlamp_resource_map_selected_sources`). This allows enabling default-disabled groups and disabling default-enabled groups to persist seamlessly across reloads.
- **Safe parsing**: Added array and string type validation for localStorage data to prevent unexpected runtime errors.
- **Tests**: Added test cases for default-disabled group preservation, selection persistence, and corrupted storage fallbacks.

illume (changes requested):
Thanks for the contribution.


harshitnub077:
@illume! i’ve rebased on latest `main` and updated the commit message and all of copilot’s feedback was addressed in the code and i’ve marked those threads resolved now also Linters and tests are all green and ready for another look

illume (changes requested):
Thanks for the contribution.


harshitnub077:
resolved all the open comment threads. The branch is rebased and good to go!

harshitnub077:
hi @illume all comments and formatting issues are resolved locally and all fifteen unit tests as well as linters are passing cleanly could you please approve the workflow runs when you have a moment thanks

PR #6909, opened 2026-08-05
0043: app: Derive the Linux executable name

illume:
## Summary

Remove the hardcoded Linux `executableName` so Electron Builder derives it from the package and product metadata

## Provenance

This upstreams patch 0043 retained by [Azure/aks-desktop#823](https://github.com/Azure/aks-desktop/pull/823). The source patch metadata identifies commit `a13a74312748b7247cc8f3487e47a10534d89abb`, authored by Joaquim Rocha on 2026-07-31. That SHA is recorded here as source provenance; this description does not claim that it resolves publicly.

Related umbrella tracking: [Azure/aks-desktop#153](https://github.com/Azure/aks-desktop/issues/153).

## Proof this still works?

`.github/workflows/app.yml` runs the Linux package verification, and `app/scripts/verify-build-linux.sh` finds and launches the packaged `headlamp` executable with `list-plugins`. That existing package smoke directly exercises the derived executable name, so no duplicate end-to-end test is added.

Assisted by copilot

### DIFF
--- a/app/electron/package.test.ts
+++ b/app/electron/package.test.ts
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


ashu8912 (commented):
This change originally came from an AKS Desktop commit, but looking at it again, I’m wondering whether we should carry it over here.

What motivated the change originally? The net effect is that we remove an explicit pin and add a test that enforces its absence. While “don’t set a value when it matches the default” is reasonable, it also replaces an explicit guarantee around a user-visible artifact name with an implicit dependency on the default behavior. Unless this was addressing a concrete issue in AKS Desktop that also applies here, I’m not sure the trade-off is worth making.

illume:
Yeah... the value is limited. Except if someone adds it back, it will break downstream. So for me, I think it's still worth it. Because this would stop breaking downstream if someone in the future configures this again.

That's my understanding anyway. Please correct me if I'm wrong, and we can close this.

ashu8912 (approved):

PR #6748, opened 2026-07-30
frontend: StatusCharts: Add interactive red/green query links for filtered list views

Suyog241005:
## Summary
Adds interactive click navigation to Cluster Overview status pie charts (`PodsStatusCircleChart`, `NodesStatusCircleChart`) so users can drill down directly to filtered list views of troubleshooting resources (e.g. unready pods, cordoned nodes, or unready nodes).

Fixes #6734

## Changes
- **`Chart.tsx`**:
  - Enhanced `PercentageCircle` to render interactive `<g role="button" tabIndex={0}>` sectors when `onClick` is provided, retaining standard focus visibility for keyboard navigation.
  - Restricted `activeIndex` to positive, finite segments (`> 0%`) to avoid rendering phantom tab stops or invalid accessible labels on 0-value sectors.
- **`StatusCharts.tsx`**:
  - Linked Pods ready / not-ready pie slices to filtered list view routes (`/c/:cluster/pods?podsfilter=Ready`, `/c/:cluster/pods?podsfilter=NotReady`).
  - Linked Nodes ready / cordoned / not-ready pie slices to filtered list view routes (`/c/:cluster/nodes?nodesfilter=Ready`, `?nodesfilter=Cordoned`, `?nodesfilter=NotReady`).
  - Set the Drained slice as non-interactive to accurately reflect that node list views filter on node object state.
- **`pod/List.tsx` & `node/List.tsx`**:
  - Added structured readiness filtering (`filterPod`, `filterNode`) supporting `Ready`, `NotReady`, and `Cordoned` query keywords.
  - Preserved initial table row pre-filtering when search query is undefined (`search === undefined`).
- **Tests**:
  - Added unit tests for `filterPod` and `filterNode` in `pod/List.test.tsx` and `node/List.test.tsx`.
  - Updated Storybook storyshots and verified full accessibility compliance with `test:a11y`.

## Steps to Test
1. Start Headlamp dev server: `npm start` (or `npm run frontend:start`).
2. Open the Cluster Overview page: `http://localhost:3000/c/:cluster`.
3. Hover over the **Pods** or **Nodes** status circle charts and observe `cursor: pointer` on interactive slices.
4. Click on chart slices:
   - Clicking Pods **Ready** slice navigates to `/c/:cluster/pods?podsfilter=Ready` with table filtered to ready pods.
   - Clicking Pods **Not Ready** slice navigates to `/c/:cluster/pods?podsfilter=NotReady` with table filtered to unready pods.
[...]

Issue #3175, opened 2025-04-25
Remember filter when going back


yajo:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
I'm investigating a problem in some pods. I filter pods. Enter each one of them. After inspection, I go back.

When I go back, I have to add the pod filter again. It's annoying.

https://github.com/user-attachments/assets/81c786f4-8bc5-4b19-8408-96703bff2e80

## Describe the solution you'd like
Filter remembered when going back in history.

## What users will benefit from this feature?
Everyone.

## Are you able to implement this feature?
No.

## Additional context
[If needed, add any other context or screenshots about the feature request here.]

joaquimrocha:
Hi @yajo . I agree this can be a non-intuitive behavior. I have a question though: do you think we should also maintain the filter if you navigate away by clicking something outside the table, i.e. Filtering, then clicking the Security sidebar tab, then clicking to go back to pods?

BTW, if you enable the details overlay pane (in the General settings), you can open the details without navigating away from the table and thus the filter remains.

yajo:
> do you think we should also maintain the filter if you navigate away by clicking something outside the table, i.e. Filtering, then clicking the Security sidebar tab, then clicking to go back to pods?

I think so. Not when entering from the left menu, but yes when going "back" from browser history or the back button rendered by HL.

> BTW, if you enable the details overlay pane (in the General settings), you can open the details without navigating away from the table and thus the filter remains.

Thanks! I'll try that

sniok:
Completed in #3354

PR #6918, opened 2026-08-05
frontend: Lists: Reflect the search filter in the URL

Adi-204:
## Summary

Enables `reflectInURL` on the list views that were missing it, so the search filter (and page) is kept in the URL on every single-table list instead of only on a handful of them.

Today `?filter=` works on Pods, Jobs, JobSets, LimitRanges, ResourceQuotas and the Workloads overview, because those views hand-roll their table and opt into `reflectInURL`. Every other list renders through `ResourceListView`, which never received the prop, so the filter is lost on reload and on back navigation, and a filtered list cannot be shared as a link.

## Related Issue

Follow-up to #3175 ("Remember filter when going back"), which introduced the behaviour in #3354 but only for the views that already had `reflectInURL` enabled.

## Changes

- Added `reflectInURL` to 35 single-table list views:
  - **Config**: ConfigMaps, Secrets, PriorityClasses, RuntimeClasses, Leases, PodDisruptionBudgets, HPAs, VPAs, MutatingWebhookConfigurations, ValidatingWebhookConfigurations
  - **Network**: Services, Ingresses, IngressClasses, Endpoints, EndpointSlices, NetworkPolicies
  - **Storage**: PersistentVolumeClaims, PersistentVolumes, StorageClasses, VolumeAttributesClasses
  - **Security**: ServiceAccounts, Roles, RoleBindings, Namespaces
  - **Gateway API**: Gateways, GatewayClasses, HTTPRoutes, GRPCRoutes, ReferenceGrants, BackendTLSPolicies, BackendTrafficPolicies
  - **Custom resources**: CRDs, CustomResourceList, CustomResourceInstancesList
  - **Cluster overview**: the Events table

The prop is passed as a bare boolean, which gives an empty prefix and therefore the same unprefixed `?filter=` key the Pods list already uses.

## Steps to Test

1. Run Headlamp against any cluster: `npm start`, then open `http://localhost:3000`.
2. Go to **Config → Secrets** and type something into the table search box.
3. Observe the URL becomes `.../secrets?filter=<text>`.
4. Reload the page: the filter is still applied and the search box is open.
5. Open one of the listed Secrets, then press **Back**: the filter is restored instead of cleared.
6. Copy the filtered URL into a new tab and confirm it opens already filtered.
7. Repeat on any other list touched here (for example Services, PersistentVolumeClaims, or a Gateway API list) — behaviour should be identical to the existing Pods list.

No visual change: the tables render exactly as before, the only difference is the query string. Verified by the storyshot suite, which passes with no snapshot changes.

## Notes for the Reviewer

- **Deliberately not included:** CronJobs, DaemonSets, Deployments, Nodes, ReplicaSets and StatefulSets. #6748 already adds the identical line to those six, so leaving them out avoids a conflict. They can be folded in here if that PR stalls.
- **Also not included:** `project/ProjectDetails.tsx` and `verticalPodAutoscaler/Details.tsx`. Both render several tables on one page, so they need distinct prefixes (`reflectInURL="..."`) rather than the bare flag, and picking those keys is a per-table judgement call rather than a mechanical change.
- **Scope of the flag:** `reflectInURL` is a single switch that controls URL persistence for the search filter *and* pagination. These pages therefore also start reflecting `p` and `perPage`. That matches the existing behaviour of the Pods list, but it is a visible URL change on 35 pages, so worth a conscious ack.
- An alternative to touching each page would be to default `reflectInURL` to true inside `ResourceListView`. I avoided that because it silently changes behaviour for every consumer, including plugins that render their own lists through it. Happy to switch approach if maintainers prefer the central default.

## Validation

- `npm run frontend:tsc` — passes
- `npm run frontend:lint` — eslint passes; the 14 Prettier warnings it reports are pre-existing on a clean tree (yaml/css/md files untouched by this PR)
- Storyshots (`src/storybook.test.tsx`) — 651 passed, no snapshot changes

## Summary by CodeRabbit

* **New Features**
  * List and filter state is now reflected in the URL across resource views, including workloads, networking, storage, access control, and configuration resources.
  * Resource list views can now be bookmarked, shared, and restored with their current state intact.

### DIFF
--- a/frontend/src/components/cluster/Overview.tsx
+++ b/frontend/src/components/cluster/Overview.tsx
@@ -175,6 +175,7 @@ function EventsSection() {
 
   return (
     <ResourceListView
+      reflectInURL
       title={t('glossary|Events')}
       headerProps={{
         noNamespaceFilter: false,
--- a/frontend/src/components/configmap/List.tsx
+++ b/frontend/src/components/configmap/List.tsx
@@ -25,6 +25,7 @@ export default function ConfigMapList() {
     <ResourceListView
       title={t('glossary|Config Maps')}
       resourceClass={ConfigMap}
+      reflectInURL
       columns={[
         'name',
         'namespace',
--- a/frontend/src/components/crd/CustomResourceInstancesList.tsx
+++ b/frontend/src/components/crd/CustomResourceInstancesList.tsx
@@ -81,6 +81,7 @@ function CrInstancesView({ crds }: { crds: CRD[]; key: string }) {
         </Alert>
       )}
       <ResourceListView
+        reflectInURL
         title={t('translation|Custom Resource Instances')}
         headerProps={{
           noNamespaceFilter: false,
--- a/frontend/src/components/crd/CustomResourceList.tsx
+++ b/frontend/src/components/crd/CustomResourceList.tsx
@@ -187,6 +187,7 @@ export function CustomResourceListTable(props: CustomResourceTableProps) {
         ),
       }}
       resourceClass={CRClass}
+      reflectInURL
       columns={cols}
     />
   );
--- a/frontend/src/components/crd/List.tsx
+++ b/frontend/src/components/crd/List.tsx
@@ -41,6 +41,7 @@ export default function CustomResourceDefinitionList() {
 
   return (
     <ResourceListView
+      reflectInURL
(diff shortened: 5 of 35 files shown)


illume (changes requested):
Thanks for this PR.


<details>

</details>

PR #6938, opened 2026-08-06
frontend: Empty: Add quick actions to empty resource list pages

### LINKED ISSUE #5708: frontend: Empty state pages should provide quick navigation actions
## Feature Request

Empty resource pages currently show limited guidance when no resources exist.

## Suggested Improvement

Add quick actions such as:
- Create Resource
- Refresh
- Change Namespace
- View Documentation

## Benefit

Improves onboarding and usability for new Kubernetes users.

WasThatRudy:
https://github.com/user-attachments/assets/78f6cf36-78b1-4bf1-9389-fc6f681a4601

Fixes #5708

## Summary
- Empty resource list pages currently show only a "No data to be shown." message with no next step. Add a "Create <Kind>" button (RBAC-gated via `AuthVisible`) and a "Refresh" button below the message so users have a next step from the empty state.
- The `Empty` component gains an optional `actions` prop that renders below the text inside a `role="group"` landmark for screen readers.
- `Table` and `ResourceTable` gain an `emptyActions` pass-through.
- `ResourceListView` auto-injects `<EmptyStateActions resourceClass={resourceClass} />` when a `resourceClass` is provided AND the caller did not set `emptyActions`. Nested embedded lists on Details pages (no `resourceClass`) do NOT get the auto-default, to avoid noise on non-top-level lists. Passing `emptyActions={null}` opts out even when a class is present.
- `CreateResourceButton` gains a `variant?: 'icon' | 'labeled'` prop. The labeled variant renders an MUI Button with icon + text and is used by the new `EmptyStateActions` helper. The default icon variant is unchanged and continues to be used by existing header toolbars.
- Refresh calls `queryClient.refetchQueries({ type: 'active' })` so only the current view's queries are refetched. Rejection is swallowed since transient K8s API errors already surface through the existing snackbar path.

## Non-goals (deferred)
- "Change Namespace" quick action — redundant with the header namespace filter already present on the same page.
- "View Documentation" quick action — headlamp has no per-resource doc anchors to link to.
- Applying the pattern to non-list empty states (detail sub-tables, etc.) — separate follow-up if desired.
- `AuthVisible` layout jank (Create button pops in after the SSAR resolves, shifting the row on wide viewports). This is existing behavior across every AuthVisible caller in the codebase; fixing it would require a Skeleton placeholder in `AuthVisible` itself. Out of scope here.

## Test plan
- [x] `npx vitest run src/components/common src/components/common/Resource src/components/common/Table` (141 tests pass across 28 files, includes 13 new tests: 4 for `EmptyStateActions`, 2 for `CreateResourceButton` variants, 4 for `ResourceListView` defaulting, plus a new `EmptyContent.WithActions` storyshot)
- [x] `npx vitest run src/storybook.test.tsx` (662 tests pass; 4 top-level list storyshots updated for DaemonSet / ServiceAccount / StatefulSet)
- [x] `npm run tsc` (clean)
- [x] `npm run i18n` (three new English keys: "Refresh", "Refresh list", "Empty state actions" synced to all 17 locale files)
- [ ] Manual: open a resource list with zero items in the selected namespace (e.g. Deployments on an empty namespace), confirm the "Create Deployment" + "Refresh" buttons appear below the "No data to be shown." message
- [ ] Manual: click Create — confirm the same Activity opens as clicking the top-bar Create button
- [ ] Manual: click Refresh — confirm React Query refetches the list without a full page reload
- [ ] Manual: on a user account that lacks `create` RBAC for the resource, confirm the Create button does not appear (only Refresh)
- [ ] Manual: open a Details page whose nested embedded list is empty (e.g. StatefulSet with no Events), confirm no auto-default actions row appears there

## Summary by CodeRabbit

- **New Features**
  - Added Create and Refresh actions to empty resource-list states.
  - Added labeled and icon-only Create button variants.
  - Added support for custom actions beneath empty-state messages.
  - Refresh actions reload active list data while handling transient errors gracefully.
- **Documentation**
  - Added Storybook coverage for empty states with actions.
- **Localization**
  - Added translation keys for empty-state actions and refresh controls.
- **Tests**
  - Added coverage for button variants, empty-state actions, refresh behavior, and customization options.


illume (changes requested):
Thanks for this PR.


There are snapshot mismatches in CI. You can fix them locally by running `cd frontend && npm run test -- -u`.

<details>

</details>

WasThatRudy:
> Thanks for this PR.
> 
> There are snapshot mismatches in CI. You can fix them locally by running `cd frontend && npm run test -- -u`.
> 
> How to update snapshots

done

illume:
Thanks for this.

Can you please add a video to the PR description?

WasThatRudy:
Added video as you asked @illume :)

frontend/src/components/common/Resource/ResourceListView.tsx
```diff
@@ -54,6 +55,16 @@ export default function ResourceListView(
   const resourceClass = (props as ResourceListViewWithResourceClassProps<any>)
     .resourceClass as KubeObjectClass;
 
+  // Default the empty-state actions to a Create + Refresh row on top-level
+  // resource list views (identified by a `resourceClass` prop). Nested
+  // embedded lists on Details pages don't get the auto-default; they can
+  // still opt in by passing `emptyActions` explicitly. Passing
+  // `emptyActions={null}` opts out even when a class is present.
+  let emptyActions = tableProps.emptyActions;
+  if (emptyActions === undefined && resourceClass) {
```
xonas1101, line 64: This condition assumes resourceClass means "top-level list", but 12 of the 47 ResourceListView call sites pass data= instead because they fetch their own items for metrics/filtering, including Pods (pod/List.tsx:553), Jobs (job/List.tsx:222), Secrets, LimitRange, ResourceQuota,
  Roles, RoleBindings, Namespaces and the CRD lists. None of them get the actions. I verified on an empty namespace: Pods and Jobs show a bare "No data to be shown.", while Network Policies shows the buttons.

  That makes the first manual check in the description - "Pods on an empty namespace, confirm the Create Pod + Refresh buttons appear" , impossible to pass as written.

  It also splits pages against themselves: Jobs and Pods pass their own CreateResourceButton via titleSideActions (job/List.tsx:141, pod/List.tsx:278), so the header offers Create while the empty state right below it doesn't. Line 77 below already handles both shapes for the header
  (headerProps?.titleSideActions || resourceClass) , keying this default off the same condition would keep the two Create affordances consistent, and Details-page embedded lists would still opt out naturally.

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -724,6 +731,7 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
         globalFilterFn="kubeObjectSearch"
         filterFunction={filterFunc}
         getRowId={getResourceRowId}
+        emptyActions={emptyActions}
```
xonas1101: Minor, and partly pre-existing: when the list itself errored, the empty state still renders the actions. On a cluster without the JobSet CRD, Job Sets shows "Resource not found" and now a labeled "Create JobSet" directly beneath it; applying from there fails with Not Found. The header + already had this on main, so it's not new behavior — but a full-size CTA under an error message reads much more like a recommendation than the icon did. Since errors is in scope here, emptyActions={errors?.length ? undefined : emptyActions} would keep Create off the  error path. Refresh arguably still belongs there

xonas1101:
Hey @WasThatRudy I have added some review comments for your PR after manual testing. A short summary:
 - The create button and the refresh button do appear in most empty resource cases, but in pods , jobs and other resources I wasnt able to find them. 
 The other is a minor UX fix, 
 - On a cluster ***without*** the job set CRD, the create jobset action must fail, given, main has that button in the header as well, but a full sized button in place of a small "+" icon is what makes it sting. 
 tsc/lint/tests/storyshots are all green on my end.

WasThatRudy:
> * The create button and the refresh button do appear in most empty resource cases, but in pods , jobs and other resources I wasnt able to find them.

I am planning to open another PR for this after this one merges, the PR is ready. Just waiting for this to merge for that.
Unless @illume you want me to add that also in this PR.

> * On a cluster without the job set CRD, the create jobset action must fail, given, main has that button in the header as well, but a full sized button in place of a small "+" icon is what makes it sting. tsc/lint/tests/storyshots are all green on my end.

Let me look into this asap.

@xonas1101 Thank you so much for reviewing it :)

xonas1101:
> > * The create button and the refresh button do appear in most empty resource cases, but in pods , jobs and other resources I wasnt able to find them.
> 
> I am planning to open another PR for this after this one merges, the PR is ready. Just waiting for this to merge for that. Unless @illume you want me to add that also in this PR.

@WasThatRudy for the time being, update the PR description testing strategy, since that still mentions pods testing. It isnt exactly a one-liner so a new PR might make sense.

WasThatRudy:
> > > * The create button and the refresh button do appear in most empty resource cases, but in pods , jobs and other resources I wasnt able to find them.
> > 
> > 
> > I am planning to open another PR for this after this one merges, the PR is ready. Just waiting for this to merge for that. Unless @illume you want me to add that also in this PR.
> 
> @WasThatRudy for the time being, update the PR description testing strategy, since that still mentions pods testing. It isnt exactly a one-liner so a new PR might make sense.

@xonas1101 Updated the test plan to use Deployments as the example. You're right that Pods wasn't the right one to reference, extending the coverage set is the separate follow-up.

frontend/src/components/common/Resource/ResourceTable.tsx
WasThatRudy: Good catch, fixed. emptyActions is now suppressed in ResourceTable when errorMessage is set or errors is non-empty, so the Create button no longer shows under a "Resource not found" or similar failure. Added tests covering the three cases (no error → passthrough, errorMessage → suppressed, errors[] → suppressed).

frontend/src/components/common/Resource/ResourceListView.tsx
WasThatRudy: Yeah, follow-up PR will key off data as well so those 12 sites get covered. Will link it here once opened.

xonas1101 (commented):
Good updates - the work looks largely fine. One fine detail - the retry looks like a no-op. There is no loading banner or anything. Also, when I change something, like in an empty deployments list, when i create  a new deployment using kubectl, the thinng appears normally, due to web socket stuff. PTAL.
The PR needs a rebase as well. Cheers!

WasThatRudy:
> Good updates - the work looks largely fine. One fine detail - the retry looks like a no-op. There is no loading banner or anything. Also, when I change something, like in an empty deployments list, when i create a new deployment using kubectl, the thinng appears normally, due to web socket stuff. PTAL

thanks for pointing it out, will take a look

WasThatRudy:
@xonas1101 
Refresh now shows a loading state (spinner + "Refreshing", disabled while in flight, min-visible so instant refetches don't flash). 
On the websocket-already-covers-this point, agreed it's a smaller win when the socket is up. Keeping it as an explicit control for cold-start / disconnected cases. 
Rebased.

WasThatRudy:
The build-linux failure is a network flake — app-builder couldn't download electron: 
Get "https://github.com/electron/electron/releases/download/v38.8.6/electron-v38.8.6-linux-x64.zip": EOF
Not related to this PR.

illume (changes requested):
Thanks for these changes.


<details>

</details>

frontend/src/components/common/Resource/EmptyStateActions.tsx
```diff
@@ -0,0 +1,87 @@
...
+    setIsRefreshing(true);
+    try {
+      // Ensure the spinner is visible even when the refetch completes
+      // instantly (e.g. no active queries or cached response). Without this
+      // the button flashes and the click reads as a no-op.
+      await Promise.all([
+        // Silence transient errors (network drop, K8s API blip). The user
+        // can retry; failures already surface through the existing snackbar
+        // path.
+        queryClient.refetchQueries({ type: 'active' }).catch(() => {}),
```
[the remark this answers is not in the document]
WasThatRudy: Right, type: 'active' also refetches auth/discovery/sidebar queries. Deferring to a follow-up: scoping the refetch to the list's queryKey prefix needs ResourceListView
  to hand EmptyStateActions the key it built for its own list, which is a small plumbing change I'd rather keep out of this PR.

frontend/src/components/common/Resource/ResourceListView.tsx
```diff
@@ -54,6 +55,16 @@ export default function ResourceListView(
...
+    emptyActions = <EmptyStateActions resourceClass={resourceClass} />;
```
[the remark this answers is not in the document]
WasThatRudy: Same follow-up I already agreed to with @xonas1101  on the earlier coverage thread: the next PR will also key off data= so Pods / Jobs / Secrets / Namespaces / Roles / RoleBindings / LimitRange / ResourceQuota / CRD lists pick up the actions

Issue #6015, opened 2026-06-13
feature: Auto-discover accessible namespaces from RBAC (OIDC / namespace-scoped users)


rrk-dev:
# GitHub Issue Draft: RBAC-based namespace discovery

Use this when opening an issue at https://github.com/kubernetes-sigs/headlamp/issues/new?template=feature_request.md

---

## Title

```
feat(frontend): Auto-discover namespaces for users without cluster-wide list namespaces permission
```

---

## Summary

Namespace-scoped users (common with OIDC/OAuth) often cannot `list namespaces` cluster-wide, but **can** access specific namespaces via RoleBindings. Today Headlamp shows an empty namespace filter and empty resource lists unless users manually configure **Settings → Allowed namespaces**.

Propose **opt-in RBAC-based namespace discovery** in the frontend: when `GET /api/v1/namespaces` is denied, derive accessible namespaces from RoleBindings whose subjects match the authenticated user (via SelfSubjectReview identity).

---

## Motivation / problem

| Symptom | Cause |
|---------|--------|
| Empty namespace dropdown | `list namespaces` returns 403 |
| Empty Pods/Deployments lists | List routing uses cluster-wide API without namespace scope |
| Manual Settings override works | Per-namespace API calls succeed; discovery is the gap |

This affects OIDC deployments where cluster admins grant namespace-scoped RBAC but not cluster-wide namespace listing.

---

### Discovery chain (frontend only, no backend changes)

1. Try `GET /api/v1/namespaces` — if success, **unchanged legacy behavior**
2. If 403/401, `GET /rolebindings` cluster-wide and match subjects to user/groups from SelfSubjectReview
3. Optional SelfSubjectAccessReview validation per candidate namespace
4. Fallback to Settings / kubeconfig default namespace
5. Manual **Settings → Additional namespaces (override)** always takes precedence

### Backward compatibility

- Cluster-admins: unchanged (`list namespaces` succeeds → legacy path)
- Token/kubeconfig users: unchanged
- Discovery only applies when namespace list API is denied **and** RoleBinding discovery returns namespaces

### Cluster RBAC required (documented, not in Headlamp code)

Namespace-scoped users need a ClusterRole granting:

- `list` on `rolebindings` (cluster-wide)
- `create` on `selfsubjectaccessreviews` (optional, for validation)

Plus namespace RoleBindings for actual workload access.

---

## Security consideration

[...]

PR #6961, opened 2026-08-06
frontend: Overview: Scope charts to the accessible namespaces

### LINKED ISSUE #6897: Overview charts show 0 for users with limited namespace access (no cluster-wide list permissions for objects)
What happened:
When a user with RBAC permissions restricted to specific namespaces (e.g., via RoleBinding in a single namespace) logs into Headlamp, the workload overview charts (circular diagrams for Pods, Deployments, StatefulSets, etc.) display 0 for all categories, even though the user can navigate into the details of those resources and see real data.

What you expected to happen:
The charts should reflect the actual state of the resources that the user has access to (i.e., only those in the allowed namespaces), instead of showing zero values.

Anything else we need to know?
This happens because the Overview component uses useList() https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/components/workload/Overview.tsx#L40 without a namespace parameter, which issues a cluster‑wide LIST request for each resource type (e.g., /api/v1/pods). Users without cluster‑wide list permissions receive either an empty list or a 403 Forbidden, resulting in zero counts.

The issue is similar to #1420, but there the charts were spinning indefinitely; here they simply show zero.

Environment:
Headlamp version: 0.44.0 via Helm chart
Kubernetes version: 1.34
Installation method: Helm chart with OIDC auth
RBAC: user limited to specific namespaces

Possible solutions / suggestions
Dynamically discover accessible namespaces – Headlamp could query the Kubernetes API to determine which namespaces the current user has list permissions on (e.g., using SelfSubjectAccessReview or by inspecting the user's RBAC). Then it would issue per‑namespace requests accordingly.

Additional context
This is a common pain point for multi‑tenant clusters where users should only see their own workloads. Providing cluster‑wide list permissions (even for just metadata) is often undesirable for security reasons. A more granular approach would greatly improve the user experience for restricted users.

Anjali-Chauhan1:
## Summary

This PR fixes the overview charts showing `0` for users whose RBAC is limited to specific namespaces. The cluster and workload overviews called `useList()` without a namespace, so every chart issued a cluster-wide LIST that such users are refused — and the empty result was drawn as a confident zero. The charts now use the same namespace filter the list pages already pass, stop charting failures as zero, and point the user at the one setting that can fix a request nothing scopes.

## Related Issue

Fixes #6897

## Changes

- **Scoped the overview lists to the selected namespaces** in `cluster/Overview.tsx` (pods) and `workload/Overview.tsx` (all eight workload kinds). `makeListRequests` then fans out per namespace; an empty selection still falls back to Allowed namespaces and finally to one cluster-wide request, so nothing changes for users who can list the cluster.
- **Stopped charting a failed list as zero.** A list that returned nothing and an error is replaced by a tile carrying the chart's own title and "Unavailable". A list that partly succeeded keeps its chart — one forbidden namespace does not make the others uncountable.
- **Reported failures once per section** instead of once per affected chart. On the cluster overview a single node failure used to mean three identical alerts (CPU, Memory, Nodes); it is now one, above the grid.
- **Narrowed the metrics-403 handling**: a forbidden node-metrics response no longer blanks the whole overview section, which would have hidden the pods chart this fix repairs.
- **Added an actionable hint** to `ClusterGroupErrorMessage`: when a namespaced list is forbidden and no namespaces are selected or configured, it links to *Settings → Cluster → Allowed namespaces*. Wired through `ResourceTable` (`namespacedResource={resourceClass.isNamespaced}`), so every list page gets it too, and never for cluster-scoped resources where namespaces would not help.
- **Tests**: new `workload/Overview.test.tsx`, extended `cluster/Overview.test.tsx` — namespace scoping on both pages, placeholder instead of zero, one report per section, no namespace hint for a cluster-scoped failure, and partial results surviving alongside the error.

## Steps to Test

1. Create a namespace-restricted user and point a context at it:

   ```bash
   kubectl create ns team-a
   kubectl create rolebinding team-a-admin --clusterrole=admin --user=restricted -n team-a
   kubectl config set-context restricted --cluster=kind-kind --user=restricted --namespace=team-a
   ```

2. Open Headlamp with that context and go to **Cluster → Overview**. Before this PR the Pods chart reads `0`; with it, the chart counts the pods in `team-a`, while CPU/Memory/Nodes show *Unavailable* with one explanation above them.
3. Go to **Workloads**. Every chart is scoped the same way; kinds that cannot be listed show *Unavailable* rather than `0`.
4. Clear the namespace filter and remove any Allowed namespaces so nothing scopes the request. The charts report the permission failure and offer a link to *Cluster settings*, instead of silently showing zeros.
5. Run the unit tests:

   ```bash
   cd frontend && npx vitest run src/components/cluster src/components/workload src/components/common/Resource
   ```

   116 tests pass, including the new namespace-scoping and placeholder cases.

## Notes for the Reviewer

- **This does not discover namespaces.** Kubernetes has no API that enumerates the namespaces a user may read — `SelfSubjectAccessReview` and `SelfSubjectRulesReview` both require the namespace up front. So counts appear only once Headlamp knows the namespaces: the filter, the kubeconfig context's namespace, or Allowed namespaces. Auto-discovery is #6015; this PR makes the gap legible rather than silent.
- **Behaviour change worth your judgement:** a forbidden list now renders a labelled "Unavailable" tile instead of a zeroed chart. I think a wrong number is worse than an absent one, but say the word if you would rather keep the chart.
- **i18n:** one new key across the 17 `translation.json` files. The chart titles and "Unavailable" reuse existing keys. I reverted the unrelated `glossary.json` churn that `npm run i18n` produces so the diff stays reviewable — worth a separate cleanup PR.
- `ResourceTable` gains one optional prop (`namespacedResource`), defaulted off, so plugin tables are unaffected.
- Verified with `eslint`, `prettier` and `tsc --noEmit`; snapshot suites under `components/cluster` pass unchanged.

## Summary by CodeRabbit

- **New Features**
  - Namespace selections now apply consistently across cluster and workload overviews.
  - Added guidance and a direct link to cluster settings when namespace access is restricted.
  - Overview charts show when data is unavailable instead of displaying misleading zero values.
  - Partial chart data remains visible when only some requests fail.
  - Pod and node information refreshes periodically.

- **Bug Fixes**
  - Improved error reporting by separating, grouping, and deduplicating resource access errors.
  - Metrics charts remain visible when related permissions are unavailable.

### DIFF
--- a/frontend/src/components/cluster/ClusterGroupErrorMessage.test.tsx
+++ b/frontend/src/components/cluster/ClusterGroupErrorMessage.test.tsx
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render, screen } from '@testing-library/react';
+import type React from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { storeClusterSettings } from '../../helpers/clusterSettings';
+import { ApiError } from '../../lib/k8s/api/v2/ApiError';
+import { ClusterGroupErrorMessage } from './ClusterGroupErrorMessage';
+
+const CLUSTER = 'test-cluster';
+
+vi.mock('react-i18next', async importOriginal => ({
+  ...(await importOriginal<typeof import('react-i18next')>()),
+  useTranslation: () => ({ t: (key: string) => key.split('|').pop() }),
+}));
+
+vi.mock('../../lib/k8s', () => ({
+  useCluster: () => CLUSTER,
+  useSelectedClusters: () => [CLUSTER],
+}));
+
+vi.mock('../../redux/hooks', () => ({
+  useTypedSelector: (selector: any) => selector({ filter: { namespaces: new Set<string>() } }),
+}));
+
+vi.mock('../common/Link', () => ({
+  default: ({ children }: { children: React.ReactNode }) => <span>{children}</span>,
+}));
+
+const HINT =
+  'Headlamp is requesting resources across the whole cluster. If you only have access to ' +
+  'some namespaces, list them under Allowed namespaces in the cluster settings.';
+
+function renderForbidden() {
+  return render(
+    <ClusterGroupErrorMessage
+      errors={[new ApiError('forbidden', { status: 403, cluster: CLUSTER })]}
+      namespacedResource
+    />
+  );
+}
+
+describe('ClusterGroupErrorMessage', () => {
+  beforeEach(() => localStorage.clear());
+  afterEach(() => localStorage.clear());
+
+  it('offers the namespace scoping hint when nothing narrows the requests', () => {
+    renderForbidden();
+
+    expect(screen.getByText(HINT, { exact: false })).toBeInTheDocument();
+  });
+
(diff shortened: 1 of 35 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

</details>

illume (changes requested):
Thanks for these changes.

PR #6988, opened 2026-08-07
frontend: resourceMap: Fix Map view hang on stuck sources

### LINKED ISSUE #6555: Map view intermittently never finishes loading; one default source can hold the whole view in the spinner state

rootp1:
## Summary

This PR fixes a bug where the resource Map view could hang on the
loading spinner forever by making `GraphSourceManager` stop waiting
indefinitely on a source whose data never resolves.

## Related Issue

Fixes #6555

## Changes

- Root cause: `GraphSourceManager`'s aggregate `isLoading` stayed
  `true` forever if any selected source's `useData()` hook never
  resolved (e.g. a watch/list that hangs for a particular resource
  kind), blocking the whole Map view from rendering even though every
  other source had already loaded.
- `SourceLoader` now gives each source `SOURCE_LOADING_TIMEOUT_MS`
  (15s) to resolve. If it's still `null` by then, it's treated as
  loaded-but-empty so the rest of the graph can render. If real data
  arrives later, it still replaces the placeholder normally through
  the existing update path.
- This is complementary to #6898 (source-selection persistence), which
  only works around the symptom by letting users permanently deselect
  the problematic source group. This PR fixes the underlying hang so
  the Map view finishes loading on its own.
- Added `GraphSources.test.tsx` covering `GraphSourceManager` (no
  tests existed for this file before), including a new test asserting
  a hung source no longer blocks `isLoading` once the timeout elapses.
- Added a Storybook story (`HungSourceRepro`) reproducing the bug (one
  source resolves, one never does), used to capture the screenshot
  evidence below.

## Steps to Test

1. Run `npm run storybook` in `frontend/` and open the `GraphView ->
   HungSourceRepro` story.
2. Observe the Map view resolves and renders the settled source's
   nodes after ~15s, instead of spinning forever.
3. Run `cd frontend && npx vitest run
   src/components/resourceMap/sources/GraphSources.test.tsx` to verify
   the new/updated tests pass.

Screen recording of the same story end to end (spinner, then resolves
at the timeout instead of hanging forever):
https://github.com/rootp1/headlamp/blob/map-hang-demo-media/docs/media/map-hang-demo.webm

Captured via Playwright against the `HungSourceRepro` story, at the
same elapsed time in both runs:

**Before the fix** — still showing only the loading spinner at 17s,
even though the "Pods (settles)" source resolved instantly; it never
recovers:

before fix - still loading forever

**After the fix** — resolves at the 15s timeout and renders the
settled source's nodes; the hung source simply contributes no nodes
instead of blocking everything:

after fix - resolves at timeout

## Notes for the Reviewer

- `SOURCE_LOADING_TIMEOUT_MS` is currently a fixed 15s constant; happy
  to make it configurable or tune the value if reviewers have a
  different preference.
- `tsc --noEmit` and `eslint` are clean on the changed files.

### DIFF
--- a/frontend/src/components/resourceMap/GraphView.stories.tsx
+++ b/frontend/src/components/resourceMap/GraphView.stories.tsx
@@ -217,6 +217,33 @@ export const BasicExample = () => (
 );
 BasicExample.args = {};
 
+// Repro for https://github.com/kubernetes-sigs/headlamp/issues/6555:
+// one source resolves normally, the other never settles (simulates a
+// stuck watch/list). Used to capture before/after screenshot evidence
+// of the SOURCE_LOADING_TIMEOUT_MS fix in GraphSources.tsx.
+const settledSource: GraphSource = {
+  id: 'settled-source',
+  label: 'Pods (settles)',
+  useData() {
+    return data;
+  },
+};
+
+const hungSource: GraphSource = {
+  id: 'hung-source',
+  label: 'Workloads (never settles)',
+  useData() {
+    return null;
+  },
+};
+
+export const HungSourceRepro = () => (
+  <TestContext>
+    <GraphView height="600px" defaultSources={[settledSource, hungSource]} />
+  </TestContext>
+);
+HungSourceRepro.args = {};
+
 /**
  * Percentage of pods that should have error status (for testing error filtering)
  */
--- a/frontend/src/components/resourceMap/__snapshots__/GraphView.HungSourceRepro.stories.storyshot
+++ b/frontend/src/components/resourceMap/__snapshots__/GraphView.HungSourceRepro.stories.storyshot
@@ -1,0 +1,380 @@
+<body>
+  <div>
+    <div
+      class="MuiBox-root css-1yvazsw"
+    >
+      <div
+        class="MuiBox-root css-2gjmrx"
+      >
+        <div
+          class="MuiBox-root css-pb8oes"
+        >
+          <div
+            class="MuiAutocomplete-root MuiAutocomplete-hasPopupIcon css-1x6bjyf-MuiAutocomplete-root"
+          >
+            <div
+              class="MuiBox-root css-1dipl1t"
+            >
+              <div
+                class="MuiFormControl-root MuiFormControl-fullWidth MuiTextField-root css-wb57ya-MuiFormControl-root-MuiTextField-root"
+                style="margin-top: 0px;"
+              >
+                <label
+                  class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-sizeSmall MuiInputLabel-outlined MuiFormLabel-colorPrimary MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-sizeSmall MuiInputLabel-outlined css-81cxhg-MuiFormLabel-root-MuiInputLabel-root"
+                  data-shrink="true"
+                  for="namespaces-filter"
+                  id="namespaces-filter-label"
+                >
+                  Namespaces
+                </label>
+                <div
+                  class="MuiInputBase-root MuiOutlinedInput-root MuiInputBase-colorPrimary MuiInputBase-fullWidth MuiInputBase-formControl MuiInputBase-sizeSmall MuiInputBase-adornedEnd MuiAutocomplete-inputRoot css-b1xigt-MuiInputBase-root-MuiOutlinedInput-root"
+                >
+                  <input
+                    aria-autocomplete="both"
+                    aria-expanded="false"
+                    aria-invalid="false"
+                    autocapitalize="none"
+                    autocomplete="off"
+                    class="MuiInputBase-input MuiOutlinedInput-input MuiInputBase-inputSizeSmall MuiInputBase-inputAdornedEnd MuiAutocomplete-input MuiAutocomplete-inputFocused css-19qh8xo-MuiInputBase-input-MuiOutlinedInput-input"
+                    id="namespaces-filter"
+                    placeholder="Filter"
+                    role="combobox"
+                    spellcheck="false"
+                    type="text"
+                    value=""
+                  />
+                  <div
+                    class="MuiAutocomplete-endAdornment css-p1olib-MuiAutocomplete-endAdornment"
+                  >
+                    <button
+                      aria-label="Open"
+                      class="MuiButtonBase-root MuiIconButton-root MuiIconButton-sizeMedium MuiAutocomplete-popupIndicator css-qzbt6i-MuiButtonBase-root-MuiIconButton-root-MuiAutocomplete-popupIndicator"
+                      tabindex="-1"
+                      title="Open"
+                      type="button"
+                    >
+                      <svg
+                        aria-hidden="true"
+                        class="MuiSvgIcon-root MuiSvgIcon-fontSizeMedium css-i4bv87-MuiSvgIcon-root"
+                        data-testid="ArrowDropDownIcon"
+                        focusable="false"
+                        viewBox="0 0 24 24"
+                      >
+                        <path
+                          d="M7 10l5 5 5-5z"
+                        />
+                      </svg>
+                      <span
+                        class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"
+                      />
+                    </button>
+                  </div>
+                  <fieldset
+                    aria-hidden="true"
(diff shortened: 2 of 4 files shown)


vyncent-t (changes requested):
The copilot comment brings a good point, try to see if we can avoid it getting stuck.

rootp1:
> The copilot comment brings a good point, try to see if we can avoid it getting stuck.

Hey @vyncent-t, the point that the timeout only exists while data === null, so it restarts on later loading cycles and is canceled as soon as the source resolves was implemented, can your please have a look whenever you have a chance to?

rootp1:
## Screenshots — live cluster repro

Captured against a real `kind` cluster + live Headlamp backend/frontend (not Storybook). To reproduce the hang for real, I added a CRD (`stuckresources.stuckwatch.example.com`) and put a TLS-terminating proxy in front of the cluster's API server that never responds to that CRD's list/watch call, then selected "Workloads" + "Custom Resources" as Map sources — one source resolves normally, the other never settles, same as the real-world scenario in #6555.

**Before** (current `main`, no fix): Map view spins forever — still stuck after 18s+ because the aggregate `isLoading` never clears while the CRD source is unresolved.

Before fix: Map view stuck on an infinite loading spinner with no nodes rendered, because one source (a CRD watch) never resolves.

**After** (this PR): once `SOURCE_LOADING_TIMEOUT_MS` (15s) elapses, the stuck CRD source is treated as loaded-but-empty and the rest of the graph — the real Workloads data (Deployments, DaemonSets, Pods) — renders normally.

After fix: Map view renders Workloads nodes (coredns, kindnet, kube-proxy, etc.) normally, 15+ seconds after selecting a source that never resolves, instead of staying stuck.

---

Also looked into the failing `test` CI job: it's a Storybook snapshot mismatch in `pvc/CreatePVCForm` (React `useId`-generated radio input names, e.g. `:r3l3:` vs `:r3la:`). This PR only touches `resourceMap/GraphSources.tsx`, `GraphView.stories.tsx`, and `GraphSources.test.tsx` — nothing related to PVC forms. Reproduced locally on this branch and the mismatch persists with a different id shift depending on which/how many other tests run first, confirming it's pre-existing base-branch flakiness (an order-dependent `useId` counter) rather than a regression introduced here.

rootp1:
@illume this is a fully tested pr, with no new comments

illume (approved):
🎉 thanks!

PR #6989, opened 2026-08-07
frontend: Add Headlamp events for project, settings and plugin views

ashu8912:
Plugins could not observe several of Headlamp's views because no `HeadlampEvent` was dispatched for them.

This adds events for:

- the project list and details views
- project creation and deletion
- the general and cluster settings views
- the plugin list and details views
- switching tabs in the project details view

Each new event type gets a payload interface, a `useEventCallback` overload and a dispatch case, and is re-exported from the plugin registry so plugins can type against it.


skoeva (commented):
linter is complaining about useEffect dependencies

skoeva (approved):
looks good, thanks!

PR #6990, opened 2026-08-07
frontend: CreateServiceForm: Add create service form

vyncent-t:
## Summary

This PR adds a Form tab for creating Services, bringing feature parity with the existing create forms for Deployments, DaemonSets, ReplicaSets, Pods, Jobs, CronJobs, and StatefulSets. It also promotes the Service ports editor to a reusable `type: 'ports'` field on `CreateResourceForm`, alongside the existing reusable field types (`labels`, `containers`, `namespace`, `select`).

The form defaults match the base YAML from `Service.getBaseObject()`:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: ''
spec:
  clusterIP: ''
  ports:
    - name: ''
      nodePort: 30000
      port: 80
      protocol: TCP
      targetPort: 80
  type: ClusterIP
  externalIPs: []
  selector: {}
```

## Changes

- Added `CreateServiceForm` component with metadata and spec (type, clusterIP, ports, externalIPs, selector) sections
- Added `type: 'ports'` as a first-class reusable field type on `CreateResourceForm`, backed by a new exported `ServicePortsTextField` component (per-row `name` / `port` / `targetPort` / `nodePort` / `protocol` with add/remove controls)
- Added a `ports` case to the form's validity check: requires a non-empty array where each entry has an integer `port` between 1 and 65535
- Registered `Service` in `RESOURCE_DEFINITIONS` in `CreateButton.tsx`
- Exported the form from the common Resource index for plugin access
- Added a Storybook story (Default) matching the base YAML
- Added tests covering the new ports editor: partial-YAML tolerance (`null`, `[null]`, string), default-row rendering, numeric coercion when editing the Port input, named-string `targetPort` preservation, and the Add Port button emit

## Steps to Test

1. Navigate to the Services list view
2. Click the "Create" button
3. Observe the Form tab appears alongside the YAML editor
4. Verify fields: Name, Namespace, Labels, Type (ClusterIP/NodePort/LoadBalancer/ExternalName), Cluster IP, Ports (name/port/targetPort/nodePort/protocol per row), External IPs, Selector labels
5. Edit the port row: change `Port`, `Target Port` (try both `8080` and a named port like `http`), `Node Port`, and `Protocol`
6. Click "New Port" to append a row, then remove it via the trash icon
7. Fill in values and confirm the YAML editor stays in sync bidirectionally
8. Apply and verify the Service is created on the cluster

## Screenshots
image

## Notes for the Reviewer

- Follows the same pattern as the other resource forms — uses `metadataSection` for the standard name/namespace/labels fields and declares `spec.ports` and `spec.selector` via the reusable field types on `CreateResourceForm`.
- The new field type is intentionally named `ports` (Service ports). It is not a generic "any ports" editor — `containerPort` editing for Pods/workloads still lives inside `ContainerTextField`.
- `ServicePortsTextField` preserves the YAML-editor tolerance pattern used by `ContainerTextField` (`null`, `[null]`, and non-array values must not crash while the user types in the editor tab).
- `targetPort` remains typed as `number | string` to match the Kubernetes schema; the editor uses a text input so named target ports are not silently coerced to numbers.
- `RESOURCE_DEFINITIONS` and the `common/Resource` barrel are re-exported into the plugin lib; the `pluginLib.snapshot` update reflects the new `Service` entry, `CreateServiceForm`, and `ServicePortsTextField`.


frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.tsx
```diff
@@ -508,6 +528,118 @@ export function NamespaceTextField(props: NamespaceTextFieldProps) {
...
+  onChange: (ports: ServicePortDraft[]) => void;
+}
+
+/** Editor for a Service's `spec.ports` array. Renders one row per port with
+ *  name / port / targetPort / nodePort / protocol inputs plus add/remove
+ *  controls. `targetPort` accepts either an integer or a string (named
+ *  target port) to match the Kubernetes schema. */
+export function ServicePortsTextField(props: ServicePortsTextFieldProps) {
+  const { value, onChange } = props;
+  const { t } = useTranslation(['translation']);
```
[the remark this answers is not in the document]
vyncent-t: fixed

frontend/src/components/service/CreateServiceForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+  resource?: ServiceDraft;
+  onChange: (resource: ServiceDraft) => void;
+  onValidChange?: (valid: boolean) => void;
+}
+
+const EMPTY_SERVICE_DRAFT: ServiceDraft = {};
+
+export default function CreateServiceForm(props: CreateServiceFormProps) {
+  const { resource = EMPTY_SERVICE_DRAFT, onChange, onValidChange } = props;
+  const { t } = useTranslation(['translation']);
```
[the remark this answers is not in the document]
vyncent-t: fixed

frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.tsx
```diff
@@ -508,6 +528,118 @@ export function NamespaceTextField(props: NamespaceTextFieldProps) {
...
+  const ports: ServicePortDraft[] = Array.isArray(value) ? value : [];
+
+  function updatePort(index: number, field: keyof ServicePortDraft, rawValue: string) {
+    const nextPorts = ports.map(port => ({ ...(port ?? {}) }));
```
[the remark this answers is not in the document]
vyncent-t: fixed

frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.test.tsx
```diff
@@ -63,155 +62,69 @@ describe('ContainerTextField partial-input tolerance', () => {
...
-        options: [
-          { value: '0', label: 'Port 0' },
-          { value: '80', label: 'Port 80' },
-        ],
-      },
-    ],
-  },
-];
-
-describe('CreateResourceForm – select field', () => {
```
ashu8912, line 104: This removes the entire `CreateResourceForm – select field` suite (8 tests, ~110 lines) along with `makeSelectSections` / `makeNumericSelectSections`. I grepped the branch and they aren't moved anywhere else, so this is a straight coverage loss on behaviour unrelated to Services:

- single-select renders with a string value
- single-select emits a string (not an array)
- multi-select renders with an array value
- multi-select normalizes a bare string
- multi-select emits an array
- native-autofill string is wrapped into a one-element array
- `''` is treated as absent, `'0'` is *not* (the falsy-numeric-string case)
- chips render option labels, not raw stored values

The `'0'` and chip-label cases in particular guard real past bugs. Could you keep this `describe` block and add the new `ServicePortsTextField` suite alongside it rather than replacing it? That also means keeping the `screen` and `CreateResourceForm`/`FormSection` imports that were dropped at the top of the file.

frontend/src/i18n/locales/pt/translation.json
```diff
@@ -0,0 +1,1049 @@
+{
```
ashu8912, line 1: This PR adds an entirely new `pt` locale — `pt/translation.json` (1049 lines), `pt/glossary.json` (339) and `pt/app.json` (57), 1445 lines total. `pt` does not exist on `main`; the repo ships `pt-br` and `pt-pt`.

This looks like an accidental artifact of the i18n tooling rather than something intended by a "create Service form" change. Please drop these three files. Adding a new locale is its own decision and should be a separate PR.

While you're regenerating: the same run also reordered keys in `glossary.json` for cs/hu/id/nl/pl/pt-br/sv/tr (moving `Service Account`), which is unrelated churn. A clean rebase + regenerate should leave only the new `Port` / `Target Port` / `Node Port` / `Protocol` / `Remove port` / `Add port` / `New Port` / `Cluster IP` / `Ports` / `External IPs` / `Leave empty to have Kubernetes assign an IP.` keys.

frontend/src/components/service/CreateServiceForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+              onChange={event =>
+                onExternalIPsChange(
+                  event.target.value
+                    .split(',')
+                    .map(ip => ip.trim())
+                    .filter(Boolean)
+                )
+              }
+              inputProps={{ 'aria-label': t('translation|External IPs') }}
+            />
```
ashu8912, line 87: This field can never hold more than one IP.

The input is fully controlled off the array (`value.join(', ')`), and `handleFieldChange` in `CreateResourceForm` writes the parsed array straight back into the resource, so every keystroke round-trips through `join`/`split`/`filter(Boolean)`.

Trace typing `192.0.2.1,` to start a second entry:

1. `'192.0.2.1,'.split(',')` → `['192.0.2.1', '']`
2. `.map(trim).filter(Boolean)` → `['192.0.2.1']`
3. re-render → `value` is `'192.0.2.1'`

The comma is swallowed the instant it's typed, so the separator can never be entered and `externalIPs` is capped at one address. Same for the trailing space.

Fix would be to hold the raw text in local state and only parse into the array on change/blur, e.g. keep a `React.useState` seeded from `value.join(', ')`, sync it when the incoming array changes identity, and call `onExternalIPsChange` with the parsed list while rendering the raw buffer.

This also isn't covered by any test — there's no test file for `CreateServiceForm` itself, only for `ServicePortsTextField`. A small test that types `192.0.2.1, 192.0.2.2` and asserts both land in `spec.externalIPs` would have caught it and would guard the fix.

ashu8912 (commented):
Review focused on blocking items only. Nice work on the form itself — the `type: 'ports'` field type and the YAML-tolerance tests follow the existing `containers` pattern well. Three things need to be resolved before this can go in; I've left them as inline comments.

1. The `CreateResourceForm – select field` test suite is deleted (unrelated coverage regression).
2. A brand-new `pt` locale is added (out of scope).
3. `spec.externalIPs` can only ever hold a single IP due to a controlled-input round-trip.

Also worth rebasing onto current `main` before the next push — the base here is a fair bit behind, and the a11y baseline / storyshot files move often.

frontend/src/components/service/CreateServiceForm.tsx
vyncent-t: resolved

frontend/src/i18n/locales/pt/translation.json
vyncent-t: resolved

frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.test.tsx
vyncent-t: resolved

frontend/src/components/service/CreateServiceForm.tsx
```diff
@@ -0,0 +1,146 @@
...
+    metadataSection(t),
+    {
+      title: t('translation|Spec'),
+      fields: [
+        {
+          key: 'type',
+          path: 'spec.type',
+          label: t('translation|Type'),
+          type: 'select',
+          options: ['ClusterIP', 'NodePort', 'LoadBalancer', 'ExternalName'].map(value => ({
```
skoeva: `ExternalName` is selectable, but the form never sets the required `spec.externalName`- I think it should be added conditionally or removed here

frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.tsx
```diff
@@ -571,6 +591,120 @@ export function NamespaceTextField(props: NamespaceTextFieldProps) {
...
+      const numberValue = Number(rawValue);
+      if (!Number.isInteger(numberValue) || numberValue < 0) return;
+      nextPort[field] = numberValue as never;
+    }
+
+    onChange(nextPorts);
+  }
+
+  function addPort() {
+    onChange([...ports, { name: '', nodePort: 30000, port: 80, protocol: 'TCP', targetPort: 80 }]);
```
skoeva: This adds `nodePort: 30000` regardless of Service type - k8s rejects `nodePort` on `ClusterIP` services, so we can leave it out by default and only expose it for `NodePort` or `LoadBalancer`

skoeva (commented):
left a couple of comments

frontend/src/components/service/CreateServiceForm.tsx
vyncent-t: fixed

frontend/src/components/common/Resource/CreateResourceForm/CreateResourceForm.tsx
vyncent-t: fixed

skoeva (approved):
lgtm, thanks for this!

PR #6995, opened 2026-08-07
plugins/headlamp-plugin: Fix storybook dev command on Windows

### LINKED ISSUE #6994: [Bug]: Headlamp-plugin storybook fails on Windows ('.' is not recognized as an internal or external command)
## Describe the bug

Running `npm run storybook` inside a plugin package (which maps to `headlamp-plugin storybook`) fails immediately on native Windows (cmd.exe/PowerShell, not WSL) with:

```
'.' is not recognized as an internal or external command,
Problem running storybook dev inside of "." abs: "<packageFolder>"
```

Storybook never starts.

**Root cause:** `storybook()` in `plugins/headlamp-plugin/bin/headlamp-plugin.js` calls:

```js
child_process.execSync(
  './node_modules/.bin/storybook dev -p 6007 -c node_modules/@kinvolk/headlamp-plugin/config/.storybook',
  { stdio: 'inherit', cwd: packageFolder, encoding: 'utf8' }
);
```

`execSync` runs this through a shell — `cmd.exe` on Windows — which doesn't understand the Unix-style `./` relative path prefix, so it fails to resolve the command at all.

Every other subcommand (`lint`, `tsc`, `test`, `i18n`, and even the sibling `storybook_build`) instead goes through the shared `runScriptOnPackages` helper, which resolves the binary cross-platform via `path.join('node_modules', '.bin', scriptCmd + binExt)` (with `binExt` = `.cmd` on Windows) and runs it with `execFileSync` — no shell involved, so it isn't affected by this. `storybook` (dev mode) is the only command that bypasses this helper with its own hand-rolled, Unix-only `execSync` call.

## To Reproduce

Steps to reproduce the bug:
1. On Windows, clone `headlamp-k8s/plugins` and `cd` into any plugin folder, e.g. `backstage/`.
2. Run `npm install`.
3. Run `npm run storybook`.
4. See the error above; Storybook never launches.

## Environment:

- Installation type: Windows (native, PowerShell) — not WSL
- Headlamp Version: N/A (this is the `@kinvolk/headlamp-plugin` CLI tool, not the Headlamp app itself)
- Other:
  - `@kinvolk/headlamp-plugin` version: 0.14.0
  - Node version: v22.19.0
  - Reproduces in any plugin under `headlamp-k8s/plugins` that has a `storybook` script (e.g. `backstage`), since they all shell out to the same CLI command.

## Additional Context

Not a regression from a specific version as far as I can tell — the Unix-only path has likely always been there since `storybook` dev mode was added; it's just never worked on native Windows.

KariraLakshya:
## Summary

This PR fixes `headlamp-plugin storybook` failing on native Windows by replacing its hand-rolled `execSync` call with the same cross-platform helper every other CLI subcommand already uses.

## Related Issue

Fixes #6994 

## Changes

- `plugins/headlamp-plugin/bin/headlamp-plugin.js`: `storybook()` previously called `child_process.execSync('./node_modules/.bin/storybook dev ...')` directly. On Windows, `execSync` spawns `cmd.exe`, which can't resolve the Unix-style `./` relative path, so the command failed immediately with `'.' is not recognized as an internal or external command`.
- Replaced it with a call to the existing `runScriptOnPackages()` helper — the same one `lint`, `tsc`, `test`, `i18n`, and the sibling `storybook_build` already use. It resolves the binary cross-platform (`.cmd` extension on Windows via `viaCmd`) and runs it with `execFileSync`, without going through a shell at all.
- Net change: removed the custom `try`/`catch`/`execSync` block, replaced with the same two-line pattern `storybook_build()` uses.

## Steps to Test

1. On Windows (native cmd.exe/PowerShell, not WSL), clone `headlamp-k8s/plugins` (or any repo with a Headlamp plugin) and `cd` into a plugin folder, e.g. `backstage/`.
2. Run `npm install`.
3. Before the fix: `npm run storybook` fails instantly with `'.' is not recognized as an internal or external command, operable program or batch file.`
4. After the fix: `npm run storybook` correctly resolves `node_modules\.bin\storybook.cmd`, Storybook starts, and webpack begins compiling — verified locally end-to-end.

## Notes for the Reviewer

- No behavior change for macOS/Linux — `runScriptOnPackages` already handled those platforms correctly; this only fixes the Windows path.
- I did not change `storybook_build()`; it already followed this pattern correctly, which is how I found the inconsistency.

### DIFF
--- a/plugins/headlamp-plugin/bin/headlamp-plugin.js
+++ b/plugins/headlamp-plugin/bin/headlamp-plugin.js
@@ -1452,23 +1452,8 @@ function tsc(packageFolder) {
  * @returns {0 | 1} Exit code, where 0 is success, 1 is failure.
  */
 function storybook(packageFolder) {
-  try {
-    child_process.execSync(
-      './node_modules/.bin/storybook dev -p 6007 -c node_modules/@kinvolk/headlamp-plugin/config/.storybook',
-      {
-        stdio: 'inherit',
-        cwd: packageFolder,
-        encoding: 'utf8',
-      }
-    );
-  } catch (e) {
-    console.error(
-      `Problem running storybook dev inside of "${packageFolder}" abs: "${resolve(packageFolder)}"`
-    );
-    return 1;
-  }
-
-  return 0;
+  const script = `storybook dev -p 6007 -c node_modules/@kinvolk/headlamp-plugin/config/.storybook`;
+  return runScriptOnPackages(packageFolder, 'storybook dev', script, {});
 }
 
 /**


illume (changes requested):
Thanks for these changes.


<details>

</details>

KariraLakshya:
@illume branch is rebased and ready for review

illume (approved):
🎉 thanks!

KariraLakshya:
thanks @illume for the approval! Let me know if anything else is needed to merge

PR #6996, opened 2026-08-07
backend: oidc: Characterize the OIDC callback success path

gambtho:
## Summary

Follow-up to #5420. That PR characterized the `/oidc` and `/oidc-callback` handlers up to the point of token exchange; the mock provider's `/token` endpoint always failed, so the entire callback **success** path was unexercised. This PR teaches the mock to be a real IdP — signing tokens with a genuine RSA key served through a real JWKS endpoint — and characterizes the rest of the flow.

Test-only. No production code, `go.mod`, or `go.sum` changes.

## Related Issue

Part of #5401.

## Coverage

`/oidc-callback` closure (`backend/cmd/headlamp.go:1010-1124`): **22/54 → 48/54 statements (88.9%)**.

`go tool cover -func` is misleading here — it folds handler closures into the enclosing `createHeadlampHandler`, so there is no per-handler number to read. These figures come from slicing the raw `-coverprofile` by the closure's line range, the same method used to produce the numbers in #5420:

```bash
go test -count=1 -run '^TestOIDC' -coverprofile=/tmp/oidc.out ./cmd
python3 -c "
import re
tot=cov=0
for line in open('/tmp/oidc.out'):
    m=re.match(r'.*headlamp\.go:(\d+)\.\d+,(\d+)\.\d+ (\d+) (\d+)',line)
    if not m: continue
    s,e,n,c = (int(x) for x in m.groups())
    if 1010<=s<=1124:
        tot+=n
        if c>0: cov+=n
print(f'{cov}/{tot} statements ({100*cov/tot:.1f}%)')
"
```

## Changes

Five commits, all in `backend/cmd/oidc_handlers_test.go`, covering:

- **Signing mock provider.** `newOIDCTestServer` now serves a real key set at `/jwks` (previously `{"keys":[]}`) and can mint signed tokens via `go-oidc`'s `oidc/oidctest`, already a direct dependency. One package-level RSA key behind `sync.Once`, shared across servers; a second key only for the verification-failure test.
- **`TestOIDCCallback_Success`** — 303 to the frontend, chunked auth cookie carries the id_token, claims decoded.
- **`TestOIDCCallback_CachesRefreshToken`** — refresh token stored under `oidc-token-<raw token>`.
- **`TestOIDCCallback_TokenMissingFromResponse`** — a 200 from `/token` with no `id_token` hits the error path.
- **`TestOIDCCallback_VerificationFailure`** — a token signed by a different key is rejected.
- **`TestOIDCCallback_UsesAccessTokenWhenConfigured`** — `OidcUseAccessToken` swaps which token reaches the cookie.
- **`TestOIDCCallback_RedirectTarget`** — table over DevMode and BaseURL.

Two fixes folded in:

- `driveOIDCStart` built a relative request, so `r.Host` was empty and `getOidcCallbackURL` emitted `redirect_uri=http:///oidc-callback`. Harmless while every callback test failed early; not once a full round trip runs. Copilot flagged this on #5420. Now uses an absolute URL.
- `authCookieValue` reassembles the chunked cookie by index rather than reading `.0`, and sanitizes the cluster name to match how `SetTokenCookie` names cookies.

## Steps to Test

```bash
cd backend
go test -count=1 -run 'TestOIDC' ./cmd/... -v   # 17 pass
go test -count=1 ./...
```

## Notes for the Reviewer

**Non-vacuity was checked for every new test**, since that was the substance of Copilot's review on #5420. Each test's target branch was broken and the test observed to fail, then restored:

| Test | Broke | Observed failure |
|---|---|---|
| `_Success` | commented out `SetTokenCookie` (`headlamp.go:1119`) | `no headlamp-auth-oidc-char-test.N cookies were set` |
| `_CachesRefreshToken` | disabled the `Cache.Set` (`:1078`) | `key not found` |
| `_TokenMissingFromResponse` | added an `id_token` back to the mock response | 303 instead of 500 |
| `_VerificationFailure` | signed with the correct key | 303 instead of 500 |
| `_UsesAccessTokenWhenConfigured` | neutered the `withAccessToken` option | id_token in the cookie, not the access_token |
| `_RedirectTarget` | corrupted one `want` | only that subtest failed |

**Three things this characterizes without endorsing:**

1. **Access-token mode cannot work with a real provider** (filed as #6998). `OidcUseAccessToken` changes only which field is *read* from the token response; the value still goes through the ID-token verifier. An opaque access token — what most IdPs actually issue — would be rejected. The test has to mint a second signed JWT for the `access_token` field to pass at all. Worth a look during #5401.

2. **A rejected token still leaves a cached refresh token** (filed as #6997). `Cache.Set` runs *before* `Verifier.Verify` (`headlamp.go:1078` vs `:1086`), so a token that fails verification leaves an `oidc-token-<raw>` entry behind. `TestOIDCCallback_VerificationFailure` pins this deliberately — it is exactly the kind of ordering a consolidation could change silently. Flipping that assertion is a one-line change once the fix lands.

3. **Two cosmetic oddities in the redirect construction**, reported but deliberately not fixed here: `fmt.Sprintf("auth?cluster=%1s", ...)` at `headlamp.go:1121` uses a no-op width-1 verb, almost certainly a typo for `%s`; and the cluster is interpolated into that URL unsanitized while the cookie name for the same cluster goes through `auth.SanitizeClusterName`. Not reachable from user input — the cluster comes from server-side state — but the asymmetry is odd.

**Knowingly left uncovered** (the six remaining statements): `Cache.Set` failure (`:1079-1084`) and claims-decode failure (`:1099-1104`). Both need injected failures, and neither is code #5401 is likely to disturb.

**Also out of scope:** the `/oidc` handler's issuer/client-ID override branches and the `config.Insecure` custom-transport block (`:886-907`). The latter needs a TLS mock — separate test infrastructure, possibly a third PR.

**Verification.** `gofmt -l cmd/`, `go vet ./cmd/...`, `golangci-lint run` (v2.12.2, the pinned version), and the full backend suite are all clean — and, via `git rebase <base> --exec`, clean at *every* commit, not just at the tip.

One note for contributors, unrelated to the change itself: `npm run backend:install:linter` installs golangci-lint with `go install`, which builds it from source. That fails on any toolchain older than the one `backend/go.mod` targets — currently Go 1.26 — with `can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.5)`, so `npm run backend:lint` is unrunnable locally on such a machine. CI is unaffected: `golangci-lint-action` uses `install-mode: binary` and downloads the prebuilt release, which is built with a matching Go. Downloading the same release binary makes local lint match CI exactly. Possibly worth changing the npm script to do the same, but that is out of scope here.

### DIFF
--- a/backend/cmd/oidc_handlers_test.go
+++ b/backend/cmd/oidc_handlers_test.go
@@ -708,3 +708,67 @@ func TestOIDCCallback_CachesRefreshToken(t *testing.T) {
 	require.NoError(t, err, "refresh token should be cached under oidc-token-<raw token>")
 	assert.Equal(t, "refresh-xyz", got)
 }
+
+// TestOIDCCallback_TokenMissingFromResponse characterizes a token endpoint
+// that succeeds but omits id_token entirely.
+func TestOIDCCallback_TokenMissingFromResponse(t *testing.T) {
+	srv := newOIDCTestServer(t, nil)
+	handler, cluster := newOIDCTestHandler(t, srv)
+
+	srv.setTokenHandler(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+
+		_, _ = io.WriteString(w, `{"access_token":"opaque-access","token_type":"Bearer"}`)
+	})
+
+	state := extractState(t, driveOIDCStart(t, handler, cluster))
+	rr := driveOIDCCallback(t, handler, state)
+
+	assert.Equal(t, http.StatusInternalServerError, rr.Code)
+	assert.Contains(t, rr.Body.String(), "No id_token field in oauth2 token")
+}
+
+// TestOIDCCallback_VerificationFailure characterizes a token that is well
+// formed but signed by a key the provider does not advertise.
+//
+// It also pins a partial write: the refresh token is cached BEFORE the
+// signature is verified (headlamp.go:1078-1092), so a token that is then
+// rejected still leaves an oidc-token-<raw> entry behind. This is recorded as
+// current behavior, not endorsed — caching credentials for a token that was
+// subsequently rejected is arguably wrong, and #5401 may want to reorder it.
+// If that happens, this assertion should be UPDATED, not deleted.
+func TestOIDCCallback_VerificationFailure(t *testing.T) {
+	srv := newOIDCTestServer(t, nil)
+	c := cache.New[interface{}]()
+	handler, cluster := newOIDCTestHandler(t, srv, withCache(c))
+
+	otherKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	require.NoError(t, err)
+
+	claims := fmt.Sprintf(
+		`{"iss":%q,"aud":"test-client-id","sub":"test-subject","exp":%s,"iat":%s}`,
+		srv.URL(),
+		strconv.FormatInt(time.Now().Add(time.Hour).Unix(), 10),
+		strconv.FormatInt(time.Now().Unix(), 10),
+	)
+	badToken := oidctest.SignIDToken(otherKey, testKeyID, oidc.RS256, claims)
+
+	srv.setTokenHandler(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+
+		_, _ = io.WriteString(w, `{"access_token":"opaque-access","token_type":"Bearer",`+
+			`"refresh_token":"refresh-orphan","id_token":"`+badToken+`"}`)
+	})
+
+	state := extractState(t, driveOIDCStart(t, handler, cluster))
+	rr := driveOIDCCallback(t, handler, state)
+
+	require.Equal(t, http.StatusInternalServerError, rr.Code)
+	assert.Contains(t, rr.Body.String(), "Failed to verify ID Token")
+
+	got, err := c.Get(context.Background(), "oidc-token-"+badToken)
+	require.NoError(t, err,
+		"current behavior: the refresh token is cached before verification, so a "+
+			"rejected token still leaves an entry")
+	assert.Equal(t, "refresh-orphan", got)
+}
--- a/backend/cmd/oidc_handlers_test.go
+++ b/backend/cmd/oidc_handlers_test.go
@@ -218,6 +218,12 @@ func withStateReader(r io.Reader) oidcTestOption {
 	return func(c *HeadlampConfig) { c.oidcStateReader = r }
 }
 
+// withCache injects a caller-held cache so a test can inspect what the
+// callback handler wrote. Without it, newOIDCTestHandler's cache is private.
+func withCache(c cache.Cache[interface{}]) oidcTestOption {
+	return func(hc *HeadlampConfig) { hc.Cache = c }
+}
+
 // newOIDCTestHandler builds a Headlamp handler with one OIDC-configured
 // kubeconfig context whose IdP issuer points at the supplied mock server.
 // Returns the handler and the cluster name registered.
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for this! 🎉 

Some suggestions for your consideration...

The PR title "backend: oidc: Characterize the OIDC callback success path" confused me a bit about what this PR was doing. Is the main goal here to finish off test coverage so that future changes are less risky? Something like this?

> backend: oidc_handlers_test: Add test coverage to OIDC callback behavior before refactoring


For the PR description:

The golint stuff can be removed, I don't think it's relevant.

>Verification. gofmt -l cmd/, go vet ./cmd/..., golangci-lint run (v2.12.2, the pinned version), and the full backend suite are all clean — and, via git rebase <base> --exec, clean at every commit, not just at the tip.

>One note for contributors, unrelated to the change itself: npm run backend:install:linter installs golangci-lint with go install, which builds it from source. That fails on any toolchain older than the one backend/go.mod targets — currently Go 1.26 — with can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26.5), so npm run backend:lint is unrunnable locally on such a machine. CI is unaffected: golangci-lint-action uses install-mode: binary and downloads the prebuilt release, which is built with a matching Go. Downloading the same release binary makes local lint match CI exactly. Possibly worth changing the npm script to do the same, but that is out of scope here.

The section "Three things this characterizes without endorsing:" I believe is about future work. This could be trimmed, or removed entirely. It confuses me.

## Recommended commit changes

Keep all five commits. Amend the first, third, and fifth commits.

The goal with commit message changes was to make them easier to read. Because they lead with the protected behavior while retaining the rationale needed to understand future failures.

"Non-vacuity" is a bit too jargon? I'm not sure many people know what that means. Maybe the commit messages could be changed so they describe the change a bit better without many of the details.

---

**note: below was copilot assisted...**

---
<br><br><br>

### KEEP AND AMEND — `5a267d4` — `backend: oidc: Add signing mock provider and callback success test`

**Compared with the original:** The original body contains valuable context
about why signed tokens and absolute request URLs are necessary. Keep that
rationale. Remove review-process details, correct `/jwks` to `/keys`, and fold
in the sanitized-cookie lookup currently added by `81738fe`.

```text
backend: oidc_handlers_test: Test successful OIDC callbacks

Existing callback tests stopped at token exchange because their token
responses failed and the mock provider exposed no signing keys. Use
oidctest.Server to publish keys at /keys and mint verifiable ID tokens.

Build /oidc requests with an absolute URL so getOidcCallbackURL receives
a host instead of producing redirect_uri=http:///oidc-callback.

Verify token exchange, verification, claims decoding, the frontend
redirect, and the chunked authentication cookie.

Part of #5401.
```

### KEEP AND REWORD — `4359554` — `backend: oidc: Characterize refresh-token caching on callback`

**Compared with the original:** The original is accurate and already explains
the cache contract. Retain that rationale, but remove the non-vacuity note and
use a subject that states the tested behavior directly.

```text
backend: oidc_handlers_test: Test OIDC refresh-token caching

Verify that a successful callback stores the refresh token under
oidc-token-<raw ID token>, preserving the token-refresh middleware lookup
contract during #5401.
```

### KEEP AND NARROW — `3ae812c` — `backend: oidc: Characterize callback error paths and partial cache write`

**Compared with the original:** The missing-token and invalid-signature tests
are useful negative coverage. Remove the partial-cache assertion and its
explanation because they make the known #6997 bug part of the regression
contract.

```text
backend: oidc_handlers_test: Test OIDC callback rejection paths

Cover a token response without an ID token and an ID token whose signature
does not match the provider's advertised keys.

Verify that both responses are rejected before the callback completes.

Part of #5401.
```

### KEEP AND REWORD — `f505de2` — `backend: oidc: Characterize OidcUseAccessToken callback mode`

**Compared with the original:** The original correctly explains why the test
needs a signed JWT in the `access_token` field. Preserve that limitation in the
body. The test still usefully verifies token selection and does not require
#6998 to remain unfixed.

```text
backend: oidc_handlers_test: Test OIDC access-token selection

Verify that OidcUseAccessToken selects access_token rather than id_token
for the authentication cookie.

The current callback applies ID-token verification to the selected value,
so the fixture must use a signed JWT. Opaque access-token support remains
tracked by #6998.

Part of #5401.
```

### KEEP AND AMEND — `81738fe` — `backend: oidc: Characterize post-login redirect target`

**Compared with the original:** Keep its explanation that `BaseURL` mounts the
entire router under a prefix. Move cookie sanitization into `5a267d4`, where
the cookie helper is introduced. Do not add special-character expectations
until the redirect construction is fixed.

```text
backend: oidc_handlers_test: Test OIDC callback redirects

Verify successful callback redirects in default, development, and base-URL
configurations.

A base URL mounts the OIDC routes under that prefix, so the test requests
both /oidc and /oidc-callback through the prefixed routes.

Part of #5401.
```

## Add before the OIDC refactor

If this PR is intended to establish sufficient coverage before refactoring the
OIDC handlers, add the following tests before changing their control flow.

### Add additional tests

- Validator issuer override uses the configured validation issuer.
- Validator client-ID override uses the configured validation audience.
- Insecure TLS configuration can discover a provider through a test TLS server.

### Add additional negative tests

- Unknown cluster returns `404`.
- Context without OIDC configuration returns `500`.
- Provider-discovery failure returns `500`.
- Missing `access_token` with `OidcUseAccessToken` returns `500`.
- Cache write failure returns `500` and sets no auth cookie.
- Invalid insecure transport configuration returns `500`.
- Claims-decoding failure returns `500` and sets no auth cookie.

Cache failure requires a failing cache implementation. If claims decoding
cannot fail after successful token verification, document that invariant and
remove the unreachable error branch instead of adding an artificial test.

## Add with production fixes

Do not encode known bugs as expected behavior. Add these tests with their
corresponding production fixes:

- **#6997:** verification failure does not retain a refresh token.
- **#6998:** opaque access tokens work with `OidcUseAccessToken`.
- **Redirect encoding:** cluster names containing `?`, `&`, `=`, `#`, `%`,
  spaces, or Unicode survive the redirect round trip unchanged.
- **Cookie handling:** special-character cluster names produce valid,
  unambiguous cookie names and paths while still resolving to the original
  cluster.

For redirect tests, parse the `Location` header and assert that
`Query().Get("cluster")` equals the original cluster name. This verifies the
behavior rather than a particular escaped representation.

PR #7003, opened 2026-08-08
app: e2e-tests: Give the desktop suite its own Playwright config

unlikelyzero:
## Summary

Gives the desktop (Electron) e2e suite its own Playwright config, removing the `PLAYWRIGHT_TEST_MODE` env var and the `test.skip()` calls that used it to opt specs in/out of app mode at runtime.

## Background

`clusterRename.spec.ts`, `namespaces.spec.ts`, and `clusterAutoConnect.spec.ts` are the only specs in `app/e2e-tests`, and all three only make sense driving the real Electron app directly (kubeconfig, minikube, a single-instance lock) — none of them had a meaningful "web mode" run. Previously the split between app mode and web mode was done at runtime via a `PLAYWRIGHT_TEST_MODE` env var and manual `test.skip()` calls scattered across every spec file, requiring a `cross-env` wrapper to set it.

## Change

Added `playwright.electron.config.ts`, scoped via `testMatch` to just the specs that drive Electron. `npm run test-app` now runs that config directly — no env var, no `cross-env` wrapper. Dropped the `PLAYWRIGHT_TEST_MODE` branching and the browserPage/electronPage fallback in all three specs, and simplified `HeadlampPage.authenticate()` to the single code path its only two callers (both Electron specs) actually use. With no specs left for the old web-mode config to run, removed `playwright.config.ts`, the `test-web` script, the now-unused `cross-env` dependency, and the corresponding README section.

## Verification

Ran `npm run test-app` against a real Electron build and a real minikube cluster: all three specs are collected and run under the new config with no env var required. All four tests fail on two separate, pre-existing bugs unrelated to this change (not introduced by it, and not fixed here — this PR is scoped to the config/env-var cleanup):

- `clusterRename.spec.ts` and `namespaces.spec.ts` hit `AxeBuilder.analyze()`'s `Protocol error (Target.createTarget): Not supported` under Electron — fixed by a separate, still-open PR (`AxeBuilder.setLegacyMode(true)`).
- All three specs are also affected by a pre-existing click-interception flake (Electron's shared default userData directory) — fixed by a separate, still-open PR.

Confirmed this is unrelated to the config move: the same failures reproduce identically against `origin/main` (before this change) using the old `PLAYWRIGHT_TEST_MODE=app` invocation.

## Test plan

- [x] `npm run test-app` collects and runs all three desktop specs under the new config with no env var set
- [x] Confirmed the remaining failures are pre-existing and reproduce the same way on `main`
- [ ] CI


illume (approved):
🎉 thanks!

PR #7008, opened 2026-08-08
frontend: Advanced Search: Match resources by API identity

### LINKED ISSUE #7007: frontend: Advanced Search: Advanced Search may resolve discovered resources using the wrong API resource class
## Describe the bug

Advanced Search resolves discovered Kubernetes resources to a registered `KubeObjectClass` using the resource `kind` alone.

This can cause a discovered resource to incorrectly reuse a built-in `ResourceClass` when another resource with the same `kind` has a different API identity, such as a different `apiVersion`, resource name, or namespace scope.

A discovered resource should only reuse a registered `ResourceClass` when its complete API identity matches. Otherwise, Advanced Search should use a generic `KubeObject` class based on the discovered resource.

## To Reproduce

Steps to reproduce the issue:

1. Use a Kubernetes cluster that exposes an API resource whose `kind` matches a resource already registered in Headlamp.
2. Make sure the discovered resource differs from the registered resource in `apiVersion`, resource name, or namespace scope.
3. Open **Advanced Search** in Headlamp.
4. Select the discovered resource.
5. Observe that Headlamp resolves it using the existing built-in `ResourceClass` based only on the matching `kind`.

This can cause the resource to be queried or handled using an incompatible resource implementation.

## Environment

* Installation type: Windows
* Headlamp Version: Current development version
* Other: Kubernetes cluster exposing an API resource whose `kind` collides with a registered Headlamp resource.

Yes (I will propose a PR).

## Additional Context

Kubernetes API resource identity cannot be determined reliably from `kind` alone. API discovery provides additional information needed to distinguish resources, including:

* `apiVersion`
* resource name (`apiName` / plural name)
* namespace scope (`isNamespaced`)

The proposed fix validates all three fields before reusing a registered `ResourceClass`. If the registered class does not match the discovered API resource, Advanced Search falls back to a generic `KubeObject` configured from the discovered resource.

This prevents Advanced Search from using an incompatible built-in resource implementation for a discovered API resource that happens to share the same `kind`.

mahesh-09-12:
## Summary

This PR fixes Advanced Search resolving discovered Kubernetes resources using only their `kind`.

A registered `ResourceClass` is now reused only when its API identity matches the discovered resource's `apiVersion`, resource name, and namespace scope. Otherwise, Advanced Search falls back to a generic `KubeObject` class for the discovered resource.

## Related Issue

Fixes #7007
## Changes

* Updated `useKubeLists` to validate registered resource classes against the discovered API resource identity.
* Prevented resources with matching `kind` but different API identity from incorrectly using a built-in `ResourceClass`.
* Corrected the `AdvancedSearch/ResourceSearch` Storybook fixture to use the correct `apps/v1` API version for Deployments.

## Steps to Test

1. Run the TypeScript check with `npm run tsc`.
2. Run the Storybook test suite with `npx vitest run src/storybook.test.tsx`.
3. Open Advanced Search and select discovered Kubernetes resources.
4. Verify that resources are queried using their discovered API identity.

Not applicable.

## Notes for the Reviewer

* The resource lookup previously relied on `kind` alone, which can be insufficient for distinguishing Kubernetes API resources.
* The fix preserves existing registered resource classes when their `apiVersion`, resource name, and namespace scope match the discovered resource.
* Existing built-in resources such as Deployments continue to use their specialized `KubeObject` implementations.

### DIFF
--- a/frontend/src/components/advancedSearch/ResourceSearch.stories.tsx
+++ b/frontend/src/components/advancedSearch/ResourceSearch.stories.tsx
@@ -132,7 +132,7 @@ Default.args = {
       isNamespaced: true,
     },
     {
-      apiVersion: 'v1',
+      apiVersion: 'apps/v1',
       version: 'v1',
       singularName: 'deployment',
       kind: 'Deployment',
--- a/frontend/src/components/advancedSearch/utils/useKubeLists.test.ts
+++ b/frontend/src/components/advancedSearch/utils/useKubeLists.test.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { useKubeLists } from './useKubeLists';
+
+const mockRegisteredUseList = vi.hoisted(() =>
+  vi.fn(() => ({
+    items: [],
+    isError: false,
+    errors: undefined,
+  }))
+);
+
+const mockGenericUseList = vi.hoisted(() =>
+  vi.fn(() => ({
+    items: [],
+    isError: false,
+    errors: undefined,
+  }))
+);
+
+vi.mock('../../../lib/k8s', () => {
+  class Deployment {
+    static apiVersion = 'apps/v1';
+    static apiName = 'deployments';
+    static isNamespaced = true;
+    static useList = mockRegisteredUseList;
+  }
+
+  return {
+    ResourceClasses: {
+      Deployment,
+    },
+  };
+});
+
+vi.mock('../../../lib/k8s/cluster', () => {
+  class KubeObject {
+    static useList = mockGenericUseList;
+  }
+
+  return {
+    KubeObject,
+  };
+});
+
+const mockUseNamespaces = vi.hoisted(() => vi.fn(() => []));
+
+vi.mock('../../../redux/filterSlice', () => ({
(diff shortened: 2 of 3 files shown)


illume (changes requested):
Thanks for the contribution.


mahesh-09-12:
@illume could you please take a look at this PR when you get a chance? I’ve addressed the review comments and added regression coverage for the API identity mismatch. Thanks!

mahesh-09-12:
Hey @illume, copilot's latest review found no new review comments. The only remaining issue is the Windows CI -- the arm64 job failed during npm install with an EEXIST npm-cache error, and the x64 Windows job was cancelled. Could you please rerun those checks when you get a chance?

mahesh-09-12:
@illume there are no new comments from Copilot on this PR. Could you please take a look when you get a chance?

unlikelyzero (changes requested):
Some things I observed:

1. The array-valued `apiVersion` branch in `useKubeLists.ts:29` accepts a match by set membership, but reusing that class doesn't actually constrain the request to the discovered version. `getWorkingEndpoint` in `frontend/src/lib/k8s/api/v2/hooks.ts:256-278` races a GET probe against every declared version with `Promise.any` and uses whichever the cluster answers first — not necessarily the discovered one. For a class like `Ingress` (present in both `extensions` and `networking.k8s.io` groups on older clusters) or Gateway API's `Gateway`/`GatewayClass`/`HTTPRoute` (multi-version by design), this means the queried API identity is nondeterministic, and can also produce duplicate rows when the same class is reached via two distinct discovered `ApiResource` entries, since `kubeObjectListQuery`'s cache key (`useKubeObjectList.ts:90-98`) is keyed on `apiVersion`/`apiName` values rather than endpoint identity. This is the same class of bug the PR is fixing for the kind-collision case — the array-version case reopens it.

2. `useKubeLists.test.ts:69` only covers the mismatch/fallback branch. There's no positive-control test proving a matching registered class is still reused, no array-version case, and no assertion that the fallback class's `apiVersion`/`apiName`/`isNamespaced` actually reflect the discovered identity — the test checks which mocked `useList` was called, not what API identity the constructed class carries. This only partially addresses the earlier Copilot comment asking for that endpoint-identity verification.

3. `ApiResourcePicker.stories.tsx:46` still declares Deployment with `apiVersion: 'v1'` — the same wrong-apiVersion issue this PR fixes in the neighboring `ResourceSearch.stories.tsx` fixture.

4. The hoisted mocks in `useKubeLists.test.ts:21` (`mockRegisteredUseList`, `mockGenericUseList`) are never reset between tests. Harmless with one test, but once positive/array-version cases are added, `not.toHaveBeenCalled()`-style assertions become order-dependent.

PR #7009, opened 2026-08-08
frontend: ClusterDialog: Add resizable paper with persisted dimensions

### LINKED ISSUE #5709: frontend: Cluster chooser dialog should remember previous window size
## Current Behavior

The cluster chooser dialog always opens with the default dimensions.

## Expected Behavior

The dialog should restore the user's previous resized dimensions locally.

## Benefit

Improves usability on ultrawide and multi monitor setups.

WasThatRudy:
Fixes #5709

## Summary
- The ClusterChooser dialog always opened at a fixed size, awkward on ultrawide monitors and multi-monitor setups. Make the dialog paper user-resizable via the browser's native corner grip and persist the chosen width/height to `localStorage` across sessions.
- New hook `useResizableDialogSize(storageKey, enabled)` attaches a `ResizeObserver` on the paper element via a callback ref, writes debounced (300ms) size updates to `localStorage[storageKey]`, and reads the stored size on mount.
- Stored sizes are clamped to `95vw × 95vh` before being applied so the drag handle never lands offscreen when the user later opens headlamp on a smaller monitor. CSS `maxWidth: 95vw` / `maxHeight: 95vh` keep the paper inside the viewport at paint time if the browser window is later resized while the dialog is open, so no runtime resize listener is needed.
- The observer skips the very first firing per mount (browser laying out the paper at MUI defaults or at the CSS-clamped stored preference) and any subsequent firing whose size is within 2px of `clampToViewport(applied)`. Without this guard, opening the dialog on a viewport smaller than the stored preference would let the CSS clamp echo into `localStorage` and clobber the user's larger preference. This preserves a big-monitor preference through smaller-monitor sessions, and only persists sizes when the user actually drags the corner handle.
- The hook is disabled when the dialog is in `fullScreen` mode (below the `sm` breakpoint), so mobile viewports get the original fullscreen behavior with no resize affordance.
- All `localStorage` reads and writes are wrapped in `try/catch` so private-browsing modes or blocked storage never crash the hook.

## Non-goals (deferred)
- Applying the hook to other dialogs in the app. The issue is specifically about the ClusterChooser; a follow-up could generalize this once maintainers signal it's desired for e.g. `EditorDialog`, `ConfirmDialog`, etc.
- Keyboard-accessible resize. The native `resize: both` corner grip is mouse-only. Documented in the hook JSDoc as an accessibility limitation with a future-follow-up note.

## Test plan
- [x] `npx vitest run src/components/cluster` (21 tests pass across 4 files; includes 15 new tests for the hook)
- [x] `npx vitest run src/storybook.test.tsx -u` (661 tests pass; 18 storyshot updates are mechanical CSS class-hash changes on cluster/authchooser dialogs)
- [x] `npm run tsc` (clean)
- [ ] Manual: open the cluster chooser dialog on a normal viewport, drag the bottom-right corner to resize, close and re-open the dialog — confirm the new size is restored
- [ ] Manual: resize the browser window smaller than the stored preference, open the dialog, confirm the paper is CSS-clamped to 95% of the viewport without overflowing
- [ ] Manual: on a fullscreen-triggering viewport (below the `sm` breakpoint), confirm no resize handle appears and no localStorage write occurs
- [ ] Manual: with a stored preference on a big monitor (e.g. 3000×2000), reopen headlamp on a smaller laptop (e.g. 1920×1080), confirm the paper renders clamped and the stored preference is NOT overwritten

### DIFF
--- a/frontend/src/components/cluster/Chooser.tsx
+++ b/frontend/src/components/cluster/Chooser.tsx
@@ -31,6 +31,7 @@ import { useTheme } from '@mui/material/styles';
 import TextField from '@mui/material/TextField';
 import Typography from '@mui/material/Typography';
 import useMediaQuery from '@mui/material/useMediaQuery';
+import useForkRef from '@mui/material/utils/useForkRef';
 import { visuallyHidden } from '@mui/utils';
 import _ from 'lodash';
 import React, { isValidElement, PropsWithChildren } from 'react';
@@ -56,6 +57,7 @@ import Loader from '../common/Loader';
 import { LightTooltip } from '../common/Tooltip';
 import ClusterChooser from './ClusterChooser';
 import ClusterChooserPopup from './ClusterChooserPopup';
+import { useResizableDialogSize } from './useResizableDialogSize';
 
 export interface ClusterTitleProps {
   clusters?: {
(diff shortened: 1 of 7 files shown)


illume (changes requested):
Thanks for these changes.


WasThatRudy:
@illume resolved all the copilot comments :)

PR #7010, opened 2026-08-08
backend: Fix cache keys for URL query parameters

### LINKED ISSUE #7004: Cache key ignores URL query parameters
## Describe the bug

When the Headlamp backend is run with the experimental `--cache-enabled=true` flag, the response cache does not include URL query parameters when generating cache keys.

`backend/pkg/k8cache/cacheStore.go:GenerateKey()` builds the cache key from the API group, resource kind, namespace, and context, but does not include `url.RawQuery`.

Because `CacheMiddleWare` uses this key to cache the HTTP response, different requests for the same resource with different query parameters can resolve to the same cache entry.

For example:

- `GET /api/v1/namespaces/default/pods?labelSelector=app=A`
- `GET /api/v1/namespaces/default/pods?labelSelector=app=B`

Both requests generate the same cache key. If the first request populates the cache, the second request can receive the cached response from the first request instead of the response for its own selector.

The same applies to query parameters used for pagination, such as `limit` and `continue`.

**Expected Outcome:**

Requests that can produce different Kubernetes API responses should not share the same response-cache entry. Query parameters that affect the response should either be represented in the cache key or the corresponding requests should bypass the cache.

## To Reproduce

Steps to reproduce the bug:

1. Start the Headlamp backend with caching enabled using `--cache-enabled=true` (or `HEADLAMP_CACHE_ENABLED=true`).
2. Make a request such as:
   `GET /api/v1/namespaces/default/pods?labelSelector=app=frontend`
3. Allow the response to populate the cache.
4. Make another request with a different selector:
   `GET /api/v1/namespaces/default/pods?labelSelector=app=backend`
5. Observe that the second request can return the cached response from the first request.

The same collision can be observed with pagination parameters, for example:

- `GET .../pods?limit=50`
- `GET .../pods?limit=50&continue=<token>`

- Installation type: Any
- Headlamp Version: Main branch
- Other: The issue requires the experimental `--cache-enabled=true` feature to be enabled.

Yes, I can propose a PR if the issue is confirmed.

## Additional Context

The issue was identified by tracing the response-cache flow from `CacheMiddleWare` to `GenerateKey()` and the cache store/load functions.

A minimal reproduction using the existing `GenerateKey()` implementation produces the same key for two different query strings:

```go
urlA, _ := url.Parse(
    "http://localhost/api/v1/namespaces/default/pods?labelSelector=app=A",
)
keyA, _ := k8cache.GenerateKey(urlA, "my-cluster")

urlB, _ := url.Parse(
    "http://localhost/api/v1/namespaces/default/pods?labelSelector=app=B",
)
keyB, _ := k8cache.GenerateKey(urlB, "my-cluster")

// keyA == keyB
// +pods+default+my-cluster

mayurgalhate:
## Summary

Fixes the Headlamp backend response cache so URL query parameters are included in cache keys. This prevents requests with different query parameters from incorrectly sharing the same cached response.

## Related Issue

Fixes #7004

## Changes

* Include canonicalized URL query parameters in cache keys using `url.Query().Encode()`.
* Ensure different selectors and pagination parameters such as `labelSelector`, `limit`, and `continue` produce distinct cache entries.
* Invalidate the base cache entry and all query variants when the underlying Kubernetes resource changes.
* Add regression tests for query-based cache keys and cache invalidation.
* Preserve existing cache-key behavior for requests without query parameters.

## Steps to Test

1. Enable the experimental backend cache with `--cache-enabled=true`.
2. Make a request with a query parameter, such as:
   `GET /api/v1/namespaces/default/pods?labelSelector=app=frontend`
3. Make another request with a different selector:
   `GET /api/v1/namespaces/default/pods?labelSelector=app=backend`
4. Verify that the requests do not share the same cache entry.
5. Verify that pagination parameters such as `limit` and `continue` also produce distinct cache keys.
6. Run the backend tests and lint checks.

Not applicable — this is a backend caching fix.

## Notes for the Reviewer

* Query parameters are canonicalized using `url.Query().Encode()` so equivalent parameter orderings produce the same cache key.
* Cache invalidation removes all query variants associated with the affected resource while preserving unrelated cache entries.
* No frontend changes are required for this backend-only fix.

### DIFF
--- a/backend/pkg/auth/auth_test.go
+++ b/backend/pkg/auth/auth_test.go
@@ -441,6 +441,10 @@ func (cacheStub) Delete(ctx context.Context, k string) error {
 	return nil
 }
 
+func (cacheStub) DeleteAll(ctx context.Context, _ cache.Matcher) error {
+	return nil
+}
+
 func (cacheStub) Get(ctx context.Context, k string) (interface{}, error) {
 	return nil, nil
 }
--- a/backend/pkg/cache/cache.go
+++ b/backend/pkg/cache/cache.go
@@ -31,6 +31,7 @@ type Cache[T any] interface {
 	Set(ctx context.Context, key string, value T) error
 	SetWithTTL(ctx context.Context, key string, value T, ttl time.Duration) error
 	Delete(ctx context.Context, key string) error
+	DeleteAll(ctx context.Context, selectFunc Matcher) error
 	Get(ctx context.Context, key string) (T, error)
 	GetAll(ctx context.Context, selectFunc Matcher) (map[string]T, error)
 	UpdateTTL(ctx context.Context, key string, ttl time.Duration) error
@@ -100,6 +101,20 @@ func (c *cache[T]) Delete(ctx context.Context, key string) error {
 	defer c.lock.Unlock()
 
 	delete(c.store, key)
+
+	return nil
+}
+
+// DeleteAll removes all values from the cache that match the given matcher function.
+func (c *cache[T]) DeleteAll(ctx context.Context, selectFunc Matcher) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	for key := range c.store {
+		if selectFunc != nil && selectFunc(key) {
+			delete(c.store, key)
+		}
+	}
 
 	return nil
 }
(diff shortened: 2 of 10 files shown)


mayurgalhate:
Hi @illume , I’ve addressed all 4 review comments and squashed the commits. Tests are passing. Please take a look when you get a chance. Thanks!

PR #7029, opened 2026-08-09
frontend: useKubeObject: Preserve cluster when handling WebSocket updates

### LINKED ISSUE #7023: frontend: useKubeObject: Preserve cluster when handling WebSocket updates
## Describe the bug

The `useKubeObject` WebSocket update handler creates updated resource instances without preserving the current cluster context.

The initial query creates objects using the active cluster, but the WebSocket update handler instantiates the object using only the resource data:

```ts
client.setQueryData(queryKey, new kubeObjectClass(update.object));
```

If the hook is being used with a non-default cluster, or if the active cluster changes while the hook remains mounted, the updated object may be associated with the wrong cluster. This can result in inconsistent resource state after WebSocket updates.

The expected behavior is for WebSocket updates to preserve the same cluster context used by the initial query.

## To Reproduce

The issue was identified through code inspection.

A possible reproduction scenario is:

1. Configure multiple Kubernetes clusters in Headlamp.
2. Open a resource view that uses `useKubeObject`.
3. Switch the active cluster or keep multiple cluster views active.
4. Trigger a WebSocket update for the resource.
5. Observe whether the updated resource instance retains the correct cluster association.

- Headlamp Version: `main`
- Other: Frontend (`useKubeObject` hook)

**Yes. I will propose a PR.**

## Additional Context

The hook already has access to the current `cluster` value, but it is not used when creating updated objects inside the WebSocket callback.

Passing the cluster to the resource constructor and including it in the callback dependencies would keep the update path consistent with the initial query path and ensure the correct cluster context is preserved after WebSocket updates.

This issue was identified during code review and would benefit from verification in a multi-cluster environment before merging a fix.

bhuvan-somisetty:
Fixes #7023

## Problem

`useKubeObject`'s initial query builds the resource instance with `new kubeObjectClass(obj, cluster)`, but the WebSocket update handler (`handleMessage`) only passed the raw object through:

```ts
client.setQueryData(queryKey, new kubeObjectClass(update.object));
```

If the hook is used against a non-default cluster, the instance produced by a watch event loses its cluster association even though the rest of the query (query key, initial fetch, endpoint resolution) stays scoped to that cluster. This shows up as inconsistent resource state after a WebSocket update in multi-cluster setups.

## Fix

Pass `cluster` through in `handleMessage` the same way the initial `queryFn` already does, and add `cluster` to the callback's dependency array so it stays correct if the hook is ever re-rendered with a different cluster.

## Testing

Added two tests to `hooks.test.tsx` covering both WebSocket transports the hook can use (legacy per-resource socket and the multiplexer), each simulating an update event and asserting the resulting instance still carries the original `cluster`. Both tests fail against the pre-fix code (confirmed locally by temporarily reverting the source change) and pass with the fix.

- `npx vitest run src/lib/k8s/api/v2/hooks.test.tsx` — 14/14 passing
- `npx tsc --noEmit` — clean
- `npm run lint` (eslint) on the changed files — clean

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.test.tsx
+++ b/frontend/src/lib/k8s/api/v2/hooks.test.tsx
@@ -435,4 +435,97 @@ describe('useKubeObject watch wiring', () => {
       expect(lastCall.connections[0]?.url).not.toContain('metadata.name%3Dpod-a');
     });
   });
+
+  it('preserves the cluster on the instance built from a legacy WebSocket update', async () => {
+    vi.stubEnv('REACT_APP_ENABLE_WEBSOCKET_MULTIPLEXER', 'false');
+    mockClusterFetch.mockResolvedValue(
+      mockJsonResponse({
+        apiVersion: 'v1',
+        kind: 'Pod',
+        metadata: { name: 'my-pod', namespace: 'my-ns' },
+      })
+    );
+
+    const { result } = renderHook(
+      () =>
+        useKubeObject({
+          kubeObjectClass: MockPod,
+          name: 'my-pod',
+          namespace: 'my-ns',
+          cluster: 'non-default-cluster',
+        }),
+      { wrapper: createWrapper() }
+    );
+
+    await waitFor(() => {
+      expect(result.current.data?.cluster).toBe('non-default-cluster');
+    });
+
+    expect(mockUseWebSockets).toHaveBeenCalled();
+    const connections = mockUseWebSockets.mock.calls.at(-1)![0].connections;
+    expect(connections).toHaveLength(1);
+
+    connections[0].onMessage({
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this PR.


<details>

- `frontend: useKubeObject: Preserve cluster when handling WebSocket updates` — Contains `Fixes #NN`; please remove the issue reference from the commit message.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for these changes.


bhuvan-somisetty:
Addressed the Copilot feedback: asserted the mock call counts before indexing into them so a failure points at the real problem instead of a raw index error. All 14 tests green, tsc and lint clean.

PR #7048, opened 2026-08-10
frontend: role: Link the service accounts in the binding list

### LINKED ISSUE #6928: Hyperlink service account in RoleBinding view
Make service account clickable in RoleBinding view. Currently have to click into the RoleBinding, where the service account is clickable.

Image

Clicking the service account in the RoleBinding list view takes you to the service account details

All kind of users

No

Anjali-Chauhan1:
## Summary

This PR makes service accounts clickable in the Role Bindings list view, so you can reach a service account without opening the binding first. It also fixes the sort comparator on that column, which never matched any subject and so did nothing.

## Related Issue

Fixes #6928

## Changes

- Added a `ServiceAccountList` renderer in `BindingList.tsx` that links each service account subject to its details page
- Fixed the Service Accounts column sort, which filtered for the kind `Service Accounts` instead of `ServiceAccount`
- Added a `ServiceAccount` subject to the ClusterRoleBinding story fixture, which previously only had a Group
- Added `BindingList.test.tsx` covering the link, the namespace fallback, and the plain text case

## Steps to Test

1. Go to Security, then Role Bindings
2. Find a binding that has something in the Service Accounts column
3. Click the service account name
4. You land on the service account details page, in the same cluster the binding belongs to

## Notes for the Reviewer

- The obvious approach of passing `<Link>` elements to `LabelListItem` does not work. Its prop type says `React.ReactNode[]`, but internally it does `labels.join(', ')`, so elements would render as `[object Object]`. I left that component alone rather than change something around 20 lists depend on, and rendered the links in the column instead, following the `RoleLink` helper already in this file. Happy to fix `LabelListItem` separately if you would prefer that.
- The namespace comes from `subject.namespace` and falls back to the binding's own namespace. If neither is present the name stays plain text, so no broken URL is generated.
- `activeCluster` is passed so the link stays on the binding's cluster. The details view does not need this, the list does.
- `getValue` is unchanged, so search, filtering and export still see plain names.
- The sort fix is a separate commit since it is a pre-existing bug and not part of the feature.
- No new translation strings, so the locale files are untouched.

### DIFF
--- a/frontend/src/components/role/BindingList.test.tsx
+++ b/frontend/src/components/role/BindingList.test.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { KubeRoleBinding } from '../../lib/k8s/roleBinding';
 import { TestContext } from '../../test';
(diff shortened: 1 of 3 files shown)


skoeva (changes requested):
this could be rebased, #7070 should fix the frontend test failures

Anjali-Chauhan1:
> this could be rebased, #7070 should fix the frontend test failures

Thanks @skoeva for review. Rebased onto main, which now includes #7070.

PR #7058, opened 2026-08-10
backend: serviceproxy: Allow optional Service port selection

### LINKED ISSUE #7057: feature: Headlamp's serviceproxy only proxies a Service when one of its ports is named `http` or `https`.
## Description

Many real Services use other port names (`web`, `metrics`, `api`) or unnamed ports (e.g. `8080`). Those Services are healthy, but:
GET /clusters/{cluster}/serviceproxy/{namespace}/{name}?request=/...

fails with 404 because port resolution rejects anything that isn't named `http`/`https`.

## Where

- `backend/pkg/serviceproxy/service.go` (`GetPort`)
- Route: `/clusters/{clusterName}/serviceproxy/{namespace}/{name}?request=...`

## Proposed solution

Add an optional query parameter:

- `?port=<port-name>` (e.g. `metrics`)
- `?port=<port-number>` (e.g. `8080`)

Behavior:

1. If `port` is omitted → keep current behavior (`http`/`https` named ports only; prefer `https`).
2. If `port` is set → select that port from the Service spec only (no arbitrary ports).
3. Unknown name/number → keep failing (404).

This stays backward compatible and limited to ports declared on the Service.

## Why this matters

Plugins / App Catalog / in-cluster HTTP apps often expose non-`http`/`https` named ports. Headlamp should be able to proxy those without forcing users to rename Service ports.

## Acceptance criteria

- [ ] Optional `port` query param selects Service port by name or number
- [ ] Omitting `port` preserves existing `http`/`https` default
- [ ] Only ports present on the Service are accepted
- [ ] Unit tests cover name/number selection and default behavior

CygnusMaximillian:
## Summary

This PR adds optional Service port selection to serviceproxy via `?port=<name|number>`, while keeping the existing `http`/`https` named-port default when `port` is omitted.

## Related Issue

Fixes #7057

## Changes

- Added `ResolvePort` and `GetServiceWithPort` in `backend/pkg/serviceproxy/service.go`
- Kept `GetService` / `GetPort` default behavior unchanged when no port is specified
- Wired optional `port` query param in `RequestHandler`
- Updated existing handler tests only as needed for the new function signatures

## Steps to Test

1. From `backend/`, run:
   ```bash
   go test ./pkg/serviceproxy/ -count=1
2. Confirm existing default behavior still works for Services with ports named http/https.
3. (Optional manual check) With Headlamp running against a cluster:
Service with port named http → GET /clusters/<cluster>/serviceproxy/<ns>/<svc>?request=/ still works without port
Service with port named web / number 8080 → GET /clusters/<cluster>/serviceproxy/<ns>/<svc>?request=/&port=web or ...&port=8080 now resolves instead of failing on port selection

### DIFF
--- a/backend/pkg/serviceproxy/service.go
+++ b/backend/pkg/serviceproxy/service.go
@@ -3,6 +3,8 @@ package serviceproxy
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/kubernetes-sigs/headlamp/backend/pkg/logger"
 
@@ -26,7 +28,21 @@ type proxyService struct {
 }
 
 // GetService returns the requested service based on the provided name and namespace.
+// It keeps the historical default of selecting a port named "http" or "https".
 func GetService(ctx context.Context, cs kubernetes.Interface, namespace string, name string) (*proxyService, error) {
+	return GetServiceWithPort(ctx, cs, namespace, name, "")
+}
+
+// GetServiceWithPort returns the requested service and resolves the Service port.
+// When portSelector is empty, behavior matches GetService (http/https named ports).
+// Otherwise portSelector may be a Service port name or port number that exists on the Service.
+func GetServiceWithPort(
+	ctx context.Context,
+	cs kubernetes.Interface,
+	namespace string,
+	name string,
+	portSelector string,
+) (*proxyService, error) {
 	service, err := cs.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
@@ -38,7 +54,7 @@ func GetService(ctx context.Context, cs kubernetes.Interface, namespace string, 
 		IsExternal: len(service.Spec.ExternalName) > 0,
 	}
 
-	port, err := GetPort(service.Spec.Ports)
+	port, err := ResolvePort(service.Spec.Ports, portSelector)
 	if err != nil {
 		logger.Log(logger.LevelError, nil, err, "service port not found")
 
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for the contribution.


backend/pkg/serviceproxy/service.go
```diff
@@ -77,6 +86,52 @@ func GetPort(ports []corev1.ServicePort) (*corev1.ServicePort, error) {
 	return nil, fmt.Errorf("no port found with the name http or https")
 }
 
+// ResolvePort selects a Service port.
+// An empty portSelector preserves GetPort behavior.
+// A non-empty selector matches a port by name, or by port number when the selector is numeric.
+// Only ports declared on the Service are accepted.
+func ResolvePort(ports []corev1.ServicePort, portSelector string) (*corev1.ServicePort, error) {
```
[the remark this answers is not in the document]
CygnusMaximillian: thanks that's expected for this PR.
This change is intentionally split: this PR is the backward-compatible `?port=` behavior only. Coverage for `ResolvePort` / `GetServiceWithPort` (name, number, unknown selector) and a non-empty `port` query case is in the follow-up tests PR: #7059.

backend/pkg/serviceproxy/service.go
```diff
@@ -38,22 +54,15 @@ func GetService(ctx context.Context, cs kubernetes.Interface, namespace string,
 		IsExternal: len(service.Spec.ExternalName) > 0,
 	}
 
-	port, err := GetPort(service.Spec.Ports)
+	port, err := ResolvePort(service.Spec.Ports, portSelector)
```
Psykii22, line 57: i think you should write some tests for the explicit usage of the name, port,443 or something like 9999 giving error and doing same for the schemeForPort testing the http,https,443 giving https

Psykii22: it would help make the code more robust and help prevent regression

CygnusMaximillian: Thankyou for your suggestion I have improved the test, looking forward to getting more valuable suggestion

PR #7073, opened 2026-08-10
frontend: units: Parse the lowercase k suffix as kilo

### LINKED ISSUE #7072: parseRam misreads the lowercase 'k' suffix, so 1k parses as 1 byte instead of 1000
### Describe the bug

`parseUnitsOfBytes` in `frontend/src/lib/units.ts` (behind `parseRam` and `parseDiskSpace`) does not accept the lowercase `k` suffix, so a valid Kubernetes memory quantity like `500k` is read as `500` bytes rather than `500000`. That is a 1000x under-report.

The regex only allows an uppercase `K`:

```ts
const groups = value.match(/(\d+(?:\.\d+)?)([BKMGTPEe])?(i)?(\d+)?/) || [];
```

Per the [Quantity reference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/), the decimal suffixes are `m | "" | k | M | G | T | P | E`, and the spec calls the capitalisation out explicitly: "Note that 1024 = 1Ki but 1000 = 1k; I didn't choose the capitalization."

### The codebase already disagrees with itself

`normalizeUnit` in `frontend/src/lib/util.ts` handles the same quantities and gets it right, with a comment citing the same spec:

```ts
/**
 * Decimal: m | n | "" | k | M | G | T | P | E
 */
...
} else if (quantity.endsWith('k')) {
  bytes *= 1000;
}
```

So the two helpers are exact mirror images of each other:

| quantity | `parseRam` | `normalizeUnit` | correct |
|---|---|---|---|
| `1k` | `1` | `1 KB` | 1000 |
| `500k` | `500` | `500 KB` | 500000 |
| `1024k` | `1024` | `1.02 MB` | 1024000 |
| `1K` | `1000` | `1 Bytes` | not a valid suffix |

The same quantity string can render as "500 KB" in one view while counting as 500 bytes in a chart or a sort in another.

### To reproduce

```ts
import { parseRam } from './units';
import { normalizeUnit } from './util';

parseRam('500k');                  // 500      <- wrong, expected 500000
normalizeUnit('memory', '500k');   // "500 KB" <- right
```

### Expected behavior

`parseRam('1k') === 1000`, matching the Kubernetes quantity spec and `normalizeUnit`.

Uppercase `K` is not a valid Kubernetes suffix, but `parseRam` has treated it as kilo for a long time, so I would keep accepting it rather than change behaviour anyone might be relying on.

### Notes

Related but separate: #6816 / #6817 covers the missing `m` (milli) suffix in the same function. That PR adds an `endsWith('m')` guard above the regex and leaves the character class untouched, so it does not cover this. The two fixes are complementary and touch different lines.

Yes, I have a fix with tests and will open a PR.

omlahore:
Fixes #7072

## Description

`parseUnitsOfBytes` in `frontend/src/lib/units.ts` only accepted an uppercase `K`, so a valid Kubernetes memory quantity like `500k` was read as `500` bytes instead of `500000`. That is a 1000x under-report, and it is silent.

Per the [Quantity reference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) the decimal suffixes are `m | "" | k | M | G | T | P | E`, and the spec calls out the capitalisation directly: "Note that 1024 = 1Ki but 1000 = 1k; I didn't choose the capitalization."

What convinced me this is a bug rather than a deliberate simplification is that `normalizeUnit` in `frontend/src/lib/util.ts` already gets it right, and cites the same spec in a comment. The two helpers were exact mirror images:

| quantity | `parseRam` before | `normalizeUnit` | correct |
|---|---|---|---|
| `1k` | `1` | `1 KB` | 1000 |
| `500k` | `500` | `500 KB` | 500000 |
| `1024k` | `1024` | `1.02 MB` | 1024000 |
| `1K` | `1000` | `1 Bytes` | not a valid suffix |

So the same quantity string could render as "500 KB" in one view while counting as 500 bytes in a chart or a sort elsewhere. `parseRam` and `parseDiskSpace` both go through this function, and `pod/List.tsx` and `node/Details.tsx` are the visible callers.

## Changes

Two lines. The lowercase `k` joins the suffix character class, and is folded onto the existing `K` before the `UNITS` lookup so the index math is unchanged:

```ts
const groups = value.match(/(\d+(?:\.\d+)?)([BkKMGTPEe])?(i)?(\d+)?/) || [];
...
const unitIndex = _.indexOf(UNITS, groups[2] === 'k' ? 'K' : groups[2]);
```

The binary `Ki` path and the `e`/`E` exponent path are untouched.

I kept uppercase `K` accepted. It is not a valid Kubernetes suffix, but it has been read as kilo here for a long time and removing it would change behaviour for anyone relying on it. Happy to drop it if you would rather be strict.

## Testing

Added a case to the existing `parseRam` suite covering `1k`, `500k`, `1.5k`, the retained `1K`, and `1Ki` to show the binary suffix stays distinct. It fails against unpatched `main`:

```
× parseRam > should parse the lowercase k suffix as kilo
  → expected 1 to be 1000
```

and passes with the change. Full file: 29 passed. `tsc`, `eslint` and `prettier --check` are clean.

## Note on overlap

#6817 fixes the missing `m` (milli) suffix in this same function. It adds an `endsWith('m')` guard above the regex and leaves the character class alone, so the two changes are complementary and touch different lines. If #6817 lands first I will rebase.

### DIFF
--- a/frontend/src/lib/units.test.ts
+++ b/frontend/src/lib/units.test.ts
@@ -47,6 +47,18 @@ describe('parseRam', () => {
     expect(parseRam('1e6')).toBe(1000000);
   });
 
+  it('should parse the lowercase decimal kilo suffix', () => {
+    expect(parseRam('1k')).toBe(1000);
+    expect(parseRam('64k')).toBe(64000);
+    expect(parseRam('1.5k')).toBe(1500);
+  });
+
+  it('should not treat the lowercase ki as a binary suffix', () => {
+    // Kubernetes binary suffixes are Ki, Mi, Gi and so on. There is no
+    // lowercase ki, so it must not be read as 1024.
+    expect(parseRam('1ki')).toBe(1);
+  });
+
   it('should parse decimal values with binary units', () => {
     expect(parseRam('1.5Ki')).toBe(1.5 * 1024);
     expect(parseRam('1.5Mi')).toBe(1.5 * 1024 * 1024);
--- a/frontend/src/lib/units.ts
+++ b/frontend/src/lib/units.ts
@@ -44,7 +44,11 @@ function parseUnitsOfBytes(value: string): number {
     return parseFloat(value.slice(0, -1)) / 1000;
   }
 
-  const groups = value.match(/(\d+(?:\.\d+)?)([BKMGTPEe])?(i)?(\d+)?/) || [];
+  // Kubernetes spells kilo with a lowercase k in the decimal suffixes
+  // (m | "" | k | M | G | T | P | E), so 1k is 1000 while 1Ki is 1024. The
+  // uppercase K is not a valid suffix, but it has been read as kilo here for a
+  // long time, so it stays accepted.
+  const groups = value.match(/(\d+(?:\.\d+)?)([BkKMGTPEe])?(i)?(\d+)?/) || [];
   const number = parseFloat(groups[1]);
 
   // number ex. 1000
@@ -57,7 +61,13 @@ function parseUnitsOfBytes(value: string): number {
     return number * 10 ** parseInt(groups[4], 10);
   }
 
-  const unitIndex = _.indexOf(UNITS, groups[2]);
+  // The binary suffixes are Ki, Mi, Gi and so on, all with an uppercase letter.
+  // There is no lowercase ki, so it is left unparsed rather than read as 1024.
+  if (groups[2] === 'k' && groups[3] !== undefined) {
+    return number;
+  }
+
+  const unitIndex = _.indexOf(UNITS, groups[2] === 'k' ? 'K' : groups[2]);
 
   // Unit + i ex. 1Ki
   if (groups[3] !== undefined) {


omlahore:
Verified this against a real cluster rather than only against the spec, since the obvious question is whether anything actually emits a lowercase `k`.

On a kind cluster (v1.36.1), a pod requesting `memory: "500k"` is accepted, and the API returns the suffix verbatim:

```console
$ kubectl -n unit-probe get pod mem-lowercase-k -o jsonpath='{.spec.containers[0].resources}'
{"limits":{"cpu":"200m","memory":"900k"},"requests":{"cpu":"100m","memory":"500k"}}
```

So the string that reaches the frontend really is `500k`, not a normalised `500000` or `488Ki`. The API server does not rewrite it. Before this change `parseRam('500k')` returns `500`, so that request is read as 500 bytes rather than 500 kilobytes wherever the parsed value is used, for example the pod list's resource columns and node details.

For contrast, the binary suffix on another pod in the same namespace round-trips as `64Mi` and parses correctly today, which is why this has gone unnoticed: `Ki`/`Mi`/`Gi` are what most manifests use, and they were never affected.

omlahore:
Ran this end to end against a live cluster, and the user-visible effect is worse than the "reads 1000x small" framing in the description. It is not a formatting nit, it is a false over-utilisation reading.

**Which quantities actually reach the frontend.** The API server canonicalises quantities, so I checked which ones keep the lowercase `k` rather than assuming. It only rewrites values that are an exact multiple of 1000k:

```
requested -> stored
1k        -> 1k
500k      -> 500k
750k      -> 750k
999k      -> 999k
1500k     -> 1500k
12345k    -> 12345k
1000k     -> 1M      (rewritten)
200000k   -> 200M    (rewritten)
```

So anything not landing exactly on a megabyte boundary arrives at the client with the suffix intact.

**What the pod list shows.** A running pod on kind v1.36.1 requesting `65432k` and limited to `131072k`, using ~224 Ki. `pod/List.tsx` feeds `parseRam` into the memory column's percentage, so hovering the memory cell gives:

Before, on `main`:
```
Request: 350.6% (224 Ki/63.9 Ki)
Limit:   175.0% (224 Ki/128 Ki)
```

After, with this PR:
```
Request: 0.4% (224 Ki/62.4 Mi)
Limit:   0.2% (224 Ki/125 Mi)
```

Same pod, same cluster, only `units.ts` differs. An idle pod using 0.4% of its request is reported as being at 350% of it, and the request itself renders as `63.9 Ki` instead of `62.4 Mi`. That is the kind of number someone acts on, by raising limits or chasing a memory leak that is not there.

Happy to add a screenshot if it helps, though the tooltip text above is the whole of it.

illume (changes requested):
Thanks for this PR.


<details>

- `fix(units): parse the lowercase k suffix as kilo` — Missing `area: description` prefix — e.g. `frontend: HomeButton: Fix so it navigates to home` or `backend: config: Add enable-dynamic-clusters flag`.

</details>

<details>

Good examples:

</details>

omlahore:
Thanks both, addressed in a force push.

**Commit guidelines:** reworded to `frontend: units: Parse the lowercase k suffix as kilo` and squashed, so there is a single commit rather than one that a later one rewrites.

**@copilot-pull-request-reviewer's point about `1ki`:** you're right, and it was a real regression I introduced. Adding lowercase `k` to the suffix class meant `1ki` matched as `k` plus `i` and returned 1024, where `main` returns 1:

| input | main | this branch, before | this branch, now |
|---|---|---|---|
| `1k` | 1 | 1000 | 1000 |
| `1K` | 1000 | 1000 | 1000 |
| `1Ki` | 1024 | 1024 | 1024 |
| `1ki` | 1 | **1024** | 1 |

The binary suffixes are all capitalised and there is no lowercase `ki`, so the lowercase branch is now guarded against a following `i` and falls back to the bare number, matching `main`. Regression test added, plus explicit cases for `1k`, `64k` and `1.5k`.

`units.test.ts` is green at 31 tests.

illume (changes requested):
Thanks for the contribution.


omlahore:
Marked it resolved. That was the `1ki` regression I already fixed and described in my last comment, just hadn't clicked resolve on the thread itself.

omlahore:
@illume ready for another look when you have a minute. The `1ki` regression Copilot caught is fixed with a test, and that thread is resolved. Head is still bf65d75d, same commit you reviewed, so the changes-requested is just sitting there.

omlahore:
@illume nudging this one, since it has been sitting at changes-requested since the 13th with nothing left outstanding on my side. The `1ki` regression Copilot caught is fixed with a test and that thread is resolved, the commits are squashed and reworded to kernel style, and all 15 checks pass. Head is still bf65d75d.

omlahore:
@vyncent-t @ashu8912 could one of you take a look at this? It has been at changes-requested since 13 Aug with nothing outstanding on my side. The `1ki` regression Copilot caught is fixed with a test, that thread is resolved, and all 15 checks pass on bf65d75d.

Also flagging that #7423 fixes the same thing. It was opened on 23 Aug, this one on 10 Aug, and both are sitting at changes-requested. Whichever you prefer, it is probably worth picking one rather than leaving both parked.

omlahore:
Dropped the duplicate lowercase-k test block that Copilot flagged. `1k` and `1.5k` were already covered a few tests up, and `1K` and `1Ki` in the decimal and binary blocks, so it was the same assertions written twice. The `1ki` case stays, that one is the only new behaviour.

PR #7076, opened 2026-08-10
frontend: Table: Revert scroll preservation (does not work)

WasThatRudy:
Reverts #6787.

The ref passed via `muiTableContainerProps.ref` never attaches because MRT in our configuration does not render a `TableContainer` element in the DOM. The layout effect in `useScrollPreservationOnDataChange` short-circuits on the null ref check, so the hook is a no-op. The element that actually scrolls is a page-level ancestor, outside the Table component's tree.

Verified in Storybook: walking up from `.MuiTable-root` shows only `MuiBox-root` divs and `<body>`, all with `overflowY: visible` and `scrollHeight === clientHeight`. No `TableContainer` in the tree.

Reverting so main is clean while I look at whether a real fix (wiring into the page-level scroller) is worth doing in a follow-up.

Bug report: 5701


illume (approved):
Thanks!

PR #7081, opened 2026-08-10
frontend: KubeObjectDetails: Fix header cut off in embedded details

Adi-204:
## Problem

Details rendered outside their own page -- the details drawer, activity windows and the resource map panel -- go through `KubeObjectDetails`, which pulls the content up with a `marginTop: '-70px'`. That margin was a way to swallow the row the back link takes at the top of the details page.

Since the details header became a sticky block of its own, that offset no longer matches the markup, and it breaks in two ways:

- **While the header is pinned**, sticky keeps it at the top of the scroll container, but the rest of the details still start 70px higher. The pinned header ends up drawn over the first rows of the metadata table -- in a small activity window the `Name` row is not reachable at all.
- **Where the header is not pinned**, the negative margin does what it says and takes the top of the resource title off the visible area.

## Fix

Drop the negative margin and let the details sit where they are laid out. The back link is visible again, the title is no longer cut off, and the pinned header no longer hides the top of the content.

Before Fix  

image

After Fix 

image

## Testing

Rendered a details view inside a 460x700 scroll container with a title bar, mimicking an activity window, and measured the geometry in a browser:

| | before | after |
| --- | --- | --- |
| back link | -12px, clipped above the panel | at the top of the panel, visible |
| resource title | 11px above the visible area | fully visible |
| first metadata row | covered by the pinned header | visible |
| scrolled 500px | -- | header stays pinned, content scrolls under it |

`vitest run` over `src/components/resourceMap` and `src/components/common/Resource` passes (41 files, 419 tests); the full suite passed as well. Prettier and CI lint are clean.


Adi-204:
@illume can you pls review this small one!

Adi-204:
@illume we don't need this anymore as it is fixed by PR https://github.com/kubernetes-sigs/headlamp/pull/7103 which is already merged!

illume (changes requested):
Thanks for the contribution.


<details>

</details>

PR #7091, opened 2026-08-10
app: Implement graceful shutdown for backend server process

### LINKED ISSUE #7090: app: Implement graceful shutdown for the backend server process
**Is your feature request related to a problem? Please describe.**
Currently, when the Headlamp desktop application quits, it forcibly kills the backend server process. In `app/electron/main.ts`, there is a `@todo` comment questioning if we should attempt a more graceful shutdown:

// @todo: should we try and end the process a bit more gracefully?
// What happens if the kill signal doesn't kill it?
serverProcess.kill();

Force-killing the process without allowing it to clean up resources or connections can lead to orphaned processes or state issues.

**Describe the solution you'd like**
Implement a graceful shutdown sequence for the backend server process in the Electron app:
1. Send a `SIGTERM` signal to `serverProcess` first to allow it to clean up and exit gracefully.
2. Introduce a timeout (e.g., 5 seconds) to wait for the process to exit.
3. If the process has not exited after the timeout, fallback to forcefully terminating it with `SIGKILL`.

**Describe alternatives you've considered**
Leaving it as a forceful kill, but this risks leaving zombie processes if the backend was in the middle of something critical.

**Additional context**
This requires modifying the `quitServerProcess` function inside `app/electron/main.ts`.

vishnukothakapu:
## Summary

This PR adds a graceful shutdown sequence for the backend server process in the Electron application.

## Related Issue
Fixes #7090 

## Changes

- Updated `quitServerProcess()` in `app/electron/main.ts` to implement coordinated shutdown. On Windows, this closes `stdin` so the backend can detect the shutdown event reliably. On non-Windows platforms, it sends a `SIGTERM` signal.
- Added a **15-second timeout fallback** to forcefully kill the process via `SIGKILL` if it hasn't exited gracefully in time. This longer deadline is intentional because backend HTTP shutdown and telemetry flushing can consume up to 10 seconds.
- Bounded MCP client cleanup with a **3-second timeout** to prevent nonresponsive transports from hanging the overall application quit sequence.
- Fixed a TOCTOU (Time-of-Check to Time-of-Use) race condition in the test server setup by injecting the listener directly into the server runner.

## Steps to Test

1. Run the desktop app using `npm run app:start`.
2. Wait for the backend server to spin up and connect successfully.
3. Close the application.
4. Verify that the backend server process shuts down correctly in the terminal logs and leaves no orphaned processes behind.

## Notes for the Reviewer

- Addresses the `@todo: should we try and end the process a bit more gracefully?` comment in `app/electron/main.ts`.

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -801,7 +801,10 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
   const options = {
     detached: true,
     windowsHide: true,
-    env: withBackendMemoryDefaults(extendedEnv),
+    env: {
+      ...withBackendMemoryDefaults(extendedEnv),
+      HEADLAMP_ELECTRON: 'true',
+    },
   };
 
   return spawn(serverFilePath, serverArgs, options);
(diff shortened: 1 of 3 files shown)


illume (changes requested):
Thanks for this PR.


<details>

</details>

PR #7094, opened 2026-08-11
frontend: auth: Clear cached resource data for the logged-out cluster in logout()

### LINKED ISSUE #7093: frontend: auth: logout() only clears auth cache, so re-login within staleTime serves stale resource data without refetching
### Describe the bug

`logout()` in `lib/auth.ts` removes `['auth']` and `['clusterMe', cluster]` entries from the React Query cache but does not touch cached resource data. Pods, Secrets, ConfigMaps etc. are cached under `['kubeObject', 'list', ...]` and `['object', ...]` keys and these survive logout.

The app-wide `staleTime` in `queryClient.ts` is 3 minutes and list queries do not override it. If a user logs back into the same cluster within that window, React Query serves the pre-logout snapshot as fresh data with no network request and no visual indicator that it is stale.

AuthRoute in RouteSwitcher.tsx prevents unauthenticated access after logout in single-cluster views (the ['auth', cluster] query is correctly cleared) — but this check is bypassed entirely for multi-cluster group views (RouteSwitcher.tsx:200-205, when more than one cluster is selected), so that path isn't covered by this issue. For the single-cluster case, this is not an unauthenticated data exposure: the gap is in the re-login path, where the user has a valid token again, AuthRoute passes, but the resource cache still holds data from the previous session.

### Environment

- Installation type: any
- Headlamp Version: main
- Other: React Query v5, default staleTime 180000ms, default gcTime 300000ms

Yes, I will propose a PR. The fix is adding a `removeQueries` call with a cluster-scoped predicate to `logout()` so all cached resource entries for that cluster are cleared alongside the existing auth cleanup. Other clusters' caches are unaffected.

### Additional Context

Not a regression. The existing `removeQueries` calls in `logout()` were each added to address specific symptoms (auth-check state, username display) and neither addressed the general resource-data cache.

beep-boopp:
Fixes #7093

`logout()` removes `['auth']` and `['clusterMe', cluster]` from the React Query cache but
leaves all resource data (`['kubeObject', 'list', ...]`, `['object', ...]`) untouched. If a
user logs back into the same cluster within the 3-minute `staleTime` window, React Query
serves the pre-logout snapshot without refetching.

Added a `removeQueries` call with a cluster-scoped predicate so all cached entries for that
cluster are cleared on logout. Other clusters' caches are not affected.

Note: this covers the resource-cache staleness described in #7093. It does not touch the
separate `AuthRoute` auth-gate bypass for multi-cluster group views mentioned there — that's
a related but distinct gap.

**Testing:** `frontend/src/lib/auth.test.ts` — new `describe('logout', ...)` block seeds
cache entries for two clusters, calls `logout('prod')`, and asserts `prod` entries are gone
while `staging` entries survive untouched.

### DIFF
--- a/frontend/src/lib/auth.test.ts
+++ b/frontend/src/lib/auth.test.ts
@@ -15,10 +15,11 @@
  */
 
 import { Base64 } from 'js-base64';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import store from '../redux/stores/store';
-import { getUserInfo, setToken } from './auth';
+import { getUserInfo, logout, setToken } from './auth';
 import { backendFetch } from './k8s/api/v2/fetch';
+import { queryClient } from './queryClient';
 
 // Mock the dependencies
 vi.mock('./k8s/api/v2/fetch');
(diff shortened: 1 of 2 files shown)


beep-boopp:
Hey @illume PTAL whenever possible 🙏

beep-boopp:
Scoped the predicate to positional matching for the two resource data families this PR targets (kubeObject at index 4, object at index 1). Added a test case where a staging cluster entry has 'prod' as its namespace to verify it survives logout('prod'). PTAL @illume

PR #7097, opened 2026-08-11
frontend: CopyableCell: Integrate copy in tables

### LINKED ISSUE #3983: Add a button to quickly copy values from table cells where it matters

rootp1:
### Summary
This is an extension of the stale PR #4459. It adds a hover copy button on table cells (IP, address columns) so users can grab values like Pod IP, Cluster IP, External IP, Internal IP, and Endpoint/EndpointSlice Addresses without selecting text manually.

It also carries the fixes requested in review on #4459: reuses ActionButton instead of a bespoke button, avoids recomputing getValue twice per cell, cleans up the copy reset timer on unmount, makes the button usable from the keyboard, and tightens the copyable column type. Rebased onto latest main to resolve the branch's merge conflict, and fixed a flaky storyshot test plus a couple of translation file issues (duplicate/missing keys) uncovered along the way.

### Related Issue
Fixes #3983

### Steps to Test

- Open the Nodes, Pods, Services, Endpoints, or EndpointSlices list.
- Hover over an IP/address column (or tab to it) and click the copy icon that appears.
- Paste to confirm the correct value was copied.
- Check a row with no IP (e.g. a node with no external IP) to confirm the fallback text still shows and nothing gets copied incorrectly.

Screen recording of the copy interaction (Cluster IP and External IP columns):

https://github.com/user-attachments/assets/b83002fc-229e-4f63-be6c-3c2f80073424

Adds a hover copy button to IP and address columns in the resource tables so values can be copied without selecting text.

### DIFF
--- a/frontend/src/components/common/Resource/CopyButton.tsx
+++ b/frontend/src/components/common/Resource/CopyButton.tsx
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { Box } from '@mui/material';
+import { visuallyHidden } from '@mui/utils';
 import type { ComponentProps, MouseEventHandler } from 'react';
 import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -46,24 +48,31 @@ export default function CopyButton(props: CopyButtonProps) {
   async function onCopy(event: Parameters<MouseEventHandler<HTMLElement>>[0]) {
     onClick?.(event);
 
+    clearTimeout(resetTimeoutRef.current);
+
     try {
       await navigator.clipboard.writeText(copyText);
       setCopied(true);
-      clearTimeout(resetTimeoutRef.current);
       resetTimeoutRef.current = setTimeout(() => setCopied(false), 1500);
     } catch (err) {
+      setCopied(false);
       console.error('Failed to copy to clipboard:', err);
     }
   }
 
   return (
-    <ActionButton
-      description={copied ? t('translation|Copied!') : t('translation|Copy to clipboard')}
-      buttonStyle={buttonStyle}
-      onClick={onCopy}
-      icon={copied ? 'mdi:check' : 'mdi:content-copy'}
-      iconButtonProps={iconButtonProps}
-      width={width}
-    />
+    <>
+      <ActionButton
+        description={copied ? t('translation|Copied!') : t('translation|Copy to clipboard')}
+        buttonStyle={buttonStyle}
+        onClick={onCopy}
+        icon={copied ? 'mdi:check' : 'mdi:content-copy'}
+        iconButtonProps={iconButtonProps}
+        width={width}
+      />
+      <Box role="status" aria-live="polite" aria-atomic="true" sx={visuallyHidden}>
+        {copied ? t('translation|Copied!') : ''}
+      </Box>
+    </>
   );
 }
--- a/frontend/src/components/common/Resource/CopyButton.tsx
+++ b/frontend/src/components/common/Resource/CopyButton.tsx
@@ -14,32 +14,56 @@
  * limitations under the License.
  */
 
+import type { ComponentProps, MouseEventHandler } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import ActionButton, { ButtonStyle } from '../ActionButton';
 
 interface CopyButtonProps {
   text?: string;
   buttonStyle?: ButtonStyle;
+  iconButtonProps?: ComponentProps<typeof ActionButton>['iconButtonProps'];
+  width?: ComponentProps<typeof ActionButton>['width'];
+  onClick?: MouseEventHandler<HTMLElement>;
 }
 
 export default function CopyButton(props: CopyButtonProps) {
-  const { text, buttonStyle } = props;
+  const { text, buttonStyle, iconButtonProps, width, onClick } = props;
   const { t } = useTranslation(['translation']);
+  const [copied, setCopied] = useState(false);
+  const resetTimeoutRef = useRef<ReturnType<typeof setTimeout>>();
+
+  useEffect(() => {
+    return () => clearTimeout(resetTimeoutRef.current);
+  }, []);
 
   if (text === undefined || text === null || text === '') {
     return <></>;
   }
 
-  async function onCopy() {
-    await navigator.clipboard.writeText(text!);
+  const copyText = text;
+
+  async function onCopy(event: Parameters<MouseEventHandler<HTMLElement>>[0]) {
+    onClick?.(event);
+
+    try {
+      await navigator.clipboard.writeText(copyText);
+      setCopied(true);
+      clearTimeout(resetTimeoutRef.current);
+      resetTimeoutRef.current = setTimeout(() => setCopied(false), 1500);
+    } catch (err) {
+      console.error('Failed to copy to clipboard:', err);
+    }
   }
 
   return (
     <ActionButton
-      description={t('translation|Copy to clipboard')}
+      description={copied ? t('translation|Copied!') : t('translation|Copy to clipboard')}
       buttonStyle={buttonStyle}
       onClick={onCopy}
-      icon="mdi:content-copy"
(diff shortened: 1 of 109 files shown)


rootp1:
Hey @skoeva 
I didn't drop in <CopyButton> asis. 
Instead I refactored CopyableCell to use ActionButton, which is the same underlying primitive CopyButton itself is built on (CopyButton = ActionButton + a hardcoded mdi:content-copy icon + clipboard write, no more). 
That replaced the original spoke Tooltip + IconButton markup, so the duplication you flagged is gone.

rootp1:
@illume please have a review whenever you have a chance to

frontend/src/i18n/locales/de/translation.json
```diff
@@ -301,14 +301,15 @@
...
-  "Fetching release information…": "",
-  "Skip": "",
-  "Failed to fetch release information": "",
-  "An update is available": "",
-  "Read more": "",
+  "Release Notes ({{ appVersion }})": "Versionshinweise ({{ appVersion }})",
+  "Fetching release information…": "Abrufen von Release-Informationen…",
+  "Skip": "Überspringen",
+  "Failed to fetch release information": "Fehler beim Abrufen von Release-Informationen",
+  "An update is available": "Ein Update ist verfügbar",
```
skoeva: This fills in hundreds of previously-empty translations - unrelated to the copy buttons, and some of them are wrong. Please drop everything except the `Copied!` key; translations need native-speaker review in their own PR

frontend/src/components/common/Resource/CopyableCell.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /** The content to display in the cell */
+  children: ReactNode;
+}
+
+/**
+ * A wrapper component that adds a copy-to-clipboard button on hover.
+ * Used for table cells containing values that users commonly need to copy,
+ * such as IP addresses, hostnames, etc.
+ */
+export default function CopyableCell({ value, children }: CopyableCellProps) {
```
skoeva, line 33: This is going into the public plugin API via `index.tsx` but has no test and no story - please add coverage here

frontend/src/components/common/Resource/CopyableCell.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return () => {
+      clearTimeout(resetTimeoutRef.current);
+    };
+  }, []);
+
+  if (!value) {
+    return <>{children}</>;
+  }
+
+  async function handleCopy(e: MouseEvent) {
```
skoeva: Fair point that `CopyButton` is thin, but this still re-implements its three jobs: the clipboard write, the empty-value guard, and the `Copy to clipboard` label - and both are exported to plugins, so authors now get two overlapping copy components. Suggest moving the copied-state feedback and an `iconButtonProps` passthrough into `CopyButton` and having this render it; that also gives the detail-view copy in `MainInfoSectionHeader` the same feedback, which it currently lacks

frontend/src/components/endpoints/List.tsx
```diff
@@ -43,6 +43,7 @@ export default function EndpointList() {
           getValue: endpoint => endpoint.getAddresses().join(', '),
```
skoeva: Copying yields `10.0.0.1, 10.0.0.2` for multi-address rows, which isn't very passable into another tool. Consider newline-joining the copy value, or only setting `copyable` when there's a single address. Same applies in `endpointSlices/List.tsx`

frontend/src/components/common/Resource/ResourceTable.tsx
```diff
@@ -438,9 +448,33 @@ function ResourceTableContent<RowItem extends KubeObject>(props: ResourceTablePr
...
+                );
+              };
+            } else {
+              mrtColumn.Cell = ({ row }: { row: MRT_Row<RowItem> }) =>
+                renderFn?.(row.original) ?? null;
+            }
+          } else if ('render' in column) {
             mrtColumn.Cell = ({ row }: { row: MRT_Row<RowItem> }) =>
               column.render?.(row.original) ?? null;
+          } else if (column.copyable && 'getValue' in column) {
```
skoeva, line 473: There's no row virtualization here, so every copyable cell - both this branch and the `render` one above - adds two `Box`, and `IconButton` and a MUI `Tooltip` per row, a few hundred extra instances per page across five copyable columns. Probably fine, but worth a check on a large Pod list

skoeva (changes requested):
thanks for picking this up, and for keeping the previous author in the commits. this PR needs a rebase and the translation changes are the main blocker for me

frontend/src/components/common/Resource/ResourceTable.tsx
rootp1: I checked this path and kept the current structure after reusing CopyButton and limiting it to the existing copyable columns.

frontend/src/components/common/Resource/CopyableCell.tsx
rootp1: I moved the copy feedback into CopyButton and reused it here.

rootp1:
I pushed the requested updates and addressed the open review points.

rootp1:
@skoeva please have a chance to have a look

illume (changes requested):
Thanks for working on this.


frontend/src/components/common/Resource/CopyButton.tsx
```diff
@@ -14,32 +14,56 @@
...
+    onClick?.(event);
+
+    try {
+      await navigator.clipboard.writeText(copyText);
+      setCopied(true);
+      clearTimeout(resetTimeoutRef.current);
+      resetTimeoutRef.current = setTimeout(() => setCopied(false), 1500);
+    } catch (err) {
+      console.error('Failed to copy to clipboard:', err);
+    }
```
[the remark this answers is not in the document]
rootp1: Fixed. The catch now resets the copied state and clears the pending timer so a failed retry does not keep showing the success icon.

frontend/src/components/common/Resource/CopyButton.tsx
```diff
@@ -14,32 +14,56 @@
...
   }
 
   return (
     <ActionButton
-      description={t('translation|Copy to clipboard')}
+      description={copied ? t('translation|Copied!') : t('translation|Copy to clipboard')}
```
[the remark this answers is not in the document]
rootp1: Added a hidden aria-live status region that announces Copied! the same way the table status feedback does.

frontend/src/components/common/Resource/CopyableCell.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    vi.useFakeTimers();
+    writeText.mockClear();
+    Object.defineProperty(window.navigator, 'clipboard', {
+      configurable: true,
+      value: { writeText },
+    });
+  });
+
+  it('copies the value and shows copied feedback', async () => {
```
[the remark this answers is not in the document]
rootp1: CI is green now on this branch, that failure was transient.

frontend/src/components/common/Resource/CopyableCell.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /** The text value to copy to clipboard */
+  value: string;
+  /** The content to display in the cell */
+  children: ReactNode;
+}
+
+/**
+ * A wrapper component that adds a copy-to-clipboard button on hover.
+ * Used for table cells containing values that users commonly need to copy,
+ * such as IP addresses, hostnames, etc.
```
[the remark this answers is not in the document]
rootp1: This repo's commit convention is a subject line only, so I'm keeping the rationale in the PR description instead.

illume (approved):
🎉 thanks!

PR #7111, opened 2026-08-11
frontend: Map: Make Back return to the map overview

### LINKED ISSUE #7110: Map: Closing the details panel leaves a dead history entry, so Back does nothing
## Describe the bug
On the Map, clicking a node pushes a `?node=<id>` history entry (`GraphView.tsx`, `setSelectedNodeId` → `useQueryParamsState` push) and opens the details activity panel. Closing the panel (X button or the taskbar tab) leaves that entry and the node highlight behind, because the panel is not tied to the URL state.

So the first Back press after closing the panel only clears the highlight and stays on the same page — it looks like Back is broken and takes a second press to actually navigate. The same desync happens the other way: pressing Back while the panel is open clears the highlight but leaves the panel open.

Expected: closing the panel returns to the plain map view, and one Back press from a selected node lands on the map overview (panel closed, nothing highlighted).

## To Reproduce
1. Go to Map.
2. Click any node — the details panel opens on the right, node is highlighted.
3. Close the panel with the X or its taskbar tab.
4. Press Back (browser button or mouse button 4) — the highlight disappears but the page does not change. Press Back again to actually navigate.

 - Installation type: Linux, dev server (`make run-frontend`)
 - Headlamp Version: main (578bb1945)

Yes (I will propose a PR).

xonas1101:
Clicking a Map node pushes a `?node=<id>` history entry and opens the details activity panel, but the panel and that entry were never kept in sync: closing the panel left the entry and the node highlight behind, so the first Back press only cleared the highlight and stayed on the page. Pressing Back with the panel open had the mirror problem — highlight gone, panel still open.

Fixes: https://github.com/kubernetes-sigs/headlamp/issues/7110

- `GraphView.tsx`: when the panel closes while its node is still selected, pop the entry the click pushed; when the selection changes (e.g. Back), close the panel of the previous selection.
- `useQueryParamsState.tsx`: setting a param to the value it already has no longer stacks an identical history entry (clicking the same node twice).
- `useQueryParamsState.test.tsx`: test for that guard.

How to test: open the Map, click a node, close the panel with the X or its taskbar tab, press Back — you should leave the Map instead of only losing the highlight. Then click a node again and press Back with the panel open: panel closes and the highlight clears in one press.

### DIFF
--- a/frontend/src/components/resourceMap/GraphView.tsx
+++ b/frontend/src/components/resourceMap/GraphView.tsx
@@ -37,10 +37,12 @@ import {
 } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
+import { useHistory } from 'react-router';
 import Namespace from '../../lib/k8s/namespace';
 import K8sNode from '../../lib/k8s/node';
 import { setNamespaceFilter } from '../../redux/filterSlice';
 import { useTypedSelector } from '../../redux/hooks';
+import { Activity } from '../activity/Activity';
 import { NamespacesAutocomplete } from '../common/NamespacesAutocomplete';
 import { filterGraph, filterGraphIncremental, GraphFilter } from './graph/graphFiltering';
 import { getGraphForNamespaceSelection } from './graph/graphForNamespaceSelection';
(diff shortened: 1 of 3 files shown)


xonas1101:
@illume PTAL

xonas1101:
@illume PTAL

PR #7116, opened 2026-08-11
frontend: vpa: re-evaluate enablement on cluster switch

yashveersinghh:
### What & Why
`VPAList` only checked `VPA.isEnabled()` on mount. Switching active clusters didn't trigger a re-check, leaving stale state if VPA wasn't installed on the new cluster.

### Changes
- Added `cluster` to `useEffect` deps in `List.tsx`
- Added unit test in `VPAList.test.tsx`

### How to test
- Run `npm test frontend/src/components/verticalPodAutoscaler/VPAList.test.tsx`
- Run `npm run lint`

All tests pass cleanly.


illume (changes requested):
Thanks for the contribution.


<details>

- `frontend: vpa: re-evaluate enablement on cluster switch` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

yashveersinghh:
Updated the commit message as requested. Thanks!

yashveersinghh:
Addressing the suppressed comments:
Updated the cluster-switch handling and added a test for stale async responses. Tests pass locally.

unlikelyzero (changes requested):
Thanks for chasing down the cluster-switch staleness bug here — the core fix (the `cancelled` flag, resetting `vpaEnabled` to `null`, and adding `cluster` to the effect deps) looks right and does close the stale-overwrite race that was flagged earlier.

Two things I'd like to see addressed before merging though:

1. The new test in `VPAList.test.tsx` doesn't actually exercise the race it's meant to cover. It fully resolves cluster-a's `isEnabled()` promise and lets the UI settle *before* switching to cluster-b, so the two requests are never in flight at the same time — this test would still pass even without the `cancelled` guard. Can you rework it so cluster-a's promise resolves *after* switching to cluster-b (and after cluster-b's own promise resolves), and assert the UI keeps cluster-b's result rather than reverting to cluster-a's?

2. There's a one-frame window where a cluster switch renders with the new `cluster` value but the previous cluster's `vpaEnabled` state, since the reset happens in `useEffect` (post-commit) rather than during render. If the previous cluster had VPA enabled, `ResourceListView`/`VPA.useList()` can briefly mount and fire a request against the new cluster before the effect flips the gate closed. Worth keying the enablement result to the cluster it was checked for (e.g. store `{cluster, enabled}` and only render the list when they match), or keying the subtree by cluster, so stale state can't drive a render of list-fetching children.

Good catch on the underlying bug — just want the test to actually prove it, and the edge case above closed out, before this goes in.

yashveersinghh:
hi @unlikelyzero thanks for the feedback!
I have updated both points:

Fixed 1-frame gap: Keyed enablement state directly to the active cluster during render (checkedState.cluster === cluster ? ...) so cluster switches reset state immediately.

Updated test: Reworked VPAList.test.tsx to simulate an out-of-order response (Cluster A resolves after Cluster B finishes) and verified late responses are discarded.

Tests are passing cleanly locally!

yashveersinghh:
Squashed intermediate commits into a single atomic commit as requested. Ready for review!

yashveersinghh:
Addressed the rapid cluster switch issue state now clears synchronously on transition to prevent stale cached results, and added a unit test covering pending switches. Should be good for review now!

yashveersinghh:
@illume The latest build discover and deploy failure is from verify-image-digests, reporting outdated architecture-specific digests in Dockerfile / Dockerfile.plugins (golang:1.26.5 and node:22). These files are outside this PR's changes, so I haven't modified them to avoid unrelated changes. The VPA changes and tests are otherwise passing.

unlikelyzero (approved):

yashveersinghh:
Hi @illume, just following up on this PR. The code changes are ready and the remaining CI failure is still the unrelated image:verify-image-digests check I mentioned above. The PR has been approved by unlikelyzero. Please let me know if any further action is needed from my side.

PR #7120, opened 2026-08-11
backend: k8cache: Fix potential DoS from gzip bombs

### LINKED ISSUE #7119: backend: Prevent gzip decompression DoS with response size limit
## Describe the bug

The `GetResponseBody()` function in `backend/pkg/k8cache/cacheStore.go` uses `io.ReadAll()` to decompress gzip-encoded Kubernetes API responses without any limit on the amount of data that can be read.

If a Kubernetes API server returns a highly compressed payload (for example, a gzip bomb), the decompressed data could become extremely large. Since `io.ReadAll()` tries to read the entire response into memory, this could cause the Headlamp backend to consume excessive memory and eventually crash with an OOM error.

The decompressed response should have a reasonable size limit so that excessively large payloads are rejected instead of being fully loaded into memory.

## To Reproduce

1. Intercept or mock a Kubernetes API response handled by Headlamp.
2. Return a highly compressed gzip payload that expands to a very large size.
3. Observe the Headlamp backend's memory usage increase significantly while decompressing the response.
4. The process may eventually run out of memory and crash.

## Environment

* Installation type: Any (Backend)
* Headlamp Version: Current main
* Other: N/A

Yes (I will propose a PR).

## Additional Context

I have a fix prepared locally that uses `io.LimitReader` to limit the decompressed response to 50 MB:

```go
io.LimitReader(reader, 50*1024*1024+1)
```

Responses exceeding the limit are rejected instead of being fully read into memory.

mayurgalhate:
## Summary

This PR fixes a Denial of Service (DoS) vulnerability in the Kubernetes API caching layer by enforcing a maximum size limit when decompressing gzip payloads. 

## Related Issue

Fixes #7119

## Changes

- Updated `GetResponseBody()` in `backend/pkg/k8cache/cacheStore.go` to wrap the gzip reader with `io.LimitReader`.
- Enforced a hard limit of 50MB for decompressed payloads to prevent malicious or misconfigured Kubernetes APIs from sending "gzip bombs" that cause Out-Of-Memory (OOM) crashes in the Headlamp backend.

## Steps to Test

1. Ensure you have a standard Headlamp backend running (`npm run backend:start`).
2. Verify that general cluster navigation and caching of large resources (like viewing a large number of Pods or large ConfigMaps) still functions normally without hitting the 50MB limit.
3. Review the code changes in `cacheStore.go` to confirm the `io.LimitReader` is appropriately placed before `io.ReadAll`.

N/A (Backend-only security fix)

## Notes for the Reviewer

- I chose 50MB (`50 * 1024 * 1024 + 1`) as a generous upper bound for a single decompressed Kubernetes API response to ensure legitimate large clusters don't experience truncation, while still preventing catastrophic OOM crashes.

### DIFF
--- a/backend/pkg/k8cache/cacheStore.go
+++ b/backend/pkg/k8cache/cacheStore.go
@@ -83,10 +83,18 @@ func GetResponseBody(bodyBytes []byte, encoding string) (string, error) {
 
 		defer func() { _ = reader.Close() }()
 
-		decompressedBody, err := io.ReadAll(reader)
+		// Use LimitReader to prevent gzip bomb DoS (max 50MB)
+		decompressedBody, err := io.ReadAll(io.LimitReader(reader, 50*1024*1024+1))
 		if err != nil {
 			logger.Log(logger.LevelError, nil, err, "failed to decompress body")
 			return "", fmt.Errorf("failed to decompress body: %w", err)
+		}
+
+		if len(decompressedBody) > 50*1024*1024 {
+			err = fmt.Errorf("response body exceeds maximum allowed size (50MB)")
+			logger.Log(logger.LevelError, nil, err, "failed to decompress body")
+
+			return "", err
 		}
 
 		dcmpBody = decompressedBody


illume (changes requested):
Thanks for these changes.

PR #7123, opened 2026-08-11
backend: Fix incorrect caching behavior for Kubernetes resources containing "Failure"

### LINKED ISSUE #7122: backend: Avoid skipping valid Kubernetes resources containing "Failure" in cache
## Describe the bug

In `backend/pkg/k8cache/cacheStore.go`, the `StoreK8sResponseInCache()` function currently checks whether the decompressed response body contains the word `"Failure"` using `strings.Contains(dcmpBody, "Failure")`.

This can cause valid Kubernetes resources to be skipped from the cache if they happen to contain the word `"Failure"` somewhere in their data, labels, or spec. For example, a ConfigMap containing `"Failure"` as part of its data would not be cached even though the response itself was successful.

This can lead to unnecessary API requests and reduce the effectiveness of the Kubernetes response cache.

Instead of searching the response body for a specific word, Kubernetes error responses should be identified using their HTTP status code.

## To Reproduce

1. Create a valid Kubernetes ConfigMap containing the string `"Failure"` in one of its data fields.
2. Access the ConfigMap through Headlamp.
3. Observe that the response is not stored in the cache.
4. Subsequent requests fetch the resource from the Kubernetes API again instead of using the cached response.

## Environment

* Installation type: Any (Backend caching layer)
* Headlamp Version: Current main
* Other: N/A

Yes (I will propose a PR).

## Additional Context

I have a fix prepared locally that removes the `strings.Contains(dcmpBody, "Failure")` check and prevents caching responses with HTTP status codes `>= 400`.

This makes the cache decision based on the response status rather than the contents of the response body.

mayurgalhate:
## Summary

This PR fixes an issue where valid Kubernetes resources containing the word `"Failure"` were incorrectly excluded from the cache.

## Related Issue

Fixes #7122 

## Changes

* Updated `StoreK8sResponseInCache()` in `backend/pkg/k8cache/cacheStore.go`.
* Changed the cache check to skip all HTTP error responses (`>= 400`).
* Removed the `strings.Contains(dcmpBody, "Failure")` check that could incorrectly reject valid resources.

## Steps to Test

1. Create a valid Kubernetes resource, such as a ConfigMap, containing `"Failure"` in its data.
2. Access the resource through Headlamp.
3. Verify that the successful response is stored in the Kubernetes cache.
4. Request the resource again and verify that the cached response is used.

Not applicable.

## Notes for the Reviewer

* Cache filtering is now based on the HTTP status code rather than searching the response body for `"Failure"`.
* This prevents valid resources from being incorrectly excluded from the cache.

### DIFF
--- a/backend/pkg/k8cache/cacheStore.go
+++ b/backend/pkg/k8cache/cacheStore.go
@@ -83,10 +83,18 @@ func GetResponseBody(bodyBytes []byte, encoding string) (string, error) {
 
 		defer func() { _ = reader.Close() }()
 
-		decompressedBody, err := io.ReadAll(reader)
+		// Use LimitReader to prevent gzip bomb DoS (max 50MB)
+		decompressedBody, err := io.ReadAll(io.LimitReader(reader, 50*1024*1024+1))
 		if err != nil {
 			logger.Log(logger.LevelError, nil, err, "failed to decompress body")
 			return "", fmt.Errorf("failed to decompress body: %w", err)
+		}
+
+		if len(decompressedBody) > 50*1024*1024 {
+			err = fmt.Errorf("response body exceeds maximum allowed size (50MB)")
+			logger.Log(logger.LevelError, nil, err, "failed to decompress body")
+
+			return "", err
 		}
 
 		dcmpBody = decompressedBody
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for the contribution.


The backend test job in CI is failing. Run `cd backend && go test ./...` to reproduce the errors locally.

<details>

</details>

illume (changes requested):
Thanks for working on this.

PR #7124, opened 2026-08-11
frontend: useKubeObjectList: Guard empty-string namespace watch predicate

gambtho:
## Summary

This PR fixes a latent regression in `useKubeObjectList`, where a watch entry with an
empty-string namespace can be dropped by a truthiness check that was reintroduced by a
long-lived branch after it had already been fixed.

`9542b5cb5` ("frontend: useKubeObjectList: Fix infinite re-render when namespace is empty
string") matched watch entries against requested namespaces with an explicit
`!== undefined` check instead of truthiness. An empty-string namespace is a real value —
it means "all namespaces" for a namespaced resource — but it is falsy, so the truthiness
check dropped the entry on every render while query data immediately re-added it, looping
`setState`.

`1f35a61bc` ("frontend: pod list: Add pagination") was authored 2026-05-15, before that
fix, and merged 2026-07-27, after it. The merge reinstated the old truthiness form.

**This is latent, not live.** On the watching path `keptListsToWatch` is discarded: the
`shouldWatch` branch recomputes `nextListsToWatch` from `query.data` and never derives
from it, so the stale predicate is only consumed when `shouldWatch` is false, where it
converges. I could not produce a user-visible failure on current `main` and do not believe
one exists today. The reason it stays inert is incidental, though — a future refactor that
feeds `keptListsToWatch` back into the watch path would reanimate the original infinite
re-render, with no test standing in the way.

## Related Issue

None — found while testing a third-party plugin that passes `namespace: ''` to `useList`,
a common `useState('')` pattern.

## Changes

- Restored the explicit `!== undefined` check in the namespace-matching branch of the
  watch-reconciliation predicate, exactly as `9542b5cb5` had it.
- Extracted that predicate into an exported `isWatchedListStillRequested` helper so it is
  directly testable, leaving the `useEffect` call site to a single filter.
- Added unit tests covering both branches of the helper, citing `9542b5cb5` in a comment.
  Following review, added a sixth case pinning an empty-string namespace against a
  *specific* requested namespace. Without it, `''` was never checked against a non-empty
  namespace list anywhere in the suite, so an implementation treating `''` as
  always-matching would have passed every other test.
- Documented that `initialData` on `useKubeObject` and `KubeObject.useGet` must be
  reference-stable. It is seeded through a layout effect, so a value rebuilt on every
  render re-seeds on every render. The one in-tree caller (`EditButton.tsx`) passes a
  stable prop; this is a comment-only change aimed at plugin authors and future callers.
  It rides along deliberately: it is a second, unrelated re-render hazard in the same
  hook family, found during the same investigation of a plugin passing `namespace: ''`.
  It has no behavioral effect — happy to split it into a follow-up PR if preferred.

## Steps to Test

1. `cd frontend && npx vitest run src/lib/k8s/api/v2/useKubeObjectList.test.tsx`
2. Observe 37 tests passing.
3. To confirm the new tests discriminate, mutate `isWatchedListStillRequested`:
   - Flip either branch back to its truthiness form (`watching.namespace === undefined` /
     `!!watching.namespace`). Each mutation fails exactly one of the two
     empty-string-namespace tests.
   - Or short-circuit the else branch with `watching.namespace === '' ||`, treating `''`
     as universal. That fails only the sixth test, and no other.

N/A — no user-visible change.

## Notes for the Reviewer

- The unfiltered branch deliberately keeps `!watching.namespace` rather than
  `=== undefined`, matching `9542b5cb5`, which changed only the else branch.
  `kubeObjectListQuery` defaults `namespace` to `''`, so a request carrying no namespace
  filter produces a watch entry whose namespace is `''`, not `undefined`; tightening that
  branch would drop it. The second test pins this case.
- The behavior change is confined to the predicate; extracting the helper is otherwise a
  pure move. `keptListsToWatch` is only consumed on the `!shouldWatch` path, which is why
  this is a defensive fix rather than a live one.
- Verified locally: `vitest` 37/37 on the touched file, `tsc --noEmit` clean, `eslint`
  clean, `prettier --check` clean.

### DIFF
--- a/frontend/src/lib/k8s/KubeObject.ts
+++ b/frontend/src/lib/k8s/KubeObject.ts
@@ -433,6 +433,11 @@ export class KubeObject<T extends KubeObjectInterface | KubeEvent = any> {
     opts?: {
       queryParams?: QueryParameters;
       cluster?: string;
+      /**
+       * Seed data to use before the query resolves. Must be reference-stable across
+       * renders (memoize it or capture it once) — an object rebuilt on every render
+       * will cause repeated re-renders.
+       */
       initialData?: K;
     }
   ) {
--- a/frontend/src/lib/k8s/api/v2/hooks.ts
+++ b/frontend/src/lib/k8s/api/v2/hooks.ts
@@ -122,6 +122,11 @@ export function useKubeObject<K extends KubeObject>({
   /** Cluster name */
   cluster?: string;
   queryParams?: QueryParameters;
+  /**
+   * Seed data to use before the query resolves. Must be reference-stable across
+   * renders (memoize it or capture it once) — an object rebuilt on every render
+   * will cause repeated re-renders.
+   */
   initialData?: K;
 }): [K | null, ApiError | null] & QueryResponse<K, ApiError> {
   type Instance = K;
--- a/frontend/src/lib/k8s/api/v2/useKubeObjectList.test.tsx
+++ b/frontend/src/lib/k8s/api/v2/useKubeObjectList.test.tsx
@@ -247,6 +247,17 @@ describe('isWatchedListStillRequested', () => {
 
     expect(isWatchedListStillRequested(watching, requests)).toBe(false);
   });
+
+  // The mirror image of the 9542b5cb5 regression: '' must not be read as "matches
+  // anything" either. Without this case, '' is never checked against a non-empty
+  // namespace list anywhere in the suite, so an implementation short-circuiting ''
+  // to a match would pass every other test here.
+  it('should drop a watch entry with an empty-string namespace when a specific namespace is requested', () => {
+    const watching = { cluster: 'default', namespace: '' };
+    const requests = [{ cluster: 'default', namespaces: ['default'] }];
+
+    expect(isWatchedListStillRequested(watching, requests)).toBe(false);
+  });
 });
 
 describe('useWatchKubeObjectLists', () => {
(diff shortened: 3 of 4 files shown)


Ralthos (commented):
The extraction is the right call. Exporting the predicate so the test imports the shipped function, with no copy of the logic beside it, is what makes these five cases worth having.

I checked the claim in the doc comment about where `''` comes from and it holds: `kubeObjectListQuery` declares `namespace: string | undefined = ''`, so a request with no namespace filter does produce a watch entry carrying `''` and not `undefined`. The unfiltered branch keeping its falsy check follows from that.

One case the table does not cover, and it is the one that pins this fix in place:

```ts
it('should drop a watch entry with an empty-string namespace when a specific namespace is requested', () => {
  const watching = { cluster: 'default', namespace: '' };
  const requests = [{ cluster: 'default', namespaces: ['default'] }];

  expect(isWatchedListStillRequested(watching, requests)).toBe(false);
});
```

The current code returns `false` there, correctly. The gap is that an implementation treating `''` as always-matching would pass all five existing tests: `should drop a watch entry whose namespace is no longer requested` uses `'gone'`, so `''` is never checked against a non-empty namespace list anywhere in the suite.

The regression being fixed here is `''` read as absent. The mirror-image mistake is `''` read as universal, and at the moment nothing would catch it.

Smaller thing: the `initialData` JSDoc added to `KubeObject.ts` and `hooks.ts` describes a different cause of re-renders from the one in the title. Worth having, just unexplained in the description.

PR #7129, opened 2026-08-11
frontend: TooltipLight: Fix dropped styles via slotProps

### LINKED ISSUE #7080: frontend: TooltipLight's sx never reaches the tooltip, so multi-line titles collapse and its styling is dropped
## Describe the bug

`TooltipLight` never applies any of its own styling. Every tooltip it renders comes out as the
standard themed tooltip - black background, `0.875rem` text - and, more visibly, multi-line titles
are collapsed onto one line because `white-space: pre-line` is lost with the rest.

`frontend/src/components/common/Tooltip/TooltipLight.tsx` styles the tooltip by passing `sx`
directly to MUI's `Tooltip`:

```tsx
<Tooltip
  disableInteractive={disableInteractive}
  TransitionComponent={Fade}
  TransitionProps={{ timeout: 0 }}
  sx={theme => ({
    backgroundColor: theme.palette.background.default,
    color: theme.palette.resourceToolTip.color,
    boxShadow: theme.shadows[1],
    fontSize: '1rem',
    whiteSpace: 'pre-line',
  })}
  {...rest}
>
  <span>{children}</span>
</Tooltip>
```

MUI's `Tooltip` has no root element of its own. `sx` is not in its `_excluded` list, so it falls
into `other` (`Tooltip.js:262`) and is then spread onto the anchor child (`Tooltip.js:470`):

```js
const childrenProps = _extends({}, nameOrDescProps, other, children.props, { ... });
```

The child here is a plain `<span>`, so `sx` is emitted as a DOM attribute and dropped by React:

```
Warning: Invalid value for prop `sx` on <span> tag. Either remove it from the element, or pass a
string or number value to keep it in the DOM.
    at span
    at Tooltip (@mui/material/Tooltip/Tooltip.js:229:59)
    at TooltipLight (frontend/src/components/common/Tooltip/TooltipLight.tsx:13:11)
```

React drops the value rather than writing it out, so the styles simply vanish and the tooltip
surface keeps its defaults. Measured in jsdom:

| | `white-space` | `background-color` | `font-size` |
| --- | --- | --- | --- |
| current | (unset) | `rgba(97, 97, 97, 0.92)` | `0.6875rem` |
| same styles via `slotProps.tooltip.sx` | `pre-line` | as specified | `1rem` |

In the app the background is the theme override (`MuiTooltip.styleOverrides.tooltip` sets
`backgroundColor: '#000'`, `color: '#fff'`, `fontSize: '0.875rem'` in `lib/themes.ts`) rather than
MUI's default grey, but the point stands: none of `TooltipLight`'s own values reach the DOM.

One consequence worth calling out: `theme.palette.resourceToolTip.color` is declared in both the
light and dark palettes, and exposed to plugins in
`plugins/headlamp-plugin/src/additional.d.ts`, but this dropped `sx` is its only reader. It has
never had an effect.

The second branch of the component (non-string children) never sets the styling at all, so both
paths produce a plain themed tooltip.

## To Reproduce

Steps to reproduce the bug:
1. Open any Deployments list with a workload that has more than one container.
2. Hover the **Containers** cell.
3. The tooltip shows the container names run together on one wrapped line.

`frontend/src/components/deployments/List.tsx:131`:

```tsx
const containerText = containers.join(', ');
const containerTooltip = containers.join('\n');
return (
  <LightTooltip title={containerTooltip} interactive>
    {containerText}
  </LightTooltip>
);
```

The title is deliberately newline-joined so each container gets its own line; `whiteSpace:
'pre-line'` was the mechanism, and it is not applied. The same `join('\n')` pattern is used for
containers and images in `cronjob/List.tsx`, `daemonset/List.tsx`, `deployments/List.tsx`,
`job/List.tsx` and `statefulset/List.tsx`, and for readiness gates in `pod/List.tsx:519`.

 - Installation type: source checkout (development)
 - Headlamp Version: main @ 233816c5
 - Other: @mui/material 5.18.0

Yes (I will propose a PR)

## Additional Context

`sx` needs to go to the tooltip slot rather than to `Tooltip` itself:

```tsx
slotProps={{
  tooltip: {
    sx: theme => ({
      backgroundColor: theme.palette.background.default,
      color: theme.palette.resourceToolTip.color,
      boxShadow: theme.shadows[1],
      fontSize: '1rem',
      whiteSpace: 'pre-line',
    }),
  },
}}
```

I confirmed the fixed shape produces `white-space: pre-line` and the intended background, against
the current code producing neither.

Two things to decide, which is why I am opening an issue rather than going straight to a PR:

1. Both branches should presumably get the styling, since a `TooltipLight` with element children is
   currently indistinguishable from a plain `Tooltip`.
2. Callers already pass their own `slotProps` (e.g. `pod/List.tsx:75` sets `slotProps.popper`), so
   the component needs to merge rather than overwrite.

Separately, `frontend/src/components/globalSearch/GlobalSearchContent.tsx:545` passes
`sx={{ opacity: query.length ? 1 : 0 }}` to a `Tooltip` in the same way. That one happens to work,
because the child is an `IconButton` and MUI components do accept a forwarded `sx` - but it is
styling the button rather than the tooltip, which is probably not what was meant either.

Pratham19106:
## Summary

This PR fixes `TooltipLight` silently dropping all its custom styling (light background, text color, font size, and `white-space: pre-line`) because `sx` was passed directly to MUI's `<Tooltip>`, which has no root DOM element and forwards unrecognised props onto the anchor child. The styles are moved to `slotProps.tooltip.sx` so MUI applies them to the actual tooltip popover surface. Both render branches (string and ReactElement children) are consolidated into one, and caller-provided `slotProps` are safely merged instead of overwritten.

## Related Issue

Fixes #7080

## Changes

- **Updated `TooltipLight.tsx`**:
  - Moved styling to `slotProps.tooltip.sx` so MUI applies custom styles (background, font size, multi-line support) directly to the tooltip popover box.
  - Safely merged any custom `slotProps` passed by parent components.
  - Simplified and combined duplicate rendering logic for text and element children into a single clean path.
- **Added `TooltipLight.test.tsx`**:
  - Added unit tests to verify style application, multi-line text rendering (`\n`), and theme compatibility.

## Steps to Test

1. Run the frontend dev server: `npm start` in `frontend/`.
2. Navigate to **Workloads → Deployments** and select a deployment with multiple containers.
3. Hover the **Containers** cell.
4. Verify the tooltip now renders with:
   - A **light/themed background** (not the default dark MUI tooltip).
   - Each container name on a **separate line** (`white-space: pre-line` working).
   - Correct **font size** (`1rem`, not the default `0.875rem`).
5. Run the unit tests to confirm all pass:
   ```bash
   cd frontend && npx vitest run src/components/common/Tooltip/TooltipLight.test.tsx
   ```

### Before:
Before - Dark Tooltip with collapsed lines

### After:
After - Styled Light Tooltip with multi-line container list

The fix restores styles that were silently dropped; the tooltip surface changes from the default dark appearance to the intended light-themed appearance.

## Notes for the Reviewer

- This is my second PR to the organization! (beacuse the bot rejected the first one 🥲)I would be delighted if you could suggest any improvements or feedback.

### DIFF
--- a/frontend/src/components/common/Tooltip/TooltipLight.test.tsx
+++ b/frontend/src/components/common/Tooltip/TooltipLight.test.tsx
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @vitest-environment jsdom
+
+import { ThemeProvider } from '@mui/material/styles';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, expect, it } from 'vitest';
+import { createMuiTheme } from '../../../lib/themes';
+import TooltipLight from './TooltipLight';
+
+const lightTheme = createMuiTheme({ name: 'light', base: 'light' });
+const darkTheme = createMuiTheme({ name: 'dark', base: 'dark' });
+
+describe('TooltipLight', () => {
+  it('does not forward sx to the anchor element', () => {
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <TooltipLight title="tip">
+          <button data-testid="trigger">hover</button>
+        </TooltipLight>
+      </ThemeProvider>
+    );
+
+    // The bug: sx was passed directly to <Tooltip>, which spread it onto the
+    // child <button> as a DOM attribute. React dropped it, so the tooltip
+    // surface styles were never applied.
+    expect(screen.getByTestId('trigger')).not.toHaveAttribute('sx');
+  });
+
+  it('renders newline-separated text preserving each line', async () => {
+    const user = userEvent.setup();
+    const multiLine = 'nginx\ncoredns\nistio-proxy';
+
+    render(
+      <ThemeProvider theme={lightTheme}>
+        <TooltipLight title={multiLine}>
+          <button data-testid="trigger">containers</button>
+        </TooltipLight>
+      </ThemeProvider>
+    );
+
+    await user.hover(screen.getByTestId('trigger'));
+
+    const tooltip = await screen.findByRole('tooltip');
+
+    // The tooltip text must preserve the newlines that the deployment
+    // container list builds with containers.join('\n').
+    expect(tooltip.textContent).toBe(multiLine);
+
+    // The tooltip surface itself is the element with the .MuiTooltip-tooltip class.
+    // That is where our slotProps.tooltip.sx styles should be applied.
+    const tooltipSurface = tooltip.className.includes('MuiTooltip-tooltip')
+      ? tooltip
+      : tooltip.querySelector('.MuiTooltip-tooltip');
+
+    // Verify that the required styling actually reached the tooltip surface
+    expect(tooltipSurface).toHaveStyle({ whiteSpace: 'pre-line' });
+  });
+
+  it('renders the tooltip under the dark theme without leaking sx', async () => {
+    const user = userEvent.setup();
+
+    render(
+      <ThemeProvider theme={darkTheme}>
+        <TooltipLight title="dark tip">
+          <button data-testid="trigger">hover</button>
+        </TooltipLight>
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for working on this.


<details>

</details>

illume (changes requested):
Thanks for this PR.


Pratham19106:
Hello @illume 

Thank you for the feedback! I’ve updated the PR to address all the comments:

Did the changes requested by copilot:-
1. **`componentsProps` fallback:** Updated `TooltipLight.tsx` to include `componentsProps?.tooltip` as a fallback when merging styles, ensuring legacy consumers aren’t broken.
2. **Style assertions in unit tests:** Updated `TooltipLight.test.tsx` to query `.MuiTooltip-tooltip` and verify that `white-space: pre-line` is correctly applied to the tooltip surface.

Did the changes requested by you:-
Rebased the branch onto the latest `main` to remove all merge commits.

Really sorry for making you review my PR twice. 🙇

PR #7135, opened 2026-08-12
frontend: api/v2: Guard WebSocket message JSON parsing against malformed frames

### LINKED ISSUE #7101: frontend: api/v2: Guard WebSocket message JSON parsing against malformed frames
## Describe the bug
In `createWebSocket` (`frontend/src/lib/k8s/api/v2/webSocket.ts`), the message
listener parses every incoming frame without error handling:

    const data = type === 'json' ? JSON.parse(body.data) : body.data;

If a frame isn't valid JSON, `JSON.parse` throws uncaught inside the message
listener and the frame is silently dropped instead of being handled cleanly.
The callback dispatch two lines below is already wrapped in try/catch, and
the equivalent handler in `multiplexer.ts` (`handleWebSocketMessage`) wraps
its own `JSON.parse` too — this path is inconsistent with both.

Expected: a malformed frame should be caught, logged, and skipped, the same
way `multiplexer.ts` already handles it — not throw uncaught inside the
listener.

## To Reproduce
Steps to reproduce the bug:
1. Open `frontend/src/lib/k8s/api/v2/webSocket.ts` and look at the
   `message` event listener inside `createWebSocket`.
2. Trigger a WebSocket connection created through this path (e.g. any
   `useWebSocket`/`useWebSockets` consumer) and have the server send a
   non-JSON frame.
3. See the parse throw uncaught inside the listener instead of being caught
   and logged.

 - Installation type: any (source/dev build)
 - Headlamp Version: current main
 - Other: `frontend/src/lib/k8s/api/v2/webSocket.ts`

Yes (I will propose a PR).

## Additional Context
Not a regression tied to a specific version — the code has looked like this
for a while. The fix is small: wrap the parse in try/catch, log the error,
and return early on failure, matching the pattern already used in
`multiplexer.ts`.

krsatyamthakur-droid:
## Summary

createWebSocket's message listener parsed every incoming frame with no error handling, so a non-JSON frame threw uncaught inside the listener instead of being handled. The callback dispatch two lines below is already wrapped in try/catch, and the equivalent handler in multiplexer.ts (handleWebSocketMessage) guards its own JSON.parse too — this path was the odd one out.

This wraps the parse in try/catch, logs the error, and returns early on failure, matching multiplexer.ts's pattern.

## Related Issue

Fixes #7101

## Changes

- Wrapped the JSON.parse in createWebSocket's message listener (frontend/src/lib/k8s/api/v2/webSocket.ts) in try/catch, logging and returning early on a malformed frame
- Added a regression test that sends a malformed frame and checks the connection keeps delivering subsequent valid messages instead of the listener staying broken

## Steps to Test

1. Run `npx vitest run src/lib/k8s/api/v2/webSocket.test.ts` from frontend/
2. Confirm the new "drops a malformed frame instead of throwing, and keeps delivering later messages" test passes
3. Or manually: open a WebSocket connection through this path and have the server send a non-JSON frame — confirm it's logged and skipped instead of throwing uncaught, and later valid messages still arrive

## Notes for the Reviewer

- This mirrors the existing guard pattern already used in multiplexer.ts's handleWebSocketMessage, so behavior stays consistent between the two WebSocket paths.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/webSocket.test.ts
+++ b/frontend/src/lib/k8s/api/v2/webSocket.test.ts
@@ -170,6 +170,36 @@ describe('useWebSockets', () => {
     });
   });
 
+  it('drops a malformed frame instead of throwing, and keeps delivering later messages', async () => {
+    const url = 'api/v1/pods?watch=1&resourceVersion=2.5';
+    const server = new WS(`${BASE_WS_URL}${url}`);
+    const onMessageA = vi.fn();
+    const consoleError = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+    renderHook(() =>
+      useWebSockets({
+        connections: [{ cluster: '', url, onMessage: onMessageA }],
+      })
+    );
+
+    await server.connected;
+    await server.send('not valid json');
+
+    await waitFor(() => {
+      expect(consoleError).toHaveBeenCalledWith(
+        'WebSocket message parse error:',
+        expect.any(SyntaxError)
+      );
+    });
+    expect(onMessageA).not.toHaveBeenCalled();
+
+    await server.send(JSON.stringify({ type: 'DELETED', object: { metadata: { uid: 'pod-c' } } }));
+
+    await waitFor(() => {
+      expect(onMessageA).toHaveBeenCalledTimes(1);
+    });
+  });
+
   it('keeps a shared websocket alive until the last listener unsubscribes', async () => {
     const url = 'api/v1/pods?watch=1&resourceVersion=3';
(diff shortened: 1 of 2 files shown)


krsatyamthakur-droid:
The `e2e` failure looks unrelated - `helmPlugins.spec.ts` waits for the flux plugin to show up after pulling it from `artifacthub`, and that download never landed, so the table only had the shipped prometheus one. this PR is just a try/catch around JSON.parse in `webSocket.ts` plus its test, nothing that could affect a plugin install.
rebased on main since the branch had drifted a long way behind, which re-ran the job too. ptal when you get a chance @illume

PR #7141, opened 2026-08-12
frontend: Deployments: Fix Pods column sort with no available replicas

### LINKED ISSUE #7117: frontend: Deployments: Pods column sorts arbitrarily when availableReplicas is absent
## Describe the bug
The Deployments list sorts the "Pods" column with `sortByPods` (`frontend/src/components/deployments/List.tsx:33`):

```js
const availSorted = avail1 - avail2;
if (availSorted === 0) {
  return r1 - r2;
}
```

Kubernetes omits `status.availableReplicas` (and `status.replicas`) when they are zero — the renderer right above compensates with `${availableReplicas || 0}/${replicas || 0}`, but the comparator does not. Any deployment with no available pods makes the subtraction `NaN`, so the comparator returns `NaN` instead of -1/0/1 and those rows end up in arbitrary order: clicking the header a second time does not reverse the list. The sibling lists (`replicaset/List.tsx:56`, `statefulset/List.tsx:64`) use `|| 0`.

The same field is used raw for `getValue` (line 97), so those rows also export/search as an empty value.

Expected: `0/3` rows sort together and by their replica count, below `1/1` and `3/3`.

## To Reproduce
1. Have a namespace with some deployments that have no available pods (e.g. a bad image, so they show `0/1`, `0/2`, `0/3`) alongside healthy ones (`1/1`, `3/3`).
2. Go to Workloads > Deployments.
3. Click the "Pods" column header.
4. The `0/x` rows are not ordered by replica count and the healthy rows land arbitrarily; clicking the header again does not produce the reverse order. The neighbouring "Replicas" column sorts correctly on the same rows.

 - Installation type: Linux, dev server (`make run-frontend` / `make run-backend`)
 - Headlamp Version: main (578bb1945)

Yes (I will propose a PR).

## Screenshots
Image

Image

Image

xonas1101:
Fixes #7117

Kubernetes omits `status.availableReplicas` and `status.replicas` when they are zero, so the Pods comparator returned `NaN` and those rows landed in an arbitrary order. Falls back to 0, matching the renderer and the ReplicaSet/StatefulSet lists.

Moved the comparator into `sortByPods.ts` with a unit test; `getValue` gets the same fallback so empty cells don't export/search as blank.

### DIFF
--- a/frontend/src/components/deployments/List.tsx
+++ b/frontend/src/components/deployments/List.tsx
@@ -21,6 +21,7 @@ import Deployment from '../../lib/k8s/deployment';
 import { StatusLabel } from '../common/Label';
 import ResourceListView from '../common/Resource/ResourceListView';
 import LightTooltip from '../common/Tooltip/TooltipLight';
+import { sortByPods } from './sortByPods';
 
 export default function DeploymentsList() {
   const { t } = useTranslation(['glossary', 'translation']);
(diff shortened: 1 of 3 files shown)


xonas1101:
@illume PTAL

xonas1101:
@illume PTAL

PR #7144, opened 2026-08-12
frontend: app: Add script permissions to the azure-aks plugin

sniok:
adds necessary permissions for the upcoming plugin
this a configuration change and doesn't modify any logic inside the app

tested locally with an unreleased plugin


vyncent-t (approved):
LGTM

PR #7150, opened 2026-08-12
frontend: plugin: Handle plugin initialization errors

kehach07:
Wrap plugin.initialize() in a try-catch block and dispatch a PLUGIN_LOADING_ERROR event to the store if it fails. This prevents a broken plugin from corrupting the dashboard state and ensures that users are notified when a plugin fails to initialize.

## Summary

This PR fixes a missing error handler during the plugin initialization lifecycle by safely catching exceptions and dispatching the failure to the global state.

## Related Issue

Fixes #ISSUE_NUMBER  *(Note: Replace ISSUE_NUMBER if you opened an issue for this, otherwise you can remove this line)*

## Changes

- Added a `try-catch` block around `plugin.initialize(new Registry())` in `frontend/src/plugin/index.ts`
- Updated the logic to dispatch a `HeadlampEventType.PLUGIN_LOADING_ERROR` Redux event when an initialization exception occurs

## Steps to Test

1. Create a local test plugin and intentionally throw an error inside the plugin's `initialize` method (e.g., `throw new Error("Test init failure");`).
2. Run the application (`npm start`).
3. Verify that the Headlamp UI successfully loads the rest of the application without crashing or hanging.
4. Check the developer console to confirm the caught error is logged, and verify the `PLUGIN_LOADING_ERROR` action is dispatched to the Redux store.

## Notes for the Reviewer

- This addresses a long-standing `@todo` in `index.ts`. It ensures a misbehaving community plugin cannot crash the core dashboard when it starts up.

### DIFF
--- a/frontend/src/plugin/index.ts
+++ b/frontend/src/plugin/index.ts
@@ -122,10 +122,18 @@ export async function initializePlugins() {
     for (const pluginName of Object.keys(window.plugins)) {
       const plugin = window.plugins[pluginName];
       try {
-        // @todo: what should happen if this fails?
         plugin.initialize(new Registry());
       } catch (e) {
         console.error(`Plugin initialize() error in ${pluginName}:`, e);
+        store.dispatch(
+          eventAction({
+            type: HeadlampEventType.PLUGIN_LOADING_ERROR,
+            data: {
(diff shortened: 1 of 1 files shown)


illume (changes requested):
Thanks for these changes.

PR #7160, opened 2026-08-13
frontend: Multiplexer: Pause background Kubernetes watches on hidden tab

### LINKED ISSUE #7139: Pause Kubernetes resource watches when the browser tab is hidden

When Headlamp is open in multiple browser tabs, each tab continues to maintain its own Kubernetes `watch` connections, even when it has been hidden or inactive for a long time.

For example, three open tabs may create three similar sets of `list/watch` requests through Headlamp to the Kubernetes API. In multi-user environments, the number of active watches grows with the number of users and open tabs.

Browsers may throttle JavaScript execution in background tabs, but they do not necessarily close active WebSocket connections. As a result, a hidden Headlamp tab continues to receive Kubernetes updates that the user is not currently viewing.

Have the maintainers considered suspending background Kubernetes resource watches for hidden tabs?

Headlamp could use the Page Visibility API (`document.visibilityState` / `visibilitychange`) or the Page Lifecycle API.

A possible lifecycle would be:

1. The browser tab becomes hidden.
2. Headlamp waits for a grace period, for example 30–60 seconds, to avoid unnecessary reconnections during brief tab switches.
3. Headlamp suspends its background Kubernetes resource watches.
4. The Headlamp backend closes the corresponding upstream watch connections to the Kubernetes API.
5. When the tab becomes visible again, Headlamp performs a fresh `list` request.
6. Watches are restored using the latest `resourceVersion`.

A fresh `list` request may be required after resuming because the previous `resourceVersion` may have expired, causing Kubernetes to return `410 Gone`.

Only passive resource watches should be suspended. User-initiated interactive connections should remain active or be handled separately, including:

- Pod terminal / `exec`;
- port forwarding;
- explicitly started log streams;
- other long-running user operations.

A small randomized delay could also be added when resuming, preventing many tabs from generating a sudden burst of requests at the same time.

This behavior could potentially be implemented centrally in the existing WebSocket manager.

This feature would benefit:

- users running Headlamp in-cluster;
- multi-user Headlamp installations;
- users who keep multiple Headlamp tabs open;
- large Kubernetes clusters;
- installations using OIDC and an authentication proxy such as `kube-oidc-proxy`;
- environments where reducing Kubernetes API connections and load is important.

No.

I would first like to get feedback from the maintainers on the following questions:

- Has suspending watches for background tabs already been considered?
- Should this behavior be implemented in the WebSocket manager or in the resource hooks?
- Which connection types should be excluded from suspension?
- Would a configurable grace period be acceptable?
- Should active React Query entries be invalidated and refetched when the tab becomes visible again?

> Do you want to help implement this feature? Please see our [contribution docs](https://headlamp.dev/docs/latest/contributing). For any non-trivial contributions, we advise you to join our [our slack channel](https://kubernetes.slack.com/messages/headlamp) to discuss the implementation with Headlamp's core developers.

## Additional context

Environment where the behavior was observed:

- Installation type: In-Cluster;
- Headlamp version: `v0.44.0`;
- Authentication: OIDC;
- Kubernetes API access: through `kube-oidc-proxy`;
- Authorization: OIDC groups with namespace-scoped Kubernetes RBAC.

The goal is not to share watches between different users, since their tokens and RBAC permissions may differ. The goal is only to close unnecessary resource watches belonging to hidden browser tabs.

SiddharthSingh-iitbhu:
## What this PR does

Pauses background Kubernetes resource watches when the browser tab is hidden.

- Suspends multiplexed WebSocket watches when the tab becomes hidden.
- Resumes watches when the tab becomes visible again.
- Uses a 30-second grace period before suspending watches.
- Handles `410 Gone` errors by invalidating the relevant React Query caches so watches can recover correctly.

## Why

Headlamp continues maintaining Kubernetes resource watches while the browser tab is not visible. Pausing these watches reduces unnecessary browser and Kubernetes API activity when the user is working in another tab.

## Steps to test

1. Open Headlamp with resources that use Kubernetes watches.
2. Switch to another browser tab.
3. Wait for the 30-second grace period.
4. Verify that the background watches are suspended.
5. Return to the Headlamp tab.
6. Verify that the watches resume and resource data continues updating.
7. Run the relevant frontend tests and verify they pass.

Fixes #7139

### DIFF
--- a/frontend/src/lib/k8s/api/v2/multiplexer.ts
+++ b/frontend/src/lib/k8s/api/v2/multiplexer.ts
@@ -49,6 +49,12 @@ export const WebSocketManager = {
   /** Map to track pending unsubscribe operations for debouncing */
   pendingUnsubscribes: new Map<string, NodeJS.Timeout>(),
 
+  /** Flag to track if the WebSocket is intentionally suspended due to tab visibility */
+  suspended: false,
+
+  /** Timer for debouncing suspension when tab is hidden */
+  suspendTimer: null as NodeJS.Timeout | null,
+
   /**
    * Creates a unique key for identifying WebSocket subscriptions
    * @param clusterId - Cluster identifier
@@ -156,6 +162,28 @@ export const WebSocketManager = {
       };
       socket.send(JSON.stringify(requestMsg));
     });
+  },
+
+  /**
+   * Suspend all active watches by closing the WebSocket connection.
+   * Does not clear active subscriptions so they can be resumed later.
+   */
+  suspendAll(): void {
+    if (this.socketMultiplexer && this.socketMultiplexer.readyState === WebSocket.OPEN) {
+      this.socketMultiplexer.close();
+    }
+    this.suspended = true;
+  },
+
+  /**
+   * Resume all active watches by reconnecting the WebSocket.
+   */
+  resumeAll(): void {
+    this.suspended = false;
+    if (this.activeSubscriptions.size > 0) {
+      // Re-connect and resubscribe if there are active subscriptions
+      this.connect().catch(err => console.error('Failed to resume WebSocket subscriptions:', err));
+    }
   },
 
   /**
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for the contribution.


sentoz:
@SiddharthSingh-iitbhu can you fix it?

Issue #7147, opened 2026-08-12
frontend: apply: Unknown fields are silently dropped without fieldValidation=Strict


xonas1101:
### Problem

Applying a resource with a misspelled or unknown field silently drops that field. The edit succeeds, Dry Run shows a clean preview, and the user's change does nothing.

`apply()` sets only `dryRun` as a query param:

https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/api/v1/apply.ts#L80

```ts
const queryParams: QueryParameters = options.dryRun ? { dryRun: 'All' } : {};
```

`fieldValidation` is never sent — no occurrences anywhere in `frontend/src/lib` or `backend/pkg`. The API server therefore uses its default, `Warn`: unknown and duplicate fields are pruned and reported only in response warning headers. Headlamp does not read those headers, so nothing reaches the user. `kubectl` sends `fieldValidation=Strict` for exactly this reason.

### Steps to reproduce

1. Edit a Deployment, change `spec.replicas` to `spec.replics` (or add any unknown field).
2. Click Dry Run — preview looks fine.
3. Apply — succeeds with no warning.
4. Reopen the resource: the field is gone and the intended change never happened.

### Expected

Invalid fields are rejected, or at minimum reported.

### Suggested fix

- Send `fieldValidation=Strict` on apply/create/update, including the dry-run path, so bad fields fail loudly instead of vanishing.
- Alternatively send `Warn` and surface the returned `Warning` headers in the editor/snackbar. Strict matches `kubectl` and is the clearer behavior.

Affects every editing surface, since they all route through `EditorDialog` → `apply()`.

xonas1101:
Hey @SiddharthSingh-iitbhu I have already raised a PR for this issue. Could you please take up some other issue instead of this? I was already assigned to the issue, so please take care of it going forward in open source in general. Thanks!

PR #7215, opened 2026-08-13
frontend: apply: Enforce strict field validation

SiddharthSingh-iitbhu:
## Summary

Enforces strict Kubernetes field validation for resource mutations so unknown or misspelled fields are rejected instead of being silently dropped.

## Related Issue

Related to #7147

## Changes

- Adds `fieldValidation=Strict` to resource mutation requests.
- Applies strict validation to create, update, and patch operations.
- Applies the validation to dry-run requests as well.
- Updates the relevant tests to verify that strict field validation is included.

## Why

Kubernetes can silently drop unknown fields when field validation is not explicitly set to `Strict`. This can cause a user's edit to appear successful even though a misspelled field was ignored.

Using strict validation makes the API server reject invalid fields and gives the user immediate feedback instead of silently losing the change.

## Steps to Test

1. Open a resource editor in Headlamp.
2. Add or rename a field to an invalid/unknown field.
3. Try a dry-run and verify that the request is rejected.
4. Try applying the invalid resource and verify that the request fails with a field validation error.
5. Verify that valid resource create/update/apply operations continue to work.
6. Run the relevant frontend tests and verify they pass.

### DIFF
--- a/frontend/src/lib/k8s/api/v1/apiProxy.test.ts
+++ b/frontend/src/lib/k8s/api/v1/apiProxy.test.ts
@@ -681,6 +681,7 @@ describe('apiProxy', () => {
     it('Successfully creates a new resource with POST', async () => {
       nock(baseApiUrl)
         .post(`/clusters/${clusterName}/api/v1/namespaces/${namespace}/configmaps`)
+        .query({ fieldValidation: 'Strict' })
         .reply(201, mockConfigMap);
 
       const response = await apiProxy.apply(mockConfigMap);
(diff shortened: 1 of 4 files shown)


xonas1101:
Hey @SiddharthSingh-iitbhu I think this is a dupe of my PR #7155 . Could you please close this PR, and look for some other issue to open a PR on. Just as an advice for moving forward in open source - do not open a duplicate PR in general, and dont take up issues in which some other person is assigned, or has already raised a PR. Basic open source etiquette. Thanks.

PR #7223, opened 2026-08-13
fix(frontend): LogsButton: Preserve cluster context for related pods

### LINKED ISSUE #7222: frontend: LogsButton loses cluster context for related pods
## Describe the bug

In a multi-cluster view, `LogsButton` can lose the cluster context of the workload when constructing its related Pods.

`getRelatedPods()` correctly fetches the Pods using the workload's `item.cluster`, but the returned Pods are constructed without passing that cluster context:

```ts
new Pod(podData)
```

When no cluster is provided, the Pod falls back to the current UI cluster. As a result, when viewing a workload from a different cluster, its log requests can be made against the wrong cluster.

### Expected behavior

Related Pods should retain the cluster of the workload they belong to, and `getLogs()` should request logs from that same cluster.

## To Reproduce

Steps to reproduce the bug:

1. Configure Headlamp with at least two Kubernetes clusters.
2. Make one cluster the current UI cluster.
3. Open a workload (Deployment, StatefulSet, DaemonSet, or Job) from another cluster.
4. Open the workload's Logs action.
5. Observe that the related Pod objects do not retain the workload's cluster context, which can cause the log request to target the current UI cluster instead.

## Environment

- Installation type: [e.g. Linux / Windows / Container / In-Cluster]
- Headlamp Version: main (current development version)
- Other: Multi-cluster configuration

Yes (I will propose a PR).

## Additional Context

The issue was confirmed by tracing the flow from `LogsButton` to `Pod.getLogs()`.

The related Pods are fetched from the workload's cluster, but are currently constructed as:

```ts
new Pod(podData)
```

The minimal fix appears to be:

```ts
new Pod(podData, item.cluster)
```

I also verified this with a regression test: the test fails with the current implementation because the Pod loses the cluster context, and passes when the workload's cluster is explicitly provided.

AbhishekMauryaGEEK:
## Summary

Fixes #7222

`LogsButton` correctly fetches related Pods using the workload's cluster, but the cluster context was lost when constructing the `Pod` objects.

Previously, related Pods were created with:

```ts
new Pod(podData)
```

When no cluster was provided, the Pod could fall back to the current UI cluster. In a multi-cluster view, this could cause log requests for a workload from another cluster to target the wrong cluster.

This change preserves the workload's cluster when constructing related Pods:

```ts
new Pod(podData, item.cluster)
```

## Testing

- Added a regression test covering cluster context preservation for related Pods.
- `npx vitest run src/components/common/Resource/LogsButton.test.tsx` — 14/14 passed
- `npm run lint` — passed

### DIFF
--- a/frontend/src/components/common/Resource/LogsButton.test.tsx
+++ b/frontend/src/components/common/Resource/LogsButton.test.tsx
@@ -35,10 +35,12 @@ const {
 } = vi.hoisted(() => {
   class MockKubeObject {
     jsonData: any;
+    _cluster: string;
     static kind = '';
     static apiGroupName = '';
-    constructor(data: any) {
+    constructor(data: any, cluster: string = '') {
       this.jsonData = data;
+      this._cluster = cluster;
     }
     get kind() {
       return this.jsonData?.kind;
@@ -59,7 +61,7 @@ const {
       return this.jsonData?.status;
     }
     get cluster() {
-      return '';
+      return this._cluster;
     }
     getName() {
       return this.jsonData?.metadata?.name ?? '';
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for working on this.


<details>

</details>

AbhishekMauryaGEEK:
Thanks! Rebased the branch onto the latest main and force-pushed the updated history. The PR now contains a single commit with no merge commit. Tests and lint are passing locally.

PR #7239, opened 2026-08-14
app: scripts: Copy plugin locales from tarballs

skoeva:
Plugins installed via `app-build-manifest.json` can ship translations in a
`locales` folder, and Headlamp fetches
`<plugin>/locales/{lang}/translation.json` at runtime. Archive extraction only
copied `main.js` and `package.json`, so translations included in these archives
never reached the installed plugin directory.

Needed for https://github.com/Azure/aks-desktop/issues/515, which localizes the
bundled plugins.

### Changes

- Copy `locales` from extracted plugin archives for both current and legacy
  tarball layouts
- Preserve archive extraction safeguards while recursively copying locale
  directories and regular files
- Add regression coverage for both supported archive layouts

Stale locale files are handled by the existing staging workflow, which builds a
fresh plugin directory and atomically replaces the previous installation.

### Testing

- [x] `cd app && npm test -- electron/build-manifest.test.ts`
- [x] `cd app && npm run tsc`
- [x] Verified locale extraction for both current and legacy tarball layouts
- [x] Plugins without a `locales` directory remain unaffected

### DIFF
--- a/app/electron/build-manifest.test.ts
+++ b/app/electron/build-manifest.test.ts
@@ -816,6 +816,51 @@ describe('plugin archive integrity', () => {
     ).rejects.toThrow();
   });
 
+  it.each([
+    {
+      layout: 'current',
+      archiveRoot: 'plugin',
+      mainPath: 'plugin/main.js',
+      packageJsonPath: 'plugin/package.json',
+      translationPath: 'plugin/locales/fr/translation.json',
+    },
+    {
+      layout: 'legacy',
+      archiveRoot: 'package',
+      mainPath: 'package/dist/main.js',
+      packageJsonPath: 'package/package.json',
+      translationPath: 'package/dist/locales/fr/translation.json',
+    },
+  ])('copies plugin locales from the $layout archive layout', async archiveLayout => {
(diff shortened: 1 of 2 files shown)


app/scripts/setup-plugins.js
```diff
@@ -11,6 +11,19 @@ const MANIFEST_FILE = path.join(__dirname, '../app-build-manifest.json');
 
 const manifest = require(MANIFEST_FILE);
 
+// Headlamp fetches plugin translations from <plugin>/locales at runtime.
+function copyLocales(fromDir, toDir) {
+  const localesDir = path.join(fromDir, 'locales');
+  const destDir = path.join(toDir, 'locales');
+  // Cleared first so locales dropped by a newer archive do not linger.
+  fs.rmSync(destDir, { recursive: true, force: true });
+  if (!fs.existsSync(localesDir)) {
```
sniok: this should probably go before the `rmSync` of the destination

skoeva: addressed by the rebase

illume (approved):
🎉 thanks!

PR #7240, opened 2026-08-14
frontend: useKubeLists: Fix maxItems boundary check and add tests

yashveersinghh:
## Summary
Fixes a boundary condition bug in `useKubeLists` where resources meeting the exact `maxItems` limit were being discarded, and adds unit tests to verify the behavior.

## What Changed
- **Bug Fix (`useKubeLists.ts`):** Updated the item threshold check to `items.length <= maxItems` so items are retained when total count equals `maxItems`.
- **Unit Tests (`useKubeLists.test.ts`):**
  - Added test verifying items are **retained** when `items.length == maxItems`.
  - Added test verifying items are **discarded** (`[]`) when `items.length > maxItems`.

## Testing Steps
Ran Vitest locally against the test file:
```bash
npm run test src/components/advancedSearch/utils/useKubeLists.test.ts


yashveersinghh:
Hi @illume @skoeva, looks like Windows CI failed on a random Chocolatey network timeout (499). 
Could you please re-run it when you get a chance? Thanks!

PR #7249, opened 2026-08-15
frontend: SimpleTable: Fix maximum page calculation bounds

yashveersinghh:
### Description
Fixes an off-by-one boundary condition in `SimpleTable` pagination calculation. 

When `displayData.length` was an exact multiple of `rowsPerPage` (e.g., 20 items with 10 per page), the boundary check `page * rowsPerPage > displayData.length` evaluated `20 > 20` as `false`, allowing page index `2` (the 3rd page, which contains 0 items) to remain active instead of clamping down to max page index `1`.

### Changes Made
- Updated the `useEffect` bounds check to derive `maxPage` using `Math.ceil(displayData.length / rowsPerPage) - 1`.
- Added unit tests in `SimpleTable.test.tsx` verifying that out-of-bounds page requests clamp properly when item counts are exact multiples of `rowsPerPage`.

### Testing
Ran unit tests locally:
`npm run test -- src/components/common/SimpleTable.test.tsx` (Passed)


unlikelyzero (approved):
LGTM. The off-by-one fix is correct for the case it targets — `maxPage` computed via `Math.ceil` and clamped properly for exact-multiple row counts.

One small note, not blocking: the bounds check still derives `maxPage` from the unfiltered `displayData` rather than `filteredData`, so the same exact-multiple edge case can still slip through when a `filterFunction` prop is in play (e.g. 30 source rows, filter yields 20, rowsPerPage=10, page=2 → no clamp, blank page renders). This is a pre-existing gap, not something this PR introduces or needs to fix, but worth a follow-up issue since `filterFunction` is used in a few places (Overview.tsx, endpoints/List.tsx, PluginSettings.tsx, ResourceTable.tsx).

PR #7271, opened 2026-08-15
frontend: ShowHideLabel: fix phantom collapse button on short text

### LINKED ISSUE #7270: ShowHideLabel: Collapse button disappears when initialized expanded with short text
`ShowHideLabel` renders a Collapse button that disappears the moment you click it, when it's initialized already-expanded with text that doesn't actually need truncating.

## Repro

```tsx
<ShowHideLabel show={true} maxChars={256}>Short message</ShowHideLabel>
```

Text is shorter than `maxChars`, so no truncation is ever needed — but a Collapse button still renders, because `needsButton` is computed like this:

```tsx
const [actualText, needsButton] = React.useMemo(() => {
  if (typeof children !== 'string') {
    return ['', false];
  }

  if (expanded) {
    return [children, true];
  }

  return [children.substr(0, maxChars), children.length > maxChars];
}, [children, expanded, maxChars]);
```

When `expanded` is `true` (from `show={true}`), `needsButton` is unconditionally `true`, regardless of whether the text is actually long enough to need a toggle. Clicking the rendered Collapse button flips `expanded` to `false`, which recomputes `needsButton` as `children.length > maxChars` — `false` for short text — so the button vanishes entirely rather than toggling back to an Expand button. The control disappears rather than behaving like a toggle.

## Fix

`needsButton` should depend only on whether the text actually exceeds `maxChars`, independent of `expanded`:

```tsx
const isLong = typeof children === 'string' && children.length > maxChars;
const actualText = expanded || !isLong ? children : children.slice(0, maxChars);
return [actualText, isLong];
```

## Where I found this

Ran into this while reviewing #7267, which touches this same component's DOM structure for an aria-attribute fix but doesn't touch this logic.

unlikelyzero:
## Summary

Fixes a phantom collapse button in `ShowHideLabel`: initializing the component with `show={true}` on text shorter than `maxChars` rendered a Collapse button that never should have appeared, and clicking it made the button vanish entirely instead of toggling back to Expand.

Fixes #7270

## Root cause

`needsButton` was unconditionally `true` whenever `expanded` was `true`:

```tsx
if (expanded) {
  return [children, true];
}
```

So `<ShowHideLabel show={true}>Short text</ShowHideLabel>` rendered a button even though the text never needed truncating. Clicking it flipped `expanded` to `false`, which recomputed `needsButton` as `children.length > maxChars` (false for short text), so the button disappeared instead of behaving like a toggle.

## Fix

`needsButton` is now derived solely from whether the text actually exceeds `maxChars`, independent of `expanded`.

## Testing

- Added a unit test file (`ShowHideLabel.test.tsx`) covering: no button for short text, no button for short text initialized expanded (the regression case), toggle behavior for long text, and toggle staying visible for long text initialized expanded.
- Added a new `ExpandedShortText` story/storyshot to cover this case visually.
- Verified all existing ShowHideLabel snapshots are unaffected.

### DIFF
--- a/frontend/src/components/common/ShowHideLabel/ShowHideLabel.stories.tsx
+++ b/frontend/src/components/common/ShowHideLabel/ShowHideLabel.stories.tsx
@@ -56,3 +56,10 @@ ShortText.args = {
   children: 'Short text',
   labelId: 'my-label2',
 };
+
+export const ExpandedShortText = Template.bind({});
+ExpandedShortText.args = {
+  children: 'Short text',
+  show: true,
+  labelId: 'my-label3',
+};
--- a/frontend/src/components/common/ShowHideLabel/ShowHideLabel.test.tsx
+++ b/frontend/src/components/common/ShowHideLabel/ShowHideLabel.test.tsx
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import ShowHideLabel from './ShowHideLabel';
+
+describe('ShowHideLabel', () => {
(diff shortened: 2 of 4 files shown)


Alok-work23 (approved):
The needsButton logic now correctly depends on whether the text exceeds maxChars, including the initially expanded case.

The regression tests cover short text, expanded short text, expanding/collapsing long text, and long text initialized with show.

No blocking issues from my side. Thanks for adding the regression coverage!

illume (changes requested):
Thanks for this PR.


<details>

- `frontend: ShowHideLabel: fix phantom collapse button on short text` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

illume (changes requested):
Thanks for this PR.


unlikelyzero:
Commit title amended to `frontend: ShowHideLabel: Fix phantom collapse button on short text` (capitalized per commit guidelines) and force-pushed. @illume

PR #7287, opened 2026-08-16
0049: frontend: Support distribution product versions

illume:
## Summary

Add an optional distribution product version for branded Headlamp builds while
retaining the underlying Headlamp version and Git commit in the version dialog.
The top-bar menu uses the distribution version when configured and keeps the
existing Headlamp version fallback otherwise.

This is independent and does not depend on another Headlamp PR.

## Changes

- read `REACT_APP_HEADLAMP_PRODUCT_VERSION` through documented product-info
  helpers
- show distinct product and Headlamp versions only when they differ
- preserve the existing single-version dialog when the product version is
  absent or empty
- translate complete product-version labels with product-name interpolation
- widen the dialog so product labels and commit identifiers remain readable
- add focused unit and end-to-end coverage before the implementation commit

The downstream patch covered only the configured helper value. This version
adds tests for configured, absent, empty, matching, distinct, and unnamed
products so fallback behavior is explicit and changed decision modules reach
100% branch coverage. It also centralizes the top-bar fallback in
`getDisplayVersion`, translates labels without assuming English word order,
extracts the new keys across all 25 locale catalogs, adds TSDoc for the new
TypeScript API, verifies the user-visible dialog path in the e2e suite, and
widens the dialog because the added labels otherwise broke in the middle of
words.

## Source history

- Azure source PR: https://github.com/Azure/aks-desktop/pull/823
- Original About-dialog commit:
  https://github.com/Azure/aks-desktop/commit/d68693b049faa97b93dd0396103ed4fdd4cf2be4
- Original top-bar commit:
  https://github.com/Azure/aks-desktop/commit/067bf68f3e31574e958a53a3fb3ae98e7206256d
- The existing Headlamp version dialog originated in:
  https://github.com/kubernetes-sigs/headlamp/pull/640

The two original Azure commits are in rebased downstream history without an
originating Azure merge PR association, so unrelated PRs that later inherited
the commits are intentionally not listed.

## Testing

- `npm test -- --run` (218 files, 2,871 tests passed)
- focused Istanbul coverage: 100% branches for `getProductInfo.ts` and
  `VersionDialog.tsx`
- `npm run tsc`
- `npm run build`
- focused ESLint and Prettier checks
- `npx playwright test tests/headlamp.spec.ts --list` (14 tests discovered,
  including the new version-dialog test)
- Storybook before/after verification at 960 x 640

The Playwright test was compiled and discovered locally but not executed because
no Headlamp server was running and the local `test` Kubernetes API endpoint was
unavailable. CI provides the required e2e environment.

### Before

Headlamp version dialog before product version configuration

### After: configured distribution version

AKS Desktop version dialog with product and Headlamp versions

Assisted by copilot

### DIFF
--- a/frontend/src/components/App/VersionDialog.tsx
+++ b/frontend/src/components/App/VersionDialog.tsx
@@ -17,7 +17,7 @@
 import DialogContent from '@mui/material/DialogContent';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
-import { getProductName, getVersion } from '../../helpers/getProductInfo';
+import { getProductName, getProductVersion, getVersion } from '../../helpers/getProductInfo';
 import { useTypedSelector } from '../../redux/hooks';
 import { uiSlice } from '../../redux/uiSlice';
 import { Dialog } from '../common/Dialog';
@@ -33,20 +33,18 @@ interface VersionInformation {
   GIT_VERSION: string;
 }
 
-/** Properties accepted by the Version dialog. */
-interface VersionDialogProps {
-  /** Optional version provider used by tests and embedding hosts. */
-  getVersion?: () => {
-    VERSION: string;
-    GIT_VERSION: string;
-  };
+export interface VersionDialogProps {
+  /** Overrides Headlamp version lookup, primarily for tests and stories. */
+  getVersion?: () => VersionInformation;
+  /** Overrides product distribution version lookup, primarily for tests and stories. */
+  getProductVersion?: typeof getProductVersion;
 }
 
 /**
- * Displays application version information and host-provided legal documents.
+ * Displays product, Headlamp, Git, and host-provided legal information.
  *
  * @param props - Version dialog dependencies.
- * @returns The application Version dialog.
+ * @returns The version information dialog.
  */
 export default function VersionDialog(props: VersionDialogProps) {
   const open = useTypedSelector(state => state.ui.isVersionDialogOpen);
@@ -55,28 +53,45 @@ export default function VersionDialog(props: VersionDialogProps) {
   const { VERSION, GIT_VERSION } = (
     props.getVersion ? props.getVersion() : getVersion()
   ) as VersionInformation;
-  const versionInformation = (
-    <NameValueTable
-      rows={[
-        {
-          name: t('translation|Version'),
-          value: VERSION,
-        },
-        {
-          name: t('Git Commit'),
-          value: GIT_VERSION,
-        },
-      ]}
-    />
-  );
+  const productVersion = (
+    props.getProductVersion ? props.getProductVersion() : getProductVersion()
+  )?.trim();
+  const showProductVersion = productVersion && productVersion !== VERSION;
+  const defaultProductName = t('translation|Product');
+  const productName = getProductName();
+  const productVersionName =
+    productName && productName !== 'Headlamp' ? productName : defaultProductName;
+  const rows = [
+    ...(showProductVersion
+      ? [
+          {
+            name: t('translation|{{ productName }} Version', {
+              productName: productVersionName,
+            }),
+            value: productVersion,
+          },
+        ]
+      : []),
+    {
+      name: showProductVersion
+        ? t('translation|{{ productName }} Version', { productName: 'Headlamp' })
+        : t('translation|Version'),
+      value: VERSION,
+    },
+    {
+      name: t('Git Commit'),
+      value: GIT_VERSION,
+    },
+  ];
+  const versionInformation = <NameValueTable rows={rows} />;
   const hasLegalDocuments = Boolean(window.desktopApi?.getLegalDocuments);
 
   return (
     <Dialog
-      maxWidth="sm"
+      maxWidth="md"
       open={open}
       onClose={() => dispatch(uiSlice.actions.setVersionDialogOpen(false))}
-      title={getProductName()}
+      title={productName}
       aria-label={t('translation|Version information')}
       // We want the dialog to show on top of the cluster chooser one if needed
       style={{ zIndex: 1900 }}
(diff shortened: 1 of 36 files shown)


frontend/src/helpers/getProductInfo.ts
```diff
@@ -29,3 +29,17 @@ export function getVersion() {
...
+ */
+export function getProductVersion(): string | undefined {
+  return import.meta.env.REACT_APP_HEADLAMP_PRODUCT_VERSION;
+}
+
+/**
+ * @returns The product distribution version when configured, otherwise the Headlamp version.
+ */
+export function getDisplayVersion(): string | undefined {
+  return getProductVersion() ?? getVersion().VERSION;
```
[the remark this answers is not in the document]
illume: Addressed in 7eac261d4. `getDisplayVersion()` now uses the same truthy fallback as `VersionDialog`, so an empty product version falls back to the Headlamp version. Added an explicit empty-value regression test in `getProductInfo.test.ts`.

frontend/src/components/App/VersionDialog.tsx
```diff
@@ -17,26 +17,54 @@
...
   const dispatch = useDispatch();
   const { t } = useTranslation(['glossary', 'translation']);
   const { VERSION, GIT_VERSION } = props.getVersion ? props.getVersion() : getVersion();
+  const productVersion = props.getProductVersion ? props.getProductVersion() : getProductVersion();
+  const showProductVersion = productVersion && productVersion !== VERSION;
+  const rows = [
+    ...(showProductVersion
+      ? [
+          {
+            name: `${getProductName() || 'Product'} ${t('translation|Version')}`,
```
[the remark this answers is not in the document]
illume: Addressed in 7eac261d4. The dialog now translates the complete `{{ productName }} Version` label with interpolation and separately translates the generic `Product` fallback. Both keys were extracted into all 18 locale catalogs, and the unit test asserts the translation key and interpolation values.

unlikelyzero (changes requested):
Some things I observed:

1. When the product name stays at its default "Headlamp" but a distinct product version is configured (the common case — only the version is customized), `VersionDialog.tsx` renders two adjacent rows both labeled "Headlamp Version" with different values: the first row's name comes from `getProductName() || defaultProductName`, the second row's name is hardcoded to `'Headlamp'`, and both go through the same `'{{ productName }} Version'` template. `NameValueTable` keys rows by array index, so there's no console warning to catch this. `TopBar.tsx:414` has the same problem one component over — it renders `{getProductName()} {getDisplayVersion()}`, so the top-bar menu item labels the product version as plain "Headlamp" too. This is the default configuration a distributor would hit, not an edge case.

2. `make-env.js` regenerates `frontend/.env` from a fixed template on every `prebuild`/`prestart`/`pretest` and never includes or preserves `REACT_APP_HEADLAMP_PRODUCT_VERSION`. The natural way to configure it — adding it to `frontend/.env` — gets silently erased at build time. Only exporting it as a process env var before the build actually works, and that's not documented anywhere in the PR.

3. `getProductInfo.ts` doesn't trim the product version read from the env var, so a whitespace-only or padded value is treated as truthy/configured. That bypasses both the empty-string fallback fix and the `productVersion !== VERSION` deduplication check — same bug class as the empty-string issue already flagged and fixed, just not fully closed.

frontend/src/helpers/getProductInfo.ts
```diff
@@ -30,6 +30,20 @@ export function getProductName(): string | undefined {
   return import.meta.env.REACT_APP_HEADLAMP_PRODUCT_NAME;
 }
 
+/**
+ * @returns The version of the product distribution, or undefined when it is not set.
+ */
+export function getProductVersion(): string | undefined {
+  return import.meta.env.REACT_APP_HEADLAMP_PRODUCT_VERSION;
```
skoeva, line 37: `make-env.js` rewrites `.env` without this key, so `.env` config is lost. maybe this can be wired into the generated envi

frontend/src/components/App/VersionDialog.tsx
```diff
@@ -55,25 +53,37 @@ export default function VersionDialog(props: VersionDialogProps) {
...
-  );
+  const productVersion = props.getProductVersion ? props.getProductVersion() : getProductVersion();
+  const showProductVersion = productVersion && productVersion !== VERSION;
+  const defaultProductName = t('translation|Product');
+  const rows = [
+    ...(showProductVersion
+      ? [
+          {
+            name: t('translation|{{ productName }} Version', {
+              productName: getProductName() || defaultProductName,
```
skoeva: with the default product name, both version rows have the same label but different values - these could be distinguished

frontend/src/helpers/getProductInfo.ts
```diff
@@ -30,6 +30,20 @@ export function getProductName(): string | undefined {
...
+}
+
+/**
+ * @returns The product distribution version when configured, otherwise the Headlamp version.
+ */
+export function getDisplayVersion(): string | undefined {
+  return getProductVersion() || getVersion().VERSION;
```
skoeva: whitespace-only values bypass this fallback. maybe the product version could be trimmed and an empty string could represent an unset value?

illume:
Thanks @skoeva 
- rebased on main
- Updated VersionDialog.tsx so distinct versions with the default product name are labeled (Product Version, Headlamp Version).

illume:
getProductVersion now takes white space as unset.
>return getProductVersion()?.trim() || getVersion().VERSION;

frontend/src/helpers/getProductInfo.ts
illume: Can you please expand on this? I don't understand.

ashu8912 (approved):
LGTM!!!

PR #7295, opened 2026-08-16
app: runCmd: Do not run the command when consent is denied

### LINKED ISSUE #6558: app: runCommand consent dialog runs the command once even when the user clicks Deny
## Describe the bug

In app (Electron) mode, the run-command consent dialog does not block the command the first time you deny it. When a plugin calls `runCommand` for a command that hasn't been consented to yet, Headlamp shows the native "Allow / Deny" dialog. If you click **Deny**, your choice is saved correctly, but the command still runs that one time. The denial only takes effect from the *next* invocation onward.

The dialog message says "Allow this local command to be executed? Your choice will be saved.", so clicking Deny reasonably implies the command will not run. It does.

The problem is in `checkCommandConsent` in [`app/electron/runCmd.ts`](https://github.com/kubernetes-sigs/headlamp/blob/main/app/electron/runCmd.ts#L78-L104). The first-time branch saves the user's answer but the function falls through to an unconditional `return true`:

```ts
if (savedCommand === false) {
  console.error(`Invalid command: ${consentKey}, command not allowed by users choice`);
  return false;
} else if (savedCommand === undefined) {
  const commandChoice = confirmCommandDialog(consentKey, mainWindow);   // false when user clicks Deny
  if (settings?.confirmedCommands === undefined) {
    settings.confirmedCommands = {};
  }
  settings.confirmedCommands[consentKey] = commandChoice;               // saved as false
  saveSettings(SETTINGS_PATH, settings);
}
return true;                                                            // ...but returns true anyway
```

`commandChoice` is discarded. On the second run the saved `false` is read back and the command is correctly blocked, so the effect is that a denial is honoured exactly one run late.

This affects every consent-gated command in app mode (the minikube management scripts, `az`, `gh`), not a single plugin.

### Flow

```mermaid
sequenceDiagram
    participant Plugin
    participant Main as Electron main<br/>(checkCommandConsent)
    participant User
    participant Shell as spawn()

    Note over Main: first request — confirmedCommands['minikube delete'] is undefined
    Plugin->>Main: runCommand('minikube', ['delete'])
    Main->>User: consent dialog (Allow / Deny)
    User-->>Main: clicks "Deny"
    Main->>Main: save confirmedCommands['minikube delete'] = false
    Main->>Shell: return true — command runs anyway
    Shell-->>Plugin: output

    Note over Main: second request — confirmedCommands['minikube delete'] === false
    Plugin->>Main: runCommand('minikube', ['delete'])
    Main--xShell: return false — blocked (one run too late)
```

| Invocation | Saved state before | User action | Returns | Command runs? |
|-----------|--------------------|-------------|---------|---------------|
| 1st | `undefined` | clicks **Deny** | `true` | **yes (wrong)** |
| 2nd | `false` | (no prompt) | `false` | no |
| 1st | `undefined` | clicks **Allow** | `true` | yes |

## To Reproduce
Steps to reproduce the bug:
1. Run the desktop (Electron) app and install a plugin that uses `runCommand` (e.g. the minikube plugin).
2. Trigger a command that has not been consented to yet, so the "Consent to command being run" dialog appears.
3. Click **Deny**.
4. The command runs this time regardless. Trigger the same command again and it is now blocked — confirming the denial was saved but applied one run late.

 - Installation type: Desktop app (Electron) — reproduced from source on Windows; the code path is platform-independent so it applies to all desktop builds.
 - Headlamp Version: 0.43.0 (current `main`)
 - Other: app mode only; this path does not exist for the in-cluster/web build.

Yes — I plan to propose a PR. The first-time branch should return the choice the user just made (a denial has to block the current run, not only later ones) instead of falling through to the unconditional `return true`.

## Additional Context
Not a regression from a specific release — the unconditional `return true` has been present since the consent mechanism was added in #1744, so every version shipping that dialog is affected.

utkarshalpha:
## Summary

In the desktop app, denying a plugin's `runCommand` request does not stop the command the first time. The consent dialog says "Allow this local command to be executed? Your choice will be saved.", but clicking **Deny** still lets the command run once. The denial only takes effect from the next invocation onwards.

`checkCommandConsent` in `app/electron/runCmd.ts` asks for consent, saves the answer, and then falls through to an unconditional `return true`, so the answer is ignored for that call.

## Related Issue

Fixes #6558

## Changes

- `app/electron/runCmd.ts`: return the answer the user actually gave instead of always returning `true`, and log the refusal the same way the already-denied branch above it does.
- `app/electron/runCmd.test.ts`: add tests for the consent path, which had no coverage. They go through the exported `handleRunCommand`, so they check what a user sees rather than a private function.

## Steps to Test

1. `npm run app:test:unit`
2. The two new tests are under `command consent`: denying the dialog means `spawn` is never called, allowing it means it is.

Without the fix, `does not run the command when the user denies consent` fails because the command is spawned anyway.

## Notes for the Reviewer

The tests need `vi.mock('electron', ...)` because this is the first test in the file that reaches `dialog.showMessageBoxSync`. Outside the Electron runtime `require('electron')` returns the path to the binary rather than the API, so `dialog` is undefined and cannot be called without mocking it.

Nothing outside the consent path changes. Commands that were already allowed or already denied behave exactly as before.

This PR was written in part with the assistance of generative AI. I have reviewed and tested the change myself.

### DIFF
--- a/app/electron/runCmd.test.ts
+++ b/app/electron/runCmd.test.ts
@@ -18,9 +18,15 @@ import { EventEmitter } from 'events';
 import path from 'path';
 import { afterEach, beforeEach, describe, expect, it, Mock, vi } from 'vitest';
 
-const { getShellEnvironmentMock, spawnMock } = vi.hoisted(() => ({
+const { getShellEnvironmentMock, spawnMock, showMessageBoxSyncMock } = vi.hoisted(() => ({
   getShellEnvironmentMock: vi.fn(),
   spawnMock: vi.fn(),
+  showMessageBoxSyncMock: vi.fn(),
+}));
+
+vi.mock('electron', () => ({
+  BrowserWindow: class {},
+  dialog: { showMessageBoxSync: showMessageBoxSyncMock },
 }));
 
 vi.mock('child_process', () => ({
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

utkarshalpha:
> 🎉 thanks!

Thanks for merging @illume 
Since looking for more contributions in headlamp, i’ll keep an eye out for more ways to contribute.

PR #7304, opened 2026-08-17
frontend: CronJob: Spawn Job on the CronJob's cluster

### LINKED ISSUE #7292: CronJob "Spawn Job" can create the Job on the wrong cluster
## Describe the bug

Spawning a Job from a CronJob does not bind the request to the cluster that owns the CronJob. The Job can be created on a different cluster than the one being viewed. Because the spawned Job carries an `ownerReferences` entry pointing at the CronJob's UID — which does not exist on the other cluster — Kubernetes garbage collects it moments later. The user sees a "Job … spawned" success message, a pod briefly runs on a cluster they were only viewing, and afterwards no Job exists on either cluster.

`SpawnJobDialog` dispatches `clusterAction(() => apply(job), …)` without passing a cluster, so `apply()` falls back to `getCluster()`, which reads the cluster from the *current URL*. That callback is not invoked at click time: `executeClusterAction` awaits `CLUSTER_ACTION_GRACE_PERIOD` (5s, the "Cancel" undo window) before running it. By then the URL may point at a different cluster, and the Job is created there.

Expected: the Job is always created on the cluster that owns the CronJob.

## To Reproduce

Prerequisites: two clusters in your kubeconfig (below: `alpha` and `beta`), and a CronJob in the `default` namespace of `alpha`. (`default` exists on both, so the create succeeds on the wrong cluster instead of erroring — which makes the bug visible rather than just noisy.)

1. Open Headlamp and select cluster `alpha`.
2. Go to Workloads → CronJobs and open your CronJob.
3. Click the "Spawn Job" action button (lightning bolt icon).
4. In the dialog, click "Spawn". A "Spawning Job …" snackbar appears with a Cancel button — this is the 5 second grace period before the request is sent.
5. **Within those 5 seconds**, switch to cluster `beta` using the cluster chooser.
6. Wait for the "Job … spawned" success snackbar.
7. Check both clusters:
   - `kubectl --context alpha -n default get jobs` → the Job is **not** there
   - `kubectl --context beta -n default get jobs` → the Job **is** there

Video Demo:

https://github.com/user-attachments/assets/c2556c6a-58a3-4621-934a-7427091692dd

Yes (I will propose a PR).

## Additional Context

Every other `clusterAction` callsite already binds its cluster correctly, either by operating on a `KubeObject` instance (`item.update(…)`, `cronJob.patch(…)`), by passing an explicit cluster (`EditorDialog`), or by capturing it at click time before dispatching — `CreateNamespaceButton` does exactly this with `const clusterData =getCluster();` outside the callback. `SpawnJobDialog` is the only one calling a bare API function with no cluster binding.

The same root cause has a second trigger that needs no timing at all. In the resource map, `KubeObjectDetails` renders `<CronJobDetails … cluster={resource.cluster} />` while the URL stays on the map, and the map aggregates across every selected cluster. `getCluster()` returns only the *first* of `/c/alpha+beta`, so acting on a CronJob from `beta` targets `alpha` deterministically.

Relatedly, `CronJobDetails` accepts a `cluster` prop and forwards it to `DetailsGrid`, but its own `CronJob.useGet(name, namespace)` call drops it, so the object backing the action buttons is fetched from the URL cluster rather than the prop cluster.

This does not look like a regression; the missing cluster argument appears to predate multi-cluster selection.

Confirmed on two local minikube clusters. Events on the *unintended* cluster show `SuccessfulCreate` for the Job and `Scheduled` for its pod, while the intended cluster shows no activity. The Job is absent from `kubectl get jobs` on both clusters afterwards, because of the dangling owner reference described above.

YousufFFFF:
## Summary

This PR fixes the CronJob "Spawn Job" action creating the Job on the wrong cluster. `SpawnJobDialog` did not bind the request to the cluster owning the CronJob, so `apply()` fell back to `getCluster()` — which reads the cluster from the current URL, and is only evaluated after `clusterAction`'s 5 second grace period has elapsed.

## Related Issue

Fixes #7292 

## Changes

- `SpawnJobDialog`: resolve the cluster from `cronJob.cluster` at click time and pass it to `apply()`, instead of letting `apply()` fall       back to the URL cluster after the grace period
- `CronJobDetails`: pass the existing `cluster` prop through to `CronJob.useGet()`, so the object backing the action buttons comes from the same cluster already forwarded to `DetailsGrid`
- Added `Details.test.tsx` covering both, verified to fail before the fix

# After Video:

https://github.com/user-attachments/assets/3e8f231c-ba60-4184-8000-cdb5b6dd8580

## Steps to Test

1. Add two clusters to your kubeconfig, and create a CronJob on the first one.
2. Open that CronJob's details page and click "Spawn Job", then "Spawn".
3. While the "Spawning Job …" snackbar with the Cancel button is still showing, switch to the second cluster.
4. Wait for the success snackbar, then check both clusters:
   `kubectl --context <first> -n <ns> get jobs` should show the Job, and
   `kubectl --context <second> -n <ns> get jobs` should be empty.

Before this change the Job was created on the second cluster, and was then garbage collected because its owner reference pointed at a CronJob UID that does not exist there — leaving no Job on either cluster.

## Notes for the Reviewer

Every other `clusterAction` callsite already binds its cluster, so this brings the one outlier in line rather than introducing a new pattern. `CreateNamespaceButton` is the closest precedent for capturing the cluster before dispatch.

The same root cause reaches the resource map, where `KubeObjectDetails` renders details panels with an explicit `cluster` while the URL stays on the map and `getCluster()` returns only the first of `/c/a+b`. That is why the `useGet` change is included here.

### DIFF
--- a/frontend/src/components/cronjob/Details.test.tsx
+++ b/frontend/src/components/cronjob/Details.test.tsx
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { fireEvent, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { TestContext } from '../../test';
+import CronJobDetails from './Details';
+
+const OWNER_CLUSTER = 'owner-cluster';
+
+const { mockApply, capturedAction, mockUseGet } = vi.hoisted(() => ({
+  mockApply: vi.fn().mockResolvedValue({}),
+  capturedAction: { current: null as null | (() => Promise<void>) },
+  mockUseGet: vi.fn(),
+}));
+
+vi.mock('../../lib/k8s/api/v1/apply', () => ({ apply: mockApply }));
+
+vi.mock('../../lib/k8s/cronJob', () => ({
+  default: { kind: 'CronJob', useGet: mockUseGet },
+}));
+
+vi.mock('../../lib/k8s/job', () => ({
+  default: {
+    kind: 'Job',
+    useList: () => ({ items: [], errors: null, isLoading: false }),
+  },
+}));
+
+vi.mock('../../redux/clusterActionSlice', async importOriginal => ({
+  ...(await importOriginal<typeof import('../../redux/clusterActionSlice')>()),
+  clusterAction: vi.fn((action: () => Promise<void>) => {
+    capturedAction.current = action;
+    return { type: 'clusterAction/test' };
+  }),
+}));
+
+vi.mock('react-i18next', async importOriginal => ({
+  ...(await importOriginal<typeof import('react-i18next')>()),
+  useTranslation: () => ({ t: (key: string) => key, i18n: { language: 'en' } }),
+}));
+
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for this PR.


YousufFFFF:
@illume Can you please run the copilot test again?
Thanks!

PR #6540, opened 2026-07-16
Fix WebSocket auth fallback for stale user context keys

rishh09:
Fixes #6539

This updates WebSocket cluster API requests so that when a user-specific/stateless context key is missing, Headlamp falls back to the normal cluster context instead of returning `404 key not found`.

The existing cookie/Authorization handling is left unchanged, so a valid auth cookie can still be promoted to the proxied request after fallback.

Added regression coverage for a stale `base64url.headlamp.authorization.k8s.io...` WebSocket subprotocol with a valid auth cookie.

Tests:
- go test ./cmd
- go test ./...

PR #7319, opened 2026-08-18
backend: cluster: Fix WebSocket auth fallback

### LINKED ISSUE #6539: WebSocket proxy returns 404 "key not found" instead of falling back to cookie auth when base64url.headlamp.authorization.k8s.io.<userId> can't be resolved
## Summary

When a WebSocket request carries the `base64url.headlamp.authorization.k8s.io.<userId>` subprotocol and that `<userId>` is not present in the backend token store, the server responds `404 Not Found` with body `key not found` and does **not** fall back to the auth cookie, even when a valid `headlamp-auth-<cluster>.<n>` cookie is sent on the same request. This breaks any binary WebSocket proxy (e.g. the KubeVirt plugin VNC/serial console) in in-cluster OIDC/cookie deployments.

## Environment

- Headlamp `v0.43.0`, in-cluster, OIDC auth, 2 replicas.
- Consumer: `naval-group/headlamp-kubevirt` v0.2.2 (VNC/serial console), which appends `base64url.headlamp.authorization.k8s.io.<userId>` (userId = `localStorage["headlamp-userId"]`) to the WS subprotocols.

## Reproduction

Proxying a KubeVirt VNC subresource WS through `/clusters/main/apis/subresources.kubevirt.io/v1/namespaces/<ns>/virtualmachineinstances/<vm>/vnc`:

| Request (same valid auth cookie) | Result |
| --- | --- |
| Subprotocols: `base64.binary.k8s.io, plain.kubevirt.io` | `101 Switching Protocols` (works) |
| Subprotocols: `... , base64url.headlamp.authorization.k8s.io.<unresolved-userId>` | `404 Not Found` / `key not found` |

The `<userId>` token store is never populated in this deployment: neither `POST /clusters/<c>/set-token` (+ `X-HEADLAMP-USER-ID` header), nor authenticated REST calls, nor the `/wsMultiplexer` connection register a `userId -> token` entry (the multiplexer echoes back `userId:""`). So the subprotocol lookup always misses.

## Expected

When the `base64url.headlamp.authorization.k8s.io.<userId>` lookup misses, the backend should fall back to the auth cookie (or `Authorization` header) rather than returning `404 key not found`. The cookie path already authenticates the exact same request successfully.

## Impact

Any plugin using a raw WebSocket that includes the userId subprotocol (KubeVirt console, terminals, port-forward) fails in multi-replica / cookie-based OIDC deployments.

gambtho:
Fixes #6539

Supersedes #6540 by @rishh09, which has been awaiting a rebase since 2026-08-13. The original commit is carried over unchanged — same author, same `Signed-off-by` — so authorship and credit remain with @rishh09. No code was modified; this PR only rebases and tidies the history that reviewers asked for on #6540.

## What this changes

WebSocket cluster API requests derive a user-specific context key from the `base64url.headlamp.authorization.k8s.io.<userId>` subprotocol. In in-cluster OIDC/cookie deployments that `userId` is never registered in the token store, so the lookup misses with `cache.ErrNotFound` and the request fails with `404 key not found` — even when the same request carries a valid `headlamp-auth-<cluster>.<n>` cookie that would authenticate it. This breaks any plugin using a raw WebSocket that includes the userId subprotocol (KubeVirt VNC/serial console, terminals, port-forward).

`getContextWithWebSocketFallback()` falls back to the route's `{clusterName}` context when a WebSocket request references a missing user-specific context, so the existing cookie and `Authorization` handling can still promote credentials onto the proxied request.

The fallback is deliberately fenced to three conditions, so no other lookup failure is masked:

- the request is a WebSocket upgrade,
- the derived context key differs from the cluster name,
- the error is `cache.ErrNotFound` specifically.

A stateless/dynamic-kubeconfig context that exists but fails for any other reason is therefore never silently downgraded to the shared cluster context.

## What changed relative to #6540

- Rebased onto current `main` (the branch was 457 commits behind; the rebase was conflict-free).
- Squashed the docs-only follow-up commit into the implementation commit, as requested in review, and wrote the stale-context/404 rationale into the commit body.

The resulting diff is byte-identical to the diff on #6540.

## Testing

- `go build ./...` — clean
- `go test ./...` — exit 0, 15 packages ok, no failures
- `go test ./cmd/ -run TestClusterRequestHandlerFallsBackToClusterContextForWebSocketCookie -v` — PASS
- `go fmt ./cmd/ ./pkg/**` — no changes
- `./tools/golangci-lint run` — 0 issues

Regression coverage is the test added by @rishh09: a stale `base64url.headlamp.authorization.k8s.io...` WebSocket subprotocol combined with a valid auth cookie now proxies successfully and promotes the cookie to `Authorization`.

### DIFF
--- a/backend/cmd/headlamp.go
+++ b/backend/cmd/headlamp.go
@@ -1232,6 +1232,30 @@ func (c *HeadlampConfig) shouldUseUnsafeServiceAccountTokenForContext(kContext *
 	return c.shouldUseUnsafeServiceAccountToken() && kContext.UsesInClusterServiceAccountToken()
 }
 
+// getContextWithWebSocketFallback returns the requested context, falling back to the cluster
+// context when a WebSocket request references a missing user-specific context.
+func (c *HeadlampConfig) getContextWithWebSocketFallback(
+	r *http.Request,
+	contextKey string,
+) (string, *kubeconfig.Context, error) {
+	kContext, err := c.KubeConfigStore.GetContext(contextKey)
+	if err == nil {
+		return contextKey, kContext, nil
+	}
+
+	clusterName := mux.Vars(r)["clusterName"]
+	if strings.EqualFold(r.Header.Get("Upgrade"), "websocket") &&
+		contextKey != clusterName &&
+		errors.Is(err, cache.ErrNotFound) {
+		kContext, fallbackErr := c.KubeConfigStore.GetContext(clusterName)
+		if fallbackErr == nil {
+			return clusterName, kContext, nil
+		}
+	}
+
+	return contextKey, nil, err
+}
+
 func tokenFromCookie(r *http.Request, clusterName string) string {
 	if cookieToken, err := auth.GetTokenFromCookie(r, clusterName); err == nil && cookieToken != "" {
 		return cookieToken
@@ -1858,7 +1882,7 @@ func clusterRequestHandler(c *HeadlampConfig) http.Handler { //nolint:funlen
 			return
 		}
 
-		kContext, err := c.KubeConfigStore.GetContext(contextKey)
+		contextKey, kContext, err := c.getContextWithWebSocketFallback(r, contextKey)
 		if err != nil {
 			c.handleError(w, ctx, span, err, "failed to get context", http.StatusNotFound)
 			return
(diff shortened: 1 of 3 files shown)


illume (approved):
🎉 thanks!

PR #7324, opened 2026-08-18
frontend: Pod: Split log stream chunks into lines

### LINKED ISSUE #7137: Log Severity filter not working properly
## Describe the bug

The new log severity filtering https://github.com/kubernetes-sigs/headlamp/issues/5323 does not classify log lines correctly.  

## To Reproduce
Steps to reproduce the bug:
1. Start a debug container and use echo to produce logs, e.g.
    ```bash
    ~ $ echo '{"level":"info","message":"test0"}'
    {"level":"info","message":"test0"}
    ~ $ echo '{"level":"warn","message":"test1"}'
    {"level":"warn","message":"test1"}
    ~ $ echo '{"level":"error","message":"test2"}'
    {"level":"error","message":"test2"}
    ~ $ echo '{"level":"warn","message":"test3"}'
    {"level":"warn","message":"test3"}
    ~ $ echo '{"level":"debug","message":"test4"}'
    {"level":"debug","message":"test4"}
    ```
2. Go to headlamp and check the logs of the container (prettify off, format off)
  * Severity = All shows all lines. correctly.
    Image
  * Filtering by severity doesn't work well. Some log entries are classified wrong.

    Image

    Image

    Image

    Image
  * Adding more log lines increases the issues. It looks like the log lines are classified as batches.
    
    Image

    Image

 - Installation type: Container-Image/In-Cluster/Helm
 - Headlamp Version: 0.44.0
 - Other: read-only root filesystem

No.

## Additional Context
none

Anjali-Chauhan1:
## Summary

This PR fixes the log severity filter selecting the wrong lines, by making each log line its own entry instead of letting one stream chunk hold several lines that were all judged by the severity of the first one. The log stream arrives chunked on arbitrary byte boundaries rather than on line ends, so one chunk could carry several lines and one line could span two chunks. That is why the misclassification got worse the more log lines there were.

## Related Issue

Fixes #7137

## Changes

- Updated `frontend/src/lib/k8s/pod.ts` — `getLogs()` splits each chunk on newlines and holds back a trailing partial line until the rest arrives, so one stored entry is always one log line
- Fixed the last line of a stream being dropped when it has no trailing newline, by flushing the held-back line from `failCb` instead of from the cancel path
- Updated `frontend/src/components/common/Resource/logSeverityFilter.ts` — documented that an entry is matched as a whole (logic unchanged)
- Added tests for chunk splitting, lines spanning two chunks, prettified entries, and the reported reproduction end to end

## Steps to Test

1. Create a pod that emits JSON logs with explicit severity levels:

   ```sh
   kubectl run log-severity-demo --image=busybox --restart=Never -- sh -c '
     echo "{\"level\":\"info\",\"message\":\"test0\"}"
     echo "{\"level\":\"warn\",\"message\":\"test1\"}"
     echo "{\"level\":\"error\",\"message\":\"test2\"}"
     echo "{\"level\":\"warn\",\"message\":\"test3\"}"
     echo "{\"level\":\"debug\",\"message\":\"test4\"}"
     sleep 3600'
   ```

2. Open the pod in Headlamp, click **Logs**, leave **Prettify** off.
3. Filter by **Error** → only `test2`. **Warning** → `test1` and `test3`.  **Info** → `test0`. **Debug** → `test4`. All severities → all five.
4. Turn **Prettify** on and repeat. Each view should show whole JSON objects, never a stray `{`, `}`, or lone field from a filtered-out log.

## Notes for the Reviewer

- The `logSeverityFilter.ts` diff is docs and a rename only; its logic is identical to `main`. The real fix is entirely in `pod.ts`.
- `failCb` is used rather than the cancel path because `connectStream` skips `onFail` when we close the socket ourselves, so it fires exactly when the stream ends on its own. Flushing from cancel pushed a debounced `setState` while the log viewer was tearing down.
- Entries are classified whole rather than re-split inside the filter, because with **Prettify** on an entry is a multi-line JSON object — splitting it would drop its `"level"` line and leave the braces and remaining fields on screen. There is a test for this.

### DIFF
--- a/frontend/src/components/common/Resource/LogsButton.test.tsx
+++ b/frontend/src/components/common/Resource/LogsButton.test.tsx
@@ -566,10 +566,15 @@ describe('LogsButton', () => {
     await waitFor(() => expect(onLogsCallback).toBeDefined());
 
     act(() => {
-      onLogsCallback({ logs: Array.from({ length: 1001 }, (_, index) => `line ${index}\n`) });
+      onLogsCallback({ logs: Array.from({ length: 2500 }, (_, index) => `line ${index}\n`) });
     });
     expect(mockXTermWrite).toHaveBeenCalledWith(expect.stringContaining('line 999'));
     expect(requestAnimationFrameSpy).toHaveBeenCalled();
+
+    // A batch larger than the chunk size has to be drained completely, even if the
+    // stream goes quiet right after it.
+    expect(mockXTermWrite).toHaveBeenCalledWith(expect.stringContaining('line 1000'));
+    expect(mockXTermWrite).toHaveBeenCalledWith(expect.stringContaining('line 2499'));
 
     mockXTermWrite.mockClear();
     unmount();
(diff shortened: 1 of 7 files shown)


illume (changes requested):
Thanks for working on this.

PR #7335, opened 2026-08-19
frontend: ClusterChooserPopup: Fix recent cluster ordering

### LINKED ISSUE #7338: frontend: ClusterChooserPopup lists recent clusters in the wrong order
## Description

The cluster chooser popup lists "Recent clusters" in the order the cluster config was built rather than the order the clusters were last used.

`getRecentClusters()` returns names most-recent-first, but `ClusterChooserPopup` builds its recent list by iterating `Object.values(clusters)` and pushing whichever entries appear in that name list, which discards the recency order. `Chooser.tsx` and `Home/RecentClusters.tsx` read the same helper by mapping over the stored names, so they are unaffected.

## Steps to reproduce

1. Visit cluster A, then cluster B, then cluster C.
2. Open the cluster chooser popup.
3. The "Recent clusters" section is ordered by the cluster config, not C, B, A.

## Expected

Recent clusters are listed most-recently-used first.

RajPrakash681:
## Summary

`getRecentClusters()` returns cluster names most-recent-first, but `ClusterChooserPopup` builds its "Recent clusters" list by iterating `Object.values(clusters)` and pushing the matching entries, so the recency order is lost and the list follows the cluster config order instead.

`Chooser.tsx` and `Home/RecentClusters.tsx` read the same helper by mapping over the stored names, which preserves the order.

## Related Issue

Fixes #7338

## Changes

- Build the recent list by walking the stored names instead of the cluster config.
- De-duplicate the stored names, since a repeated entry in `localStorage` rendered the same cluster twice.
- Keep the current cluster pinned to the top using a stable partition rather than a non-transitive comparator.
- Added `ClusterChooserPopup.test.tsx` covering recency order, duplicate entries and current-cluster pinning.

## Steps to Test

1. `cd frontend && npx vitest run src/components/cluster/ClusterChooserPopup.test.tsx`
2. Confirm 3 tests pass. They fail on `main`.

### DIFF
--- a/frontend/src/components/cluster/ClusterChooserPopup.test.tsx
+++ b/frontend/src/components/cluster/ClusterChooserPopup.test.tsx
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render, screen } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


RajPrakash681:
@illume can you re check this?

illume (changes requested):
Thanks for the contribution.


<details>

</details>

illume (changes requested):
Thanks for the contribution.

PR #7337, opened 2026-08-19
frontend: KubeList: Keep resource version on ERROR watch events

### LINKED ISSUE #7340: frontend: KubeList.applyUpdate drops the list resource version on ERROR events
## Description

`KubeList.applyUpdate` overwrites the list's `resourceVersion` with the incoming event's value for every event type, including `ERROR`. A watch `ERROR` event carries a `Status` object, which has no `metadata.resourceVersion`, so the list's resource version becomes `undefined`.

The stale-update guard at the top of the function requires `list.metadata.resourceVersion` to be set, so it stops working from that point on and an out-of-date event is applied.

## Steps to reproduce

1. Apply an `ERROR` watch event to a list.
2. `list.metadata.resourceVersion` is now `undefined`.
3. Apply a `DELETED` event with an already-seen resource version.
4. The item is removed even though the event is stale.

## Expected

An `ERROR` event leaves the list and its resource version unchanged, and later stale events are still skipped.

RajPrakash681:
## Summary

`KubeList.applyUpdate` overwrote the list's `resourceVersion` with the incoming event's value for every event type, including `ERROR`. A watch `ERROR` event carries a `Status` object, which has no `metadata.resourceVersion`, so the list's resource version became `undefined`.

That has two consequences. The next reconnect sends `resourceVersion=` with no value, which the API server reads as "start from now", so every event between the error and the reconnect is dropped and the cached list silently drifts from the cluster. And the stale-update guard at the top of the function requires `list.metadata.resourceVersion` to be set, so it stops working from that point on and an out-of-date event is applied — a replayed `DELETED` event then removes an item that still exists.

## Related Issue

Fixes #7340

## Changes

`KubeList.applyUpdate`:

- Return the list unchanged for `ERROR` and unknown event types instead of rebuilding it with an empty resource version.
- Fall back to the current resource version when an applied event does not carry one.
- Handle `BOOKMARK` explicitly: advance the resource version, leave the items alone. This is required by the change above rather than separate from it. `BOOKMARK` previously fell into the `default` branch, where it logged `Unknown update type` and then advanced the cursor by falling through to the shared return. Making `default` return early would have stopped the cursor advancing during idle periods, which is the only thing a bookmark exists to do, so the type is now named in the union and given its own case.

`useKubeObjectList`, both watch paths:

- Log every `ERROR` event, preserving the diagnostic `applyUpdate` used to emit, then handle the event instead of feeding it back into `applyUpdate`.
- On `reason: Expired` or `code: 410`, invalidate the cached list so it is refetched from a fresh resource version. The multiplexed path also drops its tracked cursor, so the resubscription uses the refreshed value rather than the expired one.

Tests — seven added, all of which fail on `main`:

- `KubeList.test.ts` (12 total): the resource version survives an `ERROR` event; a stale `DELETED` event is still skipped afterwards; a `BOOKMARK` with no resource version is ignored; an event with an empty resource version keeps the current one.
- `useKubeObjectList.test.tsx` (41 total): the legacy path logs and invalidates on an expired `ERROR`; it reports a non-expired `ERROR` without invalidating or touching the cursor; and the multiplexed path, driven through the callback passed to `WebSocketManager.subscribe`, resubscribes at the refreshed resource version after an expired `ERROR`.

## Steps to Test

1. `cd frontend && npx vitest run src/lib/k8s/api/v2/KubeList.test.ts src/lib/k8s/api/v2/useKubeObjectList.test.tsx`
2. Confirm 53 tests pass.
3. Check out `main` for `KubeList.ts` and `useKubeObjectList.ts` only, keeping both test files, and re-run: the seven new tests fail.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/KubeList.test.ts
+++ b/frontend/src/lib/k8s/api/v2/KubeList.test.ts
@@ -156,6 +156,84 @@ describe('KubeList.applyUpdate', () => {
     consoleErrorSpy.mockRestore();
   });
 
+  it('should keep the list resource version when an ERROR event carries none', () => {
+    const consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    const update = {
+      type: 'ERROR',
+      object: { apiVersion: 'v1', kind: 'Status', metadata: {} },
+    } as any;
+
+    const updatedList = KubeList.applyUpdate(initialList, update, itemClass, cluster);
+
+    expect(updatedList.metadata.resourceVersion).toBe('1');
+    consoleSpy.mockRestore();
+  });
+
+  it('should still skip stale updates after an ERROR event', () => {
+    const consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+    const afterError = KubeList.applyUpdate(
+      initialList,
+      { type: 'ERROR', object: { apiVersion: 'v1', kind: 'Status', metadata: {} } } as any,
+      itemClass,
+      cluster
+    );
+
+    const stale = KubeList.applyUpdate(
+      afterError,
+      {
+        type: 'DELETED',
+        object: {
+          apiVersion: 'v1',
+          kind: 'MockKubeObject',
+          metadata: { uid: '1', resourceVersion: '1' },
+        },
+      } as any,
+      itemClass,
+      cluster
+    );
+
+    expect(stale.items).toHaveLength(1);
+    consoleSpy.mockRestore();
+  });
+
+  it('should advance the resource version on a BOOKMARK event', () => {
+    const update = {
+      type: 'BOOKMARK',
+      object: { apiVersion: 'v1', kind: 'MockKubeObject', metadata: { resourceVersion: '9' } },
+    } as any;
+
+    const updatedList = KubeList.applyUpdate(initialList, update, itemClass, cluster);
+
+    expect(updatedList.metadata.resourceVersion).toBe('9');
+    expect(updatedList.items).toEqual(initialList.items);
+  });
+
+  it('should ignore a BOOKMARK event with no resource version', () => {
+    const update = {
+      type: 'BOOKMARK',
+      object: { apiVersion: 'v1', kind: 'MockKubeObject', metadata: {} },
+    } as any;
+
+    expect(KubeList.applyUpdate(initialList, update, itemClass, cluster)).toBe(initialList);
+  });
+
+  it('should keep the list resource version when an event carries an empty one', () => {
+    const update = {
+      type: 'ADDED',
+      object: {
+        apiVersion: 'v1',
+        kind: 'MockKubeObject',
+        metadata: { uid: '2', resourceVersion: '' },
+      },
+    } as any;
+
+    const updatedList = KubeList.applyUpdate(initialList, update, itemClass, cluster);
+
+    expect(updatedList.metadata.resourceVersion).toBe('1');
+    expect(updatedList.items).toHaveLength(2);
+  });
+
   it('should log an error on unknown event type', () => {
     const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
     const updateEvent: KubeListUpdateEvent<MockKubeObject> = {
--- a/frontend/src/lib/k8s/api/v2/KubeList.ts
+++ b/frontend/src/lib/k8s/api/v2/KubeList.ts
@@ -31,7 +31,7 @@ export interface KubeList<T extends KubeObjectInterface> {
 }
 
 export interface KubeListUpdateEvent<T extends KubeObjectInterface> {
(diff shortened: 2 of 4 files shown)


unlikelyzero (changes requested):
Thanks for the fix — this is a real improvement over the current behavior, but I think it leaves the original problem only partly solved and introduces a couple of new gaps.

1. On an ERROR watch event (most notably HTTP 410/Expired, when the requested resourceVersion falls out of etcd's compaction window), `applyUpdate` now just logs and returns the list unchanged — still carrying the now-invalid resourceVersion. Neither watch caller invalidates or refetches the list on ERROR, so a reconnect can end up reusing that same expired resourceVersion and 410 again, repeating indefinitely with the UI stuck on stale data. Can we treat ERROR (specifically Expired/410) as a signal to invalidate/refetch the list and reconnect from a fresh resourceVersion, rather than just preserving the cursor that caused the failure? [`KubeList.ts:82-84`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/api/v2/KubeList.ts#L82-L84)

2. `KubeListUpdateEvent['type']` doesn't include `'BOOKMARK'`, so a BOOKMARK event now falls into the new `default: return list` branch and its resourceVersion advance is silently dropped. Before this change, that fell through to the final return and advanced the version correctly (by accident). `allowWatchBookmarks` is already a wired query parameter, so this isn't purely hypothetical — dropping bookmark advances means reconnects resume from an unnecessarily old cursor, which increases the odds of hitting the exact 410 this PR is trying to guard against. Worth adding an explicit BOOKMARK case that advances `metadata.resourceVersion` without touching items. [`KubeList.ts:33-35`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/api/v2/KubeList.ts#L33-L35)

3. In `useKubeObjectList.ts`, the separate `latestResourceVersions` map is written before `applyUpdate`'s new stale-event guard runs, and has no monotonicity check of its own. So a stale event that `KubeList` correctly rejects can still move this cursor backward, independent of what `KubeList` actually accepted — the two trackers can end up disagreeing about the current resourceVersion. [`useKubeObjectList.ts:334-340`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/api/v2/useKubeObjectList.ts#L334-L340)

4. Smaller one: the new `?? list.metadata.resourceVersion` fallback is dead code for the ERROR/default paths — both now `return list` before reaching it — so it isn't actually what fixes the ERROR case (the early return is). For the branches where it is reachable (ADDED/MODIFIED/DELETED), `??` also won't catch an empty-string resourceVersion, only null/undefined. [`KubeList.ts:94-95`](https://github.com/kubernetes-sigs/headlamp/blob/main/frontend/src/lib/k8s/api/v2/KubeList.ts#L94-L95)

Happy to take another look once these are addressed.

RajPrakash681:
Thanks, this was a much better read of the problem than mine. All four addressed.

**1. ERROR / Expired should refetch, not preserve the bad cursor.** Agreed — preserving the cursor that just failed is not a fix on its own. Both watch paths now treat an `Expired`/410 `Status` as a signal to refetch rather than an event to apply: they invalidate the list query and return early. The multiplexer path also drops its cached cursor for that key so it cannot be reused. Non-Expired errors keep the previous behaviour of logging and leaving the list alone.

**2. BOOKMARK.** You're right that this was an accidental pass-through that my `default: return list` broke. `'BOOKMARK'` is now in `KubeListUpdateEvent['type']` with an explicit case that advances `metadata.resourceVersion` and returns `items` untouched. A bookmark carrying no resource version returns the list unchanged rather than allocating a new object. Two tests cover both.

**3. `latestResourceVersions` monotonicity.** Guard added — the cursor only moves when the incoming version is strictly newer.

Worth flagging what I found while testing it, because it changes what this fix is worth: the ref is effectively write-only. The `connections` memo reassigns `latestResourceVersions.current[key] = list.resourceVersion` at `useKubeObjectList.ts:311` and reads it back at `:318`, so the value written from the update handler is always overwritten before it reaches a URL. The real cursor is `list.resourceVersion` from the query cache. So the two trackers cannot currently disagree in a way that affects a reconnect, and I could not write a test that failed without the guard. I've kept it as hygiene — it makes the ref correct if anything ever does read it between recomputes — but I did not want to claim it fixes live behaviour. Happy to drop it, or to remove the ref entirely as a follow-up, if you'd prefer one of those.

**4. Dead fallback and empty strings.** Correct on both counts. The early return is what fixes the ERROR case. The fallback is now `||` rather than `??`, so an empty-string resource version falls back instead of being written through, and it only sits on the ADDED/MODIFIED/DELETED path where it is reachable. Test added for the empty-string case.

Verified locally: 2972 tests across 223 files pass, `tsc --noEmit` and `eslint --max-warnings 0` clean.

RajPrakash681:
Hey can you please have a look?? @unlikelyzero

illume (changes requested):
Thanks for the contribution.

PR #7341, opened 2026-08-19
frontend: LeaderWorkerSetList: Add isEnabled gating with Checking and Not-enabled states

### LINKED ISSUE #7219: frontend: LeaderWorkerSetList: Show Checking / Not-enabled when API is missing

LeaderWorkerSet is an optional API (`leaderworkerset.x-k8s.io/v1`). The Workloads sidebar always links to Leader Worker Sets, and `LeaderWorkerSetList` always mounts `ResourceListView` with no discovery probe.

JobSet already does the right thing (merged in #7016):

* `JobSet.isEnabled()` probes the API group for `jobsets`
* `JobSetList` shows **Checking…** → list → **not enabled** (+ docs link)
* Storybook covers Checking / NotEnabled / Items

LeaderWorkerSet has none of that:

* `frontend/src/lib/k8s/leaderWorkerSet.ts` — no `isEnabled()`
* `frontend/src/components/leaderworkerset/List.tsx` — always renders the list view
* `LeaderWorkerSetList.stories.tsx` — only an `Items` story

On clusters without LeaderWorkerSet installed, operators get a generic list/error experience instead of a clear “not available on this cluster” empty state. The same pattern will be needed for optional groups like DRA (`resource.k8s.io`) on Devices pages.

Mirror the JobSet / VPA pattern for LeaderWorkerSet:

1. Add `LeaderWorkerSet.isEnabled()` (probe `/apis/leaderworkerset.x-k8s.io/v1` for resource name `leaderworkersets`; treat request failure / missing resource as not enabled). Add unit tests next to `leaderWorkerSet.test.ts`.
2. Update `LeaderWorkerSetList` to show:
   * **Checking** while discovery is in flight
   * Normal list when enabled
   * **Not enabled** empty state (short message + link to LeaderWorkerSet docs, same idea as JobSet/VPA)
3. Extend `LeaderWorkerSetList.stories.tsx` with **Checking** and **NotEnabled** stories (MSW), keeping existing **Items**.

Out of scope:

* Hiding the Leader Worker Sets sidebar entry via discovery (#6762 / follow-up)
* Devices / DRA UI (#4831)
* Map source gating for JobSet/LWS (separate follow-up)
* Changing list columns or details beyond the enablement gate

Operators using Headlamp on clusters where LeaderWorkerSet is not installed. They get a clear empty state instead of a confusing failed or empty generic list. Also documents the optional-API pattern already used by VPA and JobSet for future optional groups.

-

## Additional context

Related: JobSet empty states (#7015 / #7016), VPA Storybook (#7013 / #7014), optional-API sidebar gating (#6762).

Pritz395:
## Summary

LeaderWorkerSet is an optional CRD that may not be installed on every cluster. Without gating, navigating to the LeaderWorkerSet list on a cluster without the CRD produces a raw API error. This PR adds an `isEnabled()` probe (mirroring VPA's existing pattern) so the UI shows a meaningful empty state instead.

Fixes #7219

## Changes

- **`leaderWorkerSet.ts`**: Added `LeaderWorkerSet.isEnabled()` static method that probes the `leaderworkerset.x-k8s.io/v1` API group
- **`List.tsx`**: Gated `ResourceListView` behind the `isEnabled()` check with three states: Checking, Not-enabled (with "Learn More" link), and the normal list
- **`LeaderWorkerSetList.stories.tsx`**: Added `Checking` and `NotEnabled` stories with MSW handlers; updated existing `Items` story to also mock the API group probe
- Updated i18n locale files and storyshot snapshots

## Steps to Test

1. Run `npm run frontend:test` — all tests pass
2. Run `npm run frontend:storybook` → navigate to LeaderWorkerSet/List
   - **Items**: Shows the normal list with mock data
   - **Checking**: Shows "Checking if LeaderWorkerSet is enabled…" spinner state
   - **NotEnabled**: Shows "LeaderWorkerSet is not enabled on this cluster. Learn More" with link to LWS installation docs

## Screenshots

_Storybook stories cover all three visual states._

### DIFF
--- a/frontend/src/components/leaderworkerset/LeaderWorkerSetList.stories.tsx
+++ b/frontend/src/components/leaderworkerset/LeaderWorkerSetList.stories.tsx
@@ -74,6 +74,9 @@ updatingLeaderWorkerSet.status = {
   ],
 };
 
+const lwsApiGroupUrl = `${API_BASE}/apis/leaderworkerset.x-k8s.io/v1`;
+const lwsListUrl = `${API_BASE}/apis/leaderworkerset.x-k8s.io/v1/leaderworkersets`;
+
 export default {
   title: 'LeaderWorkerSet/List',
   component: List,
@@ -87,21 +90,6 @@ export default {
       );
     },
   ],
-  parameters: {
-    msw: {
-      handlers: {
-        story: [
-          http.get(`${API_BASE}/apis/leaderworkerset.x-k8s.io/v1/leaderworkersets`, () =>
-            HttpResponse.json({
-              kind: 'LeaderWorkerSetList',
-              metadata: {},
-              items: leaderWorkerSetList,
-            })
-          ),
-        ],
-      },
-    },
-  },
 } as Meta;
 
 const Template: StoryFn = () => {
@@ -109,3 +97,54 @@ const Template: StoryFn = () => {
 };
 
 export const Items = Template.bind({});
+Items.parameters = {
+  msw: {
+    handlers: {
+      story: [
+        http.get(lwsApiGroupUrl, () =>
+          HttpResponse.json({ resources: [{ name: 'leaderworkersets' }] })
+        ),
+        http.get(lwsListUrl, () =>
+          HttpResponse.json({
+            kind: 'LeaderWorkerSetList',
+            metadata: {},
+            items: leaderWorkerSetList,
+          })
+        ),
+      ],
+    },
+  },
+};
+
+/**
+ * LeaderWorkerSet.isEnabled() stays pending so the list shows the checking state.
+ */
+export const Checking = Template.bind({});
+Checking.parameters = {
+  storyshots: { disable: true },
+  msw: {
+    handlers: {
+      story: [http.get(lwsApiGroupUrl, () => new Promise(() => {}))],
+    },
+  },
+};
+
+/**
+ * API group responds without leaderworkersets, so isEnabled() resolves false.
+ */
+export const NotEnabled = Template.bind({});
+NotEnabled.parameters = {
+  msw: {
+    handlers: {
+      story: [
+        http.get(lwsApiGroupUrl, () =>
+          HttpResponse.json({
+            kind: 'APIResourceList',
+            groupVersion: 'leaderworkerset.x-k8s.io/v1',
+            resources: [],
+          })
+        ),
+      ],
+    },
+  },
+};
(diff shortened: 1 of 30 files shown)


unlikelyzero (changes requested):
A few things I noticed going through this:

1. `LeaderWorkerSet.isEnabled()` (frontend/src/lib/k8s/leaderWorkerSet.ts:116) collapses every discovery-request failure — 403, timeout, network error, 5xx — into a plain `false`, same as "CRD not installed." A user who's denied access to the discovery endpoint, or hits a transient network error, gets told LWS isn't enabled and shown an install link, which is wrong. The error is also swallowed silently with no logging.

2. Same method only probes the "current" cluster via `getCluster()`, but `ResourceListView`/`useSelectedClusters()` is multi-cluster-aware. If a user selects clusters where LWS exists on one but not the other, the whole list gets replaced with the "not enabled" state even though valid data exists on the other cluster.

3. The `isEnabled().then(setLwsEnabled)` effect in List.tsx:44 has no unmount-cancellation guard. `JobSetList` hit this same race and fixed it in commit 4215a48e0, which already existed on this PR's base branch — worth using the same `cancelled` flag pattern here.

4. `leaderWorkerSet.test.ts` has no test coverage for `isEnabled()`, unlike the JobSet/PodGroup equivalents it's modeled after.

5. `isEnabled()` hardcodes `/apis/leaderworkerset.x-k8s.io/v1` and `leaderworkersets` instead of deriving them from the class's own `apiVersion`/`apiName` static fields a few lines above.

6. The heading in the Checking/Not-enabled states renders as h2 instead of h1, since those branches pass a plain string to `SectionBox` instead of using `SectionHeader`/`SectionFilterHeader` like the normal list state does.

7. This is now a fourth near-identical copy of this same "probe an API group" pattern (VPA, JobSet, PodGroup, now this). Might be worth pulling into a shared helper at some point.

8. The "not enabled on this cluster" wording in List.tsx:69 is inconsistent with VPA/JobSet's phrasing, and given #1 and #2 above, it's also asserting more than what was actually checked.

One more to flag: `request()`'s `autoLogoutOnAuthError` defaults to true, so a 401 during this background probe could trigger an auto-logout while the surrounding catch swallows the error. Not sure if that's intentional, but worth confirming.

illume (changes requested):
Thanks for working on this.


Looks like there are some lint errors in the frontend — `cd frontend && npm run lint` will show them.

<details>

</details>

illume (changes requested):
Thanks for these changes.

PR #7344, opened 2026-08-19
frontend: EndpointSlice: Add model class unit tests

Adi-204:
## Description

`EndpointSlice` (`frontend/src/lib/k8s/endpointSlices.ts`) had no unit
test coverage for its `ports` and `getOwnerServiceName()` getters,
unlike sibling model classes such as `Ingress` and `Service` which
already have dedicated unit tests.

Adds `frontend/src/lib/k8s/endpointSlices.test.ts` covering:
- `getBaseObject()` defaults
- `spec` getter
- `ports` getter, including empty and undefined `ports`
- `getOwnerServiceName()`: found, no owner references, no `Service`
  owner reference, and multiple owner references

No production code is changed.

## Testing

Run from `frontend`:

```bash
npx vitest run src/lib/k8s/endpointSlices.test.ts
```

All 9 tests pass.

### DIFF
--- a/frontend/src/lib/k8s/endpointSlices.test.ts
+++ b/frontend/src/lib/k8s/endpointSlices.test.ts
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, expect, it } from 'vitest';
+import App from '../../App';
+import EndpointSlice from './endpointSlices';
+
+// cyclic imports fix
+// eslint-disable-next-line no-unused-vars
+const _dont_delete_me = App;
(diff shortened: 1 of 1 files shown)


unlikelyzero (approved):
Nice addition of coverage for the EndpointSlice model. Tests pass and match the current implementation — a few things worth follow-up, none blocking:

1. `getOwnerServiceName()` (`frontend/src/lib/k8s/endpointSlices.ts:69`) only resolves the owning Service via `metadata.ownerReferences`. Kubernetes also (and for selectorless/externally-managed Services, exclusively) uses the `kubernetes.io/service-name` label for this association. A slice with the label but no Service ownerReference will resolve to `undefined` today, which means it silently drops out of the Service details page and the resource-map edge. This is a pre-existing gap in the model class rather than something wrong with these tests — could be worth a follow-up issue to add a label fallback.

2. The `ports` getter (`frontend/src/lib/k8s/endpointSlices.ts:65`) maps `port => port.port` unconditionally, but `discovery.k8s.io/v1` allows a port entry to omit `port`. None of the new tests cover that case, so a regression there (e.g. `[undefined]` leaking into the ports list/labels in `endpointSlices/List.tsx`) wouldn't be caught. Worth adding a fixture case for it, and possibly loosening `KubeEndpointSlicePort.port` to optional to match the real schema.

3. Smaller one: `getBaseObject()`'s test only checks the inherited base-class shape (kind/apiVersion), not anything EndpointSlice-specific — not a problem for this PR, just noting it doesn't add much signal on its own.

Approving as-is; items above are follow-up work, not blockers.

Adi-204:
@unlikelyzero agreed! Thanks for the review I will do followup of pointer 1 and 2. @illume can you please review 🙏

Adi-204:
@illume looks good from copilot side! can you take final look.

Issue #7046, opened 2026-08-10
backend: spa: go test fails on Windows because TestURLToRelAcceptsSafeRelativePath hardcodes forward slashes


Shivanshu-GITH:
## Describe the bug

`go test ./...` fails in the backend on Windows. One test in `backend/pkg/spa` asserts a
path using forward slashes, while the function under test deliberately returns an
OS-relative path.

```
--- FAIL: TestURLToRelAcceptsSafeRelativePath (0.00s)
    pathSafety_internal_test.go:31: unexpected relative path: got "headlamp\\assets\\main.js"
FAIL	github.com/kubernetes-sigs/headlamp/backend/pkg/spa	3.749s
FAIL
```

`urlToRel` in `backend/pkg/spa/pathSafety.go` is documented as converting "a URL path to a
relative OS path for use with `os.Root`", and calls `filepath.FromSlash` to do exactly
that:

```go
rel := filepath.Clean(filepath.FromSlash(strings.TrimLeft(urlPath, "/")))
```

Both callers - `tryBrotliSidecar` in `encoding.go` and the handler in `spaHandler.go` -
pass the result to `os.Root`. So returning `headlamp\assets\main.js` on Windows is the
intended behaviour.

The assertion in `pathSafety_internal_test.go` hardcodes the Unix form:

```go
if rel != "headlamp/assets/main.js" {
```

So the production code is correct and the test is not portable.

## To Reproduce

Steps to reproduce the bug:
1. On Windows, clone the repository and install Go.
2. Run `npm run backend:test`, or `cd backend && go test ./...`.
3. See `TestURLToRelAcceptsSafeRelativePath` fail with
   `unexpected relative path: got "headlamp\\assets\\main.js"`.

Scope of the failure on Windows:

- `backend/pkg/spa`: 57 tests pass, 1 fails
- Whole backend: 14 packages pass, 1 package fails, and this test is the only cause

## Environment (please provide info about your environment):
 - Installation type: source checkout (development)
 - Headlamp Version: main @ 99a230be (0.44.0)
 - Other: Windows 11 Home 10.0.26200, Go 1.26.5 windows/amd64

## Are you able to fix this issue?
Yes (I will propose a PR)

## Additional Context

This is not visible in CI because the backend test jobs run on Linux only -
`backend-test.yml` runs `go test ./...` on `ubuntu-latest`, as does the `go test ./...`
step in `build-other-arch.yml`.

[...]

PR #7345, opened 2026-08-19
backend: spa: Add unit tests for detectContentType

Adi-204:
## Description

`detectContentType` (`backend/pkg/spa/contentType.go`) decides the
`Content-Type` header for every file the SPA handler serves, but had
no direct unit test coverage — it was only exercised incidentally
through handler-level tests (e.g. asserting a response's
`Content-Type` contains `"javascript"`).

Adds `backend/pkg/spa/contentType_internal_test.go` covering:
- The extension-based fast path (`.html`, `.css`, `.js`, `.png`),
  asserting `open()` is never called when the extension is known
- The sniff-based fallback when the extension is unknown, using a
  PNG magic-byte signature
- `open()` returning an error
- An empty file

No production code is changed. The test file follows this package's
existing convention for testing unexported functions: `package spa`
(internal test) with plain `testing`, matching
`pathSafety_internal_test.go` and `encoding_internal_test.go`.

## Testing

Run from `backend`:

```bash
go test ./pkg/spa/... -run TestDetectContentType -v
```

All 4 tests (6 subtests) pass.

Note: `go test ./pkg/spa/...` has one pre-existing, unrelated failure
on Windows (`TestURLToRelAcceptsSafeRelativePath`, hardcoded `/` path
separator), tracked in #7046 and not touched by this PR.

### DIFF
--- a/backend/pkg/spa/contentType_internal_test.go
+++ b/backend/pkg/spa/contentType_internal_test.go
@@ -1,0 +1,77 @@
+package spa
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"strings"
+	"testing"
+)
+
+func TestDetectContentTypeUsesExtensionWithoutOpening(t *testing.T) {
+	tests := []struct {
+		name       string
+		path       string
+		wantSubstr string
+	}{
+		{name: "html", path: "index.html", wantSubstr: "html"},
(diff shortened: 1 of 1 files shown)


Adi-204:
@illume looks good from copilot side! can you take final look.

PR #7349, opened 2026-08-19
frontend: workloads: Show service account in the details view

### LINKED ISSUE #6773: Show service account in workload views
I'd like to see what service accounts my workloads (Deployments/DaemonSets/StatefulSet/CronJobs etc) are using, so I can easily audit their permissions (thanks to the new feature on 0.44 that shows all the RoleBindings of a service account)

When viewing a workload, it should list the service account, which can be clicked to view the details of the service account (currently you have to click into a pod to be able to do this). Node Selectors are already extracted from `.spec.template.spec` and displayed, which is alongside the service account name.

All type of headlamp users, who want to see what service accounts are being used by their workloads

No

## Additional context
Would look something like this:

Image

vyncent-t:
Republished PR from @krsatyamthakur-droid w/ fixes included PR #6819 

## Summary

Workload details already surface pod-template fields like the node selector, but not the service account, so to see which one a workload runs as you had to drill into one of its pods. With the service account now linking to its RoleBindings, showing it on the workload makes it easy to audit what a workload can do.

This adds a Service Account row that links to the account's details, in the shared WorkloadDetails (Deployment, DaemonSet, StatefulSet, ReplicaSet, JobSet) and in the Job and CronJob detail views. CronJob reads it from the nested jobTemplate. When no service account is set it shows "default", the account the pod actually runs as.

## Related Issue

Fixes #6773

## Changes

- Added a Service Account row (linking to the ServiceAccount details) in components/workload/Details.tsx, following the same pattern the Pod view already uses
- Added the same row to the Job (components/job/Details.tsx) and CronJob (components/cronjob/Details.tsx) detail views, with CronJob reading from spec.jobTemplate.spec.template.spec
- Falls back to "default" when no serviceAccountName is set
- Regenerated the affected Job/CronJob storyshots and updated i18n

## Steps to Test

1. Open a Deployment, DaemonSet, StatefulSet, ReplicaSet, Job or CronJob details page
2. Confirm a "Service Account" row appears next to the other pod-template fields, linking to that service account's details page
3. Click it and confirm it opens the ServiceAccount details
4. For a workload with no serviceAccountName set, confirm it shows "default"

Service account shown on a Job's details, linking to the account:
Screenshot 2026-08-03 at 12 18 01 PM
And on a CronJob (read from the jobTemplate):
image

## Notes for the Reviewer

- The router points Deployment/DaemonSet/StatefulSet/ReplicaSet/JobSet at the shared WorkloadDetails, so the row is added there once; Job and CronJob have their own detail components so they're handled separately
- The Job/CronJob storyshots cover the new row. WorkloadDetails uses the identical code but isn't exercised by a storyshot, so I verified that path manually


vyncent-t:
Contributor has addressed issued, republish no longer needed

PR #7353, opened 2026-08-20
backend: frontend: Protect desktop APIs with app token

illume:
## Summary

Protect desktop cluster REST and WebSocket APIs with the Electron per-launch backend token while preserving standalone and in-cluster behavior.

Desktop app creates a desktop-token, which prevents anyone else connecting to localhost backend services. So people on the same machine can not use their Kubernetes APIs.

## Changes

- Require the backend token only when the desktop app configures one.
- Protect desktop configuration, cluster routes, and the WebSocket multiplexer.
- Normalize and centrally add the token to v1 and v2 HTTP requests.
- Encode WebSocket credentials in a private namespaced subprotocol.
- Reject missing, invalid, duplicate, malformed, or conflicting credentials.
- Remove private credentials before forwarding requests upstream.
- Keep in-cluster deployments free of desktop token transport.
- Preserve Kubernetes authorization and public WebSocket subprotocols.

## Test coverage

- `go test ./pkg/auth ./cmd -count=1`: passed.
- Repository-pinned backend lint for `./cmd/...` and `./pkg/auth/...`: 0 issues.
- New backend token functions: 100%, 100%, and 96% statement coverage.
- Seven focused frontend suites: 125 tests passed.
- Changed frontend branches: 19/19 covered, 100% differential branch coverage.
- Electron tray suite: 9 tests passed.
- App and frontend TypeScript checks: passed.
- Desktop Playwright spec: two security flows discovered without compile or configuration errors.
- Container Playwright pod suite: eight tests discovered without configuration errors.
- Full desktop and container e2e runtimes are delegated to Linux CI because Docker is unavailable locally.

Implementation and unit tests are paired in eight coherent commits, followed by desktop e2e CI wiring.

## Manual steps to reproduce

1. Build and launch the desktop app with a configured Kubernetes cluster.
2. Find the local backend port in the app logs.
3. Request `/config` and `/clusters/<cluster>/version` without `X-HEADLAMP_BACKEND-TOKEN`; verify each returns `403`.
4. Repeat with an incorrect token; verify each returns `403`.
5. Use the desktop UI and verify cluster resources load normally with the renderer-provided token.
6. Start a resource watch and verify WebSocket updates continue while invalid or duplicate private credentials are rejected.
7. Launch the backend without `HEADLAMP_BACKEND_TOKEN`; verify standalone requests remain available without the desktop header.
8. Run an in-cluster deployment without a desktop token and verify watch updates continue.

## Screenshots

Not applicable. This changes an invisible local security boundary and has no visual difference.

Assisted by copilot

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -49,6 +49,21 @@ jobs:
         make app-linux
     - name: Verify Linux build
       run: npm run app:verify-build-linux
+    - name: Start Minikube
+      uses: medyagh/setup-minikube@d8c0eb871f6f455542491d86a574477bd3894533 # latest
+    - name: Verify Minikube
+      run: |
+        minikube status
+        kubectl config use-context minikube
+        kubectl get nodes
+    - name: Run app e2e tests
+      env:
+        HEADLAMP_APP_E2E_CONTEXT: minikube
+      run: |
+        make backend
+        npm --prefix app/e2e-tests ci
+        npm --prefix app/e2e-tests exec -- playwright install --with-deps chromium
+          xvfb-run npm run app:test:e2e -- backendToken.spec.ts
   build-windows:
     strategy:
       matrix:
--- a/app/e2e-tests/tests/backendToken.spec.ts
+++ b/app/e2e-tests/tests/backendToken.spec.ts
@@ -1,0 +1,436 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect, test } from '@playwright/test';
+import { execFileSync } from 'child_process';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { _electron, Page } from 'playwright';
+
+const clusterName = `headlamp-token-${process.pid}`;
+const watchNamespace = `headlamp-token-watch-${process.pid}`;
+const testDir = path.join(os.tmpdir(), `headlamp-e2e-backend-token-${process.pid}`);
+const minikubeKubeconfig = path.join(testDir, 'minikube.kubeconfig');
+const appKubeconfig = path.join(testDir, 'app.kubeconfig');
+const isolatedConfigDir = path.join(testDir, 'config');
+const electronExecutable = process.platform === 'win32' ? 'electron.cmd' : 'electron';
+const electronPath = path.resolve(__dirname, `../../node_modules/.bin/${electronExecutable}`);
+const appPath = path.resolve(__dirname, '../../');
+const backendPath = path.resolve(appPath, '../backend/headlamp-server');
+const backendProtocolPrefix = 'base64url.headlamp.backend.authorization.k8s.io.';
+const multiplexerProtocol = 'headlamp.multiplexer.k8s.io';
+const reusableClusterContext = process.env.HEADLAMP_APP_E2E_CONTEXT;
+
+let electronApp: Awaited<ReturnType<typeof _electron.launch>>;
+let electronPage: Page;
+
+function run(
+  command: string,
+  args: string[],
+  ignoreFailure = false,
+  env: NodeJS.ProcessEnv = process.env
+): string {
+  try {
+    return execFileSync(command, args, {
+      encoding: 'utf8',
+      env,
+      stdio: ['ignore', 'pipe', 'inherit'],
+    }).trim();
+  } catch (error) {
+    if (ignoreFailure) {
+      return '';
+    }
+    throw error;
+  }
+}
+
+/**
+ * Writes an isolated certificate-backed kubeconfig for the desktop app.
+ * Reuses a CI cluster context when configured, otherwise creates a local Minikube profile.
+ */
+function setupCertificateBackedCluster(): void {
+  fs.mkdirSync(isolatedConfigDir, { recursive: true, mode: 0o700 });
+
+  if (reusableClusterContext) {
+    fs.writeFileSync(
+      appKubeconfig,
+      run('kubectl', [
+        '--context',
+        reusableClusterContext,
+        'config',
+        'view',
+        '--minify',
+        '--raw',
+        '--flatten',
+      ]),
+      { mode: 0o600 }
+    );
+    const isolatedContext = run('kubectl', [
+      '--kubeconfig',
+      appKubeconfig,
+      'config',
+      'current-context',
+    ]);
+    run('kubectl', [
+      '--kubeconfig',
+      appKubeconfig,
+      'config',
+      'rename-context',
+      isolatedContext,
+      clusterName,
+    ]);
+  } else {
+    run('minikube', ['start', '--profile', clusterName], false, {
+      ...process.env,
+      KUBECONFIG: minikubeKubeconfig,
+    });
+  }
+
+  const kubeconfig = run('kubectl', [
+    '--kubeconfig',
+    reusableClusterContext ? appKubeconfig : minikubeKubeconfig,
+    '--context',
+    clusterName,
+    'config',
+    'view',
+    '--minify',
+    '--raw',
+    '--flatten',
+  ]);
+  if (
+    !kubeconfig.includes('certificate-authority-data:') ||
+    !kubeconfig.includes('client-certificate-data:')
+  ) {
+    throw new Error('Expected the test cluster to produce a certificate-backed kubeconfig');
+  }
+  fs.writeFileSync(appKubeconfig, kubeconfig, { mode: 0o600 });
+}
+
+/**
+ * Requests the selected backend port from the Electron main process.
+ *
+ * @returns The selected port, or undefined until one has been assigned.
+ */
+function getBackendPort(): Promise<number | undefined> {
+  return electronPage.evaluate(
+    () =>
+      new Promise<number | undefined>((resolve, reject) => {
+        const desktopApi = (
+          window as Window & {
+            desktopApi?: {
+              send: (channel: string, data?: unknown) => void;
+              receive: (
+                channel: string,
+                callback: (port: number | undefined) => void
+              ) => (() => void) | undefined;
+            };
+          }
+        ).desktopApi;
+        if (!desktopApi) {
+          reject(new Error('Desktop API is unavailable'));
+          return;
+        }
+
+        let unsubscribe: (() => void) | undefined;
+        const timeout = window.setTimeout(() => {
+          unsubscribe?.();
+          resolve(undefined);
+        }, 250);
+        unsubscribe = desktopApi.receive('backend-port', port => {
+          window.clearTimeout(timeout);
+          unsubscribe?.();
+          resolve(port);
+        });
+        desktopApi.send('request-backend-port');
+      })
+  );
+}
+
+/**
+ * Waits until the backend listener enforces authentication on its config route.
+ *
+ * @returns The port of the ready backend server.
+ */
+async function waitForBackend(): Promise<number> {
+  let backendPort: number | undefined;
+
+  await expect
+    .poll(
+      async () => {
+        backendPort = await getBackendPort();
+        if (!backendPort) {
+          return undefined;
+        }
+
+        try {
+          return (await fetch(`http://localhost:${backendPort}/config`)).status;
+        } catch {
+          return undefined;
+        }
+      },
+      { timeout: 30_000 }
+    )
+    .toBe(403);
+
+  return backendPort!;
+}
+
+function getBackendToken(): Promise<string> {
+  return electronPage.evaluate(
+    () =>
+      new Promise<string>((resolve, reject) => {
+        const desktopApi = (
+          window as Window & {
+            desktopApi?: {
+              send: (channel: string, data?: unknown) => void;
+              receive: (
+                channel: string,
+                callback: (token: string) => void
+              ) => (() => void) | undefined;
+            };
+          }
+        ).desktopApi;
+        if (!desktopApi) {
+          reject(new Error('Desktop API is unavailable'));
+          return;
+        }
+
+        let unsubscribe: (() => void) | undefined;
+        unsubscribe = desktopApi.receive('backend-token', token => {
+          unsubscribe?.();
+          resolve(token);
+        });
+        desktopApi.send('request-backend-token');
+      })
+  );
+}
+
+test.beforeAll(async () => {
+  test.setTimeout(4 * 60 * 1000);
+
+  fs.accessSync(backendPath, fs.constants.X_OK);
+  setupCertificateBackedCluster();
+  electronApp = await _electron.launch({
+    cwd: appPath,
+    executablePath: electronPath,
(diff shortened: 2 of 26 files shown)


backend/pkg/auth/backendtoken.go
```diff
@@ -0,0 +1,128 @@
...
+
+// NewBackendTokenMiddleware protects desktop API requests and removes the private
+// credential before forwarding the request to the downstream handler.
+func NewBackendTokenMiddleware(useInCluster bool) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if useInCluster {
+				next.ServeHTTP(w, r)
+				return
+			}
```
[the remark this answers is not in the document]
illume: Updated the shared token check and middleware so authentication is enabled only when HEADLAMP_BACKEND_TOKEN is configured. Standalone and in-cluster launches now bypass the desktop token boundary, with regressions covering both modes. The focused backend packages pass.

backend/pkg/auth/backendtoken_test.go
```diff
@@ -0,0 +1,228 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("HEADLAMP_BACKEND_TOKEN", tt.envToken)
+
+			req := httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/", nil)
+			req.Header.Set(backendTokenHeader, tt.headerToken)
+
+			recorder := httptest.NewRecorder()
+
+			err := auth.CheckBackendToken(tt.useInCluster, recorder, req)
```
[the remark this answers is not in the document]
illume: Rebuilt the series into ten coherent commits. Each backend, frontend, and tray implementation now lands with its unit tests in the same commit; the desktop e2e and CI wiring follow after the feature exists. This removes references to future symbols and restores bisectable ordering.

backend/cmd/headlamp.go
```diff
@@ -954,11 +954,14 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	r.HandleFunc("/config", config.getConfig).Methods("GET")
 
 	// Auth token management
-	r.HandleFunc("/auth/set-token", config.handleSetToken).Methods("POST")
+	r.Handle("/auth/set-token", auth.NewBackendTokenMiddleware(config.UseInCluster)(
+		http.HandlerFunc(config.handleSetToken))).Methods("POST")
```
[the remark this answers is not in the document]
illume: Protected /config with the shared middleware and moved frontend token injection outside the cluster-only branch, so request(/config, ..., useCluster=false) now authenticates. Unit tests assert non-cluster propagation, and the desktop e2e rejects both missing and wrong tokens for /config and cluster REST.

backend/cmd/headlamp.go
```diff
@@ -759,7 +759,7 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	addPluginRoutes(config, r)
 
 	// Setup port forwarding handlers.
-	r.HandleFunc("/clusters/{clusterName}/portforward", func(w http.ResponseWriter, r *http.Request) {
+	r.Handle("/clusters/{clusterName}/portforward", auth.NewBackendTokenMiddleware(config.UseInCluster)(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
illume: Split all four port-forward route registrations across route, middleware, and handler lines. Ran the repository-pinned golangci-lint 2.12.2 against ./cmd/... and ./pkg/auth/...; it reports 0 issues.

backend/cmd/headlamp_test.go
```diff
@@ -4066,7 +4171,12 @@ func TestClusterRequestHandlerUsesServiceAccountToken(t *testing.T) { //nolint:f
 }
 
 func TestClusterRequestHandlerFallsBackToClusterContextForWebSocketCookie(t *testing.T) {
```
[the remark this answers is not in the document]
illume: Added the targeted //nolint:funlen annotation to the expanded WebSocket cookie regression. The added setup is necessary to authenticate through the new middleware, and the repository-pinned backend lint now reports 0 issues.

frontend/src/lib/k8s/api/v1/clusterRequests.ts
```diff
@@ -141,8 +142,13 @@ export async function clusterRequest(
...
 
   const userID = getUserIdFromLocalStorage();
-  const opts: { headers: RequestHeaders } = Object.assign({ headers: {} }, otherParams);
+  const headers: RequestHeaders = {};
+  new Headers(otherParams.headers).forEach((value, key) => {
+    headers[key] = value;
+  });
+  const opts = { ...otherParams, headers };
   const cluster = paramsCluster || '';
+  Object.assign(opts.headers, getHeadlampAPIHeaders());
```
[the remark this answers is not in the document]
illume: Kept the normalized plain header object and now writes the centrally injected token to the lowercase key, overwriting any case-insensitive caller copy instead of coalescing values. Added a regression that starts with a stale uppercase caller token and verifies one normalized header with the configured value. The full focused frontend suite passes 125 tests and typechecking passes.

app/e2e-tests/tests/backendToken.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+  run('minikube', ['delete', '--profile', clusterName], true, {
+    ...process.env,
+    KUBECONFIG: minikubeKubeconfig,
+  });
+  fs.rmSync(testDir, { force: true, recursive: true });
+});
+
+test.describe('desktop backend token', () => {
+  test('rejects missing and wrong tokens for REST and multiplexer requests', async () => {
+    await expect.poll(getBackendPort).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
illume: Added waitForBackend(), which polls unauthenticated /config for the expected 403, tolerates connection errors, and allows 30 seconds for the listener to become ready. Both desktop tests now use that readiness boundary instead of only waiting for actualPort. The app Playwright configuration discovers both scenarios and app typechecking passes.

app/e2e-tests/tests/backendToken.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+    throw new Error('Expected minikube to produce a certificate-backed kubeconfig');
+  }
+  fs.writeFileSync(appKubeconfig, kubeconfig, { mode: 0o600 });
+}
+
+function getBackendPort(): Promise<number | undefined> {
+  return electronPage.evaluate(
+    () => (window as Window & { headlampBackendPort?: number }).headlampBackendPort
+  );
+}
```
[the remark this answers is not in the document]
illume: Replaced the renderer-global lookup with the existing request-backend-port/backend-port IPC contract, registering the listener before sending. Each early request now times out, unsubscribes, and returns to the outer readiness poll so it can retry after the main-process handler is registered. The Linux workflow also runs make backend explicitly before launching the source app, and the spec verifies that binary is executable before starting minikube. Playwright discovery, app typechecking, formatting, and a local make backend build all pass.

rootp1 (commented):
Found a CI blocker in the new desktop backend token e2e wiring, see inline comment.

.github/workflows/app.yml
```diff
@@ -49,6 +49,13 @@ jobs:
         make app-linux
     - name: Verify Linux build
       run: npm run app:verify-build-linux
+    - name: Test desktop backend token
```
rootp1: This step runs backendToken.spec.ts which calls minikube start, but minikube is never installed on this runner. Every other workflow in this repo that needs minikube sets it up first. Without that here, this step will fail with minikube not found.

Please check this one, @illume

backend/cmd/headlamp.go
```diff
@@ -951,14 +963,18 @@ func createHeadlampHandler(ctx context.Context, config *HeadlampConfig) http.Han
 	})
 
 	// Configuration
-	r.HandleFunc("/config", config.getConfig).Methods("GET")
+	r.Handle("/config", auth.NewBackendTokenMiddleware(config.UseInCluster)(
```
rootp1, line 967: /externalproxy (the arbitrary URL proxy route) is left outside the new token middleware, which is a gap

rootp1: @illume wdyt

backend/pkg/auth/backendtoken.go
```diff
@@ -0,0 +1,128 @@
...
+// CheckBackendToken verifies that a request carries the token configured by the desktop app.
+func CheckBackendToken(useInCluster bool, w http.ResponseWriter, r *http.Request) error {
+	backendTokenEnv := os.Getenv("HEADLAMP_BACKEND_TOKEN")
+	if useInCluster || backendTokenEnv == "" {
+		return nil
+	}
+
+	backendTokens := r.Header.Values(backendTokenHeader)
+
+	if len(backendTokens) != 1 || backendTokens[0] != backendTokenEnv {
```
rootp1: even when we are having crypto/subtle.ConstantTimeCompare we are using != here, am i undersatnding anything wrong illume

backend/cmd/headlamp.go
illume: Yeah, I think this is ok. Because that URL is responsible for authentication/authorization.

Open to arguments if I'm wrong.

rootp1: Yeah, it's not a blocker, but surely it's fine

.github/workflows/app.yml
illume: Ooops. Yeah, this should be moved.

rootp1: yup

backend/pkg/auth/backendtoken.go
illume: Ok, good catch. I'll use this.

(Not sure if over the network differences would be detectable. But better safe than sorry!)

backend/cmd/headlamp.go
illume: @ashu8912 WDYT? It's fine to leave without?

.github/workflows/app.yml
illume: Rebased onto the merged PR 7003 app e2e setup. The Linux app job now installs Minikube first and runs the standard root `npm run app:test:e2e` command with no spec filter. `backendToken.spec.ts` is registered in `playwright.electron.config.ts`, so it runs with the rest of the dedicated Electron suite (7 tests in 4 files) rather than as a separate step or in build-container.

backend/pkg/auth/backendtoken.go
illume: Updated `CheckBackendToken` to compare the supplied and configured token with `crypto/subtle.ConstantTimeCompare`. The wrong-token table case now uses an equal-length mismatch so it exercises the constant-time comparison path. Auth tests pass and the backend linter reports 0 issues.

ashu8912:
Hii @illume  the CI is failing here

backend/pkg/auth/backendtoken.go
```diff
@@ -0,0 +1,135 @@
...
+
+const (
+	backendTokenHeader         = "X-HEADLAMP_BACKEND-TOKEN"                         // #nosec G101
+	backendTokenProtocolPrefix = "base64url.headlamp.backend.authorization.k8s.io." // #nosec G101
+)
+
+// CheckBackendToken verifies that a request carries the token configured by the desktop app.
+func CheckBackendToken(useInCluster bool, w http.ResponseWriter, r *http.Request) error {
+	backendTokenEnv := os.Getenv("HEADLAMP_BACKEND_TOKEN")
+	if useInCluster || backendTokenEnv == "" {
```
ashu8912, line 38: This inverts the failure mode of the old check. Previously (headlamp.go, removed in this
PR):

    if backendToken != backendTokenEnv || backendTokenEnv == "" {
        http.Error(w, "access denied", http.StatusForbidden)   // ← denied
    }

An unset HEADLAMP_BACKEND_TOKEN with UseInCluster=false meant 403. Now it returns nil and
the request is allowed.

Routes that flip from denied to fully unauthenticated in that configuration:
  - POST   /cluster                     (addCluster)
  - DELETE /cluster/{name}              (deleteCluster)
  - DELETE /plugins/{name}
  - /clusters/{clusterName}/helm/*      (install / upgrade / uninstall)

The commit message frames this as preserving "documented standalone server mode", but
every documented standalone workflow already sets the token: package.json `backend:start`,
all the Makefile run-backend targets, backend/.air.toml, and
docs/development/cluster-inventory.md all export HEADLAMP_BACKEND_TOKEN=headlamp. So the
fail-open branch isn't serving the documented dev flow — it's serving the *undocumented*
one.

And it does have real reach: the Dockerfile ENTRYPOINT is

    ["/headlamp/headlamp-server", "-html-static-dir", ..., "-plugins-dir", ...]

with `-in-cluster` supplied separately via args in kubernetes-headlamp.yaml:32 and
charts/.../deployment.yaml:265. So `docker run` of the published image against a mounted
kubeconfig — which the docker-extension does — runs with neither `-in-cluster` nor a
token, and now exposes unauthenticated Helm install/uninstall and cluster add/delete.

I can see this is deliberate: headlamp_test.go renames
`TestRestrictedEndpointsRejectEmptyEnvToken` → `TestRestrictedEndpointsBypassedWithoutConfiguredToken`
and inverts `assert.Equal(StatusForbidden)` to `assert.NotEqual(StatusForbidden)`. So the
question isn't "is this a bug" but "is this the right trade-off" — and I don't think it is
for the mutating routes. Could you either keep those four fail-closed, or explain in the
commit body which deployment actually needs them open without a token?

ashu8912 (approved):
One finding and overall it looks good

backend/pkg/auth/backendtoken.go
illume: Your analysis of the fail-open behavior is correct. We are keeping token enforcement opt-in: an unset or empty HEADLAMP_BACKEND_TOKEN preserves direct standalone development and test servers, while setting a non-empty token enables validation on the protected routes. I added code comments, route regressions, and a dedicated backend development guide section documenting both states. The docs now explicitly warn not to expose a non-in-cluster server with an empty token because cluster, plugin, Helm, and proxy routes then lack this additional credential check. Configured tokens still use constant-time comparison; in-cluster mode continues to use its own authentication/authorization path.

illume:
Thanks @ashu8912 for the review.

I added some comments/docs to explain the backend token a bit more. I think it's ready to go?

frontend/src/lib/k8s/api/v1/clusterRequests.ts
```diff
@@ -141,7 +142,14 @@ export async function clusterRequest(
...
 
   const userID = getUserIdFromLocalStorage();
-  const opts: { headers: RequestHeaders } = Object.assign({ headers: {} }, otherParams);
+  const headers: RequestHeaders = {};
+  new Headers(otherParams.headers).forEach((value, key) => {
+    headers[key] = value;
+  });
+  Object.entries(getHeadlampAPIHeaders()).forEach(([name, value]) => {
+    headers[name.toLowerCase()] = value;
+  });
```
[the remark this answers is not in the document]
ashu8912: @illume can you check this one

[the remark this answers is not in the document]
illume: Yop, working on it...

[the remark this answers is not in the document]
illume: Fixed. `/externalproxy` now runs through `NewBackendTokenMiddleware`, so configured desktop requests are authenticated and the private `X-HEADLAMP_BACKEND-TOKEN` header is removed before the proxy handler copies headers upstream. I added coverage that includes this route in the required-token matrix and asserts the upstream server never receives the token.

illume:
Hopefully fixed the failing e2e tests.

vyncent-t (approved):
Works as intended LGTM

PR #7358, opened 2026-08-20
app: Reduce startup memory

illume:
## Summary

Reduces desktop, renderer, and Go-backend memory by deferring optional work,
shrinking retained Kubernetes objects, and tuning backend garbage collection.
The sum of reported incremental isolated startup measurements is approximately
**60.3 MiB RSS** and **9.9 MiB JavaScript heap**. These totals combine separate
component benchmarks and are directional rather than one end-to-end app
measurement.

## Changes

- Defers MCP adapter initialization until first use, including
  configured-but-unused servers.
- Splits the MCP/LangChain graph out of Electron's startup bundle.
- Lazily loads plugin archive/version helpers, process inspection, CA trust,
  tar handling, and other optional Electron resources.
- Uses a JSON-only Electron i18n backend.
- Disables GPU acceleration in detected Windows VMs unless explicitly
  overridden, with registry detection overlapped with synchronous startup work.
- Lazily loads release-note rendering, optional resource routes, and
  docs/OpenAPI helpers.
- Starts the bundled Go backend with `GOGC=25` unless the user already set
  `GOGC`.
- Strips backend binaries without changing runtime behavior.
- Documents backend profiling and the measured compiler, GC, and dynamic
  loading experiments.

### Savings by optimization

| Optimization | Measured saving | Notes |
| --- | --- | --- |
| Lazy MCP adapter initialization | ~38 MiB RSS; ~5 MiB heap | Applies with no MCP configuration; configured servers remain dormant until first use |
| Separate MCP adapter bundle | ~13.9 MiB RSS; ~2.35 MiB heap | Keeps the 1.34 MB MCP/LangChain graph out of the 1.78 MB Electron startup bundle |
| Lazy plugin dependencies and skipped normal-startup process scan | ~722 KiB heap | Isolated plugin-module initialization |
| Lazy CA, tar, and process resources plus JSON-only i18n | ~6.5 MiB RSS; ~1.86 MiB heap | Comparable isolated Electron main-process loading |
| Windows VM GPU disable | Hardware-dependent; not separately measured | Runs registry detection in a worker during synchronous startup, then performs a bounded pre-ready wait |
| Lazy release-note renderer | 156 KiB smaller initial renderer entry | Keeps a 505 KiB markdown source graph off startup |
| Lazy optional renderer modules | Not separately measured | Removes docs/OpenAPI and resource modules from the default renderer startup path |
| Desktop backend GC tuning | ~1.9 MiB startup RSS at `GOGC=50`; 4.5 MiB less private memory at `GOGC=25` after the request workload | `GOGC=25` used about 11% more CPU than `GOGC=50` in the measured workload |

### Combined totals

- **Startup RSS:** ~60.3 MiB from the additive isolated RSS measurements above
  (38 + 13.9 + 6.5 + 1.9 MiB). The backend component conservatively uses the
  measured `GOGC=50` startup delta; the app defaults to `GOGC=25`.
- **JavaScript heap:** ~9.9 MiB from the reported incremental heap measurements
  (5 + 2.35 + 1.86 MiB + 722 KiB).
- **Startup bundles:** 1.34 MB removed from Electron's main startup bundle and
  156 KiB removed from the renderer's initial entry.
- **Loaded backend:** 4.5 MiB less private memory at `GOGC=25` than the Go
  default after 20,000 local `/config` requests.
- GPU and optional frontend-module savings are intentionally excluded from the
  fixed total because they are hardware- or workload-dependent.

### Backend measurements and experiments

The backend request measurements are medians from three Linux amd64 runs of
20,000 local `/config` requests. Private dirty memory is used because
demand-paged executable mappings varied with link layout.

| Runtime configuration | Private dirty after requests | Backend CPU | Result |
| --- | ---: | ---: | --- |
| `GOGC=100` | 17.2 MiB | 3.26 s | Go default and comparison baseline |
| `GOGC=50` | 14.8 MiB | 3.37 s | Saves 2.4 MiB over the Go default |
| `GOGC=25` | 12.7 MiB | 3.73 s | Current desktop default; saves another 2.1 MiB with about 11% more CPU than `GOGC=50` |
| `GOGC=50 GOMEMLIMIT=64MiB` | 14.8 MiB | 3.43 s | No measurable saving in this workload |
| `GOGC=50 GODEBUG=disablethp=1` | 11.7 MiB | 3.44 s | Not selected because this Linux compatibility setting is scheduled for removal |
| `GOGC=50 GOMAXPROCS=1` | 13.2 MiB | 2.32 s | Not selected because serial execution can hurt concurrent workloads |

Compiler experiments used the same request workload with `GOGC=50`:

| Build option | Binary size | Private dirty | Backend CPU | Result |
| --- | ---: | ---: | ---: | --- |
| Default | 114.4 MiB | 14.8 MiB | 3.37 s | Baseline |
| `-trimpath -ldflags="-s -w"` | 81.4 MiB | 14.5 MiB | 3.38 s | Current build default; 28.8% smaller on disk with no meaningful runtime-memory change |
| `-gcflags=all=-l` | 103.5 MiB | 12.8 MiB | 4.21 s | About 2 MiB less memory at roughly 25% more CPU; not selected |
| `GOAMD64=v3` | 114.4 MiB | 14.9 MiB | 3.41 s | No memory saving and reduced CPU compatibility |
| `-buildmode=pie` | 120.9 MiB | 19.1 MiB | 3.45 s | Increased private memory |
| `GOEXPERIMENT=greenteagc` | 114.4 MiB | 14.7 MiB | 3.47 s | No meaningful saving in this workload |

Go plugins were rejected as a backend memory-saving mechanism because they are
unsupported on Windows, require exact toolchain and dependency compatibility,
and cannot be unloaded. Optional helper processes would add another Go runtime
and IPC complexity.

## Testing

- `npm run app:test:unit` (22 files, 336 tests)
- `npm run app:tsc`
- `npm run app:lint`
- Focused frontend tests (4 files, 108 tests)
- `npm run frontend:tsc`
- `npm run frontend:lint`
- `go test ./pkg/k8cache -count=1`
- `npm run app:test:e2e -- buildManifest.spec.ts` (2 tests)

Changed-line branch coverage remains above 80% for the touched Electron modules.

## Manual steps to reproduce

1. Start the desktop app with MCP enabled and no MCP servers configured.
2. Confirm the app reaches the cluster view without loading the MCP adapter.
3. Configure an MCP server and restart the desktop app.
4. Confirm the configured server remains dormant during startup.
5. Request the MCP tool inventory, invoke a tool, and confirm the server connects
   on first use.
6. Change and clear the active cluster while MCP initialization is in progress;
   confirm cluster-dependent servers use the latest context.
7. Update MCP settings during initialization and confirm stale server settings
   do not become active.
8. Verify plugins, i18n, certificates, release notes, docs, and resource routes
   still load on first use.
9. On a Windows VM, confirm GPU acceleration is disabled and
   `--disable-gpu=false` overrides detection.

## Screenshots

Not applicable. This change affects non-visual process loading and memory
retention; it does not alter the rendered interface.

## Notes for reviewers

- Memory figures are medians from isolated component or backend request
  benchmarks. The combined total is arithmetic across incremental measurements,
  not a single full-app before/after run.
- Please pay particular attention to first-use MCP discovery, cluster changes
  during initialization, settings updates during initialization, Windows VM
  detection, and shutdown cleanup during initialization.

Assisted by copilot

### DIFF
--- a/app/electron/hardwareAcceleration.ts
+++ b/app/electron/hardwareAcceleration.ts
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import path from 'node:path';
+import { Worker } from 'node:worker_threads';
+
+const DETECTION_PENDING = 0;
+const DETECTION_PHYSICAL = 1;
+const DETECTION_VIRTUAL = 2;
+const DETECTION_FAILED = 3;
+const VM_IDENTIFIERS =
+  /\b(?:amazon ec2|bochs|digitalocean|google compute engine|hyper-v|kvm|nutanix|parallels|qemu|virtual machine|virtualbox|vmware|xen)\b/i;
+
+interface DetectionWorker {
+  on(event: 'error', listener: () => void): void;
+  unref(): void;
+}
+
+/** A Windows VM detection operation running outside the Electron main thread. */
+export interface WindowsVMDetection {
+  /** Shared state updated by the detection worker. */
+  state: Int32Array;
+}
+
+/** Dependencies used to start VM detection, injectable for testing. */
+export interface WindowsVMDetectionDependencies {
+  /** Current operating system platform. */
+  platform: NodeJS.Platform;
+  /** Windows installation directory containing System32. */
+  systemRoot: string;
+  /** Starts an unreferenced worker from inline JavaScript. */
+  createWorker: (source: string, workerData: object) => DetectionWorker;
+}
+
+const defaultDependencies: WindowsVMDetectionDependencies = {
+  platform: process.platform,
+  systemRoot: process.env.SystemRoot || 'C:\\Windows',
+  createWorker(source, workerData) {
+    return new Worker(source, { eval: true, workerData });
+  },
+};
+
+/**
+ * Starts Windows VM detection while the main process continues synchronous setup.
+ *
+ * @param disableGPU - Explicit command-line preference, when provided.
+ * @param dependencies - Platform and worker operations.
+ * @returns Shared detection state, or null when automatic detection is unnecessary.
+ */
+export function startWindowsVMDetection(
+  disableGPU: boolean | undefined,
+  dependencies: WindowsVMDetectionDependencies = defaultDependencies
+): WindowsVMDetection | null {
+  if (dependencies.platform !== 'win32' || disableGPU !== undefined) {
+    return null;
+  }
+
+  const systemRoot = path.win32.isAbsolute(dependencies.systemRoot)
+    ? dependencies.systemRoot
+    : 'C:\\Windows';
+  const state = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
+  const source = `
+    const { execFileSync } = require('node:child_process');
+    const path = require('node:path');
+    const { workerData } = require('node:worker_threads');
+    const state = new Int32Array(workerData.state);
+    let result = workerData.failed;
+    try {
+      const output = execFileSync(
+        path.win32.join(workerData.systemRoot, 'System32', 'reg.exe'),
+        ['query', 'HKLM\\\\HARDWARE\\\\DESCRIPTION\\\\System\\\\BIOS'],
+        {
+          encoding: 'utf8',
+          maxBuffer: 1024 * 1024,
+          stdio: ['ignore', 'pipe', 'ignore'],
+          timeout: workerData.timeout,
+          windowsHide: true,
+        }
+      );
+      result = new RegExp(workerData.pattern, 'i').test(output)
+        ? workerData.virtual
+        : workerData.physical;
+    } catch {
+      result = workerData.failed;
+    } finally {
+      Atomics.store(state, 0, result);
+      Atomics.notify(state, 0);
+    }
+  `;
+  const worker = dependencies.createWorker(source, {
+    failed: DETECTION_FAILED,
+    pattern: VM_IDENTIFIERS.source,
+    physical: DETECTION_PHYSICAL,
+    state: state.buffer,
+    systemRoot,
+    timeout: 1000,
+    virtual: DETECTION_VIRTUAL,
+  });
+  worker.on('error', () => {
+    Atomics.store(state, 0, DETECTION_FAILED);
+    Atomics.notify(state, 0);
+  });
+  worker.unref();
+
+  return { state };
+}
+
+/**
+ * Waits for a previously started VM detection before Electron becomes ready.
+ *
+ * @param detection - Detection state returned by startWindowsVMDetection.
+ * @param timeoutMs - Maximum time to wait after synchronous startup work.
+ * @returns Whether the Windows system was identified as a virtual machine.
+ */
+export function waitForWindowsVMDetection(
+  detection: WindowsVMDetection | null,
+  timeoutMs = 1000
+): boolean {
+  if (!detection) {
+    return false;
+  }
+
+  Atomics.wait(detection.state, 0, DETECTION_PENDING, timeoutMs);
+  return Atomics.load(detection.state, 0) === DETECTION_VIRTUAL;
+}
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -29,7 +29,6 @@ import {
   shell,
 } from 'electron';
 import { IpcMainEvent, MenuItemConstructorOptions } from 'electron/main';
-import find_process from 'find-process';
 import * as fsPromises from 'fs/promises';
 import * as net from 'net';
 import { platform } from 'os';
@@ -37,7 +36,9 @@ import path from 'path';
 import url from 'url';
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
-import { setupCustomCAs, setupSystemCAs } from './certificates';
+import { withBackendMemoryDefaults } from './backendMemory';
+import { createCertificateSetup } from './certificates';
+import { startWindowsVMDetection, waitForWindowsVMDetection } from './hardwareAcceleration';
 import i18n from './i18next.config';
 import {
   getLegalDocumentsResourcePath,
@@ -109,11 +110,7 @@ const ENABLE_MCP = process.env.HEADLAMP_MCP_ENABLE !== 'false';
 dotenv.config({ path: path.join(process.resourcesPath, '.env') });
 
 const settings = loadSettings(SETTINGS_PATH);
-setupSystemCAs(settings);
-
-if (settings.customCAPath) {
-  setupCustomCAs(settings.customCAPath);
-}
+const ensureCertificates = createCertificateSetup(settings);
 
 const isDev = !!process.env.ELECTRON_DEV;
 let frontendPath = '';
@@ -192,7 +189,12 @@ if ('remote-debugging-port' in args) {
 }
 
 const isHeadlessMode = args.headless === true;
-let disableGPU = args['disable-gpu'] === true;
+const disableGPU = args['disable-gpu'];
+const windowsVMDetection = startWindowsVMDetection(disableGPU);
+if (disableGPU === true) {
+  console.info('Disabling GPU hardware acceleration. Reason: related flag is set.');
+  app.disableHardwareAcceleration();
+}
 const defaultPort = args.port || 4466;
 let actualPort = defaultPort; // Will be updated when backend starts
 const MAX_PORT_ATTEMPTS = Math.abs(Number(process.env.HEADLAMP_MAX_PORT_ATTEMPTS) || 100); // Maximum number of ports to try
@@ -361,6 +363,7 @@ class PluginManagerEventListeners {
 
     let pluginInfo: ArtifactHubHeadlampPkg | undefined = undefined;
     try {
+      ensureCertificates();
       pluginInfo = await PluginManager.fetchPluginInfo(URL, { signal: controller.signal });
     } catch (error) {
       console.error('Error fetching plugin info:', error);
@@ -456,6 +459,7 @@ class PluginManagerEventListeners {
       controller,
     };
 
+    ensureCertificates();
     PluginManager.update(
       pluginName,
       destinationFolder,
@@ -688,6 +692,17 @@ async function isPortAvailable(port: number): Promise<boolean> {
 async function findAvailablePort(startPort: number): Promise<number> {
   for (let i = 0; i < MAX_PORT_ATTEMPTS; i++) {
     const port = startPort + i;
+    // Probe the socket first so normal startup does not load and retain the
+    // process-inspection dependency when the preferred port is free.
+    const available = await isPortAvailable(port);
+
+    if (available) {
+      if (port !== startPort) {
+        console.info(`Port ${startPort} is in use, using port ${port} instead`);
+      }
+      return port;
+    }
+
     // Skip ports already used by another Headlamp instance.
     const headlampPIDs = await getHeadlampPIDsOnPort(port);
     if (headlampPIDs && headlampPIDs.length > 0) {
@@ -697,14 +712,6 @@ async function findAvailablePort(startPort: number): Promise<number> {
         )}, trying next port...`
       );
       continue;
-    }
-    const available = await isPortAvailable(port);
-
-    if (available) {
-      if (port !== startPort) {
-        console.info(`Port ${startPort} is in use, using port ${port} instead`);
-      }
-      return port;
     }
 
     console.info(`Port ${port} is occupied by another process, trying next port...`);
@@ -794,9 +801,7 @@ async function startServer(flags: string[] = []): Promise<ChildProcessWithoutNul
   const options = {
     detached: true,
     windowsHide: true,
-    env: {
-      ...extendedEnv,
-    },
+    env: withBackendMemoryDefaults(extendedEnv),
   };
 
   return spawn(serverFilePath, serverArgs, options);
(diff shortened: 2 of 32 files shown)


app/electron/mcp/MCPClient.ts
```diff
@@ -74,22 +74,33 @@ export default class MCPClient {
...
+   * stay out of memory until the first tool execution.
    */
   async initialize(): Promise<void> {
     if (this.initialized) {
       return;
     }
     this.mcpToolState = new MCPToolStateStore(this.configPath);
-
-    await this.initializeClient();
+    await this.mcpToolState.initialize();
```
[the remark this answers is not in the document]
illume: Addressed. `mcp-get-tools-config` now treats inventory retrieval as explicit first use and awaits lazy client initialization before reading the tool-state config. A regression test verifies a configured server stays dormant through app initialization, then is discovered when inventory is requested.

app/electron/mcp/MCPClient.ts
```diff
@@ -235,6 +249,13 @@ export default class MCPClient {
 
     const oldClusters = this.currentClusters;
     this.currentClusters = newClusters;
+    this.clusters = newClusters || [];
+
+    // Recording context must not start configured servers or load the adapter.
+    if (!this.client) {
+      console.log('MCP client not yet started, skipping cluster-change restart');
+      return;
+    }
```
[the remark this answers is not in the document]
illume: Addressed. Cluster changes now await any in-flight first-use initialization before deciding whether a client is active. The completed client is then closed and restarted with the latest cluster context when the server is cluster-dependent. The new concurrency test verifies the old and new contexts are used in order.

app/electron/mcp/MCPClient.ts
```diff
@@ -399,16 +421,19 @@ export default class MCPClient {
...
 
-      // Reset and reinitialize client with new config
+      const wasActive = this.client !== null;
       if (this.client && typeof this.client.close === 'function') {
         await this.client.close();
       }
       this.client = null;
+      this.clientTools = [];
       this.isInitialized = false;
       this.initializationPromise = null;
```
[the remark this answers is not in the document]
illume: Addressed. Confirmed settings updates now wait for an in-flight initialization before saving, closing, and resetting the client. Initialization also keeps the new client local until `getTools()` finishes, so partially initialized state is never published. A regression test blocks the first discovery and verifies the update waits, closes it, and reconnects with the new settings.

app/electron/mcp/MCPClient.test.ts
```diff
@@ -142,38 +142,160 @@ describe('MCPClient', () => {
...
+    }
+  );
+
+  it('records an empty cluster context without starting MCP', async () => {
+    await client.initialize();
+
+    await client.handleClustersChange(null);
+
+    expect((client as any).clusters).toEqual([]);
+    expect((client as any).client).toBeNull();
```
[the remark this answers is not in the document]
illume: Addressed. `mcpClusterChange()` now always forwards the event to `handleClustersChange()`, mapping a null cluster to a cleared context. The regression test exercises this IPC-facing method and verifies both `currentClusters` and the environment cluster list are cleared without starting MCP.

backend/pkg/k8cache/cacheInvalidation.go
```diff
@@ -340,15 +362,12 @@ func RunInformerToWatch(gvrList []schema.GroupVersionResource,
...
+	if tombstone, ok := obj.(watchCache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
 	}
 
-	if !ok {
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
 		logger.Log(logger.LevelError, nil, nil, "unexpected object type")
 		return
 	}
```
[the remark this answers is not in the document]
illume: Addressed. The cache invalidation handler now passes the concrete `meta.Accessor` error to `logger.Log`, preserving the object type/accessor failure while retaining the existing message.

app/electron/mcp/MCPClient.ts
```diff
@@ -235,14 +259,27 @@ export default class MCPClient {
...
-      console.log('No cluster-dependent MCP servers found, skipping restart');
-      return;
-    }
+    this.clusters = newClusters || [];
 
     try {
+      // An initialization already captured the previous cluster context. Wait
+      // for it to finish before deciding whether an active client needs restart.
+      if (this.initializationPromise) {
+        await this.initializationPromise;
```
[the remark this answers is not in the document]
illume: Addressed. The full cluster-change operation now runs through a promise queue, so each event waits for the previous context update, close, and replacement initialization to complete. A regression test holds the first replacement discovery open, sends two concurrent changes, and verifies only one replacement exists at a time before the final context wins.

app/electron/mcp/MCPClient.ts
```diff
@@ -158,7 +182,9 @@ export default class MCPClient {
         defaultToolTimeout: 2 * 60 * 1000, // 2 minutes
       });
       // Get and cache the tools
-      this.clientTools = await this.client.getTools();
+      const clientTools = await client.getTools();
+      this.client = client;
+      this.clientTools = clientTools;
```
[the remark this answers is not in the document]
illume: Addressed. `cleanup()` now captures and awaits any in-flight first-use initialization before closing and clearing client state. Initialization failures remain logged by the initialization path while cleanup still resets state. A regression test blocks `getTools()`, starts cleanup, verifies cleanup remains pending, then confirms the produced client is closed and cannot remain published after shutdown.

app/electron/mcp/MCPClient.ts
```diff
@@ -217,9 +253,22 @@ export default class MCPClient {
...
    * Handle clusters change notification.
    *
-   * @param clusters - The new active clusters array, or null if none.
+   * @param newClusters - The new active clusters array, or null if none.
    */
   async handleClustersChange(newClusters: string[] | null): Promise<void> {
+    const clusterChange = this.clusterChangePromise.then(() =>
+      this.applyClustersChange(newClusters)
+    );
+    this.clusterChangePromise = clusterChange.catch(() => {});
```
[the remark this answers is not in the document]
illume: Addressed. Replaced the cluster-only queue with one generic lifecycle mutation queue. Cluster changes, cleanup, confirmed resets, and confirmed settings updates now serialize their complete close/reset/reinitialize sequences; `initializeClient()` remains deduplicated underneath. A regression test starts cleanup during a blocked cluster restart and verifies cleanup waits for and closes the replacement client instead of allowing it to publish afterward.

app/electron/main.ts
```diff
@@ -192,7 +189,7 @@ if ('remote-debugging-port' in args) {
 }
 
 const isHeadlessMode = args.headless === true;
-let disableGPU = args['disable-gpu'] === true;
+const disableGPU = args['disable-gpu'];
```
gambtho, line 192: I think this changes behavior unintentionally for Linux/arm64.

With the old `=== true`, `disableGPU` was always a boolean, so the `disableGPU === undefined` guard made the Linux-on-ARM auto-disable dead code — it never fired. With the raw yargs value (`undefined` when the flag is absent) it now does, so hardware acceleration gets disabled for every Linux/arm64 user on upgrade.

The test name `preserves Linux ARM detection` suggests this path was assumed to already work. Since the change is user-visible and isn't in the PR's change list, worth either calling out explicitly or restoring the previous effective behavior.

vyncent-t (commented):
Tried it with the loadable example config and a demo config, everything is still usable and loads on first interaction, looks okay from what I can tell

app/electron/main.ts
illume: Good catch. Probably GPU support on linux+arm+chromium improved in the last 3 years too.

I'll remove this branch to keep linux+arm as is.

illume: Addressed. Removed the Linux ARM auto-disable branch entirely, along with the now-unused architecture parameter. Linux ARM now keeps hardware acceleration enabled unless `--disable-gpu=true` is explicitly supplied. Updated the regression test to assert this effective behavior; Windows VM detection and explicit flag overrides remain unchanged.

illume:
Addressed review comment from @gambtho, "Removed the Linux ARM auto-disable branch" so it works as before.

app/electron/hardwareAcceleration.ts
```diff
@@ -0,0 +1,92 @@
...
+): string | undefined {
+  if (disableGPU !== undefined) {
+    return disableGPU ? 'related flag is set.' : undefined;
+  }
+
+  // Windows VMs normally expose only a virtual display adapter, so avoid the
+  // extra Electron GPU process and its memory unless the user explicitly enables it.
+  if (currentPlatform === 'win32' && detectWindowsVM()) {
+    return 'running in a Windows virtual machine (use --disable-gpu=false to force it if needed).';
+  }
```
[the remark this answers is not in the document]
illume: No, this was noted in previous comment. This wasn't actually disabled, due to a bug.

app/electron/hardwareAcceleration.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Reads Windows BIOS metadata used to detect virtual machines.
+ *
+ * @returns Registry output containing the system BIOS metadata.
+ */
+function readWindowsBIOS(): string {
+  if (!process.env.SystemRoot) {
+    return '';
+  }
+
+  return execFileSync(
```
ashu8912: `app.disableHardwareAcceleration()` must be called before app-ready, so this detection
cannot be deferred — meaning every Windows launch now blocks the main thread on an
`execFileSync` of reg.exe (1s timeout) before the window can appear.

In a PR whose goal is faster/leaner startup, adding a blocking subprocess spawn to startup
on the most common platform is worth justifying with a measurement. The docs table covers
backend memory thoroughly but doesn't measure this.

Minor, same function: the binary path is built from `process.env.SystemRoot`. Anyone who
controls that env var already has code execution, so it's not a real vector, but
`path.join(process.env.SystemRoot ?? 'C:\\Windows', ...)` with a basic sanity check would
be tidier than trusting it outright.

frontend/src/lib/k8s/useClustersConf.ts
```diff
@@ -0,0 +1,43 @@
...
+ * when the user changes the configuration. The configuration is stored in
+ * the local storage. When stateless clusters are present, it combines the
+ * stateless clusters with the clusters from the redux store.
+ * The shallow merge reuses nested values instead of deep-cloning every cluster,
+ * avoiding duplicate long-lived configuration object graphs.
+ * @returns the clusters configuration.
+ * */
+export function useClustersConf(): ConfigState['allClusters'] {
+  const state = useTypedSelector(state => state.config);
+  return useMemo(
```
ashu8912: The old hook returned `_.cloneDeep` copies; the new one spreads, so callers now receive
references into Redux state. `useClustersConf` is re-exported from
`frontend/src/plugin/lib.ts`, so this is a contract change for plugins, not just internal
callers — and it's invisible in the diff.

Immer freezes the produced state, so a mutating caller gets a TypeError rather than silent
corruption, which limits the damage. But it turns "works today" into "throws at runtime"
for any plugin that mutated the result.

Did you audit the ~20 in-repo consumers? If so, worth saying in the commit body; the
change deserves a line in the plugin changelog either way.

Separately, and worth calling out as a *fix*: the old deps array was
`[state.clusters === null, Object.keys(allClusters).join(',')]`, which only tracked the
set of cluster *names* — a change to an existing cluster's contents returned a stale
memoised object. The new deps fix that. I'd mention it in the commit message, since it's a
behaviour improvement hiding inside a memory change.

backend/pkg/k8cache/cacheInvalidation.go
```diff
@@ -308,6 +308,12 @@ func RunInformerToWatch(gvrList []schema.GroupVersionResource,
 ) {
 	for _, gvr := range gvrList {
 		informer := factory.ForResource(gvr).Informer()
+		// Invalidation only needs object identity. Dropping specs and status before
+		// informer storage avoids retaining full Kubernetes objects for every watch.
+		if err := informer.SetTransform(metadataOnly); err != nil {
```
ashu8912: if err := informer.SetTransform(metadataOnly); err != nil {
        logger.Log(...)
        continue          // ← skips AddEventHandler entirely
    }

On error this skips registering the event handler for that GVR, so cache invalidation stops
for that resource and Headlamp serves stale data indefinitely — with only a log line.

The ordering is currently correct (SetTransform runs before factory.Start), so this is a
defensive path only. But the failure mode is inverted: falling through *without* the
transform costs memory, whereas `continue` costs correctness. I'd drop the `continue` and
keep watching.

The rest of this change is good — switching to `meta.Accessor` and unwrapping the tombstone
first is both necessary (cached objects are now `*PartialObjectMetadata`) and more robust
than the old `*unstructured.Unstructured` assertion.

frontend/src/lib/k8s/useClustersConf.ts
illume: Yeah, I think I'll remove this change. It's not a massive improvement anyway for most people.

backend/pkg/k8cache/cacheInvalidation.go
illume: I'll remove this change.

app/electron/hardwareAcceleration.ts
illume: Thanks. I'll look into doing this in a faster way.

illume: Addressed. The registry query now starts in a worker as soon as argument parsing completes, while the main process continues synchronous setup. Immediately before registering the `ready` handler, the main thread performs a bounded wait and calls `disableHardwareAcceleration()` only for a detected VM. Explicit `--disable-gpu` values skip detection. I also researched the in-process alternatives: Node CPU strings do not expose the CPUID hypervisor bit or SMBIOS model reliably, and Electron GPU information arrives after the point where disabling acceleration is allowed. Tests cover worker startup, skip paths, timeout/failure handling, shared-state results, and execution of the real inline worker script.

frontend/src/lib/k8s/useClustersConf.ts
illume: Addressed by removing this change entirely. `useClustersConf` and its tests now match current main exactly, preserving the deep-cloned plugin-facing contract and existing memoization behavior. The unrelated lazy renderer changes remain.

backend/pkg/k8cache/cacheInvalidation.go
illume: Addressed by removing the metadata-only informer change and its regression test entirely. Both k8scache files now match current main exactly, so the existing invalidation behavior and failure semantics are unchanged.

illume:
Thanks @ashu8912 for the review on the resource reduction PR... It does the check-are-we-on-windows-vm more quickly now. Also removed those two changes you found issues with... the improvements aren't worth the risk ATM. cheers.

Should be good to go.

vyncent-t (approved):
Tried on my end again, looks good

PR #6112, opened 2026-06-20
frontend: plugins: Only fetch the active locale for plugin i18n

r0hansaxena:
## Summary

Fixes a flood of 404 requests for plugin translation files. Plugin i18n eagerly fetched a `translation.json` for every locale a plugin declares, so a plugin declaring many locales whose files aren't present produced a 404 per missing locale on every render, enough to trip rate limiters like CrowdSec. Reported for the Prometheus plugin, but the fix applies to all plugins.

## Related Issue

Fixes #4854

## Changes

- Plugin i18n now eagerly loads only the active language (plus the `en` fallback) instead of every declared locale
- Other locales are loaded lazily on language switch
- Plugins that declare no locales are skipped entirely (no probing)
- Added unit tests for the above

## Steps to Test

1. Use a plugin that declares multiple locales but ships no translations
2. Check the Network tab filtered by the plugin name
3. Before: one 404 per declared locale. After: only the active language is fetched

## Screenshots (if applicable)

Before:

image

After:

image

## Notes for the Reviewer

- Only changes how the frontend loads plugin translations. If the translation files are missing from the build itself, that's a separate packaging fix.

PR #7363, opened 2026-08-20
frontend: plugins: Only fetch the active locale for plugin i18n

### LINKED ISSUE #4854: Helm deployment is requesting missing translations
## Describe the bug

The in cluster (Helm) deployment of headlamp is requesting a lot of translations for the prometheus plugin that are not part of the image. This results in lots of 404 errors and in my case causes crowdsec to block me for scanning.

## To Reproduce
Steps to reproduce the bug:
1. Install headlamp with Helm
2. Load the main page
3. See the 404 requests that the browser did (eg. https://<..>/static-plugins/prometheus/locales/en/translation.json)

 - Installation type: In Cluster Helm
 - Headlamp Version: 0.40.0

No

YousufFFFF:
## Summary

Rebase of #6112 (by @r0hansaxena) onto current `main`, with the merge conflicts resolved. Original commits and authorship are preserved — opened from my fork because #6112's branch lives on a fork I can't push to.

Fixes a flood of 404 requests for plugin translation files. Plugin i18n eagerly fetched a `translation.json` for every locale a plugin declares, so a plugin declaring many locales whose files aren't present produced a 404 per missing locale on every render, enough to trip rate limiters like CrowdSec. Reported for the Prometheus plugin, but the fix applies to all plugins.

## Related Issue

Fixes #4854

Supersedes #6112.

## Changes

- Plugin i18n now eagerly loads only the active language (plus the `en` fallback) instead of every declared locale
- Other locales are loaded lazily on language switch
- Plugins that declare no locales are skipped entirely (no probing)
- Test files are excluded from i18n extraction, so the new tests don't generate stray plugin-namespace locale files
- Added unit tests for the above

## Conflict resolution

The conflicts came from two commits that landed on `main` after #6112 was opened:

- `81e3306cb` — *frontend: i18n: Split pt locale into pt-PT and pt-BR*
- `7198c628b` — *frontend: plugin: Scope defaults to shipped plugins*

Resolved as follows:

- **Hardcoded fallback locale list** (`pluginI18n.ts`): `81e3306cb` had just updated this list to use `pt-PT`/`pt-BR`. This PR deletes the list entirely — not probing undeclared locales is the fix — so the list is gone and that update is moot here.
- **`resolvedLanguage`** (`pluginI18n.ts`): `81e3306cb` introduced `i18next.resolvedLanguage || i18next.language` in the spots this PR rewrote. Carried it into the rewritten code — both the initial `currentLanguage` and the language-sync effect now use `resolvedLanguage` first, and the effect keeps upstream's `resolvedLanguage` dependency-array entry.
- **`PluginPackageHeadlampConfig`** (`pluginI18n.ts`): the type import from `7198c628b` is kept as-is.
- **`package.json`**: adjacent-line conflict only. Kept this PR's `i18n` script change and `main`'s `"tsc": "tsgo && tsgo -p scripts/tsconfig.json"`.

## Steps to Test

1. Use a plugin that declares multiple locales but ships no translations
2. Check the Network tab filtered by the plugin name
3. Before: one 404 per declared locale. After: only the active language is fetched
4. Switch languages and confirm the newly selected locale is fetched on demand

## Notes for the Reviewer

- Credit for the fix goes to @r0hansaxena; I only rebased and resolved conflicts.
- Verified after rebase: 89/89 tests in `src/plugin/` pass (including the 3 new ones), `tsgo --noEmit` clean, eslint and prettier clean.
- `changePluginLanguage` is now `async` and is still called fire-and-forget from the `languageChanged` listener in `plugin/index.ts`. Worth a look if you'd rather it caught its own rejections.
- Only changes how the frontend loads plugin translations. If the translation files are missing from the build itself, that's a separate packaging fix.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -124,7 +124,7 @@
     "storybook": "storybook dev -p 6006",
     "build-typedoc": "typedoc",
     "build-storybook": "storybook build -o ../docs/development/storybook",
-    "i18n": "i18next 'src/**/*.{ts,tsx}' -c ./src/i18n/i18next-parser.config.js",
+    "i18n": "i18next 'src/**/*.{ts,tsx}' '!src/**/*.test.{ts,tsx}' -c ./src/i18n/i18next-parser.config.js",
     "tsc": "tsgo && tsgo -p scripts/tsconfig.json",
     "make-version": "node ./make-env.js",
     "star": "cross-env REACT_APP_HEADLAMP_BACKEND_TOKEN=headlamp rsbuild dev",
--- a/frontend/src/plugin/pluginI18n.test.ts
+++ b/frontend/src/plugin/pluginI18n.test.ts
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { changePluginLanguage, initializePluginI18n, useTranslation } from './pluginI18n';
+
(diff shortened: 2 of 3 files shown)


YousufFFFF:
@illume Should I fix the copilot comments?

illume:
@YousufFFFF if you're up for it, that would be appreciated. If not, no stress... I can take them on. Let me know?

YousufFFFF:
> @YousufFFFF if you're up for it, that would be appreciated. If not, no stress... I can take them on. Let me know?

I have resolved the comments and pushed the changes, Thankyou!

YousufFFFF:
@illume The copilot review gave no more comments.

YousufFFFF:
PR is ready for your review!
Thankyou!

illume (approved):
🎉 thanks!

PR #7367, opened 2026-08-20
backend: telemetry: Fix data race in hijack test

### LINKED ISSUE #7366: backend: Data race in TestResponseWriterHijack_SucceedsWithUnderlyingHijacker
## Description

`TestResponseWriterHijack_SucceedsWithUnderlyingHijacker` in `backend/pkg/telemetry/metrics_test.go` has a data race, and it is the only one in the backend.

The test starts an `httptest.Server`, so its handler runs on a server goroutine. The handler writes `called = true`, and the test goroutine reads `called` in the final assertion. Because the handler hijacks and closes the connection, the client request returns without establishing any happens-before relationship with that write, so the read is unsynchronised.

Two problems:

1. **Data race.** `go test -race ./pkg/telemetry/` reports a race between the write in the handler and the read in the assertion.
2. **`require` from a non-test goroutine.** The handler calls `require.True` and `require.NoError`, which call `t.FailNow()`. The `testing` docs state `FailNow` must be called from the goroutine running the test, not from goroutines created during it.

Both mean the assertion can also pass or fail nondeterministically without the race detector.

`make backend-test` runs `go test -v -p 1 ./...` without `-race`, which is why CI does not currently surface it.

## Steps to reproduce

```
cd backend
go test -race -count=1 -run TestResponseWriterHijack_SucceedsWithUnderlyingHijacker ./pkg/telemetry/
```

```
WARNING: DATA RACE
Read at 0x... by goroutine 26:
  ...TestResponseWriterHijack_SucceedsWithUnderlyingHijacker()
      pkg/telemetry/metrics_test.go:260
Previous write at 0x... by goroutine 29:
  ...TestResponseWriterHijack_SucceedsWithUnderlyingHijacker.func2()
      pkg/telemetry/metrics_test.go:236
--- FAIL: TestResponseWriterHijack_SucceedsWithUnderlyingHijacker
    testing.go:1865: race detected during execution of test
```

## Expected

The handler's result is published to the test goroutine through a synchronising primitive, and the assertions run on the test goroutine, so the test is race-free and deterministic.

RajPrakash681:
## Summary

`TestResponseWriterHijack_SucceedsWithUnderlyingHijacker` races on the `called` flag: the handler runs on an `httptest.Server` goroutine and writes it, while the test goroutine reads it in the final assertion. The handler hijacks and closes the connection, so the client request returns without any happens-before edge to that write.

The handler also calls `require.True` and `require.NoError`, which call `t.FailNow()` from a non-test goroutine. The `testing` docs require `FailNow` to be called from the goroutine running the test.

This was the only data race in the backend.

## Related Issue

Fixes #7366

## Changes

- Publish the handler's result to the test goroutine over a buffered channel instead of a shared bool, which supplies the missing happens-before edge.
- Move the assertions onto the test goroutine, so `require`/`assert` are no longer called from the server goroutine.
- Fail with a timeout if the handler never runs, rather than asserting on an unset value.

## Steps to Test

1. `cd backend && go test -race -count=1 -run TestResponseWriterHijack ./pkg/telemetry/`
2. Confirm it passes. On `main` the same command reports `WARNING: DATA RACE` and fails.
3. Whole backend is race-clean: `cd backend && go test -race -p 1 -count=1 ./...`

### DIFF
--- a/backend/pkg/telemetry/metrics_test.go
+++ b/backend/pkg/telemetry/metrics_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	tel "github.com/kubernetes-sigs/headlamp/backend/pkg/telemetry"
 	"github.com/stretchr/testify/assert"
@@ -224,19 +225,23 @@ func TestResponseWriterHijack_SucceedsWithUnderlyingHijacker(t *testing.T) {
 	metrics, err := tel.NewMetrics()
 	require.NoError(t, err)
 
-	called := false
+	type hijackOutcome struct {
+		isHijacker bool
+		hijackErr  error
+	}
+
+	outcome := make(chan hijackOutcome, 1)
 
 	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		hj, ok := w.(http.Hijacker)
-		require.True(t, ok, "wrapped writer should implement http.Hijacker")
+		if !ok {
+			outcome <- hijackOutcome{}
 
-		conn, rw, err := hj.Hijack()
-		require.NoError(t, err, "expected Hijack to succeed when underlying supports it")
+			return
+		}
 
-		called = true
-
-		// rw is intentionally unused; we just ensure Hijack succeeded.
-		_ = rw
+		conn, _, hijackErr := hj.Hijack()
+		outcome <- hijackOutcome{isHijacker: true, hijackErr: hijackErr}
 
 		if conn != nil {
 			_ = conn.Close()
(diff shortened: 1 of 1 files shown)


9p-tech:
Tried this locally against main. Without the patch, go test -race
./pkg/telemetry/ fails with a DATA RACE on metrics_test.go:236 vs :260.
With the patch applied the package passes, and the whole backend suite
passes under -race with no warnings.

Worth mentioning that CI never runs -race today (backend-test.yml runs
plain go test), so this test is the only thing standing between the
suite and a clean -race run. Once this lands, adding -race to that
workflow would be a one-line change.

PR #7385, opened 2026-08-21
backend: multiplexer: Read Connection fields under conn.mu

### LINKED ISSUE #7384: backend: multiplexer: Connection fields read without holding conn.mu
## Description

`Connection` in `backend/cmd/multiplexer.go` carries a `sync.RWMutex` and most accesses take it, but two reads bypass it. Both race with writes that do hold the lock, and both pairs run on different goroutines in normal operation.

### 1. `reconnect` reads `conn.Token` without the lock

`reconnect` passes `conn.Token` to `establishClusterConnection` at line 597 with no lock held. `refreshConnectionToken` writes `conn.Token` at line 809 under `conn.mu.Lock()`.

These are concurrent in production: `reconnect` is reached from `monitorConnection`, which is started as its own goroutine per connection (`go m.monitorConnection(connection)`), while `refreshConnectionToken` runs on the client read loop via `processClientMessage` → `getOrCreateConnection`. A heartbeat failure triggering a reconnect at the same moment a client message arrives with a refreshed token hits this.

`reconnect` also reads `ClusterID`, `UserID`, `Path`, `Query`, `Client` and `WSConn` unlocked. Those are only written before the connection is published, so they are not racing today, but they are the same unguarded access pattern.

### 2. `processClientMessage` reads `conn.Status.State` without the lock

Line 703 reads `conn.Status.State` with no lock. `updateStatus` writes `Status.State`, `Status.LastMsg` and `Status.Error` at lines 274-276 under `conn.mu.Lock()`, and is called from `monitorConnection` on every heartbeat failure and on close.

`closeClientConnections` and `sendDataMessage` already take `conn.mu` before touching `Status`, so these two reads are the outliers.

## Steps to reproduce

Both are caught by the race detector. Adding a test that drives each pair concurrently:

```
cd backend
go test -race -count=25 -run "TestReconnectDoesNotRace|TestProcessClientMessageDoesNotRace" ./cmd/
```

On `main` this reports:

```
WARNING: DATA RACE
Write at 0x... by goroutine 38:
  ...(*Multiplexer).refreshConnectionToken()
      backend/cmd/multiplexer.go:809
Previous read at 0x... by goroutine 37:
  ...(*Multiplexer).reconnect()
      backend/cmd/multiplexer.go:597
```

```
WARNING: DATA RACE
Read at 0x... by goroutine 29:
  ...(*Multiplexer).processClientMessage()
      backend/cmd/multiplexer.go:703
Previous write at 0x... by goroutine 30:
  ...(*Connection).updateStatus()
      backend/cmd/multiplexer.go:274
```

`make backend-test` runs `go test -v -p 1 ./...` without `-race`, which is why CI does not surface these.

## Expected

Every read of a mutex-protected `Connection` field takes `conn.mu`, so the multiplexer is race-free under `go test -race`.

RajPrakash681:
## Summary

`Connection` carries a `sync.RWMutex` and most accesses take it, but two reads bypass it and race with writes that hold it. Both pairs run on different goroutines in normal operation, and both are reported by the race detector.

- `reconnect` reads `conn.Token` (line 597) while `refreshConnectionToken` writes it under `conn.mu.Lock()` (line 809). `reconnect` is reached from the per-connection `monitorConnection` goroutine; `refreshConnectionToken` runs on the client read loop.
- `processClientMessage` reads `conn.Status.State` (line 703) while `updateStatus` writes `Status` under `conn.mu.Lock()`, called from `monitorConnection` on heartbeat failure and on close.

`closeClientConnections` and `sendDataMessage` already lock before touching `Status`, so these two reads were the outliers.

## Related Issue

Fixes #7384

## Changes

- `reconnect` snapshots `ClusterID`, `UserID`, `Path`, `Query`, `Client`, `Token` and `WSConn` under `conn.mu.RLock()`, releases the lock, then closes the old socket and establishes the new connection from the snapshot. The lock is not held across `establishClusterConnection`.
- `processClientMessage` reads `Status.State` under `conn.mu.RLock()` before acting on it.
- Added two regression tests that drive each pair concurrently.

## Steps to Test

1. `cd backend && go test -race -count=25 -run "TestReconnectDoesNotRace|TestProcessClientMessageDoesNotRace" ./cmd/`
2. Confirm it passes. On `main` the same command reports `WARNING: DATA RACE` for both.
3. Whole package: `cd backend && go test -race -count=1 ./cmd/`

### DIFF
--- a/backend/cmd/multiplexer.go
+++ b/backend/cmd/multiplexer.go
@@ -584,26 +584,31 @@ func (m *Multiplexer) reconnect(conn *Connection) (*Connection, error) {
 		return nil, fmt.Errorf("cannot reconnect closed connection")
 	}
 
-	if conn.WSConn != nil {
-		_ = conn.WSConn.Close()
+	conn.mu.RLock()
+	clusterID, userID, path, query := conn.ClusterID, conn.UserID, conn.Path, conn.Query
+	client, token, wsConn := conn.Client, conn.Token, conn.WSConn
+	conn.mu.RUnlock()
+
+	if wsConn != nil {
+		_ = wsConn.Close()
 	}
 
 	newConn, err := m.establishClusterConnection(
-		conn.ClusterID,
-		conn.UserID,
-		conn.Path,
-		conn.Query,
-		conn.Client,
-		conn.Token,
+		clusterID,
+		userID,
+		path,
+		query,
+		client,
+		token,
 	)
 	if err != nil {
-		logger.Log(logger.LevelError, map[string]string{logFieldClusterID: conn.ClusterID}, err, "reconnecting to cluster")
+		logger.Log(logger.LevelError, map[string]string{logFieldClusterID: clusterID}, err, "reconnecting to cluster")
 
 		return nil, err
 	}
 
 	m.mutex.Lock()
-	m.connections[m.createConnectionKey(conn.ClusterID, conn.Path, conn.UserID)] = newConn
+	m.connections[m.createConnectionKey(clusterID, path, userID)] = newConn
 	m.mutex.Unlock()
 
 	return newConn, nil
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this PR.

PR #7387, opened 2026-08-21
frontend: advancedSearch: Keep results at exactly the item limit

### LINKED ISSUE #7386: frontend: Advanced Search drops all results at exactly the item limit, and when the limit is cleared
## Description

Advanced Search drops results it should keep, in two ways, both in the same filter in `useKubeLists`:

```ts
if (it.items && it.items.length < maxItems) {
  return it.items;
} else {
  // The amount of items exceeds the limit
  return [];
}
```

### 1. A resource holding exactly the limit is discarded

The Search settings field describes the limit as: *"If a resource has more items than this amount they will be excluded from Search."* "More than" means the boundary should be included, but `< maxItems` excludes it. With the default limit of 10,000, a resource with exactly 10,000 items contributes nothing to the search, rather than contributing all 10,000.

Note that when a resource is over the limit the whole resource is dropped rather than truncated, which is intentional per the comment. So this is not "one item missing" — it silently removes every result for that resource.

### 2. Clearing the limit field silently empties the whole search

`SearchSettings` stores the value with `parseInt(maxItemsInputRef.current.value)`. The field is a `type="number"` input, so it can be cleared, and `parseInt('')` is `NaN`. `maxItems` then becomes `NaN`, and `it.items.length <= NaN` is false for every resource, so Advanced Search returns no results at all with no indication why.

The same call also omits the radix argument.

## Steps to reproduce

Exactly at the limit:

1. Open Advanced Search and set "Max items per resource" to the number of Pods in the cluster.
2. Search. The Pods are absent, though the setting says only resources with *more* items are excluded.

NaN limit:

1. Open Advanced Search settings, clear "Max items per resource", press Save.
2. Search returns nothing for every resource.

## Expected

A resource holding exactly the limit is included, and a non-numeric limit is either rejected at the input or treated as no limit, rather than silently discarding all results.

RajPrakash681:
## Summary

Two problems in the same filter in `useKubeLists`, both of which silently remove results from Advanced Search.

A resource holding exactly `maxItems` items was excluded by `it.items.length < maxItems`, although the settings field describes the limit as *"If a resource has more items than this amount they will be excluded from Search."* Since an over-limit resource is dropped whole rather than truncated, this removed every result for that resource, not just one item.

Separately, `SearchSettings` stored the limit with `parseInt(value)` on a clearable `type="number"` input, so clearing the field and saving stored `NaN`. Every comparison against `NaN` is false, so the search returned nothing for all resources with no indication why.

## Related Issue

Fixes #7386

## Changes

- Compare with `<=` so a resource holding exactly the limit is included.
- Treat a non-finite limit as no limit in `useKubeLists`, rather than excluding everything.
- Only store the parsed values in `SearchSettings` when they are finite, and pass the radix to `parseInt`.
- Added `useKubeLists.test.tsx` covering exactly-at-limit, over-limit and a `NaN` limit.

## Steps to Test

1. `cd frontend && npx vitest run src/components/advancedSearch/`
2. Confirm 35 tests pass. On `main`, the exactly-at-limit and `NaN` cases fail.
3. Manually: open Advanced Search settings, clear "Max items per resource", save, and confirm results still appear.

### DIFF
--- a/frontend/src/components/advancedSearch/SearchSettings.tsx
+++ b/frontend/src/components/advancedSearch/SearchSettings.tsx
@@ -109,8 +109,17 @@ export function SearchSettings({
             sx={{ alignSelf: 'flex-end' }}
             onClick={() => {
               if (maxItemsInputRef.current && refetchInputRef.current) {
-                setMaxItemsPerResource(parseInt(maxItemsInputRef.current.value));
-                setRefetchIntervalMs(parseInt(refetchInputRef.current.value));
+                const maxItems = parseInt(maxItemsInputRef.current.value, 10);
+                const refetchInterval = parseInt(refetchInputRef.current.value, 10);
+
+                if (Number.isFinite(maxItems)) {
+                  setMaxItemsPerResource(maxItems);
+                }
+
+                if (Number.isFinite(refetchInterval)) {
+                  setRefetchIntervalMs(refetchInterval);
+                }
+
                 setAnchorEl(null);
               }
             }}
--- a/frontend/src/components/advancedSearch/utils/useKubeLists.test.tsx
+++ b/frontend/src/components/advancedSearch/utils/useKubeLists.test.tsx
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


illume (changes requested):
Thanks for these changes.

PR #7397, opened 2026-08-22
frontend: Resource: Fix react-hooks/exhaustive-deps

SiddharthSingh-iitbhu:
## Summary

Fix multiple `react-hooks/exhaustive-deps` warnings in `frontend/src/components/common/Resource/Resource.tsx` by providing the omitted dependencies to the React hooks and removing the lint suppressions.

## Related Issue

Refs #6894

## Changes

- **`frontend/src/components/common/Resource/Resource.tsx`:**
  - `useEffect` at line ~188: Added `dispatchHeadlampEvent` and `error` to the dependency array.
  - `useEffect` at line ~209: Added `error` and `onResourceUpdate` to the dependency array.
  - `useMemo` at line ~245: Added `backLink`, `hasPreviousRoute`, `location.state?.backLink`, and `resourceType` to the dependency array for `actualBackLink`.

## Steps to Test

1. Run `npm run frontend:lint` locally to verify that no `react-hooks/exhaustive-deps` errors are reported for `frontend/src/components/common/Resource/Resource.tsx`.
2. Run `npm start`.
3. Navigate to any cluster resource details page, such as a specific Pod or Deployment.
4. Confirm that the details load properly and that the back button functions correctly.

## Screenshots

N/A — Internal hooks refactor with no UI changes.

## Notes for the Reviewer

These changes resolve part of the suppressed `react-hooks/exhaustive-deps` violations listed in #6894 without altering functional logic or causing infinite loops.

### DIFF
--- a/frontend/src/components/common/Resource/Resource.tsx
+++ b/frontend/src/components/common/Resource/Resource.tsx
@@ -187,8 +187,7 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
         },
       });
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [item]);
+  }, [item, dispatchHeadlampEvent, error]);
 
   React.useEffect(() => {
     // We cannot call this callback more than once on each version of the item, in order to avoid
@@ -208,8 +207,7 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
       error,
     };
     onResourceUpdate?.(item as InstanceType<T>, error!);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [item, error]);
+  }, [item, error, onResourceUpdate]);
 
   const actualBackLink: string | Location | undefined = React.useMemo(() => {
     // No back link in side panels
@@ -249,8 +247,7 @@ export function DetailsGrid<T extends KubeObjectClass>(props: DetailsGridProps<T
     }
 
     return createRouteURL(route);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [item, isInPanel]);
+  }, [item, backLink, hasPreviousRoute, isInPanel, location.state?.backLink, resourceType]);
 
   const sections: (DetailsViewSection | ReactNode)[] = [];
 
--- a/frontend/src/components/common/Resource/ResourceTable.tsx
+++ b/frontend/src/components/common/Resource/ResourceTable.tsx
@@ -207,19 +207,14 @@ function TableFromResourceClass<KubeClass extends KubeObjectClass>(
   // throttle the update of the table to once per second
   const throttledItems = useThrottle(items, 1000);
   const dispatchHeadlampEvent = useEventCallback(HeadlampEventType.LIST_VIEW);
-  const dispatchHeadlampEventRef = useRef(dispatchHeadlampEvent);
 
   useEffect(() => {
-    dispatchHeadlampEventRef.current = dispatchHeadlampEvent;
-  }, [dispatchHeadlampEvent]);
-
-  useEffect(() => {
-    dispatchHeadlampEventRef.current({
+    dispatchHeadlampEvent({
       resources: items ?? [],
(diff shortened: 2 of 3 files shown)


unlikelyzero (changes requested):
Some things I observed:

1. `frontend/src/components/common/Resource/Resource.tsx:190` — adding `dispatchHeadlampEvent` to this effect's dependency array causes duplicate `DETAILS_VIEW` event dispatches on every unrelated re-render. `useEventCallback()` (`frontend/src/redux/headlampEventSlice.ts`) returns a brand-new closure on every render, with no `useCallback`/`useMemo`, so its identity changes even when `item` and `error` haven't. This effect has no idempotence guard (unlike the one twelve lines below it), so on a `withEvents` page, `useObjectEvents` updating state after its fetch is enough to trigger a second dispatch for the same resource load, and every subsequent unrelated re-render triggers another.
2. This exact bug was already hit and fixed in this codebase for the sibling `LIST_VIEW` event: commit `113aedf63` ("frontend: ResourceTable: Prevent duplicate list-view event dispatches") applies a ref-latch pattern in `ResourceTable.tsx:209-222` that keeps the effect's dependency array honest with no `eslint-disable` needed. This PR should apply the same pattern here, or memoize `useEventCallback` itself at the source (`headlampEventSlice.ts`, keyed on `[dispatch, eventType]`) so all ~40 call sites get a stable reference — the latter is more durable given more `exhaustive-deps` suppressions are queued behind this one per the PR description.
3. The PR body says `Fixes #6894`, but #6894 appears to be an unrelated issue (also flagged in an inline bot comment) and, per this PR's own description, is an umbrella issue this PR only partially addresses — `Fixes` will incorrectly auto-close it on merge. `Part of #6894` or `Refs #6894` would be more accurate.

SiddharthSingh-iitbhu:
> Some things I observed:
> 
> 1. `frontend/src/components/common/Resource/Resource.tsx:190` — adding `dispatchHeadlampEvent` to this effect's dependency array causes duplicate `DETAILS_VIEW` event dispatches on every unrelated re-render. `useEventCallback()` (`frontend/src/redux/headlampEventSlice.ts`) returns a brand-new closure on every render, with no `useCallback`/`useMemo`, so its identity changes even when `item` and `error` haven't. This effect has no idempotence guard (unlike the one twelve lines below it), so on a `withEvents` page, `useObjectEvents` updating state after its fetch is enough to trigger a second dispatch for the same resource load, and every subsequent unrelated re-render triggers another.
> 2. This exact bug was already hit and fixed in this codebase for the sibling `LIST_VIEW` event: commit `113aedf63` ("frontend: ResourceTable: Prevent duplicate list-view event dispatches") applies a ref-latch pattern in `ResourceTable.tsx:209-222` that keeps the effect's dependency array honest with no `eslint-disable` needed. This PR should apply the same pattern here, or memoize `useEventCallback` itself at the source (`headlampEventSlice.ts`, keyed on `[dispatch, eventType]`) so all ~40 call sites get a stable reference — the latter is more durable given more `exhaustive-deps` suppressions are queued behind this one per the PR description.
> 3. The PR body says `Fixes #6894`, but #6894 appears to be an unrelated issue (also flagged in an inline bot comment) and, per this PR's own description, is an umbrella issue this PR only partially addresses — `Fixes` will incorrectly auto-close it on merge. `Part of #6894` or `Refs #6894` would be more accurate.

Thanks for catching this and for the detailed explanation! 

I completely agree that fixing it at the source is the more durable approach. I've updated the PR to:
1. **Memoize `useEventCallback`** inside `headlampEventSlice.ts` (using `useMemo` keyed on `[dispatch, eventType]`) so it now provides a stable reference to all its ~40 call sites.
2. **Clean up `ResourceTable.tsx`** by removing the old ref-latch workaround, since it's no longer necessary with `useEventCallback` properly memoized.
3. **Update the PR description** from `Fixes #6894` to `Part of #6894` so it doesn't incorrectly auto-close the umbrella issue.

Let me know if everything looks good to you now!

illume (changes requested):
Thanks for working on this.

PR #7399, opened 2026-08-22
frontend: useKubeObject: Keep the watched cluster on live updates

### LINKED ISSUE #7023: frontend: useKubeObject: Preserve cluster when handling WebSocket updates

Ralthos:
Closes #7023.

`useKubeObject` builds objects with the cluster it was called with in its query function:

```ts
return new kubeObjectClass(obj, cluster) as Instance;
```

The watch handler a few lines below does not:

```ts
client.setQueryData(queryKey, new kubeObjectClass(update.object));
```

`KubeObject`'s constructor treats the missing argument as "use whatever cluster the user is
looking at":

```ts
this._clusterName = cluster || getCluster() || '';
```

and `getCluster()` reads the cluster out of the current URL, keeping only the first entry of a
multi-cluster path:

```ts
if (clusterString.includes('+')) {
  return clusterString.split('+')[0];
}
```

So on `/c/prod+staging/...`, an object being watched on `staging` is rebound to `prod` the
moment a `MODIFIED` event arrives. For a Pod that is usually seconds, since status and
conditions change constantly.

### Why the wrong cluster name matters

`_clusterName` is the cluster that `delete`, `update`, `patchUpdate` and `scale` address:

```ts
return this._class().apiEndpoint.delete(...args, params, this._clusterName);
```

and the details header passes the watched object straight to the action buttons:

```tsx
<Action item={resource} />
```

`DeleteButton` then binds `delete` to that object, so the request follows `_clusterName` rather
than the cluster the page was opened for:

```ts
let callback = item!.delete;
...
clusterAction(callback.bind(item), { ... })
```

The resource map is where the two can differ. It discovers across every selected cluster
(`apiDiscovery([...selectedClusters])`) and opens details for whichever cluster the clicked node
belongs to, passing it down explicitly:

```tsx
<Component name={name} namespace={namespace} cluster={cluster} />
```

That reaches `DetailsGrid`, which calls `useGet(name, namespace, { cluster })`. With two
clusters selected and the resource living in the second one, deleting from that panel can be
sent to the first.

A plain details page under a single-cluster URL is unaffected, because there `getCluster()`
happens to return the same cluster the hook was given. `EditButton` is also unaffected on the
write path: it watches with `cluster: item.cluster` but saves through the original `item`, so
only what it renders is affected.

I have read this through the code rather than reproduced it against two live clusters, so the
end-to-end claim is a trace, not a recording. The unit test covers the propagation directly.

### The fix

Pass `cluster` to the constructor in the watch handler, matching the query function, and add it
to the callback's dependencies.

### Testing

One test in `hooks.test.tsx`. It renders the hook against `watched-cluster`, waits for the
initial fetch, fires a `MODIFIED` event through the connection's `onMessage`, and asserts the
rebuilt object still reports `watched-cluster`. Against the old code it fails with:

```
expected undefined to be 'watched-cluster'
```

`vitest run src/lib/k8s/api/v2/` passes, 144 tests, and `src/components/common/Resource/` passes,
152 tests. eslint, prettier and `tsc --noEmit` are clean.

### DIFF
--- a/frontend/src/lib/k8s/api/v2/hooks.test.tsx
+++ b/frontend/src/lib/k8s/api/v2/hooks.test.tsx
@@ -15,7 +15,7 @@
  */
 
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { renderHook, waitFor } from '@testing-library/react';
+import { act, renderHook, waitFor } from '@testing-library/react';
 import type { ReactNode } from 'react';
 import { afterEach, beforeEach, describe, expect, it, type MockedFunction, vi } from 'vitest';
 import { ApiError } from './ApiError';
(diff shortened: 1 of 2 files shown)


Ralthos:
All four of my open PRs here have zero workflow runs. Every job sits at `action_required`,
which is the first-time-contributor gate on forks, so nothing has executed and there is
nothing for a reviewer to look at.

This one is why I am asking rather than waiting. `useKubeObject` rebuilds objects from watch
updates without passing the cluster, so `KubeObject`'s constructor falls back to
`getCluster()`, which returns only the first entry of a `prod+staging` URL. `_clusterName` is
what delete, update and patch are addressed to, and the details header hands that same object
to the delete button, so a delete issued from the map can reach a different cluster than the
one on screen.

The others are #7388, where `apiDiscovery` resolves to an empty array when every discovery
source fails and callers cannot tell that from a cluster serving no APIs; #7398, where ingress
ports are sorted as strings so 80, 443 and 8080 display as "443, 80, 8080"; and #7320, where
the plugin path containment check rejects legal directory names like `.hidden` and `..foo`.

All four have tests. They pass locally along with eslint, prettier and tsc, but that is my
word rather than CI's until the runs go through. Could a maintainer approve the workflow runs
when convenient?

PR #7403, opened 2026-08-22
backend: k8cache: Skip cache for requests with a query string

### LINKED ISSUE #7401: backend: k8cache serves one request's body to another when only the query differs
## Description

With `--cache-enabled`, the backend response cache keys entries by request **path and context only**. The query string is never part of the key, and nothing in the request path skips caching when a query is present. So two GETs to the same resource that differ only in their query share one cache entry, and whichever populates it first is served to the other for the 10 minute TTL.

`GenerateKey` builds the key from `url.Path` alone:

```go
func GenerateKey(url *url.URL, contextID string) (string, error) {
	namespace, kind := ExtractNamespace(url.Path)
	apiGroup, _, err := GetAPIGroup(url.Path)
	...
	return buildCacheKey(apiGroup, k.Kind, k.Namespace, k.Context), nil
}
```

`cacheMiddlewareHandler` only gates on the path:

```go
if !k8cache.IsKubernetesAPIPath(r.URL.Path) || k8cache.IsSelfSubjectReviewAPIPath(r.URL.Path) {
	next.ServeHTTP(w, r)
	return
}
```

This is not hypothetical for Headlamp's own frontend, which sends response-narrowing queries to list endpoints:

- `labelSelector` and `fieldSelector`
- `limit`, and `continue` for pagination — `useKubeObjectList` pages with `{...perRequestQueryParams, continue: continueToken}` against the same path

### Consequences

**Missing resources.** A filtered request populates the entry, then an unfiltered list is served the filtered body, so the UI shows fewer resources than exist, with no error.

**Wrong resources.** The reverse: a filtered request is served the unfiltered body and shows resources that do not match the filter.

**Broken pagination.** Page 1 (`?limit=100`) and page 2 (`?limit=100&continue=TOKEN`) map to the same key, so paging a large list can repeatedly return page 1.

This is a correctness problem rather than an authorization one — `IsAllowed` still runs a `SelfSubjectAccessReview` per request, so a user is only ever served data they may read. But they are served the wrong subset of it.

## Steps to reproduce

Two keys that should differ but do not:

```go
path := "/clusters/kind-kind/api/v1/namespaces/test-kube/pods"

a, _ := k8cache.GenerateKey(&url.URL{Path: path}, "kind-kind")
b, _ := k8cache.GenerateKey(&url.URL{Path: path, RawQuery: "labelSelector=app%3Dnginx"}, "kind-kind")
// a == b  -> "+pods+test-kube+kind-kind"

c, _ := k8cache.GenerateKey(&url.URL{Path: path, RawQuery: "limit=100"}, "kind-kind")
d, _ := k8cache.GenerateKey(&url.URL{Path: path, RawQuery: "limit=100&continue=TOKEN"}, "kind-kind")
// c == d
```

End to end, with `--cache-enabled`: open a list view for a resource, then load the same list with a label selector applied within the TTL. The second view is served the first body.

## Expected

A request whose query changes the response is not served another request's body.

## Note on the fix

Adding the query to the cache key looks like the natural fix, but it is not a small change. Three places parse keys with `strings.SplitN(key, "+", 4)` and require the context to be `parts[3]` (`redactCacheKey`, and two spots in `context_cleanup.go`), so a fifth segment silently corrupts context extraction. `DeleteKeys` would also need to become prefix based, since a mutation generates a query-less key and would otherwise leave every query variant stale.

Not caching query-bearing requests avoids all of that and cannot serve a wrong body. Plain list requests, which are the common case and carry no query, keep being cached. Keying on a canonicalised query plus prefix-based invalidation would be a reasonable follow-up if the extra hit rate is wanted.

RajPrakash681:
## Summary

With `--cache-enabled`, cache keys are built from the request path and context only, and the middleware gates on the path alone. Two GETs to the same resource differing only in their query therefore share one entry, and the first to populate it is served to the other for the 10 minute TTL.

Headlamp's own frontend sends response-narrowing queries to list endpoints — `labelSelector`, `fieldSelector`, `limit`, and `continue` for pagination — so this shows up as resources missing from a list, resources appearing that do not match a filter, or paging returning page 1 repeatedly.

Authorization is unaffected: `IsAllowed` still runs a `SelfSubjectAccessReview` per request, so a user is only served data they may read, just the wrong subset of it.

## Related Issue

Fixes #7401

## Changes

- Added `HasQueryParameters`, and skip the cache for requests that carry a query string, so a narrowed or paged response can never be served in place of another.
- Added a table test covering a plain list request, a label selector, a field selector and a paginated page.

Adding the query to the key was the other option, but three places parse keys with `strings.SplitN(key, "+", 4)` and require the context at `parts[3]`, so a fifth segment corrupts context extraction; `DeleteKeys` would also need to become prefix based or every query variant would be left stale after a mutation. Skipping is the smaller change and cannot serve a wrong body. Plain list requests carry no query and are still cached.

## Steps to Test

1. `cd backend && go test -count=1 -run TestHasQueryParameters ./pkg/k8cache/`
2. `cd backend && go test -count=1 ./pkg/k8cache/ ./cmd/`
3. Manually, with `--cache-enabled`: open a list view, then reload it with a label selector applied inside the TTL, and confirm the filtered view is no longer served the unfiltered body.

### DIFF
--- a/backend/cmd/headlamp_test.go
+++ b/backend/cmd/headlamp_test.go
@@ -3131,6 +3131,93 @@ func TestCacheMiddleware_CachesResourceNamedVersion(t *testing.T) {
 	assert.Equal(t, "true", resp2.Header.Get("X-HEADLAMP-CACHE"))
 }
 
+// TestCacheMiddleware_QueryBearingRequests checks the gate that decides which requests the
+// cache may handle. Cache keys are built from the path and context only, so a GET narrowed by
+// a query must bypass the cache entirely, while a mutation carrying a query must still reach
+// invalidation. An unconditional query bypass would satisfy the first case and break the second.
+func TestCacheMiddleware_QueryBearingRequests(t *testing.T) {
+	oldCache := k8sResponseCache
+	k8sResponseCache = cache.New[string]()
+
+	t.Cleanup(func() { k8sResponseCache = oldCache })
+
+	fakeK8s := newFakeK8sServer(true)
+	defer fakeK8s.Close()
+
+	c := newHeadlampConfig(fakeK8s, t.Name())
+
+	var (
+		mu   sync.Mutex
+		seen []string
+	)
+
+	proxyHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		mu.Lock()
+
+		seen = append(seen, r.Method+" "+r.URL.RawQuery)
+
+		mu.Unlock()
+
+		w.WriteHeader(http.StatusOK)
+
+		_, err := w.Write([]byte(fakeK8sResourceListResponse))
+		assert.NoError(t, err)
+	})
+
+	router := mux.NewRouter()
+	router.PathPrefix("/clusters/{clusterName}/{api:.*}").Handler(
+		CacheMiddleWare(c)(proxyHandler),
+	)
+
+	ts := httptest.NewServer(router)
+	defer ts.Close()
+
+	resourceURL := ts.URL + "/clusters/test/api/v1/namespaces/ns/configmaps"
+
+	requireQueryGetsBypassCache(t, resourceURL)
+
+	mu.Lock()
+	afterGets := len(seen)
+	mu.Unlock()
+
+	assert.Equal(t, 2, afterGets, "both query-bearing GETs must reach the proxy")
+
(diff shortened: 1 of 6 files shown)


illume (changes requested):
Thanks for these changes.

PR #7412, opened 2026-08-23
frontend: EditorDialog: Report all failures on multi-resource apply

### LINKED ISSUE #7407: frontend: EditorDialog multi-resource apply silently drops all failures after the first
## Describe the bug

When applying multiple resources at once (e.g. via Upload File/URL with multiple YAML files), `applyFunc` in `EditorDialog.tsx` uses `Promise.allSettled` but then iterates results with `forEach` and `throw`s on the first rejected entry. The `throw` aborts the loop, so only the first failure in array order is ever surfaced — remaining failures are silently dropped.

This has two concrete effects: if multiple resources fail, the user only sees the first error and discovers the rest one-at-a-time on resubmission. If one resource fails and another succeeds, the dialog stays open implying total failure while the successful resource now exists on the cluster with no indication to the user.

Expected: all failures from `allSettled` are collected and surfaced together.

Affected call sites: `CreateButton.tsx` (CreateActivityContent), `CreateResourceButton.tsx` (CreateResourceActivityContent), `ProjectDetails.tsx` (Create resource quota action).

## Environment

- Not environment-specific — reproducible on any installation type
- Headlamp Version: current main (verified against `EditorDialog.tsx:505-530`)

Yes (I will propose a PR).

## Additional Context

Root cause is `EditorDialog.tsx:505-530` — `throw new Error(msg)` inside a `.forEach()` callback aborts iteration on the first rejected result. The fix replaces the throw-inside-forEach with collecting all rejections before erroring. Single-failure behavior is preserved byte-for-byte. Existing tests pass unmodified.

beep-boopp:
## Summary

This PR fixes a bug in `EditorDialog`'s apply logic where, if you're applying more than one resource at once (e.g. uploading multiple YAML files), only the first failure ever got shown to you. If two things failed, the second one just silently vanished — you'd only find out about it after fixing the first and trying again. Worse, if one resource failed and another one actually succeeded, the dialog would still act like everything failed, even though something got created on the cluster behind the scenes.

The root cause was a `throw` inside a `.forEach()` loop — throwing there stops the loop dead, so anything after the first failure never even got looked at. The fix just collects all the failures first, then reports all of them together.

## Related Issue

Fixes #7407

## Changes

- Reworked `applyFunc` in `EditorDialog.tsx` to collect every failed resource from `Promise.allSettled` instead of bailing out on the first one
- Added a test covering the case where two resources are applied together and both fail, checking that both show up in the error
- Didn't touch anything else — single-resource apply behaves exactly the same as before (confirmed the existing tests for that still pass without changes)

## Steps to Test

1. Open the create/apply dialog and use "Upload File/URL" to load two YAML files that will both fail on apply (bad spec, missing required field, whatever triggers a server-side rejection)
2. Hit Save & Apply
3. Before this fix you'd only see one error but now you should see both failures listed

Or just run the test suite for this file, it covers the scenario directly.

## Notes for the Reviewer

- This only affects the "create/apply multiple resources at once" path,  edit and view flows use their own save logic and never went through this code, so they're untouched.
- Tried to keep this minimal on purpose,  didn't touch anything about *how* errors are formatted or displayed, just made sure none of them get dropped.

### DIFF
--- a/frontend/src/components/common/Resource/EditorDialog.test.tsx
+++ b/frontend/src/components/common/Resource/EditorDialog.test.tsx
@@ -880,6 +880,39 @@ describe('EditorDialog', () => {
     expect(screen.getByText('Failed to create Node node-1 in v1.')).toBeInTheDocument();
   });
 
+  it('reports every failure when applying multiple resources that all fail', async () => {
+    mockLoadAll.mockReturnValueOnce([
+      { apiVersion: 'v1', kind: 'Node', metadata: { name: 'node-1' } },
+      { apiVersion: 'v1', kind: 'Node', metadata: { name: 'node-2' } },
+    ]);
+    mockApply.mockRejectedValueOnce({}).mockRejectedValueOnce({});
+    renderEditorDialog();
+    fireEvent.change(screen.getByRole('textbox', { name: /code/i }), {
+      target: { value: 'multiple resources' },
+    });
+    fireEvent.click(screen.getByRole('button', { name: /save & apply/i }));
+
+    let thrown: Error | undefined;
+    await act(async () => {
+      try {
+        await capturedAction.current?.();
+      } catch (err) {
+        thrown = err as Error;
+      }
+    });
+
+    expect(thrown?.message).toContain('node-1');
+    expect(thrown?.message).toContain('node-2');
+
+    const errorParagraph = screen.getByText(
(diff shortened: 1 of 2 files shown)


beep-boopp:
@illume PTAL , no code changes , just amended a commit with accordance to copilot's concern and updated the title

Issue #7158, opened 2026-08-13
Crash Report: Cannot destructure property 'namespace' of 'e.metadata' as it is undefined.


dimaulupov:
## Crash Summary
Cannot destructure property 'namespace' of 'e.metadata' as it is undefined.

## Error Stack
```
TypeError: Cannot destructure property 'namespace' of 'e.metadata' as it is undefined.
    at ton (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/index-wMBu2Jqu.js:1211:246551)
    at kd (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:38:17652)
    at Hh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:44746)
    at Fh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:40337)
    at Vw (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:40265)
    at ic (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:40118)
    at Yd (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:36403)
    at zh (file:///Applications/Headlamp.app/Contents/Resources/frontend/assets/vendor-mui-CpvGftAG.js:40:37215)
    at Lo 
... (truncated)
```

## Headlamp Version
0.44.0
(same on 0.43.0)

## Git Commit
2f983db415644d943eaa1a4d60da94594525c566

## System Information
- User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Headlamp/0.44.0 Chrome/140.0.7339.249 Electron/38.8.6 Safari/537.36
- Platform: MacIntel
- Language: en-GB

## Additional Context
Shows this on clicking on of the clusters. Used to work previously. Maybe something that I installed on the cluster? Not sure how to debug

lovelymahor:
I'd like to try reproducing and fixing this as part of LFX mentorship prep could someone confirm the scenario, and can I take this?

dimaulupov:
[...]

PR #7421, opened 2026-08-23
frontend: apply() crashes when a manifest has no metadata block

### LINKED ISSUE #7420: frontend: apply() crashes when a manifest has no metadata block
`apply()` in `frontend/src/lib/k8s/api/v1/apply.ts` does this near the top:

```ts
const { namespace } = body.metadata;
```

If `body.metadata` is undefined this throws `TypeError: Cannot destructure property 'namespace' of 'body.metadata' as it is undefined`, and the whole apply call blows up instead of failing cleanly.

That's not a hypothetical - it's an exact match for the minified crash in #7158 ("Cannot destructure property 'namespace' of 'e.metadata' as it is undefined"). `apply()` is called from `EditorDialog`, which backs Edit, Upload, and Create-from-YAML, so any manifest missing a `metadata:` block (easy to do by hand, or copy-pasting a partial example from docs) takes down the UI instead of showing a validation error.

There's also no test file for `apply.ts` at all right now.

Planning to send a fix shortly - guard the destructure and make sure `bodyToApply.metadata` exists before line 76 writes `.namespace` onto it, plus add the missing test file.

Priyanshu-rgbb:
Fixes #7420.

`apply()` was destructuring `namespace` straight off `body.metadata` with no guard. If you apply a manifest that's missing a `metadata:` block entirely (easy to hit through Upload/Edit/Create-from-YAML - just forget the block, or paste a partial example) it throws instead of doing what the rest of the function is set up to do, which is fall back to a default namespace. Matches the crash in #7158 pretty much exactly.

Fix normalizes `bodyToApply.metadata` to `{}` right after the clone, then reads/writes off that instead of the raw `body`. Everything downstream (`namespace`, `resourceVersion`, the POST/PUT conflict fallback) was already working off `bodyToApply`, so this was really the only unguarded spot.

There wasn't a test file for `apply.ts` before this, so I added one covering: normal apply with metadata, no metadata at all, metadata with no namespace, cluster-scoped resource with no metadata (shouldn't get a namespace injected), and the 409-conflict-then-PUT path.

Ran `npm run tsc`, and lint on the touched files, all clean.

### DIFF
--- a/frontend/src/lib/k8s/api/v1/apply.test.ts
+++ b/frontend/src/lib/k8s/api/v1/apply.test.ts
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { getCluster } from '../../../cluster';
+import { apply } from './apply';
+import { getClusterDefaultNamespace } from './clusterApi';
+import { resourceDefToApiFactory } from './factories';
+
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for the contribution.


<details>

- `frontend: fix apply() crash on manifests without metadata` — Description must start with a capital letter — e.g. `frontend: HomeButton: Fix the button` not `frontend: HomeButton: fix the button`.

</details>

<details>

Good examples:

</details>

Priyanshu-rgbb:
Hey @illume Do have a look whenever you have chance.

Priyanshu-rgbb:
> Copilot reviewed 2 out of 2 changed files in this pull request and generated no new comments. 

@illume PR is ready.

PR #6220, opened 2026-06-26
Fix accessibility: decorative status icons exposed to screen readers

sushilbhattarai45:
## Summary
This PR fixes an accessibility issue where decorative status and warning icons were incorrectly using `aria-label="hidden"`, which exposes "hidden" as an accessible name instead of hiding the icon from screen readers.

Since the visual status is already displayed via the text, these icons should be treated as decorative.

## Changes
- Replaced `aria-label="hidden"` with `aria-hidden="true"` where appropriate
- Ensures icons are ignored by assistive technologies when redundant

## Testing
- Ran `npm run lint`
- Ran `npm test`
- Verified UI behavior manually in browser

## Notes
This improves accessibility compliance without affecting visual layout or functionality.

PR #7434, opened 2026-08-24
frontend: Fix decorative icons wrongly using aria-label="hidden"

### LINKED ISSUE #7433: aria-label="hidden" on decorative icons announces the word "hidden" to screen readers instead of hiding them
## Describe the bug
Several decorative status icons use `aria-label="hidden"` when the intent is clearly to hide them from assistive technology. That's not what `aria-label` does: it sets the element's *accessible name* to the literal text "hidden", so a screen reader announces the word "hidden" for these icons instead of skipping them silently. The correct attribute for "skip this decorative element" is `aria-hidden="true"`, which is already used correctly elsewhere in the codebase (e.g. `ActionButton.tsx`, `StatusCharts.tsx`), confirming this is a mistake rather than an intentional convention.

Affected locations:
- `frontend/src/components/job/List.tsx` — job condition icon
- `frontend/src/components/common/Resource/Resource.tsx` (×2) — container env-var out-of-sync icon, and container state alert icon

(`frontend/src/components/pod/List.tsx` has the same bug but is already covered by open PR #6220 — not included here to avoid duplicating that work.)

## To Reproduce
1. Open a Job list where a row has a `Complete`/`Failed`/`Suspended` condition, or a Pod details view where a container has a state message tooltip.
2. Inspect the status icon in that row, or use a screen reader to read it.
3. The icon's accessible name reads as "hidden" — the opposite of the intended behavior of removing it from the accessibility tree.

- Headlamp Version: 0.45.0
- Other: found via code review while looking at accessibility-labeled components (`frontend/src/components/job/List.tsx:87`, `frontend/src/components/common/Resource/Resource.tsx:1233,1533-1534`)

Yes (I will propose a PR).

## Additional Context
Not a regression as far as I can tell — `git log -S'aria-label="hidden"'` shows this pattern has been present since Dec 2021. Related: #6220 already fixes the same bug in `pod/List.tsx`; this issue covers the two remaining spots it doesn't touch. Fix is mechanical: replace `aria-label="hidden"` with `aria-hidden="true"` on the purely-decorative `<Icon>` and the one `<Box>` that wraps only a decorative icon; remove the attribute outright on the one `<Box>` that wraps meaningful content (a `HoverInfoLabel` with real hover text), since hiding that would suppress real information instead of just the wrapper.

garvit-bhattt:
## Summary

This PR fixes a decorative-icon accessibility bug by replacing the incorrect `aria-label="hidden"` with `aria-hidden="true"` (or removing it entirely where it wraps meaningful content) in `job/List.tsx` and `Resource.tsx`.

## Related Issue

Fixes #7433

## Changes

- `frontend/src/components/job/List.tsx`: job condition icon — `aria-label="hidden"` → `aria-hidden="true"` (purely decorative; the condition type is already conveyed by adjacent text).
- `frontend/src/components/common/Resource/Resource.tsx`:
  - Container env-var out-of-sync wrapper `Box` — removed `aria-label="hidden"` outright, since it wraps a `HoverInfoLabel` with real, meaningful hover content that must stay reachable to assistive tech.
  - Container state alert icon + its wrapper `Box` — `aria-label="hidden"` → `aria-hidden="true"` on the `Box`; removed the now-redundant duplicate `aria-label="hidden"` on the inner `Icon`. The tooltip's real message is already exposed correctly via `aria-describedby` on the sibling `StatusLabel`, so this pair is genuinely decorative.

Note: `pod/List.tsx` has the same bug but is already being fixed in #6220, so it's intentionally left out of this PR to avoid duplicating that work.

## Steps to Test

1. `npm run frontend:test` — existing unit tests (`job/Details.test.tsx`, storyshots) pass unchanged.
2. `npx vitest run src/storybook.test.tsx -t "Job/List"` — storyshot for the touched Job list story passes with no snapshot diff.
3. `npm run frontend:lint` and `npm run frontend:tsc` — both clean.
4. Manually: inspect the elements in DevTools and confirm the icons now render `aria-hidden="true"` instead of `aria-label="hidden"`.

None — this is a non-visual accessibility fix (only ARIA attributes change; no visual rendering is affected).

## Notes for the Reviewer

- See #7433 for the full writeup of why `aria-label="hidden"` is wrong here (it sets the accessible *name* to the word "hidden" rather than hiding the element).
- The fix isn't uniform across the two `Resource.tsx` spots on purpose — one wraps real content (so the attribute is just removed) and one is purely decorative (so it becomes `aria-hidden`). Happy to adjust if you'd prefer a different pattern.


unlikelyzero (changes requested):
Some things I observed:

1. `frontend/src/components/common/Resource/Resource.tsx:1533` — this trades one accessibility bug for another. That `Box` is the MUI `Tooltip` trigger for `ContainerInfo`, and MUI applies `aria-label={stateDetails.message}` to it by default (since `describeChild` isn't set here, MUI uses `aria-label`, not `aria-describedby`). Setting `aria-hidden="true"` on it removes the element and its whole subtree from the accessibility tree, so the container's error/status message becomes completely unreachable to screen readers. The box isn't focusable, so hover is the only path to it — there's no AT fallback. This should follow the same pattern as the sibling hunk in `ContainerEnvironmentVariables`: drop the attribute entirely instead of converting it to `aria-hidden`.

2. `frontend/src/components/pod/List.tsx:104` still has `aria-label="hidden"` on a structurally identical icon (same `StatusLabel`/`LightTooltip` pattern as the `job/List.tsx` case this PR fixes). The PR body attributes this to #6220 already covering it — worth confirming that PR's current diff actually touches this exact line before treating it as already scheduled, since as of this diff it's still broken.

3. `frontend/src/components/common/Resource/Resource.tsx:1235` — the `aria-label="error"` prop on `HoverInfoLabel` is dead: `HoverInfoLabelProps` doesn't declare `aria-label` and the component only destructures known props, so it's silently dropped and never reaches the DOM. Worth removing while this block is already being touched for accessibility reasons.

garvit-bhattt:
> Some things I observed:
> 
> 1. `frontend/src/components/common/Resource/Resource.tsx:1533` — this trades one accessibility bug for another. That `Box` is the MUI `Tooltip` trigger for `ContainerInfo`, and MUI applies `aria-label={stateDetails.message}` to it by default (since `describeChild` isn't set here, MUI uses `aria-label`, not `aria-describedby`). Setting `aria-hidden="true"` on it removes the element and its whole subtree from the accessibility tree, so the container's error/status message becomes completely unreachable to screen readers. The box isn't focusable, so hover is the only path to it — there's no AT fallback. This should follow the same pattern as the sibling hunk in `ContainerEnvironmentVariables`: drop the attribute entirely instead of converting it to `aria-hidden`.
> 2. `frontend/src/components/pod/List.tsx:104` still has `aria-label="hidden"` on a structurally identical icon (same `StatusLabel`/`LightTooltip` pattern as the `job/List.tsx` case this PR fixes). The PR body attributes this to #6220 already covering it — worth confirming that PR's current diff actually touches this exact line before treating it as already scheduled, since as of this diff it's still broken.
> 3. `frontend/src/components/common/Resource/Resource.tsx:1235` — the `aria-label="error"` prop on `HoverInfoLabel` is dead: `HoverInfoLabelProps` doesn't declare `aria-label` and the component only destructures known props, so it's silently dropped and never reaches the DOM. Worth removing while this block is already being touched for accessibility reasons.

Thanks for the detailed review !
Yeah these three points make sense.

For `Resource.tsx:1533`, I checked MUI’s `Tooltip` implementation to confirm the behavior. Since `describeChild` isn’t set here, the tooltip adds `aria-label={title}` to its child automatically. The `aria-hidden="true"` on the Box was preventing that from being exposed to assistive technologies, so I’ve removed it. This also keeps it consistent with the `ContainerEnvironmentVariables` change. The Iconify icon already defaults to `aria-hidden`, so there’s no need to add anything explicitly there.

For `pod/List.tsx:104`, I also checked the actual diff for #6220. It does update this exact line from `aria-label="hidden"` to `aria-hidden="true"`. That PR is still open, though, so the change hasn’t landed yet. I’ll keep it out of this PR as planned so we don’t end up with both PRs touching the same line.

And for `Resource.tsx:1235`, I confirmed that `HoverInfoLabelProps` doesn’t include `aria-label`, and because the component only passes through its known props, the value never reached the DOM anyway. I’ve removed it.

I’ve made all three fixes and I’m pushing them now.

illume (changes requested):
Thanks for this PR.


<details>

</details>

illume (changes requested):
Thanks for these changes.

PR #7465, opened 2026-08-24
app: main: Contain renderer-supplied plugin folder paths

### LINKED ISSUE #6472: app: open-plugin-folder IPC allows path traversal via folderName
## Describe the bug

The open-plugin-folder IPC handler in app/electron/main.ts takes pluginInfo.folderName straight from the renderer and joins it onto the plugins base directory with no validation:

    folderPath = path.join(defaultUserPluginsDir(), pluginInfo.folderName);
    ...
    shell.openPath(folderPath)

path.join doesn't stop `..` segments from walking out of the base directory, so a folderName like ../../../../etc resolves to somewhere completely outside the plugins folder, and shell.openPath just opens whatever it lands on.

This is worse than it sounds because shell.openPath isn't limited to showing folders. Per Electron's docs, if the resolved path is a file rather than a directory, it opens that file with the OS's default handler for it. So depending on where the traversal lands, this isn't just "browses the wrong folder," it can end up launching something.

On the reachability side: the frontend normally sends a safe, backend-derived folderName, but PluginSettingsDetails.tsx has a fallback (plugin.folderName || plugin.name.split('/').pop()) that falls back to the plugin's own package.json name field when folderName isn't set, and that field is written by whoever authored the plugin, not sanitized by Headlamp. More generally, IPC is a trust boundary in Electron, so the main process handler shouldn't trust whatever the renderer sends it regardless of which frontend code path triggers it.

Expected: the resolved path should be checked to stay within the intended plugins directory before shell.openPath is ever called on it, same as any other user-supplied path reaching the filesystem.

## To Reproduce
Steps to reproduce the bug:
1. Send an open-plugin-folder IPC message with pluginInfo.folderName set to something like ../../../../etc and type set to a valid plugin type (e.g. user)
2. The handler joins it straight onto the base plugins directory and calls shell.openPath on the result
3. The opened location is well outside the plugins folder instead of being rejected

I verified the path resolution locally with a small script mirroring the exact join logic (without ever calling shell.openPath, since that would actually open something) — a normal plugin name stays inside the base directory, a folderName of ../../../../../etc resolves clean outside of it.

 - Installation type: running from source (main branch), Electron desktop app
 - Other: desktop app only, this is in app/electron/main.ts

Yes (I will propose a PR)

## Additional Context

The spot is the open-plugin-folder handler in app/electron/main.ts around line 1747. Doesn't look like a regression, this has been unvalidated for a while.

Aamod007:
## Summary

This PR contains renderer-supplied plugin folder paths in the Electron main process: the `open-plugin-folder` IPC handler and the plugin-manager `LIST` action now validate the renderer-supplied directory before use.

## Related Issue

Fixes #6472

## Changes

- Fixed `open-plugin-folder` in `app/electron/main.ts`: `pluginInfo.folderName` is validated as a single safe path segment (rejecting empty, `.`, `..`, and any `/`/`\` separators — same rule as `validatePluginName`), so `path.join(base, folderName)` can no longer escape the user/development/shipped plugin roots.
- Applied the same containment to the `destinationFolder` passed to `PluginManager.list`, restricting it to the known plugin roots; previously it read any directory on disk and returned subdirectory names plus parsed package.json fields (name/title/author/artifacthub metadata) to the renderer.
- Added unit tests for the validation helper covering traversal attempts.

## Steps to Test

1. From renderer devtools send `desktopApi.send('open-plugin-folder', { type: 'user', folderName: '..\\..\\..\\' })`.
2. Before: a directory outside the plugins root opens in Explorer/Finder (`shell.openPath`); after: the request is rejected.
3. Same for a `plugin-manager` LIST with an arbitrary `destinationFolder`.
4. `cd app && npm run tsc && npm run lint && npm run test:unit`

## Notes for the Reviewer

- Follows the containment approach already used elsewhere in this area (`legal-documents.ts`, `setup-plugins.ts`).

### DIFF
--- a/app/electron/main.ts
+++ b/app/electron/main.ts
@@ -519,6 +519,23 @@ class PluginManagerEventListeners {
 
     // If a specific folder is requested, list only from that folder
     if (destinationFolder) {
+      // destinationFolder comes from the renderer; contain it to the known
+      // plugin roots so an arbitrary directory cannot be listed and its
+      // contents disclosed to the renderer.
+      const allowedPluginRoots = [
+        defaultPluginsDir(),
+        defaultUserPluginsDir(),
+        path.join(__dirname, '.plugins'),
+        path.join(process.resourcesPath, '.plugins'),
+      ];
+      if (!isPathInsideRoots(destinationFolder, allowedPluginRoots)) {
+        console.error('plugin-manager list: destination folder outside plugin roots', {
+          destinationFolder,
+        });
+
+        return;
+      }
+
       PluginManager.list(destinationFolder, progress => {
         event.sender.send(
           'plugin-manager',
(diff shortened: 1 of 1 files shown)


krsatyamthakur-droid:
hey @Aamod007 - #6473 already fixes the `open-plugin-folder` half of this, open since Jul 11 against #6472 which I'm assigned to. it also handles the `shell.openPath` error path (it resolves with an error string rather than rejecting, so the old `.catch()` never fired) and has unit tests for the containment helper.
the `PluginManager.list `/ `destinationFolder` sink is a real find though and it's yours - that part isn't covered anywhere yet.
could you narrow this to just the LIST containment and drop the `Fixes #6472`? happy to file a separate issue for it to reference. `plugin-management.ts` already exports `isPathWithinDirectory` with tests, which does the same job as your `isPathInsideRoots` - reusing it would avoid a second helper.

PR #7471, opened 2026-08-25
frontend: AlertNotification: Stop polling health on the auth routes

### LINKED ISSUE #7469: Cluster health banner ("Lost connection to the cluster") sticks after successful OIDC login due to /healthz being called before token is applied
## Describe the bug
On page load / redirect from `/c/main/login`, Headlamp fires a request to `/clusters/main/healthz` before the OIDC token from Keycloak has been applied to the request. This request returns `401 Unauthorized`, and the UI shows a persistent "Lost connection to the cluster. Cluster main is not healthy: Check cluster settings" banner.

The login itself then completes successfully a few seconds later (`/me`,`selfsubjectrulesreviews`, and actual resource requests all return
200/201), and subsequent `/healthz` calls also return 200 — but the error banner from the earlier failed check remains visible and does not auto-clear once the cluster is confirmed healthy.

## To Reproduce
Steps to reproduce the bug:
1. Configure Headlamp with OIDC (Keycloak realm), in-cluster mode
2. Navigate directly to a deep link (e.g. a PVC list under a namespace) while not yet authenticated, so the app redirects through `/c/main/login`
3. Complete the Keycloak login
4. Observe the "Cluster is not healthy" banner appears immediately, then the page's actual data (resource list) loads and works fine
5. The banner does not go away on its own

### HAR excerpt (timeline)
04:02:06.853Z GET /clusters/main/healthz -> 401 Unauthorized
04:02:21.950Z GET /clusters/main/healthz -> 401 Unauthorized
04:02:27.473Z GET /clusters/main/me -> 200
04:02:27.475Z POST .../selfsubjectrulesreviews -> 201
04:02:27.610Z POST .../selfsubjectaccessreviews -> 201
04:02:27.610Z GET .../namespaces -> 200
04:02:27.611Z GET .../persistentvolumeclaims -> 200
04:02:42.048Z GET /clusters/main/healthz -> 200
04:02:47.142Z GET /clusters/main/healthz -> 200

The two failed `/healthz` requests both have `referer: .../c/main/login` and carry no Authorization/Cookie header — they were sent before authentication completed.

## Expected behavior
- `/healthz` polling should either wait until the auth flow completes, or a transient 401 during the login redirect should not surface a persistent error banner.
- If a later `/healthz` check succeeds, any previously-shown "cluster not healthy" banner should clear automatically.

 - Headlamp Version: 0.44.0
 - Other: Deployed behind a TCP passthrough reverse proxy + nginx-ingress; auth via OIDC/Keycloak

No.

## Additional Context
Release 0.44.0 fixed a related crash in the cache middleware for non-API paths like `/healthz` (#5798), but this specific race between the health check and the OIDC redirect on initial page load still reproduces.

Anjali-Chauhan1:
Closes #7469

## Summary

This PR fixes the "Lost connection to the cluster." banner sticking around after a successful OIDC login, by not polling the cluster health check on the auth routes and by treating a 401/403 as an authentication state rather than a lost connection.
While the user is signed out, a deep link redirects through the cluster login route, where every `/healthz` call answers 401 until the token is applied. Those failures were recorded as a lost connection, and each one backed the poller off by another 5 seconds, so once the login redirected back the banner showed the stale failure and only cleared on the next check, which by then could be far away.

## Related Issue

Fixes #7469

## Changes

* Updated `AlertNotification` to skip the health check on the routes that cannot answer it yet (`login`, `token`, `oidcAuth`, `settingsCluster`), and to clear any pending error when entering them. The route list governed only the rendering before.
* Fixed the banner being raised by 401/403 responses, which are an auth state that `AuthRoute` already handles by redirecting to the login route. They no longer back the poller off either.
* Refactored the poller from `setInterval` to a self-scheduling `setTimeout`, so a check is only scheduled once the previous one settles and the two can no longer overlap. The backoff is now capped at 35s instead of growing without bound.
* Added an immediate check whenever the route, the cluster or the connectivity changes (`online`, tab visibility, "Try Again"), so a recovery shows up right away instead of waiting out the backoff.
* Updated `testClusterHealth` to reject with an `ApiError` carrying the response `status` and `cluster`. The message is unchanged, so the `apiProxy` surface used by plugins stays compatible.

## Steps to Test

1. Configure a cluster with OIDC auth (e.g. a Keycloak realm), in-cluster.
2. While signed out, navigate directly to a deep link, e.g. a PVC list under a namespace, so the app redirects through `/c/main/login`.
3. Complete the login.
4. Observe that no "Lost connection to the cluster." banner appears at any point, and that the resource list loads as usual. On `main`, the banner shows up as the page loads and stays until a later health check happens to succeed.
5. Check the network tab: no `/clusters/main/healthz` request is sent while the login route is open.
6. Regression check: with a genuinely unreachable cluster, the banner still appears, and "Try Again" re-checks immediately. Stopping and restarting the cluster clears the banner on its own within 35s.

## Notes for the Reviewer

* No new user-facing strings, so the i18n files are untouched.
* The 401/403 rule is deliberately a second layer of defence: route gating already covers the reported flow, but the reporter runs behind a TCP passthrough proxy where the very first check after the redirect can still answer 401 before the cookie applies. There is a test for exactly that case.
* I considered gating the poller on the `['auth', cluster]` React Query cache that `AuthRoute` populates, since that is the authoritative signal. I went with route gating instead, to avoid making a presentational component depend on route-level query keys, and because the query-based approach behaves poorly in multi-cluster mode.
* Test coverage in `AlertNotification.test.tsx` (9 tests): immediate check on mount, failure then recovery, no polling on auth routes, the reported login round trip, 401 and 403, an unauthorized first check after the redirect, the backoff cap, and non-overlapping checks. Plus a case in `apiProxy.test.ts` for the preserved status.
* Not verified: I have not run this against a live Keycloak-backed cluster, only the test suite.

### DIFF
--- a/frontend/src/components/common/AlertNotification.test.tsx
+++ b/frontend/src/components/common/AlertNotification.test.tsx
@@ -14,9 +14,13 @@
  * limitations under the License.
  */
 
-import { act, cleanup, render } from '@testing-library/react';
+import { act, cleanup, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { Provider } from 'react-redux';
+import { MemoryRouter, useHistory } from 'react-router-dom';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { TestContext } from '../../test';
+import { ApiError } from '../../lib/k8s/api/v2/ApiError';
+import store from '../../redux/stores/store';
 import { PureAlertNotification } from './AlertNotification';
 
 vi.mock('../../lib/cluster', async importOriginal => ({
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for these changes.

PR #7481, opened 2026-08-25
frontend: ingress: Fix HTTPS detection for multi-host TLS rules

### LINKED ISSUE #7480: frontend: Ingress: isHttpsUsed fails when a TLS entry lists multiple hosts
## Describe the bug

On the Ingress details page, rule host links are supposed to use `https://` when the host is covered by the Ingress's TLS config, and `http://` otherwise. That check breaks when a single `tls` entry lists more than one host — the link falls back to `http://` even though TLS is configured for that exact host.

The bug is in `isHttpsUsed` in `frontend/src/components/ingress/Details.tsx`:

```ts
function isHttpsUsed(item: Ingress, url: String) {
  const hostList = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`) ?? [];
  const isHttps = hostList.includes(`${url}`);

  return isHttps;
}
```

`hosts.hosts` is a `string[]` (a `tls` entry can list multiple hosts under one secret). Template-literal-interpolating an array joins it with commas, so a TLS entry like:

```yaml
tls:
  - hosts: ["a.example.com", "b.example.com"]
    secretName: my-cert
```

turns into the single string `"a.example.com,b.example.com"` in `hostList`. Checking `hostList.includes("a.example.com")` then compares against that joined string and returns `false`, even though `a.example.com` is clearly covered by this TLS entry. A `tls` block with only one host per entry happens to still work, since the "joined" string equals the single host name — which is why this has stayed unnoticed.

## To Reproduce

1. Create/view an Ingress whose `spec.tls` has one entry with multiple hosts, e.g.:
   ```yaml
   spec:
     tls:
       - hosts: ["a.example.com", "b.example.com"]
         secretName: my-cert
     rules:
       - host: a.example.com
         http: { ... }
   ```
2. Open the Ingress details page, look at the Rules table.
3. The host link renders as `http://a.example.com` instead of `https://a.example.com`, even though TLS is configured for it.

## Expected behavior

Any host listed under any `tls[].hosts` entry should be treated as HTTPS, regardless of how many other hosts share that same entry.

## Environment

Yes (I will propose a PR). Fix is to check each `tls` entry's actual `hosts` array with `.includes()` instead of stringifying the whole array first.

Priyanshu-rgbb:
## Summary

Fixes `isHttpsUsed` on the Ingress details page so rule host links correctly show `https://` when the host is covered by a `tls` entry that lists multiple hosts.

## Related Issue

Closes #7480

## Changes

`isHttpsUsed` in `frontend/src/components/ingress/Details.tsx` built its host list like this:

```ts
const hostList = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`) ?? [];
```

`hosts.hosts` is a `string[]`, and template-literal-interpolating an array joins it with commas. So a `tls` entry with `hosts: ["a.example.com", "b.example.com"]` became the single string `"a.example.com,b.example.com"` in `hostList`, and `hostList.includes("a.example.com")` compared against that joined string instead of the individual host — always `false`. Rule links for a host covered by a multi-host TLS entry rendered `http://` instead of `https://`. Single-host `tls` entries happened to still work, which is why this went unnoticed.

Fixed by checking each `tls` entry's actual `hosts` array directly:

```ts
function isHttpsUsed(item: Ingress, url: String) {
  const tlsEntries = item.jsonData.spec?.tls ?? [];
  return tlsEntries.some(tls => (tls.hosts ?? []).includes(`${url}`));
}
```

## Steps to Test

1. `cd frontend && npx vitest run src/components/ingress/Details.test.tsx`
2. Or manually: create an Ingress with `spec.tls: [{ hosts: ["a.example.com", "b.example.com"], secretName: "..." }]` and a rule for `a.example.com`, open its details page, and check the Rules table shows an `https://` link.

Added two tests to the existing `frontend/src/components/ingress/Details.test.tsx`, using the already-exported `LinkStringFormat` component directly:
- a host covered by a multi-host TLS entry resolves to `https://`
- a host not covered by any TLS entry still falls back to `http://`

Confirmed the first of these fails on `main` (renders `http://` instead of `https://`) and passes with the fix.

I checked the two Ingress storyshots that exercise TLS (`Details.WithTLS`, `Details.WithWildcardTLS`) — neither is affected: `WithTLS` uses a single-host `tls` entry (unaffected either way), and `WithWildcardTLS`'s rules render an empty table regardless of this fix, since its fixture rules have no `http.paths` (an unrelated pre-existing quirk of that fixture, out of scope here).

## Notes for the Reviewer

Confirmed no open issue or PR covers this before filing #7480.

### DIFF
--- a/frontend/src/components/ingress/Details.test.tsx
+++ b/frontend/src/components/ingress/Details.test.tsx
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { vi } from 'vitest';
 import { TestContext } from '../../test';
-import IngressDetails from './Details';
+import IngressDetails, { LinkStringFormat } from './Details';
 
 const { mockDetailsGrid } = vi.hoisted(() => ({
   mockDetailsGrid: vi.fn(),
@@ -179,5 +179,35 @@ describe('IngressDetails', () => {
     expect(sections).toHaveLength(1);
     expect(sections[0].id).toBe('headlamp.ingress-rules');
     expect(sections[0].section).toBeTruthy();
+  });
+});
+
+describe('LinkStringFormat', () => {
+  const multiHostTlsIngress: any = {
+    jsonData: {
+      spec: {
+        tls: [
+          {
+            hosts: ['a.example.com', 'b.example.com'],
+            secretName: 'multi-host-tls',
+          },
+        ],
+      },
+    },
+    spec: {
+      rules: [],
+    },
+  };
+
+  it('detects https for a host covered by a multi-host TLS rule', () => {
+    render(<LinkStringFormat url="b.example.com" item={multiHostTlsIngress} />);
+
+    expect(screen.getByRole('link')).toHaveAttribute('href', 'https://b.example.com/');
+  });
+
+  it('falls back to http for a host not covered by any TLS rule', () => {
+    render(<LinkStringFormat url="other.example.com" item={multiHostTlsIngress} />);
+
+    expect(screen.getByRole('link')).toHaveAttribute('href', 'http://other.example.com/');
   });
 });
--- a/frontend/src/components/ingress/Details.tsx
+++ b/frontend/src/components/ingress/Details.tsx
@@ -31,10 +31,8 @@ import SimpleTable from '../common/SimpleTable';
  * Is https used in the ingress item
  */
 function isHttpsUsed(item: Ingress, url: String) {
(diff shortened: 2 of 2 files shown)


Priyanshu-rgbb:
> Reviewed changes

Hey @illume PR is ready to be merged, have a review whenever you have a chance.

Priyanshu-rgbb:
> Reviewed changes

@illume This PR is ready to be merged.

PR #7486, opened 2026-08-25
frontend: Table: Provide select filter options when faceting is off

srebb:
## What this fixes

On large clusters the select-filter dropdowns (Status, Namespace, Node, …)
open empty, so filtering a list down to e.g. the pods that are not Running is
no longer possible.

Steps to reproduce (Headlamp 0.45.0, cluster with >500 pods):
1. Open the cluster-wide Pods list.
2. Open the column filters.
3. Click the Status or Namespace filter — the dropdown has no options.

**Why it happens**
`ResourceTable` disables faceted values above 500 rows:

```const enableFacetedValues = (data?.length ?? 0) <= 500;```

That guard is correct and was added for a real problem — TanStacks default getFacetedUniqueValues builds a unique-value map for every column, which is what made large pod lists run the browser out of memory (#5660, commit 1f35a61 "frontend: pod list: Add pagination").

The side effect is that MRTs useDropdownOptions sources the select-filter options from those same faceted values, so turning faceting off leaves every dropdown without options.

**What this changes**
Instead of disabling faceting, Table now passes its own getFacetedUniqueValues implementation that only walks the columns whose filter actually renders a dropdown (select, multi-select, autocomplete) and returns an empty map for every other column. The memory guard is preserved. The expensive part was building maps for all columns, including high cardinality ones like Name and IP — while the columns users actually filter on get their options back.

The map is built lazily (on first dropdown render) and cached per faceted row model, and MRT's dropdown pipeline is reused unchanged, so filterSelectOptions precedence and option sorting behave as before. Values are kept raw rather than stringified, so they still match what the filter functions compare against. A caller provided getFacetedUniqueValues still wins.

**Relation to #7226**
#7226 addresses the same root cause, but only restores the Namespace column by computing filterSelectOptions for it in ResourceTable. This change fixes it once in Table for every select filter, so Status, Node and any future select column are covered too.

**Tests**
- Table.test.tsx: unit coverage for the implementation itself. Empty map for non dropdown columns, recompute only when the faceted rows change, caller-provided implementation preserved.
- Table.facetedFilterOptions.test.tsx (new): regression tests against the real MRT rather than the module mock, because this fix depends on two behaviors that are not part of MRTs documented API. If an MRT upgrade changes, these tests fail instead of the dropdowns silently going empty again.

### DIFF
--- a/frontend/src/components/common/Table/Table.facetedFilterOptions.test.tsx
+++ b/frontend/src/components/common/Table/Table.facetedFilterOptions.test.tsx
@@ -28,10 +28,15 @@ import { ThemeProvider } from '@mui/material/styles';
 import { fireEvent, render, screen, waitFor, within } from '@testing-library/react';
 import { describe, expect, it, vi } from 'vitest';
 import { createMuiTheme } from '../../../lib/themes';
-import Table from './Table';
+import Table, { TableColumn, TableProps } from './Table';
 
 vi.mock('react-i18next', () => ({
-  useTranslation: () => ({ t: (key: string) => key, i18n: { language: 'en' } }),
+  useTranslation: () => ({
+    // Keys stand in for their translation, with the interpolation i18next would do.
+    t: (key: string, values?: Record<string, unknown>) =>
+      key.replace(/{{(\w+)}}/g, (_match, name) => String(values?.[name])),
+    i18n: { language: 'en' },
+  }),
 }));
 
 vi.mock('../../../lib/useShortcut', () => ({ useShortcut: vi.fn() }));
@@ -51,22 +56,26 @@ const theme = createMuiTheme({ base: 'light', name: 'light' });
 
 interface StatusRow {
   status: unknown;
+  tier?: string;
 }
 
 /** Enough rows that ResourceTable-style callers turn faceted values off. */
 const largeData: StatusRow[] = Array.from({ length: 501 }, (_, index) => ({
   status: index % 2 === 0 ? 'Running' : 'Completed',
+  tier: index % 2 === 0 ? 'gold' : 'silver',
 }));
 
-/** The page-size selector is a combobox too, so take the one in the table head. */
-function statusFilterInput() {
-  const combobox = screen.getAllByRole('combobox').find(element => element.closest('th') !== null);
-  expect(combobox).toBeDefined();
-  return combobox!;
+/** The page-size selector is a combobox too, so take the ones in the table head. */
+function statusFilterInput(index = 0) {
+  const comboboxes = screen
+    .getAllByRole('combobox')
+    .filter(element => element.closest('th') !== null);
+  expect(comboboxes.length).toBeGreaterThan(index);
+  return comboboxes[index];
 }
 
-function openStatusFilterDropdown() {
-  fireEvent.mouseDown(statusFilterInput());
+function openStatusFilterDropdown(index = 0) {
+  fireEvent.mouseDown(statusFilterInput(index));
 }
 
 /** One distinct value per row, far past MAX_FILTER_OPTIONS. */
@@ -76,32 +85,40 @@ const highCardinalityData: StatusRow[] = Array.from({ length: 1500 }, (_, index)
 
 function statusTable({
   accessorFn = (row: StatusRow) => row.status,
+  columns,
   data = largeData,
   enableColumnFilterModes,
+  enableFacetedValues = false,
   filterVariant = 'multi-select' as 'multi-select' | 'select' | 'autocomplete',
+  initialState,
   showColumnFilters = true,
 }: {
   accessorFn?: (row: StatusRow) => unknown;
+  columns?: TableColumn<StatusRow>[];
   data?: StatusRow[];
   enableColumnFilterModes?: boolean;
+  enableFacetedValues?: boolean;
   filterVariant?: 'multi-select' | 'select' | 'autocomplete';
+  initialState?: TableProps<StatusRow>['initialState'];
   showColumnFilters?: boolean;
 } = {}) {
   return (
     <ThemeProvider theme={theme}>
       <Table<StatusRow>
-        columns={[
-          {
-            id: 'status',
-            header: 'Status',
-            filterVariant,
-            accessorFn,
-          },
-        ]}
+        columns={
+          columns ?? [
+            {
+              id: 'status',
+              header: 'Status',
+              filterVariant,
+              accessorFn,
+            },
+          ]
+        }
         data={data}
         enableColumnFilterModes={enableColumnFilterModes}
-        enableFacetedValues={false}
-        initialState={showColumnFilters ? { showColumnFilters } : undefined}
+        enableFacetedValues={enableFacetedValues}
+        initialState={{ ...(showColumnFilters ? { showColumnFilters } : {}), ...initialState }}
       />
     </ThemeProvider>
   );
@@ -150,21 +167,28 @@ describe('Table select filter options with real MRT', () => {
     });
   });
 
-  it('explains the empty dropdown of a column above the option cap', () => {
-    // Rendering a menu item per value is what would freeze the tab, so the values
-    // are replaced by a notice.
-    renderStatusTable({ data: highCardinalityData });
+  it('serves only the most common options for a column past the cap', () => {
+    // Ported from #7226 by @simonepri. Plain DOM queries: computing accessible names for
+    // 1000 options is too slow for jsdom, and the "value (count)" shape drops MRT's
+    // placeholder item.
+    // The common value comes last, so keeping it proves the list is cut by frequency.
+    const data: StatusRow[] = [
+      ...Array.from({ length: 1000 }, (_, index) => ({ status: `rare-${index}` })),
+      ...Array.from({ length: 500 }, () => ({ status: 'common' })),
+    ];
+    renderStatusTable({ data });
 
     openStatusFilterDropdown();
 
-    const listbox = screen.getByRole('listbox');
-    expect(listbox.textContent).not.toContain('node-');
-
-    // The notice replaces the values and cannot be picked as a filter.
-    const options = within(listbox).getAllByRole('option');
-    expect(options).toHaveLength(1);
-    expect(options[0]).toHaveTextContent('Too many values to filter');
-    expect(options[0]).toHaveAttribute('aria-disabled', 'true');
+    const options = Array.from(
+      screen.getByRole('listbox').querySelectorAll('[role="option"]'),
+      option => option.textContent?.trim()
+    ).filter(text => / \(\d+\)$/.test(text ?? ''));
+    expect(options).toHaveLength(1000);
+    expect(options).toContain('common (500)');
+    expect(options).not.toContain('rare-999 (1)');
+    // The shortened list says so, so nobody hunts for a value that is not offered.
+    expect(screen.getByText('Showing the 1000 most common values')).toBeVisible();
   });
 
   it('follows the data while the dropdown stays open', () => {
@@ -175,17 +199,19 @@ describe('Table select filter options with real MRT', () => {
     // A background refresh pushes the column past the cap while the viewer is looking at it.
     rendered.rerender(statusTable({ data: highCardinalityData }));
 
-    expect(dropdownOptions()).toEqual(['Too many values to filter']);
+    expect(dropdownOptions()).toHaveLength(1000);
+    expect(screen.getByText('Showing the 1000 most common values')).toBeVisible();
   });
 
   it('restores the options when the data drops back under the cap', () => {
     const rendered = renderStatusTable({ data: highCardinalityData });
     openStatusFilterDropdown();
-    expect(dropdownOptions()).toEqual(['Too many values to filter']);
+    expect(dropdownOptions()).toHaveLength(1000);
 
     rendered.rerender(statusTable({ data: largeData }));
 
     expect(dropdownOptions()).toEqual(['Completed (250)', 'Running (251)']);
+    expect(screen.queryByText('Showing the 1000 most common values')).not.toBeInTheDocument();
   });
 
   it('serves no options for a column MRT cannot sort', () => {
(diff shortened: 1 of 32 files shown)

### CODE REVIEW

**Review by unlikelyzero — CHANGES_REQUESTED** [review]
Some things I observed:

1. The new `getFacetedUniqueValues` in `frontend/src/components/common/Table/Table.tsx:287-325` computes eagerly at mount for every dropdown-type filter column, not lazily on first dropdown open as described. MRT renders `MRT_TableHeadCellFilterLabel` for every filterable column and calls `useDropdownOptions` unconditionally, so mounting a large list walks the full row set for every select/multi-select/autocomplete column even if the user never opens a filter. Measured directly: a 501-row table with the filter row hidden invoked the column's `accessorFn` 501 times with this change versus 10 times (visible rows only) on `main`.

2. There's no cardinality bound on the columns this does compute (same range, `Table.tsx:311-319`). Namespace and Node are already multi-select columns in this codebase and can have thousands of distinct values on large clusters — precisely the clusters where the 500-row guard exists. The resulting map is rendered as one unvirtualized MUI `MenuItem` per value with no cap, so this trades the old failure mode (empty dropdown) for a hung tab, which is a worse outcome than the bug being fixed. This needs an explicit size cap (or a short-circuit when `filterSelectOptions` is already caller-provided) before merge.

3. Faceted values are kept raw rather than stringified, and MRT sorts dropdown options with `a.localeCompare(b)` — a non-string value (e.g. a numeric column configured as select/multi-select) throws during header render. Reproduced directly against the real MRT library; `main` doesn't hit this because the map was always empty above 500 rows, so this extends an existing MRT constraint's blast radius to large lists.

4. The comment on the `enableFacetedValues` threshold in `frontend/src/components/common/Resource/ResourceTable.tsx:722-726` ("disables it once the dataset exceeds a threshold" to avoid OOM) is now inaccurate, since `Table.tsx` unconditionally installs its own `getFacetedUniqueValues` that still walks large datasets for dropdown columns. Worth updating once the cardinality bound above is in place, since the right wording depends on that fix.

5. The new unit tests (`Table.test.tsx:434-531`) check output shape (empty map for non-dropdown columns, cache reuse) but don't assert the two properties actually being claimed: that computation is lazy, and that it stays bounded on high-cardinality data. The real-MRT regression file (`Table.facetedFilterOptions.test.tsx`) is good but only exercises the `multi-select` variant, not `select` or `autocomplete`.

Ran the change past the real MRT/TanStack library and confirmed the diagnosis is right and the fix is the right layer to apply it in — items 1 and 2 are what I'd want to see closed before merge; 3-5 could reasonably be follow-ups.

**Comment by srebb** [comment]
Thank you for the feedback @unlikelyzero 
I added fixes for 1 and 2. In addition I added a small message when the dropdown stays empty, so the user knows the reason and it feels less faulty.

File: frontend/src/components/common/Table/Table.tsx [hunk 1]
```diff
@@ -269,6 +277,79 @@ export default function Table<RowItem extends Record<string, any>>({
...
+        if (
+          !column ||
+          (filterVariant !== 'select' &&
+            filterVariant !== 'multi-select' &&
+            filterVariant !== 'autocomplete') ||
+          // Caller options win in MRT, which then also drops the value counts these feed.
+          columnDef?.filterSelectOptions
+        ) {
+          return emptyMap;
+        }
```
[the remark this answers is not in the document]
srebb: The stale state is real. Anyway, I am not happy with the analysis by copilot.

- `columnsOverOptionCap` is the actual problem, not that one early return.
  Any path that skips the walk can leave a verdict behind, so clearing it there closes one hole and leaves the class open.
- The closure cache cannot go stale when a definition is switched in place.
  TanStack memoizes `getAllColumns` on `_getColumnDefs()`, so a changed columns array creates new column objects and a fresh closure, which resets `computedForRows` and `uniqueValues`.
  Only the ref survives that. The visibility early return is not observable either, since MRT mounts the filter field inside `<Collapse mountOnEnter unmountOnExit>` and showing the filter row re-runs the walk.

Proposal: I could drop the ref and derive the verdict at call time instead of remembering it...

```ts
// dropdown variant, no caller filterSelectOptions, rows present but no value left
column.getFacetedUniqueValues().size === 0 && column.getFacetedRowModel().flatRows.length > 0
```

An empty map with rows present can only mean the cap kicked in, since every row contributes a value, `null` included.
That removes the state written during render and cannot go stale on any path.
I would also skip the derivation when a caller supplies its own `getFacetedUniqueValues`, because a foreign empty map  says nothing about our cap.

Should I implement this? The PR is growing more than expected at the beginning.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -361,6 +442,29 @@ export default function Table<RowItem extends Record<string, any>>({
...
+        return callerProps ?? {};
+      }
+      return {
+        ...callerProps,
+        // MRT renders this into an array next to its placeholder item, hence the key.
+        children: (
+          <MenuItem disabled key="too-many-values">
+            {t('Too many values to filter')}
+          </MenuItem>
+        ),
```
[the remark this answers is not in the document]
srebb: Good catch, autocomplete columns do go through the same cap, and the notice only reaches the select and multi-select branches, since MRT reads `muiFilterTextFieldProps.children` only there.

The suggested mechanism will not show up though.
MRT renders the autocomplete with `freeSolo: true`, and MUI renders its no-options block only when `groupedOptions.length === 0 && !freeSolo && !loading`, so `noOptionsText` stays dead while `freeSolo` is set.
Overriding `freeSolo: false` through `muiFilterAutocompleteProps` would make it render, but it also changes the filter:
typing arbitrary text is the point of that variant, and with a capped, empty option list the filter would be unusable.

A possible solution: put the notice on the input instead of the popup.

The autocomplete branch spreads `commonTextFieldProps` into its `renderInput` TextField, so a helperText coming from `muiFilterTextFieldProps` reaches autocomplete columns too, reusing the same translated string.
That keeps `freeSolo` and free text filtering intact, and the reason is visible without opening a popup at all.

For the popup route I do not see an option that keeps the cap intact, so I am open to suggestions.
Else, if you are fine with it, I am happy to implement the helperText route described above, if that sounds right.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -269,6 +277,79 @@ export default function Table<RowItem extends Record<string, any>>({
     [tableProps.columns]
   );
 
+  // Select-filter dropdowns get their options from MRT's faceted values, but callers turn
+  // faceting off above a dataset threshold (TanStack's default builds a unique-value map for
+  // every column), which left those dropdowns empty on large lists. This walks the dropdown
+  // columns only, and only while their options are on screen, since MRT reads faceted values
+  // on every header render. It relies on two undocumented MRT behaviors, pinned in
```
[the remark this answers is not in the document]
srebb: Agreed on the end state, `docs/contributing.md` asks for atomic commits and for squashing away development history, so I will do that.

I would rather do it right before merge than now: force-pushing at this stage detaches the existing review comments from the code they point at, and there are open threads on this PR.
Once the review is settled I will rebase into one implementation commit plus the tests, keeping the messages that explain the why.

@unlikelyzero @sniok @vyncent-t if you prefer the squash now, say so and I will push it right away.

frontend/src/components/common/Table/Table.facetedFilterOptions.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const theme = createMuiTheme({ base: 'light', name: 'light' });
+
+interface StatusRow {
+  status: string;
+}
+
+/** Enough rows that ResourceTable-style callers turn faceted values off. */
+const largeData: StatusRow[] = Array.from({ length: 501 }, (_, index) => ({
+  status: index % 2 === 0 ? 'Running' : 'Completed',
```
unlikelyzero, line 64: Nice

unlikelyzero (approved):

srebb:
@illume may you please review my requests belonging to the copilot review, you requested?

illume (changes requested):
Thanks for the contribution.


frontend/src/components/common/Table/Table.tsx
```diff
@@ -269,6 +277,79 @@ export default function Table<RowItem extends Record<string, any>>({
...
+            filterVariant !== 'autocomplete') ||
+          // Caller options win in MRT, which then also drops the value counts these feed.
+          columnDef?.filterSelectOptions
+        ) {
+          return emptyMap;
+        }
+        // Faceting on: the column's faceted row model. Faceting off: TanStack falls back to
+        // the pre-filtered rows (post-filterFunction), a superset of what the table shows.
+        const { flatRows } = column.getFacetedRowModel();
+        if (flatRows !== computedForRows) {
```
rootp1: This is the same spot Copilot flagged earlier about stale faceted value caching, and I think this PR builds new user facing behavior directly on top of it. MemoHeadCell's render comparator can skip re rendering the header on data only changes, so this check against computedForRows does not reliably re run while a filter dropdown is open. Practically that means if a column crosses the 1000 item cap while its filter is already open, say data refreshes in the background, the new too many values hint can end up stale or missing until the user closes and reopens the dropdown. Since that is basically the exact case this PR is meant to handle, I would rather see it fixed or explicitly called out with a follow up issue than merged silently alongside it.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -269,6 +277,79 @@ export default function Table<RowItem extends Record<string, any>>({
...
+  // faceting off above a dataset threshold (TanStack's default builds a unique-value map for
+  // every column), which left those dropdowns empty on large lists. This walks the dropdown
+  // columns only, and only while their options are on screen, since MRT reads faceted values
+  // on every header render. It relies on two undocumented MRT behaviors, pinned in
+  // Table.facetedFilterOptions.test.tsx: caller options win over MRT's own
+  // `enableFacetedValues ? ... : undefined`, and the dropdown reads faceted values regardless
+  // of that flag.
+  // Ids of the columns whose options were dropped by MAX_FILTER_OPTIONS, so their dropdown can
+  // say so instead of just being empty. Written while MRT renders the header, hence a ref.
+  const columnsOverOptionCap = useRef(new Set<string>());
```
rootp1: columnsOverOptionCap is a plain useRef, so flipping its contents does not itself cause React to re render the filter text field, it only shows up next time muiFilterTextFieldProps happens to get invoked for some other reason. That works today because of MRT's call order, but it is fragile and non obvious to a future maintainer. Would it be possible to thread this through state or a memo dependency instead, or at least leave a comment here explaining why relying on the ref is safe given the current call order.

frontend/src/components/common/Table/Table.tsx
```diff
@@ -361,6 +442,29 @@ export default function Table<RowItem extends Record<string, any>>({
     ...tableProps,
     columns: tableColumns ?? [],
     data: tableData,
+    getFacetedUniqueValues: tableProps.getFacetedUniqueValues ?? getFacetedUniqueValues,
+    // Tell the user why a capped dropdown has no options. MRT renders `children` instead of
+    // the option list, so the entry cannot be picked as a filter value.
+    muiFilterTextFieldProps: args => {
```
rootp1, line 503: Would love a test for the transition case, a column that crosses the threshold while its filter UI is already open, and also the reverse, a column that drops back under the cap. Both are the scenarios most likely to regress silently given the ref based signal above.

rootp1 (commented):
Pulled this branch and ran the existing and new tests locally, all 45 pass and the i18n strings are complete across every locale, nice work there. Left a couple of inline notes on the part I think needs another look before merge.

srebb:
All open review items are addressed, and the history is now squashed and rebased onto current main, as I said I would once the review had settled.
That is why the commit ids changed and why some inline threads now point at detached commits.
Sorry for that noise, the trade was keeping the review anchored while it was in progress.

The branch is two commits now, since the first one is a separate concern:

- **Keep the row identity a filtered table depends on**: `useFilterFunc` handed out a new closure per render, and `Table` memoizes its rows on that function, so every `ResourceTable` rebuilt its row model on every render and discarded everything derived from those rows. The cluster overview did the same to its events table.
- **Table: Provide select filter options when faceting is off**: the feature itself.

What changed since the last round:

- The capped-options state is derived per call instead of remembered in a ref, so no path can leave a stale verdict behind, which answers Copilot's stale-ref point and @rootp1's question about threading it through state.
- Options are cached per row model in a `WeakMap`. MRT rebuilds its column array on every render and TanStack binds a fresh closure and column definition along with it, so a cache living in that closure never survived a render, which is @rootp1's point about the check not re-running reliably.
- Tests for a column crossing the cap while its dropdown is open, and for dropping back under it, both as @rootp1 asked.
- The implementation is installed only when faceting is off and the caller brought none, so TanStack's default keeps its full contract on small lists.
- Values MRT cannot sort are skipped, since it orders options with `localeCompare`, which throws on anything else.
- The autocomplete variant gets the reason under the input, since MUI never shows its no-options popup under `freeSolo`, which is where Copilot's `noOptionsText` suggestion would not have rendered.

### Relation to #7226

#7226 fixes the same problem and now uses the same mechanism.
The two conflict textually: both add `Table.facetedFilterOptions.test.tsx` and both change the same two regions of `Table.tsx`, so only one can land first and the other needs a rebase.

Functional delta as of today:

| | #7226 | this PR |
|---|---|---|
| Options with faceting off | yes | yes |
| Above the cap | keeps the 1000 most frequent values | empty dropdown plus a translated reason |
| Walk deferred until the options are on screen | no | yes |
| Cache survives a render | no, per-column closure | yes, per row model |
| Non-string values | unguarded, `localeCompare` throws | skipped |
| Autocomplete variant | no options, no reason | reason under the input |
| Row identity (`useFilterFunc`) | no | yes |

### Which one to use as the base

The important thing is that this gets fixed, not whose branch ends up on top of whose.
I have offered @simonepri both directions in #7226, either I contribute the missing items to her branch or she contributes her top-N retention to mine, so either order works and neither of us loses authorship.

If it helps you decide, I would suggest this one, for a reason about remaining work rather than authorship.
The three items #7226 is missing are not polish.
The unguarded non-string values are a new crash path, since `localeCompare` on a number throws in the header render, and with faceting off those columns never had options before.
The walk running at mount cost 501 accessor calls per dropdown column on a 501-row table with the filter row still closed, measured on this PR before the gate went in, and #7226 has no such gate.
The per-column cache does not survive a render, so the walk repeats on every header render.
Those took several review rounds here to find and pin.
Going the other way, folding #7226's top-N retention into this PR is a contained change, and it is the better answer to the cap, so I would take it either way and credit @simonepri for it.

Test coverage is the other half of it: 53 test declarations across the two files here, 66 cases once the parameterised ones expand, versus 7 there, and the ones that matter most are the regressions reviewers asked for, a column crossing the cap while its dropdown is open and the way back under it.
#7226 does cover two things this PR does not, the faceting-on contract for non-dropdown columns and small-list narrowing, and I would adopt both tests.

@illume @sniok since this spans both PRs, could one of you say which branch should be the base?
Whichever you pick, I will adapt on this side, and nobody has to wait on me.

The honest counter-argument: #7226 is older, a third of the size and easier to review, which is a legitimate reason to prefer it.
Also note the first commit here touches `useFilterFunc`, a public plugin API whose identity semantics change, which is why it is a separate commit and can be dropped or deferred if you would rather not carry it in this PR.

frontend/src/components/common/Table/Table.tsx
srebb: I could not reproduce this one.
Measured against the real MRT: with the dropdown left open, a background refresh that pushes the column past the cap updates the list right away, and so does the way back under it.
I did try threading the row identity into MemoHeadCell's comparator, then dropped it again, because removing it made no test fail, so I had no evidence it was doing anything.
Both directions are now pinned in Table.facetedFilterOptions.test.tsx, which is what makes this checkable at all, so thank you for asking for it.
The mechanism also changed since your review: there is no ref any more, and the values are cached per row model.

frontend/src/components/common/Table/Table.tsx
srebb: Agreed, and the ref is gone.
Whether a column's list was shortened is now derived at call time from the map MRT hands back, marked in a module-level WeakSet, so nothing has to be kept in sync and correctness no longer rests on MRT's call order.

frontend/src/components/common/Table/Table.tsx
srebb: Both are in, plus the same pair for the autocomplete variant.
One thing changed underneath them: past the cap the dropdown now keeps the most common values instead of going empty, adopted from #7226, and the note under the input says the list is shortened.
So the transition tests assert the shortened list and the note rather than an empty dropdown.

PR #7491, opened 2026-08-26
Improve TypeScript Typings for KubeCondition

### LINKED ISSUE #7490: Improve TypeScript Typings for KubeCondition
In frontend/src/lib/k8s/cluster.ts, there is an explicit // @todo comment regarding KubeCondition. Currently, it's defined as a catch-all interface, assuming lastProbeTime is always present. However, Kubernetes API condition types (like PodCondition, NodeCondition) have varying schemas.

Define a strict type hierarchy for KubeCondition (e.g. a base Condition type, and extending interfaces for PodCondition, NodeCondition, etc.) and apply it to the relevant files.

Plugin developers and contributors will benefit from stricter TypeScript typings.

Yes (I will propose a PR)

## Additional context
Tech Debt / Codebase TODO.

Rucha0901:
## Summary

This PR improves TypeScript typings for KubeCondition by introducing a strict type hierarchy.

## Related Issue

Fixes #7490

## Changes

- Added BaseCondition, PodCondition, and NodeCondition in k8s models
- Refactored KubeCondition to use a TypeScript union to minimize test mock disruptions

## Steps to Test

1. Run `npm run tsc` in the frontend directory.
2. Verify that there are no type errors related to KubeCondition.

## Notes for the Reviewer

The KubeCondition was changed to a union type to prevent existing test mock files that use `lastProbeTime` from failing type checks, minimizing the number of modified files to just 4 core files.


unlikelyzero (changes requested):
Some things I observed:

1. (Blocking) The new `KubeCondition` union type in cluster.ts (around line 168: `BaseCondition | (BaseCondition & { lastProbeTime: Time })`) is structurally vacuous. Because the second arm is a strict subtype of the first, TypeScript collapses it to plain `BaseCondition` for everything except fresh object-literal checks. Concretely, `kc.lastProbeTime` is now a compile error unless narrowed with an `in` check. Since `KubeCondition` is re-exported to out-of-tree plugin authors, any existing plugin code reading `condition.lastProbeTime` (previously required, previously safe) will fail to compile on upgrade with an unhelpful error. I'd drop the union entirely in favor of a plain `lastProbeTime?: Time` optional field on the interface — that gives the same mock-literal tolerance the PR is going for, without the broken type.

2. `lastUpdateTime` (a Deployment/ReplicaSet-only field per the actual Kubernetes API) now lives on the common `BaseCondition` in cluster.ts (around lines 157-165), so it gets inherited by the new "strict" Pod and Node condition subtypes even though NodeCondition previously explicitly excluded it via Omit. This inverts the type hierarchy the PR is trying to build. I'd move `lastUpdateTime` off `BaseCondition` onto a workload-specific type used only by Deployment/ReplicaSet.

3. replicaSet.ts line 36 was changed from `Omit<KubeCondition, 'lastProbeTime' | 'lastUpdateTime'>[]` to plain `KubeCondition[]`. This widening looks unnecessary — the original `Omit<...>` expression still compiles unchanged against the new union type, so this file didn't need to be touched at all, and touching it strictly lost precision by now permitting fields the ReplicaSet API never returns.

4. `NodeCondition.lastHeartbeatTime` in node.ts (around lines 25-27) was silently weakened from required `string` to optional `Time` (a wider type), which isn't mentioned in the PR description. The real Kubernetes API always returns this field as a non-empty string for Node conditions.

5. Minor: a stale comment in the Storybook fixture ConditionList.stories.tsx line 29 says lastProbeTime "is required by the type but never rendered" — under the new union it's no longer required at all, so the comment (and the `lastProbeTime: null` entries it justifies) is now unexplained dead weight.

6. Minor: the deleted @todo block in cluster.ts (was around line 155) overstated how complete the migration is. Deployment, DaemonSet, StatefulSet, Job, Gateway, Namespace, Service, and PodGroup conditions still route through the same generic (now collapsed) KubeCondition type. I'd keep a trimmed @todo naming what's still unmigrated rather than deleting it outright.

Rucha0901:
@unlikelyzero , I did the changes ,can you plz review once?

app/electron/hardwareAcceleration.test.ts
```diff
@@ -87,7 +87,7 @@ describe('startWindowsVMDetection', () => {
       })
     );
 
-    expect(waitForWindowsVMDetection(detection)).toBe(false);
+    expect(waitForWindowsVMDetection(detection, 5000)).toBe(false);
```
unlikelyzero: Let's not add arbitrary timeouts here

PR #7493, opened 2026-08-26
frontend: Settings: Use useTheme() instead of TestHelpers theme

### LINKED ISSUE #7492: frontend: Settings imports theme from TestHelpers into production code
## Describe the bug

`frontend/src/components/App/Settings/Settings.tsx` imports the app's `theme` object from `frontend/src/components/TestHelpers/theme.ts`:

```ts
import { theme } from '../../TestHelpers/theme';
```

That's a test fixture — a standalone `createTheme(...)` call meant for wrapping components in tests, living in a directory literally named `TestHelpers`. It's not the app's real, live theme (which comes from `createMuiTheme()` in `lib/themes.ts` and reacts to the user's actual theme selection, including dark mode).

Most of the `theme.` references in this file are inside `sx={theme => ({...})}` callback props, where the callback parameter happens to be named `theme` too — so those are actually shadowed and correctly use MUI's real live theme via context, despite reading as if they used the broken import. But one usage isn't inside a callback (the responsive grid layout for the theme picker, around line 294-299):

```tsx
<Box
  sx={{
    display: 'grid',
    gridTemplateColumns: 'repeat(auto-fill, minmax(180px, 1fr))',
    gap: 2,
    justifyContent: 'center',
    [theme.breakpoints.down('sm')]: {
      gridTemplateColumns: 'repeat(auto-fit, minmax(130px, 1fr))',
      gap: 2,
    },
    ...
```

This one really does read the static test theme's `breakpoints.down('sm')`, not the live one. It happens to render correctly today only because the test theme's `sm` breakpoint coincidentally still defaults to MUI's standard `600px`, same as the real app theme — so there's no visible glitch right now, but that's incidental, not guaranteed, and it'll silently drift if either theme's breakpoints change independently.

Regardless of whether it's currently visibly broken, production code pulling in a module from a directory named `TestHelpers` is a bundle-hygiene problem on its own — it ships test-only code in the production build and creates a hidden coupling that's easy to reintroduce or get subtly wrong (as the shadowing above shows).

## Expected behavior

`Settings.tsx` should get the theme via `useTheme()` from `@mui/material/styles` like other components in the app do, not from a test helper. No production file should import from `TestHelpers/`.

## Environment

Yes (I will propose a PR). Fix is to drop the `TestHelpers/theme` import, call `const theme = useTheme()` at the top of the component instead, and let the existing `theme.` references (both the shadowed ones and the one real usage) resolve to the same hook-provided value.

Priyanshu-rgbb:
## Summary

`Settings.tsx` was importing its theme from `TestHelpers/theme.ts` (a test-only fixture) instead of the app's real live theme. Switches it to `useTheme()`.

## Related Issue

Closes #7492

## Changes

Most of the `theme.` references in this file were inside `sx={theme => ({...})}` callbacks, where the callback's own `theme` parameter shadows the module-level import — those already resolved correctly against the real theme via context, despite reading like they used the broken import. One usage wasn't inside a callback though (the responsive grid layout for the theme picker), and that one genuinely read the static test theme's `breakpoints.down('sm')` instead of the live one.

The two themes actually diverge at the `md` breakpoint: the app's real theme (`lib/themes.ts`) sets `md: 960`, while `TestHelpers/theme.ts` never overrides breakpoints, so it falls back to MUI's default `md: 900`. `sm` happens to be `600` in both, which is why this bug produced no visible diff today — the only breakpoint this file actually uses is `sm`. But that's incidental, not guaranteed, and the divergence at `md` shows the two themes are not interchangeable in general.

Regardless of current visible impact, production code importing from a directory literally named `TestHelpers` is a bundle-hygiene problem on its own (ships test-only code in the prod bundle, and the shadowing above shows how easy it is to introduce this kind of hidden coupling without noticing).

Fix:
- Drop the `TestHelpers/theme` import, call `const theme = useTheme()` at the top of the component like other components in the app already do.
- Also removed the two `sx={theme => ({...})}` callback wrappers (lines ~260, ~283) in favor of plain `sx={{...}}` objects that reference the component-level `theme` directly, so every `theme.` reference in the file is now the same binding — not just resolving to the same value by coincidence of context, but literally the same variable. Removes the shadowing idiom inconsistency this file previously had.

## Steps to Test

1. `npx tsc --noEmit` — no new type errors.
2. `cd frontend && npx vitest run src/components/App/Settings/Settings.test.tsx` — passes. Added a new regression test, `resolves the theme-picker grid breakpoint from the live theme, not a static import`, that renders `Settings` inside a `ThemeProvider` with a deliberately distinctive `sm` breakpoint (733px, chosen so it can't coincidentally match either theme's default) and asserts the grid's own emitted `@media` rule uses that value. This test fails against the pre-fix code (confirmed locally by reverting `Settings.tsx` to the `TestHelpers`-import version and re-running it — it emits `599.95px` instead of `732.95px`), and passes against the fix. The previously existing test only asserted the `SETTINGS_VIEW` dispatch event and didn't exercise this path at all.
3. Manually: open Settings, resize the window across the `sm` breakpoint, confirm the theme picker grid still reflows correctly.

## Notes for the Reviewer

No behavior change is expected or intended — this replaces a dead/fragile import with the correct one everywhere it's used, and the `sm` values matching means there's no visual diff from this change today. Confirmed via grep that no other file depends on `Settings.tsx` importing from `TestHelpers/`. Confirmed no open issue or PR covers this before filing #7492.

Thanks to @unlikelyzero's review for flagging that the existing test/Storybook coverage couldn't actually catch a regression here (since both themes coincide at `sm`), and for pointing out the `md` divergence and the callback-shadowing inconsistency — addressed both above.

### DIFF
--- a/frontend/src/components/App/Settings/Settings.test.tsx
+++ b/frontend/src/components/App/Settings/Settings.test.tsx
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-import { ThemeProvider } from '@mui/material/styles';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { render, waitFor } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import { ReactNode } from 'react';
 import { afterEach, describe, expect, it } from 'vitest';
 import App from '../../../App';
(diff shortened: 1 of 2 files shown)


unlikelyzero (approved):
Some things I observed:

1. The fix (Settings.tsx, replacing a TestHelpers/theme import with useTheme()) doesn't have test coverage that locks it in — the existing test only checks a dispatch event, and the Storybook snapshot can't catch a regression here since both themes happen to render identically at the sm breakpoint today.
2. The two themes (the app's real theme and the TestHelpers test theme) actually diverge at the md breakpoint (960 vs MUI's default 900) — worth citing this concrete divergence in the PR description to strengthen the justification for the fix, since currently it only mentions the sm breakpoint coincidence.
3. Minor style note: the new useTheme() binding at line 44 is shadowed by two existing callback-scoped theme params elsewhere in the same file (lines 260, 283) — no bug, just an inconsistency in idiom worth a passing mention.

skoeva (approved):
lgtm, thanks

PR #7495, opened 2026-08-26
frontend: globalSearch: Fix keyboard scroll for recent items

### LINKED ISSUE #7494: frontend: GlobalSearch: keyboard navigation doesn't scroll the recent-items list
## Describe the bug

In Global Search, keyboard navigation (arrow up/down) through the results list is supposed to keep the virtualized list scrolled so the highlighted row stays visible. That breaks specifically when the search box is empty and you're browsing the "recent items" list instead of live search results.

The bug is in `GlobalSearchContent.tsx`:

```ts
const autocomplete = useAutocomplete<SearchResult, false, false, true>({
  options: !query ? recentItems : results,
  ...
  onHighlightChange(_, option, reason) {
    if (reason === 'keyboard' && option) {
      const index = results.indexOf(option);
      const list = listRef.current;
      list?.scrollToItem(index);
    }
  },
  ...
});
```

The `options` actually shown to the user switch between `recentItems` (empty query) and `results` (active query), but `onHighlightChange` always looks up the highlighted option's position in `results` — even when `recentItems` is what's actually rendered. When browsing recent items with an empty query, `results` is empty (fuse.js has nothing to search against), so `results.indexOf(option)` is always `-1`. `scrollToItem(-1)` is a no-op, so the virtualized list never scrolls to follow the highlight past whatever's already visible in the initial viewport.

## To Reproduce

1. Open Global Search (search bar or the configured shortcut) with an empty query, so it's showing your recent items.
2. Have enough recent items that the list needs to scroll (more than fit in the visible ~10-row window).
3. Press the down arrow repeatedly to move the highlight past the visible rows.
4. The highlighted row moves out of view — the list never auto-scrolls to keep it visible, unlike when the same navigation is done over live search results.

## Expected behavior

Keyboard navigation through the recent-items list should scroll the list to keep the highlighted row visible, the same way it already works for live search results.

## Environment

Yes (I will propose a PR). Fix is to look up the highlighted option's index in whichever list is actually displayed (`recentItems` or `results`), not always `results`.

Priyanshu-rgbb:
## Summary

Fixes keyboard navigation in Global Search so the virtualized list actually scrolls to follow the highlight when browsing recent items with an empty query.

## Related Issue

Closes #7494

## Changes

The `options` shown to the user switch between `recentItems` (empty query) and `results` (active query):

```ts
options: !query ? recentItems : results,
```

but `onHighlightChange` always looked up the highlighted option's position in `results`:

```ts
onHighlightChange(_, option, reason) {
  if (reason === 'keyboard' && option) {
    const index = results.indexOf(option);
    ...
```

With an empty query, `results` is empty (nothing to search), so this lookup was always `-1` and `scrollToItem(-1)` was a silent no-op — arrow-key navigation through recent items never scrolled the list. Fixed by looking up the index in whichever list is actually displayed:

```ts
const displayedOptions = !query ? recentItems : results;

const autocomplete = useAutocomplete<SearchResult, false, false, true>({
  options: displayedOptions,
  ...
  onHighlightChange(_, option, reason) {
    if (reason === 'keyboard' && option) {
      const index = displayedOptions.indexOf(option);
      ...
```

## Steps to Test

Manually: open Global Search with an empty query so it shows recent items (need enough recent items — more than fit in the ~10-row visible window — to see the scroll behavior), then press the down arrow repeatedly. With the fix, the list scrolls to keep the highlighted row visible; on `main` it doesn't once you arrow past the initially visible rows.

## Notes for the Reviewer

I wasn't able to add an automated test for this one. `GlobalSearchContent.tsx` imports a large number of `lib/k8s/*` resource classes at module scope, and mounting it in isolation via Vitest (outside the app's normal bootstrap/Storybook context) hits a pre-existing circular-import crash unrelated to this change:

```
TypeError: Class extends value undefined is not a constructor or null
  at src/lib/k8s/replicaSet.ts:44:26 (class ReplicaSet extends KubeObject<...>)
  at src/lib/k8s/deployment.ts:24:1
```

I confirmed this reproduces on a bare `import { GlobalSearchContent } from './GlobalSearchContent'` with no other changes, so it's a structural issue with the module's import graph, not something introduced here — happy to file it separately if useful. Verified the fix by tracing the exact code path by hand (confirmed `results` is empty whenever `recentItems` is what's rendered, so the old lookup was unconditionally `-1`) and via the existing Storybook stories for this component, which render fine.

Confirmed no open issue or PR covers this before filing #7494.

### DIFF
--- a/frontend/src/components/globalSearch/GlobalSearchContent.tsx
+++ b/frontend/src/components/globalSearch/GlobalSearchContent.tsx
@@ -485,8 +485,10 @@ export function GlobalSearchContent(props: GlobalSearchContentProps) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [recent, results, query]);
 
+  const displayedOptions = !query ? recentItems : results;
+
   const autocomplete = useAutocomplete<SearchResult, false, false, true>({
-    options: !query ? recentItems : results,
+    options: displayedOptions,
     freeSolo: true, // free user input, not just autocomplete options
     autoHighlight: true, // highlight first option on open
     openOnFocus: true,
@@ -494,7 +496,7 @@ export function GlobalSearchContent(props: GlobalSearchContentProps) {
     filterOptions: options => options, // we handle filtering ourself
     onHighlightChange(_, option, reason) {
       if (reason === 'keyboard' && option) {
-        const index = results.indexOf(option);
+        const index = displayedOptions.indexOf(option);
         const list = listRef.current;
         list?.scrollToItem(index);
       }


unlikelyzero (approved):
Some things I observed:
1. There's no regression test covering arrow-key navigation of the virtualized list in GlobalSearchContent.tsx, so a future refactor could silently reintroduce the original scroll bug with nothing to catch it.
2. When the search query changes (e.g. from empty to non-empty), the highlight reset doesn't trigger a scroll. The code only scrolls for reason === 'keyboard', but MUI's internal highlight-reset uses reason: 'auto', so the highlighted first result can render off-screen after typing following arrow-key navigation of recent items.
3. indexOf is computed against displayedOptions, which today happens to match what's actually rendered, but that's not guaranteed to hold going forward - a future PR adding grouping/filtering could silently break this scroll logic with no type error.

Priyanshu-rgbb:
Hey @illume This PR is ready to be merged, have a review whenever you have a chance.

Priyanshu-rgbb:
@illume All the copilots comments have been resolved, this PR is ready.

PR #7538, opened 2026-08-28
Add secondaryContrastText colors to built-in dark themes

tejhan:
## Summary

I noticed that our secondary buttons in the built-in Dark themes don't have sufficient contrast, making them hard to see. 

We see this surface in a few elements in dark mode, for instance in the project creation popups (See cancel button): 
imageimage

This is because we explicitly defined a secondary color for the background, but none for the button text-  which left it to use headlmap's default, which is a dark text color (for the default light secondary button with it's light background).

This PR adds `secondaryContrast` to `darkTheme` & `lightsOutTheme` that match their existing `text -> primary` color to ensure adequate contrast.

**Fixed:**
image image


illume (approved):
🎉 thanks!

PR #6562, opened 2026-07-18
frontend: Add PodGroup and Workload scheduling views

Anjali-Chauhan1:
## Summary

This PR adds support for **Workload-Aware Scheduling** in Headlamp by displaying the two new gang-scheduling APIs introduced in Kubernetes v1.36 — **PodGroup** and **Workload** (`scheduling.k8s.io/v1alpha2`). These are built-in aggregated APIs (not CRDs), so Headlamp previously had no way to surface them: they don't appear under Custom Resources or in Advanced Search. As a result, a user whose Pod is stuck `Pending` because its gang group is incomplete had no way to see why. This PR closes that gap with dedicated list/details views, a Pod → PodGroup link, and a gated sidebar section.

## Related Issue

Fixes #6488

## Changes

- **Added** `PodGroup` and `Workload` resource classes under `lib/k8s/` (`podGroup.ts`, `schedulingWorkload.ts`), registered in `ResourceClasses`.
- **Added** list + details views for both kinds — PodGroups show Policy (Gang/Basic), Min Count, parent Workload, and scheduling status; Workloads show their Pod Group Templates and controller reference.
- **Added** a **"Scheduling (alpha)"** sidebar section, gated on `PodGroup.isEnabled()` so it only shows on clusters that actually serve the alpha API.
- **Added** a **"Pod Group"** row to the Pod details page (issue requirement c), linking a Pod to its scheduling group via the new `spec.schedulingGroup.podGroupName` field.
- **Added** unit tests for the new classes and list views, plus translations for all 16 supported locales.
- **Fixed** a column-width issue so the PodGroups list matches the layout of other Headlamp resource tables.

## Steps to Test

> **Requires a Kubernetes v1.36 cluster with the `GenericWorkload` feature gate** — these APIs are alpha and not served by default. See the [demo kind config](https://github.com/kannon92/kubecon-eu-2025-demo/blob/main/kind/kind.yaml).

1. Open a cluster **with** the feature gate → the sidebar shows **Scheduling (alpha)** with **Pod Groups** and **Workloads**.
2. Go to **Pod Groups** → see the group listed with Policy, Min Count, Workload, and Status. Click it → details show the scheduling policy, template reference (linked to the Workload), and Conditions.
3. Go to **Workloads** → click a workload → its **Pod Group Templates** table lists each template's policy and min count.
4. Open a Pod that belongs to a group → the details grid shows a **Pod Group** row linking to its PodGroup.
[...]

Issue #7530, opened 2026-08-28
frontend: Visualize how the scheduling API objects relate in the resource map


Anjali-Chauhan1:
## Is your feature request related to a problem? Please describe the impact that the lack of the feature requested is creating.
The scheduling APIs are graph-shaped: a Workload templates CompositePodGroups and PodGroups, composite groups nest into each other, and PodGroups own Pods. The list and details views added in #6562 show each object on its own, so users have to click through several pages to reconstruct how one workload is put together.

## Describe the solution you'd like
a) Add Workload, PodGroup and CompositePodGroup as sources in the resource map
b) Add the edges between them: Workload to the groups created from its templates, CompositePodGroup to its child groups, and PodGroup to its Pods
c) Show scheduling state on the nodes, so an unscheduled group is visible without opening it

## What users will benefit from this feature?
Kubernetes users running gang-scheduled workloads who want to see a whole workload at once instead of one object at a time.

## Are you able to implement this feature?
Yes (I will propose a PR).

## Additional context
The other half of the follow-up to #6562. The CompositePodGroup nodes and edges depend on #7529; the Workload, PodGroup and Pod edges do not, so this can start before that lands.

These objects are not linked by ownerRefs, so the edges come from spec fields: `workloadRef`, `parentCompositePodGroupName`, and `schedulingGroup.podGroupName` on the Pod.

- KEP-6012: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6012-composite-podgroup-api

PR #7542, opened 2026-08-29
frontend: Add CompositePodGroup views for hierarchical gang scheduling

### LINKED ISSUE #7529: frontend: Add CompositePodGroup views for hierarchical gang scheduling
#6562 added PodGroup and Workload views. Kubernetes 1.37 adds a third scheduling API, CompositePodGroup, which nests pod groups into a hierarchy. Headlamp doesn't display it, so users can't tell which part of a nested workload failed to schedule.

a) Display CompositePodGroups
b) Link a PodGroup to its parent CompositePodGroup, and a CompositePodGroup to its parent and children
c) Expand the nested composite templates on the Workload details page

Kubernetes users running hierarchical gang-scheduled workloads (JobSet, LeaderWorkerSet, Kubeflow Trainer).

Yes (I will propose a PR).

## Additional context
Follow-up to #6562. The visualization part is tracked separately in #7530.

CompositePodGroup is served by `scheduling.k8s.io/v1alpha3` only and needs the `CompositePodGroup` feature gate enabled, so the views should stay hidden on clusters without it.

- KEP-6012: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6012-composite-podgroup-api
- API types: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/scheduling/v1alpha3/types.go

Anjali-Chauhan1:
## Summary

This PR adds support for `CompositePodGroup`, the third scheduling API from KEP-6012, which nests pod groups into a hierarchy. Headlamp could already show the two flat objects (`PodGroup`, `Workload`) added in #6562, but not the tree that ties them together, so a user could not tell which part of a nested workload failed to schedule.

## Related Issue

Fixes #7529

## Changes

- Added the `CompositePodGroup` resource class, modelling `spec` as served by `scheduling.k8s.io/v1alpha3`, with its own availability check
- Added list and details views, routes, and a sidebar entry in the existing Scheduling (alpha) section
- Added hierarchy navigation: a children section listing the child composite groups and pod groups, a parent link on the composite details page, and a parent link on the pod group details page
- Updated the Workload details page to show every level of nested composite templates, and to name the pod group templates instead of counting them
- Refactored the scheduling status column out of the PodGroup list so both list views share it
- Added unit tests, stories and storyshots, and translations for the two new keys across all locales

## Steps to Test

The API is alpha and off by default, so it needs a 1.37 cluster with the feature gates enabled.

1. Create the cluster:
   ```yaml
   # kind-was.yaml
   kind: Cluster
   apiVersion: kind.x-k8s.io/v1alpha4
   featureGates:
     GenericWorkload: true
     TopologyAwareWorkloadScheduling: true
     CompositePodGroup: true
   runtimeConfig:
     "scheduling.k8s.io/v1beta1": "true"
     "scheduling.k8s.io/v1alpha3": "true"
   nodes: [{role: control-plane}, {role: worker}, {role: worker}]
   ```
   ```
   kind create cluster --config kind-was.yaml --image kindest/node:v1.37.0
   ```
   Both group versions are needed: the scheduler watches `PodGroup` on `v1beta1`, while `CompositePodGroup` is only served by `v1alpha3`.
2. Apply a `Workload` with `compositePodGroupTemplates`, then the `CompositePodGroup` tree and `PodGroup` leaves it describes, and pods that reference the pod groups through `spec.schedulingGroup.podGroupName`.
3. Open **Scheduling (alpha) → Composite Pod Groups**. The entry only appears when the cluster serves the resource.
4. Open a hierarchy root: it has no Parent row, and its children are listed. Follow a child down, then follow Parent back up.
5. Open a pod group: its Composite Pod Group is now a link to the group above it.
6. Open the Workload: the composite template table shows every nesting level, indented, with the pod group templates named.
7. On a cluster without the `CompositePodGroup` feature gate, confirm the sidebar entry stays hidden.

### PodGroup details 
| Before | After |
|---|---|
| image | image |

### Workload details
| Before | After |
|---|---|
| image | image |

### New: CompositePodGroup details
image

## Notes for the Reviewer

- **`v1alpha3` only, deliberately.** `v1beta1` serves the composite templates inside a `Workload` and the parent reference on a `PodGroup`, but does not register `CompositePodGroup` itself, so this class must not fall back to the other versions the way `PodGroup` and `Workload` do.
- **The sidebar entry has its own availability check**, because a cluster can serve pod groups and workloads without serving this resource.
- **Children are found by listing the namespace and filtering on `parentCompositePodGroupName`** — the API names the parent on the child and there are no owner references to follow.
- **`CompositePodGroupInitiallyScheduled` is terminal**, so it is shown as whether the subtree ever scheduled rather than as live health. The alpha scheduler does not populate it yet, so it reads Unknown on a live cluster; the fixtures cover the Scheduled and Unschedulable cases.
- **i18n adds only two keys.** `npm run i18n` reorders every locale file, so they were inserted in place to keep the diff at one line per file.

### DIFF
--- a/frontend/src/components/scheduling/SchedulingStatus.tsx
+++ b/frontend/src/components/scheduling/SchedulingStatus.tsx
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useTranslation } from 'react-i18next';
+import type { KubeCondition } from '../../lib/k8s/cluster';
+import type { StatusLabelProps } from '../common/Label';
+import { StatusLabel } from '../common/Label';
+
+type Translate = (key: string) => string;
+
+/**
+ * Text of the scheduling condition of a pod group.
+ *
+ * These conditions are terminal: they report whether the group ever met its scheduling
+ * requirement, not whether its pods are running now.
+ * @param condition - The scheduling condition, when the group already has one.
+ * @param t - The translation function of the calling view.
+ * @returns The reason of the condition, or a localized fallback.
+ */
+export function getSchedulingStatusText(
+  condition: KubeCondition | undefined,
+  t: Translate
+): string {
+  if (!condition) {
+    return t('translation|Unknown');
+  }
+  if (condition.reason) {
+    return condition.reason;
+  }
+  if (condition.status === 'True') {
+    return t('translation|Scheduled');
+  }
+  return condition.status === 'False' ? t('translation|Pending') : t('translation|Unknown');
+}
+
+function getSchedulingStatusSeverity(condition: KubeCondition): StatusLabelProps['status'] {
+  if (condition.status === 'True') {
+    return 'success';
+  }
+  return condition.status === 'False' ? 'warning' : '';
+}
+
+/** Shows whether a group met its scheduling requirement, from its scheduling condition. */
+export function SchedulingStatus({ condition }: { condition: KubeCondition | undefined }) {
+  const { t } = useTranslation(['translation']);
+
+  if (!condition) {
+    return <span>{t('translation|Unknown')}</span>;
+  }
+
+  return (
+    <StatusLabel status={getSchedulingStatusSeverity(condition)}>
+      {getSchedulingStatusText(condition, t)}
+    </StatusLabel>
+  );
+}
--- a/frontend/src/lib/k8s/podGroup.ts
+++ b/frontend/src/lib/k8s/podGroup.ts
@@ -150,6 +150,35 @@ export function getDisruptionMode(
   return undefined;
 }
 
+/**
+ * Whether a cluster serves one of the scheduling.k8s.io resources.
+ *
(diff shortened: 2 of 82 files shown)


illume (changes requested):
Thanks for this PR.


Anjali-Chauhan1:
cc: @kannon92 , please review it whenver u get the time . Follow up of #6562 .
Also working on issue [#7530 ](https://github.com/kubernetes-sigs/headlamp/issues/7530) , will open a pr on it within a couple of days...

frontend/src/components/scheduling/SchedulingStatus.tsx
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import { useTranslation } from 'react-i18next';
+import type { KubeCondition } from '../../lib/k8s/cluster';
+import { StatusLabel } from '../common/Label';
+
+type Translate = (key: string) => string;
+
+/**
+ * Text of the scheduling condition of a pod group or a composite pod group.
```
kannon92: could you introduce this change for just pod group first?

I don't really how this is related to CompositePodGroup

Anjali-Chauhan1: I split it out now, #7568 has just the PodGroup list and the new component. Nothing composite in it, and the doc comment no longer mentions CompositePodGroup. This PR is rebased and only adds the composite views now.

frontend/src/lib/k8s/podGroup.ts
```diff
@@ -150,6 +150,35 @@ export function getDisruptionMode(
   return undefined;
 }
 
+/**
```
kannon92, line 153: Please try to limit your code to only what is necessary for the PR.

I don't really see why this cleanup is required. You can do this as a separate PR.

kannon92: ah nvm. I see about in CPG you split this as a function to avoid duplicating this function.

Anjali-Chauhan1: Yeah, CompositePodGroup needs the same version check, so I pulled it out instead of copying it into the new class.

PR #6643, opened 2026-07-26
frontend: Add Storybook stories for CreateDaemonSetForm and Resource components

Suyog241005:
## Summary

This PR expands Storybook visual regression testing coverage by adding `.stories.tsx` files and Jest `.storyshot` snapshot definitions for the newly introduced `CreateDaemonSetForm` component and core `Resource` action components (`EditButton`, `DeleteMultipleButton`, `DetailsDrawer`).

## Related Issue
N/A

## Changes
- **CreateDaemonSetForm**: Added `CreateDaemonSetForm.stories.tsx` containing interactive state wrappers for `Default` and `Filled` DaemonSet form states, along with generated `.storyshot` snapshots.
- **EditButton**: Added `EditButton.stories.tsx` covering `Default` and `CustomStyle` button render states.
- **DeleteMultipleButton**: Added `DeleteMultipleButton.stories.tsx` covering multi-item batch deletion confirmation dialogs.
- **DetailsDrawer**: Added `DetailsDrawer.stories.tsx` covering the resource details drawer component state.
- 
## Steps to Test

1. Launch Storybook locally:
   ```bash
   npm run frontend:storybook
   ```

2. Navigate to `DaemonSet > CreateDaemonSetForm` and `Resource > EditButton / DeleteMultipleButton / DetailsDrawer` in the Storybook UI sidebar.
3. Confirm that all stories render cleanly without console warnings or type errors.
4. Run the frontend test suite to verify snapshot assertions:
      ```bash
   npm run frontend: test
   ```

## Notes for the Reviewer

- [e.g., This touches the i18n layer, so please check language consistency.]

## Summary by CodeRabbit

* **Documentation**
  * Added Storybook examples for resource deletion, details viewing, and editing controls.
  * Added Storybook examples for creating DaemonSets in default and pre-filled states.

* **Tests**
  * Added visual snapshots covering resource actions, the details drawer, and DaemonSet creation forms.
  * Expanded coverage for open and closed drawer states, single-item deletion, custom edit styling, and populated form fields.

Issue #4691, opened 2026-02-12
DeleteMultipleButton (0% coverage) in `frontend/src/components/common/Resource/DeleteMultipleButton.tsx` - Create DeleteMultipleButton.stories.tsx with button enabled showing selection count, button disabled with no selection, batch delete confirmation dialog, delete progress indicator, and partial failure error state


illume:

mastermaxx03:
hey @illume  needed a bit of help
To create stories for delete progress indicator and partial failure error state. How should we go about?
**Current DeleteMultipleButton** - uses `clusterAction`, shows toasts and not in the dialog:
```typescript
dispatch(
  clusterAction(async () => { /* delete */ }, {
    startMessage: t('Deleting...'),
    errorMessage: t('Error deleting...'),
  })
);
```

**AuthChooser pattern** - manual async, shows loading/error in UI:
```typescript
const [testingAuth, setTestingAuth] = React.useState(true);
const [error, setError] = React.useState(null);

testAuth().then(...).catch(err => setError(err));

return <PureAuthChooser testingAuth={testingAuth} error={error} />;
```

Should I follow AuthChooser pattern for DeleteMultipleButton?

illume:
Good question.

I think dialogs are better than snack bars. Snack bars are problematic because they disappear after a time, which is not accessible to some people (or all people who are multitasking).

A-d-i-t-y:
"Hi! I'd like to work on this issue. I'm a new contributor."

A-d-i-t-y:
Fixed in #5345

PR #7547, opened 2026-08-29
frontend: DeleteMultipleButton: Fix invalid HTML, cluster label

### LINKED ISSUE #7546: frontend: DeleteMultipleButton nests a list inside a paragraph
## Describe the bug

`DeleteMultipleButtonDescription` in `frontend/src/components/common/Resource/DeleteMultipleButton.tsx` (the confirm-dialog body shown when bulk-deleting resources) renders a `<ul>` list nested directly inside a `<p>`:

```tsx
function DeleteMultipleButtonDescription(props: DeleteMultipleButtonDescriptionProps) {
  const clusters = uniq(props.items?.map(it => it.cluster));
  return (
    <p>
      {t('Are you sure you want to delete the following items?')}
      <ul>
        {props.items?.map(item => (
          <li key={item.metadata.uid}>
            {item.kind}: {item.metadata.name}{' '}
            {clusters.length > 0 ? `(cluster: ${item.cluster})` : ''}
          </li>
        ))}
      </ul>
    </p>
  );
}
```

`<p>` can only contain phrasing (inline) content — `<ul>` is a block element and isn't valid inside it. Browsers auto-close the `<p>` before the `<ul>` to recover, which produces a DOM structure that doesn't match what React rendered, and can affect how screen readers walk the accessibility tree.

Separately, the cluster label's condition looks off: `clusters.length > 0` is true any time there's at least one selected item, so it fires even when every item is from the *same* cluster — the label seems intended to only appear when items span multiple different clusters (to disambiguate them), not unconditionally.

## Expected behavior

- Valid HTML: the list shouldn't be nested inside the paragraph.
- The `(cluster: ...)` label should only show when the selected items actually span more than one cluster.

## Environment

Yes (I will propose a PR). Fix is to move the `<ul>` out to be a sibling of the `<p>` instead of its child, and change the cluster-label condition to `clusters.length > 1`.

Priyanshu-rgbb:
## Summary

Fixes invalid HTML in the bulk-delete confirm dialog (`<ul>` nested inside a `<p>`), and a per-item cluster label that showed unconditionally instead of only when the selection spans multiple clusters.

## Related Issue

Closes #7546

## Changes

`DeleteMultipleButtonDescription` rendered its `<ul>` as a child of a `<p>`:

```tsx
<p>
  {t('Are you sure you want to delete the following items?')}
  <ul>...</ul>
</p>
```

`<p>` only accepts phrasing (inline) content — a `<ul>` isn't valid inside it, so browsers auto-close the `<p>` before the list to recover, leaving the actual DOM out of sync with what React rendered. Fixed by making the list a sibling of the paragraph instead of nesting it:

```tsx
<>
  <p>{t('Are you sure you want to delete the following items?')}</p>
  <ul>...</ul>
</>
```

Separately, each item's `(cluster: ...)` label was gated on `clusters.length > 0`, which is true for any non-empty selection — so it showed on every item even when they're all from the same cluster. Changed to `clusters.length > 1`, so it only appears when the selection actually spans more than one cluster (its apparent purpose, since the label exists to disambiguate items that would otherwise look identical).

## Steps to Test

1. `cd frontend && npx vitest run src/components/common/Resource/DeleteMultipleButton.test.tsx`
2. Added three tests: valid DOM nesting (no `<ul>` inside a `<p>`), no cluster label when every item is from the same cluster, and the label showing on every item when they span multiple clusters. Confirmed all three fail on `main` and pass with the fix.
3. Manually: select multiple items across different clusters (if using a multi-cluster setup) and open the bulk-delete confirmation — the cluster label should appear per item; with a single cluster it shouldn't.

## Notes for the Reviewer

No dedicated Storybook stories exist yet for `DeleteMultipleButton` (tracked separately by issue #4691 / open PR #6643, which is about adding coverage, not fixing this), and no existing storyshot renders this dialog's open state, so there's nothing to regenerate here. Confirmed no open issue or PR covers this specific bug before filing #7546.

### DIFF
--- a/frontend/src/components/common/Resource/DeleteMultipleButton.test.tsx
+++ b/frontend/src/components/common/Resource/DeleteMultipleButton.test.tsx
@@ -40,7 +40,7 @@ const { MockKubeObject, MockNamespace } = vi.hoisted(() => {
       return this.jsonData?.kind ?? (this.constructor as any).kind;
     }
     get cluster() {
-      return '';
+      return this.jsonData?.cluster ?? '';
     }
     getName() {
       return this.jsonData?.metadata?.name ?? '';
@@ -87,12 +87,13 @@ vi.mock('../../../lib/k8s/pod', () => ({
 import { TestContext } from '../../../test';
 import DeleteMultipleButton from './DeleteMultipleButton';
 
-function makeNamespace(metadata: Record<string, any>) {
+function makeNamespace(metadata: Record<string, any>, cluster?: string) {
   return new (MockNamespace as any)({
     kind: 'Namespace',
     apiVersion: 'v1',
     metadata: { uid: `uid-${metadata.name}`, ...metadata },
     status: { phase: 'Active' },
+    cluster,
   });
 }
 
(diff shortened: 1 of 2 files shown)


Priyanshu-rgbb:
> Reviewed changes

@illume copilot generated no new comments and issues , PR is ready for another look.

Priyanshu-rgbb:
@illume All the copilots comments have been resolved, this PR is ready.

PR #7549, opened 2026-08-30
frontend: DeleteMultipleButton: Fix list nesting in description and cluster label condition

### LINKED ISSUE #7546: frontend: DeleteMultipleButton nests a list inside a paragraph

Bhargava-Ram-Thunga:
## Summary
Fixes invalid HTML in `DeleteMultipleButtonDescription` where a block `<ul>` list was nested inside a phrasing `<p>` tag. Also fixes the cluster disambiguation label condition so that `(cluster: ...)` is only displayed when selected items span multiple distinct clusters.

## Related Issue
Fixes #7546

## Changes
- `frontend/src/components/common/Resource/DeleteMultipleButton.tsx`:
  - Moved `<ul>` out of `<p>` to be sibling elements wrapped in a React fragment (`<></>`), producing valid HTML and preventing browser DOM tree repair discrepancies.
  - Updated cluster label condition from `clusters.length > 0` to `clusters.length > 1` so that items from a single cluster are not redundantly annotated with their cluster name.
- `frontend/src/components/common/Resource/DeleteMultipleButton.test.tsx`:
  - Added unit tests verifying valid DOM structure (sibling `<p>` and `<ul>`).
  - Added unit tests verifying that cluster labels appear only when items span multiple clusters and are omitted for single-cluster deletions.

## Steps to Test
1. Run unit tests:
   ```bash
   cd frontend && npx vitest run src/components/common/Resource/DeleteMultipleButton.test.tsx
   ```
2. Run frontend typecheck and linter:
   ```bash
   npm run frontend:tsc && npm run frontend:lint
   ```

## Notes for the Reviewer
Surgical fix for valid HTML semantics and cleaner dialog text in multi-resource deletion confirmations.


Bhargava-Ram-Thunga:
@Copilot Updated `DeleteMultipleButton.test.tsx` so the single-cluster regression test renders multiple items with the same non-empty cluster (`cluster-1`), explicitly verifying that one distinct cluster suppresses cluster labels in the dialog. Rebased on `main` and all tests/linting pass.

PR #7556, opened 2026-08-30
frontend: AppContainer.stories: Fix react-hooks/exhaustive-deps

Dyfintie:
## Summary

This PR fixes a `react-hooks/exhaustive-deps` violation in `frontend/src/components/App/AppContainer.stories.tsx`. The `prev` variable was used in the `useEffect` cleanup function but was missing from the dependency array. 

## Related Issue

Part of umbrella issue #5183

## Changes

- Fixed missing dependency in `useEffect` in `AppContainer.stories.tsx` by adding `prev` to the dependency array.
- Removed the `// eslint-disable-next-line react-hooks/exhaustive-deps` comment since the rule is no longer violated.

## Steps to Test

1. Check out the branch locally.
2. Run `cd frontend && npx eslint src/components/App/AppContainer.stories.tsx`
3. Observe that the linter passes successfully with no hook violations.
4. Run `npm run test` or check CI to ensure the storybook still functions correctly.

## Notes for the Reviewer

This addresses the second checkbox (`frontend: AppContainer.stories...`) in the `exhaustive-deps` section of #5183.


skoeva (commented):
on re-render, `prev` changes to the mocked `desktopApi`, causing the cleanup to restore the real API too early. please store the original value in a ref and restore it only when the story unmounts

Dyfintie:
Done, I've updated the story to store the initial desktopApi in a React.useRef and run the cleanup only on unmount as suggested. @skoeva  PTAL when you have a moment, thanks!

PR #7568, opened 2026-09-01
frontend: podGroup: Move the scheduling status into its own component

Anjali-Chauhan1:
## Summary

Moves the PodGroup scheduling status rendering into a shared `SchedulingStatus` component, and fixes an `Unknown` scheduling condition being reported as Pending.

## Related Issue

Part of [#7529](https://github.com/kubernetes-sigs/headlamp/issues/7529) . Split out of #7542  at review request, so that the refactor of the existing PodGroup view lands separately from the new CompositePodGroup work.

## Changes

- Added `frontend/src/components/scheduling/SchedulingStatus.tsx`, holding the status text helper and the label
- Updated `podGroup/List.tsx` to use it instead of rendering the status inline
- Fixed a condition with `status: "Unknown"` reading as Pending; it now reads as Unknown, and its label stays neutral instead of warning
- The label now takes the condition rather than the group, so it does not have to know which resource the condition came from

## Steps to Test

1. On a 1.34+ cluster with the WorkloadAwareScheduling feature gate enabled,  navigate to Scheduling -> Pod Groups.
2. Check the Scheduling column: a scheduled group reads its condition reason or Scheduled, an unscheduled one Pending, and    a group whose condition is Unknown or absent reads Unknown with a neutral label.
3. `cd frontend && npm test -- src/components/podGroup`

## Notes for the Reviewer

- Pure refactor apart from the Unknown handling; no new strings, no API changes.
- The component lives under `components/scheduling/` rather than `components/podGroup/`
  because it takes a condition, not a resource.

### DIFF
--- a/frontend/src/components/podGroup/List.tsx
+++ b/frontend/src/components/podGroup/List.tsx
@@ -16,25 +16,8 @@
 
 import { useTranslation } from 'react-i18next';
 import PodGroup from '../../lib/k8s/podGroup';
-import { StatusLabel } from '../common/Label';
 import ResourceListView from '../common/Resource/ResourceListView';
-
-/** Shows whether the group met its scheduling requirement, from its scheduling condition. */
-function SchedulingStatus({ podGroup }: { podGroup: PodGroup }) {
-  const { t } = useTranslation(['translation']);
-  const condition = podGroup.schedulingCondition;
-
-  if (!condition) {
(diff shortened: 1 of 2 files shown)


Anjali-Chauhan1:
cc: @kannon92

kannon92 (commented):
/lgtm

/assign @illume

PR #7594, opened 2026-09-04
fix(backend/serviceproxy): preserve URIPrefix base path and guard against prefix escape

zanarellidev:
### What this PR does
Ensures that `serviceproxy.Connection.Get` preserves the configured `proxyService.URIPrefix` base path and prevents relative paths with leading slashes or parent segments from escaping above the base prefix.

### Details
1. When `c.URI` contains a path prefix (e.g. `http://service:8080/api/v1/proxy`), `base.ResolveReference(rel)` with an absolute path previously resolved against the host root, dropping the base path prefix.
2. `combinedPath` joins `baseClean` and `relClean`, verifying that the resulting path remains scoped within `baseClean`.
3. Adds unit tests in `pkg/serviceproxy/connection_test.go` verifying prefix preservation and traversal protection.

### DIFF
--- a/backend/pkg/serviceproxy/connection.go
+++ b/backend/pkg/serviceproxy/connection.go
@@ -45,14 +45,32 @@ func (c *Connection) Get(ctx context.Context, requestURI string, w http.Response
 
 	if rel.Path != "" {
 		cleaned := path.Clean(rel.Path)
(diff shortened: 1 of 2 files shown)


zanarellidev:
Hi @ashu8912 and @skoeva, could you please take a look when available? I am happy to address any feedback or rebase if needed.

PR #7597, opened 2026-09-04
fix: decode Base64 secret values with js-base64 to support UTF-8 characters

### LINKED ISSUE #7596: frontend: buildEnvironmentVariables crashes on secrets with UTF-8 data
## Describe the bug

Opening the container details for a Pod that references a Secret containing UTF-8 characters causes the **Environment Variables** section to crash with:

```text
InvalidCharacterError: Failed to execute 'atob' on 'Window'
```

The issue occurs in `frontend/src/components/common/Resource/Resource.tsx`, where `buildEnvironmentVariables` uses the native `atob()` function to decode Secret values:

```ts
const value = secretData[ref.key!] ? atob(secretData[ref.key!]) : '';
```

and:

```ts
variables.set(`${prefix}${key}`, {
    value: atob(value as string),
    ...
});
```

Native `atob()` decodes Base64 into a Latin-1 string and does not correctly handle the resulting UTF-8 data. As a result, Secret values containing multibyte UTF-8 characters, such as accented characters, symbols, or non-Latin text, can cause an uncaught `InvalidCharacterError`.

The expected behavior is for Secret values to be decoded correctly regardless of their character encoding.

`Base64.decode()` from `js-base64` should be used instead, as it is already available in `Resource.tsx` and correctly handles UTF-8 strings.

## To Reproduce

Steps to reproduce the bug:

1. Create a Kubernetes Secret containing UTF-8 characters. For example:

   ```bash
   echo -n "pässwörd🚀" | base64
   ```

   This produces:

   ```text
   cMOkc3N3w7ZyZPCfmYA=
   ```
2. Reference the Secret key from a Pod using either `secretKeyRef` or `envFrom.secretRef`.
3. Open the Pod's details page in Headlamp.
4. Expand the container section.
5. Open or view the **Environment Variables** section.
6. The section crashes with an `InvalidCharacterError`.

* Installation type: Development build (main branch)
* Headlamp Version: 0.45.0
* Other: Chrome / Firefox / Safari

Yes (I will propose a PR).

## Additional Context

`Base64` from `js-base64` is already imported in `Resource.tsx` and is used elsewhere in the file, including `SecretField`.

Replacing the native `atob(...)` calls in `buildEnvironmentVariables` with `Base64.decode(...)` should resolve the issue while reusing the existing Base64 utility:

```ts
Base64.decode(value as string)
```

This also keeps Secret decoding consistent with the existing implementation in `Resource.tsx`.

This appears to affect Secrets whose decoded values contain multibyte UTF-8 characters; ASCII-only Secret values continue to work as expected.

Kunal241207:
## Summary

Fixes a crash in the Pod/container environment variables panel when Secret values
contain non-ASCII characters, such as emojis, CJK, Arabic, or accented Latin characters.

The issue was caused by using the native browser `atob()` function to decode
Kubernetes Secret data. `atob()` decodes Base64 into a Latin-1 binary string
rather than interpreting the result as UTF-8, so any Secret value containing
multi-byte characters is either silently corrupted or causes a `DOMException`,
preventing the environment variables panel from rendering correctly.

This PR replaces both `atob()` calls in `buildEnvironmentVariables` with
`Base64.decode()` from the existing `js-base64` dependency, which correctly
decodes Base64-encoded UTF-8 values into proper JavaScript strings.

## Related Issue

Fixes #7596 

## Changes

* `frontend/src/components/common/Resource/Resource.tsx`

  * Replaced the `atob()` calls used when resolving:

    * `env[].valueFrom.secretKeyRef`
    * `envFrom[].secretRef`
  * Uses `Base64.decode()` to correctly decode UTF-8 Secret values.
  * Added a null guard for empty Secret values.
  * Exported `extractEnvVarReferences` and `buildEnvironmentVariables` to make them directly unit-testable.

* `frontend/src/components/common/Resource/EnvironmentVariables.test.ts`

  * Added unit tests covering multi-byte UTF-8 Secret values for:

    * `secretKeyRef`
    * `secretRef` / `envFrom`

## Steps to Test

1. Create a Kubernetes Secret containing non-ASCII characters:

   ```bash
   kubectl create secret generic utf8-test \
     --from-literal=API_KEY='pässwörd_🚀_日本語_العربية'
   ```

2. Create a Pod that references the Secret using either `secretKeyRef` or `envFrom.secretRef`.

3. In Headlamp, open the Pod details and view the container's **Environment Variables**.

4. Verify that the Secret value is displayed correctly.

### Before

The environment variables panel can fail to render or throw a `DOMException` when resolving a Secret value containing multi-byte UTF-8 characters.

### After

The Secret value is decoded and displayed correctly, including emojis, CJK, Arabic, and accented characters.

### Unit tests

Run:

```bash
cd frontend
npm test -- src/components/common/Resource/EnvironmentVariables.test.ts --run
```

Expected result:

```text
Test Files  1 passed (1)
Tests       2 passed (2)
```

## Notes for Reviewers

* `js-base64` is already a direct frontend dependency, so no new dependency is introduced.
* Only Kubernetes Secret decoding paths are changed. ConfigMap values are plain strings and are unaffected.
* Exporting `extractEnvVarReferences` and `buildEnvironmentVariables` is limited to enabling direct unit testing; no restructuring of the existing implementation is required.

### DIFF
--- a/frontend/src/components/common/Resource/EnvironmentVariables.test.ts
+++ b/frontend/src/components/common/Resource/EnvironmentVariables.test.ts
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Base64 } from 'js-base64';
+import { describe, expect, it, vi } from 'vitest';
+import { KubeContainer } from '../../../lib/k8s/cluster';
+import { KubePod } from '../../../lib/k8s/pod';
+
+const { MockKubeObject } = vi.hoisted(() => {
+  class MockKubeObject {
+    jsonData: any;
+    static kind = '';
+    constructor(data: any) {
+      this.jsonData = data;
+    }
+  }
+  return { MockKubeObject };
+});
+
+vi.mock('../../../lib/k8s/KubeObject', () => ({ KubeObject: MockKubeObject }));
+vi.mock('../../../lib/k8s/deployment', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/replicaSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/statefulSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/daemonSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/job', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/pod', () => ({ default: class extends MockKubeObject {} }));
+
+import { buildEnvironmentVariables, extractEnvVarReferences } from './Resource';
+
+describe('buildEnvironmentVariables', () => {
+  const dummyPod = {
+    metadata: {
+      name: 'test-pod',
+      namespace: 'default',
+      creationTimestamp: '2025-01-01T00:00:00Z',
+    },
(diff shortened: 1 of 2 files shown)


Kunal241207:
@joaquimrocha @kahirokunn Could any of you please review this PR when you get a chance? Thanks!

kahirokunn (approved):
/lgtm

kahirokunn:
Thank you 🙏

illume:
Please be sure to check commit message format matches commit guidelines.

Looks like some test is failing, there's a fix in here:
- https://github.com/kubernetes-sigs/headlamp/pull/7617

PR #7605, opened 2026-09-06
frontend: resourceMap: Visualize the workload aware scheduling objects

### LINKED ISSUE #7530: frontend: Visualize how the scheduling API objects relate in the resource map

Anjali-Chauhan1:
## Summary

This PR adds the workload aware scheduling objects to the resource map, so a gang scheduled workload can be seen as one graph instead of one object at a time. It adds Workload, CompositePodGroup and PodGroup as map sources, the edges that nest them and tie them to their Pods, and a node status that marks a group that is not scheduled.

## Related Issue

Fixes #7530

## Changes

* Added `useSchedulingApisEnabled()` (`frontend/src/lib/k8s/schedulingApis.ts`): the alpha APIs are only served when the GenericWorkload feature gate is on, so views built on them stay hidden otherwise. The sidebar already probed for this, so the probe moved into one hook that both it and the map use.
* Added a `Scheduling (alpha)` source group with Workload, CompositePodGroup and PodGroup. It is off by default, matching `Gateway (beta)`, `Cluster` and `Security`.
* Added five relations, registered in `BUILT_IN_RELATION_IDS`:

  * `podgroup-workload`, from `spec.workloadRef`
  * `pod-podgroup`, from the Pod's `spec.schedulingGroup.podGroupName`
  * `compositepodgroup-workload`, for a root composite group
  * `compositepodgroup-parent`, from `spec.parentCompositePodGroupName`
  * `podgroup-compositepodgroup`, for a group nested under a composite
* Added node status for PodGroup and CompositePodGroup, derived from their scheduling conditions.
* Mapped the kinds to icons by API group and put them in the workloads colour group.
* Tests for every relation, the source gating and both status helpers.

## Steps to Test

1. On a cluster with the GenericWorkload and CompositePodGroup feature gates enabled, create a Workload whose templates nest composite groups, and let its pods schedule.
2. Open **Map** and confirm the source picker lists **Scheduling (alpha)**, unticked.
3. Tick it, then click **Expand All**. The Pods, their PodGroups, the composite groups they nest under and the Workload are drawn as one connected graph.
4. Confirm a PodGroup under a composite group hangs off that group, not off the Workload, and that a flat PodGroup with no composite parent still links straight to the Workload.
5. Create a PodGroup whose gang cannot be satisfied (e.g. `minCount: 8` with two pods). Its node is marked with a warning, and **Status: Error or Warning** keeps it.
6. On a cluster without the feature gates, confirm the group does not appear at all.

### A whole workload in one graph: Pods → PodGroups → CompositePodGroups → Workload

WhatsApp Image 2026-09-06 at 1 38 37 PM

### An unscheduled gang is marked on the map without opening it

WhatsApp Image 2026-09-06 at 1 39 27 PM

## Notes for the Reviewer

* **Builds on #7542 — please merge that first.** The CompositePodGroup work needs its resource class, so this branch merges it in and its six commits show here. They drop out once #7542 merges and this is rebased.
* Edges point child to parent, as the owner reference relations do. These objects have no owner references, so the hierarchy comes from the spec fields the issue names. A PodGroup under a composite is guarded so it does not also shortcut straight to the Workload.
* Composite status warns only on an explicit `False`, unlike PodGroup which warns when the condition is absent. The composite condition is terminal and not set by every controller, so treating a missing one as unscheduled marked healthy groups forever on a real cluster.
* Type guards check the API group as well as the kind, so an unrelated resource of the same kind (e.g. Volcano's PodGroup) is not matched. They use type-only imports — importing the classes at runtime creates a cycle that breaks `KubeObjectNode.test.tsx`.
* No new i18n keys. Icons reuse existing SVGs as there are none upstream for these kinds — happy to add dedicated assets.

### DIFF
--- a/frontend/src/components/Sidebar/useSidebarItems.tsx
+++ b/frontend/src/components/Sidebar/useSidebarItems.tsx
@@ -15,7 +15,6 @@
  */
 
 import { useTheme } from '@mui/material/styles';
-import { useQuery } from '@tanstack/react-query';
 import _ from 'lodash';
 import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -24,7 +23,7 @@ import { isElectron } from '../../helpers/isElectron';
 import { useClustersConf, useSelectedClusters } from '../../lib/k8s';
 import CRD from '../../lib/k8s/crd';
 import { useGatewayL4RouteAvailability } from '../../lib/k8s/gatewayL4RouteAvailability';
-import PodGroup from '../../lib/k8s/podGroup';
+import { useSchedulingApisEnabled } from '../../lib/k8s/schedulingApis';
 import { createRouteURL } from '../../lib/router/createRouteURL';
 import { useTypedSelector } from '../../redux/hooks';
 import { DefaultSidebars, SidebarEntryProps, SidebarItemProps } from '.';
@@ -78,18 +77,7 @@ export const useSidebarItems = (sidebarName: string = DefaultSidebars.IN_CLUSTER
     console.error('Failed to fetch CRDs:', error);
   }
 
-  // The workload aware scheduling APIs are alpha and are only served when the cluster
-  // enables the GenericWorkload feature gate, so only show them when they are available.
-  const { data: schedulingWorkloadsEnabled = false } = useQuery({
-    queryKey: ['schedulingWorkloadsEnabled', ...selectedClusters],
-    queryFn: async () => {
-      const enabledPerCluster = await Promise.all(
-        selectedClusters.map(cluster => PodGroup.isEnabled(cluster))
-      );
-      return enabledPerCluster.some(Boolean);
-    },
-    enabled: selectedClusters.length > 0,
-  });
+  const schedulingWorkloadsEnabled = useSchedulingApisEnabled();
 
   const crdsSidebarEntries = useMemo(() => {
     const crdsSidebarEntries: SidebarItemProps[] = [];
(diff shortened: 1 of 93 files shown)


Anjali-Chauhan1:
cc : @kannon92 

This branch merges [#7542,](https://github.com/kubernetes-sigs/headlamp/pull/7542) as CompositePodGroup part needs the resource class from it. That is why I made six of #7542 commits show in this diff. They drop out once it merges and I rebase this on main.

Holding so this does not merge ahead of #7542. Will cancel the hold after the rebase.

Anjali-Chauhan1:
@kannon92 , please review whenever you get the chance  :)

PR #7607, opened 2026-09-06
Remount plugin settings page when navigating between plugins

magic-peach:
## Summary

This PR fixes a bug where navigating directly from one plugin's settings page to another's could leave the previous plugin's unsaved data on screen, and persist it over the new plugin's config on Save.

## Related Issue

Found by reading the code, not tied to an existing issue.

## Changes

- Fixed `PluginSettingsDetails` so it remounts when the viewed plugin changes, instead of reusing the same component instance with stale internal state
- Added a regression test covering navigation between two different plugins' settings pages

## Steps to Test

1. Install two plugins that each expose a settings component with a save button.
2. Open the first plugin's settings page and edit a field without saving.
3. Navigate directly to the second plugin's settings page (e.g. via browser back/forward between two previously visited settings URLs).
4. Observe the second plugin's settings show its own current config, not leftover data from the first plugin.

Not applicable; behavior is only visible mid-navigation and is covered by the added test.

## Notes for the Reviewer

- `PluginSettingsDetailsPure` seeds its `data` state from the `config` prop with `useState`, which only runs once on mount. Without a remount, navigating between plugins re-renders the same instance with a new `config` but doesn't reset `data`.
- Fixed by keying `PluginSettingsDetailsInitializer` by the plugin's name and type in `PluginSettingsDetails`, so switching plugins forces React to remount it with fresh state.

### DIFF
--- a/frontend/src/components/App/PluginSettings/PluginSettingsDetails.test.tsx
+++ b/frontend/src/components/App/PluginSettings/PluginSettingsDetails.test.tsx
@@ -15,9 +15,13 @@
  */
 
 import { ThemeProvider } from '@mui/material/styles';
-import { render, waitFor } from '@testing-library/react';
+import { act, render, waitFor } from '@testing-library/react';
+import { createMemoryHistory } from 'history';
+import { Provider } from 'react-redux';
+import { Route, Router } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


skitt:

/close

PR #7610, opened 2026-09-07
fix: Ingress rules show http:// instead of https:// when a TLS entry covers multiple hosts

### LINKED ISSUE #7609: Ingress: HTTPS protocol not applied when a TLS entry lists multiple hosts
## Describe the bug

When an Ingress resource has a TLS entry that covers more than one hostname (which is standard Kubernetes YAML), Headlamp displays all of those hostnames as `http://` links in the Rules table — even though HTTPS is correctly configured for them.

The root cause is in `isHttpsUsed` in `frontend/src/components/ingress/Details.tsx`. The function maps over the TLS entries and converts each entry's `hosts` array into a string using a template literal:

```ts
const hostList = item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`) ?? [];
```

When `hosts` is `['foo.example.com', 'bar.example.com']`, the template literal coerces the array to `"foo.example.com,bar.example.com"` — a single comma-joined string — so `hostList.includes('foo.example.com')` is always `false`. The protocol falls back to `http://` for both hosts.

The bug does not affect TLS entries with a single host, because `['only.example.com']` stringifies to `"only.example.com"`, which accidentally matches.

## To Reproduce

Steps to reproduce the bug:

1. Create an Ingress with two hosts sharing one TLS entry:

```yaml
kind: Ingress
metadata:
  name: multi-host-tls
  namespace: default
spec:
  tls:
    - hosts:
        - foo.example.com
        - bar.example.com
      secretName: my-tls-secret
  rules:
    - host: foo.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: foo-svc
                port:
                  number: 80
    - host: bar.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: bar-svc
                port:
                  number: 80
```

2. Apply the manifest to a cluster.
3. Open Headlamp and navigate to **Networking → Ingresses → multi-host-tls**.
4. Observe the Rules table — both `foo.example.com` and `bar.example.com` are linked with `http://` instead of `https://`.

**Expected:** Both hosts are linked with `https://` since they are listed in the TLS spec.

**Actual:** Both hosts are linked with `http://`, despite TLS being configured.

## Environment

* Installation type: any
* Headlamp Version: current `main` (confirmed in `frontend/src/components/ingress/Details.tsx`)

Yes (I will propose a PR).

Kunal241207:
## Summary

Fixes a bug where Ingress rules displayed `http://` links for hostnames covered by TLS when multiple hostnames shared a single TLS entry.

The `isHttpsUsed` function in `frontend/src/components/ingress/Details.tsx` built the host list by mapping over TLS entries and coercing each entry's `hosts` array to a string using a template literal:

```ts
item.jsonData.spec?.tls?.map(({ ...hosts }) => `${hosts.hosts}`)
```

When a TLS entry contains multiple hosts, `hosts.hosts` is an array such as `['foo.example.com', 'bar.example.com']`. The template literal converts it to `"foo.example.com,bar.example.com"`, so `includes('foo.example.com')` never matches and both hosts fall back to `http://`.

Single-host TLS entries were unaffected because `['only.example.com']` stringifies to `"only.example.com"`, which happens to match.

The fix replaces `map` with `flatMap` so each hostname is kept as an individual element:

```ts
item.jsonData.spec?.tls?.flatMap(entry => entry.hosts ?? [])
```

## Related Issue

Fixes #7609

## Changes

* `frontend/src/components/ingress/Details.tsx`: replaced the `map` + template literal with `flatMap` in `isHttpsUsed` to correctly collect individual hostnames across TLS entries.
* `frontend/src/components/ingress/Details.test.tsx`: added a regression test verifying that both hosts in a multi-host TLS entry resolve to `https://` links.

## Steps to Test

1. Create an Ingress with two hosts under one TLS entry:

   ```yaml
   spec:
     tls:
       - hosts:
           - foo.example.com
           - bar.example.com
         secretName: my-tls-secret
     rules:
       - host: foo.example.com
         ...
       - host: bar.example.com
         ...
   ```
2. Open Headlamp → **Networking → Ingresses** → select the Ingress.
3. Check the Rules table.

**Before:** `foo.example.com` and `bar.example.com` are linked with `http://`.

**After:** Both are correctly linked with `https://`.

Alternatively, run the regression test directly:

```bash
cd frontend
npm test -- src/components/ingress/Details.test.tsx --run
```

Expected: `Test Files 1 passed (1), Tests 4 passed (4)`.

## Notes for Reviewers

* The change is a single-line fix in `isHttpsUsed`. No other behaviour is altered.
* TLS entries with a single host continue to work identically — the existing three tests confirm this.

### DIFF
--- a/frontend/src/components/ingress/Details.test.tsx
+++ b/frontend/src/components/ingress/Details.test.tsx
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { vi } from 'vitest';
 import { TestContext } from '../../test';
 import IngressDetails from './Details';
+import { LinkStringFormat } from './Details';
 
 const { mockDetailsGrid } = vi.hoisted(() => ({
   mockDetailsGrid: vi.fn(),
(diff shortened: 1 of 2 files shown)


Kunal241207:
@kahirokunn @sniok Hello! could anyone please review this PR? Thank u.

Kunal241207:
@illume Hello, can you please review this PR, Thanks!

Kunal241207:
The CI failure (`Build Frontend / test (22.x, ubuntu-22.04) (pull_request)`) appears unrelated to this PR’s Ingress TLS change. The failure is in src/plugin/pluginLib.test.ts due to a snapshot mismatch

Kunal241207:
Hey! @kahirokunn @sniok  Just bumping this in case it got buried. Would appreciate a review whenever you get a chance :)

PR #7618, opened 2026-09-07
helm: list one entry per chart, not per chart version

magic-peach:
## Summary

This PR fixes ListCharts returning one row per historical chart version instead of one row per chart.

## Related Issue

Found by reading the code, not tied to an existing issue.

## Changes

- Fixed `listCharts` to build its search index with `all: false`, so only the latest version of each chart is indexed
- Added a test asserting exactly one entry per chart name against a real repository with multiple published versions

## Steps to Test

1. Add a Helm repository that has published multiple versions of at least one chart (the existing tests use `https://kubernetes-sigs.github.io/headlamp/`, whose `headlamp` chart has many releases).
2. Call `GET /clusters/{cluster}/helm/repositories/charts`.
3. Before this fix, the response contains one entry per published version of the chart. After this fix, it contains exactly one entry per chart name.

Not applicable; this is a backend API response, not a UI change.

## Notes for the Reviewer

- `index.AddRepo(name, indexFile, true)` passes Helm's own `all: true`, which indexes every version of every chart (confirmed by reading the vendored `helm.sh/helm/v3` search package). Changed to `false` so the index keeps only the newest version per chart, matching what `ListAllChartsResponse` expects.
- This endpoint isn't currently consumed by any frontend component, so today this is a latent API defect rather than something visible through the UI.

### DIFF
--- a/backend/pkg/helm/charts.go
+++ b/backend/pkg/helm/charts.go
@@ -63,7 +63,7 @@ func listCharts(filter string, settings *cli.EnvSettings) ([]chartInfo, error) {
 			return nil, err
 		}
 
-		index.AddRepo(name, indexFile, true)
+		index.AddRepo(name, indexFile, false)
 
 		for _, chart := range index.All() {
 			if filter != "" {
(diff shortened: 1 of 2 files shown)


skitt:

/close

PR #7620, opened 2026-09-07
frontend: Resource: Encode secret env var values with js-base64

### LINKED ISSUE #7619: frontend: Container env vars from Secrets crash the details view on non-Latin-1 values
## What happened

`ContainerEnvironmentVariables` decodes a Secret's value correctly, then re-encodes it
with `btoa()` before handing it to `SecretField`.

`frontend/src/components/common/Resource/Resource.tsx:1276`:

```tsx
{data.isSecret ? (
  <SecretField
    disableUnderline
    value={btoa(data.value)}
    sx={{ fontFamily: 'monospace' }}
  />
) : (
```

`btoa()` only accepts code points in the Latin-1 range (U+0000–U+00FF). Everything
else raises `InvalidCharacterError`. The exception escapes during render, so the whole
Containers section of the Pod details page fails instead of just that one row.

The round trip through this component is:

| Step | Location | Function |
| --- | --- | --- |
| decode the Secret | `Resource.tsx:915`, `Resource.tsx:974` | `Base64.decode` (js-base64, UTF-8 aware) |
| re-encode for display | `Resource.tsx:1276` | **`btoa`** (Latin-1 only) |
| decode to reveal / copy | `Resource.tsx:646`, `Resource.tsx:602` | `Base64.decode` (js-base64, UTF-8 aware) |

Only the middle step is inconsistent, and it fails in two distinct ways.

### Failure 1 — crash

Any character outside Latin-1 throws:

```
Uncaught InvalidCharacterError: Failed to execute 'btoa' on 'Window':
The string to be encoded contains characters outside of the Latin1 range.
```

### Failure 2 — silent corruption

Characters in the Latin-1 supplement (U+00A0–U+00FF) do not throw. `btoa` emits one
byte per character, but `Base64.decode` reads the result back as UTF-8, so the byte is
not valid UTF-8 on its own and becomes a replacement character. The revealed value is
wrong, and the copy-to-clipboard button copies the wrong value too.

Comparing the current path against `Base64.encode`:

| Secret value | Rendered today | Expected |
| --- | --- | --- |
| `hello` | `hello` | `hello` |
| `café` | `caf�` | `café` |
| `пароль` | throws `InvalidCharacterError` | `пароль` |
| `日本語` | throws `InvalidCharacterError` | `日本語` |
| `pw🔐` | throws `InvalidCharacterError` | `pw🔐` |

## Steps to reproduce

```bash
kubectl create secret generic utf8-secret --from-literal=API_KEY='pässwörd_🚀_日本語'

kubectl run utf8-demo --image=nginx --restart=Never --overrides='
{
  "spec": {
    "containers": [{
      "name": "utf8-demo",
      "image": "nginx",
      "env": [{
        "name": "API_KEY",
        "valueFrom": { "secretKeyRef": { "name": "utf8-secret", "key": "API_KEY" } }
      }]
    }]
  }
}'
```

Open the Pod's details page and expand the container. The Containers section fails to
render and `InvalidCharacterError` appears in the browser console.

Replacing the emoji and CJK text with `café` shows the second failure instead: the
section renders, but revealing the value shows `caf�`.

## Why this slipped through

`frontend/src/components/common/Resource/EnvironmentVariables.test.ts` already covers
multibyte UTF-8 Secret values:

- `correctly decodes secrets with multibyte UTF-8 data for secretKeyRef`
- `correctly decodes secrets with multibyte UTF-8 data for secretRef (envFrom)`

Both pass. They exercise `buildEnvironmentVariables`, which is a pure function and does
use `Base64.decode` correctly. Neither renders the component, so the `btoa` call on the
render path is untested.

## Suggested fix

`Base64` from `js-base64` is already imported in this file at `Resource.tsx:31`, so the
change is to use it on the encode side as well:

```diff
-              value={btoa(data.value)}
+              value={Base64.encode(data.value)}
```

This matches the fix applied in #7597 for the Secret details page, which moved that
page's decoding to `js-base64` for the same reason.

## Environment

- Branch: `main` at `b127881aa`
- Affects both the web and desktop builds; nothing here is platform specific.

0PrashantYadav0:
## Summary

This PR fixes a crash and a silent data-corruption bug in the container environment
variables view by encoding Secret values with `js-base64` instead of `btoa()`.

`ContainerEnvironmentVariables` decoded a Secret's value with `Base64.decode` and then
re-encoded it with `btoa()` before handing it to `SecretField`, which decodes it again
with `Base64.decode`. `btoa()` only accepts Latin-1, so the round trip broke in two ways:

- Values containing anything outside Latin-1 (Cyrillic, CJK, emoji) raised
  `InvalidCharacterError` during render, taking down the whole Containers section of the
  Pod details page.
- Values in the Latin-1 supplement (é, ü, ñ) did not throw, but `btoa` emits one byte per
  character, which is not valid UTF-8 on its own. Revealing the value showed a
  replacement character, and the copy button copied the wrong text.

## Related Issue

Fixes #7619

## Changes

- Updated `frontend/src/components/common/Resource/Resource.tsx` to encode secret env var
  values with `Base64.encode` rather than `btoa`, so both sides of the round trip agree on
  UTF-8. `Base64` was already imported in this file, so the change is one line.
- Added `frontend/src/components/common/Resource/ContainerEnvironmentVariables.test.tsx`,
  a render test covering both failure modes with an ASCII case as a control.

## Steps to Test

1. Create a Secret with a multibyte value and a Pod that consumes it:

   ```bash
   kubectl create secret generic utf8-secret --from-literal=API_KEY='pässwörd_🚀_日本語'

   kubectl run utf8-demo --image=nginx --restart=Never --overrides='
   {
     "spec": {
       "containers": [{
         "name": "utf8-demo",
         "image": "nginx",
         "env": [{
           "name": "API_KEY",
           "valueFrom": { "secretKeyRef": { "name": "utf8-secret", "key": "API_KEY" } }
         }]
       }]
     }
   }'
   ```

2. Open the Pod's details page and expand the container. On `main` the Containers section
   fails to render and `InvalidCharacterError` appears in the browser console. With this
   change the section renders normally.
3. Click the eye icon on the `API_KEY` row. The original value is shown, and the copy
   button now copies it correctly.
4. Repeat with `--from-literal=API_KEY='café'` to see the second failure mode: on `main`
   the section renders but reveals `caf<?>`, and with this change it reveals `café`.
5. Run the tests:

   ```bash
   cd frontend
   npx vitest run src/components/common/Resource/ContainerEnvironmentVariables.test.tsx
   ```

   Reverting just the one-line change in `Resource.tsx` makes 6 of the 7 cases fail: the
   four multibyte ones with `InvalidCharacterError`, and the `café` one with
   `Unable to find an element with the display value: café` against a rendered `caf<?>`.
   The ASCII case passes either way, which is what makes it a useful control.

## Notes for the Reviewer

- This mirrors the change made for the Secret details page in #7597. The env var render
  path was the remaining place still using the platform `btoa`.
- `frontend/src/components/common/Resource/EnvironmentVariables.test.ts` already asserts
  that multibyte UTF-8 Secret values decode correctly, and those tests pass on `main`.
  They exercise `buildEnvironmentVariables`, which is pure and does use `Base64.decode`
  correctly, so they never reach the `btoa` call on the render path. That is why the bug
  survived despite apparently having coverage.
- The new test file mocks `lib/k8s/KubeObject` and the workload classes the same way the
  neighbouring `EnvironmentVariables.test.ts` does. That is not stylistic: importing
  `Resource.tsx` in a test without those mocks hits a genuine import cycle through
  `lib/k8s/index.ts` and fails with `Class extends value undefined is not a constructor`.
- `tsc --noEmit`, eslint and prettier are clean. The full
  `src/components/common/Resource/` suite passes (19 files, 159 tests).


0PrashantYadav0:
@joaquimrocha @vyncent-t Will you please review this PR when you get free time? Thanks!

PR #7623, opened 2026-09-08
frontend: Resource: Scope container env var lookups to the pod's cluster

### LINKED ISSUE #7622: frontend: Container env vars resolve Secrets and ConfigMaps against the wrong cluster
## What happened

`ContainerEnvironmentVariables` renders one fetcher per referenced Secret and ConfigMap
to resolve `secretKeyRef` / `configMapKeyRef` / `envFrom` values. Neither fetcher is
given a cluster.

`frontend/src/components/common/Resource/Resource.tsx:834`:

```tsx
function SecretFetcher(props: { ... }) {
  const { name, namespace, onResult } = props;
  const [secret, error] = Secret.useGet(name, namespace);
```

`frontend/src/components/common/Resource/Resource.tsx:856`:

```tsx
function ConfigMapFetcher(props: { ... }) {
  const { name, namespace, onResult } = props;
  const [configMap, error] = ConfigMap.useGet(name, namespace);
```

`useGet` takes the cluster in its third argument (`KubeObject.ts:463-472`). With it
omitted, both calls fall back to the cluster in the URL rather than the cluster the pod
being displayed actually belongs to.

The cluster is available at the call site. `ContainerInfo` already declares
`ContainerInfoKubeObjectProps` with `resource: KubeObject | null`, and all four callers
of `ContainersSection` (`pod`, `daemonset`, `statefulset`, `job` details) pass the
KubeObject, which carries `.cluster`. It is simply never forwarded:

`frontend/src/components/common/Resource/Resource.tsx:1685`:

```tsx
value: <ContainerEnvironmentVariables pod={resource as KubePod} container={container} />,
```

## Impact

When the pod on screen is not from the URL cluster, the values shown come from a
different cluster's Secret or ConfigMap of the same name in the same namespace:

- If both clusters have `default/app-secrets`, Headlamp shows **cluster A's secret
  material on cluster B's pod page**, with no indication anything is wrong. The
  copy-to-clipboard button copies the wrong value.
- If only the pod's own cluster has it, the row shows a spurious `404`.

The same omission affects navigation. The "From" column links to the Secret or ConfigMap
without `activeCluster`, so the link goes to the wrong cluster's resource:

`frontend/src/components/common/Resource/Resource.tsx:1307-1317`.

## Steps to reproduce

With two clusters connected, both having a `default/app-secrets` Secret with different
values for the same key, and a pod in cluster B consuming it via `secretKeyRef`:

1. Select both clusters so the resource map aggregates them.
2. Open the map and click the cluster B pod. `KubeObjectDetails` renders the details
   with `cluster={resource.cluster}` while the URL still points at cluster A.
3. Expand the container and reveal the env var.

The value shown is cluster A's. The "From" link also navigates to cluster A's Secret.

## Suggested fix

Thread the pod's cluster from `ContainerInfo` down to the two fetchers and pass it to
`useGet`, and set `activeCluster` on the "From" link. `ContainerInfo` already has the
`KubeObject` in `resource`, so nothing above it needs to change.

When `resource` is a plain `KubeObjectInterface` (the deprecated
`ContainerInfoLegacyProps` shape) the cluster is `undefined`, which is exactly the
current fallback behaviour, so nothing regresses.

## Related

Same class of bug as the CronJob details page fetching its object without the `cluster`
prop it was given. This is the container env var path of the same problem.

## Environment

- Branch: `main` at `b127881aa`

0PrashantYadav0:
## Summary

This PR fixes a multi-cluster bug where a container's environment variables were resolved
against the wrong cluster, by forwarding the pod's cluster to the Secret and ConfigMap
lookups.

`SecretFetcher` and `ConfigMapFetcher` called `useGet` without a cluster:

```tsx
const [secret, error] = Secret.useGet(name, namespace);
const [configMap, error] = ConfigMap.useGet(name, namespace);
```

`useGet` takes the cluster in its third argument (`KubeObject.ts:463-472`). With it
omitted, both fall back to the cluster in the URL rather than the cluster the pod on
screen belongs to. When those differ, the values shown come from a same-named Secret or
ConfigMap in a different cluster, with nothing to indicate it, and the copy button copies
the wrong value. Where only the pod's own cluster has the resource, the row shows a
spurious 404 instead. The "From" link had the same problem and navigated to the wrong
cluster.

The cluster was already available at the call site and simply not passed on.
`ContainerInfo` declares `ContainerInfoKubeObjectProps` with `resource: KubeObject | null`,
and all four callers of `ContainersSection` (`pod`, `daemonset`, `statefulset`, `job`
details) pass the KubeObject, which carries `.cluster`.

## Related Issue

Fixes #7622

## Changes

- Updated `frontend/src/components/common/Resource/Resource.tsx`:
  - `EnvironmentVariablesProps` gains an optional `cluster`.
  - `SecretFetcher` and `ConfigMapFetcher` accept a `cluster` and pass it to `useGet`.
  - `ContainerEnvironmentVariables` forwards it to both fetchers and sets `activeCluster`
    on the "From" link.
  - `ContainerInfo` reads the cluster off its `resource` and hands it down.
- Added `frontend/src/components/common/Resource/ContainerEnvironmentVariables.test.tsx`
  covering the cluster forwarding, the `ContainerInfo` handoff, and the no-cluster
  fallback.

Nothing above `ContainerInfo` changes. Callers on the deprecated
`ContainerInfoLegacyProps` shape pass `undefined`, which is exactly the current fallback
to the URL cluster, so no existing behaviour moves.

## Steps to Test

1. Connect two clusters. Create the same-named Secret in both, with different values:

   ```bash
   kubectl --context cluster-a create secret generic app-secrets --from-literal=API_KEY='from-cluster-a'
   kubectl --context cluster-b create secret generic app-secrets --from-literal=API_KEY='from-cluster-b'
   ```

2. Create a Pod in cluster B that consumes it via `secretKeyRef` on `app-secrets`/`API_KEY`.
3. Select both clusters so the resource map aggregates them, and open the map while the
   URL is on cluster A.
4. Click the cluster B pod. `KubeObjectDetails` renders the details with
   `cluster={resource.cluster}` while the URL still points at cluster A.
5. Expand the container and reveal the env var. On `main` it shows `from-cluster-a`; with
   this change it shows `from-cluster-b`.
6. Hover the "From" link. On `main` it points at cluster A's Secret; with this change it
   points at `/c/cluster-b/secrets/default/app-secrets`.
7. Run the tests:

   ```bash
   cd frontend
   npx vitest run src/components/common/Resource/ContainerEnvironmentVariables.test.tsx
   ```

   Reverting the `Resource.tsx` change makes 4 of the 5 cases fail. The fifth asserts that
   a fetched Secret still decodes correctly and passes either way, as a control.

## Notes for the Reviewer

- Same class of bug as the CronJob details page fetching its object without the `cluster`
  prop it was given. This is the container env var path of the same problem, so it may be
  worth a broader sweep for other `useGet`/`useList` calls inside components that already
  receive a cluster.
- Worth confirming the intent of one detail: the fix always passes an options object now,
  so `useGet` receives `{ cluster: undefined }` where it previously received no third
  argument at all. That is equivalent for the current `useKubeObject` implementation, and
  a test pins it, but say the word if you would rather the object were omitted entirely
  when there is no cluster.
- The env var section only ever renders for actual Pods, because
  `ContainerEnvironmentVariables` guards on `pod.status.containerStatuses` and `Pod` is the
  class that exposes `spec`/`status` getters. The DaemonSet, StatefulSet and Job callers of
  `ContainersSection` therefore never reach this code. That is pre-existing behaviour and
  this PR does not change it, but it is why the test mock needs those getters.
- The new test file mocks `lib/k8s/KubeObject` and the workload classes the same way the
  neighbouring `EnvironmentVariables.test.ts` does, because importing `Resource.tsx`
  without those mocks hits a genuine import cycle through `lib/k8s/index.ts`.
- `tsc --noEmit`, eslint and prettier are clean. The full
  `src/components/common/Resource/` suite passes (19 files, 159 tests).

### DIFF
--- a/frontend/src/components/common/Resource/ContainerEnvironmentVariables.test.tsx
+++ b/frontend/src/components/common/Resource/ContainerEnvironmentVariables.test.tsx
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from '@mui/material/styles';
+import { fireEvent, render } from '@testing-library/react';
+import { Base64 } from 'js-base64';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { KubeContainer } from '../../../lib/k8s/cluster';
+import { KubePod } from '../../../lib/k8s/pod';
+import { createMuiTheme } from '../../../lib/themes';
+import { TestContext } from '../../../test';
+
+// The k8s classes form an import cycle through lib/k8s/index.ts, so the base class is
+// stubbed the same way EnvironmentVariables.test.ts does it.
+const { MockKubeObject } = vi.hoisted(() => {
+  class MockKubeObject {
+    jsonData: any;
+    cluster: string | undefined;
+    static kind = '';
+    constructor(data: any, cluster?: string) {
+      this.jsonData = data;
+      this.cluster = cluster;
+    }
+    get kind() {
+      return this.jsonData?.kind;
+    }
+    get metadata() {
+      return this.jsonData?.metadata;
+    }
+    // Pod exposes spec/status off jsonData, and ContainerInfo hands its resource
+    // straight to ContainerEnvironmentVariables as a KubePod.
+    get spec() {
+      return this.jsonData?.spec;
+    }
+    get status() {
+      return this.jsonData?.status;
+    }
+  }
+  return { MockKubeObject };
+});
+
+vi.mock('../../../lib/k8s/KubeObject', () => ({ KubeObject: MockKubeObject }));
+vi.mock('../../../lib/k8s/deployment', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/replicaSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/statefulSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/daemonSet', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/job', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/pod', () => ({ default: class extends MockKubeObject {} }));
+vi.mock('../../../lib/k8s/secret', () => ({
+  default: class Secret extends MockKubeObject {
+    static kind = 'Secret';
+    static detailsRoute = 'secret';
+    static useGet = vi.fn(() => [null, null]);
+    get data() {
+      return this.jsonData?.data;
+    }
+  },
+}));
+vi.mock('../../../lib/k8s/configMap', () => ({
+  default: class ConfigMap extends MockKubeObject {
+    static kind = 'ConfigMap';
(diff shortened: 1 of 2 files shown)


0PrashantYadav0:
@ashu8912 and @sniok Please review this PR when you have free time. Thank you

PR #7627, opened 2026-09-08
github: Use Docker runtime for minikube chart tests

skoeva:
minikube v1.39.0 changed its default container runtime to containerd. The chart tests use `minikube docker-env` with Docker Buildx, which is incompatible with minikube's experimental containerd Docker API proxy and fails while inspecting the BuildKit image.

### Changes

- Explicitly use the Docker runtime for minikube chart tests


illume (approved):
🎉 thanks!

PR #7629, opened 2026-09-08
helm: scope release status cache key by namespace

magic-peach:
## Summary

This PR fixes a cache key collision in the Helm action status cache by scoping it to the release's namespace.

## Related Issue

Fixes #

## Changes

- Threaded the release's namespace into `getReleaseStatus`/`setReleaseStatus` and the action-status cache key in `backend/pkg/helm/release.go` and `handler.go`
- Updated `handler_test.go` and `release_internal_test.go` to cover the namespace-scoped key

## Steps to Test

1. Install (or upgrade) a release with the same name in two different namespaces
2. Poll the action status endpoint for each namespace's release right after
3. Before this fix, whichever write landed last would win the shared cache key, so one namespace's status could report the other's; after the fix each namespace's status is tracked independently

Not applicable, backend-only change.

## Notes for the Reviewer

- Found by reading `backend/pkg/helm/handler.go` directly rather than from a filed issue, no issue number to link
- Existing tests in `pkg/helm` all still pass

### DIFF
--- a/backend/pkg/helm/handler.go
+++ b/backend/pkg/helm/handler.go
@@ -128,8 +128,8 @@ type stat struct {
 }
 
 // getReleaseStatus returns the status of the release.
-func (h *Handler) getReleaseStatus(actionName, releaseName string) (*stat, error) {
-	key := "helm_" + actionName + "_" + releaseName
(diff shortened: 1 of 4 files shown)


skitt:

/close

PR #7645, opened 2026-09-10
backend: lock the repositories file before listing repos

magic-peach:
## Summary

This PR fixes a torn read race in ListRepo by taking a shared file lock before reading repositories.yaml.

## Related Issue

No existing issue tracks this; found by reading the locking code around AddRepo/RemoveRepo/UpdateRepository.

## Changes

- Added lockRepositoryFileForRead, which takes a shared lock instead of the exclusive lock the write paths use
- ListRepo now takes that read lock before loading repositories.yaml and releases it afterward
- Extracted the lock path computation shared by both lock functions into repositoryLockPath

## Steps to Test

1. Run go test ./pkg/helm/... -run TestLockRepositoryFile from the backend module
2. TestLockRepositoryFileForReadWaitsOnWriteLock confirms a read lock attempt fails while a write lock is held
3. TestLockRepositoryFileForReadAllowsConcurrentReaders confirms two read locks can be held at once

Not applicable, backend only change.

## Notes for the Reviewer

AddRepo, RemoveRepo, and UpdateRepository already take an exclusive lock before touching repositories.yaml, but ListRepo read it with no lock at all. Helm's repo.File.WriteFile truncates the file before writing rather than replacing it atomically, so a list request racing a concurrent add, remove, or update could previously read a truncated or partial file.

### DIFF
--- a/backend/pkg/helm/repository.go
+++ b/backend/pkg/helm/repository.go
@@ -84,20 +84,31 @@ func createFileIfNotThere(fileName string) error {
 	return err
 }
 
-func lockRepositoryFile(lockCtx context.Context, repositoryConfig string) (bool, *flock.Flock, error) {
-	var lockPath string
-
+func repositoryLockPath(repositoryConfig string) string {
 	repoFileExt := filepath.Ext(repositoryConfig)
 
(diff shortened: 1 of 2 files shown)


skitt:

/close

Issue #4539, opened 2026-02-04
Help to setup OIDC with Authentik


vacquah:
I can't get oidc setup for headlamp in kubernetes to work, using authentik. I have a talos kubernetes cluster with an external authentik server ( external to the cluster )

I have this in my talos kubernetes config:
```
      cluster:
        apiServer:
          extraArgs:
            oidc-client-id: kubernetes
            oidc-groups-claim: groups
            oidc-issuer-url: https://authentik.xxx.xx/application/o/k8s/
            oidc-username-claim: preferred_username
```

I have this in headlamp values. Yaml:
```
  values:
    # namespaceOverride: "headlamp"
    config:
      inCluster: true
      # baseURL: headlamp.xxx.xxx
      pluginsDir: /build/plugins
      enableHelm: false
      watchPlugins: false

      oidc:
        baseURL: https://headlamp.xxx.xx  
        clientID: "headlamp"
        clientSecret: "xxxx"
        issuerURL: "https://authentik.xxx.xxx/application/o/headlamp/"
        scopes: "openid,profile,email,groups"
        callbackURL: "https://headlamp.xxx.xxx/oidc-callback"
        # -- OIDC client to be used during token validation

```
Headlamp opens up ok but clicking on the oidc login button results in a loop - it seems to log in and I am brought right back to the login page. I have verified. in authentic my callback url is set properly. Not sure how to proceed.

minimalmattt:
Hi, You need to use the same OIDC client ID on both headlamp and the apiserver:

https://headlamp.dev/docs/latest/installation/in-cluster/oidc/:

> Note If you already have another static client configured for Kubernetes for the [apiserver's OIDC](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server) (OpenID Connect) configuration, use a single static client ID i.e -oidc-client-id for both Dex and Headlamp. Additionally, the redirectURIs need to be specified for each client.

I'm not sure how to check in talos but for example in k3s you can see the following (redacted) logs in journald if you try to login with a mismatched client_id

> Feb 06 09:05:01 k3s-test01 k3s[838803]: E0206 09:05:01.781121  838803 authentication.go:75] "Unable to authenticate the request" err="[invalid bearer token, oidc: verify token: oidc: expected audience \"apiserver-token\" got [\"headlamp-token\"]]"

Hope this helps :)

chr0n1x:
[...]

Issue #4198, opened 2025-11-23
Headlamp v0.38.0 OIDC In-Cluster Mode - Impersonation Not Working- EKS


hitsaurabhraj:
# Headlamp OIDC In-Cluster Mode - Impersonation Not Working

## Summary
Headlamp successfully authenticates users via OIDC (Azure AD), but fails to make API calls to the Kubernetes API server because it doesn't use impersonation headers, resulting in 401 Unauthorized errors.

## Environment
- **Headlamp Version**: v0.38.0
- **Kubernetes**: EKS (AWS)
- **OIDC Provider**: Azure AD (Azure Entra ID)
- **Deployment Mode**: In-cluster
- **Browser**: Chrome (latest)

## Related Issues
- [#3380](https://github.com/kubernetes-sigs/headlamp/issues/3380) - Missing OIDC Configuration Options for Azure Entra ID Integration
- [#3401](https://github.com/kubernetes-sigs/headlamp/issues/3401) - Unable to view logs or exec into pods with OIDC

## Problem Description

After successful OIDC authentication with Azure AD:
1. ✅ OIDC token is validated and stored in cookie
2. ✅ User is redirected back to Headlamp
3. ❌ Headlamp shows "Lost connection to cluster" error
4. ❌ All API calls to `/api/v1/` endpoints return 401 Unauthorized

**Error in Browser Console:**
```
GET https://headlamp.xxxx.xxxxxxxx.io/headlamp/clusters/main/healthz 401 (Unauthorized)
```
<img width="1570" height="599" alt="Image" src="https://github.com/user-attachments/assets/10ce2a8a-4576-46ef-82db-6b4e0d1757f6" />

Error.
Failed to verify ID Token: oidc: expected audience "<Client_ID>" got ["00000003-0000-0000-c000-000000000000"]

**Response:**
```json
{
  "kind": "Status",
  "apiVersion": "v1",
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}
```

### Headlamp Deployment Args
```yaml
args:
  - "-in-cluster"
  - "-plugins-dir=/headlamp/plugins"
  - "-base-url=/headlamp"
  - "-oidc-client-id=$(OIDC_CLIENT_ID)"
  - "-oidc-client-secret=$(OIDC_CLIENT_SECRET)"
  - "-oidc-idp-issuer-url=https://login.microsoftonline.com/<tenantID>/v2.0"
[...]

Issue #1651, opened 2024-01-10
OIDC: No permissions to list pods (dex + eks)


batbattur:
## Description
OIDC: No permissions to list pods. Similar to https://github.com/headlamp-k8s/headlamp/issues/393 and https://github.com/headlamp-k8s/headlamp/issues/1436

## Impact
Cannot use headlamp after OIDC login

## Environment and steps to reproduce
Set-up:

1. on EKS, deployed headlamp

```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: headlamp
  namespace: headlamp
spec:
  interval: 5m
  chart:
    spec:
      chart: headlamp
      reconcileStrategy: ChartVersion
      sourceRef:
        name: headlamp
        kind: HelmRepository
        namespace: headlamp
  values:
    config:
      oidc:
        baseURL: https://headlamp.someurl.com
        clientID: headlamp
        clientSecret: somesecret
        issuerURL: "https://dex.someurl.com"
        scopes: "openid email profile groups"
    ingress:
      enabled: true
      annotations:
        kubernetes.io/ingress.class: nginx
      hosts:
        - host: headlamp.someurl.com
          paths:
            - path: /
              type: Prefix
```

2. Setup Dex with Github connector:

```
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: dex
  namespace: dex
spec:
  interval: 5m
  chart:
    spec:
      chart: dex
      version: 0.6.5
      sourceRef:
        kind: HelmRepository
        name: dex
        namespace: dex
      interval: 1m
  values:
    image:
      tag: v2.31.0
    config:
      issuer: https://dex.someurl.com
      storage:
        type: memory
      oauth2:
        responseTypes: ["code", "token", "id_token"]
      staticClients:
      - name: 'Headlamp'
        id: headlamp
        secret: somesecret
        redirectURIs:
        - 'https://headlamp.someurl.com/oidc-callback'
      connectors:
      - type: github
        id: github
        name: GitHub
        config:
          clientID: github-app-client-id
          clientSecret: github-client-secret
          redirectURI: https://dex.someurl.com/callback
          orgs:
          - name: someorg
            teams:
            - somegroup
    ingress:
      enabled: true
      className: nginx
      annotations:
        kubernetes.io/ingress.class: nginx
      hosts:
        - host: dex.someurl.com
          paths:
[...]

PR #7646, opened 2026-09-10
backend: add --oidc-use-impersonation to fix 401s when the API server doesn't trust the OIDC issuer

T-Chittibabu:
## Problem

When running Headlamp with `-in-cluster` and OIDC authentication, login through the identity provider succeeds (the callback completes, the auth cookie is set), but every subsequent Kubernetes API call through Headlamp fails with `401 Unauthorized` — even though the same token works fine with `kubectl` directly.

This happens whenever the **real** Kubernetes API server does not have the OIDC issuer registered as a trusted identity provider. This is a very common situation on managed Kubernetes offerings (e.g. EKS), where:

- The control plane isn't reachable to register a native OIDC identity provider (e.g. the issuer is only resolvable/reachable from inside the cluster's own network, not from the cloud provider's control plane), or
- The cluster operator simply hasn't (or can't) register the issuer with the managed control plane.

Headlamp's `-in-cluster` OIDC mode forwards the user's raw ID token as the literal `Authorization: Bearer <token>` credential straight to whatever the API server is. If that API server has no authenticator configured for the token's issuer, the request is rejected outright — this happens at the *authentication* stage, before Kubernetes RBAC is even consulted. The failure looks identical to an expired/invalid token, and is easy to misdiagnose as an OIDC login bug when the login itself is actually working correctly.

This is the root cause of #4198, and appears to be the same underlying issue reported in #4539, #1651, and consolidated in #5401 ("Headlamp doesn't add Impersonate-User headers automatically").

Today, the only workarounds are:
1. Register the OIDC issuer natively with the managed control plane (not always possible, see above), or
2. Deploy a separate impersonating reverse proxy (e.g. `kube-oidc-proxy`) in front of the real API server, which validates the OIDC token and re-issues the request with `Impersonate-User`/`Impersonate-Group` headers using its own trusted in-cluster credential.

Both require infrastructure outside of Headlamp itself.

## Solution

This PR adds an opt-in flag, `--oidc-use-impersonation`, that lets Headlamp do what a proxy like `kube-oidc-proxy` does, natively:

When `--oidc-use-impersonation` is set together with `--in-cluster`:
- Headlamp authenticates to the API server using its **own** in-cluster ServiceAccount credential (the one the API server already trusts natively — no external issuer registration needed).
- It resolves the caller's username and groups from the already-verified OIDC token's claims, using the **same** `--me-username-path` / `--me-groups-path` JMESPath expressions already used by the `/me` endpoint — so no new claim-mapping configuration surface is introduced.
- It sets `Impersonate-User` / `Impersonate-Group` headers on the proxied request instead of forwarding the raw token.

This is exactly the mechanism Kubernetes' own impersonation feature is designed for (the same one `kubectl --as` and tools like `kube-oidc-proxy` rely on): a trusted identity re-presents a request on behalf of another identity, and the API server's RBAC authorization runs against the *impersonated* identity, not the trusted one.

**Requirement**: the Headlamp ServiceAccount must be granted RBAC `impersonate` on `users`/`groups`/`serviceaccounts` (documented in the flag's help text). This is a deliberate, explicit RBAC grant — nothing is auto-escalated.

If `--unsafe-use-service-account-token` is also set, startup now fails fast with a clear error: that flag already authenticates every request as the ServiceAccount without impersonating anyone, so combining the two would leave `--oidc-use-impersonation` silently inert.

### What this does *not* change
- Default behavior (raw token forwarding) is unchanged when the flag is off.
- `--in-cluster` without OIDC, or OIDC without `--in-cluster` (`--oidc-use-cookie`), are unaffected.
- No new claim-mapping flags — reuses `--me-username-path` / `--me-groups-path`.

## Testing

- New unit tests for the claims-to-impersonation-identity extraction (`pkg/auth`), the in-cluster context's ServiceAccount token-file wiring (`pkg/kubeconfig`), and an end-to-end test of the proxied request handler verifying that `Impersonate-User`/`Impersonate-Group` headers are set and the raw OIDC token is *not* forwarded to the upstream API server (`cmd`).
- New config validation tests for the flag's `--in-cluster` requirement and its mutual exclusivity with `--unsafe-use-service-account-token`.
- Manually validated end-to-end against a real managed Kubernetes cluster (EKS) whose control plane did not trust the OIDC issuer: before this change, every API call after a successful OIDC login returned 401; with `--oidc-use-impersonation` enabled (and the ServiceAccount granted `impersonate` RBAC), the same login flow works correctly and `Impersonate-User`/`Impersonate-Group` are set from the token's claims.
- Full existing test suite (`go test ./...`) passes unchanged.

Addresses #4198. Likely also relevant to #4539, #1651, and #5401, which describe the same underlying gap.


jagath25:
A couple of smaller notes, separate from the header-injection/token-verification points above:

- `validateOidcImpersonationFlag` checks `--in-cluster` and the `--unsafe-use-service-account-token` conflict, but doesn't check that OIDC is actually configured. If someone turns the flag on without any OIDC auth type configured, every single proxied API call will 401 forever (no cookie can ever exist), which will look like a broken installation rather than a config mistake. Not a security issue, just a rough edge — a validation error here would save someone a confusing debugging session.

- `makeUnsignedTestToken` in `backend/cmd/headlamp_test.go` duplicates what looks like it already exists as `makeTestToken` in `backend/pkg/auth/auth_test.go`. Different package, so it's not directly reusable as-is, but worth factoring into a shared test helper if this pattern shows up again.

T-Chittibabu:
Thanks for the review — both addressed:

1. `validateOidcImpersonationFlag` now also requires OIDC to actually be configured (`--oidc-client-id`, `--oidc-idp-issuer-url`, `--oidc-scopes`). Without it, no auth cookie can ever exist, so every proxied API call would have failed forever with a confusing "no valid OIDC token to resolve an identity" error — now it's a clear config-time validation error instead.

2. Extracted the duplicated JWT-builder (`makeTestToken` in `pkg/auth/auth_test.go` and `makeUnsignedTestToken` in `cmd/headlamp_test.go`) into a shared `internal/testutil.MakeUnsignedJWT` helper used by both packages.

PR #7659, opened 2026-09-14
frontend: fix KubeObjectGlance hook dependencies

kaulastudies:
## Summary

Fixes the `react-hooks/exhaustive-deps` suppression in `KubeObjectGlance` by tracking the `resource` used by the events effect.

The regression test now verifies that rerendering with the same resource reference does not refetch events, while changing to a new resource triggers a refetch for that resource.

## Related issue

Part of #5183

## Changes

- adds `resource` to the `useEffect` dependency array
- removes the `react-hooks/exhaustive-deps` suppression
- updates the existing KubeObjectGlance regression test for resource changes

## Verification

- `npx vitest run src/components/resourceMap/KubeObjectGlance/KubeObjectGlance.test.tsx`
- `npm run ci-lint`
- `npm run tsc`

A prior full frontend run passed 3271 of 3273 tests before the regression assertion was updated. The remaining environment-specific failure was `spawnSync make ENOENT` on Windows because GNU make is not installed locally. Upstream Linux CI provides the complete suite gate.


skoeva (changes requested):
please guard against stale responses so an earlier resource request cannot overwrite events for the current resource

kaulastudies:
Addressed the stale-response concern in 56f2de575.

The effect now ignores results from superseded resource requests, and I added a regression test that starts requests for two resources, resolves the newer request first, then resolves the older request afterward and verifies the stale response cannot overwrite the current resource's events.

Verified locally:
- `npx vitest run src/components/resourceMap/KubeObjectGlance/KubeObjectGlance.test.tsx` — 11/11 passed
- `npm run tsc`
- `npm run ci-lint`
- `git diff --check`

@skoeva  when you have a chance, could you please take another look?

PR #7662, opened 2026-09-14
Fix aria-label=hidden on decorative icons

### LINKED ISSUE #7433: aria-label="hidden" on decorative icons announces the word "hidden" to screen readers instead of hiding them

magic-peach:
## Summary

This PR fixes decorative icons that had aria-label="hidden", which sets the accessible name to the literal text "hidden" so screen readers announced the word instead of skipping the icon.

## Related Issue

Fixes #7433

## Changes

- Changed the alert icon in job/List.tsx to aria-hidden="true"
- Changed the alert icon in pod/List.tsx to aria-hidden="true"
- Changed one decorative wrapping Box in Resource.tsx to aria-hidden="true" and its inner icon to drop the redundant attribute
- Dropped aria-label="hidden" entirely on the other Box in Resource.tsx since it wraps a HoverInfoLabel with real hover text, so hiding it would suppress actual information

## Steps to Test

1. Search the frontend source for aria-label="hidden" and confirm there are no remaining matches
2. Open a Job or Pod list with a warning or error status and inspect the alert icon in devtools, confirming it now has aria-hidden="true" instead of aria-label="hidden"
3. Run npm run lint and npm run tsc in frontend/, both pass

Not applicable, this only changes an accessibility attribute with no visual change.

## Notes for the Reviewer

This touches four decorative icon spots across three files. Three are purely decorative and now use aria-hidden="true". The fourth, in Resource.tsx, wraps a HoverInfoLabel with real content, so the attribute was dropped instead of hidden.


skoeva (changes requested):
could you address the comments raised by copilot?

magic-peach:
@skoeva could you please check?

skitt:

/close

PR #7663, opened 2026-09-14
frontend: use PhaseLabel for namespace status in the list view

magic-peach:
## Summary

This PR replaces the hand rolled namespace status chip in the namespace list view with the shared PhaseLabel component.

## Related Issue

No related issue.

## Changes

- Removed the local makeStatusLabel helper in namespace/List.tsx, which mapped Active to success and everything else to error using StatusLabel directly
- Rendered the Status column with the shared PhaseLabel component instead, passing the raw phase string
- namespace/Details.tsx already renders the same field with PhaseLabel using its default successPhase of Active, so this makes both views consistent and removes a duplicated inline mapping

## Steps to Test

1. Run the frontend locally and open the Namespaces list
2. Confirm namespaces with phase Active show a success colored label reading Active
3. Confirm a namespace with any other phase (for example Terminating) shows an error colored label with that phase text
4. Compare against the namespace details page, which already uses PhaseLabel for the same field

Not included. This is a like for like swap of one label component for another with the same default success condition, so there is no visual change.

## Notes for the Reviewer

PhaseLabel defaults successPhase to Active, which matches the exact behavior of the removed makeStatusLabel helper, so this is intended to be behavior neutral.


skitt:

/close

PR #7668, opened 2026-09-15
frontend: node: Fix metrics with multiple clusters selected

edvorg:
With more than one cluster selected, node metrics were only fetched from the first cluster in the URL. The Nodes list only had CPU/memory bars for that cluster's nodes, and the Overview usage charts left the other clusters out.

This fetches node metrics with a `NodeMetrics` list query, the same way pod metrics are already fetched, so all selected clusters get queried. Metrics are matched to nodes by cluster too, and the "install metrics-server" hint only shows for nodes of the cluster that returned the error.

Node metrics now refresh every 5s like pod metrics, instead of every 10s.

Tested locally with 4 clusters: all nodes get CPU/memory bars and the Overview totals include every cluster.


edvorg:
This screenshot demonstrates the problem. I selected multiple clusters but metrics loads only for the first cluster in the list

image

PR #7671, opened 2026-09-15
frontend: Terminal: Keep idle exec sockets alive

### LINKED ISSUE #7650: Timeout after exec into pods
## Describe the bug

I am only been able to exec into pods for some arbitrary amount of time before the cli restarts and the session times out.

## To Reproduce
Steps to reproduce the bug:
1. Exec into any pod
2. Wait 1-2 minutes and get kicked out
3.  Message is failed to run cmd.exe

 - Installation type: Big Bang installation using Ironbank image. Linux OS
 - Headlamp Version: registry1.dso.mil/ironbank/opensource/headlamp-k8s/headlamp:v0.40.0 

No.

krsatyamthakur-droid:
## Summary

This PR fixes exec and attach terminals getting disconnected after about a minute of idling when Headlamp sits behind a proxy or load balancer with an idle timeout. The terminal now sends a small keepalive frame every 30 seconds so the connection never looks idle.

## Related Issue

Fixes #7650

## Changes

- Added a 30s keepalive in `Terminal.tsx` that re-sends the current terminal size on the exec/attach socket. Browsers can't send WebSocket ping frames, and an unchanged size is a no-op for the container, so it's just enough traffic to stop the connection being dropped. The interval matches the multiplexer's `HeartbeatInterval` in the backend.
- Moved the resize frame into a small `sendResize()` helper, since it's now sent from three places (xterm resize, first connect, keepalive).
- Added tests for the resize frame sent on connect, the keepalive firing every 30s, and the interval being cleared on unmount.

## Steps to Test

1. Run `cd frontend && npx vitest run src/components/common/Terminal.test.tsx`.
2. Run Headlamp behind a proxy with a short idle timeout (e.g. Istio or an ingress with a 60s idle timeout).
3. Open a terminal on any Linux pod and leave it alone for a few minutes.
4. The shell should stay connected. Before this change it dropped after ~60s and cycled through `bash`, `/bin/bash`, `sh`, `/bin/sh` one per minute until it showed "Failed to run cmd.exe".
5. Optionally, check the WS frames in the browser's network tab: a channel 4 resize frame should go out every 30s.

Not applicable, no UI changes.

## Notes for the Reviewer

- The reporter's Istio logs showed every shell living ~59.5s with `DC downstream_remote_disconnect`, while the rest of the UI stayed live over the same gateway thanks to the multiplexer heartbeat. `kubectl exec` didn't reproduce it because it goes straight to the API server and skips that gateway.
- `vitest.config.ts` only fakes `setTimeout`/`clearTimeout`, so the keepalive tests opt into fake intervals themselves.


krsatyamthakur-droid:
heads up, the windows arm64 failure isn't from this PR. it's a flaky e2e test that's been failing on other PRs too. i've opened #7675 to fix it

PR #7675, opened 2026-09-15
app: e2e-tests: Retry temp profile cleanup in externalBackendToken

krsatyamthakur-droid:
## Summary

This PR fixes a flaky Windows arm64 e2e failure in `externalBackendToken.spec.ts`. The "does not send the internal backend token to an unrelated port owner" test passes its assertions, then fails in cleanup with `EBUSY` while deleting its temporary Electron profile.

## Related Issue

No issue filed. It's currently failing `Build App / build-windows (arm64)` on unrelated PRs, e.g. #7671, and the runs for `add-missing-mui-locales-cs-sv`, `fix/7414-projectdetails-tabs-race` and `add-endpoints-create-form-ui`.

## Changes

- Let the `fs.rmSync` cleanup of the temporary user data directory retry (`maxRetries: 10`, `retryDelay: 200`).

On Windows the test spawns Electron through a shell, so the pid it kills and waits on is `cmd.exe`. That exits right after `taskkill`, but Electron's child processes can keep files in the profile (the `DIPS` database) open a bit longer, so the delete hits a locked file. Node only retries on errors like `EBUSY` and `EPERM`, so a real cleanup problem still fails the test.

## Steps to Test

1. Check that `Build App / build-windows (arm64, windows-11-arm, 22.x)` passes on this PR.
2. Optionally, run `externalBackendToken.spec.ts` on Linux or macOS and confirm it still passes. Cleanup behaves the same there.

Not applicable, test-only change.

## Notes for the Reviewer

- I couldn't reproduce this locally since I don't have a Windows arm64 machine, so this PR's CI run is the real check.
- The first test in the file cleans up after `electronApp.close()`, which waits for Electron to actually exit, so I left it alone.


skoeva (approved):
lgtm, thanks

PR #7696, opened 2026-09-17
frontend: HeadlampButton: Inherit navbar foreground color

skoeva:
The `HeadlampButton` explicitly uses `theme.palette.text.primary`, overriding the foreground color used by its parent AppBar.

This change makes the button inherit the AppBar foreground color, consistent with the other mobile navbar controls. Built-in themes are unaffected, while custom navbar themes and logos using `currentColor` receive the correct foreground color.

Originally observed in AKS desktop: https://github.com/Azure/aks-desktop/issues/638.

### Testing

- [x] Manually verified the mobile logo inherits the navbar foreground color
- [x] HeadlampButton Storybook tests pass (5 tests)


vyncent-t (approved):
LGTM

PR #7709, opened 2026-09-18
frontend: Translate UI strings that were hardcoded

### LINKED ISSUE #7692: frontend: Several UI strings are hardcoded and never translated
## Describe the bug

I was poking around the cluster settings pages and noticed "Enable Node Shell"
wasn't wrapped in `t()`, even though the heading right above it is. Searched for
other cases and there are a few across the app. The most visible is the heading
on the home page.

Six already have translations sitting in the locale files — the code just never
asks for them, so they go unused. In French, `Accueil`, `Enregistrer`, `Nom`,
`Paramètres` and `Chargement` are all already written.

Translations exist but never load:

- `components/App/Home/index.tsx:198` — `title="Home"`
- `components/common/Resource/Resource.tsx:571` — `<Button>Save</Button>`
- `components/namespace/CreateNamespaceButton.tsx:133` — `aria-label="Name"`
- `components/advancedSearch/SearchSettings.tsx:55` — `Settings`
- `components/resourceMap/GraphRenderer.tsx:162` — `title="Loading"`

Never extracted for translators at all:

- `components/App/Settings/NodeShellSettings.tsx:76` — `Enable Node Shell`
- `components/App/Settings/PodDebugSettings.tsx:71` — `Enable Pod Debug`
- `components/namespace/CreateNamespaceButton.tsx:126` — `aria-label="Dialog"`

## Why these look like mistakes

`Home/index.tsx:198` is the main heading on the home page, and everything else
on that screen goes through `t()` — both tabs, "No clusters found", "Add a
cluster to get started." and "Add Cluster". Only the heading doesn't:

Image

Red is line 198, the main heading on the home page. Green is line 206, a tab
eight lines below it. Same screen, same component, one wrapped and one not.

`CreateNamespaceButton.tsx:133` hardcodes `aria-label="Name"` while the line
directly below uses `t('translation|Name')` for the same word.

`Resource.tsx:571` is the only Save button in the app that isn't wrapped.
ShortcutsSettings, PluginSettingsDetails, secret/Details, configmap/Details and
SavedSearches all use `t('Save')`. It sits inside `DataField`, which doesn't
have `t` in scope, so that one needs `useTranslation` added rather than just a
wrap.

## To Reproduce

Switch Headlamp to French and open the home page. The tabs and empty state below
are translated, the heading still says "Home".

## Environment

- Installation type: dev build from main
- Headlamp Version: main

Yes (I will propose a PR).

9p-tech:
## Summary

Eight user-facing strings were rendered directly instead of going through
`t()`, so they stayed English in every locale. The most visible one is the
main heading on the home page.

Five of them already had translations sitting in the locale files that
were never used. Those load now: `Accueil`, `Enregistrer`, `Nom`,
`Paramètres` and `Chargement` in French, and the same in the other 24
locales.

The three that had never been collected (`Enable Node Shell`,
`Enable Pod Debug`, and the `Dialog` aria-label) are now in the locale
files as empty entries for translators to pick up.

`DataField` in `Resource.tsx` had no `t` in scope, so `useTranslation` is
added there too.

Fixes #7692

## Before and after

The home page heading, which is the one everyone sees. Line 198 was the
only string on that screen not going through `t()`; line 206 below it
already did.

Before:

Screenshot 2026-09-17 at 3 55 04 AM

After:

Screenshot 2026-09-18 at 6 23 56 AM

# Same fix, all eight places

Translations already existed, now they load:

- `App/Home/index.tsx:198` — `title="Home"` → `title={t('Home')}`
- `common/Resource/Resource.tsx:572` — the Save button in `DataField`
- `namespace/CreateNamespaceButton.tsx:133` — `aria-label="Name"`
- `advancedSearch/SearchSettings.tsx:55` — the Settings button
- `resourceMap/GraphRenderer.tsx:162` — `<Loader title="Loading" />`

Never extracted before, now in the locale files:

- `App/Settings/NodeShellSettings.tsx:77` — `Enable Node Shell`
- `App/Settings/PodDebugSettings.tsx:71` — `Enable Pod Debug`
- `namespace/CreateNamespaceButton.tsx:126` — `aria-label="Dialog"`

`DataField` in `Resource.tsx` had no `t` in scope, so `useTranslation` is
added at line 501 for the Save button.

## What's in the diff

7 source files plus the 25 locale files touched by `npm run i18n`. The
locale changes are just the extractor adding the three new keys and
reordering two existing ones to match source order. No translations were
removed.

## Steps to test

1. Switch Headlamp to French
2. Open the home page. The heading reads "Accueil" instead of "Home"
3. Open a resource with an editable field. The Save button reads
   "Enregistrer"

## Checks

- `npx tsc --noEmit` passes
- eslint and prettier pass on all changed files
- i18n extraction produces no further changes
- 77 tests across 12 files pass in the affected folders


skoeva (approved):
lgtm, thanks!

PR #7710, opened 2026-09-18
frontend: Sidebar/SidebarItem: Fix color-contrast in SectionHeader st…

### LINKED ISSUE #7173: **Sidebar/SidebarItem**: `frontend/src/components/Sidebar/Sidebaritem.stories.tsx` - Fix color contrast in Section Header story

vividwhisp:
# PR Description

## Summary

Fix `color-contrast` violation in `Sidebar/SidebarItem/Section Header` and `Styled Section Header` stories by adding explicit white text color.

---

## Related Issue

Fixes #7173

---

## Changes

- **`frontend/src/components/Sidebar/Sidebaritem.stories.tsx`**
  - Added explicit `color: '#ffffff'` to `SectionHeader` story args
  - Added explicit `color: '#ffffff'` to `StyledSectionHeader` story args (replaces removed `#90caf9`)
  - Both stories now use white text on the dark sidebar background, meeting WCAG AA contrast requirements

- **`frontend/.axe-storybook-baseline.test-a11y.json`**
  - Removed `"Sidebar/SidebarItem/Section Header": ["color-contrast"]` baseline entry
  - Removed `"Sidebar/SidebarItem/Styled Section Header": ["color-contrast"]` baseline entry

---

## Steps to Test

1. Run the accessibility test suite:
   ```bash
   export NODE_OPTIONS="--max-old-space-size=8192"
   npm run test:a11y
   ```

2. Verify output shows 0 new failures for:
   - `Sidebar/SidebarItem/Section Header`
   - `Sidebar/SidebarItem/Styled Section Header`

---

## Notes for the Reviewer

- **Fix approach (Option B)**: Added explicit `color: '#ffffff'` to both `SectionHeader` and `StyledSectionHeader` stories instead of fixing the component's `getContrastText` fallback. This is an isolated story-level fix; the component's `SidebarItem.tsx` still uses the `getContrastText` fallback for other consumers.
- No component logic changes-only story-level color override and baseline update.
- Test locally with `npm run test:a11y`.


vividwhisp:
all tests pass, need a review @illume

harshitnub077:
## Review Summary

Reviewed the diff and the `SidebarItem.tsx` component implementation that these stories exercise.

### Validation
- [x] Diff reviewed against component source and baseline file
- [x] Snapshot files verified
- [ ] Local a11y test — not run locally

### Findings

**🟠 [IMPORTANT] — Story-level fix masks the real component defect**

The component (`SidebarItem.tsx:99-102`) determines the subheader colour at runtime via:

```ts
theme.palette.sidebar.color ??
theme.palette.getContrastText(theme.palette.sidebar.background)
```

The `getContrastText` fallback is what fails WCAG AA in light mode or when the sidebar palette doesn't supply an explicit `sidebar.color`. This PR patches the *Storybook story args* with a hardcoded `color: '#ffffff'`, which removes the baseline entry and makes the a11y test pass — but leaves the live application rendering this path unchanged.

A user visiting a custom theme (or a future theme that doesn't set `sidebar.color`) will still hit the contrast violation. The baseline entry was accurately recording a real component bug. Removing it without fixing the component means we're now flying blind on that defect.

The correct fix is one of:
1. Fix `getContrastText` usage in `SidebarItem.tsx` to always return a sufficiently contrasted colour (the safer option).
2. If the story is intentionally testing the component in an unusual/custom colour context, document why white is the correct representation for the story rather than the component's real rendering.

**🟡 [MINOR] — Incomplete baseline removal**

The diff removes `"Sidebar/SidebarItem/Section Header": ["color-contrast"]` but the PR description also mentions `StyledSectionHeader`. Looking at the diff, `StyledSectionHeader` changes the colour from `#90caf9` (a blue that may or may not pass) to `#ffffff`. The corresponding baseline entry removal is not visible in the diff. Please confirm whether `StyledSectionHeader` also had a baseline entry that was removed.

**💡 [SUGGESTION] — Unrelated baseline removals**

The diff also removes `project/NewProjectPopup/Default` and `project/NewProjectPopup/With Existing Projects` from the baseline. These are not mentioned in the PR description. If those were fixed separately or the stories were removed, it's worth a brief note here so reviewers don't need to investigate separately.

### Positive Observations

- The PR correctly regenerates the snapshot files after the story change.
- The PR description clearly describes the chosen approach (Option B) and acknowledges it's a story-level fix rather than a component fix — transparency is appreciated.
- Baseline hygiene is a real improvement; stale baseline entries can hide regressions.

### Recommendation

**Request changes** — the baseline removal is premature while the underlying contrast defect in the live component is unfixed. The story correctly demonstrates a WCAG-compliant colour, but production users on certain themes can still encounter the bug. I'd suggest either fixing `SidebarItem.tsx` or keeping the baseline entry with a comment explaining the deferred fix.

harshitnub077:
The a11y baseline entries are stale and removing them is the right direction, but I'd push back on this approach a bit.

The contrast violation exists in `SidebarItem.tsx` itself — specifically the `getContrastText` fallback at line 102 that kicks in when `theme.palette.sidebar.color` is not set. This PR fixes it in the *stories* by hardcoding `color: '#ffffff'` in the args, which makes the a11y test pass, but leaves the component's actual rendering path unchanged. Any theme that doesn't supply an explicit `sidebar.color` will still produce insufficient contrast in the real UI.

The baseline entry was accurately recording a real bug. Removing it without fixing the component means we lose the signal. I'd either fix the fallback in `SidebarItem.tsx` (which is the real fix), or keep the baseline entry with a comment linking to a follow-up issue so it's tracked rather than silently dropped.

The unrelated `project/NewProjectPopup` entries removed from the baseline aren't mentioned in the PR description — worth a quick note on what happened to those stories so reviewers don't have to dig.

vividwhisp:
@harshitnub077 i just wanted the scope of the PR to stay to stories and shouldn't touch the frontend files directly, that's why i thought that this approach would fix this issue, which it did, But i think i need to ask one of the maintainers and get their opinion on this. @illume

skoeva (commented):
hi, could you provide screenshots for this fix? it would be super helpful!

PR #7711, opened 2026-09-18
frontend: GlobalSearch: Include Secrets in search results

edvorg:
Global search goes through a fixed list of resource kinds and Secret wasn't in it, so there was no way to find a secret from the search bar. This adds it next to ConfigMap, along with the missing mock in the GlobalSearchContent stories.

Opening the search now also lists Secrets, with the same request the Secrets page makes.


harshitnub077:
## Review Summary

Reviewed the implementation and diff against the existing `GlobalSearchContent.tsx` patterns.

### Validation
- [x] Diff reviewed against existing resource list
- [x] Storybook mock handler pattern verified
- [ ] Local lint/tests — not run locally

### Findings

**🟠 [IMPORTANT] — RBAC / 403 handling**

`Secret` is added to the `classes` array on line 102, which means `useList` will be called unconditionally for *all* users regardless of RBAC policy.

Most other resources in the list (Pods, ConfigMaps, etc.) are relatively safe to list from an RBAC perspective. Secrets are different — cluster operators commonly grant read access to Secrets only to specific service accounts, and a user who can see the Pods list in Headlamp may receive a 403 on `/api/v1/secrets`.

The current `GlobalSearchContent` renders a loading spinner per resource and silently drops 4xx errors in `useList`. It's worth confirming that a 403 on the secrets request does not surface as a visible error to the user (e.g. an error banner or a console warning that could be confusing). If the hook surfaces the error to the UI in any form, a 403 for Secrets should be silently swallowed or at most logged at debug level — not shown as a user-facing error, since this is an expected condition.

**🟡 [MINOR] — Story mock placement**

The new `http.get` handler for `/api/v1/secrets` is inserted between `configmaps` and `namespaces`. The existing handlers in the story appear to follow the order of the `classes` array. The Secret import is placed after ConfigMap in the array (line 102), which is consistent, but it may help future readers if the story handlers are also sorted consistently (e.g., alphabetically or matching array order exactly).

### Positive Observations

- The fix is minimal and surgically correct — one import, one array entry, one mock handler. No collateral changes.
- The Storybook mock was updated in the same commit, which prevents the story from making a real network request that would fail silently in CI.
- Placement next to `ConfigMap` is semantically sensible since both are namespace-scoped config/secret stores.

### Recommendation

**Approve with minor suggestions** — the core change is correct and well-scoped. I'd encourage the author to verify that a 403 on the Secrets endpoint does not produce a user-visible error banner before this merges, given how commonly Secrets are RBAC-restricted in production clusters.

harshitnub077:
The change makes sense and the mock handler update is the right thing to do alongside it.

One thing worth checking before this merges: Secrets are one of the most commonly RBAC-restricted resources in production clusters. A user who can list Pods through Headlamp may well get a 403 on `/api/v1/secrets`. Looking at `GlobalSearchContent.tsx`, `useList` results are mapped over without filtering errors — worth verifying that a 403 from the secrets request doesn't surface as a visible error banner or console warning to the user. If the hook silently swallows 4xx responses the way most Headlamp resource hooks do, this is fine. If it surfaces the error in the UI, Secrets should probably be treated like an optional resource and handled gracefully.

The story mock and the array position both look right.

edvorg:
@harshitnub077 Good point, I checked this. Global search only uses the items from each list and never looks at the errors, and 403s aren't retried, so if secrets are forbidden they just don't show up in the results. No banner, nothing extra in the console.

I verified it by rendering the `FoundSomeResults` story with secrets returning a 403:

```tsx
const { FoundSomeResults } = composeStories(stories);

await FoundSomeResults.load();
getWorker().use(
  http.get('http://localhost:4466/clusters/sample-cluster/api/v1/secrets', () =>
    HttpResponse.json({ kind: 'Status', reason: 'Forbidden', code: 403 }, { status: 403 })
  )
);
render(<FoundSomeResults />);
```

Then I waited for the requests to finish and compared the rendered DOM and the console.error/warn calls with a normal run. The DOM was identical and the 403 didn't add anything to the console.

Should I add this to the PR as a story (secrets returning 403), or is the check enough?

PR #7713, opened 2026-09-18
backend: auth: Fix OIDC refresh and /me behind a base URL

### LINKED ISSUE #7647: OIDC token refresh is skipped when Headlamp runs behind a path-based `baseURL`.
## Describe the bug

  OIDC token refresh is skipped when Headlamp runs behind a path-based `baseURL`.

  With `baseURL: /dev/internal/headlamp`, Kubernetes API requests use paths such as:

```text
  /dev/internal/headlamp/clusters/main/apis/metrics.k8s.io/v1beta1/pods
```
  After the access token expires, Headlamp returns 401 Unauthorized instead of refreshing the token and updating the authentication cookie.

  Expected behavior: Headlamp should recognize the /clusters/ request after removing the configured baseURL, refresh the token, and continue the session.

  ## To Reproduce

  Steps to reproduce the bug:

  1. Deploy Headlamp in-cluster with baseURL: /dev/internal/headlamp.
  2. Configure OIDC authentication with offline_access.
  3. Log in through OIDC.
  4. Wait for the access token to expire.
  5. Open or refresh a cluster page.
  6. Observe 401 Unauthorized responses from /dev/internal/headlamp/clusters/....

  - Headlamp Version: 0.45.0
  - Other: Azure Entra ID OIDC, Gateway API, path-based routing

  No.

  ## Additional Context

  The refresh middleware checks whether the raw request path starts with:

  strings.HasPrefix(r.URL.String(), "/clusters/")

  With baseURL configured, the path starts with /dev/internal/headlamp/clusters/, so refresh is skipped.

  Rewriting only /dev/internal/headlamp/clusters/... to /clusters/... at the Gateway works as a temporary workaround, but rewriting the entire prefix breaks the UI because Headlamp serves the UI at /
  dev/internal/headlamp/.

krsatyamthakur-droid:
## Summary

This PR fixes OIDC token refresh and the `/me` endpoint when Headlamp runs under a path-based `baseURL`. Right now both only recognise cluster requests that start with `/clusters/`, so behind something like `/dev/internal/headlamp` the token never gets refreshed and the user gets a 401 once it expires.

## Related Issue

Fixes #7647

## Changes

- Added a `StripBaseURL(path, baseURL)` helper in `pkg/auth`. It only strips on a `/` boundary, so a base URL of `/foo` doesn't also match `/foobar`.
- Added `ParseClusterAndTokenWithBaseURL`, which strips the base URL before matching `/clusters/<name>/`. `ParseClusterAndToken` calls it with an empty base URL, so existing callers behave the same.
- The OIDC refresh middleware now uses both, with the `BaseURL` it already had in its config. Fixing only the `HasPrefix` check wasn't enough, because the cluster name regex was anchored to `/clusters/` too, so the refresh would still be skipped.
- `HandleMe` gets the base URL through `MeHandlerOptions`, so `/me` reads the token cookie again and the top bar shows the user.
- Added tests for the helper, the base-aware parser, a full token refresh through the middleware under a base URL, and `/me` under a base URL.

## Steps to Test

1. Run `cd backend && go test ./pkg/auth/... ./cmd/...`.
2. Deploy Headlamp with `baseURL: /dev/internal/headlamp` and OIDC with `offline_access`.
3. Log in, wait for the access token to expire, then open or refresh a cluster page.
4. Requests to `/dev/internal/headlamp/clusters/...` should keep working, and the auth cookie should be updated with the new token.
5. The username should show in the top bar (`/dev/internal/headlamp/clusters/<name>/me` returns 200).

Not applicable, backend-only change.

## Notes for the Reviewer

- I checked that the new middleware and `/me` tests fail on the old code. `/me` returns the same 401 the reporter saw.
- The middleware test sends the old token in the `Authorization` header rather than a cookie. `SetTokenCookie` clears the existing cookie before setting the new one, so with a cookie the first matching `Set-Cookie` in the response is the cleared one. Reading the cookie under a base URL is covered by the `ParseClusterAndTokenWithBaseURL` test.
- I haven't tested this against a real Entra ID setup, so it'd be great if @jorisdebock could give it a try.


harshitnub077:
## Review Summary

Reviewed the backend auth fix for OIDC refresh and `/me` under a path-based `baseURL`. The implementation adds a `StripBaseURL` helper and threads it through two previously broken code paths.

### Validation
- [x] Diff reviewed end-to-end — helper, middleware, `HandleMe`, and all tests
- [x] `StripBaseURL` logic manually traced for edge cases
- [ ] `go test ./pkg/auth/... ./cmd/...` — not run locally

### Findings

**🟡 [MINOR] — `headlamp.go` `BaseURL` wiring not visible in the diff**

The `cmd/headlamp.go` diff shows `+1/-0` line (one addition). The PR description says `HandleMe` now reads `BaseURL` through `MeHandlerOptions`, which means the call site in `headlamp.go` must be updated to pass `config.BaseURL` into `MeHandlerOptions`. It would help to confirm in the PR that the wire-up line is present and uses the correct config field, since this is the only spot where a wrong field name or a zero value would silently leave `/me` broken in production even after this fix.

**🟡 [MINOR] — `shouldSkipOIDCRefresh` checks `r.URL.String()` but the fix uses `r.URL.Path`**

The old code was:
```go
if !strings.HasPrefix(r.URL.String(), "/clusters/")
```

The new code is:
```go
if !strings.HasPrefix(StripBaseURL(r.URL.Path, c.BaseURL), "/clusters/")
```

Using `r.URL.Path` is actually *more* correct than `r.URL.String()` (which would include the query string), so this is a silent improvement. Worth a one-line comment noting the intentional change from `.String()` to `.Path` so future readers don't revert it thinking it was accidental.

**💡 [SUGGESTION] — `TestParseClusterAndTokenWithBaseURL` only tests two paths**

The table has `"cluster request under base URL"` (positive) and `"request outside base URL"` (negative). A third case — `"base URL prefix collision, e.g. /dev/headlampx/clusters/..."` — would confirm the boundary-match semantics of `StripBaseURL` are preserved end-to-end through `ParseClusterAndTokenWithBaseURL`, not just through the unit-level `TestStripBaseURL`. This is a suggestion, not a blocker, since `TestStripBaseURL` already covers this case directly.

### Positive Observations

- The `StripBaseURL` helper is cleanly isolated, well-documented, and handles all the tricky boundary cases (trailing slash, exact-match, partial-segment prefix) correctly.
- Backward compatibility is properly preserved: `ParseClusterAndToken` now delegates to the new function with an empty `baseURL`, so existing callers are unaffected.
- The middleware and `HandleMe` test cases are end-to-end rather than just unit tests — they exercise the full refresh flow and the cookie read through the real handler, which gives high confidence the fix actually works.
- The bug analysis in the PR description is precise: the author correctly identifies that fixing only the `HasPrefix` check was not enough without also fixing the cluster-name regex path, and addresses both.

### Recommendation

**Approve with minor suggestions** — the implementation is sound and well-tested. The only ask is confirming the `headlamp.go` wiring is correct before merge.

harshitnub077:
The root cause analysis is correct and the fix covers both failure modes — the `HasPrefix` check in `shouldSkipOIDCRefresh` and the cluster-name regex in `ParseClusterAndToken`. Fixing only one of them wouldn't have been enough, and the PR description explains this clearly.

`StripBaseURL` is clean: boundary-safe, handles trailing slash, returns path unchanged when it doesn't match. The test table covers the edge cases that matter (partial prefix, exact match, no base URL).

One small note: `middleware.go` previously checked `r.URL.String()` (which includes query string) and the fix changes it to `r.URL.Path`. That's actually more correct, not just equivalent, but it's easy to misread as a typo. A brief comment on that line noting the intentional switch from `.String()` to `.Path` would help a future reader who might try to "fix" it back.

The `headlamp.go` wiring (`BaseURL: config.BaseURL`) is in place, which is the part that actually makes `/me` work in production under a base URL — good to see it's not missed.

krsatyamthakur-droid:
The two failed checks are infrastructure timeouts, not test failures: both timed out waiting for the Minikube Headlamp deployment to become available. The targeted suites pass locally (`cd backend && go test ./pkg/auth/... ./cmd/...`). I cannot rerun these jobs because GitHub requires repository admin permissions; please rerun the failed jobs from the Actions UI.

PR #7714, opened 2026-09-18
app: Bump electron to 43.7.0

skoeva:
Electron 38 is unsupported. This updates the desktop app to Electron 43 and adapts to its explicit binary installation flow.

### Changes

- Upgrade Electron from 38.8.6 to 43.7.0
- Explicitly install the Electron binary before builds and tests
- Require Node.js 22.12 or newer for the desktop app

### Testing

- [x] App unit tests pass
- [x] Electron e2e tests pass
- [x] TypeScript checks pass
- [x] Linux, macOS, and Windows x64 builds pass
- [x] macOS ARM64 packaging passes


sniok:
last time we upgraded electron, it broke rendering on linux https://github.com/kubernetes-sigs/headlamp/pull/5434

we would just need to double check that it still works after an upgrade

PR #7715, opened 2026-09-18
Do not evict an entry when bumping one that already exists

magic-peach:
## Summary
useRecent's bump() evicted the oldest recent-item entry whenever entries.length + 1 > maxItems, without checking whether the bumped id was actually new. So revisiting an item that was already in the recent list, while the list was already at maxItems, silently evicted a different entry anyway, even though the total count didn't grow.

## Related Issue
None filed, found by direct code reading.

## Changes
- Gate eviction in bump() on the entry actually being new (isNewEntry), not just on the post-bump length.
- Add useRecent.test.tsx (no test file existed before) covering: adding a new item under capacity, adding a new item at capacity (evicts oldest), re-bumping an existing item at capacity (no eviction), and bump order updates.

## Steps to Test
1. Run `npx vitest run src/components/globalSearch/useRecent.test.tsx` from `frontend/`.
2. All 4 tests pass; reverting the fix in useRecent.tsx makes the capacity re-bump test fail.

## Notes for the Reviewer
No existing test file covered this hook; new tests follow the pattern used by the sibling `useLocalStorageState.test.tsx`.


skitt:

/close

PR #7734, opened 2026-09-21
charts: pvc: Render all accessModes under a single key

### LINKED ISSUE #7732: charts: pvc: Only the last accessModes entry is applied
## Describe the bug

In the Helm chart `persistentVolumeClaim.accessModes` is a list, but the template writes the `accessModes:` key inside the `range` loop:

```yaml
{{- range .Values.persistentVolumeClaim.accessModes}}
accessModes:
  - {{ . }}
{{- end}}
```

so with two or more modes the rendered manifest has duplicate `accessModes` keys. Helm and kubectl parse that non-strictly, so only the last entry survives and the PVC is created with the wrong access modes nothing warns about it `helm lint` passes

Expected: all configured access modes end up on the PVC

Related: with the default `accessModes: []` and `enabled: true`, no `accessModes` key is rendered at all and the API server rejects the PVC (`spec.accessModes: Required value`) `size` has a `required` guard, `accessModes` doesn't

## To Reproduce

1. Save this as `values.yaml`:
   ```yaml
   persistentVolumeClaim:
     enabled: true
     size: 10Gi
     accessModes: [ReadWriteOnce, ReadOnlyMany]
   ```
2. Run `helm template hl charts/headlamp -f values.yaml --show-only templates/pvc.yaml`
3. Output has `accessModes:` twice:
   ```yaml
   spec:
     accessModes:
       - ReadWriteOnce
     accessModes:
       - ReadOnlyMany
   ```
4. Pipe the output through `kubectl create --dry-run=client -o yaml -f -` only `ReadOnlyMany` remains

## Environment

- Installation type: Helm
- Headlamp Version: chart 0.45.0 (current `main`, `charts/headlamp/templates/pvc.yaml`)
- Other: Helm v3.19.0

Yes I will propose a PR

## Additional Context

Not a regression the block has been like this since PVC support was added to the chart. Previous fixes touched `storageClassName` (#2579) and `size` (#2706) in the same template but not `accessModes`

CapThunder19:
## Summary

The PVC template put the `accessModes:` key inside the `range` loop, so a list with more than one mode produced duplicate keys and only the last one made it to the API server. This renders the list under a single key with `toYaml`, and fails at template time when the list is empty since a PVC needs at least one access mode

## Related Issue

Fixes #7732
## Changes

- `templates/pvc.yaml`: render `accessModes` with `toYaml`; `fail` if the list is empty
- `values.yaml`, `README.md`: note that `accessModes` is required when the PVC is enabled
- Added `tests/test_cases/pvc-enabled.yaml` with two access modes and its expected template (there was no PVC test case before)
- Added `tests/failing_test_cases/pvc-missing-access-modes.yaml`

## Steps to Test

1. `helm template hl charts/headlamp --set persistentVolumeClaim.enabled=true --set persistentVolumeClaim.size=10Gi --set 'persistentVolumeClaim.accessModes={ReadWriteOnce,ReadOnlyMany}' --show-only templates/pvc.yaml`
2. Both modes appear under one `accessModes:` key (on `main`, the key appears twice)
3. Drop the `accessModes` flag: rendering now fails with a clear message instead of the API server rejecting the PVC
4. `./charts/headlamp/tests/test.sh` passes

## Notes for the Reviewer

- I used `fail` instead of `required` because `required` accepts an empty list.
- No `Chart.yaml` version bump, since that looks like it's handled by the releaser.


harshitnub077:
## Review Summary

Reviewed the Helm chart template fix, the new test cases, and the expected template snapshot.

### Validation
- [x] Diff reviewed against existing chart templates and test infrastructure
- [x] Helm template logic verified manually
- [ ] Local `helm template` / `test.sh` — not run locally

### Findings

**🟠 [IMPORTANT] — `fail` guard will also reject an empty list assigned explicitly by the user**

The guard uses:

```yaml
{{- if not .Values.persistentVolumeClaim.accessModes }}
{{- fail "A valid .Values.persistentVolumeClaim.accessModes entry required!" }}
{{- end }}
```

In Helm, `if not []` evaluates to **true** (an empty list is falsy), so this correctly catches an unset or empty list. However, it also catches `accessModes: []` passed explicitly by a user, which is the same as the default. The error message is appropriate in that case, but consider whether the error text could be slightly more helpful — e.g., explicitly naming that at least one access mode is required and listing the valid values (`ReadWriteOnce`, `ReadOnlyMany`, `ReadWriteMany`).

**🟡 [MINOR] — Expected template snapshot pins chart version `0.45.0`**

`tests/expected_templates/pvc-enabled.yaml` contains hardcoded chart/app version labels (`helm.sh/chart: headlamp-0.45.0`, `app.kubernetes.io/version: "0.45.0"`). This means the test will fail on every chart version bump, which adds maintenance overhead. The existing test infrastructure may already handle this (e.g., via `sed` in `test.sh`), but if it doesn't, version-sensitive fields should either be templated out or tested differently to avoid a brittle snapshot.

**🟡 [MINOR] — `storageClassName: standard` in test case assumes a specific storage class**

The `pvc-enabled.yaml` test case uses `storageClassName: standard`. The `standard` storage class is not present in all Kubernetes distributions (e.g., EKS uses `gp2`/`gp3`). This is a test fixture, not a real deployment, so it won't fail the test — but it may mislead users who copy this as a reference. A neutral placeholder (e.g., an empty string or a comment noting it's an example) would be clearer.

**💡 [SUGGESTION]** — The `values.yaml` default for `accessModes` is still `[]` with no inline hint that enabling the PVC without setting access modes will fail at template time. A brief comment there (e.g., `# At least one access mode is required when enabled: true`) would improve discoverability without requiring README navigation.

### Positive Observations

- The root cause analysis is precise: the duplicate YAML key from the `range` loop is a real and silent data-loss bug — the Kubernetes API server would silently accept only the last entry.
- The author correctly chose `fail` over `required` to catch an empty list, and documented the rationale. That's exactly the kind of nuance reviewers look for.
- First PVC test case in the repo — the test coverage improvement is meaningful.
- Both the happy-path and the failing-path test cases were added together, which is good practice.

### Recommendation

**Approve with minor suggestions** — the core template fix is correct and well-tested. The minor points above are improvements rather than blockers.

harshitnub077:
The template fix is correct — the `range` loop was producing duplicate `accessModes:` keys, and only the last one would survive the YAML parse, which is a real silent data-loss bug. The `toYaml` approach is the right way to render a list here, and using `fail` instead of `required` to catch an empty list is the right call (as noted in the PR description, `required` doesn't reject `[]`).

One thing to be aware of: `tests/expected_templates/pvc-enabled.yaml` has `helm.sh/chart: headlamp-0.45.0` hardcoded throughout, and looking at `test.sh`, it does a direct diff against the expected file without stripping version labels. That means this test will break on the next chart version bump and someone will have to update 163 lines. The other expected templates in the repo probably have the same issue, so this isn't a new problem introduced here — but if there's an `update-version.sh` or equivalent that regenerates these, it should be the source of truth rather than a hand-edited file.

The failing test case for missing access modes is a nice addition — first time PVC behaviour has test coverage at all.

Issue #855, opened 2022-12-01
backend uses nolint and other linter skips


illume:
## Description

The backend code uses nolint and other lint skips which shouldn't be skipped. There's no justification next to where the linter is disabled either.

Lint skips shouldn't pass CI/review unless they come with a good justification (IMHO).

## Impact

The code is riskier, and harder to understand/test/change.

joaquimrocha:
@vyncent-t Maybe you can add this to your backlog (not urgent, just in case you end up having to choose items).

RISHIKESHk07:
Hey @vyncent-t are you working on this ? , would love to give it a try ,could you please tell me what i have to do and where i need to change stuff .

vyncent-t:
Hello @RISHIKESHk07

This is currently on my backlog so I have not much info to deliver!
Currently the backend has the linter disabled because its too long, if you nav to [backend/pkg/config/config.go](https://github.com/headlamp-k8s/headlamp/blob/main/backend/pkg/config/config.go#L61) you can see it has been commented out.

The linter is saying func length is too long, this code could either need to be refactored or an explanation as to why it is not in use. You may want to consider making the length shorter by using smaller separate functions.

You may also want to ref [linter info](https://golangci-lint.run/usage/linters/) for more information on [funlen](https://github.com/ultraware/funlen#description)

RISHIKESHk07:
@vyncent-t thanks for the help , will work on it in the coming days

jhalak101205-cpu:
hey @skoeva can i be assigned these issue , I have approach for solving these :)

jhalak101205-cpu:
@illume , Regarding the `//nolint:gosec` suppressions around file reading (os.ReadFile in kubeconfig and server.go): these paths originate from CLI flags or kubeconfig paths rather than untrusted user inputs. Would the project consider adding a shared sanitized file-reading helper in `pkg/` in the future to avoid repeating gosec suppressions across packages ???

PR #7735, opened 2026-09-22
backend: Audit production complexity and security lint suppressions

### LINKED ISSUE #7702: backend: Refactor or justify production complexity/length suppressions

jhalak101205-cpu:
## Summary
Audited and updated production `//nolint` suppressions in `backend/cmd/` and `backend/pkg/` to use specific linter names and concise inline justifications, addressing #7702 (part of meta-issue #855).

Fixes #7702

## Changes
- Replaced un-justified or generic `//nolint` comments across 14 production Go files with specific linter names (`gosec`, `funlen`, `gocognit`, `gocyclo`, `goconst`, `prealloc`).
- Added concise inline justification comments explaining the rationale for each suppression.
- Kept all line lengths strictly within the 120-character limit (`lll` rule) and formatted according to `gofmt`.
- Did not modify any business logic, API contracts, or error handling.

Files updated:
- `backend/cmd/headlamp.go`
- `backend/cmd/multiplexer.go`
- `backend/cmd/server.go`
- `backend/pkg/auth/auth.go`
- `backend/pkg/auth/cookies.go`
- `backend/pkg/auth/middleware.go`
- `backend/pkg/helm/repository.go`
- `backend/pkg/k8cache/cacheInvalidation.go`
- `backend/pkg/kubeconfig/kubeconfig.go`
- `backend/pkg/plugins/plugins.go`
- `backend/pkg/portforward/handler.go`
- `backend/pkg/portforward/store.go`
- `backend/pkg/serviceproxy/http.go`
- `backend/pkg/telemetry/telemetry.go`


jhalak101205-cpu:
/kind cleanup

CC @skoeva @gambtho @kahirokunn

While auditing and justifying the complexity and function length suppressions (`funlen`, `gocognit`, `gocyclo`) on handlers like `createHeadlampHandler` and `handleNodeDrain`, I added clear justifications to keep these changes atomic, surgical, and low-risk. 

Longer-term, would the maintainers prefer these larger handlers to be broken down into sub-helpers/middleware, or is keeping them consolidated preferred for readability? Happy to help with follow-up refactoring if desired!

harshitnub077:
Good cleanup — keeping suppressions specific to a single linter name is strictly better than the broad `//nolint` pattern we had before.

Two things worth a quick fix before this lands:

`cacheInvalidation.go` — the G107 justification just says "Cache invalidation request", which describes the call site but not why the taint is safe. Since `freshURL` is a copy of the already-validated incoming request URL with nothing user-injected, the comment should say that. Something like `// G107: freshURL copied from validated request, no external input` is what a security reviewer needs to close out this suppression without having to read the surrounding code.

`headlamp.go` — the `InsecureSkipVerify` comment says "Dev TLS config", which implies it's only reachable in development. It isn't — this path is live in production when a user opts out of TLS verification. The same block in `auth.go` already uses the cleaner wording "InsecureSkipVerify used when cluster skips TLS"; aligning headlamp.go to match avoids someone removing the suppression later under the assumption it was leftover debug code.

On the refactoring question: extracting helpers from the long `funlen` functions in a follow-up PR is exactly the right call. I'd open an issue first, describe which functions you want to split and how, and let the maintainers sign off before writing the code — saves a round-trip if the preferred decomposition differs.

Everything else looks fine.

PR #7743, opened 2026-09-22
frontend: PluginSettingsDetails: Use theme styling for save buttons

### LINKED ISSUE #7742: frontend: PluginSettingsDetails: Disabled Save button looks enabled due to hardcoded inline styles
### Describe the bug

In Settings → Plugins → (plugin details), the Save and Cancel buttons use hardcoded inline styles (`backgroundColor: 'silver', color: 'black'` on Save, `color: 'silver'` on Cancel). This causes two problems:

1. **A disabled Save button looks enabled.** The inline `style` overrides MUI's disabled treatment, so before any setting is changed (`disabled={!enableSaveButton}`), Save still renders as a solid silver button, indistinguishable from its enabled state. Verified in the DOM: `{"text":"Save","disabled":true,"style":"background-color: silver; color: black;"}`.
2. **The enabled Cancel button looks disabled.** Silver text on the light background reads as a disabled control, the inverse of the actual state.

The hardcoded silver also ignores the active theme in both light and dark modes, and the row's `borderColor: 'silver'` bypasses the theme divider color. Sibling Settings components (IconPicker, ColorPicker) use the standard themed `variant="contained"` buttons.

### To Reproduce

1. `npm run storybook` in `frontend/`
2. Open `Settings/PluginSettingsDetail` → `Without Auto Save`
3. Observe Save (disabled, but looks active) and Cancel (enabled, but looks inactive)
4. Type in the plugin's input: Save becomes enabled with no visual change

Same behavior in the app under Settings → Plugins for any plugin with `displaySettingsComponentWithSaveButton`.

### Screenshot

Save here is **disabled**, Cancel is **enabled**:

current

### Environment
- Headlamp version: main (`33015fbf5`)
- Storybook / browser: Chromium

Yes. PR incoming: drop the inline styles so the buttons get the standard MUI contained/text treatment (matching the other Settings pages) and use `divider` for the separator color.

Satyam087:
Fixes #7742

The Save and Cancel buttons in the plugin settings details page hardcode silver/black inline styles. The inline `style` prop overrides MUI's disabled treatment, so a disabled Save button looks identical to an enabled one, while the enabled Cancel button's silver text reads as disabled. The affordances are inverted. The hardcoded colors also ignore the active theme.

This drops the inline styles so both buttons get the standard themed treatment (same as the other Settings pages, e.g. IconPicker/ColorPicker), and switches the separator's `borderColor` from `'silver'` to the theme `divider` color. The `PluginSettingsDetails.WithoutAutoSave` storyshot is regenerated accordingly.

**Steps to test**

1. `npm run storybook` in `frontend/`, open `Settings/PluginSettingsDetail` → `Without Auto Save`
2. Save renders in MUI's disabled style (muted), Cancel as a normal text button
3. Type in the plugin's input: Save switches to the enabled contained-primary style
4. `npm run frontend:lint` and `npm test -- --run storybook.test -t "PluginSettingsDetail"` pass

| Before (Save is disabled here) | After: Save disabled | After: Save enabled |
| --- | --- | --- |
| before | after disabled | after enabled |


skoeva (approved):
lgtm, thanks for this!

